Issue #174, opened 2018-05-16
Add simple tabs tray component (+ concept)


pocmo:
We do not know the actual structure/bheavior of the tabs tray in Fenix. However a very simple implementation will help us to test the session switching behavior.

PR #361, opened 2018-07-03
Isse #174: GeckoEngineView: Release sessions before view gets detached or a new session gets assigned.

pocmo:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1470883

... that's the best we can do for now.

### DIFF
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineView.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineView.kt
@@ -20,7 +20,15 @@ class GeckoEngineView @JvmOverloads constructor(
     defStyleAttr: Int = 0
 ) : FrameLayout(context, attrs, defStyleAttr), EngineView {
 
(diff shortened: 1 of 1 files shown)


components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineView.kt
```diff
@@ -20,7 +20,15 @@ class GeckoEngineView @JvmOverloads constructor(
     defStyleAttr: Int = 0
 ) : FrameLayout(context, attrs, defStyleAttr), EngineView {
 
-    internal var currentGeckoView = GeckoView(context)
+    internal var currentGeckoView = object : GeckoView(context) {
+        override fun onDetachedFromWindow() {
+            // We are releasing the session before GeckoView gets detached from the window. Otherwise
+            // GeckoView will close the session automatically and we do not want hat.
```
csadilek: typo. want *t*hat.

pocmo: 🎩

PR #362, opened 2018-07-04
Closes #349, #340: Add FxaResult async type and FxaException

### LINKED ISSUE #349: Improve Error handling in FxA module
The module includes lots of API calls that return null on error. The Error class needs a different name too i.e. it has an `isSuccess` method which doesn't seem right.

carolkng:
Adds 
- FxaResult, a Promise-like type derived from [GeckoResult][1]. Since this is a "homegrown" promise type, the library will continue to support apps written in Java 6+ and Kotlin.
- FxaException, a general Exception class for errors from the FxA library which turns the Rust error codes into something Java-like

r? @csadilek, @eoger
ping @thomcc after pto to make sure there aren't any memory leaks

[1]: https://reviewboard.mozilla.org/r/250560/diff/3/

### DIFF
--- a/samples/fxa/src/main/java/org/mozilla/samples/fxa/MainActivity.kt
+++ b/samples/fxa/src/main/java/org/mozilla/samples/fxa/MainActivity.kt
@@ -7,18 +7,22 @@ package org.mozilla.samples.fxa
 import android.net.Uri
 import android.os.Bundle
 import android.support.v7.app.AppCompatActivity
-import mozilla.components.service.fxa.FirefoxAccount
 import android.support.customtabs.CustomTabsIntent
 import android.view.View
-import mozilla.components.service.fxa.Config
-import mozilla.components.service.fxa.FxaClient
 import android.content.Intent
 import android.widget.TextView
+import mozilla.components.service.fxa.Config
+import mozilla.components.service.fxa.FirefoxAccount
+import mozilla.components.service.fxa.FxaClient
+import mozilla.components.service.fxa.FxaResult
+import mozilla.components.service.fxa.OAuthInfo
+import mozilla.components.service.fxa.Profile
+
 
 open class MainActivity : AppCompatActivity() {
 
     private var account: FirefoxAccount? = null
-    private var config: Config? = null
+    private var scopes: Array<String> = arrayOf("profile")
 
     companion object {
         const val CLIENT_ID = "12cc4070a481bc73"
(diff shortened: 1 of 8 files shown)


pocmo:
@carolkng There seem to be a couple of build errors on taskcluster:
https://tools.taskcluster.net/groups/RZZRfM93QQyqzvIujR1wOw/tasks/QMAvErojReOiYmGPnsW49w/runs/0/logs/public%2Flogs%2Flive.log

ktlint and detekt also reported some issues:
* https://tools.taskcluster.net/groups/RZZRfM93QQyqzvIujR1wOw/tasks/Aa71zcCJR3SzQImxcggldw/runs/0/logs/public%2Flogs%2Flive.log
* https://tools.taskcluster.net/groups/RZZRfM93QQyqzvIujR1wOw/tasks/OoEh0keUSPijZevwcLglAg/runs/0/logs/public%2Flogs%2Flive.log

samples/fxa/src/main/java/org/mozilla/samples/fxa/MainActivity.kt
```diff
@@ -7,18 +7,16 @@ package org.mozilla.samples.fxa
...
 import android.os.Bundle
 import android.support.v7.app.AppCompatActivity
-import mozilla.components.service.fxa.FirefoxAccount
 import android.support.customtabs.CustomTabsIntent
 import android.view.View
-import mozilla.components.service.fxa.Config
-import mozilla.components.service.fxa.FxaClient
 import android.content.Intent
 import android.widget.TextView
+import mozilla.components.service.fxa.*
```
sputnikci: \[ktlint\] WARNING: \[no\-wildcard\-imports\] Wildcard import in column 1

samples/fxa/src/main/java/org/mozilla/samples/fxa/MainActivity.kt
```diff
@@ -78,4 +77,13 @@ open class MainActivity : AppCompatActivity() {
         customTabsIntent.intent.data = Uri.parse(url)
         customTabsIntent.launchUrl(this@MainActivity, Uri.parse(url))
     }
+
+    private fun openOAuthTab() {
+        val valueListener =  { value: String? ->
```
sputnikci: \[ktlint\] WARNING: \[no\-multi\-spaces\] Unnecessary space\(s\) in column 29

csadilek:
Test coverage is down, because we actually added the first tests here and it started tracking ;).

thomcc:
Hey, I didn't get to review this before it landed (I was on PTO until today), but I think it has some fairly large problems. Nothing prevents one thread from `close()`ing the FxAccount between when another thread calls `validPointer()` and when it gets passed to Rust.

The solution to that probably means just using a finalizer() for RustObject (since something could always preempt with a `close()`).

Additionally, it's not safe at all to call multiple methods at the same time from multiple threads on the rust fxa code (in rust terms, it's `Send` but not `Sync`), but I don't think anything prevents this? I could be wrong.

Both of these problems can lead to memory safety issues and potential security vulns.

csadilek:
Thanks @thomcc! Sorry for merging this before you had a chance to look. We're still in POC mode and are OK with incremental improvements :). I've filed a follow-up issue for the problems you mentioned: https://github.com/mozilla-mobile/android-components/issues/388

The purpose of the async calls was to detach the FxA calls from the UI thread so we don't risk blocking the UI. I don't think we need to support multiple threads calling FxA methods at the same time, or do we? Of course, we do definitely need to add guards to prevent this from happening (see the issue I've filed.)

thomcc:
Yeah, if you can ensure the API is only called from one thread at a time that should be fine. I would also override close() to be a noop, or something (e.g. finalize() should probably be the only way resources are released) unless you synchronize access in that function as well.

PR #364, opened 2018-07-04
concept-tabstray + browser-tabstray

pocmo:
The next pieces are ready for review.

I need to cleanup `feature-tabs` and the code that adds it to the sample app. I'll open a follow-up PR once that's ready too.


components/concept/tabstray/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+    }
+}
+
+dependencies {
+    implementation project(':browser-session')
+
+    implementation "org.jetbrains.kotlin:kotlin-stdlib:${rootProject.ext.dependencies['kotlin']}"
+}
+
+archivesBaseName = "abstract-toolbar"
```
csadilek: Name needs to be updated (concept-tabstray)

components/concept/tabstray/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+
+apply from: '../../../publish.gradle'
+ext.configurePublish(
+        'org.mozilla.components',
+        'concept-tabstray',
```
csadilek: We have some inconsistency here now:

concept-toolbar -> abstract-toolbar
concept-engine -> engine
concept-tabstray -> concept-tabstray

Should we just stick to the latter and file a follow-up?

components/concept/tabstray/src/main/java/mozilla/components/concept/tabstray/TabsTray.kt
```diff
@@ -0,0 +1,76 @@
...
+
+        /**
+         * A tab has been closed.
+         */
+        fun onTabClosed(session: Session)
+    }
+
+    /**
+     * Displays the given list of sessions. This method will be invoked with the initial list
+     * of sessions or if a completely new list of sessions should be displayed.
```
csadilek: Something missing after 'or' :)

components/browser/tabstray/src/test/java/mozilla/components/browser/tabstray/BrowserTabsTrayTest.kt
```diff
@@ -0,0 +1,62 @@
...
+import org.junit.runner.RunWith
+import org.mockito.Mockito.verify
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.RuntimeEnvironment
+import org.robolectric.Shadows
+
+@RunWith(RobolectricTestRunner::class)
+class BrowserTabsTrayTest {
+    @Test
+    fun `holders will unsubscrive if view gets detached`() {
```
csadilek: typo: unsubscribe.

components/browser/tabstray/src/test/java/mozilla/components/browser/tabstray/BrowserTabsTrayTest.kt
```diff
@@ -0,0 +1,62 @@
...
+        val adapter: TabsAdapter = mock()
+        val tabsTray = BrowserTabsTray(RuntimeEnvironment.application, tabsAdapter = adapter)
+
+        val shadow = Shadows.shadowOf(tabsTray)
+        shadow.callOnDetachedFromWindow()
+
+        adapter.unsubscribeHolders()
```
csadilek: this should be verify(adapter).unsubscribeHolders().

components/concept/tabstray/build.gradle
pocmo: Yeah. Let's file something and come up with a good naming scheme once we migrate from jcenter to our own repository. That's a good time to change the artifact names.

Regarding abstract-toolbar ... that's probably a relic. In the beginning I used the word "abstract" instead of "concept". But that didn't work well with Java package names that cannot contain the word "abstract". And yeah, engine exists before we had concept modules. :)

pocmo:
@csadilek PR updated!

components/concept/tabstray/build.gradle
pocmo: #368

PR #367, opened 2018-07-04
Do not build "release" variants of sample browser.

pocmo:
We only need the "debug" variants. Hopefully this will make things a little bit faster on taskcluster. Building all those variants takes quite a while.


pocmo:
Build time of the browser variants is down from ~27 minutes to ~15 minutes. 🎉

csadilek:
⏩

PR #373, opened 2018-07-05
Issue #174: Add "feature-tabs" component that connects a trabs tray implementation with the session and a toolbar implementation.

pocmo:
That's the feature connecting (tabs tray) <-> (toolbar, session). After that I only need to clean up the sample browser code and open a PR for that. :)

This code here depend on PR #371.


components/feature/tabs/src/main/java/mozilla/components/feature/tabs/tabstray/TabsTrayPresenter.kt
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * Presenter implementation for a tabs tray implementation in order to update the tabs tray whenever
+ * the state of the session manager changes.
+ */
+class TabsTrayPresenter(
+    private val tabsTray: TabsTray,
+    private val sessionManager: SessionManager
+) : SessionManager.Observer {
+    @Volatile private var sessions = sessionManager.sessions
```
csadilek: Why do these need to be volatile? A presenter instance isn't shared across threads, right?

PR #380, opened 2018-07-06
Issue #174: Add tabs tray component to sample browser..

### LINKED ISSUE #174: Add simple tabs tray component (+ concept)

pocmo:
Closes #174


pocmo:
Like with the BrowserFragment I want to see where things are going with TabsTrayFragment and whether we can move this (partially) to a component later.

Issue #396, opened 2018-07-09
Do not rely on finalize to free memory


csadilek:
Finalize is finally deprecated in Java 9. It's not a reliably way to free up resources of any kind.

PR #397, opened 2018-07-09
Closes #383 and #392: FxA fixes / refactorings.

### LINKED ISSUE #383: Refactor FxA method names to use Java/Kotlin convention (camelCase not snake_case)

csadilek:
First of many refactorings :).

This is to hide the snake_case Rust method wrappers from our public API. There's no value in having the app call `init`. If a project adds the library to the classpath, but doesn't use it, the FxaClient class won't be loaded and therefore not initialized. So, there's no need for extra control.

### DIFF
--- a/components/service/fxa/src/main/java/mozilla/components/service/fxa/RustObject.kt
+++ b/components/service/fxa/src/main/java/mozilla/components/service/fxa/RustObject.kt
@@ -29,16 +29,12 @@ abstract class RustObject<T> : Closeable {
         return p
     }
 
-    abstract fun destroyPointer(p: T)
+    protected abstract fun destroy(p: T)
 
     override fun close() {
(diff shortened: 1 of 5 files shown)


components/service/fxa/src/main/java/mozilla/components/service/fxa/RustObject.kt
```diff
@@ -29,7 +29,7 @@ abstract class RustObject<T> : Closeable {
         return p
     }
 
-    protected abstract fun destroyPointer(p: T)
+    abstract fun destroyPointer(p: T)
```
csadilek: This shouldn't be protected. It's currently called from a finalizer, which needs to be fixed as well: #396

csadilek:
I am adding another commit to address #396.

PR #398, opened 2018-07-10
Handle "no selected session" case. 

### LINKED ISSUE #379: Handling "no sessions" in sample browser.
We need to handle the case where the user closes all tabs.

What happens in this case is probably up to the app. For example in Fennec we open a new tab that points to the start page. The sample browser currently just crashes because it wants to display the selected session which doesn't exist anymore.

pocmo:
Fixes #379 (See issue for details).

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/storage/DefaultSessionStorage.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/storage/DefaultSessionStorage.kt
@@ -109,7 +109,7 @@ class DefaultSessionStorage(
         return try {
             val json = JSONObject()
             json.put(VERSION_KEY, VERSION)
-            json.put(SELECTED_SESSION_KEY, sessionManager.selectedSession.id)
+            json.put(SELECTED_SESSION_KEY, sessionManager.selectedSession?.id ?: "")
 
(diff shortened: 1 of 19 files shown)


components/browser/session/src/main/java/mozilla/components/browser/session/storage/DefaultSessionStorage.kt
```diff
@@ -109,7 +109,7 @@ class DefaultSessionStorage(
         return try {
             val json = JSONObject()
             json.put(VERSION_KEY, VERSION)
-            json.put(SELECTED_SESSION_KEY, sessionManager.selectedSession.id)
+            json.put(SELECTED_SESSION_KEY, sessionManager.selectedSession?.id)
```
csadilek: I think we should put an empty string in here instead of "null". When this value gets restored we're looking up a session with its ID using a non-nullable String.

pocmo: 👍

csadilek:
@pocmo it was a minor test change so I went ahead and updated the commit. 😎 

Just need Taskcluster to behave now :).

PR #399, opened 2018-07-10
Closes #388: Run Rust FxA calls in single thread context

### LINKED ISSUE #388: Ensure single-threaded access to Rust FxA calls
We have multiple options here e.g. synchronizing methods or maybe better a single-threaded coroutine-context.

carolkng:
Creates an FxaClient-specific single thread context, then moves all launch calls and runBlocking calls (for non-async methods calling into the Rust library) into that context. 

r? @csadilek @thomcc


csadilek:
@pocmo are you ok with this as well? I didn't find an easy way to release the thread on timeout. I filed a follow-up to dig deeper.

carolkng:
From Slack discussion with Thom: instead of starting the new thread context and worrying about when to close it, the blocks inside the coroutines can be synchronized on the FxaClient instance.

csadilek:
@thomcc I think this does it. Want to take another look? We can refactor this later to introduce `safeAsynRustCall` and `safeRustCall` methods in RustObject. So, we don't have to do the launch / synchronized dance every time, but I think this is ready to go in now.

csadilek:
@carolkng fixed the code duplication by adding the `safeAsync` and `safeSync` calls now. Merging, as we have everything we want here ;).

PR #409, opened 2018-07-11
Closes #354: Rename fxa component to provide clearer name

### LINKED ISSUE #354: Rename fxa component to provide clearer name
It's not clear to external consumers that fxa stands for Firefox Accounts. Let's keep using fxa as package and library name, but rename our component to make it easier to find.

csadilek:
This seems like a good compromise. Change the module names, but keep fxa on the source-level.


csadilek:
Header updated.

PR #413, opened 2018-07-12
Closes #412: Introduce method to complete FxA promise chain

### LINKED ISSUE #412: Introduce method to complete FxA result promise chain
Currently every value listener (lambda expression) has to return a result even when its known to be the last in line: https://github.com/mozilla-mobile/android-components/blob/master/samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/MainActivity.kt#L67

We should introduce a `whenComplete` method to simplify this API.

csadilek:
This simplifies the API as we no longer have to return a value from our listeners, if not needed (if they are the last in the chain). It cleans up the sample app and makes it easier to read.

/cc @carolkng

### DIFF
--- a/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FxaResult.kt
+++ b/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FxaResult.kt
@@ -45,7 +45,9 @@ class FxaResult<T> {
     }
 
     /**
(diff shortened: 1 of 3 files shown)


csadilek:
This includes a test fix to make sure the whole chain is actually executed. Before, if any of the callbacks aren't called, the tests would still pass.

PR #421, opened 2018-07-13
Closes #382: Increase test coverage of FxA component

### LINKED ISSUE #382: Increase test coverage for service-fxa module
Coverage is currently much lower than desired. It should be pretty easy to write unit tests for most of the involved classes.

csadilek:
Uphill battle, but this finally works, and we can build on it to add more tests in the future.

/cc @carolkng


csadilek:
eek uphill continues. need to get the tests to pass on taskcluster.

PR #423, opened 2018-07-13
Don't set composing region after removing autocomplete

### LINKED ISSUE #348: InlineAutocompleteEditText: Backspace deletes wrong chunk when deleting autocompleted URL
From: https://bugzilla.mozilla.org/show_bug.cgi?id=1471868

> 
> Using backspace to delete autocompleted URL leaves the wrong chunk of characters left in url bar.
> 
> To reproduce: Start typing "spo" and wait for the "spotify.com" as autocomplete. Then click backspace to delete the autocompleted chunk will leave "spo" again. When now typing "i" will show "spi" in the url instead of "spoi".

aperfilyev:
Seems like composing span is applied anyway.
Closes #348

### DIFF
--- a/components/ui/autocomplete/src/main/java/mozilla/components/ui/autocomplete/InlineAutocompleteEditText.kt
+++ b/components/ui/autocomplete/src/main/java/mozilla/components/ui/autocomplete/InlineAutocompleteEditText.kt
@@ -490,7 +490,6 @@ open class InlineAutocompleteEditText @JvmOverloads constructor(
                     // Make the IME aware that we interrupted the setComposingText call,
                     // by having finishComposingText() send change notifications to the IME.
                     finishComposingText()
-                    setComposingRegion(composingStart, composingEnd)
                     return true
                 }
                 return false
--- a/components/ui/autocomplete/src/test/java/mozilla/components/ui/autocomplete/InlineAutocompleteEditTextTest.kt
+++ b/components/ui/autocomplete/src/test/java/mozilla/components/ui/autocomplete/InlineAutocompleteEditTextTest.kt
@@ -321,4 +321,24 @@ class InlineAutocompleteEditTextTest {
         assertEquals(AutocompleteResult.emptyResult(), et.autocompleteResult)
         assertEquals("text", et.text.toString())
     }
+
+    @Test
+    fun testRemoveAutocompleteOnComposing() {
+        val et = InlineAutocompleteEditText(context, attributes)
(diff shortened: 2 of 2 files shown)


aperfilyev:
@pocmo yup, will do!

pocmo:
Awesome, thank you!

aperfilyev:
@pocmo so this is a bit harder than I thought because `setComposingText` with wrong indices is being triggered by the `Binder` and I don't think that it's possible to simulate this without the framework.

pocmo:
@csadilek You wrote the other tests in InlineAutocompleteEditTextTest. Do you have an idea what we could do here?

csadilek:
@aperfilyev @pocmo Yes, here's a test that fails without the fix and passes with it. It ensures that we finish the composing text when removing the autocomplete text so we re-start composing at the current cursor position.

This bug actually causes two different problems:

**Incorrect composing start (as described in https://bugzilla.mozilla.org/show_bug.cgi?id=1471868):**
*spo* -> spotify.com -> *backspace* -> spo -> *i* -> spi (should be *spoi*)

**Incorrect autocompletion:**
*spo* -> spotify.com -> *backspace* -> spo -> *backspace* -> stackoverflow.com (should be *sp*)

The fix proposed by @aperfilyev will take care of both. It just took me a while to realize that the second problem also isn't expected behaviour :). Here's the unit test I'd porpose to use and add to `InlineAutocompleteEditTextTest`.

```kotlin
@Test
fun testRemoveAutocompleteOnComposing() {
  val et = InlineAutocompleteEditText(context, attributes)
  val ic = et.onCreateInputConnection(mock(EditorInfo::class.java))

  ic?.setComposingText("text", 1)
  assertEquals("text", et.text.toString())

  et.applyAutocompleteResult(AutocompleteResult("text completed", "source", 1))
  assertEquals("text completed", et.text.toString())

  // Simulating a backspace which should remove the autocomplete and leave original text
  ic?.setComposingText("tex", 1)
  assertEquals("text", et.text.toString())

  // Verify that we finished composing
  assertEquals(-1, BaseInputConnection.getComposingSpanStart(et.text))
  assertEquals(-1, BaseInputConnection.getComposingSpanEnd(et.text))
}
```

csadilek:
@aperfilyev do you want to update your PR with the test added? Then we can review this :).

aperfilyev:
@csadilek sure

pocmo:
Thank you @aperfilyev and @csadilek. :)

I squashed, rebased and landed this in 7e56af9adb5a77071d7dc55f77231127993dafb0 :tada:

PR #425, opened 2018-07-16
RFC: Adding "consumable" properties to Session.

pocmo:
(This PR is not for merging - only for discussion).

I would like to add properties to Session that are _consumable_. This means when setting the value of a _consumable property_ an `Observer` can decide to consume it. This means the value won't be set on `Session` and the property value will be `null`. If the value is not consumed then it will be set and can be read from `Session` to be consumed later.

Some properties like a `Download` are something that an observer processes and then there is no need to keep a reference to the value anymore. Sometimes the value can't be processed immediately (e.g. we need to request a permission first) and then it is helpful to store it in `Session` to be consumed later. There will be more properties in the future that following this _consumer_ pattern.

The following PR implements this behavior for a Download. The PR is not complete to be actually merged.

I would like to make it more clear that this property can/should be consumed. For the observer case it's clear. But when consuming this property later a _processor_ would need to read (is it null?), then process, then set it to `null` which is not super nice.

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/Download.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/Download.kt
@@ -1,0 +1,16 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.session
+
+import android.os.Environment
+
+data class Download(
+    val url: String,
+    val contentDisposition: String,
+    val mimeType: String,
+    val contentLength: Long,
+    val userAgent: String, // TODO: needed??
+    val destinationDirectory: String = Environment.DIRECTORY_DOWNLOADS
+)
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
@@ -6,6 +6,7 @@ package mozilla.components.browser.session
 
 import mozilla.components.browser.session.engine.EngineSessionHolder
 import mozilla.components.browser.session.tab.CustomTabConfig
+import mozilla.components.support.utils.observer.Consumable
 import mozilla.components.support.utils.observer.Observable
 import mozilla.components.support.utils.observer.ObserverRegistry
 import java.util.UUID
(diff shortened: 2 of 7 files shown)


pocmo:
@csadilek What do you think?

csadilek:
@pocmo cool feature :).

I think if we want to force the caller to consume the property, or at least make a decision whether or not to consume it, we will have to wrap it in some reference holder. So, instead of having a `Download` property, we'd have a `Consumable<Download>` where the caller either needs to call `consume` which will return and clear the reference, or `get` which will return but keep it.

It might also make sense to create our own custom delegate type (`consumableObservable`). This might make it clearer than using `vetoable`, but that's mostly cosmetics.

pocmo:
> Consumable<Download> 

I was thinking about that too. But then writing the property would be awkward, either I'd have to wrap it in Consumable myself or I'd need to call some other setter method.

pocmo:
> It might also make sense to create our own custom delegate type (consumableObservable)

I first started to implement a *consumable* property delegate but then I noticed that it's actually exactly the same as `vetoable`. As this is only an implementation detail that is not exposed (A caller can't access the delegate) `vetoable` may be enough?

pocmo:
> fun onDownload(session: Session, download: Download): Boolean = false

* Do you think that interface is good enough (+ javadoc) or should it be more obvious that we are asking the observer to consume if it wants?

* Also: What do you think about not notifying other observers after the first one has consumed. This may cause "order" problems. And maybe this breaks use cases: e.g. I have one observer that show a popup and initiates the download and another one that  just sends telemetry for downloads etc.. - Maybe we should just notify all of them and _veto_ if at least one of them consumed?

csadilek:
> I was thinking about that too. But then writing the property would be awkward, either I'd have to wrap it in Consumable myself or I'd need to call some other setter method.

Yes, but it seems that's worth it e.g. having a Consumable.forValue(download) is pretty easy and then the consumer knows what's going on.

csadilek:
> As this is only an implementation detail that is not exposed (A caller can't access the delegate) vetoable may be enough?

True, the complexity of adding our own delegate might not be worth it.

csadilek:
> `fun onDownload(session: Session, download: Download): Boolean = false` Do you think that interface is good enough (+ javadoc) or should it be more obvious that we are asking the observer to consume if it wants?

I think introducing the `Consumable<Download>` fixes all that. Let's have the type system help us. :) I  don't really see another way right now.

> Also: What do you think about not notifying other observers after the first one has consumed. This may cause "order" problems. And maybe this breaks use cases: e.g. I have one observer that show a popup and initiates the download and another one that just sends telemetry for downloads etc.. - Maybe we should just notify all of them and veto if at least one of them consumed?

Yes, I agree, that this seems better. Call them all and make a decision after.

pocmo:
@csadilek Updated the PR with a version that uses Consumable<Download>. Let me know what you think.

pocmo:
@csadilek Okay, in this case I'll continue and implement the missing pieces. Closing this PR for now. :)

PR #428, opened 2018-07-16
Closes #343: Persist FxA state in shared prefs

### LINKED ISSUE #343: Persist and restore fxa state to avoid having to log in every time
Currently a log in is required every time the app starts.

carolkng:
r? @csadilek

### DIFF
--- a/samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/MainActivity.kt
+++ b/samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/MainActivity.kt
@@ -4,6 +4,7 @@
 
 package org.mozilla.samples.fxa
 
+import android.content.Context
 import android.net.Uri
 import android.os.Bundle
 import android.support.v7.app.AppCompatActivity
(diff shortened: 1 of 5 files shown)


samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/MainActivity.kt
```diff
@@ -27,19 +28,40 @@ open class MainActivity : AppCompatActivity() {
...
         }
 
         findViewById<View>(R.id.button).setOnClickListener {
             account?.beginOAuthFlow(scopes, false)?.whenComplete { openTab(it) }
         }
+
+        findViewById<View>(R.id.buttonPurge).setOnClickListener {
+            getSharedPreferences(FXA_STATE_PREFS_KEY, Context.MODE_PRIVATE).edit().putString(FXA_STATE_KEY, "").apply()
+            val txtView: TextView = findViewById(R.id.txtView)
+            txtView.text = "State cleared!"
```
csadilek: We need to put "State cleared!" into strings.xml

samples/firefox-accounts/src/main/res/values/strings.xml
```diff
@@ -6,4 +6,5 @@
     <string name="app_name">FxA Android Demo</string>
     <string name="sign_in">FxA Sign In</string>
     <string name="signed_in">Signed in: %1$s</string>
+    <string name="purge">Purge state</string>
```
csadilek: Should we maybe just call this "sign out" or "logout"?

carolkng: updated to "FxA Log Out"

samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/MainActivity.kt
carolkng: Updated to "Logged out!" to be consistent with the button

PR #436, opened 2018-07-17
Adds search suggestions to search

boek:


components/browser/search/src/main/java/mozilla/components/browser/search/SearchEngine.kt
```diff
@@ -32,7 +34,19 @@ class SearchEngine internal constructor(
         // The parse should have put the best URL for this device at the beginning of the list.
         val searchUri = resultsUris[0]
 
-        val template = Uri.decode(searchUri.toString())
+        return buildURL(searchUri, searchTerm)
+    }
+
+    /**
+     * Attempts to builds a Uri to get suggestions from this search engine
```
csadilek: nit: typo: Attempts to build...or maybe better "Builds a URL ...."

components/browser/search/src/main/java/mozilla/components/browser/search/suggestions/Parser.kt
```diff
@@ -0,0 +1,97 @@
...
+        }
+    }
+}
+
+/**
+ * The Parser is a function that takes a JSON Response and maps
+ * it to a Suggestion list.
+ */
+typealias JSONResponse = String
+typealias Parser = (JSONResponse) -> List<String>
```
csadilek: nit: maybe name this ResponseParser?

components/browser/search/src/main/java/mozilla/components/browser/search/suggestions/SearchSuggestionClient.kt
```diff
@@ -0,0 +1,37 @@
...
+
+            return SearchSuggestionClient(searchEngine, fetcher)
+        }
+    }
+    /**
+     * Gets search suggestions for a given query
+     */
+    suspend fun getSuggestions(query: String): List<String>? {
+        val suggestionsURL = searchEngine.buildSuggestionsURL(query) ?: return null
+        val parser = selectParser(searchEngine)
```
csadilek: I think renaming this to `selectResponseParser` would make this easier to read.

boek: @csadilek Great comments. Will revise!

components/browser/search/src/main/java/mozilla/components/browser/search/suggestions/SearchSuggestionClient.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Gets search suggestions for a given query
+     */
+    suspend fun getSuggestions(query: String): List<String>? {
+        val suggestionsURL = searchEngine.buildSuggestionsURL(query) ?: return null
+        val parser = selectResponseParser(searchEngine)
+
+        return fetcher(suggestionsURL)?.let { parser(it) }
+    }
```
pocmo: Regarding `JSONException` (and I assume `IOException`): I think we should either wrap them in a custom exception and make it explicit that this code throws it [(throws annotation)](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-throws/index.html) _or_ just fail gracefully and return an empty list. I think for apps showing suggestions returning an empty list is probably the easiest to handle.

components/browser/search/src/main/java/mozilla/components/browser/search/suggestions/SearchSuggestionClient.kt
```diff
@@ -0,0 +1,38 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.search.suggestions
+
+import mozilla.components.browser.search.SearchEngine
+
+/**
+ * Async function responsible for taking a URL and returning the results
+ */
+typealias SearchSuggestionFetcher = suspend (url: String) -> String?
```
pocmo: Do you want to leave it to the app to implement this? Do you expect apps have very different implementations of it? Should we provide a simple default implementation based on HttpUrlConnection? I assume you will need to write something for Focus anyways?

components/browser/search/src/main/java/mozilla/components/browser/search/suggestions/Parser.kt
```diff
@@ -0,0 +1,97 @@
...
+
+            override fun next(): Any {
+                val i = it.next()
+                return this@asSequence.get(i)
+            }
+
+            override fun hasNext() = it.hasNext()
+        }
+    }
+}
```
pocmo: This is nice. I think we should move this into `support-ktx` and make it public. :)

components/browser/search/src/main/java/mozilla/components/browser/search/suggestions/Parser.kt
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Selects a Parser based on a SearchEngine
+ */
+fun selectResponseParser(searchEngine: SearchEngine): ResponseParser = when (searchEngine.name) {
+    "Azerdict" -> azerdictResponseParser
+    "다음지도" -> daumResponseParser
+    "Qwant" -> qwantResponseParser
+    else -> defaultResponseParser
+}
```
pocmo: Can we make this function and the properties above `internal`? Otherwise they are visible application-wide (and will get autocompleted everywhere).

components/browser/search/src/main/java/mozilla/components/browser/search/suggestions/SearchSuggestionClient.kt
```diff
@@ -0,0 +1,38 @@
...
+    private val searchEngine: SearchEngine,
+    private val fetcher: SearchSuggestionFetcher
+) {
+    companion object {
+        fun create(searchEngine: SearchEngine, fetcher: SearchSuggestionFetcher): SearchSuggestionClient? {
+            if (!searchEngine.canProvideSearchSuggestions) { return null }
+
+            return SearchSuggestionClient(searchEngine, fetcher)
+        }
+    }
```
pocmo: Do we really need this factory method?

You can just throw from init:

```Kotlin
class SearchSuggestionClient private constructor(
    private val searchEngine: SearchEngine,
    private val fetcher: SearchSuggestionFetcher
) {
    init {
        if (!searchEngine.canProvideSearchSuggestions) {
           throw IllegalStateException(...)
       }
    }
}
```

It also avoid that you have to deal with `null` values coming out of `create()`. Especially if we make `canProvideSearchSuggestions` public then we do not need to try to create a client for this search engine in the first place.

components/browser/search/src/main/java/mozilla/components/browser/search/suggestions/Parser.kt
boek: 👍

components/browser/search/src/main/java/mozilla/components/browser/search/suggestions/SearchSuggestionClient.kt
boek: ah! I totally forgot about inits with kotlin 🤦‍♂️

components/browser/search/src/main/java/mozilla/components/browser/search/suggestions/SearchSuggestionClient.kt
boek: that was the idea. I'm not opposed to including a default fetcher (I'm also not a huge fan of the name the more I say it.. so suggestions are definitely welcome!)

components/browser/search/src/main/java/mozilla/components/browser/search/suggestions/SearchSuggestionClient.kt
```diff
@@ -0,0 +1,56 @@
...
+ */
+class SearchSuggestionClient(
+    private val searchEngine: SearchEngine,
+    private val fetcher: SearchSuggestionFetcher
+) {
+
+    /**
+     * Exception types for errors caught while getting a list of suggestions
+     */
+    class FetchException : Exception("There was a problem fetching suggestions")
```
sputnikci: \[ktlint\] WARNING: \[colon\-spacing\] Missing spacing before ":" in column 25

components/browser/search/src/main/java/mozilla/components/browser/search/suggestions/SearchSuggestionClient.kt
```diff
@@ -0,0 +1,56 @@
...
+    class ResponseParserException : Exception("There was a problem parsing the suggestion response")
```
sputnikci: \[ktlint\] WARNING: \[colon\-spacing\] Missing spacing before ":" in column 34

PR #439, opened 2018-07-17
Closes #407: Add source info to session

### LINKED ISSUE #407: Expose "source" in session
In Focus and the Fire TV app we save the "source" of a session (e.g. intent, custom tab, user entered, ...) in the session object.

I need to investigate what the Fire TV app is using it for but at least in Focus we use it for reporting it to telemetry (observer -> new session -> report with source).

csadilek:
Pretty simple but seems good enough for now?

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
@@ -16,6 +16,7 @@ import kotlin.properties.Delegates
  */
 class Session(
     initialUrl: String,
+    val source: Source = Source.NONE,
     val id: String = UUID.randomUUID().toString(),
     delegate: Observable<Session.Observer> = ObserverRegistry()
 ) : Observable<Session.Observer> by delegate {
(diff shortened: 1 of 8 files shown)


components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
```diff
@@ -48,6 +49,21 @@ class Session(
...
+    enum class Source {
+        ACTION_SEND, // Created to handle an ACTION_SEND (share) intent
+        ACTION_VIEW, // Created to handle an ACTION_VIEW intent
+        CUSTOM_TAB, // Created to handle a CustomTabs intent
+        HOME_SCREEN, // User interacted with the home screen
+        MENU, // User interacted with a menu
+        NEW_TAB, // User opened a new tab
+        NONE, // Default value and for testing purposes
+        TEXT_SELECTION, // Default value and for testing purposes
+        USER_ENTERED //  User entered a URL or search term
```
pocmo: Why not a JavaDoc (or however it's called for Kotlin code 😀) so that it shows up in the IDE when using the enum?

components/browser/session/src/main/java/mozilla/components/browser/session/storage/DefaultSessionStorage.kt
```diff
@@ -140,20 +141,21 @@ class DefaultSessionStorage(
...
         return JSONObject().apply {
             put("url", session.url)
+            put("source", session.source.name)
         }
     }
 
     @Throws(JSONException::class)
     internal fun deserializeSession(id: String, json: JSONObject): Session {
-        return Session(json.getString("url"), id)
+        return Session(json.getString("url"), Source.valueOf(json.getString("source")), id)
```
pocmo: This can get problematic if we decide to rename a source. Do you think we should handle the case where there's no matching Source? And/or should we pass a String to the enum constants so that the string representation is  independent from the name of the enum constants?

csadilek: I think going with the constants and their string representation is easier than adding a secondary String. Let's handle the case where we don't find a matching name and fall back to Source.NONE. Even if we add a secondary string for persistence, there's no guarantee that we won't forget to keep its original name :). I will update this.

pocmo: 👍

csadilek: @pocmo updated and test added.

pocmo:
🚢

PR #442, opened 2018-07-17
Closes #403: Ignore first load of about:blank in engine-gecko

### LINKED ISSUE #403: engine-gecko-*: Ignore first load of about:blank
GeckoView always loads about:blank as the first thing it does. And we end up showing this briefly in the URL bar.

Apparently that's a Gecko thing and we have to live with that for now. Let's ignore that and do not forward this to the observers of the engine session.

Fennec does all kinds of things to work around that too, e.g.:
* https://dxr.mozilla.org/mozilla-central/source/mobile/android/components/SessionStore.js#753-756
* https://dxr.mozilla.org/mozilla-central/source/mobile/android/base/java/org/mozilla/gecko/webapps/WebAppActivity.java#118

csadilek:
In our sample browser, this saves us 3 events per tab on startup.

### DIFF
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -21,6 +21,8 @@ class GeckoEngineSession(
 
     internal var geckoSession = GeckoSession()
 
+    private var initialLoad = true
+
     init {
         geckoSession.open(runtime)
 
@@ -73,13 +75,13 @@ class GeckoEngineSession(
     override fun saveState(): Map<String, Any> = runBlocking {
         val stateMap = CompletableDeferred<Map<String, Any>>()
         launch {
-            geckoSession.saveState({ state ->
+            geckoSession.saveState { state ->
                 if (state != null) {
                     stateMap.complete(mapOf(GECKO_STATE_KEY to state.toString()))
                 } else {
                     stateMap.completeExceptionally(GeckoEngineException("Failed to save state"))
                 }
-            })
+            }
         }
         stateMap.await()
     }
@@ -99,6 +101,12 @@ class GeckoEngineSession(
      */
     private fun createNavigationDelegate() = object : GeckoSession.NavigationDelegate {
         override fun onLocationChange(session: GeckoSession?, url: String) {
+            // Ignore initial load of about:blank (see https://github.com/mozilla-mobile/android-components/issues/403)
+            if (initialLoad && url == ABOUT_BLANK) {
+                return
+            }
+            initialLoad = false
+
             notifyObservers { onLocationChange(url) }
         }
 
@@ -135,6 +143,11 @@ class GeckoEngineSession(
             session: GeckoSession?,
             securityInfo: GeckoSession.ProgressDelegate.SecurityInformation?
         ) {
+            // Ignore initial load of about:blank (see https://github.com/mozilla-mobile/android-components/issues/403)
+            if (initialLoad && securityInfo?.origin?.startsWith(MOZ_NULL_PRINCIPAL) == true) {
+                return
+            }
+
             notifyObservers {
                 if (securityInfo != null) {
                     onSecurityChange(securityInfo.isSecure, securityInfo.host, securityInfo.issuerOrganization)
@@ -165,5 +178,7 @@ class GeckoEngineSession(
         internal const val PROGRESS_START = 25
         internal const val PROGRESS_STOP = 100
         internal const val GECKO_STATE_KEY = "GECKO_STATE"
+        internal const val MOZ_NULL_PRINCIPAL = "moz-nullprincipal:"
+        internal const val ABOUT_BLANK = "about:blank"
     }
 }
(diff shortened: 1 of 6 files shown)


components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -135,6 +143,12 @@ class GeckoEngineSession(
             session: GeckoSession?,
             securityInfo: GeckoSession.ProgressDelegate.SecurityInformation?
         ) {
+            // Ignore initial load of about:blank (see https://github.com/mozilla-mobile/android-components/issues/403)
+            if (initialLoad && securityInfo?.origin?.startsWith(MOZ_NULL_PRINCIPAL) == true) {
```
pocmo: Oh, what's `moz-nullprincipal:`?

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -99,6 +101,12 @@ class GeckoEngineSession(
      */
     private fun createNavigationDelegate() = object : GeckoSession.NavigationDelegate {
         override fun onLocationChange(session: GeckoSession?, url: String) {
+            // Ignore initial load of about:blank (see https://github.com/mozilla-mobile/android-components/issues/403)
+            if (initialLoad && url == ABOUT_BLANK) {
+                return
+            }
+            initialLoad = false
```
pocmo: We set `initialLoad` to `false` in onLocationChange and onSecurityChange. Is it guaranteed that only one of those happens? Will it never happen that one of them gets called, we set `initialLoad` to `false` and then do not ignore `about:blank` in the other callback?

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -135,6 +143,12 @@ class GeckoEngineSession(
             session: GeckoSession?,
             securityInfo: GeckoSession.ProgressDelegate.SecurityInformation?
         ) {
+            // Ignore initial load of about:blank (see https://github.com/mozilla-mobile/android-components/issues/403)
+            if (initialLoad && securityInfo?.origin?.startsWith(MOZ_NULL_PRINCIPAL) == true) {
+                return
+            }
+            initialLoad = false
+
```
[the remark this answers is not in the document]
csadilek: It appears the be the principal used for about:blank. So, this saves us handling one additional event. If we're not forwarding onLocationChange we also shouldn't forward the corresponding onSecurityChange. See here: https://dxr.mozilla.org/mozilla-central/source/mobile/android/base/java/org/mozilla/gecko/webapps/WebAppActivity.java#119

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -99,6 +101,12 @@ class GeckoEngineSession(
      */
     private fun createNavigationDelegate() = object : GeckoSession.NavigationDelegate {
         override fun onLocationChange(session: GeckoSession?, url: String) {
+            // Ignore initial load of about:blank (see https://github.com/mozilla-mobile/android-components/issues/403)
+            if (initialLoad && url == ABOUT_BLANK) {
+                return
+            }
+            initialLoad = false
+
```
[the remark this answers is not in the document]
csadilek: In my testing onSecurityChange always happened after, but that's not much of a guarantee. I will update this. Let's only set initialLoad = false in onLocationChange if it's not ABOUT_BLANK.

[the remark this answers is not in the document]
pocmo: Okay, sounds good!

csadilek:
@pocmo ok, updated.

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -135,6 +143,11 @@ class GeckoEngineSession(
             session: GeckoSession?,
             securityInfo: GeckoSession.ProgressDelegate.SecurityInformation?
         ) {
+            // Ignore initial load of about:blank (see https://github.com/mozilla-mobile/android-components/issues/403)
+            if (initialLoad && securityInfo?.origin?.startsWith(MOZ_NULL_PRINCIPAL) == true) {
+                return
+            }
```
pocmo: If in your tests `onSecurityChange` happened after `onLocationChange`... will this block ever get executed?

csadilek: @pocmo yes, it will be executed for about:blank. Here's the flow of events for a single open tab:

OnLocationChange (about:blank) -> ignore (initialLoad = true)
OnLocationChange (about:blank) -> ignore (initialLoad = true)
OnSecurityChange (about:blank, origin:moz-nullprincipal) -> ignore (initialLoad = true)
OnLocationChange (taburl) -> notify (initialLoad = false)
OnSecurityChange (taburl, origin:taburl) -> notify (initialLoad = false)

pocmo: Thanks. Got it!

PR #444, opened 2018-07-18
Issue #432: Fretboard: Kinto delete diffs might lead to crash.

### LINKED ISSUE #432: Fretboard: Kinto delete diffs might lead to crash
From: https://github.com/mozilla-mobile/fretboard/issues/48

> The current implementation for parsing Kinto delete diffs gets the max last modified from the list of updated experiments. Although this works fine for updates, for deletes it leads to downloading a delete diff response multiple times. The first time we parse that delete response everything works fine, but the second one it crashes due to the single function not finding any value with that id.
> 
> Also, we are downloading the delete diff multiple times, using more data for our users.

fercarcedo:
Previously, the last modified date was calculated as the max of the updated list of experiments (thus not including deleted items). This caused a problem where we'd be downloading delete diffs multiple times, with the associated data costs to the users, and also because the deleted experiment doesn't exist anymore, it could lead to a crash.

This pull request solves this, and instead of keeping around deleted experiments (maybe flagging them with a deleted attributed), with the associated memory and disk costs of this, I modified the signature of ExperimentSource and ExperimentSource from List<Experiment> to a SyncResult object, a class which contains a list of experiments and their associated last modified date.

Closes #432

### DIFF
--- a/components/service/fretboard/src/main/java/mozilla/components/service/fretboard/ExperimentSource.kt
+++ b/components/service/fretboard/src/main/java/mozilla/components/service/fretboard/ExperimentSource.kt
@@ -14,9 +14,9 @@ interface ExperimentSource {
      * parsing the response into experiments
      *
      * @param client Http client to use, provided by Fretboard
-     * @param experiments list of already downloaded experiments
+     * @param snapshot list of already downloaded experiments
      * (in order to process a diff response, for example)
      * @return modified list of experiments
      */
-    fun getExperiments(experiments: List<Experiment>): List<Experiment>
+    fun getExperiments(snapshot: ExperimentsSnapshot): ExperimentsSnapshot
 }
--- a/components/service/fretboard/src/main/java/mozilla/components/service/fretboard/ExperimentStorage.kt
+++ b/components/service/fretboard/src/main/java/mozilla/components/service/fretboard/ExperimentStorage.kt
@@ -14,12 +14,12 @@ interface ExperimentStorage {
      *
      * @param experiments list of experiments to store
      */
-    fun save(experiments: List<Experiment>)
+    fun save(snapshot: ExperimentsSnapshot)
 
     /**
      * Reads experiments from disk
      *
      * @return experiments from disk
      */
-    fun retrieve(): List<Experiment>
+    fun retrieve(): ExperimentsSnapshot
 }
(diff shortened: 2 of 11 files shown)


pocmo:
detekt complained about new lines:

```
Ruleset: style - 35min debt
	NewLineAtEndOfFile - [ExperimentStorage.kt] at /build/android-components/components/service/fretboard/src/main/java/mozilla/components/service/fretboard/ExperimentStorage.kt:25:1
	NewLineAtEndOfFile - [ExperimentsSerializer.kt] at /build/android-components/components/service/fretboard/src/main/java/mozilla/components/service/fretboard/storage/flatfile/ExperimentsSerializer.kt:64:1
	NewLineAtEndOfFile - [FlatFileExperimentStorage.kt] at /build/android-components/components/service/fretboard/src/main/java/mozilla/components/service/fretboard/storage/flatfile/FlatFileExperimentStorage.kt:31:1
	NewLineAtEndOfFile - [SyncResult.kt] at /build/android-components/components/service/fretboard/src/main/java/mozilla/components/service/fretboard/SyncResult.kt:13:1
	NewLineAtEndOfFile - [KintoExperimentSource.kt] at /build/android-components/components/service/fretboard/src/main/java/mozilla/components/service/fretboard/source/kinto/KintoExperimentSource.kt:73:1
	NewLineAtEndOfFile - [ExperimentSource.kt] at /build/android-components/components/service/fretboard/src/main/java/mozilla/components/service/fretboard/ExperimentSource.kt:22:1
	NewLineAtEndOfFile - [Fretboard.kt] at /build/android-components/components/service/fretboard/src/main/java/mozilla/components/service/fretboard/Fretboard.kt:113:1
```

components/service/fretboard/src/main/java/mozilla/components/service/fretboard/SyncResult.kt
```diff
@@ -0,0 +1,13 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.fretboard
+
+/**
+ * Represents an experiment sync result
+ */
+data class SyncResult(
```
pocmo: Let's try to find a different name. `Sync` is usually associated with Firefox Sync and that may get confusing. How about something like `ExperimentsSnapshot`?

And then in the code it's `snapshot.experiments` and `snapshot.lastModified`?

fercarcedo: Looks good!

fercarcedo:
I have changed the name to `ExperimentsSnapshot` :)

PR #446, opened 2018-07-18
Issue #433: Fretboard: Certificate pinning

### LINKED ISSUE #433: Fretboard: Certificate pinning
From https://github.com/mozilla-mobile/fretboard/issues/15

> In addition to verifying the signature of a collection (#13) we should use certificate pinning:
> 
> https://www.owasp.org/index.php/Certificate_and_Public_Key_Pinning#What_Is_Pinning.3F
> Fennec bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1329721

fercarcedo:
This pull request introduces certificate pinning. It adds a pinCertificates method to ExperimentSource with a set of base64-encoded SHA-256 of the certificate subject public key info.

The implementation first gets the list of trusted certificates as detailed, cleaning the certificate chain using X509TrustManagerExtensions as described here and here.

I also discovered there are libraries available such as TrustKit which replicate the behavior of Android N+ Network Security Configuration, but I decided against using it because it adds extra bloat to the library and also the user has to create an XML file and register it on the Manifest, I don't know if you have different opinion.

Closes #433


components/service/fretboard/src/test/java/mozilla/components/service/fretboard/source/kinto/TestTrustManager.kt
```diff
@@ -0,0 +1,23 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this
+* file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.fretboard.source.kinto
+
+import java.security.cert.X509Certificate
+import javax.net.ssl.X509TrustManager
+
+class TestTrustManager : X509TrustManager {
```
sputnikci: \[ktlint\] WARNING: \[colon\-spacing\] Missing spacing before ":" in column 23

pocmo:
Blocked:
 * Pending security review

pocmo:
@fercarcedo Can you rebase this branch? There are some conflicts here. :)

fercarcedo:
Rebased

pocmo:
Given the feedback of the security team I think we should move forward without certificate pinning. Signature validation is a stronger signal and does not have all the problems that come with cert pinning.

PR #447, opened 2018-07-18
Issue #434: Fretboard: Verify signatures of experiments collection

### LINKED ISSUE #434: Fretboard: Verify signatures of experiments collection
From: https://github.com/mozilla-mobile/fretboard/issues/13

> Whenever the experiments collection is modified the signature service signs the collection in Kinto. To make sure a configuration we downloaded is coming from an authorative source we should validate the signature. Note that this is not done in Fennec yet (Bugzilla bug).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-3669)

fercarcedo:
This pull request adds a mechanism to verify signature of experiments collection. It does so by retrieving the certificate chain from the `x5u` field from Kinto, checking the certificate validity and then verifying the signature of the canonical json of the updated experiments using the public key from the certificate and the expected signature from the Kinto response.

Closes #434


components/service/fretboard/src/main/java/mozilla/components/service/fretboard/source/kinto/KintoExperimentSource.kt
```diff
@@ -22,16 +23,24 @@ class KintoExperimentSource(
     val baseUrl: String,
     val bucketName: String,
     val collectionName: String,
-    private val client: HttpClient = HttpURLConnectionHttpClient()
+    client: HttpClient = HttpURLConnectionHttpClient()
 ) : ExperimentSource {
+    var validateSignature = true
```
pocmo: Is this something we could pass in via the constructor?

components/service/fretboard/src/main/java/mozilla/components/service/fretboard/source/kinto/SignatureVerifier.kt
```diff
@@ -0,0 +1,192 @@
...
+import java.security.PublicKey
+import java.security.Signature
+import java.security.SignatureException
+import java.security.cert.CertificateException
+import java.security.cert.CertificateFactory
+import java.security.cert.X509Certificate
+import java.util.Date
+import java.util.concurrent.TimeUnit
+
+internal class SignatureVerifier(
```
pocmo: Can you add some comments here describing what the class does? Maybe even add some comments to the code. If you have links to pages that explain more then add them too. :)

components/service/fretboard/src/main/java/mozilla/components/service/fretboard/source/kinto/KintoExperimentSource.kt
fercarcedo: Yes, it could be passed via the constructor, right. I'm going to change it

components/service/fretboard/src/main/java/mozilla/components/service/fretboard/source/kinto/SignatureVerifier.kt
fercarcedo: Okay :)

pocmo:
Blocked:
 * Pending security review

fercarcedo:
Rebased :)

components/service/fretboard/src/main/java/mozilla/components/service/fretboard/Experiment.kt
```diff
@@ -16,6 +16,7 @@ data class Experiment(
     val match: Matcher? = null,
     val bucket: Bucket? = null,
     val lastModified: Long? = null,
+    val schema: Long? = null,
     val payload: ExperimentPayload? = null
 ) {
```
pocmo: We should add some kdoc to explain what schema is. I filed #524 to add kdoc in general here.

components/service/fretboard/src/main/java/mozilla/components/service/fretboard/JSONExtensions.kt
```diff
@@ -0,0 +1,62 @@
...
+internal fun <T> JSONArray?.toList(): List<T> {
+    if (this != null) {
+        val result = ArrayList<T>()
+        for (i in 0 until length())
+            result.add(get(i) as T)
+        return result
+    }
+    return listOf()
+}
+
```
pocmo: Is this file still needed? Didn't we move them into ktx? Or are there new ones in here?

fercarcedo: No, actually it isn't needed. I have updated the pull request to use the extensions from ktx instead

fercarcedo:
Regarding the schema field: this field was added because Kinto responses have it and we need to store it in order to be able to recreate the collection later in order to validate its signature.

However, now I think this shouldn't be a field of `Experiment`, because it's only specific to Kinto, and more fields like these could arise.

So instead what I thought about doing is the following: replace it with a `Map<String,Any>` of fields, and fill that map with these custom attributes we only need in order to validate the signature.

What do you think @pocmo?

pocmo:
@fercarcedo Can you disable signature verification by default and rebase the branch? I'd like to land this - but in a disabled state. Apps will have to explicitly enable it for now.

fercarcedo:
Done :)

components/service/fretboard/src/main/java/mozilla/components/service/fretboard/Experiment.kt
```diff
@@ -37,7 +37,12 @@ data class Experiment(
     /**
      * Experiment associated metadata
      */
-    val payload: ExperimentPayload? = null
+    val payload: ExperimentPayload? = null,
+    /**
+     * Last time the experiment schema was modified
```
pocmo: nit: Looks like here is an empty line missing before the doc comment?

components/service/fretboard/src/main/java/mozilla/components/service/fretboard/source/kinto/SignatureVerifier.kt
```diff
@@ -0,0 +1,266 @@
...
+            signed = false
+        } catch (e: NoSuchProviderException) {
+            signed = false
+        } catch (e: SignatureException) {
+            signed = false
+        }
+
+        if (!signed) {
+            throw ExperimentDownloadException("Invalid certificate chain")
+        }
```
pocmo: Now that `ExperimentDownloadException` can wrap an exception I think we want to do this here. And I assume we can do that in the catch blocks directly and do not need `signed` here anymore?

components/service/fretboard/src/main/java/mozilla/components/service/fretboard/Experiment.kt
fercarcedo: I think now it should be fine...

components/service/fretboard/src/main/java/mozilla/components/service/fretboard/source/kinto/SignatureVerifier.kt
fercarcedo: I have changed that now :)

PR #452, opened 2018-07-18
Persist FxA state when logging in via Webview

carolkng:
Related to issue #343 - the custom tabs flow persist state, but the same methods were forgotten in the webview flow
 
r? @csadilek


samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/MainActivity.kt
```diff
@@ -122,6 +122,10 @@ open class MainActivity : AppCompatActivity(), LoginFragment.OnLoginCompleteList
             runOnUiThread {
                 txtView.text = getString(R.string.signed_in, "${value.displayName ?: ""} ${value.email}")
             }
+            account?.toJSONString().let {
```
csadilek: This block is exactly the same as https://github.com/carolkng/android-components/blob/b841d5d2a7a9870dbd330faef68320c1fc13eae7/samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/MainActivity.kt#L99, right? Let's refactor that to a function

Issue #400, opened 2018-07-10
Modifying sessionManager from an observer may cause observers seeing events in different order and other "in-between states"


pocmo:
For #379 I added code in the sample browser that will add a new tab if all tabs get closed. However this code is indirectly called from a session manager observer. If there are more observers then the following ones would get called with the session manager state already changed - or other calls in between.

Let's say there are two observers (A / B) registered on the session manager. And the session manager has one session X.

* `session manager` (`X`)
* `X` is removed from `session manager`.
* `A.onSessionRemoved(X)` gets called: `A` adds a new session `Y` to `session manager` and this will notify observers:
  * `A.onSessionAdded(Y)` gets called
  * `B.onSessionAdded(Y)` gets called
* B.onSessionRemoved(X) gets called.

In the end A and B have observed the same events. But in different order. B never saw an empty session manager:

* For A: (X) -> () -> (Y)
* For B: (X) -> (X,Y) -> (X)

pocmo:
I wonder what a good solution to that is.

The only solution I can come up with is that all modifying methods are not executed synchronously but on a single-threaded executor. With that modifications during a dispatch would just be queued until the current update is made and dispatched to all observers.

csadilek:
I think we run into this because we're using an reentrant lock (`synchronized(observers)`) in our `ObserverRegistry.notifyObservers`. So, when notifying one observer that notifies more, those subsequent calls can enter and execute right away (before the original observers are notified.) If we used a semaphore/mutex instead those calls would instead have to wait until all observers are executed. I think with this change we could still execute them synchronously which is simpler than introducing a thread/queue?

pocmo:
> If we used a semaphore/mutex instead those calls would instead have to wait until all observers are executed.

Wouldn't we deadlock then? It's the same thread that is notifying observers and then modifying sessionManager?

csadilek:
>Wouldn't we deadlock then? It's the same thread that is notifying observers and then modifying sessionManager?

Yes, we can't block on the mutex. I was mainly trying to look for a synchronous solution. Here's what I was thinking re: mutex/semaphore:


```
private val notifyMutex = Semaphore(1)
private val queue = LinkedList<T.() -> Unit>()
...
override fun notifyObservers(block: T.() -> Unit) {
        if (notifyMutex.tryAcquire()) {
            observers.forEach {
                it.block()
            }
            notifyMutex.release()

            if (queue.isNotEmpty()) {
[...]

PR #453, opened 2018-07-19
Issue #400: Prevent side effects / glitches in Observable implementations.

pocmo:
This is a first version of a helper to fix the issue described in #400.

It's not used in any `Observable` yet. We may want to use it in `SessionManager` and `Session` but probably not `EngineSession` implementations because observers do not really modify the internal state.

### DIFF
--- a/components/support/utils/src/main/java/mozilla/components/support/utils/observer/SideEffects.kt
+++ b/components/support/utils/src/main/java/mozilla/components/support/utils/observer/SideEffects.kt
@@ -1,0 +1,60 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, you can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.support.utils.observer
+
+import java.util.LinkedList
+
+/**
+ * This helper class prevents side effects (glitches) in an <code>Observable</code> implementation
+ * where an observer can modify the internal state of an <code>Observable</code> from an observer
+ * callback. In this situation follow-up observers may see a modified state that may conflict with
+ * the change the observer was notified about.
+ *
+ * To avoid those side effects in an <code>Observable</code> create an internal instance of
+ * <code>SideEffects</code> and wrap methods modifying the <code>Observable</code> state inside
+ * <code>modifyWithoutSideEffects()</code>. Methods that just read the current state can be wrapped
+ * in <code>readWithLock</code>.
+ *
+ * SideEffects queues modifications to an <code>Observable</code> until all observers have been
+ * notified and then performs them sequentially
+ *
+ * For an example of observer side effects / glitches see SideEffectsTest<code>.
+ */
+class SideEffects {
+    // Those properties have 'internal' visibility with @PublishedApi annotation so that they can be
+    // inlined by the compiler.
+    @PublishedApi internal val lock = Any()
+    @PublishedApi internal val queue: LinkedList<() -> Unit> = LinkedList()
+    @PublishedApi internal var isProcessing = false
+
+    inline fun modifyWithoutSideEffects(crossinline action: () -> Unit) {
+        synchronized(lock) {
+            if (isProcessing) {
+                queue.add { action() }
+            } else {
+                isProcessing = true
+
+                try {
+                    action()
+                } finally {
+                    processQueue()
+                }
+
+                isProcessing = false
+            }
+        }
+    }
+
+    @PublishedApi internal fun processQueue() {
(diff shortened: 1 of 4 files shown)


components/support/utils/src/main/java/mozilla/components/support/utils/observer/SideEffects.kt
```diff
@@ -0,0 +1,53 @@
...
+ */
+class SideEffects {
+    // Those properties have 'internal' visibility with @PublishedApi annotation so that they can be
+    // inlined by the compiler.
+
+    @PublishedApi internal val lock = Any()
+    @PublishedApi internal val queue: MutableList<() -> Unit> = mutableListOf()
+    @PublishedApi internal var isProcessing = false
+
+    inline fun modifyWithoutSideEffects(crossinline action: () -> Unit): Unit = synchronized(lock) {
```
sputnikci: \[ktlint\] WARNING: \[no\-unit\-return\] Unnecessary "Unit" return type in column 74

pocmo:
@csadilek What do you think? If you agree I'll go ahead with the patch and apply it to `SessionManager` and `Session`.

components/support/utils/src/main/java/mozilla/components/support/utils/observer/SideEffects.kt
```diff
@@ -0,0 +1,53 @@
...
+                action()
+            }
+        } else {
+            isProcessing = true
+
+            action()
+
+            queue.forEach { action ->
+                action()
+            }
```
pocmo: This may cause a concurrent modification exception if we are adding something to the queue while processing the queue. I may add a test for that and fix it. We may want to use an actual queue here.

components/support/utils/src/main/java/mozilla/components/support/utils/observer/SideEffects.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    inline fun modifyWithoutSideEffects(crossinline action: () -> Unit): Unit = synchronized(lock) {
+        if (isProcessing) {
+            queue.add {
+                action()
+            }
+        } else {
+            isProcessing = true
+
+            action()
```
pocmo: I'm afraid that a `return` statement inside the lambda will jump out of this method because it's getting inlined. I thought `crossinline` would prevent that. But I seem to be able to still use a return@label:

```
fun someMethod(bar: Foo?) = sideEffects.modifyWithoutSideEffects {
    if (bar == null) {
        // This doesn't compile because of 'crossinline'
        return
    }

    if (bar == null) {
        // This compiles... but isn't it the same as the one above?
        return@modifyWithoutSideEffects
    }

    // ...
}
```

pocmo: I wonder if every `action()` should actually be wrapped in:

```Kotlin
try {
    action()
} finally {
   processQueue()
}
```

... so that we will always continue with the next action if one of them returns after being inlined?

csadilek:
@pocmo yes, I like the solution. This way the observable can "decide" whether or not it allows side-effects.  So, let's get in hardened and landed ;).

components/support/utils/src/main/java/mozilla/components/support/utils/observer/SideEffects.kt
pocmo: This try catch doesn't really work... Every action we process in the queue would need to be wrapped in the same try-catch. We can do that with recursion - but we can't inline a method that is called recursively..

csadilek: Interesting. Shouldn't we just use noinline for the lambda here? It shouldn't have the power to jump out of the executing method :).

pocmo: I wanted to inline aggressively so that using SideEffects has no penalty. Otherwise Java (6) will need to create an object (+ the class) for every lambda :(

pocmo:
Updated the PR.

This is working nicely with `SessionManager` now. I just wanted to apply it to `Session` and then realized that we are using `Delegates.observable` here. Queuing up the change here isn't as straight-forward.

pocmo:
> Unable to instrument file with Jacoco: /home/travis/build/mozilla-mobile/android-components/components/browser/session/build/tmp/kotlin-classes/debug/mozilla/components/browser/session/SessionManager$removeSessions$$inlined$modifyWithoutSideEffects$1.class

🔥

pocmo:
Closing: I moved this PR out of the sprint. I want to think a bit more about this.

PR #462, opened 2018-07-20
Issue #456: Fretboard: Allow filtering by release channel

### LINKED ISSUE #456: Fretboard: Allow filtering by release channel
In Focus we would like to filter by release channel (e.g. have an experiment that only runs in Beta). The release channel is obviously something that is very app specific. So we would need some way to provide this data to `Fretboard`. There may be other data that we want to provide to the library in the future (or maybe even override some of the existing fields). That makes we wonder if we should pull that behavior out into its own class that `ExperimentEvaluator` uses to get the values. Basically turn the `RegionProvider` we already have into something that can provide more than just the region.

fercarcedo:
This pull request adds support to allow filtering by release channel, and also for overriding values for filters throught a new `ValuesProvider` class that users of the library can extend when creating the `Fretboard` instance

Closes #456


components/service/fretboard/src/main/java/mozilla/components/service/fretboard/ExperimentEvaluator.kt
```diff
@@ -30,19 +30,23 @@ internal class ExperimentEvaluator(private val regionProvider: RegionProvider? =
...
+            val region = valuesProvider.getRegion(context)
             val matchesRegion = !(region != null &&
                                         experiment.match.regions != null &&
                                         experiment.match.regions.isNotEmpty() &&
                                         experiment.match.regions.none { it == region })
-            val appVersion = context.packageManager.getPackageInfo(context.packageName, 0).versionName
+            val releaseChannel = valuesProvider.getReleaseChannel(context)
+            val matchesReleaseChannel = !(releaseChannel != null &&
+                                                experiment.match.releaseChannel != null &&
+                                                releaseChannel != experiment.match.releaseChannel)
```
pocmo: With [De Morgan's Law ](https://en.wikipedia.org/wiki/De_Morgan%27s_laws) you can use the complement which may be easier to read (doesn't need the double negation):

```Kotlin
 val matchesReleaseChannel = releaseChannel == null ||
    experiment.match.releaseChannel == null ||
    releaseChannel == experiment.match.releaseChannel
```

components/service/fretboard/src/main/java/mozilla/components/service/fretboard/JSONExtensions.kt
```diff
@@ -0,0 +1,50 @@
...
+        return getLong(key)
+    }
+    return null
+}
+
+fun JSONObject.putIfNotNull(key: String, value: Any?) {
+    if (value != null) {
+        put(key, value)
+    }
+}
```
pocmo: Let's create a new issue to potentially move those extensions into our support-ktx package where we collect a bunch of useful extensions.

components/service/fretboard/src/main/java/mozilla/components/service/fretboard/ValuesProvider.kt
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Class used to provide
+ * custom filter values
+ */
+open class ValuesProvider {
+    /**
+     * Provides the user's language
+     *
+     * @return language
```
pocmo: Let's either add an explanation here or remove the "@return language" doc because it doesn't add anything.

components/service/fretboard/src/main/java/mozilla/components/service/fretboard/ValuesProvider.kt
```diff
@@ -0,0 +1,87 @@
...
+     */
+    open fun getAppId(context: Context): String {
+        return context.packageName
+    }
+
+    /**
+     * Provides the user's region
+     *
+     * @return user's region
+     */
```
pocmo: I think it would be helpful if you could describe the required format here and in other methods where its helpful. Otherwise someone implementing this has no idea what kind of String to return here.

PR #473, opened 2018-07-20
Closes #450: Fetch libjnidispatch at build-time and remove from tree.

### LINKED ISSUE #450: Fetch libjnidispatch at build-time and remove from tree
We are currently checking in libjnidispatch because I couldn't find it anywhere...seems we should version and publish it?

csadilek:
Alright, that's kinda cool that this works. 💥 

The android libjnidispatch.so files are not part of the published jna.jar, but luckily there are tagged releases of the distribution on GitHub. So, fetching those instead, unzipping and putting in the right folder.

Also, I tested that this all this works :).


csadilek:
We should refactor this into a separate module, when we bring in additional libraries that use JNA...

ncalexan:
> /cc @ncalexan This landed now. I plan to refactor this to a generic android-jna module which other android components (that use Rust) can then depend on.

Oh, nice -- I will steal this.  There's at least one Android JNA module around -- that's where @thomcc grabbed the libraries from to begin with, IIUC.  @thomcc?

thomcc:
Nope, I grabbed them from https://github.com/mozilla/mentat/tree/master/sdks/android/Mentat/library/src/main/jniLibs, not sure where I'd find them otherwise.

PR #477, opened 2018-07-23
Update GeckoView Nightly (2018.07.24) and Beta (62.0b9)

pocmo:
This update introduces GeckoResult and I had to update some code.

The tests for saveState() are hanging now and I see the following exception in the test run log:
> Exception in thread "ForkJoinPool.commonPool-worker-1" java.lang.RuntimeException: Can't
> create handler inside thread that has not called Looper.prepare()

This needs to be fixed in order to land this. In this version of the patch the two hanging tests are marked with `@Ignore`.


pocmo:
@csadilek I found the issue. I did a mistake when migrating onLoadRequest(). This method returned `flase` and now I return `true` intercepting every page load 😂 . I'll update the PR!

pocmo:
This works now and we render pages again.

But saveState() is just deadlocking :(

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -74,15 +75,17 @@ class GeckoEngineSession(
     @Throws(GeckoEngineException::class)
```
csadilek: Just noticed that we don't throw this anymore. We can also delete that type now.

pocmo: I'll file a follow-up for that.

PR #484, opened 2018-07-24
Issue #115: Core ping: Report experiments

### LINKED ISSUE #115: Core ping: Report experiments
https://github.com/mozilla-mobile/telemetry-android/issues/15

`"experiments": [<string>, /* … */], // Optional, array of identifiers for the active experiments`

fercarcedo:
This pull request adds active experiments to telemetry core ping. It adds an `ExperimentsMeasurement` class and adds it to the `TelemetryCorePingBuilder`. It also adds a `recordActiveExperiments` method to the `Telemetry` class and tests it inside `TelemetryTest`, I think that's all that's needed in order for an app to provide such values.

Closes #115

### DIFF
--- a/components/service/telemetry/src/main/java/org/mozilla/telemetry/Telemetry.java
+++ b/components/service/telemetry/src/main/java/org/mozilla/telemetry/Telemetry.java
@@ -23,6 +23,7 @@ import org.mozilla.telemetry.storage.TelemetryStorage;
 
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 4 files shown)


components/service/telemetry/src/main/java/org/mozilla/telemetry/Telemetry.java
```diff
@@ -190,6 +191,27 @@ public Telemetry recordSearch(@NonNull  String location, @NonNull String identif
         return this;
     }
 
+    /**
+     * Records the list of active experiments
+     *
+     * @param activeExperimentsIds list of active experiments ids
+     */
+    public Telemetry recordActiveExperiments(List<String> activeExperimentsIds) {
```
pocmo: Let's add a helper to `Fretboard` that returns the list of active experiments so that we can easily generate this list of Ids from it.

PR #490, opened 2018-07-24
Closes #352: Expose tracking protection in session/engine

### LINKED ISSUE #352: Expose tracking protection in session/engine
The goal of this issue is working towards getting browser-session into a state where we can use it in Firefox for Fire TV (see [Google doc](https://docs.google.com/spreadsheets/d/17a4bIYB88eovF_UwiaXjL3uJa3r7Fidl8IzAzV6QEbw/edit#gid=0)). Tracking protection is the last big missing piece.

In GeckoView enabling tracking protection is done automatically for private browsing. In the WebView version of Focus we do this manually by intercepting URLs from the TP lists.

In Focus' version of `session` we save whether tracking protection is enabled and how many trackers have been blocked for the currently loaded page (That's something we do not need in Fire TV; only later for Focus).

csadilek:
This exposes tracking protection in our session/engine. It's currently only implemented for GeckoEngine. I suggest filing a follow-up for making this work with WebView so we can make some progress?

### DIFF
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -12,12 +12,14 @@ import org.mozilla.geckoview.GeckoResponse
 import org.mozilla.geckoview.GeckoRuntime
 import org.mozilla.geckoview.GeckoSession
 import org.mozilla.geckoview.GeckoResult
+import org.mozilla.geckoview.GeckoSessionSettings
 
 /**
  * Gecko-based EngineSession implementation.
  */
+@Suppress("TooManyFunctions")
 class GeckoEngineSession(
-    runtime: GeckoRuntime
+    private val runtime: GeckoRuntime
 ) : EngineSession() {
 
     internal var geckoSession = GeckoSession()
@@ -29,6 +31,7 @@ class GeckoEngineSession(
 
         geckoSession.navigationDelegate = createNavigationDelegate()
         geckoSession.progressDelegate = createProgressDelegate()
+        geckoSession.trackingProtectionDelegate = createTrackingProtectionDelegate()
     }
 
     /**
@@ -175,6 +178,23 @@ class GeckoEngineSession(
                 }
             }
         }
+    }
+
+    private fun createTrackingProtectionDelegate() = GeckoSession.TrackingProtectionDelegate {
+        session, uri, _ ->
+            session?.let { uri?.let { notifyObservers { onTrackerBlocked(it) } } }
+    }
+
+    override fun enableTrackingProtection(policy: TrackingProtectionPolicy) {
+        geckoSession.settings.setBoolean(GeckoSessionSettings.USE_TRACKING_PROTECTION, true)
+        runtime.settings.trackingProtectionCategories = policy.categories
+        notifyObservers { onTrackerBlockingEnabledChange(true) }
+    }
+
+    override fun disableTrackingProtection() {
+        geckoSession.settings.setBoolean(GeckoSessionSettings.USE_TRACKING_PROTECTION, false)
+        runtime.settings.trackingProtectionCategories = TrackingProtectionPolicy.none().categories
+        notifyObservers { onTrackerBlockingEnabledChange(false) }
     }
 
     companion object {
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
@@ -15,6 +15,7 @@ import kotlin.properties.Delegates
 /**
  * Value type that represents the state of a browser session. Changes can be observed.
  */
+@Suppress("TooManyFunctions")
 class Session(
     initialUrl: String,
     val source: Source = Source.NONE,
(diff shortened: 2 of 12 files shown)


components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
```diff
@@ -40,6 +42,8 @@ class Session(
         fun onSecurityChanged(session: Session, securityInfo: SecurityInfo) = Unit
         fun onCustomTabConfigChanged(session: Session, customTabConfig: CustomTabConfig?) = Unit
         fun onDownload(session: Session, download: Download): Boolean = false
+        fun onTrackerBlockingEnabledChanged(session: Session, blockingEnabled: Boolean): Unit = Unit
```
sputnikci: \[ktlint\] WARNING: \[no\-unit\-return\] Unnecessary "Unit" return type in column 90

components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
```diff
@@ -40,6 +42,8 @@ class Session(
...
+        fun onTrackerBlocked(session: Session, blocked: String, all: List<String>): Unit = Unit
```
sputnikci: \[ktlint\] WARNING: \[no\-unit\-return\] Unnecessary "Unit" return type in column 85

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -11,23 +11,30 @@ import org.mozilla.geckoview.GeckoResponse
...
+import org.mozilla.geckoview.GeckoSessionSettings
 
 /**
  * Gecko-based EngineSession implementation.
  */
+@Suppress("TooManyFunctions")
 class GeckoEngineSession(
-    runtime: GeckoRuntime
+    private val runtime: GeckoRuntime,
+    privateMode: Boolean = false
```
pocmo: I think we should keep private browsing and tracking protection separated. Is this possible with GeckoView?

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
```diff
@@ -19,17 +19,17 @@ class GeckoEngine(
...
     }
 
     /**
-     * Create a new Gecko-based EngineSession.
+     * Creates a new Gecko-based EngineSession.
      */
-    override fun createSession(): EngineSession {
-        return GeckoEngineSession(runtime)
+    override fun createSession(private: Boolean): EngineSession {
+        return GeckoEngineSession(runtime, private)
```
pocmo: Eventually we may want to introduce some kind of `EngineSessionSettings`?

components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
```diff
@@ -174,6 +178,20 @@ class Session(
...
+        notifyObservers(old, new) { onTrackerBlockingEnabledChanged(this@Session, trackerBlockingEnabled) }
+    }
+
+    /**
+     * List of URIs that have been blocked in this session.
+     */
+    var trackersBlocked: List<String> by Delegates.observable(emptyList()) { _, old, new ->
+        notifyObservers(old, new) { onTrackerBlocked(this@Session, trackersBlocked.last(), trackersBlocked) }
+    }
+
```
pocmo: So far we only need the `count` in our apps and not the URLs.

components/browser/session/src/main/java/mozilla/components/browser/session/engine/EngineObserver.kt
```diff
@@ -39,6 +40,14 @@ internal class EngineObserver(val session: Session) : EngineSession.Observer {
                 ?: "", issuer ?: "")
     }
 
+    override fun onTrackerBlocked(url: String) {
+        session.trackersBlocked += url
+    }
+
```
pocmo: This information is dependent on the page loaded and we clear that in Focus whenever we load a new page.

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
csadilek: OK, let's postpone private browsing. I will remove the commit.

components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
csadilek: I saw, yes, but having the list seems like a better solution. It doesn't require additional logic and we can always get the count from the list. It also seemed likely to me that we want to be able to tell *what* was blocked. So, my preference would be to keep the list.

components/browser/session/src/main/java/mozilla/components/browser/session/engine/EngineObserver.kt
csadilek: OK, I will update and add this.

components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
pocmo: Alright. We may be interested in the blocked _categories_ soon. But the GV API doesn't give us that for now anyways.

PR #522, opened 2018-07-26
Add KDocs and documentation for the FxA library

### LINKED ISSUE #426: Provide documentation for FxA lib and sample app
Let's write up some details on how to use the API and what the differences are between using CustomTabs and a browser view etc.

carolkng:
Fixes #426 
Related to mozilla/application-services#122

r? @pocmo, @vladikoff, @eoger

### DIFF
--- a/components/service/firefox-accounts/README.md
+++ b/components/service/firefox-accounts/README.md
@@ -4,7 +4,14 @@ A library for integrating with Firefox Accounts.
 
 ## Motivation
 
-The goal of this library is to provide an API for integrating with Firefox Accounts.
+The **Firefox Accounts Android Component** provides a way for Android applications to do the following:
+
+* Obtain OAuth tokens that can be used to access the user's data in Mozilla-hosted services like Firefox Sync
+* Fetch client-side scoped keys needed for end-to-end encryption of that data
+* Fetch a user's profile to personalize the application
+
+See also the [sample app](https://github.com/mozilla-mobile/android-components/tree/master/samples/firefox-accounts)
+for help with integrating this component into your application.
 
 ## Usage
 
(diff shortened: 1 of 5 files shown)


eoger:
Looks good, however `components/service/firefox-accounts/README.md` should live in mozilla/application-services

pocmo:
> Looks good, however components/service/firefox-accounts/README.md should live in mozilla/application-services

I understand that both sides have an interest in having documentation. We would like to have a documentation for the top-level Android API here in this repository - where the code lives. For internals or Rust-level documentation it's fine to just link out.

components/service/firefox-accounts/README.md
```diff
@@ -16,6 +23,102 @@ Use gradle to download the library from JCenter:
...
 
+### Start coding
+
+> This tutorial is for version 0.15 of the FxA client.
+
+First you need some OAuth information. Generate a `client_id`, `redirectUrl` and find out the scopes for your application.
+See Firefox Account documentation for that. 
+
+Once you have the OAuth info, you can start adding `FxAClient` to your Android project.
+As part of the OAuth flow your application will be opening up a Web view or open the system browser.
```
pocmo: nit: `WebView`

pocmo: Should we mention Custom Tabs here explicitly? They are using the system browser too but it allows the app to customize the look and feel.

components/service/firefox-accounts/README.md
```diff
@@ -16,6 +23,102 @@ Use gradle to download the library from JCenter:
...
+Currently the SDK does not provide the Web view, you have to write it yourself.
+
+Create a global `account` object: 
+
+```kotlin
+var account: FirefoxAccount? = null
+```
+
+You will need to save state for FxA in your app, this example just uses `SharedPreferences`. We suggest using the Android key store for this data.
```
pocmo: nit: Should we link out to https://developer.android.com/training/articles/keystore here?

eoger:
> I understand that both sides have an interest in having documentation. We would like to have a documentation for the top-level Android API here in this repository - where the code lives. For internals or Rust-level documentation it's fine to just link out.

I disagree, the doc we have right now is very much the one we have here: https://github.com/mozilla/application-services/pull/122 living [here](https://mozilla.github.io/application-services/).
I'd like to centralize the high-level documentations for the FxA Android SDK in a single place and avoid duplication/out-of-sync docs.
I believe implementers of the FxA SDK will look first on the application-services website for docs.

carolkng:
We could also consider linking to the [application-services docs](mozilla.github.io/application-services/docs/accounts/fxa-client-android.html) from the README in the components folder, once the android docs in a-s are merged and published.

pocmo:
> I disagree, the doc we have right now is very much the one we have here: mozilla/application-services#122 living here.
I'd like to centralize the high-level documentations for the FxA Android SDK in a single place and avoid duplication/out-of-sync docs.
I believe implementers of the FxA SDK will look first on the application-services website for docs.

I agree with avoiding duplication. However won't it be easier to keep the Android documentation and the code in sync if they live in the same repository? E.g. they can be updated with the same PR. It also seems to be easier to enforce that - especially if the change is made by someone not on the FxA team.

carolkng:
Sorry for letting this sit for so long. At the moment, I agree with Sebastian that the docs' source should live close to the code, since then it'll be harder for devs on the FxA component to forget about documenting major changes to the API (with the caveat that future a-c reviews will have to have an extra focus on the docs). 

I'll amend mozilla/application-services#122 to link to the android component docs.

PR #537, opened 2018-08-02
Use jacoco 0.8.1 for generating usage reports.

pocmo:
I noticed that by default the plugin uses an older version: 0.7.9.201702052155. The latest version is 0.8.1 and the good news is that this version ignores some of the generated code from the Kotlin compiler. :)


pocmo:
> Merging #537 into master will decrease coverage by 0.14%.

What .. the .. hell!

pocmo:
I was expecting it to go up - by a big chunk. 

Let's see what the 0.8.2 snapshot says -> #538.

pocmo:
> Merging #538 into master will decrease coverage by 0.11%.

A little less. But still decreasing. Very confusing.

pocmo:
I updated Kotlin and landed some other things. Let's see what it says after updating the branch.

pocmo:
> Merging #537 into master will decrease coverage by 0.54%.

3xiqswsmbjbiu

PR #540, opened 2018-08-02
Add and update photon icon svgs Closes #539

### LINKED ISSUE #539: Update Icons component with new mobile sized SVGs
I'll do this today!

ekager:
All optimized with avocado

### DIFF
--- a/components/ui/icons/src/main/res/drawable/mozac_ic_arrowhead_down.xml
+++ b/components/ui/icons/src/main/res/drawable/mozac_ic_arrowhead_down.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="18dp"
+    android:height="11dp"
+    android:viewportWidth="18"
+    android:viewportHeight="11">
+    <path
+        android:fillColor="#FFFFFF"
+        android:pathData="M1.2929,1.2929C1.6834,0.9024 2.3166,0.9024 2.7071,1.2929L9,7.5858L15.2929,1.2929C15.6834,0.9024 16.3166,0.9024 16.7071,1.2929C17.0976,1.6834 17.0976,2.3166 16.7071,2.7071L9.7071,9.7071C9.3166,10.0976 8.6834,10.0976 8.2929,9.7071L1.2929,2.7071C0.9024,2.3166 0.9024,1.6834 1.2929,1.2929Z" />
+</vector>
--- a/components/ui/icons/src/main/res/drawable/mozac_ic_arrowhead_up.xml
+++ b/components/ui/icons/src/main/res/drawable/mozac_ic_arrowhead_up.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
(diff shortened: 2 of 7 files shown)


components/ui/icons/src/main/res/drawable/mozac_ic_back.xml
```diff
@@ -2,13 +2,6 @@
...
-        android:height="24dp"
-        android:viewportWidth="24.0"
-        android:viewportHeight="24.0"
-        android:autoMirrored="true">
-    <path
-        android:fillColor="#ffffff"
-        android:pathData="M21,11H5.914l5.793,-5.793a1,1 0,0 0,-1.414 -1.414l-7.5,7.5a1,1 0,0 0,0 1.414l7.5,7.5a1,1 0,0 0,1.414 -1.414L5.914,13H21a1,1 0,0 0,0 -2Z"/>
+<vector xmlns:android="http://schemas.android.com/apk/res/android" android:width="24dp" android:height="24dp" android:viewportWidth="24" android:viewportHeight="24">
+    <path android:fillColor="#F9F9FA" android:pathData="M11.707 3.293c0.39 0.39 0.39 1.024 0 1.414L5.414 11H21c0.552 0 1 0.448 1 1s-0.448 1-1 1H5.414l6.293 6.293c0.39 0.39 0.39 1.024 0 1.414-0.39 0.39-1.024 0.39-1.414 0l-8-8c-0.39-0.39-0.39-1.024 0-1.414l8-8c0.39-0.39 1.024-0.39 1.414 0z"/>
 </vector>
```
pocmo: Mh, the new versions have a different fill color. This may break some of the apps that are already using them (assuming they haven't tinted them).

I wonder whether we should create a `color` for this. Then apps can override the color without having to tint all the icons they load.

components/ui/icons/src/main/res/drawable/mozac_ic_lock.xml
```diff
@@ -2,12 +2,6 @@
...
-<vector xmlns:android="http://schemas.android.com/apk/res/android"
-        android:width="24dp"
-        android:height="24dp"
-        android:viewportWidth="24.0"
-        android:viewportHeight="24.0">
-    <path
-        android:fillColor="#ffffff"
-        android:pathData="M17.25,11H17V8A5,5 0,0 0,7 8v3H6.75A1.814,1.814 0,0 0,5 12.871v6.257A1.814,1.814 0,0 0,6.75 21h10.5A1.814,1.814 0,0 0,19 19.129V12.871A1.814,1.814 0,0 0,17.25 11ZM9,7.739A2.739,2.739 0,0 1,11.739 5h0.523A2.739,2.739 0,0 1,15 7.739V11H9Z"/>
+<vector xmlns:android="http://schemas.android.com/apk/res/android" android:width="24dp" android:height="24dp" android:viewportWidth="24" android:viewportHeight="24">
+    <path android:fillColor="#F9F9FA" android:fillType="evenOdd" android:pathData="M9 8c0-1.657 1.343-3 3-3s3 1.343 3 3v3H9V8zm-2 3V8c0-2.761 2.239-5 5-5s5 2.239 5 5v3c1.105 0 2 0.895 2 2v6c0 1.105-0.895 2-2 2H7c-1.105 0-2-0.895-2-2v-6c0-1.105 0.895-2 2-2z"/>
```
pocmo: Those new ones use `evenOdd` which isn't supported on all versions of Android. Where is that coming from?

ekager: Hm I think that was from avocado

pocmo:
Some of those SVGs I hand-optimized carefully with various tools. Maybe we should only land the new ones for now. :D

components/ui/icons/src/main/res/drawable/mozac_ic_lock.xml
pocmo: Oh, that's weird. Sometimes it's enough to remove the keyword and it will still render fine.

Definitely test Focus on an API 21 device or similar to make sure it renders fine there too.

ekager:
Took out icons with evenOdd added. Just added a few new icons

pocmo:
Thank you!

PR #556, opened 2018-08-03
Issue #555: Fretboard: Add test for non HTTP url for HttpURLConnectionHttpClient

### LINKED ISSUE #555: Fretboard: Add test for non HTTP url for HttpURLConnectionHttpClient

fercarcedo:
Closes #555

### DIFF
--- a/components/service/fretboard/src/test/java/mozilla/components/service/fretboard/source/kinto/HttpURLConnectionHttpClientTest.kt
+++ b/components/service/fretboard/src/test/java/mozilla/components/service/fretboard/source/kinto/HttpURLConnectionHttpClientTest.kt
@@ -11,7 +11,10 @@ import okhttp3.mockwebserver.RecordedRequest
 import org.junit.Assert.assertEquals
 import org.junit.Test
 import org.junit.runner.RunWith
+import org.mockito.Mockito.`when`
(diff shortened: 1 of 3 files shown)


components/service/fretboard/src/test/java/mozilla/components/service/fretboard/source/kinto/HttpURLConnectionHttpClientTest.kt
```diff
@@ -27,14 +30,46 @@ class HttpURLConnectionHttpClientTest {
...
+    }
+
+    @Test
+    fun testGET200() {
+        testGETResponseCode(200)
+    }
+
+    @Test
+    fun testGET299() {
+        testGETResponseCode(299)
```
pocmo: `testGET200` and `testGET299` do not perform any assertions or verifications.

fercarcedo: Now it's corrected (when I wrote them I thought about the test passing if the method didn't throw any exceptions), now they have assertions. 

Turns out `testGET200` was redundant since `sinceTestGETWithoutHeaders` was testing the same situation, and I changed `testGET299` to perform assertions.

Issue #517, opened 2018-07-25
Switch to published .aar for JNA


csadilek:
There is in fact a published .aar that contains libjnidispatch for various supported architectures. Not sure how I missed that 🙈.

So, we can remove our Ivy config for fetching it and replace it with e.g:
`implementation 'net.java.dev.jna:jna:4.5.1@aar'`

ncalexan:
I'm trying to integrate a sample application for the Rust logins API component we're building over in https://github.com/mozilla/application-services.  That app depends on the logins API and the FxA client library, and both of those depend on JNA.  I'm getting duplicate dex errors, like:
```
D8: Program type already present: com.sun.jna.CallbackParameterContext
```
I'm hoping that correctly labeling the Android dependency will allow the regular Maven/Gradle dependency resolution to determine that the underlying JNA classes.jar needs to be dexed only once.  Let us hope!

ncalexan:
While I'm here, I'll link to Mentat's JNA configuration, which allows for unit testing with Robolectric and running on Android devices.  See https://github.com/mozilla/mentat/blob/f8478835a2052a0857f1052105676747fdd2cc6e/sdks/android/Mentat/library/build.gradle#L71-L92.

pocmo:
#583 -> Follow-up?

PR #568, opened 2018-08-07
Issue #517 - Switch to published AAR for JNA.

ncalexan:
This allows a consuming application to use multiple libraries that
transitively depend on JNA (@aar type).

Tested by:

- bumping the android-components version
- `./gradlew :service-firefox-accounts:install` to install bumped version locally
- adding `repository { mavenLocal() }` to consuming application
- requiring bumped version in consuming application

PR #575, opened 2018-08-09
Issue #517 - Switch to published AAR for JNA.

ncalexan:
This allows a consuming application to use multiple libraries that
transitively depend on JNA (@aar type).

Tested by:

- bumping the android-components version
- `./gradlew :service-firefox-accounts:install` to install bumped version locally
- adding `repository { mavenLocal() }` to consuming application
- requiring bumped version in consuming application

This special `jnaForTest` construction is not necessary when invoking
tests with Gradle; it is necessary when invoking tests with Android
Studio.  This is lifted from Mentat, where I initially worked out
these details -- see
https://github.com/mozilla/mentat/blob/f8478835a2052a0857f1052105676747fdd2cc6e/sdks/android/Mentat/library/build.gradle#L71-L89.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,7 @@ buildscript {
         coroutines: '0.23.4',
         supportLibraries: '27.1.1',
         fxa: '0.2.1',
-        jna: '4.5.1',
+        jna: '4.5.2',
         // Test only:
         junit: '4.12',
         robolectric: '3.8',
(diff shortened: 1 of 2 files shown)


ncalexan:
It was suggested that sputnik was timing out in #568 because I didn't push to the `mozilla/android-components` repo.  @pocmo has already reviewed this, but let's see if we can get a sputnik success so that I can, after Sebastian marks it, land the thing.

PR #582, opened 2018-08-10
Issue #576: Fretboard: Log ExperimentDownloadException

### LINKED ISSUE #576: Fretboard: Log ExperimentDownloadException
I think it would be interesting to log ExperimentDownloadException here:
https://github.com/mozilla-mobile/android-components/blob/master/components/service/fretboard/src/main/java/mozilla/components/service/fretboard/Fretboard.kt#L54

I recently landed support-base which is our new base component containing building blocks for components. One helper is a Logger that can be used by components and allows the app to control what gets logged and how:
https://github.com/mozilla-mobile/android-components/blob/master/components/support/base/README.md

Here's a (Java) example:
https://github.com/mozilla-mobile/android-components/blob/master/components/service/telemetry/src/main/java/org/mozilla/telemetry/net/HttpURLConnectionTelemetryClient.java#L28

fercarcedo:
Closes #576

### DIFF
--- a/components/service/fretboard/build.gradle
+++ b/components/service/fretboard/build.gradle
@@ -23,6 +23,7 @@ android {
 
 dependencies {
     implementation project(':support-ktx')
+    implementation project(':support-base')
     implementation "org.jetbrains.kotlin:kotlin-stdlib:${rootProject.ext.dependencies['kotlin']}"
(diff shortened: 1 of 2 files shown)


fercarcedo:
If I understood everything correctly, then the client app must call `Log.addSink` in order to make the logs visible, because otherwise they don't get shown on logcat. Is that correct?

PR #606, opened 2018-08-10
Download statically linked fxa library

eoger:
Following https://github.com/mozilla/application-services/issues/172, `openssl` is now baked into `fxa` and we don't generate the dependencies zip file anymore.


eoger:
Yup I need to make a PR to switch to 0.2.2

Issue #604, opened 2018-08-10
Engine: Add ability to stop loading


pocmo:
Something like `EngineSession.stopLoading()`.

PR #611, opened 2018-08-14
Issue #604: Add Engine ability to stop loading

jonalmeida:
Adds the ability to stop loading a session that would have already started.

Some considerations were made about whether `EngineSession.Observer#onLoadingStateChanged` should be called. In our current System Engine and Gecko Engine there isn't any explicit callback for handling a stop loading event so there wouldn't be any way to do this.


jonalmeida:
\o/

PR #613, opened 2018-08-14
Severin/fix keyboard desync

Baron-Severin:
This fixes a bug we're seeing on Fire TV, where certain actions will cause our EditText to desynchronize from the soft keyboard.


Baron-Severin:
@csadilek Thanks for looking over the code.

- Filed issue #619 
- Amended commit to include issue #
- I don't think this is reproducible in a unit test
- Added unit test to verify that `post` is called from `deleteSurroundingText`
- Verified that Focus is unaffected

Let me know if there's anything else you would like me to take care of!

csadilek:
OK, thanks @Baron-Severin.

PR #616, opened 2018-08-15
Closes #615: Add debug mode to telemetry.

### LINKED ISSUE #615: Add debug mode to telemetry
It'll be great if we can see the JSON content during local development

cnevinc:
I'm not sure if there's already such a feature, but it'll be great if we can see the output JSON during development.

### DIFF
--- a/components/service/telemetry/src/main/java/org/mozilla/telemetry/storage/FileTelemetryStorage.java
+++ b/components/service/telemetry/src/main/java/org/mozilla/telemetry/storage/FileTelemetryStorage.java
@@ -99,6 +99,10 @@ public class FileTelemetryStorage implements TelemetryStorage {
 
         final String serializedPing = serializer.serialize(ping);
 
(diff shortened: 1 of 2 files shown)


components/service/telemetry/src/main/java/org/mozilla/telemetry/storage/FileTelemetryStorage.java
```diff
@@ -99,6 +99,10 @@ private void storePing(TelemetryPing ping) {
 
         final String serializedPing = serializer.serialize(ping);
 
+        if (configuration.isDebugMode()) {
+            logger.warn(serializedPing, null);
```
csadilek: We should use logger.debug here. It's not a warning, just debugging information :).

csadilek: We should use logger.debug here. It's not a warning, just debugging information :).

PR #623, opened 2018-08-16
Add loadData to Engine

jonalmeida:
Adds loadData support to the Engine.

Defaults system WebView encoding to `UTF-8` (default for Android) and mimeType defaults to `text/html` for parity as well.


components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
```diff
@@ -65,6 +66,11 @@ abstract class EngineSession(
      */
     abstract fun loadUrl(url: String)
 
+    /**
+     * Loads the data with the given mimeType.
+     */
```
csadilek: Let's add `@param` here to document the parameters.

components/browser/session/src/test/java/mozilla/components/browser/session/engine/EngineObserverTest.kt
```diff
@@ -62,6 +68,10 @@ class EngineObserverTest {
                 return emptyMap()
             }
 
+            override fun loadData(data: String, mimeType: String) {
+                // TODO: What do we do here?
```
csadilek: If it's not needed in the test, we can just leave it empty. Did you want to keep that comment?

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
```diff
@@ -15,6 +15,7 @@ import java.lang.ref.WeakReference
  */
 class SystemEngineSession : EngineSession() {
     internal var urlToLoad: String? = null
```
csadilek: nit: finding the naming a bit confusing now `urlToLoad` isn't necessarily a URL. Should we rename it? We could also create a data class ScheduledLoad or similar which takes a url, data, and mime type?

csadilek:
@jonalmeida this is another case of a network issue. I will restart the build.

components/browser/session/src/test/java/mozilla/components/browser/session/engine/EngineObserverTest.kt
jonalmeida: This was my second question about testing but I figured it out; forgot to remove this todo. :)

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
jonalmeida: I wasn't too keen on my original implementation as well since it reduced the test coverage percentage. The data class sounds like a better idea!

jonalmeida:
After putting up this PR, I realized too late that I missed a couple of things: add SessionUseCase, identify and handling base64 encoded data.

We should be using the GeckoEngine's `loadString` everywhere, except for base64 encoded data that needs to use `loadData`.

We can match what the [`SystemWebView`][1] is doing:
>  The encoding parameter specifies whether the data is base64 or URL encoded. If the data is base64 encoded, the value of the encoding parameter must be 'base64'. 

[1]: https://developer.android.com/reference/android/webkit/WebView#loadData(java.lang.String,%20java.lang.String,%20java.lang.String)

components/browser/engine-system/src/test/java/mozilla/components/browser/engine/system/SystemEngineSessionTest.kt
```diff
@@ -52,6 +53,26 @@ class SystemEngineSessionTest {
...
+        val engineSession = spy(SystemEngineSession())
+        val webView = mock(WebView::class.java)
+
+        engineSession.loadData("<html><body>Hello!</body></html>")
+        verify(webView, never()).loadData(anyString(), eq("text/html"), eq("UTF-8"))
+
+        `when`(engineSession.currentView()).thenReturn(webView)
+
+        engineSession.loadData("<html><body>Hello!</body></html>")
+        verify(webView).loadData(anyString(), eq("text/html"), eq("UTF-8"))
```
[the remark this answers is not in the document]
jonalmeida: We can. I was mainly verifying the mimeType and encoding. I'll fix this though.

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
```diff
@@ -14,8 +14,8 @@ import java.lang.ref.WeakReference
  * WebView-based EngineSession implementation.
  */
 class SystemEngineSession : EngineSession() {
-    internal var urlToLoad: String? = null
     internal var view: WeakReference<SystemEngineView>? = null
+    internal var scheduledLoad = ScheduledLoad()
```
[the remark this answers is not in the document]
jonalmeida: I put it as an empty constructed data class to avoid the extra `?` in it's usage which I think would decrease our code coverage:
```kotlin
internalSession.scheduledLoad?.data?.let {
  //...
}
```

I've changed the assignment to make it more explicit:
```kotlin
internal var scheduledLoad = ScheduledLoad(null)
```

PR #628, opened 2018-08-17
Set version to 0.19.1

### LINKED ISSUE #622: `ShiftDrawable$1.onAnimationUpdate`, `performTraversals` uses CPU during idle
Running samples-toolbar, using the Android CPU profiler in instrumented mode, the flame chart shows `onAnimationUpdate` uses a lot of CPU:
screen shot 2018-08-15 at 16 33 58

It uses ~15% of the CPU used during this time which seems like a lot given there are no animations occurring. Other large CPU use during this time is `performTraversals`, which this might be causing (or maybe vice versa!).

csadilek:
Setting the version for our mini sprint to fix #622 .

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,7 +26,7 @@ buildscript {
 
     // Synchronized library configuration for all modules
     ext.library = [
-        version: '0.19'
+        version: '0.19.1'
     ]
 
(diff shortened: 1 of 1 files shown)


pocmo:
Oh, the label was removed! :)

Issue #508, opened 2018-07-25
Implement tracking protection in SystemEngine


csadilek:
Focus provides support for tracking protecting in WebView: https://github.com/mozilla-mobile/focus-android/blob/master/app/src/webview/java/org/mozilla/focus/webview/TrackingProtectionWebViewClient.java

Our fix for #352 did not include this.

csadilek:
This will be ready today, but moving to the next sprint so we''ll have time to review.

PR #630, opened 2018-08-17
Issue #508: Implement tracking protection in SystemEngine

csadilek:
The meat of the tracking protection logic in Focus is in the URL matching and black/whitelisting logic. I was considering making this an application-level concern, but I think, by default, our components should support tracking protection i.e. if `enableTrackingProtection` is called it should just work without the App having to provide the URL matching functionality.

So, I ported all this logic to Kotlin, refactored it and wrote tests, then added `enable|disableTrackingProtection` to our `SystemEngine`.


csadilek:
@pocmo OK, this is updated now to move the matcher functionality to the system engine.

components/browser/engine-system/build.gradle
```diff
@@ -32,6 +32,7 @@ dependencies {
     implementation project(':support-utils')
 
     implementation "org.jetbrains.kotlin:kotlin-stdlib:${rootProject.ext.dependencies['kotlin']}"
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:${rootProject.ext.dependencies['coroutines']}"
```
pocmo: I need to update this before #636 lands (or the other way around).

csadilek: Ah OK, I will do it. I will merge #636 and update this.

csadilek:
OK, updated!

Issue #631, opened 2018-08-20
ui-progress: High CPU usage


pocmo:
Follow-up from #622 

* AnimatedProgressBar uses about 15% of CPU when idle
* The animator in ShiftDrawable doesn't seem to stop
* The animator is starting by default even if AnimatedProgressBar is not visible

PR #633, opened 2018-08-20
Issue #631: Reduce CPU usage of ui-progress when idling.

pocmo:
This fixes the high CPU usage when idling. We should still merge #632. However we could land this one in 0.19.1 too because it will help Focus (and potentially Rocket if they want/can switch to the component version of the progress bar).


pocmo:
Some unit tests for the animator start/end calls would be good here. But that's something for 0.20.

PR #642, opened 2018-08-21
Use Android SDK 28

pocmo:
Work in progress.

* [x] Support libraries not final yet: ~28.0.0-rc01~ ~28.0.0-rc02~
* [ ] Robolectric 4.0 (with support for API 28) not final yet: 4.0-alpha-3
* [x] The new support libraries added a bunch of nullable annotations and we need to update some Kotlin code to deal with potential nulls now. I did that for the components code already. But there's still some test code left that needs to be updated.

### DIFF
--- a/components/ui/autocomplete/src/main/java/mozilla/components/ui/autocomplete/InlineAutocompleteEditText.kt
+++ b/components/ui/autocomplete/src/main/java/mozilla/components/ui/autocomplete/InlineAutocompleteEditText.kt
@@ -235,6 +235,10 @@ open class InlineAutocompleteEditText @JvmOverloads constructor(
         resetAutocompleteState()
     }
 
+    override fun getText(): Editable {
+        return super.getText() as Editable
+    }
+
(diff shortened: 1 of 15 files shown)


pocmo:
Also fixed the outstanding tests. Everything should pass again.

components/ui/autocomplete/src/main/java/mozilla/components/ui/autocomplete/InlineAutocompleteEditText.kt
```diff
@@ -118,10 +118,10 @@ open class InlineAutocompleteEditText @JvmOverloads constructor(
     private var discardAutoCompleteResult: Boolean = false
 
     val nonAutocompleteText: String
-        get() = getNonAutocompleteText(text)
+        get() = getNonAutocompleteText(text!!)
```
csadilek: What do you think about adding an override for `getText`? Then, at least, we don't need to add those `!!` everywhere. I understand text can only be null for API < 28 anyway.

```
override fun getText(): Editable {
        return super.getText() as Editable
}
```

Issue #521, opened 2018-07-26
Add support for modifying WebView settings from EngineSession


carolkng:
Currently, the WebView inside the EngineSession is internal, so external applications can't modify the settings on it. I'm not sure how this is implemented in GeckoView.

The Firefox Accounts login page requires Javascript and DOM storage to be enabled, so in order to incorporate EngineView into the component, would it be possible to open up some public methods for those settings?

carolkng:
(At the same time, it would probably be a huge hassle to write those public methods for _all_ the settings, since I'm not familiar with how similar the GeckoView and WebView settings API are)

pocmo:
Is the goal to replace `WebView` with the/an abstract `engine` component?

carolkng:
Yes, the idea is to be able to give an example of how to spawn a configurable Webview/Geckoview with the login page through the engine componeny

pocmo:
Ah, got it! I'll move this into our next sprint and then we can see!

PR #645, opened 2018-08-21
Issue #521: Add support for modifying engine session settings

csadilek:
Only added the settings we wanted to use so far. We'd have to add specific settings as needed. They're all implemented differently in WebView and GeckoView.

Some thoughts:
Do we need this abstraction of `Settings`? I think we need *an* abstraction as consumers and other components interact with our abstract `EngineSession` and we shouldn't leak internals of concrete implementations. Alternatively, we could simply use a `Map<String, Any>` for settings and expose it, but then we'd need to document all possible keys and map them back to the individual implementations. So, decided for this approach instead...

### DIFF
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -7,6 +7,7 @@ package mozilla.components.browser.engine.gecko
 import kotlinx.coroutines.experimental.CompletableDeferred
 import kotlinx.coroutines.experimental.runBlocking
 import mozilla.components.concept.engine.EngineSession
+import mozilla.components.concept.engine.Settings
 import org.mozilla.gecko.util.ThreadUtils
 import org.mozilla.geckoview.GeckoRuntime
 import org.mozilla.geckoview.GeckoSession
@@ -18,12 +19,11 @@ import org.mozilla.geckoview.GeckoSessionSettings
  */
 @Suppress("TooManyFunctions")
 class GeckoEngineSession(
-    private val runtime: GeckoRuntime,
+    runtime: GeckoRuntime,
     privateMode: Boolean = false
 ) : EngineSession() {
 
     internal var geckoSession = GeckoSession()
-
     private var initialLoad = true
 
     init {
@@ -120,6 +120,29 @@ class GeckoEngineSession(
     }
 
     /**
+     * See [EngineSession.enableTrackingProtection]
+     */
+    override fun enableTrackingProtection(policy: TrackingProtectionPolicy) {
+        geckoSession.settings.setBoolean(GeckoSessionSettings.USE_TRACKING_PROTECTION, true)
+        notifyObservers { onTrackerBlockingEnabledChange(true) }
+    }
+
+    /**
+     * See [EngineSession.disableTrackingProtection]
+     */
+    override fun disableTrackingProtection() {
+        geckoSession.settings.setBoolean(GeckoSessionSettings.USE_TRACKING_PROTECTION, false)
+        notifyObservers { onTrackerBlockingEnabledChange(false) }
+    }
+
+    /**
+     * See [EngineSession.settings]
+     */
+    override val settings: Settings
+        get() = throw UnsupportedOperationException("""Not supported by this implementation:
+            Use Engine.settings instead""".trimIndent())
+
+    /**
      * NavigationDelegate implementation for forwarding callbacks to observers of the session.
      */
     private fun createNavigationDelegate() = object : GeckoSession.NavigationDelegate {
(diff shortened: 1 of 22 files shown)


components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -33,6 +36,29 @@ class GeckoEngineSession(
...
 
+    /**
+     * See [EngineSession.settings]
+     */
+    override val settings: Settings = object : Settings, PrefsHelper.PrefHandlerBase() {
+        val prefs = mutableMapOf<String, Any>()
+        init { PrefsHelper.getPref(DOM_STORAGE_PREF, this) }
+
+        override var javascriptEnabled: Boolean
+            get() = runtime.settings.javaScriptEnabled
```
pocmo: Oh, JavaScript is a runtime setting? GeckoView distinguishes between Runtime and Session settings. Should we do that too?

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -33,6 +36,29 @@ class GeckoEngineSession(
...
+            set(value) { runtime.settings.javaScriptEnabled = value }
+
+        override var domStorageEnabled: Boolean
+            get() = prefs[DOM_STORAGE_PREF] as? Boolean ?: true
+            set(value) {
+                PrefsHelper.setPref(DOM_STORAGE_PREF, value)
+                prefs[DOM_STORAGE_PREF] = value
+            }
```
pocmo: Oh, I didn't know that `PrefsHelper` is exposed. For other things we can use `GeckoSessionSettings` and `GeckoRuntimeSettings`. 

@snorp Is using PrefsHelper okay?

csadilek: @snorp Heh, it *is* part of the public API. Otherwise I wouldn't be able to use it :). But, point taken, I will remove this for now. The type shouldn't be public then, I guess.

snorp: It's difficult to actually give it `package` protection currently because of how the code is organized. We'll get to it some day, I hope.

Only stuff in org.mozilla.geckoview should be considered public, with a couple caveats.

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
pocmo: Alright!

csadilek:
First update is in which introduces a property delegate for unsupported settings. A setting is unsupported unless a concrete engine provides an implementation. `domStorageEnabled` is now unsupported for GV. We didn't have a use case for it anyway in GV (only for WebView where we had to turn it on). So, this part seems fine.

csadilek:
Wow, custom delegates are not good for code coverage. Seem to cause a bunch of byte code generation.

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -33,6 +31,15 @@ class GeckoEngineSession(
         geckoSession.trackingProtectionDelegate = createTrackingProtectionDelegate()
     }
 
+    /**
+     * See [EngineSession.settings]
+     */
+    override val settings: Settings = object : Settings() {
+        override var javascriptEnabled: Boolean
+            get() = runtime.settings.javaScriptEnabled
+            set(value) { runtime.settings.javaScriptEnabled = value }
```
pocmo: I'm still unsure about modifying a runtime setting from a Session. The API and doc make it look like we are enabling/disabling JavaScript just for this session when in fact we are enabling/disabling for all of them.

csadilek: Agreed. This shouldn't be merged as is. The problem is that for GeckoView this is really an engine setting, but for our system engine, it's a session or view setting. We'd want to stay on the side of GV, but if we move this to engine that it won't work at all for WebView which is where this requirement came from. I have thought of some options we should discuss...

pocmo: @csadilek Couldn't we apply the "SystemEngineRuntimeSetting" to every WebView that we create and emulate the same behavior?

@snorp Is JavaScript really only a runtime option in Gecko or could we also expose this as a session-scoped setting?

csadilek: @pocmo yes that's an update I am working on.   Introduce engine settings and pass along to the session (and view). Then the implementations can decide how to handle it.

csadilek:
Updated to introduce both engine and session settings so that implementations can decide how to implement a specific setting (e.g. javascriptEnabled), and throw exceptions if unsupported.

Issue #499, opened 2018-07-25
Engine: Support private browsing mode


pocmo:
In GeckoView it's just setting a flag. In WebView it's 🔥.

PR #650, opened 2018-08-22
Issue #499: Support private browsing mode

csadilek:
Kept this independent of the settings work after all. Whether or not a session is private should be a direct property of the session rather than a setting, I think.


csadilek:
@pocmo It does already have a default for `createSession` on the interface: https://github.com/mozilla-mobile/android-components/pull/650/files#diff-20a2d5d389d053ecea72232700d2e678R32 or did you mean somewhere else?

pocmo:
@csadilek Oops. Somehow I missed that. This is perfect then. Non-modifiable "settings" as constructor parameters with default values is super nice. :)

Issue #476, opened 2018-07-23
Make long presses on web content observable/consumable


pocmo:
Part of #200.

pocmo:
I think that a long press should be a `Consumable` in Session like `Download`. We could create some class like `LongClick` that contains meta data about what was clicked. Observers can then handle this click and consume the object.

For Focus-parity we need at least the functionality of `HitTarget`:
https://github.com/mozilla-mobile/focus-android/blob/master/app/src/main/java/org/mozilla/focus/web/IWebView.java#L15-L36

We can probably come up with a nicer API than just a bunch of booleans and Strings. However be aware that a `HitTarget` can be multiple things: It can be a link and an image at the same time.

Additionally we can look at the APIs of GeckoView and WebView and see if there's more we can implement or whether one of them (hopefully GeckoView) implements a nicer API that we can mirror.

jonalmeida:
After some investigation I think we can support most of the [HitTestResults][3] that WebView does and do some simple checking in a-c for a few more types:
```kotlin
class HitResult(type: Int) {
    companion object {
        const val NONE_TYPE = 0
        const val IMAGE_TYPE = 1
        const val VIDEO_TYPE = 2
        const val AUDIO_TYPE = 3
        const val SRC_TYPE = 4
        const val IMAGE_SRC_TYPE = 5 
        const val PHONE_TYPE = 6 // URI prepended with 'tel:'
        const val EMAIL_TYPE = 7 // URI prepended with 'mailto:'
        const val GEO_TYPE = 8 // URI prepended with 'geo:'
    }
}
```
I can't find any other multiple types other than image + link, so having a separate `IMAGE_SRC_TYPE` for it seems fair.

For our Gecko implementation, we're given the [basic element types][1] (e.g. img, video, audio), but within the components, we can support more by checking the URI values (e.g. [geo][2]).

In System WebView, we just forward those on.

[1]: https://dxr.mozilla.org/mozilla-central/source/mobile/android/geckoview/src/main/java/org/mozilla/geckoview/GeckoSession.java?q=GeckoSession.java&redirect_type=direct#2039-2043
[2]: https://tools.ietf.org/html/rfc5870
[3]: https://developer.android.com/reference/android/webkit/WebView.HitTestResult

pocmo:
@jonalmeida Sounds great! :)

> WebView also supports long presses on edit texts but I'm not sure what use that is, nor does Gecko provide us with enough information for us to support that type.

[...]

Issue #648, opened 2018-08-22
Convert Strings to URIs at the edge of our apps


severinrudie:
#### Overview
We're working with URIs as strings across a lot of our apps.  Standardizing all of our repos to use one URI implementation, and then parsing strings at the edges of our apps and internally using this URI type would yield benefits.

#### Benefits
- Stricter type safety
- Function signature expressiveness
- Less repetition of parsing and error handling

#### Potential Class Candidates
- android.net.Uri
- java.net.URI - unintuitive interface
- java.net.URL - lots of landmines in the implementation (e.g. #equals makes a blocking network call)
- okhttp3.HttpUrl - doesn't cover our use cases, and library is not used by all of our repos
- Potentially more options?

Internal conversations have pointed to android.net.Uri being the best fit, for the above reasons.

#### Proposed Next Steps
- Finalize decision on which class to standardize around
- Identify files where this change could be made in small steps

pocmo:
Another interesting data point may be all our UrlUtils classes, here for example the one in Focus:
https://github.com/mozilla-mobile/focus-android/blob/master/app/src/main/java/org/mozilla/focus/utils/UrlUtils.java#L23

It would be nice if we could get rid of that and offer this functionality either in a wrapper or as an extension function on what we'd pick.

pocmo:
WebView internally uses `Uri` too, like here in WebResourceRequest:
https://developer.android.com/reference/android/webkit/WebResourceRequest

Baron-Severin:
The more I look at this the more I think `Uri` with some extension functions is the way to go.  I don't think we have enough custom behavior to even make a wrapper worth the effort.

NotWoods:
Reviving this conversation. If we want to converge on `Uri` let's ask contributors to help convert internal functions to use it instead of `String`. Should I open an issue for this?

pocmo:
Looking through old issues and closing them if they are not applicable anymore.

This is still interesting, but not something we will look at anytime soon, I think.

PR #659, opened 2018-08-24
Issue #476: Add long press support as observable

jonalmeida:


jonalmeida:
My take away from this PR is: don't try to mock static inner classes with PowerMock via Robolectric.

jonalmeida:
I need to add more tests.

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -224,6 +229,69 @@ class GeckoEngineSession(
...
+                    elementSrc?.let {
+                        when {
+                            it.isPhone() -> notifyObservers { onLongPress(HitResult.PHONE(it)) }
+                            it.isEmail() -> notifyObservers { onLongPress(HitResult.EMAIL(it)) }
+                            it.isGeoLocation() -> notifyObservers { onLongPress(HitResult.GEO(it)) }
+                            else -> notifyObservers { onLongPress(HitResult.UNKNOWN(it)) }
+                        }
+                    } ?: notifyObservers { onLongPress(HitResult.UNKNOWN("")) }
+                }
+            }
```
pocmo: It looks like all branches call `notifyObservers { onLongPress(...) }` so I wonder if you could make the `when` statement return the HitResult and then call notifyObservers ... with that.

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -224,6 +229,69 @@ class GeckoEngineSession(
...
+            uri: String?,
+            elementType: Int,
+            elementSrc: String?
+        ) {
+            when (elementType) {
+                ELEMENT_TYPE_AUDIO ->
+                    elementSrc?.let {
+                        notifyObservers { onLongPress(HitResult.AUDIO(it)) }
+                    }
+                GeckoSession.ContentDelegate.ELEMENT_TYPE_VIDEO ->
```
csadilek: nit: formatting: ELEMENT_TYPE_VIDEO is fully qualified, the others aren't.

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -236,4 +304,10 @@ class GeckoEngineSession(
         internal const val MOZ_NULL_PRINCIPAL = "moz-nullprincipal:"
         internal const val ABOUT_BLANK = "about:blank"
     }
+
+    fun String.isPhone(): Boolean = contains("tel:", true)
```
csadilek: Nice, let's move these to our support-ktx module as standard string extensions? Seems like they could come in handy.

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -147,6 +162,51 @@ class SystemEngineView @JvmOverloads constructor(
         }
     }
 
+    internal fun handleLongClick(type: Int, extra: String): Boolean {
+        return when (type) {
+            EMAIL_TYPE -> {
```
csadilek: nit: We could simplify this by creating a `mapOf(EMAIL_TYPE to HitResult.EMAIL, GEO_TYPE to HitResult.GEO.....)`. Then we can handle all these cases at once with a map lookup.

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
jonalmeida: I kept in here so that it wouldn't show up for all Strings outside this class. I don't think it makes sense for all strings to have an `isPhone`. I know extension functions don't add any overhead to the class since they're static methods really, but it seems misleading.

When we have a URI class (see #648) these exts should go there.

csadilek: Right, that's fair. I just wasn't sure if having a String ext function for single use makes sense...

csadilek: OTOH, our String extension also has an `isUrl` function. So, `isPhone` etc. would seem to fit there. The Strings wouldn't have a `isPhone` function unless `mozilla.components.support.ktx.kotlin.isPhone` is imported. So, it wouldn't show up for Strings by default. Anyways, don't feel strongly here, just a suggestion :).

jonalmeida: > The Strings wouldn't have a isPhone function unless `mozilla.components.support.ktx.kotlin.isPhone` is imported.

Oh, that's a good point! I'll move it there then.

jonalmeida:
Squashed! I'll merge it when the CI build is complete.

Issue #662, opened 2018-08-24
Provide API to set session defaults


csadilek:
Currently there's no way for an application to specify default session characteristics e.g. that all sessions should be private and have tracking protection enabled.

We need to look into building out an API for this.

pocmo:
While working on the `browser-engine` integration into Firefox for Fire TV I noticed that they have a quite complex settings setup for every WebView:
https://github.com/mozilla-mobile/firefox-tv/blob/master/app/src/amazonWebview/java/org/mozilla/focus/iwebview/WebViewProvider.kt#L63-L100

This may be something for a follow-up. I just wanted to bring it up here in case your implement something that configures engine (views) too.

PR #674, opened 2018-08-28
Issue #662: Provide API to set engine/session defaults

csadilek:
In #521, we introduced both engine and engine session settings and handled the differences between GeckoView and WebView settings. When providing default settings, we can hide this complexity from the consumer. An API user should simply be able to pass default settings to the engine and not worry about whether these are handled by the engine or session implementation, as these settings should be applied to all sessions anyway (by default).

So, our engine now accepts optional default settings, which will be forwarded to the engine sessions, as needed. If you wanted to build a browser that always has tracking protection on:

```
val engine: Engine by lazy {
   val runtime = GeckoRuntime.getDefault(applicationContext)
   val settings = DefaultSettings(trackingProtectionPolicy=TrackingProtectionPolicy.all())
   GeckoEngine(runtime, settings)
}
```

or

```
val engine: Engine by lazy {
   SystemEngine(DefaultSettings(trackingProtectionPolicy=TrackingProtectionPolicy.all()))
}
```

Notes:
- This also brings back setting tracking protection categories, which we lost in a recent update. It's also dealt with correctly now (as a runtime setting, not session setting)
- Why are browser sessions not affected? We handle browser sessions directly in use cases e.g. there's a `AddNewPrivateTabUseCase` which sets the corresponding properties on the browser session. It seems to me we only need this API for engine sessions. We can file a follow-up if we need it though.

### DIFF
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -28,13 +28,15 @@ import org.mozilla.geckoview.GeckoSessionSettings
 @Suppress("TooManyFunctions")
 class GeckoEngineSession(
     runtime: GeckoRuntime,
-    privateMode: Boolean = false
+    privateMode: Boolean = false,
+    defaultSettings: Settings? = null
 ) : EngineSession() {
 
     internal var geckoSession = GeckoSession()
     private var initialLoad = true
 
     init {
+        defaultSettings?.trackingProtectionPolicy?.let { enableTrackingProtection(it) }
         geckoSession.settings.setBoolean(GeckoSessionSettings.USE_PRIVATE_MODE, privateMode)
         geckoSession.open(runtime)
 
--- a/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
+++ b/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
@@ -65,6 +65,17 @@ abstract class EngineSession(
             fun select(vararg categories: Int): TrackingProtectionPolicy =
                 TrackingProtectionPolicy(categories.sum())
         }
+
+        override fun equals(other: Any?): Boolean {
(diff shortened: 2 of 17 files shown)


components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -132,6 +134,10 @@ class GeckoEngineSession(
      * See [EngineSession.enableTrackingProtection]
      */
     override fun enableTrackingProtection(policy: TrackingProtectionPolicy) {
+        if (policy == TrackingProtectionPolicy.none()) {
+            return
+        }
```
pocmo: Is this something we strictly need?

I wonder how that would work in Focus where we can have tracking protection on but all categories disabled.

components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
```diff
@@ -65,6 +65,18 @@ abstract class EngineSession(
             fun select(vararg categories: Int): TrackingProtectionPolicy =
                 TrackingProtectionPolicy(categories.sum())
         }
+
+        override fun equals(other: Any?): Boolean {
+            if (this === other) return true
+            if (javaClass != other?.javaClass) return false
```
pocmo: I think if you use the following then Kotlin will smart cast and you do not need to do that manually:

```Kotlin
if (other !is TrackingProtectionPolicy) return false
```

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
csadilek: Not strictly, no. I removed it.

components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
csadilek: 👍 this was auto-generated by AS. Good find.

Issue #673, opened 2018-08-28
Engine: Request interception and content injection


pocmo:
#### WebView

WebViewClient.shouldInterceptRequest
https://developer.android.com/reference/android/webkit/WebViewClient#shouldInterceptRequest(android.webkit.WebView,%20android.webkit.WebResourceRequest)

#### GeckoView

GeckoSession.NavigationDelegate.onLoadRequest

### Implementations

Firefox for Fire TV intercepts `firefox:home` and injects a blank page:
https://github.com/mozilla-mobile/firefox-tv/blob/c660e1938eb0f112379e38c8af8a3f892ed9ba1f/app/src/amazonWebview/java/org/mozilla/focus/webview/FocusWebViewClient.java#L117-L127

Firefox for Fire TV injects a custom page for firefox:about URLs:
https://github.com/mozilla-mobile/firefox-tv/blob/c660e1938eb0f112379e38c8af8a3f892ed9ba1f/app/src/main/java/org/mozilla/focus/browser/LocalizedContent.kt#L24

Focus intercepts internal URLs (e.g. `focus:about`) and injects generated content:
https://github.com/mozilla-mobile/focus-android/blob/master/app/src/main/java/org/mozilla/focus/web/GeckoWebViewProvider.kt#L410-L414

pocmo:
Moving this into the 0.21 sprint and assigning myself: This is needed for Firefox for Fire TV for various internal pages and I would like to get this in for the 0.21 release.

PR #680, opened 2018-08-29
Issue #673: Engine: Allow consumers to intercept requests and inject content.

pocmo:
@csadilek This is not done (I need to add KDoc and tests..). But I would like to know if the proposed Settings API looks good to you?

### DIFF
--- a/components/browser/engine-gecko-beta/build.gradle
+++ b/components/browser/engine-gecko-beta/build.gradle
@@ -37,6 +37,8 @@ dependencies {
     testImplementation Deps.testing_junit
     testImplementation Deps.testing_robolectric
     testImplementation Deps.testing_mockito
+
+    testImplementation project(':support-test')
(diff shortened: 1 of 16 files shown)


pocmo:
> The only downside is that now when you call GeckoEngineSession.settings.javascriptEnabled you no longer get an exception telling you to use the engine instead.

But I'd still get the "unsupported" exception from the interface's default implementation, right?

pocmo:
@csadilek Okay, I think this is now ready for review. :)

pocmo:
> decrease coverage by 0.16%.

Let's see if I can improve that.

csadilek:
> But I'd still get the "unsupported" exception from the interface's default implementation, right?

Yes, exactly.

pocmo:
> decrease coverage by 0.06%.

That's better.

PR #687, opened 2018-08-31
Use jacoco 0.8.2 for code coverage reports.

pocmo:
Jacoco 0.8.2 was released. Let's see how this affects code coverage. According to the release notes a bunch of Kotlin improvements landed:
https://github.com/jacoco/jacoco/releases

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -89,6 +89,10 @@ subprojects {
 
     afterEvaluate {
         if (it.hasProperty('android')) {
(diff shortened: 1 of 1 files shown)


pocmo:
> decrease coverage by 0.09%

That's unfortunate. But at least low. Looking at the report somehow all the companion objects got flagged. 

@csadilek What do you think? Should we get this in?

PR #694, opened 2018-08-31
Closes #657: Forward lifecycle events to EngineView

### LINKED ISSUE #657: Forward life cycle events to (System)EngineView
WebView needs to know about lifecycle events to know when to pause certain things.

As an example see code of the deprecated `WebViewFragment`:
https://android.googlesource.com/platform/frameworks/base/+/refs/heads/master/core/java/android/webkit/WebViewFragment.java

Our `WebView` is currently wrapped by `SystemEngineView`. Either we allow forwarding those events to `EngineView`s in general. Or we just add this to `SystemEngineView`.

csadilek:
Some thoughts / open questions:

- Focus calls `setActive` on the GeckoSession `onPause/onResume`, not `onStart/onStop`. According to the Javadoc in GV we should only call `setActive(false)` when not visible. So, this implementation seems correct.

- Should we save state by default in `onPause` or let the app decide / handle that?

### DIFF
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineView.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineView.kt
@@ -52,4 +52,12 @@ class GeckoEngineView @JvmOverloads constructor(
             currentGeckoView.session = internalSession.geckoSession
         }
     }
+
+    override fun onStart() {
+        currentGeckoView.session.setActive(true)
+    }
+
+    override fun onStop() {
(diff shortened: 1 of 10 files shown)


pocmo:
> Focus calls setActive on the GeckoSession onPause/onResume, not onStart/onStop. According to the Javadoc in GV we should only call setActive(false) when not visible. So, this implementation seems correct.

Agreed. This looks correct to me. Let's go with that.

> Should we save state by default in onPause or let the app decide / handle that?

The whole state saving story with GeckoView is still a bit wonky. For now I wouldn't add this. We have our `DefaultSessionStorage` implementation that gets triggered by `SessionFeature.stop()` already.

PR #699, opened 2018-09-04
Issue #696: browser-engine-system: Clear "X-Requested-With" header when loading URLs.

### LINKED ISSUE #696: Engine does not appear to set "X-Requested-With" to ""
i.e. https://github.com/mozilla-mobile/focus-android/issues/1196

This works for some requests but not all of them (see the issue). I determined the header is not unset by running `ag --ignore=js x-requested`

pocmo:
Yet another WebView thing.. :)

Closes #696

### DIFF
--- a/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
+++ b/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
@@ -13,6 +13,14 @@ import kotlinx.coroutines.experimental.launch
 import mozilla.components.concept.engine.Settings
 import mozilla.components.concept.engine.request.RequestInterceptor
 
+internal val additionalHeaders = mapOf(
+    // For every request WebView sends a "X-requested-with" header with the package name of the
+    // application. We can't really prevent that but we can at least send an empty value.
+    // Unfortunately the additional headers will not be propagated to subsequent requests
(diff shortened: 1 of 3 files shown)


components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
```diff
@@ -29,12 +37,12 @@ class SystemEngineSession(private val defaultSettings: Settings? = null) : Engin
         val internalView = currentView()
 
         if (internalView == null) {
-            // We can't load a URL without a WebView. So let's just remember the URL here until
+            // We can't load al URL without a WebView. So let's just remember the URL here until
```
csadilek: typo?

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
```diff
@@ -13,6 +13,14 @@ import kotlinx.coroutines.experimental.launch
 import mozilla.components.concept.engine.Settings
 import mozilla.components.concept.engine.request.RequestInterceptor
 
+internal val additionalHeaders = HashMap<String, String>().apply {
```
csadilek: Could use ```mutableMapOf/mapOf("X-Requested-With" to "")```

PR #704, opened 2018-09-04
Closes #643: FullScreen mode support...

### LINKED ISSUE #643: Engine/Session: Fullscreen mode
There are a lot of things needed for fullscreen mode. For this issue I propose to only focus on the `browser-session` and `browser-engine` components and ignore the actual UI pieces. We will need the `browser-session` implementation for Focus parity (#200). We should do the `browser-engine` work too to come up with the right state representation in `browser-session`.

The APIs of GeckoView and WebView are fundamentally different which will make this more difficult to implement:
* WebView will return a `View` object that needs to be added to the view root and that replaces/overlays the app UI. In addition to that it returns a Callback object that the app needs to use if it wants to leave full screen mode.
https://developer.android.com/reference/android/webkit/WebChromeClient.html#onShowCustomView(android.view.View,%20android.webkit.WebChromeClient.CustomViewCallback)
* GeckoView will switch itself into fullscreen mode and the app just needs to hide all other UI. Fullscreen mode can be left by calling a method. No callback is needed.

I think we should hide this callback nonsense and everything related inside the `engine-*` component. And just provide a method on the EngineSession to exit full screen mode. Internally the component can do that with the callback or on whatever it needs.

We should add some flag to `browser-session` for the full screen mode. We do not do that in Focus - but it will be needed to correctly restore the full screen state.

jonalmeida:
This was difficult to wrap my head around in the beginning for the known reasons of having different APIs for both the engines. I solved this by having an optional callback for `EngineSession#fullScreenMode` which is called from the SystemWebView.

For GeckoView, we don't need to inform the engine about entering fullscreen mode, so calling `fullScreenMode(true)` does nothing. Only exiting, requires us to notify the engine.

Another difference is that in the `SystemEngineSession` we have access to the current view attached to the EngineSession and we're able to enter and exit Immersive Mode thanks to that, but we're not able to do it for GeckoView, which requires the callers to remember to do it (if they want to?).

### DIFF
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -4,6 +4,7 @@
 
 package mozilla.components.browser.engine.gecko
 
+import android.annotation.SuppressLint
 import kotlinx.coroutines.experimental.CompletableDeferred
 import kotlinx.coroutines.experimental.runBlocking
 import mozilla.components.concept.engine.EngineSession
@@ -194,6 +195,7 @@ class GeckoEngineSession(
     /**
      * See [EngineSession.findNext]
      */
+    @SuppressLint("WrongConstant") // FinderFindFlags annotation doesn't include a 0 value.
     override fun findNext(forward: Boolean) {
         val findFlags = if (forward) 0 else GeckoSession.FINDER_FIND_BACKWARDS
         geckoSession.finder.find(null, findFlags).then { result: GeckoSession.FinderResult? ->
@@ -209,6 +211,13 @@ class GeckoEngineSession(
      */
     override fun clearFindMatches() {
         geckoSession.finder.clear()
+    }
+
+    /**
+     * See [EngineSession.exitFullScreenMode]
+     */
+    override fun exitFullScreenMode() {
+        geckoSession.exitFullScreen()
     }
 
     /**
@@ -319,7 +328,9 @@ class GeckoEngineSession(
 
         override fun onCrash(session: GeckoSession?) = Unit
 
-        override fun onFullScreen(session: GeckoSession, fullScreen: Boolean) = Unit
+        override fun onFullScreen(session: GeckoSession, fullScreen: Boolean) {
+            notifyObservers { onFullScreenChange(fullScreen) }
+        }
 
         override fun onExternalResponse(session: GeckoSession, response: GeckoSession.WebResponseInfo) {
             notifyObservers {
--- a/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
+++ b/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
@@ -6,12 +6,12 @@ package mozilla.components.browser.engine.system
 
 import android.os.Bundle
 import android.webkit.WebView
-import mozilla.components.concept.engine.EngineSession
-import mozilla.components.support.ktx.kotlin.toBundle
-import java.lang.ref.WeakReference
 import kotlinx.coroutines.experimental.launch
+import mozilla.components.concept.engine.EngineSession
 import mozilla.components.concept.engine.Settings
 import mozilla.components.concept.engine.request.RequestInterceptor
+import mozilla.components.support.ktx.kotlin.toBundle
+import java.lang.ref.WeakReference
 
 internal val additionalHeaders = mapOf(
     // For every request WebView sends a "X-requested-with" header with the package name of the
@@ -26,6 +26,7 @@ internal val additionalHeaders = mapOf(
  */
 @Suppress("TooManyFunctions")
 class SystemEngineSession(private val defaultSettings: Settings? = null) : EngineSession() {
+
     internal var view: WeakReference<SystemEngineView>? = null
     internal var scheduledLoad = ScheduledLoad(null)
     @Volatile internal var trackingProtectionEnabled = false
@@ -215,6 +216,13 @@ class SystemEngineSession(private val defaultSettings: Settings? = null) : Engin
                 view.reload()
             }
         }
+    }
+
+    /**
+     * See [EngineSession.exitFullScreenMode]
+     */
+    override fun exitFullScreenMode() {
+        // no-op
     }
 
     internal fun toggleDesktopUA(userAgent: String, requestDesktop: Boolean): String {
--- a/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
+++ b/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
@@ -45,6 +45,7 @@ class SystemEngineView @JvmOverloads constructor(
     internal var currentWebView = createWebView(context)
     internal var currentUrl = ""
     private var session: SystemEngineSession? = null
+    internal var fullScreenCallback: WebChromeClient.CustomViewCallback? = null
 
     init {
         // Currently this implementation supports only a single WebView. Eventually this
@@ -95,6 +96,7 @@ class SystemEngineView @JvmOverloads constructor(
 
     private fun createWebView(context: Context): WebView {
         val webView = WebView(context)
+        webView.tag = "mosac_system_engine_webview"
         webView.webViewClient = createWebViewClient(webView)
         webView.webChromeClient = createWebChromeClient()
         webView.setDownloadListener(createDownloadListener())
@@ -182,6 +184,16 @@ class SystemEngineView @JvmOverloads constructor(
         override fun onReceivedTitle(view: WebView, title: String?) {
             session?.internalNotifyObservers { onTitleChange(title ?: "") }
         }
+
+        override fun onShowCustomView(view: View, callback: CustomViewCallback) {
+            addFullScreenView(view, callback)
+            session?.internalNotifyObservers { onFullScreenChange(true) }
+        }
+
+        override fun onHideCustomView() {
+            removeFullScreenView()
+            session?.internalNotifyObservers { onFullScreenChange(false) }
+        }
     }
 
     internal fun createDownloadListener(): DownloadListener {
@@ -235,6 +247,26 @@ class SystemEngineView @JvmOverloads constructor(
             return true
         }
         return false
+    }
+
+    internal fun addFullScreenView(view: View, callback: WebChromeClient.CustomViewCallback) {
+        val webView = findViewWithTag<WebView>("mosac_system_engine_webview")
+        webView?.apply { this.visibility = View.GONE }
+
+        fullScreenCallback = callback
+
+        view.tag = "mosac_system_engine_fullscreen"
+        addView(view)
+    }
+
+    internal fun removeFullScreenView() {
+        val view = findViewWithTag<View>("mosac_system_engine_fullscreen")
+        val webView = findViewWithTag<WebView>("mosac_system_engine_webview")
+        view?.let {
+            fullScreenCallback?.onCustomViewHidden()
+            webView?.apply { this.visibility = View.VISIBLE }
+            removeView(view)
+        }
     }
 
     class ImageHandler(val session: SystemEngineSession?) : Handler() {
--- a/components/browser/engine-system/src/test/java/mozilla/components/browser/engine/system/SystemEngineSessionTest.kt
+++ b/components/browser/engine-system/src/test/java/mozilla/components/browser/engine/system/SystemEngineSessionTest.kt
@@ -26,9 +26,9 @@ import org.mockito.ArgumentMatchers.eq
 import org.mockito.Mockito.`when`
 import org.mockito.Mockito.doReturn
 import org.mockito.Mockito.mock
+import org.mockito.Mockito.never
 import org.mockito.Mockito.spy
 import org.mockito.Mockito.verify
-import org.mockito.Mockito.never
 import org.robolectric.RobolectricTestRunner
 import org.robolectric.RuntimeEnvironment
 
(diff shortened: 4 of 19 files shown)


components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
```diff
@@ -179,6 +185,16 @@ abstract class EngineSession(
...
 
+    /**
+     * Enables/disables fullscreen mode. It also invokes a callback when fullscreen mode is exited.
+     * For GeckoView, this shouldn't necessarily be used to
+     *
+     * @param enable Turns fullscreen mode on/off.
+     * @param exitCallback (Optional) Invoked when the engine exits fullscreen mode. This is mostly
+     * needed for only SystemWebView implementations which need to invoke WebView APIs on exiting.
+     */
+    abstract fun fullScreenMode(enable: Boolean, exitCallback: () -> Unit? = {})
```
pocmo: If we are able to hide the View inside the engine (see above) then we may be able to do the same for the callback.

components/support/ktx/src/main/java/mozilla/components/support/ktx/android/view/View.kt
```diff
@@ -83,6 +85,69 @@ fun View.hideKeyboard() {
...
+ *
+ * Examples:
+ * ```
+ * toolbar.visibleOrGone(foo != bar)
+ * imageView.visibleOrGone(foo.isEnabled)
+ * ```
+ */
+fun View.visibleOrGone(enabled: Boolean) {
+    this.visibility = if (enabled) View.VISIBLE else View.GONE
+}
```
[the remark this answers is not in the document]
jonalmeida: It was mostly for product teams that needed to change multiple View's visibility dependent on some single logic.
Example:
```kotlin
if (enabled) {
  view1.visibility = View.VISIBLE
  view2.visibility = View.VISIBLE
  view3.visibility = View.VISIBLE
} else {
  view1.visibility = View.GONE
  view2.visibility = View.GONE
  view3.visibility = View.GONE
}
```
Instead we could do just this:
```kotlin
view1.visibleOrGone(enabled)
view2.visibleOrGone(enabled)
view3.visibleOrGone(enabled)
```
It's fair though to remove this as well because I can't see many use cases (for example) in focus where we follow that pattern.

components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
```diff
@@ -47,6 +48,7 @@ class Session(
         fun onTrackerBlocked(session: Session, blocked: String, all: List<String>) = Unit
         fun onLongPress(session: Session, hitResult: HitResult): Boolean = false
         fun onFindResult(session: Session, result: FindResult) = Unit
+        fun onFullScreenChanged(session: Session, enabled: Boolean, view: View? = null) = Unit
```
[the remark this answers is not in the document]
jonalmeida: I was trying to go for giving the client the opportunity to handle the view how ever they wanted, but it makes sense to handle the View placement for them.

In terms of layout, the SystemEngineView is in a FrameLayout, so should we programmatically add a new View above the engine?

components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
```diff
@@ -5,17 +5,22 @@
...
+import android.view.View
 import mozilla.components.support.base.observer.Observable
 import mozilla.components.support.base.observer.ObserverRegistry
 
 /**
  * Class representing a single engine session.
  *
  * In browsers usually a session corresponds to a tab.
+ *
+ * @param fullScreenCallback Holds onto a callback to be invoked when fullscreen mode is exited.
```
csadilek: This should be `@property` and docs probably just `A callback to be invoked when...`

components/support/ktx/src/main/java/mozilla/components/support/ktx/android/view/View.kt
```diff
@@ -83,6 +85,69 @@ fun View.hideKeyboard() {
...
+        val window = context.window
+        window.addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON)
+        window.decorView.systemUiVisibility = (View.SYSTEM_UI_FLAG_LAYOUT_STABLE
+            or View.SYSTEM_UI_FLAG_LAYOUT_HIDE_NAVIGATION
+            or View.SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN
+            or View.SYSTEM_UI_FLAG_HIDE_NAVIGATION
+            or View.SYSTEM_UI_FLAG_FULLSCREEN
+            or View.SYSTEM_UI_FLAG_IMMERSIVE_STICKY)
+    }
+}
```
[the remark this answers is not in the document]
jonalmeida: Yeah, it's a bit weird. Your previous comment about letting the app decide to to call it, means I don't have to put it on a View anymore. I was trying to handle this for them.  ¯\\\_(ツ)\_/¯

components/browser/engine-system/src/test/java/mozilla/components/browser/engine/system/SystemEngineSessionTest.kt
```diff
@@ -393,7 +396,7 @@ class SystemEngineSessionTest {
     }
 
     @Test
-    fun testFindAll() {
+    fun findAll() {
```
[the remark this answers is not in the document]
jonalmeida: Seems fair, I'll stick with the original.

components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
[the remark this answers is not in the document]
pocmo: I'd try dropping it into the FrameLayout and hiding the WebView.

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
```diff
@@ -211,6 +214,19 @@ class SystemEngineSession(private val defaultSettings: Settings? = null) : Engin
...
+     */
+    override fun fullScreenMode(enable: Boolean, exitCallback: () -> Unit?) {
+        if (enable) {
+            currentView()?.enterToImmersiveMode()
+            fullScreenCallback = exitCallback
+        } else {
+            currentView()?.exitImmersiveModeIfNeeded()
+            fullScreenCallback.invoke()
+        }
+    }
```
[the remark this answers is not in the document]
jonalmeida: Fixed!

jonalmeida:
browser-engine-system: `Could not resolve mozilla:fxa_client_android:0.2.4.`

pocmo:
Meh. Every now and then we have problems with taskcluster or jcenter or plugins.gradle.org not being resolved.

Btw.: You can re-trigger taskcluster by updating the PR title (just add dots..)

jonalmeida:
@pocmo I tried; don't have access. :(

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -180,6 +181,7 @@ class GeckoEngineSession(
     /**
      * See [EngineSession.findNext]
      */
+    @SuppressLint("WrongConstant")
```
pocmo: Can you add a comment explaining why this falsely triggers "WrongConstant"?

components/support/ktx/src/test/java/mozilla/components/support/ktx/android/view/ViewTest.kt
```diff
@@ -4,19 +4,19 @@
...
 import org.junit.Assert.assertTrue
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.robolectric.RobolectricTestRunner
 import org.robolectric.RuntimeEnvironment
 import org.robolectric.shadows.ShadowLooper
-import android.util.DisplayMetrics
-import android.view.View
-import org.junit.Assert.assertEquals
-import org.junit.Assert.assertNotEquals
```
pocmo: nit: This file only contains "import" changes now..

components/support/ktx/src/main/java/mozilla/components/support/ktx/android/view/View.kt
```diff
@@ -68,7 +68,7 @@ fun View.showKeyboard(flags: Int = InputMethodManager.SHOW_IMPLICIT) {
  */
 fun View.hideKeyboard() {
     val imm = (context.getSystemService(Context.INPUT_METHOD_SERVICE) ?: return)
-            as InputMethodManager
+        as InputMethodManager
```
pocmo: nit: Unrelated

components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
```diff
@@ -179,6 +180,14 @@ abstract class EngineSession(
      */
     abstract fun clearFindMatches()
 
+    /**
+     * Enables/disables fullscreen mode. It also invokes a callback when fullscreen mode is exited.
+     * For GeckoView, this shouldn't necessarily be used to
```
pocmo: The comment still mentions the callback :)

components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
```diff
@@ -232,9 +239,13 @@ class Session(
      * Desktop Mode state, true if the desktop mode is requested, otherwise false.
      */
     var desktopMode: Boolean by Delegates.observable(false) { _, old, new ->
-        notifyObservers(old, new) {
-            onDesktopModeChanged(this@Session, new)
-        }
+        notifyObservers(old, new) { onDesktopModeChanged(this@Session, new) }
+    }
+    /**
+     * The fullscreen state. True if it is in fullscreen mode.
```
pocmo: nit: empty line between `}` and comment of next method.

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -237,6 +249,26 @@ class SystemEngineView @JvmOverloads constructor(
         return false
     }
 
+    internal fun addFullScreenView(view: View, callback: WebChromeClient.CustomViewCallback) {
+        val webView = findViewWithTag<WebView>("webview")
+        webView?.apply { this.visibility = View.GONE }
+
+        fullScreenCallback = callback
+
+        view.tag = "fullscreen"
```
pocmo: nit: We should make sure that we do not clash with tags/ids of the consuming apps. For most resources we use a "mozac" prefix and sometimes another "module" prefix (e.g. Strings: `mozac_browser_toolbar_menu_button)`.

components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
```diff
@@ -179,6 +180,14 @@ abstract class EngineSession(
...
+     *
+     * @param enable Turns fullscreen mode on/off.
+     */
+    abstract fun toggleFullScreenMode(enable: Boolean)
+
```
pocmo: `toggleFullScreenMode` doesn't seem to be the right thing here: It never goes to fullscreen mode. In GeckoView it only exists full screen mode. In WebView it does nothing. Should this only be a method that exits full screen mode? And shouldn't we invoke the WebView callback in the `browser-engine-system` implementation`

components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
jonalmeida: Ah, sorry!

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
jonalmeida: I used `mosac_system_engine_*`.

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
jonalmeida: Sure, it was the linter that complains since there are IntDef values for `@FinderFindFlags` that don't include the default 0 value. I'll add that in a comment by the annotation and I've filed a bug for GV ([bug 1489528][0]).

[0]: https://bugzilla.mozilla.org/show_bug.cgi?id=1489528

components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
jonalmeida: You're right. Since we're hiding the WebView implementation in the screen this method doesn't follow the original idea. Sorry, I should have thought about this earlier.

components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
```diff
@@ -122,65 +125,66 @@ class Session(
...
+        notifyObservers(old, new) { onSecurityChanged(this@Session, new) }
     }
 
     /**
      * Configuration data in case this session is used for a Custom Tab.
      */
-    var customTabConfig: CustomTabConfig? by Delegates.observable<CustomTabConfig?>(null) {
-        _, _, new -> notifyObservers { onCustomTabConfigChanged(this@Session, new) }
+    var customTabConfig: CustomTabConfig? by Delegates.observable<CustomTabConfig?>(
+        null) { _, _, new ->
```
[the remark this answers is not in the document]
jonalmeida: Auto-formatting.  ¯\\\_(ツ)\_/¯ 

I'll disable it until I figure out how to stop it from misbehaving.

PR #712, opened 2018-09-06
Data Protection API

linuxwolf:
A first-pass at implementing a data protection API using AndroidKeyStore for encryption keys.

Tests obviously need to be written, but was hoping opening the PR now would help to understand and evaluate the shape of the API.

I do have a [demo app](https://github.com/linuxwolf/android-components/compare/linuxwolf/dataprotect...linuxwolf:linuxwolf/sample-dataprotect) I could also submit, but felt that would make this PR too big to be digested.


components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/Keystore.kt
```diff
@@ -0,0 +1,98 @@
...
+private val ENCRYPTED_VERSION = 0x02
+
+private val CIPHER_ALG = KeyProperties.KEY_ALGORITHM_AES
+private val CIPHER_MOD = KeyProperties.BLOCK_MODE_GCM
+private val CIPHER_PAD = KeyProperties.ENCRYPTION_PADDING_NONE
+private val CIPHER_TAG_LEN = 128
+private val CIPHER_SPEC = "${CIPHER_ALG}/${CIPHER_MOD}/${CIPHER_PAD}"
+
+@TargetApi(Build.VERSION_CODES.M)
+class Keystore {
```
pocmo: I wonder if the API would be simpler if `label` would be a constructor parameter of `KeyStore` and you'd create multiple instances if you'd need multiple labels. But that may depend on how this is going to be used in an app, I guess.

components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/Keystore.kt
```diff
@@ -0,0 +1,98 @@
...
+import javax.crypto.spec.GCMParameterSpec
+
+private val KEYSTORE_PROVIDER = "AndroidKeyStore"
+private val ENCRYPTED_VERSION = 0x02
+
+private val CIPHER_ALG = KeyProperties.KEY_ALGORITHM_AES
+private val CIPHER_MOD = KeyProperties.BLOCK_MODE_GCM
+private val CIPHER_PAD = KeyProperties.ENCRYPTION_PADDING_NONE
+private val CIPHER_TAG_LEN = 128
+private val CIPHER_SPEC = "${CIPHER_ALG}/${CIPHER_MOD}/${CIPHER_PAD}"
```
pocmo: Are we expecting to pick values here that will work for every app or do you think this should be configurable by a consuming app (constructor parameters with default values? e.g. `val cipherMod: String = KeyProperties.BLOCK_MODE_GCM`)?

components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/Keystore.kt
linuxwolf: I was basing the API on work done in NSS for [OSKeyStore](https://dxr.mozilla.org/mozilla-central/source/security/manager/ssl/nsIOSKeyStore.idl).  We don't have to be religious about it, and I suspect the work is about the same for both models.

In Lockbox, I think we'll just use a single label.  I can foresee how multiple labels could be applied.

I'm happy to change this to a single label set on construction if that's the preference.

components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/Keystore.kt
linuxwolf: I think there's little-to-no value in making these configurable; this crypto is just on the device, for the app consuming the API.  It (the key, therefore the encrypted data) can't leave the device.

The algorithm spec chosen is the strongest available, and I think it should be enforced.

linuxwolf:
I've been working on the linting issues, and slowly adding some tests.

I should have a first pass at KDoc comments with the next round of commits, too.

components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/Keystore.kt
pocmo: If there will be only a single label in lockbox then the constructor API may be more handy not requiring you to pass the label around all the time? :)

But yeah, not religious about it either. :)

components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/Keystore.kt
pocmo: 👍

components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/Keystore.kt
csadilek: That's the only thing I noticed as well as the `generateKey` method seemed suspicious to me, since it's not returning anything. Not a big deal, but would also prefer the constructor parameter and `Keystore` being immutable.

linuxwolf: I do need to note the exceptions to be thrown, which are a lot \-:

Automatically creating the key risks some data corruption and crashing if the keystore  was reset for some reason but the data stored did not; hence the separation.

Definitely open to suggestions, though.

linuxwolf:
@pocmo @csadilek Looks like this is running into some potentially odd crypto errors.  Tests pass on my machine, but are failing in taskcluster; I'l try things out again on-device (using a sample app to-be-submitted), but I'd like some guidance for what to do:

* We could try and troubleshoot what the java platform configuration is, and see if it's something with whatever security/JCE providers are provisioned there
* I could try to get things setup and using BouncyCastle as the JCE provider and see if that fixes the issues on TaskCluster (it's hopefully a known-good state compared with whatever is configured on the Taskcluster machines)
* Try running these on a simulator instead (although I see no provisions for that today, so mabye that's a too-big hill to climb)
* Mark the specific test as ignored for now, and continue with some other option(s)

linuxwolf:
For now I marked the test as @ignored; let me know what I can help with to make it really work.

csadilek:
@linuxwolf Do you remember / capture the error we had on Taskcluster? Was it `java.security.InvalidKeyException: Illegal key size` because I get these errors locally (for all tests)?

components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/Keystore.kt
```diff
@@ -0,0 +1,190 @@
...
+internal const val CIPHER_TAG_LEN = 128
+internal const val CIPHER_SPEC = "$CIPHER_ALG/$CIPHER_MOD/$CIPHER_PAD"
+
+internal const val CIPHER_NONCE_LEN = 12
+
+/**
+ * Wraps the critical functions around a Java KeyStore to better facilitate testing
+ * and instrumenting.
+ *
+ * @suppress
```
csadilek: Probably just a left-over here?

csadilek:
@linuxwolf Would it be bad to set `Security.setProperty("crypto.policy", "unlimited")` just in the tests? This will make it work for me.

linuxwolf:
There exception I saw thrown in Taskcluster was `javax.crypto.AEADBadTagException`, but not locally.

An `InvalidKeyException` exception (in this case) is fixed by installing the "JCE Unlimited Strength Jurisdiction" policy ([can be found here](http://www.oracle.com/technetwork/java/javase/downloads/jce8-download-2133166.html)).

linuxwolf:
> Would it be bad to set Security.setProperty("crypto.policy", "unlimited") just in the tests? This will make it work for me.

That would work around issues here unlimited strength isn't set, but it doesn't solve the Taskcluster problem I saw.  I don't think it hurts, though.

csadilek:
> That would work around issues here unlimited strength isn't set, but it doesn't solve the Taskcluster problem I saw. I don't think it hurts, though.

OK, let's add it so we don't have build failures on our dev machines. 

Coverage is pretty good otherwise. I think we can squash the commits and merge it, and file a follow-up to investigate the Taskcluster problem later.

linuxwolf:
> Coverage is pretty good otherwise. I think we can squash the commits and merge it, and file a follow-up to investigate the Taskcluster problem later.

I'm addressing the configuration now, and making sure all the @throws are accounted for in KDoc.

linuxwolf:
@csadilek Everything should be up to date now: docs, test configs, lint, etc.

linuxwolf:
@csadilek not at all.  Will push that presently!

csadilek:
Filed a follow-up issue to fix tests on TaskCluster: https://github.com/mozilla-mobile/android-components/issues/786

csadilek:
Thanks @linuxwolf !

PR #723, opened 2018-09-06
Closes #691: Dokka: Generate aggregated API docs

### LINKED ISSUE #691: Dokka: Generate aggregated API docs
Currently we create separate API docs for all modules. Because of that you can't jump between modules in the documentation. For example if you are on the page about `GeckoEngine` you can't get to `Engine` because this class is in a different module.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-5891)

csadilek:


build.gradle
```diff
@@ -83,6 +83,11 @@ allprojects {
     }
 }
 
+configurations {
+    dokkapath
+    dokkapath.description = "dokka project classpath"
+}
+
```
pocmo: I assume we can remove `dokkapath` now. Nothing is added to it and we can just pass `[]` to `processConfigurations` in the dokka task?

build.gradle
```diff
@@ -163,6 +162,41 @@ subprojects {
     }
 }
 
+evaluationDependsOnChildren()
+
+task dokka(type: org.jetbrains.dokka.gradle.DokkaAndroidTask, overwrite: true) {
```
pocmo: I think we need to update .taskcluster.yml and generate_api-reference_release.sh to use the new task name?
* https://github.com/mozilla-mobile/android-components/blob/master/automation/docs/generate_api-reference_release.sh
* https://github.com/mozilla-mobile/android-components/blob/master/.taskcluster.yml#L70

Maybe we want to remove it from the release task? It's currently a manual process we invoke **before** the release.

build.gradle
csadilek: 👍 I removed this and left out `processConfigurations. Seems fine :).

build.gradle
csadilek: Renamed our new tasks to `docs`. So, it's the same as before.

csadilek:
@pocmo this is updated now! Let's see what taskcluster does now that it generates the docs.

csadilek:
🚢 ?

PR #727, opened 2018-09-07
Issue #679: Add first iteration of documentation for engine components.

### LINKED ISSUE #679: Document browser-engine component(s)
This component is making its way into more and more apps. We should start to document it in the README(s).

pocmo:
Closes #679.


components/browser/engine-gecko-beta/README.md
```diff
@@ -12,6 +14,46 @@ Use gradle to download the library from JCenter:
 implementation "org.mozilla.components:engine-gecko-beta:{latest-version}
 ```
 
+### Initializing
+
+It is recommended t create only one `GeckoEngine` instance per app. To create a `GeckoEngine` a `GeckoRuntime` and optionally `DefaultSettings` are needed.
```
Amejia481: Small improvement, the **t** missed the **o**
It is recommended **t** should be t**o**

pocmo: Thanks!

components/browser/engine-gecko-beta/README.md
```diff
@@ -12,6 +14,46 @@ Use gradle to download the library from JCenter:
 implementation "org.mozilla.components:engine-gecko-beta:{latest-version}
```
Amejia481: I'm not quite sure 🤔, if we should also include as a step adding the dependency of 
**geckoview**, I couldn't resolve **GeckoRuntime** reference by just following the instructions
I had to included. 
> implementation "org.mozilla.geckoview:{latest-version}"

components/browser/engine-gecko-beta/README.md
```diff
@@ -12,6 +14,46 @@ Use gradle to download the library from JCenter:
...
+// Create default settings (optional) and enable tracking protection for all future sessions.
+val defaultSettings = DefaultSettings().apply {
+    trackingProtectionPolicy = EngineSession.TrackingProtectionPolicy.all()
+}
+
+// Create and initialize a Gecko runtime with the default settings.
+val runtime = GeckoRuntime.getDefault(applicationContext)
+
+// Create an engine instance to be used by other components.
+val engine = GeckoEngine(runtime, defaultSettings)
```
Amejia481: Maybe in the future we could encapsulated the creation of the  **GeckoEngine** using the [builder pattern](https://en.wikipedia.org/wiki/Builder_pattern), providing some smart defaults values to easy the creating of GeckoEngine, this way the user of the API can use **GeckoEngine** right away and if want something custom just can mix and match.

components/browser/engine-gecko-beta/README.md
pocmo: Yeah, that's a great idea. And we should link to the CHANGELOG that mentions compatible versions. I'll file a follow-up.

components/browser/engine-gecko-beta/README.md
pocmo: I feel like Kotlin's ability to have default values in constructor parameters makes the builder pattern most obsolete. We could definitely add `GeckoRuntime.getDefault(applicationContext)` as a default parameter here! I'll file that separately.

Amejia481: 👍

components/browser/engine-gecko-beta/README.md
Amejia481: 👍

Issue #735, opened 2018-09-10
Adding jacoco.exec file to .gitignore file


Amejia481:
I noticed that @jonalmeida  and @csadilek  are suffering like me from the file `jacoco.exec` showing in the changed files every time that code coverage is run.

I'll summit a pull request adding it to the `.gitignore`, if it's not a problem.

![image](https://user-images.githubusercontent.com/773158/45299131-9b863680-b4d8-11e8-9c34-c50d5d162c1e.png)

PR #736, opened 2018-09-10
Closes: #735: Adding jacoco.exec and similar files to .gitignore

Amejia481:
Closes: #735


Amejia481:
You're right is better to be more specific 👍

PR #740, opened 2018-09-10
Closes #644: Clear Data (WebView only).

### LINKED ISSUE #644: Engine: cleanup/clear data
Parent: #200

We need to offer a way to clear the browsing data. For now that doesn't need to be a fine-grained API. Just remove everything.

* In Focus (WebView build) that's used for removing all data after the end of a browsing session
* In the Fire TV app there's a setting to clear the browsing data

For Focus and private browsing that may be handled in #499 already.

For our existing apps we'd only need to support this in WebView for now.

jonalmeida:
This only includes the implementation for the SystemWebView that's based
off of what we currently do in Focus/Fire TV. Since this is needed for
those apps now, we can work with this implementation until GeckoView
provides us with a nicer API to do the equivalent for it. (Bug 1489669)

We also don't notify any observers since there isn't any
confirmation/information available that we can propagate to clients.


jonalmeida:
@csadilek I'm not sure what to do with this PR, if you check Taskcluster and Travis, they've both passed successfully, but github shows that it's still pending?

PR #749, opened 2018-09-11
Closes #286: Process Share Intents (ACTION_SEND)

### LINKED ISSUE #286: Process Share Intents (ACTION_SEND)
In Focus we process share intents (ACTION_SEND):
https://github.com/mozilla-mobile/android-components/issues/282

The intent contains additional text. If the text is a URL then we load it. Otherwise we trigger a search.

Amejia481:
Closes #286: Process Share Intents (ACTION_SEND)

*Handling action ACTION_SEND on SessionIntentProcessor
    if the intent contains additional text, do the following
     If the text is a URL then we load it,
     Otherwise we trigger a search.

* Adding support-ktx module to feature-session module
    
* Updating the sample-browser

### DIFF
--- a/components/feature/session/src/main/java/mozilla/components/feature/session/SessionIntentProcessor.kt
+++ b/components/feature/session/src/main/java/mozilla/components/feature/session/SessionIntentProcessor.kt
@@ -10,21 +10,29 @@ import mozilla.components.browser.session.Session
 import mozilla.components.browser.session.Session.Source
 import mozilla.components.browser.session.SessionManager
 import mozilla.components.browser.session.tab.CustomTabConfig
+import mozilla.components.support.ktx.kotlin.isUrl
 import mozilla.components.support.utils.SafeIntent
 
 typealias IntentHandler = (Intent) -> Boolean
+
+typealias TextSearchHandler = (String, Session) -> Unit
 
 /**
  * Processor for intents which should trigger session-related actions.
  *
  * @param openNewTab Whether a processed Intent should open a new tab or open URLs in the currently
  *                   selected tab.
+ *
+ * @param textSearchHandler Will be invoked for ACTION_SEND intents when the provided EXTRA_TEXT is not a URL,
+ *                          or a empty string. A behavior must provided to handle the search,
+ *                          otherwise by default it will do nothing.
  */
 class SessionIntentProcessor(
     private val sessionUseCases: SessionUseCases,
     private val sessionManager: SessionManager,
     useDefaultHandlers: Boolean = true,
-    private val openNewTab: Boolean = true
+    private val openNewTab: Boolean = true,
+    private val textSearchHandler: TextSearchHandler = { _, _ -> }
 ) {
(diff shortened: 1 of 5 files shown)


components/feature/session/src/main/java/mozilla/components/feature/session/SessionIntentProcessor.kt
```diff
@@ -43,22 +51,42 @@ class SessionIntentProcessor(
...
+                true
+            }
             else -> {
-                val session = if (openNewTab) {
-                    Session(url, false, Source.ACTION_VIEW).also { sessionManager.add(it, selected = true) }
-                } else {
-                    sessionManager.selectedSession ?: Session(url, false, Source.ACTION_VIEW)
-                }
-
+                val session = createSession(url)
```
csadilek: Would be nicer to have a `handleActionView` method here as well, similar to `handleActionSend`

components/feature/session/src/test/java/mozilla/components/feature/session/SessionIntentProcessorTest.kt
```diff
@@ -151,6 +151,49 @@ class SessionIntentProcessorTest {
...
+        verify(engineSession).loadUrl("http://mozilla.org")
+    }
+
+    @Test
+    fun `perform search when action send with text`() {
+        val engine = mock(Engine::class.java)
+        val sessionManager = spy(SessionManager(engine))
+        doReturn(engineSession).`when`(sessionManager).getOrCreateEngineSession(anySession())
+        val useCases = SessionUseCases(sessionManager)
+        var isSearchHandlerCalled = false
```
csadilek: just `searchHandlerCalled`...not a method :)

components/feature/session/src/main/java/mozilla/components/feature/session/SessionIntentProcessor.kt
```diff
@@ -10,25 +10,33 @@ import mozilla.components.browser.session.Session
...
+typealias TextSearchHandler = (String, Session) -> Unit
+
 /**
  * Processor for intents which should trigger session-related actions.
  *
  * @param openNewTab Whether a processed Intent should open a new tab or open URLs in the currently
  *                   selected tab.
+ *
+ * @param textSearchHandler This lambda expression will be invoked when a text search should be performed,
+ *                           the consumer must provide a behavior to handle the search, by default it will do nothing.
```
csadilek: I think we can be more concrete here i.e. when exactly is this invoked. So instead of "This lambda expression..." (which can be seen based on the type) I would write "textSearchHandler Will be invoked for ACTION_SEND intents when the provided dataString is not a URL..."

components/feature/session/src/main/java/mozilla/components/feature/session/SessionIntentProcessor.kt
Amejia481: 👍

csadilek:
Just a nit: There's a also a typo in the commit message: Should be "Closes #286: ..."

PR #751, opened 2018-09-11
Closes #750: UrlMatcher incorrectly uses Java 8 Map.forEach

### LINKED ISSUE #750: UrlMatcher incorrectly uses Java 8 Map.forEach
It should be using forEach from the Kotlin stdlib.

csadilek:


csadilek:
Follow-up issue file to make your build fail in case this happens again: https://github.com/mozilla-mobile/android-components/issues/752

PR #787, opened 2018-09-14
Resolve #754: refactor View.dp() as Resources.pxToDp

### LINKED ISSUE #754: View.dp has nothing to do with View
We have a function `View.dp()` that converts px to dp:
https://github.com/mozilla-mobile/android-components/blob/d055cc58566737697b360bd08728b80a7bffa91c/components/support/ktx/src/main/java/mozilla/components/support/ktx/android/view/View.kt#L32-L33

However, this has nothing to do with a View, which makes it hard to find and clutters View with confusing methods. I would propose one of:
```kt
fun Int.pxToDp(context: Context)
fun Resources.pxToDp(pixels: Int)
```

SubhrajyotiSen:


components/support/ktx/src/test/java/mozilla/components/support/ktx/android/content/res/Resources.kt
```diff
@@ -0,0 +1,19 @@
+package mozilla.components.support.ktx.android.content.res
+
+import android.util.DisplayMetrics
+import org.junit.Assert
+import org.junit.Test
+import org.robolectric.RuntimeEnvironment
+
+@Test
```
csadilek: Similar to `ViewTest` can we rename this file to `ResourcesTest` and create a test class for it?

components/support/ktx/src/test/java/mozilla/components/support/ktx/android/content/res/Resources.kt
```diff
@@ -0,0 +1,19 @@
...
+fun `dp returns same value as manual conversion`() {
```
csadilek: Similar to ViewTest can we rename this file to ResourcesTest and create a test class for it?

csadilek:
Thanks @SubhrajyotiSen. @mcomella when you get a chance, can you take a look at this as well?

components/support/ktx/src/test/java/mozilla/components/support/ktx/android/content/res/Resources.kt
SubhrajyotiSen: I will take a look

SubhrajyotiSen:
I will fix the ktlint failure when I'm back at my computer later today

components/support/ktx/src/test/java/mozilla/components/support/ktx/android/content/res/Resources.kt
```diff
@@ -0,0 +1,25 @@
...
+
+import android.util.DisplayMetrics
+import org.junit.Assert
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.RuntimeEnvironment
+
+@RunWith(RobolectricTestRunner::class)
+class ResourcesTest {
```
csadilek: @SubhrajyotiSen I think the filename should be ResourcesTest.kt as well for consistency. Thanks!

PR #792, opened 2018-09-15
Resolve #790: Add ext function for `isScreenReaderEnabled`

### LINKED ISSUE #790: Add extension fn for `isScreenReaderEnabled`
This is currently in production in fire TV:
https://github.com/mozilla-mobile/firefox-tv/blob/366ebfb70f7d3d3e46d622e338be278cbc90e02b/app/src/main/java/org/mozilla/focus/ext/Context.kt#L26

SubhrajyotiSen:
Also, implement extension function for getAccessibilityManager to keep it uniform with the Firefox TV approach.

### DIFF
--- a/components/support/ktx/src/main/java/mozilla/components/support/ktx/android/content/Context.kt
+++ b/components/support/ktx/src/main/java/mozilla/components/support/ktx/android/content/Context.kt
@@ -6,6 +6,7 @@ package mozilla.components.support.ktx.android.content
 
 import android.app.ActivityManager
 import android.content.Context
+import android.view.accessibility.AccessibilityManager
 
 /**
  * The (visible) version name of the application, as specified by the <manifest> tag's versionName
@@ -33,3 +34,10 @@ fun Context.isOSOnLowMemory(): Boolean {
         activityManager.getMemoryInfo(memoryInfo)
     }.lowMemory
 }
+
+/**
+ * AccessibilityManager.isAccessibilityEnabled can be enabled for more things than just VoiceView. If we want
+ * VoiceView, we only need to check this one field; via comments on https://stackoverflow.com/a/12362545/582004.
+ */
+fun Context.isScreenReaderEnabled() =
+        (getSystemService(Context.ACCESSIBILITY_SERVICE) as AccessibilityManager).isTouchExplorationEnabled
--- a/components/support/ktx/src/main/java/mozilla/components/support/ktx/android/content/Context.kt
+++ b/components/support/ktx/src/main/java/mozilla/components/support/ktx/android/content/Context.kt
@@ -25,7 +25,7 @@ inline fun <reified T> Context.systemService(name: String): T {
     return getSystemService(name) as T
 }
 
-fun Context.getAccessibilityManager() = getSystemService(Context.ACCESSIBILITY_SERVICE) as AccessibilityManager
+fun Context.getAccessibilityManager() = getSystemService(Context.ACCESSIBILITY_SERVICE) as? AccessibilityManager
 
 /**
  * Returns whether or not the operating system is under low memory conditions.
@@ -41,4 +41,4 @@ fun Context.isOSOnLowMemory(): Boolean {
  * AccessibilityManager.isAccessibilityEnabled can be enabled for more things than just VoiceView. If we want
  * VoiceView, we only need to check this one field; via comments on https://stackoverflow.com/a/12362545/582004.
  */
-fun Context.isScreenReaderEnabled() = getAccessibilityManager().isTouchExplorationEnabled
+fun Context.isScreenReaderEnabled() = getAccessibilityManager()?.isTouchExplorationEnabled
--- a/components/support/ktx/src/main/java/mozilla/components/support/ktx/android/content/Context.kt
+++ b/components/support/ktx/src/main/java/mozilla/components/support/ktx/android/content/Context.kt
@@ -25,6 +25,8 @@ inline fun <reified T> Context.systemService(name: String): T {
     return getSystemService(name) as T
 }
 
+fun Context.getAccessibilityManager() = getSystemService(Context.ACCESSIBILITY_SERVICE) as AccessibilityManager
+
 /**
  * Returns whether or not the operating system is under low memory conditions.
  */
@@ -39,5 +41,4 @@ fun Context.isOSOnLowMemory(): Boolean {
  * AccessibilityManager.isAccessibilityEnabled can be enabled for more things than just VoiceView. If we want
  * VoiceView, we only need to check this one field; via comments on https://stackoverflow.com/a/12362545/582004.
  */
-fun Context.isScreenReaderEnabled() =
-        (getSystemService(Context.ACCESSIBILITY_SERVICE) as AccessibilityManager).isTouchExplorationEnabled
+fun Context.isScreenReaderEnabled() = getAccessibilityManager().isTouchExplorationEnabled
--- a/components/support/ktx/src/test/java/mozilla/components/support/ktx/android/content/ContextTest.kt
+++ b/components/support/ktx/src/test/java/mozilla/components/support/ktx/android/content/ContextTest.kt
@@ -6,29 +6,42 @@ package mozilla.components.support.ktx.android.content
 
 import android.app.ActivityManager
 import android.content.Context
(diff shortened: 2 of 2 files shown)


SubhrajyotiSen:
Can someone help me figure out how to increase code coverage on this patch?

jonalmeida:
@SubhrajyotiSen [looking at the CodeCov diff][1] I can see the cast line highlighted. So I think it might be the branch where `getSystemService(Context.ACCESSIBILITY_SERVICE) as AccessibilityManager` fails. You could try to write another test where the casting fails.

[1]: https://codecov.io/gh/mozilla-mobile/android-components/compare/7ae80694c44dd6751f9de4ebd0ab7bba9cf5e888...386a5b8c80aeb60ac6903a2e77a13fd4c70f23c6/diff

jonalmeida:
@SubhrajyotiSen if you want to compare your test changes locally, you can build the tests with coverage for that module and compare the generated jacoco report with an older version:

```bash
./gradlew :support-ktx:test -Pcoverage
open components/support/ktx/build/reports/jacoco/jacocoTestReport/html/index.html
```

SubhrajyotiSen:
Thanks for the instructions and pointing out where the error might be.
But I am unable to figure out how to write a test to check for cast failure in this case

jonalmeida:
I believe the issue lies in the return type for the extension function which should return a nullable `AccessibilityManager` (notice the question mark added to the `as`):
```kotlin
fun Context.getAccessibilityManager() =
    getSystemService(Context.ACCESSIBILITY_SERVICE) as? AccessibilityManager
```
Since we're accessing the AccessibliityManager via a Java API, we can't assert that it is non-null, so we have to assume nullability which would allow us to write a test to cover that branch.

This is a hacked up version of a test that should work, but can probably be written better:
```kotlin
@Test
fn `screenReaderSupport when AccessibiltyManager is not available`() {
    // Assume AccessibilityManager is null
    val spyContext = spy(context)
    `when`(spyContext?.getAccessibilityManager()).thenReturn(null)
    assertNull(spyContext?.isScreenReaderEnabled())

    // Assume AccessibilityManager is NOT null
    val accessibilityManager: AccessibilityManager = mock(AccessibilityManager::class.java)
    `when`(spyContext?.getAccessibilityManager()).thenReturn(accessibilityManager)
    `when`(accessibilityManager.isTouchExplorationEnabled).thenReturn(true)
    val res = spyContext?.isScreenReaderEnabled()
    assertNotNull(res)
    assertTrue(res ?: false)
}
```

SubhrajyotiSen:
Thanks a lot. It makes it much clearer now

SubhrajyotiSen:
@jonalmeida Thanks for your clarification. Coverage is fine now

components/support/ktx/src/test/java/mozilla/components/support/ktx/android/content/ContextTest.kt
```diff
@@ -6,29 +6,50 @@ package mozilla.components.support.ktx.android.content
...
-            context.getSystemService(Context.ACTIVITY_SERVICE),
-            context.systemService(Context.ACTIVITY_SERVICE))
+                context.getSystemService(Context.ACTIVITY_SERVICE),
+                context.systemService(Context.ACTIVITY_SERVICE))
 
         assertEquals(
-            context.getSystemService(Context.LOCATION_SERVICE),
-            context.systemService(Context.LOCATION_SERVICE))
+                context.getSystemService(Context.LOCATION_SERVICE),
+                context.systemService(Context.LOCATION_SERVICE))
```
pocmo: nit: All those whitespaces changes seem unnecessary?

components/support/ktx/src/main/java/mozilla/components/support/ktx/android/content/Context.kt
```diff
@@ -24,6 +25,8 @@ inline fun <reified T> Context.systemService(name: String): T {
     return getSystemService(name) as T
 }
 
+fun Context.getAccessibilityManager() = getSystemService(Context.ACCESSIBILITY_SERVICE) as? AccessibilityManager
```
pocmo: Let's add some KDoc?

Also this could either be private or move inside isScreenReaderEnabled(). It looks like we do not want to expose this publicly?

components/support/ktx/src/main/java/mozilla/components/support/ktx/android/content/Context.kt
```diff
@@ -33,3 +36,9 @@ fun Context.isOSOnLowMemory(): Boolean {
         activityManager.getMemoryInfo(memoryInfo)
     }.lowMemory
 }
+
+/**
+ * AccessibilityManager.isAccessibilityEnabled can be enabled for more things than just VoiceView. If we want
+ * VoiceView, we only need to check this one field; via comments on https://stackoverflow.com/a/12362545/582004.
+ */
+fun Context.isScreenReaderEnabled() = getAccessibilityManager()?.isTouchExplorationEnabled
```
pocmo: This method returns an nullable Boolean. This makes it awkward to use. I guess if there's no accessibility manager available then we can just return false?

components/support/ktx/src/test/java/mozilla/components/support/ktx/android/content/ContextTest.kt
```diff
@@ -44,4 +65,49 @@ class ContextTest {
 
         assertEquals(extensionFunctionResult, normalMethodResult)
     }
+
+    @Test
+    fun `isScreenReaderEnabled() returns true when TouchExploration enabled`() {
+        shadowOf(accessibilityManager).isTouchExplorationEnabled = true
+        context!!.isScreenReaderEnabled()?.let { assertTrue(it) }
```
pocmo: Tests should always make explicit checks. For this to pass `isScreenReaderEnabled` can return `null` or `true`. But the test pretends to check if it returns `true`.

components/support/ktx/src/test/java/mozilla/components/support/ktx/android/content/ContextTest.kt
```diff
@@ -44,4 +65,49 @@ class ContextTest {
...
+        val res = spyContext?.isScreenReaderEnabled()
+        assertNotNull(res)
+        assertTrue(res ?: false)
+    }
+
+    @Throws(Exception::class)
+    private fun getAccessibilityManagerInstance(): AccessibilityManager {
+        return ReflectionHelpers.callStaticMethod(AccessibilityManager::class.java, "getInstance",
+                ReflectionHelpers.ClassParameter.from(Context::class.java, RuntimeEnvironment.application))
+    }
```
pocmo: Why is reflection needed here?

components/support/ktx/src/main/java/mozilla/components/support/ktx/android/content/Context.kt
SubhrajyotiSen: Yes. It should be moved inside isScreenReaderEnabled()

components/support/ktx/src/main/java/mozilla/components/support/ktx/android/content/Context.kt
SubhrajyotiSen: That would make more sense

components/support/ktx/src/test/java/mozilla/components/support/ktx/android/content/ContextTest.kt
SubhrajyotiSen: You are right.

components/support/ktx/src/test/java/mozilla/components/support/ktx/android/content/ContextTest.kt
SubhrajyotiSen: I used the roboelctric sample as a reference https://github.com/robolectric/robolectric/blob/master/robolectric/src/test/java/org/robolectric/shadows/ShadowAccessibilityManagerTest.java

components/support/ktx/src/test/java/mozilla/components/support/ktx/android/content/ContextTest.kt
SubhrajyotiSen: Oops. My IDE did that. I will correct it

PR #793, opened 2018-09-15
Resolve #774: add lint to TaskCluster.

### LINKED ISSUE #774: Run Android Lint on TaskCluster
We currently only run `ktlint` on TaskCluster and Travis CI. We should _at least_ run android lint on TC.

This should involve only editing the `.taskcluster.yml` and/or `.travis.yml` files to include `lint` in the gradle command.

SubhrajyotiSen:


jonalmeida:
@SubhrajyotiSen Thanks for the patch! I think what's best for now, is to have only taskcluster run lint and if we see a need for it, we'll add it to Travis.

Right now, we use Travis to simply uploading our code coverage to CodeCov where our primary builds take place on TaskCluster.

PR #794, opened 2018-09-17
Support added for GeckoView error pages

keianhzo:
Added support for GeckoView category/code type errors. Also added helper creation method for the private browsing homepage. Maybe this component should be renamed to something like "browser-internalpages" so we can centralize here all the internal pages (error, private browsing, about pages, etc)?


components/browser/errorpages/build.gradle
```diff
@@ -35,6 +35,9 @@ dependencies {
     implementation Deps.support_annotations
 
     implementation Deps.kotlin_stdlib
+
+    compileOnly "org.mozilla.geckoview:geckoview-beta-armeabi-v7a:${rootProject.ext.geckoBeta['version']}"
```
jonalmeida: We want our components to stay as composable as possible, so including a specific engine as dependency for the error pages will be a problem.

Another approach would be to implement the mapping for each engine (gecko-*, system-engine) in their own components, and having them accessible via the same interface in `concept-engine`.

We could then share a unified `ErrorType` across system and gecko engines, keeping the string resources in the `errorpages` component.

To be more concrete, the steps would follow as such:
1. In `concept-engine` there would be `Engine#errorTypeFromCode`.
2. `engine-gecko-*` would implement this to provide the mapping from this patch's `ErrorPages#mapGeckoViewErrorCodeToGeckoErrorType`
3. `engine-system` would implement the `ErrorPages#mapWebViewErrorCodeToErrorType`
4. In `errorpages` we would take that ErrorType and produce the string resource
5. The other helper methods can stay in this component, but might need to be refactored to work with the above.

components/browser/errorpages/src/main/res/values/errorpages.xml
```diff
@@ -1,12 +1,8 @@
-<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
keianhzo: Fixed

components/browser/errorpages/build.gradle
keianhzo: Sorry I'm not fully familiar with the components architecture. Made the refactor, hope that looks better. Thanks for the thorough review, really appreciated.

keianhzo:
Sorry I'm not fully familiar with the components architecture yet. Made the refactor, hope that looks better. Thanks for the thorough review, really appreciated.

I've updated the RequestInterceptor too as now you can send back a URI to be shown as error page to Gecko or null to halt the load.

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
```diff
@@ -37,17 +39,117 @@ class GeckoEngine(
...
+        throw IllegalArgumentException("No error type definition for error category $errorCategory")
+    }
+
     /**
      * See [Engine.settings]
      */
     override val settings: Settings = object : Settings {
         override var javascriptEnabled: Boolean
             get() = runtime.settings.javaScriptEnabled
-            set(value) { runtime.settings.javaScriptEnabled = value }
```
jonalmeida: Let's try not to format lines that we aren't touching. We're using ktlint's formatting rules which you can apply to your problem locally with `ktlint --apply-to-idea-project --android`.

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
```diff
@@ -37,17 +39,117 @@ class GeckoEngine(
 
     override fun name(): String = "Gecko"
 
+    /**
+     * See [Engine.errorTypeFromCode]
+     */
+    override fun errorTypeFromCode(errorCode: Int, errorCategory: Int?): ErrorPages.ErrorType {
```
jonalmeida: Maybe we can move this into a helper function or function extension for `ErrorType`? It'll reduce the complexity of this class (fixing that lint error in the CI build) and also makes it easier to write tests which having to mock the entire `GeckoEngine`.

Something like: `fun ErrorType.fromErrorInt(errorCode, errorCategory)`. 

I'm bad with function naming so feel free to come up with something better!

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngine.kt
```diff
@@ -42,6 +44,49 @@ class SystemEngine(
      */
     override fun name(): String = "System"
 
+    /**
+     * See [Engine.errorTypeFromCode]
+     */
+    override fun errorTypeFromCode(errorCode: Int, errorCategory: Int?): ErrorPages.ErrorType =
+    // Chromium's mapping (internal error code, to Android WebView error code) is described at:
```
jonalmeida: Can we fix the indentation so that the comments fall in-line with the start of the `when` block?

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngine.kt
```diff
@@ -42,6 +44,49 @@ class SystemEngine(
...
+    // https://goo.gl/vspwct (ErrorCodeConversionHelper.java)
+        when (errorCode) {
+            WebViewClient.ERROR_UNKNOWN -> ErrorPages.ErrorType.UNKNOWN
+
+        // This is probably the most commonly shown error. If there's no network, we inevitably
```
jonalmeida: The same for these comments. Can we make them line up with the code lines?

components/browser/errorpages/src/main/java/mozilla/components/browser/errorpages/ErrorPages.kt
```diff
@@ -1,108 +1,176 @@
...
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 object ErrorPages {
+
     /**
      * Enum containing all supported error types that we can display an error page for.
      */
     enum class ErrorType(
```
jonalmeida: Since this `ErrorType` enum has gotten pretty large now that we've merged all the error codes together, maybe we could make this a top-level class in the file or move it to a different file? I have no strong preference between either option, but I do think it'll make it easier to read the class if we moved the enum somewhere else - that's just me though so feel free to disagree. :)

jonalmeida:
@keianhzo since this PR is untouched for a while, I hope you don't mind if I add the finishing touches to it (comments + tests) so that we can get it merged in.

PR #806, opened 2018-09-19
Document sync-logins features added in 0.4.3

thomcc:
This updates the documentation to be in line with the next version of sync-logins.

In particular, fields of `ServerPassword` that are not generally expected to be populated by users now have default values, and there are two new types of exceptions (which previously would have been emitted as a non-specific `LoginsStorageException`).

### DIFF
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -24,7 +24,7 @@ private object Versions {
     const val jna = "4.5.2"
 
     const val fxa = "0.2.4"
-    const val sync_logins = "0.4.2"
+    const val sync_logins = "0.4.3"
 }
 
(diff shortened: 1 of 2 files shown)


thomcc:
@csadilek is that an exact version test or semver?

We'll probably need to do 0.5.0, actually, because of dubious reasons. I'll push an update for it but we shouldn't land it until that build is available, obviously.

thomcc:
It's actually 0.4.3, sorry for the confusion. This can be merged now, since 0.4.3 is released. I've tested locally against the remote 0.4.3.

thomcc:
I don't think the taskcluster failure has anything to do with me?

csadilek:
@thomcc no it doesn't. TC fails frequently with network errors (when fetching dependencies). I will get this green and merged.

PR #808, opened 2018-09-19
Expose maven packages to Taskcluster Artifacts

JohanLorenzo:


JohanLorenzo:
See what could release look like in https://tools.taskcluster.net/groups/bYzy3lGJRwapL_Nj7uid9A/tasks/bYzy3lGJRwapL_Nj7uid9A/runs/0/artifacts

JohanLorenzo:
[Task group passed](https://tools.taskcluster.net/groups/VrQOyPXmQDim4WnOSk0Z2w) but still shows a red cross on Github because of https://github.com/taskcluster/taskcluster-github/issues/248

.taskcluster.yml
```diff
@@ -67,11 +68,134 @@ tasks:
...
           && git checkout {{ event.version }}
-          && ./gradlew --no-daemon clean test detektCheck ktlint assembleRelease docs lint
+          && ./gradlew --no-daemon clean test detektCheck ktlint assembleRelease docs lint uploadArchives zipMavenArtifacts
           && python automation/taskcluster/release/fetch-bintray-api-key.py
           && ./gradlew bintrayUpload --debug
       features:
         taskclusterProxy: true
+
+      artifacts:
+        public/build/lib.dataprotect.maven.zip:
```
csadilek: I assume there's isn't a way to iterate through all modules? We add new components frequently. Then we wouldn't have to remember adding them here?

.taskcluster.yml
```diff
@@ -67,11 +68,134 @@ tasks:
...
+          path: /build/android-components/components/lib/dataprotect/build/target.maven.zip
+          expires: "{{ '1 year' | $fromNow }}"
```
csadilek: Is 1 year the default? If so, can we remove the config? Technically it doesn't matter that the artifacts stay around once they're in our Maven repo, right?

.taskcluster.yml
JohanLorenzo: After talking to @MihaiTabara, this is a temporary solution while #803 is addressed. This will become more dynamic once we have what we call a "decision task".

The release task will become what a "decision task". In the context of Taskcluster, a decision task is what spawns a task group. The code of this decision task is stored in the components' github repo. We'll create a task group that contains the build+bintray task, then 1 "upload to maven.m.o" (aka beetmover) task per component. This way, if you want to add a new component, then it's a matter of updating the decision to spawn a new "beetmover" task. So yes, there is a second where you need to define that a new component was added. We can probably infer this from the tree structure. By the way, how does gradle keep track of the components @csadilek ?

This also paves the way for parallel release builds, but that's another topic.

csadilek: @JohanLorenzo OK, it's fine to add a new component to a file when it's created. We have to do that in Gradle too. I was just wondering if there's an easier way: https://github.com/mozilla-mobile/android-components/blob/master/settings.gradle

.taskcluster.yml
JohanLorenzo: You're right it is 1 year. I remember it used to be less, when we worked on Firefox Focus, 6 months ago. The documentation says the value may change again https://github.com/taskcluster/taskcluster-queue/blob/3b2960adf20ccb960a474fd68a526b4c389868b4/schemas/v1/task.yml#L132 

I'd be in favor of keeping them for a year, in case we need to audit the artifacts present on Maven. That's what we do on Fennec. 

Therefore, I think we should keep 1 year explicitly defined.

JohanLorenzo:
Yes, let's keep Bintray updated until we're 100% sure maven.mozilla.org works. Like you said on Slack, it allows us to roll out maven.m.o progressively.

From a technical perspective, I don't expect to change the `bintrayUpload` gradle task/call. This enables maven.m.o and bintray to get packages altogether.

.taskcluster.yml
JohanLorenzo: We could probably parse this file with Python. It doesn't seem too complex. What do you think @MihaiTabara ?

MihaiTabara: SGTM 👍 With the artifacts paths being so predictive, most likely we can parse the aforementioned file and:

- process the artifacts list in the "build+bintray" task
- use name + artifact to generically ramp-up another beetmover task per each of the artifacts.

Will have more info once I'm glancing better at the artifacts list (number per module, naming, etc).I'm still working on making the decision task ramp-up the build + beetmover task in my https://github.com/mozilla-mobile/android-components/pull/803

 the artifacts list within the decision task.

MihaiTabara: I'm currently tackling this problem in https://github.com/mozilla-mobile/android-components/pull/803 and after speaking with @csadilek, sounds like it might be too much effort in the end. 

If all the modules published their artifacts, we could've easily parse that `settings.gradle` file and generate that generic artifact entry. But since we have modules that don't publish, this gets slightly tricky as we need some flags to know which one is which, but we're not allowed to touch the file except maybe with comments. Overall, this beats the purpose and adds too much complexity vs. adding a new line in `.taskcluster.yml` each time a new module is added.

Agreed with @csadilek to re-evaluate in a near future, based on the future added modules, how/if problematic this will be and if it's worth optimizing. 

CC: @JohanLorenzo

MihaiTabara:
Re-iterating conversation from Slack. Was working on dynamically-generating the beetmover tasks per artifact when noticed there are only `30` module artifacts in the artifacts under `.taskcluster.yml` whereas there are `39` modules currently.

At first glance, seems that `samples-{browser, dataprotect,firefox-accounts, sync-logings,toolbar}`, `tooling-lint`, `ui-fonts`, `ui-progress` and `ui-tabcounter` are all missing. @csadilek mentioned that for `samples-*` that's expected since they aren't published. Nor is the `tooling-lint` module. But the three missing `ui-*` should be there. `ui-tabcounter` is there and the others are using `mozilla.photon` as namespace.

We might have slipped those in the first run. Will glance at this in the morning.
CC @JohanLorenzo

JohanLorenzo:
It seems I missed them when I manually transformed the output of `find . -name '*.zip'`. I confirm I do have 33 zip files locally.

PR #823, opened 2018-09-20
Closes #822: updates Telemetry error handling

### LINKED ISSUE #822: Telemetry session IllegalStateException crashes
Fire TV calls `Telemetry#recordSessionStart` in `Activity#onResume` and `Telemetry#recordSessionEnd` in `Activity#onPause`.  We're seeing a rare issue where `onPause` can sometimes be called before `onResume`, which causes the telemetry module to throw an `IllegalStateException`.  See https://github.com/mozilla-mobile/firefox-tv/issues/1159

We are seeing a relatively small number of these events (~250 / daily), so I am not particularly concerned about losing these session duration numbers but would like to prevent the hard crashes.

I have taken a look at the telemetry module, and believe there are at least a few potential solutions.
1. Update `SessionDurationMeasurement` to send an event and short circuit instead of throwing
  **Pros:** Simple. Not a breaking change.
  **Cons:** ATM the telemetry module provides tools for sending events, but doesn't send any itself. This introduces a new responsibility.  Client apps have no control over the event aside from `TelemetryConfiguration` setup.  Unlikely but possible that client code expects this exception to be thrown
2. Add overload to `Telemetry#recordSessionEnd` that accepts a callback to be executed upon failure.  Default to throwing `IllegalStateException`.
  **Pros:** Allows individual apps to define their own events. Not a breaking change.
  **Cons:** Less intuitive.  Apps that don't update to use the new method will continue to see occasional exceptions.

What option (including any I may have overlooked) would fit best with your team's philosophy?  Am I overlooking potential problems this change could introduce?  I've started a WIP branch taking approach 2, but am happy to rewrite as requested.

Baron-Severin:


pocmo:
This PR has a conflict now. I'm going to fix that and land this as PR #961.

PR #826, opened 2018-09-21
SessionStorage data format and API refactoring.

grigoryk:
A very simple (and incomplete, as is SessionManager itself) take on this, but seems like an improvement to me. See commit messages.

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
@@ -34,22 +34,54 @@ class SessionManager(
         get() = synchronized(values) { values.size }
 
     /**
-     * Returns a snapshot suitable for persisting for "session read" purposes.
+     * Produces a snapshot of this manager's state, suitable for restoring via [SessionManager.restore].
+     * Only regular sessions are included in the snapshot. Private and Custom Tab sessions are omitted.
+     *
+     * @return [SessionsSnapshot] or null if no sessions are present.
      */
-    fun createSnapshot(): SessionsSnapshot = synchronized(values) {
+    fun createSnapshot(): SessionsSnapshot? = synchronized(values) {
+        if (values.isEmpty()) {
+            return@synchronized null
+        }
+
         // Filter out CustomTab and private sessions.
         // We're using 'values' directly instead of 'sessions' to get benefits of a sequence.
-        values.asSequence()
+        val sessionStateTuples = values.asSequence()
                 .filter { !it.isCustomTabSession() }
                 .filter { !it.private }
-                .mapIndexed { index, session ->
+                .map { session ->
                     SessionWithState(
                             session,
-                            selectedIndex == index,
                             session.engineSessionHolder.engineSession
                     )
                 }
                 .toList()
+
+        // We might have some sessions (private, custom tab) but none we'd include in the snapshot.
+        if (sessionStateTuples.isEmpty()) {
+            return@synchronized null
+        }
+
+        // We need to find out the index of our selected session in the filtered list. If we have a
+        // mix of private, custom tab and regular sessions, global selectedIndex isn't good enough.
+        // We must have a selectedSession if there is at least one "regular" (non-CustomTabs) session
+        // present. Selected session might be private, in which case we reset our selection index to 0.
+        var selectedIndexAfterFiltering = 0
+        selectedSession?.takeIf { !it.private }?.let { selected ->
+            sessionStateTuples.find { it.session.id == selected.id }?.let { selectedTuple ->
+                selectedIndexAfterFiltering = sessionStateTuples.indexOf(selectedTuple)
+            }
+        }
+
+        // Sanity check to guard against producing invalid snapshots.
+        checkNotNull(sessionStateTuples.getOrNull(selectedIndexAfterFiltering)) {
+            "Selection index after filtering session must be valid"
+        }
+
+        SessionsSnapshot(
+            sessions = sessionStateTuples,
+            selectedSessionIndex = selectedIndexAfterFiltering
+        )
     }
 
     /**
@@ -128,40 +160,48 @@ class SessionManager(
 
         engineSession?.let { link(session, it) }
 
-        if (!viaRestore) {
-            notifyObservers { onSessionAdded(session) }
+        // If session is being added via restore, skip notification and auto-selection.
+        // Restore will handle these actions as appropriate.
+        if (viaRestore) {
+            return@synchronized
         }
 
-        // We're only auto-selecting session on empty selection index during a regular add (not via
-        // restore). That's because during a read, we know exactly which session will be the
-        // selected one. It does not make sense to go through selection twice in that case - when
-        // the first session was added, and then when the "selected" was added.
-        val autoSelect = !viaRestore && selectedIndex == NO_SELECTION && !session.isCustomTabSession()
-        if (selected || autoSelect) {
+        notifyObservers { onSessionAdded(session) }
+
+        // Auto-select incoming session if we don't have a currently selected one.
+        if (selected || selectedIndex == NO_SELECTION && !session.isCustomTabSession()) {
             select(session)
         }
     }
 
     /**
-     * Restores sessions from the provided snapshot.
+     * Restores sessions from the provided [SessionsSnapshot].
      * Notification behaviour is as follows:
      * - onSessionAdded notifications will not fire,
      * - onSessionSelected notification will fire exactly once if the snapshot isn't empty,
      * - once snapshot has been restored, and appropriate session has been selected, onSessionsRestored
      *   notification will fire.
-     * If a non-empty snapshot doesn't contain a selected session, first session will be selected.
+     *
+     * @param snapshot A [SessionsSnapshot] which may be produced by [createSnapshot].
+     * @throws IllegalArgumentException if an empty snapshot is passed in.
      */
-    fun restore(snapshot: SessionsSnapshot) {
-        snapshot.forEach {
-            addInternal(it.session,
-                    selected = it.selected, engineSession = it.engineSession, viaRestore = true
-            )
+    fun restore(snapshot: SessionsSnapshot) = synchronized(values) {
+        require(snapshot.sessions.isNotEmpty()) {
+            "Snapshot must contain session state tuples"
         }
 
-        // In case snapshot didn't contain a selected session, select the very first one.
-        if (selectedIndex == NO_SELECTION && snapshot.isNotEmpty()) {
-            select(snapshot[0].session)
+        // Let's be forgiving about incoming illegal selection indices, but strict about what we
+        // produce ourselves in `createSnapshot`.
+        val sessionTupleToSelect = snapshot.sessions.getOrElse(snapshot.selectedSessionIndex) {
+            // Default to the first session if we received an illegal selection index.
+            snapshot.sessions[0]
         }
+
+        snapshot.sessions.forEach {
+            addInternal(it.session, engineSession = it.engineSession, parent = null, viaRestore = true)
+        }
+
+        select(sessionTupleToSelect.session)
 
         notifyObservers { onSessionsRestored() }
     }
@@ -327,8 +367,8 @@ class SessionManager(
     fun select(session: Session) = synchronized(values) {
         val index = values.indexOf(session)
 
-        if (index == -1) {
-            throw IllegalArgumentException("Value to select is not in list")
+        require(index != -1) {
+            "Value to select is not in list"
         }
 
         selectedIndex = index
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
@@ -4,7 +4,10 @@
 
 package mozilla.components.browser.session
 
+import android.support.annotation.GuardedBy
 import mozilla.components.browser.session.engine.EngineObserver
+import mozilla.components.browser.session.storage.SessionWithState
+import mozilla.components.browser.session.storage.SessionsSnapshot
 import mozilla.components.concept.engine.Engine
 import mozilla.components.concept.engine.EngineSession
 import mozilla.components.support.base.observer.Observable
@@ -20,6 +23,8 @@ class SessionManager(
     delegate: Observable<SessionManager.Observer> = ObserverRegistry()
 ) : Observable<SessionManager.Observer> by delegate {
     private val values = mutableListOf<Session>()
+
+    @GuardedBy("values")
     private var selectedIndex: Int = NO_SELECTION
 
     /**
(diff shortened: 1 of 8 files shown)


samples/browser/src/main/java/org/mozilla/samples/browser/DefaultComponents.kt
```diff
@@ -34,13 +35,21 @@ open class DefaultComponents(private val applicationContext: Context) {
...
-        SessionManager(engine).apply {
-            sessionStorage.restore(this)
+        SessionManager(engine,
+                initialSessions = restoredSessionsOrInitial,
+                defaultSession = { Session("about:blank") }
+        )
+    }
+
+    private val restoredSessionsOrInitial by lazy {
+        val restored = sessionStorage.persistedSnapshot(engine)
```
csadilek: I think `sessionStorage.restore()` was clearer (see above).

samples/browser/src/main/java/org/mozilla/samples/browser/DefaultComponents.kt
```diff
@@ -34,13 +35,21 @@ open class DefaultComponents(private val applicationContext: Context) {
...
 
-            if (size == 0) {
-                val initialSession = Session("https://www.mozilla.org")
-                add(initialSession)
-            }
+        // If we couldn't read persisted sessions, or there were none, provide an initial session.
```
csadilek: I like the non-side-effecty persist/restore but I think this we just made harder to crasp. How about we change this to:

```Kotlin
SessionManager(engine, 
  initialSessions = sessionStorage.restore(engine), 
  defaultSession = { Session("about:blank") }
).apply {
  if (size == 0) {
    add(Session("https://www.mozilla.org"))
  }
}
```

components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
```diff
@@ -16,17 +18,50 @@ import mozilla.components.support.base.observer.ObserverRegistry
...
     private val values = mutableListOf<Session>()
+    // Access must be guarded by synchronizing on `values`.
     private var selectedIndex: Int = NO_SELECTION
 
+    init {
+        initialSessions?.forEach {
+            add(it.session, selected = it.selected, engineSession = it.engineSession)
+        }
+
+        if (defaultSession != null && size == 0) {
```
csadilek: I wouldn't apply this on init, because there the caller specifies directly what the initial sessions are. Either based on what is restored from storage or based on manually setting it (see my comment on the sample app.)

components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
```diff
@@ -16,17 +18,50 @@ import mozilla.components.support.base.observer.ObserverRegistry
...
         get() = synchronized(values) { values.size }
 
+    /**
+     * Returns a snapshot suitable for persisting for "session restore" purposes.
+     */
+    val snapshot: SessionsSnapshot
+        get() = synchronized(values) {
+            // Filter out CustomTab and private sessions.
+            // We're using 'values' directly instead of 'sessions' to get benefits of a sequence.
+            values.asSequence()
```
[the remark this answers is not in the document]
grigoryk: I've used `values` directly for the sake of some extra perf - at least that's what `sequence` claims to provide for multi-step collection manipulations. https://blog.kotlin-academy.com/effective-kotlin-use-sequence-for-bigger-collections-with-more-than-one-processing-step-649a15bb4bf

Arguably this might also be nice for other reasons - it's explicit, so you don't need to go hunting around the class while reading this code, and changes to `sessions` won't have unintended consequences on session restore.

components/browser/session/src/main/java/mozilla/components/browser/session/storage/SessionStorage.kt
```diff
@@ -4,28 +4,39 @@
...
+     * Persists the provided [SessionsSnapshot] for a given [Engine].
      *
-     * @param sessionManager the session manager to persist from.
-     * @return true if the state was persisted, otherwise false.
+     * @param engine the session in which context to persist a snapshot.
+     * @param sessions the snapshot of sessions which are to be persisted.
+     * @return true if the snapshot was persisted, otherwise false.
      */
-    fun persist(sessionManager: SessionManager): Boolean
+    fun persistSnapshot(engine: Engine, sessions: SessionsSnapshot): Boolean
```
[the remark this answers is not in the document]
grigoryk: "restore" feels a bit off to me since it's a verb, and this method no longer performs that action. I don't feel strongly about this though.

components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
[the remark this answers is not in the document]
csadilek: OK, that makes sense. It was just a nit. Marking as "resolved" :)

components/browser/session/src/main/java/mozilla/components/browser/session/storage/SessionStorage.kt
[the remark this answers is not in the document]
csadilek: `read` is another option. any verb that makes it clear it's the inverse operation of persist :).

samples/browser/src/main/java/org/mozilla/samples/browser/DefaultComponents.kt
csadilek: I should clarify my concern here. It's basically that we put 3 new concepts in front of the API user. `initialSession`, `defaultSession` and `SessionsSnapshot`. So, I was looking for a way to hide `SessionsSnapshot`...

grigoryk:
@csadilek @pocmo incorporated feedback, updated tests and rebased on top of the 'default' changes from (#933), which we're planning to land first.

grigoryk:
@pocmo hmm yeah, of course, that's no good. I'll address this and push an update.

grigoryk:
@pocmo I think that the current API is blocking in the same way as the new API! In a sense that currently if you restore a session in the `apply` block right after initializing the manager, it's blocking on the `restore` call - just like it would be if you used `initialSessions`.

I've added an `addFromSnapshot` method which takes in a `SessionsSnapshot` and adds all of the sessions+their state that's in there. This way, if a consumer wants to use session restore and not block creation of a SessionManager, they can do so - create a manager without `initialSessions`, read in the snapshot when appropriate, and use `addFromSnapshot` method to actually restore. Otherwise, just use `initialSessions` like we do in the sample browser.

What do you think?

pocmo:
> @pocmo I think that the current API is blocking in the same way as the new API! In a sense that currently if you restore a session in the apply block right after initializing the manager, it's blocking on the restore call - just like it would be if you used initialSessions.

The API is not blocking. But the way we currently use it in the sample browser is. :)

> I've added an addFromSnapshot method which takes in a SessionsSnapshot and adds all of the sessions+their state that's in there.

This sounds good. Today we looked at Rocket and that's one of the things they'd need. They launch immediately (with a then still empty session manager), load everything and then pass everything to the session manager (list + selection).

> Otherwise, just use initialSessions like we do in the sample browser.

Do we still need the initialSessions parameter then? As you said above if the app wants to block on the sessions being in the manager before continuing then it can do so already.

grigoryk:
We don't strictly need `initialSessions`, that's true. One remaining change that this needs (to `addFromSnapshot`) - from your note on the issue - "Important here is that we wouldn't notify multiple times about an add - and only after done."

pocmo:
Added some more information to the issue about the use case in Rocket.

grigoryk:
@pocmo I've addressed that feedback. Added `onSessionsRestored` observer method, tweaked when notifications are fired, added tests, etc.

grigoryk:
Taskcluster stumbled into #1083.

components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
```diff
@@ -20,6 +22,7 @@ class SessionManager(
     delegate: Observable<SessionManager.Observer> = ObserverRegistry()
 ) : Observable<SessionManager.Observer> by delegate {
     private val values = mutableListOf<Session>()
+    // Access must be guarded by synchronizing on `values`.
```
pocmo: There's `@GuardedBy` annotation that may be helpful:
https://developer.android.com/reference/android/support/annotation/GuardedBy

components/browser/session/src/main/java/mozilla/components/browser/session/storage/SessionStorage.kt
```diff
@@ -4,28 +4,39 @@
...
 import mozilla.components.browser.session.SessionManager
+import mozilla.components.concept.engine.Engine
+import mozilla.components.concept.engine.EngineSession
+
+data class SessionWithState(
+    val session: Session,
+    val selected: Boolean = false,
+    val engineSession: EngineSession? = null
+)
+typealias SessionsSnapshot = List<SessionWithState>
```
pocmo: I wonder why you added the `selected` Boolean to all `SessionWithState` objects instead of turning `SessionsSnapshot` into a class that can hold global values too. Maybe there are going to be more. :)

components/browser/session/src/main/java/mozilla/components/browser/session/storage/DefaultSessionStorage.kt
```diff
@@ -85,41 +87,43 @@ class DefaultSessionStorage(
...
-    override fun persist(sessionManager: SessionManager): Boolean {
+    override fun persist(engine: Engine, sessions: SessionsSnapshot): Boolean {
         var file: AtomicFile? = null
         var outputStream: FileOutputStream? = null
 
         return try {
             val json = JSONObject()
             json.put(VERSION_KEY, VERSION)
-            json.put(SELECTED_SESSION_KEY, sessionManager.selectedSession?.id ?: "")
+            json.put(SELECTED_SESSION_KEY, sessions.find { it.selected }?.session?.id ?: "")
```
csadilek: It seems unfortunate that we have to go search for the selected session although that is already known information. Should we just add the selected session Id to the `SessionsSnapshot`? (see @pocmo's comment below.) So, this would be another argument for turning in into a class.

components/browser/session/src/main/java/mozilla/components/browser/session/storage/SessionStorage.kt
```diff
@@ -4,28 +4,39 @@
...
  */
 interface SessionStorage {
 
     /**
-     * Persists the session state of the provided [SessionManager].
+     * Persists the provided [SessionsSnapshot] for a given [Engine].
      *
-     * @param sessionManager the session manager to persist from.
-     * @return true if the state was persisted, otherwise false.
+     * @param engine the session in which context to persist a snapshot.
```
csadilek: This line needs an update `engine the session`...

components/browser/session/src/main/java/mozilla/components/browser/session/storage/SessionStorage.kt
```diff
@@ -4,28 +4,39 @@
...
+typealias SessionsSnapshot = List<SessionWithState>
 
 /**
  * Storage component for browser and engine sessions.
  */
 interface SessionStorage {
 
     /**
-     * Persists the session state of the provided [SessionManager].
+     * Persists the provided [SessionsSnapshot] for a given [Engine].
```
csadilek: Would be nice to add that snapshot can be retrieved using [SessionManager.createSnapshot]...

components/browser/session/src/main/java/mozilla/components/browser/session/storage/SessionStorage.kt
```diff
@@ -4,28 +4,39 @@
...
+     * @param engine the session in which context to persist a snapshot.
+     * @param sessions the snapshot of sessions which are to be persisted.
+     * @return true if the snapshot was persisted, otherwise false.
      */
-    fun persist(sessionManager: SessionManager): Boolean
+    fun persist(engine: Engine, sessions: SessionsSnapshot): Boolean
 
     /**
-     * Restores the session state by reading from the latest persisted version.
+     * Returns the latest persisted sessions snapshot that may be used to restore sessions.
```
csadilek: Would be nice to link to [SessionManager.restore]

components/browser/session/src/main/java/mozilla/components/browser/session/storage/SessionStorage.kt
```diff
@@ -4,28 +4,39 @@
...
      *
-     * @param sessionManager the session manager to restore into.
-     * @return true if the state was restored, otherwise false.
+     * @param engine the engine for which to read the snapshot.
+     * @return snapshot of sessions to restore, or null if they couldn't be read.
      */
-    fun restore(sessionManager: SessionManager): Boolean
+    fun restore(engine: Engine): SessionsSnapshot?
```
csadilek: Now that `SessionManager` has a `restore` method, I prefer this one to be `read`, but that's a nit. It would give us:
```Kotlin
sessionStorage.read(engine)?.let { restore(it) }
```
vs.
```Kotlin
sessionStorage.restore(engine)?.let { restore(it) }
```
Whatever you prefer though. Just a comment.

grigoryk: Yeah, the restore & restore is a pretty weird combo - and that name isn’t factually correct, either. I’d prefer read as well.

grigoryk:
Looking a bit closer at session storage, I noticed that the storage format is wrong. It's a set of sessions, whereas it needs to be an ordered list. Currently we are depending on JSONObject.keys() to be ordered to preserve session order, and it isn't.

Unless I'm misunderstanding how this is supposed to work, I don't think we can do a sensible thing with the current format.

I propose the following changes for a `v2`:
1) move session UUID into Session's serialized representation
2) change the storage overall format to be:

```
{
 version: Int,
 selectedSessionIndex: Int
 sessionTuples: [(serializedSession, serializedEngineSession)]
}
```

v1->v2 migration looks like mapping a set to a list on first launch, and hoping for the best (as we do already!).

Snapshot creation and restoration methods in the manager would check integrity of the selection index at runtime. The rest of this refactoring would still apply.

Since we're changing a lot of this flow, I reckon this PR is as good of a place as any to change the format.

grigoryk:
I won't bother with a migration or bumping the version number then. Just changing the format should be fine. To be clear, version is already in the format, we're just not using it since there aren't any migrations in place.

grigoryk:
@pocmo @csadilek I've pushed up a revision of the storage format, along with bunch of "firm up" changes. Being explicit about the selection index in the `SessionsSnapshot` forced my hand on a lot of correctness-style changes, which I think is a good thing (removal of bunch of test code is a nice confirmation of this).

These components are slowly turning into shippable state (although session storage will see more churn).

Unfortunately, this series now diverged with the `parent` changes which I haven't noticed coming up...

grigoryk:
Rebased on top of parent changes and pushed up a patch to preserve parent information in session storage. Together with preserving session order, this should provide good basis for the "session has a parent" case.

grigoryk:
It's not strictly necessary, but it does help make things stricter internally, and I think it is helpful for consumers to be explicit about their intentions. By forcing a `clear` call, we'll make it obvious at the consumer layer what's going on.

Allowing empty snapshots will make the API less strict, which is a change we can always make in the future if it seems necessary. (much easier than going the other way).

PR #834, opened 2018-09-21
Closes #825: Session.url is not updated until page has nearly finished

### LINKED ISSUE #825: Session.url is not updated until page has nearly finished loading
slow-url

`Session.url` is updated very late in the page load process (`SystemEngineView#onPageFinished`).  In the GIF you can see that the loading URL in the bottom left shows "imdb.com" long after Rotten Tomatoes has started loading  In addition to this visual bug, this has led to problems when we compare expected and actual state (as in https://github.com/mozilla-mobile/firefox-tv/issues/1134).  

Is `Session.url` being set late intentional, and does other logic rely upon that?  If it cannot be changed, could we expose some other variable that is updated immediately?

Amejia481:
Closes #825


components/browser/engine-system/src/test/java/mozilla/components/browser/engine/system/SystemEngineViewTest.kt
```diff
@@ -73,10 +73,10 @@ class SystemEngineViewTest {
 
         engineView.currentWebView.webViewClient.onPageStarted(null, "http://mozilla.org", null)
         assertEquals(true, observedLoadingState)
+        assertEquals("http://mozilla.org", observedUrl)
 
         observedLoadingState = true
         engineView.currentWebView.webViewClient.onPageFinished(null, "http://mozilla.org")
-        assertEquals("http://mozilla.org", observedUrl)
```
pocmo: Can we use two different URLs here to make sure both changes are observed?

Amejia481: 👍

jonalmeida:
@Baron-Severin left [an interesting comment][1] in this issue as well that maybe we would also want to solve in this issue if not separate it out into two?

[1]: https://github.com/mozilla-mobile/android-components/issues/825#issuecomment-423836660

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -324,6 +327,7 @@ class GeckoEngineSession(
                 notifyObservers {
                     onProgress(PROGRESS_STOP)
                     onLoadingStateChange(false)
+                    onLoadComplete(actualURL)
```
pocmo: GeckoView already reports the URL as we expect it. Adding this new behavior here obscures the GV behavior. Especially if the URL may have changed between onPageStart() and onPageStop() then we'd report the wrong one here. I think it would be good to avoid adding logic here.

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -118,6 +118,7 @@ class SystemEngineView @JvmOverloads constructor(
                 currentUrl = url
                 session?.internalNotifyObservers {
                     onLoadingStateChange(true)
+                    onLocationChange(it)
```
pocmo: This is good and the change I would have expected here :)

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -126,10 +127,9 @@ class SystemEngineView @JvmOverloads constructor(
             url?.let {
                 val cert = view?.certificate
                 session?.internalNotifyObservers {
-                    onLocationChange(it)
```
pocmo: I think this should still be called here. We shouldn't introduce multiple ways of how the URL gets updated.

components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
```diff
@@ -51,6 +51,7 @@ class Session(
         fun onDesktopModeChanged(session: Session, enabled: Boolean) = Unit
         fun onFullScreenChanged(session: Session, enabled: Boolean) = Unit
         fun onThumbnailChanged(session: Session, bitmap: Bitmap?) = Unit
+        fun onLoadComplete(session: Session, url: String) = Unit
```
pocmo: I don't see a difference between this and `loading = false`.

I also think this breaks the contract of Session: It is expected to encapsulate the state and makes it observable. This just propagates an event and no state is saved.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
Amejia481: 👍

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
Amejia481: 👍

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
Amejia481: 👍

PR #820, opened 2018-09-20
Expose remaining maven artifacts to TaskCluster

csadilek:
There are the ones we initially forgot.

r? @JohanLorenzo @MihaiTabara

PR #836, opened 2018-09-21
Revert PR #808 and #820 to make release build pass

csadilek:
@JohanLorenzo @MihaiTabara have to temporarily revert the changes we made this week as we can't get our release build to pass. I am attaching the build.log here as well. Doesn't seem like any artifacts were generated but TC ended up in some weird low memory state (see end of log file)

[build.log.zip](https://github.com/mozilla-mobile/android-components/files/2406984/build.log.zip)

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -55,9 +55,8 @@ tasks:
     scopes:
       - "secrets:get:project/android-components/publish"
     payload:
-      maxRunTime: 7200
+      maxRunTime: 3600
       deadline: "{{ '2 hours' | $fromNow }}"
-      expires: "{{ '1 year' | $fromNow }}"
       image: 'mozillamobile/android-components:1.4'
       command:
         - /bin/bash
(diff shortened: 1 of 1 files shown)


MihaiTabara:
Thanks for the heads-up on this! Not sure why this happened, I've rebased Johan's changes in my PR ever since they landed and they work just fine on my release graphs. 

I was initially leaning towards saying this could be intermittent failure but seeing everything built green right away with changes reverted makes me reconsider. Will investigate and touch base again.

JohanLorenzo:
Thank you for letting us know! 

I'm not familiar enough with the docker-worker instances.

* Maybe all artifacts are stored in RAM before getting uploaded. However, I don't understand why I didn't see it in PR. Both PR and release run on the same containers.
* Maybe we got unlucky on Friday, and the container ran on a machine that ran too many jobs at the same time, that's why it had to dynamically shrink the RAM.
* Maybe both? 

@walac, could you point me to the specs of the github-worker worker type? Would any hypothesis explain a memory shortage?

PR #839, opened 2018-09-21
Provide extension via MIME type if none exists

gaul:
Previously `guessFileName` only replaced the extension with the MIME
type if the file name had an extension which mapped onto a valid but
different MIME type.  Suggested by:

https://issuetracker.google.com/issues/37053711

Also fix missing dot in extension, copying a fix from Android `URLUtil`.

References mozilla-mobile/focus-android#3462.


gaul:
> The change looks good to me. Can you add a test case for this to `DownloadUtilsTest`? :)

Done.

pocmo:
Thank you!

PR #850, opened 2018-09-24
Closes #828: Fullscreen mode doesn't use entire screen.

### LINKED ISSUE #828: Fullscreen window only takes up part of the screen
See mozilla-mobile/firefox-tv#938 for repro steps.

This doesn't happen for all fullscreen windows. So far we've only observed the issue with `m.youtube.com` and `bcc.com` videos.

When you enable layout bounds for FireTV devices, we can see that when a video starts playing, there is an invisible layout showing up on the screen, so when you click the video's fullscreen button, the video controls correctly take up the entire fullscreen view, but the video itself only uses the part of the screen which was seen from the invisible layout bounds.

jonalmeida:
When we add the fullscreen view into the FrameLayout, we were taking the
original WebView and hiding it with View.GONE which doesn't take up any
layout space. View.INVISIBLE doesn't show the view, but continues to
use the space that it would if it did exist (drawing in the background).

For websites like YouTube and BBC.com/ideas, the video provided to us
was of a smaller size and resolution so it didn't take the fullscreen
and ended up shrinking to it's original size view size. 🤦

### DIFF
--- a/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
+++ b/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
@@ -297,12 +297,14 @@ class SystemEngineView @JvmOverloads constructor(
 
     internal fun addFullScreenView(view: View, callback: WebChromeClient.CustomViewCallback) {
         val webView = findViewWithTag<WebView>("mosac_system_engine_webview")
-        webView?.apply { this.visibility = View.GONE }
+        val layoutParams = FrameLayout.LayoutParams(
+            FrameLayout.LayoutParams.MATCH_PARENT, FrameLayout.LayoutParams.MATCH_PARENT)
+        webView?.apply { this.visibility = View.INVISIBLE }
 
         fullScreenCallback = callback
 
         view.tag = "mosac_system_engine_fullscreen"
-        addView(view)
+        addView(view, layoutParams)
     }
 
     internal fun removeFullScreenView() {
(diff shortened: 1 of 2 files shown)


pocmo:
I saw that you call `addView(view)` for the fullscreen view. You may want to pass in some `FrameLayout.LayoutParams` as second parameter with `MATCH_PARENT` width and height?

jonalmeida:
> I saw that you call `addView(view)` for the fullscreen view. You may want to pass in some `FrameLayout.LayoutParams` as second parameter with `MATCH_PARENT` width and height?

@csadilek verified that with the layout param additions, we still have fullscreen working. 👍

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -297,12 +297,14 @@ class SystemEngineView @JvmOverloads constructor(
 
     internal fun addFullScreenView(view: View, callback: WebChromeClient.CustomViewCallback) {
         val webView = findViewWithTag<WebView>("mosac_system_engine_webview")
-        webView?.apply { this.visibility = View.GONE }
+        val layoutParams = FrameLayout.LayoutParams(
+            FrameLayout.LayoutParams.MATCH_PARENT, FrameLayout.LayoutParams.MATCH_PARENT)
+        webView?.apply { this.visibility = View.INVISIBLE }
```
pocmo: Do we still need to make the webview invisible now?

pocmo: Let's look at that in a follow-up so that we can release :)

jonalmeida: @pocmo I originally thought having the FrameLayout was enough, but it still didn't fix the issue, so instead I went back to the original way that Focus had it working.

Issue #641, opened 2018-08-21
taskcluster: Do not run for commits that include [ci skip]


pocmo:
If a commit message contains "[ci skip]" then travis won't run:
https://docs.travis-ci.com/user/customizing-the-build/

We should do something similar for taskcluster.

pocmo:
According to #taskcluster on IRC there's nothing build in for that. So we would need to do this ourselves. However that's okay - we only want that for PRs and can still build after merge and especially in releases (assuming the last commit in a release has [ci skip] because it's the changelog or whatever).

PR #862, opened 2018-09-25
Issue #641: Push is failing due to PR title check

csadilek:
Currently all our pushes are failing on TC since we landed the fix for #641. The reason is that we're using the same script but the PR_TITLE will be `None` and hence the `if SKIP_TASKS_TRIGGER in PR_TITLE:` will fail with `TypeError: argument of type 'NoneType' is not iterable`


automation/taskcluster/decision_task_pull_request.py
```diff
@@ -19,7 +19,7 @@
 REPO_URL = os.environ.get('GITHUB_HEAD_REPO_URL')
 BRANCH = os.environ.get('GITHUB_HEAD_BRANCH')
 COMMIT = os.environ.get('GITHUB_HEAD_SHA')
-PR_TITLE = os.environ.get('GITHUB_PULL_TITLE')
+PR_TITLE = os.environ.get('GITHUB_PULL_TITLE') or ""
```
ncalexan: or `.get(..., "")` -- it's a dict-like thing.

csadilek: Thanks!

PR #852, opened 2018-09-25
app-services: set version to 0.5.1

vladikoff:
@ncalexan this fails for me with:

`More than one file was found with OS independent path 'lib/arm64-v8a/libfxa_client.so`

I guess we want to sync up the `a-s` component versions, but might need to tweak something first with the `*.aar`s

```
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':samples-sync-logins:transformNativeLibsWithMergeJniLibsForDebug'.
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeActions(ExecuteActionsTaskExecuter.java:103)
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:73)
	at org.gradle.api.internal.tasks.execution.OutputDirectoryCreatingTaskExecuter.execute(OutputDirectoryCreatingTaskExecuter.java:51)
	at org.gradle.api.internal.tasks.execution.SkipUpToDateTaskExecuter.execute(SkipUpToDateTaskExecuter.java:59)
	at org.gradle.api.internal.tasks.execution.ResolveTaskOutputCachingStateExecuter.execute(ResolveTaskOutputCachingStateExecuter.java:54)
	at org.gradle.api.internal.tasks.execution.ValidatingTaskExecuter.execute(ValidatingTaskExecuter.java:59)
	at org.gradle.api.internal.tasks.execution.SkipEmptySourceFilesTaskExecuter.execute(SkipEmptySourceFilesTaskExecuter.java:101)
	at org.gradle.api.internal.tasks.execution.FinalizeInputFilePropertiesTaskExecuter.execute(FinalizeInputFilePropertiesTaskExecuter.java:44)
	at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:91)
	at org.gradle.api.internal.tasks.execution.ResolveTaskArtifactStateTaskExecuter.execute(ResolveTaskArtifactStateTaskExecuter.java:62)
	at org.gradle.api.internal.tasks.execution.SkipTaskWithNoActionsExecuter.execute(SkipTaskWithNoActionsExecuter.java:59)
	at org.gradle.api.internal.tasks.execution.SkipOnlyIfTaskExecuter.execute(SkipOnlyIfTaskExecuter.java:54)
[...]

PR #866, opened 2018-09-26
Closes #865: Use loadDataWithBaseUrl for SystemEngine..

### LINKED ISSUE #865: SystemEngineSession.loadData sets url to "data:..." instead of something user friendly
In firefox-tv, we use `SystemEngineSession.loadData` to [load error page content](https://github.com/mozilla-mobile/firefox-tv/blob/e869530c2be44344627723c83defdf3af395a2be/app/src/main/java/org/mozilla/focus/engine/CustomContentRequestInterceptor.kt#L69). However, this method sets the url to a data URI:

screenshot_1537908683

This is not very user friendly and it'd be great to override to something like, "firefox:neterror". afaik, `WebView` provides this functionality: instead of calling `loadData`, we can call [`loadDataWithBaseURL`](https://developer.android.com/reference/android/webkit/WebView#loadDataWithBaseURL(java.lang.String,%20java.lang.String,%20java.lang.String,%20java.lang.String,%20java.lang.String)) and pass in the desired URL to display. The line to change is here:
https://github.com/mozilla-mobile/android-components/blob/14fdb2d78a009bc0dc782cb28898a6f5d3128c23/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt#L71

And it's essentially a call through.

We're tracking this issue in https://github.com/mozilla-mobile/firefox-tv/issues/1193

jonalmeida:
This was added should some apps need to load data that contained 
references to relative paths and/or different history URLs.

### DIFF
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -69,7 +69,7 @@ class GeckoEngineSession(
     /**
      * See [EngineSession.loadData]
      */
-    override fun loadData(data: String, mimeType: String, encoding: String) {
+    override fun loadData(data: String, mimeType: String, encoding: String, baseUrl: String?, historyUrl: String?) {
         when (encoding) {
             "base64" -> geckoSession.loadData(data.toByteArray(), mimeType)
             else -> geckoSession.loadString(data, mimeType)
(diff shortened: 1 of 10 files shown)


jonalmeida:
> What went wrong:
> Execution failed for task ':samples-firefox-accounts:transformNativeLibsWithMergeJniLibsForDebug'.
> More than one file was found with OS independent path 'lib/arm64-v8a/libfxa_client.so'

We need to restart this build after #852 is merged.

pocmo:
Added the *release blocker* label to make a decision on whether this gets into 0.25. Will review today.

csadilek:
@jonalmeida The build problem is caused by this PR. It attempts to check in the three .so files (see diff). You will need to remove these files locally. They are not needed anymore and also are no longer on .gitignore (because they no longer need to me). Unfortunately, on our machines we have to delete these files manually, if present.

jonalmeida:
Oh shoot, sorry!

jonalmeida:
Closing this in favour of #875.

PR #882, opened 2018-09-26
Set version to 0.26

Amejia481:


pocmo:
The 0.25.1 release will cause a conflict in this PR. :)

Let's go with 0.26.0 to follow the semantic versioning we will need for maven.mozilla.org.

pocmo:
I opened #903 :)

Issue #404, opened 2018-07-11
Update to Android SDK 28


pocmo:
* Focus: https://github.com/mozilla-mobile/focus-android/issues/2909
* Fire TV: https://github.com/mozilla-mobile/firefox-tv/issues/991

pocmo:
We want to do that soon - but not too early to avoid forcing the apps to update before critical releases that may require a components update.

pocmo:
Final support libraries are not released yet. Latest version: 28.0.0-rc01.

pocmo:
Focus has moved to the new SDK.

I moved this into the 0.20 sprint to a least figure out if there are blockers. We may want to delay updating the SDK until the support libraries are released in a final version.

pocmo:
Focus has moved to the new SDK.

I moved this into the 0.20 sprint to a least figure out if there are blockers. We may want to delay updating the SDK until the support libraries are released in a final version.

pocmo:
There's a WIP PR but some libraries we depend on are not available in a final version yet. I'll move this to the next sprint.

mcomella:
FYI: apparently the public API definitions are stable in 28 and the RC is to make sure the code is bug-free:

> This release candidate of the support library is considered feature-complete and its public API surface is stable. This release will be shipped as final stable version barring any critical issues that may arise. This release should be safe to use in production. Please report any issues to the public issue tracker. 

via https://developer.android.com/topic/libraries/support-library/revisions

Driving me nuts we can't use the real KTX yet. :P

pocmo:
28.0.0 is out now! :)

Will update to the new SDK here.

Additionally filed #842 to update to androix - The new namespace is stable now too. This may happen in the next sprint though.

pocmo:
Backing out and moving to the next sprint.

PR #893, opened 2018-09-27
Issue #404: Build with and target SDK 28 (+ update support libraries).

pocmo:
Cherry-picking the SDK 28 patch to re-apply.

### DIFF
--- a/components/browser/menu/src/test/java/mozilla/components/browser/menu/BrowserMenuBuilderTest.kt
+++ b/components/browser/menu/src/test/java/mozilla/components/browser/menu/BrowserMenuBuilderTest.kt
@@ -28,7 +28,7 @@ class BrowserMenuBuilderTest {
         val recyclerView: RecyclerView = popup.contentView.findViewById(R.id.mozac_browser_menu_recyclerView)
         assertNotNull(recyclerView)
 
-        val recyclerAdapter = recyclerView.adapter
+        val recyclerAdapter = recyclerView.adapter!!
         assertNotNull(recyclerAdapter)
         assertEquals(2, recyclerAdapter.itemCount)
     }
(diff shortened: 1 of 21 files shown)


pocmo:
This should be ready to land now. I rebased the PR and fixed some small things. Let's see what taskcluster says.

pocmo:
The decision task is failing for some reason. I'll try to debug this.

pocmo:
Apart from the build issue (#1337) some new tests need to be fixed.

pocmo:
Okay, this should pass now. :)

components/browser/menu/src/test/java/mozilla/components/browser/menu/BrowserMenuBuilderTest.kt
```diff
@@ -28,7 +28,7 @@ class BrowserMenuBuilderTest {
         val recyclerView: RecyclerView = popup.contentView.findViewById(R.id.mozac_browser_menu_recyclerView)
         assertNotNull(recyclerView)
 
-        val recyclerAdapter = recyclerView.adapter
+        val recyclerAdapter = recyclerView.adapter!!
```
Amejia481: What's the difference to change it **!!**? Just to be more explicit about nullability?

components/service/telemetry/src/test/java/org/mozilla/telemetry/measurement/OperatingSystemVersionMeasurementTest.java
```diff
@@ -25,7 +25,7 @@ public void testValue() {
 
         final String version = (String) value;
         assertFalse(TextUtils.isEmpty(version));
-        assertEquals("27", version); // 16 is the default API level of this Robolectric version
+        assertEquals("28", version);
```
Amejia481: Maybe we can have a constant for that, this way we have only one point of changing

components/browser/menu/src/test/java/mozilla/components/browser/menu/BrowserMenuBuilderTest.kt
pocmo: They added @Nullable in the latest version of the support library. Now we need to deal with that :)

Amejia481: 👍

PR #895, opened 2018-09-27
Closes #874: browser-engine-system: Rename strings file.

### LINKED ISSUE #874: translatable=false strings in components/browser/engine-system/src/main/res/values/strings.xml
`components/browser/engine-system/src/main/res/values/strings.xml` has a file path that looks like a translatable file, but only contains non-localizable ones.

@pocmo, should we rename that file?

pocmo:
Background: We are going to reserve the name "strings.xml" for translatable strings. This avoids that
our L10N tools need to be able to parse and process all kinds of resources that are not strings.


pocmo:
@csadilek I just renamed the file - but I wonder whether we need this preference code at all - we do not actually read the app preferences, I guess?

csadilek:
Yeah, makes sense, the names are quite generic, but we should provide an API to customize the pref -> category mapping.

csadilek:
Or we can detach this from prefs altogether and just expose API to enable/disable categories. Seemed quite nice to support prefs, but maybe it's too much.

pocmo:
> Or we can detach this from prefs altogether and just expose API to enable/disable categories. Seemed quite nice to support prefs, but maybe it's too much.

Yeah, there may not even be a 1:1 mapping. For example Fire TV just enables a specific set when the turbo mode is enabled, while focus currently has settings for all categories - it's app specific.

PR #896, opened 2018-09-27
fxa: add pairing and QR code scanning.

vladikoff:


samples/firefox-accounts/build.gradle
```diff
@@ -36,7 +36,12 @@ android {
 }
 
 dependencies {
-    implementation project(':service-firefox-accounts')   
+    implementation project(':service-firefox-accounts')
+
+    implementation 'com.android.support.constraint:constraint-layout:1.1.3'
+    implementation ('me.dm7.barcodescanner:zxing:1.9') {
```
vladikoff: This will probably have a follow up PR to start moving this out of `samples` somehow

samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/MainActivity.kt
```diff
@@ -28,7 +28,7 @@ open class MainActivity : AppCompatActivity(), LoginFragment.OnLoginCompleteList
     companion object {
         const val CLIENT_ID = "12cc4070a481bc73"
         const val REDIRECT_URL = "fxaclient://android.redirect"
-        const val CONFIG_URL = "https://latest.dev.lcip.org"
+        const val CONFIG_URL = "https://pairsona.dev.lcip.org"
```
vladikoff: Before merging this will need to be reverted

samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/MainActivity.kt
```diff
@@ -45,6 +45,14 @@ open class MainActivity : AppCompatActivity(), LoginFragment.OnLoginCompleteList
                 Config.custom(CONFIG_URL).then { value: Config ->
                     val acct = FirefoxAccount(value, CLIENT_ID, REDIRECT_URL)
                     account = acct
+
+                    val pairingUrl = getIntent().getExtras().getString("pairingUrl") ?: ""
```
vladikoff: @csadilek would you like to use anything different here from `getIntent().getExtras()` 

Currently the flow is the following:

```java
MainActivity ------> ScanActivity ----- [set extras] ---------V
      ^                                                                                              
      |-------------------------------------------------------V
```

samples/firefox-accounts/build.gradle
pocmo: Do we know how big the library is (including its dependencies)? There used to be methodcount.com which was able to provide that number easily - but it seems to be down. :(

If it's quite big then I wonder if we should make this an supplemental "add-on" library?

samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/MainActivity.kt
```diff
@@ -45,6 +45,14 @@ open class MainActivity : AppCompatActivity(), LoginFragment.OnLoginCompleteList
...
+                    if (pairingUrl != "") {
```
csadilek: pairingUrl can't be null here. So, this check shouldn't be necessary.

samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/ScanActivity.kt
```diff
@@ -0,0 +1,48 @@
...
+import android.os.Bundle
+import android.support.v4.app.ActivityCompat
+import android.support.v4.content.ContextCompat
+import android.util.Log
+import com.google.zxing.Result
+import me.dm7.barcodescanner.zxing.ZXingScannerView
+
+class ScanActivity : AppCompatActivity(), ZXingScannerView.ResultHandler  {
+
+    private var mScannerView: ZXingScannerView? = null
```
csadilek: Let's make this a `lateinit` var, then we don't need the `!!` or `?` operators every time.

samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/ScanActivity.kt
```diff
@@ -0,0 +1,48 @@
...
+            ActivityCompat.requestPermissions(this, arrayOf(Manifest.permission.CAMERA), 1)
+        }
+
+        mScannerView = ZXingScannerView(this)
+        setContentView(mScannerView)
+    }
+
+    public override fun onResume() {
+        super.onResume()
+        mScannerView!!.setResultHandler(this)
```
csadilek: remove `!!` and make `lateinit` var.

samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/ScanActivity.kt
```diff
@@ -0,0 +1,48 @@
...
+        mScannerView!!.startCamera()
+    }
+
+    public override fun onPause() {
+        super.onPause()
+        mScannerView!!.stopCamera()
```
csadilek: same as above to remove `!!`

samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/MainActivity.kt
csadilek: Seems OK to me! I would probably introduce a constant for "pairingUrl" though.

samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/MainActivity.kt
vladikoff: to land this we can also introduce 2 URLs

samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/MainActivity.kt
vladikoff: from mtg: let's try to use `let` instead

vladikoff:
@csadilek updated!

pocmo:
Yep, definitely!

PR #900, opened 2018-09-27
Closes #876: Lifecycle bound observers should pause/resume.

### LINKED ISSUE #876: Observers bound to a LifecycleOwner should pause when the lifecycle is paused
In apps that used `LiveData` before and now register an `Observer` from the components there's a behavior change now: LiveData pauses observing while the app is in the background (paused state) and resumes it once the app comes to the foreground (caching the last value). Our observer does not do that and continues to fire updates.

That's a problem for apps that do things like fragment transactions in the callback and can't do that while the app is in the background.

It would be nice if our observers would have a similar behavior if they are bound to a `LifecycleOwner`.

Related is #665: LiveData passes the last value to the observer on resume.

csadilek:
First stab at this which pauses/resumes observers. When an observer is resumed only the last value/notification is sent to get it up-to-date. I want to do more testing but putting it up for review / discussion.

### DIFF
--- a/components/feature/tabs/src/test/java/mozilla/components/feature/tabs/tabstray/TabsTrayPresenterTest.kt
+++ b/components/feature/tabs/src/test/java/mozilla/components/feature/tabs/tabstray/TabsTrayPresenterTest.kt
@@ -256,7 +256,7 @@ private class MockedTabsTray : TabsTray {
 
     override fun register(observer: TabsTray.Observer) {}
 
-    override fun register(observer: TabsTray.Observer, owner: LifecycleOwner) {}
+    override fun register(observer: TabsTray.Observer, owner: LifecycleOwner, autoPause: Boolean) {}
 
     override fun register(observer: TabsTray.Observer, view: View) {}
 
@@ -267,4 +267,8 @@ private class MockedTabsTray : TabsTray {
     override fun notifyObservers(block: TabsTray.Observer.() -> Unit) {}
 
     override fun <R> wrapConsumers(block: TabsTray.Observer.(R) -> Boolean): List<(R) -> Boolean> = emptyList()
+
+    override fun pauseObserver(observer: TabsTray.Observer) {}
+
+    override fun resumeObserver(observer: TabsTray.Observer) {}
 }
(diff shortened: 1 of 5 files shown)


csadilek:
@pocmo this is updated now to:

- Remove the functionality to deliver the last notification ON_RESUME
- Add a new autoPause flag (default is false) that needs to be provided to turn on this functionality for individual observers.

components/feature/tabs/src/test/java/mozilla/components/feature/tabs/tabstray/TabsTrayPresenterTest.kt
```diff
@@ -256,7 +256,7 @@ private class MockedTabsTray : TabsTray {
 
     override fun register(observer: TabsTray.Observer) {}
 
-    override fun register(observer: TabsTray.Observer, owner: LifecycleOwner) {}
+    override fun register(observer: TabsTray.Observer, owner: LifecycleOwner, autoPlay: Boolean) {}
```
pocmo: Here it is called `autoPlay` :)

components/support/base/src/main/java/mozilla/components/support/base/observer/ObserverRegistry.kt
```diff
@@ -44,18 +37,14 @@ class ObserverRegistry<T> : Observable<T> {
...
         lifecycleObservers[observer] = lifecycleObserver
 
         owner.lifecycle.addObserver(lifecycleObserver)
     }
 
-    /**
-     * Registers an observer to get notified about changes.
-     *
-     * The observer will automatically unsubscribe if the provided view gets detached.
-     */
```
pocmo: We remove all the kdoc here because the method is documented in the interface? Do you know how dokka will handle that?

components/feature/tabs/src/test/java/mozilla/components/feature/tabs/tabstray/TabsTrayPresenterTest.kt
csadilek: Oops. Fixed.

components/support/base/src/main/java/mozilla/components/support/base/observer/ObserverRegistry.kt
csadilek: Yes, exactly. JavaDocs and Dokka will correctly show the docs on the implementations as well. IDEs also. So, we don't need to repeat the docs on the implementation.

csadilek:
@pocmo OK, updated, thanks.

components/support/base/src/main/java/mozilla/components/support/base/observer/ObserverRegistry.kt
pocmo: Nice :)

PR #902, opened 2018-09-28
Update Docker image to version 1.5.

pocmo:
Updated Docker image to include the latest dependencies. Hopefully this will reduce some of the "couldn't resolve dependency" issues we are seeing. Our dependencies will change and taskcluster will need to download new ones over time - but this will happen slowly and as soon as it gets worse we can update the image again.


pocmo:
This seems to speed up things too :)

PR #912, opened 2018-09-29
Closes #911 - ui-autocomplete: Do not restart input method in deleteS…

### LINKED ISSUE #911: Keyboard still desyncs in InlineAutocompleteEditText
I'm still seeing #619:
> We've encountered desyncs between the soft keyboard and our InlineAutocompleteEditText on certain devices. Specifically, this happens when autocompleted text is deleted.

In particular, the STR:
- Tap the toolbar
- Either:
  - Type s, type backspace
  - Type st, type backspace x2
- Repeat previous step, selecting one of these at random

It's intermittent and takes me 3-4 times to reproduce.

---

I'm looking at the `deleteSurroundingText` method to investigate why our previous fix worked. I feel suspicious that we're restarting the input method at all:
https://github.com/mozilla-mobile/android-components/blob/c67f8813a5e4c300359563da6ddf8b8e128ebd0e/components/ui/autocomplete/src/main/java/mozilla/components/ui/autocomplete/InlineAutocompleteEditText.kt#L480-L481

The reasoning in the code comments is:
```kotlin
// Make the IME aware that we interrupted the deleteSurroundingText call,
// by restarting the IME.
```

However, the return value for `deleteSurroundingText` means: "true on success, false if the input connection is no longer valid." Being theoretical here, if the input connection is no longer valid, the input method will stop taking actions until the input connection is recreated. To me, this seems like a very similar effect to directly restarting the input method.

Relatedly, it seems like restarting the input method and invalidating the input connection could create a race condition, which seems like the problem we were experiencing in #619 because of the intermittent nature of the restart. Our fix, posting the restart of the input method, may reduce the occurrence of a race with a negative outcome because it prevents the two from starting simultaneously.

In practice, I don't see any negative effects on my Android emulator or on the Fire device by removing the input method restart nor do I see any tests covering this to verify why it's important.

I'll do some more blame spelunking to figure out why this was added initially (it appears to be from focus).

mcomella:
…urroundingText.

This fix is speculative. There seems to be theoretical reason we need
to restart the input method and I think it's creating a race condition
that causes the keyboard to stop working on some devices. In practice,
removing this call fixes the bug but otherwise doesn't seem to make a
difference in autocomplete behavior on device or in the emulator.

### DIFF
--- a/components/ui/autocomplete/src/main/java/mozilla/components/ui/autocomplete/InlineAutocompleteEditText.kt
+++ b/components/ui/autocomplete/src/main/java/mozilla/components/ui/autocomplete/InlineAutocompleteEditText.kt
@@ -468,18 +468,6 @@ open class InlineAutocompleteEditText @JvmOverloads constructor(
                     // If we have autocomplete text, the cursor is at the boundary between
                     // regular and autocomplete text. So regardless of which direction we
                     // are deleting, we should delete the autocomplete text first.
-                    // Make the IME aware that we interrupted the deleteSurroundingText call,
-                    // by restarting the IME.
-                    post {
-                        // On some devices this method can lead to race conditions (see Javadoc on
-                        // [InputConnection#deleteSurroundingText], and in at least one case this
-                        // caused our soft keyboard to desynchronize from its backing EditText.
-                        //
-                        // It is not known exactly how this method interacts with platform code
-                        // to create problems, but posting has resolved the reported issue.
-                        val imm = ctx.getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager
-                        imm.restartInput(this@InlineAutocompleteEditText)
-                    }
                     return false
                 }
                 return super.deleteSurroundingText(beforeLength, afterLength)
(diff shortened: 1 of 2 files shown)


pocmo:
@mcomella There's a ktlint issue preventing this from landing (unused import).

mcomella:
TaskCluster failure:

> "\nSchema Validation Failed!\nRejecting Schema: https://schemas.taskcluster.net/docker-worker/v1/payload.json#\nErrors:\n  * data should NOT have additional properties: \"deadline\""

But I didn't change TaskCluster; will try re-running.

mcomella:
It sounds like the TaskCluster schema changed: https://mozilla.logbot.info/taskcluster/20181002#c15405235 and https://groups.google.com/forum/#!topic/mozilla.tools.taskcluster/OHTPYz_X_gI

I think they said they're going to roll this change back and make a more formal announcement email but we can fix this by removing `deadline` from our TaskCluster configs.

mcomella:
Actually, they're not going to rollback but there's a fix in https://github.com/mozilla-mobile/android-components/pull/951

mcomella:
@pocmo Ready to merge, assuming travis passes too.

pocmo:
Awesome, thanks!

Issue #899, opened 2018-09-27
Provide access to default WebView user agent for mutation to custom user agent


mcomella:
In firefox-tv, we take the system's default user agent string and modifying it for our purposes: https://github.com/mozilla-mobile/firefox-tv/blob/591dc154df2b1de023336e0a8ec7a0d6701aa75f/app/src/main/java/org/mozilla/focus/browser/UserAgent.kt#L59

However, there is not a good way to access the system user agent at settings init time:
- DefaultSettings sets `userAgentString` to false if not provided
- `SystemEngine.settings` is not implemented so I can't get the value immediately after creating `DefaultSettings`

We mutate the system's user agent in order to provide a user agent specific to the version of the WebView that's installed while enabling users to have the experience we want them to have (based on how servers serve different content to different user agents). It's also helpful for debugging (i.e. to find out which version of WebView users are running).

csadilek:
The solution in this case was to use the static `WebSettings.getDefaultUserAgent(context)` instead of `userAgentString` on a `WebSettings` instance. This isn't instance-specific anyway so the static method makes more sense and our `DefaultSettings` can then be initialized with this value or a customized version of it.

This is the relevant code in FireTV: https://github.com/mozilla-mobile/firefox-tv/blob/7877efdd6cc5961cd90eb7447b16dbbe07f22db5/app/src/main/java/org/mozilla/focus/browser/UserAgent.kt#L71 which needed to change to use the static method instead.

However, we shouldn't override the WebView default if no `userAgentString` default is provided on the component level. So, let's keep this issue to fix this, and maybe bring in customization functionality as well.

mcomella:
FYI that WebSettings.getDefaultUserAgent causes the tests to fail since it isn't mocked by robolectric so I had to wrap it in a function (i.e. a provider) that I mutate at runtime during testing to a innocuous version. Ew. :(

csadilek:
OK, that isn't great, static methods are never fun when testing. We can improve this in the next release.

mcomella:
Changed to a different methodology: we have the production function in `FocusApplication.getSystemUserAgent` and override this in `TestFocusApplication`, used automatically by mockito.

PR #931, opened 2018-10-01
Fixes #854 and #899: Adds remote debugging enabled setting and allows customization of user agent string

### LINKED ISSUE #854: Engine: Remote debugging mode
Expose flag/setting for enabling remote debugging mode (engine-wide).

Gecko:

```Kotlin
GeckoRuntimeSettings.setRemoteDebuggingEnabled()
```

WebView:
```
WebView.setWebContentsDebuggingEnabled()
```

csadilek:
Contains two commits to close both #854 and #899.

Similar to engine-gecko, we now have to differentiate between engine and engine session settings (for both remoteDebugging and the global user agent string). The user agent string now defautls to WebView's default, if not provided, and to the user's default, if provided. It can also be read and altered i.e.

```Kotlin
// Using WebView's default
val engine = SystemEngine(context)
```

```Kotlin
// Using customized WebView default
val engine = SystemEngine(context)
engine.settings.userAgentString = buildUserAgentString(engine.settings.userAgentString)
```

```Kotlin
// Using custom default
val engine = SystemEngine(context, userAgentString="foo")
```


mcomella:
@csadilek Sorry for the late notice 😛 but the solution presented may be insufficient for our use case.
```kotlin
engine.settings.userAgentString = buildUserAgentString(engine.settings.userAgentString)
```

This assumes that we want to mutate the current engine's user agent (i.e. System or Gecko) as opposed to always mutating the SystemEngine's default user agent (e.g. if we want to ship GeckoView but with a WebView-based user agent). We're not shipping GV right now so I don't know our actual requirements but just FYI.

I think there's nothing to do for a-c right now: we'll just modify and pass in the modified WebView user agent by accessing `WebSettings` directly in our code.

csadilek:
@mcomella yes, setting a modified SystemEngine (WebView) user agent on a GeckoEngine is not something we have considered at the time. I am also not sure how that should work given that you  wouldn't have both `engine-system` and `engine-gecko` on the classpath, and we'd need that to read the defaults. I guess we could move that part of engine-system to a different (support) module.

However, as you suggested:
```kotlin
(gecko)engine.settings.userAgentString = buildUserAgentString(WebSettings.getDefaultUserAgent(context))
``` 
should work for now.

PR #933, opened 2018-10-01
Closes #764: Add 'defaultSession' to SessionManager

### LINKED ISSUE #764: Improving SessionManager's interface
(cross-posted, with slight modifications, from https://github.com/mozilla-mobile/fenix/issues/63)

There is a bit of an "action at a distance" problem with how session manager's state as it is perceived by various components. A specific example is the tabs tray interaction with the toolbar in Fenix. If one tab is currently opened (one session is present in the manager), closing that tab will leave session manager with no sessions.
However, toolbar component assumes presence of a selected session - so if you were to open a url, it'll crash as it hits a null session.

At first glance, this seems like a problem at an interface level between the session manager and its consumers. There are two methods to get a session, one of which will throw if set of sessions is empty. So the implicit contract of the manager to its consumers is "you decide if I may be empty". Inevitably this leads to problems like the one I described above, where two parts of the system make different assumptions.

As consuming application grows in complexity, I imagine these kinds of issues will keep popping up, and if not addressed at a lower level will require bunch of cumbersome state management boiler plate code.

I'd be nice if the interface between components made it impossible to represent an illegal state.

One way to address this would be to enforce a behaviour at the manager level - i.e. a "must have a default session" type of a session manager, which would be responsible for enforcing the correct behaviour. ISTM that we'd need this anyway for something like a "new tab" panel, but more generally it should help avoid having components step on each other's toes.

This kind of pattern likely generalizes to other types of "in-between" components.

grigoryk:
Fixed #764. Split from #826.

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
@@ -16,6 +16,7 @@ import mozilla.components.support.base.observer.ObserverRegistry
 @Suppress("TooManyFunctions")
 class SessionManager(
     val engine: Engine,
+    val defaultSession: (() -> Session)? = null,
     delegate: Observable<SessionManager.Observer> = ObserverRegistry()
 ) : Observable<SessionManager.Observer> by delegate {
     private val values = mutableListOf<Session>()
@@ -158,6 +159,13 @@ class SessionManager(
 
         notifyObservers { onSessionRemoved(session) }
 
+        // NB: we're not explicitly calling notifyObservers here, since adding a session when none
+        // are present will notify observers with onSessionSelected.
+        if (selectedIndex == NO_SELECTION && defaultSession != null) {
+            add(defaultSession.invoke())
+            return
+        }
+
         if (selectionUpdated && selectedIndex != NO_SELECTION) {
             notifyObservers { onSessionSelected(selectedSessionOrThrow) }
         }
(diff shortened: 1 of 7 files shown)


components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
```diff
@@ -173,18 +181,31 @@ class SessionManager(
         }
 
         selectedIndex = NO_SELECTION
-        notifyObservers { onAllSessionsRemoved() }
+
+        if (defaultSession != null) {
+            add(defaultSession.invoke())
+        } else {
+            notifyObservers { onAllSessionsRemoved() }
+        }
```
pocmo: That's the part I'm not sure about: An observer would never see the removal, just an add. Why wouldn't we first notify about removal and then add the new session?

csadilek:
@grigoryk the test `default session not used when all sessions were removed and they were all CustomTab` seems to have failed on Travis.

components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
grigoryk: We could! With a default session present, we're never really in an "empty" state that's observable to the outside world - that's why I didn't notify observers. Reflecting on the docstring of "all sessions removed" callback, it seems that the intent there isn't to indicate an empty manager, but to indicate that one of the blanket remove* functions were invoked. That interpretation would dictate that yeah, they should be notified here.

grigoryk: ISTM that the observer "onAllSessionsRemoved" method should be named differently, since it doesn't actually fire when "all sessions are removed". Filed a follow up #948.

grigoryk:
@csadilek not sure why, I think I forgot to push something since it is passing locally. I've addressed @pocmo question and changed the behaviour to fire the "onAllSessionsRemoved" callbacks, and filed a follow-up to clarify this (I think it's a confusing API). See #948.

... and now TC is failing for entirely unrelated reasons :-)

PR #936, opened 2018-10-02
Closes #923: SystemEngineSession.exitFullScreenMode() is a no-op...

### LINKED ISSUE #923: SystemEngineSession.exitFullScreenMode() is a no-op
I wonder why `SystemEngineSession.exitFullScreenMode()` is not implemented. I was expecting us to remove the full screen view and notifying the callback.

Downstream issue: https://github.com/mozilla-mobile/firefox-tv/issues/1224

jonalmeida:
This was an error where we weren't calling the WebView's fullscreen view
callback in the EngineSession#exitFullScreenMode.

### DIFF
--- a/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
+++ b/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
@@ -297,7 +297,7 @@ class SystemEngineSession(private val defaultSettings: Settings? = null) : Engin
      * See [EngineSession.exitFullScreenMode]
      */
     override fun exitFullScreenMode() {
-        // no-op
+        view?.get()?.fullScreenCallback?.onCustomViewHidden()
     }
 
     override fun captureThumbnail(): Bitmap? {
--- a/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
+++ b/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
@@ -332,7 +332,6 @@ class SystemEngineView @JvmOverloads constructor(
         val view = findViewWithTag<View>("mosac_system_engine_fullscreen")
         val webView = findViewWithTag<WebView>("mosac_system_engine_webview")
         view?.let {
-            fullScreenCallback?.onCustomViewHidden()
             webView?.apply { this.visibility = View.VISIBLE }
             removeView(view)
         }
--- a/components/browser/engine-system/src/test/java/mozilla/components/browser/engine/system/SystemEngineSessionTest.kt
+++ b/components/browser/engine-system/src/test/java/mozilla/components/browser/engine/system/SystemEngineSessionTest.kt
@@ -4,6 +4,7 @@ import android.content.Context
 import android.graphics.Bitmap
 import android.net.Uri
 import android.os.Bundle
+import android.webkit.WebChromeClient
 import android.webkit.WebResourceRequest
 import android.webkit.WebSettings
 import android.webkit.WebStorage
@@ -643,5 +644,17 @@ class SystemEngineSessionTest {
 
         `when`(webView.drawingCache).thenReturn(ShadowBitmap.createBitmap(10, 10, Bitmap.Config.RGB_565))
         assertNotNull(engineSession.captureThumbnail())
+    }
+
+    @Test
+    fun testExitFullscreenModeWithWebViewAndCallBack() {
+        val engineSession = SystemEngineSession()
+        val engineView = SystemEngineView(RuntimeEnvironment.application)
+        val customViewCallback = mock(WebChromeClient.CustomViewCallback::class.java)
+
+        engineView.render(engineSession)
+        engineView.fullScreenCallback = customViewCallback
+        engineSession.exitFullScreenMode()
+        verify(customViewCallback).onCustomViewHidden()
     }
 }


jonalmeida:
Coverage from running jacoco locally is very confusing from what Codecov is showing.

Locally, having just this one test increases the branch coverage to 5/6 branches, but _adding_ another test, reduces it to 2/6 branches.

Codecov on the other hand, shows both tests as providing zero coverage (???).

This is the test that decreases coverage locally:
```kotlin
@Test
fun testExitFullscreenModeWithWebViewAndCallBack() {
    val engineSession = SystemEngineSession()
    val engineView = SystemEngineView(RuntimeEnvironment.application)
    val customViewCallback = mock(WebChromeClient.CustomViewCallback::class.java)

    engineView.fullScreenCallback = customViewCallback
    engineView.render(engineSession)

    engineSession.exitFullScreenMode()
    verify(customViewCallback).onCustomViewHidden()
}
```

Looking for advice; I've spent too long on this already.

@pocmo / @csadilek

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
```diff
@@ -304,7 +304,7 @@ class SystemEngineSession(private val defaultSettings: Settings? = null) : Engin
      * See [EngineSession.exitFullScreenMode]
      */
     override fun exitFullScreenMode() {
-        // no-op
+        view?.get()?.fullScreenCallback?.onCustomViewHidden()
```
Baron-Severin: I'm no expert on this codebase, but it looks like the only other place this method is called (in `SystemEngineView#createWebChromeClient`, override `onHideCustomView` calls `removeFullScreenView`), a few other things happen at the same time.

- `onFullScreenChange` observer is notified
- View with tag 'mosac_system_engine_fullscreen' is removed
- View with tag 'mosac_system_engine_webview' is made visible

Should we be doing any of these here as well?

csadilek:
@jonalmeida I hit the codecov 0% issue before. It's because it just sees the same line hit before and after. So, if you were to write your diff like this, it would fix it:

```Kotlin
 view?.
  get()?.
  fullScreenCallback?.
  onCustomViewHidden()
```

I don't recommend doing that, but this will give you a >0% diff hit.

On the coverage for this in general.
```Kotlin
 view?.get()?.fullScreenCallback?.onCustomViewHidden()
```
has a few branches that aren't hit. `View` (null and not null), `get()` (null and not null), `fullScreenCallback` (null and not null).

I can't reproduce the problem of locally reduced coverage. Add your test definitely increases coverage for me.

jonalmeida:
@pocmo I'm not sure where exactly we should add the `exitFullScreenMode` function call to since the back button already respects this interaction natively. 

I think the main issue I didn't consider, is that FireTVs 'back' isn't the same as a regular system back button press. I'm not sure if that's the real fix that FireTV should be making, but regardless this issue should be fixed.

jonalmeida:
@csadilek In that case, I'm going to remove my commit that removes the extra test I wrote even though Codecov is acting funny.

I'll add some tests in some other part of the codebase to make up for the lost code coverage.  ;)

jonalmeida:
So.. this isn't good. I'm restarting the build for now, but I'm hoping this isn't an intermittent issue or a low memory thing?
```
:browser-engine-system:testDebugUnitTest

mozilla.components.browser.engine.system.SystemEngineSessionTest > testExitFullscreenModeWithoutWebViewAndCallBack FAILED
    java.lang.StackOverflowError

mozilla.components.browser.engine.system.SystemEngineSessionTest > testFindAll FAILED
    java.lang.IllegalStateException

mozilla.components.browser.engine.system.SystemEngineSessionTest > testWebChromeClientNotifiesObservers FAILED
    java.lang.IllegalStateException

mozilla.components.browser.engine.system.SystemEngineSessionTest > testDefaultSettings FAILED
    java.lang.IllegalStateException

mozilla.components.browser.engine.system.SystemEngineSessionTest > testClearDataMakingExpectedCalls FAILED
    java.lang.IllegalStateException

mozilla.components.browser.engine.system.SystemEngineSessionTest > testOnLoadRequestWithoutInterceptor FAILED
    java.lang.IllegalStateException

mozilla.components.browser.engine.system.SystemEngineSessionTest > testCaptureThumbnail FAILED
    java.lang.IllegalStateException

mozilla.components.browser.engine.system.SystemEngineSessionTest > testDisableTrackingProtection FAILED
    java.lang.IllegalStateException

mozilla.components.browser.engine.system.SystemEngineSessionTest > testLoadUrl FAILED
    java.lang.IllegalStateException

mozilla.components.browser.engine.system.SystemEngineSessionTest > testFindNext FAILED
    java.lang.IllegalStateException

mozilla.components.browser.engine.system.SystemEngineSessionTest > testEnableTrackingProtection FAILED
    java.lang.IllegalStateException

mozilla.components.browser.engine.system.SystemEngineSessionTest > testGoForward FAILED
    java.lang.IllegalStateException

mozilla.components.browser.engine.system.SystemEngineSessionTest > testInitSettings FAILED
    java.lang.IllegalStateException

mozilla.components.browser.engine.system.SystemEngineSessionTest > testOnLoadRequestWithInterceptorThatDoesNotIntercept FAILED
    java.lang.IllegalStateException

mozilla.components.browser.engine.system.SystemEngineSessionTest > testSaveState FAILED
    java.lang.IllegalStateException

mozilla.components.browser.engine.system.SystemEngineSessionTest > desktop mode FAILED
    java.lang.IllegalStateException

mozilla.components.browser.engine.system.SystemEngineSessionTest > testGoBack FAILED
    java.lang.IllegalStateException

mozilla.components.browser.engine.system.SystemEngineSessionTest > testReload FAILED
    java.lang.IllegalStateException

mozilla.components.browser.engine.system.SystemEngineSessionTest > testSharedFieldsAreVolatile FAILED
    java.lang.IllegalStateException

mozilla.components.browser.engine.system.SystemEngineTest > testSettings FAILED
    java.lang.IllegalStateException

mozilla.components.browser.engine.system.SystemEngineTest > testName FAILED
    java.lang.IllegalStateException

mozilla.components.browser.engine.system.SystemEngineTest > testCreateView FAILED
    java.lang.IllegalStateException

mozilla.components.browser.engine.system.SystemEngineTest > testCreateSession FAILED
    java.lang.IllegalStateException

73 tests completed, 23 failed
:browser-engine-system:testDebugUnitTest FAILED
```

pocmo:
> @pocmo I'm not sure where exactly we should add the exitFullScreenMode function call to since the back button already respects this interaction natively.

In `BrowserFragment.onBackPressed()` I think. Currently we seem to just navigate back in history instead of only closing fullscreen mode?

jonalmeida:
> In `BrowserFragment.onBackPressed()` I think. Currently we seem to just navigate back in history instead of only closing fullscreen mode?

I better understand the problem with FireTV now after testing it out on a device. The system back button does work correctly as expected, however, when you're in fullscreen mode and you enter the "overlay mode" they have a UI back button that needs to detect their in fullscreen mode (which they can since there is an observable) and then call `exitFullScreenMode` for the **overlay back button**.

I still need to fix this weird test framework failure.

jonalmeida:
Using firefox-tv/#1251 I was able to verify that my fix does actually work.

jonalmeida:
@pocmo yes!

Issue #260, opened 2018-05-30
In "Icons" readme, provide sample images of included icons


mcomella:
Workaround is to clone components and open each XML file in Android Studio, which is doable but a pain.

PR #937, opened 2018-10-02
Added Icons PNGs and Updated Icons ReadMe. Fixes Issue #260

wajahatkarim3:
This Pull Request adds:
* Icon PNGs of Vector XMLs for easily viewing
* Updated Icons Read me by adding a list of all icons with links to the respective vector XML files
* Fixes issue #260


pocmo:
We have [a website](https://mozilla-mobile.github.io/android-components/) now (we didn't at the time the issue was filed). So I wonder if there are ways to render the drawable XML files directly with the help of some JavaScript.

wajahatkarim3:
That's a good idea. I think I will look into it and see if anything can work about it. 👍

pocmo:
Awesome, thank you! :)

Alex Lockwood as written a bunch of tools that work with vector drawables and SVGs, maybe there's something helpful:
https://github.com/alexjlockwood

pocmo:
Closing this for now. Let's re-open if you find some way to make this easier to update and keep in sync. :)

PR #946, opened 2018-10-02
Issue #370: browser-engine-gecko-*: Add default runtime as default pa….

### LINKED ISSUE #730: browser-engine-gecko-*: Add default runtime as default parameter to Engine
GeckoEngine requires a runtime:

```Kotlin
val runtime = GeckoRuntime.getDefault(applicationContext)
// Create an engine instance to be used by other components.
val engine = GeckoEngine(runtime)
```

`GeckoRuntime.getDefault(applicationContext)` should be the default value of the runtime parameter. Only if an app needs a non default runtime it should need to specify that.

fercarcedo:
Closes #730

### DIFF
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
@@ -17,8 +17,9 @@ import org.mozilla.geckoview.GeckoRuntime
  * Gecko-based implementation of Engine interface.
  */
 class GeckoEngine(
-    private val runtime: GeckoRuntime,
-    private val defaultSettings: Settings? = null
+    context: Context,
+    private val defaultSettings: Settings? = null,
+    private val runtime: GeckoRuntime = GeckoRuntime.getDefault(context)
 ) : Engine {
 
     /**
--- a/samples/browser/src/geckoNightly/java/org/mozilla/samples/browser/Components.kt
+++ b/samples/browser/src/geckoNightly/java/org/mozilla/samples/browser/Components.kt
@@ -18,8 +18,6 @@ class Components(private val applicationContext: Context) : DefaultComponents(ap
         val defaultSettings = DefaultSettings().apply {
             requestInterceptor = SampleRequestInterceptor()
         }
-
-        val runtime = GeckoRuntime.getDefault(applicationContext)
-        GeckoEngine(runtime, defaultSettings)
+        GeckoEngine(applicationContext, defaultSettings)
     }
 }
--- a/components/browser/engine-gecko-beta/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineTest.kt
+++ b/components/browser/engine-gecko-beta/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineTest.kt
@@ -4,6 +4,7 @@
 
 package mozilla.components.browser.engine.gecko
 
+import android.content.Context
 import mozilla.components.concept.engine.DefaultSettings
 import mozilla.components.concept.engine.EngineSession.TrackingProtectionPolicy
 import mozilla.components.concept.engine.UnsupportedSettingException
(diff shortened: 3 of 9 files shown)


pocmo:
#370 is a PR and not an issue :)

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
```diff
@@ -21,6 +21,8 @@ class GeckoEngine(
     private val defaultSettings: Settings? = null
 ) : Engine {
 
+    constructor(context: Context, defaultSettings: Settings? = null) : this(GeckoRuntime.getDefault(context), defaultSettings)
+
```
pocmo: Ah, we can't do that in the existing constructor because we do not have a `Context` here. Mhh, maybe it's not worth adding another constructor for this then. @csadilek what do you think?

csadilek: @pocmo as we just added a `context` parameter to `SystemEngine`, we could now do the same here. I think that would be good, as it would make the API symmetric.

So, we wouldn't need a separate constructor but change the existing one:
```Kotlin
class GeckoEngine(
    context: Context,
    private val runtime: GeckoRuntime = GeckoRuntime.getDefault(context),
    private val defaultSettings: Settings? = null
)
```

Then constructing engines will be the same in both cases:

```Kotlin
engine = SystemEngine(context)
engine = SystemEngine(context, DefaultSettings(...))
engine = GeckoEngine(context)
engine = GeckoEngine(context, DefaultSettings(...))
```

Wdyt?

pocmo: Sounds good!

@fercarcedo Can you update the PR to add this new constructor? :)

fercarcedo: Updated :)

csadilek: @fercarcedo Thank you! We have code for Gecko build variants in the our sample-browser app (beta|nightly|release) that needs to be updated now though to use this new constructor (pass in the context):

https://github.com/mozilla-mobile/android-components/blob/master/samples/browser/src/geckoBeta/java/org/mozilla/samples/browser/Components.kt#L24

https://github.com/mozilla-mobile/android-components/blob/master/samples/browser/src/geckoNightly/java/org/mozilla/samples/browser/Components.kt#L23

https://github.com/mozilla-mobile/android-components/blob/master/samples/browser/src/geckoRelease/java/org/mozilla/samples/browser/Components.kt#L23

Would you mind updating these as well?

fercarcedo: Yes, those files were also updated as part of the PR :) (along with the tests)

samples/browser/src/geckoNightly/java/org/mozilla/samples/browser/Components.kt
```diff
@@ -20,6 +20,6 @@ class Components(private val applicationContext: Context) : DefaultComponents(ap
         }
 
         val runtime = GeckoRuntime.getDefault(applicationContext)
-        GeckoEngine(runtime, defaultSettings)
+        GeckoEngine(applicationContext, runtime, defaultSettings)
```
csadilek: Since we now no longer need to specify the runtime, we have two options:

```Kotlin
GeckoEngine(context = applicationContext, defaultSettings = defaulSettings)
```

or we could change the parameter ordering:

```Kotlin
GeckoEngine(applicationContext, defaultSettings)
```

csadilek: Since we now no longer need to specify the runtime, we have two options:

```Kotlin
GeckoEngine(context = applicationContext, defaultSettings = defaulSettings)
```

or we could change the parameter ordering:

```Kotlin
GeckoEngine(applicationContext, defaultSettings)
```

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
csadilek: Perfect, thanks!

samples/browser/src/geckoNightly/java/org/mozilla/samples/browser/Components.kt
fercarcedo: Changed

fercarcedo:
Solved :)

PR #949, opened 2018-10-02
Issue #843: Fretboard: Deprecate JobSchedulerSyncScheduler

### LINKED ISSUE #843: Fretboard: Deprecate JobSchedulerSyncScheduler
Now that the work manager implementation has landed we can deprecate JobSchedulerSyncScheduler.

fercarcedo:
Closes #843

### DIFF
--- a/components/service/fretboard/README.md
+++ b/components/service/fretboard/README.md
@@ -13,7 +13,7 @@ implementation "org.mozilla.components:fretboard:{latest-version}"
 ```
 
 ### Creating Fretboard instance
-In order to use the library, first you have to create a new `Fretboard` instance. You do this once per app launch 
+In order to use the library, first you have to create a new `Fretboard` instance. You do this once per app launch
 (typically in your `Application` class `onCreate` method). You simply have to instantiate the `Fretboard` class and
 provide the `ExperimentStorage` and `ExperimentSource` implementations, like this:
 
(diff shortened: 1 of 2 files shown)


Amejia481:
I believe we also should update the documentation to use  WorkManagerSyncScheduler too!

> The second one is to use the provided JobScheduler-based scheduler, like this:
https://github.com/mozilla-mobile/android-components/blob/master/components/service/fretboard/README.md#updating-experiment-list

pocmo:
The change looks good to me. However I added the "blocked" label because I think we are not ready to do that just yet: We haven't fully migrated to androidx yet and this can cause problems for apps still on the "old" support libraries.

grigoryk:
@fercarcedo how well did you test the WorkManager implementation across different Android API levels (it'll use different scheduling strategies under the hood)? I'm seeing some issues on certain devices while using it with sync manager stuff (https://github.com/mozilla-mobile/reference-browser/pull/425), while at the same time it works great on other devices/api versions.

Don't have a clear picture yet, but just wanted to call it out.

fercarcedo:
For my testing I used the latest API (28) for the JobScheduler-based implementation, and Android 4.4 with and without Play Services, but I'll take a look at it

fercarcedo:
I think I have found a possible explanation to the issues: if enqueue is called on app start, maybe more than one sync worker is being scheduled at the same time. I have created a pull request setting the work as unique: [https://github.com/mozilla-mobile/android-components/pull/1900](https://github.com/mozilla-mobile/android-components/pull/1900)

PR #951, opened 2018-10-02
Move deadline to task definition per TC example

vladikoff:
Per https://docs.taskcluster.net/docs/tutorial/hello-world

This might be a better fix than https://github.com/mozilla-mobile/android-components/pull/950


vladikoff:
@csadilek r?

PR #956, opened 2018-10-03
Issue #122: JobSchedulerSyncScheduler: Make job id configurable

### LINKED ISSUE #122: JobSchedulerTelemetryScheduler: Make job id configurable
Original issue: https://github.com/mozilla-mobile/telemetry-android/issues/24

Currently we just use `42`. However the job id must be unique across all clients of the same uid (not just the same package). As it's pretty much impossible to guess a number that won't be used by the application, we should make this configurable by the client.

fercarcedo:
Closes #122

### DIFF
--- a/components/service/telemetry/src/main/java/org/mozilla/telemetry/schedule/jobscheduler/JobSchedulerTelemetryScheduler.java
+++ b/components/service/telemetry/src/main/java/org/mozilla/telemetry/schedule/jobscheduler/JobSchedulerTelemetryScheduler.java
@@ -8,7 +8,6 @@ import android.app.job.JobInfo;
 import android.app.job.JobScheduler;
 import android.content.ComponentName;
 import android.content.Context;
-import android.os.PersistableBundle;
 
 import org.mozilla.telemetry.config.TelemetryConfiguration;
 import org.mozilla.telemetry.schedule.TelemetryScheduler;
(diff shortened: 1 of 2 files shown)


components/service/telemetry/src/main/java/org/mozilla/telemetry/schedule/jobscheduler/JobSchedulerTelemetryScheduler.java
```diff
@@ -34,4 +35,8 @@ public void scheduleUpload(TelemetryConfiguration configuration) {
 
         scheduler.schedule(jobInfo);
     }
+
+    public void setJobId(int jobId) {
+        this.jobId = jobId;
+    }
```
pocmo: As this ID is only meant to be set once I wonder whether this could be a constructor parameter (secondray, we still need to keep the default constructor with no parameters). Too bad that this isn't Kotlin. :)

components/service/telemetry/src/main/java/org/mozilla/telemetry/schedule/jobscheduler/JobSchedulerTelemetryScheduler.java
```diff
@@ -17,13 +16,23 @@
  * TelemetryScheduler implementation that uses Android's JobScheduler API to schedule ping uploads.
  */
 public class JobSchedulerTelemetryScheduler implements TelemetryScheduler {
-    public static final int JOB_ID = 42; //  TODO: Make this configurable (Issue #24)
+    private static final int DEFAULT_JOB_ID = 42;
+
+    private int jobId;
```
pocmo: nit: Can be `final` now? :)

components/service/telemetry/src/test/java/org/mozilla/telemetry/schedule/jobscheduler/JobSchedulerTelemetrySchedulerTest.java
```diff
@@ -0,0 +1,35 @@
...
+public class JobSchedulerTelemetrySchedulerTest {
+    @Test
+    public void checkId() {
+        Context context = RuntimeEnvironment.application;
+        int jobId = 27;
+        JobSchedulerTelemetryScheduler scheduler = new JobSchedulerTelemetryScheduler(jobId);
+        scheduler.scheduleUpload(new TelemetryConfiguration(context));
+        JobScheduler jobScheduler = context.getSystemService(JobScheduler.class);
+        List<JobInfo> pendingJobs = jobScheduler.getAllPendingJobs();
+        assertEquals(1, pendingJobs.size());
```
pocmo: nit: Let's add an assert that checks that its 0 before :)

fercarcedo: Changed :)

components/service/telemetry/src/main/java/org/mozilla/telemetry/schedule/jobscheduler/JobSchedulerTelemetryScheduler.java
fercarcedo: Changed :)

PR #977, opened 2018-10-05
Issue #788: Add StrictMode resetAfter/use extensions

### LINKED ISSUE #788: Add StrictMode resetAfter/use extensions
Like `Closeable.use`, this helps you to remember to reset the StrictMode state after making a brief exception. For example, 
```kotlin
            StrictMode.allowThreadDiskReads().resetAfter {
                FormattedDomain.format(holder.itemView.context, itemURI, false, 0)
            }
```

This is currently used in Fire TV: https://github.com/mozilla-mobile/firefox-tv/blob/1abbaeb4a7d27b6500212114dff0d453d734e4eb/app/src/main/java/org/mozilla/focus/home/pocket/PocketVideoFragment.kt#L151-L153

And defined here: https://github.com/mozilla-mobile/firefox-tv/blob/c660e1938eb0f112379e38c8af8a3f892ed9ba1f/app/src/main/java/org/mozilla/focus/ext/StrictMode.kt#L16

fercarcedo:
Closes #788

### DIFF
--- a/components/support/ktx/src/test/java/mozilla/components/support/ktx/android/os/StrictModeTest.kt
+++ b/components/support/ktx/src/test/java/mozilla/components/support/ktx/android/os/StrictModeTest.kt
@@ -1,0 +1,31 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.support.ktx.android.os
+
+import android.os.StrictMode
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNotEquals
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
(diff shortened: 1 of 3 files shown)


pocmo:
> Merging #977 into master will decrease coverage by 0.05%.

Unfortunately jacoco can't see the usage of inlined functions in tests.

components/support/ktx/src/test/java/mozilla/components/support/ktx/android/os/StrictModeTest.kt
```diff
@@ -0,0 +1,31 @@
...
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNotEquals
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
+class StrictModeTest {
+    @Test
+    fun `strict mode policy should be restored`() {
```
mcomella: We should also ensure the thread policy is restored if the function block throws an error. :)

Issue #925, opened 2018-10-01
Cannot use dependencies with Java 8 bytecode


pocmo:
```
The dependency contains Java 8 bytecode. Please enable desugaring by adding the following to build.gradle
android {
    compileOptions {
        sourceCompatibility 1.8
        targetCompatibility 1.8
    }
}
```

mcomella:
Are bintray assets distributed by bytecode or source? I assumed the latter but it seems I might be mistaken. It seems like it's a problem for us as an open source organization if they're distributed by bytecode (in addition to problems like these).

pocmo:
AARs contain compiled bytecode. There are a lot of reasons why this is good. Imagine having to build GeckoView from source :)

The problem here is solvable by adding the `compileOptions` above and letting the Android's build tools take care of de-sugaring:
https://developer.android.com/studio/write/java8-support

In the case here we actually own the dependency (`ServoView`) and could change this upstream as well.

What other problems are you seeing?

mcomella:
> AARs contain compiled bytecode. There are a lot of reasons why this is good. Imagine having to build GeckoView from source :)

I talked to Nick about this: see https://github.com/mozilla-mobile/android-automation-tools/issues/21

> What other problems are you seeing?

The problem I was concerned with boils down to the problem [reproducible builds](https://en.wikipedia.org/wiki/Deterministic_compilation) is trying to solve. https://github.com/mozilla-mobile/android-automation-tools/issues/21 covers that so no more concerns from me on this bug.

pocmo:
GeckoView switched to Java 8. We need to follow and apply the same changes:
https://bugzilla.mozilla.org/show_bug.cgi?id=1485439

pocmo:
It may be enough to just add this to `browser-engine-gecko-nightly` and `samples-browser` for now. This avoids that this replicates to all components.

PR #978, opened 2018-10-05
DO NOT MERGE: Set sourceCompatibility and targetCompatibility to Java 1.8.

pocmo:
Do not merge this. Just trying to see if everything still passes with that.

#925


pocmo:
Closing this. I'll try to only add this to the modules that are affected (browser-engine-gecko-nightly and the sample browser).

PR #998, opened 2018-10-05
Closes #997: SystemEngine fails to initialize with Robolectric

### LINKED ISSUE #997: SystemEngine fails to initialize with Robolectric
This was caused by the fix for #899. Since we now call `WebSettings.getDefaultUserAgent` when initializing the `SystemEngine`. In our Robolectric tests, we can set the internal `defaultUserAgent` to avoid this problem, but we didn't make this accessible for consumer tests.

```
java.lang.NullPointerException
	at android.webkit.WebViewFactory.getWebViewContextAndSetProvider(WebViewFactory.java:313)
	at android.webkit.WebViewFactory.getProviderClass(WebViewFactory.java:383)
	at android.webkit.WebViewFactory.getProvider(WebViewFactory.java:194)
	at android.webkit.WebSettings.getDefaultUserAgent(WebSettings.java:1259)
	at mozilla.components.browser.engine.system.SystemEngine$Companion.initDefaultUserAgent(SystemEngine.kt:88)
	at mozilla.components.browser.engine.system.SystemEngine$Companion.access$initDefaultUserAgent(SystemEngine.kt:83)
	at mozilla.components.browser.engine.system.SystemEngine.<init>(SystemEngine.kt:27)
```

While this is just a testing problem, it prevents upgrading FireTV.

csadilek:


pocmo:
Does this need a changelog entry? 🙃

csadilek:
@pocmo changelog entry added :)

pocmo:
> Not really happy about having this visible and apps having to set that for tests - but seems to be the best we can do right now.

I guess the actual fix should be in Roboletric. :)

csadilek:
@pocmo yeah, I think it's best we can do. I thought of lazily initializing the `defaultUserAgent` i.e. only call `getDefaultUserAgent` if no `DefaultSetting` for it was provided. Then we could simply add `DefaultSetting(userAgentString = "test-ua")` when testing, but this will break things again as we want to support reading the "system default" when no settings was provided....see #899 .

Issue #454, opened 2018-07-19
Use JEXL/CEL in fretboard to evaluate experiments?


pocmo:
In Fretboard/Switchboard we sometimes had the problem that we couldn't have more complex configurations because we can only match against certain properties (with Regex) and not write more complex rules involving `AND`, `OR`, `>`, `<`. Something like "Users with version 58 and higher" is already ridiculously complex when trying to express this as a regular expression that we match against the actual version number.

The desktop version of Firefox (SHIELD, Normandy) uses JEXL (Javascript Expression Language) for that purpose.

* https://github.com/TechnologyAdvice/Jexl
* Mozilla JS JEXL: https://github.com/mozilla/mozjexl
* Mozilla Python JEXL: https://github.com/mozilla/pyjexl
* ~~Apache has something: (Is this actually the same?) https://commons.apache.org/proper/commons-jexl/download_jexl.cgi~~ (This is another JEXL)

pocmo:
https://remote-settings.readthedocs.io/en/latest/target-filters.html

fercarcedo:
I can take a look at this and modify the evaluator

pocmo:
@fercarcedo I think this is a rather big project. As far as I can see there's no implementation of JEXL in Java/Kotlin and this alone is a project in itself. Let's focus on the other fretboard tasks first. :)

pocmo:
Another option may be "Common Expression Language" (CEL):
* https://github.com/google/cel-spec
* https://github.com/google/cel-cpp

mostlygeek:
I also opened this on our side: https://github.com/mozilla/product-delivery/issues/6 to track feedback / comments.

pocmo:
I have been using some time here and there to write a Kotlin version of mozjexl. It's quite far already. I may hack a bit on that during my PTO next week because it's fun. I'll move this issue into the next sprint to share progress.

PR #1001, opened 2018-10-08
Issue #454: Add lib-jexl: Javascript Expression Language - Context-based expression parser and evaluator.

pocmo:
Needs some more tests and there are detekt/ktlint issues to fix.

In addition to that there are some things I would like to make nicer / clean up. But because of the size I would prefer to do that after landing the initial version.


automation/taskcluster/artifacts.yml
```diff
@@ -3,6 +3,11 @@
           name: lib-dataprotect
           old_artifact_id: dataprotect
           new_group_id: org.mozilla.components
+        public/build/lib.dataprotect.maven.zip:
```
pocmo: Oops. Popy & Caste error.

components/lib/jexl/src/main/java/mozilla/components/lib/jexl/value/JexlValue.kt
```diff
@@ -0,0 +1,304 @@
...
+ * A JEXL value type.
+ */
+sealed class JexlValue {
+    abstract operator fun plus(other: JexlValue): JexlValue
+    abstract operator fun times(other: JexlValue): JexlValue
+    abstract operator fun div(other: JexlValue): JexlValue
+    abstract operator fun compareTo(other: JexlValue): Int
+
+    abstract fun toBoolean(): Boolean
+    abstract fun toKotlin(): Any
```
csadilek: I found `toKotlin` a bit confusing. Maybe just `value` (to map the constructor param)?

components/lib/jexl/src/main/java/mozilla/components/lib/jexl/value/JexlValue.kt
```diff
@@ -0,0 +1,304 @@
...
+
+    override fun toString() = value.toString()
+
+    override fun hashCode() = value.hashCode()
+}
+
+/**
+ * JEXL Double type.
+ */
+class JexlDouble(val value: Double): JexlValue() {
```
csadilek: Looks like a `JexlNumber` base class would make sense to reduce the duplication in `div/plus/minus/times`.

components/lib/jexl/src/main/java/mozilla/components/lib/jexl/lexer/Lexer.kt
```diff
@@ -0,0 +1,220 @@
...
+            }
+
+            input.proceed()
+        }
+
+        input.proceed()
+
+        val raw = input.emit()
+
+        // TODO: Validate string is closed?
```
csadilek: Maybe a follow-up issue?

components/lib/jexl/src/main/java/mozilla/components/lib/jexl/lexer/LexerInput.kt
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        return value.substring(position, position + steps)
+    }
+
+    fun character(): Char = value[position]
+
+    fun end() = position == value.length
+
+    fun next(): Char = if (last()) ' ' else value[position + 1]
```
csadilek: `next` and `last` seem like good candidates to increase code coverage (easy to test).

components/lib/jexl/src/main/java/mozilla/components/lib/jexl/ast/AstNode.kt
```diff
@@ -0,0 +1,143 @@
...
+/**
+ * A node of the abstract syntax tree.
+ *
+ * This class has a lot of properties because it needs to represent all types of nodes (See [AstType]). It should be
+ * possible to make this much simpler by creating a dedicated class for each type.
+ */
+class AstNode(
+    val type: AstType,
+    var value: Any? = null,
+    val operator: String? = null,
```
csadilek: Operator should probably be typed as well e.g. an enum as there's a fixed set of supported operators...

components/lib/jexl/src/main/java/mozilla/components/lib/jexl/parser/Parser.kt
```diff
@@ -0,0 +1,250 @@
...
+import mozilla.components.lib.jexl.ast.AstType
+import mozilla.components.lib.jexl.grammar.Grammar
+import mozilla.components.lib.jexl.lexer.Token
+
+/**
+ * JEXL parser.
+ *
+ * Takes a list of tokens from the lexer and transforms them into an abstract syntax tree (AST).
+ */
+internal class Parser(
```
csadilek: Lexer|Parser|Evaluator are all `internal`. Should they be? What is the main API entry point for other components?

components/lib/jexl/src/main/java/mozilla/components/lib/jexl/evaluator/JexlContext.kt
```diff
@@ -0,0 +1,51 @@
...
+ *  )
+ * </code>
+ *
+ * This context can be accessed in an JEXL expression like this:
+ *
+ * <code>
+ *     employees[.age >= 30 && .age < 90][.age < 35]
+ * </code>
+ */
+class JexlContext(
```
csadilek: `toString` will be helpful here for debugging at one point.

components/lib/jexl/src/main/java/mozilla/components/lib/jexl/value/JexlValue.kt
csadilek: This would also be an easy to way to get coverage up. `JexlDouble` has no coverage right now and then would have quite a bit. ;)

components/lib/jexl/src/main/java/mozilla/components/lib/jexl/evaluator/EvaluatorHandlers.kt
```diff
@@ -0,0 +1,145 @@
...
+                is Boolean -> JexlBoolean(value)
+                else -> throw EvaluatorException("Unknown value type: ${value!!::class}")
+            }
+        },
+
+        AstType.BINARY_EXPRESSION to { evaluator, node ->
+            val left = evaluator.evaluate(node.left!!)
+            val right = evaluator.evaluate(node.right!!)
+
+            val operator = evaluator.grammar.elements[node.operator!!]!!
```
csadilek: We could fix this `[node.operator!!]!!` when we refactor `AstNode` into different types. Then can we verify that node is a BinaryExpressionNode where `operator` is not nullable.

components/lib/jexl/src/main/java/mozilla/components/lib/jexl/value/JexlValue.kt
pocmo: Yeah, I plan to raise the coverage before landing!

The `Number` refactoring is something I would like to file a follow-up for.

components/lib/jexl/src/main/java/mozilla/components/lib/jexl/ast/AstNode.kt
pocmo: This iteration does not yet, but mozjexl supports adding custom operators:
https://github.com/mozilla/mozjexl#api

components/lib/jexl/src/main/java/mozilla/components/lib/jexl/parser/Parser.kt
pocmo: The public API is the `Jexl` class which allows consumers to evaluate expressions, add custom transforms and inject a custom/extended Grammar (not implemented yet).

components/lib/jexl/src/main/java/mozilla/components/lib/jexl/evaluator/EvaluatorHandlers.kt
pocmo: Yeah, that's my biggest complaint about the current implementation too. I would like to have subtypes for `AstNode`. Some properties may still need to be nullable though as they get filled while parsing and are not available yet at the point of creating the node.

However that's a very big refactoring and I would like to do that after landing the first version. :)

pocmo:
✅ Fixed the detekt and ktlint issues.

pocmo:
Added some more tests. Locally now at 83% coverage. But more is possible. :)

pocmo:
Updated changelog, readme and added some more tests. Local coverage is at 84% now.

@csadilek I think from my side this is ready to land now.

csadilek:
🛬

PR #1003, opened 2018-10-08
Add experimental browser-engine-servo component.

### LINKED ISSUE #1002: Add experimental browser-engine-servo component
The Servo project publishes an AAR including ServoView. Let's try to wrap that in a `concept-engine` implementation.

The latest AAR can be downloaded from:
https://download.servo.org/nightly/android/servo-latest.aar

Eventually ServoView should get published on maven.mozilla.org:
https://bugzilla.mozilla.org/show_bug.cgi?id=1494913

Sample app:
https://github.com/servo/servo/tree/master/support/android/apk/servoapp

pocmo:
Closes #1002.

### DIFF
--- a/automation/taskcluster/artifacts.yml
+++ b/automation/taskcluster/artifacts.yml
@@ -53,6 +53,11 @@
           name: browser-engine-gecko-nightly
           old_artifact_id: engine-gecko-nightly
           new_group_id: org.mozilla.components
+        public/build/browser.engine-servo.maven.zip:
+          path: /build/android-components/components/browser/engine-servo/build/target.maven.zip
+          name: browser-engine-servo
+          old_artifact_id: browser-engine-servo
+          new_group_id: org.mozilla.components
         public/build/browser.engine-system.maven.zip:
           path: /build/android-components/components/browser/engine-system/build/target.maven.zip
           name: browser-engine-system
--- a/components/browser/engine-servo/README.md
+++ b/components/browser/engine-servo/README.md
@@ -1,0 +1,21 @@
+# [Android Components](../../../README.md) > Browser > Engine-Servo
+
+[*Engine*](../../concept/engine/README.md) implementation based on the [Servo Browser Engine](https://servo.org/).
+
+## Usage
+
+See [concept-engine](../../concept/engine/README.md) for a documentation of the abstract engine API this component implements.
+
+### Setting up the dependency
+
+Use gradle to download the library from JCenter:
+
+```Groovy
+implementation "org.mozilla.components:browser-engine-servo:{latest-version}
+```
+
+## License
+
+    This Source Code Form is subject to the terms of the Mozilla Public
+    License, v. 2.0. If a copy of the MPL was not distributed with this
+    file, You can obtain one at http://mozilla.org/MPL/2.0/
--- a/components/browser/engine-servo/build.gradle
+++ b/components/browser/engine-servo/build.gradle
@@ -1,0 +1,63 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+apply plugin: 'com.android.library'
+apply plugin: 'kotlin-android'
+
+android {
+    compileSdkVersion Config.compileSdkVersion
+
+    defaultConfig {
+        minSdkVersion Config.minSdkVersion
+        targetSdkVersion Config.targetSdkVersion
+    }
+
+    lintOptions {
+        warningsAsErrors true
+        abortOnError true
+    }
+
+    buildTypes {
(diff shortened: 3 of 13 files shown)


components/browser/engine-servo/src/main/java/mozilla/components/browser/engine/servo/ServoEngineSession.kt
```diff
@@ -0,0 +1,150 @@
...
+@Suppress("TooManyFunctions")
+class ServoEngineSession(
+    private val defaultSettings: Settings? = null
+) : EngineSession() {
+    override val settings: Settings
+        get() = defaultSettings ?: DefaultSettings()
+
+    private var urlToLoad: String? = null
+    private var viewReference: WeakReference<ServoEngineView>? = null
+    private val view: ServoView?
```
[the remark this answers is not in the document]
pocmo: Isn't that the same with the `view` property here?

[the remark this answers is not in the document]
jonalmeida: Yes, but since it's a private field, mocking that wouldn't be as easy, right? 

Maybe I'm just a bit more fond of our `SystemEngineSession#currentView` after trying to write a test for the exitFullScreen callback to cover all the branches with something like `weakRef?.get()?.anotherNullableObject?.invoke()`.

```kotlin
internal fun currentView(): WebView? {
        return view?.get()?.currentWebView
}
```

With that, it's easier to just mock that function and let those branches not add to the branches that we can't cover (because mocking WeakReference is hard).

This is also just opinion. 👌

components/browser/engine-servo/build.gradle
```diff
@@ -0,0 +1,63 @@
...
+    implementation Deps.kotlin_coroutines
+
+    testImplementation project(':support-test')
+
+    testImplementation Deps.testing_junit
+    testImplementation Deps.testing_robolectric
+    testImplementation Deps.testing_mockito
+}
+
+archivesBaseName = "browser-ngine-servo"
```
csadilek: there's a typo here "ngine"

components/browser/engine-servo/src/main/java/mozilla/components/browser/engine/servo/ServoEngineSession.kt
```diff
@@ -0,0 +1,150 @@
...
+
+        override fun onHistoryChanged(canGoBack: Boolean, canGoForward: Boolean) {
+            notifyObservers {
+                onNavigationStateChange(canGoBack, canGoForward)
+            }
+        }
+    }
+
+    override fun loadUrl(url: String) {
+        val view = view
```
csadilek: This seems like an unneeded assignment?

automation/taskcluster/artifacts.yml
```diff
@@ -53,6 +53,11 @@
           name: browser-engine-gecko-nightly
           old_artifact_id: engine-gecko-nightly
           new_group_id: org.mozilla.components
+        public/build/browser.engine-servo.maven.zip:
```
MihaiTabara: 👍

components/browser/engine-servo/src/main/java/mozilla/components/browser/engine/servo/ServoEngineSession.kt
pocmo: The `view` property is nullable and with the local reference I circumvent the issue that it may change after the null check. Usually we do this with `view?.let {}` but I need an `else` branch here as well.

components/browser/engine-servo/build.gradle
pocmo: Thanks, updated!

components/browser/engine-servo/README.md
```diff
@@ -0,0 +1,21 @@
+# [Android Components](../../../README.md) > Browser > Engine-Servo
+
+[*Engine*](../../concept/engine/README.md) implementation based on the [Servo Browser Engine](https://servo.org/).
```
Amejia481: Should we add some basic docs with the new component?

pocmo: Yeah, let's wait a bit. The Servo team is working an actual View/Session split so this may change some things here and there. For now no one is probably seriously going to use this component. :)

components/browser/engine-servo/src/main/java/mozilla/components/browser/engine/servo/ServoEngineSession.kt
csadilek: heh, yes, that's how we usually end up with `view?.let {} ?: (run) {}`. Still on the fence what's easier to read...

pocmo:
> heh, yes, that's how we usually end up with view?.let {} ?: (run) {}. Still on the fence what's easier to read...

It definitely took me a while to understand what you did in that line in the other PR :)

Issue #47, opened 2018-04-05
Remove 'test' prefix from test method names


csadilek:
The prefix isn't needed anymore and just makes the names harder to read.

PR #1005, opened 2018-10-08
Removed test prefix from test method names

santimar:
I renamed test methods removing the prefix, like asked in #47


santimar:
Squashed and renamed

csadilek:
Thanks @santimar. 

@pocmo OK for you as well?

PR #1009, opened 2018-10-10
Closes #988: Create feature-download (Non UI).

### LINKED ISSUE #988: Create feature-downloads (Non UI)
Breaking in small parts this issue #459

- [ ] Subscribe to downloads in the currently selected session (I assume we do not need to subscribe to it in all Sessions?)

- [ ] Consume Download objects and pass the data to Androids Download Manager

- [ ] Call back into the app to notify if a download is done

Amejia481:
#988


components/feature/download/README.md
```diff
@@ -0,0 +1,19 @@
+# [Android Components](../../../README.md) > Feature > Download
+
+Feature implementation for apps that want to use Android downloads manager.
```
pocmo: Let's link to the Android docs: [Android's Download Manager](https://developer.android.com/reference/android/app/DownloadManager)

components/feature/download/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Usage
```
pocmo: Let's try to add some basic docs with new components: Let's describe what this actually does (forward downloads requests of the currently selected session to Android's Download Manager) and how to use it.

components/feature/download/build.gradle
```diff
@@ -0,0 +1,45 @@
...
+    testImplementation Deps.testing_robolectric
+    testImplementation Deps.testing_mockito
+    testImplementation project(':support-test')
+    testImplementation project(':concept-engine')
+}
+
+archivesBaseName = "feature-download"
+
+apply from: '../../../publish.gradle'
+apply plugin: 'kotlin-android-extensions'
```
pocmo: nit: this should go to the top to the other plugins

components/feature/download/src/main/java/mozilla/components/feature/download/DownloadManager.kt
```diff
@@ -0,0 +1,139 @@
...
+import android.content.Context.DOWNLOAD_SERVICE
+
+typealias OnDownloadFinish = (Long) -> Unit
+typealias AndroidDownloadManager = android.app.DownloadManager
+
+/**
+ * Handles the interactions with the [AndroidDownloadManager].
+ * @property context A reference to [Context] applicationContext.
+ */
+class DownloadManager(private val context: Context) {
```
pocmo: nit: If we want this to be an applicationContext then let's make this a bit more obvious and name the parameter `applicationContext`.

components/feature/download/src/main/java/mozilla/components/feature/download/DownloadManager.kt
```diff
@@ -0,0 +1,139 @@
...
+
+        val fileName = getFileName(download)
+
+        val request = Request(Uri.parse(download.url))
+            .setNotificationVisibility(VISIBILITY_VISIBLE_NOTIFY_COMPLETED)
+            .setMimeType(download.contentType)
+            .addRequestHeader("User-Agent", download.userAgent)
+
+        if (refererURL.isNotEmpty())
+            request.addRequestHeader("Cookie", cookie)
```
pocmo: You are checking the `refererURL` and add the `cookie`. I guess we need to add both if they are not null?

components/feature/download/src/main/java/mozilla/components/feature/download/DownloadManager.kt
```diff
@@ -0,0 +1,139 @@
...
+
+        val downloadID: Long
+
+        request.setDestinationInExternalPublicDir(download.destinationDirectory, fileName)
+
+        downloadID = androidDownloadManager.enqueue(request)
```
pocmo: nit: It looks like downloadID can be created and assigned at the same place:

```Kotlin
val downloadID = androidDownloadManager.enqueue(request)
```

components/feature/download/src/main/java/mozilla/components/feature/download/DownloadManager.kt
```diff
@@ -0,0 +1,139 @@
...
+        queuedDownloads[downloadID] = downloadListener
+
+        if (!isSubscribedReceiver)
+            registerBroadcastReceiver(context)
```
pocmo: nit: Let's always wrap blocks in `{ }`. We may want to enable a detekt rule for that.

components/feature/download/src/main/java/mozilla/components/feature/download/DownloadManager.kt
```diff
@@ -0,0 +1,139 @@
...
+
+    private fun getFileName(download: Download): String? {
+        return if (!download.fileName.isNullOrEmpty())
+            download.fileName
+        else
+            DownloadUtils.guessFileName(
+                "",
+                download.url,
+                download.contentType
+            )
```
pocmo: I think we should file a follow-up to make sure fileName is never null. No reason to let consumers call `guessFileName`. And I think in `browser-engine-system` we already do that. We just need to make sure that the gecko-flavors do that too.

samples/firefox-accounts/build.gradle
```diff
@@ -38,7 +38,7 @@ android {
 dependencies {
     implementation project(':service-firefox-accounts')
 
-    implementation 'com.android.support.constraint:constraint-layout:1.1.3'
+    implementation Deps.support_constraintlayout
```
pocmo: nit: unrelated change. Let's always file a separate issue for that to not make the review diff too large.

components/feature/download/README.md
Amejia481: Should I also update the diagram in the main README.md?
I noticed that Intent and Tabs features aren't there too.

image

Amejia481: 👍

components/feature/download/src/main/java/mozilla/components/feature/download/DownloadManager.kt
Amejia481: Oops my mistake, thanks for catching this one!

components/feature/download/src/main/java/mozilla/components/feature/download/DownloadManager.kt
Amejia481: I'm a bit confused, can you help me to clarify please.

1.Do you want that I change Download.fileName property to Non-null? **e.g fileName: String** instead of  **fileName: String?** ?

2.Or are you referring to `private fun getFileName(download: Download): String?` type return  **String?**

pocmo: Option 1: We should always come up with a file name. But that's something for a separate issue.

Amejia481: https://github.com/mozilla-mobile/android-components/issues/1017

Amejia481:
I've added all the suggestion to the PR!

components/feature/download/src/main/java/mozilla/components/feature/download/DownloadManagerFeature.kt
```diff
@@ -0,0 +1,61 @@
...
+import mozilla.components.browser.session.Session
+import mozilla.components.browser.session.SessionManager
+
+/**
+ * Feature implementation for proving download functionality for the selected session.
+ */
+class DownloadManagerFeature(
+    private val downloadManager: DownloadManager,
+    private val sessionManager: SessionManager
+) : SessionManager.Observer, Session.Observer {
```
pocmo: As there is only one `DownloadManager` how about adding that as default parameter:

```Kotlin
class DownloadManagerFeature(
    private val downloadManager: DownloadManager = DownloadManager()
) : SessionManager.Observer, Session.Observer {
```

If you see the need for multiple download manager implementations in the future (I guess that makes sense) then I'd expect `DownloadManager` to be an interface and your implementation to have a different name (`AndroidDownloadManager` - but then your `typealias` may break 😜).

components/feature/download/src/main/java/mozilla/components/feature/download/DownloadManagerFeature.kt
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /**
+     * Notifies to the [DownloadManager] that a new download must be processed.
+     */
+    @SuppressLint("MissingPermission")
+    override fun onDownload(session: Session, download: Download): Boolean {
+        downloadManager.download(download)
+        return true
+    }
```
pocmo: We need a way to let the app request the permission only when needed.

I'm thinking of something like this:
* `onDowload()` is getting invoked
* If we do **not** have the permissions:
   * let the app know via a callback that it needs to prompt
   * Return false here so that the `Download` object will be retained
   * Add a method that the app can call when a permission was granted
   * Start the download if we have all permissions now
* If we do have the permissions:
   * Start the download immediately

If this is a too big change then let's work on that in a follow-up issue. We will need that to use such a component in an app like Focus. None of the apps will want to ask the user upfront to accept those permissions. At the time of a download coming in is the best in-context place to ask.

components/feature/download/src/main/java/mozilla/components/feature/download/DownloadManager.kt
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    /**
+     * Remove all the listeners.
+     */
+    fun unregisterListeners() {
+        applicationContext.unregisterReceiver(onDownloadComplete)
+        isSubscribedReceiver = false
+        queuedDownloads.clear()
+    }
```
pocmo: Your listener setup is very fine-grained and I wonder if we can do it simpler. The apps we have are not really interested in subscribing to specific downloads. For example Focus shows a `Snackbar` whenever a download is completed. For that some simple callback would be enough:

```
onDownloadCompleted(download: Download, ... additional meta data if needed)
```

This callback (if registered) could get invoked for all downloads and I think I'd want to register that on the `Feature` when initializing it.

Does this make sense?

components/feature/download/README.md
```diff
@@ -0,0 +1,69 @@
...
+        )
+
+        /*Adding a listener to download in process
+        * If the download I've already had a listener
+        * it will be updated with the new one*/
+        manager.registerListener(downloadId, onDownloadComplete)
+
+        /*Unregistering all listeners*/
+        manager.unregisterListeners()
+```
```
pocmo: nit: I see the `DownloadManager` class more as an implementation detail. Do you think there's a need to use it in a standalone way? Anyhow, I'd at least move this to the end and explain `DownloadManagerFeature` first. :)

components/feature/download/src/main/java/mozilla/components/feature/download/DownloadManagerFeature.kt
```diff
@@ -0,0 +1,61 @@
...
+import android.annotation.SuppressLint
+import android.support.annotation.RequiresPermission
+import mozilla.components.browser.session.Download
+import mozilla.components.browser.session.Session
+import mozilla.components.browser.session.SessionManager
+
+/**
+ * Feature implementation for proving download functionality for the selected session.
+ */
+class DownloadManagerFeature(
```
pocmo: nit: I think for consistency it would be nice to use the same name as the module: `DownloadFeature`.

Originally I was thinking of `feature-downloads` (s) and `DownloadsFeature`. But the *s* is not that important :D

components/feature/download/src/main/java/mozilla/components/feature/download/DownloadManager.kt
```diff
@@ -0,0 +1,143 @@
...
+     * @param downloadID the id that reference to the schedule download.
+     * @param listener a listener to be notify when the download is completed.
+     */
+    @Suppress("unused")
+    fun registerListener(downloadID: Long, listener: OnDownloadFinish) {
+        if (downloadID in queuedDownloads)
+            queuedDownloads[downloadID] = listener
+        else
+            throw IllegalArgumentException("Unable to find $downloadID on the queued downloads")
+    }
```
pocmo: nit: There are multiple if statements here and above without `{ }`

components/support/ktx/src/main/java/mozilla/components/support/ktx/android/content/Context.kt
```diff
@@ -33,3 +35,11 @@ fun Context.isOSOnLowMemory(): Boolean {
...
     }.lowMemory
 }
+
+/**
+ * Returns if a list of permission have been granted, if all the permission have been granted
+ * returns true otherwise false.
+ */
+fun Context.isPermissionGranted(vararg permission: String): Boolean {
+    return permission.all { checkSelfPermission(this, it) == PERMISSION_GRANTED }
+}
```
pocmo: Let's make sure to add this (and of course the new component!) to the changelog. :)

components/feature/download/src/main/java/mozilla/components/feature/download/DownloadManagerFeature.kt
Amejia481: Let's keep it simple for now :), I will the default parameter 😄

Amejia481: But I will need an extra parameter (context) for creating a` DownloadManager(private val applicationContext: Context)`

Amejia481: @pocmo it wont be worst to have an extra parameter, just for creating a default value. 

```kotlin
class DownloadManagerFeature(
    context:Context,
    private val downloadManager: DownloadManager = DownloadManager(context),
    private val sessionManager: SessionManager
) : SessionManager.Observer, Session.Observer
```

👍 
``` kotlin
       val feature = DownloadManagerFeature(context, sessionManager = sessionManager)
```

👎  the parameter is never been used
```kotlin
        val manager = DownloadManager(context)
        val feature = DownloadManagerFeature(context, manager, sessionManager)
```

Should I just add a context to the constructor and internally create the download manager? 

```kotlin
class DownloadManagerFeature(
    context: Context,
    private val sessionManager: SessionManager
) : SessionManager.Observer, Session.Observer {
    private val downloadManager: DownloadManager = DownloadManager(context)
```

pocmo: Ah, I see. Mh. Yeah, I think creating the `DownloadManager` internally may be a good workaround - as long as this doesn't affect testability? :)

Once we have the need to inject other DownloadManager things we can open that up again.

components/feature/download/src/main/java/mozilla/components/feature/download/DownloadManagerFeature.kt
Amejia481: What about this?
I add a new callback to the constructor `needToRequestPermissions`,
when I new download comes out and the app doesn't have the permissions I will call it.

When the app gets the permissions granted, the app can call the FeatureDownload.onDownload function with the same parameters that `needToRequestPermissions` gave it. 

What do you think?

```kotlin

class DownloadManagerFeature(
    private val context: Context,
    private val needToRequestPermissions: (session: Session, download: Download) -> Unit = { _, _ -> },
    private val downloadManager: DownloadManager = DownloadManager(context)
) : SessionManager.Observer, Session.Observer {

/*.....*/

 override fun onDownload(session: Session, download: Download): Boolean {
        return if (context.isPermissionGranted(INTERNET, WRITE_EXTERNAL_STORAGE)) {
            downloadManager.download(download)
            true
        } else {
            needToRequestPermissions(session, download)
            false
        }
    }
}
```

components/feature/download/src/main/java/mozilla/components/feature/download/DownloadManagerFeature.kt
Amejia481: 👍

components/feature/download/src/main/java/mozilla/components/feature/download/DownloadManagerFeature.kt
pocmo: I wonder if we can make it easier for the app and not require and parameters, something like: `onPermissionsGranted()`. Permission granting is asynchronously and it's quite some overhead for the app to keep a reference to the `Session` and `Download` across potential configuration changes etc.

Could we just check (in the callback) the selected session and see if it has a download attached and then re-trigger this ourselves?

Amejia481: Got it! 👍

components/feature/download/README.md
Amejia481: Done!
I improved the documentation and put `DownloadsFeature` first.
I added the documentation of `DownloadManager` someone has a different use case that is not cover by `DownloadsFeature`.

components/feature/download/src/main/java/mozilla/components/feature/download/DownloadManagerFeature.kt
Amejia481: ^_^
I renamed the component to **feature-downloads** and `DownloadsFeature` 🕺 .

components/feature/download/src/main/java/mozilla/components/feature/download/DownloadManager.kt
Amejia481: Done `{ }` 👍

components/support/ktx/src/main/java/mozilla/components/support/ktx/android/content/Context.kt
Amejia481: Done!
https://github.com/mozilla-mobile/android-components/pull/1032

components/support/test/build.gradle
```diff
@@ -24,6 +24,7 @@ android {
 
     lintOptions {
         lintConfig file("lint.xml")
+        disable 'InvalidPackage'
```
Amejia481: I had to suppress this check to avoid  the lint error.
To be able to reuse Robolectric code.
image

pocmo: See lint.xml in the module. You can just list those jars to be excluded. There should already be one listed in the file.

Amejia481: 👍

Amejia481: Done! 🕺

Amejia481:
@pocmo  should I integrate this one with  #1020?

pocmo:
@Amejia481 Yeah, I think that would be a good idea. But it doesn't need to be now. Do you want to land that first or update the PR?

Amejia481:
Ok, lets land this first 😄 !

Amejia481:
@csadilek  helped me to add this feature and we did and less that 2 minutes. I'm updating the PR 😆

pocmo:
Alright 🤣

components/browser/session/src/main/java/mozilla/components/browser/session/SelectionAwareSessionObserver.kt
```diff
@@ -47,7 +47,7 @@ abstract class SelectionAwareSessionObserver(
     /**
      * Stops the observer.
      */
-    fun stop() {
+    open fun stop() {
```
Amejia481: We had to add this one to be able to override the method.

pocmo: Instead of extending `SelectionAwareSessionObserver` you could also create a local instance inside `DownloadsFeature`. May make it easier to understand what's going on too. :)

pocmo:
>  sample browser

I meant reference browser! :)

samples/browser/src/main/java/org/mozilla/samples/browser/BrowserFragment.kt
```diff
@@ -46,6 +51,13 @@ class BrowserFragment : Fragment(), BackHandler {
                 sessionId)
 
         tabsToolbarFeature = TabsToolbarFeature(context!!, toolbar, ::showTabs)
+
+        val ctx = context!!.applicationContext
```
pocmo: * nit: You can use `requireContext` in Fragments if you want to avoid `context!!` and are expecting to always have. a `Context`.
* nit: I feel the shorter `ctx` is harder to skim then just using `context` or `applicationContext` :)

Amejia481: Nice I didn't know about `requireContext()`, thanks for the suggestion!

components/browser/session/src/main/java/mozilla/components/browser/session/SelectionAwareSessionObserver.kt
Amejia481: But it will also over-complicate things a bit.

Because then I will have to implement `Session.Observer` for getting notify on `onDownload` and I will need the `activeSession` and `SelectionAwareSessionObserver` just expose the `activeSession` for their children 😿  .

https://github.com/mozilla-mobile/android-components/blob/996c9c493beb21a182b08680889fa5df6e4b2a36/components/browser/session/src/main/java/mozilla/components/browser/session/SelectionAwareSessionObserver.kt#L23

PR #1019, opened 2018-10-10
Added required uri parameter to createErrorPage.

keianhzo:
Some error pages like [Safe Browsing](https://testsafebrowsing.appspot.com/) error pages require a URI parameter.

### DIFF
--- a/components/browser/errorpages/src/main/java/mozilla/components/browser/errorpages/ErrorPages.kt
+++ b/components/browser/errorpages/src/main/java/mozilla/components/browser/errorpages/ErrorPages.kt
@@ -16,6 +16,7 @@ object ErrorPages {
     fun createErrorPage(
         context: Context,
         errorType: ErrorType,
+        uri: String? = null,
         @RawRes htmlResource: Int = R.raw.error_pages,
         @RawRes cssResource: Int = R.raw.error_style
     ): String {
@@ -29,7 +30,7 @@ object ErrorPages {
             .replace("%pageTitle%", context.getString(R.string.errorpage_title))
             .replace("%button%", context.getString(R.string.errorpage_refresh))
             .replace("%messageShort%", context.getString(errorType.titleRes))
-            .replace("%messageLong%", context.getString(errorType.messageRes))
+            .replace("%messageLong%", context.getString(errorType.messageRes, uri))
             .replace("%css%", css)
     }
 }
--- a/components/browser/errorpages/src/test/java/mozilla/components/browser/errorpages/ErrorPagesTest.kt
+++ b/components/browser/errorpages/src/test/java/mozilla/components/browser/errorpages/ErrorPagesTest.kt
@@ -9,6 +9,7 @@ package mozilla.components.browser.errorpages
 import org.junit.Assert
 import org.junit.Test
 import org.junit.runner.RunWith
+import org.mockito.ArgumentMatchers.nullable
 import org.mockito.Mockito.anyInt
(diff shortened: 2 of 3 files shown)


components/browser/errorpages/src/main/java/mozilla/components/browser/errorpages/ErrorPages.kt
```diff
@@ -29,7 +30,7 @@ object ErrorPages {
             .replace("%pageTitle%", context.getString(R.string.errorpage_title))
             .replace("%button%", context.getString(R.string.errorpage_refresh))
             .replace("%messageShort%", context.getString(errorType.titleRes))
-            .replace("%messageLong%", context.getString(errorType.messageRes))
+            .replace("%messageLong%", context.getString(errorType.messageRes, uri ?: ""))
```
jonalmeida: Since you've added a default value of empty string for this parameter, then you don't need to use the elvis operator to do a null-check.

That said, in my test, `Context#getString` already handles nullable values being passed to it, so you don't have to do this check.

Keeping it as just `context.getString(errorType.messageRes, uri)` will work along with your changes above.

components/browser/errorpages/src/main/java/mozilla/components/browser/errorpages/ErrorPages.kt
```diff
@@ -16,6 +16,7 @@ object ErrorPages {
     fun createErrorPage(
         context: Context,
         errorType: ErrorType,
+        uri: String? = "",
```
jonalmeida: If you're adding a default parameter, then we can assert that the string is non-null. If we _don't_ have it, then it's nullable.

So you can choose to either have `uri: String = ""` or `uri: String?`. You don't need both together :).

keianhzo: Sorry, not a Kotlin expert. If I make it `uri: String?` (optional without default value) then I **must** pass a parameter when calling createErrorPage. Can I avoid that without providing a default value?

components/browser/errorpages/src/main/java/mozilla/components/browser/errorpages/ErrorPages.kt
keianhzo: Thanks, that can be removed.

components/browser/errorpages/src/main/java/mozilla/components/browser/errorpages/ErrorPages.kt
pocmo: > I must pass a parameter when calling createErrorPage. Can I avoid that without providing a default value?

The default can be null too (`uri: String? = null?`) if that is helpful here.

keianhzo:
Thanks @pocmo! @jonalmeida as I think I can't remove the default value in this case, I've just make the default value be null.

jonalmeida:
@keianhzo that's great! If you could fix the failing test to reflect that change and squash your commits into one, it looks good to me. :)

keianhzo:
@jonalmeida Fixed, this should be the good one

jonalmeida:
🍾 🎉

pocmo:
@jonalmeida Should we mention that in the changelog? Should apps like Firefox for Fire TV pass in the URL now?

jonalmeida:
@pocmo No, I don't think so. This was more of a bugfix where we were showing "%s" in some error pages when we were supposed to show the error url.

PR #1020, opened 2018-10-10
Closes #962: Add reusable functionality for observing selected session

### LINKED ISSUE #962: Add (front-end) feature skeleton to support-base component
At least our front-end feature components always have start() / stop() methods. They also have a presenter and an interactor that get started and stopped.

I am in a situation where I want to write a "feature" that like ToolbarFeature/ToolbarPresenter/ToolbarInteractor needs to track the currently selected session (See code in ToolbarPresenter). This already feels like a common use case and boilerplate code that I would like to avoid writing again.

csadilek:
This provides a reusable solution for consumers to observe the currently selected session, even if it changes. I went with a simple approach:

We already have both `Session.Observer` and `SessionManager.Observer` in our `browser-session` component. We can hide these from consumers and alternatively let them implement the new `SessionObserver`. The consumer can then decide whether they want to observe a single session or always observe the selected session. Both cases are needed in our `ToolbarPresenter` to support CustomTabs

This cleaned up the `ToolbarPresenter` and also seems to work well for the download feature under development in #1009 .

The reason I went with an interface instead of an (abstract) class is to not limit the possible consumers (e.g. a consumer could already by a subclass and still need this functionality).

### DIFF
--- a/components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
+++ b/components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
@@ -4,6 +4,7 @@
 
 package mozilla.components.feature.toolbar
 
+import mozilla.components.browser.session.SelectionAwareSessionObserver
 import mozilla.components.browser.session.Session
 import mozilla.components.browser.session.SessionManager
 import mozilla.components.concept.toolbar.Toolbar
@@ -16,41 +17,22 @@ class ToolbarPresenter(
     private val toolbar: Toolbar,
     private val sessionManager: SessionManager,
     private val sessionId: String? = null
-) : SessionManager.Observer, Session.Observer {
-    private var activeSession: Session? = null
+) : SelectionAwareSessionObserver(sessionManager) {
 
     /**
      * Start presenter: Display data in toolbar.
      */
     fun start() {
-        activeSession = sessionId?.let {
-            sessionManager.findSessionById(sessionId)
-        } ?: run {
-            sessionManager.register(this)
-            sessionManager.selectedSession
-        }
-
-        activeSession?.register(this)
+        val session = sessionId?.let { sessionManager.findSessionById(sessionId) }
+        session?.let { observeFixed(it) } ?: observeSelected()
         initializeView()
-    }
-
-    /**
-     * Stop presenter from updating the view.
-     */
-    fun stop() {
-        sessionManager.unregister(this)
-        activeSession?.unregister(this)
     }
 
     /**
      * A new session has been selected: Update toolbar to display data of new session.
      */
     override fun onSessionSelected(session: Session) {
-        activeSession?.unregister(this)
-
-        activeSession = session
-        session.register(this)
-
+        super.onSessionSelected(session)
         initializeView()
     }
 
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/SelectionAwareSessionObserver.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/SelectionAwareSessionObserver.kt
@@ -1,0 +1,61 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.session
+
+import android.support.annotation.CallSuper
+
+/**
+ * This class is a combination of [Session.Observer] and
+ * [SessionManager.Observer]. It provides functionality to observe changes to a
+ * specified or selected session, and can automatically take care of switching
+ * over the observer in case a different session gets selected (see
+ * [observeFixed] and [observeSelected]).
+ *
+ * @property activeSession the currently observed session
+ * @property sessionManager the application's session manager
+ */
+abstract class SelectionAwareSessionObserver(
+    private val sessionManager: SessionManager
+) : SessionManager.Observer, Session.Observer {
+
+    protected open var activeSession: Session? = null
+
+    /**
+     * Starts observing changes to the specified session.
+     *
+     * @param session the session to observe.
+     */
+    fun observeFixed(session: Session) {
+        activeSession = session
+        session.register(this)
+    }
+
+    /**
+     * Starts observing changes to the selected session (see
+     * [SessionManager.selectedSession]). If a different session is selected
+     * the observer will automatically be switched over and only notified of
+     * changes to the newly selected session.
+     */
+    fun observeSelected() {
+        activeSession = sessionManager.selectedSession
+        sessionManager.register(this)
+        activeSession?.register(this)
+    }
+
+    /**
+     * Stops the observer.
+     */
+    fun stop() {
+        sessionManager.unregister(this)
+        activeSession?.unregister(this)
+    }
+
+    @CallSuper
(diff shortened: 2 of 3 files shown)


csadilek:
@Amejia481 yes, you're right, that was an oversight. updated now.

csadilek:
Moved this back to using an abstract class, as there is no way to limit the visibility of interface properties and this would otherwise expose `activeSession` which....isn't good.

components/browser/session/src/main/java/mozilla/components/browser/session/SessionObserver.kt
```diff
@@ -0,0 +1,58 @@
...
+ * This class is a combination of [Session.Observer] and
+ * [SessionManager.Observer]. It provides functionality to observe changes to a
+ * specified or selected session, and automatically takes care of switching
+ * over the observer in case a different session is selected (see [observe] and
+ * [observeSelected]).
+ *
+ * @property activeSession the currently observed session
+ * @property sessionManager the application's session manager
+ */
+abstract class SessionObserver(
```
pocmo: nit: I wonder if we can find a more explicit name. The difference between `Session.Observer` and `SessionObserver` is very subtle. And you can actually accidentally use `SessionObserver` instead of `Session.Observer` because it implements that. :)

Not sure what a good name would be though. `SingleSessionObserver`, `FollowingSessionObserver`, `AutoSessionObserver`, `SelectedSessionObserver` (it can do more than that though).

components/browser/session/src/main/java/mozilla/components/browser/session/SessionObserver.kt
```diff
@@ -0,0 +1,58 @@
...
+) : SessionManager.Observer, Session.Observer {
+
+    protected open var activeSession: Session? = null
+
+    /**
+     * Starts observing changes to the specified session.
+     *
+     * @param session the session to observe.
+     */
+    fun observe(session: Session) {
```
pocmo: API nit: Would it be even more obvious if we name this something like `observeOnly()` or `observeAlways`?

components/browser/session/src/main/java/mozilla/components/browser/session/SessionObserver.kt
```diff
@@ -0,0 +1,58 @@
...
+        session.register(this)
+    }
+
+    /**
+     * Starts observing changes to the selected session (see
+     * [SessionManager.selectedSession]). If a different session is selected
+     * the observer will automatically be switched over and only notified of
+     * changes to the newly selected session.
+     */
+    fun observeSelected() {
```
pocmo: API nit: I wonder if we can make it even more obvious that this is not only observing the *currently* selected session but updates? `observeSelection()`, `alwaysObserveSelected()`, ... not sure.

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
```diff
@@ -16,41 +17,22 @@ class ToolbarPresenter(
...
      * A new session has been selected: Update toolbar to display data of new session.
      */
     override fun onSessionSelected(session: Session) {
-        activeSession?.unregister(this)
-
-        activeSession = session
-        session.register(this)
-
+        super.onSessionSelected(session)
         initializeView()
```
pocmo: This `super` call is something that is easy to forget and feels like it is not the best solution yet. We could try to at least annotate `SessionObserver.onSessionSelected()` with [@CallSuper](https://developer.android.com/reference/android/support/annotation/CallSuper). This will show a warning in the IDE if you forget that. Haven't tried whether that works in Kotlin though.

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
```diff
@@ -16,41 +17,22 @@ class ToolbarPresenter(
...
-        activeSession = sessionId?.let {
-            sessionManager.findSessionById(sessionId)
-        } ?: run {
-            sessionManager.register(this)
-            sessionManager.selectedSession
-        }
-
-        activeSession?.register(this)
+        val session = sessionId?.let { sessionManager.findSessionById(sessionId) }
+        session?.let { observe(it) } ?: run { observeSelected() }
```
pocmo: Is the `run {}` block really needed here?

components/browser/session/src/main/java/mozilla/components/browser/session/SessionObserver.kt
```diff
@@ -0,0 +1,58 @@
...
+    fun stop() {
+        sessionManager.unregister(this)
+        activeSession?.unregister(this)
+    }
+
+    override fun onSessionSelected(session: Session) {
+        activeSession?.unregister(this)
+        activeSession = session
+        session.register(this)
+    }
```
pocmo: I'm currently reviewing the feature-downloads patch and I wonder if there's a race condition here: Can `stop()` and `onSessionSelected()` be called in such an order that we are still registered to a session?

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
csadilek: Yes, good idea, `CallSuper` works fine. I've added it. This override isn't actually needed for any of our samples, as we're always starting/stopping the fragment and never actually have to observe `onSessionSelected`. I opted for keeping it in `ToolbarPresenter` anyway as it should work regardless. Wdyt?

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
csadilek: No, it's not needed. I just found it easier to read `let {} ?: run {}`, but I'll remove it if it's confusing.

components/browser/session/src/main/java/mozilla/components/browser/session/SessionObserver.kt
csadilek: @pocmo So, In the current design, I don't see a race condition:

- Opening the TabsTray stops the BrowserFragment which stops the ToolbarFeature->Presenter ->SessionObserver
-  A new session is selected in the tabs tray -> the ToolbarPresenter no longer observes that
- The TabsTrayFragment stops and the BrowserFragment starts again re-registering the observer

So, no race condition, but it also means that `onSessionSelected` might not even be needed? Should we not start/stop the ToolbarPresenter in this case?

csadilek: Marking this as resolved as per our chat. Even if an app doesn't use fragments and therefore uses `onSessionSelected` there still won't be a race conditions as all methods in `ObserverRegistry` are synchronized on `observers`.

csadilek:
@pocmo I think everything is resolved here now. I've also found new names which I think make it very clear, for both the class and the two methods.

So, the class is a `SelectionAwareSessionObserver`, which describes exactly what it offers and consumers can either observe a fixed session (`observeFixed`) or observe whatever session is selected (`observeSelected`). I think in combination this makes it all clear?

pocmo:
@csadilek Changelog worthy? :)

csadilek:
@pocmo yes, I think it makes sense to start adding features / API to the changelog even if it's "just" for component developers. Added it here: https://github.com/mozilla-mobile/android-components/pull/1035

PR #1023, opened 2018-10-11
Fixes #1022: LocalizationProvider not accepting the right locale.

### LINKED ISSUE #1022: LocalizationProvider not accepting the right locale
I have tracked down this bug from [here](https://github.com/mozilla-mobile/focus-android/issues/1899). The localizationProvider would have the previous locale after the change. Making a PR for this

levimatheri:
Enforced localizationProvider vars enabling them to have the right locale. This fixes the bug whereby changing locale uses search engines from the previous locale

### DIFF
--- a/components/browser/search/src/main/java/mozilla/components/browser/search/provider/AssetsSearchEngineProvider.kt
+++ b/components/browser/search/src/main/java/mozilla/components/browser/search/provider/AssetsSearchEngineProvider.kt
@@ -15,7 +15,6 @@ import mozilla.components.browser.search.provider.localization.SearchLocalizatio
 import mozilla.components.support.ktx.android.content.res.readJSONObject
 import org.json.JSONArray
 import org.json.JSONObject
-import java.util.*
 
 /**
  * SearchEngineProvider implementation to load the included search engines from assets.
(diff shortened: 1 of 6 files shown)


components/browser/search/src/main/java/mozilla/components/browser/search/provider/AssetsSearchEngineProvider.kt
```diff
@@ -46,6 +48,12 @@ class AssetsSearchEngineProvider(
         return loadSearchEnginesFromList(context, searchEngineIdentifiers.distinct())
     }
 
+    // Bug fix (hack!) ensure localization provider vars get the right locale
+    private fun enforceLocaleChange() {
+        localizationProvider.language = Locale.getDefault().language
+        localizationProvider.country = Locale.getDefault().country
```
pocmo: The localization provider can have different implementations so you do not want to always override that with the default locale here.

components/browser/search/src/main/java/mozilla/components/browser/search/provider/localization/LocaleSearchLocalizationProvider.kt
```diff
@@ -11,9 +11,9 @@ import java.util.Locale
  * default languageTag.
  */
 class LocaleSearchLocalizationProvider : SearchLocalizationProvider() {
-    override val language: String = Locale.getDefault().language
+    override var language: String = Locale.getDefault().language
 
-    override val country: String = Locale.getDefault().country
+    override var country: String = Locale.getDefault().country
```
pocmo: I think an easier way to fix the issue would be to avoid using a backing field for this property (which will be set only once) and instead use a getter (that will be evaluated on every read) here:

```
override var language: String
    get() = Locale.getDefault().language

override val country: String
    get() = Locale.getDefault().country
```

components/browser/search/src/main/java/mozilla/components/browser/search/provider/AssetsSearchEngineProvider.kt
levimatheri: I agree since localizationProvider is of "type" SearchLocalizationProvider. I hadn't thought of that

components/browser/search/src/main/java/mozilla/components/browser/search/provider/localization/LocaleSearchLocalizationProvider.kt
levimatheri: Will make the change

pocmo:
Landed as ed6de28ffdb22412f363eda0e39816873c40b13e

PR #1030, opened 2018-10-11
closes #1017: makes Download file name not nullable

### LINKED ISSUE #1017: Change the property fileName to No-null on Download class
https://github.com/mozilla-mobile/android-components/blob/5e2251df0bb5e0e8c95d18f430a3fdc9486f589c/components/browser/session/src/main/java/mozilla/components/browser/session/Download.kt#L21

renanrbs:
resolves issue #1017


Amejia481:
Don't worry @renanrbs !
I will help you out!

Amejia481:
Travis had time out trying to resolve dependencies from [jcenter.bintray.com](jcenter.bintray.com), sometimes is not reliable, resolving dependencies. I just restart the build.

I believe if you sign on Travis using your GitHub account, you can do it too.
image

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -359,11 +359,13 @@ class GeckoEngineSession(
...
-                    contentLength = response.contentLength,
-                    contentType = response.contentType,
-                    fileName = response.filename)
+                response.filename?.let {
+                    onExternalResource(
+                            url = response.uri,
+                            contentLength = response.contentLength,
+                            contentType = response.contentType,
+                            fileName = it)
+                }
```
pocmo: ⚠️ We **always** want to invoke the callback! If filename is null then please use `DownloadUtils.guessFileName()`.

pocmo:
In the case here there seems to be an actual build/test error:

```
:browser-engine-system:compileDebugUnitTestKotline: /home/travis/build/mozilla-mobile/android-components/components/browser/engine-system/src/test/java/mozilla/components/browser/engine/system/SystemEngineViewTest.kt: (253, 13): 'onExternalResource' overrides nothing
 FAILED
```

Amejia481:
When you enter to the Taskcluster page it will display all the tasks that have been run and it will put a label indicating the status.

image

Just click the one that you want to take a look, for example  `Android Components - detekt`, go to the tab **Run logs** and click `public/logs/live.log`
image

Then you can see the whole log.

image

renanrbs:
Should i make unit tests for the Download class?

pocmo:
> Should i make unit tests for the Download class?

Taskcluster and Travis do that - and it seems to have passed there. But yeah, usually it's good to do that before pushing; you'll get feedback faster that way. :)

renanrbs:
i've squashed the commits

PR #1031, opened 2018-10-11
Engine history delegate API

### LINKED ISSUE #1025: Engine: Expose API for receiving history/visits
We will need an API for exposing history/visits. That's what we store as browsing history. This data is different from just URLs we load. For example it doesn't include error pages or redirects.

WebView API:
https://developer.android.com/reference/android/webkit/WebViewClient.html#doUpdateVisitedHistory(android.webkit.WebView,%20java.lang.String,%20boolean)

The GeckoView API is not implemented yet:
https://bugzilla.mozilla.org/show_bug.cgi?id=1494713

grigoryk:
Closes #1025.

The goal is to provide our engines with ability to query for browser history
and to record visits and page metadata updates (e.g. title changes).

GeckoView's APIs are still being implemented in https://bugzilla.mozilla.org/show_bug.cgi?id=1494713,
so this works from the implementation notes. WebView's APIs are well known.

History delegate is optional. Absence of it in settings when configuring an engine as akin to
disabling history tracking.

This patch explicitly doesn't provide an implementation of a delegate.


components/concept/engine/src/main/java/mozilla/components/concept/engine/Settings.kt
```diff
@@ -40,6 +41,11 @@ abstract class Settings {
      */
     open var requestInterceptor: RequestInterceptor? by UnsupportedSetting()
 
+    /**
+     * Setting to provide a history delegate to the engine.
+     */
+    open var historyTrackingDelegate: HistoryTrackingDelegate? by UnsupportedSetting()
```
pocmo: I assume you went with a delegate in `Settings` instead of exposing that in the `EngineSession.Observer` because you want a single observer for all engine sessions?

components/concept/engine/src/main/java/mozilla/components/concept/engine/history/HistoryTrackingDelegate.kt
```diff
@@ -0,0 +1,34 @@
...
+
+    /**
+     * An engine needs to know "visited" (true/false) status for provided URIs.
+     */
+    fun getVisited(uris: List<String>, callback: (List<Boolean>) -> Unit)
+
+    /**
+     * An engine needs to know a list of all visited URIs.
+     */
+    fun getVisited(callback: (List<String>) -> Unit)
```
pocmo: I assume we need those two lists because WebView needs all URLs upfront and GeckoView will only ask for the URLs of the current document?

components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -335,6 +339,11 @@ class GeckoEngineSession(
         override fun onCloseRequest(session: GeckoSession) = Unit
 
         override fun onTitleChange(session: GeckoSession, title: String) {
+            // Only track history meta information if we're not in a private mode.
+            if (!privateMode) {
+                settings.historyTrackingDelegate?.onTitleChanged(currentUrl, title)
+            }
```
pocmo: Ignoring private pages is something the storage wants. Is it correct to assume that every `HistoryTrackingDelegate` implementation will not care about them?

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -34,24 +35,27 @@ import org.mozilla.geckoview.GeckoSessionSettings
 @Suppress("TooManyFunctions")
 class GeckoEngineSession(
     runtime: GeckoRuntime,
-    privateMode: Boolean = false,
+    val privateMode: Boolean = false,
```
csadilek: If me make this a `val/property` I think we should make it `private`.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -34,17 +35,19 @@ import org.mozilla.geckoview.GeckoSessionSettings
 @Suppress("TooManyFunctions")
 class GeckoEngineSession(
     runtime: GeckoRuntime,
-    privateMode: Boolean = false,
+    val privateMode: Boolean = false,
```
csadilek: `private val...` see above.

components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -30,17 +31,19 @@ import org.mozilla.geckoview.GeckoSessionSettings
 @Suppress("TooManyFunctions")
 class GeckoEngineSession(
     runtime: GeckoRuntime,
-    privateMode: Boolean = false,
+    val privateMode: Boolean = false,
```
csadilek: `private val...` see above.

components/concept/engine/src/main/java/mozilla/components/concept/engine/history/HistoryTrackingDelegate.kt
grigoryk: That’s exactly right yeah.

components/concept/engine/src/main/java/mozilla/components/concept/engine/Settings.kt
grigoryk: Basically, yes. This seemed like the most straightforward approach once you start wiring things together, configuring a delagate this way becomes trivial.

Going through enginesession would also basically mean going through the SessionManager, I think, and that didn’t seem like the right fit here.

components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
grigoryk: Probably not! For general browsing scenarios this is exactly what we want though, I think?

We could expand this to include “isPrivate” flag, and have the delegate implementation do what it wants.

What do you think?

grigoryk: Changed the delegate interface to require passing `privateMode`. This gives us the most flexibility around different types of delegate implementations, and keeps the engine stuff simpler.

grigoryk:
Cleaned up things, added more tests, addressed @pocmo's feedback around privateMode flag.

grigoryk:
Failing an unrelated test:
```
mozilla.components.browser.engine.gecko.GeckoEngineSessionTest > saveState FAILED
    java.lang.UnsatisfiedLinkError at GeckoEngineSessionTest.kt:273
```

PR #1034, opened 2018-10-12
Update gradle (and wrapper) to 4.10.2..

pocmo:
Don't want to do that just yet. I'd like to update the Docker image too - so that not every task needs to re-download Gradle. For now I just want to see if everything still works.


pocmo:
All green. Nice! :)

pocmo:
Okay, created a new Docker image with gradle 4.10.2 and am pushing it to Docker hub right now. Will update the PR afterwards.

pocmo:
Updated PR with new Docker image.

PR #1037, opened 2018-10-12
Issue #506: Add new generic crash reporting component (lib-crash)..

### LINKED ISSUE #506: Component Idea: Crash reporting (Sentry / Socorro)
In Firefox for iOS we have a small wrapper around Sentry. Not super big but enough to add some more functionality.

* it will check the build you are on
* it will grab the Sentry DSN from a standard place
* it makes sure to not enable it when running in the simulator
* it implements its own device id generation to avoid a sticky ID

See also https://github.com/mozilla-mobile/firefox-ios/blob/master/Shared/SentryIntegration.swift

pocmo:
Closes #506.


components/lib/crash/build.gradle
```diff
@@ -0,0 +1,52 @@
...
+    testImplementation Deps.testing_mockito
+}
+
+archivesBaseName = "lib-crash"
+
+apply from: '../../../publish.gradle'
+ext.configurePublish(
+        'org.mozilla.components',
+        'lib-crash',
+        'A generic crash repoorting library that can submitReport crashes to multiple services.')
```
csadilek: Some typos here the `repoo` :), and `submitReport` which was probably caused by a refactoring.

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/CrashReporterService.kt
```diff
@@ -0,0 +1,16 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.lib.crash.service
+
+import mozilla.components.lib.crash.crash.Crash
```
csadilek: `crash.crash.Crash` seems like one `crash` too many. Let's move the `Crash` class to the root package so it's `lib.crash.Crash`?

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/CrashReporterService.kt
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Interface to be implemented by external services that
```
csadilek: // TODO complete docs :)

components/lib/crash/README.md
```diff
@@ -0,0 +1,155 @@
...
+CrashReporter(
+    services = listOf(
+        // List the crash reporting services you want to use
+    )
+).install(this)
+```
+
+### Sending crash reports to Sentry
+
+⚠️ Note: The crash reporter library is compiled against the Sentry SDK but it doesn't require it as a dependency. The app using the component is responsible for adding the Sentry dependency to its build files in order to use Sentry crash reporting.
```
csadilek: Interesting, I was wondering whether this needs to be another concept and multiple implementation case. Here we're hiding the implementations from the class loader (not creating instances or referencing any static fields/methods). Don't see a problem other than being careful not to break this as we wouldn't detect it automatically.

components/lib/crash/src/test/java/mozilla/components/lib/crash/crash/CrashTest.kt
```diff
@@ -0,0 +1,59 @@
...
+            "/data/data/org.mozilla.samples.browser/files/mozilla/Crash Reports/pending/3ba5f665-8422-dc8e-a88e-fc65c081d304.dmp",
+            recoveredCrash.minidumpPath
+        )
+        assertEquals(
+            "/data/data/org.mozilla.samples.browser/files/mozilla/Crash Reports/pending/3ba5f665-8422-dc8e-a88e-fc65c081d304.extra",
+            recoveredCrash.extrasPath
+        )
+    }
+
+    fun `Test serializing and de-serializing UncaughtExceptionCrash`() {
```
csadilek: Maybe just `serialize and deserialize UncaughtExceptionCrash`

components/lib/crash/src/test/java/mozilla/components/lib/crash/crash/NativeCodeCrashTest.kt
```diff
@@ -0,0 +1,42 @@
...
+import android.content.Intent
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
+class NativeCodeCrashTest {
+    @Test
+    fun `test creating crash object from sample GeckoView intent`() {
```
csadilek: let's skip the test prefix ;)

components/lib/crash/build.gradle
pocmo: fail2

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/CrashReporterService.kt
pocmo: Oh yeah, that make sense. In the IDE this looked fine because it only showed the sub tree. :)

pocmo:
Added the last code pieces and wrote some more tests (coverage for the component locally is at 91%).

Need to document the prompt API in the README and will do some more on-device testing.

components/lib/crash/README.md
pocmo: Yeah, that would be the most clean way. However it feels super over-engineered here. I'd have to create two separate gradle modules (+ concept module) and artifacts just for two files. Our first consumer will be Focus and Focus actually requires both (and other GV apps in the future probably too). Luckily we will still be able to move the code to new artifacts later if we decide so. :)

pocmo:
I'm going to disable CrashReporterActivity because of timing issues.

PR #1040, opened 2018-10-14
Issue #1015: Enable an detekt rule for wrap blocks in { }

### LINKED ISSUE #1015: Enable an detekt rule for wrap blocks in { }
e.g

🔴 Bad
```kotlin
        if (!successful)
            println("Hoorah!")
```

🍏  Good
```kotlin
        if (!successful){
            println("Hoorah!")
       }
```

vadimsemenov:
Closes #1015


components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
```diff
@@ -312,6 +310,7 @@ internal class DisplayToolbar(
         //   +-------------+------------------------------------------------+
 
         val navigationActionsWidth = navigationActions
+            .asSequence()
```
pocmo: Those two `asSequence()` changes seem unrelated?

vadimsemenov: I've noticed it because of IDEA inspection.

If we apply call chain on `List`, on each call new `ArrayList` will be created. So in worst case `O((number of calls in chain) * (size of initial list))` aux memory will be allocated. `asSequence` fixes it and creates only (more on less) a new `Sequence` on each call. 
It should improve both memory consumption and performance. 

In case it seems unrelated, I can extract it to another pull request. 
If you'd like me to do this, should I also extract fixes in documentation? As It seems equally unrelated to the original issue...

vadimsemenov: @pocmo I've run that inspection on the whole project and found another warning here: https://github.com/mozilla-mobile/android-components/blob/47bde0ed44b3c32c7fae399c26048476ff16f9e1/components/service/fretboard/src/main/java/mozilla/components/service/fretboard/Fretboard.kt#L107

So it may worth to extract it to separate issue and pull request. What do you think?

vadimsemenov: @pocmo, @Amejia481 any updates on this?

PR #1060, opened 2018-10-16
Blog Post: Release updates only in the changelog [ci skip]

jonalmeida:
It felt a bit odd to continue having releases while our blog looked like we discontinued working on a-c with no further release announcements. This blog post is just there to inform anyone entering from the blog that release announcements will only be on the changelog.

Just a nice-to-have.  ¯\\\_(ツ)\_/¯


pocmo:
I was thinking about just moving the existing posts to the draft folder. But that's an option too! :)

jonalmeida:
That works too! Will it break any links in mailing lists?

pocmo:
> That works too! Will it break any links in mailing lists?

🤷‍♂️

jonalmeida:
I took a look at some of our mailing list emails and I don't see any direct links to the blog posts so I'm going to close this PR and we can retire our previous posts to the drafts like you said.

PR #1071, opened 2018-10-16
Closes #1057: Add localization notes to Error Pages [ci skip].

### LINKED ISSUE #1057: Localization comments for strings
We've heard from the Firefox Reality folks that they'll be using strings from Error Pages: https://github.com/mozilla-mobile/android-components/blob/master/components/browser/errorpages/src/main/res/values/strings.xml

In order to meet their timeline for launch in a subset of languages, we'll have to kick-off localization this week. In order for us to do this, localization comments are needed for each string in order to give context to translators (and preferably not using too much technical jargon so they can understand how to translate).

Here are guidelines for writing localization notes: https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_content_best_practices#Add_localization_notes

I can review that once it's done. Please reach out should you have any questions about this.

In the meantime, I'll be going through each string and checking for any specific issues that may need to be fixed.
(also, CC @Pike)

jonalmeida:
@Delphine I've taken a look at the localization notes for strings that are similar in Fennec and Focus, and I've added my own in accordance. Could you verify if this is fine, or if I'm missing something else? :)


components/browser/errorpages/src/main/res/values/strings.xml
```diff
@@ -32,6 +33,7 @@
 
 
     <string name="error_malformedURI_title">The address isn’t valid</string>
+    <!-- LOCALIZATION NOTE (error_malformedURI_message) This string contains markup. The URL should not be localized. -->
```
Pike: Just `<!-- This string contains markup. The URL should not be localized. -->`

Let's drop the gecko-isms about `LOCALIZATION NOTE`.

components/browser/errorpages/src/main/res/values/strings.xml
```diff
@@ -67,13 +69,13 @@ be temporary, and you can try again later.</li>
     <string name="errorpage_title">Problem loading page</string>
     <string name="errorpage_refresh">Try Again</string>
 
-    <!-- NavigationDelegate.ERROR_CATEGORY_UNKNOWN -->
+    <!-- Category: NavigationDelegate.ERROR_CATEGORY_UNKNOWN -->
```
Pike: This comment will only be shown for `error_generic_title`, and doesn't help there.

If the comment isn't vital for the actual code maintenance, I'd drop it.

Comments that are actually localizer-facing should be one per `<string>`.

components/browser/errorpages/src/main/res/values/strings.xml
```diff
@@ -185,12 +187,13 @@ be temporary, and you can try again later.</li>
...
 
-    <!-- NavigationDelegate.ERROR_CATEGORY_URI -->
+    <!-- Category: NavigationDelegate.ERROR_CATEGORY_URI -->
     <string name="error_unknown_host_title">Address Not Found</string>
+    <!-- LOCALIZATION NOTE (error_unknown_host_message) in the example, the two URLs in markup do not need to be translated. -->
     <string name="error_unknown_host_message"><![CDATA[
 <p>The browser could not find the host server for the provided address.</p>
 <ul>
-<li>Did you make a mistake when typing the domain? (e.g. <q><strong>ww</strong>.mozilla.org</q> instead of <q><strong>www</strong>.mozilla.org</q>)</li><li>Are you certain this domain address exists?  Its registration may have expired.</li>
+<li>Did you make a mistake when typing the domain? (e.g. <q><strong>ww</strong>.mozilla.org</q> instead of <q><strong>www</strong>.mozilla.org</q>)</li><li>Are you certain this domain address exists?  It\'s registration may have expired.</li>
```
Pike: Should this be a proper English apostrophe?

Also, if this changes the string from their corresponding messages in Focus etc, maybe we should fix these once we've set up l10n etc? We'll like look into the English string to carry forward localizations so that we don't loose existing work, and this might make that trickier.

Delphine: I confirm this should not be a straight apostrophe. 
(' vs ’)

components/browser/errorpages/src/main/res/values/strings.xml
jonalmeida: The comment was more for us to categorize sections for different error types. I'll remove it if it provides confusion.

components/browser/errorpages/src/main/res/values/strings.xml
jonalmeida: I assumed it was an error, but I see your point as well. :)

jonalmeida:
@Pike I've updated my PR with your comments in mind, thanks!

components/browser/errorpages/src/main/res/values/strings.xml
jonalmeida: Thanks for confirming Delphine!

components/browser/errorpages/src/main/res/values/strings.xml
```diff
@@ -88,7 +90,7 @@ be temporary, and you can try again later.</li>
 </ul>
```
Delphine: Curly apostrophe needed here again :)
’

jonalmeida: I think I got them all. :)

components/browser/errorpages/src/main/res/values/strings.xml
```diff
@@ -114,7 +116,7 @@ be temporary, and you can try again later.</li>
 <p>Though the site seems valid, the browser was unable to establish a connection.</p>
 <ul>
 <li>Could the site be temporarily unavailable? Try again later.</li>
-<li>Are you unable to browse other sites?  Check the computer\'s network connection.</li>
+<li>Are you unable to browse other sites?  Check the computer\’s network connection.</li>
```
cvan: you can remove the backslash (previously for escaping the `'`)  now, correct?

components/browser/errorpages/src/main/res/values/strings.xml
```diff
@@ -21,6 +21,7 @@
     <string name="error_timeout_message">@string/error_connectionfailure_message</string>
```
cvan: it's not covered here, but I have a question about the first string [defined at the top of this file](https://github.com/mozilla-mobile/android-components/blob/b308888/components/browser/errorpages/src/main/res/values/strings.xml#L6):

```xml
    <!-- The page html title (i.e. the <title> tag content) -->
    <string name="error_connectionfailure_title">Unable to connect</string>
    <string name="error_connectionfailure_message">
    <![CDATA[
      <ul>
        <li>The site could be temporarily unavailable or too busy. Try again in a few moments.</li>
        <li>If you are unable to load any pages, check your mobile device’s data or Wi-Fi connection.</li>
      </ul>
    ]]>
    </string>
```

components/browser/errorpages/src/main/res/values/strings.xml
```diff
@@ -129,7 +131,7 @@ be temporary, and you can try again later.</li>
 <p>The browser has stopped trying to retrieve the requested item. The site is redirecting the request in a way that will never complete.</p>
 <ul>
 <li>Have you disabled or blocked cookies required by this site?</li>
-<li><em>NOTE</em>: If accepting the site\'s cookies does not resolve the problem, it is likely a server configuration issue and not your computer.</li>
+<li><em>NOTE</em>: If accepting the site\’s cookies does not resolve the problem, it is likely a server configuration issue and not your computer.</li>
 </ul>
      ]]></string>
```
cvan: in the [`Offline Mode` string](https://github.com/mozilla-mobile/android-components/blob/b30888/components/browser/errorpages/src/main/res/values/strings.xml#L143) can we change `&quot;Try Again&quot;` to `“Try Again”`?

components/browser/errorpages/src/main/res/values/strings.xml
Delphine: I second that :)

components/browser/errorpages/src/main/res/values/strings.xml
```diff
@@ -32,6 +33,7 @@
 
 
     <string name="error_malformedURI_title">The address isn’t valid</string>
```
cvan: I should likely file a separate issue, but I notice that the formatting of these names is inconsistent. a few examples from this file:

- `error_connectionfailure_title`
- `error_malformedURI_title`
- `error_redirectLoop_title`
- `error_unsupportedprotocol_title`
- `error_sslhandshake_title`
- `errorpage_title`
- `error_generic_title`
- `error_security_bad_cert_title`
- `error_connection_failure_title`
- `error_safe_browsing_malware_uri_title`

components/browser/errorpages/src/main/res/values/strings.xml
```diff
@@ -67,13 +69,13 @@ be temporary, and you can try again later.</li>
     <string name="errorpage_title">Problem loading page</string>
     <string name="errorpage_refresh">Try Again</string>
```
cvan: IMO, there should be a newline above this and an L10n comment like this:

```xml
<!-- The button that appears at the bottom of an error page. -->
```

components/browser/errorpages/src/main/res/values/strings.xml
```diff
@@ -67,13 +69,13 @@ be temporary, and you can try again later.</li>
     <string name="errorpage_title">Problem loading page</string>
```
cvan: the comment above might read better like this:

```xml
<!-- The document title for an error page (i.e., the contents of the `<title>` tag). -->
```

components/browser/errorpages/src/main/res/values/strings.xml
```diff
@@ -103,7 +105,7 @@ be temporary, and you can try again later.</li>
 <p>The requested site did not respond to a connection request and the browser has stopped waiting for a reply.</p>
 <ul>
 <li>Could the server be experiencing high demand or a temporary outage?  Try again later.</li>
```
cvan: can we collapse the whitespace here? `?[space][space]Try` -> `?[space]Try`

components/browser/errorpages/src/main/res/values/strings.xml
cvan: can we remove the extra space after the `?` here too?

components/browser/errorpages/src/main/res/values/strings.xml
```diff
@@ -185,12 +187,13 @@ be temporary, and you can try again later.</li>
...
 
-    <!-- NavigationDelegate.ERROR_CATEGORY_URI -->
+
     <string name="error_unknown_host_title">Address Not Found</string>
+    <!-- In the example, the two URLs in markup do not need to be translated. -->
     <string name="error_unknown_host_message"><![CDATA[
 <p>The browser could not find the host server for the provided address.</p>
 <ul>
-<li>Did you make a mistake when typing the domain? (e.g. <q><strong>ww</strong>.mozilla.org</q> instead of <q><strong>www</strong>.mozilla.org</q>)</li><li>Are you certain this domain address exists?  Its registration may have expired.</li>
+<li>Did you make a mistake when typing the domain? (e.g. <q><strong>ww</strong>.mozilla.org</q> instead of <q><strong>www</strong>.mozilla.org</q>)</li><li>Are you certain this domain address exists?  It’s registration may have expired.</li>
```
cvan: can we remove the extra space after the `?` here too (and on the following lines)?

components/browser/errorpages/src/main/res/values/strings.xml
cvan: grammar nit: can we change `i.e.` to `i.e.,`?

components/browser/errorpages/src/main/res/values/strings.xml
cvan: grammar nit: can we change `e.g.` to `e.g.,` (assuming it's not referenced in other products like this)?

components/browser/errorpages/src/main/res/values/strings.xml
jonalmeida: That's a valid point - I'm not sure there's any value for the localizer in knowing that this string is a page title as well. I'd argue that removing it would be better.

Delphine:
FWIW I feel like I'm done reviewing this from an l10n-PM point of view. I'll sit back and wait for changes before reviewing again (and not sure @Pike + cvan are finished commenting either :) )

components/browser/errorpages/src/main/res/values/strings.xml
jonalmeida: Looking at cvan's comment below, instead I'm going to change this to the text below for each title:
```
<!-- The document title for an error page (i.e., the contents of the `<title>` tag). -->
```

jonalmeida:
> Also, the `@string/` redirects aren't really localizable, so the compare-locales checks we're adding will fail on those. We'll want to move them to a different resources file that we don't expose to l10n.

@Pike Those strings don't currently have any use cases right now afaik. The others are mapped to an `ErrorType` enum, but these don't seem to have any mapping. It might be easier remove them entirely if it causes a problem, or we can have a comment above telling the localizers not to translate these.

Thoughts?

components/browser/errorpages/src/main/res/values/strings.xml
jonalmeida: I'd like to fix this now, but it seems like there's a name collision if I remove the camel casing and for `error_connectionfailure_title` and it's not yet clear to me why this string resource is needed.

jonalmeida: In order to keep these strings in and fix this issue, I'm just going to append `_alternative` to the end of the duplicates that aren't being used by the `ErrorType` enum, and we can still have these translated for now.

jonalmeida:
Hello all, thanks for taking the time to review these strings so that we can get them right straight off the bat. :)

I've addressed all the latest feedback and all added some formatting as well. Take a look at the individual commits if the overall diff is a bit too messy to read.

@cvan @Delphine if you could take another pass at these, that would be great!

components/browser/errorpages/src/main/res/values/strings.xml
```diff
@@ -31,248 +27,273 @@
...
+      <p>The site at %1$s has been reported as serving unwanted software and has been blocked based on your security preferences.</p>
      ]]></string>
 
+    <!-- The document title for an error page (i.e., the contents of the `<title>` tag). -->
     <string name="error_safe_harmful_uri_title">Harmful site issue</string>
     <!-- The %1$s will be replaced by the malicious website URL-->
     <string name="error_safe_harmful_uri_message"><![CDATA[
-<p>The site at %1$s has been reported as a potentially harmful site and has been blocked based on your security preferences.</p>
+      <p>The site at %1$s has been reported as a potentially harmful site and has been blocked based on your security preferences.</p>
      ]]></string>
```
cvan: while you are fixing the indents, all of these are off by a single space

jonalmeida: I wish our formatter handled these CDATA tags so we didn't have to do it by hand - fixed!

pocmo:
> @Pike Those strings don't currently have any use cases right now afaik. The others are mapped to an ErrorType enum, but these don't seem to have any mapping. It might be easier remove them entirely if it causes a problem, or we can have a comment above telling the localizers not to translate these.

You can still have string references if you want to. You just need to move them to a separate file (e.g. `string_references.xml`). Now that we will have L10N we will need to write some docs. :)

jonalmeida:
@pocmo I wanted to get this merged before I added namespacing. Lots of conflicts would occur if I put both PRs together.

Pike:
Yes, let's settle on IDs before exposing the strings to l10n.

We'll want to recreate the translations from existing translations as much as possible, we'll need to figure out how to use TM in an automated fashion here, but that's something the l10n team needs to get done. It'll take a few days, though.

Delphine:
Yes I have a few meetings this morning, will check again right after them. thanks

components/browser/errorpages/src/main/res/values/strings.xml
```diff
@@ -14,14 +14,11 @@
...
+    <!-- The document title for an error page (i.e., the contents of the `<title>` tag). -->
+    <string name="error_host_lookup_title">Server not found</string>
 
-    <string name="error_timeout_title">The connection timed out</string>
-    <string name="error_timeout_message">@string/error_connectionfailure_message</string>
-
-    <string name="error_hostLookup_title">Server not found</string>
-    <string name="error_hostLookup_message"><![CDATA[
+    <!-- In the example, the two URLs in markup do not need to be translated. -->
+    <string name="error_host_lookup_message"><![CDATA[
```
cvan: I couldn't find where this string was used

jonalmeida: Yes, there are some strings which aren't being mapped to an`ErrorType` value yet, but we allowed them to be added to resource file to have them translated until we figure out if we should keep or remove it (better to have it translated than not?). We can update our API and release much quicker if we didn't have to wait for new translations if we decide we need it later or simply remove it otherwise.

components/browser/errorpages/src/main/res/values/strings.xml
```diff
@@ -30,242 +27,273 @@
...
+      </ul>
     ]]></string>
 
-    <string name="error_redirectLoop_title">The page isn’t redirecting properly</string>
-    <string name="error_redirectLoop_message"><![CDATA[
-<ul>
-  <li>This problem can sometimes be caused by disabling or refusing to accept cookies.</li>
-</ul>
+    <!-- The document title for an error page (i.e., the contents of the `<title>` tag). -->
+    <string name="error_unsupported_protocol_title">The address wasn’t understood</string>
```
cvan: it looks like `mozac_browser_errorpages_unknown_protocol_title` and `mozac_browser_errorpages_unknown_protocol_message` are [the ones used now](https://github.com/mozilla-mobile/android-components/blob/af881f8/components/browser/errorpages/src/main/java/mozilla/components/browser/errorpages/ErrorPages.kt#L114-L115) (I'll make this change in a PR I'm about to open)

components/browser/errorpages/src/main/res/values/strings.xml
```diff
@@ -30,242 +27,273 @@
...
+      </ul>
     ]]></string>
 
-    <string name="error_unsupportedprotocol_title">The address wasn’t understood</string>
-    <string name="error_unsupportedprotocol_message"><![CDATA[
-<ul>
-  <li>You might need to install other software to open this address.</li>
-</ul>
+    <!-- The document title for an error page (i.e., the contents of the `<title>` tag). -->
+    <string name="error_ssl_handshake_title">Secure connection failed</string>
```
cvan: I can't tell when these were ever [used](https://github.com/mozilla-mobile/android-components/blob/master/components/browser/errorpages/src/main/java/mozilla/components/browser/errorpages/ErrorPages.kt#L114-L115)? I assume it's not an oversight in the [GeckoView error codes](https://mozilla.github.io/geckoview/javadoc/mozilla-central/org/mozilla/geckoview/GeckoSession.NavigationDelegate.html#field.summary)

I don't see them here:

- https://dxr.mozilla.org/mozilla-central/source/mobile/locales/en-US/overrides/netError.dtd
- https://dxr.mozilla.org/mozilla-central/source/dom/locales/en-US/chrome/netError.dtd
- https://dxr.mozilla.org/mozilla-central/source/browser/locales/en-US/chrome/overrides/netError.dtd

PR #1073, opened 2018-10-17
Setup codeowners for glean component.

pocmo:
@Dexterp37 @mdboom I would like to add you as codeowners for glean. This means every change of code inside the component will require approval from one of you. Does this make sense to you? Is there anyone else I should add?


Dexterp37:
> Does this make sense to you? Is there anyone else I should add?

It does, thank you so much! Would you mind also adding @georgf (Georg Fritzsche) to this?

PR #1088, opened 2018-10-17
Closes #980: Add pressed state for unselected toggle buttons

### LINKED ISSUE #980: Add pressed state for unselected toggle buttons
> When you press on a toggle button, there's no animation to indicate that you've pressed it. There is no effect right now.

Amejia481:
Is this the expected behavior?
[Before](https://drive.google.com/file/d/1sZjC3bDktQzPhzm_fWH-XYtyYKiKVJzE/view)
[Now](https://drive.google.com/file/d/1uyLT9QfxB_kAiMzxu0hVflrqq7_onmjv/view)


jonalmeida:
> Is this the expected behavior?
> [Before](https://drive.google.com/file/d/1sZjC3bDktQzPhzm_fWH-XYtyYKiKVJzE/view)
> [Now](https://drive.google.com/file/d/1uyLT9QfxB_kAiMzxu0hVflrqq7_onmjv/view)

From the video, I'd say no since they ripple animation just seems to cause be vastly different from the selected state surrounding circle, and the animation is confined to the layout of the icon.

This might be more complicated than just adding an attribute then. 🤔

Amejia481:
@jonalmeida thanks 😀  for the clarification I believe that I got the right believable.
You can take a look [here](https://drive.google.com/file/d/1JHNngZKzESB-SQSU7jNkKOdSLNRaYVBd/view) and let me know.

samples/toolbar/src/main/res/drawable/toggle_background.xml
```diff
@@ -6,6 +6,12 @@
     xmlns:android="http://schemas.android.com/apk/res/android"
     android:color="@color/photonGrey90">
 
+    <item android:id="@android:id/mask">
```
Amejia481: @mcomella this approach fix the issue for your use case?

pocmo:
Regardless of whether that is the correct pressed state: Looking at your change it seems like this is something the app customizing the background needs to do and not a functionality we need to provide in the toolbar API. Right?

Amejia481:
I guess so. We haven't been exposing this behavior through our API. We only have been showing how to do it in our sample apps(samples-toolbar). But Correct me if I'm wrong, I'm not quite sure if we should do it.

samples/toolbar/src/main/res/drawable/toggle_background.xml
mcomella: I'm not sure – I was specifically trying to address the problem that the ripple appears on non-toggle ToolbarAction but only appears in the enabled -> disabled state on toggleable actions: https://drive.google.com/file/d/1fS_A4HtXQFB6AT_StketLG-gpRf89d7_/view?usp=sharing

However, I didn't realize that some part of this is done in the application – we set the ripple on our elements – but I don't know why the ripple effect shouldn't just work going disable -> enabled (maybe because there needs to be a background color for that to work?). I'm not sure if this is a bug in the product or in components.

csadilek:
@Amejia481 the changelog has a conflict now. Would you mind updating when you get a chance? ;)

csadilek:
@Amejia481 actually, never mind, I was able to resolve this here directly.

Amejia481:
👍

csadilek:
See https://github.com/mozilla-mobile/android-components/pull/1134/

PR #1090, opened 2018-10-17
Bump mozilla_app_services to 0.6.0, and update FxA component to match

thomcc:
We changed some things about how errors are handled internally, which required becoming consistent about which value is used for the panic error code (-1 everywhere now).

This also changes how the `FxaException` implemented `fromConsuming` because IMO it's much better to be explicit about what the error code values are (and also one of them is -1, which isn't a valid enum index, so it wouldn't have worked for that one anymore anyway).

### DIFF
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -26,7 +26,7 @@ private object Versions {
 
     const val sentry = "1.7.10"
 
-    const val mozilla_app_services = "0.5.1"
+    const val mozilla_app_services = "0.6.0"
     const val servo = "0.0.1.20181017.aa95911"
 }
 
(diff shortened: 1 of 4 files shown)


pocmo:
This seems to break on taskcluster:

```
:service-firefox-accounts:assemble
:service-firefox-accounts:compileDebugUnitTestKotline: /build/android-components/components/service/firefox-accounts/src/test/java/mozilla/components/service/fxa/RustObjectTest.kt: (101, 52): Unresolved reference: ErrorCode
e: /build/android-components/components/service/firefox-accounts/src/test/java/mozilla/components/service/fxa/RustObjectTest.kt: (122, 55): Unresolved reference: ErrorCode
e: /build/android-components/components/service/firefox-accounts/src/test/java/mozilla/components/service/fxa/RustObjectTest.kt: (127, 55): Unresolved reference: ErrorCode
e: /build/android-components/components/service/firefox-accounts/src/test/java/mozilla/components/service/fxa/RustObjectTest.kt: (132, 55): Unresolved reference: ErrorCode
 FAILED
```

thomcc:
Sorry about that, totally missed this reply (and those tests >_>).

To be honest, I'm not a huge fan of exposing these codes but ultimately it doesn't matter that much.

PR #1099, opened 2018-10-18
Closes #1072: Make Keystore class open.

### LINKED ISSUE #1072: DataProtect: Make `Keystore` class `open` for client tests
We need to mock the `Keystore` class in DataProtect for our tests. To do this, the class needs to be `open` rather than `final`.

pocmo:
CC @linuxwolf

### DIFF
--- a/components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/Keystore.kt
+++ b/components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/Keystore.kt
@@ -135,7 +135,7 @@ internal open class KeyStoreWrapper {
  * @constructor Creates a new instance around a key identified by the given label
  */
 @TargetApi(Build.VERSION_CODES.M)
-class Keystore(val label: String) {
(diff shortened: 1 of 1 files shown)


linuxwolf:
@sashei how may of the methods within do you need to be `open`?  My preference would be to keep that set to a minimum ...

sashei:
specifically `encryptBytes` and `decrypteBytes` would be good.

pocmo:
Oh, sorry, I forgot that the method will be `final` too. I'll open up the two methods and land.

pocmo:
@sashei @linuxwolf I updated the PR and opened up encryptBytes() and decryptBytes() too. Will this help you with testing?

Another options may be that we introduce an interface that `Keystore` implements and Lockbox code just cares about the interface which will be easier to mock. But would that be worth it?

sashei:
@pocmo I think the implementation details are up to you. I'm generally opposed to interfaces for the sake of interfaces but if you feel the pattern makes sense here we're happy to take advantage of it for testing!!

PR #1103, opened 2018-10-18
Close #1064: add Parent id to Session

### LINKED ISSUE #1064: Keep a reference to "parent" session
Rocket keeps a reference (only ID) to the parent `Session`. The parent is the `Session` the `Session` was opened from.

walkingice:
When creating new session from an existing session, we need to the
relationship between two sessions. For example, to click a href link in
web-page to open new tab.

This commit add parent id to Session. If a session is added without
parent, it will be append to tail of sessions. Otherwise it will sit
next to parent.

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
@@ -72,7 +72,21 @@ class SessionManager(
     /**
      * Adds the provided session.
      */
-    fun add(session: Session, selected: Boolean = false, engineSession: EngineSession? = null) = synchronized(values) {
+    fun add(
+        session: Session,
+        selected: Boolean = false,
+        engineSession: EngineSession? = null,
+        parent: Session? = null
+    ) = synchronized(values) {
+
+        parent?.let {
+            if (!values.contains(it)) {
+                throw IllegalArgumentException("The parent does not exist")
+            }
+
+            session.parentId = parent.id
+        }
+
         values.add(session)
 
         engineSession?.let { link(session, it) }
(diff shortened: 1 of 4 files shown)


components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
```diff
@@ -72,7 +72,21 @@ class SessionManager(
...
-    fun add(session: Session, selected: Boolean = false, engineSession: EngineSession? = null) = synchronized(values) {
+    fun add(
+        session: Session,
+        selected: Boolean = false,
+        engineSession: EngineSession? = null,
+        parent: Session? = null
+    ) = synchronized(values) {
+
+        parent?.let {
+            if (values.indexOf(it) == -1) {
```
pocmo: Maybe with "contains" and named parameter?

```Kotlin
parent?.let { session -> 
            if (!values.contains(session)) {
```

components/browser/session/src/test/java/mozilla/components/browser/session/SessionManagerTest.kt
```diff
@@ -64,6 +64,25 @@ class SessionManagerTest {
...
+        manager.add(session2)
+        manager.add(session3, parent = session1)
+        manager.add(session4, parent = session2)
+
+        assertNull(manager.sessions[0].parentId)
+        assertNull(manager.sessions[1].parentId)
+        assertEquals(session1.id, manager.sessions[2].parentId)
+        assertEquals(session2.id, manager.sessions[3].parentId)
+    }
+
```
pocmo: We could add another test that makes sure we are throwing an exception, e.g.:

```Kotlin
@Test(expected = IllegalArgumentException::class)
fun `session manager throws exception if parent is not in session manager`() {
    val parent = Session("https://www.mozilla.org")
    val session = Session("https://www.firefox.com")

    val manager = SessionManager(mock())
    manager.add(session, parent = parent)
}
```

pocmo:
This seems to have broken the IntentProcessor?

https://travis-ci.org/mozilla-mobile/android-components/builds/443117876?utm_source=github_status&utm_medium=notification

PR #1107, opened 2018-10-19
Do Not Merge WIP towards a places component

mhammond:
This is a simple WIP which creates a simple "places" component. It's very thin as code which currently lives in the application-services repo hasn't been copied here yet - see ./components/service/sync-places/README.md for more information.

This has been created primarily so a related PR in reference-browser makes sense and can be reproduced. We'll work with the a-c team to bash this into better shape soon.


samples/browser/build.gradle
```diff
@@ -132,6 +132,8 @@ dependencies {
 
     servoImplementation project(':browser-engine-servo')
 
+    implementation project(':service-sync-places')
```
csadilek: For sync-logins we have a separate sample app. We'd likely not add all sync features to the sample browser, I think. So, we'd have a sync specific sample app and the reference-browser (having all features.)

pocmo:
Closing in favor of #1240

PR #1111, opened 2018-10-19
Use annotation processor to generate telemetry doc in csv and markdown

cnevinc:


pocmo:
There are some detekt/ktlint issues we will need to fix before landing.

PR #1116, opened 2018-10-20
Add document for debugging Telemetry library.

cnevinc:


components/service/telemetry/README.md
```diff
@@ -16,6 +16,9 @@ Use gradle to download the library from JCenter:
 implementation "org.mozilla.components:service-telemetry:{latest-version}"
 ```
 
+### Debug
+
+If you want to debug your Telemetry code, use ``DebugLogClient`` instead of ``HttpURLConnectionTelemetryClient``. And add ``Log.addSink(AndroidLogSink())`` before you check it in logcat. Beware, the tag you set in ``DebugLogClient`` won't be used. See [Logging](../../support/base/README.md#logging) section for more details.
```
pocmo: Can you file an issue about the tag? We should fix that. :)

cnevinc: > Can you file an issue about the tag? We should fix that. :)

created! see #1121

PR #1138, opened 2018-10-23
Update detekt

pocmo:


build.gradle
```diff
@@ -198,15 +198,20 @@ task clean(type: Delete) {
...
-    version = "1.0.0.RC7-3"
-    profile("main") {
-        input = "$projectDir/components"
-        config = "$projectDir/config/detekt.yml"
-        filters = ".*test.*,.*/resources/.*,.*/tmp/.*"
-        output = "$projectDir/build/reports/detekt"
-        baseline = "$projectDir/config/detekt-baseline.xml"
+    version = "1.0.0.RC9.2"
+    input = files("$projectDir/components")
+    config = files("$projectDir/config/detekt.yml")
```
grigoryk: We have just one config, so this could be `file` instead?

automation/docker/Dockerfile
```diff
@@ -82,6 +82,13 @@ RUN ./gradlew clean \
...
+    && ./gradlew --no-daemon detekt \
     && ./gradlew --no-daemon ktlint \
     && ./gradlew --no-daemon docs
+
+#----------------------------------------------------------------------------------------------------------------------
+# -- Cleanup ----------------------------------------------------------------------------------------------------------
+#----------------------------------------------------------------------------------------------------------------------
+
+RUN apt-get clean
+>>>>>>> Use new detekt command in automation scripts.
```
grigoryk: This snuck in somehow! I guess Docker didn't care though.

pocmo: Oops.

build.gradle
pocmo: detekt somehow expects a collection 🤷‍♂️ 

```
> Cannot cast object '/Users/skaspari/Projects/Mozilla/android-components/config/detekt.yml' with class 'java.io.File' to class 'org.gradle.api.file.FileCollection'
```

Issue #1012, opened 2018-10-10
Concept component hiding (http) networking stack


pocmo:
Currently our component code uses `HttpUrlConnection` - mostly because that is offered by the Android System and doesn't require any third-party dependency.

However `HttpUrlConnection` has its issues:
* The API doesn't really separate requests/responses and makes it unpredictable at what point it actually performs a request. It's not pleasant to use.
* It only has a synchronous API so you almost always want to wrap it
* It may have different underlying implementations based on Android version / flavor.

Other options are:
* [OkHttp](http://square.github.io/okhttp/) is the de-facto standard used in Android apps. A bunch of [other libraries use it](https://github.com/square/okhttp/wiki/Works-with-OkHttp) internally. HttpUrlConnection is actually backed by it on newer versions of Android.
* [Necko](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/Necko) - There has been an ongoing discussion on whether Necko can/should/will be turned into something that can live outside of Firefox. This was mostly interesting for Rust components. However with a nice Kotlin layer on top that this would be interesting for app code as well.
* `WebExecutor` [API in GeckoView](https://bugzilla.mozilla.org/show_bug.cgi?id=1405081): The GeckoView team is working on an API that allows apps using GV to send arbitrary http requests.

The reality is:

* Apps may have different requirements.
  * An app already using GV may want to send (some) http requests through it as well (Favicons?).
  * An app already using okHttp may want to do that for component code too (because it allows them to use debugging tools like stetho).
  * An app optimized for size may want to continue using HttpUrlConnection (e.g. Rocket)
  * Apps not using GV definitely do not want to add it just for the http client.
* Android Components do not want to force any (third-party) dependencies on apps if not absolutely necessary.

So should we:

* Add a `concept` component with the necessary interfaces for doing http requests (something like [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API); `concept-fetch`?)

* Provide components implementing the `concept` component. At least HttpUrlConnection and GeckoView?

* Modify components doing http requests to depend on the `concept` component and let apps inject a dependency implementing it.

pocmo:
@ncalexan You mentioned that you are working on something similar in the Rust world. You also mentioned that it may be worthwhile to put things together in a way that Rust and Kotlin code uses the same http client implementation.

pocmo:
Components that do some kind of http request or require a client:

* service-glean: Ping upload
[...]

PR #1142, opened 2018-10-23
Bug 1497809 - Implement a simple ping uploader

Dexterp37:
This implements a ping uploader that attempts to send a ping once and never again. This also implements the `Configuration` object and a `DefaultConfiguration` for glean.

Please note that this is a copy of the `HttpURLConnectionTelemetryClient.java` from `telemetry-android`, converted to Kotlin to have an uniform codebase.

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
@@ -1,0 +1,24 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.glean.config
+
+import mozilla.components.service.glean.BuildConfig
+
+/**
+ * The Configuration class describes how to configure the Glean.
+ *
+ * @property serverEndpoint the server pings are sent to
+ * @property userAgent the user agent used when sending pings
+ * @property connectionTimeout the timeout, in milliseconds, to use when connecting to
+ *           the [serverEndpoint]
+ * @property readTimeout the timeout, in milliseconds, to use when connecting to
+ *           the [serverEndpoint]
+ */
+data class Configuration(
+    val serverEndpoint: String = "https://incoming.telemetry.mozilla.org",
+    val userAgent: String = "Glean/${BuildConfig.LIBRARY_VERSION} (Android)",
+    val connectionTimeout: Int = 10000,
+    val readTimeout: Int = 30000
+)
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/net/HttpPingUploader.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/net/HttpPingUploader.kt
@@ -1,0 +1,120 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.glean.net
+
+import android.support.annotation.VisibleForTesting
+import mozilla.components.service.glean.BuildConfig
+import mozilla.components.service.glean.config.Configuration
+import mozilla.components.support.base.log.logger.Logger
+import java.io.IOException
+import java.net.HttpURLConnection
+import java.net.MalformedURLException
+import java.net.URL
+
+/**
+ * A simple ping Uploader, which implements a "send once" policy, never
+ * storing or attempting to send the ping again.
+ */
+class HttpPingUploader(configuration: Configuration) : PingUploader {
+    private val config = configuration
+    private val logger = Logger("glean/HttpPingUploader")
+
+    /**
+     * Synchronously upload a ping to Mozilla servers.
+     * Note that the `X-Client-Type`: `Glean` and `X-Client-Version`: <SDK version>
+     * headers are added to the HTTP request in addition to the UserAgent. This allows
+     * us to easily handle pings coming from glean on the legacy Mozilla pipeline.
+     *
+     * @param path the URL path to append to the server address
+     * @param data the serialized text data to send
+     *
+     * @return true if the ping was correctly dealt with (sent successfully
+     *         or faced an unrecoverable error), false if there was a recoverable
+     *         error callers can deal with.
+     */
+    @Suppress("ReturnCount", "MagicNumber")
+    override fun upload(path: String, data: String): Boolean {
+        var connection: HttpURLConnection? = null
+        try {
+            connection = openConnection(config.serverEndpoint, path)
+            connection.requestMethod = "POST"
+            connection.connectTimeout = config.connectionTimeout
+            connection.readTimeout = config.readTimeout
+            connection.doOutput = true
+
+            connection.setRequestProperty("Content-Type", "application/json; charset=utf-8")
+            connection.setRequestProperty("User-Agent", config.userAgent)
+            connection.setRequestProperty("Date", createDateHeaderValue())
+
+            // Add headers for supporting the legacy pipeline.
+            connection.setRequestProperty("X-Client-Type", "Glean")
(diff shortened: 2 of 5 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
```diff
@@ -0,0 +1,18 @@
...
+ *           the [serverEndpoint]
+ * @property readTimeout the timeout, in milliseconds, to use when connecting to
+ *           the [serverEndpoint]
+ */
+public data class Configuration(
+    val serverEndpoint: String,
+    val userAgent: String,
+    val connectionTimeout: Int,
+    val readTimeout: Int
+)
```
pocmo: I wonder if `DefaultConfiguration` is really needed. You can define default values for the properties here. It will allow the app to override specific ones if needed. But I guess most of the time the default will be good enough here?

E.g.:

```Kotlin
data class Configuration(
    val serverEndpoint: String = "https://incoming.telemetry.mozilla.org",
    val userAgent: String = "Glean/1.0 (Android)",
    val connectionTimeout: Int = 1000,
    val readTimeout = 30000
}
```

Also note that `public` is redundant here. By default everything is public in Kotlin.

components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
```diff
@@ -0,0 +1,18 @@
+package mozilla.components.service.glean.config
```
pocmo: nit: license header here and in the other files :)

components/service/glean/src/main/java/mozilla/components/service/glean/config/DefaultConfiguration.kt
```diff
@@ -0,0 +1,8 @@
+package mozilla.components.service.glean.config
+
+public val DefaultConfiguration = Configuration(
+    serverEndpoint = "https://incoming.telemetry.mozilla.org",
+    userAgent = "Glean/1.0 (Android)",
```
pocmo: "1.0" doesn't match the release version number. Is this something you would like to use? Currently we do not expose the version as a code constant. But we could.

pocmo:
> Please note that this is a copy of the HttpURLConnectionTelemetryClient.java from telemetry-android, converted to Kotlin to have an uniform codebase.

Sounds good for now. In #1012 we want to provide an http client to components.

components/service/glean/src/main/java/mozilla/components/service/glean/config/DefaultConfiguration.kt
Dexterp37: @mreid-moz to the rescue! Do you think exposing the Glean version through the "UserAgent" header is of any use? Otherwise I'll gladly morph this to `"Glean (Android"`

components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
Dexterp37: What is wrong with my memory!!? 🤦‍♂️ Addressing this :D

components/service/glean/src/main/java/mozilla/components/service/glean/config/DefaultConfiguration.kt
mreid-moz: Generally I think it will be very useful to know the Glean SDK version for a given submission, since it will allow us to handle the possible future case where we want to make changes to the client-server protocol (or server-side routing/processing). Putting it in the `UserAgent` string seems like a good approach.

The current Telemetry ingestion pipeline does not parse or pass along the `UserAgent` string (though support code for that exists), but the next gen [gcp-based ingestion infra](https://github.com/mozilla/gcp-ingestion/blob/b289340aae93fee5f6673cee2e715433f1c50f08/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseUserAgent.java) does UA-parsing.

One piece of prior art: when developing PingSender, we added a specific header `X-PingSender-Version`, maybe because we didn't already have UserAgent parsing at the time? If we want to go with this approach, I would prefer to generalize to something like a pair of `X-Client-Type` and `X-Client-Version` headers.

Dexterp37: > One piece of prior art: when developing PingSender, we added a specific header `X-PingSender-Version`, maybe because we didn't already have UserAgent parsing at the time? If we want to go with this approach, I would prefer to generalize to something like a pair of `X-Client-Type` and `X-Client-Version` headers.

Good point! I didn't remember that, you're right. We added these two headers because the pipeline, at the time, didn't support UserAgent parsing (see [the docs](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/internals/pingsender.html)). I will add the two headers for backwards compatibility with the current pipeline, and still send the UserAgent, as done by the `pingsender` on desktop.

@pocmo turns out we might need to export the version as a code constant. Is this something that's trivial to do? Can I take this as a separate PR (I will file a bug to track this)?

pocmo: Yes, this should be trivial. I can try that and open a PR.

components/service/glean/src/main/java/mozilla/components/service/glean/net/HttpPingUploader.kt
```diff
@@ -0,0 +1,100 @@
...
+ * A simple ping Uploader, which implements a "send once" policy, never
+ * storing or attempting to send the ping again.
+ */
+class HttpPingUploader(configuration: Configuration) : PingUploader {
+    private val config = configuration
+    private val logger = Logger("glean/HttpPingUploader")
+
+    override fun upload(path:String, data: String): Boolean {
+        try {
+            with(openConnection(config.serverEndpoint, path) as HttpURLConnection) {
```
pocmo: One thing that is missing here is that we never seem to call `disconnect()` on the connection. Usually that's part of the `finally` block of the `try` block to make sure it's always executed.

components/service/glean/src/main/java/mozilla/components/service/glean/net/PingUploader.kt
```diff
@@ -0,0 +1,18 @@
+package mozilla.components.service.glean.net
```
pocmo: License header :D

components/service/glean/src/main/java/mozilla/components/service/glean/config/DefaultConfiguration.kt
Dexterp37: Thank you!

components/service/glean/src/main/java/mozilla/components/service/glean/net/PingUploader.kt
Dexterp37: 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️

components/service/glean/src/main/java/mozilla/components/service/glean/config/DefaultConfiguration.kt
mreid-moz: > > One piece of prior art: when developing PingSender, we added a specific header `X-PingSender-Version`, maybe because we didn't already have UserAgent parsing at the time? If we want to go with this approach, I would prefer to generalize to something like a pair of `X-Client-Type` and `X-Client-Version` headers.
> 
> Good point! I didn't remember that, you're right. We added these two headers because the pipeline, at the time, didn't support UserAgent parsing (see [the docs](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/internals/pingsender.html)). I will add the two headers for backwards compatibility with the current pipeline, and still send the UserAgent, as done by the `pingsender` on desktop.

FWIW we don't currently use the two headers I mentioned, we only use the pingsender-specific one. I just figured that to be future proof, we could use the same thing here and elsewhere. I think sending the version in the `UserAgent` is probably sufficient if we want to go that route.

Dexterp37: Oh! @mreid-moz , should I remove the 2 additional headers or leave them?

PR #1146, opened 2018-10-23
Issue #1011: Fretboard: ValuesProvider.getCountry(): MissingResourceException

### LINKED ISSUE #1011: Fretboard: ValuesProvider.getCountry(): MissingResourceException
From Focus:

```
Caused by: java.util.MissingResourceException: 
  at java.util.Locale.getISO3Country (Locale.java:1589)
  at mozilla.components.service.fretboard.ValuesProvider.getCountry (ValuesProvider.kt:76)
  at mozilla.components.service.fretboard.ExperimentEvaluator.matches (ExperimentEvaluator.kt:71)
  at mozilla.components.service.fretboard.ExperimentEvaluator.evaluate (ExperimentEvaluator.kt:38)
  at mozilla.components.service.fretboard.ExperimentEvaluator.evaluate$default (ExperimentEvaluator.kt:35)
  at mozilla.components.service.fretboard.Fretboard.isInExperiment (Fretboard.kt:74)
  at mozilla.components.service.fretboard.Fretboard.getExperimentsMap (Fretboard.kt:120)
  at org.mozilla.focus.utils.ExperimentsKt.getActiveExperimentNames (Experiments.kt:27)
  at org.mozilla.focus.telemetry.TelemetryWrapper.recordActiveExperiments (TelemetryWrapper.kt:821)
  at org.mozilla.focus.telemetry.TelemetryWrapper.init (TelemetryWrapper.kt:262)
```

fercarcedo:
The `Locale` class [documentation](https://developer.android.com/reference/java/util/Locale.html#getISO3Country()) says the `getISO3Country` method throws a `MissingResourceException` if the three-letter country abbreviation is not available for a `Locale`.

What this pull request does is in this case it returns the result of calling `getCountry`, which could return a 2 letter country code or an empty string (but not throw an exception). It seems more reasonable to me to do this rather than, for example, simply return an empty string in that case.

This pull request also does the same for the language, since it's the same problem.

Closes #1011

### DIFF
--- a/components/service/fretboard/src/main/java/mozilla/components/service/fretboard/ValuesProvider.kt
+++ b/components/service/fretboard/src/main/java/mozilla/components/service/fretboard/ValuesProvider.kt
@@ -7,6 +7,7 @@ package mozilla.components.service.fretboard
 import android.content.Context
 import android.os.Build
 import java.util.Locale
+import java.util.MissingResourceException
 
 /**
  * Class used to provide
@@ -19,7 +20,11 @@ open class ValuesProvider {
      * @return user's language as a three-letter abbreviation
      */
     open fun getLanguage(context: Context): String {
-        return Locale.getDefault().isO3Language
+        return try {
+            Locale.getDefault().isO3Language
+        } catch (e: MissingResourceException) {
+            Locale.getDefault().language
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


fercarcedo:
I have added some tests :)

Issue #1085, opened 2018-10-17
API feedback and improvements for service/firefox-accounts


csadilek:
I am opening this issue to collect feedback I have while integrating [service/firefox-accounts](https://github.com/mozilla-mobile/android-components/tree/master/components/service/firefox-accounts) into our [reference-browser](https://github.com/mozilla-mobile/reference-browser/) and to start discussing future improvements. Nothing here needs to be addressed immediately. Should I run into a bug that can't be worked around I will file a separate issue.

~**Configuration:**~ (**Fixed**)
The first step when using the API is to create a `Config` instance and pass it to the `FirefoxAccount` constructor. In addition, a `CLIENT_ID` and `REDIRECT_URL` need to be provided separately:
```Kotlin
 Config.custom(CONFIG_URL).then { 
   config -> FirefoxAccount(config, CLIENT_ID, REDIRECT_URL)
 } ...
```
Later, when the OAuth flow is started, `scopes` and the `wantsKeys` boolean need to be provided:
```Kotlin
account.beginOAuthFlow(scopes, wantsKeys)
```
At least `CLIENT_ID` and `REDIRECT_URL` (and probably `scopes`) are client-side config properties as they won't change at run-time (if at all). We could extend the `Config` class to contain both server-side and client-side configurations. This would clean up the API quite a bit.

~**Result types:**~ (**Fixed**)
We've introduced `FxaResult` to have a promise type in Java < 8. This has now been duplicated in `SyncResult`. It doesn't appear to be a goal anymore to support old versions of Java. GeckoView switched to Java 8 as well. If that's the case, I'd vote for removing these types and replace them with `CompletableFuture` (Java SDK) or `Deferred` (Kotlin), where we need Promise types, and a simple synchronous result type otherwise.

**Higher-Level API:**
At least from a Java/Kotlin dev perspective, I originally expected `FirefoxAccount` to represent an account i.e. to be a data class or value object. In reality, it's the entry point of the API and contains almost all functionality of the library including the OAuth flow. After bringing in `service-firefox-accounts` as a component, I would have expected something along the lines of:
```Kotlin
val fxaService = FirefoxAccountsService(config)
val loginUrl = fxaService.createLoginUrl()
```
We could even provide login methods for the various supported use cases e.g.:
- Login using a CustomTab
  ```Kotlin
  fxaService.login()
  ```
- Login using a fragment with a WebView or GeckoView
  ```Kotlin
  fxaService.login(fragment)
  ```
[...]

PR #1148, opened 2018-10-23
Use the new fxa lib exported from application-services

thomcc:
In https://github.com/mozilla/application-services/pull/311 we moved the unsafe portion of the FxA kotlin code into application-services. This is the android-components side of that work.

The primary changes here are: 

- Move the code to use `async`/`await`/`launch` instead of a dubious and buggy hand-rolled thing (`FxaResult` is dead, long live `Deferred`).
- The FxA component is entirely safe code and you're completely free to change it in whatever ways you want. I don't feel any need to be flagged on reviews for it in the future, although am completely willing to review things if you'd like.
    - This clears out a lot of rust-specific cruft and moves it to our repo.
    - And also means that if the versions mismatch in a compatibility-breaking way, we'll get a compilation failure instead of memory corruption and security vulnerabilities. Progress!
- Some methods that would crash if they were ever called were removed.
    - `FirefoxAccounts.Companion.from` and `FirefoxAccounts.getSyncKeys` only existed when compilation flags we don't provide were enabled, and really only make sense if we were going to integrate this into the pre-oauth Firefox desktop sync key flow
    - As a result, the `SyncKeys` type is removed.
- `OAuthInfo` and `Profile` are now `data class`es, although I don't know how much this actually matters to anybody.
- `FirefoxAccount.getOAuthToken` got renamed to `FirefoxAccount.getCachedOAuthToken` to better reflect it's behavior, and now returns null on error instead of throwing.
- `FirefoxAccount` and `Config` both use `java.lang.AutoCloseable` instead of `java.io.Closeable` now.
- The race in the FxA sample mentioned in #1085 is fixed, as is a similar race in the logins sample (This doesn't use the fix suggested in the bug, which still had a race, just a smaller one).

There are also a number of fixes that happened on the other part of this patch (https://github.com/mozilla/application-services/pull/311) such as fixing the fact that we'd leak whenever things returned errors.

I've also updated the documentation in the README's to match the changes. 

r? @csadilek

Note: The release last night of 0.7.0 actually failed because the build took too long, so I'm retriggering it, but this won't work until that retrigger finishes (it works when I do everything locally, of course). I'd hold off on submitting the PR until then, but I've already written it so... Yeah.


samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/MainActivity.kt
```diff
@@ -37,44 +38,50 @@ open class MainActivity : AppCompatActivity(), LoginFragment.OnLoginCompleteList
...
 
-                        acct.getProfile()
+                val profile = whenAccount.await().getProfile(true).await()
+                val txtView: TextView = findViewById(R.id.txtView)
+                launch(UI) {
+                    txtView.text = getString(R.string.signed_in,
+                            "${profile.displayName ?: ""} ${profile.email}")
+                }
+                acct
+            } catch (e: FxaException) {
```
csadilek: It's not ideal that we handle the main auth flow in an Exception block. I think that's something we should fix, but we can file a follow-up issue if that's a big change right now.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/OAuthInfo.kt
```diff
@@ -4,36 +4,20 @@
...
-    class Raw(p: Pointer) : Structure(p) {
-        @JvmField var access_token: Pointer? = null
-        @JvmField var keys: Pointer? = null
-        @JvmField var scope: Pointer? = null
-
-        init {
-            read()
+import org.mozilla.fxaclient.internal.OAuthInfo as InternalOAuthInfo;
+
+data class OAuthInfo(
```
csadilek: Can we add some KDocs here using `@property` to explain what keys, scope etc. are?

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FirefoxAccount.kt
```diff
@@ -3,117 +3,94 @@
...
+import kotlinx.coroutines.experimental.async
+import org.mozilla.fxaclient.internal.FirefoxAccount as InternalFxAcct
 
 /**
  * FirefoxAccount represents the authentication state of a client.
- *
- * @param <T> The type of the value delivered via the FxaResult.
  */
-class FirefoxAccount(override var rawPointer: RawFxAccount?) : RustObject<RawFxAccount>() {
+class FirefoxAccount internal constructor(private val inner: InternalFxAcct): AutoCloseable {
```
csadilek: We'd want some test coverage for this type, but it'll probably make more sense to add tests to the higher-level API when we build it.

thomcc: Test coverage is a bit tricky right now due to https://github.com/mozilla/application-services/issues/262

We could run tests in e.g. androidTest, but not as part of unit tests.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/OAuthInfo.kt
thomcc: Sure. This also points out that we really should be decoding these. `scopes` is a space-separated list, and `keys` is a JSON object. I've done that in the next patch.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/Config.kt
```diff
@@ -3,24 +3,27 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 package mozilla.components.service.fxa
```
csadilek: That's a linting error here. We need a newline between package and imports.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FirefoxAccount.kt
```diff
@@ -3,117 +3,94 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 package mozilla.components.service.fxa
```
csadilek: same linting error.

csadilek:
After upgrading the coroutines lib we also have the plain `launch` deprecated now. Here's are some more details:

https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-scope/index.html

https://medium.com/@elizarov/structured-concurrency-722d765aa952

The first link shows and example activity with the new API. Smaller change, but we can chat about it or land this as is an fix later...

samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/MainActivity.kt
thomcc: I think this should be cleaned up in a follow up.

thomcc:
Linter errors are fixed, everything should be good to go as far as I know.

samples/sync-logins/src/main/java/org/mozilla/samples/sync/logins/MainActivity.kt
```diff
@@ -101,33 +107,34 @@ open class MainActivity : AppCompatActivity(), LoginFragment.OnLoginCompleteList
...
+            val appFiles = this@MainActivity.applicationContext.getExternalFilesDir(null)
             val logins = SyncLoginsClient(appFiles.absolutePath + "/logins.sqlite")
-            val tokenServer = account!!.getTokenServerEndpointURL()!!
+            val tokenServer = account.getTokenServerEndpointURL()
             logins.syncAndGetPasswords(oauthInfo, tokenServer).thenCatch { e ->
                 SyncResult.fromException(e)
             }.whenComplete { syncLogins ->
-                runOnUiThread {
-                    Toast.makeText(this, "Logins success", Toast.LENGTH_SHORT).show()
+                launch(UI) {
```
csadilek: This needs to stay `runOnUiThread` (for now). Otherwise, the block below is not being executed. We have an issue where we're tracking upgrading coroutines #1181 . Let's switch this back to `runOnUiThread` and address later? There's more here that needs to change.

samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/MainActivity.kt
csadilek: Follow-up filed: https://github.com/mozilla-mobile/android-components/issues/1230

PR #1149, opened 2018-10-23
 Closes #772: BrowserToolbar.Button adds inconvenient padding

### LINKED ISSUE #772: BrowserToolbar.Button adds inconvenient padding around image
I found my toolbar butons assets were offset when I changed the toolbar height and it seems the reason for that is that there is padding added to the view:

https://github.com/mozilla-mobile/android-components/blob/f93998788c670b86707d8dee9451a81198a5f15c/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt#L311-L316

I want to be able to declare the assets to be a specific size (with #741?) and center them in the toolbar, rather than having confusing padding added.

---

fwiw, this padding seems to be set in proportion to the toolbar height (which I changed):
https://github.com/mozilla-mobile/android-components/blob/f93998788c670b86707d8dee9451a81198a5f15c/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt#L363-L364

Amejia481:


components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
```diff
@@ -301,20 +301,31 @@ class BrowserToolbar @JvmOverloads constructor(
...
      */
     open class Button(
         imageResource: Int,
         contentDescription: String,
         visible: () -> Boolean = { true },
         @DrawableRes background: Int? = null,
+        private val paddingStart: Int = ACTION_PADDING_DP,
+        private val paddingTop: Int = ACTION_PADDING_DP,
+        private val paddingEnd: Int = ACTION_PADDING_DP,
+        private val paddingBottom: Int = ACTION_PADDING_DP,
```
pocmo: I wonder if we should just go with one padding for all of them for now. It's good enough for our default padding and for removing padding completely.

Amejia481: Yes, we can. I just wanted to be as flexible as possible, for the consumers. Should I change it to just one value?

Amejia481: What about having overloaded constructors? One with the 4 parameters and another that takes just one add the same value for start, top, end, and bottom?

pocmo: > Yes, we can. I just wanted to be as flexible as possible, for the consumers. 

Yeah, this is more flexible. But it's also more noisy by adding 4 more parameters. I wonder if a `Padding` data class in the base component would be helpful?

Amejia481: Let's add the class `Padding`, I was thinking the same!

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
```diff
@@ -317,13 +318,26 @@ class BrowserToolbar @JvmOverloads constructor(
...
         override fun createView(parent: ViewGroup): View {
             val view = super.createView(parent)
-
-            val padding = view.resources.pxToDp(ACTION_PADDING_DP)
-            view.setPadding(padding, padding, padding, padding)
+            val (start, top, end, bottom) = padding
+
+            with(view.resources) {
+                view.setPadding(
+                    pxToDp(start),
```
mcomella: The Android API is `setPadding(left, ...)` not `setPadding(start, ...)`, right? So I think you'd have to manually swap left/right to start/end based on the locale (unless this method exists already and I haven't seen it).

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
```diff
@@ -317,13 +318,26 @@ class BrowserToolbar @JvmOverloads constructor(
         contentDescription: String,
         visible: () -> Boolean = { true },
         @DrawableRes background: Int? = null,
+        private val padding: Padding = Padding(
```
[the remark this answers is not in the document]
mcomella: Yes but we additionally need this in:
- `ToggleButton`
- `ActionImage`

Also, is there any reason this shouldn't be support in the lowest level class, like `Toolbar.ActionButton`, rather than `BrowserToolbar.Button`?

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
Amejia481: 👍

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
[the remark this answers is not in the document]
Amejia481: I updated the PR with the changes!

PR #1112, opened 2018-10-19
Bug 1497804 - Integrate Glean with Gradle

Dexterp37:
This PR enables Glean's integration with the application's Gradle build system. This allows parsing the application's `metrics.yaml` file and producing the related specific metrics API.
A very basic sample application is also provided, in order to test the integration of the systems.

PR #1154, opened 2018-10-23
Integrate glean_parser into the build

mdboom:
While #1112 integrated a build step for the new "glean" telemetry, it only contained a dummy telemetry parser.  This pulls the "real" telemetry parser from https://github.com/mozilla/glean_parser and uses that.

NOTE: This will not work until https://github.com/mozilla/glean_parser/pull/1 is merged and tagged.


samples/glean/metrics.yaml
```diff
@@ -1,22 +1,17 @@
-# This file contains a definition of the scalar probes that are recorded in Telemetry.
```
Dexterp37: Yeah, this comment was weird :) Can we still retain some similar comment at the top of the file?

components/service/glean/src/main/java/mozilla/components/service/glean/EventMetricType.kt
```diff
@@ -24,12 +24,13 @@ internal const val MAX_LENGTH_OBJECT_ID = 20
  * data and making sure that limits are enforced.
  */
 data class EventMetricType(
-    override val applicationProperty: Boolean,
-    override val disabled: Boolean,
-    override val group: String,
+    override val applicationProperty: Boolean = false,
```
Dexterp37: Does it make sense for these defaults to be in `CommonMetricData` instead?

samples/glean/metrics.yaml
```diff
@@ -1,22 +1,17 @@
...
-      - accessibility@mozilla.com
-      - jteh@mozilla.com
-    record_in_processes:
-      - 'main'
-    release_channel_collection: opt-out
-    keyed: false
-    cpp_guard: 'ACCESSIBILITY'
\ No newline at end of file
+      - telemetry-client-dev@mozilla.com
+    objects:
```
Dexterp37: Can we update the sample application as well to make use of these properties?

components/service/glean/src/main/java/mozilla/components/service/glean/EventMetricType.kt
mdboom: If I do that, I get the message "Property initializers are not allowed on interfaces".

An alternative might be to not skip defaults in the `glean_parser` generator, but always specify them explicitly.  Then at least the defaults are only defined in one place (glean_parser).  I don't know if there are code size implications to that, though.

Dexterp37: Mh, @pocmo , do you have any advice for this? I would be for skipping defaults and always generate named arguments for all the parameters. Do you think this would affect the code size too much?

components/service/glean/src/main/java/mozilla/components/service/glean/EventMetricType.kt
```diff
@@ -24,12 +24,13 @@ internal const val MAX_LENGTH_OBJECT_ID = 20
...
+    override val disabled: Boolean = false,
+    override val category: String,
     override val name: String,
-    override val sendInPings: List<String>,
-    override val userProperty: Boolean,
+    override val sendInPings: List<String> = listOf("default"),
+    override val userProperty: Boolean = false,
+    val objects: List<String>,
```
Dexterp37: Darn! You just found a bug in my implementation of events. `objectId` from the `record()` function of the EventMetricType must be in the list of `objects`. We can either fix it here or we can file a bug and fix it later, up to you!

mdboom: I can fix it here.  Good chance for me to learn more Kotlin.

components/service/glean/src/main/java/mozilla/components/service/glean/EventMetricType.kt
```diff
@@ -62,12 +63,11 @@ data class EventMetricType(
             return
         }
 
-        val truncatedObjectId = objectId.let {
-            if (it.length > MAX_LENGTH_OBJECT_ID) {
-                logger.warn("objectId parameter exceeds maximum string length, truncating.")
-                return@let it.substring(0, MAX_LENGTH_OBJECT_ID)
-            }
-            it
+        // We don't need to check that the objectId is short, since that
```
Dexterp37: Good point. This means we can get rid of `MAX_LENGTH_OBJECT_ID` as well at the top of this file, right?

mdboom: Good point.

samples/glean/metrics.yaml
mdboom: Done.

samples/glean/metrics.yaml
mdboom: I've put a comment back in here.

pocmo:
This is ready to land now, right? Do we want to squash the commits or do we want to keep them separate?

mdboom:
Yep.  This is ready to merge.  I've squashed the commits.

pocmo:
Awesome, thanks! Merging...

PR #1156, opened 2018-10-23
Closes #1106: Enhance RequestInterceptor to load alternative URL

### LINKED ISSUE #1106: Enhance RequestInterceptor to load alternative URL
We already support providing an `InterceptionResponse` to load custom data instead of loading the requested URL. This enhancement would allow returning an alternative URL to be loaded. This is really useful when integrating FxA in the reference browser where we want to show a custom success page after successfully intercepting the OAuth redirect URL. I can do that with a content redirect as a workaround, but using Gecko that will briefly display the data URL in the toolbar.

csadilek:
Turned `InterceptionResponse` into a sealed class. Adding just a `url` field would have allowed for incorrect API usage i.e. one could've provided both a `url` and `data` which wouldn't make sense.

This is a breaking API change, but it's minimal: 
```Kotlin
return InterceptionResponse("data") 
```
becomes
```Kotlin
return InterceptionResponse.Content("data")
```

I will add this to the changelog if we decide to land this.

### DIFF
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -14,6 +14,7 @@ import mozilla.components.concept.engine.HitResult
 import mozilla.components.concept.engine.Settings
 import mozilla.components.concept.engine.history.HistoryTrackingDelegate
 import mozilla.components.concept.engine.request.RequestInterceptor
+import mozilla.components.concept.engine.request.RequestInterceptor.InterceptionResponse
 import mozilla.components.support.ktx.android.util.Base64
 import mozilla.components.support.ktx.kotlin.isEmail
 import mozilla.components.support.ktx.kotlin.isGeoLocation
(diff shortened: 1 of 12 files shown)


csadilek:
There is one potential footgun here: Unconditional interceptor logic could redirect indefinitely. We could prevent that by checking if the currentUrl is the same as redirect url and simply do nothing in that case.

csadilek:
> We could prevent that by checking if the currentUrl is the same as redirect url and simply do nothing in that case.
Actually, that's probably not worth doing. We'd have to get much smarter in checking if the URLs are the same (e.g. trailing slash or none) and we will likely not be able to prevent all cases of this.

jonalmeida:
> > We could prevent that by checking if the currentUrl is the same as redirect url and simply do nothing in that case.
> 
> Actually, that's probably not worth doing. We'd have to get much smarter in checking if the URLs are the same (e.g. trailing slash or none) and we will likely not be able to prevent all cases of this.

SystemEngine's interceptor throws an error if there are too many redirects, maybe that's something we could ask Gecko to also handle?

csadilek:
> SystemEngine's interceptor throws an error if there are too many redirects, maybe that's something we could ask Gecko to also handle?

This isn't technically a redirect though. I tested this with SystemEngine as well. It will just go on forever :). But of course I would have to write an interceptor that unconditionally redirects. I no longer expect the framework to catch that for me...

jonalmeida:
Just some tests to fix. 👌

csadilek:
Fixed the test and added a changelog entry.

PR #1164, opened 2018-10-25
WIP first pass on the HistoryTrackingDelegate implementation backed by Rust Places

grigoryk:
A quick pass to see how things might fit together. It works!


grigoryk:
https://github.com/mozilla/application-services/pull/327 adds an API we'd need here for the system engine, and fixes a bug around titles not being recorded.

samples/browser/src/geckoBeta/java/org/mozilla/samples/browser/Components.kt
```diff
@@ -14,10 +15,12 @@ import org.mozilla.samples.browser.request.SampleRequestInterceptor
...
-        GeckoEngine(applicationContext, defaultSettings)
-    }
+    // TODO uncomment this; testing this with the SystemEngine because it has a more complete API.
+//    override val engine: Engine by lazy {
+//        val defaultSettings = DefaultSettings().apply {
+//            requestInterceptor = SampleRequestInterceptor(applicationContext)
+//            historyTrackingDelegate = PlacesHistoryTrackingDelegate(applicationContext)
+//        }
+//        GeckoEngine(applicationContext, defaultSettings)
+//    }
```
pocmo: FYI: In the bottom left corner in Android Studio there should be a "build variant" side bar. There you should be able to switch the build variant of `samples-browser`. In this case you can just use `systemUniversalDebug`.

grigoryk:
@pocmo @csadilek this is ready for a review. This works for me locally with a HistoryTrackingFeature enabled in the sample browser with a SystemEngine, and will track history with GeckoView once #1238 lands.

To get tests to work for `PlacesHistoryStorage` - or anything that uses `PlacesConnection` - we'd probably first need to change how it loads the native library, e.g. wrap it in a try/catch to let tests work and mock it, like how we do in the `FxaClient`. Otherwise, as long as `PlacesConnection` is initialized at all during a test, it fails since it can't load the library.

Alternatively, we could supply library artifacts for the test machine architectures, and wire that up...? Not entirely sure how that would look.

grigoryk:
Relevant app services issue: https://github.com/mozilla/application-services/issues/342

grigoryk:
Pushed out an app-services PR to help out with tests and docs. https://github.com/mozilla/application-services/pull/343

grigoryk:
Still wrapping my head around coroutines, dispatchers and whatnot. Regarding the `Deferred` change, it complicates the implementations a little bit. The onus is on the implementations to schedule jobs. Places layer doesn't inherit a CoroutineScope (and the dispatcher) of the caller, and instead uses `Dispatchers.IO` (or some other thread pool in the future). While this helps with predictability of behaviour, it removes flexibility from the callers around scheduling and cancelling work. On another hand, it helps prevent some footguns of doing blocking IO on a UI thread, it simplifies interacting with the storage layer, and forces use of suspend functions on the callers whenever appropriate.

We don't always care about what happens with the work we schedule. For recording a visit, we really just want to kick off that work and move on. For retrieving visits, we obviously need to wait for it to complete. It's possible we might also want to be able to cancel that work (if an engine view is destroyed?). For autocomplete results, we care about the results and definitely want to have cancellation machinery in place (how deep it should go is another question... down into the places rust lib..? see https://github.com/mozilla/application-services/issues/265). But, this can also be accomplished at the consumer level, and not by using `Deferred` or callbacks. E.g. that's the approach taken in `SuggestionProvider.onInputChanged` - see https://github.com/pocmo/android-components/commit/e9df44aaab4d817a285533900cf928851f379517.

For now, I think it makes most sense to approach this on a method-by-method basis.

csadilek:
> While this helps with predictability of behaviour, it removes flexibility from the callers around scheduling and cancelling work

@grigoryk @pocmo and I talked about this as well today. It seems to me that it should be perfectly reasonable for a library API to return `Deferred`, as opposed to making all calls synchronous. Maybe the sweet spot here is that the library should use a default coroutine scope and dispatcher, but also allow callers to specify an alternative? Still wrapping my head around it as well...

csadilek:
> While this helps with predictability of behaviour, it removes flexibility from the callers around scheduling and cancelling work

@grigoryk @pocmo and I talked about this as well today. It seems to me that it should be perfectly reasonable for a library API to return `Deferred`, as opposed to making all calls synchronous. Maybe the sweet spot here is that the library should use a default coroutine scope and dispatcher, but also allow callers to specify an alternative? Still wrapping my head around it as well...

components/browser/storage-sync-places/src/main/java/mozilla/components/browser/storage/places/Conn.kt
```diff
@@ -0,0 +1,25 @@
...
+import org.mozilla.places.PlacesConnection
+
+const val DB_NAME = "places.sqlite"
+
+/**
+ * A singleton which holds a shared reference to [PlacesConnection].
+ */
+internal object Conn {
+    @Volatile
+    private var placesConnection: PlacesConnection? = null
```
csadilek: It seems we need to call `close` on `PlacesConnection`. Need to think about how we do this otherwise we'd leak every time the app process is killed/restarted.  Maybe expose a `close` method on `Conn`?

grigoryk: @csadilek yup, good call. I've updated `HistoryStorage` to let us do that.

components/browser/storage-sync-places/src/main/java/mozilla/components/browser/storage/places/Conn.kt
```diff
@@ -0,0 +1,32 @@
...
+        placesConnection = PlacesConnection(dbPath)
+        return@synchronized placesConnection!!
+    }
+
+    fun close() = synchronized(this) {
+        if (placesConnection != null) {
+            placesConnection!!.close()
+            placesConnection = null
+        }
+    }
```
pocmo: Slightly unclean: We have a `close()` here and the user of `PlacesConnection` could call `close` too; creating a situation where our connection holder will return a closed instance.

components/browser/storage-sync-places/src/main/java/mozilla/components/browser/storage/places/PlacesHistoryStorage.kt
```diff
@@ -0,0 +1,79 @@
...
+import kotlinx.coroutines.experimental.Dispatchers
+import kotlinx.coroutines.experimental.launch
+import mozilla.components.concept.storage.HistoryStorage
+import mozilla.components.concept.storage.PageObservation
+import mozilla.components.concept.storage.VisitType
+import org.mozilla.places.PlacesAPI
+import org.mozilla.places.VisitObservation
+
+class PlacesHistoryStorage(context: Context)
+    : TestablePlacesHistoryStorage(Conn.getInstance(context.getDatabasePath(DB_NAME).canonicalPath))
```
pocmo: What exactly is the reason for `TestablePlacesHistoryStorage`? I assume that you do not want to make `PlacesHistoryStorage` open. If this is just the case for mocking then creating this file in the component will allow you to just mock the non-open class: https://github.com/mozilla-mobile/android-components/blob/master/components/browser/engine-gecko/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker

components/browser/storage-sync-places/src/main/java/mozilla/components/browser/storage/places/PlacesHistoryStorage.kt
```diff
@@ -0,0 +1,79 @@
...
+            places.getVisitedUrlsInRange(
+                start = 0,
+                end = System.currentTimeMillis(),
+                includeRemote = true
+            )
+        }
+    }
+
+    override fun cleanup() {
+        Conn.close()
```
pocmo: nit: if we close the connection here: Do we want to cancel all jobs of this coroutine scope too? You may close a connection that is in use here.

settings.gradle
```diff
@@ -51,6 +51,7 @@ setupProject(':browser-menu', 'components/browser/menu', 'A customizable menu fo
 setupProject(':browser-search', 'components/browser/search', 'Search plugins and companion code to load, parse and use them.')
 setupProject(':browser-session', 'components/browser/session', 'An abstract layer hiding the actual browser engine implementation.')
 setupProject(':browser-storage-memory', 'components/browser/storage-memory', 'A memory-backed implementation of core data storage.')
+setupProject(':browser-storage-sync-places', 'components/browser/storage-sync-places', 'A syncable, Rust Places-backed implementation of core data storage.')
```
pocmo: This needs to be added to `artifacts.yml`.

pocmo: And a changelog entry.. :)

grigoryk:
I've changed how the underlying places lib is wrapped (it's much clearer now than when @csadilek and @pocmo took a look), added tests and did some other cleanup work. Will land this once TC is green.

Issue #1109, opened 2018-10-19
Awesomebar component


pocmo:
We are going to need some kind of awesomebar component. This here is more like a meta issue to write some thoughts down and discuss them.

## Awesomewhat?

For context: With awesomebar I mean the area that appears below the URL input field as soon as you start typing. It usually shows related results from history, bookmarks, other synced data, extensions, search engines and other things. In Chrome this part of the browser is called omnibox.

## Concept component

The awesomebar is a place where the different browser products may be opinionated and therefore it is unlikely that we will have a single component/UI for it. It is likely that we want to define a shared awesombar interface in a `concept-awesombar` like we did for the toolbar and tabstray. With that we can have multiple implementations while other components only know about the concept component interfaces.

The WebExtension omnibox API maybe be a similar API that would be worth looking at:
* https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/omnibox
* https://developer.chrome.com/extensions/omnibox

## Composition

The awesombar will display data from a variety of sources. So there will be some setup and plumbing required. So far we try to make this easier for apps by hiding this "additional code" in a feature component (e.g. `feature-toolbar` takes care of getting all the data and updating the toolbar if needed; `feature-tabs` does the same with a tabstray). So in the case here we could create a `feature-awesomebar` component. However this may require that we add some kind of abstraction on top of the components that provide data for the awesomebar in order to avoid having the feature component having a hard dependency on everything. Something like `concept-awesomebar-provider`?

In addition to that the feature component would be helpful to setup the interaction with the toolbar. The awesomebar needs to appear/disappear based on toolbar events and we may not necessarily keep a strong dependency on the toolbar component for that.

PR #1166, opened 2018-10-25
Issue #1109: Add concept component for implementing awesome bar components.

pocmo:
This is the concept piece of #1109. Locally I already have `browser-awesomebar` and `feature-awesomebar` components and this is the "concept" that was required to implement the awesomebar and connect it to the other components.

### DIFF
--- a/components/concept/awesomebar/README.md
+++ b/components/concept/awesomebar/README.md
@@ -1,0 +1,47 @@
+# [Android Components](../../../README.md) > Concept > Awesomebar
+
+An abstract definition of an awesome bar component.
+
+## Usage
+
+### Setting up the dependency
+
+Use Gradle to download the library from [maven.mozilla.org](https://maven.mozilla.org/) ([Setup repository](../../../README.md)):
+
+```Groovy
+implementation "org.mozilla.components:concept-awesomebar:{latest-version}"
+```
+
+### Implementing an Awesome Bar
+
+An Awesome Bar can be any [Android View](https://developer.android.com/reference/android/view/View.html) that implements the `AwesomeBar` interface.
+
+An `AwesomeBar` implementation needs to react to the following events:
+
+* `onInputStarted()`: The user starts interacting with the awesome bar by entering text in the [toolbar](../toolbar/README.md). This callback is a good place to initialize code that will be required once the user starts typing.
+* `onInputChanged(text: String)`: The user changed the text in the [toolbar](../toolbar/README.md). The awesome bar implementation should update its suggestions based on the text entered now.
+* `onInputCancelled()`: The user has cancelled their interaction with the awesome bar. This callback is a good place to free resources that are no longer needed.
+
+The suggestions an awesome bar displays are provided by an `SuggestionProvider`. Those providers are passed by the app (or another component) to the awesome bar by calling `addProviders()`. Once the text changes the awesome bar queries the `SuggestionProvider` instances and receives a list of `Suggestion` objects.
+
+Once the user selects a suggestion and the awesome bar wants to stop the interaction it can invoke the callback provided via the `setOnStopListener()` method. This is required as the awesome bar implementation is unaware of how it gets displayed and how interaction with it should be stopped (e.g. leaving the [toolbar's](../toolbar/README.md) editing mode).
+
+### Suggestions
+
+A `Suggestion` object contains the data required to be displayed and callbacks for when a suggestion was selected by the user.
+
+It is up to the suggestion or its provider to define the behavior that should happen in that situation (e.g. loading a URL, performing a search, switching tabs..).
+
+All data in the `Suggestion` object is optional. It is up to the awesome bar implementation to handle missing data (e.g. show the `description` instead of a missing `title`).
+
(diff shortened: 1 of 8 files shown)


components/concept/awesomebar/README.md
```diff
@@ -0,0 +1,47 @@
...
+* `onInputChanged(text: String)`: The user changed the text in the [toolbar](../toolbar/README.md). The awesome bar implementation should update its suggestions based on the text entered now.
+* `onInputCancelled()`: The user has cancelled their interaction with the awesome bar. This callback is a good place to free resources that are no longer needed.
+
+The suggestions an awesome bar displays are provided by an `SuggestionProvider`. Those providers are passed by the app (or another component) to the awesome bar by calling `addProviders()`. Once the text changes the awesome bar queries the `SuggestionProvider` instances and receives a list of `Suggestion` objects.
+
+Once the user selects a suggestion and the awesome bar wants to stop the interaction it can invoke the callback provided via the `setOnStopListener()` method. This is required as the awesome bar implementation is unaware of how it gets displayed and how ht interaction should be stopped (e.g. leaving the [toolbar's](../toolbar/README.md) editing mode).
```
jonalmeida: Typo: ..and how **ht** interaction should..

> to stop the interaction it can invoke the callback..
In the next sentence, it says this is required, so maybe replace the word 'can' with 'will'?

components/concept/awesomebar/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Suggestions
+
+A `Suggestion` object contains the data required to be displayed and callbacks for when a suggestion was selected by the user.
+
+It is up to the suggestion or its provider to define the behavior that should happen in that situation (e.g. loading a URL, performing a search, switching tabs..).
+
+All data the in the `Suggestion` object is optional. It is up to the awesome bar implementation to handle missing data (e.g. show the `description` instead of a missing `title`).
```
jonalmeida: typo: All data ~the~ in the `Suggestion` object..

components/concept/awesomebar/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+Every `Suggestion` has an `id`. By default the `Suggestion` will generate a random ID. This ID can be used by the awesome bar to determine whether two suggestions are the same even though they are containing different/updated data. For example a `Suggestion` showing search suggestions from a search engine might use a constant ID when it is showing new search suggestions - to avoid the awesome bar implementation animating the previous suggestion leaving and a new suggestion appearing.
+
+### Implementing a Suggestion Provider
+
+For implementing a Suggestion Provider the `SuggestionProvider` interface needs to be implemented. The awesome bar forwards the events it receives to every provider: `onInputStarted()`, `onInputCancelled()`, `onInputChanged(text: String)`. A provider is required to return a list of `Suggestion` objects from `onInputChanged()`. This implementation can be synchronous. The awesome bar implementation takes care of performing the requests from worker threads.
```
jonalmeida: Random thoughts: would there be a need to provide ordering from different Suggestion Providers? e.g. bookmarks provider results comes before search provider results. Should a `Suggestion` also hold a provider ID so we can sort them that way?

components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
```diff
@@ -0,0 +1,121 @@
...
+     */
+    data class Suggestion(
+        val id: String = UUID.randomUUID().toString(),
+        val title: String? = null,
+        val description: String? = null,
+        val icon: ((width: Int, height: Int) -> Bitmap?) = { _, _ -> null },
+        val chips: List<Chip> = emptyList(),
+        val flags: Set<Flag> = emptySet(),
+        val onSuggestionClicked: (() -> Unit)? = null,
+        val onChipClicked: ((Chip) -> Unit)? = null
```
jonalmeida: More random thoughts: When a chip is clicked, in the future implementation for the Awesome Bar, will that also invoke the AwesomeBar's onStopListener or would that only happen when a suggestion is clicked?

When a chip is clicked, will it need reference to the suggestion it's coming from in the lambda? 🤔

components/concept/awesomebar/README.md
pocmo: Yeah, I want to land this in a follow-up. I do have a TODO for that in the `browser-awesomebar` component. I think we will need some kind of ranking on the `Suggestion` level to be able to mix suggestions. However I am not really sure how that should look like.

components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
pocmo: Yep! Currently `browser-awesomebar` invokes the listener for both a chip click or a suggestion click. I was wondering whether to return a `Boolean` from the lambda to let the suggestion/chip/provider control that - but so far this wasn't needed.

> When a chip is clicked, will it need reference to the suggestion it's coming from in the lambda? 🤔

Right now that wasn't needed. The `Suggestion` with the `Chip`s are created in the `SuggestionProvider` and there the actual lambda has all the information.

Here are the two providers I implemented so far. One for "open tabs" and one for search suggestions:
https://gist.github.com/pocmo/7f1405c4a8491b09b91e994b23226cdf

PR #1169, opened 2018-10-25
Fix broken link in new architecture doc [ci skip]

csadilek:


pocmo:
Thanks for catching that! :)

PR #1171, opened 2018-10-25
Add string metric type and string storage backend [glean]

mdboom:
Also refactors EventMetricType to reduce code duplication.

Fixes:

https://bugzilla.mozilla.org/show_bug.cgi?id=1497817
https://bugzilla.mozilla.org/show_bug.cgi?id=1497826

Depends on: #1154


components/service/glean/src/main/java/mozilla/components/service/glean/CommonMetricData.kt
```diff
@@ -4,15 +4,34 @@
...
-    val userProperty: Boolean?
+    val applicationProperty: Boolean
+    val disabled: Boolean
+    val category: String
+    val name: String
+    val sendInPings: List<String>
+    val userProperty: Boolean
+}
+
+public fun shouldRecord(metric: CommonMetricData, logger: Logger): Boolean {
```
Dexterp37: This could be a protected function on the `CommonMetricData`. What do you think?

components/service/glean/src/main/java/mozilla/components/service/glean/StringMetricType.kt
```diff
@@ -0,0 +1,59 @@
...
+ * This implements the developer facing API for recording string metrics.
+ *
+ * Instances of this class type are automatically generated by the parsers at build time,
+ * allowing developers to record values that were previously registered in the metrics.yaml file.
+ *
+ * The string API only exposes the [record] method, which takes care of validating the input
+ * data and making sure that limits are enforced.
+ */
+data class StringMetricType(
+    override val applicationProperty: Boolean = false,
```
Dexterp37: Mh, I think I found a Kotlin way to have defaults declared on the `CommonMetricData` interface. According to [the docs](https://kotlinlang.org/docs/reference/interfaces.html#properties-in-interfaces), we can do something like this:

```
interface CommonMetricData {
   ...
    val disabled: Boolean?
        get() = false
    val category: String?
    ....
}
```

Since we're always using the same defaults for each metric type, maybe we should just refactor this right now? What do you think? Do you want to address that in this PR or should we file a follow-up and take it there, before we use other types?

samples/glean/src/main/java/org/mozilla/samples/glean/MainActivity.kt
```diff
@@ -14,9 +14,18 @@ open class MainActivity : AppCompatActivity() {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_main)
 
+        GleanMetrics.Basic.os.setValue("Android")
```
Dexterp37: This looks so cool. 🆒

samples/glean/metrics.yaml
```diff
@@ -1,22 +1,33 @@
...
+    notification_emails:
+      - telemetry-client-dev@mozilla.com
+    objects:
+      - object1
+      - object2
+    extra_keys:
+      key1: "This is key one"
+      key2: "This is key two"
+
+basic:
```
Dexterp37: Should this group end up in a separate file? This is not for this PR, but we should clarify the expected behaviour

components/service/glean/src/main/java/mozilla/components/service/glean/StringMetricType.kt
```diff
@@ -0,0 +1,59 @@
...
+) : CommonMetricData {
+    private val logger = Logger("glean/StringMetricType")
+
+    /**
+     * Set a string value.
+     *
+     * @param value This is a user defined string value. The maximum length of
+     * this string is defined the metrics.yaml file by max_length.
+     */
+    public fun setValue(value: String) {
```
Dexterp37: nit: what do you think about using simple verbs for this? for example, we could name this method `set`, similarly to the `record` from the `EventMetricType`

components/service/glean/src/main/java/mozilla/components/service/glean/StringMetricType.kt
```diff
@@ -0,0 +1,59 @@
...
+ * data and making sure that limits are enforced.
+ */
+data class StringMetricType(
+    override val applicationProperty: Boolean = false,
+    override val disabled: Boolean = false,
+    override val category: String,
+    override val name: String,
+    override val sendInPings: List<String> = listOf("default"),
+    override val userProperty: Boolean = false,
+    val maxLength: Int
```
Dexterp37: Mh, are we supposed to let users define a limit here or should we settle for one ourselves? This makes sense, but I wonder if that's what we intend to do. I think the spec is not clear wrt this.

components/service/glean/src/main/java/mozilla/components/service/glean/CommonMetricData.kt
mdboom: Sure -- I didn't realise interfaces could have method implementations, but it seems they can.  I agree that would be cleaner.

components/service/glean/src/main/java/mozilla/components/service/glean/StringMetricType.kt
mdboom: Thanks for finding that.  I actually looked and didn't come across this in my searching.  Let's try to do this now -- it will make adding more metric types easier going forward.

mdboom: Trying this, I'm not sure this makes things any better.  Putting the `get() = false` in the interface indeed sets default behavior for the interface, but you still have to override it in the data class so that you can give kotlin a place to store the actual value for the case where it's different from the default.  And you still need to provide the default in the data class to generate a default value for the constructor's arguments.  So in short, I don't think making this change is useful for us unless there were specific metric types that wouldn't need to touch / store values for these fields (which I don't think will ever be the case).

This leads me back to my earlier thought: we could have `glean_parser` *always* write out the values for arguments, even if they are the default.  Then we wouldn't have to declare the defaults in each metric type implementation -- the defaults would only be stored and applied by the glean_parser when loading a `metrics.yaml` file...

What do you think?

samples/glean/metrics.yaml
mdboom: This does end up in a separate generated Kotlin file.  Are you thinking it should live in a separate (input) `metrics.yaml` as well?

components/service/glean/src/main/java/mozilla/components/service/glean/StringMetricType.kt
mdboom: Sure.  No problem.  I was going by the API docs which say `setValue`.  I can update the docs as well.

components/service/glean/src/main/java/mozilla/components/service/glean/StringMetricType.kt
Dexterp37: Oh, good point. I like the idea of the parsers always writing values for arguments. But let's tackle this right after this PR is merged and not block on it. Meanwhile, let's leave the defaults as you did initially.

samples/glean/metrics.yaml
Dexterp37: 🤦‍♂️ sorry, my bad! No, I meant separate generated Kotlin files :)

components/service/glean/src/main/java/mozilla/components/service/glean/StringMetricType.kt
mdboom: It looks like the spec had this at one point, but it looks like @georgf removed it on 2018-09-23.  I guess the question is whether we should just have a fixed global limit or we want the complexity of enforcing it from a configurable variable.

Dexterp37: I would be for removing it: otherwise users of the API might be tempted to use unreasonable limits to force the collection of some specific strings. Let's define a limit ourselves: we can start by using the one from [Desktop scalars](https://searchfox.org/mozilla-central/rev/e0c879c86b95bdc752b1dbff6088169735674e4a/toolkit/components/telemetry/core/TelemetryScalar.cpp#104), `50`. We can always revisit later.

components/service/glean/src/main/java/mozilla/components/service/glean/StringMetricType.kt
mdboom: Sounds like a plan.

components/service/glean/src/main/java/mozilla/components/service/glean/StringMetricType.kt
mdboom: Ok.  Going with that...

mdboom:
I think this addresses all of @Dexterp37's comments now and is ready for a final review.

components/service/glean/src/main/java/mozilla/components/service/glean/CommonMetricData.kt
```diff
@@ -4,15 +4,34 @@
...
-    val sendInPings: List<String>?
-    val userProperty: Boolean?
+    val applicationProperty: Boolean
+    val disabled: Boolean
+    val category: String
+    val name: String
+    val sendInPings: List<String>
+    val userProperty: Boolean
+
+    public fun shouldRecord(logger: Logger): Boolean {
```
pocmo: nit: `public` is redundant and the default. :)

components/service/glean/src/main/java/mozilla/components/service/glean/StringMetricType.kt
```diff
@@ -0,0 +1,61 @@
...
+) : CommonMetricData {
+    private val logger = Logger("glean/StringMetricType")
+
+    /**
+     * Set a string value.
+     *
+     * @param value This is a user defined string value. The maximum length of
+     * this string is defined the metrics.yaml file by max_length.
+     */
+    public fun set(value: String) {
```
pocmo: nit: `public` :)

Issue #1083, opened 2018-10-17
Flaky test on engine-gecko-nightly GeckoEngineSessionTest.kt:272


Amejia481:
This test fun `saveState()` have been failing intermittently  GeckoEngineSessionTest.kt:272
```
:support-base:packageDebugAssets
:support-ktx:mergeDebugShaders
:support-ktx:compileDebugShaders
:support-ktx:generateDebugAssets
:support-ktx:packageDebugAssets
:support-utils:mergeDebugShaders
:support-utils:compileDebugShaders
:support-utils:generateDebugAssets
:support-utils:packageDebugAssets
:browser-engine-gecko-nightly:mergeDebugAssets
:browser-engine-gecko-nightly:generateDebugUnitTestConfig
:browser-engine-gecko-nightly:mockableAndroidJar
:browser-engine-gecko-nightly:processDebugUnitTestJavaRes
:browser-engine-gecko-nightly:transformClassesAndResourcesWithPrepareIntermediateJarsForDebug
:browser-engine-gecko-nightly:testDebugUnitTest

mozilla.components.browser.engine.gecko.GeckoEngineSessionTest > saveState FAILED
    java.lang.UnsatisfiedLinkError at GeckoEngineSessionTest.kt:272
42 tests completed, 1 failed
:browser-engine-gecko-nightly:testDebugUnitTest FAILED
FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':browser-engine-gecko-nightly:testDebugUnitTest'.
> There were failing tests. See the report at: file:///build/android-components/components/browser/engine-gecko-nightly/build/reports/tests/testDebugUnitTest/index.html
* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
* Get more help at https://help.gradle.org
BUILD FAILED in 5m 47s
265 actionable tasks: 261 executed, 4 up-to-date
[taskcluster 2018-10-17 14:29:09.572Z] === Task Finished ===
[taskcluster 2018-10-17 14:29:09.572Z] Unsuccessful task run with exit code: 1 completed in 559.804 seconds
```

pocmo:
Locally or on taskcluster? Or both? :)

Amejia481:
It's only happening on taskcluster.
I've tried to simulated it on my machine without luck

Amejia481:
[...]

PR #1172, opened 2018-10-25
Issue #1083: Improved test fix to be independent of post order.....

csadilek:
I think we already saw a test timeout due to this on a master push. This should be a better way to workaround the problem where we are ignoring failing runnables. If our runnable fails, the assertion will fail anyway.


csadilek:
6 successful runs / no timeouts. Landing...

Amejia481:
🚢 !

Issue #1177, opened 2018-10-26
CrashReporter: Allow setting theme for dialog


pocmo:
This will give the app more control over how the dialog should look like.

PR #1179, opened 2018-10-26
Crash reporter improvements

### LINKED ISSUE #1042: Crash Reporter Prompt: Add ability to display extra text.
Focus wants to display a prompt like this:

image

Currently we do not have the ability to show extra text ("As a private browser..."). This is something we could add to `PromptConfiguration`.

pocmo:
Closes #1042, #1177, #1178.

Ads a sample app `samples-crash` to show integration and for testing crash reporter.


pocmo:
I should add those things to the changelog :)

components/lib/crash/src/main/java/mozilla/components/lib/crash/prompt/CrashReporterActivity.kt
```diff
@@ -70,12 +77,12 @@ class CrashReporterActivity : AppCompatActivity() {
             return
         }
 
-        launch(UI) {
-            launch(CommonPool) {
-                crashReporter.submitReport(crash)
-            }.join()
+        GlobalScope.launch {
```
csadilek: Did some reading on the new APIs. Seems like we should define a CoroutineScope now? Basically implement `CoroutineScope` and override `coroutineContext`. So that execution lifecycle can be tied to this activity: 
https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-scope/index.html
https://medium.com/@elizarov/structured-concurrency-722d765aa952

Use GlobalScope seems to be discouraged....

components/lib/crash/src/test/java/mozilla/components/lib/crash/prompt/CrashReporterActivityTest.kt
```diff
@@ -38,13 +40,18 @@ class CrashReporterActivityTest {
...
             .start()
             .resume()
             .get()
 
         val closeButton = activity.findViewById<Button>(R.id.closeButton)
         closeButton.performClick()
 
+        runBlocking {
+            delay(100)
+        }
```
csadilek: Applying the changes in the comment above would give us an internal `job` field which we could `join`?
https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-scope/index.html

components/lib/crash/src/main/java/mozilla/components/lib/crash/prompt/CrashReporterActivity.kt
pocmo: Yeah, watched the new `Coroutines` talk of this year's KotlinConf. This makes a lot of sense.

components/lib/crash/src/test/java/mozilla/components/lib/crash/prompt/CrashReporterActivityTest.kt
pocmo: That's nice! 👍

pocmo:
Filed follow-up: #1196

PR #1180, opened 2018-10-26
Don't specify defaults in metric type destructors [glean]

mdboom:
As mentioned [here](https://github.com/mozilla-mobile/android-components/pull/1171#discussion_r228514134), we shouldn't specify any default argument values in the metric type constructors, but leave the specification of defaults to one place at build type (in `glean_parser`).

This is a follow-on to #1171, which should be merged first.

The `glean-parser` side of this change is in https://github.com/mozilla/glean_parser/pull/5


components/service/glean/src/main/java/mozilla/components/service/glean/EventMetricType.kt
```diff
@@ -10,7 +10,7 @@ import mozilla.components.support.base.log.logger.Logger
 // Maximum length of any string value in the extra dictionary, in UTF8 byte sequence length.
 internal const val MAX_LENGTH_EXTRA_KEY_VALUE = 80
 // Maximum length of any passed value string, in UTF8 byte sequence length.
-internal const val MAX_LENGTH_VALUE = 80
+internal const val MAX_LENGTH_EVENT_VALUE = 80
```
Dexterp37: Are we renaming due to name clashing from the consts in `StringMetricType`? If so, maybe a better option is to use [companion objects](https://kotlinlang.org/docs/reference/object-declarations.html#companion-objects) to hold the constants (see [here](https://github.com/mozilla-mobile/android-components/blob/c874c343cab18fe64f45ba8eb9a726f829fe06b9/components/service/fretboard/src/main/java/mozilla/components/service/fretboard/DeviceUuidFactory.kt#L40-L43) for an example)? I'm not a language expert, so I'm not sure if there are implications in doing this. Maybe @pocmo knows better! Regardless, I'd stick with your renaming in this PR and eventually change this in a new bug/PR, so we're not blocked.

components/service/glean/src/main/java/mozilla/components/service/glean/StringMetricType.kt
```diff
@@ -0,0 +1,61 @@
...
+// Maximum length of any passed value string, in UTF8 byte sequence length.
+internal const val MAX_LENGTH_STRING_VALUE = 50
+
+/**
+ * This implements the developer facing API for recording string metrics.
+ *
+ * Instances of this class type are automatically generated by the parsers at build time,
+ * allowing developers to record values that were previously registered in the metrics.yaml file.
+ *
+ * The string API only exposes the [record] method, which takes care of validating the input
```
Dexterp37: nit: "... exposes the [set] ..."

components/service/glean/src/main/java/mozilla/components/service/glean/StringMetricType.kt
```diff
@@ -0,0 +1,61 @@
...
+
+        val truncatedValue = value.let {
+            if (it.length > MAX_LENGTH_STRING_VALUE) {
+                logger.warn("String length ${it.length} > $MAX_LENGTH_STRING_VALUE")
+                return@let it.substring(0, MAX_LENGTH_STRING_VALUE)
+            }
+            it
+        }
+
+        // Delegate storing the event to the storage engine.
```
Dexterp37: nit: "the string to"

components/service/glean/src/main/java/mozilla/components/service/glean/EventMetricType.kt
mdboom: Yep -- I was surprised that these constants were in the same namespace (and thus clashed).  I think members of the class is probably better, but I can do that as a follow-on to this.

PR #1183, opened 2018-10-26
Fix ktlint checking

mdboom:
This is a fix for the issue I reported today in Slack.

If an .editorconfig file exists in any parent directory of the checkout
directory, ktlint will fail because it uses those settings to determine what
the indentation should be.

Where this happens in ktlint:

https://github.com/shyiko/ktlint/blob/41a5067ad04bfee1ae5975ff774b9076a5bb37c0/ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/IndentationRule.kt#L23

(So ktlint is "configuration file free" except when it reads other tools' configuration files...)

This creates an .editorconfig that ktlint will "stop" at, effectively getting
the .editorconfig defaults. It should only be necessary to add settings to this
file if we want to override the defaults.


jonalmeida:
Not sure if #1193 is related to this.

Issue #1121, opened 2018-10-22
[Telemetry] Tag for DebugLogClient is not working


cnevinc:
After using `Log.addSink(AndroidLogSink())`, logcat output no longer honor the tag set in DebugLogClient.

pocmo:
All that should be needed here is passing the tag to the Logger class here:

https://github.com/mozilla-mobile/android-components/blob/c1b23d064dec60721ff21ce154d64a38f4728769/components/service/telemetry/src/main/java/org/mozilla/telemetry/net/DebugLogClient.kt#L16-L17

PR #1189, opened 2018-10-27
Issue #1121: make Logger use the tag set

SubhrajyotiSen:


jonalmeida:
Not sure if it's enough to just update the PR title to have it linked to the ticket or if we need the commit message to also have the '#' in there as well.

SubhrajyotiSen:
@jonalmeida The commit message needs the '#'. I have updated the commit message.

PR #1190, opened 2018-10-27
Issue #1094: Use JEXL in fretboard

### LINKED ISSUE #1094: Use JEXL in fretboard
#454 landed support for JEXL. Now we could consider using it in fretboard.

fercarcedo:
This pull request adds a field `jexl` to the `Experiment.Matcher` class which contains a JEXL expression to be evaluated. The experiment will only be enabled if that expression evaluates to true, along with the rest of matchers

Closes #1094

### DIFF
--- a/components/service/fretboard/build.gradle
+++ b/components/service/fretboard/build.gradle
@@ -28,6 +28,7 @@ dependencies {
 
     implementation project(':support-ktx')
     implementation project(':support-base')
+    implementation project(':lib-jexl')
 
 
     testImplementation Deps.testing_junit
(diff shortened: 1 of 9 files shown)


pocmo:
@fercarcedo That's pretty cool! Thank you for this work. :)

Let me think about the JEXL context. I wonder two things:
* Should we add more things, like the user bucket to it?
* I would like to give the app the ability to add more values to the context. For example Focus may have additional values that it wants to use in the expressions.

fercarcedo:
I have added the user bucket to the JEXL context, and also added a `put` method to the `ValuesProvider` class which lets the app add more values as needed. I thought this was the simplest approach, let me know what you think :)

pocmo:
Okay, next week I'll finally take the time to think about how we want to use JEXL from the app side. @fercarcedo Are you still interested in contributing this piece? :)

fercarcedo:
Yes, if you find it useful, of course. If not, feel free to take another approach :)

PR #1194, opened 2018-10-28
Closes #756: Add new BrowserToolbar text attributes

### LINKED ISSUE #756: Provide a way to customize Toolbar url bar text
In particular, we need to change the:
- Color
- Text size
- Hint color

We also managed to change the font family by setting the default font family across our app but for other apps, it could be customized more directly.

jonalmeida:
**Notes:**
- The typeface isn't accessible from XML.
- Writing a unit test for null attributes in the init block was more work that value so I omitted it.

### DIFF
--- a/components/browser/toolbar/src/main/res/values/attrs_browser_toolbar.xml
+++ b/components/browser/toolbar/src/main/res/values/attrs_browser_toolbar.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+
+<resources>
+  <declare-styleable name="BrowserToolbar">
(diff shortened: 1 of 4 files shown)


components/browser/toolbar/src/main/res/values/attrs_browser_toolbar.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+  ~ This Source Code Form is subject to the terms of the Mozilla Public
+  ~  License, v. 2.0. If a copy of the MPL was not distributed with this
+  ~  file, You can obtain one at http://mozilla.org/MPL/2.0/.
+  -->
```
pocmo: nit: This uses a different format (`~`) than in our other XML files 🤔 

```
```

jonalmeida:
> Maybe a good first issue? 
I wrote a simple test that fixes the coverage, but doesn't really avoid a regression from happening if you remove the 'getColor` for each of those attributes. I don't think it's a 'good first issue', but it's definitely something that is 'help wanted'.

> The getters seem to be untested too.
I believe this was the coverage/kotlin bug which counts get coverage incorrectly?

PR #1209, opened 2018-10-29
Improvements to lib-dataprotect usability

### LINKED ISSUE #1206: Dataprotect: Improve API usability
Feedback from use in Lockbox:

* Automatically create the underlying crypto key if it does not exist (during `init`)
* Simplify and annotate error cases (e.g., KeystoreException to wrap possible errors)
* Improve documentation on the above

linuxwolf:
This is a small collection of changes based on feedback from using dataprotect in Lockbox.

Resolves #1206


linuxwolf:
CC @sashei @jhugman

components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/Keystore.kt
```diff
@@ -133,10 +134,21 @@ internal open class KeyStoreWrapper {
  *
  * @property label The label the cryptographic key is identified as
  * @constructor Creates a new instance around a key identified by the given label
+ *
+ * Unless `manual` is `true`, the key is created if not already present in the
+ * platform's key storage.
```
pocmo: This sounds like a good `@param` description

PR #1224, opened 2018-10-30
[glean]  Bug 1497808 - Create the PingMaker to collect data and assemble pings.

Dexterp37:
This adds the PingMaker and slightly refactors the storage engines to have implement the same interface, which is useful for handling the ping generation. Additionally, this stubs out the "ping_info" section
that must be send with every ping. Even though a better solution would have been for the fields of that section to dogfood our field types, some of them are not planned for the MVP (e.g. DateTime), so I'm manually handling this.

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/CommonMetricData.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/CommonMetricData.kt
@@ -35,7 +35,11 @@ interface CommonMetricData {
     val name: String
     val sendInPings: List<String>
 
-    public fun shouldRecord(logger: Logger): Boolean {
+    companion object {
+        internal const val DEFAULT_STORAGE_NAME = "default"
+    }
+
+    fun shouldRecord(logger: Logger): Boolean {
         // Silently drop recording for disabled events.
         if (disabled) {
             return false
@@ -50,5 +54,26 @@ interface CommonMetricData {
         }
 
         return true
+    }
+
+    /**
+     * Defines the names of the storages the metric defaults to when
+     * "default" is used as the destination storage.
+     * Note that every metric type will need to implement this.
+     */
+    fun defaultStorageDestinations(): List<String>
+
+    /**
+     * Get the list of storage names the metric will record to. This
+     * automatically expands [DEFAULT_STORAGE_NAME] to the list of default
+     * storages for the metric.
+     */
+    fun getStorageNames(): List<String> {
+        if (DEFAULT_STORAGE_NAME !in sendInPings) {
+            return sendInPings
+        }
+
+        val filteredNames = sendInPings.filter { it != DEFAULT_STORAGE_NAME }
+        return filteredNames + defaultStorageDestinations()
     }
 }
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/ping/PingMaker.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/ping/PingMaker.kt
@@ -1,0 +1,73 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.glean.ping
+
+import android.support.annotation.VisibleForTesting
+import mozilla.components.service.glean.BuildConfig
+import mozilla.components.service.glean.storages.StorageEngineManager
+import org.json.JSONObject
+import java.text.SimpleDateFormat
+import java.util.Date
+import java.util.Locale
+
+internal class PingMaker(
+    private val storageManager: StorageEngineManager
+) {
+    private val pingStartTimes: MutableMap<String, String> = mutableMapOf()
+    private val objectStartTime = getISOTimeString()
+
+    /**
+     * Generate an ISO8601 compliant time string for the current time.
+     *
+     * @return a string containing the date and time.
+     */
+    private fun getISOTimeString(): String {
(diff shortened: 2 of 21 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/ping/PingMaker.kt
```diff
@@ -0,0 +1,75 @@
...
+     * @param storage the name of the storage containing the data for the ping.
+     *        This usually matches with the name of the ping
+     * @return a string holding the data for the ping.
+     */
+    fun collect(storage: String) : String {
+        val jsonPing = JSONObject()
+
+        // Assemble the JSON ping.
+        jsonPing.put("ping_info", getPingInfo(storage))
+        jsonPing.put("metrics", StorageEngineManager.collect(storage))
```
Dexterp37: @pocmo I'm having hard times testing with `object`, but this really is a singleton that is not accessible to users. Do you have any tip on how to properly test this? I would like to fake the response of `StorageEngineManager.collect` to return test data and test `PingMaker` in isolation :\

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
```diff
@@ -79,7 +80,32 @@ internal object EventsStorageEngine {
...
+                val eventData = JSONArray()
+                eventData.put(it.msSinceStart)
+                eventData.put(it.category)
+                eventData.put(it.name)
+                eventData.put(it.objectId)
+                eventData.put(it.value)
+                eventData.put(it.extra)
+                eventsArray.put(eventData)
+            }
+            return eventsArray
```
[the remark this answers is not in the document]
Dexterp37: The expected format here should be the same as the one we currently send from telemetry-android and on Desktop. The serialization format can be found [here](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/collection/events.html#serialization-format)

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
```diff
@@ -79,7 +80,32 @@ internal object EventsStorageEngine {
             return eventStores.remove(storeName)
```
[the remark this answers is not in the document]
Dexterp37: Yes, this makes sense. We can tackle this as part of the "persistence" bug 1499756 (or as a separate, follow-up bug)

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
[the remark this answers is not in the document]
mdboom: Ok.  In that case, https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/210 is currently wrong.  I'll update that.

Dexterp37:
@pocmo @mdboom this is ready for another look!

Dexterp37:
@pocmo the failure here looks unrelated here (but I can't retrigger it) :(

pocmo:
Updating the PR title will retrigger taskcluster :)

components/service/glean/src/main/java/mozilla/components/service/glean/CommonMetricData.kt
```diff
@@ -51,4 +55,25 @@ interface CommonMetricData {
 
         return true
     }
+
+    /**
+     * Defines the names of the storages the metric defaults to when
+     * "default" is used as the destination storage.
+     * Note that every metric type will need to implement this.
+     */
+    fun defaultStorageDestinations(): List<String>
```
pocmo: nit: Those "getters" can be properties (`val`) too and that's a little bit more idiomatic Kotlin:

```Kotlin
interface CommonMetricData {
    val defaultStorageDestinations: List<String>
}

class SomeMetricData : CommonMetricData {
    override val defaultStorageDestinations: List<String>
        get() = ...
        // or:
        get() {
           ...
       }
    
   // or with a backing field:
   override val defaultStorageDestinations: List<String> = listOf(".....")
}
```

components/service/glean/src/main/java/mozilla/components/service/glean/ping/PingMaker.kt
```diff
@@ -0,0 +1,73 @@
...
+        // This needs to be a bit more involved for start-end times. "start_time" is
+        // the time the ping was generated the last time. If not available, we use the
+        // date the object was initialized.
+        val startTime = if (pingName in pingStartTimes) pingStartTimes[pingName] else objectStartTime
+        pingInfo.put("start_time", startTime)
+        val endTime = getISOTimeString()
+        pingInfo.put("end_time", endTime)
+        // Update the start time with the current time.
+        pingStartTimes[pingName] = endTime
+        return pingInfo
```
pocmo: Just FYI: Almost all of `JSONObject` methods can throw `JSONException`. In Java that was a bit annoying because they are all checked exceptions. In Kotlin there are no checked exceptions and you do not need to catch them. I think that's no problem here (because they actually only throw in specific edge cases) but just so that you know.

Dexterp37: Ah, thanks for telling me, I didn't notice 😭 I agree with you that this might not be a problem at this specific location: I'll keep an eye in other places for similar issues.

components/service/glean/src/main/java/mozilla/components/service/glean/CommonMetricData.kt
Dexterp37: Oh, interesting. I had pushed an updated version with this fixed but it's not showing up. I guess I'll have a separate PR for this.

PR #1232, opened 2018-10-30
Issue #1109: concept-awesomebar: Introduce weights, suspend onInputChanged() and add shouldClearSuggestions..

pocmo:
Some more improvements in the concept component before landing `browser-awesomebar`.. :)

* This change introduces the concept of "weights". Weights are used to rank suggestions against each other.
  This allows the awesomebar to mix suggestions from different providers.
* AwesomeBar.SuggestionProvider.onInputChanged() is now a suspending function.
* A provider can define `shouldClearSuggestions` to avoid suggestions getting cleared immediatelly once
  the entered text changes.

### DIFF
--- a/components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
+++ b/components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
@@ -55,16 +55,18 @@ interface AwesomeBar {
      * A [Suggestion] to be displayed by an [AwesomeBar] implementation.
      *
      * @property id A unique ID identifying this [Suggestion]. A stable ID but different data indicates to the
-     *              [AwesomeBar] that this is the same [Suggestion] with new data. This will affect how the [AwesomeBar]
-     *              animates showing the new suggestion.
+     * [AwesomeBar] that this is the same [Suggestion] with new data. This will affect how the [AwesomeBar]
+     * animates showing the new suggestion.
      * @property title A user-readable title for the [Suggestion].
      * @property description A user-readable description for the [Suggestion].
      * @property icon A lambda that can be invoked by the [AwesomeBar] implementation to receive an icon [Bitmap] for
-     *                this [Suggestion]. The [AwesomeBar] will pass in its desired width and height for the Bitmap.
+     * this [Suggestion]. The [AwesomeBar] will pass in its desired width and height for the Bitmap.
      * @property chips A list of [Chip] instances to be displayed.
      * @property flags A set of [Flag] values for this [Suggestion].
      * @property onSuggestionClicked A callback to be executed when the [Suggestion] was clicked by the user.
      * @property onChipClicked A callback to be executed when a [Chip] was clicked by the user.
+     * @property score A score used to rank suggestions of this provider against each other. A suggestion with a higher
+     * score will be shown on top of suggestions with a lower score.
      */
     data class Suggestion(
         val id: String = UUID.randomUUID().toString(),
@@ -74,7 +76,8 @@ interface AwesomeBar {
         val chips: List<Chip> = emptyList(),
         val flags: Set<Flag> = emptySet(),
         val onSuggestionClicked: (() -> Unit)? = null,
-        val onChipClicked: ((Chip) -> Unit)? = null
+        val onChipClicked: ((Chip) -> Unit)? = null,
+        val score: Int = 0
     ) {
         /**
          * Chips are compact actions that are shown as part of a suggestion. For example a [Suggestion] from a search
@@ -93,6 +96,19 @@ interface AwesomeBar {
             BOOKMARK,
             OPEN_TAB
         }
+
+        /**
+         * Returns true if the content of the two suggestions is the same.
+         *
+         * This is used by [AwesomeBar] implementations to decide whether an updated suggestion (same id) needs its
+         * view to be updated in order to display new data.
+         */
+        fun areContentsTheSame(other: Suggestion): Boolean {
+            return title == other.title &&
+                description == other.description &&
+                chips == other.chips &&
+                flags == other.flags
+        }
     }
 
     /**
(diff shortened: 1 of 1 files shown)


components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
```diff
@@ -93,6 +96,27 @@ interface AwesomeBar {
             BOOKMARK,
             OPEN_TAB
         }
+
+        /**
+         * Constants for defining a shared understanding of weights.
+         */
+        object Weight {
```
grigoryk: Scoring - and weights - need to somehow mesh across different providers in a sensible way, and I am not sure that "suggested weights" is the right approach here.

I don't see a way to keep components acting as "suggestion providers" conceptually separate from each other, and have them somehow produce compatible (weighted) scores. Also, these components need to be free to iterate on the way they weight search results internally (e.g. places tweaking its frecency algorithm shouldn't tie into in any way).

Another aspect here is dependency between a storage layer (for example) which provides an implementation of some kind of "search" function, and a `concept-awesomebar`. I don't think it makes sense for something like `concept-storage` to depend on `concept-awesomebar` - the `Suggestion` interface is too specific to the awesomebar itself (chips, actions, etc. don't have any relation to storage, neither can a storage implementation meaningfully fulfill an "onClicked" action). So we'll end up with a separate `SearchResult (id, url, title?, descrition?, icon?)` data class in `concept-storage` which defines a subset of a `Suggestion`.

Thinking out-loud, one way to address scoring across multiple independent providers is to let them produce any scores they please, and equalize the scores at the Awesomebar implementation layer. 

- Convert suggestions from each provider onto a common scale, e.g. [0, 100] (if we can define min/max for scores it'll be easier, but not that important since those could be figured out dynamically)
- Weight suggestions by their providers (like, anything from browsing history gets higher weight than common domain suggestions)
- Finally, sort suggestions

components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
```diff
@@ -108,31 +132,29 @@ interface AwesomeBar {
...
-         * as they are no longer needed. For the provider this makes the implementation easier as it does not need to
-         * deal with asynchronous cancellable tasks.
+         * Coroutine cancellation is cooperative. A coroutine code has to cooperate to be cancellable:
+         * https://github.com/Kotlin/kotlinx.coroutines/blob/master/docs/cancellation-and-timeouts.md
          *
          * @param text The current user input in the toolbar.
          * @return A list of suggestions to be displayed by the [AwesomeBar].
          */
-        fun onInputChanged(text: String): List<Suggestion>
+        suspend fun onInputChanged(text: String): List<Suggestion>
```
grigoryk: Why not do something like `Deferred<List<Suggestion>>`? `await` is a suspending function, which allows for the same usage pattern. It kind of matches the intent a bit better, too? This is still new to me, so it'd be great to see why you chose plain `suspend`.

One reason I can think of is that the implementation wouldn't be forced to use `CompleteableDeferred` or something similar if it's non-blocking (like an in-memory thing..).

components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
```diff
@@ -108,31 +132,29 @@ interface AwesomeBar {
...
          * Fired when the user has cancelled their interaction with the awesome bar.
          */
         fun onInputCancelled() = Unit
+
+        /**
+         * If true an [AwesomeBar] implementation can clear the previous suggestions of this provider as soon as the
+         * user continues to type. If this is false an [AwesomeBar] implementation is allowed to keep the previous
+         * suggestions around until the provider returns a new list of suggestions for the updated text.
+         */
+        val shouldClearSuggestions
```
grigoryk: Not sure why you need this, but I guess you do!

components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
```diff
@@ -74,7 +76,8 @@ interface AwesomeBar {
         val chips: List<Chip> = emptyList(),
         val flags: Set<Flag> = emptySet(),
         val onSuggestionClicked: (() -> Unit)? = null,
-        val onChipClicked: ((Chip) -> Unit)? = null
+        val onChipClicked: ((Chip) -> Unit)? = null,
+        val weight: Int = Weight.DEFAULT
```
grigoryk: I'd rename this to something like `score`. A provider overall might have a weight, which would use to "weight" its suggestions' scores (once equalized) against other providers. See my other comment.

components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
pocmo: I do not expect any storage to implement this. This is something for `feature-awesomebar` (not pushed yet) or other feature modules. Storage should return whatever ranked results it produces. The provider implementation (not in storage) translates that into something that is suitable for the awesomebar.

components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
pocmo: Deferred means the provider will be responsible for threading. That's not what I want. The caller will wrap that in async *in its own scope* so that it can shutdown everything once it is done.

components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
pocmo: Some suggestions (like the search engine suggestions) should not disappear / re-appear whenever the text changes.

components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
pocmo: Sounds good!

components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
pocmo: As discussed on Slack I'll open a follow-up for implementing "provider relative ranking" + "weighting providers".

Issue #315, opened 2018-06-20
Move codecov into separate taskcluster tasks


pocmo:
Since we run one task per module (#304) taskcluster is pretty fast (~ 7 mins) and travis is now the bottleneck (~ 17-20mins). I wonder if we can speed this up by running two tasks for ~sputnik~ and codecov on taskcluster.

For codecov we need to run **all** tests with code coverage and this will probably take long no matter what we do. It would be interesting to know how long it runs on taskcluster.

In theory we could generate coverage reports in the module tasks and then have a codecov task that just collects them and uploads them. That would be pretty fast.

jonalmeida:
> travis is now the bottleneck (~ 17-20mins)

BUILD SUCCESSFUL in 37m 56s

😅

pocmo:
Holy coverage report merging! I thought codecov would only merge multiple code coverage reports uploaded at once. But it's also possible to do multiple uploads and codecov will merge them together:
https://docs.codecov.io/docs/merging-reports

That's pretty awesome! With that we do not need to collect all the reports together to upload them. Instead every task can upload itself and once that's done codecov will generate the overall report (it looks at the pull request status to know when all tasks are done).

With that we can kill travis! I'll move that in. :)

pocmo:
The state is: Uploading works in parallel from taskcluster. However codecov posts the report too early because it doesn't wait for taskcluster. Normally it'd wait for CI tasks but it doesn't detect taskcluster as CI because of:

> Codecov detects CI services through the Commit Status API provided by GitHub/Bitbucket/GitLab. We review the target url, and context, in order to determine if the status is a "continuous integration service" that uploads coverage reports.
> There is a match when ci, semaphoreci, pull request validator (cloudbees), continuous-integration, or buildkite is found in the context.

Taskcluster's context is `Taskcluster (pull_request)` and doesn't contain any of those keywords.

pocmo:
Reached out to the taskcluster folks to see if we can do anything with our context.

And reached out to codecov to see if they are willing to add another keyword to check for.

> Thanks for reaching out to Codecov Support. We are sorry to see that you are running into some issues. In an effort to expedite a resolution, I have forwarded your support ticket to our Senior Support Engineers who will circle back with their resolution ASAP

PR #1235, opened 2018-10-30
WIP: Issue #315: Generate code coverage and upload to codecov from taskcluster..

pocmo:
Not ready yet. Will likely fail.


pocmo:
The upload and merging is working. What's missing is that codecov looks at the taskcluster check status before posting the comment. Right now it posts the comment as soon as one component has been tested and then updates the comment multiple times.

PR #1238, opened 2018-10-31
Wire up the history tracking delegate for GeckoView

linacambridge:
I tested this with my local GeckoView build, the sample browser, and @grigoryk's in-memory history storage, and it works! 🎉 There's also a commit that points the Nightly x86 target to m-c instead, for folks to try out if they'd like. 😄


pocmo:
GV bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1494713

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -347,6 +359,66 @@ class GeckoEngineSession(
...
+        override fun getVisited(
+            session: GeckoSession,
+            urls: Array<out String>
+        ): GeckoResult<BooleanArray>? {
+            val result = GeckoResult<BooleanArray>()
+            scope.launch {
+                var exception: Throwable? = null
+                val visits: List<Boolean>? = try {
+                    settings.historyTrackingDelegate?.getVisited(urls.toList(), privateMode)?.await()
+                } catch (t: Throwable) {
```
grigoryk: Generally you watch to catch as narrow set of exceptions as possible. `Throwable` covers unchecked exceptions (subclasses of `RuntimeException` and `Error`, such as `NullPointerException`, `ThreadDeath`, etc), which is not what we want here. If there's nothing better, use `Exception` instead.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -347,6 +359,66 @@ class GeckoEngineSession(
...
+                    exception = t
+                    null
```
grigoryk: Remove this?

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -347,6 +359,66 @@ class GeckoEngineSession(
...
+                    val markAsVisited = !privateMode
+                    result.complete(markAsVisited)
+                } else {
+                    result.completeExceptionally(exception)
+                }
+            }
+            return result
+        }
+
+        @Suppress("TooGenericExceptionCaught")
```
grigoryk: Hopefully this will go away if you don't catch `throwable`.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -347,6 +359,66 @@ class GeckoEngineSession(
...
+
+                return GeckoResult.fromValue(false)
+            }
+
+            val isReload = lastVisitedURL?.let { it == url } ?: false
+            val result = GeckoResult<Boolean>()
+            scope.launch {
+                var exception: Throwable? = null
+                try {
+                    settings.historyTrackingDelegate?.onVisited(url, isReload, privateMode)
```
grigoryk: We shouldn't even call into `historyTrackingDelegate` if we're in a `privateMode`.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -57,6 +61,13 @@ class GeckoEngineSession(
 
     private var initialLoad = true
 
+    private val scope by lazy { CoroutineScope(Dispatchers.IO) }
```
grigoryk: nit: Can you change other users of the history delegate to use this scope as well?

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -347,6 +359,66 @@ class GeckoEngineSession(
...
+        ): GeckoResult<Boolean>? {
+            if ((flags and GeckoSession.HistoryDelegate.VISIT_TOP_LEVEL) == 0 ||
+                (flags and GeckoSession.HistoryDelegate.VISIT_UNRECOVERABLE_ERROR) != 0) {
+
+                return GeckoResult.fromValue(false)
+            }
+
+            val isReload = lastVisitedURL?.let { it == url } ?: false
+            val result = GeckoResult<Boolean>()
+            scope.launch {
```
grigoryk: This should not do any work if: 
- we're in privateMode
- there's no configured `historyTrackingDelegate`

In those two cases, you'd want to resolve GeckoResult right away without doing anything with scopes.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -347,6 +359,66 @@ class GeckoEngineSession(
...
+                    result.complete(markAsVisited)
+                } else {
+                    result.completeExceptionally(exception)
+                }
+            }
+            return result
+        }
+
+        @Suppress("TooGenericExceptionCaught")
+        override fun getVisited(
```
grigoryk: Same as in `onVisited`, resolve this right away if we don't have a configured delegate or if we're in a private mode.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
linacambridge: Hmm, I'm not sure what all we'd like to catch here; IIUC, `getVisited` can throw anything. On the one hand, I don't think we want to crash if `getVisited` throws; on the other, I don't see other handlers doing the same, and, given that Gecko's [`OnError` handlers do nothing](https://searchfox.org/mozilla-central/rev/0859e6b10fb901875c80de8f8fc33cbb77b2505e/mobile/android/components/geckoview/GeckoViewHistory.cpp#363-367,565-569), maybe it's best to remove this entirely.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
linacambridge: I'm assigning to `val visits`, so I think I still need the `null`, but it's moot if I remove the catches.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
linacambridge: In that case, should we also remove the `privateMode` argument from `historyTrackingDelegate.{onVisited, getVisited}`?

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
linacambridge: Do you mean I should make this public and have `PlacesHistoryStorage` use this delegate? How would that work if `PlacesHistoryStorage` doesn't access the session; should I pass the scope as an argument to `GeckoEngineSession` and `PlacesHistoryStorage` instead?

linacambridge:
@grigoryk Whew! I think this is ready for re-review. Changes since last time:

* Removed the `privateMode` argument from `HistoryTrackingDelegate`. Instead, we'll skip the call to the delegate entirely.
* Bail early instead of launching a coroutine if a delegate isn't set.
* Tests actually work! 🎉
* Made both overloads of `getVisited` `suspend` functions.

linacambridge:
Looks like the `browser-awesomebar` failure is #1300; I didn't see it fail on previous runs.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
grigoryk: On one hand, `privateMode` is useful on a conceptual level in the delegate as it allows to easily segment history into two buckets, and let the storage layer do something sensible with those buckets. On another hand, we don't actually follow through on this, and it's not clear that we'll want to. I'll file a follow-up to consider cleaning this up.

grigoryk: Ah! I see that you killed that argument. That's fine!

grigoryk:
This landed!

Issue #1126, opened 2018-10-22
Rust component- backed implementation of the HistoryTrackingDelegate


grigoryk:
Placeholder to track upcoming work.

grigoryk:
WIP in https://github.com/mozilla-mobile/android-components/pull/1164

PR #1240, opened 2018-10-31
WIP towards a places component

mhammond:
This effectively replaces #1107 as this is based on the v0.28.0 tag so the reference browser patch still works. As @grigoryk mentioned, I did create a new PlacesAwesomeBarProvider.kt where the interactions with the low-level components are abstracted.

I haven't created a sample app because I can't picture what it would do - ie, it's not clear how it would populate the database, or really, what an "autocomplete sample that's not a browser" looks like. Once Grisha's history component works it might make sense to work on that - but I don't think it makes sense for me to work on that alone.

I'm not sure how to proceed here:
* Because the reference browser is still on 0.28.0, this PR is too - but that means it can't be merged.
* If I stick to the master branch, we don't get to integrate it with the reference browser until that moves to the most recent version of these components.

@pocmo @csadilek can you please give me some guidance on what concrete things I need to do to move this forward? I opened https://github.com/mozilla/application-services/issues/344 to arrange for the low-level component to be published which will need to happen first, but beyond that I'm not sure how to proceed.


grigoryk:
To be clear, that structure (concept-storage, feature-storage) is already in place. #1164 just adds the places storage component.

grigoryk:
As always, @pocmo is a step ahead :) See https://github.com/mozilla-mobile/android-components/issues/1242 and https://github.com/mozilla-mobile/android-components/issues/1241.

grigoryk:
From a conversation with @mhammond, I will fold this alongside my work in #1248 and #1126.

PR #1244, opened 2018-10-31
Add Uuid metric type (#1501318)

mdboom:
This adds a Uuid metric type to glean.


components/service/glean/src/test/java/mozilla/components/service/glean/UuidMetricTypeTest.kt
```diff
@@ -0,0 +1,95 @@
...
+        // Record two uuids of the same type, with a little delay.
+        val uuid = uuidMetric.generateAndSet()
+
+        // Check that data was properly recorded.
+        val snapshot = UuidsStorageEngine.getSnapshot(storeName = "store1", clearStore = false)
+        assertEquals(1, snapshot!!.size)
+        assertEquals(true, snapshot.containsKey("telemetry.uuid_metric"))
+        assertEquals(uuid, snapshot.get("telemetry.uuid_metric"))
+
+        val uuid2 = UUID.randomUUID()
```
Dexterp37: Instead of using something random, let's use `UUID.fromString("a valid UUID here")`. It's not that different, but more closely resembles what the final usage scenario would probably be.

samples/glean/metrics.yaml
```diff
@@ -2,7 +2,7 @@
 # automatically converted to Kotlin code at build time using the `glean_parser`
 # PyPI package.
 
-$schema: http://mozilla.org/schemas/glean/metrics
+$schema: moz://mozilla.org/schemas/glean/metrics/1-0-0
```
Dexterp37: That's changing due to the version bump for the parsers, right?

mdboom: Yes.

Issue #1181, opened 2018-10-26
Update deprecated Coroutines code


pocmo:
We recently updated the Coroutines library and now we have a bunch of deprecated code. I assume that code will get removed once Kotlin 1.3 gets released and Coroutines are no longer experimental. So let's update that.

The biggest change seems to be that we need to scope Coroutines now.

Building the tree will show a bunch of warnings that we'll need to fix here, e.g.:

> w: components/browser/domains/src/main/java/mozilla/components/browser/domains/DomainAutoCompleteProvider.kt: (120, 37): 'CommonPool' is deprecated. Use Dispatchers.Default
> ...

pocmo:
Coroutines 1.0 removes this deprecated code. So this is a blocker for updating Kotlin.

grigoryk:
Remaining bits closed in #1281.

PR #1245, opened 2018-10-31
Issue #1181: Update deprecated coroutines for FxA

csadilek:
This brings in coroutine scopes and makes sure we call cancel on the corresponding "Job" when the activity is destroyed and/or the FirefoxAccount/Config are closed.

With @thomcc's changes plus these refactorings, things look much cleaner than before, imo. `onCreate` is still somewhat convoluted, but I think it's best to clean this up when we attempt to build the higher level API. This is mainly caused by `onCreate` handling 3 different use cases in one async block.

Thanks @pocmo @thomcc for talking through the scope->context->dispatcher->job coroutine fun :)

### DIFF
--- a/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/Config.kt
+++ b/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/Config.kt
@@ -5,6 +5,8 @@
 package mozilla.components.service.fxa
 
 import kotlinx.coroutines.experimental.Deferred
+import kotlinx.coroutines.experimental.Dispatchers
+import kotlinx.coroutines.experimental.GlobalScope
 import kotlinx.coroutines.experimental.async
 import org.mozilla.fxaclient.internal.Config as InternalConfig
 
@@ -23,7 +25,7 @@ class Config internal constructor(internal val inner: InternalConfig) : AutoClos
          * Set up endpoints used in the production Firefox Accounts instance.
          */
         fun release(): Deferred<Config> {
-            return async {
+            return GlobalScope.async(Dispatchers.IO) {
                 Config(InternalConfig.release())
             }
         }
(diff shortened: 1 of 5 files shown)


components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/Config.kt
```diff
@@ -15,15 +19,19 @@ import org.mozilla.fxaclient.internal.Config as InternalConfig
 class Config internal constructor(internal val inner: InternalConfig) : AutoCloseable {
 
     override fun close() {
+        job.cancel()
         inner.close()
```
pocmo: With the scope as part of the companion object wouldn't this cancel **all** Config related jobs? Not sure, maybe we want this?

csadilek: Yes, it will. This seems OK as we're only having one short-lived Config instance per App. My concern was that `cancel` might prevent future submissions, but it worked when testing.

csadilek: Since an instance-based scope wouldn't make sense here, an alternative to this solution would be to use `GlobalScope.async`, but then technically coroutines could be around longer. This is all just for a one-time method call though, so it seems fine either way?

csadilek: Actually, I am wrong about `job.cancel`. Once cancelled, it's over. I will push an update for this.

csadilek: OK, updated, since this is a single static method call to create the singleton config, I think `GlobalScope` is fine in this case. In the future, either the higher-level API needs to provide a single managed scope, or we push this to the consumer and make all calls synchronous. Either way, we don't want the app having to coordinate multiple scopes.

csadilek:
> My only concern is about the fxa sample: given that it is sample code, some comments explaining what's happening might be warranted

Yeah, We have the README which describes all this code. I will update it to reflect the latest changes, and add some details.

csadilek:
Updated docs.

csadilek:
@pocmo r? / ok landing this? :)

PR #1248, opened 2018-11-01
Closes #1242: Add suggestion API to concept-storage + basic implementation

### LINKED ISSUE #1242: Add API to concept-storage for providing awesomebar suggestions
Depends on #1109. Related to #1241.

For getting awesomebar suggestions we will need some kind of API(s) in `concept-storage` so that we do not have a hard dependency on the storage implementation.

The actual implementation of the `SuggestionProvider` (using the APIs) may live in `feature-autocomplete` (not in tree yet) where I have some other implementations for Sessions (open tabs) and Search Engine Suggestions.

grigoryk:
Added implementation for the in-memory storage which is already in-tree.

### DIFF
--- a/components/concept/storage/src/main/java/mozilla/components/concept/storage/HistoryStorage.kt
+++ b/components/concept/storage/src/main/java/mozilla/components/concept/storage/HistoryStorage.kt
@@ -35,6 +35,13 @@ interface HistoryStorage {
      * @param callback will be invoked with a list of all visited page URIs.
      */
     fun getVisited(callback: (List<String>) -> Unit)
+
+    /**
+     * Retrieves suggestions matching the [query].
+     * @param query A query by which to query the underlying store.
+     * @return A List of [SearchResult] matching the query, in no particular order.
+     */
+    fun getSuggestions(query: String): List<SearchResult>
 }
 
 data class PageObservation(val title: String?)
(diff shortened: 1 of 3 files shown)


grigoryk:
This assumes "independent scoring" as suggested in my comment regarding weights on #1232.

pocmo:
> This assumes "independent scoring" as suggested in my comment regarding weights on #1232.

I changed the code to just return a `score` and removed the `Weight` class. Now every provider can return its local score and we will deal with normalizing and weighting in a follow-up.

components/concept/storage/src/main/java/mozilla/components/concept/storage/HistoryStorage.kt
```diff
@@ -50,3 +59,10 @@ enum class VisitType(val type: Int) {
     TYPED(2),
     RELOAD(9)
 }
+
+data class SearchResult(
+    val id: String = UUID.randomUUID().toString(),
+    val url: String,
+    val score: Int,
+    val title: String? = null
```
pocmo: Maybe we could add some KDoc for `SearchResult`.

components/concept/storage/src/main/java/mozilla/components/concept/storage/HistoryStorage.kt
```diff
@@ -50,3 +59,10 @@ enum class VisitType(val type: Int) {
     TYPED(2),
     RELOAD(9)
 }
+
+data class SearchResult(
+    val id: String = UUID.randomUUID().toString(),
```
pocmo: Do we need the UUID default? Looks like this will always be set?

PR #1251, opened 2018-11-01
Closes #755: Replace Toolbar Actions from resouces to Drawables.

### LINKED ISSUE #755: Provide a way to tint toolbar icons
Right now toolbar buttons take resource IDs so we're unable to tint the images and are stuck using the asset's raw color. Perhaps the toolbar should take drawables, which are more flexible, instead?

jonalmeida:
To be more versatile, we've added a secondary constructor to `BrowserToolbar.Button` and `BrowserToolbar.ToggleButton`.

```kotlin
// Before
BrowserToolbar.Button(R.drawable.image, "image description") {
  // perform an action on click.
}

// Now
val imageDrawable: Drawable = Drawable()
BrowserToolbar.Button(imageDrawable, "image description") {
  // perform an action on click.
}
```

```kotlin
// Before
BrowserToolbar.ToggleButton(
  R.drawable.image, 
  R.drawable.image_selected, 
  "image description", 
  "image selected description") {
  // perform an action on click.
}

// Now
val imageDrawable: Drawable = Drawable()
val imageSelectedDrawable: Drawable = Drawable()
BrowserToolbar.ToggleButton(
  imageDrawable, 
  imageSelectedDrawable, 
  "image description", 
  "image selected description") {
  // perform an action on click.
}
```

(These commits will be squashed before merging)

### DIFF
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
@@ -6,6 +6,7 @@ package mozilla.components.browser.toolbar
 
 import android.content.Context
 import android.graphics.Typeface
+import android.graphics.drawable.Drawable
 import android.support.annotation.DrawableRes
 import android.support.annotation.VisibleForTesting
 import android.util.AttributeSet
@@ -366,7 +367,7 @@ class BrowserToolbar @JvmOverloads constructor(
     /**
      * An action button to be added to the toolbar.
      *
-     * @param imageResource The drawable to be shown.
+     * @param imageDrawable The drawable to be shown.
      * @param contentDescription The content description to use.
      * @param visible Lambda that returns true or false to indicate whether this button should be shown.
      * @param background A custom (stateful) background drawable resource to be used.
@@ -374,17 +375,18 @@ class BrowserToolbar @JvmOverloads constructor(
      * @param listener Callback that will be invoked whenever the button is pressed
      */
     open class Button(
-        imageResource: Int,
+        imageDrawable: Drawable,
         contentDescription: String,
         visible: () -> Boolean = { true },
-        @DrawableRes background: Int? = null,
-        private val padding: Padding = defaultActionPadding,
+        @DrawableRes background: Int = 0,
+        val padding: Padding = DEFAULT_PADDING,
         listener: () -> Unit
-    ) : Toolbar.ActionButton(imageResource, contentDescription, visible, background, listener = listener) {
+    ) : Toolbar.ActionButton(imageDrawable, contentDescription, visible, background, padding, listener) {
+
         override fun createView(parent: ViewGroup): View {
-            val view = super.createView(parent)
-            view.setPadding(padding)
-            return view
+            return super.createView(parent).apply {
+                setPadding(padding)
+            }
         }
     }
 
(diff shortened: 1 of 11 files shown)


jonalmeida:
> ..will increase coverage by 0.55%.

💪 💪 💪

jonalmeida:
Closed the PR in error.

pocmo:
> The reason I kept it around was to avoid breaking compatibility with existing users.

I agree with that in most cases. Here the "migration" seems to be so cheap that I'm not sure whether it's worth having to maintain two versions on our side is worth it. :)

csadilek:
> That's what I said to @csadilek ! The reason I kept it around was to avoid breaking compatibility with existing users.

Umm, I remember I +1

jonalmeida:
> > That's what I said to @csadilek ! The reason I kept it around was to avoid breaking compatibility with existing users.
> 
> Umm, I remember I also +1'ed breaking API :)

Debatable! 😛

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
```diff
@@ -366,34 +367,37 @@ class BrowserToolbar @JvmOverloads constructor(
...
+
         override fun createView(parent: ViewGroup): View {
-            val view = super.createView(parent)
-            view.setPadding(padding)
-            return view
+            return super.createView(parent).apply {
+                padding?.let {
+                    setPadding(it)
+                }
+            }
```
pocmo: This is very confusing. `padding?` is referencing the property of the parent class. Since you are calling `super.createView()` here.. shouldn't the padding be applied already?

jonalmeida: Hmm.. I didn't quite look at the reasoning behind why this original code was here when I refactored it. I'll look into it. 👍

jonalmeida: This was my error in terms of visibility scope. The original change has the super padding as private, and the browser padding as a val so that it couldn't be optional for BrowserToolbar I imagine. I've fixed this now to reflect the original changes.

pocmo: But we are still applying the padding twice, aren't we? 🤔

jonalmeida: From what I understand, it's to have a default padding only in browser which does look odd. I can file a follow up to fix this.

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
```diff
@@ -403,25 +405,27 @@ class BrowserToolbar @JvmOverloads constructor(
...
         selected,
         background,
-        listener = listener
+        padding,
+        listener
     ) {
+
         override fun createView(parent: ViewGroup): View {
             return super.createView(parent).apply {
                 setPadding(padding)
```
jonalmeida: @pocmo it's the same thing here.

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
jonalmeida: Created https://github.com/mozilla-mobile/android-components/issues/1270

jonalmeida:
@pocmo I've updated the changelog to the other PR that's dependant on this one instead so that merging is easer.

PR #1253, opened 2018-11-02
Closes #1182: Crash reporter prompt: Persist checkbox state

### LINKED ISSUE #1182: Crash reporter prompt: Persist checkbox state
I assume we want to persist the state of the checkbox ("[ ] Send crash report to Mozilla"). Currently it will always be [x] when showing the prompt.

Shubhamurkade:
Modified the file CrashReporterActivity to add the code for the persistence of checkbox state.


components/lib/crash/src/main/java/mozilla/components/lib/crash/prompt/CrashReporterActivity.kt
```diff
@@ -40,6 +44,10 @@ class CrashReporterActivity : AppCompatActivity() {
 
         titleView.text = getString(R.string.mozac_lib_crash_dialog_title, appName)
         sendCheckbox.text = getString(R.string.mozac_lib_crash_dialog_checkbox, organizationName)
+
+        sharedPreferences = PreferenceManager.getDefaultSharedPreferences(this)
```
pocmo: We should avoid using the default shared preferences here: We do not want to interfere with the app. Let's use a different file here. Something like `mozac_lib_crash_settings.xml`.

components/lib/crash/src/main/java/mozilla/components/lib/crash/prompt/CrashReporterActivity.kt
```diff
@@ -85,4 +93,13 @@ class CrashReporterActivity : AppCompatActivity() {
...
         }
     }
+
+    override fun onStop() {
+        super.onStop()
+
+        /* Save the state of send checkbox into shared preferences. We can't use onSaveInstanceState here
+            because the state of the checkbox should also be preserved upon app restart.
+        */
+        sharedPreferences.edit().putBoolean(sendCheckboxKey, sendCheckbox.isChecked).apply()
```
pocmo: Let's do that from inside `sendCrashReportIfNeeded()`.

components/lib/crash/src/main/java/mozilla/components/lib/crash/prompt/CrashReporterActivity.kt
```diff
@@ -85,4 +93,13 @@ class CrashReporterActivity : AppCompatActivity() {
             }
         }
     }
+
+    override fun onStop() {
+        super.onStop()
+
+        /* Save the state of send checkbox into shared preferences. We can't use onSaveInstanceState here
+            because the state of the checkbox should also be preserved upon app restart.
+        */
```
pocmo: Mentioning `onSaveInstanceState` may be confusing here. Just some short comment that we are remembering this for the next time the app crashes should be enough :)

components/lib/crash/src/main/java/mozilla/components/lib/crash/prompt/CrashReporterActivity.kt
Shubhamurkade: Could do please explain me what do I do with this file?

components/lib/crash/src/main/java/mozilla/components/lib/crash/prompt/CrashReporterActivity.kt
```diff
@@ -40,6 +44,28 @@ class CrashReporterActivity : AppCompatActivity() {
...
+            try {
+                newFile.createNewFile()
+                /* After creating a new file, store a default value of 1 so that the
+                checkbox remains checked initially when the file is not present.
+                 */
+                newFile.writeText(true.toString())
+            } catch (e: IOException) {
+                e.printStackTrace()
+            }
+        }
```
pocmo: Oh, sorry, I think my comment was misleading. With a separate file I still meant a shared preference:
* [SharedPreferences Context.getSharedPreferences (String name, 
                int mode)](https://developer.android.com/reference/android/content/Context#getSharedPreferences(java.lang.String,%20int))

pocmo:
Closing for #1267

PR #1255, opened 2018-11-02
Change applicationProperty/userProperty to lifetime [glean]

mdboom:
Related to https://github.com/mozilla/glean_parser/pull/10.

[Bugzilla 1503569](https://bugzilla.mozilla.org/show_bug.cgi?id=1503569)

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/CommonMetricData.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/CommonMetricData.kt
@@ -6,17 +6,22 @@ package mozilla.components.service.glean
 
 import mozilla.components.support.base.log.logger.Logger
 
+enum class Lifetime {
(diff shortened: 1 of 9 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/CommonMetricData.kt
```diff
@@ -6,17 +6,22 @@ package mozilla.components.service.glean
 
 import mozilla.components.support.base.log.logger.Logger
 
+enum class Lifetime {
+    Ping,        // Metric is reset with each sent ping
+    Application, // Metric is reset on application restart
+    User         // Metric is reset with each user profile
```
pocmo: nit: As KDoc those comments would be visible from wherever `Lifetime` is used. :)

PR #1257, opened 2018-11-02
Closes #1256: Update ic_device_desktop

### LINKED ISSUE #1256: Update ic_device_desktop to match photon icon
Photon icon: https://github.com/FirefoxUX/photon-icons/blob/master/icons/android/device-desktop-24.xml

dnarcese:


components/ui/icons/src/main/res/drawable/mozac_ic_device_desktop.xml
```diff
@@ -3,11 +3,12 @@
...
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
 <vector xmlns:android="http://schemas.android.com/apk/res/android"
-    android:width="16dp"
-    android:height="16dp"
-    android:viewportHeight="16"
-    android:viewportWidth="16">
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
```
pocmo: This icon has a different size. Is this intentional?

components/ui/icons/src/main/res/drawable/mozac_ic_device_desktop.xml
```diff
@@ -3,11 +3,12 @@
...
   <path
       android:fillColor="#FFFFFF"
-      android:pathData="M15.5,12L15,12L15,3a1,1 0,0 0,-1 -1L2,2a1,1 0,0 0,-1 1v9L0.5,12a0.5,0.5 0,0 0,-0.5 0.5v1a0.5,0.5 0,0 0,0.5 0.5h15a0.5,0.5 0,0 0,0.5 -0.5v-1a0.5,0.5 0,0 0,-0.5 -0.5zM10,13L6,13v-1h4zM13,11L3,11L3,4h10z" />
+      android:fillType="evenOdd"
```
pocmo: Wasn't there an issue with "evenOdd" not being supported on all versions of Android?

components/ui/icons/src/main/res/drawable/mozac_ic_device_desktop.xml
dnarcese: Yes, I needed the size to match the other icons FFTV is using (which are all 24x24).

pocmo: Oh, I just noticed that this was 16dp **before**. Yep, makes sense!

components/ui/icons/src/main/res/drawable/mozac_ic_device_desktop.xml
dnarcese: I regenerated the SVG so it wouldn't use `evenOdd` 👍

pocmo: Awesome. Thank you! I was trying to find any information on whether the support library can handle "evenOdd". The system will only be able to load it with API 24+ but maybe the support library can do it. Anyhow, now we can just merge it. :)

PR #1260, opened 2018-11-02
Issue #1109: browser-awesomebar: A customizable awesome bar implementation for browsers.

pocmo:
Here's the `browser-awesomebar` implementation.

The adapter does a lot of work currently (e.g. diffing). I'm thinking about moving some of that to the feature in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,8 @@ repositories {
 
 High-level components for building browser(-like) apps.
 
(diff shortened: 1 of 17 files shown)


pocmo:
> Should be we add it to the sample browser?

I have the code for that locally. I'm not sure yet whether I should open a PR for it or whether to create a separate sample app. I would like to avoid sample-browser becoming a complex browser app mirroring reference-browser. I see `browser-sample` more as a sample app to test/show engine related things (maybe we should rename the sample!).

Amejia481:
I like more the idea of creating a separate sample app as we did with Toolbar. As you mentioned it keep things straightforward. 👍

PR #1261, opened 2018-11-02
Issue #1109: feature-awesomebar: Connecting a concept-awesomebar with…

pocmo:
Depends on `browser-awesomebar` (See PR #1260).

~~The build will fail until PR #1260 is in.~~

~~In addition to that I still need to write some tests. However I wanted to put this up for review already.~~

### DIFF
--- a/components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/SearchSuggestionProvider.kt
+++ b/components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/SearchSuggestionProvider.kt
@@ -1,0 +1,124 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.feature.awesomebar.provider
+
+import mozilla.components.browser.search.SearchEngine
+import mozilla.components.browser.search.suggestions.SearchSuggestionClient
+import mozilla.components.concept.awesomebar.AwesomeBar
+import mozilla.components.feature.search.SearchUseCases
+import mozilla.components.support.base.log.logger.Logger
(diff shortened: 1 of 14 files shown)


pocmo:
> The build will fail until PR #1260 is in.

Heh, fooled by `concept` components. Of course this does not depend on the actual implementation. :)

pocmo:
Added some tests.

components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/SearchSuggestionProvider.kt
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * A [AwesomeBar.SuggestionProvider] implementation that provides a suggestion containing search engine suggestions (as
+ * chips) from the passed in [SearchEngine].
+ */
+class SearchSuggestionProvider(
+    private val searchEngine: SearchEngine,
+    private val searchUseCase: SearchUseCases.DefaultSearchUrlUseCase
+) : AwesomeBar.SuggestionProvider {
+    private val client = if (searchEngine.canProvideSearchSuggestions) {
+        SearchSuggestionClient(searchEngine, { url -> fetch(url) })
```
Amejia481: nit: As the lambda is the last parameter  you could pull it the  outside of the parenthesis.
```kotlin
SearchSuggestionClient(searchEngine){ url -> 
    fetch(url) 
} 
```

pocmo: Oh yeah, that makes sense. I'll update the patch.

pocmo:
Updated!

Amejia481:
```
There's a small format issue
> Task :ktlint
/build/android-components/components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/SearchSuggestionProvider.kt:25:44: Unexpected spacing before ")"

```

PR #1265, opened 2018-11-04
Closes #260: In "Icons" readme, provide sample images of included icons

### LINKED ISSUE #260: In "Icons" readme, provide sample images of included icons

santimar:
I wrote a simple script that fetch all the Android Vector Drawable files inside the drawable folder, makes a simple conversion to transform them in regular svg (that browsers can render) and then displays them inside a table.
Files are retrieved via GitHub API, so everything should stay in sync when adding future icons or after making changes.
The API is limited with max 60 requests per hour per IP, but the script only makes one.
Android SVG to standard SVG conversion is made via regex, by striping off/renaming elements.

Everything is working, but probably needs some improvements.

I made a first version, then a second one with Promise.
I don't know which one is better.
The other problem is that in ```docs/ui/icons.md``` i wrote some css style and HTML tags directly inside the page.
Let me know what I should change

Preview: https://santimar.github.io/android-components/components/ui/icons

Closes: #260

### DIFF
--- a/docs/assets/js/icon-js.js
+++ b/docs/assets/js/icon-js.js
@@ -1,0 +1,100 @@
+let icons = {};
+let needed = 0;
+let got = 0;
+
+/**
+ * @param html {String} HTML representing a single element
+ * @return {Element}
+ */
+function htmlToElement(html) {
+    let template = document.createElement('template');
+    html = html.trim(); // Never return a text node of whitespace as the result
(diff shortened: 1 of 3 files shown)


pocmo:
Wow. That's pretty awesome. Thank you! :)

PR #1267, opened 2018-11-05
Closes #1182: Crash reporter prompt: Persist checkbox state

### LINKED ISSUE #1182: Crash reporter prompt: Persist checkbox state

### LINKED ISSUE #1215: Correct README.md file for ui-colors component dependency
In  [ui-colors component README.md file](https://github.com/mozilla-mobile/android-components/blob/master/components/ui/colors/README.md) the gradle dependency is incorrect.

WRONG ❌ 
`implementation "org.mozilla.photon:ui-colors:{latest-version}"`

Must be 🔵 
`implementation "org.mozilla.components:ui-colors:{latest-version}"`

Shubhamurkade:
Modified the file CrashResporterActivity.kt. The value of the send checkbox will be stored in a shared preference.

### DIFF
--- a/components/lib/crash/src/main/java/mozilla/components/lib/crash/prompt/CrashReporterActivity.kt
+++ b/components/lib/crash/src/main/java/mozilla/components/lib/crash/prompt/CrashReporterActivity.kt
@@ -4,6 +4,7 @@
 
 package mozilla.components.lib.crash.prompt
 
+import android.content.Context
 import android.content.Intent
 import android.os.Bundle
 import android.support.v7.app.AppCompatActivity
@@ -17,12 +18,17 @@ import mozilla.components.lib.crash.Crash
 import mozilla.components.lib.crash.CrashReporter
 import mozilla.components.lib.crash.R
 
+private const val PREFERENCE_KEY_SEND_REPORT = "sendCrashReport"
+
 /**
  * Activity showing the crash reporter prompt asking the user for confirmation before submitting a crash report.
  */
 class CrashReporterActivity : AppCompatActivity() {
     private val crashReporter: CrashReporter by lazy { CrashReporter.requireInstance }
     private val crash: Crash by lazy { Crash.fromIntent(intent) }
+    private val sharedPreferences by lazy {
+        getSharedPreferences("mozac_lib_crash_settings", Context.MODE_PRIVATE)
+    }
 
     override fun onCreate(savedInstanceState: Bundle?) {
         setTheme(crashReporter.promptConfiguration.theme)
@@ -40,6 +46,9 @@ class CrashReporterActivity : AppCompatActivity() {
 
         titleView.text = getString(R.string.mozac_lib_crash_dialog_title, appName)
         sendCheckbox.text = getString(R.string.mozac_lib_crash_dialog_checkbox, organizationName)
+
+        sendCheckbox.isChecked = sharedPreferences.getBoolean(PREFERENCE_KEY_SEND_REPORT, true)
+
         restartButton.apply {
             text = getString(R.string.mozac_lib_crash_dialog_button_restart, appName)
             setOnClickListener { restart() }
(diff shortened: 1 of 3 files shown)


components/lib/crash/src/main/java/mozilla/components/lib/crash/prompt/CrashReporterActivity.kt
```diff
@@ -23,6 +25,9 @@ import mozilla.components.lib.crash.R
 class CrashReporterActivity : AppCompatActivity() {
     private val crashReporter: CrashReporter by lazy { CrashReporter.requireInstance }
     private val crash: Crash by lazy { Crash.fromIntent(intent) }
+    private val sendCheckboxKey = "send_checkbox_value"
```
pocmo: this can be a constant above the class definition:

```Kotlin
private const val PREFERENCE_KEY_SEND_REPORT = "sendCrashReport"
```
```

File: components/lib/crash/src/main/java/mozilla/components/lib/crash/prompt/CrashReporterActivity.kt [hunk 2]
```diff
@@ -23,6 +25,9 @@ import mozilla.components.lib.crash.R
...
+    private lateinit var sharedPreferences: SharedPreferences
+    private val sendCheckboxFile = "mozac_lib_crash_settings.xml"
```
pocmo: nit: This can be a lazy property:

```Kotlin
private val sharedPreferences by lazy {
    getSharedPreferences("mozac_lib_crash_settings", Context.MODE_PRIVATE)
}
```

components/lib/crash/src/main/java/mozilla/components/lib/crash/prompt/CrashReporterActivity.kt
```diff
@@ -72,6 +81,11 @@ class CrashReporterActivity : AppCompatActivity() {
     }
 
     private fun sendCrashReportIfNeeded(then: () -> Unit) {
+
+        /* Save the state of send checkbox into shared preferences.
+        */
```
pocmo: nit: I think we can get rid of this comment now. The following line is self explanatory :)

pocmo:
After fixing the last issues, can you please rebase and squash this PR so that there's only a single commit? :)

components/lib/crash/src/main/java/mozilla/components/lib/crash/prompt/CrashReporterActivity.kt
Shubhamurkade: Don't I need to add a separate variable for "mozac_lib_crash_settings" file?

Shubhamurkade:
> After fixing the last issues, can you please rebase and squash this PR so that there's only a single commit

components/lib/crash/src/main/java/mozilla/components/lib/crash/prompt/CrashReporterActivity.kt
pocmo: If it's just in this place then it may be fine.

If you want to you can create a const above the class though:

```Kotlin
private const val SETTINGS_FILE = "mozac_lib_crash_settings"
```

Shubhamurkade:
I think I've screwed my branch during rebase. I'm new to VCS. Could you tell me what can I do about it? I thought deleting this branch and creating a new one would help. But, seems like it'll mess up more things

Shubhamurkade:
> > After fixing the last issues, can you please rebase and squash this PR so that there's only a single commit? :)
> 
> Do I need another review after making the latest changes?

> I'm new to git. I've squashed the commits in my forked repository's branch. What do I need to do next?

PR #1280, opened 2018-11-06
Add domain suggestions to concept-storage and its implementations

grigoryk:
Basic matching logic for now, follow-up in https://github.com/mozilla-mobile/android-components/issues/1279.


components/support/utils/src/main/java/mozilla/components/support/utils/DomainMatcher.kt
```diff
@@ -0,0 +1,49 @@
...
+
+package mozilla.components.support.utils
+
+import java.net.URL
+
+const val WWW_PREFIX_OFFSET = 4
+
+// FIXME implement Fennec-style segment matching logic
+// See https://github.com/mozilla-mobile/android-components/issues/1279
+object DomainMatcher {
```
pocmo: I wonder whether we actually need this Singleton object or if we could just create a function that can be imported :)

PR #1281, opened 2018-11-06
Cleanup deprecations and warnings throughout the components

grigoryk:


components/browser/domains/src/main/java/mozilla/components/browser/domains/DomainAutoCompleteProvider.kt
```diff
@@ -116,11 +117,14 @@ class DomainAutoCompleteProvider {
...
         if (loadDomainsFromDisk) {
-            launch(UI) {
-                val domains = async(CommonPool) { Domains.load(context) }
-                val customDomains = async(CommonPool) { CustomDomains.load(context) }
-
-                onDomainsLoaded(domains.await(), customDomains.await())
+            CoroutineScope(Dispatchers.Main).launch {
+                // If a 'load' call throws, cancel the other one and re-throw.
+                coroutineScope {
+                    val domains = async { Domains.load(context) }
```
csadilek: Hm this one doesn't seem right to me. Before, we launched on the UI but had the `load`s run in a separate thread pool. Now they're in a nested scope but still run on the Main/UI thread, no?

components/browser/search/src/main/java/mozilla/components/browser/search/SearchEngineManager.kt
```diff
@@ -94,8 +99,8 @@ class SearchEngineManager(
     private val localeChangedReceiver by lazy {
         object : BroadcastReceiver() {
             override fun onReceive(context: Context, intent: Intent?) {
-                launch(CommonPool) {
-                    load(context.applicationContext)
+                CoroutineScope(Dispatchers.Default).launch {
```
pocmo: Is there any difference (here and in the other places) between creating a scope with the default dispatcher and just using the `GlobalScope`? If not then I think we can replace those generic scope creations with just the global one until we actually make use of some local scope?

components/browser/domains/src/main/java/mozilla/components/browser/domains/DomainAutoCompleteProvider.kt
```diff
@@ -116,11 +117,14 @@ class DomainAutoCompleteProvider {
...
 
         if (loadDomainsFromDisk) {
-            launch(UI) {
-                val domains = async(CommonPool) { Domains.load(context) }
-                val customDomains = async(CommonPool) { CustomDomains.load(context) }
-
-                onDomainsLoaded(domains.await(), customDomains.await())
+            CoroutineScope(Dispatchers.Main).launch {
+                // If a 'load' call throws, cancel the other one and re-throw.
+                coroutineScope {
```
pocmo: Those async calls are already part of the outer scope you create with `CoroutineScope()`. There shouldn't be any need to wrap them in another scope afaik?

components/browser/domains/src/main/java/mozilla/components/browser/domains/DomainAutoCompleteProvider.kt
grigoryk: Yeah, good call - the new semantics are different.

components/browser/domains/src/main/java/mozilla/components/browser/domains/DomainAutoCompleteProvider.kt
pocmo: But this should work with the outer scope already?

grigoryk:
Ended up rewriting the contentious bit entirely - take another look, @pocmo @csadilek ?

grigoryk:
Note that the loading behaviour is slightly different (lists of domains will become available for matching one by one), but I think that's fine.

PR #1282, opened 2018-11-06
Add browser-storage-sync readme and mention it in global readme. [ci skip]

grigoryk:
Forgot to do that in the original PR.


grigoryk:
Does codecov want tests for the readme files now? :)

jonalmeida:
Maybe it wants you to assert the links are correct. :)

I'm going to restart the build since it's comparing your branch against my last change which isn't at the top of the tree, so it's just confused itself.

jonalmeida:
Ideally for readme updates we can also just add `[ci skip]` to get an automatic green.

PR #1284, opened 2018-11-06
Fetch artifact information from gradle

mitchhentges:
[Bugzilla 1500149](https://bugzilla.mozilla.org/show_bug.cgi?id=1500149)

First PR! :wave: :tada:


automation/taskcluster/decision_task_release.py
```diff
@@ -29,15 +31,26 @@
 )
 
 
-def load_artifacts_manifest():
-    return yaml.safe_load(open('automation/taskcluster/artifacts.yml', 'r'))
+def fetch_artifact_metadata():
```
MihaiTabara: Nit (not blocking): maybe it's worth extracting L35-40 into a function somewhere in an `utils.py` module or alike and reuse in both the decision tasks? (https://github.com/mozilla-mobile/android-components/pull/1284/files#diff-168d551d2c0830884a5caeb50ae576e9R29 too)

build.gradle
```diff
@@ -209,7 +209,7 @@ task ktlint(type: JavaExec, group: "verification") {
 task printModules {
     doLast {
         subprojects.each { p ->
-            println "module: " + p.path
+            println "module: name=" + p.path + " buildPath=" + p.buildDir
```
ncalexan: Maybe just make this JSON or YAML or something and make parsing really, really straightforward?

automation/taskcluster/lib/artifacts.py
```diff
@@ -0,0 +1,22 @@
...
+
+    if exit_code is not 0:
+        print "Gradle command returned error:", exit_code
+
+    tuples = re.findall('module: name=(.*) buildPath=(.*)', output, re.M)
+    return map(lambda (name, build_path): {
+        'name': name[1:],
+        'artifact': 'public/build/' + name[1:].replace('-', '.', 1) + '.maven.zip',
+        'path': build_path + '/target.maven.zip'
+    }, tuples)
```
pocmo: Note that artifacts.yml is smaller than all gradle modules. All gradle modules also contains sample apps and modules we do not publish (e.g. tooling-lint). As far as I can see we do not filter those out yet. There may be no good automated way to figure out what module should be published - so I wonder whether we should just add an additional flag to settings.gradle, where we configure other things already like the description.

MihaiTabara: I had this in mind too last night but wasn't 100% sure it's accurate so I'm happy you raised it. I think the solutions with additional flags in `settings.gradle` is great as it scales better if we wanted to add more tasks in the graph (signing, checksums, etc) later on.

automation/taskcluster/lib/artifacts.py
```diff
@@ -0,0 +1,22 @@
...
+    (output, err) = process.communicate()
+    exit_code = process.wait()
+
+    if exit_code is not 0:
+        print "Gradle command returned error:", exit_code
+
+    tuples = re.findall('module: name=(.*) buildPath=(.*)', output, re.M)
+    return map(lambda (name, build_path): {
+        'name': name[1:],
+        'artifact': 'public/build/' + name[1:].replace('-', '.', 1) + '.maven.zip',
```
JohanLorenzo: Nit: Even though string concatenation (with `+`) is faster, `''.format()` is generally preferred when dealing with a handful of strings https://stackoverflow.com/questions/376461/string-concatenation-vs-string-substitution-in-python. It's a generally a matter of readability. 

We can also get rid of `.replace('-', '.', 1)`. The build and the beetmover task are the only consumers of this line (if I remember correctly). The reason we made it this way is because:
* artifacts.yml was a quick hack to get a-c deployed on maven.mozilla.org, a month and a half ago (https://github.com/mozilla-mobile/android-components/pull/803)
* some components got their name changed, in the meantime. 

This would become:

```py
# 
'artifact': 'public/build/{}.maven.zip'.format(name[1:]),
```

build.gradle
JohanLorenzo: +1

Another note: I love the fact we don't hardcode buildPaths anymore \o/

automation/taskcluster/decision_task.py
```diff
@@ -126,7 +115,7 @@ def create_compare_locales_task():
 
     queue = taskcluster.Queue({ 'baseUrl': 'http://taskcluster/queue/v1' })
```
srfraser: `baseUrl` is deprecated in taskcluster-client.py for 4.*, and removed in 5.* (https://github.com/taskcluster/taskcluster-client.py/blob/master/taskcluster/client.py#L58).  Maybe this is a good time to change to `rootUrl`, the new option. I don't know where the module version is specified for this deployment.

build.gradle
mitchhentges: My concern with using JSON/YAML is that it's not really consistent with the rest of the `gradle` output - there'd be a bunch of human-readable output, followed by a clump of JSON. I'd need to deliminate the JSON line, so that the importing tools wouldn't accidentally use the other `gradle` output as part of the JSON parse. Considering this, it seemed simpler and more elegant to follow the same style as the rest of the output

-----

Looking into this further, `gradle` has a `--quiet/-q` option which means that we can have `gradle` _only_ output our task's output. This means that executing `gradle -q printModulesJson` could strictly dump a block of machine-readable JSON, which is nice. I'll follow that approach

automation/taskcluster/lib/artifacts.py
mitchhentges: Really great point! I should do a more-complete diff of tasks created before vs. after this change.
Great catch finding this in code review, thanks! :100:

automation/taskcluster/lib/artifacts.py
```diff
@@ -0,0 +1,22 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import re
+import subprocess
+
+
+def from_gradle():
```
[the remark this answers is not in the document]
mitchhentges: I was leaning on the filename here to provide greater context of what this was about.
For example, reading a usage of this function gives a better hint:

> `artifacts.from_gradle()`

Though, I can see how it looks like it's doing something regarding artifacts themselves, not just artifact metadata/module definitions.

What if I called the file `module_definitions`, and had it like:

> `artifacts_info = module_definitions.from_gradle()`

?

automation/taskcluster/lib/artifacts.py
```diff
@@ -0,0 +1,22 @@
...
+    process = subprocess.Popen(["./gradlew", "--no-daemon", "printModules"], stdout=subprocess.PIPE)
+    (output, err) = process.communicate()
+    exit_code = process.wait()
+
+    if exit_code is not 0:
+        print "Gradle command returned error:", exit_code
+
+    tuples = re.findall('module: name=(.*) buildPath=(.*)', output, re.M)
+    return map(lambda (name, build_path): {
+        'name': name[1:],
```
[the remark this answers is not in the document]
mitchhentges: I'll add a comment, great point!

Unrelated, but I noticed that `decision_task` and `decision_task_release` both handle the name a little differently: `decision_task` loves to keep the `:`, but `decision_task_release` uses name without the `:` in a lot of places. I didn't investigate fully, but since, in the end, it's just creating a graph for `taskcluster`, I'd love to make these more consistent. Probably not something for this PR, but a bit of a :thinking:

automation/taskcluster/lib/artifacts.py
mitchhentges: Awesome, I wanted to avoid breaking changes, but if I've got your go-ahead, then I'd love to make this simpler!

automation/taskcluster/decision_task.py
mitchhentges: IMHO, I'd like to keep this PR specific to updating how we source artifact metadata.
Is there a good place to note that this needs to be updated? @MihaiTabara mentioned a trello this morning, or is bugzilla a better approach? @JohanLorenzo too :)

automation/taskcluster/decision_task.py
```diff
@@ -126,7 +115,7 @@ def create_compare_locales_task():
...
 
-    modules = fetch_module_names()
+    modules = map(lambda artifact: ':' + artifact['name'], lib.artifacts.from_gradle())
```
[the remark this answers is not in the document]
mitchhentges: List comprehensions are neat!

settings.gradle
```diff
@@ -11,7 +11,17 @@ def setupProject(name, path, description) {
...
 
+    // TODO instead of using global "componentDescriptions" map, attach description to "project.ext"
     gradle.ext.componentDescriptions[name.substring(1)] = description
+
+    // project(name) gives us a skeleton project that we can't set ext.* on
+    gradle.afterProject { project ->
+        // However, the "afterProject" listener iterates over every project and gives us the actual project
+        // So, once we filter for the project we care about, we can set whatever we want
+        if (project.name == name.substring(1)) {
+            project.ext.enableTaskcluster = enableTaskcluster
```
mitchhentges: @pocmo I think you'll really like this part: with a bit of `gradle` dancing, we can assign custom properties right on the project itself!

pocmo: @mitchhentges Nice! That's something I failed to do since there's no `Project` defined at this point. I didn't know the right Gradle idiom for that. :)

My only nit here would be to rename `enableTaskcluster` to something more task oriented like `publishLibrary` or `publishToMavenRepository` since those other modules still get build and tested on taskcluster and this may be confusing.

settings.gradle
```diff
@@ -99,16 +109,16 @@ setupProject(':lib-jexl', 'components/lib/jexl', 'Javascript Expression Language
 // Tooling
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
-setupProject(':tooling-lint', 'components/tooling/lint', 'Custom Lint checks for using and writing components.')
+setupProject(':tooling-lint', 'components/tooling/lint', 'Custom Lint checks for using and writing components.', false)
```
pocmo: Does `Groovy` allow to use named parameters to make it more obvious what `false` means?

automation/taskcluster/lib/artifacts.py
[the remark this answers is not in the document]
JohanLorenzo: `module_definitions.from_gradle()` is great! Good idea :wink:

automation/taskcluster/lib/artifacts.py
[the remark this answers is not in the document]
JohanLorenzo: Good catch on making these consistent! +1 to that (in a followup PR). For the record, we haven't investigated too much into making these graphs generic across Android projects. That's something we'd like to invest on.

automation/taskcluster/decision_task.py
JohanLorenzo: Agreed on keeping this PR specific :+1: 

https://bugzilla.mozilla.org/show_bug.cgi?id=1460015 tracks this issue. IMO, this is good enough for now. We can revisit this, if this becomes too much of a hassle.

MihaiTabara: I've added an item in trello as well to keep this under the radar.  https://trello.com/c/aWjXAAke/10-once-the-taskcluster-github-integration-is-solved-with-taskcluster-5-we-should-update-it

settings.gradle
mitchhentges: Unfortunately not, no.
It _does_ allow [named parameters](http://docs.groovy-lang.org/latest/html/documentation/#_named_arguments), but that'd require `setupProject` to take a single parameter (a map), but I believe that that will reduce usability

pocmo: Okay, no problem. With Gradle 5 I'll try to migrate all that nasty Groovy to Kotlin :)

PR #1285, opened 2018-11-07
Closes #1157: Add API for permission requests.

### LINKED ISSUE #1157: Add API for permission requests
Like #1078 for geolocation requests.

See #1110 for related "events that require callback" discussion.

We currently allow protected media requests (EME APIs) by default in `browser-engine-system` (#1128).

csadilek:


components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/permission/GeckoPermissionRequest.kt
```diff
@@ -0,0 +1,150 @@
...
+import org.mozilla.geckoview.GeckoSession.PermissionDelegate.MediaSource.SOURCE_WINDOW
+import android.Manifest.permission.ACCESS_COARSE_LOCATION
+import android.Manifest.permission.ACCESS_FINE_LOCATION
+import android.Manifest.permission.CAMERA
+import android.Manifest.permission.RECORD_AUDIO
+
+/**
+ * Gecko-based implementation of [PermissionRequest].
+ *
+ * @property uri the URI of the content requesting the permissions.
```
Amejia481: Nit: This `@property uri` is duplicated in line 42 and also other variants

samples/browser/src/main/java/org/mozilla/samples/browser/BrowserFragment.kt
```diff
@@ -69,6 +73,12 @@ class BrowserFragment : Fragment(), BackHandler, DownloadDialogListener {
         downloadsFeature.onNeedToRequestPermissions = { _, _ ->
             requestPermissions(arrayOf(WRITE_EXTERNAL_STORAGE), PERMISSION_WRITE_STORAGE_REQUEST)
         }
+
+        // Accept all DRM content
+        object : SelectionAwareSessionObserver(components.sessionManager) {
```
Amejia481: Should this be a member class variable and call `observeSelected()` method on `onStop()`?

components/concept/engine/src/main/java/mozilla/components/concept/engine/permission/PermissionRequest.kt
```diff
@@ -0,0 +1,80 @@
...
+    fun reject()
+}
+
+/**
+ * Represents all the different supported permission types.
+ *
+ * @property id an optional native engine-specific ID of this permission.
+ * @property desc an optional description of what this permission type is for.
+ */
+sealed class Permission(open val id: String? = "", open val desc: String? = "") {
```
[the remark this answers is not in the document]
csadilek: @jonalmeida yes, I found that annoying too, but it's needed. Basically if you don't provide the parameter you will get an empty string, which is fine as we don't always have/need an `id` and `description`, but for Gecko media permission requests both `id` and `name` can technically be null, and then we don't want to blow up:

https://github.com/mozilla-mobile/android-components/pull/1285/files#diff-4c2633d99f27336160a07ca2ae6f45f9R129

We could add the checks there but then we'd sprinkle `?: ""` all over the place which isn't much nicer (harder to read)?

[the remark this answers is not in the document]
jonalmeida: That's a good judgement call. 👍

components/concept/engine/src/main/java/mozilla/components/concept/engine/permission/PermissionRequest.kt
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Represents a permission request, used when engines need access to protected
+ * resources. Every request must be handled by either calling [grant] or [reject].
+ */
+interface PermissionRequest {
+    /**
+     * The origin URI which caused the permissions to be requested.
+     */
+    val uri: String
```
pocmo: nit: This may not be 100% correct for an App Permission. They are not for a specific origin and they may not be caused by the "current URL". I think it's okay to have this for the sake of having a shared permission request interface - however I wonder whether we should make this nullable and have this set to null for App Permissions.

samples/browser/src/main/java/org/mozilla/samples/browser/BrowserFragment.kt
```diff
@@ -69,6 +74,12 @@ class BrowserFragment : Fragment(), BackHandler, DownloadDialogListener {
         downloadsFeature.onNeedToRequestPermissions = { _, _ ->
             requestPermissions(arrayOf(WRITE_EXTERNAL_STORAGE), PERMISSION_WRITE_STORAGE_REQUEST)
         }
+
+        // Accept all DRM content
+        permissionObserver = object : SelectionAwareSessionObserver(components.sessionManager) {
+            override fun onContentPermissionRequested(session: Session, permissionRequest: PermissionRequest): Boolean =
+                permissionRequest.grantIf { it is Permission.ContentProtectedMediaId }
+        }
```
pocmo: Not sure whether we need this in the sample? :)

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -302,14 +303,12 @@ class SystemEngineView @JvmOverloads constructor(
...
+
         override fun onPermissionRequest(request: PermissionRequest) {
-            if (request.resources.contains(PermissionRequest.RESOURCE_PROTECTED_MEDIA_ID)) {
-                // For now we automatically grant playing protected media (EME APIs).
-                // See https://github.com/mozilla-mobile/android-components/issues/1128
-                request.grant(arrayOf(PermissionRequest.RESOURCE_PROTECTED_MEDIA_ID))
-            } else {
-                super.onPermissionRequest(request)
-            }
+            session?.internalNotifyObservers { onContentPermissionRequest(SystemPermissionRequest(request)) }
```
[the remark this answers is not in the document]
csadilek: Yes, definitely, I will add a commit here.

components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
```diff
@@ -258,6 +261,29 @@ class Session(
...
+    /**
+     * [Consumable] permission request for the device. A [PermissionRequest]
+     * must be consumed i.e. either [PermissionRequest.grant] or
+     * [PermissionRequest.reject] must be called.
+     */
+    var devicePermissionRequest: Consumable<PermissionRequest> by Delegates.vetoable(Consumable.empty()) {
+        _, _, request ->
+            val consumers = wrapConsumers<PermissionRequest> { onDevicePermissionRequested(this@Session, it) }
+            !request.consumeBy(consumers)
+    }
```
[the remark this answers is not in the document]
csadilek: Yeah, makes sense, turning the refactoring wheels.

samples/browser/src/main/java/org/mozilla/samples/browser/BrowserFragment.kt
csadilek: We don't, it was mostly for testing and showing the API. Also, Gecko doesn't send this permission request anyway. It's a setting in about:config whether or not DRM / EME content is allowed.

components/concept/engine/src/main/java/mozilla/components/concept/engine/permission/PermissionRequest.kt
csadilek: I will take a look!

csadilek:
@pocmo this is updated now. Can you take quick peek to approve/merge?

PR #1294, opened 2018-11-07
Update tooling-lint to Java 8..

jonalmeida:
Noticed that tooling-lint was the only module left that wasn't updated yet.


pocmo:
Afaik only browser-engine-gecko nightly/beta and the sample browser are on Java 8? :)

jonalmeida:
Yeah, that's true, but while I was doing other things, I noticed that everything else that had target/source compatibility was already on Java 8 and since 8 is the standard I think it should be safe to upgrade this too for consistency.

No strong feelings here, so feel free to close otherwise.

Issue #1295, opened 2018-11-07
Sample with Servo engine doesn't work


Dmitry-Borodin:
Current master, trying to run sample-browser with Servo on ARM

The reason is ServoEngine.kt:35

> Process: org.mozilla.samples.browser, PID: 7616
>     java.lang.RuntimeException: Unable to start activity ComponentInfo{org.mozilla.samples.browser/org.mozilla.samples.browser.BrowserActivity}: java.lang.UnsupportedOperationException: Not supported by this implementation:
>                 Use EngineSession.settings instead
>         at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2778)
>         at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2856)
>         at android.app.ActivityThread.-wrap11(Unknown Source:0)
>         at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1589)
>         at android.os.Handler.dispatchMessage(Handler.java:106)
>         at android.os.Looper.loop(Looper.java:164)
>         at android.app.ActivityThread.main(ActivityThread.java:6494)
>         at java.lang.reflect.Method.invoke(Native Method)
>         at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:438)
>         at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:807)
>      Caused by: java.lang.UnsupportedOperationException: Not supported by this implementation:
>                 Use EngineSession.settings instead
>         at mozilla.components.browser.engine.servo.ServoEngine.getSettings(ServoEngine.kt:35)
>         at mozilla.components.feature.storage.HistoryTrackingFeature.<init>(HistoryTrackingFeature.kt:20)
>         at org.mozilla.samples.browser.BrowserFragment.onViewCreated(BrowserFragment.kt:42)
>  <...>

Issue #1028, opened 2018-10-11
Add Servo X86 flavor to sample browser


pocmo:
After #1027 we could add an x86 flavor to make testing on the emulator nicer.

PR #1296, opened 2018-11-07
Issue #1028 adding servo x86 flavor.

Dmitry-Borodin:
Added servo flavors similar to how gecko flavors implemented.

Tested with emulator - is it installed and crashing the same was as current arm version crashing on arm ( https://github.com/mozilla-mobile/android-components/issues/1295 )
Arm version is not even installing, this is the only way I found to test my changes.

I assumed, that servo, similar to gecko, has the same API for all flavors, so we can compile module only against ARM and run with x86 (and I didn't see this promise in the servo docs)

Another thing, in gecko modules there is a comment in gradle config
    // We only compile against the ARM artifact. We let the app decide which artifact to actually
    // use at runtime. As the Kotlin/Java API is the same for all ABIs it is not important which one
    // we import here.
While as I can see, artifact is chosen in build flavor configuration, during build time, not at runtime. Should I change this comments as well?


samples/browser/build.gradle
```diff
@@ -80,7 +80,7 @@ android {
             setIgnore(true)
         }
 
-        if (flavors.contains("servo") && !flavors.contains("arm")) {
+        if (flavors.contains("servo") && (flavors.contains("universal") || flavors.contains("aarch64"))) {
```
pocmo: Let's reverse that check to check for ARM or x86. Otherwise this needs to be updated for every new ABI we add (x86_64 may get added soon).

Dmitry-Borodin: Will do in 8 hours.

Dmitry-Borodin: fixed

components/browser/engine-gecko-beta/build.gradle
```diff
@@ -26,9 +26,8 @@ dependencies {
     implementation project(':support-ktx')
     implementation project(':support-utils')
 
-    // We only compile against the ARM artifact. We let the app decide which artifact to actually
-    // use at runtime. As the Kotlin/Java API is the same for all ABIs it is not important which one
-    // we import here.
+    // We only compile against the ARM artifact. External module will decide which module to provide by build configuration.
```
Dmitry-Borodin: Note, I changed comments for gecko engine as well

Dmitry-Borodin:
Resolved merge conflicts to master, 
checked browser compiling.

pocmo:
PR 1638 landed so this can finally get in :)

pocmo:
Finally landed this as #1979. Thank you for those patches. :)

PR #1304, opened 2018-11-08
Update GeckoView Nightly to 65.0.20181107100135.

pocmo:
Needed to fix the error API.


pocmo:
> Should we also update the Reference-browser version?

Nope. It will need the changes in the component. We should wait for 0.31.0 :)

Amejia481:
Yeah, you're right! 👍

PR #1306, opened 2018-11-08
Fix mozilla/glean_parser#12 Upgrade glean_parser if at the incorrect version

mdboom:
This adds a little bootstrap script to ensure that glean_parser is at the desired version, optionally upgrading it using pip, before running the glean_parser script.


jonalmeida:
Thanks for the quick turn around @mdboom !

PR #1307, opened 2018-11-08
Closes #177: Show site security icon when page security changes

### LINKED ISSUE #177: Add support to show security indicator in toolbar
DisplayToolbar structure:

```
 +------+-----+--------------------+---------+------+
 | icon | nav |         url        | actions | menu |
 +------+-----+--------------------+---------+------+
```

We want to show a security indicator (lock / globe) in the toolbar depending on the security level of the current site.

* Engine: Hook up security level callbacks
* Session: Expose security level state
* Toolbar: Add option to display security levels

jonalmeida:
TODO: Report GV bugs mentioned in #177.

### DIFF
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
@@ -187,6 +187,12 @@ class BrowserToolbar @JvmOverloads constructor(
             field = value
         }
 
+    override var siteSecure: Toolbar.SiteSecurity = Toolbar.SiteSecurity.INSECURE
+        set(value) {
+            displayToolbar.setSiteSecurity(value)
+            field = value
+        }
+
     init {
         context.obtainStyledAttributes(attrs, R.styleable.BrowserToolbar, defStyleAttr, 0).apply {
             attrs?.let {
--- a/components/concept/toolbar/src/main/java/mozilla/components/concept/toolbar/Toolbar.kt
+++ b/components/concept/toolbar/src/main/java/mozilla/components/concept/toolbar/Toolbar.kt
@@ -27,6 +27,11 @@ interface Toolbar {
     var url: String
 
     /**
+     * Sets/Gets the site security to be displayed on the toolbar.
+     */
+    var siteSecure: SiteSecurity
+
+    /**
      * Displays the currently used search terms as part of this Toolbar.
      *
      * @param searchTerms the search terms used by the current session
@@ -312,5 +317,10 @@ interface Toolbar {
         }
 
         override fun bind(view: View) = Unit
+    }
+
+    enum class SiteSecurity {
+        INSECURE,
+        SECURE,
     }
 }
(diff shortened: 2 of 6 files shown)


components/concept/toolbar/src/main/java/mozilla/components/concept/toolbar/Toolbar.kt
```diff
@@ -26,6 +26,11 @@ interface Toolbar {
      */
     var url: String
 
+    /**
+     * Sets/Gets the site security to be displayed on the toolbar.
+     */
+    var siteSecure: Boolean
```
pocmo: In preparation for the future we may want to turn this into an enum. It's likely that we want to show other states here too. :)

Amejia481: In that case, it won't be better to use sealed classes ?

jonalmeida: @Amejia481 I don't see anything that we'd want to hold in the sealed class. The enum value should be good enough.

Amejia481: 👍

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
```diff
@@ -187,6 +187,12 @@ class BrowserToolbar @JvmOverloads constructor(
             field = value
         }
 
+    override var siteSecure: Toolbar.SiteSecurity = Toolbar.SiteSecurity.INSECURE
+        set(value) {
+            displayToolbar.setSiteSecurity(value)
+            field = value
+        }
```
pocmo: I wonder if we should turn this into `setSiteSecurity` too. Not sure if we want to keep this as a readable property?

jonalmeida: That's a fair point. I want to say it'll be useful for browser apps that want to read the the toolbar state if they want to style it differently, but they should really be using the observers in that case.

I'll remove the read ability and merge.

jonalmeida: I thought about it a bit more. It's not that bad to have it visible (for now) and it also makes testing a little easier, so I'm going to leave it as is.

pocmo:
> I thought about it a bit more. It's not that bad to have it visible (for now) and it also makes testing a little easier, so I'm going to leave it as is.

I'd argue that tests should verify things on the View.(s) The property is meaningless. :)

jonalmeida:
> > I thought about it a bit more. It's not that bad to have it visible (for now) and it also makes testing a little easier, so I'm going to leave it as is.
> 
> I'd argue that tests should verify things on the View.(s) The property is meaningless. :)

In BrowserToolbar, we don't have direct access to the Views. 🤔 

We can access them though by going one layer deeper with `toolbar.displayToolbar.iconView` but it feels like that should be done by DisplayToolbar's tests (which is does).

I guess it could have been better written to only rely on the mock the displayToolbar making the call.

PR #1308, opened 2018-11-08
[Glean] Enable 'user' persistence on supported metrics.

Dexterp37:
This PR enables support for all the `Lifetime` values (`Ping`, `User` and `Application`) in the metrics. See the [spec](https://docs.google.com/document/d/186GI_2ljdqWqWTBk5P3QjHD3BwpMyd8kRayRAuz1-dk/edit#) for additional details.

**`Lifetime` tl;dr**

- the `Ping` lifetime will clear the value stored for the metric whenever a ping in which it is contained is collected;
- the `Application` lifetime will clear the value when the application is closed;
- the `User` lifetime will store the values to disk making them available again next time the application is started again (think of the `client_id`).

**Rationale**
The backend support for storing metrics with `User` lifetime is provided by the [SharedPreferences](https://developer.android.com/reference/android/content/SharedPreferences#getAll()). This requires us to have a valid `Context` from which to get the `SharedPreferences`, which is passed on to the API `StorageEngine`(s) by the `StorageEngineManager`.

The backing storage is loaded off the disk the first time we attempt to write a metric with `User` lifetime or when a snapshot is requested.

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/EventMetricType.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/EventMetricType.kt
@@ -60,6 +60,11 @@ data class EventMetricType(
             return
         }
 
+        if (lifetime != Lifetime.Ping) {
+            logger.warn("$category.$name can only have a Ping lifetime")
+            return
+        }
+
         // We don't need to check that the objectId is short, since that
         // has already been determined at build time for each of the valid objectId values.
         if (!objects.contains(objectId)) {
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/UuidMetricType.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/UuidMetricType.kt
@@ -53,6 +53,7 @@ data class UuidMetricType(
                 stores = sendInPings,
                 category = category,
                 name = name,
+                lifetime = lifetime,
                 value = value
         )
     }
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
@@ -4,6 +4,8 @@
 
 package mozilla.components.service.glean.storages
 
+import android.annotation.SuppressLint
+import android.content.Context
 import android.os.SystemClock
 import android.support.annotation.VisibleForTesting
 import org.json.JSONArray
@@ -12,8 +14,15 @@ import org.json.JSONArray
  * This singleton handles the in-memory storage logic for events. It is meant to be used by
  * the Specific Events API and the ping assembling objects. No validation on the stored data
  * is performed at this point: validation must be performed by the Specific Events API.
+ *
+ * This class contains a reference to the Android application Context. While the IDE warns
+ * us that this could leak, the application context lives as long as the application and this
+ * object. For this reason, we should be safe to suppress the IDE warning.
  */
+@SuppressLint("StaticFieldLeak")
 internal object EventsStorageEngine : StorageEngine {
+    override lateinit var applicationContext: Context
+
     // Events recorded within the list should be reasonably sorted by timestamp, assuming
     // the sequence of calls to [record] has not been messed with. However, please only
     // trust the recorded timestamp values.
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
@@ -182,22 +182,17 @@ abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
         @SuppressLint("CommitPrefEdits")
         val userPrefs: SharedPreferences.Editor? =
             if (lifetime == Lifetime.User) userLifetimeStorage.edit() else null
-        for (storeName in stores) {
-            val storeData = dataStores[lifetime.ordinal].getOrPut(storeName) { mutableMapOf() }
+        stores.forEach {
+            val storeData = dataStores[lifetime.ordinal].getOrPut(it) { mutableMapOf() }
             val entryName = "$category.$name"
             storeData[entryName] = value
             // Persist data with "user" lifetime
             if (lifetime == Lifetime.User) {
-                userPrefs?.putString("$storeName#$entryName", serializeSingleMetric(value))
+                userPrefs?.putString("$it#$entryName", serializeSingleMetric(value))
             }
         }
         userPrefs?.apply()
     }
 
-    @VisibleForTesting
-    internal open fun clearAllStores() {
-        for (store in dataStores) {
-            store.clear()
-        }
-    }
+    internal open fun clearAllStores() = dataStores.forEach { it.clear() }
 }
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
@@ -94,6 +94,7 @@ abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
 
         // Make sure data with "user" lifetime is loaded before getting the snapshot.
         // We still need to catch exceptions here, as `getAll()` might throw.
+        @Suppress("TooGenericExceptionCaught")
         try {
             userLifetimeStorage.all
         } catch (e: NullPointerException) {
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
@@ -7,7 +7,6 @@ package mozilla.components.service.glean.storages
 import android.annotation.SuppressLint
 import android.content.Context
 import android.content.SharedPreferences
-import android.support.annotation.VisibleForTesting
 import mozilla.components.service.glean.Lifetime
 import mozilla.components.support.base.log.logger.Logger
 import org.json.JSONObject
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
@@ -1,0 +1,160 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.glean.storages
+
+import android.annotation.SuppressLint
+import android.content.Context
+import android.content.SharedPreferences
+import android.support.annotation.VisibleForTesting
+import mozilla.components.service.glean.Lifetime
+import mozilla.components.support.base.log.logger.Logger
+import org.json.JSONObject
+
+internal typealias GenericDataStorage<T> = MutableMap<String, T>
+internal typealias GenericStorageMap<T> = MutableMap<String, GenericDataStorage<T>>
+
+abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
+    override lateinit var applicationContext: Context
+
+    // Let derived class define a logger so that they can provide a proper name,
+    // useful when debugging weird behaviours.
+    abstract val logger: Logger
+
+    protected val userLifetimeStorage: SharedPreferences by lazy { deserializeUserLifetime() }
+
+    // Use a multi-level map to store the data for the different lifetimes and
+    // stores: Map[Lifetime, Map[StorageName, ScalarType]].
+    protected val dataStores: Map<String, GenericStorageMap<ScalarType>> =
+        Lifetime.values().associateBy(
+            { it.toString() },
+            { mutableMapOf<String, GenericDataStorage<ScalarType>>() }
+        )
+
+    /**
+     * Implementor's provided function to convert deserialized 'user' lifetime
+     * data to the destination ScalarType.
+     *
+     * @param value loaded from the storage as [Any]
+     *
+     * @return data as [ScalarType] or null if deserialization failed
+     */
+    abstract fun singleMetricDeserializer(value: Any?): ScalarType?
+
+    /**
+     * Deserialize the metrics with a lifetime = User that are on disk.
+     * This will be called the first time a metric is used or before a snapshot is
+     * taken.
+     *
+     * @return A [SharedPreferences] reference that will be used to inititialize [userLifetimeStorage]
+     */
+    @Suppress("TooGenericExceptionCaught")
+    open fun deserializeUserLifetime(): SharedPreferences {
+        val prefs =
+            applicationContext.getSharedPreferences(this.javaClass.simpleName, Context.MODE_PRIVATE)
+
+        return try {
+            for ((metricName, metricValue) in prefs.all.entries) {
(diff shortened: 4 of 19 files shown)


Dexterp37:
@pocmo could you or somebody on the android-components team give a first pass on this? I'm specifically interested in knowing if there's a more Kotlin idiomatic way to do what I'm doing in `UuidsStorageEngineImplementation` .

components/service/glean/src/main/java/mozilla/components/service/glean/UuidMetricType.kt
```diff
@@ -53,6 +53,7 @@ data class UuidMetricType(
                 stores = sendInPings,
                 category = category,
                 name = name,
+                lifetime = lifetime,
```
mdboom: Random thought: Would it be better to pass `this as CommonMetricData` to the storage engine rather than passing all of this common metadata independently.  Should require fewer updates to all of the storage engines as we iterate on this stuff.

Dexterp37: That is a great idea! Let's take this to a separate bug though, and defer after MVP (I filed bug 1506061).

Dexterp37:
> 
> 
> This looks good.
> 
> My biggest concern is that we have some pretty nitpicky logic here that we should try not to duplicate across all of the storage engines. We plan to have lots of those and keeping them in sync will be difficult.

I agree with you on this. We should find a proper way to embed this behaviour in some class.

> What if, instead, lifetimes were handled by the `StorageEngineManager` rather than each storage engine. I'm thinking the `StorageEngineManager` would have three sets of storage engines (for each lifetime), each having one of each of the storage engines. (That implies the storage engines would no longer by singletons). Then the storage manager handles which storage engine recording needs to go to, based on the lifetime, and could handle setting off the serializing/deserializing of all of the storage engines in the user group only.

As it is design, the current Specific Metric API requires the `StorageEngine`(s) to be objects, otherwise it would not be possible to reference them from the dynamically generated Kotlin code (unless I'm missing something). Moreover, different engines might have different ways of persisting metrics with `user` lifetime. Think of `UUID` vs `Use counters`: while the former requires only one entry in the `SharedPreferences`, the latter might require something more complex than that. For this reason, I thought to delegate the specifics of serialization to each engine. 

However, nothing prevents us from having shared engine implementations. For example, we could have a `Generic` "scalar" storage engine that would work for all scalar types (Strings, Numbers, UUIDs) with minor adjustments.  @mdboom what do you think?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/UuidsStorageEngine.kt
```diff
@@ -49,16 +89,76 @@ internal object UuidsStorageEngine {
...
 
-        return uuidStores.get(storeName)
+        return if (allLifetimes.isNotEmpty()) allLifetimes else null
+    }
+
+    /**
+     * Get a snapshot of the stored data as a JSON object.
+     *
+     * @param storeName the name of the desired store
+     * @param clearStore whether or not to clearStore the requested store
```
badboy:
```suggestion
     * @param clearStore whether or not to clear the requested store
```

components/service/glean/src/main/java/mozilla/components/service/glean/storages/UuidsStorageEngine.kt
```diff
@@ -4,17 +4,47 @@
...
+@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+open class UuidsStorageEngineImplementation : StorageEngine {
+    override lateinit var applicationContext: Context
+
+    private val userLifetimeStorage: SharedPreferences by lazy { deserializeUserLifetime() }
+    private val logger = Logger("glean/HttpPingUploader")
+
+    // Use a multi-level map to store the data for the different lifetimes and
+    // stores: Map[Lifetime, Map[StorageName, UUID]].
+    private val uuidStores: Map<String, UUIDMap> =
```
badboy: How expensive are strings? I wonder if this can be cheaper when it's just a `Map<Lifetime, UUIDMap>`.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/UuidsStorageEngine.kt
badboy: Should this doc contain that it only clears metrics with a Ping lifetime?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/UuidsStorageEngine.kt
mdboom: Since `Lifetime` is an enum which we know is a small number of contiguous integers, we can probably use `Array<UUIDMap>` here instead and save all of the hashing / toStringing of enumerations.

Dexterp37: I've resorted to using an `Array`, as suggested by @mdboom

components/service/glean/src/main/java/mozilla/components/service/glean/storages/UuidsStorageEngine.kt
Dexterp37: Good point. I've moved this to `GenericScalarStorageEngine` and made it clearer.

components/service/glean/src/main/java/mozilla/components/service/glean/EventMetricType.kt
```diff
@@ -60,6 +60,11 @@ data class EventMetricType(
             return
         }
 
+        if (lifetime != Lifetime.Ping) {
+            logger.warn("$category.$name can only have a Ping lifetime")
+            return
+        }
```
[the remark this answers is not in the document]
Dexterp37: Good point, we should probably do that in both places!

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
```diff
@@ -12,8 +14,15 @@ import org.json.JSONArray
...
  * the Specific Events API and the ping assembling objects. No validation on the stored data
  * is performed at this point: validation must be performed by the Specific Events API.
+ *
+ * This class contains a reference to the Android application Context. While the IDE warns
+ * us that this could leak, the application context lives as long as the application and this
+ * object. For this reason, we should be safe to suppress the IDE warning.
  */
+@SuppressLint("StaticFieldLeak")
 internal object EventsStorageEngine : StorageEngine {
+    override lateinit var applicationContext: Context
```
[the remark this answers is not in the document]
Dexterp37: Compilation will fail if it's not overridden, as it's defined in the `StorageEngine` interface and interfaces [can't have backing fields](https://kotlinlang.org/docs/reference/interfaces.html#properties-in-interfaces)

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -0,0 +1,167 @@
...
+     * Get a snapshot of the stored data as a JSON object.
+     *
+     * @param storeName the name of the desired store
+     * @param clearStore whether or not to clearStore the requested store
+     *
+     * @return the [JSONObject] containing the recorded data.
+     */
+    override fun getSnapshotAsJSON(storeName: String, clearStore: Boolean): Any? {
+        return getSnapshot(storeName, clearStore)?.let { dataMap ->
+            return JSONObject(dataMap)
```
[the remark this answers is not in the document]
Dexterp37: Let me do this as part of this PR, it will make things less messy in the future!

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -0,0 +1,183 @@
...
+            // Split the stored name in 2: we expect it to be in the format
+            // store#metric.name
+            val parts = metricName.split('#', limit = 2)
+            if (parts[0].length <= 1) {
+                continue
+            }
+
+            val storeData = dataStores[Lifetime.User.ordinal].getOrPut(parts[0]) { mutableMapOf() }
+            // Only set the stored value if we're able to deserialize the persisted data.
+            deserializeSingleMetric(metricValue)?.let {
```
georgf: If `deserializeSingleMetric()` returns `null`, are we logging an error somewhere?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -0,0 +1,183 @@
...
+     * Deserialize the metrics with a lifetime = User that are on disk.
+     * This will be called the first time a metric is used or before a snapshot is
+     * taken.
+     *
+     * @return A [SharedPreferences] reference that will be used to inititialize [userLifetimeStorage]
+     */
+    @Suppress("TooGenericExceptionCaught")
+    open fun deserializeUserLifetime(): SharedPreferences {
+        val prefs =
+            applicationContext.getSharedPreferences(this.javaClass.simpleName, Context.MODE_PRIVATE)
```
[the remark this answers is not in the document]
Dexterp37: Sure, filed bug 1507081.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -0,0 +1,203 @@
...
+    @Synchronized
+    fun getSnapshot(storeName: String, clearStore: Boolean): GenericDataStorage<ScalarType>? {
+        val allLifetimes: GenericDataStorage<ScalarType> = mutableMapOf()
+
+        // Make sure data with "user" lifetime is loaded before getting the snapshot.
+        // We still need to catch exceptions here, as `getAll()` might throw.
+        @Suppress("TooGenericExceptionCaught")
+        try {
+            userLifetimeStorage.all
+        } catch (e: NullPointerException) {
```
badboy: Is `NullPointerException` the most generic one here? (I have no clue about Kotlin, so just trying to understand it)

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -0,0 +1,203 @@
...
+        stores: List<String>,
+        category: String,
+        name: String,
+        lifetime: Lifetime,
+        value: ScalarType
+    ) {
+        checkNotNull(applicationContext) { "No recording can take place without an application context" }
+
+        // Record a copy of the metric in all the needed stores.
+        @SuppressLint("CommitPrefEdits")
```
badboy: Why is suppressing the lint needed? Is it not smart enough to see that `apply` is called at the end if it is non-null?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
Dexterp37: There's the Java `RuntimeException`, which this one inherits from. However, I'm only interested in catching this one, since this is the only one that `all` can throw.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
Dexterp37: No, it's not :( The IDE will keep telling me `apply` is missing without that suppression.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/StorageEngineManager.kt
```diff
@@ -14,9 +15,17 @@ import org.json.JSONObject
 internal class StorageEngineManager(
     private val storageEngines: Map<String, StorageEngine> = mapOf(
         "events" to EventsStorageEngine,
-        "strings" to StringsStorageEngine
-    )
+        "strings" to StringsStorageEngine,
+        "uuids" to UuidsStorageEngine
+    ),
+    applicationContext: Context
```
pocmo: Maybe unrelated to this PR: `StorageEngineManager` seems to be responsible to keep the map of storage engines. Do the engines themselves need to be singletons then? Or can they get accessed through the storage engine manager? This would simplify this setup here too; where you need to pass the applicationContext to the singletons.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
pocmo: Having the `applicationContext` as lateinit here in the Singleton and then assigning it manually from the manager is a bit confusing to follow. See comment in the manager class.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -0,0 +1,203 @@
...
+    open fun deserializeUserLifetime(): SharedPreferences {
+        val prefs =
+            applicationContext.getSharedPreferences(this.javaClass.simpleName, Context.MODE_PRIVATE)
+
+        val metrics = try {
+            prefs.all.entries
+        } catch (e: NullPointerException) {
+            // If we fail to deserialize, we can log the problem but keep on going.
+            logger.error("Failed to deserialize metric with 'user' lifetime")
+            return prefs
```
pocmo: Oh wow, it's super annoying that `prefs.all` throws a NullPointerException 😞

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -0,0 +1,203 @@
...
+        }
+        userPrefs?.apply()
+    }
+
+    @VisibleForTesting
+    internal open fun clearAllStores() {
+        for (store in dataStores) {
+            store.clear()
+        }
+    }
```
pocmo: nit: Some of the functions here you can "shrink" by using some of the various list extension functions, e.g.:

```Kotlin
internal open fun clearAllStores() = dataStores.forEach { it.clear() }
```

components/service/glean/src/main/java/mozilla/components/service/glean/storages/StorageEngineManager.kt
Dexterp37: Good point. The downside of this would be that `StorageEngineManager` would then need to be a singleton (the *specific metric APIs* require either this or the engines to be one). Moreover, when calling the engine from the API (e.g. `UuidMetricType`), we'd need to manually upcast from the `StorageEngine` returned by the manager (each engine may implements custom methods not available on the interface, which are used by the APIs).

We might revisit this post MVP though.

pocmo: `StorageEngineManager` sounds like a Singleton tbh :)

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
Dexterp37: Yeah, it's a bit weird :\ I wish there was something like "getApplicationContext()" somewhere

components/service/glean/src/main/java/mozilla/components/service/glean/storages/StorageEngineManager.kt
Dexterp37: Fair enough! Let me file a bug to deal with that (we're tracking our current tech debt on bugzilla :) ).  I filed bug 1491345 for this!

PR #1310, opened 2018-11-08
Closes #1252: Webfont blocking is controlled by both policy and setting.

### LINKED ISSUE #1252: Webfont blocking is controlled via tracking protection policy and setting
I was looking at https://github.com/mozilla-mobile/firefox-tv/issues/1264 and noticed two things in our code base:

* (1) For "blocking" web fonts there are two ways: `Settings.webFontsEnabled` and `TrackingProtectionPolicy.WEBFONTS`.

* (2) In `browser-engine-system` webfont blocking is enabled by default (`UrlMatcher.blockWebFonts` is true. I didn't verify this on a device/website).

---

I think we should:

* Only use `Settings.webFontsEnabled`. There's actually no webfont tracker category. It's our best effort to emulate disabling web fonts by not loading resources that look like a web font. This should be `browser-engine-system`'s implementation of `Settings.webFontsEnabled`.

* We shouldn't block anything by default.

csadilek:
Small change. We're basically removing the tracking protection policy, and making sure we're not blocking webfonts by default. The rest of the functionality can stay the same: If `engine.settings.webFontsEnabled = false`, we're blocking webfonts as we did before. This change only affects `SystemEngine`, which used both the policy and the setting.

### DIFF
--- a/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/matcher/UrlMatcher.kt
+++ b/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/matcher/UrlMatcher.kt
@@ -23,7 +23,6 @@ class UrlMatcher {
     private val whiteList: WhiteList?
     private val previouslyMatched = HashSet<String>()
     private val previouslyUnmatched = HashSet<String>()
-    private var blockWebfonts = true
 
     constructor(patterns: Array<String>) {
         categories = HashMap()
@@ -66,11 +65,6 @@ class UrlMatcher {
     }
 
     internal fun setCategoryEnabled(category: String, enabled: Boolean) {
-        if (WEBFONTS == category) {
-            blockWebfonts = enabled
-            return
-        }
-
         if (enabled) {
             if (enabledCategories.contains(category)) {
                 return
@@ -114,10 +108,6 @@ class UrlMatcher {
         val resourceHost = resourceURI.host
         val pageHost = pageURI.host
 
-        if (blockWebfonts && isWebFont(resourceURI)) {
-            return true
-        }
-
         if (previouslyUnmatched.contains(resourceURLString)) {
             return false
         }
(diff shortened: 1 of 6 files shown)


components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/matcher/UrlMatcher.kt
```diff
@@ -23,7 +23,7 @@ class UrlMatcher {
     private val whiteList: WhiteList?
     private val previouslyMatched = HashSet<String>()
     private val previouslyUnmatched = HashSet<String>()
-    private var blockWebfonts = true
+    private var blockWebfonts = false
```
pocmo: I don't see how `DefaultSettings(webFontsEnabled = ...)` will lead to this value being true/false? Isn't that part missing here?

csadilek: @pocmo we didn't need that as we've never matched against that URL matching category WEBFONT, but checked for webfonts explicitly:

https://github.com/mozilla-mobile/android-components/blob/master/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt#L162

Alternatively, we could also add a parameter to `matches` to only check certain (URL matching) categories, but it made sense to me to keep this separate, now that it's not a tracking protection category.

csadilek: Since it was confusing to get rid of the TrackingProtectionPolicy.WEBFONTS but keep the UrlMatcherCategory WEBFONTS, I got rid of both now.

I think this is easier to read. We can either configure TrackingProtectionPolicies which directly map to UrlMatcher categories, our we can enable/disable webfonts using the engine setting (same logic as on master except that we no longer block web fonts by default/mistake).

PR #1318, opened 2018-11-09
[glean] Add General API...

mdboom:
This adds the top-level API for glean telemetry -- basically a singleton to handle all of the global functionality.

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
@@ -1,0 +1,141 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.glean
+
+import android.content.Context
+import android.support.annotation.VisibleForTesting
+
+import java.util.UUID
+
+import mozilla.components.service.glean.config.Configuration
+import mozilla.components.service.glean.net.HttpPingUploader
+import mozilla.components.service.glean.ping.PingMaker
+import mozilla.components.service.glean.storages.ExperimentsStorageEngine
+import mozilla.components.service.glean.storages.StorageEngineManager
+import mozilla.components.support.base.log.logger.Logger
+
+object Glean {
+    /**
+    * Enumeration of different metric lifetimes.
+    */
+    enum class PingEvent {
+        /**
+        * When the application goes into the background
+        */
+        Background,
+        /**
+        * A periodic event to send the default ping
+        */
+        Default
+    }
+
+    private val logger = Logger("glean/Glean")
+
+    internal const val SCHEMA_VERSION = 1
+
+    // Include our singletons of StorageEngineManager and PingMaker
+    private lateinit var storageEngineManager: StorageEngineManager
+    private lateinit var pingMaker: PingMaker
+    private lateinit var configuration: Configuration
+    // `internal` so this can be modified for testing
+    internal lateinit var httpPingUploader: HttpPingUploader
+
+    private var initialized = false
+    private var metricsEnabled = true
+
+    /**
+     * Initialize glean.
+     */
+    fun initialize(applicationContext: Context, configuration: Configuration) {
+        storageEngineManager = StorageEngineManager(applicationContext = applicationContext)
+        pingMaker = PingMaker(storageEngineManager)
+        this.configuration = configuration
+        httpPingUploader = HttpPingUploader(configuration)
+        initialized = true
+    }
+
+    /**
+     * Enable or disable metric collection.
+     *
+     * Metric collection is emabled by default.
+     *
+     * When disabled, metrics aren't recorded at all.
+     *
+     * @param enabled When true, enable metric collection.
+     */
+    fun setMetricsEnabled(enabled: Boolean) {
+        logger.info("Metrics enabled: $enabled")
+        metricsEnabled = enabled
+    }
+
+    /**
+     * Get the status of metrics enabled.
+     */
+    fun getMetricsEnabled(): Boolean {
+        return metricsEnabled
+    }
+
+    /**
+     * Indicate that an experiment is running. Glean will then add an
+     * experiment annotation to the environment which is sent with pings. This
+     * information is not persisted between runs.
+     *
+     * @param experimentId The id of the active experiment (maximum
+     *     30 bytes)
+     * @param branch The experiment branch (maximum 30 bytes)
+     * @param extra Optional metadata to output with the ping
+     */
+    fun setExperimentActive(
+        experimentId: String,
+        branch: String,
+        extra: Map<String, String>? = null
+    ) {
+        ExperimentsStorageEngine.setExperimentActive(experimentId, branch, extra)
+    }
+
+    /**
+     * Indicate that an experiment is no longer running.
+     *
+     * @param experimentId The id of the experiment to deactivate.
+     */
+    fun setExperimentInactive(experimentId: String) {
+        ExperimentsStorageEngine.setExperimentInactive(experimentId)
+    }
+
+    @VisibleForTesting
+    internal fun clearExperiments() {
+        ExperimentsStorageEngine.clearAllStores()
+    }
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun makePath(docType: String, uuid: UUID): String {
+        return "/submit/glean/$docType/$SCHEMA_VERSION/$uuid"
+    }
+
+    private fun sendPing(store: String, docType: String) {
+        val pingContent = pingMaker.collect(store)
+        val uuid = UUID.randomUUID()
+        val path = makePath(docType, uuid)
+        this.httpPingUploader.upload(pingContent, path)
+    }
+
+    /**
+     * Handle an event and send the appropriate pings.
+     *
+     * Valid events are:
+     *
+     *   - Background: When the application goes to the background.
+     *       This triggers sending of the baseline ping.
+     *   - Default: Event that triggers the default pings.
+     *
+     * @param pingEvent The type of the event.
+     */
+    fun handleEvent(pingEvent: PingEvent) {
+        when (pingEvent) {
+            PingEvent.Background -> sendPing("baseline", "baseline")
+            PingEvent.Default -> sendPing("metrics", "metrics")
+        }
+    }
+}
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/storages/ExperimentsStorageEngine.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/storages/ExperimentsStorageEngine.kt
@@ -1,0 +1,144 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.glean.storages
+
+import android.annotation.SuppressLint
+import mozilla.components.support.base.log.logger.Logger
+import mozilla.components.support.ktx.android.org.json.toJSON
+
+import android.content.Context
+import android.support.annotation.VisibleForTesting
+import org.json.JSONObject
+
+/**
+ * This singleton handles the in-memory storage logic for keeping track of
+ * the active experiments. It is meant to be used by through the methods in
+ * the Glean class (General API). No validation on the stored data is
+ * performed at this point: validation must be performed by the General API.
+ */
+@SuppressLint("StaticFieldLeak")
+internal object ExperimentsStorageEngine : StorageEngine {
+    override lateinit var applicationContext: Context
+
+    private val logger = Logger("glean/ExperimentsStorageEngine")
+
+    // Maximum length of the experiment and branch names, in UTF8 byte
+    // sequence length
+    private const val MAX_ID_LENGTH = 30
+
+    data class RecordedExperimentData(
+        val branch: String,
+        val extra: Map<String, String>? = null
+    )
+
+    private val experiments: MutableMap<String, RecordedExperimentData> = mutableMapOf()
+
+    /**
+     * Record an experiment as active.
+     *
+     * @param experimentId the id of the experiment
+     * @param branch the active branch of the experiment
+     * @param extra an optional, user defined String to String map used to provide richer experiment
+     *              context if needed
+     */
+    fun setExperimentActive(
+        experimentId: String,
+        branch: String,
+        extra: Map<String, String>? = null
+    ) {
+        val truncatedExperimentId = experimentId.let {
+            if (it.length > MAX_ID_LENGTH) {
+                logger.warn("experimentId ${it.length} > $MAX_ID_LENGTH")
+                return@let it.substring(0, MAX_ID_LENGTH)
+            }
+            it
+        }
+
+        val truncatedBranch = branch.let {
+            if (it.length > MAX_ID_LENGTH) {
+                logger.warn("branch ${it.length} > $MAX_ID_LENGTH")
+                return@let it.substring(0, MAX_ID_LENGTH)
+            }
+            it
+        }
+
+        val experimentData = RecordedExperimentData(truncatedBranch, extra)
+        experiments.put(truncatedExperimentId, experimentData)
+    }
+
+    /**
+     * Set an active experiment as inactive.
+     *
+     * Warns to the logger (but does not raise) if the experiment is not already active.
+     *
+     * @param experimentId the id of the experiment to set as inactive
+     */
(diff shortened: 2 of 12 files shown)


components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
```diff
@@ -0,0 +1,163 @@
...
+    fun `disabling metrics should record nothing`() {
+        val stringMetric = StringMetricType(
+                disabled = false,
+                category = "telemetry",
+                lifetime = Lifetime.Application,
+                name = "string_metric",
+                sendInPings = listOf("store1")
+        )
+        Glean.setMetricsEnabled(false)
+        try {
```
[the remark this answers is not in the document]
mdboom: Because we need to reset the global state that changes here (`setMetricsEnabled`).  Otherwise other unrelated tests start failing randomly (depending on order of test run).

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -0,0 +1,103 @@
...
+     *
+     *   - background: When the application goes to the background.
+     *       This triggers sending of the default ping.
+     *   - default: Event that triggers the default pings.
+     *
+     * @param name The name of the event.
+     */
+    fun handleEvent(name: String) {
+        if (name == "background") {
+            sendPing("baseline", "/baseline")
```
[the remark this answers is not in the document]
mdboom: I've updated this to store the UUID in `ping_info/ping_id` for now, and updated this PR to WIP until to wait to hear from @mreid-moz on how best to handle that.

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
```diff
@@ -0,0 +1,163 @@
...
+                    "Metrics should not be recorded if glean is disabled",
+                    StringsStorageEngine.getSnapshot(storeName = "store1", clearStore = false)
+            )
+        } finally {
+            Glean.setMetricsEnabled(true)
+        }
+    }
+
+    @Test
+    fun `disabling event metrics should record only when enabled`() {
```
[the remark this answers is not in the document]
mdboom: Done.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/ExperimentsStorageEngine.kt
```diff
@@ -0,0 +1,137 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
Dexterp37: I think we're missing test coverage for this engine :)

components/service/glean/src/main/java/mozilla/components/service/glean/storages/ExperimentsStorageEngine.kt
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * Record an experiment as active.
+     *
+     * @param experimentId the id of the experiment
+     * @param branch the active branch of the experiment
+     * @param extra an optional, user defined String to String map used to provide richer experiment
+     *              context if needed
+     */
+    fun setExperimentActive(
```
Dexterp37: We should have test coverage for this function, testing truncation and edge cases, such us what happens if we try to record the same experiment id twice.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/ExperimentsStorageEngine.kt
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    /**
+     * Set an active experiment as inactive.
+     *
+     * Warns to the logger (but does not raise) if the experiment is not already active.
+     *
+     * @param experimentId the id of the experiment to set as inactive
+     */
+    fun setExperimentInactive(experimentId: String) {
```
Dexterp37: We should have test coverage for this and also check on edge cases, such as removing a non-existing experiment.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/ExperimentsStorageEngine.kt
```diff
@@ -0,0 +1,137 @@
...
+     * Retrieves the data about active experiments
+     *
+     * @return the map of experiment ids to [RecordedExperimentData]
+     */
+    @Synchronized
+    fun getSnapshot(): Map<String, RecordedExperimentData> {
+        return experiments
+    }
+
+    private fun convertExtra(extra: Map<String, String>): JSONObject {
```
Dexterp37: Let's have some documentation for this function. Maybe we can also add direct test coverage for it, using `@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)` (if you see fit).

components/service/glean/src/main/java/mozilla/components/service/glean/storages/ExperimentsStorageEngine.kt
```diff
@@ -0,0 +1,137 @@
...
+     *
+     * @param storeName the name of the desired store.  Since experiments are
+     *   included in all stores, this parameter is effectively ignored, but
+     *   must be provided to comply with the StorageEngine interface.
+     * @param clearStore whether or not to clearStore the requested store.
+     *   This parameter is ignored.
+     *
+     * @return the JSONArray containing information about the active experiments
+     */
+    override fun getSnapshotAsJSON(storeName: String, clearStore: Boolean): Any? {
```
Dexterp37: Let's also have test coverage for this :) You can explicitly mark `storeName` as unused by prefixing it with ` @Suppress("UNUSED_PARAMETER")`, so the signature would become 

`override fun getSnapshotAsJSON(@Suppress("UNUSED_PARAMETER") storeName: String, ...`

components/service/glean/src/main/java/mozilla/components/service/glean/storages/ExperimentsStorageEngine.kt
```diff
@@ -0,0 +1,137 @@
...
+    /**
+     * Get a snapshot of the stored data as a JSON object.
+     *
+     * @param storeName the name of the desired store.  Since experiments are
+     *   included in all stores, this parameter is effectively ignored, but
+     *   must be provided to comply with the StorageEngine interface.
+     * @param clearStore whether or not to clearStore the requested store.
+     *   This parameter is ignored.
+     *
+     * @return the JSONArray containing information about the active experiments
```
Dexterp37: Isn't this returning a `JSONObject` rather than a `JSONArray`? For such nested types, it would be better to have an example in the docs.

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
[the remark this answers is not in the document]
mdboom: Nothing is throwing on purpose.  What I'm indicating is that no matter what happens in this test function, the global state needs to be reset at the end (in the `finally` block).  The behavior that this prevents is a failure in this test leading to a failure in other tests simply because they come after this one.

If there is a way to set a `@Before` on every test in the test suite, that would be a viable alternative.  But I think `@Before` just affects tests in the same fixture or file.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -0,0 +1,109 @@
...
+     * Indicate that an experiment is no longer running.
+     *
+     * @param experimentId The id of the experiment to deactivate.
+     */
+    fun setExperimentInactive(experimentId: String) {
+        ExperimentsStorageEngine.setExperimentInactive(experimentId)
+    }
+
+    private fun sendPing(store: String, docType: String) {
+        val uuid = UUID.randomUUID()
```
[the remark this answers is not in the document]
mdboom: And I guess that implies `pingMaker.collect` should return a pair of `content, documentId`.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -0,0 +1,109 @@
...
+import mozilla.components.service.glean.storages.StorageEngineManager
+
+object Glean {
+    internal const val SCHEMA_VERSION = 1
+
+    // Include our singletons of StorageEngineManager and PingMaker
+    private val storageEngineManager = StorageEngineManager()
+    private val pingMaker = PingMaker(storageEngineManager)
+    private lateinit var configuration: Configuration
+    internal lateinit var httpPingUploader: HttpPingUploader
```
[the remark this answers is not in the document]
mdboom: Because it needs to be modified for testing purposes.  AFAICT, that's the only way, short of adding a `setter` to make that work.

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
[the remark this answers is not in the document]
mdboom: It does like like, however, that we could use `@After`, which might be cleaner / more obvious than `try/finally`.  I'll play with that.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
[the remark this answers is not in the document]
Dexterp37: Yes, that would make things simpler.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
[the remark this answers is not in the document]
Dexterp37: Ah, good point! Thanks for clarifying!

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
```diff
@@ -0,0 +1,172 @@
...
+            eventMetric.record("buttonA", "event3")
+            val snapshot3 = EventsStorageEngine.getSnapshot(storeName = "store1", clearStore = false)
+            assertEquals(2, snapshot3!!.size)
+        } finally {
+            Glean.setMetricsEnabled(true)
+        }
+    }
+
+    @Test
+    fun `test sending of default pings`() {
```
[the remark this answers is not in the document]
mdboom: This test is a lot smaller now because it's not sending the event ping.  But I think it's still worth testing the path here (as well as more directly) because it's the only place where all the pieces involved that get integrated and you can confirm that the UUID in the path matches the UUID in the ping payload.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/ExperimentsStorageEngine.kt
mdboom: Oops.  Done.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/ExperimentsStorageEngine.kt
mdboom: I've put most of the testing for this in `ExperimentsStorageEngineTests` since that's where the real work is done, and just made sure that these delegation functions are covered by tests.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/ExperimentsStorageEngine.kt
mdboom: I added documentation and improved the name a bit.  It's really just a helper function for `getSnapshotAsJSON`, where `detekt` was complaining about excessive nesting, so I think it's fine to just test it as an implementation detail of `getSnapshotAsJSON`.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/ExperimentsStorageEngine.kt
mdboom: Yes, indeed.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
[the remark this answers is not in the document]
mreid-moz: It's needed in the URL path for skipping duplicate submissions. It's not required inside the ping itself (for example, the [core ping](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/dev/schemas/telemetry/core/core.9.schema.json) doesn't include it inside the document). If you can't think of a reason why it would be useful on the client-side, I'd lean towards leaving it out.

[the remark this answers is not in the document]
mdboom: Thanks.  I'll leave it out of the ping content itself then.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/ExperimentsStorageEngine.kt
```diff
@@ -0,0 +1,150 @@
...
+     * @param clearStore whether or not to clearStore the requested store.
+     *   This parameter is ignored.
+     *
+     * @return the JSONObject containing information about the active experiments
+     */
+    override fun getSnapshotAsJSON(
+        @Suppress("UNUSED_PARAMETER") storeName: String,
+        @Suppress("UNUSED_PARAMETER") clearStore: Boolean
+    ): Any? {
+        return getSnapshot().let { pingExperiments ->
```
pocmo: Android Studio complains that this outer `return` is unreachable code and that seems to be correct since you already return from inside the block in all cases. Since `getSnapshot()` never returns `null` the block may not be needed:

```Kotlin
    override fun getSnapshotAsJSON(
        @Suppress("UNUSED_PARAMETER") storeName: String,
        @Suppress("UNUSED_PARAMETER") clearStore: Boolean
    ): Any? {
        val pingExperiments = getSnapshot()

        if (pingExperiments.count() == 0) {
            return null
        }

      ...
```

components/service/glean/src/main/java/mozilla/components/service/glean/storages/ExperimentsStorageEngine.kt
```diff
@@ -0,0 +1,150 @@
...
+                    experimentData.put(getExtraAsJSON(value.extra))
+                }
+                experimentsMap.put(key, experimentData)
+            }
+            return experimentsMap
+        }
+    }
+
+    override val sendAsTopLevelField: Boolean
+        get() = true
```
pocmo: nit: If this is always true you can also create a backing field for it and set it once:

```Kotlin
override val sendAsTopLevelField: Boolean = true
```

components/service/glean/src/main/java/mozilla/components/service/glean/storages/StorageEngine.kt
```diff
@@ -20,4 +20,11 @@ internal interface StorageEngine {
      *         ancestor is [Object], so we need to return [Any].
      */
     fun getSnapshotAsJSON(storeName: String, clearStore: Boolean): Any?
+
+    /**
+     * Indicate whether this storage engine is sent at the top level of the ping
+     * (rather than in the metrics section).
+     */
+    val sendAsTopLevelField: Boolean
+        get() = false
```
pocmo: See above

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -0,0 +1,125 @@
...
+     *
+     * Valid events are:
+     *
+     *   - background: When the application goes to the background.
+     *       This triggers sending of the default ping.
+     *   - default: Event that triggers the default pings.
+     *
+     * @param name The name of the event.
+     */
+    fun handleEvent(name: String) {
```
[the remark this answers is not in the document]
mdboom: It definitely is limited to a small enum for now, but the design document includes a plan for "custom pings", post MVP, which the application would define in `pings.yaml` file.  That's the reason for using something more open-ended here.  But I actually think we should probably handle that as an (enum == CUSTOM, string) pair, so I'll update this to use an enum for now.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/StorageEngine.kt
mdboom: Unfortunately with
```
    val sendAsTopLevelField: Boolean = false
```
it says `Property initializers are not allowed in interfaces`.

pocmo: Ops, yes, makes sense. This should only work in `ExperimentsStorageEngine`.

mdboom:
@pocmo: I've addressed all of your comments and rebased on master.  It looks like the failing tests maybe are unrelated, though.  Thoughts?

Dexterp37:
Hey @mdboom , looks like in addition to the unrelated problems, this is also failing for legit reasons:

> e: /build/android-components/components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt: (106, 9): Unresolved reference: ExperimentsStorageEngine

mdboom:
Great.  Rebased and squashed!

pocmo:
🚀

PR #1321, opened 2018-11-10
Closes #1319: Conceptualize and featurize toolbar autocompletion

### LINKED ISSUE #1319: Conceptualize/featurize domain autocompletion
For our toolbar we'd like to have multiple providers of auto-complete data. Currently `HistoryStorage` interface has a `getDomainSuggestions` method (landed in #1280), and we also have `browser-domains/DomainAutoCompleteProvider` which encapsulates two data sources - "shipped" and "custom" sets of domain, with logic to prioritize these data sources (use "custom" if available/configured, and "shipped" as a fallback).

Currently, autocomplete wiring and prioritization of autocomplete sources needs to happen at the application level.

If we describe a source of domains suggestions as a separate concept, we can implement source prioritization and autocompletion and source privatization logic within the `feature-toolbar` layer, which will operate over a set of DomainProviders. 

A `concept-domains` could describe a `DomainProvider::getDomainSuggestion(query: String): Result`, and we'll have four concrete implementations to begin with:
- browser-storage-sync
- browser-storage-memory
- custom domains
- shipped domains (these last two are the same except for the source on disk, so they'll share a base class)

The `feature-toolbar/ToolbarAutocompleteFeature` can have a single `addProvider` method which follows a builder pattern, with an order in which providers are added guiding internal source prioritization logic. Autocompletion glue can then live within the feature, and query configured domain providers.

These changes will be backwards compatible, as consumers don't have to use the new functionality. We can mark existing `DomainAutoCompleteProvider` as deprecated and remove it once we're confident existing users moved off of it. (I think there's just a single consumer currently). I don't think anyone's using `HistoryStorage.getDomainSuggestion` yet, so we can just move that method.

Questions/concerns:
- `browser-domains` has a concept of an "autocomplete source", which is used to distinguish between "shipped" and "custom" autocompletion results. The above plan suggests a potentially arbitrary set of autocomplete sources, without a globally defined set of all types of sources. As far as I can tell, we don't use this identifier for anything. It's plumbed and ignored in `ui-autocomplete/InlineAutocompleteEditText`.
  - Since it's the application that will be responsible for adding individual domain providers to the autocomplete feature, it can also name them.
  - `toolbarAutocompleteFeature.addProvider("coolProvider", provider)`.
  - If we'd like custom treatment of different autocomplete sources at the UI level, an application will be able to configure that:
   - `toolbarAutocompleteUiView.setAutocompleteSourceRules("customProvider", CustomProviderRules)`
- in the future, we're likely grow some kind of "blocked domains" functionality (e.g. #313). This might just plug in at the `feature` level (e.g. a "blocked list" to negate results from other sources). This needs more thought.

cc @pocmo @csadilek

grigoryk:
First pass at this. Take a look at the sample browser change - in the end, it's trivial to wire in toolbar autocompletion with browsing history as a autocompletion provider as the primary source, and define backup data sources.


grigoryk:
The "WIP" bit is that I need to update any tests this broke, and likely write new ones. Other than that, it works and feels pretty well.

components/browser/domains/src/main/java/mozilla/components/browser/domains/Providers.kt
```diff
@@ -0,0 +1,114 @@
...
+        scope.launch {
+            domains = async { CustomDomains.load(context).into() }.await()
+        }
+    }
+}
+
+/**
+ * Provides common autocomplete functionality powered by domain lists.
+ */
+abstract class Provider : Autocomplete {
```
csadilek: I'd find `AutocompleteProvider` and `DomainAutocompleteProvider` (for the common subtype) easier to read.

components/browser/domains/src/main/java/mozilla/components/browser/domains/Providers.kt
```diff
@@ -0,0 +1,114 @@
...
+
+                    return Domain(protocol, hasWww, host)
+                } ?: throw IllegalStateException()
+            }
+        }
+    }
+
+    abstract fun initialize(context: Context)
+
+    // We compute these on worker thread; make sure results are immediately visible on the UI thread.
```
csadilek: Looks like a left-over comment.

components/concept/toolbar/src/main/java/mozilla/components/concept/toolbar/Autocomplete.kt
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * Describes an object to which a [Result] may be applied. Usually a text view.
+     */
+    interface ResultAcceptor {
+        fun applyAutocompleteResult(result: Result)
+    }
+
+    /**
+     * Describes an autocompletion result.
+     * @param text Result of autocompletion, text to be displayed.
```
csadilek: nit: these are `@property`

grigoryk:
@csadilek take another look when you get a chance: updated tests, renamed/cleaned up stuff, kept 'source' and 'url' in the result to make focus update straightforward.

grigoryk:
TC tasks failed because of codecov service outage: https://github.com/mozilla-mobile/android-components/issues/1364

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarAutocompleteFeature.kt
```diff
@@ -0,0 +1,28 @@
...
+import mozilla.components.concept.toolbar.AutocompleteProvider
+import mozilla.components.concept.toolbar.Toolbar
+
+/**
+ * Feature implementation for connecting a toolbar with a list of autocomplete providers.
+ */
+class ToolbarAutocompleteFeature(val toolbar: Toolbar, private val providers: List<AutocompleteProvider>) {
+    init {
+        toolbar.setAutocompleteListener { value, acceptor ->
+            acceptor?.let {
```
pocmo: I wonder if we should ever end up here with a `null` acceptor. This seems to be unneeded work. I remember that this may come from InlineAutocomplete; I wonder if we should stop that event ("autocomplete for nobody") as early as possible. Doesn't need to be here though.

grigoryk: Yeah, that crept in from some existing code, and I didn't clean it up.

grigoryk:
We've discussed this offline, and option (B) as proposed by @pocmo is worth the effort to re-align this patch series.

grigoryk:
@pocmo take another look, this is following your suggested pattern (B). A bit of boilerplate here and there, and I've kept part of the `concept-toolbar` changes (the "acceptor" side of things) to make interacting with the ui components simpler.

components/concept/toolbar/src/main/java/mozilla/components/concept/toolbar/Toolbar.kt
```diff
@@ -58,6 +58,14 @@ interface Toolbar {
      */
     fun setOnUrlCommitListener(listener: (String) -> Unit)
 
+    /**
+     * Registers the given function to be invoked when users changes text in the toolbar.
+     *
+     * @param filter A functions which is will perform autocompletion and send results to
+     * [AutocompleteProvider.ResultAcceptor].
+     */
+    fun setAutocompleteListener(filter: (String, AutocompleteAcceptor?) -> Unit)
```
csadilek: We need to update the parameter name now.

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarAutocompleteFeature.kt
```diff
@@ -0,0 +1,69 @@
...
+    fun addDomainProvider(provider: DomainAutocompleteProvider) {
+        domainProviders.add(provider)
+    }
+
+    // The rest is a little bit of boilerplate to tie things together.
+    // We terminate our dependency chains at the 'feature' level. This is an alternative to creating
+    // an "AutocompleteProvider" concept. This pattern allows a storage component, for example, to
+    // be standalone and not depend on something like 'concept-domains'.
+    // This pattern also helps prevent circular dependency situations from arising over time.
+
```
csadilek: I get the reason we're doing this, but it is a bit unfortunate :). Wondering if we should create a single AutocompleteResult (instead of 3-4 distinct types) in a root module e.g. support-base? I think this would simplify this quite a bit!

components/ui/autocomplete/build.gradle
```diff
@@ -27,6 +27,7 @@ dependencies {
     implementation Dependencies.kotlin_stdlib
 
     implementation Dependencies.support_appcompat
+    implementation project(':concept-toolbar')
```
pocmo: I think we should avoid having this dependency too:

* `ui-*` components are dumb view components without any knowledge about the context they are used in. This ties this to the toolbar.

* The autocomplete component can be used for autocompleting all kinds of things. You could throw it into some app and autocomplete phone numbers or user names. However the concept leaks the idea of "URLs" into it now.

`BrowserToolbar` accepts a an `AutocompleteAcceptor`and in `setAutocompleteListener()` it could create a wrapper/delegate that implements `AutocompleteAcceptor` and delegates to `InlineAutocompleteEditText`. What do you think?

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarAutocompleteFeature.kt
grigoryk: @csadilek One of the reasons *for* doing this boilerplate that's not currently obvious is that it enables individual providers to define a "result" as they please, and expose a greater set of functionality as they see fit. The `feature` component is then responsible for massaging different results together. If we have a common definition in `support-*`, we introduce an implicit dependency between all of our providers.

The other side of having this implicit dependency is that it has a benefit of imposing a "growing in lockstep" limitation on the different provider implementations, which could be quite useful.

pocmo: Boilerplate is the price for abstractions and a layered architecture. This seems fine to me here since Features are translating between two worlds. Anyhow, boilerplate is something we can always optimize later. :)

csadilek: Yeah, if the results were different I wouldn't have noticed. It's just that it's the same type with 4 copies, which tipped me off. So, this boilerplate is caused by a lack of abstraction (over the result) not because of it. :)

Anyways, like I wrote, I don't feel strongly. Just a comment...

components/ui/autocomplete/build.gradle
grigoryk: I think that's OK. This creates an annoying delegate chain, and I really, really, really dislike delegate chains (you can blame Fennec sync for that...). But it doesn't feel very out of place here, as long as we make sure this doesn't set a wider precedent for these damn things ;)

grigoryk:
Failed an unrelated test due to an OOM, see https://github.com/mozilla-mobile/android-components/issues/1467.

pocmo:
I tried to fix the conflict and land this, but I'm seeing some other unresolved reference issues after rebasing. :(

Issue #1327, opened 2018-11-12
Create generic test suite for concept-fetch implementations


pocmo:
Breaking down: #1012

PR #1330, opened 2018-11-12
Issue #1327: (tooling-fetch-tests): A generic test suite for components that implement concept-fetch..

pocmo:


components/tooling/fetch-tests/src/main/java/mozilla/components/tooling/fetch/tests/FetchTestCases.kt
```diff
@@ -0,0 +1,416 @@
...
+import java.util.concurrent.TimeUnit
+
+/**
+ * Generic test cases for concept-fetch implementations.
+ *
+ * We expect any implementation of concept-fetch to pass all test cases here.
+ */
+@Suppress("IllegalIdentifier", "FunctionName", "unused")
+abstract class FetchTestCases {
+    abstract fun createClient(): Client
```
Amejia481: 🤔  Could this method introduce flakiness ? If the implementer always returns the same instance, this could be share between tests. Should we document this function about or change the method to createNewClient()?

components/tooling/fetch-tests/test/src/main/java/mozilla/components/support/test/Expect.kt
```diff
@@ -0,0 +1,22 @@
...
+
+package mozilla.components.support.test
+
+import org.junit.Assert.fail
+import kotlin.reflect.KClass
+
+/**
+ * Expect [block] to throw an exception. Otherwise fail the test (junit).
+ */
+inline fun <reified T : Throwable> expectException(clazz: KClass<T>, block: () -> Unit) {
```
Amejia481: 🤔  I'm not quite sure why we're copying some classes from  `support-test` instead of bring them as a dependency?

components/tooling/fetch-tests/test/README.md
```diff
@@ -0,0 +1,19 @@
...
+A collection of helpers for testing components.
+
+## Usage
+
+### Setting up the dependency
+
+Use Gradle to download the library from [maven.mozilla.org](https://maven.mozilla.org/) ([Setup repository](../../../README.md#maven-repository)):
+
+```Groovy
+implementation "org.mozilla.components:support-test:{latest-version}"
```
Amejia481: This should be `fetch-tests`

components/tooling/fetch-tests/test/src/main/java/mozilla/components/support/test/fetch/FetchTestCases.kt
```diff
@@ -0,0 +1,276 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.support.test.fetch
+
+/**
+ * Generic test cases for concept-fetch implementations.
+ *
+ * We expect any implementation of concept-fetch to pass all test cases here.
+ */
+/*
```
Amejia481: 🤔  I not sure why this class is completely commented?

components/tooling/fetch-tests/test/src/main/java/mozilla/components/support/test/Expect.kt
pocmo: Ops, that was a copy error. I started implementing this in `support-test` and then moved it over to its own component later. :)

Amejia481: 👍

pocmo:
@Amejia481 Sorry for all this nonsense in this PR. Now it should be up to date and only contain the things it should.

Amejia481:
No worries! :)

Issue #1325, opened 2018-11-12
Create HttpUrlConnection implementation of concept-fetch


pocmo:
Breaking down: #1012

PR #1331, opened 2018-11-12
 Issue #1325: (lib-fetch-httpurlconnection): A concept-fetch implementation using HttpURLConnection.

pocmo:
~~Depends on PR #1330.~~

~~**Let's review and land PR #1330 before this one (after rebasing this PR will just have one commit)**.~~

Ready for review!

### DIFF
--- a/components/lib/fetch-httpurlconnection/src/main/java/mozilla/components/lib/fetch/httpurlconnection/HttpURLConnectionClient.kt
+++ b/components/lib/fetch-httpurlconnection/src/main/java/mozilla/components/lib/fetch/httpurlconnection/HttpURLConnectionClient.kt
@@ -1,0 +1,152 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.lib.fetch.httpurlconnection
+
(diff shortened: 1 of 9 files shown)


pocmo:
It looks like one of the timeout tests is running indefinitely on taskcluster :)

components/lib/fetch-httpurlconnection/src/main/java/mozilla/components/lib/fetch/httpurlconnection/HttpURLConnectionClient.kt
```diff
@@ -0,0 +1,154 @@
...
+        createBody(this)
+    )
+}
+
+private fun translateHeaders(connection: HttpURLConnection): Headers {
+    val headers = MutableHeaders()
+
+    var index = 0
+
+    while (connection.getHeaderField(index) != null) {
```
csadilek: Would it be easier to use `connection.headerFields` and iterate over it, rather than using the index?

pocmo:
Addressed the review comments. Still need to figure out why one of the tests is hanging. It doesn't locally and the same test passes for the OkHttp implementation.

pocmo:
I had to remove the connect timeout test. It was behaving as I expected (it looks like my assumption that this scenario would trigger a connect timeout is wrong) and it was super flaky even locally.

Only an unrelated component failed (jcenter yay) so I'm going ahead and merge.

PR #1352, opened 2018-11-13
Added debounce() method and tests

colintheshots:


colintheshots:
Huh? Tests failed? I ran it thousands of times to verify the timings weren't an issue. I'll look into it.

PR #1355, opened 2018-11-13
Split application services gradle projects versions

eoger:
This allows for more flexibility on the `a-s` side.

cc @thomcc @ncalexan

### DIFF
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -25,7 +25,9 @@ private object Versions {
 
     const val sentry = "1.7.10"
 
-    const val mozilla_app_services = "0.10.0"
+    const val mozilla_fxa = "0.10.0"
+    const val mozilla_sync_logins = "0.10.0"
+    const val mozilla_places = "0.10.0"
     const val servo = "0.0.1.20181017.aa95911"
 }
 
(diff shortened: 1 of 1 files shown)


grigoryk:
Won't this dramatically increase amount of weirdness/bugs that will be possible in theory?

ncalexan:
> Won't this dramatically increase amount of weirdness/bugs that will be possible in theory?

Possibly.  We expect to use this for a while and manage the "interaction windows" manually.  Fine by me if it unblocks us for the immediate time.

csadilek:
I think using a single app_services version is preferable, but right now development productivity matters most. So, if this is better for you guys, definitely r+.

@vladikoff r?

vladikoff:
> I think using a single app_services version is preferable, but right now development productivity matters most. So, if this is better for you guys, definitely r+.

> @vladikoff r?

Yea correct, this is dev only and if there is a PR that changes one of the versions, but not the rest it might lead to weird artifacts. Previously I would just manually do this change while I was working on component updates. This should be fine for now

eoger:
Can we also make sure the right people are being asked when bumping the 3 versions in the future? For example right now I'd like `mozilla_fxa` to stay 0.10.0 until https://github.com/mozilla/application-services/issues/303 is done.

pocmo:
> Can we also make sure the right people are being asked when bumping the 3 versions in the future? 

Can we add this (and which people) as a comment in the Dependencies.kt file right next to those numbers? Imho that's the only place where we will see this every time we are making or reviewing changes.

PR #1363, opened 2018-11-14
[glean] Create a module to check if that was the first run for glean

Dexterp37:
This introduces a module that checks if a file exists: if that's the case, then we're not in a first run. If it's a first run, after the check, the first run file is created in the application's data dir.

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/firstrun/FileFirstRunDetector.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/firstrun/FileFirstRunDetector.kt
@@ -1,0 +1,81 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.glean.firstrun
+
+import android.support.annotation.VisibleForTesting
+import mozilla.components.support.base.log.logger.Logger
+import java.io.File
+import java.io.IOException
+
+/**
+ * An implementation that uses the presence of a specific file in the
+ * User's app directory to determine if this is a first run or not.
+ *
+ * @param gleanDataDir the directory, within the application's data dir, that
+ *        contains glean files.
+ */
+internal class FileFirstRunDetector(
+    private val gleanDataDir: File
+) {
+    private val logger = Logger("glean/FileFirstRunDetector")
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    companion object {
+        const val FIRST_RUN_FILENAME = "glean_already_ran"
+    }
+
+    private val firstRun: Boolean by lazy {
+        !checkFirstRunFileExists()
+    }
+
+    /**
+     * Whether or not the first run file exists in the glean's data directory.
+     *
+     * @return true if the file exists (regardless if that's a dir or a file), false
+     *         if it doesn't exist or an exception was thrown.
+     */
+    private fun checkFirstRunFileExists(): Boolean {
+        val f = File(gleanDataDir, FIRST_RUN_FILENAME)
+        return f.exists()
+    }
+
+    /**
+     * Create the first run file
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    fun createFirstRunFile() {
+        val f = File(gleanDataDir, FIRST_RUN_FILENAME)
+        try {
+            f.createNewFile()
+        } catch (e: IOException) {
+            logger.error("There was a problem creating the first run file ${f.absolutePath}")
+        }
+    }
+
+    /**
+     * Check if this is the first time glean was run.
+     *
+     * @return true if it was, false otherwise
+     */
+    fun isFirstRun(): Boolean {
+        if (firstRun) {
+            createFirstRunFile()
+            return true
+        }
+
+        return false
+    }
+
+    /**
+     * Delete the first run file. This is useful for resetting
+     * the first run state and run again any dependant initialization
+     * code (e.g. in case of profile resets or opting out and then opting
+     * back in).
+     */
+    fun reset() {
+        val f = File(gleanDataDir, FIRST_RUN_FILENAME)
+        f.delete()
+    }
+}
--- a/components/service/glean/src/test/java/mozilla/components/service/glean/firstrun/FileFirstRunDetectorTest.kt
+++ b/components/service/glean/src/test/java/mozilla/components/service/glean/firstrun/FileFirstRunDetectorTest.kt
@@ -1,0 +1,78 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.glean.firstrun
+
+import org.junit.Test
+
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.RuntimeEnvironment
+import java.io.File
+import java.io.IOException
+
+@RunWith(RobolectricTestRunner::class)
+class FileFirstRunDetectorTest {
+    private val dataDir: File = File(RuntimeEnvironment.application.applicationInfo.dataDir)
+
+    @Before
+    fun setUp() {
+        // Delete any first run file, if it's there.
+        val gleanFirstRunFile = File(dataDir, FileFirstRunDetector.FIRST_RUN_FILENAME)
+        try {
+            gleanFirstRunFile.delete()
+        } catch (e: IOException) {
+            // Do nothing, it's a test!
+        }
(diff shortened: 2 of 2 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/firstrun/FileFirstRunDetector.kt
```diff
@@ -0,0 +1,85 @@
...
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    fun checkFirstRunFileExists(): Boolean {
+        val f = File(gleanDataDir, FIRST_RUN_FILENAME)
+        return try {
+            f.exists()
+        } catch (e: SecurityException) {
+            logger.error("There was a problem testing for ${f.absolutePath}")
+            false
+        }
```
pocmo: Why are you catching `SecurityException` here? Are you expecting this file to (sometimes?) live in the "public" storage (requiring a permission) instead of the app local one? If this is always stored in the app directory then I don't think we should see a SecurityException.

components/service/glean/src/main/java/mozilla/components/service/glean/firstrun/FileFirstRunDetector.kt
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /**
+     * Whether or not the first run file exists in the glean's data directory.
+     *
+     * @return true if the file exists (regardless if that's a dir or a file), false
+     *         if it doesn't exist or an exception was thrown.
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    fun checkFirstRunFileExists(): Boolean {
```
pocmo: nit: If they are only visible for testing then I'd pick `internal`. But it's not super important here since the whole class is `internal` too. (Same for the property in the companion object and `createFirstRunFile`.

components/service/glean/src/main/java/mozilla/components/service/glean/firstrun/FileFirstRunDetector.kt
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * A [FirstRunDetector] implementation that uses the presence
+ * of a specific file in the User's app directory to determine
+ * if this is a first run or not.
+ *
+ * @param gleanDataDir the directory, within the application's data dir, that
+ *        contains glean files.
+ */
+internal class FileFirstRunDetector(
+    private val gleanDataDir: String
```
pocmo: nit: Could be a `File`object. Would still allow you to do `val file = File(gleanDataDir, ..)`..

components/service/glean/src/main/java/mozilla/components/service/glean/firstrun/FileFirstRunDetector.kt
```diff
@@ -0,0 +1,85 @@
...
+
+import android.support.annotation.VisibleForTesting
+import mozilla.components.support.base.log.logger.Logger
+import java.io.File
+import java.io.IOException
+
+/**
+ * A [FirstRunDetector] implementation that uses the presence
+ * of a specific file in the User's app directory to determine
+ * if this is a first run or not.
```
pocmo: Nice! This is so generic that I wonder whether this could be moved to `support-utils` (in a later PR). Every component that needs something like this could intiialize this with a dir/file. I assume we do not want to share this "firstrun" state because it's very component-specific though, right?

components/service/glean/src/main/java/mozilla/components/service/glean/firstrun/FirstRunDetector.kt
```diff
@@ -0,0 +1,26 @@
...
+     * @return true if this is the first run, false otherwise.
+     */
+    fun isFirstRun(): Boolean
+
+    /**
+     * Clears up the "next run" state. The next
+     * time [isFirstRun] is called, it will report
+     * a first run.
+     */
+    fun reset()
```
pocmo: The `reset` method confused me: Is this about the first run after an app isntallation or first run in a process or a very component specific meaning of first run?

components/service/glean/src/main/java/mozilla/components/service/glean/firstrun/FirstRunDetector.kt
```diff
@@ -0,0 +1,26 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.glean.firstrun
+
+/**
+ * Defines the interface for a service detecting if this
+ * is the first time the glean is used by the app.
+ */
+internal interface FirstRunDetector {
```
pocmo: Are you expecting this to have different implementations? I wonder if this level of abstraction is needed?

components/service/glean/src/main/java/mozilla/components/service/glean/firstrun/FileFirstRunDetector.kt
Dexterp37: Ah, no. This should only live in the app directory :) Thanks for pointing this out!

components/service/glean/src/main/java/mozilla/components/service/glean/firstrun/FileFirstRunDetector.kt
Dexterp37: Ah, nice!

components/service/glean/src/main/java/mozilla/components/service/glean/firstrun/FirstRunDetector.kt
Dexterp37: Mh, this is a good point. Not right now, I'd say. Ideally, this could use some other backend implementation (shared configuration file? sqlite due to other reasons?). But not now. Maybe I should just remove the interface and reconsider later.

components/service/glean/src/main/java/mozilla/components/service/glean/firstrun/FirstRunDetector.kt
Dexterp37: Right! So, say the application wants to "remove the user profile" or "user data", or opt-out of `glean`. This will make sure, next time glean is opted in again, we go through any "first time" configuration again. In our specific case, this would internally trigger generating a new `client id`.

Dexterp37:
> Looks good to me. Mostly curious about what this will be used for and whether this is something that could be shared.

As I stated in one of the comments (repeating it here so it's a bit more visible :D), we're planning to use this piece of code for generating the `client id` once. We don't really want to change the API to be smarter than it is, so we don't really want something like `setUUIDIfItWasntSetBeforeAlready`. This object will allow us to just do initialization once.

This might also be useful for other apps that might require onboarding.

components/service/glean/src/main/java/mozilla/components/service/glean/firstrun/FileFirstRunDetector.kt
pocmo: Yeah. I was thinking about the opposite: If we provide a FirstRunThingy for everyone in support-utils or support-base... would this be good enough or does "first run" mean something different to you (specifically because you have a clear method here).

components/service/glean/src/main/java/mozilla/components/service/glean/firstrun/FirstRunDetector.kt
pocmo: Yeah, depends on how you intent to use it. Every class/interface adds a little bit to the APK size.

components/service/glean/src/main/java/mozilla/components/service/glean/firstrun/FileFirstRunDetector.kt
Dexterp37: Ahhhh! Sorry, I'm a bit slow today. Yes, it would be good enough, as long as it could be `reset` :)

components/service/glean/src/main/java/mozilla/components/service/glean/firstrun/FirstRunDetector.kt
Dexterp37: I see, good point. I removed it, there's no need for it to be here now.

pocmo:
> As I stated in one of the comments (repeating it here so it's a bit more visible :D), we're planning to use this piece of code for generating the client id once. We don't really want to change the API to be smarter than it is, so we don't really want something like setUUIDIfItWasntSetBeforeAlready. This object will allow us to just do initialization once.

Not sure how this improves the situation though: Now you check for a file instead of the UUID / file containing the UUID. Isn't that another layer and adds the risk of de-syncing both states?

Dexterp37:
> Not sure how this improves the situation though: Now you check for a file instead of the UUID / file containing the UUID. Isn't that another layer and adds the risk of de-syncing both states?

Right, let me expand a bit. In general, having something 'smarter' than push data to the engine conflicts with our idea of having a *push only* API . Data should flow one way from the API to the storage. Adding an API method to "set if not previously set" would require the API to get data out of the engine, violating that design principle.

Moreover, while my example was mainly for the client id, UUID is not the only metric type that would allow such behaviour. Also Dates, for example, will require something similar. And other use cases might come in the future.

But I hear your point about de-syncing: are you thinking of any specific issue or problem we could try to fix here in the `FirstRunThinghy` (I like this name :D).

PR #1377, opened 2018-11-15
Closes #1046: CrashReporterActivity: Handle back button presses

### LINKED ISSUE #1046: CrashReporterActivity: Handle back button presses
In Fennec on back button press we ask in a dialog whether the user wants to "exit without sending a crash report". We may want to do something similar. However a dialog on top of a dialog may look weird?

Or we just continue to not send anything at all?

Fennec:
https://searchfox.org/mozilla-central/source/mobile/android/base/java/org/mozilla/gecko/CrashReporterActivity.java#268-284

Shubhamurkade:
Added anew method onBackPressed to handle back button presses.


components/lib/crash/src/main/java/mozilla/components/lib/crash/prompt/CrashReporterActivity.kt
```diff
@@ -96,4 +96,11 @@ class CrashReporterActivity : AppCompatActivity() {
             }
         }
     }
+	
+	override fun onBackPressed() {
+        super.onBackPressed()
```
pocmo: Let's not call super here. Otherwise we finish immediately.

PR #1389, opened 2018-11-15
Closes #1356: Introduce consumable stream for session fields

### LINKED ISSUE #1356: Introduce consumable stream/list for (observable) session fields
We currently have `Consumable` fields which are used for downloads, long presses, and permission requests. While it's possible to observe multiple (changing) values of these consumables, the session will only hold the last one. Observers would have to save and queue these consumables themselves in order to make sure all of them get processed.

I have an initial patch for this which I need to document and test more, but I will put it up for discussion soon.

csadilek:
This builds on our existing `Consumable` concept and enhances it to support streams of consumable values.

Say we wanted to have multiple concurrent downloads in a session, our `EngineObserver` would set the field as:

```Kotlin
session.downloads = Consumable.stream(download1)
```
and if it later observes more engine download events, it can keep appending or cancelling:
```Kotlin
session.downloads = session.downloads.append(download2)
session.downloads = session.downloads.remove(download2)
```

Within our browser session, we can use `Delegates.vetoable` just as we do now where we veto the new values if that are all consumed, and otherwise keep them to be consumed later:

```Kotlin
 var downloads: ConsumableStream<Download> by Delegates.vetoable(Consumable.stream()) {
    _, _, consumables ->
      val consumers = wrapConsumers<Download> { onDownload(this@Session, it) }
      !consumables.consumeAllBy(consumers)
 }
``` 

The implemenation was pretty simple with our existing `Consumable`. I've added a few more methods that allow us to consume individual values as well as all at once.

### DIFF
--- a/components/support/base/src/main/java/mozilla/components/support/base/observer/Consumable.kt
+++ b/components/support/base/src/main/java/mozilla/components/support/base/observer/Consumable.kt
@@ -4,18 +4,12 @@
 
 package mozilla.components.support.base.observer
 
-import android.support.annotation.VisibleForTesting
-
 /**
  * A generic wrapper for values that can get consumed.
  */
 class Consumable<T> private constructor(
-    private var value: T?
+    internal var value: T?
 ) {
-    // Internal getter for testing only.
-    @VisibleForTesting internal val internalValue: T?
-        get() = value
-
     /**
      * Invokes the given lambda and marks the value as consumed if the lambda returns true.
      */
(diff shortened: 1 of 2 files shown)


components/support/base/src/main/java/mozilla/components/support/base/observer/Consumable.kt
```diff
@@ -4,16 +4,13 @@
...
 /**
  * A generic wrapper for values that can get consumed.
  */
 class Consumable<T> private constructor(
     private var value: T?
 ) {
-    // Internal getter for testing only.
-    @VisibleForTesting internal val internalValue: T?
+    internal val internalValue: T?
         get() = value
```
pocmo: Now that we use this internally we could just make the `value` property itself internal.

csadilek: True, I will update this.

csadilek:
@pocmo Updated, and removed the `Robolectric` test runner as well. We don't need it and the test runs much faster now :)

pocmo:
Nice :)

PR #1391, opened 2018-11-16
Closes #1339: Add scroll behavior support for full screen pages on SystemEngineView.

### LINKED ISSUE #1339: Fullscreen pages can be scrolled (because height of toolbar can be scrolled offscreen)
Our reference browser and sample browser have the same issue that is described in this Focus issue:
https://github.com/mozilla-mobile/focus-android/issues/3815

Let's investigate if we can find a solution for that.

Amejia481:


components/feature/scroll/src/main/java/mozilla/components/feature/scroll/ScrollFeature.kt
```diff
@@ -0,0 +1,55 @@
...
+import mozilla.components.browser.session.SelectionAwareSessionObserver
+import mozilla.components.browser.session.Session
+import mozilla.components.browser.session.SessionManager
+import mozilla.components.concept.engine.EngineView
+
+/**
+ * Feature implementation for connecting an [EngineView] with any View that you want to coordinate scrolling
+ * behavior with.
+ */
+class ScrollFeature(
```
pocmo: naming nit: `Scroll` is not precise enough yet, I think. Can we find something that is describing better was that feature does? I was thinking of something like "ToolbarHidingFeature" but its more generic than just about the toolbar (however we should mention the toolbar use case in the comment maybe!).

pocmo: CoordinateScrollingFeature? SyncScrollingFeature? ScrollFlagsFeature?

Amejia481: Yeah name is hard 😅. I like `CoordinateScrollingFeature` 👍 .

README.md
```diff
@@ -114,6 +114,8 @@ _Combined components to implement feature-specific use cases._
 
 * 🔴 [**Toolbar**](components/feature/toolbar/README.md) - A component that connects a (concept) toolbar implementation with the browser session module.
 
+* 🔴 [**Scroll**](components/feature/scroll/README.md) - A component coordinates scroll behavior with an `EngineView`.
```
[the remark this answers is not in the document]
Amejia481: Fair enough! 👍

PR #1410, opened 2018-11-19
[glean] Enable sending the "baseline" ping from the sample app

Dexterp37:
This PR additionally fixes a few bugs in the ping submission code:
- HTTP I/O needs to happen off the main thread otherwise we crash;
- `path` and `payload` arguments where swapped in `upload()`

Moreover, this PR also introduces:

- the internal object that manages the set of core metrics used by
Glean. This file will be removed and automatically generated on build
in future PRs;
- the (internal) ability to have metrics without a category name, for
using built-in metrics when assembling the "ping_info" section of the ping;
- the "mergeWith" `JSONObject` extension function, to merge two `JSONObject`
instances.

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
@@ -6,15 +6,22 @@ package mozilla.components.service.glean
 
 import android.content.Context
 import android.support.annotation.VisibleForTesting
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.launch
 
 import java.util.UUID
 
 import mozilla.components.service.glean.config.Configuration
+import mozilla.components.service.glean.firstrun.FileFirstRunDetector
+import mozilla.components.service.glean.metrics.GleanInternalMetrics
 import mozilla.components.service.glean.net.HttpPingUploader
 import mozilla.components.service.glean.ping.PingMaker
 import mozilla.components.service.glean.storages.ExperimentsStorageEngine
 import mozilla.components.service.glean.storages.StorageEngineManager
 import mozilla.components.support.base.log.logger.Logger
+import java.io.File
 
 object Glean {
     /**
@@ -35,6 +42,12 @@ object Glean {
 
     internal const val SCHEMA_VERSION = 1
 
+    /**
+     * The name of the directory, inside the application's directory,
+     * in which Glean files are stored.
+     */
+    internal const val GLEAN_DATA_DIR = "glean_data"
+
     // Include our singletons of StorageEngineManager and PingMaker
     private lateinit var storageEngineManager: StorageEngineManager
     private lateinit var pingMaker: PingMaker
@@ -46,6 +59,11 @@ object Glean {
     private var metricsEnabled = true
 
     /**
+     * The instance holding lazy references to the Glean's core metrics.
+     */
+    private val gleanInternalMetrics by lazy { GleanInternalMetrics() }
+
+    /**
      * Initialize glean.
      */
     fun initialize(applicationContext: Context, configuration: Configuration) {
@@ -54,12 +72,14 @@ object Glean {
         this.configuration = configuration
         httpPingUploader = HttpPingUploader(configuration)
         initialized = true
+
+        initializeCoreMetrics(applicationContext)
     }
 
     /**
      * Enable or disable metric collection.
      *
-     * Metric collection is emabled by default.
+     * Metric collection is enabled by default.
      *
      * When disabled, metrics aren't recorded at all.
      *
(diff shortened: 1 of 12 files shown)


samples/glean/src/main/java/org/mozilla/samples/glean/MainActivity.kt
```diff
@@ -18,11 +20,15 @@ open class MainActivity : AppCompatActivity() {
 
         Log.addSink(AndroidLogSink())
 
+        // Initialize the Glean library. Ideally, this is the first thing that
+        // must be done right after enabling logging.
+        Glean.initialize(applicationContext, Configuration())
```
pocmo: Eventually we may want to move this code from the activity to an application object to showcase the intended usage. But obviously we do not need to do that now. :)

samples/glean/src/main/java/org/mozilla/samples/glean/MainActivity.kt
```diff
@@ -32,5 +38,10 @@ open class MainActivity : AppCompatActivity() {
                     )
             )
         }
+
+        // Generate a "baseline" ping on click.
+        findViewById<View>(R.id.buttonSendPing).setOnClickListener {
+            Glean.handleEvent(Glean.PingEvent.Background)
+        }
```
pocmo: nit: Kotlin Android Extensions are enabled for this sample app. You should be able to just use `buttonSendPing` and `buttonWebView`. Android Studio should then offer to import that from `kotlinx.android.synthetic.main.activity_main.`. With that no `findViewById` is needed. :)

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -114,11 +134,45 @@ object Glean {
...
+     */
+    private fun sendPing(store: String, docType: String): Job {
         val pingContent = pingMaker.collect(store)
         val uuid = UUID.randomUUID()
         val path = makePath(docType, uuid)
-        this.httpPingUploader.upload(pingContent, path)
+        // Asynchronously perform the HTTP upload off the main thread.
+        return CoroutineScope(Dispatchers.IO).launch {
+            httpPingUploader.upload(path = path, data = pingContent)
+        }
```
pocmo: Since you asked about Coroutines specifically: This looks good to me. You do not need any fancy cancellation (+ automatic children cancellation) so I think this is good. Instead of creating a new scope on the fly you could also consider using [GlobalScope](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-global-scope/) here. Usually that's something that is discouraged because this launches an unscoped global job that is never cancelled; but that's exactly what you want here.

With GlobalScope that would be something like:

```Kotlin
    // ...
}
```

If you ever intend to do more work in the background then you could consider creating your own dispatcher. But for now that doesn't seem to be needed.

samples/glean/src/main/java/org/mozilla/samples/glean/MainActivity.kt
Dexterp37: Ah, this is a good point. But I have no clue on how that would work. Do you think you or anyone in your team would have time to mentor/stub out the changes to the sample application? It would be great to showcase the "best" way to use the API, but I'm not sure I have the titles to say what that is :)

samples/glean/src/main/java/org/mozilla/samples/glean/MainActivity.kt
Dexterp37: Ah, nice! AndroidStudio didn't offer me that, but everything works fine after manually importing it!

samples/glean/src/main/java/org/mozilla/samples/glean/MainActivity.kt
Dexterp37: I filed bug 1508227 for this.

pocmo: In a nutshell: 
* Create a class that extends `android.app.Application`: https://github.com/mozilla-mobile/reference-browser/blob/master/app/src/main/java/org/mozilla/reference/browser/BrowserApplication.kt#L13
* Reference it from the application tag (name) in AndroidManifest.xml: https://github.com/mozilla-mobile/reference-browser/blob/master/app/src/main/AndroidManifest.xml#L20
* Do you initialization from onCreate() of the application object.

pocmo:
@mdboom Oh sorry for merging before flagging you for review.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -114,11 +134,45 @@ object Glean {
...
+
+        if (!gleanDataDir.isDirectory || !gleanDataDir.canWrite()) {
+            logger.error("Glean's data directory is not a writable directory ${gleanDataDir.absolutePath}")
+            return
+        }
+
+        // The first time Glean runs, we set the client id and other internal
+        // one-time only metrics.
+        val firstRunDetector = FileFirstRunDetector(gleanDataDir)
```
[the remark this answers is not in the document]
Dexterp37: Good question, I should have left a comment there to clarify. Ideally, we will have more than just the "first run file" living in the glean data dir: think of pings we fail to send and that we want to retry sending next time. For this reason, handling the directory creation shouldn't probably live in the `FileFirstRunDetector`, which should be agnostic to "dirs". It should also probably live somewhere else: maybe we should have a concept of "GleanFileSystem" once we have ping persistence in place.

mdboom:
@pocmo No problem!  My point is very minor...

Issue #743, opened 2018-09-10
Sample Browser: Clicking the tab menu icon freezes the entire app with SystemEngine


jonalmeida:
Steps to reproduce:
1. Load https://www.microsoft.com/en-ca/
2. Click the tab menu icon.
3. Observe the delay in switching from the current tab to the tab view.

GIF of the issue: https://gfycat.com/UnluckyLikelyIggypops

jonalmeida:
Looking at the memory profiler, there's a large de-allocation taking place when we click the tab icon.
In the image below, every dot is a click on the tab icon, and every other dot is a click on the thumbnail back to the WebView.

<img width="1721" alt="screen shot 2018-09-10 at 6 09 02 pm" src="https://user-images.githubusercontent.com/1370580/45327305-03626e80-b525-11e8-9282-f93472728062.png">

jonalmeida:
It looks like the issue is related to the allocations from the `TabsAdapter`. There are a lot of [`Rect`][1] objects being created from the view inflation.

My guess is that our implementation of the tabs with the lazy initialization might be related to this.

[1]: https://developer.android.com/reference/android/graphics/Rect.html

csadilek:
Hm I am actually not able to switch between tabs more than once with SystemEngine (Gecko is fine.)...

pocmo:
When adding the tabs fragment and removing the browser fragment we save the state of the sessions to disk (`BrowserFragment.onStop() -> SessionFeature.stop() -> SessionStorage.persist()`). Can this be causing this?

jonalmeida:
We should get this fixed soon since more and more people are using the reference browser.

pocmo:
> We should get this fixed soon since more and more people are using the reference browser.

As long as it is in the Gecko builds too.. :)

pocmo:
Okay, I have an idea and a prototype. Will propose something. :)

PR #1413, opened 2018-11-19
Issue #743: SessionStorage: Customizable UI-independent session storage.....

pocmo:
This PR contains multiple changes to our session storage API:

* It removes session storage handling from SessionFeature: Previously we blocked the UI thread whenever the feature was stopped (e.g. fragment transaction). That caused UI freezes and also some "dead locks" where saving the session took more than 3 seconds (while less than 10 milliseconds is normal).

* Because of the decoupling from the UI I removed the interface and renamed DefaultSessionStorage to SessionStorage. There will very likely be only one implementation of session storage. And even if more are needed then this can be done without a shared interface.

* In addition to the periodic save this PR introduces two additional mechanisms: Saving when the app goes to the background and saving when the sessions change.

* Automatic saving of sessions can be configured by the app:

```Kotlin
sessionStorage.autoSave(this)
	.periodicallyInForeground(interval = 10, unit = TimeUnit.SECONDS)
	.whenGoingToBackground()
	.whenSessionsChange()
```

* Automatic saving will take care of keeping a minimum interval between saves and will delay saves if needed. Additionally it will avoid saving if a save is already in flight.


components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
```diff
@@ -40,9 +38,9 @@ class SessionManager(
      *
      * @return [SessionsSnapshot] or null if no sessions are present.
```
csadilek: KDocs need an update `@return [Snapshot]....`

components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
```diff
@@ -186,7 +184,7 @@ class SessionManager(
      * @param snapshot A [SessionsSnapshot] which may be produced by [createSnapshot].
```
csadilek: KDoc update...[Snapshot]

components/browser/session/src/main/java/mozilla/components/browser/session/storage/SessionStorage.kt
```diff
@@ -4,59 +4,428 @@
...
+ */
+private class AutoSaveBackground(
+    private val autoSave: AutoSave
+) : LifecycleObserver {
+    @OnLifecycleEvent(Lifecycle.Event.ON_STOP)
+    fun stop() {
+        autoSave.logger.info("Save: Background")
+
+        runBlocking {
+            autoSave.triggerSave().join()
```
csadilek: Do we need a flag here or similar to prevent a potential scheduled delay? If a (periodic) save happened less than 2 seconds (minimum interval) ago, and the app goes to the background we would delay the save, right? So, we'd be blocking for longer than necessary.

pocmo: Oh yes, that's a good point. I'll update the PR to add some kind of "no delay flag".

pocmo:
Damn, that test is still failing. I'll need to investigate that a bit. Was hoping to get that in 0.32.0 but I'll trigger that release now.

pocmo:
This seems to work now. Merging this in. The 0.32.0 release just failed; so I can still sneak that in. ;)

PR #1417, opened 2018-11-19
[glean] Check generated ping content against their json schema

mdboom:
The purpose of this change is to be able to write unit tests that validate the generated pings (in JSON) against the JSON schema that will be validating the pings on the backend in production (and thus catch any schema violations early, prior to merging).

@Dexterp37 did some [initial investigation](https://bugzilla.mozilla.org/show_bug.cgi?id=1504200) doing this with Java/Kotlin. tl;dr: there are two candidate libraries for JSON schema available, one doesn't work with Android due to differences in the `org.json` API, and one is basically a dormant project that doesn't support recent versions of JSON schema.

The approach here is to use Python (which we are already using for code generation at build time) to do the schema validation. This has the added benefit that we can use exactly the same [schema validation library](https://github.com/Julian/jsonschema) that we use in production, so we don't have to "hope" that it has the same corner case bugs as some other JSON schema validation library.

Prior to this change, the Python environment was installed to the application build directory (`samples/glean/build`).  In order to use it for unit testing, we need to be able to use it in the from the library, so it now installs the Python environment to the library's build directory (`components/services/glean/build`).  The same Python environment is thus used for both purposes: doing JSON schema validation as part of the library's unit tests and build-time code generation for the application.  Additionally, the same Python library/script (`glean_parser`) is used for both purposes.

One wrinkle is that the library needs to know how to find the Python executable within the library build directory, even when run as part of the application's build process.  There doesn't seem to be a way to write a Gradle plugin that looks for things relative to the plugin's location, only to the build or project root.  Therefore, using Glean's custom gradle plugin is a two step process: import the plugin using `apply from:` and then calling a function `setupGlean` and passing in the path to the library relative to the project root.  That definitely should be documented, but I thought I would solicit feedback on the approach first.

Requires: 
- [x] https://github.com/mozilla/glean_parser/pull/15 to be merged first

### DIFF
--- a/components/service/glean/build.gradle
+++ b/components/service/glean/build.gradle
@@ -2,8 +2,20 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+plugins {
+    id "com.jetbrains.python.envs" version "0.0.25"
+}
+
 apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
+
+/*
+ * This defines the location of the JSON schema used to validate the pings
+ * created during unit testing.
+ * This uses a specific version of the schema identified by a git commit hash.
+ */
+String GLEAN_PING_SCHEMA_GIT_HASH = "3a15121c"
+String GLEAN_PING_SCHEMA_URL = "https://raw.githubusercontent.com/mozilla-services/mozilla-pipeline-schemas/$GLEAN_PING_SCHEMA_GIT_HASH/schemas/glean/baseline/baseline.1.schema.json"
 
 android {
     compileSdkVersion Config.compileSdkVersion
@@ -13,6 +25,7 @@ android {
         targetSdkVersion Config.targetSdkVersion
 
         buildConfigField("String", "LIBRARY_VERSION", "\"" + Config.componentsVersion + "\"")
+        buildConfigField("String", "GLEAN_PING_SCHEMA_URL", "\"" + GLEAN_PING_SCHEMA_URL + "\"")
     }
 
     buildTypes {
@@ -39,3 +52,5 @@ dependencies {
 
 apply from: '../../../publish.gradle'
 ext.configurePublish(Config.componentsGroupId, archivesBaseName, gradle.componentDescriptions[archivesBaseName])
+
+apply from: './scripts/sdk_generator.gradle'
--- a/components/service/glean/scripts/sdk_generator.gradle
+++ b/components/service/glean/scripts/sdk_generator.gradle
@@ -4,8 +4,7 @@
 
 import org.apache.tools.ant.taskdefs.condition.Os
 
-String GLEAN_PARSER_VERSION
-GLEAN_PARSER_VERSION = "0.6.0"
+String GLEAN_PARSER_VERSION = "0.7.0"
 
 /* This script runs a given Python module as a "main" module, like
  * `python -m module`. However, it first checks that the installed
@@ -49,7 +48,8 @@ envs {
     envsDirectory = new File(buildDir, 'envs')
     pipInstallOptions = "--trusted-host pypi.python.org --no-cache-dir"
 
-    // Setup a miniconda environment.
+    // Setup a miniconda environment. conda is used because it works
+    // non-interactively on Windows, unlike the other options
     conda "Miniconda3", "Miniconda3-latest", "64", ["glean_parser==${GLEAN_PARSER_VERSION}"]
 }
 
@@ -74,7 +74,7 @@ preBuild.dependsOn(createBuildDir)
 preBuild.finalizedBy("build_envs")
 
 // Generate the Metrics API
-android.applicationVariants.all { variant ->
+ext.gleanGenerateMetricsAPI = { variant ->
     def sourceOutputDir =  "$buildDir/telemetry/src/${variant.name}/kotlin"
     def generateKotlinAPI = task("generateMetricsSourceFor${variant.name.capitalize()}", type: Exec) {
         description = "Generate the Kotlin code for the Metrics API"
@@ -115,4 +115,8 @@ android.applicationVariants.all { variant ->
     // The added directory doesn't appear in the paths listed by the
     // `sourceSets` task, for reasons unknown.
     variant.registerJavaGeneratingTask(generateKotlinAPI, new File(sourceOutputDir))
+}
+
+if (android.hasProperty('applicationVariants')) {
+    android.applicationVariants.all(ext.gleanGenerateMetricsAPI)
 }
--- a/components/service/glean/src/test/java/mozilla/components/service/glean/TestUtil.kt
+++ b/components/service/glean/src/test/java/mozilla/components/service/glean/TestUtil.kt
@@ -1,0 +1,40 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this
+* file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.glean
+
+import org.json.JSONObject
+
+fun checkPingSchema(content: JSONObject) {
+    val os = System.getProperty("os.name")?.toLowerCase()
+    val pythonExecutable =
+        if (os?.indexOf("win")?.compareTo(0) == 0)
+            "./build/bootstrap/Miniconda3/python"
+        else
+            "./build/bootstrap/Miniconda3/bin/python"
+
+    val proc = ProcessBuilder(
+        listOf(
+            pythonExecutable,
+            "-m",
+            "glean_parser",
+            "check",
+            "-s",
+            "${BuildConfig.GLEAN_PING_SCHEMA_URL}"
+        )
+    ).redirectOutput(ProcessBuilder.Redirect.INHERIT)
+        .redirectError(ProcessBuilder.Redirect.INHERIT)
+    val process = proc.start()
+
+    val jsonString = content.toString()
+    with(process.outputStream.bufferedWriter()) {
+        write(jsonString)
+        newLine()
+        flush()
+        close()
+    }
+
+    val exitCode = process.waitFor()
+    assert(exitCode == 0)
+}
--- a/components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
+++ b/components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
@@ -143,6 +143,7 @@ class GleanTest {
 
             val (metricsJsonData, metricsPath) = payloads.get("metrics")!!
             val metricsJson = JSONObject(metricsJsonData)
+            checkPingSchema(metricsJson)
             assertEquals(
                 "foo",
                 metricsJson.getJSONObject("metrics")


components/service/glean/build.gradle
```diff
@@ -2,9 +2,43 @@
...
 
+plugins {
+    id "com.jetbrains.python.envs" version "0.0.25"
+}
+
 apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 
+// Configure the Python environments
+envs {
```
Dexterp37: Even after discussing this with you I admit I'm still a bit confused.

We're moving the setup code from `sdk_generator.gradle` to the library's `build.gradle`. I think that doing this would prevent 3rd party applications to process YAML files: if you create an application external to `android-components`, include *glean* and trigger the build process, it would now probably fail: I'm not sure we'd be able to call the tasks from the *glean*'s gradle.

I claim we should think the sample application as an external entity even though it lives in this repo: it should have no way to access things outside its direct reach, other than through gradle inclusion.

In short, I think it's ok for us to download the environment both in the app and when building the library, and have it in both places.

components/service/glean/src/test/java/mozilla/components/service/glean/TestUtil.kt
```diff
@@ -0,0 +1,25 @@
...
+
+fun checkPingSchema(content: JSONObject) {
+    val proc = ProcessBuilder(
+        listOf("./build/bootstrap/Miniconda3/bin/python", "-m", "glean_parser", "check")
+    ).redirectOutput(ProcessBuilder.Redirect.INHERIT)
+        .redirectError(ProcessBuilder.Redirect.INHERIT)
+    val process = proc.start()
+
+    val stdinWriter = process.outputStream.bufferedWriter()
+    val jsonString = content.toString()
```
Dexterp37: If you put `val jsonString =`  on the line before `val stdinWriter` then you can do something like:

```
with (stdinWriter) {
  write(jsonString)
  newLine()
  flush()
  close()
}
```

components/service/glean/src/test/java/mozilla/components/service/glean/TestUtil.kt
```diff
@@ -0,0 +1,25 @@
...
+* License, v. 2.0. If a copy of the MPL was not distributed with this
+* file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.glean
+
+import org.json.JSONObject
+
+fun checkPingSchema(content: JSONObject) {
+    val proc = ProcessBuilder(
+        listOf("./build/bootstrap/Miniconda3/bin/python", "-m", "glean_parser", "check")
```
Dexterp37: I wonder if there's a better way to do this, without hardcoding paths. Maybe adding this to the resources? @pocmo , thoughts?

samples/glean/build.gradle
```diff
@@ -43,3 +39,5 @@ dependencies {
 }
 
 apply from: '../../components/service/glean/scripts/sdk_generator.gradle'
+
+ext.setupGlean('./components/service/glean')
```
Dexterp37: As stated above, this woldn't be possible if the sample app lived outside of this repo. Let's keep the scripts as they are, even if we end up with two conda environments. That's ok, as the library and the app are two distinct things, as if they were different repos.

@pocmo is there something we could do to cache the conda installer/packages to be nicer to them?

mdboom:
Thanks.  You are right -- gradle clearly doesn't work how I assumed, and if using an external dependency that may get shipped as a binary and thus wouldn't have the Python environment in the build products.

I've updated this so there is a separate Python environment for the library and the application.  It seems to work well at the cost of some disk space.  Each conda environment takes about 305MB.  Using a standard upstream Python environment is a little smaller at 220MB.  Since I don't see any need for conda for this project, we might as well save the space.

@Dexterp37: Of particular note, we should probably test this on Windows if you can prior to merging.  It's not expected to work until mozilla/glean_parser#15 is merged and a new release tagged, however.

components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -115,4 +114,8 @@ android.applicationVariants.all { variant ->
     // The added directory doesn't appear in the paths listed by the
     // `sourceSets` task, for reasons unknown.
     variant.registerJavaGeneratingTask(generateKotlinAPI, new File(sourceOutputDir))
+}}
+
+if (android.hasProperty('applicationVariants')) {
+    ext.gleanGenerateMetricsAPI(android.applicationVariants.all)
```
mdboom: This is where the library-local metrics.yaml for the baseline telemetry will go (in a future PR).

components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -49,8 +48,8 @@ envs {
     envsDirectory = new File(buildDir, 'envs')
     pipInstallOptions = "--trusted-host pypi.python.org --no-cache-dir"
 
-    // Setup a miniconda environment.
-    conda "Miniconda3", "Miniconda3-latest", "64", ["glean_parser==${GLEAN_PARSER_VERSION}"]
+    // Setup a standard python environment.
+    python "python37_64", "3.7.1", ["glean_parser==${GLEAN_PARSER_VERSION}"]
```
Dexterp37: Good point in fixing the version for the environment. Why are we switching from Miniconda3 to Python? I think it makes sense, but, as you said, we should double check about any weirdness when installing on Windows. I seem to remember (gah, I should have documented that!) that there were issues (e.g. environment not being installed and devs prompted with dialogs) with the barebone environment on Windows :(

**tl;dr** - is this change strictly required in this PR?

components/service/glean/src/test/java/mozilla/components/service/glean/TestUtil.kt
```diff
@@ -0,0 +1,43 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this
+* file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.glean
+
+import org.json.JSONObject
+
+const val PING_SCHEMA_GIT_HASH = "3a15121c"
```
Dexterp37: We could probably fetch the script using Gradle and pin the hash in the build configuration too, to avoid hardcoding this in the source. Downloading the schema could be done through the `task` described in the related [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1504200#c0)

components/service/glean/scripts/sdk_generator.gradle
Dexterp37: Sorry for being pedantic about this 🤦‍♂️ But why is this change needed? (along with all the changes in this file)

For installing the environment, we just need to include the `plugins` and `sdk_generator.gradle` in our gradle file, right? If these changes are not strictly required for this PR, I'd be for dropping them: it would make it easier to reason about them in the follow-up PR related to the library `metrics.yaml`.

components/service/glean/scripts/sdk_generator.gradle
mdboom: This change isn't strictly required -- it was just an attempt to reduce disk space usage, since conda is overkill for this purpose.  But obviously, if this doesn't work non-interactively on Windows, that's a non-starter and I can revert it.

Dexterp37: Yeah, sorry about this, I should have documented my choice in the first place. There's something quirky going on in the environments plugin when the plain python is picked, on Windows. Something that has to do with permissions and stuff, making the installation process interactive (showing the UI), which is undesired. Let's revert this and face the problem as a separate issue, if/when needed.

components/service/glean/src/test/java/mozilla/components/service/glean/TestUtil.kt
mdboom: I don't think downloading the schema using gradle is a good idea.  We would run into the same versioning issues we had with glean_parser, but even worse, without a built-in way to manage record, check and manage those versions.

I would be fine with storing the git hash in `sdk_generator.gradle` and injecting that value here, assuming that's possible.

components/service/glean/scripts/sdk_generator.gradle
mdboom: When using this plugin in the context of the library, there is to `android.applicationVariants` (makes sense, because we aren't building an application in that context), so line 77 of the original gives an "undefined variable" exception.  This turns the chunk that runs the generator into a function so it can be run only when building the application, and in a future PR, it will be called differently to handle the needs of the library.  And it allows the rest of the file (the setting up of the Python environment) to be shared in both the library and application contexts.

components/service/glean/src/test/java/mozilla/components/service/glean/TestUtil.kt
Dexterp37: > I would be fine with storing the git hash in `sdk_generator.gradle` and injecting that value here, assuming that's possible.

@pocmo any idea if this is possible or not?

components/service/glean/scripts/sdk_generator.gradle
Dexterp37: Ah, gotcha! Thanks for the clarification, it makes sense!

components/service/glean/scripts/sdk_generator.gradle
mdboom: I've reverted and added a comment about why conda is the better choice on Windows.

components/service/glean/src/test/java/mozilla/components/service/glean/TestUtil.kt
mdboom: I've submitted a method that sets the URL to a resource string and then accesses that from the code.  If there's a better method, let me know.

samples/glean/build.gradle
pocmo: Would it be possible to use the `buildDir` of `rootProject` to share the Python environment? I'm not sure whether this could cause weird side effects when building parallel though (although some parts of samples-glean needs to wait for services-glean to complete).

Having the environment twice is not the best for saving disk space; however that's how the setup with real apps will be too. So i'm not too concerned.

components/service/glean/src/test/java/mozilla/components/service/glean/TestUtil.kt
pocmo: With those paths being used in the gradle scripts and the tests I can't think of a shared place where both world have access. From the gradle script you could inject it into a constant; but that constant would be available to all glean code and not limited to tests.

components/service/glean/build.gradle
```diff
@@ -19,6 +31,11 @@ android {
         release {
             minifyEnabled false
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+            resValue("string", "GLEAN_PING_SCHEMA_URL", GLEAN_PING_SCHEMA_URL)
+        }
+
+        debug {
+            resValue("string", "GLEAN_PING_SCHEMA_URL", GLEAN_PING_SCHEMA_URL)
```
pocmo: Maybe instead of a String resource you want to use just a `BuildConfig` constant. If you define it in the `defaultConfig` block then you won't need to duplicate it (See `LIBRARY_VERSION` example above).

samples/glean/build.gradle
Dexterp37: > Would it be possible to use the `buildDir` of `rootProject` to share the Python environment? I'm not sure whether this could cause weird side effects when building parallel though (although some parts of samples-glean needs to wait for services-glean to complete).

Mh, good point. @pocmo, is the following reasoning correct:

- `rootProject.buildDir == buildDir` for most, common apps
- `rootProject.buildDir != buildDir` if glean is used by a library/component in an app with multiple components

I guess that could be ok. Moreover, this saves us some trouble when we'll start collecting telemetry from libraries, preventing them from requiring an environment each when building (think of components in `android-components`, maybe some of them want to be instrumented and we don't really want to have an environment for each of them). Right?

> Having the environment twice is not the best for saving disk space; however that's how the setup with real apps will be too. So i'm not too concerned.

Agreed! However, in general, one wouldn't really need to install the *glean* development environment. App developer would just need to have **one** environment. I think the sample within this repo is a special case, but a corner case as well :)

mdboom:
This is passing CI, and I think is ready for a final look and merge. :whew:

samples/glean/build.gradle
mdboom: Sorry I missed this comment the first time around.  As it stands, this PR is creating a Python environment for the library itself, and each application that uses it.  This could be unnecessarily large when we start using this in libraries, as @Dexterp37 points out.  But I would argue for merging this as-is with a follow-on bug to address that when we get there.

PR #1419, opened 2018-11-19
Fixes #1366: Don't use private org.mozilla.gecko.

### LINKED ISSUE #1366: a-c tests should not use EventDispatcher
Discussed with @pocmo a little bit. `EventDispatcher` is a private interface to GV and should not be used in tests like [0]. We are trying to hide it so I'm cleaning up every package that use it. I'll send a pull request for this soon.

[0] https://github.com/mozilla-mobile/android-components/blob/386fc6758d2ee1780f72e62c74a29891c42a08d4/components/browser/engine-gecko-nightly/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineSessionTest.kt#L229-L240

agi90:
This commit removes most uses of `org.mozilla.gecko` in a-c. Namely:

- EventDispatcher is not used anymore, instead GeckoSession is mocked
- org.mozilla.gecko.ThreadUtils is replaced with utils.ThreadUtils

org.mozilla.gecko.GeckoBundle uses will be fixed in a follow-up commit.

### DIFF
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -21,7 +21,7 @@ import mozilla.components.support.ktx.kotlin.isEmail
 import mozilla.components.support.ktx.kotlin.isGeoLocation
 import mozilla.components.support.ktx.kotlin.isPhone
 import mozilla.components.support.utils.DownloadUtils
-import org.mozilla.gecko.util.ThreadUtils
+import mozilla.components.support.utils.ThreadUtils
 import org.mozilla.geckoview.AllowOrDeny
 import org.mozilla.geckoview.GeckoResult
 import org.mozilla.geckoview.GeckoRuntime
@@ -41,10 +41,11 @@ import org.mozilla.geckoview.WebRequestError
 class GeckoEngineSession(
     private val runtime: GeckoRuntime,
     private val privateMode: Boolean = false,
-    private val defaultSettings: Settings? = null
+    private val defaultSettings: Settings? = null,
+    private val geckoSessionProvider: () -> GeckoSession = { GeckoSession() }
 ) : EngineSession() {
 
-    internal var geckoSession = GeckoSession()
+    internal var geckoSession: GeckoSession = geckoSessionProvider()
     internal var currentUrl: String? = null
 
     /**
@@ -122,7 +123,7 @@ class GeckoEngineSession(
     override fun saveState(): Map<String, Any> = runBlocking {
         val stateMap = CompletableDeferred<Map<String, Any>>()
 
-        ThreadUtils.sGeckoHandler.post {
+        ThreadUtils.postToBackgroundThread {
             geckoSession.saveState().then({ state ->
                 stateMap.complete(mapOf(GECKO_STATE_KEY to state.toString()))
                 GeckoResult<Void>()
@@ -475,7 +476,7 @@ class GeckoEngineSession(
     }
 
     private fun initGeckoSession() {
-        this.geckoSession = GeckoSession()
+        this.geckoSession = geckoSessionProvider()
         defaultSettings?.trackingProtectionPolicy?.let { enableTrackingProtection(it) }
         defaultSettings?.requestInterceptor?.let { settings.requestInterceptor = it }
         defaultSettings?.historyTrackingDelegate?.let { settings.historyTrackingDelegate = it }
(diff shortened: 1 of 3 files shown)


components/browser/engine-gecko-nightly/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineSessionTest.kt
```diff
@@ -60,24 +56,48 @@ import org.robolectric.RobolectricTestRunner
 @RunWith(RobolectricTestRunner::class)
 class GeckoEngineSessionTest {
 
+    private val navigationDelegate = ArgumentCaptor.forClass(
```
csadilek: nit: I would `lateinit` these and initialize in `setup`. Makes it less likely we'll see some test ordering issues / sideeffects in the future.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -133,7 +133,7 @@ class GeckoEngineSession(
     override fun saveState(): Map<String, Any> = runBlocking {
         val stateMap = CompletableDeferred<Map<String, Any>>()
 
-        ThreadUtils.sGeckoHandler.post {
+        ThreadUtils.postToBackgroundThread {
             geckoSession.saveState().then({ state ->
```
csadilek: Same concern as @pocmo. I remember this not working on a background thread. Does it now?

components/browser/engine-gecko-nightly/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineSessionTest.kt
```diff
@@ -26,20 +26,18 @@ import org.junit.Assert.assertNotNull
 import org.junit.Assert.assertNull
 import org.junit.Assert.assertTrue
 import org.junit.Before
-import org.junit.Test
+import org.junit.Test;
```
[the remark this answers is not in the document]
agi90: Fixed all ktlint issues :)

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -133,7 +133,7 @@ class GeckoEngineSession(
     override fun saveState(): Map<String, Any> = runBlocking {
         val stateMap = CompletableDeferred<Map<String, Any>>()
 
-        ThreadUtils.sGeckoHandler.post {
+        ThreadUtils.postToBackgroundThread {
```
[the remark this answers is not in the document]
agi90: Eh you are of course right this doesn't work! I changed `ThreadUtils` to use a `HandlerThread` which has a `Looper`. Let me know if you have any concern about that.

components/browser/engine-gecko-nightly/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineSessionTest.kt
agi90: TIL `lateinit` (this is my first kt commit :)) fixed thanks!

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
agi90: Yup you are both right! See my update in `ThreadUtils`

agi90:
This is ready for re-review, thank you @pocmo @csadilek !

components/support/utils/src/main/java/mozilla/components/support/utils/ThreadUtils.kt
```diff
@@ -5,18 +5,40 @@
...
+        looperBackgroundThread.start()
+        Handler(looperBackgroundThread.looper)
+    }
     private val handler = Handler(Looper.getMainLooper())
     private val uiThread = Looper.getMainLooper().thread
 
+    private var handlerForTest: Handler? = null
+
+    private fun backgroundHandler(): Handler {
+        if (handlerForTest != null) {
```
csadilek: nit: This could be one line: `return handlerForTest ?: looperBackgroundHandler`

agi90: nice! yay Kotlin. Fixed!

Issue #1428, opened 2018-11-20
Flaky test on SessionStorageTest "AutoSave - when going to background FAILED"


Amejia481:
```
> Task :browser-session:testDebugUnitTest
mozilla.components.browser.session.storage.SessionStorageTest > AutoSave - when going to background FAILED
    org.mockito.exceptions.verification.WantedButNotInvoked at SessionStorageTest.kt:175

```

pocmo:
The only explanation I have so far is that something inside the Coroutine may throw. Since we are using GlobalScope that exception is probably lost.

pocmo:
I pushed a PR that is a bit more verbose.

It turns out that `autoSave.saveJob` is null in that case and `autoSave.saveJob?.join()` doesn't wait for anyhting.

I have no clue why it is null though. It seems like the LifecycleObserver is never invoked.

pocmo:
After reading through the code of `LifecycleRegistry` I finally understand what's going on.

At the beginning of our test we create a lifecycle for testing:

```Kotlin
val lifecycle = LifecycleRegistry(mock(LifecycleOwner::class.java))
```

Deep in `LifecycleRegistry` it checks whether the owner still exists:

```
LifecycleOwner lifecycleOwner = mLifecycleOwner.get();
        if (lifecycleOwner == null) {
            Log.w(LOG_TAG, "LifecycleOwner is garbage collected, you shouldn't try dispatch "
                    + "new events from it.");
            return;
        }
```


We do not keep a reference to the `LifecycleOwner` so if it was garbage collected at this time then our observer would never get notified. If garbage collection happens depends on a bunch of things and this makes this flaky here: Locally everything is fine but if the taskcluster worker may be a bit low on memory it may GC earlier.

Calling `System.gc()` in the test makes this 100% reproducible. Just keeping a reference to the owner should fix the test.

PR #1431, opened 2018-11-20
1506061: [glean] Refactor to pass whole CommonMetricType rather than fields

mdboom:
https://bugzilla.mozilla.org/show_bug.cgi?id=1506061

This is just a refactor.  Rather than passing individual fields of the CommonMetricType from the MetricType down into the storage engines, we just pass the CommonMetricType.  Since we expect the number of fields and metric types to grow, this should make it easier to do that for multiple metric types in the future.


mdboom:
Thanks for the review.  Pushed changes to address your comments, @Dexterp37

pocmo:
> Failure seems like it might be unrelated? (@pocmo?)

Yep, that's #1428. I'll re-trigger.

PR #1433, opened 2018-11-21
Closes #1383: App crash caused by invalid URI in GeckoSession 

### LINKED ISSUE #1383: (GeckoEngineSession$createNavigationDelegate) IllegalArgumentException: Parameter specified as non-null is null
From the reference browser: https://github.com/mozilla-mobile/reference-browser/issues/193

```
java.lang.IllegalArgumentException: Parameter specified as non-null is null: method kotlin.jvm.internal.Intrinsics.checkParameterIsNotNull, parameter uri
    at mozilla.components.browser.engine.gecko.GeckoEngineSession$createNavigationDelegate$1.onLoadError
    at org.mozilla.geckoview.GeckoSession$2.handleMessage(GeckoSession.java:345)
    at org.mozilla.geckoview.GeckoSession$2.handleMessage(GeckoSession.java:176)
    at org.mozilla.geckoview.GeckoSessionHandler.handleMessage(GeckoSessionHandler.java:79)
    at org.mozilla.gecko.EventDispatcher$3.run(EventDispatcher.java:356)
    at android.os.Handler.handleCallback(Handler.java:873)
    at android.os.Handler.dispatchMessage(Handler.java:99)
    at android.os.Looper.loop(Looper.java:193)
    at android.app.ActivityThread.main(ActivityThread.java:6680)
    at java.lang.reflect.Method.invoke(Method.java)
    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
```

csadilek:
See https://github.com/mozilla-mobile/android-components/issues/1383#issuecomment-440468994 for all details.

The first commit makes sure we tolerate a MALFROMED_URI error from GeckoView. We do this because our ACTION_SEND handling problem might have caused malformed URIs to be persisted, and more importantly, to make sure we don't crash should a similar problem occur in the future. For engine-gecko release this already worked coincidentally.

The second commit fixes the cause of the malformed URL. When processing ACTION_SEND intents with EXTRA_TEXT that is not a URL, we no longer create a session with that text as URL. It also fixes the problem that the search wasn't properly triggered (using GeckoView) because of the two `loadUri` calls in quick succession. And while we're at it :), this also fixes the behaviour of how to decide whether to load the URL or search to align with Focus/Fennec.

### DIFF
--- a/components/feature/intent/src/main/java/mozilla/components/feature/intent/IntentProcessor.kt
+++ b/components/feature/intent/src/main/java/mozilla/components/feature/intent/IntentProcessor.kt
@@ -67,15 +67,16 @@ class IntentProcessor(
         when {
             TextUtils.isEmpty(extraText.trim()) -> false
 
-            extraText.isUrl() -> {
-                val session = createSession(extraText, source = Source.ACTION_SEND)
-                sessionUseCases.loadUrl.invoke(extraText, session)
-                true
-            }
             else -> {
-                val session = createSession(extraText, source = Source.ACTION_SEND)
-                searchUseCases.defaultSearch.invoke(extraText, session)
-                true
+                val url = extraText.split(" ").find { it.isUrl() }
+                if (url != null) {
+                    val session = createSession(url, source = Source.ACTION_SEND)
+                    sessionUseCases.loadUrl.invoke(url, session)
+                    true
+                } else {
+                    searchUseCases.defaultSearch.invoke(extraText, Source.ACTION_SEND, openNewTab)
+                    true
+                }
             }
         }
     }
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -276,7 +276,7 @@ class GeckoEngineSession(
 
         override fun onLoadError(
             session: GeckoSession?,
-            uri: String,
+            uri: String?,
             category: Int,
             error: Int
         ): GeckoResult<String> {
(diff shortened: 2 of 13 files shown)


components/feature/intent/src/main/java/mozilla/components/feature/intent/IntentProcessor.kt
```diff
@@ -67,15 +67,16 @@ class IntentProcessor(
...
-            extraText.isUrl() -> {
-                val session = createSession(extraText, source = Source.ACTION_SEND)
-                sessionUseCases.loadUrl.invoke(extraText, session)
-                true
-            }
             else -> {
-                val session = createSession(extraText, source = Source.ACTION_SEND)
-                searchUseCases.defaultSearch.invoke(extraText, session)
-                true
+                val url = extraText.split(" ").find { it.isUrl() }
```
pocmo: Note that `isUrl()` is a very basic implementation to only determine whether something entered into the URL bar should trigger a search or a load. I doesn't actually look for somewhat valid URLs. I fear that it will do unexpected things. For example if I share the text `Hello world.` then from the implementation of `isUrl()` I guess it would try to load `world.`.

Fennec has a more complex `WebURLFinder` for that:
* https://searchfox.org/mozilla-central/source/mobile/android/services/src/main/java/org/mozilla/gecko/sync/setup/activities/WebURLFinder.java
* https://searchfox.org/mozilla-central/source/mobile/android/services/src/test/java/org/mozilla/gecko/sync/setup/activities/WebURLFinderTest.java

Not something we would need to fix now though; we could also look at landing something like `WebURLFinder` in a follow-up.

csadilek: OK, I've filed a follow-up: https://github.com/mozilla-mobile/android-components/issues/1438

csadilek:
Sill need to push a test update here.

PR #1437, opened 2018-11-21
Closes #1434: Add support for choice html element for GeckoEngine

### LINKED ISSUE #1434: Add support for select html element for GeckoEngine
User Story: https://github.com/mozilla-mobile/reference-browser/issues/223

Amejia481:
This is a work in progress but I want to have your early feedback because I'm not sure if I'm in the right track with the architecture.

Right now, I'm taking the same route as `feature-downlads` implementing the UI on a feature(`PromptFeature`) and trigger it when is notify  from the session observer.

As far as I know, this functionality is not consumable from the WebView side. I will be only implemented on Gecko.

[Sample Browser using `PromptFeature`](https://drive.google.com/file/d/1lwgEAiEsIzKTNNfnTQC2M8kkGbvz3dz6/view)


Amejia481:
cc @csadilek  @pocmo

pocmo:
For choice/select specifically: Since this is a basic engine feature and since `WebView` just does it on its own too: Could we add this to `browser-engine-gecko-*` directly (assuming we have everything we need there)? Right now I do not see any architectural reason for moving this out to a feature.

For some of the other prompt callbacks we will need to handle them outside though. :)

Amejia481:
To create a dialog inside `browser-engine-gecko-*`, I need a `Activity context` and from `GeckoEngineSession` and `GeckoEngine` I'm getting a `application context`. Without a feature implementation can't find a way to get the `Activity context` any suggestions?

image

Amejia481:
PR updated!

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/prompt/GeckoPromptDelegate.kt
```diff
@@ -0,0 +1,178 @@
...
+import org.mozilla.geckoview.GeckoSession.PromptDelegate.FileCallback
+import org.mozilla.geckoview.GeckoSession.PromptDelegate.TextCallback
+
+typealias GeckoChoice = GeckoSession.PromptDelegate.Choice
+
+/**
+ * Gecko-based PromptDelegate implementation.
+ */
+@Suppress("TooManyFunctions")
+class GeckoPromptDelegate(private val geckoEngineSession: GeckoEngineSession) :
```
pocmo: nit: `internal`. This class is not part of the public API.

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/prompt/GeckoPromptDelegate.kt
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    override fun onButtonPrompt(
+        session: GeckoSession?,
+        title: String?,
+        msg: String?,
+        btnMsg: Array<out String>?,
+        callback: ButtonCallback?
+    ) {
+    }
```
pocmo: nit: `= Unit` avoids you to have empty blocks :)

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/prompt/GeckoPromptDelegate.kt
```diff
@@ -0,0 +1,178 @@
...
+        session: GeckoSession?,
+        title: String?,
+        msg: String?,
+        value: String?,
+        callback: TextCallback?
+    ) {
+    }
+
+    override fun onPopupRequest(session: GeckoSession?, targetUri: String?): GeckoResult<AllowOrDeny> {
+        TODO("not implemented")
```
pocmo: Let's not use `TODO()` because this will throw an exception when called.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -489,8 +490,11 @@ class GeckoEngineSession(
         geckoSession.contentDelegate = createContentDelegate()
         geckoSession.trackingProtectionDelegate = createTrackingProtectionDelegate()
         geckoSession.permissionDelegate = createPermissionDelegate()
+        geckoSession.promptDelegate = createPromptDelegate()
     }
 
+    private fun createPromptDelegate() = GeckoPromptDelegate(this)
```
pocmo: Do we need this method or could we create the delegate in `initGeckoSession` directly? Afaik the other create() methods only exist because we are creating all the instances inline and that separates the code.

components/concept/engine/src/main/java/mozilla/components/concept/engine/prompt/Choice.kt
```diff
@@ -0,0 +1,63 @@
...
+    var label: String,
+    var selected: Boolean = false,
+    val isASeparator:
+    Boolean = false,
+    val children: Array<Choice>? = null
+) : Parcelable {
+
+    val isGroupType get() = children != null
+
+    constructor(parcel: Parcel) : this(
```
pocmo: `internal`?

components/concept/engine/src/main/java/mozilla/components/concept/engine/prompt/Choice.kt
```diff
@@ -0,0 +1,63 @@
...
+ * @property label The label for displaying the option, optgroup or menuitem.
+ * @property selected Indicate if the item should be pre-selected.
+ * @property isASeparator Indicating if the item should be a menu separator (only valid for menus).
+ * @property children Sub-items in a group, or null if not a group.
+ */
+data class Choice(
+    val id: String,
+    var enable: Boolean = true,
+    var label: String,
+    var selected: Boolean = false,
```
pocmo: Do those need to be mutable?

components/concept/engine/src/test/java/mozilla/components/concept/engine/prompt/PromptRequestTest.kt
```diff
@@ -0,0 +1,14 @@
+package mozilla.components.concept.engine.prompt
```
pocmo: nit: license

components/feature/session/src/main/java/mozilla/components/feature/session/prompt/ChoiceAdapter.kt
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    override fun onCreateViewHolder(parent: ViewGroup, type: Int): ChoiceViewHolder {
+        return ChoiceViewHolder(inflater.inflate(getLayoutId(type), parent, false))
+    }
+
+    override fun getItemCount(): Int = choices.size
+
+    @Suppress("ComplexMethod")
+    override fun onBindViewHolder(holder: ChoiceViewHolder, position: Int) {
```
pocmo: This is indeed a complex method. I wonder if it would be simpler if we would use multiple view holders types?

components/feature/session/src/main/java/mozilla/components/feature/session/prompt/ChoiceDialogFragment.kt
```diff
@@ -0,0 +1,135 @@
...
+        }
+        return view
+    }
+
+    fun onSelect(selectedChoice: Choice) {
+        feature?.onSingleChoiceSelect(sessionId, selectedChoice)
+
+        if (!isATest) {
+            dismiss()
+        }
```
pocmo: This is dangerous since you start testing something else than what is running in production. If you do not want to call `dismiss()` in tests, can you stub it? E.g.:

```Kotlin
val fragment = spy(ChoiceDialogFragment.newInstance(..))
doNothing().when(fragment).dismiss()

...
```

(See `ContextMenuFragmentTest`).

components/feature/session/src/main/java/mozilla/components/feature/session/prompt/PromptFeature.kt
```diff
@@ -0,0 +1,142 @@
...
+ * [Session.Observer.onPromptRequested] events. Once the dialog is closed or the user selects an item from the dialog
+ * the related [PromptFeature] will be consumed.
+ *
+ * @property sessionManager The [SessionManager] instance in order to subscribe to the selected [Session].
+ * @property fragmentManager The [FragmentManager] to be used when displaying a dialog (fragment).
+ */
+class PromptFeature(
+    private val sessionManager: SessionManager,
+    private val fragmentManager: FragmentManager
+) : SelectionAwareSessionObserver(sessionManager) {
```
pocmo: nit: I'd prefer to implement the observer in a separate property. Otherwise the methods on `SelectionAwareSessionObserver`, `SessionManager.Observer` and `Session.Observer` will all be part of the public API of the feature class.

components/feature/session/src/main/java/mozilla/components/feature/session/prompt/PromptFeature.kt
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    /**
+     * Event that is triggered when a native dialog needs to be shown.
+     * Displays suitable dialog for the type of the [promptRequest].
+     *
+     * @property session The session the request the dialog.
+     * @property promptRequest The session the request the dialog.
+     */
+    override fun onPromptRequested(session: Session, promptRequest: PromptRequest): Boolean {
```
pocmo: nit: `internal`?

components/feature/session/src/main/java/mozilla/components/feature/session/prompt/PromptFeature.kt
```diff
@@ -0,0 +1,142 @@
...
+        dialog.show(fragmentManager, FRAGMENT_TAG)
+        return false
+    }
+
+    /**
+     * Event that is triggered when a single choice or menu item is selected in a dialog.
+     * @property sessionId the id for which session was selected.
+     * @property choice the selection from the dialog.
+     */
+    fun onSingleChoiceSelect(sessionId: String, choice: Choice) {
```
pocmo: nit: `internal`?

components/feature/session/src/main/java/mozilla/components/feature/session/prompt/PromptFeature.kt
```diff
@@ -0,0 +1,142 @@
...
+            true
+        }
+    }
+
+    /**
+     * Event that is triggered when a multiple choice items are selected in a dialog.
+     * @property sessionId the id for which session was selected.
+     * @property choices the selected items from the dialog.
+     */
+    fun onMultipleChoiceSelect(sessionId: String, choices: Array<Choice>) {
```
pocmo: nit: `internal`?

components/feature/session/src/main/java/mozilla/components/feature/session/prompt/PromptFeature.kt
```diff
@@ -0,0 +1,142 @@
...
+            true
+        }
+    }
+
+    /**
+     * Event that is called when a dialog is dismissed.
+     * This consumes the [PromptFeature] value from the [Session] indicated by [sessionId].
+     * @property sessionId the id for which session was dismissed.
+     */
+    fun onCancel(sessionId: String) {
```
pocmo: nit: `internal`?

components/feature/session/src/main/res/layout/mozac_feature_choice_group_item.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?><!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+
+<TextView xmlns:android="http://schemas.android.com/apk/res/android"
+          style="?android:attr/listSeparatorTextViewStyle"
+          android:id="@android:id/text1"/>
```
pocmo: Let's prefer using our own Ids?

components/feature/session/src/main/res/layout/mozac_feature_menu_choice_item.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<TextView
+        xmlns:android="http://schemas.android.com/apk/res/android"
+        android:id="@android:id/text1"
```
pocmo: Let's prefer using our own Ids?

components/feature/session/src/main/res/layout/mozac_feature_menu_separator_choice_item.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<View xmlns:android="http://schemas.android.com/apk/res/android"
+      android:id="@android:id/text1"
```
pocmo: Let's prefer using our own Ids?

components/feature/session/src/main/res/layout/mozac_feature_multiple_choice_item.xml
```diff
@@ -0,0 +1,21 @@
...
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+
+<LinearLayout
+        xmlns:android="http://schemas.android.com/apk/res/android"
+        android:layout_width="match_parent"
+        android:layout_height="?android:attr/listPreferredItemHeightSmall"
+        android:background="?android:attr/selectableItemBackground">
+
+    <CheckedTextView
+            android:id="@android:id/text1"
```
pocmo: Let's prefer using our own Ids?

components/feature/session/src/main/res/layout/mozac_feature_single_choice_item.xml
```diff
@@ -0,0 +1,20 @@
...
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+
+<LinearLayout
+        xmlns:android="http://schemas.android.com/apk/res/android"
+        android:layout_width="match_parent"
+        android:layout_height="?android:attr/listPreferredItemHeightSmall"
+        android:background="?android:attr/selectableItemBackground">
+
+    <CheckedTextView
+            android:id="@android:id/text1"
```
pocmo: Let's prefer using our own Ids?

components/feature/session/src/main/java/mozilla/components/feature/session/prompt/PromptFeature.kt
```diff
@@ -0,0 +1,142 @@
...
+ * input type date,file,time,color, option, menu, authentication, confirmation and other alerts.
+ *
+ * This feature will subscribe to the currently selected [Session] and display a suitable native dialog based on
+ * [Session.Observer.onPromptRequested] events. Once the dialog is closed or the user selects an item from the dialog
+ * the related [PromptFeature] will be consumed.
+ *
+ * @property sessionManager The [SessionManager] instance in order to subscribe to the selected [Session].
+ * @property fragmentManager The [FragmentManager] to be used when displaying a dialog (fragment).
+ */
+class PromptFeature(
```
pocmo: I wonder if we should move this into its own feature (like `feature-prompt`). Mostly because some of those prompt implementation that we are going to build may have more dependencies on other things and then we wouldn't want to add those to `browser-session`.

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/prompt/GeckoPromptDelegate.kt
Amejia481: 👍

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/prompt/GeckoPromptDelegate.kt
Amejia481: Yeah, this will be implemented in coming PRs. How should I handle it ?

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/prompt/GeckoPromptDelegate.kt
Amejia481: 👍 I will return a dummy value!

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/prompt/GeckoPromptDelegate.kt
pocmo: Yeah, that's fine. `) = Unit` looks more like it is an empty implementation on purpose instead of "Oops, forgot the code" :D But feel free to land as-is or throw in a comment linking to the issue if there's one already.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
Amejia481: You are right. First I did the inline implementation but it was to much to be in the same class, I just forgot about `createPromptDelegate()`

components/concept/engine/src/main/java/mozilla/components/concept/engine/prompt/Choice.kt
Amejia481: 👍

components/concept/engine/src/main/java/mozilla/components/concept/engine/prompt/Choice.kt
Amejia481: Yes, I'm mutating them when is multiple choice dialog, instead of creating new new objects.

https://github.com/mozilla-mobile/android-components/blob/ffdd513f100c0c119940694f728d5279d3cc4144/components/feature/session/src/main/java/mozilla/components/feature/session/prompt/ChoiceAdapter.kt#L125

image

components/feature/session/src/main/java/mozilla/components/feature/session/prompt/ChoiceDialogFragment.kt
Amejia481: Thanks I didn't know about `doNothing()`!

components/feature/session/src/main/res/layout/mozac_feature_choice_group_item.xml
Amejia481: I'm using Android's  String should I also created ours too?
```
android.R.string.cancel
android.R.string.ok
```

components/feature/session/src/main/java/mozilla/components/feature/session/prompt/PromptFeature.kt
Amejia481: Agree! Do you have an opinions @csadilek ?

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/prompt/GeckoPromptDelegate.kt
Amejia481: 👍

components/feature/session/src/main/res/layout/mozac_feature_choice_group_item.xml
Amejia481: I updated it to use our strings too!

components/feature/session/src/main/java/mozilla/components/feature/session/prompt/PromptFeature.kt
Amejia481: I moved it to its own feature, if you have any suggestions just let me know.

components/feature/session/src/main/java/mozilla/components/feature/session/prompt/ChoiceAdapter.kt
Amejia481: Fixed. Yes, it really helped!

PR #1440, opened 2018-11-21
[WIP] Use new combined fxa config

vbudhram:
This PR connects with https://github.com/mozilla/application-services/pull/403 and updates the sample FxA application to use the new config. 

A new config objected is created with 
```
Config.custom(CONFIG_URL, CLIENT_ID, REDIRECT_URL).await()...

val acct = FirefoxAccount(config)...
```

With the introduction of https://github.com/mozilla/application-services/pull/388, the project wouldn't compile. I briefly looked at fixing the issues but wasn't able to get it properly updated so I just commented out the code that broke.

Depends on https://github.com/mozilla-mobile/android-components/pull/1354

cc @eoger


csadilek:
@vbudhram great! Just a heads-up that the sync-logins demo would also need this update: https://github.com/mozilla-mobile/android-components/blob/master/samples/sync-logins/src/main/java/org/mozilla/samples/sync/logins/MainActivity.kt#L70

eoger:
This should be closed in favor of 1453

PR #1448, opened 2018-11-21
Closes #739: Move SyncLoginsClient to services-sync-logins

### LINKED ISSUE #739: Move MinimalLoginsExample to services-sync-logins
Ref: https://github.com/mozilla-mobile/android-components/pull/676/files#r215347368

csadilek:
@thomcc @vladikoff As discussed, just moving SyncLoginsClient into the lib and some related cleanup. 

Keeping this small as we'll have to do another pass once the FxA PRs land.


components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/SyncLoginsClient.kt
```diff
@@ -0,0 +1,62 @@
+package mozilla.components.service.sync.logins
+
+import mozilla.appservices.logins.ServerPassword
+import mozilla.appservices.logins.SyncUnlockInfo
+
+/**
+ * Entrypoint for integrating with Firefox Sync for logins.
```
thomcc: Probably it would be better to indicate that this is a helper class. Users who don't need its functionality are free to ignore it.

Just changing "Entrypoint" to "Helper" is probably enough.

samples/sync-logins/src/main/java/org/mozilla/samples/sync/logins/MainActivity.kt
```diff
@@ -83,7 +92,12 @@ open class MainActivity : AppCompatActivity(), LoginFragment.OnLoginCompleteList
 
     override fun onDestroy() {
         super.onDestroy()
-        runBlocking { whenAccount.await().close() }
+
+        if (whenAccount.isCompleted) {
+            whenAccount.getCompleted().close()
+        } else {
+            whenAccount.cancel()
```
thomcc: Ditto.

components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/SyncLoginsClient.kt
```diff
@@ -0,0 +1,62 @@
...
+            tokenserverURL = tokenServerURL
+        )
+
+        val store = getUnlockedStore(encryptionKey)
+        store.sync(unlockInfo).await()
+
+        return getLocalPasswordList(encryptionKey)
+    }
+
+    private suspend fun getLocalPasswordList(encryptionKey: String): List<ServerPassword> {
```
thomcc: This probably should be public, most users will want it, even if the sample app doesn't need it.

PR #1449, opened 2018-11-22
 Closes #1446: Giving correct parameters to guessFileName when a file name is null in Gecko Release/nightly/beta

### LINKED ISSUE #1446: APK download uses "downloadfile.bin" as file name
I wanted to download the reference browser from the reference browser (inception!)

But it didn't use the file name and instead offered to download "downloadfile.bin".

screenshot_20181120-195001

Amejia481:


Amejia481:
👍

PR #1454, opened 2018-11-23
Closes #1451: Provide tab filter to TabsFeature.

### LINKED ISSUE #1451: Add ability to provide a filter to the TabsFeature
We may want to show certain tabs based on our own filter. 

e.g. Providing this would allow us to tell our TabsTray when to display private and when not to depending a button clicked.

jonalmeida:
~~Also exposes an invalidate so that clients can call it when they expect
to see changes.~~

We have a filter method on `TabsFeature` that allows us to filter the sessions and update the TabsTray accordingly.

**Q. Why do this instead of just adding a 'private' flag?**
**A:** It's more versatile to pass in a filter. Currently we may have three possible options: all, regular, private. With this alone a boolean doesn't work.
In the future, this allows new features for tab grouping to be done by the application by their own definition.

### DIFF
--- a/components/feature/tabs/src/main/java/mozilla/components/feature/tabs/tabstray/TabsFeature.kt
+++ b/components/feature/tabs/src/main/java/mozilla/components/feature/tabs/tabstray/TabsFeature.kt
@@ -4,6 +4,8 @@
 
 package mozilla.components.feature.tabs.tabstray
 
+import android.support.annotation.VisibleForTesting
+import mozilla.components.browser.session.Session
 import mozilla.components.browser.session.SessionManager
 import mozilla.components.concept.tabstray.TabsTray
 import mozilla.components.feature.tabs.TabsUseCases
@@ -17,12 +19,15 @@ class TabsFeature(
     tabsUseCases: TabsUseCases,
     closeTabsTray: () -> Unit
 ) {
-    private val presenter = TabsTrayPresenter(
+    @VisibleForTesting
+    internal var presenter = TabsTrayPresenter(
         tabsTray,
         sessionManager,
-        closeTabsTray)
+        closeTabsTray
+    )
 
-    private val interactor = TabsTrayInteractor(
+    @VisibleForTesting
+    internal var interactor = TabsTrayInteractor(
         tabsTray,
         tabsUseCases.selectSession,
         tabsUseCases.removeSession,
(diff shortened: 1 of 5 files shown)


components/feature/tabs/src/main/java/mozilla/components/feature/tabs/tabstray/TabsFeature.kt
```diff
@@ -15,12 +16,14 @@ class TabsFeature(
     tabsTray: TabsTray,
     sessionManager: SessionManager,
     tabsUseCases: TabsUseCases,
-    closeTabsTray: () -> Unit
+    closeTabsTray: () -> Unit,
+    filter: (Session) -> Boolean = { true }
```
csadilek: nit: naming `sessionFilter`

components/feature/tabs/src/main/java/mozilla/components/feature/tabs/tabstray/TabsFeature.kt
```diff
@@ -37,4 +40,8 @@ class TabsFeature(
         presenter.stop()
         interactor.stop()
     }
+
+    fun invalidate() {
```
csadilek: Naming: We're not "invalidating" the feature. We just want it to update the displayed sessions / invalidate the UI.

Testing: This doesn't seem to be covered by a test?

Design: I think providing a lambda to filter sessions is fine, but the fact that we need a public API call on the feature to make it re-render the UI seems a bit weird. It seems the presenter should be able to observe some state and decide that it needs to that instead of getting nudge from the outside? But maybe that's for a future refactoring. Just a thought.

jonalmeida: > Design: I think providing a lambda to filter sessions is fine, but the fact that we need a public API call on the feature to make it re-render the UI seems a bit weird. It seems the presenter should be able to observe some state and decide that it needs to that instead of getting nudge from the outside? But maybe that's for a future refactoring. Just a thought.

We talked about this offline and decided the best way to avoid having to expose and 'invalidate' is to pass in the filter to feature when we need it to be invoked. So instead of providing the filter in the constructor AND exposing an invalidation method, we just have an `filterTabs` method that takes a filter and invokes the invalidation after that.

jonalmeida:
> OpenJDK 64-Bit Server VM warning: INFO: os::commit_memory(0x0000000723853000, 158695424, 0) failed; error='Cannot allocate memory' (errno=12)

Restarting.

Issue #1457, opened 2018-11-23
Add new TabsUseCases for closing all sessions


jonalmeida:
We need a use case that lets you close:
- all
- regular tabs only
- private tabs only

pocmo:
FYI: In the context menu PR that I'm going to open now I renamed some of the use case classes inside `TabsUseCases`. There may be conflicts! :)

chikka:
@jonalmeida @pocmo can i take it up?

jonalmeida:
@chikka Definitely! Although you should wait for #1459 to get merged. If you're keen to work on this ASAP, then you can work off that branch.

Have a look at [`TabsUseCases.kt`](https://github.com/mozilla-mobile/android-components/blob/master/components/feature/tabs/src/main/java/mozilla/components/feature/tabs/TabsUseCases.kt) to get an idea of how to write one.

chikka:
@jonalmeida Thanks!. I have looked into it and I think we need to create two classes for it

   ```
 class RemoveAllTabsUseCase internal constructor(
        private val sessionManager: SessionManager
    ) {
       
        fun invoke() {
            
        }
    }
```

 ```
   class RemoveTabsUseCase internal constructor(
        private val sessionManager: SessionManager
    ) {
       
//send privateOnly=true for closing only private tabs otherwise false to close all normal tabs
        fun invoke(privateOnly: Boolean = false) {
            
        }
    }
```
Please let me know your thoughts on it

jonalmeida:
@chikka looks great! Feel free to create a PR and we talk review it there as well.

Small nit:
```kotlin
fun invoke(privateOnly: Boolean = false) { }
```
I'd change the attribute name from `privateOnly` to `private`. It may imply that it would remove _all_ tabs vs only private tabs.

PR #1460, opened 2018-11-23
close #1270: Updated BrowserToolbar.ToggleButton and BrowserToolbar.Button .

### LINKED ISSUE #1270: Set Padding only once in BrowserToolbar's Button and ToggleButton
In both BrowserToolbar classes, we're re-setting padding twice, once here and then in the super.

We should fix this so that we can have a default padding for the Browser buttons while still making the padding optional in the Toolbar buttons.

chikka:
Removed redundant setPadding and  overriding methods


jonalmeida:
🎉 

Thanks @chikka! If you're interested in trying out another issue I filed #1457 recently as well that's a little bit more challenging (only a little!) than this one. :)

Otherwise, check out our '[help wanted][0]' label if you're interested in something.

[0]: https://github.com/mozilla-mobile/android-components/labels/help%20wanted

Issue #842, opened 2018-09-22
Migrate to androidx


pocmo:
https://developer.android.com/topic/libraries/support-library/androidx-rn?linkId=57203177

- [ ] Remove Custom Tabs Jetifier hack: https://github.com/mozilla-mobile/android-components/issues/1884

pocmo:
I tested this with our reference browser: It's a pain and doesn't work yet.

After changing the dependencies and all imports I had to use a bleeding edge (alpha) Android gradle plugin and the latest Gradle version to get it working.

And at the end I got stuck with:

```
> Transform browser.aar (androidx.browser:browser:1.0.0-rc01) with DexingTransform
AGPBI: {"kind":"error","text":"Class descriptor \u0027Landroid/support/customtabs/ICustomTabsCallback/;\u0027 cannot be represented in dex format.","sources":[{}],"tool":"D8"}

> Task :app:mergeExtDexGeckoBetaArmDebug FAILED
AGPBI: {"kind":"error","text":"Class descriptor \u0027Landroid/support/customtabs/ICustomTabsCallback/;\u0027 cannot be represented in dex format.","sources":[{}],"tool":"D8"}

FAILURE: Build failed with an exception.

* What went wrong:
Could not resolve all files for configuration ':app:geckoBetaArmDebugRuntimeClasspath'.
> Failed to transform file 'jetified-classes.jar' to match attributes {artifactType=android-dex, dexing-is-debuggable=true, dexing-min-sdk=21} using transform DexingTransform
   > Error while dexing.
```

jonalmeida:
There's no solution here, but at least we're not the only ones running into this? 

https://stackoverflow.com/questions/52639577/class-descriptor-landroid-support-customtabs-icustomtabscallback-cannot-be-r

pocmo:
I'm thinking about moving the service to a feature component for a variety of reasons. Currently the service doesn't do match. As soon as we implement more we may not want that to stay in `browser-session`.

Amejia481:
We also  should a look at `androidx.webkit.WebViewCompat`

References:
[WebViewCompat class](https://developer.android.com/reference/androidx/webkit/WebViewCompat)
[Modern WebView Best Practices Android Dev Summit '18](https://youtu.be/HGZYtDZhOEQ)

pocmo:
Update:

On AndroidX:
* Reference Browser
* Fenix
* Firefox Reality
* Lockbox

On support libraries:
[...]

PR #1475, opened 2018-11-26
[WIP] Update to Robolectric 4.0.2

### LINKED ISSUE #1176: Migrate to Robolectric 4.0
http://robolectric.org/blog/2018/10/25/robolectric-4-0/

Dexterp37:
This attempts to fix #1176 .

This is currently broken for two problems in glean which I'm not sure how to fix, so any input is more than welcome!

** Problem 1 **

>     [Robolectric] NOTICE: legacy resources mode is deprecated; see http://robolectric.org/migrating/#migrating-to-40

** Problem 2 **

>DefNotFound android/arch/lifecycle/ReportFragment$ActivityInitializationListener

I think this has probably something to do with the suggested migration to `androidx`, see [the migration docs](http://robolectric.org/migrating/#migrating-to-40).

### DIFF
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -7,8 +7,9 @@ private object Versions {
     const val kotlin = "1.3.10"
     const val coroutines = "1.0.1"
 
+    const val androidx_test = "1.0.0"
     const val junit = "4.12"
-    const val robolectric = "4.0-alpha-3"
+    const val robolectric = "4.0.2"//""4.0-alpha-3"
     const val mockito = "2.23.0"
 
     const val mockwebserver = "3.10.0"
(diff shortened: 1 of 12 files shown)


Dexterp37:
@pocmo this sounds a bit more involved than initially anticipated. I'll drop this for now and use a workaround for my tests in Glean.

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -56,7 +58,7 @@ object Dependencies {
     const val support_constraintlayout = "com.android.support.constraint:constraint-layout:${Versions.constraint_layout}"
     const val support_compat = "com.android.support:support-compat:${Versions.support_libraries}"
 
-    const val arch_lifecycle = "android.arch.lifecycle:extensions:${Versions.lifecycle}"
+    const val arch_lifecycle = "androidx.lifecycle:lifecycle-extensions:${Versions.lifecycle}"
```
pocmo: This is definitely something we won't be able to push at this time: When migrating this to AndroidX then we will have to migrate all other dependencies to AndroidX too. And this will force all apps to migrate too. This is too aggressive and we need to get existing apps to migrate first before we can start using AndroidX in our existing components (No problem for glean though!). #842

pocmo:
Picking this up and continuing in #1479.

PR #1479, opened 2018-11-26
Use Robolectric 4.0.2

pocmo:
On top of #1463. Continuing #1475.


pocmo:
This seems to pass now.

@Dexterp37 Can you verify the glean changes? Those should be mostly from your PR. I only had to go back to lifecycle 1.1.1 for now.

pocmo:
> @pocmo , before merging this, I have a question: you mentioned that other apps would need to use androidx as well. Would having this requirement for glean be a sane thing? We don't really want to have an high barrier to its use.

* In the current version we only use androidx in testing.

* Going forward: Yes, I absolutely think glean should go with androidx from the beginning. The reference browser is already migrated to use androidx and eventually the other apps will have to migrate too. We have to hold back for our existing components to not force the migration on the apps *right now*. But glean isn't integrated in any existing apps yet. And new apps can start with androidx from the beginning (like the reference browser does now).

FYI: I'm tracking various migrations like androidx in this spreadsheet:
https://docs.google.com/spreadsheets/d/18nRwYboUouBWn3jnrnjIn8dynIjD2Ie45oLbphwRbvs/edit#gid=0

PR #1484, opened 2018-11-27
close #1457 add tabsusecases for removing all/privateonly/nonprivate sessions

### LINKED ISSUE #1457: Add new TabsUseCases for closing all sessions

chikka:
add tabsusecases for removing all/privateonly/nonprivate sessions


components/feature/tabs/src/main/java/mozilla/components/feature/tabs/TabsUseCases.kt
```diff
@@ -93,8 +93,48 @@ class TabsUseCases(
...
+            sessionManager.removeAll()
+
+        }
+    }
+
+    class RemovePrivateTabsUseCase internal constructor(
+            private val sessionManager: SessionManager
+    ) {
+
+        fun invoke() {
```
jonalmeida: You could consider pass a boolean argument to the invoke function to differ between removing regular/private tabs.

components/feature/tabs/src/main/java/mozilla/components/feature/tabs/TabsUseCases.kt
```diff
@@ -93,8 +93,48 @@ class TabsUseCases(
         }
     }
 
+    class RemoveAllTabsUseCase internal constructor(
```
jonalmeida: If you could write a few basic tests for these use cases as well, that would be great too. (: 

We have some handy ones in [`TabsUseCasesTest.kt`](https://github.com/mozilla-mobile/android-components/blob/master/components/feature/tabs/src/test/java/mozilla/components/feature/tabs/TabsUseCasesTest.kt) for reference.

components/feature/tabs/src/main/java/mozilla/components/feature/tabs/TabsUseCases.kt
chikka: ok.will add

components/feature/tabs/src/main/java/mozilla/components/feature/tabs/TabsUseCases.kt
chikka: sure

components/feature/tabs/src/main/java/mozilla/components/feature/tabs/TabsUseCases.kt
chikka: sure

PR #1487, opened 2018-11-27
[WIP] [glean] 1507743: Add an internal metrics.yaml

mdboom:
This uses an internal `metrics.yaml` to generate metrics for the baseline ping and one of the `ping_info` entries.

There are some comments inline below.  This is basically blocked by some other work, but I thought it would be good to get feedback early, and potentially we want to merge it as-is anyway before we're ready to implement all of the baseline ping stuff.


components/service/glean/metrics.yaml
```diff
@@ -0,0 +1,96 @@
+# This file defines the metrics that are recorded by glean telemetry. They are
+# automatically converted to Kotlin code at build time using the `glean_parser`
+# PyPI package.
+
+$schema: moz://mozilla.org/schemas/glean/metrics/1-0-0
+
+baseline:
+  # sessions:
```
mdboom: These are commented out since we don't have the underlying metric types implemented yet.

components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -74,49 +74,57 @@ preBuild.dependsOn(createBuildDir)
...
-        description = "Generate the Kotlin code for the Metrics API"
-
-        workingDir  rootDir
-        // Note that the command line is OS dependant: on linux/mac is Miniconda3/bin/python.
-        if (Os.isFamily(Os.FAMILY_WINDOWS)) {
-            commandLine "$buildDir/bootstrap/Miniconda3/python"
-        } else {
-            commandLine "$buildDir/bootstrap/Miniconda3/bin/python"
-        }
+ext.gleanGenerateMetricsAPI = {
```
mdboom: This diff is less big than it seems -- it's mostly indenting to add the outer closure.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -86,6 +82,8 @@ object Glean {
 
         initializeCoreMetrics(applicationContext)
 
+        Baseline.os.set("Android")
```
mdboom: Just an example to test that we can use a generated metric.  This isn't really where we would do this.

components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -74,49 +74,57 @@ preBuild.dependsOn(createBuildDir)
...
+        // `sourceSets` task, for reasons unknown.
+        variant.registerJavaGeneratingTask(generateKotlinAPI, new File(sourceOutputDir))
+    }
 }
 
 if (android.hasProperty('applicationVariants')) {
-    android.applicationVariants.all(ext.gleanGenerateMetricsAPI)
+    android.applicationVariants.all(ext.gleanGenerateMetricsAPI('GleanMetrics'))
+} else {
+    android.libraryVariants.all(ext.gleanGenerateMetricsAPI('mozilla.components.service.glean.metrics'))
```
Dexterp37: What if we're collecting from a library? Is there any better way to assert that this is `glean`?

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
Dexterp37: Can we move this inside `initializeCoreMetrics`, right before the end?
Moreover, can we initialize (or comment out or remove) the other metrics defined in the `metrics.yaml` file? I'm fine with commenting them out and taking their initialization as part of a separate bug. Initializing these metrics could be part of [bug 1497894](https://bugzilla.mozilla.org/show_bug.cgi?id=1497894)

components/service/glean/metrics.yaml
Dexterp37: Let's add this as a comment before the probes. Also mention the bug numbers they depend on (e.g. 1497894)

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
Dexterp37: Ah, ok. Let's either remove it or leave a comment stating that.

components/service/glean/scripts/sdk_generator.gradle
mdboom: Good question. I'll have to go spelunking into gradle internals, I guess, though it might be enough to inspect `buildDir` which probably has `glean` near the end...

Dexterp37: Mh, you know what? Maybe this can be as easy as additionally checking if we have `GLEAN_PING_SCHEMA_URL` in `BuildConfig`. That should not be true for non glean, right?

mdboom: I couldn't get that to work, but it does seem you can do `project.name == 'service-glean'`, which I think should do the trick.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
mdboom: I've moved this to `initializeCoreMetrics` and commented out the other ones.

components/service/glean/scripts/sdk_generator.gradle
Dexterp37: Sounds good to me!

PR #1488, opened 2018-11-27
[glean] 1510308: Move experiments inside of ping_info.

mdboom:
This just moves the location of the `experiments` section of the ping inside of `ping_info`.

Schema side of the change is [here](https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/222)


components/service/glean/src/main/java/mozilla/components/service/glean/ping/PingMaker.kt
```diff
@@ -44,6 +45,8 @@ internal class PingMaker(
         pingInfo.put("ping_type", pingName)
         pingInfo.put("telemetry_sdk_build", BuildConfig.LIBRARY_VERSION)
 
+        pingInfo.put("experiments", ExperimentsStorageEngine.getSnapshotAsJSON("", false))
```
Dexterp37: Let's comment on why we need to special case the `ExperimentsStorageEngine` here.

mdboom:
I think the test failure in `feature-downloads` is unrelated.

PR #1490, opened 2018-11-27
closes #1457: added tabsusecases for all/private/normal tabs and related test cases

### LINKED ISSUE #1457: Add new TabsUseCases for closing all sessions

chikka:
added tabsusecases for all/private/normal tabs and related test cases


components/feature/tabs/src/main/java/mozilla/components/feature/tabs/TabsUseCases.kt
```diff
@@ -93,8 +93,38 @@ class TabsUseCases(
...
+            }
+        }
+        }
+
     val selectTab: SelectTabUseCase by lazy { SelectTabUseCase(sessionManager) }
     val removeTab: RemoveTabUseCase by lazy { RemoveTabUseCase(sessionManager) }
     val addTab: AddNewTabUseCase by lazy { AddNewTabUseCase(sessionManager) }
     val addPrivateTab: AddNewPrivateTabUseCase by lazy { AddNewPrivateTabUseCase(sessionManager) }
+    val removeAllTabs: RemoveAllTabsUseCase by lazy { RemoveAllTabsUseCase(sessionManager) }
+    val removeAllTabsOfTypeUseCase: RemoveAllTabsOfTypeUseCase by lazy {
```
jonalmeida: We don't need to add 'UseCase' to the attribute binding. :)

`removeAllTabsOfTypeUseCase` -> `removeAllTabsOfType`

chikka: ok.sure will do change

Issue #1195, opened 2018-10-29
Refactor browser-engine-system to keep WebView reference inside SystemEngineSession


pocmo:
### Plan

Refactor `browser-engine-system` to keep `WebView` reference inside `SystemEngineSession`. `SystemEngineView.render()` will attach/detach the internal `WebView` of the `SystemEngineSession`.

### Why?

This will simplify things a lot: Methods on `SystemEngineSession` can call directly into `WebView` even if it is not attached to a `SystemEngineView`. We do not need to deal with nullability and caching values until the next time we see a WebView. Rocket is using a similar approach and this verifies that this approach is feasible.

PR #1495, opened 2018-11-27
Closes #1067: Open / Close window support

### LINKED ISSUE #1067: Engine: Add open window / close window API
Websites can open links in new windows/tabs - via JavaScript or via `target="_blank"` links. We will need the necessary APIs for that to `EngineSession`:

WebView APIs:
* https://developer.android.com/reference/android/webkit/WebChromeClient#onCreateWindow(android.webkit.WebView,%20boolean,%20boolean,%20android.os.Message)
* https://developer.android.com/reference/android/webkit/WebChromeClient#onCloseWindow(android.webkit.WebView)

WebView's `onCreateWindow` API is potentially problematic: It requires to return a boolean to indicate whether the app is going to open a new window and has an asynchronous result message parameter.

csadilek:
This brings in new engine API to support opening and closing windows, as well as a feature that uses it.

This works for both `window.open` and `target="_blank"` (and closing windows), but only using engine-system. For Gecko, we're waiting for: https://bugzilla.mozilla.org/show_bug.cgi?id=1510314

This can be further cleaned up once #1195 lands, which is similar to what's being done here for window requests. `WindowRequest` has a prepare method which allows attaching state to an engine session. This will either be the new native `GeckoSession` or the newly created `WebView`.

The second commit exposes a new setting `supportMultipleWindows`, which I kept `false` by default for now, until we can implement the Gecko side of things. This way we don't need to update existing clients.

### DIFF
--- a/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/window/SystemWindowRequest.kt
+++ b/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/window/SystemWindowRequest.kt
@@ -1,0 +1,44 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.engine.system.window
+
+import android.os.Message
+import android.webkit.WebView
+import mozilla.components.browser.engine.system.SystemEngineSession
+import mozilla.components.concept.engine.EngineSession
+import mozilla.components.concept.engine.window.WindowRequest
+
+/**
+ * WebView-based implementation of [WindowRequest].
+ *
+ * @property webView the WebView from which the request originated.
+ * @property newWebView the WebView to use for opening a new window, may be null for close requests.
+ * @property openAsDialog whether or not the window should be opened as a dialog, defaults to false.
+ * @property triggeredByUser whether or not the request was triggered by the user, defaults to false.
+ * @property resultMsg the message to send to the new WebView, may be null.
+ */
+class SystemWindowRequest(
+    private val webView: WebView,
+    private val newWebView: WebView? = null,
+    val openAsDialog: Boolean = false,
+    val triggeredByUser: Boolean = false,
+    private val resultMsg: Message? = null
+) : WindowRequest {
+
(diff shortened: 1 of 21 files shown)


components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/window/SystemWindowRequest.kt
```diff
@@ -0,0 +1,44 @@
...
+ * @property newWebView the WebView to use for opening a new window, may be null for close requests.
+ * @property openAsDialog whether or not the window should be opened as a dialog, defaults to false.
+ * @property triggeredByUser whether or not the request was triggered by the user, defaults to false.
+ * @property resultMsg the message to send to the new WebView, may be null.
+ */
+class SystemWindowRequest(
+    private val webView: WebView,
+    private val newWebView: WebView? = null,
+    val openAsDialog: Boolean = false,
+    val triggeredByUser: Boolean = false,
```
pocmo: Currently we do not use `triggeredByUser`, right? I wonder whether we should think about this in a follow-up. `WindowRequest`s that are not triggered by any user action is something we probably want to block (popups). It would be interesting to know whether GeckoView does that by default; or whether we would want such a flag from them.

csadilek:
> Do you think we should schedule the SystemEngineSession refactoring for one of our next sprints?

Yes, agreed. I moved it to triage.

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/window/SystemWindowRequest.kt
csadilek: I think GV blocks by default, right? In Webview, we can control this using the `javaScriptCanOpenWindowsAutomatically` setting it seems. So, yes, I guess a follow-up to decide how we want to handle this. I will file one.

pocmo: Yeah, I talked to snorp. GV blocks by default. There's a strong indication that we should just drop everything where `triggeredByUser` is false.

csadilek: Made a note of it in the follow-up: https://github.com/mozilla-mobile/android-components/issues/1503

PR #1453, opened 2018-11-22
Config and tokens refactor.

eoger:
This supersedes #1440 and #1354.

PR #1496, opened 2018-11-28
Do not depend on JNA directly.

ncalexan:
I think this is just vestigial: as things developed, we expected to
use JNA directly.  Now all such access is an implementation detail of
the Application Services delivery vehicle.

The Proguard rules have been pushed to the application-services
repository, where they belong.


ncalexan:
This should be reviewed by @eoger, @vladikoff, or @thomcc as well, since there may be some subtle expectation about JNA access in a-c that I'm not aware of.  But I don't think so; everything should be in a-s, and transitive dependencies should Just Work.

ncalexan:
https://github.com/mozilla/application-services/pull/432 is a prerequisite for this work.

vladikoff:
AFAIU we bake JNA into our JARs now, so this makes sense

ncalexan:
> LGTM, now that we're happy with the a-s / a-c code split. However, we need a 0.12.0 release of a-s to get the proguard rules. So, I think we should wait for the before merging?

Yes, we should wait.  It's easy to cut new versions on the a-s side; let's do it as part of #1453, which will bump the version.

PR #1497, opened 2018-11-28
[glean] Run the API off the main thread by default.

Dexterp37:
This introduces the internal `Dispatchers` object, which allows to use a predefined scope for launching
functions off the main thread. Moreover, this also adds the test only `FakeDispatchersInTest` class, which allows to block on API calls during tests, ensuring they don't fail due to racing.

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Dispatchers.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Dispatchers.kt
@@ -1,0 +1,18 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.glean
+
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.ObsoleteCoroutinesApi
+import kotlinx.coroutines.newSingleThreadContext
+
+@ObsoleteCoroutinesApi
+internal object Dispatchers {
+    /**
+     * A coroutine scope to make it easy to dispatch API calls off the main thread.
+     * This needs to be a `var` so that our tests can override this.
+     */
+    var API = CoroutineScope(newSingleThreadContext("GleanAPIPool"))
+}
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/UuidMetricType.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/UuidMetricType.kt
@@ -4,6 +4,7 @@
 
 package mozilla.components.service.glean
 
+import kotlinx.coroutines.launch
 import java.util.UUID
 
 import mozilla.components.service.glean.storages.UuidsStorageEngine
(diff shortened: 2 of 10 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/Dispatchers.kt
```diff
@@ -0,0 +1,18 @@
...
+import kotlinx.coroutines.ObsoleteCoroutinesApi
+import kotlinx.coroutines.newSingleThreadContext
+
+@ObsoleteCoroutinesApi
+internal object Dispatchers {
+    /**
+     * A coroutine scope to make it easy to dispatch API calls off the main thread.
+     * This needs to be a `var` so that our tests can override this.
+     */
+    var APIScope = CoroutineScope(newSingleThreadContext("GleanAPIPool"))
```
pocmo: Nit: To follow the naming of Kotlin's `Dispatchers` class, I'd just use `API` here (without `Scope`).

components/service/glean/src/test/java/mozilla/components/service/glean/DispatchersTest.kt
```diff
@@ -0,0 +1,25 @@
...
+
+class DispatchersTest {
+
+    @Test
+    fun `API scope runs off the main thread`() {
+        val mainThread = Thread.currentThread()
+
+        Dispatchers.APIScope.launch {
+            assertNotSame(mainThread, Thread.currentThread())
+        }
```
pocmo: You may want to join() here. Otherwise this may not get executed before the end of the test. Maybe you even want to have some boolean var that you set to `true` inside so that you can verify that it did something at all.

components/service/glean/src/test/java/mozilla/components/service/glean/FakeDispatchersInTest.kt
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * A class that stubs the coroutine scopes in [Dispatchers] to allow blocking
+ * on async API calls when running tests.
+ *
+ * Note: according to [kotlinx.coroutines.Dispatchers.Unconfined], this will work
+ * as long as the API doesn't use any suspending function.
+ */
+@ExperimentalCoroutinesApi @ObsoleteCoroutinesApi
+open class FakeDispatchersInTest {
```
pocmo: A more idiomatic way would be to write this as a JUnit `Rule` that you can apply to tests:

* https://github.com/junit-team/junit4/wiki/rules
* https://medium.com/@vanniktech/writing-your-own-junit-rule-3df41997b10c

components/service/glean/src/main/java/mozilla/components/service/glean/UuidMetricType.kt
```diff
@@ -31,15 +32,26 @@ data class UuidMetricType(
...
+     * @return a [UUID] or [null] if we're not allowed to record.
      */
-    fun generateAndSet(): UUID {
+    fun generateAndSet(): UUID? {
+        // Even if `set` is already checking if we're allowed to record,
+        // we need to check here as well otherwise we'd return a `UUID`
+        // that won't be stored anywhere.
+        if (!shouldRecord(logger)) {
+            return null
+        }
```
mdboom: Good catch!

PR #1505, opened 2018-11-28
Bug 1508227 - Provide a better example for using Glean

travis79:
Added GleanApplication which extends Application and moved the Glean initialization from the Activity onCreate to the application onCreate.

### DIFF
--- a/samples/glean/src/main/java/org/mozilla/samples/glean/GleanApplication.kt
+++ b/samples/glean/src/main/java/org/mozilla/samples/glean/GleanApplication.kt
@@ -1,0 +1,28 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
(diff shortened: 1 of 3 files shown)


samples/glean/src/main/java/org/mozilla/samples/glean/GleanApplication.kt
```diff
@@ -0,0 +1,32 @@
...
+        // must be done right after enabling logging.
+        Glean.initialize(applicationContext, Configuration())
+
+        // Set a sample value for a metric.
+        GleanMetrics.Basic.os.set("Android")
+    }
+
+    companion object {
+        const val NON_FATAL_CRASH_BROADCAST = "org.mozilla.samples.glean"
+    }
```
pocmo: This companion object and the constant should not be needed here.

travis79:
Requested changes have been made.

pocmo:
Please squash that PR before merging. :)

travis79:
I believe I have squashed it (correctly) now.  Had to expand my git-fu a bit but I think I got it straightened out now.

PR #1525, opened 2018-11-29
Closes #1395: Implement CustomTabsService.warmup()..

### LINKED ISSUE #1395: CustomTabsService: Implement warmup()
`warmUp()`:

> Warms up the browser process asynchronously.

Let's make sure for engine-gecko that the engine is already running. In Fennec this will start `GeckoService`. Let's investigate if there's more we can do here to avoid a startup penalty once the app is actually started.

pocmo:


pocmo:
:+1: Added changelog entry.

PR #1527, opened 2018-11-30
Allow custom combiner functions in GenericScalarStorageEngine.kt

Dexterp37:
This introduces the combiner lambda function in `recordScalar`, allowing inheriting classes to define the behaviour with which new values are combined with existing values in the same storage, instead of simply overwriting the last value.

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
@@ -27,6 +27,12 @@ internal typealias GenericDataStorage<T> = MutableMap<String, T>
 internal typealias GenericStorageMap<T> = MutableMap<String, GenericDataStorage<T>>
 
 /**
+ * Defines the typealias for the combiner function to be used by
+ * [GenericScalarStorageEngine.recordScalar] when recording new values.
+ */
+internal typealias ScalarRecordingCombiner<T> = (currentValue: T?, newValue: T) -> T
+
+/**
  * A base class for 'scalar' like metrics. This allows sharing the common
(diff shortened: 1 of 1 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -167,11 +181,30 @@ abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
...
+    ) = recordScalar(metricData, value) { _, v -> v }
+
+    /**
+     * Helper function for the derived classes. It can be used to record
+     * simple scalars to the internal storage.
+     *
+     * @param metricData the information about the metric
+     * @param value the new value
+     * @param combine an optional lambda function to combine the currently stored value and
+     *        the new one; this allows to implement new behaviours such as adding.
```
pocmo: nit: `optional` is a bit misleading since the lambda is not optional for *this* method.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -167,11 +181,30 @@ abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
...
+     */
+    protected fun recordScalar(
+        metricData: CommonMetricData,
+        value: ScalarType,
+        combine: (currentValue: ScalarType?, newValue: ScalarType) -> ScalarType
```
pocmo: maybe define a typealias for this lambda at the top and use the alias here, e.g.

```Kotlin
typealias combinator = (currentValue: ScalarType?, newValue: ScalarType) -> ScalarType
```

PR #1531, opened 2018-11-30
Closes #1523: Remove references to org.mozilla.gecko.

### LINKED ISSUE #1523: Remove last bits of `org.mozilla.gecko` from Nightly
Follow up from #1366. There are still some bits of `org.mozilla.gecko` in use in A-C. Those are private APIs and will be removed soon. I'll send a patch for this soon.

agi90:
This removes all remaining references to `org.mozilla.gecko` in
`engine-gecko-nightly`. It also adds a `ReflectionUtils#setField`
method to set final fields in tests.

### DIFF
--- a/components/browser/engine-gecko-nightly/src/test/java/mozilla/components/test/ReflectionUtils.kt
+++ b/components/browser/engine-gecko-nightly/src/test/java/mozilla/components/test/ReflectionUtils.kt
@@ -1,0 +1,20 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.test
+
+import java.lang.reflect.Field
+import java.lang.reflect.Modifier
+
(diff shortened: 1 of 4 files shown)


components/browser/engine-gecko-nightly/src/test/java/mozilla/components/test/ReflectionUtils.kt
```diff
@@ -0,0 +1,16 @@
+package mozilla.components.test
```
pocmo: nit: License header

agi90: We need a lint :) Fixed, thanks for the review!

agi90:
Fixed `ktlint` failures.

p.s. I think `./gradlew lint` doesn't run `ktlint` is that intentional?

pocmo:
> p.s. I think ./gradlew lint doesn't run ktlint is that intentional?

Yep, `lint` just runs Android's linter.

pocmo:
The taskcluster failure is unrelated. I'll force-merge.

PR #1548, opened 2018-12-07
Add locales.

Pike:
We've got some localizations for error pages for FirefoxReality, let's uplift them to android-components.

Also found a bug in zh-TW.

CC @Delphine

### DIFF
--- a/l10n.toml
+++ b/l10n.toml
@@ -2,8 +2,15 @@
 basepath = "."
 
 locales = [
-
-  ]
+    "de",
+    "es",
+    "fr",
+    "it",
+    "ja",
+    "ko",
+    "zh-CN",
+    "zh-TW",
+]
 [env]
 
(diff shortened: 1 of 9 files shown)


l10n.toml
```diff
@@ -2,8 +2,15 @@
...
-  ]
+    "de",
+    "es",
+    "fr",
+    "it",
+    "ja",
+    "ko",
+    "zh-CN",
+    "zh-TW",
+]
```
pocmo: Is this how we will control what kind of locales get imported?

Pike: This is what c-l uses by default as the locales it tests.

The concrete flow of adding a new locale or removing one is something we need to drill into still. In particular around the role of the pontoon admin UI, and who's driving this. And if there's more than one way to do it.

PR #1554, opened 2018-12-10
[glean] Don't fail if no metrics yaml is present

Dexterp37:
Consumer application might just need glean core metrics, without defining any custom metric on they own. For this reason, they wouldn't need any `metrics.yaml` file. This patch makes sure the application build process doesn't fail when using glean and no such file is present.

This additionally fixes a minor problem in the CountersStorageEngine.

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/storages/CountersStorageEngine.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/storages/CountersStorageEngine.kt
@@ -43,8 +43,8 @@ internal open class CountersStorageEngineImplementation(
         amount: Int
     ) {
         // Use a custom combiner to add the amount to the existing counters rather than overwriting
-        super.recordScalar(metricData, amount) { currentValue, amount ->
-            currentValue?.let { it + amount } ?: amount
+        super.recordScalar(metricData, amount) { currentValue, newAmount ->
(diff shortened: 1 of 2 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/storages/CountersStorageEngine.kt
```diff
@@ -43,8 +43,8 @@ internal open class CountersStorageEngineImplementation(
         amount: Int
     ) {
         // Use a custom combiner to add the amount to the existing counters rather than overwriting
-        super.recordScalar(metricData, amount) { currentValue, amount ->
-            currentValue?.let { it + amount } ?: amount
+        super.recordScalar(metricData, amount) { currentValue, newAmount ->
+            currentValue?.let { it + newAmount } ?: newAmount
```
mdboom: Is it possible this is in this PR accidentally?

Dexterp37: This fixes the variable shadowing as mentioned in the initial commit message. Should I have a separate PR for this?

mdboom: Oh, I just missed that in the PR comment.  Not a big deal to include here.

PR #1555, opened 2018-12-10
Document how to integrate glean in consumer applications

Dexterp37:
This provides basic documentation for adding glean to the build system, enabling custom metrics.

**Potential improvements**

- Add the glean logo at the top of the readme.
- Document the `metrics.yaml` format.
- Provide an example on how to add custom metrics.
- Document the core metrics (this is **required** for the data review).
- Document the pings (this is **required** for the data review).

We might want to start with something simple though, and move other improvements to separate PRs.


Amejia481:
Looking forward to see the GLEAN icon in the Readme :)

image

Is this PR ready to be 🚢 ?

Dexterp37:
> Looking forward to see the GLEAN icon in the Readme :)

Let me do that now!

> Is this PR ready to be 🚢 ?

I'll wait for @mdboom to give this a look and then merge ;) Thanks!

components/service/glean/README.md
```diff
@@ -1,17 +1,59 @@
...
+```Groovy
+plugins {
+    id "com.jetbrains.python.envs" version "0.0.25"
+}
+```
+
+Right before the end of the same file, the glean build script must be included, as shown below:
+
+```Groovy
+apply from: 'https://github.com/mozilla-mobile/android-components/raw/master/components/service/glean/scripts/sdk_generator.gradle'
```
mdboom: Do we want to specify a specific tag in this URL?  Problems could occur if this doesn't match the version of `service-glean` specified above.  I think that (for now) means we'll have to track along with `android-components` versions...

Dexterp37: Great point! Thanks for spotting this. I've updated the docs, can you check if the current phrasing makes any sense?

Dexterp37:
> > Looking forward to see the GLEAN icon in the Readme :)
> 
> Let me do that now!

hah, actually, let's do this after I land a few more docs here ;)

components/service/glean/README.md
```diff
@@ -1,17 +1,68 @@
...
+        super.onCreate()
+
+        // Initialize the Glean library. Ideally, this is the first thing that
+        // must be done right after enabling logging.
+        Glean.initialize(applicationContext)
+    }
+}
+```
+
+Once initialized, glean will automatically start collecting and sending its [core metrics](metrics.yaml).
```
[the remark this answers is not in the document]
Dexterp37: > This doesn't seem very user friendly. A user using this for the first time might need a little more information than just the raw `metrics.yaml` file provides. Would it be better to have a separate metrics README.md that lists the types and intended usage or just a bullet list right here? The YAML file doesn't explain the metrics, it just shows an implementation of them and, as a user, I would like to know what a metric type is and should be used for as part of the documentation.

These are all great points! We'll need to document the data we collect out of the box and we will also need to document the YAML file format. However, I think we shouldn't block integrating to the reference-browser integration with that. I just filed bug 1513215 for this.

> Also, does this actually _automatically_ collect metrics or does the user need to call the generated API function? It might be better to show an example of both the implementation in the YAML file and how to call the API functions in the user's code.

Yes, it means that Glean does that automatically, without any other code required.

PR #1558, opened 2018-12-10
 Closes #1442: Add Setting for reading/setting GeckoView user agent

### LINKED ISSUE #1442: Add Setting for reading/setting GeckoView user agent
So far we could only do that for WebView. But now we can read and set the user agent for GeckoView builds too:

* https://bugzilla.mozilla.org/show_bug.cgi?id=1454444
* https://bugzilla.mozilla.org/show_bug.cgi?id=1507369

csadilek:


Amejia481:
@csadilek Should we have a pending issue for adding this to beta/stable when this feature is promoted to those channels?

csadilek:
> Should we have a pending issue for adding this to beta/stable when this feature is promoted to those channels?

Hm, I don't think so. We have many cases of this and with your plugin we always move Nightly forward to beta and release. So, this should just happen anyway, right?

PR #1559, opened 2018-12-10
[glean] 1512664: Don't record metrics or send pings if not initialized

mdboom:
`Glean.initialize` makes sure that upload and storage engines work properly. Unfortunately, the API is not checking whether or not glean was initialized. We need a way to enforce and write the relevant test coverage.

I don't think we want the main metric APIs to throw exceptions ever, so this just spits out error logs if trying to record metrics before initialization.

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/CommonMetricData.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/CommonMetricData.kt
@@ -40,6 +40,12 @@ interface CommonMetricData {
     }
 
     fun shouldRecord(logger: Logger): Boolean {
+        // Don't record metrics if we aren't initialized
+        if (!Glean.isInitialized()) {
+            logger.error("Glean must be initialized before metrics are recorded")
+            return false
+        }
+
         // Silently drop if metrics are turned off globally or locally
         if (!Glean.getMetricsEnabled() || disabled) {
             return false
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
@@ -26,30 +26,9 @@ import mozilla.components.service.glean.metrics.Baseline
 import mozilla.components.support.base.log.logger.Logger
 import java.io.File
 
-object Glean {
-    /**
-    * Enumeration of different metric lifetimes.
-    */
-    enum class PingEvent {
-        /**
-        * When the application goes into the background
-        */
-        Background,
-        /**
-        * A periodic event to send the default ping
-        */
-        Default
-    }
-
+@Suppress("TooManyFunctions")
+open class GleanInternalAPI {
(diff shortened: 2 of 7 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/CommonMetricData.kt
```diff
@@ -24,6 +24,8 @@ enum class Lifetime {
     User
 }
 
+private val logger = Logger("glean/CommonMetricData")
```
Dexterp37: Mh, I don't think we need this: `shouldRecord` already has a `logger` parameter that can be used for this.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -85,6 +91,13 @@ object Glean {
         ProcessLifecycleOwner.get().lifecycle.addObserver(gleanLifecycleObserver)
     }
 
+    /**
+     * Returns true if the Glean library has been initialized.
+     */
+    fun isInitialized(): Boolean {
```
Dexterp37: I'm not sure if this is possible at all, but can this be marked as `internal`? I'm wondering if we even need this function, but we certainly don't want it to be accessible as part of the outer facing API.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -201,7 +214,12 @@ object Glean {
      * @return A [Job], **only** to be used when testing, which allows to wait on
      *         the ping submission.
      */
-    fun handleEvent(pingEvent: PingEvent): Job {
+    fun handleEvent(pingEvent: PingEvent): Job? {
+        if (!isInitialized()) {
+            logger.error("Glean must be initialized before handling events.")
+            return null
```
Dexterp37: I'm not sure about this, but would it be possible to do something like `return Job()`, instead of making the return type nullable?

components/service/glean/src/main/java/mozilla/components/service/glean/CommonMetricData.kt
mdboom: Thanks, I missed that.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
mdboom: I think it's helpful to have it as a function in case we need to override its semantics later, but I agree it probably makes sense to make it internal.

Dexterp37: Sounds good to me!

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
mdboom: Yep.  Looks like that works.

PR #1566, opened 2018-12-12
[glean] Document the pings available out of the box [ci skip]

Dexterp37:
This adds documentation for the baseline ping and the metric it collects out of the box. This
additionally lays out the structure for the remaining documentation, so that all pings docs have a similar
structure.

### DIFF
--- a/components/service/glean/README.md
+++ b/components/service/glean/README.md
@@ -61,7 +61,8 @@ class SampleApplication : Application() {
 }
 ```
 
-Once initialized, glean will automatically start collecting and sending its [core metrics](metrics.yaml).
+Once initialized, if collection is enabled, glean will automatically start collecting [baseline metrics](metrics.yaml)
+and sending its [pings](docs/pings.md).
 
 ## License
 
--- a/components/service/glean/docs/baseline.md
+++ b/components/service/glean/docs/baseline.md
@@ -1,0 +1,18 @@
+# The `baseline` ping
+This ping is intended to provide metrics that are managed by the library itself, and not explicitly
+set by the application or included in the application's `metrics.yaml` file.
+
+The `baseline` ping is automatically sent when the application is moved to the *background* and it includes
+the following fields:
+
+| Field name | Type | Description |
+|---|---|---|
+| `sessions` | Counter | The number of foreground sessions since the last upload |
+| `durations` | Timespan | The combined duration of all the foreground sessions since the last upload |
+| `os` | String | The name of the operating system (e.g. "linux", "android", "ios") |
+| `os_version` | String | The version of the operating system |
+| `device` | String | Manufacturer and model |
+| `architecture` | String | The architecture of the device (e.g. "arm", "x86") |
+| `timezone` | Number | The timezone of the device, as an offset from UTC |
+| `accessibility_services` | StringList | List of accessibility services enabled on the device |
+| `locale` | String | The locale of the application |
--- a/components/service/glean/docs/pings.md
+++ b/components/service/glean/docs/pings.md
@@ -1,0 +1,46 @@
+# glean pings
+
+If data collection is enabled, glean provides a set of pings that are assembled out of the box
+without any developer intervention.
+Every glean ping is in JSON format and contains a [common section](#the-common-information-section)
+with shared information data included under the `ping_info` key.
+The following is a list of the pings along with the conditions that triggers them.
+
+- [`baseline` ping](baseline.md)
+- `events` ping (TBD)
+- `metrics` ping (TBD)
+
+## The `ping_info` section
+The following fields are always included in the `ping_info` section, for every ping.
+
+| Field name | Type | Description |
+|---|---|---|
+| `ping_type` | String | The name of the ping type (e.g. "baseline", "metrics") |
+| `app_build` | String | TBD |
+| `telemetry_sdk_build` | String | The version of the glean library |
+| `client_id` | UUID |  A UUID identifying a profile and allowing user-oriented correlation of data |
+| `seq` | Counter | A running counter of the number of times pings of this type have been sent |
+| `experiments` | Object | A dictionary of [active experiments](#the-experiments-object) |
+| `start_time` | Datetime | The time of the start of collection of the data in the ping |
+| `end_time` | Datetime | The time of the end of collection of the data in the ping. This is also the time this ping was generated and is likely well before ping transmission time |
+| `profile_age` | Datetime | TBD |
+
+All the metrics surviving application restarts (e.g. `client_id`, `seq`, ...) are removed once the
+application using glean is uninstalled.
+
+### The `experiments` object
+This object contains experiments keyed by the experiment `id`. Each listed experiment contains the
+`branch` the client is enrolled in and may contain a string to string map with additional data in the
+`extra` key. Both the `id` and `branch` are truncated to 30 characters.
+
+```json
+{
+  "<id>": {
+    "branch": "branch-id",
+    "extra": {
+      "some-key": "a-value"
+    }
+  }
+}
+```
+


components/service/glean/docs/baseline.md
```diff
@@ -0,0 +1,18 @@
+# The `baseline` ping
+This ping is intended to provide metrics that are managed by the library itself, and not explicitly
+set from the application or included in the application's `metrics.yaml` file.
```
chutten:
```suggestion
set by the application or included in the application's `metrics.yaml` file.
```

components/service/glean/docs/baseline.md
```diff
@@ -0,0 +1,18 @@
...
+
+The `baseline` ping is automatically sent when the application is moved to *background* and it includes
```
chutten:
```suggestion
The `baseline` ping is automatically sent when the application is moved to the *background* and it includes
```

components/service/glean/docs/baseline.md
```diff
@@ -0,0 +1,18 @@
...
+
+| Field name | Type | Description |
+|---|---|---|
+| `sessions` | Counter | The number of foreground sessions since the last upload |
+| `durations` | Timespan | The combined duration of all the foreground sessions since the last upload |
+| `os` | String | The name of the operating system (e.g. "linux", "android", "ios") |
+| `os_version` | String | The version of the operating system |
+| `device` | String | Manufacturer and model |
+| `architecture` | String | The architecture of the device (e.g. "arm", "x86") |
+| `timezone` | Number | The timezone of the device, in offset from UTC |
```
chutten:
```suggestion
| `timezone` | Number | The timezone of the device, as an offset from UTC |
```

components/service/glean/docs/baseline.md
```diff
@@ -0,0 +1,18 @@
...
+| `accessibility_services` | StringList | List of accessibility services enabled on the device |
```
chutten: This is on the `baseline` ping? ...that's interesting. I wouldn't have expected this to be baseline data.

components/service/glean/docs/pings.md
```diff
@@ -0,0 +1,42 @@
+# glean pings
+
+glean provides a set of pings that are assembled, out of the box, without any developer intervention.
```
chutten: Should we highlight again that this is only when collection is enabled?

components/service/glean/docs/pings.md
```diff
@@ -0,0 +1,42 @@
...
+Every glean ping is in JSON format and contains a [common section](#the-common-information-section)
+with shared information data, included under the `ping_info` key.
+The following is a list of the pings, along with the conditions that triggers them.
+
+- [`baseline` ping](baseline.md)
+- `events` ping (TBD)
+- `metrics` ping (TBD)
+
+## The `common` information section
+The following fields are always included in this section, for every ping.
```
chutten:
```suggestion
The following fields are always included in the `ping_info` section, for every ping.
```

components/service/glean/docs/pings.md
```diff
@@ -0,0 +1,42 @@
...
+
+| Field name | Type | Description |
+|---|---|---|
+| `ping_type` | String | The name of the ping |
```
chutten:
```suggestion
| `ping_type` | String | The name of the ping type (e.g. "baseline", "metrics") ||
```

components/service/glean/docs/pings.md
```diff
@@ -0,0 +1,42 @@
...
+| `app_build` | String | TBD |
+| `telemetry_sdk_build` | String | An identifier for the version of the glean library |
```
chutten:
```suggestion
| `telemetry_sdk_build` | String | The version of the glean library |
```

components/service/glean/docs/pings.md
```diff
@@ -0,0 +1,42 @@
...
+| `client_id` | String |  A UUID identifying a profile and allowing user-oriented correlation of data |
+| `seq` | Counter | A running counter of the number of times this ping type has been sent |
```
chutten:
```suggestion
| `seq` | Counter | A running counter of the number of times pings of this type have been sent |
```

components/service/glean/docs/pings.md
```diff
@@ -0,0 +1,42 @@
...
+with shared information data, included under the `ping_info` key.
+The following is a list of the pings, along with the conditions that triggers them.
+
+- [`baseline` ping](baseline.md)
+- `events` ping (TBD)
+- `metrics` ping (TBD)
+
+## The `common` information section
+The following fields are always included in this section, for every ping.
+
```
chutten: Do we want a section detailing when/how these are generated, cleared, and regenerated? For instance, I thought I remembered that there was a profile reset or opt-out mechanism that reset the clientId or reset  `seq` to 0 or something...

components/service/glean/docs/pings.md
```diff
@@ -0,0 +1,42 @@
...
+| Field name | Type | Description |
+|---|---|---|
+| `ping_type` | String | The name of the ping |
+| `app_build` | String | TBD |
+| `telemetry_sdk_build` | String | An identifier for the version of the glean library |
+| `client_id` | String |  A UUID identifying a profile and allowing user-oriented correlation of data |
+| `seq` | Counter | A running counter of the number of times this ping type has been sent |
+| `experiments` | Object | A dictionary of [active experiments](#the-experiments-object) |
+| `start_time` | Datetime | The time of the start of collection of the data in the ping |
+| `end_time` | Datetime | The time of the end of collection of the data in the ping, and ping generation time (which is likely well before ping transmission time) |
```
chutten:
```suggestion
| `end_time` | Datetime | The time of the end of collection of the data in the ping. This is also the time this ping was generated and is likely well before ping transmission time |
```

components/service/glean/README.md
```diff
@@ -61,7 +61,8 @@ class SampleApplication : Application() {
 }
 ```
 
-Once initialized, glean will automatically start collecting and sending its [core metrics](metrics.yaml).
+Once initialized, if collection is enabled, glean will automatically start collecting [core metrics](metrics.yaml)
```
mdboom: Shouldn't "core metrics" be "baseline metrics" here?  (I'm always a little unclear about core vs. baseline).

components/service/glean/docs/pings.md
```diff
@@ -0,0 +1,42 @@
...
+
+## The `common` information section
+The following fields are always included in this section, for every ping.
+
+| Field name | Type | Description |
+|---|---|---|
+| `ping_type` | String | The name of the ping |
+| `app_build` | String | TBD |
+| `telemetry_sdk_build` | String | An identifier for the version of the glean library |
+| `client_id` | String |  A UUID identifying a profile and allowing user-oriented correlation of data |
```
chutten: Isn't `UUID` a top-level type we could use here?

components/service/glean/README.md
mdboom: Shouldn't we say "baseline metrics" instead of "core metrics" here? (I'm always a little unclear on this).

mdboom: Or I guess core == "baseline" + "ping_info"?  What about saying "internally collected metrics"?

components/service/glean/docs/pings.md
```diff
@@ -0,0 +1,42 @@
...
+| `seq` | Counter | A running counter of the number of times this ping type has been sent |
+| `experiments` | Object | A dictionary of [active experiments](#the-experiments-object) |
+| `start_time` | Datetime | The time of the start of collection of the data in the ping |
+| `end_time` | Datetime | The time of the end of collection of the data in the ping, and ping generation time (which is likely well before ping transmission time) |
+| `profile_age` | Datetime | TBD |
+
+### The `experiments` object
+This object contains experiments, keyed by the experiment `id`. Each listed experiment contains the
+`branch` the client is enrolled in and may contain a string to string map with additional data in the
+`extra` key. Both the `id` and `branch` are truncated to 30 characters.
```
chutten: Limits like these remind me that there might be implicit datatype limits that might be helpful to link to from this documentation. Maybe a section "These data types are explained in further detail in \[the data types documentation](...). Any limits or exemptions to limits are documented here with the fields that have or are exempt from the limitations."

components/service/glean/docs/baseline.md
```diff
@@ -0,0 +1,18 @@
...
+| `locale` | String | The locale of the application |
```
mdboom: I'm a bit worried that this table and the descriptions in the `metrics.yaml` will get out of sync, and sort of feel that the `metrics.yaml` should be the one source of truth.  But that's probably going to require a kotlindoc plugin that I'd rather not write, so we should probably not worry about it for now.

components/service/glean/docs/pings.md
```diff
@@ -0,0 +1,42 @@
...
+glean provides a set of pings that are assembled, out of the box, without any developer intervention.
+Every glean ping is in JSON format and contains a [common section](#the-common-information-section)
+with shared information data, included under the `ping_info` key.
+The following is a list of the pings, along with the conditions that triggers them.
+
+- [`baseline` ping](baseline.md)
+- `events` ping (TBD)
+- `metrics` ping (TBD)
+
+## The `common` information section
```
mdboom: Should we just call this the `ping_info` section to avoid confusion?

components/service/glean/README.md
Dexterp37: Gah, good point. Looks like I'm getting confused too :)

components/service/glean/docs/baseline.md
Dexterp37: Thanks Chris, this is very helpful! I'm making the changes locally and then I'll force-push here (vs accepting the changes here online).

components/service/glean/docs/baseline.md
Dexterp37: Yeah :E

components/service/glean/docs/pings.md
Dexterp37: Good point! I've slightly rephrased this.

components/service/glean/docs/baseline.md
chutten: It's tricky because we have a process requirement here to be clear about the things glean collects and why. If we could embed the relevant portions it would get the job done but it would also mean that you don't have to consider keeping the docs up to date when you add new baseline metrics.

It may be that we find that "it's hard to keep the two up-to-date" is a feature not a bug, forcing us to be more deliberative and process-heavy when we add collections to this core library.

components/service/glean/docs/pings.md
Dexterp37: Good point. We don't have, right now, anything to reset user-level metrics. They go away when user uninstalls the product using glean. I filed bug 1513616 in case some products want to have a way to clear stuff manually without uninstalling the app.

components/service/glean/docs/pings.md
Dexterp37: Great point. I just filed [a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1513618) about having in-tree documentation for the different metric types.

components/service/glean/docs/baseline.md
Dexterp37: @mdboom I see your concern. On the other hand, the fact that new collection *must* require a data-review could be a reminder for keeping things in sync :)

components/service/glean/README.md
Dexterp37: No, that was my bad, it's "baseline metrics", as written in the spec. I just got confused :)

components/service/glean/docs/pings.md
```diff
@@ -0,0 +1,46 @@
+# glean pings
+
+If data collection is enabled, glean provides a set of pings that are assembled, out of the box,
```
travis79: nit:  As a native English speaker the commas around `out of the box` are a little awkward here.  It should still be grammatically correct to say:

`If data collection is enabled, glean provides a set of pings that are assembled out of the box without any developer intervention.`

components/service/glean/docs/pings.md
```diff
@@ -0,0 +1,46 @@
...
+without any developer intervention.
+Every glean ping is in JSON format and contains a [common section](#the-common-information-section)
+with shared information data, included under the `ping_info` key.
```
travis79: nit: comma after ` information data` not necessary.

components/service/glean/docs/pings.md
```diff
@@ -0,0 +1,46 @@
...
+The following is a list of the pings, along with the conditions that triggers them.
```
travis79: nit:  comma after ` the pings` not necessary

components/service/glean/docs/pings.md
```diff
@@ -0,0 +1,46 @@
...
+| `experiments` | Object | A dictionary of [active experiments](#the-experiments-object) |
+| `start_time` | Datetime | The time of the start of collection of the data in the ping |
+| `end_time` | Datetime | The time of the end of collection of the data in the ping. This is also the time this ping was generated and is likely well before ping transmission time |
+| `profile_age` | Datetime | TBD |
+
+All the metrics surviving application restarts (e.g. `client_id`, `seq`, ...) are removed once the
+application using glean is uninstalled.
+
+### The `experiments` object
+This object contains experiments, keyed by the experiment `id`. Each listed experiment contains the
```
travis79: nit: comma after ` experiments` not necessary.

PR #1568, opened 2018-12-12
Closes #1567: Don't show private sessions in awesomebar

### LINKED ISSUE #1567: SessionSuggestionProvider shows private sessions
This might be a product question, but it seems reasonable to not show private sessions in the awesomebar (this is also seen in Fennec as well).

jonalmeida:
We've been showing private sessions in the awesome bar which doesn't seem like our intended behaviour (it may give the impression that we're also saving the session).

I've added a filter check if the session is private, but would we also want to provide an external session filter (similar to `TabsFeature#filterTabs`)?


jonalmeida:
@pocmo When I changed the if-statement, the detekt was failing with a `ComplexCondition` which I think is fair for this logic:
```kotlin
if ((session.url.contains(text, ignoreCase = true) && !session.private) ||
    (session.title.contains(text, ignoreCase = true) && !session.private)
) 
```

I went back to the previous early return method instead which also passes the tests.

pocmo:
Did it fail with only one "(..) && !session.private"  too? :)

jonalmeida:
I guess not! 😅

PR #1571, opened 2018-12-12
Idea: Introduce a FeatureManager to handle lifecycle events

jonalmeida:
This was an idea I had when we talked about all the start/stop calls we've been making. A FeatureManager can take care of making those calls without having to remember to do it.

This is a rough implementation of how it would work since it was quick to implement. :)


samples/browser/src/main/java/org/mozilla/samples/browser/BrowserFragment.kt
```diff
@@ -104,6 +107,18 @@ class BrowserFragment : Fragment(), BackHandler {
         promptFeature = PromptFeature(components.sessionManager, requireFragmentManager())
 
         windowFeature = WindowFeature(components.engine, components.sessionManager)
+
+        featureManager.apply {
+            register(sessionFeature)
```
csadilek: I guess `register` could take a vararg? Then we could call `register(sessionFeature, toolbarFeature)` etc.

csadilek: There could also be an `unregisterAll`

components/support/base/src/main/java/mozilla/components/support/base/feature/FeatureManager.kt
```diff
@@ -0,0 +1,32 @@
...
+
+package mozilla.components.support.base.feature
+
+import android.arch.lifecycle.Lifecycle
+import android.arch.lifecycle.LifecycleObserver
+import android.arch.lifecycle.OnLifecycleEvent
+
+class FeatureManager(private val featureList: MutableList<Feature> = mutableListOf()) : LifecycleObserver {
+
+    fun register(feature: Feature) {
```
Amejia481: Idea 💡: We could overload the `plusAssign` and `minusAssign` for a more idiomatic way to adding or removing features:
 
```kotlin
class FeatureManager {

        operator fun minusAssign(feature: Feature) {
            register(feature)
        }

        operator fun plusAssign(feature: Feature) {
            unregister(feature)
        }
    }

    fun main() {
        val manager = FeatureManager()
        manager += feature1
        manager -= feature1
    }
```

jonalmeida: I think the varargs solution should be good enough. I can see us mostly using the manager for (un)registering all features.

jonalmeida:
@Amejia481 this isn't ready for merging, it was just easier to write the code to show the idea. It still needs tests as well. :)

components/support/base/src/main/java/mozilla/components/support/base/feature/FeatureManager.kt
Amejia481: 👍

jonalmeida:
More thoughts: the alternative to this would be to make the features themselves lifecycle aware, but I'm not sure if that's the dependency we want to include on each of them.

pocmo:
Instead of adding a manager and enforcing features to have a specific API I would like to see us go into a direction where there's a `LifecycleAwareFeature`. With the Lifecycle APIs it's way more straight-forward to have the feature annotate it's methods as it likes and the application code just registering the `feature` on a lifecycle owner. This also makes it very obvious what and when things are started/stopped.

jonalmeida:
> Instead of adding a manager and enforcing features to have a specific API I would like to see us go into a direction where there's a `LifecycleAwareFeature`. With the Lifecycle APIs it's way more straight-forward to have the feature annotate it's methods as it likes and the application code just registering the `feature` on a lifecycle owner. This also makes it very obvious what and when things are started/stopped.
> 
> Edit: Heh, just saw that this is something you wrote too in the last comment. I agree with that :)

I initially thought we would have introduce the lifecycle aware library to each feature, but it looks like we inherited that already. I have a patch that I updated already, but I'll test it a bit more today to make sure everything works fine.

jonalmeida:
@csadilek had an excellent point that we should try to enforce a lifecycle contract with our features, so instead we're extending the `LifecycleObserver` interface where we add the appropriate annotations and also avoid polluting each feature class with the annotations as well.

Maybe in the future we can add a lint rule to hint feature users to not forget to observe the lifecycle when creating one in an app?

PR #1572, opened 2018-12-13
[glean] 1510552 Send the event ping

mdboom:
This sets up the event ping to send when (a) going to background and (b) when the event storage engine of a particular store exceeds `MAX_EVENTS` items.

This includes a refactor of the `Glean` object into a class and a singleton instance to make it easier to create one-off instances for testing.

This also rewrites an unrelated test to send pings to a testing webserver rather than catching the contents in a mock.  I wasn't aware we already had this infrastructure when I wrote that original test and I think it's much better to mock less.

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
@@ -9,7 +9,6 @@ import android.content.Context
 import android.support.annotation.VisibleForTesting
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.GlobalScope
-import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
 
 import java.util.UUID
@@ -138,15 +137,13 @@ open class GleanInternalAPI {
     /**
      * Collect and assemble the ping. Asynchronously submits the assembled
      * payload to the designated server using [httpPingUploader].
-     *
-     * @return The [Job] created by the ping submission.
      */
-    internal fun sendPing(store: String, docType: String): Job {
+    internal fun sendPing(store: String, docType: String) {
         val pingContent = pingMaker.collect(store)
         val uuid = UUID.randomUUID()
         val path = makePath(docType, uuid)
         // Asynchronously perform the HTTP upload off the main thread.
-        return GlobalScope.launch(Dispatchers.IO) {
+        GlobalScope.launch(Dispatchers.IO) {
             httpPingUploader.upload(path = path, data = pingContent)
         }
     }
@@ -189,18 +186,18 @@ open class GleanInternalAPI {
      *   - Default: Event that triggers the default pings.
      *
      * @param pingEvent The type of the event.
-     *
-     * @return A [Job], **only** to be used when testing, which allows to wait on
-     *         the ping submission.
      */
-    fun handleEvent(pingEvent: Glean.PingEvent): Job {
+    fun handleEvent(pingEvent: Glean.PingEvent) {
         if (!isInitialized()) {
             logger.error("Glean must be initialized before handling events.")
-            return Job()
+            return
         }
 
-        return when (pingEvent) {
-            Glean.PingEvent.Background -> sendPing("baseline", "baseline")
+        when (pingEvent) {
+            Glean.PingEvent.Background -> {
+                sendPing("baseline", "baseline")
+                sendPing("events", "events")
+            }
             Glean.PingEvent.Default -> sendPing("metrics", "metrics")
         }
     }
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
@@ -4,6 +4,7 @@
 
 package mozilla.components.service.glean.storages
 
+import mozilla.components.service.glean.Glean
 import android.annotation.SuppressLint
 import android.content.Context
 import android.os.SystemClock
@@ -70,6 +71,9 @@ internal object EventsStorageEngine : StorageEngine {
             for (storeName in stores) {
                 val storeData = eventStores.getOrPut(storeName) { mutableListOf() }
                 storeData.add(event.copy())
+                if (storeData.size == Glean.configuration.maxEvents) {
+                    Glean.sendPing(storeName, "events")
+                }
             }
         }
     }
--- a/components/service/glean/src/test/java/mozilla/components/service/glean/storages/EventsStorageEngineTest.kt
+++ b/components/service/glean/src/test/java/mozilla/components/service/glean/storages/EventsStorageEngineTest.kt
@@ -4,18 +4,39 @@
 package mozilla.components.service.glean.storages
 
 import android.os.SystemClock
+import androidx.test.core.app.ApplicationProvider
+import mozilla.components.service.glean.checkPingSchema
+import mozilla.components.service.glean.config.Configuration
+import mozilla.components.service.glean.FakeDispatchersInTest
+import mozilla.components.service.glean.Glean
+import mozilla.components.service.glean.EventMetricType
+import mozilla.components.service.glean.Lifetime
+import mozilla.components.service.glean.net.HttpPingUploader
+import okhttp3.mockwebserver.MockResponse
+import okhttp3.mockwebserver.MockWebServer
+import org.json.JSONObject
 import org.junit.Before
 import org.junit.Test
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertNull
+import org.junit.Rule
 import org.junit.runner.RunWith
 import org.robolectric.RobolectricTestRunner
 
 @RunWith(RobolectricTestRunner::class)
 class EventsStorageEngineTest {
 
+    @get:Rule
+    val fakeDispatchers = FakeDispatchersInTest()
+
     @Before
     fun setUp() {
+        Glean.initialized = false
+        Glean.initialize(
+            applicationContext = ApplicationProvider.getApplicationContext(),
+            configuration = Configuration(applicationId = "test")
+        )
+        assert(Glean.initialized)
         EventsStorageEngine.clearAllStores()
     }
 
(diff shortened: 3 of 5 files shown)


components/service/glean/src/test/java/mozilla/components/service/glean/storages/EventsStorageEngineTest.kt
```diff
@@ -154,4 +167,46 @@ class EventsStorageEngineTest {
...
+
+        val realClient = Glean.httpPingUploader
+        val testConfig = Glean.configuration.copy(
+            serverEndpoint = "http://" + server.hostName + ":" + server.port
+        )
+        Glean.httpPingUploader = HttpPingUploader(testConfig)
+
+        try {
+            for (i in 0..499) {
+                click.record("buttonA")
```
Dexterp37: Can we encode `i` in some field of the event? To make sure we're actually sending different events. Maybe the `extra` field might be a good option.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -33,7 +32,7 @@ open class GleanInternalAPI {
     // Include our singletons of StorageEngineManager and PingMaker
     private lateinit var storageEngineManager: StorageEngineManager
     private lateinit var pingMaker: PingMaker
-    private lateinit var configuration: Configuration
+    internal lateinit var configuration: Configuration
```
Dexterp37: Instead of changing this, can't we pass the default configuration to the uploader in the test that needs it? We're always using the default configuration in the tests anyway

components/service/glean/src/test/java/mozilla/components/service/glean/storages/EventsStorageEngineTest.kt
```diff
@@ -154,4 +167,46 @@ class EventsStorageEngineTest {
...
+            disabled = false,
+            category = "ui",
+            lifetime = Lifetime.Ping,
+            name = "click",
+            sendInPings = listOf("default"),
+            objects = listOf("buttonA")
+        )
+
+        val realClient = Glean.httpPingUploader
+        val testConfig = Glean.configuration.copy(
```
Dexterp37: Instead of doing this, we can pass in the default `with val testConfig = Configuration(...)`, leaving the definitions untouched.

components/service/glean/src/test/java/mozilla/components/service/glean/storages/EventsStorageEngineTest.kt
```diff
@@ -154,4 +167,46 @@ class EventsStorageEngineTest {
...
+            serverEndpoint = "http://" + server.hostName + ":" + server.port
+        )
+        Glean.httpPingUploader = HttpPingUploader(testConfig)
+
+        try {
+            for (i in 0..499) {
```
Dexterp37: Mh, I would do something different: trigger more than 500 events and make sure that just 500 make it into the ping, checking that the rest is still there when taking a snapsnot.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
```diff
@@ -70,6 +71,9 @@ internal object EventsStorageEngine : StorageEngine {
             for (storeName in stores) {
                 val storeData = eventStores.getOrPut(storeName) { mutableListOf() }
                 storeData.add(event.copy())
+                if (storeData.size >= Glean.configuration.maxEvents) {
+                    Glean.sendPing(storeName, "events")
```
Dexterp37: Would this guarantee that only `maxEvents` events make it into the ping?

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
mdboom: This change is needed so the EventsStorageEngine has access to `MAX_EVENTS`.  We could provide an API in Glean to get it or pass the Configuration object along to all storage engines instead...

components/service/glean/src/test/java/mozilla/components/service/glean/storages/EventsStorageEngineTest.kt
mdboom: That's a good idea.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
Dexterp37: Ah, good point, never mind my comments about this then!

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
mdboom: In practice, yes.  Using `>=` rather than `==` is just insurance against another future code path adding events without checking and sending, but hopefully that would be caught be testing.

components/service/glean/src/test/java/mozilla/components/service/glean/storages/EventsStorageEngineTest.kt
mdboom: I'm not sure I follow...

Dexterp37: Yeah, don't bother about this. It was related to [this one](https://github.com/mozilla-mobile/android-components/pull/1572#discussion_r241403006), and it doesn't make sense :)

components/service/glean/src/test/java/mozilla/components/service/glean/storages/EventsStorageEngineTest.kt
mdboom: Good idea.

mdboom:
@Dexterp37 : I think I've addressed all of the comments and added an additional test to increase coverage.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
mdboom: Thinking this through again.  Given that ping building is async, there is a theoretical condition where:

1) The 500th event is recorded, triggering an eventual ping build
2) A 501st event is recorded
3) The ping is built

This would lead to a ping with 501 events, which I think is fine and preferable to data loss and preferable to blocking on ping building.

But this does lead me to feel that the `>=` should change to `==` to prevent two pings from being sent in this scenario.

Dexterp37: Good point. I think we always prefer 501 to data loss and blocking. Maybe we could check how focus and telemetry-android deal with this?

Dexterp37: So, looks like telemetry-android has a [similar behaviour](https://github.com/mozilla-mobile/android-components/blob/bc0f57b553a41f1e93213937118f6c84a442803e/components/service/telemetry/src/main/java/org/mozilla/telemetry/Telemetry.java#L112-L113). Moreover, by looking at [FFTV](https://sql.telemetry.mozilla.org/queries/60515/source#156227), it seems that we don't have that many clients hitting the limit... so is an edge case of a limit case? Which means we should be pretty much ok?

mdboom: Yeah.  I think particularly on mobile where we are sending/clearing the event queue every time we go to background, we're even more likely to not fill up the queue than on desktop.  So I'm pretty confident we are good here.

Dexterp37:
This is also failing due to #1576

mdboom:
Unfortunately, #1576 doesn't seem to have fixed these tests reliably (they worked once, but not now).  Back to the drawing board on this.

Dexterp37:
> 
> 
> Unfortunately, #1576 doesn't seem to have fixed these tests reliably (they worked once, but not now). Back to the drawing board on this.

This is weird. Are you able to reproduce locally?

mdboom:
service-glean is not passing, but many of the other tests aren't.  That's a good sign, I guess.

components/service/glean/src/test/java/mozilla/components/service/glean/storages/EventsStorageEngineTest.kt
```diff
@@ -4,18 +4,39 @@
...
 import org.junit.Assert.assertNull
+import org.junit.Rule
 import org.junit.runner.RunWith
 import org.robolectric.RobolectricTestRunner
 
 @RunWith(RobolectricTestRunner::class)
 class EventsStorageEngineTest {
 
+    @get:Rule
+    val fakeDispatchers = FakeDispatchersInTest()
```
mdboom: This is, I think, the key to making the tests work.  This PR adds some `Dispatchers.API.launch` calls to this file for the first time (in the use of `EventMetricType.record`), and therefore needs the shim to make that stuff synchronous.

travis79:
These `504` errors are similar to the errors I commonly saw with my last PR.  It was rare that it would work the first time and I would sometimes have to retrigger TC a couple of times for it to pass.  I wonder if it is related to server load?  I noticed it seemed to work a lot better earlier in the AM while I was in Berlin when most of the US would still be asleep.

pocmo:
Yeah, we have been seeing those issues frequently and more often when there are multiple PRs ongoing (during US work time). We'll update our Docker image soon (which includes the cached dependencies) and then hopefully we'll have less problems for some time again.

components/service/glean/src/test/java/mozilla/components/service/glean/storages/EventsStorageEngineTest.kt
Dexterp37: I think so too. But that should have made everything sync: at least I'm fairly sure that was working for other tests.

mdboom:
Having rebased this on upstream, I've rerun the tests 5 times and can't get them to fail (for the service-glean test, at least).

mdboom:
Given now 7 times passing, I'm convinced that the failures have been resolved by upstream changes, and this is ready to merge.

PR #1573, opened 2018-12-13
Closes #1489: Fix backspace issue when suggestion strip is off

### LINKED ISSUE #1489: Backspace in URL bar skips character when autocomplete suggestion given
## Steps to Reproduce:
0. Disable all text correction features for Gboard (Can keep "auto-capitalization" on)
1. Type "Moz" in the url bar
2. Press backspace
3. Press backspace

## Expected
"Mo" is left in the url bar 

## Actual
"Mz" is left in the url bar

Demo of this can be found [here](https://github.com/mozilla-mobile/focus-android/issues/3892#issuecomment-441253070).

Amejia481:
I also added a workaround for [an issue with Sony keyboards](https://bugzilla.mozilla.org/show_bug.cgi?id=1344464) that was missing from the fennec port. 

https://github.com/mozilla-mobile/focus-android/issues/3892

### DIFF
--- a/components/ui/autocomplete/src/main/java/mozilla/components/ui/autocomplete/InlineAutocompleteEditText.kt
+++ b/components/ui/autocomplete/src/main/java/mozilla/components/ui/autocomplete/InlineAutocompleteEditText.kt
@@ -5,9 +5,12 @@
 package mozilla.components.ui.autocomplete
 
 import android.content.Context
+import android.content.Context.INPUT_METHOD_SERVICE
 import android.graphics.Color.parseColor
 import android.graphics.Rect
 import android.os.Build
+import android.provider.Settings.Secure.DEFAULT_INPUT_METHOD
+import android.provider.Settings.Secure.getString
 import android.support.v7.widget.AppCompatEditText
 import android.text.Editable
 import android.text.NoCopySpan
@@ -123,6 +126,8 @@ open class InlineAutocompleteEditText @JvmOverloads constructor(
         color
     }()
 
+    private val inputMethodManger get() = context.getSystemService(INPUT_METHOD_SERVICE) as InputMethodManager?
+
     private val onKeyPreIme = fun (_: View, keyCode: Int, event: KeyEvent): Boolean {
         // We only want to process one event per tap
         if (event.action != KeyEvent.ACTION_DOWN) {
@@ -184,6 +189,9 @@ open class InlineAutocompleteEditText @JvmOverloads constructor(
         }
     }
 
+    private val isSonyKeyboard: Boolean
+        get() = INPUT_METHOD_SONY == getCurrentInputMethod()
+
     public override fun onAttachedToWindow() {
         super.onAttachedToWindow()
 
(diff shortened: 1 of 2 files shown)


Amejia481:
Hi @sblatz, could take a look at Sample Browser and see if the issue is fixed ? 
If you need any help just let know :)

Thanks!

Amejia481:
I'm unable to replicate the same issue.
I tested on a physical phone (Pixel 3) and an emulator with Android 9.

Could you me which environment settings did you use? Thanks :)
 
**Demo**
example

sblatz:
I was testing on a Pixel 2 XL API 28 (emulator). I just tested again today and am unable to reproduce. Maybe it was some strange emulator bug that day? I think since we're now both unable to reproduce this, it's safe to merge in!

Amejia481:
Awesome :). If you encounter this issue again feel free to open an issue.

components/ui/autocomplete/src/main/java/mozilla/components/ui/autocomplete/InlineAutocompleteEditText.kt
```diff
@@ -469,6 +476,7 @@ open class InlineAutocompleteEditText @JvmOverloads constructor(
                     // If we have autocomplete text, the cursor is at the boundary between
                     // regular and autocomplete text. So regardless of which direction we
                     // are deleting, we should delete the autocomplete text first.
+                    restartInput()
```
pocmo: We recently removed exactly this restart to fix an issue. See #911 and f789c0653b67557bf6525817b6d9e902cdcf56de.

Amejia481: What a dilemma!
Without the restarting, we're having issues when suggestion strip is off.

We have to find a way to convey solutions.

Amejia481: @pocmo  What do you think about disabling this functionally only on Amazon devices?  Both sides could be happy.

The [Build class](https://developer.android.com/reference/android/os/Build#MANUFACTURER) can provide us with enough info to do that.

pocmo: I don't remember whether this problem only showed on Amazon devices. @mcomella Do you remember whether you were able to reproduce #911 on other devices too?

mcomella: #911 was created from https://github.com/mozilla-mobile-skunkworks/firefox-connect/issues/484 – iirc, this was only reproducible on the Echo Show.

Amejia481: Should I proceed with this approach ?

pocmo: It may be our only option at this point. Let's document this decision in code and a test though. :)

Amejia481: 👍

Sdaswani:
@mcomella we actually need this fix in for Focus soon so can you test if this regresses Echo Show next week?

components/ui/autocomplete/src/main/java/mozilla/components/ui/autocomplete/InlineAutocompleteEditText.kt
mcomella: I opened https://github.com/mozilla-mobile/android-components/pull/1666 with the Echo Show fix.

I chose to go with the IME because an IME with the same ID should have similar behavior so it seemed less fragile than using the build information.

pocmo:
Closing in favor of #1666.

PR #1575, opened 2018-12-14
[glean] Require an 'app_id' for initializing glean

Dexterp37:
We need the concept of 'app_id' to tell data streams apart on the pipeline side. This PR adds the field and
changes the glean initialization so that the application is required to provide one.

### DIFF
--- a/components/service/glean/README.md
+++ b/components/service/glean/README.md
@@ -56,10 +56,13 @@ class SampleApplication : Application() {
 
         // Initialize the Glean library. Ideally, this is the first thing that
         // must be done right after enabling logging.
-        Glean.initialize(applicationContext)
+        Glean.initialize(applicationContext, Configuration(applicationId = "YOUR_APPLICATION_NAME_HERE"))
     }
 }
 ```
+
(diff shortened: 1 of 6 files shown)


Dexterp37:
@mdboom all right, this is ready for another look! We're now using the `app-id` to build the submission URL :)

components/service/glean/README.md
```diff
@@ -56,11 +56,14 @@ class SampleApplication : Application() {
 
         // Initialize the Glean library. Ideally, this is the first thing that
         // must be done right after enabling logging.
-        Glean.initialize(applicationContext)
+        Glean.initialize(applicationContext, Configuration(applicationId = "glean"))
```
mdboom: Maybe to make the note below more obvious this should be `applicationId = "YOUR_APPLICATION_NAME_HERE"`?

Dexterp37: Good point!

Dexterp37:
```suggestion
        Glean.initialize(applicationContext, Configuration(applicationId = "YOUR_APPLICATION_NAME_HERE"))
```

PR #1578, opened 2018-12-14
Closes #1300: Intermittent failure in BrowserAwesomeBarTest [#5]

### LINKED ISSUE #1300: Intermittently failing test: BrowserAwesomeBar cancels previous jobs
```
mozilla.components.browser.awesomebar.BrowserAwesomeBarTest > BrowserAwesomeBar cancels previous jobs if onInputStarted gets called again FAILED
    java.lang.AssertionError at BrowserAwesomeBarTest.kt:170
```

See also #1288.

csadilek:
Basically this comment on `scheduleResumeAfterDelay`: `Continuation **must be scheduled** to resume even if it is already cancelled,...`

Made me think if it was possible we cancelled the `Job` in between calls to `delay` so instead of an infinite loop I am testing for `isActive`. Let's see if this makes a difference.


Amejia481:
Yeap :) 🚢 5/6 💚 !

csadilek:
Back to the drawing port on this one again.

PR #1579, opened 2018-12-14
[glean] 1512947: Upgrade JetBrains Python plugin to 0.0.26..

mdboom:
This allows us to remove the workaround to forcibly create the build directory first.


mdboom:
I'm not sure how coverage decreased, but it's by a very small amount...

Dexterp37:
> 
> 
> I'm not sure how coverage decreased, but it's by a very small amount...

I think that's fine, sometimes codecov just... acts weirdly.

pocmo:
Yep, we have noticed that too!

PR #1580, opened 2018-12-15
Improvements around FirefoxSyncFeature state management and ergonomics.

grigoryk:
This patch includes a few changes:
- it changes 'sync' to be a 'suspend' function, dropping requirement for a CoroutineContext.
  We are starting to understand what we care about vis-a-vis scopes, so let's make this
  more flexible so that realistic usage patterns may emerge at the consumer level (reference browser).
- it introduces a mutex, which guards sync operation exposed by this feauture.
  We don't want multiple sync operations to be running at once, so the feature shouldn't
  expose that footgun to its users.
- doesn't throw AuthException, instead wrapping it in SyncError. This makes dealing with outcomes
  of sync a bit simpler.
  We still have work to do around unifying errors coming out of PlacesConnection, as some
  of those could be auth-related.
- introduces sync start/stop listeners that can be registered and unregistered. This allows
  for much simpler UI code which needs to reflect sync's internal state.

Sample apps that use the feature are updated accordingly.

### DIFF
--- a/components/feature/sync/src/main/java/mozilla/components/feature/sync/FirefoxSyncFeature.kt
+++ b/components/feature/sync/src/main/java/mozilla/components/feature/sync/FirefoxSyncFeature.kt
@@ -4,15 +4,31 @@
 
 package mozilla.components.feature.sync
 
-import kotlinx.coroutines.CompletableDeferred
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Deferred
-import kotlinx.coroutines.async
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
 import mozilla.components.concept.storage.SyncError
 import mozilla.components.concept.storage.SyncableStore
 import mozilla.components.service.fxa.FirefoxAccountShaped
 import mozilla.components.support.base.log.logger.Logger
-import kotlin.coroutines.CoroutineContext
+import mozilla.components.support.base.observer.Observable
+import mozilla.components.support.base.observer.ObserverRegistry
+
+/**
+ * An interface for consumers that wish to observer "sync lifecycle" events.
+ */
+interface SyncStatusObserver {
+    /**
+     * Gets called at the start of a sync, before any configured syncable is synchronized.
+     */
+    fun onStarted()
+
+    /**
+     * Gets called at the end of a sync, after every configured syncable has been synchronized.
+     */
+    fun onIdle()
+}
+
+val registry = ObserverRegistry<SyncStatusObserver>()
 
 /**
  * A feature implementation which orchestrates data synchronization of a set of [SyncableStore] which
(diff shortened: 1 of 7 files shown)


grigoryk:
@csadilek @pocmo I'd like to get this into 0.35.0, as it makes life much easier for the reference browser.

samples/sync-history/src/main/java/org/mozilla/samples/sync/history/MainActivity.kt
```diff
@@ -85,18 +85,28 @@ class MainActivity : AppCompatActivity(), LoginFragment.OnLoginCompleteListener,
             txtView.text = getString(R.string.logged_out)
         }
 
+        featureSync.onSyncRunning(this, object : onSyncRunning { override fun invoke() {
+            launch {
+                val txtView: TextView = findViewById(R.id.historySyncStatus)
+                txtView.text = getString(R.string.syncing)
+            }
+        } })
```
pocmo: Since this is a class with just one method you can write it as a lambda. And then you can move the lambda outside of the function call and this will make it much nicer to read:

```Kotlin
featureSync.onSyncRunning(this) {
    launch {
        val txtView: TextView = findViewById(R.id.historySyncStatus)
        txtView.text = getString(R.string.syncing)
    }
}
```

Finally, Kotlin Android Extensions are enabled for this sample app and you do not need to find the view yourself. The extensions will generate the code for you. Just import the property with the same id from (in this case) `kotlinx.android.synthetic.main.activity_main.`. Android Studio should do that automatically for you once you start typing and autocompleting `historySyncStatus`.

```Kotlin
featureSync.onSyncRunning(this) {
    launch {
        historySyncStatus?.text = getString(R.string.syncing)
     }
}
```

I also wonder whether you want to explicitly use the Main dispatcher here to run this on the UI thread. Afaik that's not guaranteed here.

components/feature/sync/src/main/java/mozilla/components/feature/sync/FirefoxSyncFeature.kt
```diff
@@ -24,51 +24,94 @@ import kotlin.coroutines.CoroutineContext
  * (e.g. places and logins depend on native libraries), and we do not want to force a consumer which
  * only cares about syncing logins to have to import a places native library.
  *
- * @param coroutineContext A [CoroutineContext] that will be used to perform synchronization work.
  * @param reifyAuth A conversion method which reifies a generic [FxaAuthInfo] into an object of
  * type [AuthType].
  */
 class FirefoxSyncFeature<AuthType>(
```
pocmo: For the listeners here: Consider using `ObserverRegistry` (if applicable) which is used by some of our other classes. It takes care of dealign with listeners for you. More importantly it has hooks for binding subscriptions to a lifecycle/view or pausing them when the app is in the background. This avoids you having to deal with all those weak references and stuff in your app code: It will automatically pause or unsubscribe for you.

grigoryk:
Configuration of which syncables are to be synced is in a follow-up: https://github.com/mozilla-mobile/android-components/issues/1584

grigoryk:
@pocmo @csadilek addressed feedback; now with more tests and backed by an `ObserverRegistry`.

PR #1589, opened 2018-12-18
[glean] Automatically discover the applicationId in glean

Dexterp37:
This changes the glean initialization so that a default `applicationId` is discovered and provided as the default
parameter for the configuration. This additionally fixes tests and the documentation.

This leaves the configuration option in place for supporting applications that might need to tweak this id value (e.g. Focus/Klar like)

### DIFF
--- a/components/service/glean/docs/pings.md
+++ b/components/service/glean/docs/pings.md
@@ -44,3 +44,13 @@ This object contains experiments keyed by the experiment `id`. Each listed exper
 }
 ```
 
+## Ping submission
+The pings glean generates are submitted to the Mozilla servers at specific paths, in order to provide
+additional metadata without the need of unpacking the ping payload. A typical submission URL looks like
+`"<server-address>/submit/<application-id>/<doc-type>/<glean-schema-version>/<ping-uuid>"`, with:
+ 
+- `<server-address>`: the address of the server that receives the pings;
+- `<application-id>`: a unique application id, automatically detected by glean; this is the value returned by [`Context.getPackageName()`](http://developer.android.com/reference/android/content/Context.html#getPackageName());
+- `<doc-type>`: the name of the ping; this can be one of the pings available out of the box with glean, or a custom ping;
+- `<glean-schema-version>`: the version of the glean ping schema;
+- `<ping-uuid>`: a unique identifier for this ping.
--- a/components/service/glean/README.md
+++ b/components/service/glean/README.md
@@ -56,13 +56,10 @@ class SampleApplication : Application() {
 
         // Initialize the Glean library. Ideally, this is the first thing that
         // must be done right after enabling logging.
-        Glean.initialize(applicationContext, Configuration(applicationId = "YOUR_APPLICATION_NAME_HERE"))
+        Glean.initialize(applicationContext)
     }
 }
 ```
-
-**Important:** the `applicationId` will be used to properly route data in the pipeline, so it must be unique.
-The `glean` value is reserved for testing.
 
 Once initialized, if collection is enabled, glean will automatically start collecting [baseline metrics](metrics.yaml)
 and sending its [pings](docs/pings.md).
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
@@ -42,6 +42,10 @@ open class GleanInternalAPI {
     internal var initialized = false
     private var metricsEnabled = true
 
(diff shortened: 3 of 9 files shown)


Dexterp37:
> Is there any reason to leave the Configuration as a parameter to the initialize function? Does this expose a potential way someone could spoof another valid application ID? I'm not saying this is a bad thing, I'm just curious as to whether we are doing this by design or not.

Yes, someone intentionally changing the configuration would still send data with the same id, that's expected. It's not something we can really watch out for in the library alone, because nothing would stop an individual to just use curl and send whatever they want :) What we really want is to prevent unintentional mistakes to happen: forking the repo to send data with the same id.

We still need the flexibility to change that as part of the configuration, if needed, to support some internal use cases: for example, Focus and Klar really being the same product but having different application ids.

> edit: I just read your comment and got the answer to this, I'm just uncertain whether this would have any negative aspects yet.

georgf:
> We still need the flexibility to change that as part of the configuration, if needed, to support some internal use cases: for example, Focus and Klar really being the same product but having different application ids.

Could we just handle those id differences internally in the library?
And thus avoid exposing this for configuration generally?

Dexterp37:
> > We still need the flexibility to change that as part of the configuration, if needed, to support some internal use cases: for example, Focus and Klar really being the same product but having different application ids.
> 
> Could we just handle those id differences internally in the library?
> And thus avoid exposing this for configuration generally?

I would personally try not to do that, as that would mean us managing some sort of hardcoded edge-case list within the library itself. I'd rather defer this to the application, *only when and if needed*.

I think, generally speaking, the `Configuration` should probably never be changed, unless it's needed to address special cases or in test scenarios.

georgf:
Talking through this, i caught up on the motivating issue for the override again.
However, for me it seems the main question is if allowing an override is the right way to support the Focus/Klar issue. We should have conversations about that.

However that doesn't block landing this PR and we can move that to a separate bug.
Cheers for figuring out the automatic discovery!

pocmo:
Another option may be to just cross that bridge when we get to it. Focus is using `service-telemetry` and it may be unlikely that it will migrate to `glean` any time soon (if ever tbh). So we could just make it not configurable and add that option once we are confident that we will really really need it (so far it seems to be cheap to add later on?). Overall the Focus/Klar split was such a pain that I doubt this is something we will try to do again for a new product.

pocmo:
Oh, and there may be other options. For example I could use [ContextWrapper](https://developer.android.com/reference/android/content/ContextWrapper#getPackageName()) to pass a wrapped `Context` to Glean that just overrides what `packageName` returns. :)

Dexterp37:
All right! @mdboom @travis79  this is ready for another look!

components/service/glean/docs/pings.md
```diff
@@ -44,3 +44,13 @@ This object contains experiments keyed by the experiment `id`. Each listed exper
 }
 ```
 
+## Ping submission
+The pings glean generates are submitted to the Mozilla servers at specific paths, in order to provide
+additional metadata without the need of unpacking the ping payload. A typical submission URL looks like
+ `"<server-address>/submit/<application-id>/<doc-type>/<glean-schema-version>/<ping-uuid>"`, with:
```
mdboom:
```suggestion
`"<server-address>/submit/<application-id>/<doc-type>/<glean-schema-version>/<ping-uuid>"`, with:
```

components/service/glean/docs/pings.md
```diff
@@ -44,3 +44,13 @@ This object contains experiments keyed by the experiment `id`. Each listed exper
...
+ 
+- `<server-address>`: the address of the server that receives the pings;
+- `<application-id>`: an unique application id, automatically detected by glean; this is the value returned by [`Context.getPackageName()`](http://developer.android.com/reference/android/content/Context.html#getPackageName());
```
mdboom:
```suggestion
- `<application-id>`: a unique application id, automatically detected by glean; this is the value returned by [`Context.getPackageName()`](http://developer.android.com/reference/android/content/Context.html#getPackageName());
```

English rules about an with "u" are weird.

Dexterp37:
> I agree that the common and safest case is to pull this from the Android packageName, but do we have a way to override that if we need to use something else? I thought we had a known use case for that (Focus and Klar)?

Yes, but turns out that they won't probably be using glean. The general consensus in the previous comments was to just scrape this and revisit later, as we might have other options to deal with similar cases.

mdboom:
> The general consensus in the previous comments was to just scrape this and revisit later, as we might have other options to deal with similar cases.

Ok.  LGTM then.

PR #1590, opened 2018-12-18
Closes #1577: Request Desktop checkbox doesn't show selected state

### LINKED ISSUE #1577: 'Request Desktop' checkbox remains unchecked in desktop mode
Clicking on "Request Desktop" checkbox from the browser menu in sample browser loads the web page in desktop mode. But, the checkbox remains unchecked. So, essentially I can't go back to mobile mode.

jonalmeida:


components/browser/menu/src/main/java/mozilla/components/browser/menu/item/SimpleBrowserMenuCheckbox.kt
```diff
@@ -19,15 +19,18 @@ import mozilla.components.browser.menu.R
  */
 class SimpleBrowserMenuCheckbox(
     private val label: String,
-    private var initialState: Boolean = false,
+    private val initialState: () -> Boolean = { false },
```
Amejia481: Is no completely clear to me, why we need a lambda instead of variable here? 🤔

PR #1595, opened 2018-12-19
Issue #1396: CustomTabsService: Implement mayLaunchUrl()

### LINKED ISSUE #1396: CustomTabsService: Implement mayLaunchUrl()
`mayLaunchUrl()`

> The method warmup(long) has to be called beforehand. The most likely URL has to be specified explicitly. Optionally, a list of other likely URLs can be provided. They are treated as less likely than the first one, and have to be sorted in decreasing priority order. These additional URLs may be ignored. All previous calls to this method will be deprioritized.

In Fennec this will call `GeckoThread.speculativeConnect()`. We may need something like that in GeckoView?

pocmo:
Closes #1396.


pocmo:
>  Changelog?

Oh, yeah, I always forget that :D

PR #1598, opened 2018-12-19
Issue #1085: FirefoxAccount no longer needs to be a Deferred type..

csadilek:
This is just some cleanup we forgot after applying the changes for #1085. Now that creating a `Config` and `toJSONString`/`fromJSONString` are no longer async we don't need to use a `Deferred` for `FirefoxAccount`.

It gets rid of a bunch of unneeded `await` calls and also removes the potential deadlocking code on `onDestroy`.

The reference browser is already updated.

More improvements on the way in:
https://github.com/mozilla/application-services/issues/483

### DIFF
--- a/samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/MainActivity.kt
+++ b/samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/MainActivity.kt
@@ -14,12 +14,9 @@ import android.content.Intent
 import android.widget.CheckBox
 import android.widget.TextView
 import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Deferred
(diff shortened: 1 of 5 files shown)


samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/MainActivity.kt
```diff
@@ -51,34 +48,18 @@ open class MainActivity : AppCompatActivity(), LoginFragment.OnLoginCompleteList
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_main)
         job = Job()
-
-        whenAccount = async {
```
vladikoff: 👍

PR #1599, opened 2018-12-19
[glean] Update PingMaker.kt to remove format specifier not supported in Android SDK-21

travis79:
SimpleDateFormat, prior to SDK 24 does not support format specifier 'X' to output the time zone offset.  Switched to use format specifier 'Z' which is supported, although slightly different in it's output.  Instead of hh:mm it outputs hhmm for the timezone offset.  In light of this, I had to alter the tests as DateTimeFormatter.ISO_OFFSET_DATE_TIME is strict in requiring the ':' even though the actual ISO8601 spec is not.  This should allow us to fully support the specified minimum SDK version for both android-components, and the reference browser.

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/ping/PingMaker.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/ping/PingMaker.kt
@@ -12,6 +12,7 @@ import mozilla.components.support.base.log.logger.Logger
 import mozilla.components.support.ktx.android.org.json.mergeWith
 import org.json.JSONException
 import org.json.JSONObject
+import java.lang.StringBuilder
 import java.text.SimpleDateFormat
 import java.util.Date
 import java.util.Locale
@@ -29,8 +30,18 @@ internal class PingMaker(
      * @return a string containing the date and time.
      */
     private fun getISOTimeString(): String {
-        val dateFormat = SimpleDateFormat("yyyy-MM-dd'T'HH:mmXXX", Locale.US)
-        return dateFormat.format(Date())
+        val dateFormat = SimpleDateFormat("yyyy-MM-dd'T'HH:mmZ", Locale.US)
+        val timeString = StringBuilder(dateFormat.format(Date()))
+
+        // Due to limitations of SDK version 21, there isn't a way to properly output the time
+        // offset with a ':' character:
+        // 2018-12-19T12:36-0600    --  This is what we get
+        // 2018-12-19T12:36-06:00   -- This is what GCP will expect
+        //
+        // In order to satisfy time offset requirements of GCP, we manually insert the ":"
+        timeString.insert(timeString.length - 2, ":")
+
+        return timeString.toString()
     }
 
     /**
--- a/components/service/glean/src/test/java/mozilla/components/service/glean/ping/PingMakerTest.kt
+++ b/components/service/glean/src/test/java/mozilla/components/service/glean/ping/PingMakerTest.kt
@@ -19,6 +19,7 @@ import org.junit.Assert.assertTrue
 import org.junit.runner.RunWith
 import org.mockito.Mockito.mock
 import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
 import java.time.OffsetDateTime
 import java.time.format.DateTimeFormatter
 
@@ -27,6 +28,7 @@ class PingMakerTest {
     private val mockApplicationContext = mock(Context::class.java)
 
     @Test
+    @Config(minSdk = 21)
     fun `"ping_info" must contain a non-empty start_time and end_time`() {
         val maker = PingMaker(StorageEngineManager(storageEngines = mapOf(
                 "engine2" to MockStorageEngine(JSONObject())
@@ -47,7 +49,7 @@ class PingMakerTest {
         DateTimeFormatter.ISO_OFFSET_DATE_TIME.parse(pingInfo.getString("end_time"))
         OffsetDateTime.parse(pingInfo.getString("start_time"))
         assertTrue(OffsetDateTime.parse(pingInfo.getString("start_time"))
-            >= OffsetDateTime.parse(pingInfo.getString("end_time")))
+            <= OffsetDateTime.parse(pingInfo.getString("end_time")))
     }
 
     @Test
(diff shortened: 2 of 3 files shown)


travis79:
So, according to @fbertsch it appears that the ":" will be required.

mdboom:
It occurs to me that maybe we should try to merge this asap to unblock the reference browser work, and file a bug about coming back to address the backend's needs with the `:` a bit later.

travis79:
I agree, I don’t want to hold up the reference browser integration either.  I should have it complete with the : and all with my last commit from earlier today.

Dexterp37:
@pocmo flagging you there as well, since this seems an Android specific detail. In general, do you have any suggestion on how a policy should look like wrt SDK versions? Which ones should we keep compatible to? What's the minimum SDK version we should support? How does this policy look like for the &-browser and android-components?

pocmo:
> @pocmo flagging you there as well, since this seems an Android specific detail. In general, do you have any suggestion on how a policy should look like wrt SDK versions? Which ones should we keep compatible to? What's the minimum SDK version we should support? How does this policy look like for the &-browser and android-components?

First of all, the following is a helpful link when discussing this. It shows the Android platform distribution as Google sees it from their metrics. The page is updated every couple of month:
https://developer.android.com/about/dashboards/

With Android Components we currently support SDK 21 (Android 5, Lollipop) or higher (with only one exception for lib-dataprotect [used by Lockbox] which *requires* a new level 23 API). According to the dashboard this means we support roughly ~88% of the Android market.

There are at times conflicting expectations:
* The product team wants to support as much as possible so that the potential user base is as large as possible
* The engineering team wants to use a high minimum SDK level to be guaranteed to be able to use new platform features for every installation and to reduce the engineering overhead.

SDK 21 was the lowest one we could agree on for Focus. We required some new APIs that were only available on this SDK level and this was also the first version that came with Material design and some of the related UI features.

From my point of view it's unlikely that we will ever go lower than 21 in future products. For one the market has "upgraded" enough that we cover a very large potion with that. And there's some significant overhead involved.

Since telemetry/glean will be part of all our products your `minSdkLevel` will have a direct impact on all our products. You should try to follow our global `minSdkLevel` (21) [which you do via `Config.minSdkVersion` in `build.gradle`] and avoid anything above that as much as possible. :)

The reference browser just follows what the components do and is currently API 21+.

components/service/glean/src/main/java/mozilla/components/service/glean/ping/PingMaker.kt
```diff
@@ -29,8 +30,22 @@ internal class PingMaker(
...
+        // 2018-12-19T12:36-06:00   -- This is what GCP will expect
+        //
+        // In order to satisfy time offset requirements of GCP, we manually insert the ":" when
+        // needed, which should be for every timezone except UTC which is signified by the trailing
+        // 'Z' character
+        if (!timeString.endsWith("Z")) {
+            timeString.insert(timeString.length - 2, ":")
+        }
+
+        return timeString.toString()
```
pocmo: I wonder if we could move parts of this to an `internal` helper function so that we can have a unit test call this with various dates (including the UTC special case you mention) and verify the correct output.

components/service/glean/src/test/java/mozilla/components/service/glean/ping/PingMakerTest.kt
```diff
@@ -27,6 +28,7 @@ class PingMakerTest {
     private val mockApplicationContext = mock(Context::class.java)
 
     @Test
+    @Config(minSdk = 21)
```
pocmo: Since we know that this behavior can be different based on the Android version I wonder whether we should at more API levels here - or at least `24` where we know that `Z` was introduced.

components/service/glean/src/main/java/mozilla/components/service/glean/ping/PingMaker.kt
```diff
@@ -29,8 +30,22 @@ internal class PingMaker(
...
+        // offset with a ':' character:
+        // 2018-12-19T12:36-0600    --  This is what we get
+        // 2018-12-19T12:36-06:00   -- This is what GCP will expect
+        //
+        // In order to satisfy time offset requirements of GCP, we manually insert the ":" when
+        // needed, which should be for every timezone except UTC which is signified by the trailing
+        // 'Z' character
+        if (!timeString.endsWith("Z")) {
+            timeString.insert(timeString.length - 2, ":")
+        }
```
pocmo: Just FYI: In case you ever want to write code that behaves differently on various SDK levels:
* You can read the current SDK level with: https://developer.android.com/reference/android/os/Build.VERSION.html#SDK_INT
* And compare it against constants from: https://developer.android.com/reference/android/os/Build.VERSION_CODES

Here it doesn't matter I guess and we can just use the workaround on all devices.

components/service/glean/src/main/java/mozilla/components/service/glean/ping/PingMaker.kt
```diff
@@ -29,8 +30,22 @@ internal class PingMaker(
      * @return a string containing the date and time.
      */
     private fun getISOTimeString(): String {
-        val dateFormat = SimpleDateFormat("yyyy-MM-dd'T'HH:mmXXX", Locale.US)
-        return dateFormat.format(Date())
+        val dateFormat = SimpleDateFormat("yyyy-MM-dd'T'HH:mmZ", Locale.US)
+        val timeString = StringBuilder(dateFormat.format(Date()))
+
+        // Due to limitations of SDK version 21, there isn't a way to properly output the time
```
Dexterp37: Can we document this in the `docs/pings.md`, when we talk about the fields using this date? We can create a new section in the doc, and reference it from each field.

components/service/glean/src/main/java/mozilla/components/service/glean/ping/PingMaker.kt
Dexterp37: You would still need to make this `internal` in order for it to be called from tests :)

components/service/glean/src/test/java/mozilla/components/service/glean/ping/PingMakerTest.kt
pocmo: Oh, I didn't know about the difference between `sdk` and `minSdk`. This is perfect then! :)

pocmo:
> @pocmo and @mdboom, it would be good to have your feedback on this. I don't like declining a request to change but in this case I was spending a lot of time just trying to replicate a situation that shouldn't be able to occur.

I'm totally okay with not doing that - use your own judgment here. I also fully trust that this is going to work. I was mostly concerned about changes (either by us or the platform) - but the existing test may be good enough for catching that.

PR #1608, opened 2018-12-21
Clean up gradle build files for some of the components

grigoryk:
- Remove a few unnecessary dependencies
- Add some comments around usage of 'api'
- Mark some of the dependencies as 'api' if they're necessary for use of the module

### DIFF
--- a/components/concept/storage/build.gradle
+++ b/components/concept/storage/build.gradle
@@ -22,19 +22,10 @@ android {
 }
 
 dependencies {
-    implementation Dependencies.kotlin_stdlib
-    implementation Dependencies.support_annotations
-    implementation Dependencies.kotlin_coroutines
-
-    // We expose this as API because we are using Observable in our public API and do not want every
-    // consumer to have to manually import "base".
-    api project(':support-base')
-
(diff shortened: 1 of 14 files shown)


components/concept/storage/build.gradle
```diff
@@ -22,19 +22,10 @@ android {
...
-
-    testImplementation Dependencies.testing_junit
-    testImplementation Dependencies.testing_robolectric
-    testImplementation Dependencies.testing_mockito
-
-    testImplementation project(':support-test')
+    // Necessary because we use 'suspend'. Fun fact: this module will compile just fine without this
+    // dependency, but it will crash at runtime.
+    // Included via 'api' because this module is unusable without coroutines.
+    api Dependencies.kotlin_coroutines
```
pocmo: This one I do not really understand. `suspend` is a language feature and shouldn't require the coroutines library. The library code itself also doesn't reference any code from the package. What kind of crash are you seeing?

grigoryk: `Continuation` will be missing from the classpath.

PR #1609, opened 2018-12-21
Issue #1541: feature-session-bundling: New component that saves sessions in grouped bundles.

pocmo:
*I'm currently writing tests for this. But I wanted to share this for review already.*

This is the absolute small MVP for session bundling. Sessions are saved as a bundle in a database. You can define a lifetime and during that lifetime a bundle will be restored and re-used. Otherwise a new bundle will be started. I added some APIs to build UI for listing and restoring other bundles. Once released I want to build a little prototype UI in the reference browser.

The previous refactoring was helpful. Now I could just move `AutoSave` out and re-use all that logic for a new "session storage" implementation. I did the same with the code that de/serializes the snapshots and with that I had almost all building blocks for implementing a different session storage.

### DIFF
--- a/components/feature/session-bundling/README.md
+++ b/components/feature/session-bundling/README.md
@@ -1,0 +1,78 @@
+# [Android Components](../../../README.md) > Feature > Session-Bundling
+
+A session storage implementation that saves the state of sessions in grouped bundles (e.g. by time) in a database.
+
+## Usage
+
+### Setting up the dependency
+
+Use Gradle to download the library from [maven.mozilla.org](https://maven.mozilla.org/) ([Setup repository](../../../README.md#maven-repository)):
+
+```Groovy
+implementation "org.mozilla.components:feature-session-bundling:{latest-version}"
+```
+
+### Setting up
+
+When creating a `SessionBundleStorage` instance provide a bundle lifetime:
+
+```Kotlin
+val storage = SessionBundleStorage(
+    applicationContext,
+    bundleLifetime = Pair(1, TimeUnit.HOURS))
+```
+
+The provided lifetime will control how long a session bundle is active. The last session bundle will only be restored if it is not older than the provided lifetime. In case that no recent bundle could be found a new empty session bundle will be created.
+
+### Saving state automatically
+
+An app can call `save(SessionManager.Snapshot)` at any time to save a snapshot in the current session bundle. However in most scenarios it is recommended to setup the automatic saving of state:
+
+```Kotlin
+storage.autoSave(sessionManager)
+    .periodicallyInForeground() // interval can be configured
+    .whenGoingToBackground()    // app goes to the background
+    .whenSessionsChange()       // New sessions, pages load, etc.
+```
+
+### Restoring state
+
+Calling `restore()` will restore the last active session bundle if it is still active (lifetime has not expired) or null if no active session bundle is available. This bundle can be used to restore the state of the bundle:
+
+```Kotlin
+val bundle = storage.restore()
+
+bundle?.restoreSnapshot(engine)?.let { snapshot -> restore(snapshot) }
+```
+
+### Working with bundles
+
+The `SessionBundleStorage` class offers multiple methods for accessing and using `SessionBundle` instances:
+
+```Kotlin
+// LiveData list of the last 20 bundles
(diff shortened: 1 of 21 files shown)


components/feature/session-bundling/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+Use Gradle to download the library from [maven.mozilla.org](https://maven.mozilla.org/) ([Setup repository](../../../README.md#maven-repository)):
+
+```Groovy
+implementation "org.mozilla.components:feature-session-bundling:{latest-version}"
+```
+
+### Setting up
+
+When creating a `SessionBundleStorage` instace provide a bundle lifetime:
```
csadilek: nit: type: `instance`

components/feature/session-bundling/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+```Kotlin
+val storage = SessionBundleStorage(
+    applicationContext,
+    bundleLifetime = Pair(1, TimeUnit.HOURS))
+```
+
+The provided lifetime will control how long a session is active. The last session bundle will only restored if it is not older than the provided lifetime. In case that no recent bundle could be found a new empty session bundle will be created.
```
csadilek: nit: session bundle is active?

csadilek: nit: only be restored.

components/feature/session-bundling/src/main/java/mozilla/components/feature/session/bundling/db/BundleEntity.kt
```diff
@@ -0,0 +1,48 @@
...
+ * which only exposes the part of the API we want to make public.
+ */
+@Entity(tableName = "bundles")
+internal data class BundleEntity(
+    @PrimaryKey(autoGenerate = true) var id: Long?,
+    var state: String,
+    @ColumnInfo(name = "saved_at") var savedAt: Long
+) : SessionBundle {
+    /**
+     * Update this entity with the value from the given snapshot.
```
csadilek: nit: Updates...

components/feature/session-bundling/src/main/java/mozilla/components/feature/session/bundling/SessionBundleStorage.kt
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * A [Session] storage implementation that saves snapshots as a [SessionBundle].
+ *
+ * @param bundleLifetime The lifetime of a bundle controls whether a bundle will be restored or whether this bundle is
+ * considered expired and a new bundle will be used.
+ */
+class SessionBundleStorage(
+    context: Context,
+    private val bundleLifetime: Pair<Long, TimeUnit>
```
csadilek: Passing a `bundeLifetime` to the constructor will have all bundles have the same lifetime. I am wondering if we should instead add a `createBundle` call to this class. Then apps / or other components could create bundles either based on different times/intervals (~3pm, this afternoon) or even based on topics etc. I might be getting ahead of myself, but it seems a method to create new bundles might be better? Basically the `SessionBundleStorage` would just expose method to manage bundles. Then it's up to the app / other components to define when that happens based on whatever algorithm they want to use.

csadilek: This will make `restore` more complicated though as we'd need to specify what `latest` means, but it seems doable as long as we still sort by `saved_at`.

pocmo: Yeah, currently it's not very flexible. I was thinking of turning this into some kind of builder-pattern-like thing that you can configure to create bundles the way you like (or customize). However that's something I would like to do in a future iteration.

pocmo:
@csadilek Updated the PR! Now with some tests. :)

components/feature/session-bundling/src/main/java/mozilla/components/feature/session/bundling/SessionBundleStorage.kt
csadilek: OK, sounds good!

PR #1611, opened 2018-12-21
[glean] 1497894: First set of metrics from baseline ping

mdboom:
This adds some (but not all) of the metrics from the baseline ping.  The remaining ones are blocked by either more metric types or some more design consideration.

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
@@ -6,6 +6,7 @@ package mozilla.components.service.glean
 
 import android.arch.lifecycle.ProcessLifecycleOwner
 import android.content.Context
+import android.os.Build
 import android.support.annotation.VisibleForTesting
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.GlobalScope
@@ -22,6 +23,7 @@ import mozilla.components.service.glean.scheduler.GleanLifecycleObserver
 import mozilla.components.service.glean.storages.ExperimentsStorageEngine
 import mozilla.components.service.glean.storages.StorageEngineManager
 import mozilla.components.service.glean.metrics.Baseline
+import mozilla.components.service.glean.utils.StringUtils
 import mozilla.components.support.base.log.logger.Logger
 import java.io.File
 
@@ -182,6 +184,26 @@ open class GleanInternalAPI {
 
         // Set the OS type
         Baseline.os.set("Android")
+
+        // Set the OS version
+        // https://developer.android.com/reference/android/os/Build.VERSION
+        Baseline.osVersion.set(Build.VERSION.SDK_INT.toString())
+
+        // Set the device string
+        // https://developer.android.com/reference/android/os/Build
+        // We limit the device descriptor to 32 characters because it can get long. We give fewer
+        // characters to the manufacturer because we're less likely to have manufacturers with
+        // similar names than we are for a manufacturer to have two devices with the similar names
+        // (e.g. Galaxy S6 vs. Galaxy Note 6).
+        @Suppress("MagicNumber")
+        Baseline.device.set(
+            StringUtils.safeSubstring(Build.MANUFACTURER, 0, 12) +
+            '-' +
+            StringUtils.safeSubstring(Build.MODEL, 0, 19)
+        )
+
+        // Set the CPU architecture
+        Baseline.architecture.set(Build.SUPPORTED_ABIS[0])
     }
 
     /**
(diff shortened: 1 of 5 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -182,6 +182,15 @@ open class GleanInternalAPI {
 
         // Set the OS type
         Baseline.os.set("Android")
+        // Set the OS version
+        // https://developer.android.com/reference/android/os/Build.VERSION
+        // TODO: We should probably send the SDK version as well
+        Baseline.osVersion.set(android.os.Build.VERSION.RELEASE)
```
Dexterp37: The legacy telemetry library for android uses [something different](https://github.com/mozilla-mobile/android-components/blob/9e2b38f8b98545baaee8d2dd5bb68dd4b9e6b98f/components/service/telemetry/src/main/java/org/mozilla/telemetry/measurement/OperatingSystemVersionMeasurement.java#L13): are they equivalent? If not, I think we should prefer using what the legacy library was sending for data equivalency.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -182,6 +182,15 @@ open class GleanInternalAPI {
...
+        // Set the device string
+        // https://developer.android.com/reference/android/os/Build
+        Baseline.device.set("${android.os.Build.MANUFACTURER} ${android.os.Build.MODEL}")
```
Dexterp37: The legacy library builds it a bit [differently](https://github.com/mozilla-mobile/android-components/blob/9e2b38f8b98545baaee8d2dd5bb68dd4b9e6b98f/components/service/telemetry/src/main/java/org/mozilla/telemetry/measurement/DeviceMeasurement.java#L26): let's stay equivalent

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -182,6 +182,15 @@ open class GleanInternalAPI {
...
+        // Set the CPU architecture
+        Baseline.architecture.set(System.getProperty("os.arch"))
```
Dexterp37: As above, see [here](https://github.com/mozilla-mobile/android-components/blob/9e2b38f8b98545baaee8d2dd5bb68dd4b9e6b98f/components/service/telemetry/src/main/java/org/mozilla/telemetry/measurement/ArchMeasurement.java#L24-L31)

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -182,6 +184,31 @@ open class GleanInternalAPI {
...
+        // https://developer.android.com/reference/android/os/Build.VERSION
+        Baseline.osVersion.set(Build.VERSION.SDK_INT.toString())
+
+        // Set the device string
+        // https://developer.android.com/reference/android/os/Build
+        // We limit the device descriptor to 32 characters because it can get long. We give fewer
+        // characters to the manufacturer because we're less likely to have manufacturers with
+        // similar names than we are for a manufacturer to have two devices with the similar names
+        // (e.g. Galaxy S6 vs. Galaxy Note 6).
+        @Suppress(MagicNumber)
```
travis79: I believe `MagicNumber` need quotes in the `@Suppress`

mdboom:
Totally forgot there was precedence here!  Should be more in line with legacy now.

components/service/glean/src/main/java/mozilla/components/service/glean/utils/StringUtils.kt
```diff
@@ -0,0 +1,14 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this
+* file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.glean.utils
+
+object StringUtils {
+    public fun safeSubstring(str: String, start: Int, end: Int): String {
```
travis79: `public` is redundant and can be omitted

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -182,6 +184,31 @@ open class GleanInternalAPI {
...
+        // (e.g. Galaxy S6 vs. Galaxy Note 6).
+        @Suppress("MagicNumber")
+        Baseline.device.set(
+            StringUtils.safeSubstring(Build.MANUFACTURER, 0, 12) +
+            '-' +
+            StringUtils.safeSubstring(Build.MODEL, 0, 19)
+        )
+
+        // Set the CPU architecture
+        val arch = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
```
travis79: FYI, our minimum supported SDK version is 21, which _is_ LOLLIPOP, so this should always be true.

Dexterp37: Good catch!

Issue #1293, opened 2018-11-07
Sample apps: Build variants in parallel


pocmo:
Building the sample browser takes a lot of time compared to all other modules. This is mostly because it has a lot of build variants. We could speed the overall builds up significantly by building the variants in parallel (like we build components in parallel).

```
    - variant: geckoBetaAarch64Debug
    - variant: geckoBetaX86Debug
    - variant: geckoBetaArmDebug
    - variant: systemUniversalDebug
    - variant: geckoNightlyAarch64Debug
    - variant: geckoNightlyX86Debug
    - variant: geckoNightlyArmDebug
    - variant: servoArmDebug
    - variant: geckoReleaseAarch64Debug
    - variant: geckoReleaseX86Debug
    - variant: geckoReleaseArmDebug
```

pocmo:
```Groovy
task printVariants {
    doLast {
        subprojects.each { project ->
            println "Looking at: " + project.name

            if (project.hasProperty('android') && project.name.startsWith("samples-")) {
                println "  - is Android"

                project.android.applicationVariants.all { variant ->
                    println "    - variant: " + variant.name
                }
            }
        }
    }
}
```

grigoryk:
Doing something similar with https://github.com/mozilla-mobile/android-components/pull/1614, might as well take this on too.

pocmo:
Since filing this we started doing that for the reference browser too :)

PR #1614, opened 2018-12-21
Run lint only for a single variant

grigoryk:
This should speed up our TC tasks significantly. In local tests, this speeds up 'sample-browser' task by about 40%. I imagine there will be similar gains for every other module.


grigoryk:
My rough baseline (looking at green TC runs from the recent past) is about 25 minutes for `sample-browser` and about 8-9 minutes for other modules.

grigoryk:
From a productivity perspective we mostly just care about how long the longest task takes - which is the `samples-browser`, and it's longer by about 100-200% than others. It'll be nice to get the rest of them faster from an environmental/cost reduction perspective though.

grigoryk:
(one task OOM'd, but that seems unrelated).

samples-browser task run was 18m, down from 25m. Others seem to have went down a bit as well.

grigoryk:
Okay, now it's just hitting intermittents.

A sample run before (timings in minutes):
min: 0.18
max: 25.33
70th percentile: 12.5
99th percentile: 21.17

A sample run after:
min: 0.2
max: 17.97
70th percentile: 9.14
99th percentile: 16.64

grigoryk:
This will happen in https://github.com/mozilla-mobile/android-components/issues/1293 instead.

PR #1615, opened 2018-12-21
Closes #496: Adding support for html input types file for SystemEngine

### LINKED ISSUE #496: Add support for file uploads
In GeckoView this may work out of the box. In WebView we need to implement a callback and then show a file chooser.

Amejia481:
[Here](https://drive.google.com/file/d/1LJZpB-cXxbofXZCM_M-d_u8Aylm1owf8/view). you can take a look at this feature.

To be able to test this feature, I had to comment the `sessionStorage.autoSave` on `DefaultComponents.kt`, because in the `systemUniversal` variant, the app gets frozen every time that leave and return to the app.

```kotlin
//            sessionStorage.autoSave(this)
//                .periodicallyInForeground(interval = 30, unit = TimeUnit.SECONDS)
//                .whenGoingToBackground()
//                .whenSessionsChange()
```

### DIFF
--- a/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
+++ b/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
@@ -21,6 +21,7 @@ import android.webkit.PermissionRequest
 import android.webkit.SslErrorHandler
 import android.webkit.ValueCallback
 import android.webkit.WebChromeClient
+import android.webkit.WebChromeClient.FileChooserParams.MODE_OPEN_MULTIPLE
 import android.webkit.WebResourceError
 import android.webkit.WebResourceRequest
 import android.webkit.WebResourceResponse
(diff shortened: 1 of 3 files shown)


components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -322,6 +324,47 @@ class SystemEngineView @JvmOverloads constructor(
             session?.internalNotifyObservers { onContentPermissionRequest(SystemPermissionRequest(request)) }
         }
 
+        override fun onShowFileChooser(
+            webView: WebView?,
+            filePathCallback: ValueCallback<Array<Uri>>?,
+            fileChooserParams: FileChooserParams?
```
csadilek: nit: Doesn't look like these params can be null. This will also get the test coverage up which is otherwise fine!

csadilek:
Let's land this once green 🚢 !

PR #1618, opened 2018-12-22
Remove ErrorPage dependency

jonalmeida:
This was from a previous patch which had an ErrorType that existed in
the errorpages feature but was later hidden internally within all the
engine implementations.


jonalmeida:
Made an error, we moved this to the request interceptor which still depends on the `ErrorType`. Closing for now until we have a better solution.

PR #1619, opened 2018-12-22
Clean only module before building.

jonalmeida:


jonalmeida:
Looking at @grigoryk's lint savings (https://github.com/mozilla-mobile/android-components/pull/1614), along with this, might make the sample builds a lot faster! 🎉 

Sample run before:
min: 3.53
max: 29.12
70th percentile: 14.32
99th percentile: 23.13

Sample run after:
min: 0.2
max: 25.12
70th percentile: 9.84
99th percentile: 17.68

These results aren't very accurate though since the original max can vary from (25-30 mins).

jonalmeida:
I'm not sure if there are downsides to doing this though so maybe someone with a better understanding of our build system could shed more light. :)

jonalmeida:
This was a test PR based on the conversation we were having in-channel at the time. It sounds like you're already working a more robust solution.

PR #1623, opened 2018-12-27
[glean] 1513563: Log pings to the console.

mdboom:
This works.  However, it's currently being turned on for debug builds, and off for release builds, which is maybe not exactly how we want this to work.  Once we determine how we want this turned on/off, it should be pretty easy to make that change.

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/net/HttpPingUploader.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/net/HttpPingUploader.kt
@@ -12,6 +12,8 @@ import java.io.IOException
 import java.net.HttpURLConnection
 import java.net.MalformedURLException
 import java.net.URL
+import org.json.JSONException
+import org.json.JSONObject
 
 /**
  * A simple ping Uploader, which implements a "send once" policy, never
@@ -20,6 +22,29 @@ import java.net.URL
 class HttpPingUploader(configuration: Configuration) : PingUploader {
     private val config = configuration
     private val logger = Logger("glean/HttpPingUploader")
+
+    /**
+     * Log the contents of a ping to the console, if configured to do so in
+     * config.logPings.
+     *
+     * @param path the URL path to append to the server address
+     * @param data the serialized text data to send
+     */
+    private fun logPing(path: String, data: String) {
+        if (config.logPings) {
+            // Parse and reserialize the JSON so it has indentation and is human-readable.
+            var indented = try {
+                var json = JSONObject(data)
+                json.toString(2)
+            } catch (e: JSONException) {
+                logger.debug("Exception parsing ping as JSON: $e")
+                null
+            }
+            indented?.let {
+                logger.debug("Glean ping to URL: ${path}\n$it")
+            }
+        }
+    }
 
     /**
      * Synchronously upload a ping to Mozilla servers.
@@ -36,6 +61,8 @@ class HttpPingUploader(configuration: Configuration) : PingUploader {
      */
     @Suppress("ReturnCount", "MagicNumber")
     override fun upload(path: String, data: String): Boolean {
+        logPing(path, data)
+
         var connection: HttpURLConnection? = null
         try {
             connection = openConnection(config.serverEndpoint, path)
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/net/HttpPingUploader.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/net/HttpPingUploader.kt
@@ -37,8 +37,8 @@ class HttpPingUploader(configuration: Configuration) : PingUploader {
                 var json = JSONObject(data)
                 json.toString(2)
             } catch (e: JSONException) {
-                logger.debug("Exception parsing ping as JSON: $e")
-                null
+                logger.debug("Exception parsing ping as JSON: $e") // $COVERAGE-IGNORE$
+                null // $COVERAGE-IGNORE$
             }
             indented?.let {
                 logger.debug("Glean ping to URL: ${path}\n$it")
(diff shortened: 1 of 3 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -70,7 +72,11 @@ open class GleanInternalAPI {
         storageEngineManager = StorageEngineManager(applicationContext = applicationContext)
         pingMaker = PingMaker(storageEngineManager)
         this.configuration = configuration
-        httpPingUploader = HttpPingUploader(configuration)
+        if (BuildConfig.DEBUG) {
+            pingUploader = LogPingUploader(configuration)
+        } else {
+            pingUploader = HttpPingUploader(configuration)
+        }
```
pocmo: Note that `BuildConfig.DEBUG` means it is the `debug` variant of the library. Apps consuming the library from maven.mozilla.org will always get a release artifact and this path will never get triggered.
That's different for the sample app in the repository: It knows of the variants of the library and If you build the "debug" variant  of the app it will use a matching "debug" variant of the library.

mdboom: Got it.  So this is definitely not the way we want to do it.

Do you have any suggestions for creating a flag that developers could easily flip to turn this on/off?  It's not just as simple as just changing logging settings -- we also want to *not* send pings over Http when debugging ping content...

pocmo: That's a good question.

If you really want to attach this to the debug flag: You can read that at runtime. Here is an example: https://medium.com/@elye.project/checking-debug-build-the-right-way-d12da1098120

In the old telemetry library this was mostly left to the app and via the configuration the app could either:
* Enable/disable telemetry collection/uploading
* Or: Set a different uploader (that only logs instead of actually uploads)

I like your approach of hiding this behind a flag. I wonder if this is something you could just expose and leave it to the app to decide what builds need this setup?

Dexterp37: I would also slightly change the behaviour compared to the old telemetry a little bit: I would make it possible to dump pings to console in any case, without necessarily dropping the http uploader. This would allow easier debugging of weirder things on release. What do you think, @pocmo ?

pocmo: Yes, absolutely! And in addition to that: I would like to see events at the time they happen :)

> 
> Yes, absolutely! And in addition to that: I would like to see events at the time they happen :)

This could get a bit verbose, but we can consider doing it! It only takes a single log line :)

@mdboom , given the previous discussion, I think this PR can really boil down to just dump to logcat in the http uploader, in addition to uploading, if the configuration flag is true.

mdboom:
I've updated this to hopefully reflect the feedback here.

components/service/glean/src/main/java/mozilla/components/service/glean/net/HttpPingUploader.kt
```diff
@@ -36,71 +37,82 @@ class HttpPingUploader(configuration: Configuration) : PingUploader {
...
-                    // successful upload to the service.
-                    logger.error("Server returned client error code: $responseCode")
-                    return true
-                }
-                else -> {
-                    // Known other errors:
-                    // 500 - internal error
+        if (config.logPings) {
+            // Parse and reserialize the JSON so it has indentation and is human-readable.
+            var json = JSONObject(data)
```
Dexterp37: This and the next line can throw. I'll wrap them in a try catch and watch out for JSONException.

components/service/glean/src/main/java/mozilla/components/service/glean/net/HttpPingUploader.kt
```diff
@@ -36,71 +37,82 @@ class HttpPingUploader(configuration: Configuration) : PingUploader {
...
+            var indented = json.toString(2)
+            logger.debug("Glean ping to URL: ${path}\n$indented")
+        }
+
+        config.serverEndpoint?.let {
```
Dexterp37: Why do we need this? We will always have an endpoint.

components/service/glean/src/main/java/mozilla/components/service/glean/EventMetricType.kt
```diff
@@ -102,6 +102,13 @@ data class EventMetricType(
             eventKeys
         }
 
+        if (Glean.configuration.logEvents) {
```
Dexterp37: Let's leave this out from this PR for now. Let's consider doing this as part of future improvement, when needed (do you mind filing a bug?).

mdboom: https://bugzilla.mozilla.org/show_bug.cgi?id=1518505

mdboom:
Ok.  I've pared this down to just the minimum.

mdboom:
@pocmo: Would you mind having another review pass at this?

components/service/glean/src/main/java/mozilla/components/service/glean/net/HttpPingUploader.kt
```diff
@@ -21,6 +23,29 @@ class HttpPingUploader(configuration: Configuration) : PingUploader {
...
+            var indented = try {
+                var json = JSONObject(data)
+                json.toString(2)
+            } catch (e: JSONException) {
+                logger.debug("Exception parsing ping as JSON: $e") // $COVERAGE-IGNORE$
+                null // $COVERAGE-IGNORE$
+            }
+            indented?.let {
+                logger.debug("Glean ping to URL: ${path}\n$it")
+            }
```
pocmo: nit: The code may be a bit shorter if you just log from inside the `try` block without returning anything? :)

PR #1632, opened 2018-12-31
Closes #1603: Add switch menu item.

### LINKED ISSUE #1603: Menu: Add switch item
We do have a checkbox item. We do not have a switch item yet. :)

jonalmeida:


jonalmeida:
> 81.25% of diff hit (target 82.73%) 

I think this patch has enough tests that cover this class, it's the duplication of the default values for the `initialState` that is mostly showing up as untested.

PR #1634, opened 2018-12-31
Closes #1633: Add Bitmap.withRoundedCorners.

### LINKED ISSUE #1633: Add Bitmap.withRoundedCorners function
We're using this in FFES: https://github.com/mozilla-mobile/firefox-echo-show/blob/986826e0782b3df4babde905b4549feec438d28f/app/src/main/java/org/mozilla/focus/ext/Bitmap.kt#L37

And seems like a good utility to share across all of our apps.

mcomella:
I verified this was working correctly by adding code locally to the
sample browser.

screenshot_1546288023

### DIFF
--- a/components/support/ktx/src/test/java/mozilla/components/support/ktx/android/graphics/BitmapKtTest.kt
+++ b/components/support/ktx/src/test/java/mozilla/components/support/ktx/android/graphics/BitmapKtTest.kt
@@ -1,0 +1,49 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.support.ktx.android.graphics
+
+import android.graphics.Bitmap
+import android.graphics.Color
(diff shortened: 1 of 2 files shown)


components/support/ktx/src/test/java/mozilla/components/support/ktx/android/graphics/BitmapKtTest.kt
```diff
@@ -0,0 +1,49 @@
...
+import org.junit.Assert.assertNotSame
+import org.junit.Ignore
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
+class BitmapKtTest {
+
+    @Ignore // TODO: convert to integration test. Robolectric's shadows are incomplete and cause this to fail.
```
pocmo: That's something we could do in a follow-up, right?

mcomella: Yeah – it doesn't look like components supports integration tests right now so I didn't try to set it up. Filed https://github.com/mozilla-mobile/android-components/issues/1637

PR #1635, opened 2019-01-02
Closes #1195: Move WebView reference to SystemEngineSession

### LINKED ISSUE #1195: Refactor browser-engine-system to keep WebView reference inside SystemEngineSession

csadilek:
Basically, this moves the WebView to `SystemEngineSession` and removes the circular dependency between SystemEngineView and SystemEngineSession and the corresponding weak reference. 

As hoped, this simplified things quite a bit. Nice net negative LOC, plus two fixed issues:

We can now switch back and forth between tabs (no more white tabs when using the system engine), and there was also an occasional crash with window requests that's fixed by this as well (both were caused by calling `loadUrl` on the WebView at the wrong time).

I kept the view logic in the `SystemEngineView`.

Small changeset (except the tests). Let's see what codecov thinks.


csadilek:
Yay codecov is fine, but one thing to double-check in Focus/FireTv is that SystemEngineSessions are only created on the UI thread (I tested our samples which were fine.) Now that we're creating the WebView when the EngineSession is created (as opposed to when the view is created), we have to watch out for that. WebView doesn't allow interactions from multiple threads. Of course, we could go back to lazily intializing the WebView (triggered by the view), but then I am not sure we gain much from this refactoring?

samples/browser/src/main/java/org/mozilla/samples/browser/DefaultComponents.kt
```diff
@@ -62,9 +61,9 @@ open class DefaultComponents(private val applicationContext: Context) {
             }
 
             sessionStorage.autoSave(this)
-                .periodicallyInForeground(interval = 30, unit = TimeUnit.SECONDS)
+                // .periodicallyInForeground(interval = 30, unit = TimeUnit.SECONDS)
                 .whenGoingToBackground()
-                .whenSessionsChange()
+                // .whenSessionsChange()
```
pocmo: I guess we do not want to land this part? :)

csadilek:
I am going to test these changes on FireTV now (before merging.)

pocmo:
> but one thing to double-check in Focus/FireTv 

There's one important thing we should make sure to not break: The Fire TV app has some extension functions defined that fish the `WebView` out of the `EngineView` to implement some functionality that we do not offer yet:
https://github.com/mozilla-mobile/firefox-tv/blob/master/app/src/main/java/org/mozilla/tv/firefox/ext/EngineView.kt

In the best case that just continues to work. However the extensions keep a reference here:
https://github.com/mozilla-mobile/firefox-tv/blob/master/app/src/main/java/org/mozilla/tv/firefox/ext/EngineView.kt#L140

Instead of the reference we may want to look up the WebView dynamically every time?

Another option may be to implement this workaround in a way that they can get the WebView from the `(System)EngineSession`?

The long term solution is to get rid of the extensions as much as possible and upstream the functionality. Severin created a list here:
https://github.com/mozilla-mobile/android-components/issues/1616

samples/browser/src/main/java/org/mozilla/samples/browser/DefaultComponents.kt
csadilek: Nope. :) That's just to prevent the app from freezing when testing. We can schedule a storage/autosave fix for the next sprint....

csadilek:
@pocmo looking good! The easiest way to make things work for FireTV is to expose the `webView` on `SystemEngineSession` (just the impl, not the interface) temporarily until https://github.com/mozilla-mobile/android-components/issues/1616 is resolved. I've added this change to this PR.

Then in FFTV it only needs this one-line change:
https://github.com/csadilek/firefox-tv/commit/07a10bd3f368df1fc8bf6956827314e3b4ce631e

Tested quite a bit and didn't see any problems so far. Wdyt? Should we land it?

pocmo:
> Tested quite a bit and didn't see any problems so far. Wdyt? Should we land it?

Yes, that looks great!

> Of course, we could also still decide to do that, but it doesn't affect this PR. Basically, we just need to get the WebView from the session now as the EngineView won't have it until render is called.

It doesn't affect this PR, so 🚢 👍 

However I think there may still be an issue with caching the WebView in Fire TV with the new code: Previously it was **always** the same `WebView` that was used by `EngineView`. Now it can be different ones based on the `Session`. Even though FFfTV is a single-tab browser it still creates new sessions at times. That may create a new EngineView (or fragment) as well and that may save us here though...

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
```diff
@@ -39,76 +38,73 @@ internal val additionalHeaders = mapOf(
...
+    @Volatile internal var currentUrl = ""
+    @Volatile internal var fullScreenCallback: WebChromeClient.CustomViewCallback? = null
+
+    // This is public for FFTV which needs access to the WebView instance. We can mark it internal once
+    // https://github.com/mozilla-mobile/android-components/issues/1616 is resolved.
+    @Volatile var webView: WebView = NestedWebView(context)
+        set(value) {
+            field = value
+            initSettings()
+        }
```
pocmo: Does this need to be `var`? Would it be possible to only expose a `val` outside of the component (for FFTV?). Can we add a ⚠️☢️☣️💥🧨 warning KDoc that this property is only a workaround and should not be used? :)

csadilek: Yeah, we need this mutable for the window request support, but we can hide it (make the setter internal). Updated!

csadilek:
> That may create a new EngineView (or fragment) as well and that may save us here though...

Yes, but with my earlier changes we would've ended up with the same `WebView` instance. So, we need to make sure we always have the right one from the currently selected session. Required changes in FireTV: https://github.com/csadilek/firefox-tv/commit/d284b72d73699979f78d1aed5a1176645dffa811

Will do some more testing, but so far so good.

PR #1636, opened 2019-01-02
Closes #1504: Add support for alert popup in WebView

### LINKED ISSUE #1504: Add support for alert popup in WebView.
This issue includes adding for handling when a session should be prevented for showing more alerts.

Amejia481:
@jonalmeida, @csadilek and @pocmo. I want hear what do you think about this is this is an early implementation for preventing a page showing many JS alerts. This implementation is already provided by GeckoView for this reason we have to provide a similar one on SystemView.

Here a small [demo](https://drive.google.com/file/d/18G15NiBfMu44arAhCHTJ7lWmb1mj0q_n/view)

### DIFF
--- a/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
+++ b/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
@@ -17,6 +17,7 @@ import android.util.AttributeSet
 import android.view.View
 import android.webkit.CookieManager
 import android.webkit.DownloadListener
+import android.webkit.JsResult
 import android.webkit.PermissionRequest
 import android.webkit.SslErrorHandler
 import android.webkit.ValueCallback
@@ -47,6 +48,7 @@ import mozilla.components.concept.engine.prompt.PromptRequest
 import mozilla.components.concept.engine.request.RequestInterceptor.InterceptionResponse
 import mozilla.components.support.ktx.android.content.isOSOnLowMemory
 import mozilla.components.support.utils.DownloadUtils
+import java.util.Date
 
 /**
  * WebView-based implementation of EngineView.
@@ -59,6 +61,9 @@ class SystemEngineView @JvmOverloads constructor(
 ) : FrameLayout(context, attrs, defStyleAttr), EngineView, View.OnLongClickListener {
 
     private var session: SystemEngineSession? = null
+    internal var jsAlertCount = 0
+    internal var shouldShowMoreDialogs = true
+    internal var lastDialogShownAt = Date()
 
     /**
      * Render the content of the given session.
@@ -123,6 +128,7 @@ class SystemEngineView @JvmOverloads constructor(
                     onNavigationStateChange(view.canGoBack(), view.canGoForward())
                 }
             }
+            resetJSAlertAbuseState()
         }
 
         override fun onPageFinished(view: WebView?, url: String?) {
@@ -300,6 +306,44 @@ class SystemEngineView @JvmOverloads constructor(
             session?.internalNotifyObservers { onContentPermissionRequest(SystemPermissionRequest(request)) }
         }
 
+        override fun onJsAlert(view: WebView, url: String?, message: String?, result: JsResult): Boolean {
+            val session = session ?: return super.onJsAlert(view, url, message, result)
+
+            // When an alert is triggered from a iframe, url is equals to about:blank, using currentUrl as a fallback.
+            val safeUrl = if (url.isNullOrBlank()) {
+                session.currentUrl
+            } else {
+                if (url.contains("about")) session.currentUrl else url
+            }
+
+            val title = context.getString(R.string.mozac_browser_engine_system_alert_title, safeUrl)
+
+            val onDismiss: () -> Unit = {
+                result.cancel()
+            }
+
+            if (shouldShowMoreDialogs) {
+
+                session.notifyObservers {
+                    onPromptRequest(
+                        PromptRequest.Alert(
+                            title,
+                            message ?: "",
+                            areDialogsBeingAbused(),
+                            onDismiss
+                        ) { shouldNotShowMoreDialogs ->
+                            shouldShowMoreDialogs = !shouldNotShowMoreDialogs
+                            result.confirm()
+                        })
+                }
+            } else {
+                result.cancel()
+            }
+
+            updateJSDialogAbusedState()
+            return true
+        }
+
         override fun onShowFileChooser(
             webView: WebView?,
             filePathCallback: ValueCallback<Array<Uri>>?,
@@ -449,7 +493,46 @@ class SystemEngineView @JvmOverloads constructor(
 
     override fun canScrollVerticallyDown() = session?.webView?.canScrollVertically(1) ?: false
 
+    private fun resetJSAlertAbuseState() {
+        jsAlertCount = 0
+        shouldShowMoreDialogs = true
+    }
+
+    internal fun updateJSDialogAbusedState() {
+        if (!areDialogsAbusedByTime()) {
+            jsAlertCount = 0
+        }
+        ++jsAlertCount
+        lastDialogShownAt = Date()
+    }
+
+    internal fun areDialogsBeingAbused(): Boolean {
+        return areDialogsAbusedByTime() || areDialogsAbusedByCount()
+    }
+
+    @Suppress("MagicNumber")
+    internal fun areDialogsAbusedByTime(): Boolean {
+        return if (jsAlertCount == 0) {
+            false
+        } else {
+            val now = Date()
+            val diffInSeconds = (now.time - lastDialogShownAt.time) / 1000 // 1 second equal to 1000 milliseconds
+            diffInSeconds < MAX_SUCCESSIVE_DIALOG_SECONDS_LIMIT
+        }
+    }
+
+    internal fun areDialogsAbusedByCount(): Boolean {
+        return jsAlertCount > MAX_SUCCESSIVE_DIALOG_COUNT
+    }
+
     companion object {
+
+        // Maximum number of successive dialogs before we prompt users to disable dialogs.
+        internal const val MAX_SUCCESSIVE_DIALOG_COUNT: Int = 2
+
+        // Minimum time required between dialogs in seconds before enabling the stop dialog.
+        internal const val MAX_SUCCESSIVE_DIALOG_SECONDS_LIMIT: Int = 3
+
         @Volatile
         internal var URL_MATCHER: UrlMatcher? = null
 
--- a/components/browser/engine-system/src/main/res/values/strings.xml
+++ b/components/browser/engine-system/src/main/res/values/strings.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<resources>
+    <!-- Text for a title dialog on web page. -->
+    <string name="mozac_browser_engine_system_alert_title">The page at %1$s says:</string>
+</resources>
--- a/components/browser/engine-system/src/test/java/mozilla/components/browser/engine/system/SystemEngineViewTest.kt
+++ b/components/browser/engine-system/src/test/java/mozilla/components/browser/engine/system/SystemEngineViewTest.kt
@@ -12,6 +12,7 @@ import android.net.http.SslError
 import android.os.Bundle
 import android.os.Message
 import android.view.View
+import android.webkit.JsResult
 import android.webkit.SslErrorHandler
 import android.webkit.WebChromeClient
 import android.webkit.WebResourceError
@@ -24,6 +25,8 @@ import android.webkit.WebView.HitTestResult
 import androidx.test.core.app.ApplicationProvider
 import androidx.test.core.app.ApplicationProvider.getApplicationContext
 import kotlinx.coroutines.runBlocking
+import mozilla.components.browser.engine.system.SystemEngineView.Companion.MAX_SUCCESSIVE_DIALOG_COUNT
+import mozilla.components.browser.engine.system.SystemEngineView.Companion.MAX_SUCCESSIVE_DIALOG_SECONDS_LIMIT
 import mozilla.components.browser.engine.system.matcher.UrlMatcher
 import mozilla.components.browser.errorpages.ErrorType
 import mozilla.components.concept.engine.EngineSession
@@ -55,6 +58,10 @@ import org.mockito.Mockito.verify
 import org.mockito.Mockito.verifyZeroInteractions
 import org.robolectric.RobolectricTestRunner
 import org.robolectric.RuntimeEnvironment
+import java.util.Calendar
+import java.util.Calendar.SECOND
+import java.util.Calendar.YEAR
+import java.util.Date
 
 @RunWith(RobolectricTestRunner::class)
 class SystemEngineViewTest {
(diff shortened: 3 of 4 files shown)


components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
```diff
@@ -47,6 +47,8 @@ class SystemEngineSession(private val defaultSettings: Settings? = null) : Engin
     @Volatile internal var trackingProtectionPolicy: TrackingProtectionPolicy? = null
     @Volatile internal var webFontsEnabled = true
     @Volatile internal var internalSettings: Settings? = null
+    internal var alertJSCount = 0
```
jonalmeida: Probably needs `@Volatile` for these attributes as well.

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -324,6 +325,34 @@ class SystemEngineView @JvmOverloads constructor(
...
         }
 
+        override fun onJsAlert(view: WebView?, url: String?, message: String?, result: JsResult): Boolean {
+            session?.apply {
+                val hasShownManyDialogs = (++alertJSCount) > MAX_JS_ALERT
+                val onDismiss: () -> Unit = {
+                    result.cancel()
+                }
+
+                if (shouldShowNoMoreJSDialogs) {
```
jonalmeida: If you _should not_ show more dialogs, then you notify observers?

subjective nit: `shouldShowDialogs` seems easier to read.

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -324,6 +325,34 @@ class SystemEngineView @JvmOverloads constructor(
...
+
+                    notifyObservers {
+                        onPromptRequest(
+                            PromptRequest.Alert(
+                                "",
+                                message ?: "",
+                                hasShownManyDialogs,
```
jonalmeida: subjective nit: maybe something like `showMultipleDialogs`, `tooManyDialogs`? 4+ words make it hard to grok for me at least. 😅

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
Amejia481: 👍

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
Amejia481: I had it locally :P, didn't notice that I didn't pushed it

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
```diff
@@ -369,6 +377,11 @@ class SystemEngineSession(private val defaultSettings: Settings? = null) : Engin
         super.notifyObservers(block)
     }
 
+    private fun resetJSAlertData() {
+        alertJSCount = 0
+        shouldShowMoreJSDialogs = true
+    }
+
```
pocmo: I wonder if we can keep this out of the Session and do that in SystemEngineView (or a referenced helper) internally. In SystemEngineView we do have the callbacks from the clients where we may be able to reset certain things (instead of in the actions)? Is that feasible?

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -324,6 +325,44 @@ class SystemEngineView @JvmOverloads constructor(
             session?.internalNotifyObservers { onContentPermissionRequest(SystemPermissionRequest(request)) }
         }
 
+        override fun onJsAlert(view: WebView?, url: String?, message: String?, result: JsResult): Boolean {
+            session?.apply {
```
pocmo: `session?.apply {}` is very misleading here since you are not really applying something to the session. Since `this` in the block is the `Session` it is really hard to know whether `url` for example accesses the parameter or the url inside `Session`.

If this is just for nullability reasons then I'd prefer something short like:

```Kotlin
val session = session ?: return
```

What do you think?

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -324,6 +325,44 @@ class SystemEngineView @JvmOverloads constructor(
             session?.internalNotifyObservers { onContentPermissionRequest(SystemPermissionRequest(request)) }
         }
 
+        override fun onJsAlert(view: WebView?, url: String?, message: String?, result: JsResult): Boolean {
```
pocmo: Can all those parameters really be null?

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -324,6 +325,44 @@ class SystemEngineView @JvmOverloads constructor(
...
+            session?.apply {
+
+                val safeUrl = Uri.parse(
+                    if (url.isNullOrBlank()) {
+                        currentUrl
+                    } else {
+                        if (url.contains("about")) currentUrl else url
+                    }
+                )
```
pocmo: I do not really understand what this is doing. Why the "about" check? This block may need a comment :)

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -474,6 +513,9 @@ class SystemEngineView @JvmOverloads constructor(
     override fun canScrollVerticallyDown() = currentWebView.canScrollVertically(1)
 
     companion object {
+
+        internal const val MAX_JS_ALERT: Int = 2
```
pocmo: I feel like this needs some kind of timer to measure the time between alerts. `3` alerts are fine too if they for example are triggered by user actions. However X alerts in Y seconds is almost always something you do not want. Maybe it's worth exploring the Gecko internals to see how the flag gets set there?

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
Amejia481: Unfortunately, this Apis are not annotated for null-availability and the [docs](https://developer.android.com/reference/android/webkit/WebChromeClient.html#onJsAlert(android.webkit.WebView,%20java.lang.String,%20java.lang.String,%20android.webkit.JsResult)) don't tell much neither :(

jonalmeida: @Amejia481 I [looked into this before](https://github.com/mozilla-mobile/android-components/pull/1466#discussion_r236026122) for nullability; the WebView is non-null for sure.

Amejia481: Thanks @jonalmeida !

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
Amejia481: Yes, Gecko is using a time-based implementation, I was trying to find the code that does that but was a daunting path, I will give it a second chance.

Amejia481: Many `.jsm`,  `.h` and `.cpp` files :P

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
Amejia481: Is this not an expected behavior ?

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
Amejia481: Yeah,  initially I had everything on `SystemEngineView` but couldn't find callbacks for `goBack`,  `goForward` and `reload`, but I believe that resetting on `onPageStarted` will be sufficient, what do you think?

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
pocmo: Heh, yeah, that's how it feels to work on Fennec at times. If you do not know about it yet, I mostly read the code via this page: https://searchfox.org/

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
Amejia481: Yes, I did it only for null-availability purpose, it will go with your approach thanks!

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
pocmo: Yeah, I assume that should be good enough.

Maybe another option would be to look at the URL in `onJsAlert()` and reset if the callback is invoked for a new URL?

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
Amejia481: Thanks, Now I know that I'm not alone :)

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
Amejia481: I'm getting the `url` from `onJsAlert`. At least in this context `session` doesn't expose an `url`
image

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
Amejia481: That could be a good approach too, but `url` that I'm getting from `onJsAlert` is  always **"about:blank"**. Is this an expected behavior?

pocmo: This doesn't sound right. Is this reproducible with a plain WebView? Hopefully we are not running into a WebView/Chrome bug :)

Amejia481: The [web page that I was using for testing ](https://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_alert) was triggering the alert from an `<iframe>` for this reason the url has the value of **"about:blank"** in other cases is returning the correct url.

Chrome
image

Fennec
image

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
Amejia481: [I found it ](https://dxr.mozilla.org/mozilla-central/source/dom/base/nsGlobalWindowInner.cpp#2026) 🎉 
Working on a similar solution on kt :)

Amejia481:
PR updated, with tests and a time based implementation for blocking abuse js alerts.

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -471,9 +516,51 @@ class SystemEngineView @JvmOverloads constructor(
...
+    @Suppress("MagicNumber")
+    internal fun areDialogsBeingAbusedByTime(): Boolean {
+        return if (alertJSCount == 0) {
+            false
+        } else {
+            val now = Date()
+            val diffInSeconds =
+                (now.time - lastTimeWhenAJSDialogWasShown.time) / 1000 // 1 second equal to 1000 milliseconds
+
+            diffInSeconds > MAX_SUCCESSIVE_DIALOG_SECONDS_LIMIT
```
csadilek: Hm interesting. Is this the right logic? We're saying if there's too much time between alerts then the site is abusing it. I would have expected this to be the inverse. If the time between alerts is really short i.e. it shows alerts in short succession then that's what we want to prevent?

nit: I also don't quite like the `abuseBy` naming. Maybe we can find alternatives e.g. `dialogsWithinTimeLimit` or `dialogsWithinLimit` etc. Not sure.

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -62,6 +64,10 @@ class SystemEngineView @JvmOverloads constructor(
     internal var currentUrl = ""
     private var session: SystemEngineSession? = null
     internal var fullScreenCallback: WebChromeClient.CustomViewCallback? = null
+    internal var alertJSCount = 0
+    internal var userWantToShowMoreDialogs = true
```
csadilek: nit: maybe just `showMoreDialogs`

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -62,6 +64,10 @@ class SystemEngineView @JvmOverloads constructor(
...
+    internal var lastTimeWhenAJSDialogWasShown = Date()
```
csadilek: nit: mayne just `lastDialogShownAt`?

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -471,9 +516,51 @@ class SystemEngineView @JvmOverloads constructor(
...
+    }
+
     override fun canScrollVerticallyDown() = currentWebView.canScrollVertically(1)
 
     companion object {
+
+        // Maximum number of successive dialogs before we prompt users to disable dialogs.
+        internal const val MAX_SUCCESSIVE_DIALOG_COUNT: Int = 2
+
+        // Maximum Amount of seconds allowed between js alert before enabling the stop dialog.
```
csadilek: // Minimum time required between dialogs in seconds ... would make this clearer see comment above about the logic being inverted.

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
Amejia481: @csadilek `diffInSeconds` tells how many seconds has passed from the actual time and the last time that dialog was shown. If the amount of seconds is bigger that limit (`MAX_SUCCESSIVE_DIALOG_SECONDS_LIMIT`) then the page is being abusive.

For instance if you have:
now = 11:59:40
lastDialog = 11:59:32

Then **diffInSeconds** will be **8**, 8 is bigger than 

`8  > MAX_SUCCESSIVE_DIALOG_SECONDS_LIMIT (3) ` then page is been abusive.

 ```kotlin
   @Test
    fun `are dialogs being abused by time`() {
        val context = ApplicationProvider.getApplicationContext<Context>()
        val engineView = SystemEngineView(context)

        with(engineView) {

            assertFalse(areDialogsBeingAbusedByTime())

            alertJSCount = 1

            // Decreasing lastTimeWhenAJSDialogWasShown by a year
            lastTimeWhenAJSDialogWasShown = lastTimeWhenAJSDialogWasShown.add(SECOND, - (MAX_SUCCESSIVE_DIALOG_SECONDS_LIMIT+1))

            assertTrue(areDialogsBeingAbusedByTime())
        }
    }
```
If the algorithm is wrong please let me know :)

Comment (Amejia481) [reply components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt]: @jonalmeida   I had to change the logic a bit because they were doing something that I don't like. Changing state in a method that was supposed to give a value, in my opinion, this is misleading.

```c
TimeDuration dialogInterval(TimeStamp::Now() - mLastDialogQuitTime);
  if (dialogInterval.ToSeconds() <
      Preferences::GetInt("dom.successive_dialog_time_limit",
                          DEFAULT_SUCCESSIVE_DIALOG_TIME_LIMIT)) {
    mDialogAbuseCount++;

    return GetPopupControlState() > openAllowed ||
           mDialogAbuseCount > MAX_SUCCESSIVE_DIALOG_COUNT;
  } 
```

For this reason the operator are inverted, `areDialogsBeingAbusedByTime()` is just querying and `updateJSDialogAbusedState()` is mutating state.

jonalmeida: > If the amount of seconds is **bigger that limit** (MAX_SUCCESSIVE_DIALOG_SECONDS_LIMIT) then the page is being abusive.

I think you mean if the seconds are _less than_ the limit right? That's the abusive behaviour.

If there are a long delays between the two dialog requests this should be considered reasonable behaviour.

Amejia481: @csadilek and @jonalmeida  you are right, I will inverter the condition, thanks for catching it up. I was taking **MAX_SUCCESSIVE_DIALOG_SECONDS_LIMIT** as a max diff between the too times.

Amejia481:
PR updated

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -471,9 +516,48 @@ class SystemEngineView @JvmOverloads constructor(
...
+    internal fun updateJSDialogAbusedState() {
+        if (!areDialogsWithTimeLimit()) {
+            jsAlertCount = 0
+        }
+        ++jsAlertCount
+        lastDialogShownAt = Date()
+    }
+
+    internal fun areDialogsBeingAbused(): Boolean {
+        return areDialogsWithTimeLimit() || areDialogsWithinCountLimit()
```
csadilek: @Amejia481 yes, that fixed the logic! However, the new names have inverse meaning now:

You're asking `areDialogsBeingAbused` here and that should be the case if dialogs are over (not within) the time limit `or` over (not within) the count limit. Easiest fix: let's keep this new fixed logic and just rename back to `areDialogsAbusedByTime` and `areDialogsAbusedByCount`. Alternatively `areDialogsOverTimeLimit` and `areDialogsOverCountLimit` would also work. So, logic is fine now, but naming needs an update :). Then 🚢 !

Amejia481: 👍

Amejia481:
@csadilek  PR updated: with new webview refactor and names updated

PR #1536, opened 2018-11-30
Use `android-components-g` Taskcluster worker type.

ncalexan:

PR #1638, opened 2019-01-02
Parallelize samples-browser variants; switch to 'android-compontents-g' worker type 

grigoryk:


grigoryk:
Prior to this, samples-browser would take around 25m. Now that the tasks are split up, it's looking much better:

selection_007

min: 0.21
max: 13.91
70th percentile: 9.35
99th percentile: 12.88

grigoryk:
One caveat about this change is that a greater number of tasks increases chance of some individual task getting stuck for whatever reason and needing to be rescheduled. I've hit that in an earlier green run with three tasks, and even though each successful run took at most 13 minutes, due to the timing of how tasks are timed out and re-run, total time to full resolution was about 35 minutes, which is worse than before.

Hopefully task runs aren't going to do that very often, and situation should improve further once we have control over our own worker type in #1536.

grigoryk:
Let's just switch to our owner worker-type already :) This is looking way faster now, maximum 6 minute runtime:

selection_008

grigoryk:
Changed my TC visualization extension to take into account timing from the moment a task is scheduled (not just look at the time it takes to execute it). This paints a more realistic picture, which is still pretty good!

selection_009

grigoryk:
`detekt` task seems to be failing consistently, but for slightly different reasons:
```
Failed to capture fingerprint of input files for task ':detekt' property 'input' during up-to-date check.
> Failed to create MD5 hash for file '/build/android-components/components/service/fretboard/build/intermediates/res/merged/release/values-lv/values-lv.xml' as it does not exist.
```

then again:
```
Failed to capture fingerprint of input files for task ':detekt' property 'input' during up-to-date check.
> Could not read path '/build/android-components/components/service/glean/build/bootstrap/Miniconda3/pkgs/ncurses-6.1-hf484d3e_0/include/ncursesw/unctrl.h'.
```

grigoryk:
For comparison, this the parallelization commit running using `github-worker`. Pretty consistent results - everything is fast, until a few task runs die and need to be re-scheduled, which just kills the "total time to completion" time.

selection_010

grigoryk:
That kind of worked! `detekt` passed, but another task failed with a somewhat similar IO problem (failed to delete /build/android-components while cleaning). Perhaps cleaning isn't interacting well with parallel execution, and some property of the new worker type is triggering that bad behaviour? (e.g. it has more cores, is fast enough, etc...).

Trying another approach: running `gradlew clean && gradlew {task}` as opposed to `gradlew clean {task}`, with parallel builds enabled. This should work if `clean` was interacting badly with the module task - although from the logs, it seems like `clean` is interacting badly with itself...

grigoryk:
This seems to be working out (TC is looking pretty green). Pushed a reviewable patch series.

grigoryk:
:green_heart: bam!

automation/taskcluster/decision_task.py
```diff
@@ -27,7 +27,7 @@
 
 
 def create_task(name, description, command, scopes = []):
-    return create_raw_task(name, description, "./gradlew --no-daemon clean %s" % command, scopes)
+    return create_raw_task(name, description, "./gradlew --no-daemon clean && ./gradlew --no-daemon %s" % command, scopes)
```
pocmo: What's the reason for running gradle twice? Isn't there some startup penalty since we are not using the daemon?

automation/taskcluster/decision_task.py
```diff
@@ -76,16 +76,89 @@ def create_raw_task(name, description, full_command, scopes = []):
...
+def create_module_tasks(module):
+    module_overrides = {
+        ":samples-browser": {
+            "assembleOnly/Test": (
+                ["ServoArm", "SystemUniversal"],
+                [
+                    "GeckoBetaAarch64", "GeckoBetaArm", "GeckoBetaX86",
+                    "GeckoNightlyAarch64", "GeckoNightlyArm", "GeckoNightlyX86",
+                    "GeckoReleaseAarch64", "GeckoReleaseArm", "GeckoReleaseX86"
+                ]
```
pocmo: Can we somehow make this dynamic like in the reference browser?

In theory could this be: library project: assemble, app project: assemble variants in parallel

automation/taskcluster/decision_task.py
```diff
@@ -76,16 +76,89 @@ def create_raw_task(name, description, full_command, scopes = []):
...
+                subtitle = subtitle + "AndLintDebug"
+                scheduled_lint = True
+
+            tasks.append(make_task(task, subtitle=subtitle + variant))
+
+        # If we failed to find a matching assemble/assembleTest task to sneak lint into, schedule a separate one for it.
+        # Overhead of a separate task just for linting is quite significant, but this should be a rare case.
+        if not scheduled_lint:
+            lint_task = override.get("lintTask", "lintDebug")
+            tasks.append(make_task(as_module_task(module, lint_task), subtitle="onlyLintDebug"))
```
pocmo: tbh: Since the modules can have different source sets / overrides per release type I'd feel more comfortable running `lint` for all by default. However if that speeds up things noticeable then that may be fine. But since we are shipping `release` shouldn't we lint release instead of debug?

pocmo:
This looks like something RelEng should look at too. I'll flag them. :)

pocmo:
Is getting a higher capacity (or more workers) are requirement for landing?

automation/taskcluster/decision_task.py
mitchhentges: I remember that older versions of gradle had an issue where `clean` wouldn't provide the same "fresh-slate" guarantees if `gradle` was run in parallel, maybe that problem still exists today?

ncalexan:
> Is getting a higher capacity (or more workers) are requirement for landing?

Probably not.  Looking at https://tools.taskcluster.net/aws-provisioner/android-components-g -- which is secured by default -- I see that we have capacity between 2 and 12 (inclusive).  My understanding is that is parallel jobs.  So we can run 12 jobs in parallel.  We have something like 60 tasks, IIRC, so we might want to bump that up, but I expect with a little tweaking we won't need to scale too much further.  In any case, #taskcluster can bump the definition easily.

automation/taskcluster/decision_task.py
grigoryk: @mitchhentges @pocmo I think that's what I'm seeing, yeah. I've captured my observations in the commit that made this change. Copy/pasting here:

> On the new worker type, the old pattern of running 'clean' as part of the
module task is interacting poorly with parallel builds, and we're getting
IO errors. This change ensures we wait until clean is done before doing
any other work.
> Locally this makes no difference (on Grisha's machine, at least) - and the
old pattern worked perfectly fine with the shared worker type.

grigoryk:
> Is getting a higher capacity (or more workers) are requirement for landing?

The scopes we currently have do require us to specify _a_ capacity (medium, high, low, very-high, etc). I can specify a different one if necessary.

automation/taskcluster/decision_task.py
grigoryk: Yeah, we could do that. I've elected a static list to have more control over what exactly we do (e.g. the split b/n test and assemble only tasks), but that of course comes at a cost of having to maintain this list. If things were fully dynamic, we either loose granularity, or need to implement support for it via some other means.

automation/taskcluster/decision_task.py
grigoryk: Linting everything is about half of the runtime of the samples-browser task, so it is a big overhead. I'll switch this to `release`.

automation/taskcluster/decision_task.py
```diff
@@ -76,16 +76,89 @@ def create_raw_task(name, description, full_command, scopes = []):
...
+        # Schedule "assemble and test" tasks; if a lintTask matches any variants and wasn't scheduled already, schedule it.
+        for variant in assemble_and_test:
+            task = as_module_task(module, "assemble" + variant) + " " + as_module_task(module, "test" + variant + "DebugUnitTest")
+            subtitle = "assembleAndTest"
+            if not scheduled_lint and variant in lint_task:
+                task = task + " " + as_module_task(module, lint_task)
+                subtitle = subtitle + "AndLintDebug"
+                scheduled_lint = True
+
+            tasks.append(make_task(task, subtitle=subtitle + variant))
```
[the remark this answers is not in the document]
grigoryk: Looks good to me, I'll break it apart a bit.

automation/taskcluster/decision_task.py
```diff
@@ -76,16 +76,89 @@ def create_raw_task(name, description, full_command, scopes = []):
...
+
+        # Schedule "assemble only" tasks; if a lintTask matches any variants, schedule it as well.
+        for variant in assemble_only:
+            task = as_module_task(module, "assemble" + variant)
+            subtitle = "assemble"
+            if variant in lint_task:
+                task = task + " " + as_module_task(module, lint_task)
+                subtitle = subtitle + "AndLintDebug"
+                scheduled_lint = True
+            tasks.append(make_task(task, subtitle=subtitle + variant))
```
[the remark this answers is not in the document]
grigoryk: I like the more verbose naming.

grigoryk:
Here are some run timings:

"Cold run" - decision task hit a `github-worker` without our image, and most of `android-components-g` workers weren't running yet:

selection_012

A lukewarm run, decision task had to download the image which took a while:
selection_014

Finally, a properly warm run. Tasks are almost in a hot tub! :drum: Much faster, probably our happiest path at the moment. Caching should make this quite a bit faster still:
selection_013

PR #1639, opened 2019-01-02
Adds "arrow" as a python dependency to docker image

mitchhentges:


mitchhentges:
:thinking: perhaps this kind of dependency should live in the hook's `pip install ...` instead

mitchhentges:
The hook's `pip install` affects the task that _schedules_ the decision task, but not the decision task itself.
We still need to add `arrow` to the docker image

Issue #1640, opened 2019-01-02
Share test code across applications: isEnabled(Boolean) & friends


mcomella:
Espresso expects you to match like, `isEnabled()` and `not(isEnabled())` but that means every time you want to wrap these calls and handle both enabled and disabled, you need to write two methods, with some repetition:
```kotlin
fun assertViewIsShown()
fun assertViewIsNotShown()
```

I want to write methods like `fun assertViewIsShown(isShown: Boolean)` meaning we need matchers like, `isDisplayed(isDisplayed: Boolean)`. We have this code in Fire TV and I'd like to use it on Echo Show: https://github.com/mozilla-mobile/firefox-tv/blob/6b36f79c030b186feba68562fb999415031cae68/app/src/androidTest/java/org/mozilla/tv/firefox/helpers/Matchers.kt#L17

PR #1641, opened 2019-01-03
Issue #1640: WIP: add isDisplayed(Boolean) and friends

mcomella:
This is taken nearly verbatim from functional code in firefox-tv:
  https://github.com/mozilla-mobile/firefox-tv/blob/6b36f79c030b186feba68562fb999415031cae68/app/src/androidTest/java/org/mozilla/tv/firefox/helpers/Matchers.kt#L17

I'm stuck trying to import espresso: it doesn't seem to want to work. I wonder if it only works in 1) an android application or 2) androidTest dependency. Please suggest a better approach or take this on for yourself!

Note: I was wondering if this should be its own module because anyone who includes this dep has to take a dep on espresso too but didn't want to put in the effort to create a new module if it was unnecessary and I'm concerned about the mental overload of having so many module dependencies.

### DIFF
--- a/components/support/test/src/main/java/mozilla/components/support/test/espresso/Matchers.kt
+++ b/components/support/test/src/main/java/mozilla/components/support/test/espresso/Matchers.kt
@@ -1,0 +1,38 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.support.test.espresso
+
+import android.view.View
+import org.hamcrest.CoreMatchers.not
+import org.hamcrest.Matcher
+import android.support.test.espresso.matcher.ViewMatchers.isChecked as espressoIsChecked
+import android.support.test.espresso.matcher.ViewMatchers.isDisplayed as espressoIsDisplayed
+import android.support.test.espresso.matcher.ViewMatchers.isEnabled as espressoIsEnabled
+import android.support.test.espresso.matcher.ViewMatchers.isSelected as espressoIsSelected
+
+/**
+ * The [espressoIsChecked] function that can also handle unchecked state through the boolean argument.
+ */
+fun isChecked(isChecked: Boolean): Matcher<View> = maybeInvertMatcher(espressoIsChecked(), isChecked)
+
+/**
+ * The [espressoIsDisplayed] function that can also handle not selected state through the boolean argument.
+ */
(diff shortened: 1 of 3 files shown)


components/support/test/src/main/java/mozilla/components/support/test/espresso/Matchers.kt
```diff
@@ -0,0 +1,38 @@
...
+
+package mozilla.components.support.test.espresso
+
+import android.view.View
+import org.hamcrest.CoreMatchers.not
+import org.hamcrest.Matcher
+import android.support.test.espresso.matcher.ViewMatchers.isChecked as espressoIsChecked
+import android.support.test.espresso.matcher.ViewMatchers.isDisplayed as espressoIsDisplayed
+import android.support.test.espresso.matcher.ViewMatchers.isEnabled as espressoIsEnabled
+import android.support.test.espresso.matcher.ViewMatchers.isSelected as espressoIsSelected
```
Amejia481: I believe the issue is that your importing **android.support** and should import from **androidx.test**
```kotlin
import androidx.test.espresso.matcher.ViewMatchers.isChecked as espressoIsChecked
import androidx.test.espresso.matcher.ViewMatchers.isDisplayed as espressoIsDisplayed
import androidx.test.espresso.matcher.ViewMatchers.isEnabled as espressoIsEnabled
import androidx.test.espresso.matcher.ViewMatchers.isSelected as espressoIsSelected

```

In your[ initial implementation](https://github.com/mozilla-mobile/firefox-tv/blob/6b36f79c030b186feba68562fb999415031cae68/app/src/androidTest/java/org/mozilla/tv/firefox/helpers/Matchers.kt#L17) you wasn't using **androidx** but here you are.

Amejia481: Built it locally and it worked.

Amejia481:
In my opinion, having a separate module could be a good idea, to separate references that need a physical device **(android test tests)** from **jvm tests**. But it would be nice to hear others approach too :).

pocmo:
Yeah, let's definitely test if that works. I *think* I tried using `support-test` in a UI test (`androidTest`) and that caused some issues (I don't remember the details. May be robolectric related .. which is not really needed on a device). Maybe we can just create a dummy UI test somewhere and see if that works with `support-test`? Otherwise we may need to create a second module.

pocmo:
@mcomella Is this PR still something you want to push forward? In an other PR I'm going to land `support-android-test` for helpers for on-device tests (`androidTest´).

mcomella:
> @mcomella Is this PR still something you want to push forward? In an other PR I'm going to land `support-android-test` for helpers for on-device tests (`androidTest´).

Yes but I'll do it after your other PR lands.

mcomella:
Closing in favor of a new PR in the module Sebastian recently added.

PR #1642, opened 2019-01-03
Issue #1012: service-telemetry: Replace http client with concept-fetch

pocmo:


components/service/telemetry/src/main/java/org/mozilla/telemetry/net/TelemetryClient.kt
```diff
@@ -0,0 +1,91 @@
...
+                // 405 - wrong request type (anything other than POST/PUT)
+                // 411 - missing content-length header
+                // 413 - request body too large (Note that if we have badly-behaved clients that
+                //       retry on 4XX, we should send back 202 on body/path too long).
+                // 414 - request path too long (See above)
+
+                // Something our client did is not correct. It's unlikely that the client is going
+                // to recover from this by re-trying again, so we just log and error and report a
+                // successful upload to the service.
+                logger.error("Server returned client error code: $status", null)
```
Amejia481: Nit: You could remove the second parameter is optional and is null by default.

components/service/telemetry/src/main/java/org/mozilla/telemetry/net/TelemetryClient.kt
```diff
@@ -0,0 +1,91 @@
...
+                return true
+            }
+
+            else -> {
+                // Known other errors:
+                // 500 - internal error
+
+                // For all other errors we log a warning an try again at a later time.
+                logger.warn("Server returned response code: $status", null)
```
Amejia481: Nit: You could remove the second parameter is optional and is null by default.

components/service/telemetry/src/main/java/org/mozilla/telemetry/net/TelemetryClient.kt
```diff
@@ -0,0 +1,91 @@
...
+            body = Request.Body.fromString(serializedPing))
+
+        val status = try {
+            client.fetch(request).use { response -> response.status }
+        } catch (e: IOException) {
+            logger.warn("IOException while uploading ping", e)
+            return false
+        }
+
+        logger.debug("Ping upload: $status", null)
```
Amejia481: Nit: You could remove the second parameter is optional and is null by default.

pocmo:
> Should this be included in the changelog or it will be added later?

Oh, yes, this should definitely be added :)

Amejia481:
🚢 !

PR #1643, opened 2019-01-03
feature-customtabs: Execute warmup() on main thread.

pocmo:
This fixes https://github.com/mozilla-mobile/reference-browser/issues/366


pocmo:
There's a test for warmup(). Not sure why this changes things 🤷‍♂️

PR #1650, opened 2019-01-04
Bug 1517732 - Expose routes.statuses because taskcluster-github now a…

JohanLorenzo:
…dds it

It was added in https://github.com/taskcluster/taskcluster-github/pull/323


JohanLorenzo:
@owlishDeveloper, I don't know if it's the right fix. Although, we need it temporarily to keep nightlies and releases working. Please let me know what you think.

JohanLorenzo:
Yeah, we depend on https://github.com/taskcluster/taskcluster/pull/29 to be deployed.

PR #1653, opened 2019-01-06
Closes #1606: Add Pocket Client on FireTV telemetry events ping

### LINKED ISSUE #1606: New telemetry ping for Pocket + FireTV events
Need client implementation to support the new ping added [here](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/dc458113a7a523e60a9ba50e1174a3b1e0cfdc24/schemas/pocket/fire-tv-events/fire-tv-events.1.schema.json)

psymoon:
To understand video impressions, clicks, views, and duration for Pocket on Firefox for Fire TV.

Note: `pocketId` is distinct from the telemetry `clientId`. The `clientId` should not be able to be tied to the `pocketId` in any way.

### DIFF
--- a/components/service/telemetry/src/main/java/org/mozilla/telemetry/Telemetry.java
+++ b/components/service/telemetry/src/main/java/org/mozilla/telemetry/Telemetry.java
@@ -20,6 +20,7 @@ import org.mozilla.telemetry.ping.TelemetryEventPingBuilder;
 import org.mozilla.telemetry.ping.TelemetryMobileEventPingBuilder;
 import org.mozilla.telemetry.ping.TelemetryPing;
 import org.mozilla.telemetry.ping.TelemetryPingBuilder;
+import org.mozilla.telemetry.ping.TelemetryPocketEventPingBuilder;
 import org.mozilla.telemetry.schedule.TelemetryScheduler;
 import org.mozilla.telemetry.storage.TelemetryStorage;
 
@@ -95,6 +96,7 @@ public class Telemetry {
                 // ping type and then falling back on the legacy ping type.
                 final TelemetryPingBuilder mobileEventBuilder = pingBuilders.get(TelemetryMobileEventPingBuilder.TYPE);
                 final TelemetryPingBuilder focusEventBuilder = pingBuilders.get(TelemetryEventPingBuilder.TYPE);
+                final TelemetryPingBuilder pocketEventBuilder = pingBuilders.get(TelemetryPocketEventPingBuilder.TYPE);
                 final EventsMeasurement measurement;
                 final String addedPingType;
                 if (mobileEventBuilder != null) {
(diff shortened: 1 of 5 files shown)


components/service/telemetry/src/main/java/org/mozilla/telemetry/ping/TelemetryPingBuilder.java
```diff
@@ -30,9 +30,13 @@ public TelemetryPingBuilder(@NonNull TelemetryConfiguration configuration, @NonN
...
 
-        // All pings contain a version and a client id
+        // All pings contain a version and a client id (with exception below)
         addMeasurement(new VersionMeasurement(version));
-        addMeasurement(new ClientIdMeasurement(configuration));
+
+        // Fire-tv pocket telemetry ping should not include client-id (see #1606)
+        if (!type.equals(TelemetryPocketEventPingBuilder.TYPE)) {
+            addMeasurement(new ClientIdMeasurement(configuration));
+        }
```
pocmo: Instead of requiring `TelemetryPingBuilder` to know about ping types that extend it.. how about introducing something like `shouldIncludeClientId()` which returns `true` by default and is overriden by `TelemetryPocketEventPingBuilder` to return `false`?

components/service/telemetry/src/main/java/org/mozilla/telemetry/ping/TelemetryPocketEventPingBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+import org.mozilla.telemetry.measurement.SequenceMeasurement;
+import org.mozilla.telemetry.measurement.TimezoneOffsetMeasurement;
+
+/**
+ * A telemetry ping builder for events of type "fire-tv-events".
+ *
+ * See the schema for more details:
+ *   https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/dc458113a7a523e60a9ba50e1174a3b1e0cfdc24/schemas/pocket/fire-tv-events/fire-tv-events.1.schema.json
+ */
+public class TelemetryPocketEventPingBuilder extends TelemetryPingBuilder {
```
pocmo: Can we add a unit test for this class? Take a look at the other ping builder tests. Just some sanity checks would be good. And for this ping builder in particular we may want to explicitly test that it does **not** contain the client id.

components/service/telemetry/src/main/java/org/mozilla/telemetry/Telemetry.java
```diff
@@ -103,11 +105,13 @@ public void run() {
                 } else if (focusEventBuilder != null) {
                     measurement = ((TelemetryEventPingBuilder) focusEventBuilder).getEventsMeasurement();
                     addedPingType = focusEventBuilder.getType();
+                } else if (pocketEventBuilder != null) {
+                    measurement = ((TelemetryPocketEventPingBuilder) pocketEventBuilder).getEventsMeasurement();
+                    addedPingType = pocketEventBuilder.getType();
```
pocmo: I missed this part before.

I think there's a problem here: As I understand it you do not want to replace the normal event ping with the new ping type. But instead add an additional ping type. Your regular events would still send a regular mobile event ping and some "pocket" events would use the new ping type.

My suggestion would be to remove this code here in `queueEvent()` and instead create an additional `queuePocketEvent()` path in the Fire TV repo (I assume in [TelemetryIntegration](https://github.com/mozilla-mobile/firefox-tv/blob/master/app/src/main/java/org/mozilla/tv/firefox/telemetry/TelemetryIntegration.kt)).

You'll either need to keep a reference to your pocket ping builder ([getting added here](https://github.com/mozilla-mobile/firefox-tv/blob/6b36f79c030b186feba68562fb999415031cae68/app/src/main/java/org/mozilla/tv/firefox/telemetry/TelemetryFactory.kt#L51-L52)) or maybe add some kind of public `getPingBuilder(type: Int): TelemetryPingBuilder?` helper to  `Telemetry`.

pocmo:
You know what... I'm preparing the 0.37 release right now and I'll land your changes (I'll remove the part I mentioned above and add a getter). In the best case this means you can start adding it to FFFTV today. Otherwise we will need another iteration.

pocmo:
Landed.

PR #1654, opened 2019-01-07
Closes #1397: Add custom tab toolbar color support..

### LINKED ISSUE #1397: Custom Tab: Add ability to set toolbar color
User story: https://github.com/mozilla-mobile/reference-browser/issues/209

`CustomTabsIntent.Builder.setToolbarColor(int)`

Note that based on the color we will need to adapt the text and icon colors too. There's some code for that in Fennec and Focus that we could borrow.

jonalmeida:
I ended up finishing most of the requirements for a custom tabs toolbar in https://github.com/mozilla-mobile/reference-browser/issues/209 all at once. This PR only contains the minimum required for setting up the feature and toolbar background color, the rest will come through in a follow up PR when I'm done writing tests for them.

### DIFF
--- a/components/feature/customtabs/build.gradle
+++ b/components/feature/customtabs/build.gradle
@@ -22,8 +22,11 @@ android {
 }
 
 dependencies {
+    implementation project(':browser-session')
+    implementation project(':browser-toolbar')
     implementation project(':concept-engine')
     implementation project(':support-base')
+    implementation project(':support-utils')
 
     implementation Dependencies.kotlin_stdlib
     implementation Dependencies.kotlin_coroutines
--- a/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
+++ b/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
@@ -1,0 +1,67 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ *  License, v. 2.0. If a copy of the MPL was not distributed with this
+ *  file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package mozilla.components.feature.customtabs
+
+import android.graphics.Color
+import mozilla.components.browser.session.Session
+import mozilla.components.browser.session.SessionManager
+import mozilla.components.browser.toolbar.BrowserToolbar
+import mozilla.components.support.base.feature.LifecycleAwareFeature
+
+/**
+ * Initializes and resets the Toolbar for a Custom Tab based on the CustomTabConfig.
+ */
+class CustomTabsToolbarFeature(
+    private val sessionManager: SessionManager,
+    private val toolbar: BrowserToolbar,
+    private val sessionId: String? = null
+) : LifecycleAwareFeature {
+
+    override fun start() {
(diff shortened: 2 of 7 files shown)


components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabToolbarFeature.kt
```diff
@@ -0,0 +1,67 @@
...
+import android.graphics.Color
+import mozilla.components.browser.session.Session
+import mozilla.components.browser.session.SessionManager
+import mozilla.components.browser.toolbar.BrowserToolbar
+import mozilla.components.support.base.feature.LifecycleAwareFeature
+
+/**
+ * Initializes and resets the Toolbar for a Custom Tab based on the CustomTabConfig.
+ */
+class CustomTabToolbarFeature(
```
jonalmeida: nit: should be CustomTab**s**.

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabToolbarFeature.kt
```diff
@@ -0,0 +1,67 @@
...
+    private val sessionManager: SessionManager,
+    private val toolbar: BrowserToolbar,
+    private val sessionId: String? = null
```
pocmo: In theory a custom tab should **always** have a session id. But I guess it's easier to implement that way? :)

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabToolbarFeature.kt
```diff
@@ -0,0 +1,67 @@
...
+
+        @Suppress("MagicNumber")
+        private fun greyscaleFromRGB(color: Int): Int {
+            val red = Color.red(color)
+            val green = Color.green(color)
+            val blue = Color.blue(color)
+            // Magic weighting taken from a stackoverflow post, supposedly related to how
+            // humans perceive color.
+            return (0.299 * red + 0.587 * green + 0.114 * blue).toInt()
+        }
```
pocmo: I wonder what a good pattern is.. I usually move those helpers outside the class as private methods inside the file (since they are somewhat unrelated to the class).

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabToolbarFeature.kt
pocmo: Why hold back? :D

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabToolbarFeature.kt
jonalmeida: It should! Although we tend to get the sessionId from the Bundle which makes the ID optional. So yeah, it's easier this way. :)

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabToolbarFeature.kt
jonalmeida: I can only do one rain dance a day to get a green build.

PR #1655, opened 2019-01-07
[glean] Mention the data-review process in the glean README [ci skip]

Dexterp37:
This additionally adds links to the data-review related to the fields in the `ping_info` section of the pings.

### DIFF
--- a/components/service/glean/README.md
+++ b/components/service/glean/README.md
@@ -3,6 +3,22 @@
 A client-side telemetry SDK for collecting metrics and sending them to Mozilla's telemetry service
 (eventually replacing [service-telemetry](../telemetry/README.md)).
 
+- [Before using the library](#before-using-the-library)
+- [Usage](#usage)
+    - [Setting up the dependency](#setting-up-the-dependency)
+    - [Integrating with the build system](#integrating-with-the-build-system)
+    - [Initializing glean](#initializing-glean)
+    - [Defining metrics](#defining-metrics)
+- [License](#license)
+
+## Before using the library
+Products using glean to collect telemetry **must**:
+
+- add documentation for any new metric collected with the library in its repository (see [an example](https://github.com/mozilla-mobile/android-components/blob/df429df1a193516f796f2330863af384cce820bc/components/service/glean/docs/pings.md));
+- go through data review for the newly collected data by following [this process](https://wiki.mozilla.org/Firefox/Data_Collection);
+- provide a way for users to turn data collection off (e.g. providing settings to control
+  `Glean.setMetricsEnabled()`).
+
 ## Usage
 
 ### Setting up the dependency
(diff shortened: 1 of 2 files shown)


components/service/glean/README.md
```diff
@@ -3,6 +3,21 @@
...
+    - [Setting up the dependency](#setting-up-the-dependency)
+    - [Integrating with the build system](#integrating-with-the-build-system)
+    - [Initializing glean](#initializing-glean)
+    - [Defining metrics](#defining-metrics)
+- [License](#license)
+
+## Before using the library
+Products using glean to collect telemetry **must**:
+
+- add documentation for any new metric collected with the library in its repository (see [an example](https://github.com/mozilla-mobile/android-components/commit/df429df1a193516f796f2330863af384cce820bc));
```
mdboom: This commit includes a number of things that aren't really relevant to what's being talked about here.  I assume you just mean the documentation chunk being added there?  Maybe we could link to that instead of the commit?

components/service/glean/README.md
```diff
@@ -3,6 +3,21 @@
...
+- go through data review for the newly collected data by following [this process](https://wiki.mozilla.org/Firefox/Data_Collection);
+- provide a way for users to turn data collection off.
```
mdboom: Should we mention that this code should call `Glean.setMetricsEnabled(...)`?

components/service/glean/metrics.yaml
```diff
@@ -16,6 +16,8 @@ baseline:
       - baseline
     bugs:
       - 1497894
+    reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1512938#c3
```
mdboom: Should we make `reviews` required in the `metrics.yaml` schema (in glean-parser)?  That would have enforced these changes all along.  Or is that too heavy and there are legitimate use cases to leave these out?

components/service/glean/README.md
travis79: I agree with @mdboom that it would be better to link to just the documentation example here.

components/service/glean/metrics.yaml
Dexterp37: This is a good point. We should check back with @georgf on this. I feel like we really should, though, as it would prevent bad things from happening (i.e. unreviewed collections slipping in ) :)

pocmo: I like the forced `review` property too. As a developer I could at least link to an explanation/discussion why no review was needed.

Dexterp37: Perfect :) I filed [a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1518182) to discuss this! I'll flag Georg on it so that it doesn't slip through the cracks

PR #1658, opened 2019-01-07
[glean]Prevent ping upload when metrics disabled or ping empty

travis79:
Pings were previously still being uploaded even when metrics were disabled. Added a check to test for metrics being enabled before building and uploading ping.

Pings could also be uploaded if they contained no data.  In order to address this, the collect function was changed to return an empty string to be able to intelligently determine whether or not to send a ping.

Add log warning message when attempting to send an empty ping.

Add test cases ensuring pings aren't sent for metrics disabled, and empty pings

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
@@ -146,15 +146,25 @@ open class GleanInternalAPI {
 
     /**
      * Collect and assemble the ping. Asynchronously submits the assembled
-     * payload to the designated server using [httpPingUploader].
+     * payload to the designated server using [httpPingUploader] but only
+     * if metrics are enabled, and there is ping data to send.
      */
     internal fun sendPing(store: String, docType: String) {
-        val pingContent = pingMaker.collect(store)
-        val uuid = UUID.randomUUID()
-        val path = makePath(docType, uuid)
-        // Asynchronously perform the HTTP upload off the main thread.
-        GlobalScope.launch(Dispatchers.IO) {
-            httpPingUploader.upload(path = path, data = pingContent)
+        // Do not send ping if metrics disabled
+        if (!getMetricsEnabled()) {
+            logger.debug("Attempt to send ping \"$store\" but metrics disabled, aborting send.")
+            return
+        }
+
+        // Build and send the ping on an async thread. Since the pingMaker.collect() function
+        // returns null if there is nothing to send we can use this to avoid sending an empty ping
+        pingMaker.collect(store)?.let { pingContent ->
+            val uuid = UUID.randomUUID()
+            val path = makePath(docType, uuid)
+            // Asynchronously perform the HTTP upload off the main thread.
+            GlobalScope.launch(Dispatchers.IO) {
+                httpPingUploader.upload(path = path, data = pingContent)
+            }
         }
     }
 
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/ping/PingMaker.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/ping/PingMaker.kt
@@ -108,12 +108,20 @@ internal class PingMaker(
      * Collects the relevant data and assembles the requested ping.
      *
      * @param storage the name of the storage containing the data for the ping.
-     *        This usually matches with the name of the ping
-     * @return a string holding the data for the ping.
+     *        This usually matches with the name of the ping.
+     * @return a string holding the data for the ping, or null if there is no data to send.
      */
-    fun collect(storage: String): String {
+    fun collect(storage: String): String? {
         val jsonPing = storageManager.collect(storage)
+
+        // Return null if there is nothing in the jsonPing object so that this can be used by
+        // consuming functions (i.e. sendPing()) to indicate no ping data is available to send.
+        if (jsonPing.length() == 0) {
+            return null
+        }
+
         jsonPing.put("ping_info", getPingInfo(storage))
+
         return jsonPing.toString()
     }
 }
--- a/components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
+++ b/components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
@@ -31,6 +31,7 @@ import org.mockito.Mockito.spy
 import org.robolectric.RobolectricTestRunner
 import java.io.File
 import java.util.UUID
+import java.util.concurrent.TimeUnit
 
 @ObsoleteCoroutinesApi
 @ExperimentalCoroutinesApi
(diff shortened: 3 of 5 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/ping/PingMaker.kt
```diff
@@ -113,7 +113,16 @@ internal class PingMaker(
      */
     fun collect(storage: String): String {
         val jsonPing = storageManager.collect(storage)
+
+        // TODO, maybe need to not add the 'ping_info' to the jsonPing if the jsonPing is already empty
+        // This could then be used by consuming functions (i.e. sendPing()) to selectively send the
+        // ping or not, depending on whether it was empty or not.
+        if (jsonPing.length() == 0) {
```
Dexterp37: Instead of relying on special values, we could make this function return a nullable String (i.e. `String?`). This way we could simply null-check at the call site and return null here.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -146,15 +146,26 @@ open class GleanInternalAPI {
...
     /**
      * Collect and assemble the ping. Asynchronously submits the assembled
-     * payload to the designated server using [httpPingUploader].
+     * payload to the designated server using [httpPingUploader] but only
+     * if metrics are enabled.
      */
     internal fun sendPing(store: String, docType: String) {
+        // Do not send ping if metrics disabled
+        if (!getMetricsEnabled()) {
+            return
```
Dexterp37: nit: let's print a log line here, trace or debug level, mentioning what we're trying to send and that we're bailing out because upload is disabled.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -146,15 +146,26 @@ open class GleanInternalAPI {
...
+        // Since the pingMaker.collect() function sends an empty string if there is nothing recorded
+        // we can skip sending an empty ping quietly, otherwise the ping will be built and sent
         val pingContent = pingMaker.collect(store)
-        val uuid = UUID.randomUUID()
-        val path = makePath(docType, uuid)
-        // Asynchronously perform the HTTP upload off the main thread.
-        GlobalScope.launch(Dispatchers.IO) {
-            httpPingUploader.upload(path = path, data = pingContent)
+
+        if (pingContent.isNotEmpty()) {
```
Dexterp37: If we make the return value of `collect` nullable, we can do something like:

```
pingMaker.collect(store).?let { pingContent ->
 ...
}
```

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
```diff
@@ -288,4 +289,32 @@ class GleanTest {
         gleanSpy.initialized = false
         gleanSpy.handleEvent(Glean.PingEvent.Default)
     }
+
+    @Test
+    fun `Don't send metrics if metrics disabled`() {
```
Dexterp37: nit: `Don't send pings if recording is disabled`?

samples/glean/src/main/java/org/mozilla/samples/glean/GleanApplication.kt
```diff
@@ -20,6 +20,7 @@ class GleanApplication : Application() {
         // Initialize the Glean library. Ideally, this is the first thing that
         // must be done right after enabling logging.
         Glean.initialize(applicationContext)
+        Glean.setMetricsEnabled(false)
```
Dexterp37: Why do we want/need this?

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
```diff
@@ -288,4 +289,32 @@ class GleanTest {
...
+        )
+        Glean.httpPingUploader = HttpPingUploader(testConfig)
+        Glean.setMetricsEnabled(false)
+
+        try {
+            Glean.handleEvent(Glean.PingEvent.Background)
+
+            val requests: MutableMap<String, String> = mutableMapOf()
+
+            val request = server.takeRequest(2, TimeUnit.SECONDS)
```
Dexterp37: Please leave a comment mentioning that this test will only work since we're faking the dispatchers with the `@Rule` above, otherwise it would fail due to async weirdness!

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
```diff
@@ -288,4 +289,32 @@ class GleanTest {
         gleanSpy.initialized = false
         gleanSpy.handleEvent(Glean.PingEvent.Default)
     }
+
+    @Test
```
Dexterp37: We should also add a test for making sure pings are not sent if there's no content.

components/service/glean/src/main/java/mozilla/components/service/glean/ping/PingMaker.kt
travis79: Sure, nullable is good and can make the calling function more Kotlin-like.

samples/glean/src/main/java/org/mozilla/samples/glean/GleanApplication.kt
travis79: We don't it shouldn't be there, it was for testing something and I neglected to filter it out when I committed.

components/service/glean/src/main/java/mozilla/components/service/glean/ping/PingMaker.kt
mdboom: We should probably add a comment to `StorageEngineManager.collect` that it should return a zero-length `JSONObject` whenever there is no actual content.  And I'm not sure it actually currently works that way.  (See below).

components/service/glean/src/test/java/mozilla/components/service/glean/ping/PingMakerTest.kt
```diff
@@ -93,4 +93,13 @@ class PingMakerTest {
         assertEquals(engine2Data, metricsData.getJSONArray("engine2"))
         assertFalse(metricsData.has("wontCollect"))
     }
+
+    @Test
+    fun `collect() must report an empty string when no data is stored`() {
+        val maker = PingMaker(StorageEngineManager(emptyMap(), applicationContext = mockApplicationContext))
```
mdboom: I think passing in `emptyMap()` here is kind of cheating.  In the normal case, there will be metric types here, but their storage engines will be empty.  I think in that case, we'll still be getting content from the `StorageEngineManager` that looks like:

```
{ "metrics": { "string": {} } }
```

(This is just from looking at the code -- I'm going to confirm my hunch momentarily, but wanted you to get the review asap)...

components/service/glean/src/test/java/mozilla/components/service/glean/ping/PingMakerTest.kt
```diff
@@ -93,4 +93,13 @@ class PingMakerTest {
...
+
+        // Gather the data. We expect an empty ping with the "ping_info" information
```
mdboom:
```suggestion
        // Gather the data. We expect an empty ping without the "ping_info" information
```

components/service/glean/src/test/java/mozilla/components/service/glean/ping/PingMakerTest.kt
mdboom: So, I'm happily mistaken, but it's subtle.  If you add a key/value pair to a JSONObject, where the value is `null`, nothing ends up being added.

So, I think I'd be happy with (a) removing the `emptyMap()` argument on this line, and (b) adding a note in StorageEngineManager.collect that we need to be careful to return a completely empty (zero length) JSONObject in order to ensure that pings without any real metrics aren't sent.

travis79: This is a little trickier than it appears since the GenericScalarStorageEngine.userLifetimeStorage is lazily instantiated.  Without assigning a map here, empty or otherwise, it's provoking a null reference exception when it tries to getSnapshot() since the userLifetimeStorage isn't instantiated yet.  I'm currently looking for a good way around this.

mdboom: Interesting, I'm not seeing that locally, but it could be a test ordering (i.e. "I'm getting lucky") issue.

travis79: I'm just running this test all by it's lonesome.  Let me try letting them run all in order.

travis79: I have refactored and pushed a slight change to have data in the map and just request a store that doesn't exist (i.e. it was disabled or nothing was recorded for it).  I hope this covers a little better what your concern is.

mdboom: Yes, indeed that does address my concerns.  As long as we have *some* storage engines in the manager, we're confirming that we don't add the results of empty storage engines.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -146,15 +146,25 @@ open class GleanInternalAPI {
...
     internal fun sendPing(store: String, docType: String) {
-        val pingContent = pingMaker.collect(store)
-        val uuid = UUID.randomUUID()
-        val path = makePath(docType, uuid)
-        // Asynchronously perform the HTTP upload off the main thread.
-        GlobalScope.launch(Dispatchers.IO) {
-            httpPingUploader.upload(path = path, data = pingContent)
+        // Do not send ping if metrics disabled
+        if (!getMetricsEnabled()) {
+            logger.debug("Attempt to send ping [$store] but metrics disabled, aborting send.")
```
Dexterp37: nit: let's drop the square brackets around `$store` and use double quotes `"` instead, for consistency with other messages.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -146,15 +146,25 @@ open class GleanInternalAPI {
...
+            return
+        }
+
+        // Since the pingMaker.collect() function sends an empty string if there is nothing recorded
```
Dexterp37: nit: this comment needs to be updated since we're not returning an empty string anymore from `collect`

components/service/glean/src/main/java/mozilla/components/service/glean/ping/PingMaker.kt
```diff
@@ -111,9 +111,18 @@ internal class PingMaker(
      *        This usually matches with the name of the ping
      * @return a string holding the data for the ping.
      */
-    fun collect(storage: String): String {
+    fun collect(storage: String): String? {
         val jsonPing = storageManager.collect(storage)
+
+        // TODO, maybe need to not add the 'ping_info' to the jsonPing if the jsonPing is already empty
```
Dexterp37: The TODO says that we should not add `ping_info` to the JSON payload if the latter is empty. Am I interpreting this correctly? Because it seems to be what we're doing, since we're bailing out early when the length is null.

travis79: Just me neglecting to clean up a comment.  I placed this here when I started on this bug to help identify all of the places I would need to make changes and I just didn't update this one to reflect the implementation.

Dexterp37:
This should fix the last outstanding items for mozilla-mobile/reference-browser#199.

PR #1666, opened 2019-01-08
Issue #1489: Fix backspace issue when suggestion strip is off

mcomella:
This PR should replace https://github.com/mozilla-mobile/android-components/pull/1573 (or the top-most commit should be taken into that PR).

@Amejia481 I verified this works correctly on Echo Show gen 1 and 2. Please verify it works correctly on Focus.


pocmo:
Ah, there's a lint failure too. I'll fix that before landing.

In a follow-up we could try to extend `InlineAutocompleteEditTextTest` to cover those special cases. Seems like otherwise we may accidentally remove them to fix another issue. :)

Issue #42, opened 2018-03-29
How can we incorporate telemetry without making it a requirement for consumers?


pocmo:
If you look at our current set of monolith apps then you will see that telemetry UI probes are scattered all over the code base. A reasonable chunk of the code emitting those telemetry events will be inside our components. How can we make sure that consuming apps can send telemetry events they are interested in without forcing something on all consumers using the app.

Some thoughts:
* Could we unify the telemetry events across all apps? (e.g. one standard event for "Open toolbar menu" that is the same for all apps)
* How can we make sure the app is in control what is send and how?

csadilek:
I've had good results with AspectJ in the past. Hugo I didn't know yet (looks good). The one difficulty with any AOP solution is how to handle conditional logic within a method. We can intercept methods and execute logic before/after/around (if successful or if failed), but how would we inject conditional logic e.g. the telemetry events fired here: https://github.com/mozilla-mobile/focus-android/blob/master/app/src/main/java/org/mozilla/focus/activity/MainActivity.java#L159

Of course, we could refactor all these methods to make them easier to intercept.

PR #1668, opened 2019-01-08
Introduce simple, generic API for collecting "facts" emmitted by components.

pocmo:
@csadilek I'd like to get your feedback here.

This is about #42: Having a way for components to emit telemetry events.

I think we want:
* The app to be in control about what gets send/collected
* Not depend on the telemetry library directly from all components (which is also impossible now that we have two of them)
* Have events that are independent from telemetry (There are some use cases that make it interesting to evaluate the events for other reasons, e.g.: A "contextual hints" FEATURE will need to answer questions like "did the user use the share feature in the past? did the user load a url at least 3 times?" and the answers can be derived from the events we observed)

I followed a similar approach like we did with "logs". There's a class that collects them and processors that can do something with them.

Since the events are generic we would need a way to translate between the generic event and a telemetry event. I want to look into this as soon as we have glean and some events in the reference browser. I expect this to not be super hard since the events follow a similar structure.

The word `event` is already quite overloaded. And I want to avoid that someone thinks those are events you listen to in order to implement app logic. So I tried to find something else. I came up with `Fact`, `FactCollector`, `FactProcessor`. Components emit facts. Something collects them. And something else processes them.

To simplify testing in this PR I made the FactCollector something that we pass down to our components that need it. I wanted to avoid having a singleton that every component accesses. However the current approach seems to be super verbose. I have to pass this thing around everywhere and it's pretty painful and makes our interfaces more verbose. I now tend more to using a Singleton again. It will make testing a bit more effortful - but it's still doable.

What do you think?

### DIFF
--- a/components/support/base/src/main/java/mozilla/components/support/base/facts/Action.kt
+++ b/components/support/base/src/main/java/mozilla/components/support/base/facts/Action.kt
@@ -1,0 +1,35 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.support.base.facts
+
+/**
+ * A user or system action that causes [Fact] instances to be emitted.
+ */
+enum class Action {
+    /**
+     * The user has clicked on something.
+     */
+    CLICK,
+
+    /**
+     * The user has toggled something.
+     *
+     * Other than a click this action is performed on items that have a distinct number of states. For example clicking
+     * a checkbox or switch widget could emit a [Fact] with a [TOGGLE] action instead of a [CLICK] action.
+     */
+    TOGGLE,
+
+    /**
+     * The user has committed an input (e.g. entered text into an input field and then pressed enter).
+     */
+    COMMIT,
+
+    /**
+     * A generic interaction that can be caused by a previous action (e.g. the user clicks on a button which causes a
+     * [Fact] with ][CLICK] action to be emitted. This click may causes something to load which emits a follow-up a
+     * [Fact] with [INTERACTION] action,
+     */
+    INTERACTION
+}
--- a/components/support/base/src/main/java/mozilla/components/support/base/facts/Fact.kt
+++ b/components/support/base/src/main/java/mozilla/components/support/base/facts/Fact.kt
@@ -1,0 +1,29 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.support.base.facts
+
+import mozilla.components.support.base.Component
+
+/**
+ * A fact describing a generic event that has occurred in a component.
+ *
+ * @property component Component that emitted this fact.
+ * @property action A user or system action that caused this fact (e.g. Action.CLICK).
+ * @property item An item that caused the action or that the action was performed on (e.g. "toolbar").
+ * @property value An optional value providing more context.
+ * @property metadata A key/value map for facts where additional richer context is needed.
+ */
+data class Fact(
+    val component: Component,
+    val action: Action,
+    val item: String,
+    val value: String? = null,
+    val metadata: Map<String, Any>? = null
+)
+
+/**
+ * Collect this fact through the [Facts] singleton.
+ */
+fun Fact.collect() = Facts.collect(this)
--- a/components/support/base/src/main/java/mozilla/components/support/base/facts/FactProcessor.kt
+++ b/components/support/base/src/main/java/mozilla/components/support/base/facts/FactProcessor.kt
@@ -1,0 +1,20 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.support.base.facts
+
+/**
+ * A [FactProcessor] receives [Fact] instances from the [Facts] singleton to process them further.
+ */
+interface FactProcessor {
+    /**
+     * Passes the given [Fact] to the [FactProcessor] for processing.
+     */
+    fun process(fact: Fact)
+}
+
+/**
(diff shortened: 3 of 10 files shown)


Amejia481:
I like `FactCollector`, but if the normal use case for getting these facts (telemetry events) is for sending them through a telemetry tool? If that is the case we're delegating this work to the client, and will get duplicated across apps, could we have an abstraction similar to what we have in [Crash](https://github.com/mozilla-mobile/android-components/blob/master/components/lib/crash/README.md) to sending our self or it doesn't worth it?

jonalmeida:
> Since the events are generic we would need a way to translate between the generic event and a telemetry event. I want to look into this as soon as we have glean and some events in the reference browser. I expect this to not be super hard since the events follow a similar structure.
> 
> The word event is already quite overloaded. And I want to avoid that someone thinks those are events you listen to in order to implement app logic. So I tried to find something else. I came up with Fact, FactCollector, FactProcessor. Components emit facts. Something collects them. And something else processes them.

Sounds similar to event tracking pipelines. :)



We don't do step 6, and our step 5 is sending it to (multiple?) providers. 

'Storage' would be so that you could bulk save and send the events, until then they stay on persistent storage?

pocmo:
> I like FactsCollector, but if the normal use case for getting these facts (telemetry events) is sending them through a telemetry tool? If that is the case we're delegating this work to the client, and this code will get duplicated across apps.

Yeah. I think the next steps would be to provide some (abstract?) implementations of common processors.

> We don't do step 6, and our step 5 is sending it to (multiple?) providers.

Storage would be up to whatever implements a processor. Telemetry has its own storage and let's say some "ContextualHintsAnalytics" thingy will also have its own storage optimized for the queries it needs to do.

> @pocmo I like the naming and concept. I am also in favour of passing the FactCollector around to avoid introducing a singleton for all the reasons we agree on (hard to test etc.).

I was in favor of it until I had to add it to add it to all the constructors or adding setters. And I only added a handful of probes. I almost prefer the singleton now. :)

> Maybe this means that in the future we'll have a component context that we pass around and which contains the FactCollector among other things that all components need? (to make it less verbose)

Oooh. That sounds interesting. I would feel way less concerned about passing such a context around :)

In a way this context reminds me of our scaffolding / code generation idea. Imagine this context could give you access to other components/usecases.

pocmo:
Thank you all for the feedback so far.

I added a follow-up commit that shows the singleton variant I had in mind. I like that it is less obtrusive and doesn't require the whole world to change. It's more or less the same pattern that we already use for our logging. It's a singleton, but still manageable.

pocmo:
@csadilek Okay, this is now ready to get in! :)

components/support/base/src/main/java/mozilla/components/support/base/facts/Action.kt
```diff
@@ -0,0 +1,35 @@
...
+    TOGGLE,
+
+    /**
+     * The user has committed an input (e.g. entered text into an input field and then pressed enter).
+     */
+    COMMIT,
+
+    /**
+     * A generic interaction that can be caused by a previous action (e.g. the user clicks on a button which causes a
+     * [Fact] with ][CLICK] action to be emitted. This click may causes something to load which emits a follow-up a
```
csadilek: nit: `] [CLICK`

components/support/base/src/main/java/mozilla/components/support/base/facts/Action.kt
```diff
@@ -0,0 +1,35 @@
...
+     * [Fact] with [INTERACTION] action,
```
csadilek: nit `.`->`,`

components/support/base/src/main/java/mozilla/components/support/base/facts/Fact.kt
```diff
@@ -0,0 +1,29 @@
...
+ * @property component Component that emitted this fact.
+ * @property action A user or system action that caused this fact (e.g. Action.CLICK).
+ * @property item An item that caused the action or that the action was performed on (e.g. "toolbar").
+ * @property value An optional value providing more context.
+ * @property metadata A key/value map for facts where additional richer context is needed.
+ */
+data class Fact(
+    val component: Component,
+    val action: Action,
+    val item: String,
```
csadilek: nit: would `source` be a better name for `item`?

components/support/base/src/main/java/mozilla/components/support/base/facts/FactProcessor.kt
```diff
@@ -0,0 +1,20 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.support.base.facts
+
+/**
+ * A [FactProcessor] receives [Fact] instances from the [Facts] singleton to process them further.
```
csadilek: nit: I would remove the `from the [Facts] singleton` part :)

components/support/base/src/test/java/mozilla/components/support/base/facts/FactsTest.kt
```diff
@@ -0,0 +1,59 @@
...
+class FactsTest {
+    @Before
+    @After
+    fun cleanUp() {
+        Facts.clearProcessors()
+    }
+
+    @Test
+    fun `Collecting fact without processor`() {
+        Facts.collect(Fact(
```
csadilek: 🤔 hm there's no assertion here. I guess we want to verify that this is a NOOP an doesn't throw an exception?

csadilek:
Let's land this. All nits are super minor and can be addressed later.

components/support/base/src/main/java/mozilla/components/support/base/facts/Fact.kt
pocmo: Yeah, maybe, or `origin`? [telemetry events](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/collection/events.html) are using `object` here, but that's super awkward to use in Kotlin since that is a reserved keyword.

components/support/base/src/test/java/mozilla/components/support/base/facts/FactsTest.kt
pocmo: Not sure how do check that something is a NOOP here though? Running it at least checks that there's no exception - that would fail the test. :)

PR #1673, opened 2019-01-08
Issue 1195: Remove webview if EngineView instance is reused....

csadilek:
FireTV is using a cached single instance of SystemEngineView. There are cases though where it uses different sessions i.e. settings->clear all data will create a new session with a new webview.

This makes sure we remove the old webview from the engine view so it can get gc'ed in this case. Our code (samples, and r-b) should not be affected.

### DIFF
--- a/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
+++ b/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
@@ -69,9 +69,10 @@ class SystemEngineView @JvmOverloads constructor(
      * Render the content of the given session.
      */
     override fun render(session: EngineSession) {
-        this.session = session as SystemEngineSession
+        removeAllViews()
 
-        (session.webView?.parent as? SystemEngineView)?.removeView(session.webView)
+        this.session = session as SystemEngineSession
+        (session.webView.parent as? SystemEngineView)?.removeView(session.webView)
         addView(initWebView(session.webView))
     }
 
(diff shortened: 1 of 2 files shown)


components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -69,8 +69,11 @@ class SystemEngineView @JvmOverloads constructor(
      * Render the content of the given session.
      */
     override fun render(session: EngineSession) {
-        this.session = session as SystemEngineSession
+        this.session?.let {
+            removeView(it.webView)
+        }
```
pocmo: Does the method need to be synchronized?

We could also use a hammer here and remove all child views.

csadilek: Yeah, I had this with `removeAllViews` but thought it might be too much. I will change this back.

csadilek: OK, updated.

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -69,8 +69,9 @@ class SystemEngineView @JvmOverloads constructor(
      * Render the content of the given session.
      */
     override fun render(session: EngineSession) {
-        this.session = session as SystemEngineSession
+        removeAllViews()
 
+        this.session = session as SystemEngineSession
         (session.webView?.parent as? SystemEngineView)?.removeView(session.webView)
         addView(initWebView(session.webView))
```
pocmo: And `@Synchronized` to protect against sessions switching fast?

pocmo: Mh. Yeah, I was just looking at this method in isolation - not thinking about how the other components may use it.

PR #1677, opened 2019-01-08
Closes #1400: Add close button to custom tabs...

### LINKED ISSUE #1400: Custom Tab: Add ability to set close button icon
User Story: https://github.com/mozilla-mobile/reference-browser/issues/209

`CustomTabsIntentBuilder.setCloseButtonIcon()`.

This will require changes in the toolbar too.

jonalmeida:
Also makes the toolbar unclickable so it can't be switching into
edit mode, and removes all other actions.

Moved the custom tab back button bitmap checking to the feature from the
CustomTabConfig since we needed to know the DisplayMetrics for the
current screen to properly compare it to the max dp value.

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
@@ -532,3 +532,20 @@ class SessionManager(
         )
     }
 }
+
+/**
+ * Tries to find a session with the provided session ID and runs the block if found.
+ *
+ * @return True if the session was found and run successfully.
+ */
+fun SessionManager.runWithSession(
+    sessionId: String?,
+    block: SessionManager.(Session) -> Boolean
+): Boolean {
+    sessionId?.let {
+        findSessionById(sessionId)?.let { session ->
+            return block(session)
+        }
+    }
+    return false
+}
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/tab/CustomTabConfig.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/tab/CustomTabConfig.kt
@@ -22,6 +22,7 @@ import android.support.customtabs.CustomTabsIntent.KEY_MENU_ITEM_TITLE
 import android.support.customtabs.CustomTabsIntent.EXTRA_TINT_ACTION_BUTTON
 import android.support.customtabs.CustomTabsIntent.EXTRA_REMOTEVIEWS
 import android.support.customtabs.CustomTabsIntent.EXTRA_TOOLBAR_ITEMS
+import android.util.DisplayMetrics
 
 import mozilla.components.support.utils.SafeBundle
 import mozilla.components.support.utils.SafeIntent
@@ -54,6 +55,7 @@ class CustomTabConfig internal constructor(
         internal const val BOTTOM_TOOLBAR_OPTION = "hasBottomToolbar"
         internal const val EXIT_ANIMATION_OPTION = "hasExitAnimation"
         internal const val PAGE_TITLE_OPTION = "hasPageTitle"
+        private const val MAX_CLOSE_BUTTON_SIZE_DP = 24
 
         /**
          * Checks if the provided intent is a custom tab intent.
@@ -70,10 +72,11 @@ class CustomTabConfig internal constructor(
          *
          * @param intent the intent, wrapped as a SafeIntent, which is processed
          * to extract configuration data.
+         * @param displayMetrics needed in-order to verify that icons of a max size are only provided.
          * @return the CustomTabConfig instance.
          */
         @Suppress("ComplexMethod")
-        fun createFromIntent(intent: SafeIntent): CustomTabConfig {
+        fun createFromIntent(intent: SafeIntent, displayMetrics: DisplayMetrics? = null): CustomTabConfig {
             val id = UUID.randomUUID().toString()
 
             val options = mutableListOf<String>()
@@ -87,7 +90,11 @@ class CustomTabConfig internal constructor(
 
             val closeButtonIcon = run {
                 val icon = intent.getParcelableExtra(EXTRA_CLOSE_BUTTON_ICON) as? Bitmap
-                if (icon != null) {
+                val density = displayMetrics?.density ?: 1f
+                if (icon != null &&
+                    icon.width / density <= MAX_CLOSE_BUTTON_SIZE_DP &&
+                    icon.height / density <= MAX_CLOSE_BUTTON_SIZE_DP
+                ) {
                     options.add(CLOSE_BUTTON_OPTION)
                     icon
                 } else {
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/tab/CustomTabConfig.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/tab/CustomTabConfig.kt
@@ -54,7 +54,6 @@ class CustomTabConfig internal constructor(
         internal const val BOTTOM_TOOLBAR_OPTION = "hasBottomToolbar"
         internal const val EXIT_ANIMATION_OPTION = "hasExitAnimation"
         internal const val PAGE_TITLE_OPTION = "hasPageTitle"
-        private const val MAX_CLOSE_BUTTON_SIZE_DP = 24
 
         /**
          * Checks if the provided intent is a custom tab intent.
@@ -88,7 +87,7 @@ class CustomTabConfig internal constructor(
 
             val closeButtonIcon = run {
                 val icon = intent.getParcelableExtra(EXTRA_CLOSE_BUTTON_ICON) as? Bitmap
-                if (icon != null && icon.width <= MAX_CLOSE_BUTTON_SIZE_DP && icon.height <= MAX_CLOSE_BUTTON_SIZE_DP) {
+                if (icon != null) {
                     options.add(CLOSE_BUTTON_OPTION)
                     icon
                 } else {
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
@@ -24,7 +24,6 @@ import mozilla.components.support.base.android.Padding
 import mozilla.components.support.ktx.android.content.res.pxToDp
 import mozilla.components.support.ktx.android.view.forEach
 import mozilla.components.support.ktx.android.view.isVisible
-import mozilla.components.support.ktx.android.view.setPadding
 import mozilla.components.ui.autocomplete.InlineAutocompleteEditText
 
 /**
--- a/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
+++ b/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
@@ -75,6 +75,19 @@ class CustomTabsToolbarFeature(
 
     override fun stop() {}
 
+    /**
+     * Removes the current Custom Tabs session when the back button is pressed and returns true.
+     * Should be called when the back button is pressed.
+     */
+    fun onBackPressed(): Boolean {
+        val result = sessionManager.runWithSession(sessionId) {
+            remove(it)
+            true
+        }
+        closeListener.invoke()
+        return result
+    }
+
     internal fun Drawable.isSmallerThan(maxSize: Int): Boolean = with(context.resources) {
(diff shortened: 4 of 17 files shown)


components/browser/session/src/main/java/mozilla/components/browser/session/tab/CustomTabConfig.kt
```diff
@@ -88,7 +87,7 @@ class CustomTabConfig internal constructor(
 
             val closeButtonIcon = run {
                 val icon = intent.getParcelableExtra(EXTRA_CLOSE_BUTTON_ICON) as? Bitmap
-                if (icon != null && icon.width <= MAX_CLOSE_BUTTON_SIZE_DP && icon.height <= MAX_CLOSE_BUTTON_SIZE_DP) {
+                if (icon != null) {
```
pocmo: What was the reason for removing the check?

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
```diff
@@ -40,9 +56,30 @@ class CustomTabsToolbarFeature(
...
+                if (it.isSmallerThan(MAX_CLOSE_BUTTON_SIZE_DP)) return@also
+            }
+        } ?: ContextCompat.getDrawable(context, R.drawable.mozac_ic_close)
+
+        drawableIcon?.let {
+            val button = Toolbar.ActionButton(
+                it, context.getString(R.string.mozac_feature_customtabs_exit_button)
+            ) { closeListener.invoke() }
+            toolbar.addNavigationAction(button)
+        }
```
pocmo: `drawableIcon?` confuses me here. Can we really end up in a situation where we do not have the icon and are not adding a button?

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
```diff
@@ -330,6 +329,13 @@ class BrowserToolbar @JvmOverloads constructor(
         displayToolbar.addNavigationAction(action)
     }
 
+    /**
+     * Removes all added actions on the left, right sides and the menu in display mode.
+     */
+    override fun removeAllActions() {
+        displayToolbar.removeAllActions()
+    }
```
pocmo: This is going to be a problem!

You can't know what the app added and whether the app wants those actions be removed in a custom tab (That decision is probably up to the app). Imagine a tablet layout where you have back/forward buttons in the toolbar. Maybe we want to keep those. Chrome for example shows a bookmark icon in custom tabs on a tablet. Your code may remove that if it's not happening in or after `CustomTabsToolbarFeature`.

Regarding the "tabs tray" button. I think we should make `TabsToolbarFeature` smarter to not add a button if it's not needed - instead of adding it and then removing all again.

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
```diff
@@ -28,8 +36,16 @@ class CustomTabsToolbarFeature(
 
     internal fun initialize(session: Session) {
         session.customTabConfig?.let { config ->
+            // Hide tab counter
+            toolbar.removeAllActions()
+            // Put the toolbar only in display mode
+            toolbar.displayMode()
```
pocmo: Isn't that the default? Did you run into issues if not doing that explicitly?

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
```diff
@@ -40,9 +56,30 @@ class CustomTabsToolbarFeature(
...
+            ) { closeListener.invoke() }
+            toolbar.addNavigationAction(button)
+        }
+    }
+
     override fun stop() {}
 
+    internal fun Drawable.isSmallerThan(maxSize: Int): Boolean = with(context.resources) {
+        pxToDp(minimumWidth) <= maxSize && pxToDp(minimumHeight) <= maxSize
+    }
```
pocmo: Let's move those into `support-ktx`.

samples/browser/src/main/java/org/mozilla/samples/browser/BrowserFragment.kt
```diff
@@ -114,7 +114,7 @@ class BrowserFragment : Fragment(), BackHandler {
             components.sessionManager,
             toolbar,
             sessionId
-        )
+        ) { activity?.finish() }
```
pocmo: Let's also remove the session! Otherwise it will stay somewhere in memory in the session manager :)

We'll also make sure the same happens when pressing "back" i think (although that may be out of the scope for this PR; but important).

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
jonalmeida: No I didn't, but since there wasn `editMode()` as well, I thought it would be better to be explicit?

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
jonalmeida: I considered this as well, but I'd need to pass the resources into it as well and I thought that didn't make a great extension function to be public so I had it internal i.e. `Drawable.isSmallerThan(Int)` seems better than `Drawable.isSmallerThan(Resources, Int)`

If we don't mind that, then sure!

components/browser/session/src/main/java/mozilla/components/browser/session/tab/CustomTabConfig.kt
jonalmeida: This check wasn't exactly working here. The bitmap width was the measured value and we were comparing it to the density pixel value. The close icons provided were actually 24dp but the measured width was 86 so we were always getting the icon as null.

Instead of making the CustomTabConfig depend on Resources/DisplayMetrics to be provided to it, I decided to put that check into the feature itself since we have the context for it to make the internal function `Drawable.isSmallerThan`.

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
jonalmeida: We could because the bitmap is nullable, and if we use `ContextCompat.getDrawable` to get `mozac_ic_close`, it also has a nullable annotation on that function.

samples/browser/src/main/java/org/mozilla/samples/browser/BrowserFragment.kt
jonalmeida: Oh interesting, I didn't know we needed to manually remove a session. I should also check if the sessionId needs to be cleared as well. Thanks!

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
jonalmeida: Fair point!

The reason I also added this is when we support adding browser actions from a `CustomTabConfig` (e.g. a share icon instead of a menu option) I needed to do some clean up. This might need rethinking after consider that we might be adding ids to the menu items.

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
pocmo: Maybe instead of converting the values of `Drawable` you could convert `maxSize` to what Drawable uses when passing it in?

jonalmeida: I could, but I would still need the Resources/DisplayMetrics to know the screen density to make that calculation. I'll try to think of a better way to do this... 🤔

samples/browser/src/main/java/org/mozilla/samples/browser/BrowserFragment.kt
jonalmeida: > We'll also make sure the same happens when pressing "back" i think (although that may be out of the scope for this PR; but important).

I want to handle this case (and I have an updated patch ready for it) but I don't if the CustomTabs**Toolbar**Feature is the right place. Will there be other CT features that would need to have back press handling?

jonalmeida: @pocmo I've updated the PR to handle back presses, but if you think it should be in another feature, I've kept in a separate commit so it can be easily removed.

components/feature/tabs/src/main/java/mozilla/components/feature/tabs/toolbar/TabsToolbarFeature.kt
```diff
@@ -14,13 +15,23 @@ import mozilla.components.concept.toolbar.Toolbar
...
+            }
 
-        toolbar.addBrowserAction(tabsAction)
+            val tabsAction = TabCounterToolbarButton(
+                sessionManager,
+                showTabs
+            )
+
+            toolbar.addBrowserAction(tabsAction)
+        }
```
pocmo: Why is this wrapped in a `run` block?

components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
```diff
@@ -501,3 +501,16 @@ class SessionManager(
         )
     }
 }
+
+inline fun SessionManager.runWithSessionId(
```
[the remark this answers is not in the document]
jonalmeida: Wanted to get an opinion first in case we didn't really want this. :)

components/feature/tabs/src/main/java/mozilla/components/feature/tabs/toolbar/TabsToolbarFeature.kt
jonalmeida: I was adding an early return if it's a custom tab session, but you can't "return" in an init block, so [some recommendations](https://stackoverflow.com/questions/46244409/why-is-return-not-allowed-in-kotlin-init-block) were to wrap it in a run block instead and return that.

components/browser/session/src/main/java/mozilla/components/browser/session/tab/CustomTabConfig.kt
jonalmeida: That's a good point, I didn't consider that..

components/feature/tabs/src/main/java/mozilla/components/feature/tabs/toolbar/TabsToolbarFeature.kt
jonalmeida: So early returns _are_ hard to read, you say? ;)

That's fair though. How about..

```suggestion
    init {
        sessionManager.runWithSession(sessionId) {
            if (!it.isCustomTabSession()) {
                return@runWithSession
            }
           
            val tabsAction = TabCounterToolbarButton(
                sessionManager,
                showTabs
            )

            toolbar.addBrowserAction(tabsAction)
        }
    }
```

jonalmeida:
>  pocmo 3 days ago
> But doesn't that cause a "too large" bitmap to be kept in memory without us using it for anything?
>
> jonalmeida 3 days ago
> That's a good point, I didn't consider that..

I've tried a couple of approaches to get the size check into the CustomTabConfig but I'm unable to find a good one at the moment. I recommend merging this for now so we can continue implementing the rest of the custom tab feature work. I've filed https://github.com/mozilla-mobile/android-components/issues/1724 as well to track getting this check back in.

Things tried:

1. Passing density via the intent:

   - Required passing the DisplayMetrics/density to the IntentProcessor (`feature-intent`)
   - `feature-intent` needed to depend on `browser-session` which seemed like an overkill.
2. Using a hard coded max value for the bytes the image can take.
   - I tried this naive solution but it still felt too permissive since the average max value is still quite high for a high density screen to be allowed for a device with a low density screen.

| Device     | `getAllocatedBytes` | `getRowBytes` |
| ---------- | ------------------- | ------------- |
| Pixel 2    | 15876               | 252           |
| Nexus 5X   | 15876               | 252           |
| Pixel XL   | **28224**           | **336**       |
| Android TV | 4096                | 128           |

jonalmeida:
@pocmo fixed #1724 here since this hasn't been reviewed yet.

PR #1678, opened 2019-01-09
Temporarily disable intermittent failing test until fixed

jonalmeida:
I'm seeing a lot of build failures due to this test. How do we feel about temporarily disabling this test until we get this fixed? (I'd like to work on it next sprint.)


pocmo:
But please move the issue into "in progress". My experience is that disabled tests get forgotten pretty fast when priorities change. :)

PR #1683, opened 2019-01-09
[glean] Sanitize the application id to be pipeline-friendly

Dexterp37:
This makes sure the automatically detected application is processed in order to remove all non alphanumeric
characters and replace them with dashes.


Dexterp37:
> Is a trailing or leading '-' valid in the pipeline? Your regex sanitizer could replace leading or trailing non-alphanumeric characters with a dash. I'm not sure this is possible in an application ID or not...
> 
> After looking up the rules for android app id's, it couldn't be a leading dash as all segments of app id need to start with a letter, but it could result in a trailing dash since _ character is allowed in app id.

I don't think the pipeline cares too much, as long as the character can be used. @fbertsch, thoughts?

Dexterp37:
> > After looking up the rules for android app id's, it couldn't be a leading dash as all segments of app id need to start with a letter, but it could result in a trailing dash since _ character is allowed in app id.
> 
> I don't think the pipeline cares too much, as long as the character can be used. @fbertsch, thoughts?

I had a conversation with Frank yesterday. While trailing dashes can look *weird*, they are not a technical problem for the pipeline, so we can move forward with this.

PR #1685, opened 2019-01-09
WIP: [glean] 1518163: Add the a11y_services metric in the baseline ping.

mdboom:
This works, and I was able to test it in the sample app.  However, I'm uncertain how to write a test for it, since it doesn't seem that the accessibility service is enabled in the testing environment, and I can't figure out how to programmatically do that...

### DIFF
--- a/components/service/glean/metrics.yaml
+++ b/components/service/glean/metrics.yaml
@@ -102,14 +102,18 @@ baseline:
   #   notification_emails:
   #     - telemetry-client-dev@mozilla.com
 
-  # accessibility_services:
-  #   type: string_list
-  #   description: |
-  #     List of accessibility services enabled on the device.
-  #   bugs:
-  #     - 1497894
-  #   notification_emails:
-  #     - telemetry-client-dev@mozilla.com
+  a11y_services:
+    type: string_list
+    description: |
+      List of accessibility services enabled on the device.
+    send_in_pings:
+      - baseline
+    bugs:
+      - 1497894
+    reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1512938#c3
+    notification_emails:
+      - telemetry-client-dev@mozilla.com
 
 glean.internal.metrics:
   client_id:
@@ -125,4 +129,3 @@ glean.internal.metrics:
       - https://bugzilla.mozilla.org/show_bug.cgi?id=1512938#c3
     notification_emails:
       - telemetry-client-dev@mozilla.com
-
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
@@ -4,10 +4,14 @@
 
 package mozilla.components.service.glean
 
+import android.view.accessibility.AccessibilityManager
+import android.accessibilityservice.AccessibilityServiceInfo
 import android.arch.lifecycle.ProcessLifecycleOwner
 import android.content.Context
+import android.content.pm.PackageManager
 import android.os.Build
 import android.support.annotation.VisibleForTesting
+import android.support.v4.view.accessibility.AccessibilityManagerCompat.getEnabledAccessibilityServiceList
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.launch
@@ -169,6 +173,32 @@ open class GleanInternalAPI {
     }
 
     /**
+     * Records a list of the currently enabled accessibility services.
+     *
+     * https://developer.android.com/reference/android/view/accessibility/AccessibilityManager.html
+     * @param accessibilityManager The system's [AccessibilityManager] as
+     * returned from applicationContext.getSystemService
+     * @param packageManager The system's [PackageManager], as from
+     * applicationContext.packageManager
+     * @returns services A list of strings describing the enabled services. If
+     *     the accessibility manager is disabled, returns null.
+     */
+    internal fun getEnabledAccessibilityServices(
+        accessibilityManager: AccessibilityManager,
+        packageManager: PackageManager
+    ): List<String>? {
+        if (!accessibilityManager.isEnabled) {
+            logger.info("AccessibilityManager is disabled")
+            return null
+        }
+        return accessibilityManager.getEnabledAccessibilityServiceList(
+            AccessibilityServiceInfo.FEEDBACK_ALL_MASK
+        ).map {
+            it.loadDescription(packageManager)
+        }
+    }
+
+    /**
      * Initialize the core metrics internally managed by Glean (e.g. client id).
      */
     private fun initializeCoreMetrics(applicationContext: Context) {
@@ -214,6 +244,14 @@ open class GleanInternalAPI {
 
         // Set the CPU architecture
         Baseline.architecture.set(Build.SUPPORTED_ABIS[0])
+
+        // Set the enabled accessibility services
+        getEnabledAccessibilityServices(
+            applicationContext.getSystemService(Context.ACCESSIBILITY_SERVICE) as AccessibilityManager,
+            applicationContext.packageManager
+        ) ?.let {
+            Baseline.a11yServices.set(it)
+        }
     }
 
     /**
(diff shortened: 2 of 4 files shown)


Dexterp37:
@pocmo have you ever dealt with accessibility services enumeration? We need to list them as part of the `baseline` ping but not me nor others working on glean have much experience on this. Is this the right approach?

Dexterp37:
> This works, and I was able to test it in the sample app. However, I'm uncertain how to write a test for it, since it doesn't seem that the accessibility service is enabled in the testing environment, and I can't figure out how to programmatically do that...

The accessibility services manager seems to have a [Robolectric Shadow](http://robolectric.org/javadoc/4.0/org/robolectric/shadows/ShadowAccessibilityManager.html), so it appears to be testable by using the `setEnabledAccessibilityServiceList` method (and others). See my comments in the review :)

mdboom:
Thanks -- my Googling-fu wasn't getting me anywhere.  That looks promising.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -168,6 +171,32 @@ open class GleanInternalAPI {
         }
     }
 
+    /**
+     * Records a list of the currently enabled accessibility services.
+     *
+     * https://developer.android.com/reference/android/view/accessibility/AccessibilityManager.html
+     */
+    private fun recordEnabledAccessibilityServices(applicationContext: Context) {
```
Dexterp37: Let's make this function `internal` rather than `private`, to make it accessible from tests. Moreover, if instead of passing the `Context` we pass the `accessibilityManager` and `packageManager`, then we could pass in mocks/shadows and unit test this function.

We could additionally make this function return a List, to make it easier to test. We could always set the data in the baseline ping from the call site.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -168,6 +171,32 @@ open class GleanInternalAPI {
         }
     }
 
+    /**
+     * Records a list of the currently enabled accessibility services.
+     *
+     * https://developer.android.com/reference/android/view/accessibility/AccessibilityManager.html
```
Dexterp37: nit: let's document parameters

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -168,6 +171,32 @@ open class GleanInternalAPI {
...
+     */
+    private fun recordEnabledAccessibilityServices(applicationContext: Context) {
+        val accessibilityManager = applicationContext.getSystemService(
+            Context.ACCESSIBILITY_SERVICE
+        ) as AccessibilityManager
+        if (!accessibilityManager.isEnabled()) {
+            return
```
Dexterp37: Let's add a debug log line here mentioning that the service is disabled.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -168,6 +171,32 @@ open class GleanInternalAPI {
...
+        }
+        val serviceInfos: List<AccessibilityServiceInfo>
+        serviceInfos = accessibilityManager.getEnabledAccessibilityServiceList(
```
Dexterp37: Let's make this more Kotlin...ionic? We could use `map` or something along the lines of:

```
return accessibilityManager.getEnabledAccesibilityServiceList(...).map {
  it.loadDescription(packageManager)
}
```

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -168,6 +171,32 @@ open class GleanInternalAPI {
...
+            AccessibilityServiceInfo.FEEDBACK_ALL_MASK
+        )
+
+        val packageManager = applicationContext.packageManager
+        val accessibilityNames: MutableList<String> = mutableListOf()
+        for (serviceInfo in serviceInfos) {
+            accessibilityNames.add(serviceInfo.loadDescription(packageManager))
+        }
+
+        Baseline.a11yServices.set(accessibilityNames)
```
Dexterp37: If we move this outside of the function, it would make it easier to test it.

pocmo:
> The accessibility services manager seems to have a Robolectric Shadow, so it appears to be testable by using the setEnabledAccessibilityServiceList method (and others). See my comments in the review :)

Yeah, usually one of:

* Use a Robolectic shadow if it exists
* Write a custom shadow with the Robolectric system (Which basically means replacing a system class with your own implementation at test runtime)
* Refactor your code so that you can inject your own mocks or so that you can mock a "helper" function that returns the data you are interested in - without needing to talk to any system classes in your test.

> @pocmo have you ever dealt with accessibility services enumeration? We need to list them as part of the baseline ping but not me nor others working on glean have much experience on this. Is this the right approach?

This looks reasonable. But I think @eeejay can answer that way better than I.

components/service/glean/metrics.yaml
```diff
@@ -99,14 +99,18 @@ baseline:
...
 
-  # accessibility_services:
-  #   type: string_list
-  #   description: |
-  #     List of accessibility services enabled on the device.
-  #   bugs:
-  #     - 1497894
-  #   notification_emails:
-  #     - telemetry-client-dev@mozilla.com
+  a11y_services:
```
Dexterp37: We also need to rename this field in `baseline.md`, where's it's written as `accessibility_services`

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -168,6 +172,23 @@ open class GleanInternalAPI {
         }
     }
 
+    /**
+     * Records a list of the currently enabled accessibility services.
+     *
+     * https://developer.android.com/reference/android/view/accessibility/AccessibilityManager.html
+     * @param accessibilityManager The system's AccessibilityManager as returned from applicationContext.getSystemService
+     * @param packageManager The system's PackageManager, as from applicationContext.packageManager
+     */
```
Dexterp37: Let's also document the return value.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -168,6 +172,23 @@ open class GleanInternalAPI {
         }
     }
 
+    /**
+     * Records a list of the currently enabled accessibility services.
+     *
+     * https://developer.android.com/reference/android/view/accessibility/AccessibilityManager.html
+     * @param accessibilityManager The system's AccessibilityManager as returned from applicationContext.getSystemService
```
Dexterp37: nit: I think Android's api objects can be referenced in the docs by using `[AccessibilityManager]` (adding the square brackets)

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -168,6 +172,23 @@ open class GleanInternalAPI {
...
+     * @param packageManager The system's PackageManager, as from applicationContext.packageManager
```
Dexterp37: nit: same here, square brackets

pocmo:
> It doesn't appear that ShadowAccessibilityManager actually inherits from AccessibilityManager, so I can't pass it in to my class as an AccessibilityManager. If that's the case, how is this useful?

You continue to use the AccessibilityManager from the context / system. The shadow is a way to "manipulate" how it will behave in this case.

Here is an example:

```Kotlin
    @Test
    fun shadowTest() {
        // Actual AccessibilityManager:

        val accessibilityManager = RuntimeEnvironment.application.getSystemService(Context.ACCESSIBILITY_SERVICE)
            as AccessibilityManager

        // Setting up shadow:

        val shadow: ShadowAccessibilityManager = Shadows.shadowOf(accessibilityManager)

        val serviceMock1: AccessibilityServiceInfo = mock()
        val serviceMock2: AccessibilityServiceInfo = mock()

        shadow.setInstalledAccessibilityServiceList(listOf(serviceMock1, serviceMock2))

        // Using actual accessibility manager:

        val list = accessibilityManager.installedAccessibilityServiceList
        assertEquals(2, list.size)
        assertEquals(serviceMock1, list[0])
        assertEquals(serviceMock2, list[1])
    }
```

mdboom:
Thanks for the help, everyone.  I think I have a test that is reasonably testing this new code now.

PR #1687, opened 2019-01-09
Update Dockerfile to create smaller image..

pocmo:
There are three major changes:

* Do not download any SDK bits with `sdkmanager`. Gradle will download whatever it needs itself nowadays. This will make sure that we do not download things that we won't need.
* Use `--depth=1` to not clone the whole git history that we do not need.
* Run `clean` after executing all the gradle commands. There's no need to keep all the build artifacts around. They will be outdated pretty fast and are just huuuuuge. If we want build artifacts then they should be setup as a gradle build cache in a worker cache (instead of the build folders)

The image size dropped from 6 GB to 3 GB on docker hub (And 11.2 GB to 5.16 GB locally according to `docker images`).

### DIFF
--- a/automation/docker/Dockerfile
+++ b/automation/docker/Dockerfile
@@ -66,17 +66,13 @@ RUN curl -L https://dl.google.com/android/repository/sdk-tools-linux-${ANDROID_S
     && rm sdk.zip \
     && mkdir -p /build/android-sdk/.android/ \
     && touch /build/android-sdk/.android/repositories.cfg \
-    && yes | sdkmanager --licenses \
-    && sdkmanager --verbose "platform-tools" \
-        "platforms;android-${ANDROID_PLATFORM_VERSION}" \
-        "build-tools;${ANDROID_BUILD_TOOLS}" \
-        "extras;android;m2repository" \
-        "extras;google;m2repository"
+    && yes | sdkmanager --licenses
+
 #----------------------------------------------------------------------------------------------------------------------
 #-- Project -----------------------------------------------------------------------------------------------------------
 #----------------------------------------------------------------------------------------------------------------------
 
-RUN git clone $PROJECT_REPOSITORY
+RUN git clone --depth=1 $PROJECT_REPOSITORY
 
 WORKDIR /build/android-components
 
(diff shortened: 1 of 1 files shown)


pocmo:
Pushed the image as 1.14. PR to update is: #1686

pocmo:
@mitchhentges This should include "arrow" now too :)

automation/docker/Dockerfile
```diff
@@ -66,17 +66,13 @@ RUN curl -L https://dl.google.com/android/repository/sdk-tools-linux-${ANDROID_S
     && rm sdk.zip \
     && mkdir -p /build/android-sdk/.android/ \
     && touch /build/android-sdk/.android/repositories.cfg \
-    && yes | sdkmanager --licenses \
-    && sdkmanager --verbose "platform-tools" \
```
mitchhentges: I didn't know that `gradle` will install these on its own now, that's pretty awesome!
However, by having them installed, we have them cached for our builds - are we sure we want to download these fresh from Google's repo each build?

automation/docker/Dockerfile
```diff
@@ -85,13 +81,9 @@ RUN ./gradlew clean \
     && ./gradlew androidDependencies \
     && ./gradlew --stop \
     && ./gradlew --no-daemon assemble \
-    && ./gradlew --no-daemon test \
+    && ./gradlew --no-daemon -Pcoverage test \
```
mitchhentges: If the point of the `./gradle` commands here are to download dependencies (so that they'll be cached in the image), I'm not sure that we need to do `-Pcoverage` (it doesn't influence the downloaded dependencies, does it?)

automation/docker/Dockerfile
pocmo: Later in the Dockerfile we invoke gradle which will perform the download. So it will be cached in the image!

automation/docker/Dockerfile
pocmo: Without `-Pcoverage` we didn't download the dependencies needed for generating the coverage reports. That's primarily jacoco and its transitive dependencies. I recently updated the Docker image and then a task failed because it couldn't download jacoco and then I realized that we never do that when building the image. :)

PR #1689, opened 2019-01-09
Allow easy configuration of gradle dependency substitution for A-S 

grigoryk:
Same thing as we have in the `reference-browser`, except we don't megazord in a-c so don't need those hoops. https://github.com/mozilla-mobile/reference-browser/pull/356

To configure local a-s substitution, place `substitutions.application-services.dir=../application-services/path` in `local.properties`. (relative or absolute path is fine)


pocmo:
🎉 Awesome!

Let's document this somewhere here:
https://mozilla-mobile.github.io/android-components/contributing/

PR #1692, opened 2019-01-09
Bump app services dep, and include logins API changes

### LINKED ISSUE #1690: Bump A-S to 0.13.1
A-S just cut a new release which fixes a few issues that we notice in `reference-browser`. Let's bump our dependency to fix these in `r-b`, and look out for any regressions.

thomcc:
@csadilek, @grigoryk 

Fixes #1690

### DIFF
--- a/components/service/sync-logins/README.md
+++ b/components/service/sync-logins/README.md
@@ -161,6 +161,12 @@ Currently there is no way to change the key, once set (see https://github.com/mo
 
 Part is in this tree, but most is in https://github.com/mozilla/application-services. Specifically, there are two pieces not present in this repository, an [android-specific piece written in Kotlin](https://github.com/mozilla/application-services/tree/master/logins-sql/tree/master/logins-api/android), and a [cross-platform piece written in Rust](https://github.com/mozilla/application-services/tree/master/logins-sql).
 
+### What's the difference between `wipe` and `wipeLocal`?
+
+- `wipe` deletes all records, but remembers that it has deleted them. The next time we sync, we'll delete the remote versions of the deleted records too (e.g. if you wipe, then another client uploads a record, and then you sync, the other clients record will not be deleted).
+- `wipeLocal` deletes all local data from the database, bringing us back to the state prior to the first local write (or sync). That is, it returns it to an empty database.
+- Previous versions of this library offered a third variant, `reset`, which deletes last sync timestamps, and what the database believes is stored remotely. This is almost never what you want, and so it was removed, we mention it only to assist when updating this library.
+
 ## License
 
     This Source Code Form is subject to the terms of the Mozilla Public
--- a/components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/AsyncLoginsStorage.kt
+++ b/components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/AsyncLoginsStorage.kt
@@ -103,16 +103,30 @@ typealias RequestFailedException = mozilla.appservices.logins.RequestFailedExcep
 interface AsyncLoginsStorage : AutoCloseable {
     /** Locks the logins storage.
      *
-     * @throws MismatchedLockException if we're already locked
+     * @rejectsWith [MismatchedLockException] if we're already locked
      */
     fun lock(): Deferred<Unit>
 
     /** Unlocks the logins storage using the provided key.
      *
-     * @throws InvalidKeyException if the encryption key is wrong, or the db is corrupt
-     * @throws MismatchedLockException if we're already unlocked
+     * @rejectsWith [InvalidKeyException] if the encryption key is wrong, or the db is corrupt
+     * @rejectsWith [MismatchedLockException] if we're already unlocked
+     * @rejectsWith [LoginsStorageException] if the storage is locked, and on unexpected
+     *              errors (IO failure, rust panics, etc)
      */
     fun unlock(encryptionKey: String): Deferred<Unit>
+
+    /**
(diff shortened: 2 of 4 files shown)


grigoryk:
0.13.0 breaks FxA (manifests as IO exceptions when talking to network), @thomcc is investigating.

components/service/sync-logins/README.md
```diff
@@ -161,6 +161,12 @@ Currently there is no way to change the key, once set (see https://github.com/mo
 
 Part is in this tree, but most is in https://github.com/mozilla/application-services. Specifically, there are two pieces not present in this repository, an [android-specific piece written in Kotlin](https://github.com/mozilla/application-services/tree/master/logins-sql/tree/master/logins-api/android), and a [cross-platform piece written in Rust](https://github.com/mozilla/application-services/tree/master/logins-sql).
 
+### What's the difference between `reset`, `wipe` and `wipeLocal`?
+
+- `reset` deletes last sync timestamps, and what the database believes is stored remotely. This is almost never what you want, and so it is deprecated.
+- `wipe` deletes all records, but remembers that it has deleted them. The next time we sync, we'll delete the remote versions of the deleted records too (e.g. if you wipe, then another client uploads a record, and then you sync, the other clients record will not be deleted).
+- `wipeLocal` deletes all local data from the database, bringing us back to the state prior to the first sync.
```
grigoryk: ... and prior to first local write, as well? E.g. it'll be just an empty database? As written, one might thing that only "from sync" data will be wiped.

components/service/sync-logins/README.md
```diff
@@ -161,6 +161,12 @@ Currently there is no way to change the key, once set (see https://github.com/mo
 
 Part is in this tree, but most is in https://github.com/mozilla/application-services. Specifically, there are two pieces not present in this repository, an [android-specific piece written in Kotlin](https://github.com/mozilla/application-services/tree/master/logins-sql/tree/master/logins-api/android), and a [cross-platform piece written in Rust](https://github.com/mozilla/application-services/tree/master/logins-sql).
 
+### What's the difference between `reset`, `wipe` and `wipeLocal`?
```
grigoryk: Thanks for documenting these!

components/service/sync-logins/README.md
```diff
@@ -161,6 +161,12 @@ Currently there is no way to change the key, once set (see https://github.com/mo
...
+
+- `reset` deletes last sync timestamps, and what the database believes is stored remotely. This is almost never what you want, and so it is deprecated.
```
grigoryk: Do we have any consumers even using it `reset`? I'd just remove it if not...

components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/AsyncLoginsStorage.kt
```diff
@@ -111,9 +111,21 @@ interface AsyncLoginsStorage : AutoCloseable {
      *
      * @throws InvalidKeyException if the encryption key is wrong, or the db is corrupt
      * @throws MismatchedLockException if we're already unlocked
+     * @throws LoginsStorageException On unexpected errors (IO failure, rust panics, etc)
      */
```
grigoryk: Perhaps we actually want to use `@Throws` annotation?
> This annotation indicates what exceptions should be declared by a function when compiled to a JVM method.

components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/AsyncLoginsStorage.kt
```diff
@@ -311,24 +391,13 @@ data class SyncableLoginsStore(
...
-        // Instead of synchronizing on 'store' let's be optimistic and just swallow mismatched locks.
 
-        if (store.isLocked()) {
-            try {
-                store.unlock(key().await()).await()
-            } catch (e: MismatchedLockException) {
-                // Oh well, it's already unlocked!
-            }
-        }
+        store.ensureUnlocked(key().await()).await()
```
grigoryk: At face value, this looks unsafe to me. Does anything prevent another thread from locking the store after you've ensured that it's unlocked and before you operate on it?

I'd prefer to see something like...
```
store.whileUnlocked(key().await()) {
    return block(store)
}
```
... and let the underlying implementation deal with locking, mutexing, whatever.

components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/AsyncLoginsStorage.kt
thomcc: I'll rename it to @rejectsWith, or something.

components/service/sync-logins/README.md
thomcc: Hm, well, yes, lockbox uses it. But they should use `wipeLocal` instead. I don't mind taking it out here.

components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/AsyncLoginsStorage.kt
thomcc: After discussion in slack, this is tricky and should be done in a follow up if desired.

components/service/sync-logins/README.md
thomcc: Okay, I'm keeping the mention of `reset` to make it clear that it's not the case that reset was renamed to wipeLocal (while being very clear that it no longer exists).

PR #1694, opened 2019-01-10
Don't invoke gradle in decision task

### LINKED ISSUE #1649: Decision task should not invoke gradle
The decision task should not be a heavyweight operation - it collects enough information to figure out which tasks need to run, then delegates them.

However, right now, we invoke gradle to get module definitions and the current version. This increases the minimum size of instance we can use to run the decision task, and increases overall costs and waiting time.

(Inspired by a conversation with ncalexan in Orlando)

mitchhentges:
Fixes #1649.

I've verified that `lib.build_config.module_definitions()` returns an identical payload to before this refactor.


build.gradle
```diff
@@ -187,22 +187,5 @@ task ktlint(type: JavaExec, group: "verification") {
...
-        println "modules: " + groovy.json.JsonOutput.toJson(modules)
-    }
-}
-
-task printVersion {
-    doLast {
-        println("version: " + Config.componentsVersion)
-    }
-}
-
```
pocmo: Note that `printVersion` is used by `utomation/docs/generate_api-reference_release.sh` too.

.buildconfig.yml
```diff
@@ -0,0 +1,254 @@
...
+    description: 'A library for integrating with Firefox Sync - Logins.'
+    publish: true
+  service-fretboard:
+    path: components/service/fretboard
+    description: 'An Android framework for segmenting users in order to run A/B tests and rollout features gradually.'
+    publish: true
+  service-glean:
+    path: components/service/glean
+    description: 'A client-side telemetry SDK for collecting metrics and sending them to the Mozilla telemetry service'
+    publish: true
```
Dexterp37: @pocmo does this mean the library will be officially published? Is this retaining the same behaviour as the old code?

Dexterp37: > Define "officially". :)

Good point :) Together with other libraries from android-components.

> But yes, it's getting published since the beginning :)
> https://maven.mozilla.org/?prefix=maven2/org/mozilla/components/service-glean/

Perfect then!

build.gradle
mitchhentges: I've updated [`generate_api-reference_release.sh`](https://github.com/mozilla-mobile/android-components/pull/1694/files#diff-e570e00f7f642277c394446ff507c8deL9) as part of the PR :smile:

pocmo: Okay, I'm blind 😎

mitchhentges:
Sorry for re-ping, I've realized that force-pushing clears code review.
Only the [last commit](https://github.com/mozilla-mobile/android-components/pull/1694/commits/3aa162147f96746109d74411fd9bb9edf125a18e) is new

PR #1695, opened 2019-01-10
Adds pyyaml to docker image

mitchhentges:
Multiple tasks now need `pyyaml`, so might as well add it to the shared docker image

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -88,7 +88,7 @@ tasks:
             - statuses  # Automatically added by taskcluster-github. It must be explicit because of Chain of Trust
           payload:
             maxRunTime: 600   # Decision should remain fast enough to schedule a handful of tasks
-            image: mozillamobile/android-components:1.14
+            image: mozillamobile/android-components:1.15
             command:
               - /bin/bash
               - --login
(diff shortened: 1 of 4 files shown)


.taskcluster.yml
```diff
@@ -207,7 +207,6 @@ tasks:
                               git fetch ${repository} --tags
                               && git config advice.detachedHead false
                               && git checkout ${tag}
-                              && pip install pyyaml
```
pocmo: I assume that we need the new Docker image on Docker Hub (and update the version here) before we can land this part -- or why pyyaml never needed here?. However to re-build the image it would be nice to land the Dockerfile changes first. Maybe we should separate those changes? :)

pocmo:
I applied the changes locally, re-built the image and am pushing it to Docker hub right now (will be 1.15)

.taskcluster.yml
mitchhentges: You're right, `pyyaml` is needed for that task, too.
It's true that there's a circular dependency-ish thing going on here where it's weird to deploy the docker image before it's merged, but you can't merge until the new docker image is deployed.

I see below that you rebuilt the docker image, but next time I'll separate such changes :+1:

PR #1696, opened 2019-01-10
[glean] Fix 1518835: Clarify issues around character length vs. utf8 bytes.

mdboom:
Also cleans up some duplicate constants and an obsolete comment.

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/StringMetricType.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/StringMetricType.kt
@@ -30,15 +30,15 @@ data class StringMetricType(
     private val logger = Logger("glean/StringMetricType")
 
     companion object {
-        // Maximum length of any passed value string, in UTF8 byte sequence length.
+        // Maximum length of any passed value string, in characters.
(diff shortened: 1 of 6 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/StringMetricType.kt
```diff
@@ -30,15 +30,15 @@ data class StringMetricType(
...
+        // Maximum length of any passed value string, in characters.
         private const val MAX_LENGTH_VALUE = 50
     }
 
     /**
      * Set a string value.
      *
      * @param value This is a user defined string value. The maximum length of
-     * this string is defined the metrics.yaml file by max_length.
+     * this string must not exceed [MAX_LENGTH_VALUE] characters.
```
badboy: Sorry for the drive-by, but maybe this doc string should include that it's truncated when longer? "must not exceed" sounds like this method fails if this condition is not met.

mdboom: Good point!  No problem with drive-bys :)

PR #1701, opened 2019-01-10
Switch to android.net.Uri in domain matching logic

grigoryk:
It's more forgiving than java.net.URI, which is what we want here. One downside of android.net.Uri is that it will vary by API version, so we might get different behaviours. 

This is intended to help with https://github.com/mozilla-mobile/reference-browser/issues/396,
although I wasn't able to add a failing unit test that would match that crash. @pocmo do you know what toolbar input crashed r-b for you?

### DIFF
--- a/components/support/utils/src/main/java/mozilla/components/support/utils/DomainMatcher.kt
+++ b/components/support/utils/src/main/java/mozilla/components/support/utils/DomainMatcher.kt
@@ -5,8 +5,8 @@
 
 package mozilla.components.support.utils
 
+import android.net.Uri
 import java.net.MalformedURLException
-import java.net.URL
 
 data class DomainMatch(val url: String, val matchedSegment: String)
 
@@ -31,7 +31,7 @@ private fun basicMatch(query: String, urls: Sequence<String>): String? {
         }
 
         val url = try {
-            URL(rawUrl)
+            Uri.parse(rawUrl)
         } catch (e: MalformedURLException) {
             null
         }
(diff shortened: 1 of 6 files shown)


pocmo:
> @pocmo do you know what toolbar input crashed r-b for you?

Unfortunately I do not. The stacktrace mentions the "about" protocol and the only pages I remember being on were `about:version` and `about:crashes`. And neither crashes the browser.

components/support/utils/src/main/java/mozilla/components/support/utils/DomainMatcher.kt
```diff
@@ -31,7 +31,7 @@ private fun basicMatch(query: String, urls: Sequence<String>): String? {
         }
 
         val url = try {
-            URL(rawUrl)
+            Uri.parse(rawUrl)
         } catch (e: MalformedURLException) {
             null
         }
```
pocmo: nit: It looks like this could bail out early and we do not need to continue with a null `Uri`.

pocmo: Actually it looks like `Uri` will never throw `MalformedURLException`. You just won't get values for certain getters.

components/support/utils/src/main/java/mozilla/components/support/utils/DomainMatcher.kt
```diff
@@ -62,17 +62,19 @@ private fun matchSegment(query: String, rawUrl: String): String? {
         return rawUrl
     }
 
-    val url = URL(rawUrl)
-    if (url.host.startsWith(query)) {
-        return url.host + url.path + url.port.orEmpty()
-    }
+    val url = Uri.parse(rawUrl)
+    url.host?.let { host ->
```
pocmo: nit: I usually prefer something like:

```Kotlin
val host = url.host ?: return null
```

.. instead of a block causing indentation. :)

PR #1704, opened 2019-01-11
Closes #1518: Select session of same type when session is removed

### LINKED ISSUE #1518: SessionManager: On removing tab select private/normal tab of same type
There's another case when removing tabs that we haven't implemented yet: When removing the selected tab then the new selected tab should be of the same type (private / normal). Only when no (private) tab exists anymore we'd select a normal tab.

csadilek:
Mostly re-useing our `findNearby` (oscillating) logic for this :).


components/browser/session/src/test/java/mozilla/components/browser/session/SessionManagerTest.kt
```diff
@@ -972,4 +972,68 @@ class SessionManagerTest {
...
+        manager.select(regular2)
+        manager.remove(regular2)
+        assertEquals(regular3, manager.selectedSession)
+
+        manager.remove(regular3)
+        assertEquals(regular1, manager.selectedSession)
+
+        // Removing the last regular session should NOT cause a private session to be selected
+        manager.remove(regular1)
+        assertNull(manager.selectedSession)
```
pocmo: I just compared that with Fennec and saw that it indeed does not select a regular session in this case. However they create a new regular tab. So I wonder if we should invoke the `defaultSession` lambda in this case and select the new session?

csadilek: Good idea, yes! I will update this.

csadilek: @pocmo actually, this already works, because if we return NO_SELECTION from `recalculateSelectionIndex` we invoke the defaultsession lambda: https://github.com/mozilla-mobile/android-components/blob/master/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt#L292

pocmo: Perfect :)

pocmo: Can we write a test for that? 🙃

csadilek: Already on it.

csadilek: @pocmo test added now!

PR #1713, opened 2019-01-11
WIP - demonstrate how to unit test Application Services native code.

ncalexan:


components/service/sync-logins/src/test/java/mozilla/components/service/sync/logins/AsyncLoginsStorageTest.kt
```diff
@@ -0,0 +1,70 @@
...
+import kotlinx.coroutines.newSingleThreadContext
+import kotlinx.coroutines.runBlocking
+import java.io.File
+
+@RunWith(RobolectricTestRunner::class)
+class AsyncLoginsStorageTest {
+    private val testMainScope = CoroutineScope(newSingleThreadContext("Test"))
+
+    @Test
+    fun `can do some stuff`() {
```
grigoryk: A bit more concise:
```
fun `can do some stuff`() = runBlocking {
...
}
```

build.gradle
```diff
@@ -33,6 +33,11 @@ allprojects {
             url "https://maven.mozilla.org/maven2"
         }
 
+        maven {
+            // Application Services `-forUnitTests` variants.
+            url "https://dl.bintray.com/ncalexander/application-services"
```
grigoryk: How different are these artifacts from what non-unit test environments consume?

components/service/sync-logins/src/test/java/mozilla/components/service/sync/logins/AsyncLoginsStorageTest.kt
ncalexan: I needed to put the `runBlocking` _inside_ the `use { }` block.  I would be happy to be proven wrong.

ncalexan:
> Very cool. I'd like to learn more about `-forUnitTests` variants! How different are they from regular? Why do we need different ones?

There are two major differences:

1. The `-forUnitTests` variants are, well, for host JVMs, not target VMs.  They target Linux, Windows, and macOS.  The regular variants target ARM, ARM64, and x86 (we don't publish x86_64 yet).  So the they include additional `.dll` and `.dylib` (for Windows and macOS, respectively) that make no sense on Android.

1. The `-forUnitTests` variants pack those libraries into the "Java resources" part of the package, not into the "Android resources" or the "Android assets" parts of the package.  JNA has a pattern for looking for native libraries across platforms; this is how we conform to it.

If we put the additional libraries into every Android AAR, they would "turn up" in APKs produced in awkward ways.  In fact, they conflict with some special GDB library processing that the Android-Gradle plugin itself does.

Under the hood, the code is basically identical: it's really just `cargo --target ...` differences.  There may, of course, be platform differences internally; but in the a-s stuff, there's not much difference.

All of this gets more exciting, of course, if and when a-s code starts to really rely on the Android functionality from its own native code.  Logging to Android logcat with the native `liblog`, for example.  We don't do that now, and hopefully we won't -- 'cuz that would mean we couldn't do this kind of testing, and I think this testing is critical to our long-term success story.  In fact we require Robocop (for `Log` and `JSONObject`), but both of those are happening at the JVM-level, not at the native library level.

ncalexan:
I've just realized that these `-forUnitTests` variants don't play with Gradle composite builds... at all.  Limitations in Maven and Gradle itself make this very challenging.  I'll keep thinking about it.

ncalexan:
@csadilek here's a rebased version; `./gradlew :service-sync-logins:testDebugUnitTest` passes for me locally.  https://github.com/mozilla-mobile/android-components/pull/1713#issuecomment-453659952 is still the same as it was; I have had no bright ideas here.  The way to improve this is probably https://github.com/mozilla/application-services/issues/562, which I won't get to any time soon.

grigoryk:
Even with problems with composite builds (which may go away soon?), exercising native code in our tests has a ton of value.

This is happening for history storage in https://github.com/mozilla-mobile/android-components/pull/2943.

PR #1721, opened 2019-01-14
Closes #1681 - Add ClipboardSuggestionProvider for providing URL suggestions from clipboard.

### LINKED ISSUE #1681: Implement AwesomeBar.SuggestionProvider providing suggestion from clipboard
In our products we want to make it easy to navigate to a URL that is in the clipboard. We could implement a suggestion provider that provides the URL from the clipboard (if there's any).

pocmo:


components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/ClipboardSuggestionProvider.kt
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    if (clipboardManager.primaryClipDescription?.hasMimeType(MIME_TYPE_TEXT_PLAIN) != true) {
+        return null
+    }
+
+    val item = clipboardManager.primaryClip?.getItemAt(0)
+        ?: return null
+
+    return item.text?.toString()
```
Amejia481: I found this function a bit difficult to read, because of multiple returns. Maybe smaller functions could help :)

```kotlin
private fun getTextFromClipboard(clipboardManager: ClipboardManager): String? {
    return with(clipboardManager) {
        if (hasNotPlainText() or isPrimaryEmpty()) {
            null
        } else {
            first().toString()
        }
    }
}

fun ClipboardManager.hasNotPlainText(): Boolean {
    return !(primaryClipDescription?.hasMimeType(MIME_TYPE_TEXT_PLAIN) ?: false)
}

fun ClipboardManager.isPrimaryEmpty(): Boolean {
    return primaryClip?.itemCount == 0
}

fun ClipboardManager.first(): ClipData.Item? {
    return primaryClip?.getItemAt(0)
}
```

PR #1588, opened 2018-12-18
Bug 1512631 - Use mobile-{1,3}-decision worker types and allow staging releases from forks

JohanLorenzo:

PR #1729, opened 2019-01-15
Bug 1519374 - Rename build worker types into mobile-X-b-andrcmp

JohanLorenzo:
Follows #1536 and #1638 up

`mobile-X-b-components` maps what exists on Firefox (e.g.: https://tools.taskcluster.net/aws-provisioner/gecko-1-b-android/view). It stands for: 

* `mobile` refers to the mobile projects. On Firefox, the name is `gecko`. We use it for the decision workers already. See #1588
* `X` is a trust level. It maps the levels in Gecko: 1 is the least trusted, 3 is the most. See #1588
* `b` means "build", like on Gecko.
* `components` refers to the project. For other mobile projects, we would have `fenix`, `focus`, `ref-browser`.

Please note the whole name is limited to 22 characters. 

Note: I didn't put `-g` like we had in `android-components-g`. The reasons are:
* `mobile-X-b-components` is already 22-char-long. Putting `-g` in would mean we had to truncate components somehow. I couldn't find an appealing mix of strings.
* If I'm not mistaken, all mobile build/lint tasks run on gradle, for now. Moreover, they all load the same docker image. Therefore, `-g` doesn't discriminate worker types. 

@ncalexan @grigoryk @pocmo @MihaiTabara @mitchhentges: What do you think about it?

Please do not merge until we agree on the name. I need to create the level 3 worker type before then.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -52,9 +52,9 @@ tasks:
         else: mobile-1-decision
 
       build_worker_type:
-        $if: 'is_repo_trusted && tasks_for in ["github-release", "cron"]'
-        then: gecko-focus
-        else: android-components-g
+        $if: 'is_repo_trusted'
+        then: mobile-3-b-andrcmp
+        else: mobile-1-b-andrcmp
 
       tasks_priority:
         $if: 'is_repo_trusted'
@@ -98,6 +98,7 @@ tasks:
               TASK_ID: ${decision_task_id}
               TASKS_PRIORITY: ${tasks_priority}
               SCHEDULER_ID: ${scheduler_id}
+              BUILD_WORKER_TYPE: ${build_worker_type}
               MOBILE_HEAD_REPOSITORY: ${repository}
               MOBILE_HEAD_BRANCH: ${head_branch}
               MOBILE_HEAD_REV: ${head_rev}
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -180,7 +180,6 @@ tasks:
                   scopes:
                     - project:mobile:android-components:releng:beetmover:action:push-to-maven
                     - project:mobile:android-components:releng:beetmover:bucket:${beetmover_bucket}
-                    - queue:create-task:${tasks_priority}:aws-provisioner-v1/github-worker      # Still used by non build tasks
                     - queue:create-task:${tasks_priority}:scriptworker-prov-v1/${beetmover_worker_type}
                   payload:
                     env:
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -158,84 +158,86 @@ tasks:
                 then: ''
                 else: '--staging'
 
-              nightly_or_release_definition:
-                scopes:
-                  $flatten:
-                    - project:mobile:android-components:releng:beetmover:action:push-to-maven
-                    - queue:create-task:${tasks_priority}:aws-provisioner-v1/github-worker      # Still used by non build tasks
-                    - $if: 'is_repo_trusted'
-                      then:
-                        - queue:create-task:${tasks_priority}:scriptworker-prov-v1/mobile-beetmover-v1
-                      else:
-                        - queue:create-task:${tasks_priority}:scriptworker-prov-v1/mobile-beetmover-dev
-                payload:
-                  env:
-                    MOBILE_TRIGGERED_BY: ${user}
-                  features:
-                    chainOfTrust: true
-                  artifacts:
-                    public/task-graph.json:
-                      type: file
-                      path: /build/android-components/task-graph.json
-                      expires: ${expires_in}
-                    public/actions.json:
-                      type: file
-                      path: /build/android-components/actions.json
-                      expires: ${expires_in}
-                    public/parameters.yml:
-                      type: file
-                      path: /build/android-components/parameters.yml
-                      expires: ${expires_in}
-            in:
-              - $if: 'tasks_for == "github-release"'
+              beetmover_worker_type:
+                $if: 'is_repo_trusted'
+                then: mobile-beetmover-v1
+                else: mobile-beetmover-dev
+
+              beetmover_bucket:
+                $if: 'tasks_for == "github-release"'
                 then:
-                  $let:
-                    tag: ${event.release.tag_name}
-                  in:
+                  $if: 'is_repo_trusted'
+                  then: maven-production
+                  else: maven-staging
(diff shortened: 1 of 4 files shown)


JohanLorenzo:
Tested as:
* PR (see badge) 
* Regular push: https://tools.taskcluster.net/groups/SqD49BBRSyuh9_4n25Mv8Q
* Staging release: https://tools.taskcluster.net/groups/eE6rgCyfTNOrtfl9v4jmOQ
* Staging snapshot: https://tools.taskcluster.net/groups/PSWZ_2HoSrCuS4TeM0E-XQ

Errors in graphs looks intermittent.

automation/taskcluster/decision_task_publish.py
```diff
@@ -128,11 +128,10 @@ def generate_raw_task(name, description, command_to_run):
     checkout = ("export TERM=dumb && git fetch {} {} && "
                 "git config advice.detachedHead false && "
                 "git checkout {} && ".format(REPO_URL, BRANCH, HEAD_REV))
-    return taskcluster.slugId(), BUILDER.raw_task(
+    return taskcluster.slugId(), BUILDER.craft_build_like_task(
```
MihaiTabara: 👍

automation/taskcluster/lib/tasks.py
```diff
@@ -8,143 +8,110 @@
...
-            "metadata": {
-                "name": name,
-                "description": description,
-                "owner": self.owner,
-                "source": self.source
-            }
+        payload = {
+            "features": features,
+            "maxRunTime": 7200,
+            "image": "mozillamobile/android-components:1.15",
```
MihaiTabara: Unrelated with this PR, I was wondering, we might want to revisit this at some point and rather read this Docker image from a config file or something. So that each time we bump it, we no longer have to `grep` and replace everywhere.

ncalexan:
> @ncalexan <snip>: What do you think about it?

I have no opinion.  My expectation is that all consumers will consider these strings opaque, or will consult a releng oracle to determine the appropriate strings :)

.taskcluster.yml
```diff
@@ -157,84 +158,86 @@ tasks:
...
-                    public/parameters.yml:
-                      type: file
-                      path: /build/android-components/parameters.yml
-                      expires: ${expires_in}
-            in:
-              - $if: 'tasks_for == "github-release"'
+              beetmover_worker_type:
+                $if: 'is_repo_trusted'
+                then: mobile-beetmover-v1
+                else: mobile-beetmover-dev
```
mitchhentges: Just curious: is "dev" in this context the same as "dep"?

.taskcluster.yml
```diff
@@ -157,84 +158,86 @@ tasks:
...
+                  $if: 'is_repo_trusted'
+                  then: maven-snapshot-production
+                  else: maven-snapshot-staging
+
+            in:
+              $let:
+                nightly_or_release_definition:
+                  scopes:
+                    $flatten:
+                      - project:mobile:android-components:releng:beetmover:action:push-to-maven
```
mitchhentges: I don't think you need to flatten this

automation/taskcluster/lib/tasks.py
```diff
@@ -8,143 +8,110 @@
...
+        self.beetmover_worker_type = beetmover_worker_type
         self.tasks_priority = tasks_priority
 
-    def raw_task(self, name, description, command, dependencies=[],
-                   artifacts={}, scopes=[], routes=[], features={},
-                   worker_type='github-worker'):
-        created = datetime.datetime.now()
-        expires = taskcluster.fromNow('1 year')
-        deadline = taskcluster.fromNow('1 day')
+    def craft_build_like_task(
```
mitchhentges: I'm having issues parsing this name, is it:
* Craft a build _like_ (e.g.: similar to) a task?
* Craft a task that's build-like (e.g.: very close to a build task)?

mitchhentges:
The name sounds as good as we can get with the 22-character limit :)

escapewindow:
`s,components,andrcmp,` or similar?

mitchhentges:
What value does the word "components" add to the name? Would we lose any context if it was `mobile-X-b-android`? This applies to all mobile projects, right?

Or, is there a worker for each repository? Will there be a separate, dedicated worker for `reference-browser`, and one for `fenix`?

escapewindow:
`s,components,andrcmp,` or similar?

> What value does the word "components" add to the name? Would we lose any context if it was `mobile-X-b-android`? This applies to all mobile android projects, right?
> 
> Or, is there a worker for each repository? Will there be a separate, dedicated worker for `reference-browser`, and one for `fenix`?

Good questions. I don't know definitive answers here. It's quite possible the `mobile-` prefix is sufficient, and `components` is extraneous.

JohanLorenzo:
Thank you Mitch for filing [bug 1520579](https://bugzilla.mozilla.org/show_bug.cgi?id=1520579). We will eventually have 40ish characters for worker types. While this gets implemented, `andrcmp` sounds like a good compromise. 
I think `android` is too generic, because this worker type is created for android-components only, and fenix or focus are both android project technically. So people may confuse them. 
The reason I think we should not share the same worker types between focus (for example) and android-components is: we currently define different hardware. android-components has some pretty big machines (since https://github.com/mozilla-mobile/android-components/pull/1638), but focus doesn't. Maybe we should use the same hardware everywhere. In which case `android` is great.

mitchhentges:
Ah, I didn't realize we had different workers per repository - it absolutely makes sense, `android-components` is a chungus compared to `fenix`. Big :+1: :)

.taskcluster.yml
JohanLorenzo: Fixed

automation/taskcluster/lib/tasks.py
JohanLorenzo: Good call. That's the latter. I then change the name to `craft_build_ish_task()`

JohanLorenzo:
Tested as:
* PR (see badge) 
* Regular push: https://tools.taskcluster.net/groups/XVNYmSVGRpm7ae4yr9Y6pA
* Staging release: https://tools.taskcluster.net/groups/OF358ZJQR-am3J7D2hrKlg
* Staging snapshot: https://tools.taskcluster.net/groups/S9X0SGZ8TvSvDv46kSnJCQ

Errors in graphs look intermittent.

PR #1731, opened 2019-01-15
Closes #1545 - Add support for menu dividers.

### LINKED ISSUE #1545: Menu: Support for dividers

Amejia481:


components/browser/menu/src/main/java/mozilla/components/browser/menu/item/BrowserMenuDivider.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import android.view.View
+import mozilla.components.browser.menu.BrowserMenu
+import mozilla.components.browser.menu.BrowserMenuItem
+import mozilla.components.browser.menu.R
+
+/**
+ * A browser menu item to display a horizontal divider.
+ */
+class BrowserMenuDivider : BrowserMenuItem {
```
pocmo: Looking at the other classes I wonder if this should be `BrowserMenuDividerItem`?

components/browser/menu/src/main/res/layout/mozac_browser_menu_item_divider.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?><!-- This Source Code Form is subject to the terms of the Mozilla Public
```
pocmo: nit: (Probably the auto-formatter's fault): There's a line break missing here.

components/browser/menu/src/main/res/values/style.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<resources>
+    <style name="mozacBrowserMenuItemDividerHorizontal">
```
pocmo: My gut feeling is that this should start with an uppercase character (`Mozac`). Is there a style guide for styles?

components/browser/menu/src/main/java/mozilla/components/browser/menu/item/BrowserMenuDivider.kt
Amejia481: I was a bit confused with this one because there were more classes that are not adding the item prefix than the ones that have it, 3 vs 2.  I thought they didn't because they already are in the `item` package.

Should I add it too?

image

pocmo: Alright. In this case let's leave it as-is. :)

components/browser/menu/src/main/res/values/style.xml
Amejia481: You're right, I'll fix it.

> Is there a style guide for styles?
No that I'm aware of, but I was checking the Android ones and they follow the upper case pattern.

components/browser/menu/src/main/java/mozilla/components/browser/menu/item/BrowserMenuDivider.kt
Amejia481: Should I rename the others ?

components/browser/menu/src/main/res/values/style.xml
pocmo: Looking at this it seems like we want `Mozac.Browser.Menu.Item.Divider.Horizontal`? :)

Amejia481: Thanks for keep asking. I learned something new :)

I took for granted that it wasn't possible to do that. Without having something like this:
image

Because the way that styles work, it's that every time that you have a dot in a new style name you are inheriting from it. (For this reason, every style that you are referring with a dot should it should exist before using it)

image

To have a good answer for you, I kept searching and I found [themes_base.xml,](https://github.com/aosp-mirror/platform_frameworks_support/blob/master/v7/appcompat/res/values/styles_base.xml) the namespace for *Base.AppCompat.**.

When they want to define a new style without having a style for every dot they use `parent=""`
image

Now we can have :) 

image

A good resource about [Style and themes.](https://guides.codepath.com/android/Styles-and-Themes)

components/browser/menu/src/main/java/mozilla/components/browser/menu/item/BrowserMenuDivider.kt
Amejia481: Oops that could be an Api break, now we have to live with that

PR #1732, opened 2019-01-15
Add changelog entries for 0.32.1 and 0.32.2 [ci skip]

csadilek:


docs/changelog.md
```diff
@@ -472,6 +472,37 @@ permalink: /changelog/
...
+  * Android (SDK: **28** 🔺, Support Libraries: **28.0.0** 🔺)
+  * Kotlin (Stdlib: 1.3.0, Coroutines: 1.0.1)
+  * GeckoView (Nightly: **65.0.20181116100120** 🔺, Beta: 64.0.20181022150107, Release: 63.0.20181018182531)
+  * Mozilla App Services (FxA: **0.10.0** 🔺, Sync Logins: **0.10.0** 🔺, Places: **0.10.0** 🔺)
+
+* **ui-autocomplete**
+  * Fixed problem handling backspaces as described in [Issue 1489](https://github.com/mozilla-mobile/android-components/issues/1489)
+
+* **browser-search**  
+  * Updated search codes (see [Issue 1563](https://github.com/mozilla-mobile/android-components/issues/1563) for details)
```
Amejia481: @csadilek  should this entry be removed from **0.37.0**? Because they are going to be duplicated. In **0.32.2** and **0.37.0**

docs/changelog.md
```diff
@@ -472,6 +472,37 @@ permalink: /changelog/
...
+
+# 0.32.1
```
Amejia481: Are we going to need a release for **0.32.1** because don't have one [in the release page.](https://github.com/mozilla-mobile/android-components/releases?after=v0.34.1)
image

Amejia481: Sorry I found it, it below of 0.36.1  🙃

docs/changelog.md
csadilek: @Amejia481 I think this is fine. 0.32.2 was released today (after 0.37.0). So, in my opinion, they should both have this entry. In other words, why alter the changelog of 0.37 just because we had to do a point release of an old branch...0.33 etc. don't have this. So, worth pointing out that 0.37 has it.

PR #1733, opened 2019-01-15
Issue #1604: Add BrowserToolbarBottomBehavior to be used when placing a BrowserToolbar at the bottom of a CoordinatorLayout.

### LINKED ISSUE #1604: Support scrollable bottom toolbar
Since #1205 we support scrolling the toolbar away automatically.

This works fine with the toolbar at the top. But this doesn't work when the toolbar is at the bottom.

pocmo:
Closes #1604.


pocmo:
>  Should this pull request showcase it in the sample browser app?

I have those patches locally and did that for testing. Maybe! I'll definitely want to do that in the reference browser so that we get this tested in the wild. :)

PR #1734, opened 2019-01-15
Closes #1714: Support enabling tracking protection per session type

### LINKED ISSUE #1714: Support enable tracking protection per session type
Fennec and iOS have more tracking protection enabled states than we currently support. There tracking protection can be enabled for all, private, and for regular sessions separately (the latter is only supported by Firefox for iOS). 

One way to solve this is with an additional setting (`TrackingProtectionEnabled`):

- Off -> `TrackingProtectionPolicy.NONE`, `TrackingProtectionEnabled.NONE`

- On for private sessions -> `TrackingProtectionPolicy.ALL`, `TrackingProtectionEnabled.PRIVATE`

- On for regular sessions -> `TrackingProtectionPolicy.ALL`, `TrackingProtectionEnabled.REGULAR`

- On for all sessions -> `TrackingProtectionPolicy.ALL`, `TrackingProtectionEnabled.ALL`

csadilek:
I didn't want to implement yet another setting for this, as it would have introduced some awkward states (see https://github.com/mozilla-mobile/android-components/issues/1714#issuecomment-454475044)

Instead, I opted for merging this functionality in one setting and creating a mini fluent API that prevents these states in the first place.

So, these examples are unchanged and work as before (by default we apply to both private and non-private sessions):
```Kotlin
DefaultSettings(trackingProtectionPolicy = TrackingProtectionPolicy.all())
DefaultSettings(trackingProtectionPolicy = TrackingProtectionPolicy.select(TrackingProtectionPolicy.AD))
```

Examples of new configurations:
```Kotlin
DefaultSettings(trackingProtectionPolicy = TrackingProtectionPolicy
    .all()
    .forPrivateSessionsOnly()
)

DefaultSettings(trackingProtectionPolicy = TrackingProtectionPolicy
    .select(TrackingProtectionPolicy.SOCIAL)
    .forRegularSessionsOnly()
)
```

This prevents the awkward (ALL + NONE) configuration. In addition calls like this on `none()` aren't possible / won't compile: 
DefaultSettings(trackingProtectionPolicy = TrackingProtectionPolicy).none()~.forPrivateSessionsOnly()~


pocmo:
Nice!

PR #1735, opened 2019-01-16
[glean] Remove cookies from HTTP requests when uploading pings

Dexterp37:
This makes sure no extra data gets to our server when performing HTTP uploads for glean pings. We do so
by removing all cookies from the global store that are related to the ingestion endpoint.


Dexterp37:
> This solution makes me feel a bit icky, but having seen the discussion that this is the best of the available options, LGTM.

Yes, this feels the same to me :(

Dexterp37:
I made this WIP again, as we need to make sure that we're not nuking cookies "mozilla.org", "because mozilla.org cookies bubble up to incoming.telemetry.mozilla.org"

georgf:
Have we manually sanity-tested this on a real device?
A simple test-case that might work is:
- Navigate to "mozilla.org"
- Verify some cookies are set (specifically i'd expect GA).
- Force a Glean ping upload.
- Navigate again to "mozilla.org".
- Verify the cookies on the first request are set (specifically the same GA id).

PR #1736, opened 2019-01-16
[glean] Pin the Miniconda3 version to 4.5.11

Dexterp37:
The latest version, 4.5.12, is breaking ssl for glean on Windows.


mdboom:
I *think* to really get the benefits of this, users might need to `git clean -fxd` so that this pinned version is installed rather than whatever they may already have.  (Just leaving that to help others finding this bug).

PR #1737, opened 2019-01-16
[glean]  Implement the `duration` field

Dexterp37:
This field counts how much time user spent on foreground since the application started. Moreover,
this changes the `sessions` field to have an `application` lifetime rather than `ping`. This
additionally tries to use the `LifecycleRegistry` to interact with `GleanLifecycleObserver` and trigger
background pings.

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/GleanLifecycleObserver.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/GleanLifecycleObserver.kt
@@ -20,6 +20,9 @@ internal class GleanLifecycleObserver : LifecycleObserver {
      */
     @OnLifecycleEvent(Lifecycle.Event.ON_STOP)
     fun onEnterBackground() {
+        // We're going to background, so store how much time we spent
+        // on foreground.
+        Baseline.duration.stopAndSum()
         Glean.handleEvent(Glean.PingEvent.Background)
     }
 
@@ -34,5 +37,9 @@ internal class GleanLifecycleObserver : LifecycleObserver {
     @OnLifecycleEvent(Lifecycle.Event.ON_RESUME)
     fun onEnterForeground() {
         Baseline.sessions.add()
+        // Note that this is sending the length of the last foreground session
+        // because it belongs to the baseline ping and that ping is sent every
+        // time the app goes to background.
+        Baseline.duration.start()
     }
 }
--- a/components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
+++ b/components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
@@ -4,6 +4,9 @@
 
 package mozilla.components.service.glean
 
+import android.arch.lifecycle.Lifecycle
+import android.arch.lifecycle.LifecycleOwner
+import android.arch.lifecycle.LifecycleRegistry
 import android.content.Context
 import androidx.test.core.app.ApplicationProvider
 import kotlinx.coroutines.ExperimentalCoroutinesApi
@@ -13,6 +16,7 @@ import mozilla.components.service.glean.storages.EventsStorageEngine
 import mozilla.components.service.glean.storages.ExperimentsStorageEngine
 import mozilla.components.service.glean.storages.StringsStorageEngine
 import mozilla.components.service.glean.metrics.Baseline
+import mozilla.components.service.glean.scheduler.GleanLifecycleObserver
 import okhttp3.mockwebserver.MockResponse
 import okhttp3.mockwebserver.MockWebServer
 import org.json.JSONObject
@@ -27,6 +31,7 @@ import org.junit.Test
 import org.junit.runner.RunWith
 import org.mockito.ArgumentMatchers.anyString
 import org.mockito.Mockito.doThrow
+import org.mockito.Mockito.mock
 import org.mockito.Mockito.spy
 import org.robolectric.RobolectricTestRunner
 import java.io.File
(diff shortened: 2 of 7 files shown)


components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
```diff
@@ -26,8 +30,7 @@ import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.mockito.ArgumentMatchers.anyString
-import org.mockito.Mockito.doThrow
-import org.mockito.Mockito.spy
+import org.mockito.Mockito.*
```
travis79: I'm not sure importing `*` is done anywhere else in the lib, can you just import the modules you need?  It's probably okay since it's in a test but the lib source files get hit for this by detekt.

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
```diff
@@ -224,14 +235,20 @@ class GleanTest {
...
-            val baselineCounterMetrics = baselineJson.getJSONObject("metrics")!!.getJSONObject("counter")!!
+            val baselineCounterMetrics = baselineMetricsObject.getJSONObject("counter")!!
             assertEquals(expectedBaselineCounterMetrics.size, baselineCounterMetrics.length())
             for (metric in expectedBaselineCounterMetrics) {
                 assertNotNull(baselineCounterMetrics.get(metric))
             }
+
+            val baselineTimespanMetrics = baselineMetricsObject.getJSONObject("timespan")!!
+            assertEquals(1, baselineTimespanMetrics.length())
+            assertNotNull(baselineTimespanMetrics.get("duration"))
```
travis79: Can we add a little manual delay, a 2-3 seconds maybe, between the ON_START and ON_STOP above and then validate that the recorded duration is valid?  This adds a little time to our test overhead but I think it's good to validate that it's recording a reasonable value.

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
Dexterp37: Yeah, I didn't iron out this issue, as the PR was marked as WIP. Good catch though, I've fixed this.

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
Dexterp37: In general, we don't really want `sleep` in tests. I tried to fake the system clock using the `ShadowSystemClock`, but for some reason it's not updating the value returned by `elapsedRealtimeNanos`, which is used under the hood. I'd rather leave this as it is, since the timespan mechanism is being tested elsewhere.

Dexterp37:
@mdboom @travis79 this is ready for another look!

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
travis79: Okay, as long as it's being tested somewhere I'm good with that

Dexterp37:
@mdboom @travis79 I just had a conversation with Frank and @georgf a few minutes ago. We settled on making the `duration` field measure the length of the last foreground session (and just that!). I've dropped the changes to the `sessions` field, as I'll need to remove it in a separate PR. Can you give this PR a last look?

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/GleanLifecycleObserver.kt
```diff
@@ -34,5 +37,6 @@ internal class GleanLifecycleObserver : LifecycleObserver {
     @OnLifecycleEvent(Lifecycle.Event.ON_RESUME)
     fun onEnterForeground() {
         Baseline.sessions.add()
+        Baseline.duration.start()
```
mdboom: I think this works to send the duration of the last foreground session *only* because we send the baseline ping on going to background and *only* because ping data collection happens synchronously.  Just wondering if we should document that dependency somehow / somewhere.

Dexterp37: Good point. I've added a comment above this line.

pocmo:
@Dexterp37 You are not using matching/synchronized/paired events here: Either you want to use resume/pause or start/stop here. With the difference being:
* start/stop: App is visible/invisible
* resume/pause: App is in the foreground, the user can interact with it.

An example for an app can going through those methods/state

* User starts the app
  * onCreate()
  * onStart()
  * onResume()
* User pulls down the notification shade (-> The app is still visible, but the user is interacting with the notification drawn on top)
  * onPause()
* User closes the notifaction shade and is back in the app
  * onResume()
* User leaves the app to go to the home screen:
  * onPause()
  * onStop()
* User goes back to the app:
  * onStart()
 * onResume()
* User presses back to close the app:
  * onPause()
  * onStop()
  * onDestroy()

See:
 * https://developer.android.com/guide/components/activities/activity-lifecycle#onstart
 * https://developer.android.com/guide/components/activities/state-changes

Also see: https://developer.android.com/reference/android/app/Activity#ActivityLifecycle

Dexterp37:
> 
> 
> @Dexterp37 You are not using matching/synchronized/paired events here: Either you want to use resume/pause or start/stop here. With the difference being:
> 
>     * start/stop: App is visible/invisible
> 
>     * resume/pause: App is in the foreground, the user can interact with it.

Gosh, thanks for spotting this. I filed [bug 1491345](https://bugzilla.mozilla.org/show_bug.cgi?id=1520754) for fixing it.

PR #1741, opened 2019-01-17
[glean] Remove the "sessions" field from the baseline ping

Dexterp37:
This removes the "sessions" field, since it was not adding any useful insight to the data we already
collect. This is due to the fact that the container ping is generated and sent every time we go to
background, so the value of this field would always be 1.

### DIFF
--- a/components/service/glean/docs/baseline.md
+++ b/components/service/glean/docs/baseline.md
@@ -7,7 +7,6 @@ the following fields:
 
 | Field name | Type | Description |
 |---|---|---|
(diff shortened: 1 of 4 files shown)


pocmo:
I assume there can be situations where generating+uploading didn't happen (Crashes, Out of battery, ..). But that may still not justify having the field. :)

Dexterp37:
> 
> 
> We might want to drop a comment or note in the spec document to reflect the discussion and change but this LGTM.

Sure, already done!

PR #1747, opened 2019-01-17
Closes #1544: Added support for image and text in browser menus

### LINKED ISSUE #1544: Menu: Add icons to items
Currently `SimpleBrowserMenuItem` only supports text. For some future apps we will need the ability to show an icon next to the text.

Amejia481:
To avoid API break with consumers of `SimpleBrowserMenuItem`. I added `BrowserMenuImage`.

### Demo
screenshot_20190117-103134

### DIFF
--- a/components/browser/menu/src/main/java/mozilla/components/browser/menu/item/BrowserMenuImageText.kt
+++ b/components/browser/menu/src/main/java/mozilla/components/browser/menu/item/BrowserMenuImageText.kt
@@ -1,0 +1,71 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.menu.item
+
+import android.support.annotation.ColorRes
+import android.support.annotation.DrawableRes
+import android.support.v4.content.ContextCompat
+import android.support.v7.widget.AppCompatImageView
+import android.view.View
+import android.widget.TextView
+import mozilla.components.browser.menu.BrowserMenu
+import mozilla.components.browser.menu.BrowserMenuItem
+import mozilla.components.browser.menu.R
+
+private const val NO_ID = -1
+
+/**
+ * A menu item for displaying text with an image icon.
+ *
+ * @param label The visible label of this menu item.
+ * @param imageResource ID of a drawable resource to be shown as icon.
+ * @param contentDescription The image's content description, used for accessibility support.
+ * @param iconTintColorResource Optional ID of color resource to tint the icon.
+ * @param listener Callback to be invoked when this menu item is clicked.
+ */
+class BrowserMenuImageText(
+    private val label: String,
+    @DrawableRes
+    private val imageResource: Int,
+    private val contentDescription: String,
+    @ColorRes
+    private val iconTintColorResource: Int = NO_ID,
+    private val listener: () -> Unit = {}
+) : BrowserMenuItem {
+
+    override var visible: () -> Boolean = { true }
+
+    override fun getLayoutResource() = R.layout.mozac_browser_menu_item_image_text
+
+    override fun bind(menu: BrowserMenu, view: View) {
+
+        bindText(view)
+
+        bindImage(view)
+
+        view.setOnClickListener {
+            listener.invoke()
+            menu.dismiss()
+        }
+    }
+
+    private fun bindText(view: View) {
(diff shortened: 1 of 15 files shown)


pocmo:
> To avoid API break with consumers of SimpleBrowserMenuItem. I added BrowserMenuImage.

An optional image wasn't possible? With `BrowserMenuImageText` I wonder whether `SimpleBrowserMenuItem` should be `BrowserMenuText` now though! We could keep `SimpleBrowserMenuItem`, deprecate it and just make it extend `BrowserMenuText`.

Amejia481:
> An optional image wasn't possible?

I tried it but It could cause API break in some cases:

Imagine that we add the optional parameter`imageResource`:

```kotlin
class SimpleBrowserMenuItem(
    private val label: String,
    private val imageResource: Int = -1,
    private val listener: () -> Unit
) : BrowserMenuItem {
    override var visible: () -> Boolean = { true }

    override fun getLayoutResource() = R.layout.mozac_browser_menu_item_simple

    override fun bind(menu: BrowserMenu, view: View) {
        (view as TextView).text = label
        view.setOnClickListener {
            listener.invoke()
            menu.dismiss()
        }
    }
}
```

If consumers constructed it using the lambda parameter outside of the parenthesis. No problem, it's OK it compiles ✅  _This is probably what most of the people are doing_.
```kotlin
    val item = SimpleBrowserMenuItem("Settings") {
        // Do something
    }
```

But if there is some code that is passing the lambda as a parameter will break 💣 ,unless they have named parameters:
image

> I wonder whether SimpleBrowserMenuItem should be BrowserMenuText now though! We could keep SimpleBrowserMenuItem, deprecate it and just make it extend BrowserMenuText.

I like the idea of having `BrowserMenuText` and deprecate `SimpleBrowserMenuItem` 
`BrowserMenuText` is a much nicer name.

Amejia481:
> Does this item follow the material guidelines?

Oops! 🤦‍♂️ ...Fixing it!

Amejia481:
Alt Text

Amejia481:
@pocmo  I changed it locally, the below image it following the spec but the other items on the menu are not following the spec image above.

**Spec**
image

**Screen Shot**
image


**They are items following this spec, I guess because they didn't include an icon.**

image

Amejia481:
I updated the PR to refactor all common text styles in browser text menus into a style item.

Now when we want to change something we have one point of change.

```xml
<style name="Mozac.Browser.Menu.Item.Text" parent="@android:style/TextAppearance.Material.Menu">
        <item name="android:background">?android:attr/selectableItemBackground</item>
        <item name="android:textSize">@dimen/mozac_browser_menu_item_text_size</item>
        <item name="android:ellipsize">end</item>
        <item name="android:lines">1</item>
        <item name="android:focusable">true</item>
        <item name="android:clickable">true</item>
    </style>

```

components/browser/menu/src/test/java/mozilla/components/browser/menu/item/BrowserMenuImageTextTest.kt
```diff
@@ -0,0 +1,93 @@
...
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.Mockito.mock
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
+class BrowserMenuImageTextTest {
+
+    private val context: Context = ApplicationProvider.getApplicationContext()
```
pocmo: Is it guaranteed that robolectric will use the same application context for all test methods? Otherwise I'd turn this into:

```Kotlin
    private val context: Context
        get() = ApplicationProvider.getApplicationContext()
```

components/browser/menu/src/main/java/mozilla/components/browser/menu/item/BrowserMenuImageText.kt
```diff
@@ -0,0 +1,71 @@
...
+    private val iconTintColorResource: Int = NO_ID,
+    private val listener: () -> Unit = {}
+) : BrowserMenuItem {
+
+    override var visible: () -> Boolean = { true }
+
+    override fun getLayoutResource() = R.layout.mozac_browser_menu_item_image_text
+
+    override fun bind(menu: BrowserMenu, view: View) {
+
```
pocmo: nit: empty line?

pocmo:
Nice!

Let's look at other items in follow-up issue(s) if needed.

Amejia481:
`+++ curl -s https://codecov.io/bash `
😞 https://codecov.io/bash  503 Service Temporarily Unavailable

PR #1755, opened 2019-01-18
Add the "app_build_id" and "app_display_version" to pings

Dexterp37:
This adds the application build id and display version to the ping_info section, using the glean internal metrics definition.


components/service/glean/build.gradle
```diff
@@ -14,7 +14,7 @@ apply plugin: 'kotlin-android'
  * created during unit testing.
  * This uses a specific version of the schema identified by a git commit hash.
  */
-String GLEAN_PING_SCHEMA_GIT_HASH = "6041aa6"
+String GLEAN_PING_SCHEMA_GIT_HASH = "6041aa6" // TO UPDATE BEFORE MERGE
```
Dexterp37: This schema hash will need to be updated once mozilla-services/mozilla-pipeline-schemas#245 is merged.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -56,11 +56,15 @@ open class GleanInternalAPI {
...
      * @param applicationContext [Context] to access application features, such
      * as shared preferences
+     * @param appBuild The build id of the application.
+     * @param appDisplayVersion The user visible version number of the application.
      * @param configuration A Glean [Configuration] object with global settings.
      * @raises Exception if called more than once
      */
     fun initialize(
         applicationContext: Context,
+        appBuild: String,
```
Dexterp37: @georgf , we're requiring the build id and display version to be passed to the `Glean.initialize` function. What's your take on this?

pocmo: I wonder what I'd pass in here as a developer though. I'd guess the `versionCode` of the app in most cases.

Dexterp37: @pocmo is it always like that for Android? Do you know how Focus or similar products work wrt this? If all that's used is `versionCode` and `versionNumber`, I think we can drop the parameters here and just always set them internally. That would be better, IMHO, so that no change is required by consumers :)

pocmo: The version code is what we report in the current telemetry library as "build id" (in the upload path). It's the same ID google play uses to identify a build. It's incremental and generated on taskcluster when building the application.

components/service/glean/docs/pings.md
```diff
@@ -16,7 +16,8 @@ The following fields are always included in the `ping_info` section, for every p
 | Field name | Type | Description |
 |---|---|---|
 | `ping_type` | String | The name of the ping type (e.g. "baseline", "metrics") |
-| `app_build` | String | TBD |
+| `app_build` | String | The build identifier generated by the CI system (e.g. "1234/A") |
+| `app_display_version` | String | The user visible version string (e.g. "1.0.3") |
```
mdboom:
```suggestion
| `app_display_version` | String | The user-visible version string (e.g. "1.0.3") |
```

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
```diff
@@ -45,10 +46,19 @@ class GleanTest {
     @get:Rule
     val fakeDispatchers = FakeDispatchersInTest()
 
+    private fun resetGlean(config: Configuration = Configuration()) {
```
mdboom: This is a nice refactor.  Eventually it probably will need be used by other test modules so maybe belongs in a set of test utilities?  (But that doesn't need to be done for this PR...)

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
Dexterp37: Thanks Sebastian! I've changed the code so that we automatically get the version name and code without requiring input at init.

Dexterp37:
@pocmo @mdboom  this is ready for another look!

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
Dexterp37: This is a good point, but let's wait for the testing API proposal to come up. This might be affected by that API as well.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -192,6 +192,9 @@ open class GleanInternalAPI {
             GleanInternalMetrics.clientId.generateAndSet()
         }
 
+        GleanInternalMetrics.appBuild.set(BuildConfig.VERSION_CODE.toString())
+        GleanInternalMetrics.appDisplayVersion.set(BuildConfig.VERSION_NAME)
```
pocmo: This is referencing the `BuildConfig` class of the library. I have no clue why this class has those static fields since they have no meaning in a library - but they are not pointing to the code/name of the application.

In the library you'll have to query the package manager at runtime. See here:
https://github.com/mozilla-mobile/android-components/blob/master/components/service/telemetry/src/main/java/org/mozilla/telemetry/util/ContextUtils.java#L19-L26

Dexterp37: Good point. I changed that, this is ready for another look, @pocmo . For some reason, in tests, `versionName` is always null. Do you know how can I set a proper test value with Robolectirc?

Dexterp37: I was able to mock that in the tests.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -220,6 +221,18 @@ open class GleanInternalAPI {
...
+        try {
+            val packageInfo = applicationContext.packageManager.getPackageInfo(
+                applicationContext.packageName, 0
+            )
+            GleanInternalMetrics.appBuild.set(packageInfo.versionCode.toString())
+            packageInfo.versionName?.let {
+                GleanInternalMetrics.appDisplayVersion.set(it)
+            }
+        } catch (e: PackageManager.NameNotFoundException) {
+            logger.error("Could not get own package info, unable to report build id and display version")
```
pocmo: nit: In similar situations in the app we throw `AssertionError` here since we never expect this to happen.

Dexterp37: Good point. Changing this.

PR #1756, opened 2019-01-18
Closes #1744: Support "UP" and "DOWN" orientations in browser-menu

### LINKED ISSUE #1744: Toolbar: Fix menu position for bottom toolbar
When placing browser toolbar at the bottom of the screen the menu still opens downwards (instead of upwards). Because there's not much space we end up with a single visible row that can be scrolled to reveal the rest of the menu.

pocmo:
This will add support for bottom toolbars with a menu.

This unblocks PR https://github.com/mozilla-mobile/reference-browser/pull/429.

### DIFF
--- a/components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenu.kt
+++ b/components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenu.kt
@@ -7,8 +7,10 @@ package mozilla.components.browser.menu
 import android.annotation.SuppressLint
 import android.graphics.Color
 import android.graphics.drawable.ColorDrawable
+import android.support.design.widget.CoordinatorLayout
 import android.support.v7.widget.LinearLayoutManager
 import android.support.v7.widget.RecyclerView
+import android.view.Gravity
 import android.view.LayoutInflater
 import android.view.View
 import android.view.WindowManager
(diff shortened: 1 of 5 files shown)


components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenu.kt
```diff
@@ -61,5 +64,38 @@ class BrowserMenu internal constructor(
...
     companion object {
         private const val MENU_ELEVATION_DP = 8
+
+        /**
+         * Determines the orientation to be used for a menu based on the positioning of the [parent] in the layout.
+         */
+        fun determineMenuOrientation(parent: View): BrowserMenu.Orientation {
+            val params = parent.layoutParams
+            return if (params is CoordinatorLayout.LayoutParams) {
+                if (params.gravity and Gravity.BOTTOM == Gravity.BOTTOM) {
```
Amejia481: This line is a bit confusing. 🤔  `Gravity.BOTTOM == Gravity.BOTTOM` Looks a bit strange, took me a bit to understand that `params.gravity` bit-field int and you're checking that `params.gravity` contains the value `BOTTOM`.

I was wonder if is there a way to enunciate better the intent.

Maybe something like:

```kotlin 

if (params.gravity in Gravity.BOTTOM) {
}
//Or
if (params.gravity.contains(Gravity.BOTTOM)) {
}
private operator fun Int.contains(mask: Int): Boolean {
    return (this and mask) == mask
}
```

pocmo: Would it be more obvious with `()`?

```Kotlin
if ((params.gravity and Gravity.BOTTOM) == Gravity.BOTTOM) {
...
```

Amejia481: It helps 👍 !

PR #1757, opened 2019-01-18
[glean] Correctly pair events in GleanLifecycleObserver

travis79:
Changes ON_RESUME to ON_START to correctly match with the ON_STOP event.  This should help prevent incorrectly recording metrics when doing things like interacting with the navigation shade, etc.


travis79:
@Dexterp37 simple one line change, ready for a look.

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/GleanLifecycleObserver.kt
```diff
@@ -34,7 +34,7 @@ internal class GleanLifecycleObserver : LifecycleObserver {
      *
      * https://developer.android.com/reference/android/app/Activity.html#onResume()
      */
-    @OnLifecycleEvent(Lifecycle.Event.ON_RESUME)
+    @OnLifecycleEvent(Lifecycle.Event.ON_START)
```
pocmo: The comment above mentions `ON_RESUME` :)

travis79: 🤦‍♂️ Updated and amended, thanks!

PR #1760, opened 2019-01-18
[glean] 1519402: Implement the ping_info seq metric

mdboom:
Rather than reusing a storage engine for this purpose, this is just doing the simplest thing possible.  It uses SharedPreferences directly to store a value keyed as `${pingName}_seq`.  This does mean we are updating a shared preference on each ping collection, but it's off the main thread already anyway, so hopefully not too problematic.


components/service/glean/src/main/java/mozilla/components/service/glean/ping/PingMaker.kt
```diff
@@ -44,6 +46,33 @@ internal class PingMaker(
...
+    /**
+     * Get the ping sequence number for a given ping. This is a
+     * monotonically-increasing value that is updated every time a particular ping
+     * type is sent.
+     *
+     * @param pingName The name of the ping
+     * @return sequence number
+     */
+    private fun getPingSeq(pingName: String): Int {
+        applicationContext.getSharedPreferences(
```
Dexterp37: Instead of opening the `sharedPreferences` every time, why don't we open them once lazily? Similarly to how it's done [here](https://github.com/mozilla-mobile/android-components/blob/25287691c8b78f7a911436a1de854864e0233e2c/components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt#L46). This way we don't pay the I/O for reading every time we need to increment the sequence numbers.

components/service/glean/src/test/java/mozilla/components/service/glean/ping/PingMakerTest.kt
```diff
@@ -100,13 +119,46 @@ class PingMakerTest {
...
+            StorageEngineManager(
+                storageEngines = mapOf(
+                    "engine1" to MockStorageEngine(JSONObject())
+                ),
+                applicationContext = applicationContext
+            ),
+            applicationContext
+        )
+
+        // Collect two pings, and make sure their seq numbers are sequential
```
Dexterp37: We should expand this test case or create other tests cases to make sure that:

- sequence numbers start from 0;
- that we gracefully handle corrupted data (e.g. strings instead of numbers when unpersisting sequence numbers);
- that two different ping types have difference sequences, both starting from 0.

mdboom:
@Dexterp37 : This has been updated to address your comments.

PR #1762, opened 2019-01-18
[WIP] [glean] Implement glean metrics ping

travis79:
Implementation of the 'metrics' ping and the JobScheduler for handling the scheduled upload of the ping.

### DIFF
--- a/components/service/glean/docs/pings.md
+++ b/components/service/glean/docs/pings.md
@@ -8,7 +8,7 @@ The following is a list of the pings along with the conditions that triggers the
 
 - [`baseline` ping](baseline.md)
 - [`events` ping](events.md)
-- `metrics` ping (TBD)
+- [`metrics` ping](metrics.md)
 
 ## The `ping_info` section
 The following fields are always included in the `ping_info` section, for every ping.
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
@@ -26,6 +26,7 @@ import mozilla.components.service.glean.scheduler.GleanLifecycleObserver
 import mozilla.components.service.glean.storages.ExperimentsStorageEngine
 import mozilla.components.service.glean.storages.StorageEngineManager
 import mozilla.components.service.glean.metrics.GleanBaseline
+import mozilla.components.service.glean.scheduler.MetricsPingScheduler
 import mozilla.components.service.glean.utils.StringUtils
 import mozilla.components.support.base.log.logger.Logger
 import java.io.File
@@ -50,6 +51,10 @@ open class GleanInternalAPI {
     // The application id detected by glean to be used as part of the submission
     // endpoint.
     internal lateinit var applicationId: String
+
+    // This object holds data related to persistent information about the metrics ping, such as
+    // the last time it was sent
+    internal lateinit var metricsPingScheduler: MetricsPingScheduler
 
     /**
      * Initialize glean.
@@ -233,6 +238,8 @@ open class GleanInternalAPI {
             throw AssertionError("Could not get own package info, aborting init")
         }
 
+        metricsPingScheduler = MetricsPingScheduler(applicationContext)
+
         // Set the OS type
         GleanBaseline.os.set("Android")
 
@@ -298,7 +305,12 @@ open class GleanInternalAPI {
                 sendPing("baseline", "baseline")
                 sendPing("events", "events")
             }
-            Glean.PingEvent.Default -> sendPing("metrics", "metrics")
+            Glean.PingEvent.Default -> {
+                if (metricsPingScheduler.canSendPing()) {
+                    sendPing(MetricsPingScheduler.STORE_NAME, MetricsPingScheduler.STORE_NAME)
+                    metricsPingScheduler.updateSentTimestamp()
+                }
+            }
         }
     }
 }
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/GleanLifecycleObserver.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/GleanLifecycleObserver.kt
@@ -24,6 +24,10 @@ internal class GleanLifecycleObserver : LifecycleObserver {
         // on foreground.
         GleanBaseline.duration.stopAndSum()
         Glean.handleEvent(Glean.PingEvent.Background)
+
+        // This will check the metrics ping schedule to determine whether it's time to send a new
+        // metrics ping or not
+        Glean.handleEvent(Glean.PingEvent.Default)
     }
 
     /**
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
@@ -1,0 +1,82 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.glean.scheduler
+
+import android.content.Context
+import android.content.SharedPreferences
+import mozilla.components.support.base.log.logger.Logger
+import java.util.concurrent.TimeUnit
+
+/**
+ * MetricsPingScheduler facilitates sending and scheduling of metrics pings when the application
+ * enters the background state (see [GleanLifecycleObserver]), but the ping is only sent if the
+ * scheduled time interval [PING_INTERVAL_HOURS] has elapsed.
+ */
+internal class MetricsPingScheduler(val applicationContext: Context) {
+    private val logger = Logger("glean/MetricsPingScheduler")
+    internal val sharedPreferences: SharedPreferences by lazy { getMetricsPrefs() }
+
+    companion object {
+        const val LAST_METRICS_PING_TIMESTAMP_KEY = "last_metrics_ping_timestamp"
+        const val STORE_NAME = "metrics"
+        const val PING_INTERVAL_HOURS = 24L
+    }
+
+    /**
+     * Check to see if a metrics ping can be sent.
+     *
+     * This will compare the current system time to the last ping time stored in [SharedPreferences]
+     * to determine whether [PING_INTERVAL_HOURS] has elapsed.
+     *
+     * @return true if the time interval has elapsed or if this is the first attempt to send a
+     *         metrics ping, otherwise false.
+     */
+    fun canSendPing(): Boolean {
+        val lastPingTime = try {
+            sharedPreferences.getLong(LAST_METRICS_PING_TIMESTAMP_KEY, 0L)
+        } catch (e: ClassCastException) {
+            // Intentionally doing nothing here, if LAST_METRICS_PING_TIMESTAMP_KEY's value is
+            // not a Long, then we will let lastPingTime be 0L for the purposes of comparison in
+            // order to allow the ping to be sent immediately
+            logger.error("MetricsPingScheduler last ping time stored was not a Long value")
+            0L
+        }
+
+        val currentTime = System.currentTimeMillis()
+
+        if (lastPingTime == 0L ||
+            currentTime - lastPingTime >= TimeUnit.HOURS.toMillis(PING_INTERVAL_HOURS)) {
+            return true
+        }
+
+        return false
+    }
+
+    /**
+     * Update the persisted timestamp when the metrics ping is sent.
+     *
+     * This is called after sending a metrics ping in order to timestamp when the last ping was
+     * sent in order to maintain the proper interval between pings. See [canSendPing].
+     *
+     * @param timestamp The timestamp to store, which defaults to [System.currentTimeMillis].
+     */
+    fun updateSentTimestamp(timestamp: Long = System.currentTimeMillis()) {
+        sharedPreferences.edit()?.putLong(LAST_METRICS_PING_TIMESTAMP_KEY, timestamp)?.apply()
+    }
+
+    /**
+     * Initializes the [SharedPreferences] where persisted metrics ping information is stored.
+     */
+    private fun getMetricsPrefs(): SharedPreferences {
+        return applicationContext.getSharedPreferences(this.javaClass.simpleName, Context.MODE_PRIVATE)
+    }
+
+    /**
+     * Test only function to clear stored data
+     */
+    internal fun clearSchedulerData() {
+        sharedPreferences.edit()?.clear()?.apply()
+    }
+}
--- a/components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
+++ b/components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
@@ -17,6 +17,7 @@ import mozilla.components.service.glean.storages.EventsStorageEngine
 import mozilla.components.service.glean.storages.ExperimentsStorageEngine
 import mozilla.components.service.glean.storages.StringsStorageEngine
 import mozilla.components.service.glean.scheduler.GleanLifecycleObserver
+import mozilla.components.service.glean.scheduler.MetricsPingScheduler
 import okhttp3.mockwebserver.MockResponse
 import okhttp3.mockwebserver.MockWebServer
 import org.json.JSONObject
@@ -235,6 +236,72 @@ class GleanTest {
             val baselineTimespanMetrics = baselineMetricsObject.getJSONObject("timespan")!!
             assertEquals(1, baselineTimespanMetrics.length())
             assertNotNull(baselineTimespanMetrics.get("glean.baseline.duration"))
+        } finally {
+            server.shutdown()
+            lifecycleRegistry.removeObserver(gleanLifecycleObserver)
+        }
+    }
+
+    @Test
+    fun `test sending of metrics ping`() {
+        val server = MockWebServer()
+        server.enqueue(MockResponse().setBody("OK"))
+
+        StringsStorageEngine.clearAllStores()
+        val metricToSend = StringMetricType(
+                disabled = false,
+                category = "telemetry",
+                lifetime = Lifetime.Ping,
+                name = "metrics_ping",
+                sendInPings = listOf("default")
+        )
+
+        val metricsPingScheduler = MetricsPingScheduler(getContextWithMockedInfo())
+        metricsPingScheduler.clearSchedulerData()
+
+        // Store a ping timestamp from 25 hours ago to compare to so that the ping will get sent
+        metricsPingScheduler.updateSentTimestamp(
+                System.currentTimeMillis() - TimeUnit.HOURS.toMillis(25))
+
+        resetGlean(getContextWithMockedInfo(), Glean.configuration.copy(
+                serverEndpoint = "http://" + server.hostName + ":" + server.port,
+                logPings = true
+        ))
+
+        // Fake calling the lifecycle observer.
+        val lifecycleRegistry = LifecycleRegistry(mock(LifecycleOwner::class.java))
+        val gleanLifecycleObserver = GleanLifecycleObserver()
+        lifecycleRegistry.addObserver(gleanLifecycleObserver)
+
+        try {
+            // Simulate the first foreground event after the application starts.
+            lifecycleRegistry.handleLifecycleEvent(Lifecycle.Event.ON_RESUME)
+
+            // Write a value to the metric to check later
+            metricToSend.set("Test metrics ping")
+
+            // Simulate going to background so that the metrics ping schedule will be checked
+            lifecycleRegistry.handleLifecycleEvent(Lifecycle.Event.ON_STOP)
+
+            val requests: MutableMap<String, String> = mutableMapOf()
+            for (i in 0..1) {
+                val request = server.takeRequest()
+                val docType = request.path.split("/")[3]
+                requests[docType] = request.body.readUtf8()
+            }
+
+            val metricsJson = JSONObject(requests["metrics"])
+            checkPingSchema(metricsJson)
+
+            // Check ping type
+            val pingInfo = metricsJson.getJSONObject("ping_info")!!
+            assertEquals(pingInfo["ping_type"], "metrics")
+
+            // Retrieve the string metric that was stored
+            val metricsObject = metricsJson.getJSONObject("metrics")!!
+            val stringMetrics = metricsObject.getJSONObject("string")!!
+
+            assertEquals(stringMetrics["telemetry.metrics_ping"], "Test metrics ping")
         } finally {
             server.shutdown()
             lifecycleRegistry.removeObserver(gleanLifecycleObserver)
--- a/components/service/glean/docs/metrics.md
+++ b/components/service/glean/docs/metrics.md
@@ -1,0 +1,10 @@
+# The `metrics` ping
+The `metrics` ping contains all of the metrics defined in `metrics.yaml` that don't specify a ping or 
+where `default` is specified in their [`send in pings`](https://mozilla.github.io/glean_parser/metrics-yaml.html#send-in-pings) property.  
+The ping shall also include the common ping data found in all pings.
+
+The `metrics` ping is automatically sent when the application is moved to the *background* provided
+that the scheduled time interval has elapsed, which is currently 24 hours.  This limits `metrics`
+pings to be sent not more than once per day. The schedule is only checked when the application
+enters the background state so pings may be sent after an interval longer than 24 hours but never
+shorter than 24 hours.
--- a/components/service/glean/src/test/java/mozilla/components/service/glean/scheduler/MetricsPingSchedulerTest.kt
+++ b/components/service/glean/src/test/java/mozilla/components/service/glean/scheduler/MetricsPingSchedulerTest.kt
@@ -1,0 +1,80 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.glean.scheduler
+
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertNotNull
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.RuntimeEnvironment
+import java.util.concurrent.TimeUnit
+
+@RunWith(RobolectricTestRunner::class)
+class MetricsPingSchedulerTest {
+    // NOTE: Using a "real" ApplicationContext here so that it will have
+    // a working SharedPreferences implementation
+    private val metricsPingScheduler = MetricsPingScheduler(RuntimeEnvironment.application.applicationContext)
+
+    @Before
+    fun setup() {
+        metricsPingScheduler.clearSchedulerData()
+    }
+
+    @Test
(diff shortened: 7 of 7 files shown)


travis79:
This is still a work in progress, I'm looking at service.telemetry to emulate the best way to test JobScheduler tasks.

samples/glean/metrics.yaml
```diff
@@ -34,9 +34,7 @@ test:
   test_string_list:
     type: string_list
     description: >
-      Testing StringList ping
-    send_in_pings:
-      - test_string_list
```
Dexterp37: Why did you change this? Doesn't this make the test metric available in all pings?

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * Called by the [JobScheduler] to asynchronously send the metrics ping and then calls
+     * [.jobFinished] to release the wake lock.
+     *
+     * @param params Parameters specifying info about this job
+     */
+    override fun onStartJob(params: JobParameters?): Boolean {
+        // handleEvent() should return a Job which we will need in order to interact with the
+        // async task uploading the ping
+        metricsPingJob = Glean.handleEvent(Glean.PingEvent.Default)
```
Dexterp37: Mh, so, I'm not sure how would this work: does this only work when the application is on and in the foreground? If it does work when the application is closed as well, what kind of metrics will it contain?

@pocmo have you ever used the job scheduler? What's your thought on this?

pocmo: Yes, we have been using `JobScheduler` in the old telemetry library. For all new code please use `WorkManager` from androix:
* https://developer.android.com/reference/androidx/work/WorkManager
* https://developer.android.com/topic/libraries/architecture/workmanager/

It uses `JobScheduler` and other mechanisms under the hood and avoids various issues.

Anyhow, regarding your question: At some point you schedule a job with some constraints (e.g. "I need network") (see `schedule` method here and JobScheduler will run your task in the background at some point (exceture `JobService`).

In our old telemetry we do the following: When we want to upload a ping, we generate the ping and write it to disk, we then schedule a background job (via JobScheduler) that will look at the stored pings on disk and try to upload them.

Is that helpful? Happy to write more :)

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
```diff
@@ -0,0 +1,99 @@
...
+class MetricsPingScheduler : JobService() {
+    private var metricsPingJob: Job? = null
+
+    companion object {
+        // Job ID is an arbitrary identifier for our job once it's been scheduled.  It is the key
+        // that allows us to cancel or make changes to the job.  If you start a new job with the
+        // same ID, it will replace the existing job with the new one.  In order to prevent any
+        // future ping types from having the same job ID, the ordinal value of the ping event
+        // enum is used for unique identification.
+        val JOB_ID = Glean.PingEvent.Default.ordinal
```
pocmo: Note that job IDs are application scoped. This may need some kind of negotiating with the app - or use very predictive IDs. Otherwise we may end up overriding jobs from the app or the other way around.

This is one of the pitfalls `WorkManager` handles for you :)

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
Dexterp37: Ah, this is very, very helpful, thanks!

I think we should scope this PR down and retain the same behaviour we have for other pings:

- check if some hours have passed since the last ping upload (stick to 24hrs for now)
- try to send the ping.

We have [bug 1508965](https://bugzilla.mozilla.org/show_bug.cgi?id=1508965) for using the `WorkManager`. We should probably do as the old telemetry was doing: serialize to disk and upload using the scheduler.

travis79: I’m still waiting on final decisions about when the ping should be scheduled so I started this in order to start getting feedback.  Work manager sounds like a better API so will go that route.  JobScheduler here should work regardless of whether the app is in the foreground and should persist over device reboots, sending on a period of no more than once a day.  It will contain everything in the “default” ping or any metric that doesn’t specify a ping, as the spec indicated (unless I misread what the metrics ping was to send).

Since our ping data is already serialized to disk in SharedPreferences this JobScheduler using the provided JobInfo just triggers the handler onStart() which builds the ping just like the baseline and events pings.  I’d be happy to go through my thinking on this to make sure I’m not overlooking anything on this.

samples/glean/metrics.yaml
travis79: I did this to get more data in the default or metrics ping for testing.

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
Dexterp37: > I’m still waiting on final decisions about when the ping should be scheduled so I started this in order to start getting feedback. Work manager sounds like a better API so will go that route. JobScheduler here should work regardless of whether the app is in the foreground and should persist over device reboots, sending on a period of no more than once a day. It will contain everything in the “default” ping or any metric that doesn’t specify a ping, as the spec indicated (unless I misread what the metrics ping was to send).

Let's stick to sending this ping every 24hrs, retaining the behaviour that Focus has for now.
Did you check how this behaves if the application is closed and there are ping lifetime metrics in memory?

> Since our ping data is already serialized to disk in SharedPreferences this JobScheduler using the provided JobInfo just triggers the handler onStart() which builds the ping just like the baseline and events pings. I’d be happy to go through my thinking on this to make sure I’m not overlooking anything on this.

Most of the metrics that will live in the `metrics` ping will have `Ping` lifetime, so they will not be persisted to disk. Moreover, relying on an internal implementation detail to power the scheduling logic is something we shouldn't probably do.

I think eventually persisting the whole ping to disk in order to enable background upload, retrying and different upload policies is the right way to move forward-

travis79: This does stick to the every 24 hours, just not at a specific time of day.  I went ahead with this per @georgf suggestion in our meeting last week while we wait on the data team to determine when the ping should be sent.

Unless I'm missing something, even the Application lifetime pings get written to SharedPreferences, just the lifetime indicates whether or not they get cleared out when they are collected, where User lifetime is not cleared.  So when the schedule is triggered by JobScheduler, the metrics ping data is collected, formatted, and then cleared if the lifetime requires it.  This isn't an internal implementation just relying on [JobScheduler](https://developer.android.com/reference/android/app/job/JobScheduler) API to trigger when the ping is collected and sent.  @Dexterp37 I'd be happy to go through this on Vidyo with you if that helps to explain what my thinking was on this, it seems like we have a bit of a disconnect on this.

travis79: I was overlooking the line where glean wasn't writing to SharedPreferences based on the ping lifetime, so this will need a little re-engineering per my conversation with @Dexterp37 just now.  Going to go the simple route and refactor to get glean sending metrics ping ASAP.

travis79:
I took a step back and went a different direction with this since the implementation of WorkManager will happen in another bug, so I pared this down to just the implementation of checking and uploading the ping no more than once every 24 hours.

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/GleanLifecycleObserver.kt
```diff
@@ -24,6 +24,10 @@ internal class GleanLifecycleObserver : LifecycleObserver {
         // on foreground.
         Baseline.duration.stopAndSum()
         Glean.handleEvent(Glean.PingEvent.Background)
+
+        // This will check the metrics ping schedule to determine whether it's time to send a new
```
Dexterp37: Let's use the comment as a way to explain why we're not simply adding this to `Glean.PingEvent.Background`

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -51,6 +52,10 @@ open class GleanInternalAPI {
     // endpoint.
     internal lateinit var applicationId: String
 
+    // This object holds data related to persistent information about the metrics ping, such as
+    // the last time it was sent, etc.
+    internal lateinit var metricsPing: MetricsPing
```
Dexterp37: nit: `metricsPingScheduler`?

components/service/glean/src/test/java/mozilla/components/service/glean/ping/MetricsPingTest.kt
```diff
@@ -0,0 +1,57 @@
...
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.RuntimeEnvironment
+import java.util.concurrent.TimeUnit
+
+@RunWith(RobolectricTestRunner::class)
+class MetricsPingTest {
+    // NOTE: Using a "real" ApplicationContext here so that it will have
+    // a working SharedPreferences implementation
+    var applicationContext: Context = RuntimeEnvironment.application.applicationContext
```
Dexterp37: nit: this doesn't need to be `var`

components/service/glean/src/test/java/mozilla/components/service/glean/ping/MetricsPingTest.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    @Test
+    fun `metrics ping has24HoursPassed() returns true if no value in shared prefs`() {
```
Dexterp37: nit: "if no timestamp was recorded in the shared prefs"

components/service/glean/src/main/java/mozilla/components/service/glean/ping/MetricsPing.kt
```diff
@@ -0,0 +1,40 @@
+package mozilla.components.service.glean.ping
+
+import android.content.Context
+import android.content.SharedPreferences
+import android.support.annotation.VisibleForTesting
+import java.util.concurrent.TimeUnit
+
+class MetricsPing(var applicationContext: Context) {
+    @VisibleForTesting
+    internal val metricsPreferences: SharedPreferences by lazy { getMetricsPrefs() }
```
Dexterp37: Shouldn't this be `SharedPreferences?`?

components/service/glean/src/test/java/mozilla/components/service/glean/ping/MetricsPingTest.kt
```diff
@@ -0,0 +1,57 @@
...
+        val metricsPing = MetricsPing(applicationContext)
+
+        // Clear the sharedPreferences on the MetricsPing so there is no last ping timestamp
+        metricsPing.metricsPreferences?.edit().clear().apply()
```
Dexterp37: Let's provide an `internal`, test only function in `MetricsPingScheduler` called `clearSchedulerData` that clears the prefs.

Since we want to clear the data between tests, let's add a [`@Before` rule](https://github.com/mozilla-mobile/android-components/blob/74f0ae6e7ab84f18222518aadb445c51ea0af153/components/service/glean/src/test/java/mozilla/components/service/glean/CounterMetricTypeTest.kt#L28-L39) that gets called before each test.

components/service/glean/src/test/java/mozilla/components/service/glean/ping/MetricsPingTest.kt
```diff
@@ -0,0 +1,57 @@
...
+
+        assertTrue(metricsPing.has24HoursPassed())
+    }
+
+    @Test
+    fun `metrics ping has24HoursPassed() returns false if 24 hours hasn't elapsed`() {
```
Dexterp37: nit: "hours haven't passed"?

components/service/glean/src/test/java/mozilla/components/service/glean/ping/MetricsPingTest.kt
```diff
@@ -0,0 +1,57 @@
...
+        val metricsPing = MetricsPing(applicationContext)
+
+        // Clear the sharedPreferences on the MetricsPing so there is no last ping timestamp
+        metricsPing.metricsPreferences?.edit().clear().apply()
+
+        // Store a ping timestamp to compare to
+        metricsPing.updatePingTimestamp()
```
Dexterp37: I think you need to control the value you're writing here, otherwise you might end up with a flaky test. Moreover, since writing is done off-the main thread, I'd also make sure the expected value is written before doing the check below. Another way to do this more reliably would be to mock the preferences as [done here](https://github.com/mozilla-mobile/android-components/blob/74f0ae6e7ab84f18222518aadb445c51ea0af153/components/service/glean/src/test/java/mozilla/components/service/glean/storages/GenericScalarStorageEngineTest.kt#L152-L158), so that you're sure the value you want sits in there, avoiding potentially intermittent tests in the future.

components/service/glean/src/test/java/mozilla/components/service/glean/ping/MetricsPingTest.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    @Test
+    fun `metrics ping has24HoursPassed() returns true if 24 hours has elapsed`() {
+        val metricsPing = MetricsPing(applicationContext)
+
+        // Clear the sharedPreferences on the MetricsPing so there is no last ping timestamp
+        metricsPing.metricsPreferences?.edit().clear().apply()
+
+        // Store a ping timestamp from 25 hours ago to compare to
+        metricsPing.updatePingTimestamp(System.currentTimeMillis() - TimeUnit.HOURS.toMillis(25))
```
Dexterp37: The same consideration from the previous tests applies, we need to make sure that the expected value is being saved.

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/GleanLifecycleObserver.kt
```diff
@@ -24,6 +24,10 @@ internal class GleanLifecycleObserver : LifecycleObserver {
...
+        // metrics ping or not
+        Glean.handleEvent(Glean.PingEvent.Default)
```
Dexterp37: Let's also add a test, similarly to how it's [done here](https://github.com/mozilla-mobile/android-components/blob/74f0ae6e7ab84f18222518aadb445c51ea0af153/components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt#L196-L206), to make sure this line is covered. This should also make sure that a `metrics` ping is generated and that the expected data is there. To do so, let's create a sample metric with Ping lifetime, record it, and make sure it's collected.

components/service/glean/src/main/java/mozilla/components/service/glean/ping/MetricsPing.kt
```diff
@@ -0,0 +1,40 @@
...
+
+    companion object {
+        const val LAST_METRICS_PING_TIMESTAMP_KEY = "last_metrics_ping_timestamp"
+        const val STORE_NAME = "metrics"
+        const val DOC_TYPE = "metrics"
+    }
+
+    @Suppress("MagicNumber")
+    fun has24HoursPassed(): Boolean {
+        val lastPingTime = metricsPreferences.getLong(LAST_METRICS_PING_TIMESTAMP_KEY, 0L)
```
Dexterp37: Let's also add a test to make sure that we don't break if `LAST_METRICS_PING_TIMESTAMP_KEY` contains garbage (e.g. a string)

components/service/glean/src/test/java/mozilla/components/service/glean/ping/MetricsPingTest.kt
travis79: According to the documentation for [.apply()](https://developer.android.com/reference/android/content/SharedPreferences.Editor#apply()) the in memory SharedPreferences are updated immediately, it's just the write to disk that is async so there should be no way for the following read to fail since the in memory preferences are updated synchronously.

components/service/glean/src/test/java/mozilla/components/service/glean/ping/MetricsPingTest.kt
travis79: I added some tests around storing of values to validate this.

components/service/glean/src/main/java/mozilla/components/service/glean/ping/MetricsPing.kt
travis79: I've changed this to be protected with a try/catch which will catch a ClassCastException which is what getLong throws if the value is not a Long.

components/service/glean/src/main/java/mozilla/components/service/glean/ping/MetricsPing.kt
```diff
@@ -0,0 +1,40 @@
...
+
+        if (lastPingTime == 0L ||
+            currentTime - lastPingTime >= TimeUnit.HOURS.toMillis(24)) {
+            return true
+        }
+
+        return false
+    }
+
+    fun updatePingTimestamp(timestamp: Long = System.currentTimeMillis()) {
```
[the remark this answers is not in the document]
travis79: Edit shouldn't be able to return a null, unless something went horribly wrong somewhere else.  I'm consulting the docs now to check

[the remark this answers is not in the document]
Dexterp37: Let's just assume that's a nullable and use `edit()?.`.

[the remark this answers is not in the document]
travis79: Docs do not indicate that null could be returned or any exceptions thrown at the point of .edit()

components/service/glean/docs/metrics_ping.md
```diff
@@ -0,0 +1,6 @@
+# The `metrics` ping
+The metrics ping contains all of the metrics defined in metrics.yaml that don't specify a ping or
```
Dexterp37:
```suggestion
The `metrics` ping contains all of the metrics defined in `metrics.yaml` that don't specify a ping or 
```

components/service/glean/docs/metrics_ping.md
```diff
@@ -0,0 +1,6 @@
+# The `metrics` ping
+The metrics ping contains all of the metrics defined in metrics.yaml that don't specify a ping or 
+where "default" is specified in their pings list.
```
Dexterp37:
```suggestion
where `default` is specified in their [`send_in_pings`](https://mozilla.github.io/glean_parser/metrics-yaml.html#send-in-pings) property.
```

components/service/glean/docs/metrics_ping.md
```diff
@@ -0,0 +1,6 @@
...
+
+The `metrics` ping is automatically sent when the application is moved to the *background* and it
```
Dexterp37: Please describe the full logic here, the fact that it's sent only once every 24 hours and how that works by checking the time when we get to background.

components/service/glean/docs/pings.md
```diff
@@ -8,7 +8,7 @@ The following is a list of the pings along with the conditions that triggers the
 
 - [`baseline` ping](baseline.md)
 - [`events` ping](events.md)
-- `metrics` ping (TBD)
+- [`metrics` ping](metrics_ping.md)
```
Dexterp37: Let's call the file `metrics.md` for consistency with the other doc files.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -233,6 +238,9 @@ open class GleanInternalAPI {
             throw AssertionError("Could not get own package info, aborting init")
         }
 
+        // Create a MetricsPing object
```
Dexterp37: nit: it's MetricsPingScheduler now, but the comment doesn't add much to the code, so you can remove it I guess :)

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
```diff
@@ -0,0 +1,85 @@
...
+
+package mozilla.components.service.glean.scheduler
+
+import android.content.Context
+import android.content.SharedPreferences
+import mozilla.components.support.base.log.logger.Logger
+import java.util.concurrent.TimeUnit
+
+/**
+ * MetricsPingScheduler facilitates sending and scheduling of metrics pings
```
Dexterp37: nit: when? how often?

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
```diff
@@ -0,0 +1,85 @@
...
+     * This will compare the current system time to the last ping time stored in [SharedPreferences]
+     * to determine whether [PING_INTERVAL_HOURS] has elapsed.
+     *
+     * @return true if the time interval has elapsed or if this is the first attempt to send a
+     *         metrics ping, otherwise false.
+     */
+    fun canSendPing(): Boolean {
+        var lastPingTime = 0L
+
+        try {
```
Dexterp37: Instead of using a `var`, this can be written as follows:

```
val lastPingTime = try {
 sharedPreferences.getLong(..:)
} catch (...) {
  logger...
  0L
}
```

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
```diff
@@ -0,0 +1,85 @@
...
+     */
+    private fun getMetricsPrefs(): SharedPreferences {
+        return applicationContext.getSharedPreferences(this.javaClass.simpleName, Context.MODE_PRIVATE)
+    }
+
+    /**
+     * Test only function to clear stored data
+     */
+    internal fun clearSchedulerData() {
+        sharedPreferences?.edit().clear().apply()
```
Dexterp37: Why is it assuming a nullable here? Should we do the same at the other callsites?

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
```diff
@@ -241,6 +242,67 @@ class GleanTest {
         }
     }
 
+    @Test
+    fun `test sending of metrics ping`() {
+        val server = MockWebServer()
+        server.enqueue(MockResponse().setBody("OK"))
+
+        EventsStorageEngine.clearAllStores()
```
Dexterp37: You don't need to clear the `EventsStorageEngine` but rather the `StringMetricStorageEngine` here :)

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
```diff
@@ -241,6 +242,67 @@ class GleanTest {
...
+            // Simulate the first foreground event after the application starts.
+            lifecycleRegistry.handleLifecycleEvent(Lifecycle.Event.ON_RESUME)
+
+            // Write a value to the metric to check later
+            metricToSend.set("Test metrics ping")
+
+            // Simulate going to background so that the metrics ping schedule will be checked
+            lifecycleRegistry.handleLifecycleEvent(Lifecycle.Event.ON_STOP)
+
+            val requests: MutableMap<String, String> = mutableMapOf()
```
Dexterp37: Mh, I think you should also be receiving a `baseline` ping here along with the `metrics` ping. In order to avoid intermittent failures in the future, can you please check the ping type before moving on?

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
```diff
@@ -241,6 +242,67 @@ class GleanTest {
...
+            for (i in 0..1) {
+                val request = server.takeRequest()
+                val docType = request.path.split("/")[3]
+                requests[docType] = request.body.readUtf8()
+            }
+
+            val eventsJson = JSONObject(requests["metrics"])
```
Dexterp37: nit: these are not events, so I guess `metricsJson`

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * Update the persisted timestamp when the metrics ping is sent.
+     *
+     * This is called after sending a metrics ping in order to timestamp when the last ping was
+     * sent in order to maintain the proper interval between pings. See [canSendPing].
+     *
+     * @param timestamp The timestamp to store, which defaults to [System.currentTimeMillis].
+     * @return Returns the time that was stored in [SharedPreferences] for comparision.
+     */
+    fun updateSentTimestamp(timestamp: Long = System.currentTimeMillis()): Long {
```
Dexterp37: Is this return value only for tests? I'm not sure we need this though. See my comments below.

components/service/glean/src/test/java/mozilla/components/service/glean/scheduler/MetricsPingTest.kt
```diff
@@ -0,0 +1,73 @@
...
+
+    @Test
+    fun `metrics ping canSendPing() returns true if no timestamp was recorded in shared prefs`() {
+        assertTrue(metricsPingScheduler.canSendPing())
+    }
+
+    @Test
+    fun `MetricsPingScheduler updatePingTimestamp() correctly stores values`() {
+        // First, try storing the default which is the current system time
+        val ts = metricsPingScheduler.updateSentTimestamp()
```
Dexterp37: Here we're testing the default value, which I think can be left out of the test coverage :) This way we could drop the return value from the function.

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
travis79: True dat!  There was also already a test that checked that the `defaults` ping was sent:
`fun 'test sending of default pings'()`

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
travis79: Yes this return value is only for tests, it was a simple way to test the default and have a way to compare it.

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
travis79: It is receiving both a `metrics` ping and a `baseline` ping, but since it's creating a `mutableMapOf()` the pings are maped by the doc_type which is why `requests["metrics"]` would be null if this were a baseline ping.  I'm okay with digging in and checking the ping type but I followed the function that checks the background pings and it doesn't verify the ping types either (aside from looking for things that would only be in those ping types).

Dexterp37: Ah, I see. Thanks for clarifying, makes sense.!

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /**
+     * Update the persisted timestamp when the metrics ping is sent.
+     *
+     * This is called after sending a metrics ping in order to timestamp when the last ping was
+     * sent in order to maintain the proper interval between pings. See [canSendPing].
+     *
+     * @param timestamp The timestamp to store, which defaults to [System.currentTimeMillis].
+     * @return Returns the time that was stored in [SharedPreferences] for comparision.
```
Dexterp37: nit: this doc line no longer applies and can be removed

PR #1764, opened 2019-01-19
Closes #1780 account manager 

### LINKED ISSUE #1780: Account manager
In https://github.com/mozilla-mobile/reference-browser/pull/425 I've prototyped a new state-machine driven account manager in order to support richer interactions with the accounts system. This bug is to track lifting of that work into a component.

grigoryk:
This is a first piece of functionality from https://github.com/mozilla-mobile/reference-browser/pull/425.

Still more tests to write for the manager. SyncManager port from https://github.com/mozilla-mobile/reference-browser/pull/425 is coming up next.

cc @vladikoff, @eoger as this is relevant to your interests.

### DIFF
--- a/components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FirefoxAccountsAuthFeature.kt
+++ b/components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FirefoxAccountsAuthFeature.kt
@@ -1,0 +1,65 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.feature.accounts
+
+import android.net.Uri
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import mozilla.components.concept.engine.EngineSession
+import mozilla.components.concept.engine.request.RequestInterceptor
+import mozilla.components.feature.tabs.TabsUseCases
+import mozilla.components.service.fxa.FxaAccountManager
+import mozilla.components.service.fxa.FxaException
+import kotlin.coroutines.CoroutineContext
+
+/**
+ * Ties together an account manager with a session manager/tabs implementation, facilitating an
+ * authentication flow.
+ */
+class FirefoxAccountsAuthFeature(
+    private val accountManager: FxaAccountManager,
+    private val tabsUseCases: TabsUseCases,
+    private val redirectUrl: String,
+    private val successPath: String,
+    private val coroutineContext: CoroutineContext = Dispatchers.Main
+) {
+    fun beginAuthentication() {
+        CoroutineScope(coroutineContext).launch {
+            val authUrl = try {
+                accountManager.beginAuthentication().await()
+            } catch (e: FxaException) {
+                // FIXME return a fallback URL provided by Config...
+                "https://accounts.firefox.com/signin"
+            }
+            // TODO
+            // We may fail to obtain an authentication URL, for example due to transient network errors.
+            // If that happens, open up a fallback URL in order to present some kind of a "no network"
+            // UI to the user.
+            // It's possible that the underlying problem will go away by the time the tab actually
+            // loads, resulting in a confusing experience.
+            tabsUseCases.addTab.invoke(authUrl)
+        }
+    }
+
+    val interceptor = object : RequestInterceptor {
+        override fun onLoadRequest(session: EngineSession, uri: String): RequestInterceptor.InterceptionResponse? {
+            if (!uri.startsWith(redirectUrl)) {
+                return null
+            }
+
+            val parsedUri = Uri.parse(uri)
+            val code = parsedUri.getQueryParameter("code") as String
+            val state = parsedUri.getQueryParameter("state") as String
+
+            // Notify the state machine about our success.
+            accountManager.finishAuthentication(code, state)
+
+            // TODO this can be simplified once https://github.com/mozilla/application-services/issues/305 lands
+            val successUrl = "${parsedUri.scheme}://${parsedUri.host}/$successPath"
+            return RequestInterceptor.InterceptionResponse.Url(successUrl)
+        }
+    }
+}
--- a/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FxaAccountManager.kt
+++ b/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FxaAccountManager.kt
@@ -1,0 +1,426 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.fxa
+
+import android.content.Context
+import android.support.annotation.VisibleForTesting
+import kotlinx.coroutines.CompletableDeferred
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Deferred
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.asCoroutineDispatcher
+import kotlinx.coroutines.async
+import java.util.concurrent.ConcurrentLinkedQueue
+import kotlin.coroutines.CoroutineContext
+
+import mozilla.components.support.base.log.Log
+import mozilla.components.support.base.observer.Observable
+import mozilla.components.support.base.observer.ObserverRegistry
+import java.io.Closeable
+import java.util.concurrent.Executors
+
+enum class AccountState {
+    Start,
+    NotAuthenticated,
+    AuthenticatedNoProfile,
+    AuthenticatedWithProfile
+}
+
+/**
+ * Base class for account manager state machine events.
+ * Events aren't a simple enum class because we might want to pass data along with some of the events.
+ */
+internal sealed class Event {
+    override fun toString(): String {
+        // For a better logcat experience.
+        return this.javaClass.simpleName
+    }
+
+    internal object Init : Event()
+
+    object AccountNotFound : Event()
+    object AccountRestored : Event()
+
+    object Authenticate : Event()
+    data class Authenticated(val code: String, val state: String) : Event() {
+        override fun toString(): String {
+            // data classes define their own toString, so we override it here as well as in the base
+            // class to avoid exposing 'code' and 'state' in logs.
+            return this.javaClass.simpleName
+        }
+    }
+    object FetchProfile : Event()
+    object FetchedProfile : Event()
+
+    object FailedToAuthenticate : Event()
+    object FailedToFetchProfile : Event()
+
+    object Logout : Event()
+}
+
+/**
+ * Propagated via [AccountObserver.onError] if we fail to load a locally stored account during
+ * initialization. No action is necessary from consumers.
+ * Account state has been re-initialized.
+ *
+ * @param cause Optional original cause of failure.
+ */
+class FailedToLoadAccountException(cause: Exception?) : Exception(cause)
+
+/**
+ * Observer interface which lets its consumers respond to authentication requests.
+ */
+private interface OAuthObserver {
+    /**
+     * Account manager is requesting for an OAUTH flow to begin.
+     * @param authUrl Starting point for the OAUTH flow.
+     */
+    fun onBeginOAuthFlow(authUrl: String)
+
+    /**
+     * Account manager encountered an error during authentication. Inspect [error] for details.
+     * @param error A specific FxA error encountered.
+     */
+    fun onError(error: FxaException)
+}
+
+/**
+ * Observer interface which lets its users monitor account state changes and major events.
+ */
+interface AccountObserver {
+    /**
+     * Account just got logged out.
+     */
+    fun onLoggedOut()
+
+    /**
+     * Account was successfully authenticated.
+     * @param account An authenticated instance of a [FirefoxAccountShaped].
+     */
+    fun onAuthenticated(account: FirefoxAccountShaped)
+
+    /**
+     * Account's profile is now available.
+     * @param profile A fresh version of account's [Profile].
+     */
+    fun onProfileUpdated(profile: Profile)
+
+    /**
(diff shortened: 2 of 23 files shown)


grigoryk:
My core assumption around investing time into this is that the complexity of our interactions with an account will grow over time. Already with the early SyncManager implementation in https://github.com/mozilla-mobile/reference-browser/pull/425, _some_ kind of higher level encapsulation was necessary to make things sane.

Prior art shows that this stuff tends to grow long tentacles all over the application. Managing an account in Fennec - and understanding that code well enough to be able to modify it confidently - was quite a troublesome endeavour, with a lot of nuance.

My hope is that by starting off with a principled implementation at the core we will be able to manage complexity much better over time. The current state machine isn't intended to be exhaustive - but rather a solid starting point.

Another goal is to make this component quite transparent to a casual reader. They should be able to look at the states and transitions between them to get a good sense of how things fit together.

grigoryk:
What's current getting in the way of tests is that fxaclient's FirefoxAccount is now hitting JNA's methods earlier during initialization than it used to, which is causing JNA to blow up in a test environment with `java.lang.UnsatisfiedLinkError: Native library (com/sun/jna/linux-x86-64/libjnidispatch.so)`.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/AccountManager.kt
```diff
@@ -0,0 +1,383 @@
...
+object Logout : AccountEvent()
+
+/**
+ * Observer interface which lets its users monitor account state changes and major events.
+ */
+interface AccountObserver {
+    /**
+     * Account just got logged out.
+     */
+    fun loggedOut()
```
csadilek: All our other observers use `on*` as a convention. `onLoggedOut` etc.

csadilek: I think this would read better here as well looking at the sample code.

samples/sync-history/src/main/java/org/mozilla/samples/sync/history/MainActivity.kt
```diff
@@ -46,98 +55,75 @@ class MainActivity : AppCompatActivity(), LoginFragment.OnLoginCompleteListener,
...
-            txtView.text = getString(R.string.logged_out)
+            launch { accountManager.logout().await() }
         }
 
         // NB: ObserverRegistry takes care of unregistering this observer when appropriate, and
         // cleaning up any internal references to 'observer' and 'owner'.
         featureSync.register(syncObserver, owner = this, autoPause = true)
+        accountManager.register(accountObserver, owner = this, autoPause = true)
+
+        // How that our account state observers are registered, we can kick off the account manager.
```
csadilek: nit: type: Now..

samples/sync-history/src/main/java/org/mozilla/samples/sync/history/MainActivity.kt
```diff
@@ -149,21 +135,48 @@ class MainActivity : AppCompatActivity(), LoginFragment.OnLoginCompleteListener,
...
-    private fun displayAndPersistProfile(code: String, state: String) {
-        launch {
-            account.completeOAuthFlow(code, state).await()
-            val profile = account.getProfile().await()
-            displayProfile(profile)
-            account.toJSONString().let {
-                getSharedPreferences(FXA_STATE_PREFS_KEY, Context.MODE_PRIVATE)
-                        .edit().putString(FXA_STATE_KEY, it).apply()
+    private val accountObserver = object : AccountObserver {
+        override fun beginOathFlow(authUrl: String) {
```
csadilek: We still plan on the `FirefoxAccountFeature`, right? It would observe these events e.g. `onBeginOAuthFlow` and open a new tab by default?

samples/sync-logins/src/main/java/org/mozilla/samples/sync/logins/MainActivity.kt
```diff
@@ -56,44 +60,56 @@ open class MainActivity : AppCompatActivity(), LoginFragment.OnLoginCompleteList
...
         setContentView(R.layout.activity_main)
         job = Job()
 
         listView = findViewById(R.id.logins_list_view)
         adapter = ArrayAdapter(this, android.R.layout.simple_list_item_1)
         listView.adapter = adapter
         activityContext = this
-        account = getAuthenticatedAccount() ?: FirefoxAccount(Config.release(CLIENT_ID, REDIRECT_URL))
+
+        accountManager.register(accountObserver, owner = this, autoPause = true)
```
csadilek: Can we register the observer as part of `init`? This would save an API call.

samples/sync-history/src/main/java/org/mozilla/samples/sync/history/MainActivity.kt
grigoryk: Yes, as an optional feature. I won't upgrade sample apps to use it, but we can use it in r-b.

grigoryk:
@csadilek how does the new API look like to you?

components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FirefoxAccountsAuthFeature.kt
```diff
@@ -0,0 +1,27 @@
...
+import kotlin.coroutines.CoroutineContext
+
+/**
+ * Ties together an account manager with a session manager/tabs implementation, facilitating an
+ * authentication flow.
+ */
+class FirefoxAccountsAuthFeature(
+    private val accountManager: FxaAccountManager,
+    private val tabsUseCases: TabsUseCases,
+    private val coroutineContext: CoroutineContext
```
csadilek: Should this have a default value: Dispatchers.MAIN?

components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FirefoxAccountsAuthFeature.kt
```diff
@@ -0,0 +1,27 @@
...
+) {
+    fun beginAuthentication() {
+        CoroutineScope(coroutineContext).launch {
+            tabsUseCases.addTab.invoke(accountManager.beginAuthentication().await())
```
csadilek: 👍

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/Profile.kt
```diff
@@ -4,7 +4,7 @@
 
 package mozilla.components.service.fxa
 
-data class Profile(
+open class Profile(
```
csadilek: Did you end up needing this `open`?

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FxaAccountManager.kt
```diff
@@ -0,0 +1,423 @@
...
+                }
+                AccountState.NotAuthenticated -> {
+                    when (event) {
+                        Authenticate -> AccountState.NotAuthenticated
+                        FailedToAuthenticate -> AccountState.NotAuthenticated
+                        is Authenticated -> AccountState.AuthenticatedNoProfile
+                        else -> null
+                    }
+                }
+                AccountState.AuthenticatedNoProfile -> {
```
vladikoff: Is there any way we can do to get rid of this state? I think we were hoping to treat the failure to get profile as a non-logged in case. 
cc @eoger

docs/changelog.md
```diff
@@ -31,6 +31,49 @@ permalink: /changelog/
 * **feature-prompts**, **browser-engine-gecko***
   * Added support for [Pop-up windows dialog](https://support.mozilla.org/en-US/kb/pop-blocker-settings-exceptions-troubleshooting#w_what-are-pop-ups).
 
+* **service-firefox-accounts**
+  * Added `FxaAccountManager`, which encapsulates a lower level accounts API and provides an observable interface for consumers that wish to be notified of account and profile changes.
+  * Background-worker friendly.
+  ```kotlin
+  // Long-lived instance, pinned on an application.
+  val accountManager = FxaAccountManager(context, Config.release(CLIENT_ID, REDIRECT_URL), arrayOf("profile"))
```
vladikoff: @eoger @grigoryk can anything be done in the FxaAccountManager right now to lay the foundation for the upcoming push / send-tab work? For example, can we briefly outline the events that SendTab <--> Push pipes will handle and brainstorm that?

docs/changelog.md
```diff
@@ -31,6 +31,49 @@ permalink: /changelog/
...
+  launch { accountManager.init() }
```
vladikoff: would consumers ever need to handle exceptions here?

docs/changelog.md
grigoryk: Yes; let’s iterate on this in the tree once the first pass lands.

docs/changelog.md
grigoryk: They should be emitted via onError; internally the state machine will end up in a serviceable state one way or another. This could encounter a wide set of exceptions - failures to read/parse a persisted account, failures coming out of fxaclient, etc. I’d like the FxaAccountManager to express the error conditions more explicitly, but it’s not yet clear to me which ones are worth capturing. Plan is to let these emerge naturally from how we’ll end up using it in the consuming code, and remove pain points over time by reifying states.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FxaAccountManager.kt
grigoryk: You could fail here for many reasons not related to authentication (intermittent networking problems for the device, server, profile endpoint issues, whatever else). At the same time, Sync could continue to work happily.
A better approach would be to capture explicitly auth-related problems while fetching a profile as events and have the state machine deal with them.

docs/changelog.md
grigoryk: Reason for landing first then iterating with push-specific events/states is that I need to start landing sync-related pieces that make use of this work (and are mostly ready to go), and they're currently higher priority than send tab stuff. Happy to talk about the push/sendtab next week.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FxaAccountManager.kt
```diff
@@ -0,0 +1,423 @@
...
+        return when (state) {
+            AccountState.AuthenticatedWithProfile -> profile
+            else -> null
+        }
+    }
+
+    fun beginAuthentication(): Deferred<String> {
+        val deferredAuthUrl: CompletableDeferred<String> = CompletableDeferred()
+
+        oauth.register(object : OAuthObserver {
```
eoger: This looks a bit weird to be honest, could we get rid of the `OAuthObserver` altogether? @csadilek told me it's more like a leftover implementation detail

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FxaAccountManager.kt
```diff
@@ -0,0 +1,423 @@
...
+    NotAuthenticated,
+    AuthenticatedNoProfile,
+    AuthenticatedWithProfile
+}
+
+/**
+ * Base class for account events.
+ * Events aren't a simple enum class because we might want to pass data along with some of the events.
+ */
+sealed class AccountEvent {
```
eoger: Nit: A state-machine event can be mistaken for a more classic observable-pattern event. It took me a little bit of time to realize these events were just consumed by the state machine. Could we use another term such as `Input`?

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FxaAccountManager.kt
```diff
@@ -0,0 +1,423 @@
...
+    override fun toString(): String {
+        // For a better logcat experience.
+        return this.javaClass.simpleName
+    }
+}
+object Init : AccountEvent()
```
eoger: Not a kotlin expert, but can we somehow put these events together in the same block or class?

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FxaAccountManager.kt
eoger: I tend to agree with @vladikoff, I think having a profile shouldn't be a state: in a way profile fetching is its own little state machine (no profile, loading, has profile, error fetch etc).
I'm also concerned about the upcoming combinations of states in the future (Is logged in, has profile, has push subscription but needs to register its instance again).

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FxaAccountManager.kt
grigoryk: I'm not too happy with this, but it's a straightforward way to implement this, and both `oauth` and `OAuthObserver` are private to this class so the outside world know about them.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FxaAccountManager.kt
```diff
@@ -0,0 +1,424 @@
...
+         * State transition matrix. It's in the companion object to enforce purity.
+         * @return An optional [AccountState] if provided state+event combination results in a
+         * state transition. Note that states may transition into themselves.
+         */
+        @Suppress("ComplexMethod")
+        internal fun nextState(state: AccountState, event: SMEvent): AccountState? {
+            return when (state) {
+                AccountState.Start -> {
+                    when (event) {
+                        SMEvent.Init -> AccountState.Start
```
csadilek: nit (don't change if nothing else needs an update): since this is private I would probably just go with `Event`. Not sure it's obvious what SM stands for :)

grigoryk:
The fxa samples app is has an example of using a pairing flow. I have a patch which mostly adds support for that additional flow to the account manager, but afaik support for that flow is not yet complete on the backend (@vladikoff would know more). I'll file a follow-up to update the sample app once all of the ducks are in a row.

vladikoff:
@grigoryk we can make it work in samples fxa next week, we have the dev servers deployed and ready

grigoryk:
Follow ups: https://github.com/mozilla-mobile/android-components/issues/1859, https://github.com/mozilla-mobile/android-components/issues/1860, https://github.com/mozilla-mobile/android-components/issues/1861.

PR #1765, opened 2019-01-19
Add support for hooking logs from app-services into the log system

thomcc:
This adds a new component that allows logs emitted by rust code from https://github.com/mozilla/application-services to be directed to the `Log` object in the base component. Previously, they were always written to logcat.

It's worth noting that this only works if you're using a [megazord](https://mozilla.github.io/application-services/docs/applications/consuming-megazord-libraries.html) build (Unfortunately, it's more or less impossible for us to make it work outside a megazord). Code running outside a megazord will continue to use logcat (realistically, we don't expect to ship non-megazords, so this basically means the android-components samples and the like).

I avoided doing anything fancy in `Log` to let it know about this type, although I could see an argument for having it be aware to automatically set the max level of RustLog or something, IDK (I don't think we want a hard dependency though, so there would need to be some annoying indirection that I punted on).

Anyway, I didn't really know where to put this. It's a lot less general purpose than the other stuff in `support`, but it doesn't really fit anywhere else either. LMK, I'm open to move it.

### Note:

This won't work until this is included in an app services release, which hasn't happened yet because it hasn't landed in app services yet (https://github.com/mozilla/application-services/pull/472).

Once it does I'll update the versions and add release notes.

### DIFF
--- a/.buildconfig.yml
+++ b/.buildconfig.yml
@@ -192,6 +192,10 @@ projects:
     path: components/support/utils
     description: 'A collection of generic helper classes.'
     publish: true
+  support-rustlog:
+    path: components/support/rustlog
+    description: 'A bridge allowing log messages from Rust code to be sent to the log system in support-base'
+    publish: true
   lib-crash:
     path: components/lib/crash
     description: 'A generic crash reporter library that can report crashes to multiple services.'
--- a/components/support/rustlog/src/main/java/mozilla/components/support/rustlog/RustLog.kt
+++ b/components/support/rustlog/src/main/java/mozilla/components/support/rustlog/RustLog.kt
@@ -1,0 +1,94 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.support.rustlog
+
+import mozilla.appservices.rustlog.LogLevelFilter
+import mozilla.appservices.rustlog.RustLogAdapter
+import mozilla.components.support.base.log.Log
+
+object RustLog {
+
+    /**
+     * Enable the Rust log adapter.
+     *
+     * This does almost (see below) nothing if you are not in a megazord build.
+     * After calling this, logs emitted by Rust code are forwarded to any
+     * LogSinks attached to [Log].
+     *
+     * Megazording is required due to each dynamically loaded Rust library having
+     * its own internal/private version of the Rust logging framework. When
+     * megazording, this is still true, but there's only a single dynamically
+     * loaded library, and so it is redirected properly.
+     *
+     * Note that non-megazord versions of the Rust libraries will log directly to
+     * logcat by default (at DEBUG level), so while they cannot hook into the base
+     * component log system, they still have logs available for development use.
+     *
+     * (We say "almost" nothing, as calling this will hook up logging for the dynamic
+     * library containing the Rust log hooking (and only that), as well as logging
+     * a single message indicating that it completed initialization).
+     */
+    fun enable() {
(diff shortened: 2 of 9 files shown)


thomcc:
@csadilek This is no longer blocked and should be landable. There's an issue that means we'll have to point things at nick's bintray for a-s things, the forseeable future.

build.gradle
```diff
@@ -40,6 +40,14 @@ allprojects {
...
             url "https://download.servo.org/nightly/maven"
         }
+
+        // Application services is currently published out of nalexander's bintray for reasons.
+        // FIXME: Remove after https://github.com/mozilla/application-services/issues/252
+        // is resolved.
+        maven {
+            name 'appservices'
+            url 'https://dl.bintray.com/ncalexander/application-services'
+        }
```
pocmo: Will we synchronize the artifact from there to jcenter like the others?

thomcc: I don't fully understand the question. If I'm being honest, I was instructed to do this by @ncalexan, who hopefully does.

ncalexan: It looks like we can just roll back https://github.com/mozilla-mobile/android-components/pull/1765/commits/eabc299a0f4e2bfa40093fda6c580413d9ba9dd1 'cuz `org.mozilla.appservices:rustlog` made it to jcenter: https://jcenter.bintray.com/org/mozilla/appservices/rustlog/.  I don't understand why it told me it couldn't be mirrored... maybe I did it without realizing it the first time?

thomcc: That commit is gone now, and things work. Resolving.

.buildconfig.yml
```diff
@@ -192,6 +192,10 @@ projects:
     path: components/support/utils
     description: 'A collection of generic helper classes.'
     publish: true
+  support-rustlog:
+    path: components/support/rustlog
+    description: 'A bridge allowing log messages from rust code to be sent to the the log system in support-base'
```
csadilek: nit: 'the the'

components/support/rustlog/src/main/java/mozilla/components/support/rustlog/RustLog.kt
```diff
@@ -0,0 +1,79 @@
...
+package mozilla.components.support.rustlog
+
+import mozilla.appservices.rustlog.LogLevelFilter
+import mozilla.appservices.rustlog.RustLogAdapter
+import mozilla.components.support.base.log.Log
+
+object RustLog {
+
+    /**
+     * Enable the rust log adapter. This does (almost) nothing if you are not
```
csadilek: This note (almost nothing) about megazord wasn't clear to me right away (before Nick explained it to me). Can we clarify this or simply write that it's only supported in Megazord for now?

docs/changelog.md
```diff
@@ -38,6 +39,28 @@ permalink: /changelog/
...
+  // Note: By default this is enabled at level DEBUG, which can be adjusted.
+  // (It is recommended you do this for performance if you adjust
+  // `Log.logLevel`).
+  RustLog.setMaxLevel(Log.Priority.INFO)
+  // You can also enable "trace logs", which may include PII
+  // (but can assist debugging) as follows. It is recommended
+  // you not do this in builds you distribute to users.
+  RustLog.setMaxLevel(Log.Priority.DEBUG, true)
+  ```
+  * This is pointless to do when not using a megazord (in which case logs from these libraries will be written directly to logcat automatically).
```
csadilek: Same as above. A clarification would be nice. Why?

thomcc:
I added clarification. It's unfortunate, since ideally megazording would just be an optimization, but it actually has semantic ramifications due to each `.so`'s internal variables (such as the global current logger inside the log crate) being hidden from other .so files.

csadilek:
@thomcc thanks! 🚢

PR #1774, opened 2019-01-21
Expect null strings when querying the accessibility service

Dexterp37:
Any reference in Java may be null, so we should check for null values when querying the accessibility manager.
Moreover, this changes Glean so that the enabled accessibility services ids are reported rather than their description

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
@@ -8,7 +8,6 @@ import android.view.accessibility.AccessibilityManager
 import android.accessibilityservice.AccessibilityServiceInfo
 import android.arch.lifecycle.ProcessLifecycleOwner
 import android.content.Context
-import android.content.pm.PackageManager
 import android.os.Build
 import android.support.annotation.VisibleForTesting
 import android.support.v4.view.accessibility.AccessibilityManagerCompat.getEnabledAccessibilityServiceList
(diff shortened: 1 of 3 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -178,23 +177,22 @@ open class GleanInternalAPI {
...
             return null
         }
         return accessibilityManager.getEnabledAccessibilityServiceList(
             AccessibilityServiceInfo.FEEDBACK_ALL_MASK
-        ).map {
-            it.loadDescription(packageManager)
+        ).mapNotNull {
+            // Note that any reference in java code can be null, so we'd better
+            // check for null values here as well.
+            it.id
```
pocmo: Looking at the code of `AccessibilityServiceInfo` it looks like the id can never be null. But it's probably safer to keep the check.

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
```diff
@@ -389,26 +387,31 @@ class GleanTest {
...
-        doReturn("service2").`when`(serviceSpy2).loadDescription(anyObject())
+        doReturn("service2").`when`(serviceSpy2).id
+
+        val nullDescription = spy<AccessibilityServiceInfo>(AccessibilityServiceInfo::class.java)
+        doReturn(null).`when`(nullDescription).id
 
-        shadowAccessibilityManager.setEnabledAccessibilityServiceList(listOf(serviceSpy1, serviceSpy2))
+        shadowAccessibilityManager.setEnabledAccessibilityServiceList(
+            listOf(serviceSpy1, serviceSpy2, nullDescription)
+        )
```
pocmo: nit: The name `nullDescription` may be misleading now that we are not using the description anymore. :)

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
mdboom: I see -- so we're using `id`s rather than descriptions, since descriptions are localized.  Makes sense.

Issue #95, opened 2018-04-19
module: Public Suffix List


pocmo:
A bunch of our apps use the public suffix list to split domains into parts. It would be helpful to wrap the list into a component and add some code to use it.

pocmo:
okhttp comes with a PublicSuffixDatabase:
https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/internal/publicsuffix/PublicSuffixDatabase.java

ekager:
Focus could use this to fix this bug (no priority assigned)
https://bugzilla.mozilla.org/show_bug.cgi?id=1445758

pocmo:
Related code in Fennec: https://searchfox.org/mozilla-central/source/mobile/android/base/java/org/mozilla/gecko/util/publicsuffix/PublicSuffix.java

In Fennec Gecko sends the base domain (public suffix + 1) to the app whenever the URL changes.

pocmo:
Assigning myself: Weekend hacking project :)

pocmo:
Code landed. Follow-up issues have the "publixsuffixlist" label.

PR #1775, opened 2019-01-21
Issue #95: Introduce new component for reading and accessing the public suffix list.

pocmo:
~~Work in progress from the weekend.~~

~~I still want to:~~
* [x] Get rid of the `okio` dependency
* [x]  Add some more documentation to the README.
* [x]  Debug an issue with a specific IDN domain (Moved to follow-up: #1776).
* [x]  Add even more test cases.

Ready for review!


components/lib/publicsuffixlist/src/main/java/mozilla/components/lib/publicsuffixlist/PublicSuffixList.kt
```diff
@@ -0,0 +1,120 @@
...
+     * E.g.:
+     * ```
+     *   co.uk       -> true
+     *   com         -> true
+     *   mozilla.org -> false
+     *   org         -> true
+     * ```
+     *
+     * Note that this method ignores the default "prevailing rule" described in the formal public suffix list algorithm:
+     * If a
```
csadilek: nit: leftover or incomplete sentence :)

csadilek:
I'll land this and add it to the changelog in the release prep PR.

PR #1812, opened 2019-01-23
Change Profile avatar representation

eoger:
Depends on https://github.com/mozilla/application-services/pull/579

### DIFF
--- a/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FirefoxAccount.kt
+++ b/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FirefoxAccount.kt
@@ -69,13 +69,7 @@ class FirefoxAccount internal constructor(private val inner: InternalFxAcct) : A
      */
     fun getProfile(ignoreCache: Boolean): Deferred<Profile> {
         return scope.async {
-            val internalProfile = inner.getProfile(ignoreCache)
-            Profile(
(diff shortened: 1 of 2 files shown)


components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FirefoxAccount.kt
```diff
@@ -69,13 +69,7 @@ class FirefoxAccount internal constructor(private val inner: InternalFxAcct) : A
...
     fun getProfile(ignoreCache: Boolean): Deferred<Profile> {
         return scope.async {
-            val internalProfile = inner.getProfile(ignoreCache)
-            Profile(
-                    uid = internalProfile.uid,
-                    email = internalProfile.email,
-                    avatar = internalProfile.avatar,
-                    displayName = internalProfile.displayName
-            )
+            inner.getProfile(ignoreCache).let { Profile.fromInternal(it) }
```
csadilek: nit (no need to change): you don't really need a separate scope here and could just write:
```kotlin
Profile.fromInternal(inner.getProfile(ignoreCache))
```
or
```kotlin
val internalProfile = inner.getProfile(ignoreCache)
Profile.fromInternal(internalProfile)
```

grigoryk:
I landed this in https://github.com/mozilla-mobile/android-components/pull/1932.

PR #1833, opened 2019-01-24
1522155: Implement "boolean" metric type

mdboom:
To support the Fenix north star metrics, we need a boolean metric type.


mdboom:
> :shipit: - does this require any chance to the parsers?

To the `glean_parser` parsers?  No -- they've had a placeholder for boolean for quite some time.

mdboom:
Thanks, @travis79!  I've updated the list in `StorageEngineManagers`.

PR #1834, opened 2019-01-24
Issue #1813: AwesomeBarFeature: Allow defining custom start/stop hooks.

pocmo:
### Pull Request checklist

- [X] **Changelog**: This PR includes a changelog entry or does not need one

### DIFF
--- a/components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/AwesomeBarFeature.kt
+++ b/components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/AwesomeBarFeature.kt
@@ -26,15 +26,24 @@ import mozilla.components.feature.tabs.TabsUseCases
 class AwesomeBarFeature(
     private val awesomeBar: AwesomeBar,
     private val toolbar: Toolbar,
-    private val engineView: EngineView? = null
+    private val engineView: EngineView? = null,
+    private val onEditStart: (() -> Unit)? = null,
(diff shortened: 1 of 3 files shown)


components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/AwesomeBarFeature.kt
```diff
@@ -26,15 +26,24 @@ import mozilla.components.feature.tabs.TabsUseCases
 class AwesomeBarFeature(
     private val awesomeBar: AwesomeBar,
     private val toolbar: Toolbar,
-    private val engineView: EngineView? = null
+    private val engineView: EngineView? = null,
+    private val onEditStart: (() -> Unit)? = null,
```
csadilek: nit: We could specify the default behaviour here e.g. ```= { hideAwesomeBar }```.

pocmo: That's what I tried first but it wasn't visible from there... and I was pretty sure I used something like this before. 🤔

boek: @pocmo I think if you define val inside the class declaration, not as a parameter you can do it
`private val onEditStart = onEditStart ?: ::hideAwesomeBar`

PR #1837, opened 2019-01-24
1522552: Separate out manufacturer and model into separate metrics

mdboom:
The manufacturer and model of the device are currently combined into the same metric.  This separates them out.

The complex subsetting of each part seems no longer necessary -- that was to make sure that enough of the model always made it into the string in preference over the manufacturer.  Now, I think we can just let the built-in default string truncation handle each.


components/service/glean/metrics.yaml
```diff
@@ -50,15 +50,29 @@ glean.baseline:
     notification_emails:
       - telemetry-client-dev@mozilla.com
 
-  device:
+  device_manufacturer:
```
Dexterp37: Let's also update the `baseline.md` file.

Dexterp37:
@mdboom I don't think this needs a new data review, but would you mind checking with @liuche before merging?

mdboom:
Sure.  I've reached out to @liuche for confirmation.

PR #1854, opened 2019-01-25
Closes #1840: Add menu and security icon color attributes

### LINKED ISSUE #1840: Add properties for changing the colour of menu and security icons
Fenix wants the ability to stylize the icons in the toolbar a bit more and we don't expose the ability to do so.

cc: @ekager

jonalmeida:
I've added some attributes to the `BrowserToolbar` to be able to colour the attributes.

The diff is a bit messy because I renamed the security icon value: `iconView` -> `siteSecurityIconView`.

### Pull Request checklist

- [ ] **Accessibility**: N/A ~The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features~

### DIFF
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
@@ -161,6 +161,12 @@ internal class DisplayToolbar(
 
     internal var securityIconColor = Pair(defaultColor, defaultColor)
 
+    internal var menuViewColor = defaultColor
+        set(value) {
+            field = value
+            menuView.setColorFilter(value)
+        }
+
     init {
         addView(siteSecurityIconView)
         addView(urlView)
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
@@ -6,6 +6,7 @@ package mozilla.components.browser.toolbar.display
 
 import android.annotation.SuppressLint
 import android.content.Context
+import android.support.v4.content.ContextCompat
 import android.support.v7.widget.AppCompatImageButton
 import android.support.v7.widget.AppCompatImageView
 import android.support.v7.widget.AppCompatTextView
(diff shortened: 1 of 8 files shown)


jonalmeida:
cc: @ekager

docs/changelog.md
```diff
@@ -29,6 +29,19 @@ permalink: /changelog/
...
+  * Added `BrowserToolbar` attributes to color the menu.
+  ```
+  <mozilla.components.browser.toolbar.BrowserToolbar
+      android:id="@+id/toolbar"
+      android:layout_width="match_parent"
+      android:layout_height="56dp"
+      android:background="#aaaaaa"
+      app:browserToolbarMenuColor="@color/photonBlue50"
+      app:browserToolbarInsecureColor="@color/photonRed50"
+      app:browserToolbarSecureColor="@color/photonGreen50" />
```
pocmo: I wonder if we need XML attributes or if we can define styles without them? (@Amejia481)

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
```diff
@@ -151,8 +152,18 @@ internal class DisplayToolbar(
...
 
+    private val defaultColor = ContextCompat.getColor(context, R.color.photonWhite)
+
+    internal var securityIconColor = Pair(defaultColor, defaultColor)
+
+    internal var menuViewColor = defaultColor
+        set(value) {
+            field = value
+            menuView.setColorFilter(value)
+        }
```
pocmo: Does this property need a backing field, it looks like we never read it?

docs/changelog.md
Amejia481: We can define styles without XML attributes,  but they're handy for configuring single properties.

Amejia481: Having both gives consumers more flexibility  of choosing which level of configuration they want

pocmo: That's definitely true. I was mostly concerned about having a gigantic XML API that we need to maintain.

jonalmeida: It's definitely not required like mentioned above. In this case, can't get the attribute set from `BrowserToolbar` in `DisplayToolbar` so most of the work is to enable access to the display toolbar, the rest is a nice-to-have. 🤷‍♂️

jonalmeida:
I'd like to get this merged since this PR is getting stale. @pocmo shall I remove the XML api?

PR #448, opened 2018-07-18
Issue #435: Fretboard: Documentation and guides

fercarcedo:
This pull request extends Fretboard README detailing how to use the library.

Closes #435

PR #1863, opened 2019-01-28
Closes #448 R-B:Not allowing text behind the security icon be selectable

Amejia481:
### Pull Request checklist

Reference Browser issue https://github.com/mozilla-mobile/reference-browser/issues/448


Amejia481:
After doing many tests, I realized that AppCompatImageView allows text behind it to be selectable, when it doesn't have click listener set. I believe in the future we are going to expose the click event for consumers know, when the security icon was clicked, but right now, I'm not sure if this the right moment, for this reason I prevent the issue for keep happening.

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
```diff
@@ -148,6 +153,9 @@ internal class DisplayToolbar(
             field = value
         }
 
+    // Optional callback to be call when the icon.
+    internal var onIconClicked: (() -> Unit)? = null
+
```
pocmo: This one seems unneeded?

Amejia481: Sorry a leftover from some previous experiments :P

pocmo: Why are you not onboarding?? :P

Amejia481:
PR updated :)

PR #1878, opened 2019-01-28
Use a single Miniconda installation for all glean-using libraries.

mdboom:
As part of 1505395, where we want to start collecting metrics from a number of different libraries that will all be collected by Glean, it was noticed that each of these libraries would install their own copy of Miniconda in order to run the glean_parser to generate code for the metrics.  Each Miniconda tree is approx 300MB, so it's a non-trivial amount of space.

By installing Miniconda to the "Gradle User Dir" (usually `~/.gradle` on Unixy systems) rather than the project build directory, we can have a single copy of Miniconda that is shared among all libraries that use glean, including glean itself.  Unlike earlier approaches, this still works if glean is depended on by a library outside of the android-components tree and is used as a pre-compiled binary.

This directory is versioned by the version of miniconda that was installed, so we can upgrade it at a later date and still support switching between branches of android-components to both the old and new version etc.  The version of `glean_parser` installed within the miniconda installation is already version checked to support branch hopping.

### DIFF
--- a/components/service/glean/scripts/sdk_generator.gradle
+++ b/components/service/glean/scripts/sdk_generator.gradle
@@ -4,7 +4,34 @@
 
 import org.apache.tools.ant.taskdefs.condition.Os
 
+// This installs a Miniconda3 environment into the gradle user home directory
+// so that it will be shared between all libraries that use glean.  This is
+// important because it is approximately 300MB in installed size.
+
 String GLEAN_PARSER_VERSION = "0.12.0"
+// The version of Miniconda is explicitly specified.
+// Miniconda3-4.5.12 is known to not work on Windows.
+String MINICONDA_VERSION = "4.5.11"
+File GLEAN_BOOTSTRAP_DIR = new File(
+    project.getGradle().gradleUserHomeDir,
+    "glean/bootstrap-${MINICONDA_VERSION}"
+)
+File GLEAN_MINICONDA_DIR = new File(
+    GLEAN_BOOTSTRAP_DIR,
+    "Miniconda3"
+)
+
+// Define the path to the Miniconda3 installation as a buildConfig variable so
+// that tests that need to run JSON through our JSON schema can do so.
+android {
+    defaultConfig {
+        buildConfigField(
+            "String",
+            "GLEAN_MINICONDA_DIR",
+            "\"" + GLEAN_MINICONDA_DIR + "\""
+        )
+    }
+}
 
 /* This script runs a given Python module as a "main" module, like
  * `python -m module`. However, it first checks that the installed
@@ -44,16 +71,38 @@ subprocess.check_call([
 
 // Configure the Python environments
 envs {
-    bootstrapDirectory = new File(buildDir, 'bootstrap')
-    envsDirectory = new File(buildDir, 'envs')
+    bootstrapDirectory = GLEAN_BOOTSTRAP_DIR
     pipInstallOptions = "--trusted-host pypi.python.org --no-cache-dir"
 
     // Setup a miniconda environment. conda is used because it works
     // non-interactively on Windows, unlike the other options
-    conda "Miniconda3", "Miniconda3-4.5.11", "64", ["glean_parser==${GLEAN_PARSER_VERSION}"]
+    conda "Miniconda3", "Miniconda3-${MINICONDA_VERSION}", "64", ["glean_parser==${GLEAN_PARSER_VERSION}"]
 }
 
+// Even though we are installing the Miniconda environment to the gradle user
+// home directory, the gradle-python-envs plugin is hardcoded to download the
+// installer to the project's build directory. Doing so will fail if the
+// project's build directory doesn't already exist. This task ensures that
+// the project's build directory exists before downloading and installing the
+// Miniconda environment.
+// See https://github.com/JetBrains/gradle-python-envs/issues/26
+task createBuildDir {
+    description = "Make sure the build dir exists before creating the Python Environments"
+    onlyIf {
+        !file(buildDir).exists()
+    }
+    doLast {
+        println "Creating build directory:" + buildDir.getPath()
+        buildDir.mkdir()
+    }
+}
+
+preBuild.dependsOn(createBuildDir)
 preBuild.finalizedBy("build_envs")
+
+clean {
+    delete GLEAN_BOOTSTRAP_DIR
+}
 
 // Generate the Metrics API
 ext.gleanGenerateMetricsAPI = {
@@ -67,9 +116,9 @@ ext.gleanGenerateMetricsAPI = {
             workingDir  rootDir
             // Note that the command line is OS dependant: on linux/mac is Miniconda3/bin/python.
             if (Os.isFamily(Os.FAMILY_WINDOWS)) {
-                commandLine "$buildDir/bootstrap/Miniconda3/python"
+                commandLine new File(GLEAN_MINICONDA_DIR, "python")
             } else {
-                commandLine "$buildDir/bootstrap/Miniconda3/bin/python"
+                commandLine new File(GLEAN_MINICONDA_DIR, "bin/python")
             }
 
             args "-c"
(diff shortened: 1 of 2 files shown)


components/service/glean/build.gradle
```diff
@@ -16,6 +16,12 @@ apply plugin: 'kotlin-android'
  */
 String GLEAN_PING_SCHEMA_GIT_HASH = "8939bcb"
 String GLEAN_PING_SCHEMA_URL = "https://raw.githubusercontent.com/mozilla-services/mozilla-pipeline-schemas/$GLEAN_PING_SCHEMA_GIT_HASH/schemas/glean/baseline/baseline.1.schema.json"
+// NOTE: this is also defined in scripts/sdk_generator.gradle
+String MINICONDA_VERSION = "4.5.11"
```
Dexterp37: Is there any way to share this instead of declaring it twice?

components/service/glean/build.gradle
```diff
@@ -16,6 +16,12 @@ apply plugin: 'kotlin-android'
...
+String GLEAN_MINICONDA_DIR = new File(
+    project.getGradle().gradleUserHomeDir,
```
Dexterp37: It feels weird to put stuff in the gradle homedir :-\ Maybe another good location could be "parent's project build dir"? I'm not sure if such a thing is even possible though. @pocmo , any thought on that? The idea is to have a place where to download the environment for glean python stuff, to avoid having every a-c component using glean to have its own copy of the environment.

components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -44,15 +46,42 @@ subprocess.check_call([
...
     pipInstallOptions = "--trusted-host pypi.python.org --no-cache-dir"
 
     // Setup a miniconda environment. conda is used because it works
     // non-interactively on Windows, unlike the other options
-    conda "Miniconda3", "Miniconda3-4.5.11", "64", ["glean_parser==${GLEAN_PARSER_VERSION}"]
+    conda "Miniconda3", "Miniconda3-${MINICONDA_VERSION}", "64", ["glean_parser==${GLEAN_PARSER_VERSION}"]
 }
 
+// Setup the python environment before the build starts.
+task createBuildDir {
```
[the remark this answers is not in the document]
mdboom: Because the "fix" in `com.jetbrains.python.envs` only works if you're installing the build environment under build (`bootstrapDirectory` is under build).  If `bootstrapDirectory` points to another location, it still (temporarily) downloads the Miniconda installer to the `build` directory, which in this case now doesn't exist.  It's a pain, but only a minor one, given the advantages, IMHO.

components/service/glean/build.gradle
mdboom: I actually think it's not a bad place given what else already goes in `~/.gradle`: gradle itself, gradle plugins, caches downloads of packages.  It feels like the natural place to me -- moreso than the build directory, which I generally would think includes build products, not build-time tools.

Remember also, that this will be used by applications (and potentially libraries) that exist outside of a-c as well.

components/service/glean/build.gradle
mdboom: Not that I was able to find -- Groovy/Gradle is weird -- but I'll look at it again today with fresh eyes.

mdboom: I think I found a way forward.  The `sdk_generator.gradle` file can specify the buildConfig variable used to find the Python installation during testing (to validate against schemas).  Then we don't have to do it from glean's `build.gradle`.

Dexterp37: Ah, interesting. I wonder how that would look in code, but sounds like a nice alternative to the current state!

components/service/glean/scripts/sdk_generator.gradle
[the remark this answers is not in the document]
Dexterp37: I see, makes sense!

components/service/glean/build.gradle
Dexterp37: Mh, I see. I still think @pocmo (or maybe @ncalexan ) might want to review this for sanity.

components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -44,15 +63,32 @@ subprocess.check_call([
...
     pipInstallOptions = "--trusted-host pypi.python.org --no-cache-dir"
 
     // Setup a miniconda environment. conda is used because it works
     // non-interactively on Windows, unlike the other options
-    conda "Miniconda3", "Miniconda3-4.5.11", "64", ["glean_parser==${GLEAN_PARSER_VERSION}"]
+    conda "Miniconda3", "Miniconda3-${MINICONDA_VERSION}", "64", ["glean_parser==${GLEAN_PARSER_VERSION}"]
+}
+
+// Setup the python environment before the build starts.
+task createBuildDir {
```
Dexterp37: Let's rename this to something more specific: are we creating the miniconda dir or the main glean directory under the gradle home dir here?

components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -44,15 +63,32 @@ subprocess.check_call([
...
+    onlyIf {
+        !file(buildDir).exists()
+    }
+    doLast {
+        println "Creating build directory:" + buildDir.getPath()
+        buildDir.mkdir()
+    }
 }
 
+// Make sure that a build directory exists before attempting to download
```
Dexterp37: Is this comment still valid? At the very least, we should mention that this is failing because we're not creating the dir under the project build dir, right?

components/service/glean/build.gradle
mdboom: This PR has been updated with those changes.  Let me know what you think.

components/service/glean/build.gradle
mdboom: I totally agree.  I'm a total newbie with gradle.

components/service/glean/scripts/sdk_generator.gradle
mdboom: We're creating the `build` directory for the project here.  That's where the Python plugin insists on downloading the installer for Miniconda.  It ultimately installs it *to* where we tell it to, but it's hardcoded to use the project's build dir as its scratch space for the installer.

components/service/glean/scripts/sdk_generator.gradle
mdboom: Sure.  I'll clarify that we're talking about the project's build directory here.

mdboom:
I've added significant comments to make it clearer what's going on here.

components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -5,6 +5,25 @@
...
+)
+
+android {
+    defaultConfig {
+        buildConfigField(
+            "String",
+            "GLEAN_MINICONDA_DIR",
+            "\"" + GLEAN_MINICONDA_DIR + "\""
+        )
+    }
```
pocmo: This will be part of the public API of the library. Not necessarily problematic though.

mdboom: Yes. Do you know off-hand if there's a way of making it "for testing only"?  It really should only ever be used from running unit tests.  I figured out how to do debug vs. release, but not test vs. non-test.

mdboom:
@pocmo: That's a good idea to add a cleanup task.  I'll look into that and update this.

mdboom:
@pocmo: I added a `clean` task to delete the Miniconda installation.  (I think that's what we want -- `cleanup` seems to be intended for code-formatting cleanups).  It seems to work fine if the directory is already missing (which will be the case when multiple libraries use glean).

components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -44,17 +71,39 @@ subprocess.check_call([
...
+        buildDir.mkdir()
+    }
+}
+
+preBuild.dependsOn(createBuildDir)
 preBuild.finalizedBy("build_envs")
 
+clean {
+    delete GLEAN_BOOTSTRAP_DIR
+}
```
pocmo: We execute `clean` for a bunch of reasons. I wonder if we should give that a different name?

Also: Doesn't that need to be wrapped in `doLast` to not get executed whenever the script gets evaluated (Not sure about that)?

PR #1879, opened 2019-01-28
[glean] 1505395, 1523331: Generate specific metrics into the library's namespace.

mdboom:
This generates the specific metrics defined in a `metrics.yaml` into the package namespace defined by the library (which comes from `variant.applicationId`).  It turns out we already had support for this on the `glean_parser` side, we just weren't using it properly -- we had sort of hardcoded the cases as "in an app" or "in glean itself".  This generalizes it to any app or third-party library using Glean.

As you can see from the changes required to the sample app, this requires changes to app code, since the metrics will now exist in the app's namespace, not glean's.  This might have affected the reference browser, but it doesn't look like the reference browser defines any of its own metrics yet, so I don't think there's anything to update there.


components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -57,10 +57,11 @@ preBuild.finalizedBy("build_envs")
 
 // Generate the Metrics API
 ext.gleanGenerateMetricsAPI = {
-    namespace -> {
+    {
         variant ->
 
         def sourceOutputDir =  "$buildDir/telemetry/src/${variant.name}/kotlin"
+        def fullNamespace = "${variant.applicationId}.metrics"
```
Dexterp37: I'd rather have the sob package be more specific to avoid problems, e.g `GleanMetrics`

travis79: I agree with Dexter here, that glean should be in the sub-package name.  It's also good advertising :)

mdboom: Sure.  Done.

components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -106,68 +106,67 @@ clean {
...
-            // Note that the command line is OS dependant: on linux/mac is Miniconda3/bin/python.
-            if (Os.isFamily(Os.FAMILY_WINDOWS)) {
-                commandLine new File(GLEAN_MINICONDA_DIR, "python")
-            } else {
-                commandLine new File(GLEAN_MINICONDA_DIR, "bin/python")
-            }
+    variant ->
+
+    def sourceOutputDir =  "$buildDir/telemetry/src/${variant.name}/kotlin"
+    def fullNamespace = "${variant.applicationId}.GleanMetrics"
```
pocmo: Is this going to work with libraries .. they do not really define an `applicationId`?

mdboom: I was surprised by this, too, but it seems that they do (somehow).  For the glean library it is set to `org.mozilla.components.service.glean`.  In another follow-on PR (#1926) I added another library and it did the right thing.  I also added glean support to fretboard in an exploratory branch and it did the right thing.

PR #1882, opened 2019-01-29
Closes #1880: Clear download consumable when permission is denied

### LINKED ISSUE #1880: Clear download consumable when permission is denied
We currently react to granted permissions and consume the download in `DownloadFeature.onPermissionGranted`, but if the permission wasn't granted we keep the download unconsumed/pending.

It seems cleaner to set the consumable to empty in that case.

csadilek:
@pocmo Maybe for tomorrow's release? We could do that in the app but it's nicer to have a symmetric API.  

### Pull Request checklist

### DIFF
--- a/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt
+++ b/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt
@@ -15,25 +15,28 @@ import mozilla.components.browser.session.Session
 import mozilla.components.browser.session.SessionManager
 import mozilla.components.feature.downloads.DownloadDialogFragment.Companion.FRAGMENT_TAG
 import mozilla.components.support.base.feature.LifecycleAwareFeature
+import mozilla.components.support.base.observer.Consumable
 import mozilla.components.support.ktx.android.content.isPermissionGranted
 
 typealias OnNeedToRequestPermissions = (session: Session, download: Download) -> Unit
 
 /**
- * Feature implementation for proving download functionality for the selected session.
- * It will subscribe to the selected session and will listening for downloads.
+ * Feature implementation to provide download functionality for the selected
+ * session. The feature will subscribe to the selected session and listen
+ * for downloads.
  *
- * @property applicationContext a reference to [Context] applicationContext.
- * @property onNeedToRequestPermissions it will be call when you need to request permission,
- * before a download can be performed.
- * @property onDownloadCompleted a callback to be notified when a download is completed.
- * @property downloadManager a reference of [DownloadManager] who is in charge of performing the downloads.
- * @property sessionManager a reference of [SessionManager].
- * @property fragmentManager a reference of [FragmentManager]. If you provide a [fragmentManager]
- * a dialog will show before every download.
- * @property fragmentManager a reference of [FragmentManager].
- * @property dialog a reference of [DownloadDialogFragment]. If it is not provided, it will be
- * initialized with an instance of [SimpleDownloadDialogFragment].
+ * @property applicationContext a reference to the application context.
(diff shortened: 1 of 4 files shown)


csadilek:
Not sure what's up with codecov. Diff is covered and was green on previous run.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt
```diff
@@ -98,6 +99,14 @@ class DownloadsFeature(
         }
     }
```
pocmo: nit: While we are here, this code seems to not need `let` and could be written as concise as your `onPermissionDenied()` code.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt
```diff
@@ -98,6 +99,14 @@ class DownloadsFeature(
...
 
+    /**
+     * Notifies the feature that the permissions were denied. It will then
+     * clear the pending download.
+     */
+    fun onPermissionsDenied() {
+        activeSession?.download = Consumable.empty()
+    }
```
pocmo: Is this something we need to do in order to avoid re-requesting the permission when the app resumes (is that something we do?).

I could see an app showing an indicator for a pending download that still needs a permission. But that's something that is planned and we do not need to support it now. And we could add this later by letting the app injecting a different behavior (and ours being the default implementation).

pocmo:
> @pocmo Maybe for tomorrow's release? We could do that in the app but it's nicer to have a symmetric API.

Sorry, I didn't see that before starting the release. Is this needed in an app that is not on snapshots?

docs/changelog.md
```diff
@@ -165,6 +165,9 @@ awesomeBar.layout = CustomSuggestionLayout()
 * **browser-toolbar**
   * Fixing bug that allowed text behind the security icon being selectable. [Issue #448](https://github.com/mozilla-mobile/reference-browser/issues/448)
 
+* **feature-downloads**
+  * Introduced a new `onPermissionsDenied` method, which similar to `onPermissionsGranted`, should be invoked in response to permission requests. If the requested permissions were denied the feature will clear the pending download.
+
```
pocmo: Nit: Because of my failure to merge this before the release; this now needs to be moved up to 0.41.0 :)

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt
csadilek: @pocmo hm, but we need to capture a non-null session. So, this won't work:
```kotlin
 activeSession?.download?.consume {
   onDownload(activeSession, it)
  }
```
because `activeSession` is a mutable property that could have been changed by this time :). So, I think let is the best way here?

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt
csadilek: Yes, exactly. When the request is denied, I think we should get rid of it to prevent re-requesting it (if someone wants that behaviour) and also to free up the consumable. Mostly found the API missing when I handled permission granted and there was no `else` block.

csadilek:
@pocmo 
>Sorry, I didn't see that before starting the release. Is this needed in an app that is not on snapshots?

No, all good :). We can always publish snapshots manually.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt
pocmo: Oh, yeah, makes sense!

PR #1858, opened 2019-01-26
Closes #1079: Adding find in page feature component

Amejia481:
### Pull Request checklist

- [x] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [x] **Changelog**: This PR includes a changelog entry or does not need one
- [x] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features

[GeckoView Demo](https://drive.google.com/file/d/1JXJsIqJTZMi-0pmoESk2af2WOjU0zGE3/view?usp=sharing)
[SystemView Demo](https://drive.google.com/file/d/12JyNg1B9Qfez6xA7vpvxSEfwTaS25lzo/view?usp=sharing)

PR #1888, opened 2019-01-29
Closes #1887: Addressing some nits from the find in page feature component that were pending in #1858.

### LINKED ISSUE #1887: Follow-up fix nits on find in page feature component
There were some nits that need to be addressed, related https://github.com/mozilla-mobile/android-components/pull/1858

Amejia481:
### Pull Request checklist

- [ ] ~~**Changelog**: This PR includes a changelog entry or does not need one~~
**No need it, this PR just rewording some text to make it clear.**

### DIFF
--- a/components/feature/findinpage/README.md
+++ b/components/feature/findinpage/README.md
@@ -12,7 +12,7 @@ implementation "org.mozilla.components:feature-findinpage:{latest-version}"
 ```
 
 ### Adding feature find in page.
-To use feature find in page you have to do two things:
+To use this feature you have to do two things:
 
 **1. Add the `FindInPageView` widget to you layout.**
 
--- a/components/feature/findinpage/src/main/java/mozilla/components/feature/findinpage/FindInPageView.kt
+++ b/components/feature/findinpage/src/main/java/mozilla/components/feature/findinpage/FindInPageView.kt
@@ -26,7 +26,6 @@ private const val DEFAULT_VALUE = 0
  * A customizable FindInPage UI widget.
  */
 @Suppress("TooManyFunctions")
-
 class FindInPageView @JvmOverloads constructor(
     context: Context,
     attrs: AttributeSet? = null,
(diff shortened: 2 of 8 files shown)


samples/browser/src/main/java/org/mozilla/samples/browser/DefaultComponents.kt
```diff
@@ -120,7 +120,7 @@ open class DefaultComponents(private val applicationContext: Context) {
         )
     }
 
-    var onFindPageClicked: (() -> Unit)? = null
+    var onFindInPageClicked: (() -> Unit)? = null
```
Amejia481: @csadilek Do you want to use this PR for chatting about this lambda?

components/feature/findinpage/README.md
```diff
@@ -12,7 +12,7 @@ implementation "org.mozilla.components:feature-findinpage:{latest-version}"
 ```
 
 ### Adding feature find in page.
-To use feature find in page you have to do two things:
+To use this feature have to do two things:
```
csadilek: nit: To use this feature *you*, or just "Using this feature requires:"

components/feature/findinpage/src/main/java/mozilla/components/feature/findinpage/FindInPageView.kt
```diff
@@ -39,10 +38,10 @@ class FindInPageView @JvmOverloads constructor(
     var onCloseButtonPressListener: (() -> Unit)? = null
 
     /**
-     * Gives access to the underneath engine, to start the search and to control the navigation between
+     * Gives access to the underneath sessionEngine to start the search and to control the navigation between
```
csadilek: nit: Comment still references `sessionEngine`. So, should be "..access to the engine session to start...."

samples/browser/src/main/java/org/mozilla/samples/browser/DefaultComponents.kt
csadilek: @Amejia481 let's talk through options when you're back. I don't have a quick solution that I like.

Amejia481: 👍

pocmo: I did it a bit differently when integrating this into the ref browser (see PR). It's still not super nice but it avoids leaking UI by attaching it to the components object.

PR #1896, opened 2019-01-30
AutoSave: Save state after removing all sessions.

pocmo:
### Pull Request checklist

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/storage/AutoSave.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/storage/AutoSave.kt
@@ -203,4 +203,9 @@ private class AutoSaveSessionChange(
         autoSave.logger.info("Save: Session removed")
         autoSave.triggerSave()
     }
+
+    override fun onAllSessionsRemoved() {
+        autoSave.logger.info("Save: All sessions removed")
+        autoSave.triggerSave()
+    }
 }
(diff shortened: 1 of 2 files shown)


pocmo:
In the reference browser this is noticeable if you remove all sessions in the tabs tray and then kill the app. After that the previous state with all sessions is restored. :)

pocmo:
@csadilek Good catch. Let's move that to a new issue! That seems to be an issue regardless of this PR.

AutoSave should either persist the empty state or removed the saved state. I really do not like that the `SessionManager` returns null snapshots :(

csadilek:
@pocmo yeah, true it would also happen if one closes the last session. I've filed: https://github.com/mozilla-mobile/android-components/issues/1908

But we should land the test here though with a defaultsession lambda as it proves that your changes work:
```kotlin
@Test
fun `AutoSave - when session gets removed`() {
        runBlocking {
            val sessionManager = SessionManager(mock(), { Session("https://getpocket.com" )})
            sessionManager.add(Session("https://www.firefox.com"))
            val session = Session("https://www.mozilla.org").also {
                sessionManager.add(it)
            }

            val sessionStorage: SessionStorage = mock()

            val autoSave = AutoSave(
                sessionManager = sessionManager,
                sessionStorage = sessionStorage,
                minimumIntervalMs = 0
            ).whenSessionsChange()

            assertNull(autoSave.saveJob)
            verify(sessionStorage, never()).save(any())

            sessionManager.removeAll()

            autoSave.saveJob?.join()

            verify(sessionStorage).save(any())
        }
    }

PR #1897, opened 2019-01-30
Closes #1890: PromptFeature - Allow app to specify permission request code

### LINKED ISSUE #1890: PromptFeature should not specify permission request code
We just updated the `DownloadsFeature` API to let the app know what permissions to [request](https://github.com/mozilla-mobile/android-components/blob/master/samples/browser/src/main/java/org/mozilla/samples/browser/BrowserFragment.kt#L90) and provide a call to process the permission request [result](https://github.com/mozilla-mobile/android-components/blob/master/samples/browser/src/main/java/org/mozilla/samples/browser/BrowserFragment.kt#L202). The application should be in control when requesting permissions.

The `PomptFeature` is close to this behaviour, but it specifies a request code which isn't ideal. Theoretically, it could collide with a code that the app uses elsewhere. I am thinking we should let the app control the request codes similar to our new `DownloadFeature` solution. This would also make all permission requests work in the same way.

csadilek:
This is so we have a consistent way for how our features request permissions (see updated sample code in this commit) and to put the app in control of permission request codes. Otherwise, we'd risk code collisions and bugs.

### Pull Request checklist

### DIFF
--- a/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt
+++ b/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt
@@ -40,31 +40,38 @@ import java.util.Date
 @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
 internal const val FRAGMENT_TAG = "mozac_feature_prompt_dialog"
 
-typealias OnNeedToRequestPermissions = (session: Session, permissions: Array<String>, requestCode: Int) -> Unit
+typealias OnNeedToRequestPermissions = (permissions: Array<String>) -> Unit
 
 /**
- * Feature for displaying native dialogs for html elements like:
- * input type date,file,time,color, option, menu, authentication, confirmation and other alerts.
+ * Feature for displaying native dialogs for html elements like: input type
+ * date, file, time, color, option, menu, authentication, confirmation and alerts.
  *
- * There are some requests that are not handled with dialogs instead with intents like file choosers and others.
- * For this reason, you have to keep the feature aware of flow of requesting data from other apps, overriding
- * onActivityResult on your [Activity] or [Fragment] and forward its calls to [onActivityResult].
+ * There are some requests that are handled with intents instead of dialogs,
+ * like file choosers and others. For this reason, you have to keep the feature
+ * aware of the flow of requesting data from other apps, overriding
+ * onActivityResult in your [Activity] or [Fragment] and forward its calls
+ * to [onActivityResult].
  *
- * This feature will subscribe to the currently selected [Session] and display a suitable native dialog based on
- * [Session.Observer.onPromptRequested] events. Once the dialog is closed or the user selects an item from the dialog
- * the related [PromptFeature] will be consumed.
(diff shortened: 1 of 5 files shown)


components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt
```diff
@@ -40,31 +40,38 @@ import java.util.Date
...
 /**
- * Feature for displaying native dialogs for html elements like:
- * input type date,file,time,color, option, menu, authentication, confirmation and other alerts.
+ * Feature for displaying native dialogs for html elements like: input type
+ * date, file, time, color, option, menu, authentication, confirmation and alerts.
  *
- * There are some requests that are not handled with dialogs instead with intents like file choosers and others.
- * For this reason, you have to keep the feature aware of flow of requesting data from other apps, overriding
- * onActivityResult on your [Activity] or [Fragment] and forward its calls to [onActivityResult].
+ * There are some requests that are handled with intents instead of dialogs,
```
jonalmeida: Less negation! 👏

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt
```diff
@@ -135,30 +143,27 @@ class PromptFeature(
...
 
     /**
-     * Use in conjunction with [onNeedToRequestPermissions], to notify the feature that all the required permissions
-     * have been granted, and it can perform the pending [PromptRequest] in the selected session.
+     * Used in conjunction with [onNeedToRequestPermissions], to notify the feature
+     * that all the required permissions have been granted, and the pending [PromptRequest]
+     * can be performed.
      *
-     * If the required permission has not been completely granted [onNeedToRequestPermissions] will be called.
+     * If the required permission has not granted [onNeedToRequestPermissions] will be called.
```
jonalmeida: nit: has not _been_ granted.

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt
```diff
@@ -479,6 +484,6 @@ class PromptFeature(
     }
 
     companion object {
-        const val FILE_PICKER_REQUEST = 1234
+        const val FILE_PICKER_ACTIVITY_REQUEST_CODE = 1234
```
jonalmeida: This is outside of this PR, but it's probably better to allow the app to pass their own request code to avoid collisions with other request codes that may use the same values.

This may have been talked about before when initially landing and this was the best solution at the time.

csadilek: Yes, agreed. We need to do the same for activity request codes.

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt
csadilek: 👍 fixed.

PR #1909, opened 2019-01-31
[glean] 1523979: Clean up some loose ends about initialization

mdboom:
Now that glean supports being used from multiple libraries, we need
to be careful that initialization happens only once, and that code
depending on initialization checks its status.

- I haven't been able to find a way to make `Glean.initialize` only available to application code, but we can document that requirement clearly.

- Based on an analysis of the actual side effects of `Glean.initialize`, there was only one method requiring initialization that wasn't checking for initialization -- `Glean.sendPing` -- so an explicit check has been added there.


components/service/glean/README.md
```diff
@@ -80,6 +80,10 @@ class SampleApplication : Application() {
 Once initialized, if collection is enabled, glean will automatically start collecting [baseline metrics](metrics.yaml)
 and sending its [pings](docs/pings.md).
 
+Glean should be initialized as soon as possible, and importantly, before any
+other libraries in the application start using Glean. Library code should never
```
pocmo: We setup logging and crash reporting before glean :P

mdboom: I think logging would be fine.  If crash reporting ever started using glean, we might need to change that.  Would that be hard to do?

pocmo: It's probably never going to call Glean from its own initializing code. But let's see once we get there. :)

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
```diff
@@ -347,9 +348,15 @@ class GleanTest {
         Glean.initialized = true
     }
 
-    @Test(expected = IllegalStateException::class)
-    fun `Don't initialize twice`() {
+    @Test
+    fun `Initializing twice is a no-op`() {
+        val beforeConfig = Glean.configuration
+
         Glean.initialize(ApplicationProvider.getApplicationContext())
```
Dexterp37: nit: Should we explicitly provide a configuration with test values here (e.g. the server url to something like "https:\\no.init.should.take.place.with.this"?)?

PR #1922, opened 2019-02-01
Perform browser-toolbar autocompletion off the UI thread

grigoryk:
At this point in the stack, we're not in control over what our
autocomplete providers are, what actions they'll do in order to
field our queries, etc. For example, some providers may hit the disk
and perform expensive DB queries internally. Some may even hit the
network, in theory!

In order to keep things perceptively speedy, let's run the actual work
off the main thread. This patch sets up a new pool thread to process
autocomplete requests. More than one thread is selected so that we maintain
liveliness during quick user input. Background tasks are cancelled as new
queries come in, and stale results are discarded.

### Pull Request checklist

### DIFF
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
@@ -14,6 +14,14 @@ import android.view.View
 import android.view.View.OnFocusChangeListener
 import android.view.ViewGroup
 import android.widget.ImageButton
+import kotlinx.coroutines.CoroutineExceptionHandler
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.asCoroutineDispatcher
+import kotlinx.coroutines.cancelChildren
+import kotlinx.coroutines.isActive
+import kotlinx.coroutines.launch
 import mozilla.components.browser.menu.BrowserMenuBuilder
 import mozilla.components.browser.toolbar.display.DisplayToolbar
 import mozilla.components.browser.toolbar.edit.EditToolbar
@@ -21,10 +29,17 @@ import mozilla.components.concept.toolbar.AutocompleteDelegate
 import mozilla.components.concept.toolbar.AutocompleteResult
 import mozilla.components.concept.toolbar.Toolbar
 import mozilla.components.support.base.android.Padding
+import mozilla.components.support.base.log.logger.Logger
 import mozilla.components.support.ktx.android.content.res.pxToDp
 import mozilla.components.support.ktx.android.view.forEach
 import mozilla.components.support.ktx.android.view.isVisible
+import mozilla.components.ui.autocomplete.AutocompleteView
 import mozilla.components.ui.autocomplete.InlineAutocompleteEditText
+import mozilla.components.ui.autocomplete.OnFilterListener
+import java.util.concurrent.Executors
+import kotlin.coroutines.CoroutineContext
+
+private const val AUTOCOMPLETE_QUERY_THREADS = 3
 
 /**
  * A customizable toolbar for browsers.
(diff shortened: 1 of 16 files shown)


grigoryk:
Need to fix up `EditToolbarTest` before this can land.

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
```diff
@@ -163,17 +190,51 @@ class BrowserToolbar @JvmOverloads constructor(
...
+                        if (result.input != latestAutocompleteInput) {
+                            Log.log(
+                                tag = logTag,
+                                priority = Log.Priority.INFO,
+                                message = "Discarding result for \"${result.input}\" input"
+                            )
+                            return
+                        }
+                        // Jump back to the UI thread to process the results.
+                        CoroutineScope(Dispatchers.Main).launch {
```
pocmo: You are not actually jumping back to the UI thread but launching a new asynchronous coroutine here; of which there can be multiple. Maybe you want to use [withContext](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/with-context.html) here?

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
```diff
@@ -50,11 +61,27 @@ class BrowserToolbar @JvmOverloads constructor(
...
     @VisibleForTesting internal var displayToolbar = DisplayToolbar(context, this)
     @VisibleForTesting internal var editToolbar = EditToolbar(context, this)
 
+    private val autocompleteExceptionHandler = CoroutineExceptionHandler { _, throwable ->
+        Log.log(
+            tag = logTag,
+            priority = Log.Priority.ERROR,
+            message = "Error while processing autocomplete input: $throwable"
+        )
+    }
```
pocmo: nit: `Logger` is way more fun than the low level API of `Log`. :)

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
```diff
@@ -50,11 +61,27 @@ class BrowserToolbar @JvmOverloads constructor(
...
     // displayToolbar and editToolbar are only visible internally and mutable so that we can mock
     // them in tests.
     @VisibleForTesting internal var displayToolbar = DisplayToolbar(context, this)
     @VisibleForTesting internal var editToolbar = EditToolbar(context, this)
 
+    private val autocompleteExceptionHandler = CoroutineExceptionHandler { _, throwable ->
+        Log.log(
+            tag = logTag,
+            priority = Log.Priority.ERROR,
+            message = "Error while processing autocomplete input: $throwable"
```
pocmo: Please pass in the `throwable` too so that we can get a stracktrace in the log!

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
```diff
@@ -14,17 +14,28 @@ import android.view.View
...
 import mozilla.components.concept.toolbar.Toolbar
 import mozilla.components.support.base.android.Padding
+import mozilla.components.support.base.log.Log
 import mozilla.components.support.ktx.android.content.res.pxToDp
 import mozilla.components.support.ktx.android.view.forEach
 import mozilla.components.support.ktx.android.view.isVisible
 import mozilla.components.ui.autocomplete.InlineAutocompleteEditText
+import java.util.concurrent.Executors
+
+private const val AUTOCOMPLETE_QUERY_THREADS = 3
```
pocmo: I wonder why we are using 3 threads here if we are just using one background thread and cancelling?

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
```diff
@@ -163,17 +190,51 @@ class BrowserToolbar @JvmOverloads constructor(
...
+            // irrelevant. We cancel them, but do not depend on cancellation to take place.
+            autocompleteSupervisorJob.cancelChildren()
+
+            // Our various autocomplete providers may perform expensive tasks to field our queries,
+            // such as disk access. Let's do the actual autocompletion work off the UI thread.
+            CoroutineScope(autocompleteDispatcher).launch {
+                filter.invoke(text, object : AutocompleteDelegate {
+                    override fun applyAutocompleteResult(result: AutocompleteResult) {
+                        // Ignore this result if the query is stale.
+                        if (result.input != latestAutocompleteInput) {
```
pocmo: Wouldn't it be safer to do this check when you are back on the UI thread (since that's where that value gets modified) **and** compare against the actual text in the widget instead of your cached value?

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
grigoryk: Cancellation is a cooperative process, and our providers might not cooperate. We can't depend on them honouring our cancel requests. So with a single background thread and a non-cooperative provider that's taking too long, we might get liveliness problems. Multiple threads are essentially an insurance policy against bad providers.

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
grigoryk: I think the current pattern is fine, although caching of the value does give me pause. Your suggestion is simpler to reason about, so I'll change this.

grigoryk:
All right, the desperately needed tests are now in place. Phew! @pocmo take another look?

PR #1926, opened 2019-02-01
[glean] Add a third-party library to the glean sample application

mdboom:
This is a proof-of-concept to show that we can now define metrics in a third-party library and glean can combine them across an integrated application.

I have tested the ping content manually, and I'm convinced this all works.  However, there are no unit tests here (there's actually none in `samples-glean` already), and I think writing those tests will actually require the glean testing API work that @travis79 is working on.  I'm not sure how best to handle that, so I'm filing this PR as a starting point for discussion.

(Also, this PR is based upon #1879, so contains parts of that, too).

In the meantime, this provides the basis for me to move forward with 1523977 to prevent the definition of the same metric in multiple places.

### Pull Request checklist


samples/glean/samples-glean-library/README.md
```diff
@@ -0,0 +1,4 @@
+sample_glean_library is an unnecessary third-party library that uses Glean to
+record metrics. It exists simply to show how libraries that are neither the
+application or Glean itself can record metrics, and those metrics will be stored
+and sent as part of the main pings that Glean provides.
```
Dexterp37: nit: let's make `main pings` a link to the ping docs. Also, glean should be all lowercase :) `glean`

samples/glean/samples-glean-library/metrics.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+$schema: moz://mozilla.org/schemas/glean/metrics/1-0-0
+
+sample_metrics:
+  test:
+    type: counter
+    description: >
+      A simple counter defined in a third-party library
+    send_in_pings:
+      - baseline
```
Dexterp37: Since this is an example, let's not put it into the `baseline` ping, to not set a dangerous precedent :) Let's omit this so that it shows up in the `metrics` ping and also call this out in the readme file.

mdboom: Sure.  It's easier to test the baseline ping because it goes out when you press the button -- you don't have to wait 24 hours :)  Maybe we should make the metrics ping go out then, too?

Dexterp37: Yeah, that's a good idea

Dexterp37: So, since this is blocked on [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1525333), I would still change this to `metrics` and eventually change the sample to trigger the "metrics" ping once that bug settles. Let's not block on this PR for that bug!

Dexterp37: Ok, we discussed this and Mike suggested a good compromise: let's leave this in the `metrics` ping but add a comment stating that, for testing this manually, it's better for it to be in the `baseline` ping.

PR #1931, opened 2019-02-01
Closes #1915: Memory leak caused by long-lived WebView instance

### LINKED ISSUE #1915: Memory leak: Session holds long-lived reference to SystemEngineView and its Activity Context
## FFTV background context
In FFTV, we separate our model from view state: model state is held by the `Application` while view state is held by `Activity`/`Fragments`. As part of the model, we hold our the `SessionManager` in `WebRenderComponents` which is held by the `Application`: the Engine, SessionManager, and other components are initialized with the application context. When our Activity is destroyed, e.g. under memory pressure, we keep our model and discard the view state. When the views are restored, we reactively push the data from the model to the view state (via architecture components style MVVM).

A consequence of this is that we keep the session manager, and its sessions, around for the duration of the `Application`. This is intentional: for as long as the application is alive, we don't see any reason to discard the user's browsing state, including tabs.

## Problem: the memory leak
However, we have the following reference hierarchy:
```
FirefoxApplication
- SessionManager (via WebRenderComponents)
 - Session
  - EngineSessionHolder
   - EngineSession (mutable reference but never cleared for lifecycle, only session add/removed)
     is a SystemEngineSession
    - NestedWebView (mutable reference but only written for window feature)
     - Activity Context
```

To emphasize, this means the Application is holding a long-lived reference to an Activity Context beyond the Activity's lifecycle: a common Android memory leak. Analysis with LeakCanary has backed up this theory.

We found this issue as part of the investigations in https://github.com/mozilla-mobile/firefox-tv/issues/1628

## Proposed solution
From a user's perspective, we want to keep their tabs around even when the system kills us to free up resources. Note: we could rely on session restore but that's unperformant and we haven't implemented it yet. As an application, that leaves us with two options:
- Fix the issue within the session: null the WebView reference when the Activity's lifecycle ends
- Fix the issue outside of the sessions: remove the sessions when the Activity's lifecycle ends and restore later (i.e. session restore)

Speaking with Christian, we noticed that `Session.link` and `unlink` remove the `EngineSession` from the `Session`, nulling the WebView reference. These are private so the app can't call them but `Session.add` and `remove` will call through to these methods. As such, we decided that at the end of the Activity's lifecycle, we can:
- Remove all the sessions, unlinking them from the SystemEngineView
- Cache the sessions in the application's memory
- Restore these cached sessions when the new Activity begins

Here's the code:
```kotlin
class WebRenderComponents {
    ...
    val cacheSessions = mutableListOf<Session>()
}

class MainActivity {
    ...
    override fun onStart() {
        with(application.webRenderComponents) {
            if (cacheSessions.size != 0) {
                cacheSessions.forEach {
                    sessionManager.add(it)
                }
                cacheSessions.clear()
            }
        }
    }

    override fun onStop() {
        with(application.webRenderComponents) {
            cacheSessions.addAll(sessionManager.all)
            sessionManager.removeAll()
        }
    }
}
```

Note: we still need to decide which lifecycle methods to use.

**We decided to test this code change in the application, in case it has bugs, for a few weeks before moving this implementation to components** because the application should not be handling this.

LeakCanary is catching other leaks so it's hard to say if it's fixed on LeakCanary alone. However, fixing this issue also fixed https://github.com/mozilla-mobile/firefox-tv/issues/1761, identifying that there is indeed a problem.

## Discarded solutions
The WebView javadocs recommend constructing the WebView with an Activity context and we do (in `WebViewCache.getWebView` when we create the `SystemEngineView`) so we can't pass in an `Application` context.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-1346)

csadilek:
- The WebView instance is long-lived, as it's referenced in
SystemEngineSession, and we want to be able to keep engine
sessions around for the lifespan of the application.

- The WebView instance is created by the engine and using its
context (the application context).

- The EngineView, rendering engine sessions and WebViews,
is created using an activity context.

- The problem was that when the activity and EngineView were
destroyed, the WebView was not being detached and still
referenced the parent which referenced the activity context,
which would then not get gc'ed.

- We now detach the WebView when the EngineView is destroyed
and only destroy the WebView when the session is closed.

- This will allow re-using of engine sessions and their
WebViews in different activities and allow for proper
destruction.

### Pull Request checklist


csadilek:
This is WIP. We need to test this with FireTV and likely need to make changes there e.g. the view is not created through the engine so the WebView instances there are created using the activity context?

csadilek:
Removing "work in progress" label. Looking at this again, I was wrong above about FireTV. It does use the same pattern where the engine is created with the application context and therefore  (System)EngineSessions and WebViews and created using the application context as well. So, this fix will apply there as well.

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -98,7 +98,7 @@ class SystemEngineView @JvmOverloads constructor(
 
     override fun onDestroy() {
         session?.apply {
-            webView.destroy()
+            (webView.parent as? SystemEngineView)?.removeView(webView)
```
mcomella: Please add a comment to explain why we're doing this: specifically, the cyclical parent-child reference and that some apps hold long-lived references to this view, and thus the WebView, via the session manager

mcomella: Please also explain why `destroy` should not be called and where it gets called instead.

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
```diff
@@ -154,6 +154,14 @@ class SystemEngineSession(
         notifyObservers { onTrackerBlockingEnabledChange(false) }
     }
 
+    /**
+     * See [EngineSession.close]
+     */
+    override fun close() {
+        super.close()
+        webView.destroy()
```
mcomella: Please explain why destroy is called here rather than as part of the lifecycle. Maybe mention destroying a WebView is okay to do here because each EngineSession holds its own WebView reference?

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
mcomella: I noticed we also call `removeView` in SystemEngineView.render:
https://github.com/mozilla-mobile/android-components/blob/a58f4da8606770094e59ba5e70fcce9fef0dbdc6/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt#L76

e.g. when switching tabs. I don't think there is, but maybe there's a way to consolidate these calls?

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
csadilek: Done.

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
csadilek: OK, added comment.

csadilek: Yeah, didn't find one that's a big improvement, but we can address if we look into #1945 .

PR #1933, opened 2019-02-02
Use stable IDs for awesomebar suggestions; retain suggestions on the screen 

grigoryk:
These patches make sure we have stable IDs coming out of the suggestion providers, configure the underlying recycler view to rely on stable IDs, and re-configures suggestion providers to not erase their suggestions from the screen as user is typing.

### Pull Request checklist

### DIFF
--- a/components/browser/awesomebar/src/main/java/mozilla/components/browser/awesomebar/BrowserAwesomeBar.kt
+++ b/components/browser/awesomebar/src/main/java/mozilla/components/browser/awesomebar/BrowserAwesomeBar.kt
@@ -80,6 +80,11 @@ class BrowserAwesomeBar @JvmOverloads constructor(
 
     @Synchronized
     override fun onInputStarted() {
+        // Make sure we're always displaying first suggestions at the top of the screen after input
+        // changes. Without this manual scroll, we might end with UI "scrolled" to a middle of the
+        // suggestions list.
+        scrollToPosition(0)
+
         providers.forEach { provider -> provider.onInputStarted() }
     }
 
@@ -87,7 +92,12 @@ class BrowserAwesomeBar @JvmOverloads constructor(
     override fun onInputChanged(text: String) {
         job?.cancel()
 
-        suggestionsAdapter.clearSuggestions()
+        // Make sure we're always displaying first suggestions at the top of the screen after input
+        // changes. Without this manual scroll, we might end with UI "scrolled" to a middle of the
+        // suggestions list.
+        scrollToPosition(0)
+
+        suggestionsAdapter.optionallyClearSuggestions()
 
         job = scope.launch {
             providers.forEach { provider ->
--- a/components/browser/awesomebar/src/main/java/mozilla/components/browser/awesomebar/SuggestionsAdapter.kt
+++ b/components/browser/awesomebar/src/main/java/mozilla/components/browser/awesomebar/SuggestionsAdapter.kt
@@ -24,6 +24,10 @@ internal class SuggestionsAdapter(
 ) : RecyclerView.Adapter<ViewHolderWrapper>() {
     internal var layout: SuggestionLayout = DefaultSuggestionLayout()
 
+    init {
+        setHasStableIds(true)
+    }
+
     /**
      * List of suggestions to be displayed by this adapter.
      */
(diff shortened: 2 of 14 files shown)


grigoryk:
This makes the awesomebar experience in r-b much smoother, and things are non longer flashing on the screen a lot time. With the stable IDs, we should be doing much less work underneath to paint the suggestions, as well.

grigoryk:
Ah, this needs a `getItemId` implementation!

grigoryk:
This still leaves poor behaviour of chips; I'll address those in a follow-up.

components/browser/awesomebar/src/main/java/mozilla/components/browser/awesomebar/SuggestionsAdapter.kt
```diff
@@ -95,8 +99,13 @@ internal class SuggestionsAdapter(
         return ViewHolderWrapper(layout.createViewHolder(awesomeBar, view, viewType), view)
     }
 
+    override fun getItemId(position: Int): Long {
+        val suggestion = suggestions[position]
+        return suggestion.id.hashCode().toLong()
```
pocmo: Isn't using `hashCode` problematic? It will guarantee **stable** ids. But it doesn't guarantee you **unique** ids. Two suggestions *could* have the same hashcode? And that would lead to weird bugs when updating the adapter?

grigoryk: You're correct. Currently this depends on the `id`s coming out of the providers to actually be uniue. They're prefixed with a provider identifier, and the actual ID is a URL usually, so they _should_ be unique at this point.

I agree that the current setup is a little brittle and leaks implementation details across components. It's also the "easy thing to do" here. An alternative would be to enforce uniqueness higher up the stack, closer to where it actually starts to matter. But at the end of the day, we're still depending on providers to give us good IDs to work with!

I think a reasonable thing to do here would be to document requirements around suggestion IDs clearly, static that they must include a provider identifier along with a unique-per-suggestion string (e.g. a url).

components/browser/awesomebar/src/main/java/mozilla/components/browser/awesomebar/BrowserAwesomeBar.kt
```diff
@@ -79,6 +79,11 @@ class BrowserAwesomeBar @JvmOverloads constructor(
 
     @Synchronized
     override fun onInputStarted() {
+        // Make sure we're always displaying first suggestions at the top of the screen after input
+        // changes. Without this manual scroll, we might end with UI "scrolled" to a middle of the
+        // suggestions list.
+        smoothScrollToPosition(0)
```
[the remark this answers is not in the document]
grigoryk: Is that something we're likely to do? Not sure how this would behave, tbh.

grigoryk:
@pocmo take another look, I've updated the interface to make things a bit clearer and "unique suggestion IDs" more likely to be observed.

components/browser/awesomebar/src/main/java/mozilla/components/browser/awesomebar/SuggestionsAdapter.kt
pocmo: Even with unique ids coming out of the providers we can end up with hashcode collisions, no? At the end of the day you are "mapping"/hashing Strings with "indefinite" length to a 64 bit Long.
https://android.googlesource.com/platform/libcore/+/refs/heads/master/ojluni/src/main/java/java/lang/String.java#1449

Given that hashCode aims for a uniformly random distribution of values it's probably unlikely but not impossible to have collisions?

grigoryk:
@pocmo i've updated this to use crc32 for hashing, which I think should fit our needs here. We don't have a lot of built-in choice of hashing functions that are not cryptographic (and are much slower). So IIUC, it's either CRC32, Adler32 (even less suitable as a hashing function), or rolling our own hashing schemes.

sblatz:
Bumping this thread. Would be nice to have this merged in for our Fenix demo on Friday :)

pocmo:
@sblatz I'll move the "scroll up" part to a separate PR.

pocmo:
Change of plans: I think we should land this and not "scroll up" earlier. Just scrolling up looks awful. I did make a small change to the code (removing the requirement that providers are on the hook for unique ids). From my side this is now ready to land. Let's see what taskcluster says.

ekager:
Thanks @pocmo! So is there a follow-up ticket for the scroll up investigation/alternative? Or do we think that it is something unrelated for https://github.com/mozilla-mobile/fenix/issues/443 ?

pocmo:
The "scroll up" code landed with this PR. I think mozilla-mobile/fenix#443 should be fixed with that. If not, then let's file a new issue. :)

sblatz:
Verified as fixed. Thanks @pocmo!

PR #1936, opened 2019-02-04
Closes #1934: Fix ContextMenu not working with Custom Tabs

### LINKED ISSUE #1934: Context Menus don't work in Custom Tabs
We need to pass in the sessionId to the feature.

Related: https://github.com/mozilla-mobile/reference-browser/issues/540

jonalmeida:
### Pull Request checklist

### DIFF
--- a/components/feature/contextmenu/src/main/java/mozilla/components/feature/contextmenu/ContextMenuFeature.kt
+++ b/components/feature/contextmenu/src/main/java/mozilla/components/feature/contextmenu/ContextMenuFeature.kt
@@ -32,8 +32,8 @@ internal const val FRAGMENT_TAG = "mozac_feature_contextmenu_dialog"
 class ContextMenuFeature(
     private val fragmentManager: FragmentManager,
     private val sessionManager: SessionManager,
-    private val candidates: List<ContextMenuCandidate>
-
+    private val candidates: List<ContextMenuCandidate>,
+    private val sessionId: String? = null
 ) : LifecycleAwareFeature {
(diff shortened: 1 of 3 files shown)


components/feature/contextmenu/src/main/java/mozilla/components/feature/contextmenu/ContextMenuFeature.kt
```diff
@@ -119,11 +119,16 @@ class ContextMenuFeature(
...
     private val feature: ContextMenuFeature
 ) : SelectionAwareSessionObserver(sessionManager) {
     override fun onLongPress(session: Session, hitResult: HitResult): Boolean {
         feature.onLongPress(session, hitResult)
         return false
     }
+
+    fun start(sessionId: String?) {
+        val session = sessionId?.let { sessionManager.findSessionById(sessionId) }
+        session?.let { observeFixed(it) } ?: observeSelected()
```
pocmo: nit: I guess this could have been just an if-else too, or not fancy enough? :D

jonalmeida: I was actually going to file an issue that we do this check for the `observeFixed`/`observeSelected` quite often for our features that we should abstract out off needing to write these lines over and over again.

I've also grown to dislike typing 'null' so I try to get away from it and let the compiler write that in the bytecode instead. 😅

pocmo: Heh. Optimize for reading not writing 😜

PR #1937, opened 2019-02-04
Fix 1524993: Handle Windows paths finding glean's Python install

mdboom:
The glean `build.gradle` file stores the path to its Python installation in a `buildConfig` variable so that its Python install can be used at testing time to test the pings it generates against a JSON schema.

The way this `buildConfig` variable was set was broken for Windows paths, because it would translate, e.g. `C:\Windows\Path` to the Java literal `"C:\Windows\Path"`, when it should be escaping the `\` characters as `"C:\\Windows\\Path"`.

The solution here is to use `JsonOutput.toJson` to produce a properly escaped string.  I had tried Groovy's builtin method to produce a "representation of a given object", `Object.inspect()`, but that produces a Groovy string, which can be inside single quotes, rather than a Java string with must use double quotes -- which is required since the string is being output verbatim to the generated `buildConfig.java` file.

There are other potential instances of this problem in the code base, but none using local file paths, so the problem is probably more theoretical there.

### Pull Request checklist


components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import org.apache.tools.ant.taskdefs.condition.Os
+import groovy.json.JsonOutput
 
 // This installs a Miniconda3 environment into the gradle user home directory
 // so that it will be shared between all libraries that use glean.  This is
```
Dexterp37: Let's also explain here why this change is required.

PR #1938, opened 2019-02-04
Closes #1935: Fix Downloads not working with Custom Tabs.

### LINKED ISSUE #1935: Downloads do not work in Custom Tabs
We need to pass in the sessionId to the feature.

Related: mozilla-mobile/reference-browser#538

jonalmeida:
### Pull Request checklist


components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt
```diff
@@ -53,7 +54,8 @@ class DownloadsFeature(
      * to be processed.
      */
     override fun start() {
-        super.observeSelected()
+        val session = sessionId?.let { sessionManager.findSessionById(sessionId) }
+        session?.let { observeFixed(it) } ?: observeSelected()
```
pocmo: Since that is exactly the same code as in the other PR. Maybe that could just be a method on `SelectionAwareSessionObserver` (which takes an optional session id)?

jonalmeida: Heh, left this comment in the other PR. I'll file an issue to do that.

PR #1939, opened 2019-02-04
[glean] Rename `setMetricsEnabled` to `setUploadEnabled`

Dexterp37:
The new denomination is less cryptic and clearly states what's the purpose of the function.

Please note that this will break consumers of the API, so they will need to be notified.

### Pull Request checklist

### DIFF
--- a/components/service/glean/README.md
+++ b/components/service/glean/README.md
@@ -17,7 +17,7 @@ Products using glean to collect telemetry **must**:
 - add documentation for any new metric collected with the library in its repository (see [an example](https://github.com/mozilla-mobile/android-components/blob/df429df1a193516f796f2330863af384cce820bc/components/service/glean/docs/pings.md));
 - go through data review for the newly collected data by following [this process](https://wiki.mozilla.org/Firefox/Data_Collection);
 - provide a way for users to turn data collection off (e.g. providing settings to control
-  `Glean.setMetricsEnabled()`).
+  `Glean.setUploadEnabled()`).
 
 ## Usage
 
@@ -97,8 +97,8 @@ by glean.
 
 ### Providing UI to enable / disable metrics
 
-Every application must provide a way to disable and re-enable metrics. This is
-controlled with the `glean.setMetricsEnabled()` method. The application should
+Every application must provide a way to disable and re-enable data collection and upload. This is
+controlled with the `glean.setUploadEnabled()` method. The application should
 provide some form of user interface to call this method.
 
 ## Contact
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/CommonMetricData.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/CommonMetricData.kt
@@ -47,7 +47,7 @@ interface CommonMetricData {
         }
 
         // Silently drop if metrics are turned off globally or locally
-        if (!Glean.getMetricsEnabled() || disabled) {
+        if (!Glean.getUploadEnabled() || disabled) {
             return false
         }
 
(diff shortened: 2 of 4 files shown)


chutten:
Bikeshedding time...

`setUploadEnabled()` controls both recording _and_ upload? Hm. This could be confused if consumers are coming from Desktop where there are two switches.

`setGleanEnabled()`? Or just `setEnabled()`? Are there operations that glean can/will perform if this bool is set to false?

Dexterp37:
> 
> 
> Bikeshedding time...
> 
> `setUploadEnabled()` controls both recording _and_ upload? Hm. This could be confused if consumers are coming from Desktop where there are two switches.

We have a [bug filed](https://bugzilla.mozilla.org/show_bug.cgi?id=1518503) for considering splitting things. However, looks like this might not be desirable on mobile.

> `setGleanEnabled()`? Or just `setEnabled()`? Are there operations that glean can/will perform if this bool is set to false?

I'd rather explicitly mention "upload" in the API, so that consumers will know exactly what's this about

georgf:
From a higher-level perspective:
This function guarantees that we won't upload data, just like on desktop.
Every other implementation side-effect this currently has is not guaranteed.

This being said, it is bike-shedding. Let's pick one and move on to other things.

Dexterp37:
> Travis:
> My vote is for `setEnabled`
> 
> `setGleanEnabled()` is redundant, since this will be called off the singleton named `Glean` -> `Glean.setEnabled()` seems more concise than `Glean.setGleanEnabled()`

> Mike: 
> I actually had the same reaction as @chutten. I think `setEnabled` may be clearer here since it controls both collection and uploading. But it's all bikeshedding, and I won't hold this up if I'm in the minority opinion there...
> 
> Implementation-wise, no issues.

I feel strongly about retaining the `setUploadEnabled` name, for these reasons:

- including the *upload* word makes it obvious what it guarantees (heh, re-using Georg's expression :D);
- we are still discussing about whether or not this should control only upload or upload and recording; we even need to discuss if there's value in controlling recording *at all*;
- once we make a call about the internals of that, we don't really want to enforce a new API change on consumers; worst case, they will still need to toggle upload on or off, even if we split the recording out and use a different function for that; this would allow us to be more light-hearted about changing the internals in the future. 

If nobody feels too strongly about this, I would go ahead and proceed with the current naming.

chutten:
Oh, if it only guarantees upload then the API name makes a lot of sense. 

But if that's the case we'll need to be specific about that in the README. Its assertion that 

> Every application must provide a way to disable and re-enable data collection and upload. This is
> controlled with the `glean.setUploadEnabled()` method. 

is the reason I was so confused that the API mentioned Upload only. :S

travis79:
@chutten I agree... if it only prevented upload, but right now it prevents both recording and uploading.  Every function that records a method is shielded by a call to [CommonMetricData.shouldRecord()](https://github.com/mozilla-mobile/android-components/blob/30077c34b49a98008f9bfcd78b33d48e10fd7628/components/service/glean/src/main/java/mozilla/components/service/glean/CommonMetricData.kt#L42) which uses this flag as a determining factor, and so naming of the function to `setUploadEnabled()`/`getUploadEnabled()` is still a misnomer.  

That being said, my feelings aren't _that_ strong on this, so I'm okay with changing it but it doesn't make it any clearer, in my mind.

Dexterp37:
> But if that's the case we'll need to be specific about that in the README. Its assertion that
> 
> > Every application must provide a way to disable and re-enable data collection and upload. This is
> > controlled with the `glean.setUploadEnabled()` method.
> 
> is the reason I was so confused that the API mentioned Upload only. :S

True, once we either decouple recording/uploading or completely remove disabling "recording" from that function. If you think of that, we're now only explicitly changing what's guaranteed by the function: the current name is `setMetricsEnabled`, which doesn't directly imply recording (which we might want to drop for consistency with desktop) nor upload. The new name is `setUploadEnabled`, which at least mentions the one guarantee we'll surely want to keep. Please keep in mind that changing this has implications downstream to the consumers :)

PR #1965, opened 2019-02-06
[glean] Add the `locale` field to the baseline ping

Dexterp37:
While this field was mentioned in the [ping docs](https://github.com/mozilla-mobile/android-components/blob/master/components/service/glean/docs/baseline.md) and also requested a [data-review](https://bugzilla.mozilla.org/show_bug.cgi?id=1512938#c3) for, this slipped through the cracks and its implementation never landed.

### Pull Request checklist

### DIFF
--- a/components/service/glean/metrics.yaml
+++ b/components/service/glean/metrics.yaml
@@ -102,6 +102,21 @@ glean.baseline:
     notification_emails:
       - telemetry-client-dev@mozilla.com
 
+  locale:
+    type: string
+    lifetime: application
+    send_in_pings:
+      - baseline
+    description: |
+      The locale of the application (e.g. "es-ES").
+    bugs:
+      - 1512938
+      - 1525540
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1512938#c3
+    notification_emails:
+      - telemetry-client-dev@mozilla.com
+
 glean.internal.metrics:
   client_id:
     type: uuid
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/ping/BaselinePing.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/ping/BaselinePing.kt
@@ -10,6 +10,7 @@ import android.content.Context
 import android.os.Build
 import mozilla.components.service.glean.GleanMetrics.GleanBaseline
 import mozilla.components.support.base.log.logger.Logger
+import java.util.Locale
 
 /**
  * BaselinePing facilitates setup and initialization of baseline pings.
@@ -43,6 +44,8 @@ internal class BaselinePing(applicationContext: Context) {
         ) ?.let {
             GleanBaseline.a11yServices.set(it)
         }
+
+        GleanBaseline.locale.set(getLanguageTag())
     }
 
     /**
@@ -67,6 +70,48 @@ internal class BaselinePing(applicationContext: Context) {
             // Note that any reference in java code can be null, so we'd better
             // check for null values here as well.
             it.id
+        }
+    }
+
+    /**
+     * Gets a gecko-compatible locale string (e.g. "es-ES" instead of Java [Locale]
+     * "es_ES") for the default locale.
+     *
+     * This method approximates the API21 method [Locale.toLanguageTag].
+     *
+     * @return a locale string that supports custom injected languages.
+     */
+    internal fun getLanguageTag(): String {
+        // Thanks to toLanguageTag() being introduced in API21, we could have
+        // simple returned `locale.toLanguageTag();` from this function. However
+        // what kind of languages the Android build supports is up to the manufacturer
+        // and our apps usually support translations for more rare languages, through
+        // our custom locale injector. For this reason, we can't use `toLanguageTag`
+        // and must try to replicate its logic ourselves.
+        val locale = Locale.getDefault()
+        val language = getLanguage(locale)
+        val country = locale.country // Can be an empty string.
+        return if (country.isEmpty()) language else "$language-$country"
+    }
+
+    /**
+     * Sometimes we want just the language for a locale, not the entire language
+     * tag. But Java's .getLanguage method is wrong. A reference to the deprecated
+     * ISO language codes and their mapping can be found in [Locale.toLanguageTag] docs.
+     *
+     * @param locale a [Locale] object to be stringified.
+     * @return a language string, such as "he" for the Hebrew locales.
+     */
+    internal fun getLanguage(locale: Locale): String {
+        // Can, but should never be, an empty string.
+        val language = locale.language
+
+        // Modernize certain language codes.
+        return when (language) {
+            "iw" -> "he"
+            "in" -> "id"
+            "ji" -> "yi"
+            else -> language
         }
     }
 }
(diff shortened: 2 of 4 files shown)


components/service/glean/metrics.yaml
```diff
@@ -102,6 +102,21 @@ glean.baseline:
     notification_emails:
       - telemetry-client-dev@mozilla.com
 
+  locale:
+    type: string
+    lifetime: application
+    send_in_pings:
+    - baseline
```
mdboom: Let's indent these as per the rest of the file.  (YAML doesn't require it, but I think it's clearer).

components/service/glean/src/main/java/mozilla/components/service/glean/ping/BaselinePing.kt
```diff
@@ -69,4 +72,45 @@ internal class BaselinePing(applicationContext: Context) {
...
+     * "es_ES") for the default locale.
+     *
+     * This method approximates the API21 method [Locale.toLanguageTag].
+     *
+     * @return a locale string that supports custom injected languages.
+     */
+    internal fun getLanguageTag(): String {
+        // Thanks to toLanguageTag() being introduced in API21, we could have
+        // simple returned `locale.toLanguageTag();` from this function. However
+        // what kind of languages the Android build support is up to the manufacturer
```
mdboom:
```suggestion
        // what kind of languages the Android build supports is up to the manufacturer
```

components/service/glean/src/main/java/mozilla/components/service/glean/ping/BaselinePing.kt
```diff
@@ -69,4 +72,45 @@ internal class BaselinePing(applicationContext: Context) {
...
+        // and our apps usually support translations for more rare languages, through
+        // our custom locale injector. For this reason, we can't use `toLanguageTag`
+        // and must try to replicate its logic ourselves.
+        val locale = Locale.getDefault()
+        val language = getLanguage(locale)
+        val country = locale.country // Can be an empty string.
+        return if (country.isEmpty()) language else "$language-$country"
```
mdboom: TIL: Linux seems to put an `_` between language and country, but it looks like the [IETF standard](https://tools.ietf.org/html/bcp47#section-2.1.1) is to use hypens, so :+1: to this.

components/service/glean/src/main/java/mozilla/components/service/glean/ping/BaselinePing.kt
```diff
@@ -69,4 +72,45 @@ internal class BaselinePing(applicationContext: Context) {
...
+        // Can, but should never be, an empty string.
+        val language = locale.language
+
+        // Modernize certain language codes.
+        return when (language) {
+            "iw" -> "he"
+            "in" -> "id"
+            "ji" -> "yi"
+            else -> language
+        }
```
mdboom: I assume this mapping is brought over from desktop...?  Do we want to reference it so we don't loose track of why this is needed?

components/service/glean/src/main/java/mozilla/components/service/glean/ping/BaselinePing.kt
```diff
@@ -69,4 +72,45 @@ internal class BaselinePing(applicationContext: Context) {
...
+     *
+     * This method approximates the API21 method [Locale.toLanguageTag].
+     *
+     * @return a locale string that supports custom injected languages.
+     */
+    internal fun getLanguageTag(): String {
+        // Thanks to toLanguageTag() being introduced in API21, we could have
+        // simple returned `locale.toLanguageTag();` from this function. However
+        // what kind of languages the Android build support is up to the manufacturer
+        // and our apps usually support translations for more rare languages, through
```
mdboom: Does this imply that the browser (or android-components) allows selection of a locale that isn't available on the system, or isn't the same as what the user has selected system-wide?  Just want to clarify because the code here is using the system's default (which I assume is what the user selected in the system-wide settings) and not some android-components-specific API.

components/service/glean/src/test/java/mozilla/components/service/glean/ping/BaselinePingTest.kt
```diff
@@ -47,4 +50,42 @@ class BaselinePingTest {
...
     }
+
+    @Test
+    fun `getLanguageTag() reports the tag for the default locale`() {
+        val baselinePing = BaselinePing(ApplicationProvider.getApplicationContext<Context>())
+        val defaultLanguageTag = baselinePing.getLanguageTag()
+
+        assertNotNull(defaultLanguageTag)
+        assertFalse(defaultLanguageTag.isEmpty())
+        assertEquals("en-US", defaultLanguageTag)
```
mdboom: Will this fail if the testing machine's locale is set to something non-English?  Should we instead mock `Locale.getDefault()`?

components/service/glean/metrics.yaml
Dexterp37: Gah, I hate it when Android Studio kills indentation when copy pasting.

components/service/glean/src/main/java/mozilla/components/service/glean/ping/BaselinePing.kt
Dexterp37: This whole code was ported over from the legacy Telemetry library that lives in a-c. I should have mentioned that, sorry. Here's the [original code](https://github.com/mozilla-mobile/android-components/blob/345f45448311d8c64f89e9c2a1fb435d2d46d853/components/service/telemetry/src/main/java/org/mozilla/telemetry/measurement/LocaleMeasurement.java#L48-L65). I've updated the docstring to mention that the mapping is available in `toLanguageTag` docs.

components/service/glean/src/main/java/mozilla/components/service/glean/ping/BaselinePing.kt
Dexterp37: As far as I understood from @pocmo , a-c can inject rare languages with its own language picker, that are then picked up by this. @pocmo is also reviewing this to make sure I didn't misunderstand :)

components/service/glean/src/test/java/mozilla/components/service/glean/ping/BaselinePingTest.kt
Dexterp37: Good point. I cross checked this, and turns out that we can't have anything other than "en-US" since we're runnign the tests in Robolectric. [See here](http://robolectric.org/javadoc/4.0/org/robolectric/shadows/ShadowLocaleData.html).

components/service/glean/src/main/java/mozilla/components/service/glean/ping/BaselinePing.kt
pocmo: Yeah, correct. In order to support languages that are not installed on the system (our apps have a custom language setting) we inject a different Locale at runtime. That's the code in Focus:
https://github.com/mozilla-mobile/focus-android/blob/master/app/src/main/java/org/mozilla/focus/locale/LocaleManager.java

We _could_ test and verify how `toLanguageTag()` behaves with that and then just use it here. But it may be hard to make an assumption over all variations of Android we may see.

PR #1973, opened 2019-02-06
[glean]Use WorkManager for uploading glean pings

travis79:
### Pull Request checklist


travis79:
This is still work in progress, mostly because I found a limitation on how I was passing data to the WorkManager worker that restricts the data to 10240 bytes.  This probably isn't an issue for baseline pings but I have concerns with events and metrics pings exceeding this size.  If that is the case, another method of serializing the ping payload that is accessible to the worker may be required.

Dexterp37:
> 
> 
> This is still work in progress, mostly because I found a limitation on how I was passing data to the WorkManager worker that restricts the data to 10240 bytes because it relies on [androidx.work.Data](https://developer.android.com/reference/androidx/work/Data) object.

I think this requires a bit of discussion around the design: we probably want to follow a similar approach to how the legacy telemetry library was handling things. This would allow us to work around the size limitation by serializing pings to disk.

> This probably isn't an issue for baseline pings but I have concerns with events and metrics pings exceeding this size. If that is the case, another method of serializing the ping payload that is accessible to the worker may be required.

Yeah, that depends on how the work about "extending the baseline ping goes". I doubt we'll ever go over 10kB with the baseline ping, but we will probably do for other pings.

pocmo:
> > This is still work in progress, mostly because I found a limitation on how I was passing data to the WorkManager worker that restricts the data to 10240 bytes because it relies on [androidx.work.Data](https://developer.android.com/reference/androidx/work/Data) object.
> 
> I think this requires a bit of discussion around the design: we probably want to follow a similar approach to how the legacy telemetry library was handling things. This would allow us to work around the size limitation by serializing pings to disk.

Yeah, the old library writes pings to disk and then schedules an upload job. There will always only be one job scheduled. Once the job runs it uploads all pings that are waiting on disk.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -175,13 +180,36 @@ open class GleanInternalAPI {
...
             val uuid = UUID.randomUUID()
             val path = makePath(docType, uuid)
-            // Asynchronously perform the HTTP upload off the main thread.
-            GlobalScope.launch(Dispatchers.IO) {
-                httpPingUploader.upload(path = path, data = pingContent)
-            }
+
+            // This holds input data for the WorkManager to access when it runs the worker task
+            // It's *important* to note Data has a hard size limit of Data.MAX_DATA_BYTES which
+            // has a constant value of 10240 according to the documentation.
```
mdboom: This seems awfully small.  What's the solution -- serialize the ping content to disk and use it right before sending?

(I vaguely recall this already being discussed, so sorry if I'm retreading ground).

mdboom: Oh geeze -- this discussion is right up there.  Ignore me...  We're all on the same page :)

travis79: I'm working on spreading out the architecture to a ping storage engine to separate the serialization and the upload worker.  The hard part of this is testing WorkManager class, it's a bit new and the test helper library is still in beta and lacking documentation that I can find.  The best I've been able to do so far is to try and copy a minimal example embedded in the example for the WorkManager provided by the Android devs.

travis79:
Okay, I finally feel pretty good about the test coverage and functionality.  I was able to pull most/all of the scheduling code out of Glean.kt and implement it into the PingScheduler class (which also assumed the role of the LifecycleObserver), as well as implement a PingStorageEngine for managing writing the pings to file.  This change in things forced me to do a lot of changes to the way we test the pings and unfortunately its now much harder(impossible?) to test using a mock http server where we were using it so it now tests that the work gets scheduled and that the worker does what it's supposed to do working with the HttpUploader class.

I expect some criticism/feedback over the size/breadth of the refactor so please don't hold back.  I think that overall this cleans up the things that needed cleaned up and puts things where they should be.  I'll be leaving this a WIP until everyone weighs in.

travis79:
Gah, this is so weird.  All tests pass locally but I'm randomly getting different test in taskcluster that fail.  I've done this twice now, first time 12 tests failed, this time 2 tests failed (all in service-glean).  Going to try it again and see what happens...

components/service/glean/src/main/java/mozilla/components/service/glean/storages/PingStorageEngine.kt
```diff
@@ -0,0 +1,133 @@
...
+        private const val STORAGE_DIRECTORY = "glean/storage"
+    }
+
+    @Synchronized
+    fun store(pingType: String, pingPath: String, ping: String) {
+        val pingStorageDirectory = File(storageDirectory, pingType)
+        assertDirectory(pingStorageDirectory)
+
+        // Build the file path for the ping, using the UUID from the path for the file name
+        val pingFile = File(pingStorageDirectory, pingPath.substring(pingPath.lastIndexOf('/')))
```
mdboom: Maybe pass in the UUID and use it directly.  I feel like parsing the URL path might a bit brittle if the backend ever changes how they want paths to be handled.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/PingStorageEngine.kt
```diff
@@ -0,0 +1,133 @@
...
+ */
+internal open class PingStorageEngine(context: Context) {
+
+    private val logger: Logger = Logger("glean/PingStorageEngine")
+
+    @VisibleForTesting
+    internal val storageDirectory: File = File(context.applicationInfo.dataDir, STORAGE_DIRECTORY)
+
+    companion object {
+        private const val FILE_PATTERN = "[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}"
```
mdboom: Maybe just a comment here that filenames are UUIDs, and that's what this regex matches.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/PingStorageEngine.kt
```diff
@@ -0,0 +1,133 @@
...
+                if (processed) {
+                    file?.delete()
+                } else {
+                    // The callback couldn't process this file.
+                    return false
+                }
+            } catch (e: FileNotFoundException) {
+                // This shouldn't happen after we queried the directory.
+            } catch (e: IOException) {
+                // Something is not right.
```
mdboom: Let's log these errors.  Down the road, we may want to record metrics about them, too, so maybe add a TODO?

mdboom: On second thought, if this is failing, we're unlikely to get any metrics out of this client anyway, so maybe not worth recording.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/PingStorageEngine.kt
```diff
@@ -0,0 +1,133 @@
...
+    private fun assertDirectory(directory: File) {
+        if (!directory.exists() && !directory.mkdirs()) {
+            throw IllegalStateException(
+                "Directory doesn't exist and can't be created: " + directory.absolutePath)
+        }
+
+        if (!directory.isDirectory || !directory.canWrite()) {
+            throw IllegalStateException(
+                "Directory is not writable directory: " + directory.absolutePath)
+        }
```
mdboom: I know we've been pretty shy about exceptions.  Can you confirm my assumption that because this is called from a Worker job that all this would do is essentially cancel the worker job, but the exception wouldn't bubble up to the application?

components/service/glean/src/test/java/mozilla/components/service/glean/EventMetricTypeTest.kt
```diff
@@ -30,9 +29,7 @@ class EventMetricTypeTest {
 
     @Before
     fun setUp() {
-        Glean.initialized = true
-        Glean.applicationId = "test"
-        Glean.configuration = Configuration()
+        resetGlean()
```
mdboom: I just filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1527665) about cleaning up all of these `@Before` functions.  I guess we were thinking alike!

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
```diff
@@ -105,77 +134,7 @@ class GleanTest {
...
-                    .getJSONObject("string")
-                    .getString("telemetry.string_metric")
-            )
-            assertNull(metricsJson.opt("events"))
-            assertNotNull(metricsJson.opt("ping_info"))
-            assertNotNull(metricsJson.getJSONObject("ping_info").opt("experiments"))
-            val applicationId = "mozilla-components-service-glean"
-            assert(
-                request.path.startsWith("/submit/$applicationId/metrics/${Glean.SCHEMA_VERSION}/")
-            )
```
mdboom: I worry that the validation of the actual ping content is being lost here.  (Unless it was moved somewhere and I just can't find it).  I don't care too much about testing with a mock HTTP server and making sure pings get sent over HTTP (we can assume the HTTP client is working -- we didn't write it :wink:).  But it would be good to check that:

1) If an event is triggered, the expected pings are queued (since there might be more than one).
2) The contents of the pings contains what we expect them too
3) The contents of the pings validates against the schema

If an event is fired that generates and schedules some pings, is it true that those pings would be sitting on disk if there's nothing in the test that explicitly sends them out?  If so, maybe it would it be possible to add a testing-only function to the `PingStorageEngine` that returns pings that have been queued.  Something like:

`testGetQueuedPing(pingName: string): List<String>` (or `List<JSONObject>`)

This would go through all of the queued pings for the given ping type and return their contents.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -236,52 +181,15 @@ open class GleanInternalAPI internal constructor () {
...
-     *
-     *   - Background: When the application goes to the background.
-     *       This triggers sending of the baseline ping.
-     *   - Default: Event that triggers the default pings.
-     *
-     * @param pingEvent The type of the event.
+     * Helper function to force scheduling of pings immediately. Note that the 'metrics' ping
+     * schedule is still dependent on it's internal schedule of no more than once every 24 hours.
      */
-    fun handleEvent(pingEvent: Glean.PingEvent) {
```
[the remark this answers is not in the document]
travis79: I tried to allow this with the `schedulePingsNow()` function in Glean.kt.  I really wasn't sure about the best way to approach this but I knew the sample app would need a way to force that when pressing the button, and perhaps useful elsewhere.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/PingStorageEngine.kt
travis79: This actually gets called from the scheduler which runs in the client process when the ping is serialized, not the worker process.  I copied some of this from the previous telemetry ping serialization and this does bear changing to follow the 'no throw' policy of glean.

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
travis79: Some of this is broken out into other places now. You may have noticed the `isWorkScheduled()` helper function I use in the `PingSchedulerTest`.  What we did lose was the checking of the ping contents and schema.  I totally agree with the idea of sanity checking those and I'm working on a way to do that now.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/PingStorageEngine.kt
```diff
@@ -0,0 +1,178 @@
...
+            // Build the file path for the ping, using the UUID from the path for the file name
+            val pingFile = File(pingStorageDirectory, uuidFileName.toString())
+
+            var stream: FileOutputStream? = null
+
+            try {
+                stream = FileOutputStream(pingFile, true)
+
+                val writer = BufferedWriter(OutputStreamWriter(stream))
+                writer.write(pingPath)
```
Dexterp37: Can we comment on what this is doing? Did you check with @pocmo why were they doing this and if there's any drawback?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/PingStorageEngine.kt
```diff
@@ -0,0 +1,178 @@
...
+     * ping type
+     *
+     * @param pingType String representing the ping type: baseline, events, metrics (or custom ping)
+     * @param callback Callback function to do the actual process action on the ping.  Typically
+     *                 this will be [HttpPingUploader.upload]
+     * @return Boolean representing the success of the upload task. This may be the value bubbled up
+     *         from the callback, or a local value if there was an error reading the files. The
+     *         [PingUploadWorker] uses this to determine whether or not to retry or declare success.
+     */
+    fun process(pingType: String, callback: (String, String) -> Boolean): Boolean {
```
Dexterp37: nit: callback->processingCallback?

Dexterp37: Why do we need to process per ping type? What's the advantage over processing all the files in a flat structure?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/PingStorageEngine.kt
```diff
@@ -0,0 +1,178 @@
...
+        return true
+    }
+
+    /**
+     * Helper function to list all ping files for a particular ping type
+     *
+     * @param pingType String representing the ping type: baseline, events, metrics (or custom ping)
+     * @return An Array<File?> containing the list of ping files
+     */
+    @VisibleForTesting
```
Dexterp37: nit: either the visibility or internal

components/service/glean/src/main/java/mozilla/components/service/glean/storages/PingStorageEngine.kt
```diff
@@ -0,0 +1,178 @@
...
+    internal fun listPingFiles(pingType: String): Array<File?> {
+        val pingStorageDirectory = File(storageDirectory, pingType)
+
+        val uuidPattern = Pattern.compile(FILE_PATTERN)
+
+        val uuidFilenameFilter = FilenameRegexFilter(uuidPattern)
+        return pingStorageDirectory.listFiles(uuidFilenameFilter) ?: return arrayOfNulls(0)
```
Dexterp37: Instead o returning an array of nulls, can we return a nullable array and check at call site?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/PingStorageEngine.kt
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    /**
+     * Helper function to determine if the ping directory exists which attempts to create them if
+     * they don't
+     *
+     * @param directory File representing the directory path in which to store pings.
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    fun assertDirectory(directory: File) {
```
Dexterp37: nit: can we call this `ensureDirectoryExists`?

components/service/glean/src/test/java/mozilla/components/service/glean/TestLifeCycleOwner.kt
```diff
@@ -0,0 +1,18 @@
+package mozilla.components.service.glean
```
Dexterp37: This is missing the license header.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/PingStorageEngine.kt
```diff
@@ -0,0 +1,178 @@
...
+     * Function to store the ping to file for the [PingUploadWorker] to read and send when it is run
+     *
+     * @param pingType String representing the ping type: baseline, events, metrics (or custom ping)
+     * @param pingPath String representing the upload path used for uploading the ping
+     * @param ping Serialized JSON string representing the ping payload
+     */
+    @ObsoleteCoroutinesApi
+    @Synchronized
+    fun store(pingType: String, uuidFileName: UUID, pingPath: String, ping: String) {
+        Dispatchers.API.launch {
```
Dexterp37: Let's not use the API dispatchers, since this is IO. We could use [this](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-dispatchers/-i-o.html)

components/service/glean/src/main/java/mozilla/components/service/glean/storages/PingStorageEngine.kt
```diff
@@ -0,0 +1,178 @@
...
+
+    /**
+     * Function to store the ping to file for the [PingUploadWorker] to read and send when it is run
+     *
+     * @param pingType String representing the ping type: baseline, events, metrics (or custom ping)
+     * @param pingPath String representing the upload path used for uploading the ping
+     * @param ping Serialized JSON string representing the ping payload
+     */
+    @ObsoleteCoroutinesApi
+    @Synchronized
```
Dexterp37: Why do we need this to be synchronized?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/PingStorageEngine.kt
```diff
@@ -0,0 +1,178 @@
...
+        Dispatchers.API.launch {
+            val pingStorageDirectory = File(storageDirectory, pingType)
+            assertDirectory(pingStorageDirectory)
+
+            // Build the file path for the ping, using the UUID from the path for the file name
+            val pingFile = File(pingStorageDirectory, uuidFileName.toString())
+
+            var stream: FileOutputStream? = null
+
+            try {
```
Dexterp37: Instead of using try, we could use kotlin-use. See [here](https://stackoverflow.com/a/26979481/261698)

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/PingUploadWorker.kt
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * This method is called on a background thread - you are required to **synchronously** do your
+     * work and return the [androidx.work.ListenableWorker.Result] from this method.  Once you
+     * return from this method, the Worker is considered to have finished what its doing and will be
+     * destroyed.
+     *
+     * A Worker is given a maximum of ten minutes to finish its execution and return a
+     * [androidx.work.ListenableWorker.Result].  After this time has expired, the Worker will
+     * be signalled to stop.
```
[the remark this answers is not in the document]
travis79: there is another override method `onStopped()` where we could do this.  The function docs state that we can't do anything time consuming in this function but we should have enough time to log an error or something.

components/service/glean/src/test/java/mozilla/components/service/glean/EventMetricTypeTest.kt
travis79: I went ahead and cleaned them up here (the ones I could find). I hope that's okay.  If it is, we can point that bug at this PR and close it when it lands.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -8,64 +8,44 @@ import android.arch.lifecycle.ProcessLifecycleOwner
...
     // This object encapsulates initialization and information related to the baseline ping
     internal lateinit var baselinePing: BaselinePing
 
     /**
      * Initialize glean.
      *
-     * This should only be initialized once by the application, and not by
-     * libraries using glean. A message is logged to error and no changes are made
-     * to the state if initialize is called a more than once.
+     * This should only be initialized once by the application, and not by libraries using glean.
```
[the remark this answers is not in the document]
travis79: I thought I had messed up the formatting, I can revert.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -236,52 +181,16 @@ open class GleanInternalAPI internal constructor () {
...
-                sendPing("events", "events")
-            }
-            Glean.PingEvent.Default -> {
-                if (metricsPingScheduler.canSendPing()) {
-                    sendPing(MetricsPingScheduler.STORE_NAME, MetricsPingScheduler.STORE_NAME)
-                    metricsPingScheduler.updateSentTimestamp()
-                }
-            }
-        }
+    fun schedulePingsNow() {
```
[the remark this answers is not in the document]
travis79: Purely to support the sample application.  I can mark it as internal or remove it entirely but we may have to refactor the samples-glean to indicate that the pings will only be sent when the app is backgrounded.

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/PingScheduler.kt
```diff
@@ -0,0 +1,164 @@
...
+/**
+ * This class is responsible for scheduling of glean pings. In order to do this in association
+ * with lifecycle events, it implements [LifecycleObserver].  On entering the background
+ * [Lifecycle.Event.ON_STOP] state, it will use [WorkManager] to schedule a [PingUploadWorker] task
+ * which is responsible for uploading the ping.
+ */
+internal class PingScheduler(
+    applicationContext: Context,
+    storageEngineManager: StorageEngineManager
+) : LifecycleObserver {
```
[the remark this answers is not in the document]
travis79: It made sense to have them internal to the `PingScheduler` since that was the only place most of them were used, aside from the `PingUploadWorker` needing access to the `pingStorageEngine` in order to process the files.  I'm open to moving them back if it turns out they are better suited to live somewhere else.

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/PingScheduler.kt
```diff
@@ -0,0 +1,164 @@
...
+import java.util.UUID
+import java.util.concurrent.TimeUnit
+
+/**
+ * This class is responsible for scheduling of glean pings. In order to do this in association
+ * with lifecycle events, it implements [LifecycleObserver].  On entering the background
+ * [Lifecycle.Event.ON_STOP] state, it will use [WorkManager] to schedule a [PingUploadWorker] task
+ * which is responsible for uploading the ping.
+ */
+internal class PingScheduler(
```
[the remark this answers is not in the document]
travis79: No, I can make it work leaving this in Glean.kt, but one of the things we discussed in our [notes](https://docs.google.com/document/d/1D6jACr9suJQ2y6FSonrHPifMZoOKYJ0DcB-DuU1IE5Y) was creating a class to handle the interaction with `WorkManager`

components/service/glean/src/main/java/mozilla/components/service/glean/storages/PingStorageEngine.kt
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    /**
+     * Function to store the ping to file for the [PingUploadWorker] to read and send when it is run
+     *
+     * @param pingType String representing the ping type: baseline, events, metrics (or custom ping)
+     * @param pingPath String representing the upload path used for uploading the ping
+     * @param ping Serialized JSON string representing the ping payload
+     */
+    @ObsoleteCoroutinesApi
```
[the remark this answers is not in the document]
travis79: The IDE complained with a warning, said it should be marked with this annotation

components/service/glean/src/main/java/mozilla/components/service/glean/storages/PingStorageEngine.kt
travis79: It's serializing the upload path on the first line of the file, and then it serializes the payload on the second line of the file.  I suppose it would be able to to recreate the path when it deserialized the ping, based on the filename and the ping's docType.  This was just how Telemetry lib did it so I followed suit.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/PingStorageEngine.kt
travis79: I thought it would be much easier to deal with the ping types (baseline, events, metrics) separately to reduce the IO overhead of a single worker.  I figured the OS would be less likely to kill 3 shorter running worker processes than one longer one.  If this was a single worker process and it exceptioned on a particular ping upload and got cancelled, we may lose or never upload valid pings.  Separating the work across more workers seemed like a way to isolate that and give us a better chance of getting the data to the server.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/PingStorageEngine.kt
travis79: Certainly!  This was another artifact from duplicating what previous Telemetry was doing.  Much more kotlin-esqe to return a nullable Array<>.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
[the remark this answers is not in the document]
Dexterp37: Let's leave this as you did, but also file a bug to make it internal and remove manual ping sending from the sample app. There shouldn't be any special behaviour in the sample app compared to other apps.

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/PingScheduler.kt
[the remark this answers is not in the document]
Dexterp37: What I mean is we can inject them as dependencies through the constructor, making testing easier, rather than relying on interna, e.g. `PingScheduler(context, StorageEngineManager(), PingMaker(), ...)`

components/service/glean/src/main/java/mozilla/components/service/glean/storages/PingStorageEngine.kt
Dexterp37: Sorry I was cryptic. I meant, can you *add* a comment explaining that in the code? :)

components/service/glean/src/main/java/mozilla/components/service/glean/storages/PingStorageEngine.kt
Dexterp37: I see. However, we wouldn't be uploading pings until a new ping is enqueued (e.g. when we get to background), right? Moreover, we don't really expect to have tons of pending pings, that would likely be a sign of something going wrong. Anyway, I'd leave this up to you to decide/figure out :)

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/PingScheduler.kt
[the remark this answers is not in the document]
travis79: Ahh yes, I see where you are going now.  Yes, absolutely.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/PingStorageEngine.kt
travis79: I think we still want the try/catch for the call to `write()`, but I have no problem with doing this with `.use {}`

components/service/glean/src/main/java/mozilla/components/service/glean/storages/PingStorageEngine.kt
travis79: I need a CoroutineScope to call `launch` from, but I can use the `Dispatchers.IO` context by passing it into the `launch()` function.

travis79: The Dispatchers.IO must run at a lower priority than the Default CoroutinesContext used by the API dispatchers.  

I am now getting more and more async race conditions interfering with testing of this class, even locally.  It all works for the samples-glean app just fine, but trying to check if something was stored by reading the directory immediately after calling `store()` is failing fairly consistently in tests.  If I return the `Job` object from `launch()` and then use `invokeOnCompletion` handlers, it works flawlessly every time.  Looking for better solution...

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/PingScheduler.kt
```diff
@@ -0,0 +1,206 @@
...
+
+        // Do not send ping if upload is disabled
+        if (!Glean.getUploadEnabled()) {
+            logger.debug("Glean upload not enabled, abort serialize and send ping.")
+            return
+        }
+
+        // Schedule the baseline and event pings and save the result to determine whether or not
+        // we need to schedule the upload worker
+        var pingsToSchedule = serializePing(BaselinePing.STORE_NAME)
```
Dexterp37: nit: this variable is not telling us which pings to schedule, but whether or not *any* ping was assembled/is ready to be sent. Maybe `availableForSending` is a more suitable name?

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/PingScheduler.kt
```diff
@@ -0,0 +1,206 @@
...
+        // metrics ping or not
+        if (metricsPingScheduler.canSendPing()) {
+            pingsToSchedule = pingsToSchedule || serializePing(MetricsPingScheduler.STORE_NAME)
+            metricsPingScheduler.updateSentTimestamp()
+        }
+
+        // It should only take a single PingUploadWorker to process all queued pings, so we only
+        // want to have one scheduled at a time.
+        if (pingsToSchedule && !isWorkScheduled()) {
+            WorkManager.getInstance().enqueue(buildWorkRequest())
```
Dexterp37: I think we need to make sure that this happens *after* pings are written to disk. Otherwise, we might have some delay in the ping submission. Think of this case:

- we assemble ping A and async write it to disk
- we enqueue/start the work request but A wasn't written to disk yet
- the worker goes through the files, doesn't find anything and bails out
- the worker is enqueued again next time a ping is sent (which could be a bit away in time)

Let's *not* address this problem in this bug. Let's file a bug, add a comment here with the bug number, and address it later, if required by data quality analysis.

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/PingUploadWorker.kt
```diff
@@ -0,0 +1,43 @@
...
+     *
+     * @return The [androidx.work.ListenableWorker.Result] of the computation
+     */
+    override fun doWork(): Result {
+        val httpPingUploader = HttpPingUploader()
+
+        if (!Glean.pingScheduler.pingStorageEngine.process(httpPingUploader::upload)) {
```
Dexterp37: I think we need to file yet another bug for follow-up work :) If the application toggles upload off, we should kill the worker and delete all the pending pings (probably). Let's not tackle this now, as it will require discussion, but still file a bug.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/PingStorageEngine.kt
```diff
@@ -0,0 +1,180 @@
...
+        }
+    }
+
+    /**
+     * Helper function to list all serialized ping files or returns null if none
+     *
+     * @return An Array<File?> containing the list of ping files
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    fun listPingFiles(): Array<File>? {
```
Dexterp37: nit: does having this function buy us anything now? Can't we simply use `storageDirectory.lsitFiles()` at the call site?

components/service/glean/src/test/java/mozilla/components/service/glean/TimespanMetricTypeTest.kt
```diff
@@ -1,7 +1,5 @@
 package mozilla.components.service.glean
```
Dexterp37: GAH! Not your fault, but this is missing the license header too :'( Any chance you could add it?

components/service/glean/src/test/java/mozilla/components/service/glean/storages/PingStorageEngineTest.kt
```diff
@@ -0,0 +1,176 @@
...
+import org.robolectric.RobolectricTestRunner
+import java.io.File
+import java.io.FileReader
+import java.io.BufferedReader
+import java.io.FileOutputStream
+import java.util.UUID
+
+@RunWith(RobolectricTestRunner::class)
+class PingStorageEngineTest {
+    // Filenames and paths to test with
```
Dexterp37: nit: also mention that they will be generated before each test runs

components/service/glean/src/test/java/mozilla/components/service/glean/storages/PingStorageEngineTest.kt
```diff
@@ -0,0 +1,176 @@
...
+        fileNames = arrayOf(fileName1.toString(), fileName2.toString(), fileName3.toString())
+        pathNames = arrayOf(path1, path2, path3)
+    }
+
+    @Test
+    fun `storage engine correctly stores pings to file`() {
+        runBlocking(Dispatchers.IO) {
+            pingStorageEngine.store(fileName1, path1, "dummy data")
+        }
+        Thread.sleep(10)
```
Dexterp37: Let's not have `Thread.sleep` stuff, as it will make tests intermittent if CI is under pressure. Let's investigate why this isn't working. Here and in the other tests below.

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/PingScheduler.kt
```diff
@@ -0,0 +1,206 @@
...
+/**
+ * This class is responsible for scheduling of glean pings. In order to do this in association
+ * with lifecycle events, it implements [LifecycleObserver].  On entering the background
+ * [Lifecycle.Event.ON_STOP] state, it will use [WorkManager] to schedule a [PingUploadWorker] task
+ * which is responsible for uploading the ping.
+ */
+internal class PingScheduler(
+    applicationContext: Context,
+    storageEngineManager: StorageEngineManager,
+    internal var pingMaker: PingMaker = PingMaker(storageEngineManager, applicationContext),
```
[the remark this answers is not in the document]
travis79: It was mostly to have access to them for testing.  Let me move the assignment out of the constructor so I can mark them with `@VisibleForTesting` instead.

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/PingScheduler.kt
```diff
@@ -0,0 +1,206 @@
...
+         */
+        internal fun isWorkScheduled(): Boolean {
+            val instance = WorkManager.getInstance()
+            val statuses = instance.getWorkInfosByTag(PingUploadWorker.PING_WORKER_TAG)
+            return try {
+                var running = false
+                val workInfoList = statuses.get()
+                for (workInfo in workInfoList) {
+                    val state = workInfo.state
+                    running = (state === WorkInfo.State.RUNNING) or (state === WorkInfo.State.ENQUEUED)
```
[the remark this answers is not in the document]
travis79: I think it went awry when I attempted to lift the return out of the try/catch.  I believe I have it fixed up with this next push

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/PingScheduler.kt
travis79: Simple enough, will file bug about it.

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/PingUploadWorker.kt
travis79: Will do!

components/service/glean/src/main/java/mozilla/components/service/glean/storages/PingStorageEngine.kt
travis79: True

components/service/glean/src/test/java/mozilla/components/service/glean/storages/PingStorageEngineTest.kt
travis79: Found a better way in a trick I learned from PR #2120

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
```diff
@@ -283,28 +249,74 @@ class GleanTest {
...
+            pingList?.forEach { file ->
+                val bufferedReader = BufferedReader(FileReader(file))
+                val path = bufferedReader.readLine()
+                val serializedPing = bufferedReader.readLine()
+
+                // Compare the path UUID to the filename
+                assertEquals(path.substring(path.lastIndexOf("/") + 1), file.name)
+                // Validate the ping schema
+                checkPingSchema(serializedPing)
+            }
```
mdboom: Awesome.  Thanks for providing a way to check ping contents.  I wonder if we should provide a test helper function to handle the loading of contents (i.e. the first three lines in this block).

(Also, I liked it better before when the `PingStorageEngine` abstracted away how to get the individual files, but that's *extreme* bikeshedding).

travis79:
@Dexterp37 ready for another look.  I believe I have addressed all of the changes you requested.

travis79:
FYI The merging in of the async await stuff for the test API functions helped the flakiness with passing of tests.

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/PingScheduler.kt
```diff
@@ -0,0 +1,213 @@
...
+    // The application id detected by glean to be used as part of the submission
+    // endpoint.
+    internal var applicationId = sanitizeApplicationId(applicationContext.packageName)
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    val pingMaker: PingMaker = pingMaker
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    val pingStorageEngine = pingStorageEngine
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    val metricsPingScheduler = metricsPingScheduler
```
pocmo: nit: You can mark them as `val` inside the constructor above. (maybe `internal` too?)

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/PingScheduler.kt
```diff
@@ -0,0 +1,213 @@
...
+        // metrics ping or not
+        if (metricsPingScheduler.canSendPing()) {
+            availableToSend = availableToSend || assembleAndSerializePing(MetricsPingScheduler.STORE_NAME)
+            metricsPingScheduler.updateSentTimestamp()
+        }
+
+        // It should only take a single PingUploadWorker to process all queued pings, so we only
+        // want to have one scheduled at a time.
+        if (availableToSend && !isWorkScheduled()) {
+            WorkManager.getInstance().enqueue(buildWorkRequest())
```
pocmo: With the `isWorkScheduled` check it looks like you've rebuild [enqueueUniqueWork](https://developer.android.com/reference/androidx/work/WorkManager.html#enqueue(androidx.work.WorkRequest))?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/PingStorageEngine.kt
```diff
@@ -0,0 +1,180 @@
...
+
+//    /**
+//     * Helper function to list all serialized ping files or returns null if none
+//     *
+//     * @return An Array<File?> containing the list of ping files
+//     */
+//    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+//    fun listPingFiles(): Array<File>? {
+//        return storageDirectory.listFiles()
+//    }
```
pocmo: nit: Can this be deleted?

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/PingScheduler.kt
travis79: Sweet!  Yes it does, and that allows me to move the `isWorkScheduled` to be a test only function in TestUtil.kt   Much better!

PR #1977, opened 2019-02-07
Fixes #1976: Fenix needs SessionUseCases which can always use a new tab

### LINKED ISSUE #1976: Fenix needs SessionUseCases which can always use a new tab
The Fenix multitasking design is such that certain UI interactions require always creating a new tab for SuggestionProviders. This is not only for cases where no Session already exists.

https://github.com/mozilla-mobile/fenix/issues/239#issuecomment-459578181

colintheshots:


components/feature/session/src/main/java/mozilla/components/feature/session/SessionUseCases.kt
```diff
@@ -42,7 +44,8 @@ class SessionUseCases(
          * @param url The URL to be loaded using the selected session.
          */
         override fun invoke(url: String) {
-            this.invoke(url, sessionManager.selectedSession)
+            val loadSession = if (useNewSession) onNoSession.invoke(url) else sessionManager.selectedSession
+            this.invoke(url, loadSession)
```
pocmo: This is already covered by the `TabsUseCases` (`AddNewTabUseCase`, `AddNewPrivateTabUseCase`) and they implement `LoadUrlUseCase` so that you can use them instead of the `DefaultLoadUrlUseCase` here.

pocmo:
Added something for search in #1990

PR #1984, opened 2019-02-07
[glean] 1516527: Add labeled metrics.

mdboom:
This is the hopefully not-too-big PR to add support for labeled metrics.  (I initially had a more invasive change that I decided wasn't worth the effort to try to review).

Marked as WIP, since this will not pass the tests until the [glean_parser side of this](https://github.com/mozilla/glean_parser/pull/29) is merged and a new release is made.

The general design is:

1. Labeled metrics aren't stored any differently from other metrics.  The only difference is that the key in the storage engine is of the form `$category.$name/$label`.  Using `/` as a special character should be fine given the existing constraints on categories, names and labels that they are all `snake_case` (coming from pipeline side requirements).

2. The `LabeledMetricType` class really just serves as a factory to create specific metric type objects for each of the labels, as it doesn't have a storage engine of its own.  Making this a factory that can return different types based on a template parameter is why `CommonMetricType` has grown a template parameter and most of the random changes around the code base are due to that.

KNOWN ISSUES:

The code to make sure that when labels aren't defined that are only 16 labels doesn't work if we're killed and we reload pre-saved metrics from disk, since it won't know about them.  The alternative seemed to be to go through the whole storage engine and count the number of labels already being used each time, which seemed kind of expensive.  An alternative would be to do this counting just once when reloading from disk, but it wasn't clear to me how then to communicate that back to the `LabeledMetricType` instance for the correct category.  As this is kind of a corner case, I thought I would submit this knowing that and keep working on it in the meantime.  This is filed as [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1527312) and will be addressed in follow-on work.

### Pull Request checklist


components/service/glean/build.gradle
```diff
@@ -14,8 +14,8 @@ apply plugin: 'kotlin-android'
  * created during unit testing.
  * This uses a specific version of the schema identified by a git commit hash.
  */
-String GLEAN_PING_SCHEMA_GIT_HASH = "8939bcb"
-String GLEAN_PING_SCHEMA_URL = "https://raw.githubusercontent.com/mozilla-services/mozilla-pipeline-schemas/$GLEAN_PING_SCHEMA_GIT_HASH/schemas/glean/baseline/baseline.1.schema.json"
+String GLEAN_PING_SCHEMA_GIT_HASH = "a2c1ee1"
+String GLEAN_PING_SCHEMA_URL = "https://raw.githubusercontent.com/mdboom/mozilla-pipeline-schemas/$GLEAN_PING_SCHEMA_GIT_HASH/schemas/glean/baseline/baseline.1.schema.json"
```
mdboom: This is updating the schemas we're checking against to use the work-in-progress in my personal fork [this PR](https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/254).  We will need to change this back to the mainline fork and a new git commit before merging this.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -188,6 +190,13 @@ open class GleanInternalAPI {
      * Initialize the core metrics internally managed by Glean (e.g. client id).
      */
     private fun initializeCoreMetrics(applicationContext: Context) {
+        // Since all of the ping_info properties are required, we can't
+        // use the normal metrics API to set them, since those work
+        // asynchronously, and there is a race condition between when they
+        // are set and the possible sending of the first ping upon startup.
+        // Therefore, this uses the lower-level internal storage engine API
+        // to set these metrics, which is synchronous.
```
mdboom: I discovered this because I updated the schema.  Currently on master, the `appBuild` and `appDisplayVersion` aren't mentioned in the schema we are checking against, so frequently in tests they are missing but it's not complaining.  (That's my bad for not remembering to update the schema git commit when these keys were added to the schema).  If you'd prefer this was a separate PR, let me know.

components/service/glean/src/main/java/mozilla/components/service/glean/CommonMetricData.kt
```diff
@@ -74,4 +74,18 @@ interface CommonMetricData {
...
+     * Returns a copy of this metric with the name changed to `newName`.
+     *
+     * This is used by |LabeledMetricType| to create metrics that belong
+     * in labeled metric groups.  Must be overridden by all metric types
+     * that can be labeled.
+     *
+     * @param newName The new name
+     * @returns A copy of this metric with the name changed to `newName`.
+     */
+    fun getWithName(newName: String): T {
```
Dexterp37: I wish it were possible to make this `internal`, but apparently it isn't. As a result, this is showing up in the metric API. Do you have any idea on how to fix that? I would also call this `getWithNewName` :) But that's bikeshedding, up to you

components/service/glean/src/main/java/mozilla/components/service/glean/LabeledMetricType.kt
```diff
@@ -0,0 +1,108 @@
...
+
+    /**
+     * Handles the label in the case where labels are predefined.
+     *
+     * If the given label is not in the predefined set of labels, returns "__other__", otherwise
+     * returns the label verbatim.
+     *
+     * @param label
+     */
+    private fun handleDefinedLabels(label: String): String {
```
Dexterp37: nit: let's stick to the naming convention used in the spec and refer to these as "static labels". This would become `handleStaticLabels`. But since this is returning the label that we'll eventually use, what about `getFinalStaticLabel` or something like that?

components/service/glean/src/main/java/mozilla/components/service/glean/LabeledMetricType.kt
```diff
@@ -0,0 +1,108 @@
...
+        private const val OTHER_LABEL = "__other__"
+        private val labelRegex = Regex("^[a-z_][a-z0-9_]{0,29}$")
+    }
+
+    private val seenLabels: MutableSet<String> = mutableSetOf()
+
+    /**
+     * Handles the label in the case where labels are predefined.
+     *
+     * If the given label is not in the predefined set of labels, returns "__other__", otherwise
```
Dexterp37: The "returns... " part should go after `@param`, in a `@returns`. Let's also add some description to `@param`

components/service/glean/src/main/java/mozilla/components/service/glean/LabeledMetricType.kt
```diff
@@ -0,0 +1,108 @@
...
+     * @param label
+     */
+    private fun handleDefinedLabels(label: String): String {
+        return if (labels!!.contains(label)) label else OTHER_LABEL
+    }
+
+    /**
+     * Handles the label in the case where labels aren't predefined.
+     *
+     * If we've already seen more than |MAX_LABELS| unique labels, returns "__other__", otherwise
```
Dexterp37: The "returns... " part should go after `@param`, in a `@returns`. Let's also add some description to `@param`

components/service/glean/src/main/java/mozilla/components/service/glean/LabeledMetricType.kt
```diff
@@ -0,0 +1,108 @@
...
+     * @param label
+     */
+    private fun handleUndefinedLabels(label: String): String {
+        if (!seenLabels.contains(label)) {
+            if (seenLabels.size >= MAX_LABELS) {
+                return OTHER_LABEL
+            } else {
+                // Labels must be snake_case.
+                if (!labelRegex.matches(label)) {
+                    throw IllegalArgumentException(
```
Dexterp37: Our API should never throw. Let's log an error instead and return the `OTHER_LABEL` (unless we want to return nullables, which would be a bit painful downstream). Let's also clearly document this behaviour in the `get` operator, which I think shows up in the IDE?

components/service/glean/src/main/java/mozilla/components/service/glean/LabeledMetricType.kt
```diff
@@ -0,0 +1,108 @@
...
+     * Handles the label in the case where labels aren't predefined.
+     *
+     * If we've already seen more than |MAX_LABELS| unique labels, returns "__other__", otherwise
+     * returns the label verbatim.
+     *
+     * Also validates any unseen labels to make sure they are snake_case and under 30 characters.
+     *
+     * @param label
+     */
+    private fun handleUndefinedLabels(label: String): String {
```
Dexterp37: `handleDynamicLabels` or `getFinalDynamicLabel`?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/StorageEngineManager.kt
```diff
@@ -30,6 +32,52 @@ internal class StorageEngineManager(
         }
     }
 
+    /**
+     * Splits a labeled metric back into its name/label parts.
+     *
+     * If not a labeled metric, the second part of the Pair will be null.
+     */
```
Dexterp37: nit: let's docuemnt the parameter and return

components/service/glean/src/main/java/mozilla/components/service/glean/storages/StorageEngineManager.kt
```diff
@@ -43,10 +91,11 @@ internal class StorageEngineManager(
         val metricsSection = JSONObject()
         for ((sectionName, engine) in storageEngines) {
             val engineData = engine.getSnapshotAsJSON(storeName, clearStore = true)
-            if (engine.sendAsTopLevelField) {
-                jsonPing.put(sectionName, engineData)
-            } else {
-                metricsSection.put(sectionName, engineData)
+            val dst = if (engine.sendAsTopLevelField) jsonPing else metricsSection
+            if (engineData is JSONObject) {
```
Dexterp37: Why do we need this distinction? Can you add a comment about that?

components/service/glean/src/test/java/mozilla/components/service/glean/LabeledMetricTypeTest.kt
```diff
@@ -0,0 +1,388 @@
...
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNull
+import java.util.UUID
+
+@RunWith(RobolectricTestRunner::class)
+class LabeledMetricTypeTest {
```
Dexterp37: What happens when using empty strings as keys? Can we add test coverage for that? Should we also add test coverage to make sure we catch attempts to use "/" in the name?

components/service/glean/src/main/java/mozilla/components/service/glean/CounterMetricType.kt
```diff
@@ -55,4 +55,14 @@ data class CounterMetricType(
             )
         }
     }
+
+    override fun getWithName(newName: String): CounterMetricType {
+        return CounterMetricType(
```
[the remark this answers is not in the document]
mdboom: Sure, I'll play with this some more.  Couldn't find another way in our "data-classes-that-inherit-from-an-interface" framing we currently have.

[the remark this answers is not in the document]
Dexterp37: Sure! Even if he'll hate me a bit, I'd feel more comfortable with this if @pocmo (or anyone else with Kotlin experience) would give at least the `LabeledMetricType.kt` file an in-depth review, so maybe worth asking him too!

components/service/glean/src/test/java/mozilla/components/service/glean/LabeledMetricTypeTest.kt
mdboom: These should be caught at build time, so I'll probably add the tests for this in `glean_parser`.  Since the second part of `glean` is `lean`, I'd rather not add runtime checks.

mdboom: Though, I guess users could make their own metrics without going through `metrics.yaml`, but we discourage that.  (Don't know if there's any way to prevent that).  Hmmm...  Maybe we need runtime checks after all...

components/service/glean/src/main/java/mozilla/components/service/glean/CounterMetricType.kt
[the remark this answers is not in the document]
mdboom: The gotcha here is that generics in Kotlin require a common interface, since it doesn't generate a new function for each type T.  Since data classes don't inherit from a common interface or base class, we can't call `copy` on them in this context.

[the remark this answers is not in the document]
mdboom: You can do this:

```kotlin
    private fun getMetric(name: String): T {
        return when (subMetric) {
            is BooleanMetricType -> subMetric.copy(name = name) as T
            is CounterMetricType -> subMetric.copy(name = name) as T
            is StringListMetricType -> subMetric.copy(name = name) as T
            is StringMetricType -> subMetric.copy(name = name) as T
            is TimespanMetricType -> subMetric.copy(name = name) as T
            is UuidMetricType -> subMetric.copy(name = name) as T
            else -> throw IllegalStateException()
        }
    }
```

which is a lot less verbose and puts all of the weirdness in one place, and removes the need for the `getWithName` method and the template parameter on `CommonMetricData`, but it's probably not quite as efficient (though I'm only guessing).

components/service/glean/src/test/java/mozilla/components/service/glean/TestUtil.kt
```diff
@@ -46,6 +58,40 @@ internal fun checkPingSchema(content: JSONObject) {
...
+ * @returns the ping contents, in a JSONObject
+ * @raises AssertionError If the JSON content is not valid
+ */
+internal fun collectAndCheckPingSchema(storeName: String): JSONObject {
+    val appContext = ApplicationProvider.getApplicationContext<Context>()
+    val jsonString = PingMaker(
+        StorageEngineManager(applicationContext = appContext),
+        appContext
+    ).collect(storeName)!!
+    val jsonContent = checkPingSchema(jsonString)
```
Dexterp37: nit: can't this be `return checkPingSchema(jsonString)`?

components/service/glean/src/main/java/mozilla/components/service/glean/CommonMetricData.kt
mdboom: Thanks for the suggestion, @travis79.  I ended up solving this in a different way with a big delegation switch function (which can easily be private so encapsulates things the way we want).

components/service/glean/src/test/java/mozilla/components/service/glean/LabeledMetricTypeTest.kt
mdboom: Sorry -- I was thinking you were talking about the storage engine keys, not just the labels.  I've added tests for these things on labels.  I think for now, I'm comfortable with relying on the build-time checks for the other things that make up the storage engine key (name and category).

components/service/glean/src/main/java/mozilla/components/service/glean/LabeledMetricType.kt
```diff
@@ -0,0 +1,112 @@
...
+
+    private val seenLabels: MutableSet<String> = mutableSetOf()
+
+    /**
+     * Handles the label in the case where labels are predefined.
+     *
+     * If the given label is not in the predefined set of labels, returns "__other__", otherwise
+     * returns the label verbatim.
+     *
+     * @param label
```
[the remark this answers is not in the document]
mdboom: While we could make them work, I don't think enumerations would make sense.  Even in the case of static labels, we want to be able to transparently handle "other" values, and we don't want to put the onus on the user to determine which values are "other".

components/service/glean/src/main/java/mozilla/components/service/glean/LabeledMetricType.kt
```diff
@@ -0,0 +1,130 @@
...
+    private fun getMetricWithNewName(newName: String): T {
+        // Every metric that supports labels needs an entry here
+        return when (subMetric) {
+            is BooleanMetricType -> subMetric.copy(name = newName) as T
+            is CounterMetricType -> subMetric.copy(name = newName) as T
+            is StringListMetricType -> subMetric.copy(name = newName) as T
+            is StringMetricType -> subMetric.copy(name = newName) as T
+            is TimespanMetricType -> subMetric.copy(name = newName) as T
+            is UuidMetricType -> subMetric.copy(name = newName) as T
+            else -> throw IllegalStateException()
```
Dexterp37: nit: since this is throwing, should we add an error message to `IllegalStateException` so that we know what's happening while we panic ? :P

components/service/glean/src/main/java/mozilla/components/service/glean/LabeledMetricType.kt
```diff
@@ -0,0 +1,130 @@
...
+    private val seenLabels: MutableSet<String> = mutableSetOf()
+
+    /**
+     * Handles the label in the case where labels are predefined.
+     *
+     * If the given label is not in the predefined set of labels, returns [OTHER_LABEL], otherwise
+     * returns the label verbatim.
+     *
+     * @param label The label, as specified by the user
+     * @param adjusted label, possibly set to [OTHER_LABEL]
```
Dexterp37: Hey Mike, sorry to nitpick this, but I think that this line should be ` @returns adjusted label, possibly set to [OTHER_LABEL]` (`@returns` rather than `@param`)

components/service/glean/src/main/java/mozilla/components/service/glean/LabeledMetricType.kt
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * Handles the label in the case where labels aren't predefined.
+     *
+     * If we've already seen more than [MAX_LABELS] unique labels, returns [OTHER_LABEL].
+     *
+     * Also validates any unseen labels to make sure they are snake_case and under 30 characters.
+     * If not, returns [OTHER_LABEL].
+     *
+     * @param label The label, as specified by the user
+     * @param adjusted label, possibly set to [OTHER_LABEL]
```
Dexterp37: `@returns` rather than `@param`

components/service/glean/src/test/java/mozilla/components/service/glean/LabeledMetricTypeTest.kt
Dexterp37: My bad, I should have been less cryptic and much clearer :)

components/service/glean/src/main/java/mozilla/components/service/glean/storages/StorageEngineManager.kt
```diff
@@ -30,6 +32,56 @@ internal class StorageEngineManager(
...
+     * Splits a labeled metric back into its name/label parts.
+     *
+     * If not a labeled metric, the second part of the Pair will be null.
+     *
+     * @param key The key for the metric value in the flattened storage engine
+     * @returns A pair (metricName, label).  label is null if key is not
+     * from a labeled metric.
+     */
+    private fun parseLabeledMetric(key: String): Pair<String, String?> {
+        var divider = key.indexOf('/', 1)
```
Dexterp37: nit: is `var` really required here? We're not ever changing this, so I guess it could be a `val`

components/service/glean/src/main/java/mozilla/components/service/glean/storages/StorageEngineManager.kt
```diff
@@ -30,6 +32,56 @@ internal class StorageEngineManager(
...
+        if (divider >= 0) {
```
Dexterp37: optional nit: probably more idiomatic:

```
val splits = key.split('/`, limit=2)
return if (splits.size == 2) splits.zipWithNext().first() else Pair(key, null)
```

mdboom:
This has been updated with a solution to the problem of tracking only *n* labels that does the right thing for user lifetime labels that were deserialized from disk.  See `LabeledMetricType.getFinalDynamicLabel` and `GenericScalarStorageEngine.getLabelsForMetric` for the implementation.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -177,6 +187,35 @@ internal abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
...
+    /**
+     * Return a set of the all the labels in use for a given labeled metric.
+     * This includes labels in all of the pings specified by the metric, and
+     * for all lifetimes.
+     *
+     * @param metricData The labeled metric to find labels for.
+     * @return The set of labels currently in use.
+     */
+    @Suppress("NestedBlockDepth")
+    override fun getLabelsForMetric(metricData: CommonMetricData): Set<String> {
```
Dexterp37: This shouldn't really live in the GenericStorageEngine, IMHO. I'd suggest to return the PR to the previous state (with the knonw issue about persisted labelled data), merge that, and approach this as follow-up work as this might require some discussion.

mdboom: Ok.  I've filed https://bugzilla.mozilla.org/show_bug.cgi?id=1527312 to track this.

PR #1988, opened 2019-02-07
 Closes #1987: Make back button presses sessionId aware 

### LINKED ISSUE #1987: SessionFeature.handleBackPressed isn't sessionId aware
If we press back in a Custom Tab, `SessionFeature#handleBackPressed` should be handling this call first before the session is closed.

jonalmeida:
### Pull Request checklist

### DIFF
--- a/components/feature/session/src/main/java/mozilla/components/feature/session/SessionFeature.kt
+++ b/components/feature/session/src/main/java/mozilla/components/feature/session/SessionFeature.kt
@@ -16,7 +16,7 @@ class SessionFeature(
     private val sessionManager: SessionManager,
     private val sessionUseCases: SessionUseCases,
     engineView: EngineView,
-    sessionId: String? = null
+    private val sessionId: String? = null
 ) : LifecycleAwareFeature, BackHandler {
     internal val presenter = EngineViewPresenter(sessionManager, engineView, sessionId)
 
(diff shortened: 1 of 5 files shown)


components/feature/session/src/main/java/mozilla/components/feature/session/SessionFeature.kt
```diff
@@ -32,9 +32,13 @@ class SessionFeature(
...
-                sessionUseCases.goBack.invoke()
+        val session = sessionId?.let {
+            sessionManager.findSessionById(it)
+        } ?: sessionManager.selectedSession
+
+        session?.let { it ->
+            if (it.canGoBack) {
+                sessionUseCases.goBack.invoke(it)
                 return true
             }
```
pocmo: nit: Since the `session` reference is immutable you do not really need `let` here (since Kotlin can smartcast), e.g.:

```Kotlin
if (session != null && session.canGoBack) {
    sessionUseCases.goBack.invoke(session)
}
```

or maybe:

```kotlin
if (session?.canGoBack == true) {
    sessionUseCases.goBack.invoke(session)
}
```

jonalmeida: Nice, I like the latter (because no nulls). :)

pocmo: I agree that `null` can be a pain. But your refusal to even type it is hilarious :)

PR #1992, opened 2019-02-07
Closes #1876: Tint all icons in custom tab

### LINKED ISSUE #1876: Custom Tabs: close, share, menu and security icons show be tinted for light toolbar colors
screenshot_20190127-134105

jonalmeida:
For #1876 

### Pull Request checklist

### DIFF
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
@@ -56,7 +56,7 @@ class BrowserToolbar @JvmOverloads constructor(
     @VisibleForTesting internal var editToolbar = EditToolbar(context, this)
 
     /**
-     * Set/Get whether a site security icon (usually a lock or globe icon) should be next to the URL.
+     * Set/Get whether a site security icon (usually a lock or globe icon) should be visible next to the URL.
      */
     var displaySiteSecurityIcon: Boolean
         get() = displayToolbar.siteSecurityIconView.isVisible()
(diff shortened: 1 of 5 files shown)


components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
```diff
@@ -58,19 +58,27 @@ class BrowserToolbar @JvmOverloads constructor(
     /**
      * Set/Get whether a site security icon (usually a lock or globe icon) should be next to the URL.
      */
-    var displaySiteSecurityIcon: Boolean
+    var siteSecurityIcon: Boolean
```
Amejia481: My first thought to this variable is that it will be to it give me access to siteSecurityIcon not is visibility. What about including the visibility on the name to clarify its purpose?
something like`isSiteSecurityIconVisible`.

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
```diff
@@ -58,19 +58,27 @@ class BrowserToolbar @JvmOverloads constructor(
...
         get() = displayToolbar.siteSecurityIconView.isVisible()
         set(value) {
             displayToolbar.siteSecurityIconView.visibility = if (value) View.VISIBLE else View.GONE
         }
 
+    var siteSecurityColor: Pair<Int, Int>
```
Amejia481: We can improve the name of the variable because it is not clearly conveying its purpose.
As it is a pair it will contain two colours but I'm not quite sure which state is which. Maybe adding a comment?

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
pocmo: I'm also confused about this change. Why is it needed? :)

jonalmeida: Thanks for the catch, this wasn't intended to be committed!

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
jonalmeida: Added comments!

PR #2007, opened 2019-02-11
Closes #1328: Provide GeckoView based implementation of concept-fetch.

### LINKED ISSUE #1328: Create GeckoView implementation of concept-fetch
Breaking down: #1012

csadilek:
This brings in the GV-based implementation of concept-fetch. The implementation is simple and following the same pattern as our `okhttp` and `httpurlconnection` implementations. Instead of a separate lib module, this is part of `engine-gecko-*` so we can easily deal with API changes / let it ride the trains.

The tests are more involved. I made small changes to our `tooling-fetch-test` suite to be able to run the tests as both instrumented (androidTest) and standard unit tests. `okhttp` and `httpurlconnection` still run as JVM unit tests, while the GV (GWE) implemenation runs as both! The instrumented tests cover the full end-to-end functionality. The unit tests mock both the reqeust and response side, but do provide value by testing the mapping logic in between including the buffer/stream. This should also give us full coverage.

**ToDos / Questions**
- I've only added this to engine-gecko-nightly for now until reviewed
- The instrumented unit tests are only set up to run on ARM devices. We need to make a decision how we want to run those in automation, but since we have JVM unit tests too, we could file this a follow-up?


components/tooling/fetch-tests/src/main/java/mozilla/components/tooling/fetch/tests/FetchTestCases.kt
```diff
@@ -363,16 +396,27 @@ abstract class FetchTestCases {
...
 
         try {
-            server.start()
-            server.block(createNewClient())
+            val client = createNewClient()
+            // Subclasses (implementation specific tests) might be instrumented
+            // and run on a device so we need to avoid network requests on the
+            // main thread.
+            runBlocking {
+                launch(Dispatchers.IO) {
```
pocmo: Does `runBlocking(Dispatchers.IO)` work here too?

pocmo:
> I've only added this to engine-gecko-nightly for now until reviewed

I think that's fine. Not necessarily needed to uplift.

components/tooling/fetch-tests/src/main/java/mozilla/components/tooling/fetch/tests/FetchTestCases.kt
csadilek: Ah yes, updated!  :)

csadilek:
Changelog entry added.

PR #2014, opened 2019-02-11
[glean] 1525045: Implement "first_run_date" and dependencies

mdboom:
This adds the "first_run_date" metric and the underlying datetime metric type needed to implement it.

Of note is that the datetime metric stores its values as ISO strings.  There is no way with the pre-Java 8 API to roundtrip the timezone offset through an ISO date string.  (It's read correctly, and the date is correct, but there is no way to preserve what the timezone offset was).  Since knowing the timezone offset at the time the value is recorded is a requirement, it was easiest to just store datetime as strings.  We could make this round-tripping work, but that would require either adding a dependency or writing specialized code to parse it that could be fiddly.  Since we don't really need to use these values again for anything (other than testing), I thought it was simplest to just convert to strings early and keep them that way.  But obviously, I'm open to other ideas.

### Pull Request checklist


components/service/glean/build.gradle
```diff
@@ -14,8 +14,8 @@ apply plugin: 'kotlin-android'
  * created during unit testing.
  * This uses a specific version of the schema identified by a git commit hash.
  */
-String GLEAN_PING_SCHEMA_GIT_HASH = "8939bcb"
-String GLEAN_PING_SCHEMA_URL = "https://raw.githubusercontent.com/mozilla-services/mozilla-pipeline-schemas/$GLEAN_PING_SCHEMA_GIT_HASH/schemas/glean/baseline/baseline.1.schema.json"
+String GLEAN_PING_SCHEMA_GIT_HASH = "0f6b6f"
+String GLEAN_PING_SCHEMA_URL = "https://raw.githubusercontent.com/mdboom/mozilla-pipeline-schemas/$GLEAN_PING_SCHEMA_GIT_HASH/schemas/glean/baseline/baseline.2.schema.json"
```
mdboom: We need to switch this back to the main fork once [this PR](https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/259) is merged.

components/service/glean/metrics.yaml
```diff
@@ -117,6 +117,21 @@ glean.baseline:
...
+    lifetime: user
+    send_in_pings:
+      - baseline
+    time_unit: day
+    description: |
+      The date of the first run of the application.
+    bugs:
+      - 1525045
+    data_reviews:
+      - NEEDED
```
mdboom: I guess this needs a data review.  I'm follow up on that in the bug.

components/service/glean/src/main/java/mozilla/components/service/glean/DatetimeMetricType.kt
```diff
@@ -0,0 +1,127 @@
...
+ * Instances of this class type are automatically generated by the parsers at build time,
+ * allowing developers to record values that were previously registered in the metrics.yaml file.
+ */
+data class DatetimeMetricType(
+    override val disabled: Boolean,
+    override val category: String,
+    override val lifetime: Lifetime,
+    override val name: String,
+    override val sendInPings: List<String>,
+    val timeUnit: TimeUnit = TimeUnit.Nanosecond
```
travis79: Since `TimeUnit.Second` is the minimum resolution, should we make this default to that?

components/service/glean/src/main/java/mozilla/components/service/glean/DatetimeMetricType.kt
```diff
@@ -0,0 +1,127 @@
...
+    /**
+     * Set a datetime value.
+     *
+     * This is provided as an internal-only function so that we can test that timezones
+     * are passed through correctly.  The normal public interface uses [Date] objects which
+     * are always in the local timezone.
+     *
+     * @param value The [Calendar] value to set. If not provided, will record the current time.
+     */
+    internal fun set(value: Calendar) {
```
travis79: Is this something that we can mark with `@VisibleForTesting(otherwise = VisibleForTesting.NONE)`?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/DatetimesStorageEngine.kt
```diff
@@ -0,0 +1,76 @@
...
+        extraSerializationData: Any?
+    ) {
+        userPreferences?.putString(storeName, value)
+    }
+
+    /**
+     * Set the metric to the provided date/time, truncating it to the
+     * metric's resolution.
+     *
+     * @param date the date value to set this metric to
```
travis79: nit: params listed in reverse order

components/service/glean/src/main/java/mozilla/components/service/glean/storages/DatetimesStorageEngine.kt
```diff
@@ -0,0 +1,76 @@
...
+     */
+    fun set(metricData: CommonMetricData, date: Date = Date()) {
+        super.recordScalar(metricData, getISOTimeString(date))
+    }
+
+    /**
+     * Set the metric to the provided date/time, truncating it to the
+     * metric's resolution.
+     *
+     * @param date the date value to set this metric to
```
travis79: `@param date` should be `@param calendar`

nit: params listed in reverse order

components/service/glean/src/test/java/mozilla/components/service/glean/storages/DatetimesStorageEngineTest.kt
```diff
@@ -0,0 +1,167 @@
...
+                lifetime = Lifetime.User,
+                name = "datetime_metric",
+                sendInPings = listOf("store1")
+            )
+
+            // Record the datetime in the store, without providing optional arguments.
+            storageEngine.set(metric, value)
+
+            // Get the snapshot from "store1"
+            val snapshot = storageEngine.getSnapshotAsJSON(storeName = "store1",
```
travis79: Since you created `metric` above as a `DatetimeMetricType` can you use the test API functions here?

components/service/glean/src/test/java/mozilla/components/service/glean/storages/DatetimesStorageEngineTest.kt
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        // Create a new instance of storage engine to verify serialization to storage rather than
+        // to the cache
+        run {
+            val storageEngine = DatetimesStorageEngineImplementation()
+            storageEngine.applicationContext = ApplicationProvider.getApplicationContext()
+
+            // Get the snapshot from "store1"
+            val snapshot = storageEngine.getSnapshotAsJSON(storeName = "store1",
```
travis79: Same nit as above, can we use the test API here since you have created a `DatetimeMetricType` since you aren't specifically testing `getSnapshot()`?

components/service/glean/src/test/java/mozilla/components/service/glean/storages/DatetimesStorageEngineTest.kt
```diff
@@ -0,0 +1,167 @@
...
+            name = "datetime_metric",
+            sendInPings = storeNames
+        )
+
+        // Record the datetime in the stores, without providing optional arguments.
+        DatetimesStorageEngine.set(metric)
+
+        // Check that the data was correctly set in each store.
+        for (storeName in storeNames) {
+            val snapshot = DatetimesStorageEngine.getSnapshot(storeName = storeName, clearStore = false)
```
travis79: Can we use the test API functions here since you have the `DatetimeMetricType` instance above since you aren't specifically testing `getSnapshot()`?

components/service/glean/src/main/java/mozilla/components/service/glean/DatetimeMetricType.kt
mdboom: Good point.  Given discussion elsewhere (on Slack mainly), I think the appropriate default here is "minute" (for minimal collection reasons).

components/service/glean/src/main/java/mozilla/components/service/glean/DatetimeMetricType.kt
mdboom: Good idea.

components/service/glean/src/test/java/mozilla/components/service/glean/storages/DatetimesStorageEngineTest.kt
mdboom: Here I want to check that the JSON is exactly what we expect it to be.  It's a subtle difference, but given that we've changed the spec to say datetimes are `(value, unit)` pairs to just `value`, I think it's good to have at least one test that looks directly at the JSON.  All of these in here probably don't need to do that though.

mdboom:
I've addressed @travis79's comments.

I also redid the truncation so it happens at the string level, i.e. if we don't want to send seconds, the string simply doesn't contain them.  This is a nice side effect of dealing with everything at the string level.

components/service/glean/src/test/java/mozilla/components/service/glean/storages/DatetimesStorageEngineTest.kt
travis79: Makes sense to me :)

components/service/glean/docs/pings.md
```diff
@@ -24,7 +24,7 @@ The following fields are always included in the `ping_info` section, for every p
 | `experiments` | Object | A dictionary of [active experiments](#the-experiments-object) |
 | `start_time` | Datetime | The time of the start of collection of the data in the ping | See [note](#a-note-about-time-formats) |
 | `end_time` | Datetime | The time of the end of collection of the data in the ping. This is also the time this ping was generated and is likely well before ping transmission time | See [note](#a-note-about-time-formats) |
-| `profile_age` | Datetime | TBD |
+| `first_run_date` | Datetime | The date of the first run of the application |
```
Dexterp37: nit: let's mention the date format paragraph as done for `end_time`

components/service/glean/metrics.yaml
```diff
@@ -162,3 +147,17 @@ glean.internal.metrics:
...
+    lifetime: user
+    send_in_pings:
+      - glean_ping_info
+    time_unit: day
+    description: |
+      The date of the first run of the application.
+    bugs:
+      - 1525045
+    data_reviews:
+      - NEEDED
```
Dexterp37: Let's add the link to the data review

PR #2019, opened 2019-02-12
dokka: Link to GeckoView javadoc.

pocmo:
This will make classes from GeckoView link to the GeckoView javadocs.


pocmo:
This patch is now part of #2021

PR #2027, opened 2019-02-12
Closes #2026: Ignoring schemes that are not https or http when downloading a file.

### LINKED ISSUE #2026: Ignore schemes that are not https or http when downloading a file
Related issue: https://github.com/mozilla-mobile/reference-browser/issues/554

Amejia481:
### Pull Request checklist

- [X] **Changelog**: This PR includes a changelog entry or does not need one.

### DIFF
--- a/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadManager.kt
+++ b/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadManager.kt
@@ -17,12 +17,15 @@ import android.content.Intent
 import android.content.IntentFilter
 import android.net.Uri
 import android.support.annotation.RequiresPermission
+import android.widget.Toast
 import mozilla.components.browser.session.Download
 import mozilla.components.support.ktx.android.content.isPermissionGranted
 import mozilla.components.support.utils.DownloadUtils
 
 typealias OnDownloadCompleted = (Download, Long) -> Unit
 typealias AndroidDownloadManager = android.app.DownloadManager
+
+internal const val FILE_NOT_SUPPORTED = -1L
 
 /**
  * Handles the interactions with the [AndroidDownloadManager].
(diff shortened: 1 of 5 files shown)


components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadManager.kt
```diff
@@ -52,6 +55,14 @@ class DownloadManager(
         cookie: String = ""
     ): Long {
 
+        if (download.hasAValidProtocol()) {
```
pocmo: nit: `isSupportedProtocol()` is a bit easier to read (`hasAValid" looks weird 😃)

components/feature/downloads/src/main/res/values/strings.xml
```diff
@@ -11,4 +11,7 @@
 
     <!-- Alert dialog confirmation before download a file, this is the negative action. -->
     <string name="mozac_feature_downloads_dialog_cancel">Cancel</string>
+
+    <!-- Error shown when the user is trying to download a invalid file. -->
+    <string name="mozac_feature_downloads_file_not_supported">Download protocol not supported</string>
```
pocmo: Is that the message that Focus uses? Otherwise I'm wondering if we should include the protocol in the error message.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadManager.kt
Amejia481: 👍

components/feature/downloads/src/main/res/values/strings.xml
Amejia481: It really came from [FireFoxLite](https://github.com/mozilla-tw/FirefoxLite/blob/fb7a5e18043d88378be3db7e350470a125513cb3/app/src/main/java/org/mozilla/focus/download/EnqueueDownloadTask.java#L139), [focus is just ignoring the download and not showing an error message](https://github.com/mozilla-mobile/focus-android/blob/0bc2d3dc8db69364fa4715f6382e5b039984bb5b/app/src/main/java/org/mozilla/focus/web/GeckoWebViewProvider.kt#L387). Should I update the message? If that it's the case which format should I use for the message?

pocmo: Okay, let's keep the string you have here for now! :)

Amejia481: 👍

PR #2029, opened 2019-02-13
Closes #2025: concept-fetch: Add API to control if cookies should be sent

### LINKED ISSUE #2025: concept-fetch: Provide API to control if cookies should be sent

csadilek:
This brings in a cookie flag/policy to specify whether or not cookies should be sent with a request. The two policies work for all our implementations now. 🎉 

Test was added to our `tooling-fetch-test` module and runs as a JVM and instrumented test for GV.

While testing I realized that for okhttp and httpurlconnection we don't have a persistent cookie store that will survive app restarts. I am not sure that's needed as we have GV for that, and the way this is implemented here, applications can also provide their own `CookieHandler`. If we want a built-in one anyway, I can file a follow-up (we have a few options there).

### Pull Request checklist


components/lib/fetch-httpurlconnection/src/main/java/mozilla/components/lib/fetch/httpurlconnection/HttpURLConnectionClient.kt
```diff
@@ -62,6 +74,14 @@ private fun HttpURLConnection.setupWith(request: Request) {
     request.readTimeout?.let { (timeout, unit) ->
         readTimeout = unit.toMillis(timeout).toInt()
     }
+
+    val cookieManager = getOrCreateCookieManager()
+    if (request.cookiePolicy == Request.CookiePolicy.OMIT) {
+        val uri = URL(request.url).toURI()
+        for (cookie in cookieManager.cookieStore.get(uri)) {
+            cookieManager.cookieStore.remove(uri, cookie)
+        }
```
pocmo: Let's maybe add a comment here why we are doing this - because it looks weird if you do not know the background. :)

components/lib/fetch-okhttp/src/main/java/mozilla/components/lib/fetch/okhttp/OkHttpClient.kt
```diff
@@ -48,6 +58,10 @@ private fun okhttp3.OkHttpClient.rebuildFor(request: Request): okhttp3.OkHttpCli
             clientBuilder.followRedirects(false)
         }
 
+        if (request.cookiePolicy == Request.CookiePolicy.INCLUDE) {
+            clientBuilder.cookieJar(JavaNetCookieJar(cookieHandler))
+        }
```
pocmo: Is the Java cookie handler the only option here or could we also use an isolated cookie jar (not sure if okhttp comes with multiple implementations?)

components/lib/fetch-httpurlconnection/src/main/java/mozilla/components/lib/fetch/httpurlconnection/HttpURLConnectionClient.kt
csadilek: OK, comment added.

components/lib/fetch-okhttp/src/main/java/mozilla/components/lib/fetch/okhttp/OkHttpClient.kt
csadilek: Nothing else built-in but we could provide our own implementation later if we want to e.g. if we want a persistent one (but it's tricky).

PR #2050, opened 2019-02-13
[glean] 1499761: Track errors on metrics

mdboom:
This implements error tracking that is reported in pings, as defined in the spec.

Questions for discussion:
~~- Are these the right categories of errors?  `OutOfRange` and `InvalidValue` seem perhaps redundant.  `InvalidType` is never used, but maybe that would be useful, e.g. for a Javascript API someday.~~
~~- There's a "backdoor" here to make error reporting synchronous, which makes writing tests for this possible.  This is a "category" of problem we run into all the time -- maybe there's a better solution here.  See [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1527670) for one thing that might help with this specific problem.~~

### Pull Request checklist


components/service/glean/build.gradle
```diff
@@ -14,8 +14,8 @@ apply plugin: 'kotlin-android'
  * created during unit testing.
  * This uses a specific version of the schema identified by a git commit hash.
  */
-String GLEAN_PING_SCHEMA_GIT_HASH = "04e203c"
-String GLEAN_PING_SCHEMA_URL = "https://raw.githubusercontent.com/mozilla-services/mozilla-pipeline-schemas/$GLEAN_PING_SCHEMA_GIT_HASH/schemas/glean/baseline/baseline.1.schema.json"
+String GLEAN_PING_SCHEMA_GIT_HASH = "175920c"
+String GLEAN_PING_SCHEMA_URL = "https://raw.githubusercontent.com/mdboom/mozilla-pipeline-schemas/$GLEAN_PING_SCHEMA_GIT_HASH/schemas/glean/baseline/baseline.1.schema.json"
```
mdboom: We will need to move this back to the base after merging https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/260

components/service/glean/metrics.yaml
```diff
@@ -147,3 +147,55 @@ glean.internal.metrics:
...
+  out_of_range:
+    type: counter
+    labeled: true
+    description:
+      Counts the number of times a metric was set to an out-of-range value.
+      The labels are the `category.name` identifier of the metric.
+    bugs:
+      - 1499761
+    data_reviews:
+      - TODO
```
mdboom: Going to file a data review on the bug.

mdboom: Filed a data review request.

liuche: Thanks for filing data review! Can you update these `data_reviews` fields to point to the bug you filed?

mdboom: Sure thing.  Done.

mdboom:
Now that we have the required schema changes merged, I think this is ready for final review.  I think my force pushing erased the earlier reviews, but not much has changed since on this (other than the schema git hash).

components/service/glean/metrics.yaml
```diff
@@ -161,3 +161,56 @@ glean.internal.metrics:
       - https://bugzilla.mozilla.org/show_bug.cgi?id=1525045#c18
     notification_emails:
       - telemetry-client-dev@mozilla.com
+      
+glean.error:
```
Dexterp37: Mh, one thing just occurred to me: why can't we use `GleanError.OutOfRange["label"].add()" and instead have to use the `recordError` trick?

components/service/glean/src/main/java/mozilla/components/service/glean/error/ErrorRecording.kt
```diff
@@ -0,0 +1,120 @@
...
+    message: String,
+    logger: Logger
+) {
+    val errorName = GLEAN_ERROR_NAMES[errorType]!!
+
+    val errorMetric = CounterMetricType(
+        disabled = false,
+        category = "glean.error",
+        lifetime = Lifetime.Ping,
+        name = "$errorName/${metricData.identifier.replace('/', '.')}",
```
Dexterp37: Is this the reason why we can't use the generated metrics? If so, given that we'll *always* want to not have `/` in the metirc name, can't this be in the general implementation of the storage engine? Moreover, wouldn't this still leave the `/` right after `$errorName/` in the name?

components/service/glean/src/main/java/mozilla/components/service/glean/error/ErrorRecording.kt
```diff
@@ -0,0 +1,120 @@
...
+
+// The labeled counter metrics that store the errors are defined in
+// the `metrics.yaml` for documentation purposes, but are not actually
+// used directly, since the `sendInPings` value needs to match the
+// pings of the metric that is erroring, not some constant value that
+// we could define in `metrics.yaml`.
+
+internal enum class ErrorType {
+    OutOfRange,
+    InvalidType,
```
Dexterp37: This seems to not be ever used: are we missing some instrumentation? If not, maybe we could remove it and add it later.

components/service/glean/metrics.yaml
mdboom: Because the errors need to be reported in the pings that the metric that caused the error are sent on.  In other words, we need to use the metric's `sendInPings`, not `GleanError.OutOfRange`'s `sendInPings`.

components/service/glean/src/main/java/mozilla/components/service/glean/error/ErrorRecording.kt
mdboom: The "/" acts as the boundary between the metric name and the label.  The error metric is a labeled metric where the labels are the name of the metric that caused the error.  That metric may itself be a labeled metric, hence the need to convert the "/" and make sure there's only one "/" anywhere in the name.

For example, if you have a labeled string metric and you do:

```
Category.MyLabeledMetric["foo"].set("too_long_string...")
```

the error should be reported using the following key in the storage engine:

```
glean.error.out_of_range/category.my_labeled_metric.foo
```

One could argue that we don't care about which specific label caused the error, in which case we would instead want to lop off the "/" and everything after it from `metricData.identifier` in the line above.  I could argue that either way.

components/service/glean/src/main/java/mozilla/components/service/glean/error/ErrorRecording.kt
mdboom: It was defined in the spec, but I'm happy to remove it for now.  There's also the question as to whether `OutOfRange` and `InvalidValue` are too redundant.  (See the open questions at the top of the PR).

components/service/glean/metrics.yaml
Dexterp37: This is interesting. I was assuming we'd only get error metrics in the `metrics` ping, but you raise a good point. I think we didn't spec this, hence the confusion :) I feel like having errors in the same ping metrics live might make more sense, but I'm a bit concerned about its behaviour: say that metric `A` goes in 3 pings, and there's an error with that metric. Do we send a count of 1 with each ping? Moreover, if some error happens on a metric that goes in the `baseline` or `events` ping, we'll increase the ping size there (though it's negligible).

components/service/glean/src/main/java/mozilla/components/service/glean/error/ErrorRecording.kt
Dexterp37: I see. Given the outcome of our "dots meeting", is this a problem? Do we need to watch out for `/` in keys, for this case?

mdboom: We do need to watch out for `/` in keys in the ping (and the testing we do against the schema is hopefully a guard against that).  But we still use `/` in the keys in the storage engines to markate the underlying labeled metric from the label.

So this shouldn't be a problem, but we do need to do at least one of the two options here:
- replacing `/` with `.` and thus reporting errors to the specific label for labeled metrics.
- cut the `/` and everything after it: thus reporting all errors to the "parent" labeled metric.

I'm starting to think the second option may be better (why would someone care that the error happened for search count on "Google" vs. another search engine, for example), but I don't fully understand the use cases.

components/service/glean/metrics.yaml
mdboom: > say that metric `A` goes in 3 pings, and there's an error with that metric. Do we send a count of 1 with each ping?

I'd say yes, with the added advantage that the error counters will be reset on the same cadence as the ping itself.  If we sent all errors in the metrics ping, we wouldn't know which ping timespan the errors happened in.  (Maybe that level of detail isn't useful, though).

But if the use case is just to collect some general health metrics on which metrics might have problems, it probably doesn't matter too much where we send them and having only one place to look for that might actually be an advantage.

Should we take this to a proposal?

components/service/glean/src/main/java/mozilla/components/service/glean/error/ErrorRecording.kt
Dexterp37: > I'm starting to think the second option may be better (why would someone care that the error happened for search count on "Google" vs. another search engine, for example), but I don't fully understand the use cases.

Yeah, plus this might get tricky privacy wise: say that we use user input as a key (we shouldn't, but we can't guarantee for the consumers of the API), and the key then we might end up sending stuff we don't want to.

components/service/glean/metrics.yaml
Dexterp37: > 
> 
> > say that metric `A` goes in 3 pings, and there's an error with that metric. Do we send a count of 1 with each ping?
> 
> I'd say yes, with the added advantage that the error counters will be reset on the same cadence as the ping itself. If we sent all errors in the metrics ping, we wouldn't know which ping timespan the errors happened in. (Maybe that level of detail isn't useful, though).
> 
> But if the use case is just to collect some general health metrics on which metrics might have problems, it probably doesn't matter too much where we send them and having only one place to look for that might actually be an advantage.

Good catch, different advantages depending on who's the consumer of the data: if it's us, trying to answer the question "is the API for metric X error-prone?", maybe it makes sense to have everything in one place and just have the absolute counts. If it's for developers, trying to see "is my metric behaving correctly?" then it makes sense to have that in their metrics' pings.

> Should we take this to a proposal?

Yes, I think this is good material for a proposal.

components/service/glean/src/main/java/mozilla/components/service/glean/error/ErrorRecording.kt
Dexterp37: Yeah, they feel redundant. +1 on removing that for now, and always using the same.

mdboom: Great.  On this point, I'll merge `OutOfRange` into `InvalidValue` and remove `InvalidType`.

mdboom:
I've updated this to take into account the feedback in the proposal, and I think it's ready for another look.

mdboom:
I've updated this to take into account the feedback in the proposal.

components/service/glean/src/test/java/mozilla/components/service/glean/TestUtil.kt
```diff
@@ -120,6 +121,7 @@ internal fun resetGlean(
     // Init glean.
     Glean.initialized = false
     Glean.initialize(context, config)
+    CountersStorageEngine.clearAllStores()
```
Dexterp37: Why is this needed? We're clearing all the stores at the beginning of this function already. If there's a good reason to do that, would you kindly add a line of comment about why (so that we don't forget about it?)?

mdboom: Oops.  It's not needed.

mdboom:
@Dexterp37: I've added the documentation to this PR.

Issue #1515, opened 2018-11-29
Awesome bar search suggestions: Encoding issue


pocmo:
![awesomebar](https://user-images.githubusercontent.com/89638/49215197-8aa6dc00-f3c8-11e8-8736-880768676a14.png)

pocmo:
The content type Google returns: `text/javascript; charset=ISO-8859-1`. We currently treat everything as UTF-8.

That's something to consider when switching to `concept-fetch`. OkHttp will use the charset automatically afaik. For the `HttpURLConnection` we'd have to implement that ourselves.

PR #2056, opened 2019-02-13
concept-fetch for awesomebar and response decoding fix

csadilek:
This brings in two commits/changes:

- Refactor our Awesomebar to use concept-fetch (tested with all implementations)
- A fix that makes sure all implementations of concept-fetch properly decode the response body based on the provided `content-type` header. This also fixes the encoding issues described in #1515 which happened (and now works) in all implementations (see attached screenshot)

screenshot_20190213-153003

### Pull Request checklist

### DIFF
--- a/components/concept/fetch/src/main/java/mozilla/components/concept/fetch/Response.kt
+++ b/components/concept/fetch/src/main/java/mozilla/components/concept/fetch/Response.kt
@@ -44,10 +44,28 @@ data class Response(
      * A [Body] returned along with the [Request].
      *
      * **The response body can be consumed only once.**.
+     *
+     * @param stream the input stream from which the response body can be read.
+     * @param contentType optional content-type as provided in the response
+     * header. If specified, an attempt will be made to look up the charset
+     * which will be used for decoding the body. If not specified, or if the
+     * charset can't be found, UTF-8 will be used for decoding.
      */
     open class Body(
-        private val stream: InputStream
+        private val stream: InputStream,
+        contentType: String? = null
     ) : Closeable, AutoCloseable {
+
+        @Suppress("TooGenericExceptionCaught")
+        private val charset = contentType?.let {
+            val charset = it.substringAfter("charset=")
+            try {
+                Charset.forName(charset)
+            } catch (e: Exception) {
+                Charsets.UTF_8
+            }
+        } ?: Charsets.UTF_8
+
         /**
          * Creates a usable stream from this body.
          *
@@ -64,8 +82,8 @@ data class Response(
          * Executes the given [block] function with the buffered reader as parameter and then closes it down correctly
          * whether an exception is thrown or not.
          */
-        fun <R> useBufferedReader(charset: Charset = Charsets.UTF_8, block: (BufferedReader) -> R): R = use {
-            block(stream.bufferedReader(charset))
+        fun <R> useBufferedReader(charset: Charset? = null, block: (BufferedReader) -> R): R = use {
+            block(stream.bufferedReader(charset ?: this.charset))
         }
 
         /**
@@ -73,7 +91,7 @@ data class Response(
          *
          * Takes care of closing the body down correctly whether an exception is thrown or not.
          */
-        fun string(charset: Charset = Charsets.UTF_8): String = useBufferedReader(charset) { it.readText() }
+        fun string(): String = useBufferedReader { it.readText() }
 
         /**
          * Closes this [Body] and releases any system resources associated with it.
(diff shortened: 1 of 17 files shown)


components/concept/fetch/src/main/java/mozilla/components/concept/fetch/Response.kt
```diff
@@ -44,10 +44,28 @@ data class Response(
...
      */
     open class Body(
-        private val stream: InputStream
+        private val stream: InputStream,
+        contentType: String? = null
     ) : Closeable, AutoCloseable {
+
+        @Suppress("TooGenericExceptionCaught")
+        private val charset = contentType?.let {
+            val charset = it.substringAfter("charset=")
```
pocmo: Ah, that's nice and simple. I wonder whether there are situations where this could not be the last value? Something like:

`Content-Type: text/html; charset=utf-8; foobar`

But I can't find any evidence that this is the case. :)

components/concept/fetch/src/main/java/mozilla/components/concept/fetch/Response.kt
```diff
@@ -44,10 +44,28 @@ data class Response(
...
+            try {
+                Charset.forName(charset)
+            } catch (e: Exception) {
+                Charsets.UTF_8
+            }
+        } ?: Charsets.UTF_8
```
pocmo: Interestingly this document says that the default for HTTP 1.1 is `ISO-8859-1`. I wonder what okhttp does?

https://www.w3.org/International/articles/http-charset/index#charset

components/concept/fetch/src/main/java/mozilla/components/concept/fetch/Response.kt
```diff
@@ -64,7 +82,7 @@ data class Response(
          * Executes the given [block] function with the buffered reader as parameter and then closes it down correctly
          * whether an exception is thrown or not.
          */
-        fun <R> useBufferedReader(charset: Charset = Charsets.UTF_8, block: (BufferedReader) -> R): R = use {
+        fun <R> useBufferedReader(block: (BufferedReader) -> R): R = use {
```
pocmo: What if the server does not return a `Content-Type` header but I want to read the response with a specific charset (because I expect something specific). Does it make sense to still allow to override the charset here (Instead of UTF_8 by default make it nullable)?

components/concept/fetch/src/main/java/mozilla/components/concept/fetch/Response.kt
csadilek: Yeah, I was looking too and couldn't find a case where this should break!

components/concept/fetch/src/main/java/mozilla/components/concept/fetch/Response.kt
csadilek: @pocmo okhttp uses the charset specified in the content-type header and if absent tries byte order mark (https://en.wikipedia.org/wiki/Byte_order_mark), then falls back to UTF-8.

csadilek: @pocmo I think we should stick with UTF-8 as fallback. Here's a good discussion why this is better (for XHR and JSON): https://github.com/AsyncHttpClient/async-http-client/issues/1439

csadilek: TL;DR: There's a bit of a misunderstanding/contradiction in the spec and the default charset depends on the mimetype. For application/JSON this should be UTF-8.

components/concept/fetch/src/main/java/mozilla/components/concept/fetch/Response.kt
csadilek: OK, I will add that.

csadilek:
@pocmo OK, PR updated and more tests added.

components/concept/fetch/src/main/java/mozilla/components/concept/fetch/Response.kt
pocmo: Alright :+1:

components/concept/fetch/src/main/java/mozilla/components/concept/fetch/Response.kt
```diff
@@ -64,16 +82,16 @@ data class Response(
...
+        fun <R> useBufferedReader(charset: Charset? = null, block: (BufferedReader) -> R): R = use {
+            block(stream.bufferedReader(charset ?: this.charset))
         }
 
         /**
          * Reads this body completely as a String.
          *
          * Takes care of closing the body down correctly whether an exception is thrown or not.
          */
-        fun string(charset: Charset = Charsets.UTF_8): String = useBufferedReader(charset) { it.readText() }
```
pocmo: Should we allow the charset override here too?

I'm going to merge now; we can do that in a follow-up if we want to. :)

csadilek: @pocmo OK, I will send a follow-up PR.

PR #2074, opened 2019-02-14
concept-fetch: Add API/flag to force network request..

csadilek:
Simple solution here (just a flag) to force a network request (ignore caches). We could add more API for various cache modes later, but there are significant differences between the three implementations plus we already have fine-grained control for caching behaviour using request headers.

### Pull Request checklist


pocmo:
> We could add more API for various cache modes later, but there are significant differences between the three implementations plus we already have fine-grained control for caching behaviour using request headers.

Fully agree with having only the "simple" solution for now. All those differences (and [flags in the JS fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Request/cache)) were the reason I omitted this in the initial implementation. :)

Most of the time a developer only wants to only declare: "From cache is okay" or "Never from cache". And that's totally possible now. :)

PR #2076, opened 2019-02-15
SyncManager: concept-sync, type shuffling, background execution

grigoryk:
This is a WIP. Reshuffling of Account types is stable, but the SyncManager interface is a bit fluid at the moment.


grigoryk:
There are still TODOs in the code, but this is ready for a preliminary review pass. I've updated history-sync sample app to demonstrate how this will function in practice.
Following the commits in their sequence should make reviewing these changes easier.

grigoryk:
Reference-browser integration companion PR: https://github.com/mozilla-mobile/reference-browser/pull/618

.buildconfig.yml
```diff
@@ -184,6 +188,10 @@ projects:
     path: components/service/sync-logins
     description: 'A library for integrating with Firefox Sync - Logins.'
     publish: true
+  service-sync:
```
pocmo: Discussed on Slack: I think that is a `feature` :)

components/service/sync/src/test/java/mozilla/components/service/sync/GeneralSyncManagerTest.kt
```diff
@@ -0,0 +1,287 @@
...
+//        verify(dispatcher, times(3)).startPeriodicSync(any(), any())
+
+        manager.removeStore("history")
+        assertEquals(1, storez!!.size)
+        assertTrue(storez!!.contains("logins"))
+        assertEquals(mockAccount, akaunt)
+
+        verify(dispatcher, times(4)).syncNow()
+//        verify(dispatcher, times(4)).startPeriodicSync(any(), any())
+    }
```
pocmo: A couple of `verify` calls are commented out?

PR #2082, opened 2019-02-15
Closes #2077: Toolbar: Allow returning Boolean from "URL commit listener" to control switching to display mode.

### LINKED ISSUE #2077: BrowserToolbar enters Display Mode onURLEntered flashing in Fenix SearchFragment
Currently, BrowserToolbar enters display mode when a URL is entered, but in Fenix we have two separate fragments for display + edit (SearchFragment and BrowserFragment) and on slow phones you can see a flash of display mode in the SearchFragment right after the URL is entered. Would it be possible to not automatically enter displayMode here or we can brainstorm other solutions too! 

```
internal fun onUrlEntered(url: String) {
        displayMode()

        urlCommitListener?.invoke(url)
    }
```

pocmo:
### Pull Request checklist

### DIFF
--- a/components/browser/toolbar/src/test/java/mozilla/components/browser/toolbar/BrowserToolbarTest.kt
+++ b/components/browser/toolbar/src/test/java/mozilla/components/browser/toolbar/BrowserToolbarTest.kt
@@ -21,6 +21,7 @@ import mozilla.components.browser.toolbar.edit.EditToolbar
 import mozilla.components.concept.toolbar.Toolbar
 import mozilla.components.concept.toolbar.Toolbar.SiteSecurity
 import mozilla.components.support.base.android.Padding
+import mozilla.components.support.ktx.android.view.isGone
 import mozilla.components.support.ktx.android.view.isVisible
 import mozilla.components.support.test.mock
 import org.junit.Assert.assertEquals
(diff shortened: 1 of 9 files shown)


components/browser/toolbar/src/test/java/mozilla/components/browser/toolbar/BrowserToolbarTest.kt
```diff
@@ -229,6 +230,36 @@ class BrowserToolbarTest {
         assertTrue(toolbar.editToolbar.visibility == View.GONE)
     }
 
+    @Test
+    fun `toolbar will switch back to display mode if URL commit listener returns true`() {
+        val toolbar = BrowserToolbar(RuntimeEnvironment.application)
+        toolbar.setOnUrlCommitListener { true }
+        toolbar.editMode()
+
+        assertTrue(toolbar.displayToolbar.visibility == View.GONE)
```
Amejia481: nit: You could use `toolbar.editToolbar.isGone()` and `toolbar.editToolbar.isVisible()` extension functions :)

pocmo: Yeah, true. I'm less concerned about the "complicated, verbose way" in tests since I want to test a very specific thing and not call into half our utilities. :)

pocmo:
@Amejia481 Okay, updated the PR and now it's using isGone/isVisible. :)

PR #2083, opened 2019-02-15
Closes #2023: browser-toolbar: Allow theming of autocomplete suggestions foreground/background colors..

### LINKED ISSUE #2023: Allow theming of autocomplete highlight color
Currently I am unable to change the autocomplete color from purple, which means it doesn't match the rest of my design. Is it possible to add the ability to change this color?

Example:

pocmo:
### Pull Request checklist

### DIFF
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
@@ -5,6 +5,7 @@
 package mozilla.components.browser.toolbar
 
 import android.content.Context
+import android.graphics.Color
 import android.graphics.Typeface
 import android.graphics.drawable.Drawable
 import android.support.annotation.DrawableRes
@@ -123,6 +124,9 @@ class BrowserToolbar @JvmOverloads constructor(
         get() = displayToolbar.urlBoxView
         set(value) { displayToolbar.urlBoxView = value }
 
+    /**
+     * Gets/Sets the color tint of the menu button.
+     */
     var menuViewColor: Int
         get() = displayToolbar.menuViewColor
         set(value) { displayToolbar.menuViewColor = value }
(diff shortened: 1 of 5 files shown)


pocmo:
> @pocmo the XML attributes in the `README.md` files are a good solution, is there are way to integrate them into our API docs as the android documentation does?

That's a great question. I don't think dokka has any clue about the XML attributes.

This seems related:
https://github.com/Kotlin/dokka/issues/338

We could link from the class doc to the other doc though.

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
```diff
@@ -190,6 +194,20 @@ class BrowserToolbar @JvmOverloads constructor(
...
+     * The background color used for autocomplete suggestions in edit mode.
+     */
+    var suggestionBackgroundColor: Int
+        get() = editToolbar.urlView.autoCompleteBackgroundColor
+        set(value) { editToolbar.urlView.autoCompleteBackgroundColor = value }
+
+    /**
+     * The foreground color used for autocomplete suggestions in edit mode.
+     */
+    var suggestionForegroundColor: Int?
```
sblatz: Can you explain the differences between foreground and background here? I only see one purplish color right now and want to make sure we're referencing the same thing 😄

pocmo: Sure!

Stealing images from the Android docs (since we are using `BackgroundColorSpan` and `ForegroundColorSpan` under the hood):

Background:
image

Foreground:
image

By default the toolbar will use **the same color** for the suggestion text (foreground) as it uses for the other text. But imagine you want to have black text and also a black suggestion background. In this case you want to change the foreground/text color of the suggestion to something else.

For example this:

```
            app:browserToolbarTextColor="#ff880E4F"
            app:browserToolbarSuggestionBackgroundColor="#F50057"
            app:browserToolbarSuggestionForegroundColor="#F48FB1"
```

turns into:

mozilla

Or this:

```
            app:browserToolbarTextColor="#000000"
            app:browserToolbarSuggestionBackgroundColor="#000000"
            app:browserToolbarSuggestionForegroundColor="#FFFFFF"
```

into:

moz2

PR #2091, opened 2019-02-15
Closes #2078: Add "Browsers" helper class.

### LINKED ISSUE #2078: Add Browser Utils for Default Browser Logic
[Discussed in Fenix](https://github.com/mozilla-mobile/fenix/issues/420), we want to be able to determine if we are the current default browser. Focus [had a Util File for it](https://github.com/mozilla-mobile/focus-android/blob/master/app/src/main/java/org/mozilla/focus/utils/Browsers.java), if this could be useful across multiple Android browsers, should we make it part of AC?

pocmo:
This is a nicer Kotlin version of the code we have in Focus. Regarding tests I'm going to file a follow-up. Mocking the packageManager parts is a bit annoying. But I tested the code on a device.

### Pull Request checklist

### DIFF
--- a/components/support/utils/src/main/java/mozilla/components/support/utils/Browsers.kt
+++ b/components/support/utils/src/main/java/mozilla/components/support/utils/Browsers.kt
@@ -1,0 +1,289 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.support.utils
+
+import android.content.Context
+import android.content.Intent
+import android.content.pm.ActivityInfo
+import android.content.pm.PackageManager
+import android.net.Uri
+import java.util.HashMap
+
+/**
+ * Helpful tools for dealing with other browsers on this device.
+ *
+ * ```
+ * // Collect information about all installed browsers:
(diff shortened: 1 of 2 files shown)


components/support/utils/src/main/java/mozilla/components/support/utils/Browsers.kt
```diff
@@ -0,0 +1,282 @@
...
+        SAMSUNG_INTERNET("com.sec.android.app.sbrowser"),
+        DOLPHIN_BROWSER("mobi.mgeek.TunnyBrowser"),
+        BRAVE_BROWSER("com.brave.browser"),
+        LINK_BUBBLE("com.linkbubble.playstore"),
+        ADBLOCK_BROWSER("org.adblockplus.browser"),
+        CHROMER("arun.com.chromer"),
+        FLYNX("com.flynx"),
+        GHOSTERY_BROWSER("com.ghostery.android.ghostery"),
+        DUCKDUCKGO("com.duckduckgo.mobile.android")
+    }
```
jonalmeida: Microsoft Edge-ium?

Amejia481: This is the applicationId   `com.microsoft.emmx`

pocmo: Added!

Issue #1329, opened 2018-11-12
Replace custom http client implementations in components with concept-fetch


pocmo:
Breaking down: #1012 

* [x] service-glean: Ping upload
* [x]  service-telemetry: Ping upload
* [x]  service-fretboard: Configuration download (Kinto)
* [x]  feature-awesomebar: search suggestions
* [ ] maybe: lib-crash (we are currently using the uploader from GeckoView; but maybe we want to include that functionality oursleves and avoid having that dependency)

PR #2092, opened 2019-02-15
Issue #1329: Update service-fretboard to use concept-fetch

csadilek:
Took a stab at updating kinto/service-fretboard to use concept-fetch. I didn't like the fact that I have to specify a `client` when constructing the `KintoExperimentSource` so I made it default to `HttpUrlConncectionClient`. Not sure here? This way it comes without API changes though. 

### Pull Request checklist


components/service/fretboard/src/main/java/mozilla/components/service/fretboard/source/kinto/KintoExperimentSource.kt
```diff
@@ -18,14 +20,14 @@ import org.json.JSONObject
...
- * @property client http client to use
+ * @param client http client to use
  */
 class KintoExperimentSource(
     private val baseUrl: String,
     private val bucketName: String,
     private val collectionName: String,
     private val validateSignature: Boolean = false,
-    client: HttpClient = HttpURLConnectionHttpClient()
+    client: Client = HttpURLConnectionClient()
```
pocmo: Now with `concept-fetch` I'd stop using `HttpURLConnection` by default (and having the dependency) and let the app define the `Client` (`HttpURLConnectionClient` is not something we actually recommend using unless you need a very small APK size). Currently an app already needs to construct and pass in `KintoExperimentSource` - so asking for a `Client` here doesn't seem an issue.

csadilek: @pocmo OK, updated and changelog / README note added for breaking API.

PR #2090, opened 2019-02-15
Sitepermission microphone

Amejia481:
[Demo 🌮 ](https://drive.google.com/file/d/14EJu0bTlAqCibeEaHCoC2e_tKPUiGDK3/view?usp=sharing)

### Pull Request checklist

- [x] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes a changelog entry or does not need one
- [x] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features

PR #2093, opened 2019-02-16
Site permissions camera.

Amejia481:
[Demo 🌮](https://drive.google.com/file/d/13fVZbWiT_Ao1uccNpIabxRlTmdVDxfwB/view?usp=sharing)
Depends on 🔴  : https://github.com/mozilla-mobile/android-components/pull/2090

### Pull Request checklist


components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
```diff
@@ -32,13 +40,17 @@ typealias OnNeedToRequestPermissions = (permissions: Array<String>) -> Unit
...
  * @property sessionManager the [SessionManager] instance in order to subscribe
  * to the selected [Session].
  * @property onNeedToRequestPermissions a callback invoked when permissions
  * need to be requested. Once the request is completed, [onPermissionsResult] needs to be invoked.
  **/
+
+@Suppress("TooManyFunctions")
 class SitePermissionsFeature(
     private val anchorView: View,
+    @ColorRes private val iconsTintColorResource: Int = android.R.color.black,
```
pocmo: I wonder if we can find a better solution here: Maybe start with just using the same tint as the theme text color? Overall this looks like a good candidate for a theme/style since we want to style **ALL** instances and not pass in a customized color for every single prompt.

components/feature/sitepermissions/src/main/res/values/strings.xml
```diff
@@ -38,8 +38,11 @@
     access to this permission-->
     <string name="mozac_feature_sitepermissions_not_allow">Don\'t allow</string>
 
-    <!-- Accessibility text for indicating the permission icon in a permission request dialog -->
-    <string name="mozac_feature_sitepermissions_accessibility_permission_icon">Permission Icon</string>
+    <!-- Accessibility text for indicating the permission notification icon in a permission request dialog -->
+    <string name="mozac_feature_sitepermissions_accessibility_permission_location_icon">Location Icon</string>
+
+    <!-- Accessibility text for indicating the permission notification icon in a permission request dialog -->
+    <string name="mozac_feature_sitepermissions_accessibility_permission_notification_icon">Notification Icon</string>
```
pocmo: I think those content descriptions serve no purpose. Note that the ImageViews are set to ` android:importantForAccessibility="no"`. So they shouldn't require a content description.

components/ui/doorhanger/src/main/java/mozilla/components/ui/doorhanger/DoorhangerPrompt.kt
```diff
@@ -129,6 +140,9 @@ class DoorhangerPrompt(
                 return (view as android.widget.RadioButton).apply {
                     this.id = viewId
                     this.text = label
+                    this.setOnClickListener {
+                        onClick?.invoke()
+                    }
```
pocmo: This shouldn't be needed since the group already updates the checked state. You can get the checked item from the control group.

components/feature/sitepermissions/src/main/res/values/strings.xml
Amejia481: Good catch I didn't notice it 👍

components/ui/doorhanger/src/main/java/mozilla/components/ui/doorhanger/DoorhangerPrompt.kt
Amejia481: 👍

Amejia481:
@pocmo  most of these issue came from this parent PR #2090 I've already fixed them there, could you take a look :)

Amejia481:
@pocmo I updated the PR!

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
```diff
@@ -180,6 +185,26 @@ class SitePermissionsFeature(
...
+            }
+            is ContentVideoCamera -> {
+                createPromptForCameraPermission(
+                    context,
+                    uriString,
+                    permissionRequest,
+                    session,
+                    allowButtonTitle,
+                    denyButton
+                )
```
pocmo: Is this exactly the same? Could this be reduced to `is ContentVideoCapture or is ContentVideoCamera` or something like that?

Amejia481: I thought the same thing, but I couldn't find the right syntax to make it work, I've tried multiple alternatives without any luck. Any suggestions? :)



image

image

image

pocmo: Oh, okay :(

Amejia481: The logic says it should be right to do it, because in a java `switch` is possible, but maybe it has something to do with compiler want a explicit declaration per branch.

PR #2104, opened 2019-02-18
[glean] Enable runtime glean debugging options

Dexterp37:
This allows developers to enable glean debugging options, at runtime, using adb. It currently only supports logging pings to logcat, but more options will be enabled in the future.

This introduces the concept of `NullConfiguration`, which is a `Configuration`-like object with all entries defaulting to `null`. This is used for simplifying merging between debug provided configuration and the product provided config.

### Pull Request checklist


components/service/glean/README.md
```diff
@@ -101,6 +103,22 @@ Every application must provide a way to disable and re-enable data collection an
...
+Glean exports the [`GleanDebugActivity`](src/main/java/mozilla/components/service/glean/debug/GleanDebugActivity.kt)
+that can be used to toggle debugging features on or off. Users can invoke this special activity, at
+run-time, using the following [`adb`](https://developer.android.com/studio/command-line/adb) command:
+
+`adb shell am start -n [package.name]/mozilla.components.service.glean.debug.GleanDebugActivity [extra keys]`
+
+Where the `[package.name]` is the product's package name as exported in the manifest file (e.g.
+`org.mozilla.samples.glean` for the glean sample application) and `[extra keys]` is a list of
+extra keys to be passed to the debug activity. See the [documentation](https://developer.android.com/studio/command-line/adb#IntentSpec)
+for the command line switches used to pass the extra keys. These are the currently supported keys:
```
georgf: Can you give an example of a full working command line (that i could just copy & paste)?

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -80,10 +93,17 @@ open class GleanInternalAPI internal constructor () {
             return
         }
 
+        // Monkey-patch the product-provided configuration with the data coming
+        // from the debug configuration, if available. This is required in addition to
+        // `debugConfiguration`'s setter in case the debug configuration got set before
+        // `configuration` was available.
+        this.configuration = if (debugConfiguration != null) {
+            configuration.setConfigOverride(debugConfiguration!!)
+        } else { configuration }
```
pocmo: Can this ever happen? The only situation I can imagine is when Glean was not initialized in the application object. But that seems like an error?

components/service/glean/src/main/java/mozilla/components/service/glean/net/HttpPingUploader.kt
```diff
@@ -33,17 +32,17 @@ internal class HttpPingUploader(configuration: Configuration) : PingUploader {
      * @param data the serialized text data to send
      */
     private fun logPing(path: String, data: String) {
-        if (config.logPings) {
+        if (Glean.configuration.logPings == true) {
```
pocmo: Currently you are copying the values over; so I assume you could still use the passed in configuration? Anyhow, instead of accessing the configuration via the singleton (which is awkward in tests etc.) could we pass the configuration that should be used to `upload()`? After all that's the part that requires a configuration.

components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugActivity.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        // Enable debugging options and start the application.
+        this.intent.extras?.let {
+            var debugConfig = NullConfiguration()
+            if (it.getBoolean("logPings")) {
+                debugConfig = debugConfig.copy(logPings = true)
+            }
```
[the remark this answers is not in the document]
Dexterp37: Oh, this is a smart one. I like this!

components/service/glean/src/main/java/mozilla/components/service/glean/net/HttpPingUploader.kt
Dexterp37: This is a good point. I'm now changing the configuration object as you suggested early on in the review, so it's ok to pass the `configuration` object to upload :)

PR #2120, opened 2019-02-20
[glean] Create a way for glean test API functions to await async IO operations

travis79:
Use the Job returned from calls to Dispatchers.API.launch to wait for the IO operation to complete before test functions return values.
Add a way out to avoid infinite loop situations

### Pull Request checklist

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/BooleanMetricType.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/BooleanMetricType.kt
@@ -5,6 +5,7 @@
 package mozilla.components.service.glean
 
 import android.support.annotation.VisibleForTesting
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
 import mozilla.components.service.glean.storages.BooleansStorageEngine
 import mozilla.components.support.base.log.logger.Logger
@@ -29,6 +30,10 @@ data class BooleanMetricType(
 
     private val logger = Logger("glean/BooleanMetricType")
 
+    // Holds the Job returned from launch{} for awaiting purposes
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    var ioTask: Job? = null
+
     /**
      * Set a boolean value.
      *
@@ -41,7 +46,7 @@ data class BooleanMetricType(
             return
         }
 
-        Dispatchers.API.launch {
+        ioTask = Dispatchers.API.launch {
             // Delegate storing the boolean to the storage engine.
             BooleansStorageEngine.record(
                 this@BooleanMetricType,
@@ -51,7 +56,9 @@ data class BooleanMetricType(
     }
 
     /**
-     * Tests whether a value is stored for the metric for testing purposes only
+     * Tests whether a value is stored for the metric for testing purposes only. This function will
+     * attempt to await the last task (if any) writing to the the metric's storage engine before
+     * returning a value.
      *
      * @param pingName represents the name of the ping to retrieve the metric for.  Defaults
      *                 to the either the first value in [defaultStorageDestinations] or the first
@@ -60,11 +67,14 @@ data class BooleanMetricType(
      */
     @VisibleForTesting(otherwise = VisibleForTesting.NONE)
     fun testHasValue(pingName: String = getStorageNames().first()): Boolean {
+        ioTask?.let { awaitJob(it) }
+
         return BooleansStorageEngine.getSnapshot(pingName, false)?.get(identifier) != null
     }
 
     /**
-     * Returns the stored value for testing purposes only
+     * Returns the stored value for testing purposes only. This function will attempt to await the
+     * last task (if any) writing to the the metric's storage engine before returning a value.
      *
      * @param pingName represents the name of the ping to retrieve the metric for.  Defaults
      *                 to the either the first value in [defaultStorageDestinations] or the first
@@ -74,6 +84,8 @@ data class BooleanMetricType(
      */
     @VisibleForTesting(otherwise = VisibleForTesting.NONE)
     fun testGetValue(pingName: String = getStorageNames().first()): Boolean {
+        ioTask?.let { awaitJob(it) }
+
         return BooleansStorageEngine.getSnapshot(pingName, false)!![identifier]!!
     }
 }
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/CommonMetricData.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/CommonMetricData.kt
@@ -4,6 +4,11 @@
 
 package mozilla.components.service.glean
 
+import android.support.annotation.VisibleForTesting
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.TimeoutCancellationException
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withTimeout
 import mozilla.components.support.base.log.logger.Logger
 
 /**
@@ -39,6 +44,7 @@ interface CommonMetricData {
 
     companion object {
         internal const val DEFAULT_STORAGE_NAME = "default"
+        private const val JOB_TIMEOUT_MS = 250L
     }
 
     fun shouldRecord(logger: Logger): Boolean {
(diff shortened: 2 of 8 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/BooleanMetricType.kt
```diff
@@ -58,8 +62,16 @@ data class BooleanMetricType(
      *                 value in [sendInPings]
      * @return true if metric value exists, otherwise false
      */
+    @Suppress("MagicNumber")
     @VisibleForTesting(otherwise = VisibleForTesting.NONE)
     fun testHasValue(pingName: String = getStorageNames().first()): Boolean {
+        if (ioTask != null) {
+            var loopCnt = 50
+            while (ioTask!!.isActive && loopCnt > 0) {
+                --loopCnt
```
mdboom: Would it help to have a short "sleep" here?  I'm sure Java is different, but at least on Python, if you deadspin while waiting on another thread like this, you're much less likely to give the other thread a chance to finish, so `sleep`'ing can help.

components/service/glean/src/main/java/mozilla/components/service/glean/BooleanMetricType.kt
```diff
@@ -58,8 +62,16 @@ data class BooleanMetricType(
...
+            }
+        }
```
mdboom: If `loopCnt` is `0` here, maybe log a message to the console that we timed out waiting for metric recording to finish.  That will help sort out failures for some other reason on the line below from failures due to timing out.  (And if we started seeing a lot of those messages next to seemingly random failures, we'd know that we'd need to revisit this solution.)

mdboom: Also, maybe this code block could be put in a helper function (where `ioTask` is passed in) to avoid the copy-and-paste.

components/service/glean/src/main/java/mozilla/components/service/glean/BooleanMetricType.kt
travis79: I considered that, but Thread.sleep() is usually frowned upon for busy loops.  I totally agree that relinquishing the processor would be a good idea here, maybe since this is only for tests it would be acceptable to have a 5 ms sleep or something?

components/service/glean/src/main/java/mozilla/components/service/glean/BooleanMetricType.kt
travis79: I like it!

travis79:
@mdboom I am also a little uneasy about trusting the scheduling of tasks to be done in any order, but since this should only really be used in tests I think it's safe enough.  If someone wants to test how quickly they can update a metric value, then they need to take precautions in the test to ensure that they all get completed by using `runBlocking {}` or similar.

components/service/glean/src/main/java/mozilla/components/service/glean/CommonMetricData.kt
```diff
@@ -76,4 +79,25 @@ interface CommonMetricData {
...
+     *
+     * @param job Job that is to be awaited
+     * @return true if job completed in the given time, otherwise false if it timed out
+     */
+    fun awaitJob(job: Job): Boolean {
+        var loopCnt = MAX_WAIT_LOOP_CNT
+        while (job.isActive && loopCnt > 0) {
+            --loopCnt
+            Thread.sleep(MAX_WAIT_SLEEP_DELAY)
+        }
```
pocmo: Since you already have a `Job`here. Can't you [join](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-job/join.html) on that? You may need to wrap that in [runBlocking](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/run-blocking.html). Is that something that would work here?

travis79: I initially looked at doing this but was nervous about somehow blocking indefinitely with `join()`, which is why I went with a polled solution that had an out after a reasonable amount of time.

pocmo: You may be able to work around that by using [withTimeout](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/with-timeout.html), I think.

pocmo: You may be able to work around that by using [withTimeout](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/with-timeout.html), I think.

travis79: Thanks!  I like this much better now 👍

mdboom: Yeah.  It feels like at least if something isn't right it's not our problem ;)

components/service/glean/src/main/java/mozilla/components/service/glean/BooleanMetricType.kt
travis79: Using `withTimeout` made this a little more difficult to accomplis.   It should throw a TimeoutCancellationException which I suppose we could catch and log, but since this is only used from a test then maybe throwing is better than logging?

components/service/glean/src/main/java/mozilla/components/service/glean/BooleanMetricType.kt
```diff
@@ -60,6 +65,8 @@ data class BooleanMetricType(
      */
     @VisibleForTesting(otherwise = VisibleForTesting.NONE)
     fun testHasValue(pingName: String = getStorageNames().first()): Boolean {
+        ioTask?.let { awaitJob(it) }
```
Dexterp37: Mh, what if we want to test that something was **not** written? Then await will throw, right? I think waiting should really be a separate operation to enable such test-cases.

**EDIT** - Woha. Reading the code again, looks like that test-case would still be possible, at least the first time the metric is used! That's because if nothing is written, then `ioTask` would be `null`. *However*, we're never resetting `ioTask` to `null`. But that is still not a problem, as we'd try to wait on a job that has already completed. The case I'm thinking of, is:

- `testHasValue` (with `ioTask = null`), should return `false`
- Record
- `testHasValue` (with `ioTask` as ongoing job), will return `true` after the job completes
- `testHasValue` (with `ioTask` as completed job), will return `true` immediately, because the job has already completed

Is my understanding correct? If so, I think we should at least mention in the docstring that this function waits for any outstanding recording operation to complete.

components/service/glean/src/main/java/mozilla/components/service/glean/BooleanMetricType.kt
```diff
@@ -74,6 +81,8 @@ data class BooleanMetricType(
      */
     @VisibleForTesting(otherwise = VisibleForTesting.NONE)
     fun testGetValue(pingName: String = getStorageNames().first()): Boolean {
+        ioTask?.let { awaitJob(it) }
```
Dexterp37: nit: let's mention in the docstring that this waits for any outstanding recording operation to complete.

components/service/glean/src/main/java/mozilla/components/service/glean/CommonMetricData.kt
```diff
@@ -39,6 +44,7 @@ interface CommonMetricData {
 
     companion object {
         internal const val DEFAULT_STORAGE_NAME = "default"
+        private const val JOB_TIMEOUT = 250L
```
Dexterp37: What's the time unit for this? Milliseconds? if so, let's rename this to `JOB_TIMEOUT_MS` (i.e. adding the `_MS` postfix)

components/service/glean/src/main/java/mozilla/components/service/glean/CommonMetricData.kt
Dexterp37: This looks good!

components/service/glean/src/main/java/mozilla/components/service/glean/BooleanMetricType.kt
travis79: I think that's a great idea to add the awaiting information to the docstring for the test functions, and yes, your understanding is correct.  If no one has called a metric function that uses `Dispatchers.API` which would assign the `Job` value to `ioTask`, then it will just skip the `awaitJob` and should return false with `testHasValue` or throw (as intended) with `testGetValue`.

travis79:
@Dexterp37 this is ready for another look.  I've updated the docstrings to indicate that about the awaiting behavior and renamed the job timeout constant to include the unit.

PR #2136, opened 2019-02-20
Closes #2024: Adds theming to menu background

### LINKED ISSUE #2024: Allow background color theming of toolbar menu
It'd be great if we could theme this background color to better match the theme of Fenix!

sblatz:
### Pull Request checklist

_Note: This is my first a-c contribution, so don't hold back if I messed some things up!_


components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenu.kt
```diff
@@ -22,7 +23,8 @@ import mozilla.components.support.ktx.android.view.isRTL
  * A popup menu composed of BrowserMenuItem objects.
  */
 class BrowserMenu internal constructor(
-    private val adapter: BrowserMenuAdapter
+    private val adapter: BrowserMenuAdapter,
+    private val cardBackgroundColor: Int? = null
```
jonalmeida: Maybe we can do this differently by retrieving the background colour from an attribute instead of having it passed through the constructor.

Consider what would happen if we wanted to add more styling attributes, do we continue adding more constructor args? Is a background colour really needed for constructing a BrowserMenu?

There is an `init` block in [BrowserToolbar.kt](https://github.com/mozilla-mobile/android-components/blob/master/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt#L275-L320) that gets styled attributes from XML which might be an option. Although, I don't think you need the Kotlin API for this that class does.

jonalmeida:
Congrats on your first contribution! 🎉 🎉

components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenu.kt
pocmo: I agree that it would be nice if we could reduce the number of parameters here. I don't think we can use XML attributes though: The menu is created by the builder and never placed in XML. Having those attributes on the toolbar may not make sense if the menu component can be used independently. Even if we add it to the toolbar: We would still need to pass it to the builder and/or menu.

So I wonder if this could be a great candidate for a style that can be overridden in the app theme though.

(nit: Let's not call it `cardBackgroundColor` since the usage of `CardView` is an implementation detail here)

pocmo: This looks like a great article describing the approach:
https://blog.danlew.net/2011/03/03/custom_styles_in_android_libraries/

pocmo: Just noticed that the menu already has a bunch of styles :)

https://github.com/mozilla-mobile/android-components/blob/master/components/browser/menu/src/main/res/values/style.xml

sblatz: My initial thought was the same as yours, @jonalmeida but as @pocmo mentioned the "menu" was never sticking around to be modified (simply created by the builder). I will take a look at the style file you linked, Sebastian.

Thanks for all the great feedback everyone 😄

sblatz:
Made the requested changes. Ready for review again 😄

components/browser/menu/src/main/res/values/style.xml
```diff
@@ -3,6 +3,11 @@
    - License, v. 2.0. If a copy of the MPL was not distributed with this
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
 <resources>
+
+    <style name="Mozac.Browser.Menu" parent="">
+        <item name="cardBackgroundColor">?attr/menuBackgroundColor</item>
```
Amejia481: This is could cause a silent run time crash, If the app doesn't override the `Mozac.Browser.Menu` style and define the value of `?attr/menuBackgroundColor`.

To see the issue, run the `samples-browsers`app  and tap the 3 dots menu:

An alternative could be instead of using the attribute `menuBackgroundColor`, we could have a default color and let the app override it.

Something like:

**menu/colors.xml**
```xml
<resources>
    <color name="mozac_browser_menu_background">#E91E63</color>
</resources>
```
**menu/style.xml**
```xml
 <style name="Mozac.Browser.Menu" parent="">
      <item name="cardBackgroundColor">@color/mozac_browser_menu_background</item>
</style>
```

In the app
**color.xml**
```xml
<resources>
    <color name="mozac_browser_menu_background" tools:ignore="UnusedResources">YOU_COLOR</color>
</resources>
```
If you go with this approach please update the [docs](https://github.com/mozilla-mobile/android-components/blob/master/components/browser/menu/README.md) :)

```
2019-02-21 11:33:58.082 4476-4476/org.mozilla.samples.browser E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.mozilla.samples.browser, PID: 4476
    android.view.InflateException: Binary XML file line #5: Binary XML file line #5: Error inflating class android.support.v7.widget.CardView
    Caused by: android.view.InflateException: Binary XML file line #5: Error inflating class android.support.v7.widget.CardView
    Caused by: java.lang.reflect.InvocationTargetException
        at java.lang.reflect.Constructor.newInstance0(Native Method)
        at java.lang.reflect.Constructor.newInstance(Constructor.java:343)
        at android.view.LayoutInflater.createView(LayoutInflater.java:647)
        at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:790)
        at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:730)
        at android.view.LayoutInflater.inflate(LayoutInflater.java:492)
        at android.view.LayoutInflater.inflate(LayoutInflater.java:423)
        at android.view.LayoutInflater.inflate(LayoutInflater.java:374)
        at mozilla.components.browser.menu.BrowserMenu.show(BrowserMenu.kt:31)
        at mozilla.components.browser.toolbar.display.DisplayToolbar$$special$$inlined$apply$lambda$2.onClick(DisplayToolbar.kt:121)
        at android.view.View.performClick(View.java:6597)
        at android.view.View.performClickInternal(View.java:6574)
        at android.view.View.access$3100(View.java:778)
        at android.view.View$PerformClick.run(View.java:25885)
        at android.os.Handler.handleCallback(Handler.java:873)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:193)
        at android.app.ActivityThread.main(ActivityThread.java:6718)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
     Caused by: java.lang.UnsupportedOperationException: Failed to resolve attribute at index 2: TypedValue{t=0x2/d=0x7f03017b a=3}
        at android.content.res.TypedArray.getColorStateList(TypedArray.java:546)
        at android.support.v7.widget.CardView.<init>(CardView.java:127)
        at android.support.v7.widget.CardView.<init>(CardView.java:117)
        at java.lang.reflect.Constructor.newInstance0(Native Method) 
        at java.lang.reflect.Constructor.newInstance(Constructor.java:343) 
        at android.view.LayoutInflater.createView(LayoutInflater.java:647) 
        at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:790) 
        at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:730) 
        at android.view.LayoutInflater.inflate(LayoutInflater.java:492) 
        at android.view.LayoutInflater.inflate(LayoutInflater.java:423) 
        at android.view.LayoutInflater.inflate(LayoutInflater.java:374) 
        at mozilla.components.browser.menu.BrowserMenu.show(BrowserMenu.kt:31) 
        at mozilla.components.browser.toolbar.display.DisplayToolbar$$special$$inlined$apply$lambda$2.onClick(DisplayToolbar.kt:121) 
        at android.view.View.performClick(View.java:6597) 
        at android.view.View.performClickInternal(View.java:6574) 
        at android.view.View.access$3100(View.java:778) 
        at android.view.View$PerformClick.run(View.java:25885) 
        at android.os.Handler.handleCallback(Handler.java:873) 
        at android.os.Handler.dispatchMessage(Handler.java:99) 
        at android.os.Looper.loop(Looper.java:193) 
        at android.app.ActivityThread.main(ActivityThread.java:6718) 
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493) 
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858) 
    

    --------- beginning of system
```

sblatz: Good catch on this. I'm still a bit new to applying themes through XML and struggle with understanding the linking--this was really helpful!

Amejia481: :)

pocmo: Does the color approach really work since Fenix uses different themes and therefore needs to support multiple colors.

sblatz: @pocmo I think the problem is even if you *don't* override the style you have a crash. So you're required to set a style for it. It still seems like a nicer solution than overriding a color, but it seems confusing that a user would have to set it.

Amejia481: @pocmo  if you `mozac_browser_menu_background` you will only need to change the color in the app, but if you override the style you will have to define all the values.

Amejia481: I suggested the approach of override the color, because if you override the style you will have to define all the values, even the ones that you are not interested in changing.

For instance, we  have:

**In AC**
```XML
 <style name="Mozac.Browser.Menu" parent="">
      <item name="cardBackgroundColor">#A87687</item>
        <item name="android:paddingEnd">10dp</item>
</style>
```

**In the app:**
```XML
 <style name="Mozac.Browser.Menu" parent="">
      <item name="cardBackgroundColor">NEW_COLOR</item>
</style>
```

The `paddingEnd` won't be applied just the `cardBackgroundColor`.

If want to do both you have to copy the complete style from **AC**

 ```xml
<style name="Mozac.Browser.Menu" parent="">
      <item name="cardBackgroundColor">NEW_COLOR</item>
      <item name="android:paddingEnd">10dp</item>
</style>
```

**Comment by sblatz** [comment]
Now using the updated "override color" solution that @Amejia481 suggested.

File: docs/changelog.md [hunk 3]
```diff
@@ -12,6 +12,13 @@ permalink: /changelog/
 * [Gecko](https://github.com/mozilla-mobile/android-components/blob/master/buildSrc/src/main/java/Gecko.kt)
 * [Configuration](https://github.com/mozilla-mobile/android-components/blob/master/buildSrc/src/main/java/Config.kt)
 
+* **browser-menu**
+  * Added option to set background color by overriding `mozac_browser_menu_background` color resource.
+    ```xml
+        <color name="mozac_browser_menu_background">DESIRED_COLOR</color>
+    ```
+    
+
```
Amejia481: nit: I think there are some extra white lines

Amejia481:
What do you think @pocmo?

sblatz:
The only major issue I see with this solution is it's kind of backwards of how we have our styling in Fenix.

In Fenix our styles.xml attributes point to colors in colors.xml, but this would require us to create `mozac_browser_menu_background` in colors.xml which points to an attribute in styles.xml.

Amejia481:
@sblatz  You can do as follow:

**fenix/styles.xml**

```xml
 <style name="Mozac.Browser.Menu" parent="" tools:ignore="UnusedResources">
      <item name="cardBackgroundColor">YOUR_COLOR</item>
</style>
```
The only caveat is if `Mozac.Browser.Menu` is updated in ac, you're not going to see the changes, you will have to manually update it.

components/browser/menu/src/main/res/layout/mozac_browser_menu.xml
```diff
@@ -4,6 +4,8 @@
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
 <android.support.v7.widget.CardView xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
+    style="@style/Mozac.Browser.Menu"
+    android:id="@+id/mozac_browser_menu_cardView"
```
pocmo: nit: The Id doesn't seem to be needed?

components/browser/menu/src/main/res/values/colors.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<resources xmlns:tools="http://schemas.android.com/tools" >
```
pocmo: nit: the "tools" namespace doesn't seem to be used here.

components/browser/menu/src/main/res/values/colors.xml
```diff
@@ -0,0 +1,7 @@
...
+    <color name="mozac_browser_menu_background"></color>
```
pocmo: Oh, empty just works? Let's add a comment why we leave it empty though. :)

pocmo:
> The only caveat is if `Mozac.Browser.Menu` is updated in ac, you're not going to see the changes, you will have to manually update it.

I tried to override the color in an app project with an attr and then having that attr in the theme sue a specific color. But that didn't work unfortunately. That would have been nice.

Amejia481:
@pocmo 
> I tried to override the color in an app project with an attr (mozac_browser_menu_background -> my_attr) and then having that attr in the theme use a specific color (my_attr -> my_color). But that didn't work unfortunately (runtime crash). That would have been nice.

It didn't work because they are not under same theme. We could consider the idea of having a base theme in ac (Like component theme), this way we could define attribute there give them default values and let the app overridden via theme like `colorPrimary` and `colorPrimaryDark`.

```xml
<resources>
    <attr name="mozac_browser_menu_background" format="reference|color" />

    <style name="Theme.Mozac.Light.NoActionBar" parent="Theme.AppCompat.Light.NoActionBar">
        <item name="mozac_browser_menu_background">#A87687</item>
    </style>

    <style name="Theme.Mozac.Light" parent="Theme.AppCompat.Light">
        <item name="mozac_browser_menu_background">#A87687</item>
    </style>

    <style name="Theme.Mozac.NoActionBar" parent="Theme.AppCompat.NoActionBar">
        <item name="mozac_browser_menu_background">#A87687</item>
    </style>
</resources>
```

Amejia481:
@pocmo are we keeping the color solution ?

pocmo:
Let's land to unblock and then we can discuss on #2172 a better solution.

PR #2143, opened 2019-02-21
glean: 1527312: Reinstantiate dynamic labels from disk

mdboom:
This is a work-in-progress that I thought made sense to file now to get some brainstorming help and feedback on.

This makes sure that the set of "already seen" dynamic labels includes labels that have been serialized to disk.  Failure to do so would result in meaningless/unusable stats on labels, since the `__other__` label might include values that also have their own label.

I like this approach because it directly uses the existing code to reinstantiate a storage engine, so it should automatically grow to support other things that we'll need to serialize/deserialize (such as ping-lifetime metrics to survive the app being killed by the os).

Known downsides of the current implementation:

- This loading happens lazily on the first use of a dynamically labeled metric during the current execution.  This means that I/O might happen on the main thread.  Unfortunately, this isn't easily resolved, since the `operator get()` function needs to be synchronous, and that's where we do the selection of the specific label we are going to use.  Possible solutions are (1) to handle this during startup (tricky because there's no reference from the storage engine back to the labeled metric object, and it would change the current behavior that storage engines are reinstantiated lazily). Or (2) handle the mapping of the given label to the actual label (where the change to `__other__` might occur) inside the storage engine rather than the metric type object, so it could all be done off the main thread.  This would require a much larger refactor and a deeper understanding of labels in the storage engine than we currently have so this mapping could be done later.

- There's a `O(n)` process for each labeled metric to go through all of the metric values in the storage engine to find its own relevant labels.  Maybe this isn't too bad, but given that the storage engines aren't organized by labeled metric objects, it's suboptimal.  Again, I don't know if we want to refactor the storage engines so that "understand" labeled metrics at a deeper level.

- The set of "seen" labels is tracked per labeled metric object, not per-labeled-metric-object-per-store as they probably should be.  I'm actually not sure what the right behavior is there, tbh.  It's always the same thing if the timings of the stores is same, but if it's in multiple pings with different lifecycles, all bets are off.

### Pull Request checklist

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/LabeledMetricType.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/LabeledMetricType.kt
@@ -4,6 +4,8 @@
 
 package mozilla.components.service.glean
 
+import mozilla.components.service.glean.storages.StorageEngine
+import mozilla.components.service.glean.storages.StorageEngineManager
 import mozilla.components.support.base.log.logger.Logger
 
 /**
@@ -67,6 +69,20 @@ data class LabeledMetricType<T>(
      * @return adjusted label, possibly set to [OTHER_LABEL]
      */
     private fun getFinalDynamicLabel(label: String): String {
+        if (lifetime != Lifetime.Application && seenLabels.size == 0) {
+            // TODO 1530733: This might cause I/O on the main thread if this is the
+            // first thing being stored to the given storage engine after app restart.
+            getStorageEngineForMetric()?.let {
+                val identifier = (subMetric as CommonMetricData).identifier
+                val prefix = "$identifier/"
+                seenLabels.addAll(
+                    it.getIdentifiersInStores((subMetric as CommonMetricData).sendInPings)
+                        .filter { it.startsWith(prefix) }
+                        .map { it.substring(prefix.length) }
+                )
+            }
+        }
+
         if (!seenLabels.contains(label)) {
             if (seenLabels.size >= MAX_LABELS) {
                 return OTHER_LABEL
@@ -96,6 +112,7 @@ data class LabeledMetricType<T>(
         return when (subMetric) {
             is BooleanMetricType -> subMetric.copy(name = newName) as T
             is CounterMetricType -> subMetric.copy(name = newName) as T
+            is DatetimeMetricType -> subMetric.copy(name = newName) as T
             is StringListMetricType -> subMetric.copy(name = newName) as T
             is StringMetricType -> subMetric.copy(name = newName) as T
             is TimespanMetricType -> subMetric.copy(name = newName) as T
@@ -104,6 +121,14 @@ data class LabeledMetricType<T>(
                 "Can not create a labeled version of this metric type"
             )
         }
+    }
+
+    /**
+     * Delegates to [StorageEngineManager.getStorageEngineForMetric].
+     * Provided here so it can be mocked for testing.
+     */
+    internal fun getStorageEngineForMetric(): StorageEngine? {
+        return StorageEngineManager.getStorageEngineForMetric(subMetric)
     }
 
     /**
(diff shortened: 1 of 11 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -177,6 +185,35 @@ internal abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
...
+     *
+     * @param metricData The labeled metric to find labels for.
+     * @return The set of labels currently in use.
+     */
+    @Suppress("NestedBlockDepth")
+    override fun getLabelsForMetric(metricData: CommonMetricData): Set<String> {
+        // Make sure data with "user" lifetime is loaded before getting the snapshot.
+        ensureUserLifetimeLoaded()
+
+        val prefix = "${metricData.identifier}/"
```
Dexterp37: Instead of doing this and creating a special case, is there any way we could use `extraSerializationData` for this purpose? If it's too tricky, then I'm fine with just moving ahead with the current approach.

mdboom: I'm not seeing a way, but I might be missing something.  Since labeled metrics don't have their own storage engine, we can't override `extraSerializationData`.

mdboom:
Random question: Do we care so much about conserving memory, or conserving bandwidth?  If we only care about conserving bandwidth, we could just deal with `__other__` at ping creation time.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
Dexterp37: Oh, good point. You know what? let's move forward with this and maybe file a bug to see if there's a better way (e.g. removing labelled info from leaking into the generic scalar storage engine)

components/service/glean/src/main/java/mozilla/components/service/glean/LabeledMetricType.kt
```diff
@@ -106,6 +119,24 @@ data class LabeledMetricType<T>(
...
     }
 
+    /**
+     * Get the storage engine associated with a given metric type.
+     *
+     * @return A storage engine
+     */
+    internal fun getStorageEngineForMetric(): StorageEngine? {
+        // Every metric that supports labels needs an entry here
+        return when (subMetric) {
```
Dexterp37: nit: What if we had this kind of functionality as part of the `StorageEngineManager`? It would be a bit less weird as the manager should know about the engines. What do you think?

Dexterp37:
> 
> 
> Random question: Do we care so much about conserving memory, or conserving bandwidth? If we only care about conserving bandwidth, we could just deal with `__other__` at ping creation time.

Ideally, right now we care about both. We don't which one we'll care the most until we'll have... telemetry for that :°(

components/service/glean/src/main/java/mozilla/components/service/glean/LabeledMetricType.kt
mdboom: That makes sense.  Easy enough to move it.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
mdboom: I thought of a way to make this feel less "special-casey" and not have understanding of labeled metrics in `GenericScalarStorageEngine`.  If `GenericScalarStorageEngine` just provided an iterator over all of the "names" in the store, `LabeledMetricType` could filter on the ones it cares about...

Dexterp37: Oh! That sounds smart, I like it! Do you prefer landing this as it is and working on that as a follow-up or just deal with it here?

mdboom: I'll probably just deal with that here and ping you when it's ready for another look.

mdboom:
I think I've made this better by keeping the logic about how labeled metrics actually work in `LabeledMetricType`.

The issue that this might cause I/O on the main thread has been moved out to a [new bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1530733) with a comment here.

I think this is ready for final review.

PR #2152, opened 2019-02-21
Closes #1018: Allow reading GV default user agent via engine settings

### LINKED ISSUE #1018: Expose user agent in browser-engine-gecko-* components
There will be an API to get the user agent soon:
https://bugzilla.mozilla.org/show_bug.cgi?id=1454444

We won't be able to set it - but maybe we want to make it readable from the engine's settings?

csadilek:
We already had support to get/set the user agent string override but so far we couldn't return a default user agent string when none was provided (for system engine we already can). We now have GV's default ua-string accessible via GeckoSession, so bringing this last piece in with this PR.


csadilek:
Coverage is fine here (not affected.)

PR #2155, opened 2019-02-21
Closes #2150: Add engine API to install WebExtensions

### LINKED ISSUE #2150: Add engine API to install WebExtensions
We're getting this from GeckoView soon: https://phabricator.services.mozilla.com/D16268

We need to provide an API in our engine abstraction.

csadilek:
This is a draft showing how to install web extensions using our engine. It's based and dependent on https://bugzilla.mozilla.org/show_bug.cgi?id=1518841 / https://phabricator.services.mozilla.com/D16268 landing.

Apps should be able to install extensions directly and we will also have android components wrapping extensions (for ReaderView, AdBlocking), as well as a feature-webextensions module to manage extensions. So, this is just a first step.


components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
```diff
@@ -59,6 +60,19 @@ class GeckoEngine(
         executor.speculativeConnect(url)
     }
 
+    /**
+     * See [Engine.installWebExtension].
+     */
+    override fun installWebExtension(ext: WebExtension) {
+        fakeRuntime.registerWebExtension(org.mozilla.geckoview.WebExtension(ext.id, ext.url))
```
pocmo: If you want to avoid the name confusion here then you can import the type with a different name:

```
import org.mozilla.geckoview.WebExtension as GeckoWebExtension
```

https://kotlinlang.org/docs/reference/packages.html#imports

samples/browser/src/main/assets/extensions/helloworld/helloworld.js
```diff
@@ -0,0 +1 @@
+console.log("Hello World!");
```
pocmo: If we want something more visible then we could do something like `borderify.js` from the WebExtension samples:
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Your_first_WebExtension#borderify.js

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
csadilek: OK, cleaned this up!

samples/browser/src/main/assets/extensions/helloworld/helloworld.js
csadilek: Add a border to the helloworld extension for now! :)

csadilek:
This works now and is ready for review! 🎉

PR #2173, opened 2019-02-22
[glean] Log events to console when they are recorded

travis79:
### Pull Request checklist

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/EventMetricType.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/EventMetricType.kt
@@ -114,6 +114,11 @@ data class EventMetricType(
                 extra = truncatedExtraKeys
             )
         }
+
+        // Log event to console
+        if (Glean.configuration.logPings) {
+            logger.info("Glean event recorded | $category.$name | $truncatedExtraKeys")
+        }
     }
 
     /**


components/service/glean/src/main/java/mozilla/components/service/glean/EventMetricType.kt
```diff
@@ -114,6 +114,11 @@ data class EventMetricType(
                 extra = truncatedExtraKeys
             )
         }
+
+        // Log event to console
+        if (Glean.configuration.logPings) {
+            logger.info("Glean event recorded | $category.$name | $truncatedExtraKeys")
```
mdboom: We should probably log everything (including the object and value).  But I think object is going away as part of https://bugzilla.mozilla.org/show_bug.cgi?id=1529511 anyway, so maybe we hold off until that's done?

travis79: That was my understanding as well, that `objectId` and `value` were both going away.

Dexterp37: Moreover, we should probably hang this off a separate option (maybe add a runtime `adb` switch for that?): we might want to turn this on independently of pings, but let's ask @pocmo about that, since it was his feature request :)

travis79:
I closed this because it was not that much work invested and it was outdated quite a bit.  I figured we can take this up again when time allows, and with a fresher branch

PR #2177, opened 2019-02-22
Issue #1328: Support redirect flags in Gecko fetch client

csadilek:
Another draft PR to enable redirect flags as GV API becomes available: https://bugzilla.mozilla.org/show_bug.cgi?id=1526327 and https://phabricator.services.mozilla.com/D19523 (landed and got backed out since).


csadilek:
This is ready for review now! I've included the upgrade to the latest GV nightly which required some other fetch related changes (body is now an inputstream, not a bytebuffer which simplified our code too).

PR #2181, opened 2019-02-25
[glean] Implement the new events API for glean

Dexterp37:
The current events API requires parameters that are not used so often. The new API allows to
record events with no parameters and optionally allows to use the `extra` key-value map for all
the other use cases.

### Pull Request checklist


components/service/glean/src/main/java/mozilla/components/service/glean/EventMetricType.kt
```diff
@@ -103,14 +82,17 @@ data class EventMetricType(
             eventKeys
         }
 
+        // We capture the event time now, since we don't know when the async code below
+        // might get executed.
+        val monotonicElapsed = SystemClock.elapsedRealtime()
```
mdboom: :+1:

Issue #1940, opened 2019-02-04
Observe session based on a provided session ID


jonalmeida:
We tend to use the same bit of code for observing the correct session in most of our features to support Custom Tabs.

We should add this to `SelectionAwareSessionObserver` as an extension function or something similar that abstracts out this logic:
```kotlin
val session = sessionId?.let { sessionManager.findSessionById(sessionId) }
session?.let { observeFixed(it) } ?: observeSelected()
```

RomanWuattier:
Hey, I would enjoy working on this issue. Can I?

jonalmeida:
Go for it! :)

PR #2184, opened 2019-02-25
Observe session based on the session ID

RomanWuattier:
Create the function `observeFromId` to start observing changes based on
the given `sessionId`.

Closes: #1940

### Changelog
* Create the function `observeFromId` to abstract the following logic:
```
val session = sessionId?.let { sessionManager.findSessionById(sessionId) }
session?.let { observeFixed(it) } ?: observeSelected()
```
with
```
observeFromId(sessionId)
```
* Update impacted feature' and presenter' unit tests. 
### Accessibility
Does not include any user-facing features.

### Pull Request checklist

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/SelectionAwareSessionObserver.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/SelectionAwareSessionObserver.kt
@@ -45,6 +45,17 @@ abstract class SelectionAwareSessionObserver(
     }
 
     /**
+     * Starts observing changes to the session matching the [sessionId]. If
+     * the session does not exist, then observe the selected session.
+     *
+     * @param sessionId the session ID to observe.
+     */
+    fun observeIdOrSelected(sessionId: String?) {
+        val session = sessionId?.let { sessionManager.findSessionById(sessionId) }
+        session?.let { observeFixed(it) } ?: observeSelected()
+    }
+
+    /**
      * Stops the observer.
      */
     @CallSuper
--- a/components/browser/session/src/test/java/mozilla/components/browser/session/SelectionAwareSessionObserverTest.kt
+++ b/components/browser/session/src/test/java/mozilla/components/browser/session/SelectionAwareSessionObserverTest.kt
@@ -17,6 +17,7 @@ class SelectionAwareSessionObserverTest {
     private lateinit var sessionManager: SessionManager
     private lateinit var session1: Session
     private lateinit var session2: Session
+    private lateinit var session3: Session
     private lateinit var observer: MockObserver
 
     @Before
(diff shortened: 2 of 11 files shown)


components/browser/session/src/main/java/mozilla/components/browser/session/SelectionAwareSessionObserver.kt
```diff
@@ -44,6 +44,19 @@ abstract class SelectionAwareSessionObserver(
         activeSession?.register(this)
     }
 
+    /**
+     * Starts observing changes to the session matching the [sessionId].
+     *
+     * @param sessionId the session ID to observe.
+     */
+    fun observeFromId(sessionId: String?) {
+        sessionId?.let {
```
jonalmeida: This is definitely correct, but do you mind going back to the old version of the code? I find it reads better.

components/browser/session/src/main/java/mozilla/components/browser/session/SelectionAwareSessionObserver.kt
```diff
@@ -44,6 +44,19 @@ abstract class SelectionAwareSessionObserver(
         activeSession?.register(this)
     }
 
+    /**
+     * Starts observing changes to the session matching the [sessionId].
+     *
+     * @param sessionId the session ID to observe.
+     */
+    fun observeFromId(sessionId: String?) {
```
jonalmeida: Since this method does a lot more than expected, i.e. "if the session with an ID doesn't exist, observe the selected session.", we could name it `observeIdOrSelected`?

`observeFromId` signifies, it will _only_ observe the session with the ID.

components/browser/session/src/main/java/mozilla/components/browser/session/SelectionAwareSessionObserver.kt
```diff
@@ -44,6 +44,19 @@ abstract class SelectionAwareSessionObserver(
         activeSession?.register(this)
     }
 
+    /**
+     * Starts observing changes to the session matching the [sessionId].
```
jonalmeida: Lets add some more detailed docs since it does more than just observing changes to the selected session.

components/browser/session/src/main/java/mozilla/components/browser/session/SelectionAwareSessionObserver.kt
jonalmeida: Let's add this to the changelog! 🎉 

It's in `docs/changelog.md`

jonalmeida:
Some tests would be nice as well. :)

components/browser/session/src/main/java/mozilla/components/browser/session/SelectionAwareSessionObserver.kt
RomanWuattier: Agreed. `observeIdOrSelected` is better :)

components/browser/session/src/main/java/mozilla/components/browser/session/SelectionAwareSessionObserver.kt
RomanWuattier: Done

RomanWuattier:
Thanks for the review! I updated the PR and unit tests the function `observeIdOrSelected`.

jonalmeida:
🎉

PR #2185, opened 2019-02-25
[glean] 1525600: Round up timespan values

mdboom:
This rounds up all timespan values (rather than truncating), so e.g. 400ms is reported as 1s if the `time_unit` is `second`.  This should prevent the high number of 0 values.

### Pull Request checklist


mdboom:
Maybe so, but I'm not sure it matters as much as never returning 0.  But if ceil() is strictly what we want, we'll need to write the conversions ourselves and not use the Android API.

Dexterp37:
I have mixed feelings about this: this is probably better than the current state, for smaller time units. However, I feel like, if it comes to `Days`, having a potentially misleading +1 could be a problem :\ Maybe we should ask Frank too :)

mdboom:
Remember, we're not rounding up the base value (which remains in nanoseconds).  So if you timed something for say, 3 hours on one day, and 3 hours the next day, you'd still round to "1 day", not "2 days".  I think short of using floating point for this (which would negate the purpose of the rounding) this is probably the best we can do.

Dexterp37:
@mdboom should we document this somewhere? The metric docs?

mdboom:
Good point about documentation.  Probably should document in the SDK docs (I think there's a section on TimeUnit there), as well as the API docs here.

mdboom:
I've suggested changes to the SDK docs (seems to be closed for editing).

@fbertsch: It looks like I can't add you as a reviewer, but I'd appreciate your feedback.

fbertsch:
From a dashboarding perspective, I think we want to truncate. My rationale being that a bar chart will probably have bars at e.g. `0, 1, 2, ...`. The 0 bar will stretch from 0 -> 1, the 1 bar from 1 -> 2. It would be confusing to see that bar stretch from 1 -> 2 but represent values from 0 -> 1.

Thoughts?

fbertsch:
If there are an unusually high number of 0s, perhaps we should change the time unit? Additionally, we could record a boolean that counts the number of 0 instances?

mdboom:
I feel a bit out of my depth here about what analysts really would want.  It's all a reaction to @chutten's data analysis, which ultimately ended up in [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1525600) -- basically that `0` for a timespan was surprising (because it represents something between 0 and 1 in that case).  I'm totally happy to merge or close this depending on the consensus.

(`null` timespans shouldn't be occurring, and I'm sort of treating that as a separate issue and investigation).

fbertsch:
My only concern is the one @travis79 brought up. Why might a client report an actual 0 for a timespan? Is that possible with the current API? If it is we may want to separate that case out, or include exact values of the value is an integer already.

mdboom:
If the 0 is truly "no time at all", then the metric simply won't be recorded and reported.  The only exception to that might be if the timing is truly less than one nanosecond.

fbertsch:
If that is impossible with the current API then I would stick with truncation. The engineer should be able to test out the probe and see if the results look reasonable; e.g. is it getting too many 0s? If so they can change the unit. Does that seem reasonable?

mdboom:
Sure -- that seems reasonable.  Perhaps it's an issue of documentation and education around 0 meaning '0 - 1'.

One counterargument the just occurred to me, however: For analysis involving `n / time`, if we allow zeros, analysts will have to care about the possibility of divide-by-zero errors.  If we consider 1 the minimum value (where there's any data at all), that's a little simpler...

Dexterp37:
> 
> 
> If that is impossible with the current API then I would stick with truncation. The engineer should be able to test out the probe and see if the results look reasonable; e.g. is it getting too many 0s? If so they can change the unit. Does that seem reasonable?

Looks like a reasonable approach to me.

mdboom:
Ok.  It sounds like the action plan is to close this bug but document the surprising behavior.  Is that correct?

Dexterp37:
> 
> 
> Ok. It sounds like the action plan is to close this bug but document the surprising behavior. Is that correct?

Yes !

mdboom:
Ok.  I've documented the current behavior in the SDK doc -- specifically how `0` values work.

Therefore, closing this.

Dexterp37:
> 
> 
> Ok. I've documented the current behavior in the SDK doc -- specifically how `0` values work.

@mdboom should we update this in the `metrics.yaml` schema file in [here](https://github.com/mozilla/glean_parser/blob/master/glean_parser/schemas/metrics.1-0-0.schema.yaml)? So that it gets in the automatically generated metrics docs?

mdboom:
Yes, good point.

PR #2187, opened 2019-02-25
Closes #2186: Adds private mode to IntentProcessor

### LINKED ISSUE #2186: Update IntentProcessor to respect private mode
See [here](https://github.com/mozilla-mobile/fenix/issues/664#issuecomment-467169680) for more information.

Basically the IntentProcessor needs to check if the selected session isPrivate before creating a tab.

sblatz:
### Pull Request checklist

I'm assuming I don't need a test for creating a flag, but please let me know if I do.

I chose to make this a public variable as, in Fenix, we want to set the isPrivate flag before processing without creating a whole new IntentProcessor. Perhaps there's a cleaner solution that allows this variable to set private, though? I also wanted to avoid using a custom processor since we wanted all other functionality to be the same.

### DIFF
--- a/components/feature/intent/src/main/java/mozilla/components/feature/intent/IntentProcessor.kt
+++ b/components/feature/intent/src/main/java/mozilla/components/feature/intent/IntentProcessor.kt
@@ -28,6 +28,7 @@ typealias IntentHandler = (Intent) -> Boolean
  * @property useDefaultHandlers Whether or not the built-in handlers should be used.
  * @property openNewTab Whether a processed intent should open a new tab or
  * open URLs in the currently selected tab.
+ * @property isPrivate Whether a processed intent should open a new tab as private
  */
 class IntentProcessor(
     private val sessionUseCases: SessionUseCases,
@@ -35,7 +36,8 @@ class IntentProcessor(
     private val searchUseCases: SearchUseCases,
     private val context: Context,
     private val useDefaultHandlers: Boolean = true,
-    private val openNewTab: Boolean = true
+    private val openNewTab: Boolean = true,
+    private val isPrivate: Boolean = false
 ) {
     private val defaultActionViewHandler = { intent: Intent ->
         val safeIntent = SafeIntent(intent)
@@ -56,7 +58,7 @@ class IntentProcessor(
             }
 
             else -> {
-                val session = createSession(url, source = Source.ACTION_VIEW)
+                val session = createSession(url, private = isPrivate, source = Source.ACTION_VIEW)
                 sessionUseCases.loadUrl.invoke(url, session)
                 true
             }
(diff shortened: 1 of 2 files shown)


components/feature/intent/src/main/java/mozilla/components/feature/intent/IntentProcessor.kt
```diff
@@ -28,14 +28,16 @@ typealias IntentHandler = (Intent) -> Boolean
...
  */
 class IntentProcessor(
     private val sessionUseCases: SessionUseCases,
     private val sessionManager: SessionManager,
     private val searchUseCases: SearchUseCases,
     private val context: Context,
     private val useDefaultHandlers: Boolean = true,
-    private val openNewTab: Boolean = true
+    private val openNewTab: Boolean = true,
+    var isPrivate: Boolean = false
```
jonalmeida: Apps usually have one IntentProcessor to handle all incoming intents, so if this becomes a contructor argument, you would need an IntentProcessor instantiated for private and non-private sessions.

This might be a case where we want to allow the app to delegate whether to make this a private session:
```kotlin
class IntentProcessor(
  /* other deps */
  private val privateSession: () -> Boolean = { false }
) {}
```

This would allow the app to decide (maybe based on a pref?) when a session should be private/regular. At the calling site you'd have this:

```kotlin
val session = createSession(url, private = privateSession.invoke(), source = Source.ACTION_VIEW)
```

I think we can do better than this as well, but I'm not quite sure what.. 🤔

pocmo: This is currently the case but actually there's no reason for the `IntentProcessor` to be a global singleton. It could just live in the scope of the activity that handles the incoming Intents?

pocmo: Does this need to be `var` though? I assume this can be a `val` (especially if we do not treat the `IntentProcessor` as something global`.

components/feature/intent/src/main/java/mozilla/components/feature/intent/IntentProcessor.kt
```diff
@@ -73,7 +75,7 @@ class IntentProcessor(
             else -> {
                 val url = extraText.split(" ").find { it.isUrl() }
                 if (url != null) {
-                    val session = createSession(url, source = Source.ACTION_SEND)
+                    val session = createSession(url, private = isPrivate, source = Source.ACTION_SEND)
                     sessionUseCases.loadUrl.invoke(url, session)
```
pocmo: Instead of adding `openNewTab` and `isPrivate` parameters we probably should just replace this with a `LoadUrlUseCase`. This allows an app to specify `AddNewTabUseCase`, `AddNewPrivateTabUseCase` or `DefaultLoadUrlUseCase`.

sblatz: Are you suggesting I make a loadUrlCase a parameter (to replace `openNewTab` and `isPrivate`) that is passed in with the constructor that is then used to open sessions with the createSession method?

pocmo: Would that work? We'd need to pass the `source` along though.

PR #2188, opened 2019-02-25
New component: concept-push

jonalmeida:
A concept component that provides a minimal base. More to follow..

### Pull Request checklist


components/concept/push/src/main/java/mozilla/components/concept/push/Push.kt
```diff
@@ -0,0 +1,45 @@
...
+class Push(
+    internal val service: PushService
+) {
+    /**
+     * Start the Push component.
+     */
+    fun init(): Push {
+        Push.instance = this
+
+        service.start()
```
Amejia481: Should also `Push` have a function to be stopped ? Because `service` has a stop function, we're starting it but never stopping it. :)

components/concept/push/src/main/java/mozilla/components/concept/push/Push.kt
```diff
@@ -0,0 +1,45 @@
...
+
+        return this
+    }
+
+    internal fun onNewToken(newToken: String) {}
```
Amejia481: nit: We could refer to the issues number that this work is related, and the functions could  be equal to `Unit`until we provide an implementation.

components/concept/push/src/main/java/mozilla/components/concept/push/Push.kt
```diff
@@ -0,0 +1,45 @@
...
+ *  file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package mozilla.components.concept.push
+
+/**
+ * The Push component that will handle push messages from the [PushService] provided.
+ * Starting Push in the Application's onCreate is recommended.
+ */
+class Push(
```
pocmo: Since this class seems to contain logic (or will in the future) it doesn't really feel right to have this in a `concept` component. This may be something for a `feature` component.

jonalmeida:
@csadilek could you take a peek at this again when you get a chance? I remember you had some comments about the tense in the documentation, which I think I've addressed.

components/concept/push/src/main/java/mozilla/components/concept/push/Bus.kt
```diff
@@ -0,0 +1,51 @@
...
+ *          }
+ *     }
+ *
+ *     bus.register(T.A, MyObservableClass())
+ *
+ *     // In some other class, we can notify the MyObservableClass.
+ *     bus.notifyObservers(Type.A, Message("Hello!"))
+ * </code>
+ */
+interface Bus<T, M> {
```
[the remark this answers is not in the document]
jonalmeida: descrived them both!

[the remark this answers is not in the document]
csadilek: ✏️ 👓

PR #2199, opened 2019-02-26
Closes #2121: Adding db for saving site permissions.

### LINKED ISSUE #2121: Add database to remember site permissions granted
sitepermissions database:
- save user choices (site - permissions)
- query sites
- query site
- modify site permissions
- remove site permissions

Amejia481:
@pocmo before adding some tests I want your input about db and especially the `SitePermissionsStorage` api, I'm not quite sure if we should handle the threading or if  the api should be blocking and the app handle itself.


components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsStorage.kt
```diff
@@ -0,0 +1,75 @@
...
+                    sitePermissions.toSitePermissionsEntity()
+                )
+        }
+    }
+
+    fun update(sitePermissions: SitePermissions) {
+        save(sitePermissions)
+    }
+
+    fun findSitePermissionsBy(origin: String): Deferred<SitePermissions?> {
```
jonalmeida: Can I make a request for `findOriginsBySitePermission`? 😃 

I like Android's list system of permissions. You can see a single app's permissions or all the apps that have a permission. Maybe this can be something for the future, if not now.

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissions.kt
```diff
@@ -0,0 +1,20 @@
...
+    val origin: String,
+    val location: Boolean = false,
+    val notification: Boolean = false,
+    val microphone: Boolean = false,
+    val cameraBack: Boolean = false,
+    val cameraFront: Boolean = false,
+    val bluetooth: Boolean = false,
+    val localStorage: Boolean = false,
+    val savedAt: Long
+)
```
pocmo: I don't think a `Boolean` is enough here; since we need to model "Allowed", "Blocked", "No decision". For example with `microphone = false` it's unclear whether the user has decided to block this or whether the site has never requested this.

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsStorage.kt
```diff
@@ -0,0 +1,75 @@
...
+    fun all(limit: Int = 20): Deferred<List<SitePermissions>> {
+        return GlobalScope.async(Dispatchers.IO) {
+            database
+                .sitePermissionsDao()
+                .getSitePermissions(limit)
+                .map {
+                    it.toSitePermission()
+                }
+        }
+    }
```
pocmo: I think I would make all those methods blocking and let the calling app decide if it wants to use coroutines or rxjava or whatever. This also allows an app  using coroutines to define the scope too. For example here the call to `all()` should probably be in the scope of the activity/fragment and not in the global one.

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/db/SitePermissionEntity.kt
```diff
@@ -0,0 +1,74 @@
...
+    var cameraBack: Boolean,
+
+    @ColumnInfo(name = "camera_front")
+    var cameraFront: Boolean,
+
+    @ColumnInfo(name = "bluetooth")
+    var bluetooth: Boolean,
+
+    @ColumnInfo(name = "local_storage")
+    var localStorage: Boolean,
```
pocmo: Like above: I don't think `Boolean`s are good enough here. Maybe an `enum` instead? (You'll need a type converter for Room in that case, I guess).

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsStorage.kt
Amejia481: To do something similar to this but with sites permissions?
screenshot_20190226-183814


screenshot_20190226-183914

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissions.kt
Amejia481: I didn't know that we will store **"No decision"** as a state.

I thought that we only will store the site permissions when they get **granted** or **denied**, and we will assume that if the `origin` is not in the db is because it hasn't be asked before (**"No decision"**). As this state only happens one time. Am I wrong?

Is this approach enough? Or should we still need to store  **"No decision"** as a state.
 Could you help me to clarify the use case for "**No decision"** and when we save it to the db?

This is the flow that I initially thought:

### Case 1:
A user enters for first time to a site that requires a notification permission:

We check if the `origin` is in the db using `store.findSitePermissionsBy(origin: String)`,
as it's the first time for the user, the `origin` is not in the db. Then we show the prompt.

 1. If the user hit the **"Allow"** button, then we save the origin using `store.save(sitePermissions: SitePermissions)`  with the `sitePermissions.notification` marked as  **"Allowed"** .

`Otherwise`

2.  If the user hit the **"Don't Allow"** button, then we save the origin using `store.save(sitePermissions: SitePermissions)`  with the `sitePermissions.notification` marked as  **"Blocked"** .


### Case 2:
The same user enters to the same site, but this time the `origin` is in the db with the `sitePermissions.notification` marked as  **"Allowed"**. 

1. We check if the `origin` is in the db using `store.findSitePermissionsBy(origin: String)`,
and **it's there** with the `sitePermissions.notification` marked as  **"Allowed"**. Then  we  don't show the prompt and grant the permission.

### Case 3:
The same user enters to the same site the `origin` is in the db with the `sitePermissions.notification` marked as  **"Blocked"**. 

1. We check if the `origin` is in the db using `store.findSitePermissionsBy(origin: String)`,
and **it's there** with the `sitePermissions.notification` marked as  **"Blocked"**. Then  we do show the prompt and let the user chooses.

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsStorage.kt
Amejia481: 👍

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsStorage.kt
Amejia481: Looks like that this is what chrome is doing right now, you can mange the permissions by site or by type
screenshot_20190226-201441

By specify permission.
screenshot_20190226-201519

Amejia481: @jonalmeida  I will try to include it 👍  in this version

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissions.kt
Amejia481: @jonalmeida helped me to clarify why  **"No decision"** is need it :)

Amejia481:
Ready for review!

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsStorage.kt
```diff
@@ -0,0 +1,124 @@
...
+            .sitePermissionsDao()
+            .getSitePermissionsBy(origin)
+            ?.toSitePermission()
+    }
+
+    /**
+     * Finds all SitePermissions grouped by [Permission].
+     * @return a map of site grouped by [Permission].
+     */
+    fun findAllSitePermissionsGroupedByPermission(): MutableMap<Permission, SitePermissions> {
```
Amejia481: @jonalmeida I changed the name from your original suggestion

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/db/SitePermissionsDatabase.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    @TypeConverter
+    fun toInt(status: SitePermissions.Status): Int {
+        return statusArray.indexOf(status)
+    }
+
+    @TypeConverter
+    fun toStatus(index: Int): SitePermissions.Status {
+        return statusArray[index]
+    }
```
pocmo: This is dangerous since the enum value order is something that is not static, can change and is pretty much not in your control. Someone changing the enum order wouldn't expect other parts of the app to blow up. :)

Since enums can have values. What do you think about defining the int value inside the enum.

Something like:

```kotlin
    enum class Status(
        internal val value: Int
    ) {
        ALLOWED(1),
        BLOCKED(-1),
       NO_DECISION(0)
    }
```

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsStorage.kt
```diff
@@ -0,0 +1,130 @@
...
+            )
+    }
+
+    /**
+     * Replaces an existing SitePermissions with the values of [sitePermissions] provided as a parameter.
+     * @param sitePermissions the sitePermissions to be updated.
+     */
+    fun update(sitePermissions: SitePermissions) {
+        save(sitePermissions)
+    }
```
pocmo: Do we really need `update()` if it only calls `save()`?

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/db/SitePermissionsDao.kt
```diff
@@ -0,0 +1,30 @@
...
+import android.arch.persistence.room.Query
+
+/**
+ * Internal dao for accessing and modifying sitePermissions in the database.
+ */
+@Dao
+internal interface SitePermissionsDao {
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    fun insertOrUpdateSitePermission(entity: SitePermissionsEntity): Long
```
pocmo: This is convenient but can also be the source of bugs if you start "replacing" values when you actually expect to "insert" (e.g. Insert "Allow microphone" replaces "Allow camera" in database, instead of updating the existing entry).

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsStorage.kt
```diff
@@ -0,0 +1,130 @@
...
+            .sitePermissionsDao()
+            .getSitePermissionsBy(origin)
+            ?.toSitePermission()
+    }
+
+    /**
+     * Finds all SitePermissions grouped by [Permission].
+     * @return a map of site grouped by [Permission].
+     */
+    fun findAllSitePermissionsGroupedByPermission(): MutableMap<Permission, MutableList<SitePermissions>> {
```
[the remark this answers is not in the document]
Amejia481: [1] The thing is that as I'm adding element to the list of each key on the map, I need it to be mutable. 

[2] If I change the type to Map<.., List<>>, I will need to iterate over each key on the map, and its permissions list to create a new immutable list(List<>) .

Is it not a bit overkill ?

```kotlin

    fun findAllSitePermissionsGroupedByPermission(): List<Permission>, MutableList<SitePermissions>> {
        val sitePermissions = all()
        val map = mutableMapOf<Permission, MutableList<SitePermissions>>()

        sitePermissions.forEach { permission ->
            with(permission) {
                map.putIfAllowed(BLUETOOTH, bluetooth, permission)
                map.putIfAllowed(MICROPHONE, microphone, permission)
                map.putIfAllowed(CAMERA, cameraFront, permission)
                map.putIfAllowed(CAMERA, cameraBack, permission)
                map.putIfAllowed(LOCAL_STORAGE, localStorage, permission)
                map.putIfAllowed(NOTIFICATION, notification, permission)
                map.putIfAllowed(LOCATION, location, permission)
            }
        }
       // [2]
        val newMap = convertToImmutableList(map)
        return newMap
    }

    private fun MutableMap<Permission, MutableList<SitePermissions>>.putIfAllowed(
        permission: Permission,
        status: Status,
        sitePermissions: SitePermissions
    ) {
        if (status == ALLOWED) {
            if (this.containsKey(permission)) {
//[1]
                this[permission]?.add(sitePermissions)
            } else {
                this[permission] = mutableListOf(sitePermissions)
            }
        }
```
    }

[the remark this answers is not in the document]
pocmo: I don't think you have to convert (or am I missing something)? This just works:

```kotlin
fun foo(): Map<String, List<Int>> {
        val value = mutableMapOf<String, MutableList<Int>>()

        // Do something with the mutable type..

        return value
}
```

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsStorage.kt
```diff
@@ -0,0 +1,130 @@
...
+     * @return all the site permissions in the storage
+     */
+    fun all(limit: Int = 20): List<SitePermissions> {
+        return database
+            .sitePermissionsDao()
+            .getSitePermissions(limit)
+            .map {
+                it.toSitePermission()
+            }
+    }
```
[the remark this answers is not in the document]
Amejia481: I thought some apps will need to do something similar to this, but you are right, let's focus on what is need it for the MVP.

image

screenshot_20190305-085417

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsStorage.kt
```diff
@@ -0,0 +1,130 @@
...
+            .getSitePermissionsBy(origin)
+            ?.toSitePermission()
+    }
+
+    /**
+     * Finds all SitePermissions grouped by [Permission].
+     * @return a map of site grouped by [Permission].
+     */
+    fun findAllSitePermissionsGroupedByPermission(): MutableMap<Permission, MutableList<SitePermissions>> {
+        val sitePermissions = all()
```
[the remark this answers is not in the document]
Amejia481: Thanks for the caught 👍

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsStorage.kt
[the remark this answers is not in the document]
Amejia481: 👍

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/db/SitePermissionsDatabase.kt
Amejia481: 👍

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/db/SitePermissionsDao.kt
Amejia481: 👍

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsStorage.kt
Amejia481: Now we have independent functions for saving and updating in `SitePermissionsDao`.

PR #2200, opened 2019-02-26
[glean] 1529204: Fix compile warnings

mdboom:
### Pull Request checklist

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/BooleanMetricType.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/BooleanMetricType.kt
@@ -46,6 +46,7 @@ data class BooleanMetricType(
             return
         }
 
+        @Suppress("EXPERIMENTAL_API_USAGE")
         ioTask = Dispatchers.API.launch {
             // Delegate storing the boolean to the storage engine.
             BooleansStorageEngine.record(
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
@@ -194,6 +194,7 @@ open class GleanInternalAPI internal constructor () {
             val packageInfo = applicationContext.packageManager.getPackageInfo(
                 applicationContext.packageName, 0
             )
+            @Suppress("DEPRECATION")
             StringsStorageEngine.record(
                 GleanInternalMetrics.appBuild,
                 packageInfo.versionCode.toString()
@@ -209,7 +210,7 @@ open class GleanInternalAPI internal constructor () {
 
         // Set up information and scheduling for glean owned pings
         metricsPingScheduler = MetricsPingScheduler(applicationContext)
-        baselinePing = BaselinePing(applicationContext)
+        baselinePing = BaselinePing()
     }
 
     /**
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/ping/BaselinePing.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/ping/BaselinePing.kt
@@ -4,7 +4,6 @@
 
 package mozilla.components.service.glean.ping
 
-import android.content.Context
 import android.os.Build
 import mozilla.components.service.glean.GleanMetrics.GleanBaseline
 import mozilla.components.support.base.log.logger.Logger
@@ -13,7 +12,7 @@ import java.util.Locale
 /**
  * BaselinePing facilitates setup and initialization of baseline pings.
  */
-internal class BaselinePing(applicationContext: Context) {
+internal class BaselinePing {
     private val logger = Logger("glean/BaselinePing")
 
     companion object {
--- a/components/service/glean/src/test/java/mozilla/components/service/glean/ping/PingMakerTest.kt
+++ b/components/service/glean/src/test/java/mozilla/components/service/glean/ping/PingMakerTest.kt
@@ -5,6 +5,7 @@
 package mozilla.components.service.glean.ping
 
 import android.content.Context
+import androidx.test.core.app.ApplicationProvider
 import mozilla.components.service.glean.BuildConfig
 import mozilla.components.service.glean.storages.MockStorageEngine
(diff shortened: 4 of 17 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/ping/BaselinePing.kt
```diff
@@ -13,7 +13,7 @@ import java.util.Locale
 /**
  * BaselinePing facilitates setup and initialization of baseline pings.
  */
-internal class BaselinePing(applicationContext: Context) {
+internal class BaselinePing(@Suppress("UNUSED_PARAMETER") applicationContext: Context) {
```
Dexterp37: Let's remove `applicationContext` if we are not using it

components/service/glean/src/test/java/mozilla/components/service/glean/ping/PingMakerTest.kt
```diff
@@ -139,6 +139,7 @@ class PingMakerTest {
     fun `seq number must be sequential`() {
         // NOTE: Using a "real" ApplicationContext here so that it will have
         // a working SharedPreferences implementation
+        @Suppress("DEPRECATION")
         var applicationContext = RuntimeEnvironment.application.applicationContext
```
Dexterp37: I think this can become: `var applicationContext = ApplicationProvider.getApplicationContext<Context>()` or simply `var applicationContext = ApplicationProvider.getApplicationContext()` (if it doesn't complain about the casting)

components/service/glean/src/test/java/mozilla/components/service/glean/scheduler/MetricsPingSchedulerTest.kt
```diff
@@ -19,6 +19,7 @@ import java.util.concurrent.TimeUnit
 class MetricsPingSchedulerTest {
     // NOTE: Using a "real" ApplicationContext here so that it will have
     // a working SharedPreferences implementation
+    @Suppress("DEPRECATION")
     private val metricsPingScheduler = MetricsPingScheduler(RuntimeEnvironment.application.applicationContext)
```
Dexterp37: I think we can pass in `ApplicationProvider.getApplicationContext()`

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -226,6 +226,7 @@ open class GleanInternalAPI internal constructor () {
             val packageInfo = applicationContext.packageManager.getPackageInfo(
                 applicationContext.packageName, 0
             )
+            @Suppress("DEPRECATION")
```
Dexterp37: Mh, woha, `versionCode` is being deprecated? I wonder if there's anything we should do here before we start relying on this data.

components/service/glean/src/main/java/mozilla/components/service/glean/ping/BaselinePing.kt
mdboom: Ok.  Easy enough.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
mdboom: Yes -- but unfortunately, its replacement (`getLongVersionCode()`) isn't available in the earliest version we need to support, so it fails in testing.

Dexterp37: Ouch, I see :) Suppressing is fine then :)

components/service/glean/src/test/java/mozilla/components/service/glean/scheduler/MetricsPingSchedulerTest.kt
mdboom: Indeed.  That works.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
travis79: There are ways of doing version dependent code branches, is that something worth looking at doing here?

components/service/glean/src/main/java/mozilla/components/service/glean/BooleanMetricType.kt
```diff
@@ -46,6 +46,7 @@ data class BooleanMetricType(
             return
         }
 
+        @Suppress("EXPERIMENTAL_API_USAGE")
```
travis79: Is there any advantage/disadvantage to go with the IDE's suggestion to use either `@UseExperimental(ObsoleteCoroutinesApi::class)` or `@ObsoleteCoroutinesApi` rather than just suppressing the warning?  This applies to all of the places that this is used.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
mdboom: The whole thing is a bit weird.  It seems like they changed the version identifier from an `int` to a `long`, with the latter having what's called a "major version id" in the upper 32 bits.  We don't appear use or care about that, so we'd be masking those bits out.  All that seemed much more brittle to me than just using the deprecated API (that's sure to do the right thing) as long as it exists.

components/service/glean/src/main/java/mozilla/components/service/glean/BooleanMetricType.kt
mdboom: I started with that.  Unfortunately, those flags are "viral", in that all callers of these methods would also have to declare that.  Meaning, anyone using the Glean API would also have to add those tags.

travis79: That makes sense!

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
travis79: That's perfectly reasonable to me, I doubt they ever do more than deprecate it.

PR #2202, opened 2019-02-26
[glean] 1525333: Provide way to send pings immediately

mdboom:
This adds a "BroadcastReceiver" so that we can send glean messages at runtime using adb.

A couple of things to note:

- This is only enabled in a DEBUG build.  I don't know if that's required, but it seems like good practice to avoid the ability to expose information in a release build.

- I added an intent to turn logging of pings to logCat on.  I know we already have a way to do that, but I think this might be better since it doesn't require restarting the app.  Happy to remove if others disagree, but we probably should have only-one-way-to-do-it on that.

### Pull Request checklist

### DIFF
--- a/components/service/glean/README.md
+++ b/components/service/glean/README.md
@@ -117,12 +117,16 @@ for the command line switches used to pass the extra keys. These are the current
 
 |key|type|description|
 |---|----|-----------|
-| logPings | boolean | if set to `true`, glean dumps pings to logcat; defaults to `false`|
+| logPings | boolean (--ez) | If set to `true`, glean dumps pings to logcat; defaults to `false` |
+| sendPing | string (--es) | Sends the ping with the given name immediately. |
 
-For example, to start the glean sample application with the option to dump pings to logcat turned on,
-the following command can be used:
+For example, to direct the glean sample application to dump pings to logcat, and send the "metrics" ping immediately, the following command can be used:
 
-`adb shell am start -n org.mozilla.samples.glean/mozilla.components.service.glean.debug.GleanDebugActivity --ez logPings true`
+```
+adb shell am start -n org.mozilla.samples.glean/mozilla.components.service.glean.debug.GleanDebugActivity \
+  --ez logPings true \
+  --es sendPing metrics
+```
 
 ## Contact
 
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
@@ -257,8 +257,10 @@ open class GleanInternalAPI internal constructor () {
      * Send a list of pings by name.
      *
      * Both the ping collection and ping uploading happens asyncronously.
+     * If the ping currently contains no content, it will not be sent.
      *
      * @param pingNames List of pings to send.
+     * @return true if any pings were actually sent.
      */
     internal fun sendPings(pingNames: List<String>): Boolean {
         if (!isInitialized()) {
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
@@ -237,44 +237,53 @@ open class GleanInternalAPI internal constructor () {
      * @param pingEvent The type of the event.
      */
     fun handleEvent(pingEvent: Glean.PingEvent) {
-        if (!isInitialized()) {
-            logger.error("Glean must be initialized before handling events.")
-            return
-        }
-
-        if (!uploadEnabled) {
-            logger.error("Glean must be enabled before handling events.")
-            return
-        }
-
-        val availableToSend = when (pingEvent) {
+        when (pingEvent) {
             Glean.PingEvent.Background -> {
-                // Schedule the baseline and event pings and save the result to determine whether or not
-                // we need to schedule the upload worker
-                assembleAndSerializePing(BaselinePing.STORE_NAME)
-                assembleAndSerializePing("events")
-
-                // Always true since baseline is always sent
-                true
+                // Schedule the baseline and event pings
+                sendPings(listOf(BaselinePing.STORE_NAME, "events"))
             }
             Glean.PingEvent.Default -> {
                 // Check the metrics ping schedule to determine whether it's time to schedule a new
                 // metrics ping or not
                 if (metricsPingScheduler.canSendPing() &&
-                    assembleAndSerializePing(MetricsPingScheduler.STORE_NAME)) {
+                    sendPings(listOf(MetricsPingScheduler.STORE_NAME))) {
                     metricsPingScheduler.updateSentTimestamp()
-                    true
-                } else {
-                    false
                 }
             }
+        }
+    }
+
+    /**
+     * Send a list of pings by name.
+     *
+     * Both the ping collection and ping uploading happens asyncronously.
+     *
+     * @param pingNames List of pings to send.
+     */
+    internal fun sendPings(pingNames: List<String>): Boolean {
+        if (!isInitialized()) {
+            logger.error("Glean must be initialized before sending pings.")
+            return false
+        }
+
+        if (!uploadEnabled) {
+            logger.error("Glean must be enabled before sending pings.")
+            return false
+        }
+
+        var hasPingContent = false
+        for (pingName in pingNames) {
+            val hasContent = assembleAndSerializePing(pingName)
+            hasPingContent = hasPingContent || hasContent
         }
 
         // It should only take a single PingUploadWorker to process all queued pings, so we only
         // want to have one scheduled at a time.
-        if (availableToSend) {
+        if (hasPingContent) {
             PingUploadWorker.enqueueWorker()
         }
+
+        return hasPingContent
     }
 
     /**
--- a/components/service/glean/src/test/java/mozilla/components/service/glean/debug/GleanDebugActivityTest.kt
+++ b/components/service/glean/src/test/java/mozilla/components/service/glean/debug/GleanDebugActivityTest.kt
@@ -16,7 +16,15 @@ import org.junit.Before
 import org.robolectric.Robolectric
 import android.content.pm.ActivityInfo
 import android.content.pm.ResolveInfo
+import androidx.work.testing.WorkManagerTestInitHelper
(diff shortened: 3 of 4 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -92,6 +93,10 @@ open class GleanInternalAPI internal constructor () {
         initializeCoreMetrics(applicationContext)
 
         ProcessLifecycleOwner.get().lifecycle.addObserver(gleanLifecycleObserver)
+
+        if (BuildConfig.DEBUG) {
```
Dexterp37: We don't want this to only work in debug builds. We'll need this for QAing final builds as well.

components/service/glean/README.md
```diff
@@ -124,6 +127,33 @@ the following command can be used:
 
 `adb shell am start -n org.mozilla.samples.glean/mozilla.components.service.glean.debug.GleanDebugActivity --ez logPings true`
 
+### Debug intent receiver
```
Dexterp37: Mh, how is this mechanism different than the one above? Do we prefer this one for some reason? If so, we can get rid of the other one, or mention when to use one or the other.

components/service/glean/README.md
```diff
@@ -124,6 +127,33 @@ the following command can be used:
...
+
+`adb shell am broadcast -a {NAME_OF_INTENT} [extra keys]`
+
+The following intents are supported:
+
+- `org.mozilla.components.service.glean.sendPing`: Sends a ping immediately.
+
+  It has two parameters:
+  - pingName: The name of the ping
+  - docType: The docType to use in the URL (optional: if not provided,
```
Dexterp37: Do we have a specific use-case for this? I can't think of any off the top of my head: I'd vote for dropping this :)

components/service/glean/README.md
```diff
@@ -124,6 +127,33 @@ the following command can be used:
...
 
+### Debug intent receiver
+
+Glean also contains an intent receiver that can be sent intents (messages) to
+perform certain actions in a running application.
+
+Users can trigger these intents using the following
+[`adb`](https://developer.android.com/studio/command-line/adb) command:
+
+`adb shell am broadcast -a {NAME_OF_INTENT} [extra keys]`
```
Dexterp37: Can we use this `broadcast` command for doing two actions at time? For example, sending pings and turning on logging on a single command? We will have an upcoming use-case for this, [see the upcoming proposal](https://docs.google.com/document/d/1w0aJK6AqqGAo_wr-iXMyBsmLYQTMKc5BXuSVkBS-MxU/edit?ts=5c6c335a#heading=h.h0ynf4clj2l3).

components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugBroadcastReceiver.kt
```diff
@@ -0,0 +1,68 @@
...
+ *   - org.mozilla.components.service.glean.sendPing
+ *         Sends a ping immediately.  It has two charSequence parameters:
+ *             - pingName: The name of the ping
+ *             - docType: The docType to use in the URL (optional: if not provided,
+ *               equals pingName)
+ *
+ *   - org.mozilla.components.service.glean.logPings
+ *         Turns on logging of ping contents to logCat.
+ */
+class GleanDebugBroadcastReceiver : BroadcastReceiver() {
```
Dexterp37: This is lacking test coverage. Moreover, I'd favour extending `GleanDebugActivity` instead of introducing a new mechanism, unless there's a good reason to do so. Moreover, `GleanDebugActivityTest` already gives you the framework to do test coverage :)

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
mdboom: Ok.  No problem.

components/service/glean/README.md
mdboom: I mentioned this in the PR comment.  I think we should prefer using intents because it doesn't require restarting the app which may cause unnecessary side effects when trying to debug.

components/service/glean/README.md
mdboom: We will when we have custom pings where the user will probably put things on a custom ping but send to a generic doctype endpoint.

components/service/glean/README.md
mdboom: We could put them as different arguments on the same intent, but I don't see a way to send multiple intents with the same broadcast command.

components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugBroadcastReceiver.kt
mdboom: I agree it needs testing. I was hoping to get feedback on the general approach before doing testing.

components/service/glean/README.md
Dexterp37: Ouch, sorry, I should have been more explicit, it's been a hard morning :) I mean, the current approach doesn't require an app restart as well - did you try and it worked differently?

components/service/glean/README.md
Dexterp37: Do we even allow that? A specific custom ping name should end up in a specific doctype endpoint, not a generic one. Moreover, even if we were to do that, tagging in the debug view will address this?

components/service/glean/README.md
Dexterp37: This use case will be required for the debug view :( The other activity allows for that, since we use a single intent

components/service/glean/README.md
travis79: +1 for this!

components/service/glean/README.md
mdboom: It looked like that to me.  I'll double-check, though.  If it doesn't cause a restart, I totally agree there's no need to using a different approach and I'll refactor this PR to use `Activity`.

components/service/glean/README.md
mdboom: I was going off of the signature for `sendPing(pingName, docType)`.  Maybe we should refactor that as well?  We currently only ever use it with `pingName == docType`.

components/service/glean/README.md
mdboom: You're right -- it doesn't seem to cause a restart.  I think I was testing it wrong before.

components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugBroadcastReceiver.kt
travis79: I like the idea of using a broadcast receiver here, are there any security implications if this is allowed in more than just debug though?

components/service/glean/README.md
Dexterp37: Yeah, we should probably refactor that: I don't think a future dev-facing API would allow for custom doctype != pingName.

components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugBroadcastReceiver.kt
mdboom: I've updated this to just use the `Activity` we already had, instead of the broadcast intent.  The biggest practical difference is the Activity is directed at a specific application rather than all applications using glean, which is probably a good thing, though less convenient.

components/service/glean/README.md
mdboom: Ok.  I've filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1531038) for that.  Should be simple enough, but we don't need to confuse this PR with that.

mdboom:
This has been redone to use the existing `Activity` rather than `BroadcastReceiver`.  Now that I got off my wild goose chase, this boils down to a 3 line code change.

colintheshots:
Security-wise, this seems fairly safe. The Intent used is an explicit intent to launch an Activity. Other than appearing in LogCat, it shouldn't leak to other applications on the device unless the device is rooted and has malware installed with root permissions granted.

components/service/glean/src/test/java/mozilla/components/service/glean/debug/GleanDebugActivityTest.kt
```diff
@@ -88,4 +95,40 @@ class GleanDebugActivityTest {
...
+        val booleanMetric = BooleanMetricType(
+            disabled = false,
+            category = "telemetry",
+            lifetime = Lifetime.Application,
+            name = "boolean_metric",
+            sendInPings = listOf("store1")
+        )
+
+        booleanMetric.set(true)
+
```
Dexterp37: I'd add an `assertTrue(booleanMetric.testHasValue())` to be safe: that will wait and make sure that there's data there before we attempt to send. Otherwise, depending on CI, we might have intermittents.

Dexterp37:
> 
> 
> This looks good to me, and given [this comment](https://github.com/mozilla-mobile/android-components/pull/2202#issuecomment-467949264) this shouldn't be a problem. However let's hold this off until we hear back from security.

After discussing this internally and with @georgf , we decided to merge this and ask for a security review of the feature. We could always disable this later in the release cycle.

@mdboom would you mind adding a comment about *not* allowing any command to change the telemeetry server address for security reasons?

mdboom:
@Dexterp37 : I added the comment.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -237,44 +237,53 @@ open class GleanInternalAPI internal constructor () {
...
+        }
 
         // It should only take a single PingUploadWorker to process all queued pings, so we only
         // want to have one scheduled at a time.
-        if (availableToSend) {
+        if (hasPingContent) {
             PingUploadWorker.enqueueWorker()
         }
+
+        return hasPingContent
```
travis79: Are we using the return for this anywhere?  If not, I'd say just remove it since this function handles that internally.

mdboom: Yep -- in `handleEvent`.  It's used to decide whether to reset the metrics ping scheduler counter.

PR #2209, opened 2019-02-27
Closes #2208: Adds isPrivate flag to NewTabSearchUseCase

### LINKED ISSUE #2208: Add NewPrivateTabSearchUseCase
In Fenix we need to be able to create a new search as a private tab through awesomebar. Because of this, we'll want a NewPrivateTabSearchUseCase

sblatz:
### Pull Request checklist

### DIFF
--- a/components/feature/search/src/main/java/mozilla/components/feature/search/SearchUseCases.kt
+++ b/components/feature/search/src/main/java/mozilla/components/feature/search/SearchUseCases.kt
@@ -62,10 +62,11 @@ class SearchUseCases(
     class NewTabSearchUseCase(
         private val context: Context,
         private val searchEngineManager: SearchEngineManager,
-        private val sessionManager: SessionManager
+        private val sessionManager: SessionManager,
+        private val isPrivate: Boolean
     ) : SearchUseCase {
         override fun invoke(searchTerms: String) {
-            invoke(searchTerms, source = Session.Source.NONE, selected = true)
+            invoke(searchTerms, source = Session.Source.NONE, selected = true, private = isPrivate)
         }
 
         /**
@@ -97,6 +98,10 @@ class SearchUseCases(
     }
 
     val newTabSearch: NewTabSearchUseCase by lazy {
-        NewTabSearchUseCase(context, searchEngineManager, sessionManager)
+        NewTabSearchUseCase(context, searchEngineManager, sessionManager, false)
+    }
+
+    val newPrivateTabSearch: NewTabSearchUseCase by lazy {
+        NewTabSearchUseCase(context, searchEngineManager, sessionManager, true)
     }
 }
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -22,6 +22,9 @@ permalink: /changelog/
   }
   ```
 
+* **feature-search**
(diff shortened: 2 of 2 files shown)


components/feature/search/src/main/java/mozilla/components/feature/search/SearchUseCases.kt
```diff
@@ -97,6 +99,6 @@ class SearchUseCases(
     }
 
     val newTabSearch: NewTabSearchUseCase by lazy {
-        NewTabSearchUseCase(context, searchEngineManager, sessionManager)
+        NewTabSearchUseCase(context, searchEngineManager, sessionManager, isPrivate)
     }
```
pocmo: In the issue you mention `NewPrivateTabSearchUseCase`. But it looks like instead of adding a new property here, you decided to add a `private´ parameter. Does this work better in your situation?

sblatz: I realized that the NewPrivateTabSearchUseCase would be the exact same excpet for having an isPrivate flag. To avoid code duplication, I think it makes more sense to just instantiate a type of NewTabSearchUseCase that "isPrivate." Let me know if you disagree!

pocmo: I wonder if we should just introduce another property though, like

```kotlin
    val newPrivateTabSearch by lazy {
        NewTabSearchUseCase(context, searchEngineManager, sessionManager)
    }
```

... which avoids having multiple `SearchUseCases` instances?

pocmo: I wonder if we should just introduce another property though, like

```kotlin
    val newPrivateTabSearch by lazy {
        NewTabSearchUseCase(context, searchEngineManager, sessionManager)
    }
```

... which avoids having multiple `SearchUseCases` instances?

sblatz: Ah, I see. That's a much better solution!

sblatz: I've updated the PR to reflect this. Is this more what you had in mind?

csadilek: If we want a separate use case, another option would be to mark `NewTabSearchUseCase` (and its invoke method) as `open` and subclass it:

```
 class NewPrivateTabSearchUseCase(
        context: Context,
        searchEngineManager: SearchEngineManager,
        sessionManager: SessionManager
    ) : NewTabSearchUseCase(context, searchEngineManager, sessionManager) {

        override fun invoke(searchTerms: String, source: Session.Source, selected: Boolean, private: Boolean) {
            super.invoke(searchTerms, source, selected, true)
        }
    }
```

csadilek: Doesn't make any difference the caller though as they'll just use the `newPrivateTabSearch` field. So, matter of preference mostly.

Issue #1402, opened 2018-11-16
Menu: Allow menu items to be enabled/disabled


pocmo:
For: https://github.com/mozilla-mobile/reference-browser/issues/212

NotWoods:
Duplicate of #169

Issue #1380, opened 2018-11-15
browser-menu: Add ability for adding a two-state button


pocmo:
*Maybe that's already possible with what we have. Let's try and add functionality if needed.*

For some menu buttons we want to have two states that can change dynamically. One example of that is the reload button: If the page is loaded you can "reload" and if the page is still loading you can "stop".

jonalmeida:
We have `BrowserToolbar.TwoStateButton` already! 🎉

Amejia481:
@jonalmeida then Can we close this issue?

pocmo:
@Amejia481 No, in this case we need something similar for inside the menu! :)

ekager:
So should #2201 close this issue, or is this still different?

Amejia481:
I think they are the same!

PR #2211, opened 2019-02-27
Close #2201 - Adds TwoStateButton in BrowserMenuItemToolbar

### LINKED ISSUE #2201: Add TwoStateButton with ability to disable in BrowserMenuItemToolbar
In Fenix, we need a button in BrowserMenuItemToolbar that can have 2 states, enabled and disabled, and have an option to be unclickable (disabled) when in the visibly disabled state. 
We will use this for refresh/stop, back, and forward.
See https://github.com/mozilla-mobile/fenix/issues/654 for context.

ekager:
One note that I think we need some AC input on, this works well for the enabled/disabled back/forward button with stylization, but for the refresh/stop button, I don't think we currently have a way to refresh the menu in the same way that the toolbar has an invalidate method, so it won't change state while the menu is open and the page finishes loading. Should I open a new issue for that?

### Pull Request checklist

### DIFF
--- a/components/browser/menu/src/main/java/mozilla/components/browser/menu/item/BrowserMenuItemToolbar.kt
+++ b/components/browser/menu/src/main/java/mozilla/components/browser/menu/item/BrowserMenuItemToolbar.kt
@@ -30,27 +30,42 @@ class BrowserMenuItemToolbar(
 
         for (item in items) {
             val button = AppCompatImageButton(view.context)
-            button.setImageResource(item.imageResource)
+            if (item is TwoStateButton && !item.isInPrimaryState.invoke()) {
+                button.setImageResource(item.secondaryImageResource)
+                button.contentDescription = item.secondaryContentDescription
+                if (item.secondaryImageTintResource != 0) {
+                    button.imageTintList = ContextCompat.getColorStateList(
+                        view.context,
+                        item.secondaryImageTintResource
+                    )
+                }
+                button.isEnabled = !item.disableInSecondaryState
+            } else {
+                button.setImageResource(item.imageResource)
+                button.contentDescription = item.contentDescription
+                if (item.iconTintColorResource != 0) {
+                    button.imageTintList =
+                        ContextCompat.getColorStateList(view.context, item.iconTintColorResource)
+                }
+            }
 
             val outValue = TypedValue()
             view.context.theme.resolveAttribute(
-                    android.R.attr.selectableItemBackgroundBorderless,
-                    outValue,
-                    true)
+                android.R.attr.selectableItemBackgroundBorderless,
+                outValue,
+                true
+            )
 
             button.setBackgroundResource(outValue.resourceId)
-            button.contentDescription = item.contentDescription
             button.setOnClickListener {
                 item.listener.invoke()
                 menu.dismiss()
             }
 
-            if (item.iconTintColorResource != 0) {
-                button.imageTintList = ContextCompat.getColorStateList(view.context, item.iconTintColorResource)
-            }
-
-            layout.addView(button,
-                LinearLayout.LayoutParams(0, view.resources.pxToDp(ICON_HEIGHT_DP), 1f))
+            layout.addView(
+                button,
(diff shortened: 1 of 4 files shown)


pocmo:
> Should I open a new issue for that?

Yeah, let's do that! We also have #1380 and #1402. Feel free to close or re-file as needed.

ekager:
Ah okay, sorry! Yeah this definitely looks like #1380 and #1402

components/browser/menu/src/main/java/mozilla/components/browser/menu/item/BrowserMenuItemToolbar.kt
```diff
@@ -62,13 +77,44 @@ class BrowserMenuItemToolbar(
...
+     * @param primaryImageTintResource Optional ID of color resource to tint the primary icon.
+     * @param secondaryImageResource Optional ID of a different drawable resource to be shown as secondary icon.
+     * @param secondaryContentDescription Optional secondary content description for button, for accessibility support.
+     * @param secondaryImageTintResource Optional ID of secondary color resource to tint the icon.
+     * @param isInPrimaryState Lambda to return true/false to indicate if this button should be primary or secondary.
+     * @param disableInSecondaryState Optional boolean to disable the button when in secondary state.
+     * @param listener Callback to be invoked when the button is pressed.
+     */
+    open class TwoStateButton(
+        val primaryImageResource: Int,
```
jonalmeida: nit: I might be wrong, but we might be using 'initial' and 'secondary' as the correct terms. If not, ignore.

components/browser/menu/src/main/java/mozilla/components/browser/menu/item/BrowserMenuItemToolbar.kt
```diff
@@ -62,13 +77,44 @@ class BrowserMenuItemToolbar(
...
+        val primaryContentDescription: String,
+        val primaryImageTintResource: Int = 0,
+        val secondaryImageResource: Int = primaryImageResource,
+        val secondaryContentDescription: String = primaryContentDescription,
+        val secondaryImageTintResource: Int = primaryImageTintResource,
+        val isInPrimaryState: () -> Boolean = { true },
```
jonalmeida: We could pass the current button's state in the listener below: `listener: (Boolean) -> Unit`

components/browser/menu/src/main/java/mozilla/components/browser/menu/item/BrowserMenuItemToolbar.kt
```diff
@@ -30,27 +30,42 @@ class BrowserMenuItemToolbar(
 
         for (item in items) {
             val button = AppCompatImageButton(view.context)
-            button.setImageResource(item.imageResource)
+            if (item is TwoStateButton && !item.isInPrimaryState.invoke()) {
```
jonalmeida: This can get complex really quickly if we do type and state checks. Maybe we can create a `BrowserMenuTwoStateButton`? We have an abstract `BrowserMenuCompoundButton` that might be useful. It's currently only used for `BrowserMenuCheckbox` and `BrowserMenuSwitch`.

We also have `BrowserToolbar.TwoStateButton` that might have an API similar to one that you need that can be copied for the BrowserMenu.

ekager:
Thanks for the review @jonalmeida ! 
So I did look at the original implementation of `TwoStateButton` in the toolbar and reused a lot of parts here. I decided to change the names from `enabled`/`disabled` to `primary`/`secondary` to reflect that the two states may have different image resources/tints, and that the secondary state button may not be disabled (unclickable) (i.e. the case of the refresh/stop button). `initialState` is used for the `BrowserMenuCompoundButton` to refer to the checked/unchecked status. Hope this gives some context and let me know if you still think it's unclear! 

https://github.com/mozilla-mobile/android-components/blob/b63ffeee6ab3026d57e376d3fc20001d6bd9fd3a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt#L561-L591

PR #2215, opened 2019-02-27
Implement persistent storage for Ping lifetime metrics

travis79:
Add ping lifetime storage using SharedPreferences similar to how User lifetime storage is used, except that when Ping lifetime data is snapshot, it will be cleared from the cache and the SharedPreferences.

### Pull Request checklist

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
@@ -44,7 +44,12 @@ internal abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
     // useful when debugging weird behaviours.
     abstract val logger: Logger
 
-    protected val userLifetimeStorage: SharedPreferences by lazy { deserializeUserLifetime() }
+    protected val userLifetimeStorage: SharedPreferences by lazy {
+        deserializeLifetime(Lifetime.User)
+    }
+    protected val pingLifetimeStorage: SharedPreferences by lazy {
+        deserializeLifetime(Lifetime.Ping)
+    }
 
     // Store a map for each lifetime as an array element:
     // Array[Lifetime] = Map[StorageName, ScalarType].
@@ -79,22 +84,32 @@ internal abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
     )
 
     /**
-     * Deserialize the metrics with a lifetime = User that are on disk.
+     * Deserialize the metrics with a particular lifetime that are on disk.
      * This will be called the first time a metric is used or before a snapshot is
      * taken.
      *
-     * @return A [SharedPreferences] reference that will be used to initialize [userLifetimeStorage]
+     * @param lifetime a [Lifetime] to deserialize
+     * @return A [SharedPreferences] reference that will be used to initialize [pingLifetimeStorage]
+     *         or [userLifetimeStorage] or null if an invalid lifetime is used.
      */
-    @Suppress("TooGenericExceptionCaught")
-    open fun deserializeUserLifetime(): SharedPreferences {
-        val prefs =
-            applicationContext.getSharedPreferences(this.javaClass.canonicalName, Context.MODE_PRIVATE)
+    @Suppress("TooGenericExceptionCaught", "ComplexMethod")
+    open fun deserializeLifetime(lifetime: Lifetime): SharedPreferences {
+        require(lifetime == Lifetime.Ping || lifetime == Lifetime.User) {
+            "deserializeLifetime does not support Lifetime.Application"
+        }
+
+        val prefsName = if (lifetime == Lifetime.Ping) {
+            "${this.javaClass.canonicalName}.PingLifetime"
+        } else {
+            this.javaClass.canonicalName
+        }
+        val prefs = applicationContext.getSharedPreferences(prefsName, Context.MODE_PRIVATE)
 
         val metrics = try {
             prefs.all.entries
         } catch (e: NullPointerException) {
             // If we fail to deserialize, we can log the problem but keep on going.
-            logger.error("Failed to deserialize metric with 'user' lifetime")
+            logger.error("Failed to deserialize metric with ${lifetime.name} lifetime")
             return prefs
         }
 
@@ -111,10 +126,12 @@ internal abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
                 continue
             }
 
-            val storeData = dataStores[Lifetime.User.ordinal].getOrPut(storeName) { mutableMapOf() }
+            val storeData = dataStores[lifetime.ordinal].getOrPut(storeName) { mutableMapOf() }
             // Only set the stored value if we're able to deserialize the persisted data.
-            deserializeSingleMetric(metricName, metricValue)?.let {
-                storeData[metricName] = it
+            deserializeSingleMetric(metricName, metricValue)?.let { value ->
+                storeData?.let {
+                    it[metricName] = value
+                }
             } ?: logger.warn("Failed to deserialize $metricStoragePath")
         }
 
@@ -122,14 +139,15 @@ internal abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
     }
 
     /**
-     * Ensures that the user lifetime metrics (in [userLifetimeStorage]) is
+     * Ensures that the lifetime metrics in [pingLifetimeStorage] and [userLifetimeStorage] is
      * loaded.  This is a no-op if they are already loaded.
      */
-    private fun ensureUserLifetimeLoaded() {
-        // Make sure data with "user" lifetime is loaded.
+    private fun ensureAllLifetimesLoaded() {
+        // Make sure data with the provided lifetime is loaded.
         // We still need to catch exceptions here, as `getAll()` might throw.
         @Suppress("TooGenericExceptionCaught")
         try {
+            pingLifetimeStorage.all
             userLifetimeStorage.all
         } catch (e: NullPointerException) {
             // Intentionally left blank. We just want to fall through.
@@ -154,7 +172,7 @@ internal abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
     fun getSnapshot(storeName: String, clearStore: Boolean): GenericDataStorage<ScalarType>? {
         val allLifetimes: GenericDataStorage<ScalarType> = mutableMapOf()
 
-        ensureUserLifetimeLoaded()
+        ensureAllLifetimesLoaded()
 
         // Get the metrics for all the supported lifetimes.
         for (store in dataStores) {
@@ -165,6 +183,11 @@ internal abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
 
         if (clearStore) {
             // We only allow clearing metrics with the "ping" lifetime.
+            val editor = pingLifetimeStorage.edit()
+            dataStores[Lifetime.Ping.ordinal][storeName]?.keys?.forEach { key ->
+                editor.remove("$storeName#$key")
+            }
+            editor.apply()
             dataStores[Lifetime.Ping.ordinal].remove(storeName)
         }
 
@@ -194,8 +217,8 @@ internal abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
      *     those stores.
      */
     override fun getIdentifiersInStores(stores: List<String>) = sequence {
-        // Make sure data with "user" lifetime is loaded before getting the snapshot.
-        ensureUserLifetimeLoaded()
+        // Make sure data with "user" and "ping" lifetimes are loaded before getting the snapshot.
+        ensureAllLifetimesLoaded()
 
         dataStores.forEach { lifetime ->
             stores.forEach {
@@ -241,31 +264,35 @@ internal abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
 
         // Record a copy of the metric in all the needed stores.
         @SuppressLint("CommitPrefEdits")
-        val userPrefs: SharedPreferences.Editor? =
-            if (metricData.lifetime == Lifetime.User) userLifetimeStorage.edit() else null
-        metricData.getStorageNames().forEach {
-            val storeData = dataStores[metricData.lifetime.ordinal].getOrPut(it) { mutableMapOf() }
+        val editor: SharedPreferences.Editor? = when (metricData.lifetime) {
+            Lifetime.User -> userLifetimeStorage.edit()
+            Lifetime.Ping -> pingLifetimeStorage.edit()
+            else -> null
+        }
+        metricData.getStorageNames().forEach { store ->
+            val storeData = dataStores[metricData.lifetime.ordinal].getOrPut(store) { mutableMapOf() }
             // We support empty categories for enabling the internal use of metrics
             // when assembling pings in [PingMaker].
             val entryName = metricData.identifier
             val combinedValue = combine(storeData[entryName], value)
             storeData[entryName] = combinedValue
-            // Persist data with "user" lifetime
-            if (metricData.lifetime == Lifetime.User) {
+            // Persist data with "user" or "ping" lifetimes
+            editor?.let {
                 serializeSingleMetric(
-                    userPrefs,
-                    "$it#$entryName",
+                    it,
+                    "$store#$entryName",
                     combinedValue,
                     extraSerializationData
                 )
             }
         }
-        userPrefs?.apply()
+        editor?.apply()
     }
 
     @VisibleForTesting(otherwise = VisibleForTesting.NONE)
     override fun clearAllStores() {
         userLifetimeStorage.edit().clear().apply()
+        pingLifetimeStorage.edit().clear().apply()
         dataStores.forEach { it.clear() }
     }
 }
--- a/components/service/glean/src/test/java/mozilla/components/service/glean/storages/GenericScalarStorageEngineTest.kt
+++ b/components/service/glean/src/test/java/mozilla/components/service/glean/storages/GenericScalarStorageEngineTest.kt
@@ -11,6 +11,7 @@ import mozilla.components.service.glean.CommonMetricData
 import mozilla.components.service.glean.Lifetime
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertFalse
+import org.junit.Assert.assertNull
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.mockito.ArgumentMatchers.eq
@@ -165,6 +166,12 @@ class GenericScalarStorageEngineTest {
             eq(MockScalarStorageEngine::class.java.canonicalName),
             eq(Context.MODE_PRIVATE)
         )).thenReturn(sharedPreferences)
+        `when`(context.getSharedPreferences(
+            eq("${MockScalarStorageEngine::class.java.canonicalName}.PingLifetime"),
+            eq(Context.MODE_PRIVATE)
+        )).thenReturn(ApplicationProvider.getApplicationContext<Context>()
+            .getSharedPreferences("${MockScalarStorageEngine::class.java.canonicalName}.PingLifetime",
+                Context.MODE_PRIVATE))
 
         // Instantiate our mock engine and check that it correctly unpersists the
         // data and makes it available in the snapshot.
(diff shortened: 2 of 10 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -105,13 +112,14 @@ internal abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
 
             // Split the stored name in 2: we expect it to be in the format
             // store#metric.name
-            val (storeName, metricName) =
-                metricStoragePath.split('#', limit = 2)
+            val (storeName, metricName) = metricStoragePath.split('#', limit = 2)
```
Dexterp37: nit: no need to change this and the next line

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -105,13 +112,14 @@ internal abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
...
+
             if (storeName.isEmpty()) {
                 continue
             }
 
-            val storeData = dataStores[Lifetime.User.ordinal].getOrPut(storeName) { mutableMapOf() }
+            val storeData = getDataStoreForLifetime(lifetime, storeName)
+
```
Dexterp37: nit: extra blank line

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -122,15 +130,54 @@ internal abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
     }
 
     /**
-     * Ensures that the user lifetime metrics (in [userLifetimeStorage]) is
+     * Gets SharedPreference for a particular lifetime
+     *
+     * @param lifetime [Lifetime] for which to get the preferences for
+     * @return [SharedPreferences]
```
Dexterp37: nit: this is incorrect: we either return a valid `[SharedPreferences]` or `null`

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -37,14 +37,20 @@ internal typealias ScalarRecordingCombiner<T> = (currentValue: T?, newValue: T)
...
 internal abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
     override lateinit var applicationContext: Context
 
     // Let derived class define a logger so that they can provide a proper name,
     // useful when debugging weird behaviours.
     abstract val logger: Logger
 
-    protected val userLifetimeStorage: SharedPreferences by lazy { deserializeUserLifetime() }
+    protected val userLifetimeStorage: SharedPreferences by lazy {
+        deserializeLifetime(Lifetime.User)!!
```
Dexterp37: The `!!` after this is scary. However, given that the function only returns `null` if `Lifetime.Application`, which is not the case here, we should be ok. However, I'd add a comment above this and line 52.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -79,22 +85,23 @@ internal abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
...
     @Suppress("TooGenericExceptionCaught")
-    open fun deserializeUserLifetime(): SharedPreferences {
-        val prefs =
-            applicationContext.getSharedPreferences(this.javaClass.canonicalName, Context.MODE_PRIVATE)
+    open fun deserializeLifetime(lifetime: Lifetime): SharedPreferences? {
+        val prefs = getPrefsForLifetime(lifetime)
 
         val metrics = try {
-            prefs.all.entries
+            prefs!!.all.entries
```
Dexterp37: We're intentionally causing a `NullPointerException` here if `lifetime == Lifetime.Application`. However, `prefs` is a nullable and the else branch is really just for the `.all.entries` part. So what about we change this to `prefs?.all.entries` and add a comment there mentioning why are we using `?` in the first place?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -105,13 +112,14 @@ internal abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
...
-            val (storeName, metricName) =
-                metricStoragePath.split('#', limit = 2)
+            val (storeName, metricName) = metricStoragePath.split('#', limit = 2)
+
             if (storeName.isEmpty()) {
                 continue
             }
 
-            val storeData = dataStores[Lifetime.User.ordinal].getOrPut(storeName) { mutableMapOf() }
+            val storeData = getDataStoreForLifetime(lifetime, storeName)
```
Dexterp37: I think this could just become `val storeData = dataStores[lifetime.ordinal].getOrPut(storeName) { mutableMapOf() }`: this way we can get rid of the new function, possibly removing the need for the suppression at the top of the class too.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -79,22 +85,23 @@ internal abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
...
+     * @param lifetime a [Lifetime] to deserialize
+     * @return A [SharedPreferences] reference that will be used to initialize [pingLifetimeStorage]
+     *         or [userLifetimeStorage] or null if an invalid lifetime is used.
      */
     @Suppress("TooGenericExceptionCaught")
-    open fun deserializeUserLifetime(): SharedPreferences {
-        val prefs =
-            applicationContext.getSharedPreferences(this.javaClass.canonicalName, Context.MODE_PRIVATE)
+    open fun deserializeLifetime(lifetime: Lifetime): SharedPreferences? {
+        val prefs = getPrefsForLifetime(lifetime)
```
Dexterp37: If we add `require` above, then this could become:

```
val prefsName = if (lifetime == Lifetime.Ping) { 
  "${this.javaClass.canonicalName}.PingLifetime" 
} else {
  this.javaClass.canonicalName
}
val prefs = applicationContext.getSharedPreferences(prefsName, Context.MODE_PRIVATE)
```

This way you won't need to add a new function and the code would be covered by other tests.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -122,15 +130,54 @@ internal abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
...
+        }
+    }
+
+    /**
+     * Gets correct data store for a particular lifetime
+     *
+     * @param lifetime [Lifetime] for which to get the preferences for
+     * @return [SharedPreferences]
+     */
+    private fun getDataStoreForLifetime(lifetime: Lifetime, storeName: String):
```
Dexterp37: We can do without this function, check my previous comments.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -122,15 +130,54 @@ internal abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
...
+    }
+
+    /**
+     * Ensures that the lifetime metrics in [pingLifetimeStorage] or [userLifetimeStorage] is
      * loaded.  This is a no-op if they are already loaded.
      */
-    private fun ensureUserLifetimeLoaded() {
-        // Make sure data with "user" lifetime is loaded.
+    private fun ensureLifetimeLoaded(lifetime: Lifetime) {
+        // Make sure data with "lifetime" is loaded.
```
Dexterp37: nit: "with the provided lifetime is loaded."?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -79,22 +85,23 @@ internal abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
...
-     * @return A [SharedPreferences] reference that will be used to initialize [userLifetimeStorage]
+     * @param lifetime a [Lifetime] to deserialize
+     * @return A [SharedPreferences] reference that will be used to initialize [pingLifetimeStorage]
+     *         or [userLifetimeStorage] or null if an invalid lifetime is used.
      */
     @Suppress("TooGenericExceptionCaught")
-    open fun deserializeUserLifetime(): SharedPreferences {
-        val prefs =
-            applicationContext.getSharedPreferences(this.javaClass.canonicalName, Context.MODE_PRIVATE)
+    open fun deserializeLifetime(lifetime: Lifetime): SharedPreferences? {
```
Dexterp37: This function really _requires us_ to have `lifetime` either be `Lifetime.User || Lifetime.Ping`. We can tell Kotlin about that, by using the `require` function (see the [docs](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/require.html))

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -122,15 +130,54 @@ internal abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
...
+    private fun ensureLifetimeLoaded(lifetime: Lifetime) {
+        // Make sure data with "lifetime" is loaded.
         // We still need to catch exceptions here, as `getAll()` might throw.
         @Suppress("TooGenericExceptionCaught")
         try {
-            userLifetimeStorage.all
+            when (lifetime) {
+                Lifetime.Ping -> pingLifetimeStorage.all
+                Lifetime.User -> userLifetimeStorage.all
+                Lifetime.Application -> return
```
Dexterp37: CI coverage tells us that this line is not covered by tests :)

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -166,6 +214,15 @@ internal abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
...
             dataStores[Lifetime.Ping.ordinal].remove(storeName)
+
+            // Also remove from SharedPreferences
+            val editor = pingLifetimeStorage.edit()
+            pingLifetimeStorage.all.forEach {
+                if (it.key.startsWith("$storeName#")) {
+                    editor.remove(it.key)
+                }
+            }
+            editor.apply()
```
mdboom: If you iterated through the keys in `dataStores[Lifetime.Ping.ordinal]` and removed `$storeName#$key` from `pingLifetimeStorage`, this would be `O(n)` where `n` is the number of keys in the current store, rather than `O(m)` where `m` is the number of keys in all stores.  `m > n` should always hold true, and you have to do the hash lookups to remove the entries in the SharedPreferences either way.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -242,16 +300,20 @@ internal abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
         // Record a copy of the metric in all the needed stores.
         @SuppressLint("CommitPrefEdits")
         val userPrefs: SharedPreferences.Editor? =
-            if (metricData.lifetime == Lifetime.User) userLifetimeStorage.edit() else null
+            when {
+                metricData.lifetime == Lifetime.User -> userLifetimeStorage.edit()
+                metricData.lifetime == Lifetime.Ping -> pingLifetimeStorage.edit()
+                else -> null
+            }
```
mdboom: I think you could do:

```
when (metricData.lifetime) {
  Lifetime.User -> ...
  Lifetime.Ping -> ...
  else ->
}
```

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -242,16 +300,20 @@ internal abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
         // Record a copy of the metric in all the needed stores.
         @SuppressLint("CommitPrefEdits")
         val userPrefs: SharedPreferences.Editor? =
```
mdboom: Maybe rename `userPrefs` since it can now be `user` or `ping`?

mdboom:
> If products record "startup time", and we have, say 1000k metrics stored, this would cause startup jank.

In most cases, this will run inside our `Dispatcher` thread, though, which hopefully ameliorates it somewhat.

For dynamic labeled metrics, though, I couldn't find a way to do this off the main thread and we have [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1530733).  That's somewhat orthogonal to this PR.  I think to fix that bug we either (1) move to rkv, or (2) would have to do the determination of whether to use an `__other__` label deeper in the stack, which would require a lot more "knowledge" about labeled metrics everwhere.

Dexterp37:
> 
> 
> > If products record "startup time", and we have, say 1000k metrics stored, this would cause startup jank.
> 
> In most cases, this will run inside our `Dispatcher` thread, though, which hopefully ameliorates it somewhat.
> 
> For dynamic labeled metrics, though, I couldn't find a way to do this off the main thread and we have [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1530733). That's somewhat orthogonal to this PR. I think to fix that bug we either (1) move to rkv, or (2) would have to do the determination of whether to use an `__other__` label deeper in the stack, which would require a lot more "knowledge" about labeled metrics everwhere.

Yes, that's true. It's probably worth waiting for Rurst/rkv a bit more before getting to scared about this.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
travis79: I refactored it into a function because detekt didn't like the function complexity.  I can suppress the warning and do away with the extra functions.

Dexterp37: Ok, as long as you add unit tests for it

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
travis79: Fixed so I don't need to use either !! or ?.  It looks like `getSharedPreferences()` guarantees that it returns something, according to the docs.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
travis79: I like this approach, but I had to iterate through `dataStores[Lifetime.Ping.ordinal][storeName].keys` to get to the right level to accomplish this.

travis79:
@Dexterp37 in regards to how this behaves at startup?  It actually does pretty good.  I've spent a little time researching to see if there's a better API than SharedPreferences for this and there isn't much out there that's not a database or database DAO type wrapper, and those are worse on performance as they have overhead of DB engine when we don't need the power that entails.

Other than that, ready for another look!

Dexterp37:
> 
> 
> @Dexterp37 in regards to how this behaves at startup? It actually does pretty good.

I mean, if we had 10k "ping" lifetime metrics persisted, then we started the app using glean and attempted to modify the value of one of these metrics right at startup, we would currently block on that. Which might not be ideal. However, let's file a bug about this and defer fixing it when it becomes a problem.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -241,31 +264,35 @@ internal abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
...
+            editor?.let {
                 serializeSingleMetric(
-                    userPrefs,
-                    "$it#$entryName",
+                    it,
+                    "$store#$entryName",
                     combinedValue,
                     extraSerializationData
                 )
+                it.apply()
```
Dexterp37: This will trigger the I/O write on each iteration of the `forEach`. Let's leave this back where it was, so that the flush is only triggered once.

travis79:
> > @Dexterp37 in regards to how this behaves at startup? It actually does pretty good.
> 
> I mean, if we had 10k "ping" lifetime metrics persisted, then we started the app using glean and attempted to modify the value of one of these metrics right at startup, we would currently block on that. Which might not be ideal. However, let's file a bug about this and defer fixing it when it becomes a problem.

@Dexterp37 I believe there is already a [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1505741) filed that covers this.

travis79:
> I think this is still lacking this test coverage: we need to at least cover: missing "ping lifetime" file, corrupted "ping lifetime" file; making sure that the local storage is correctly cleared when sending a ping.

I don't think there's any way to test for a missing ping lifetime file, since `getSharedPreferences` creates it if it doesn't exist.  As for corrupted, how can I tell that from the `SharedPreferences` API?  The calls to `getBoolean` or `getString` either work or they don't, I guess I don't understand what you want me to test for.  Is there a similar test applied to the User storage that you would like to see implemented for the Ping storage?  I will add a test that explicitly tests the ping lifetime storage getting cleared out if possible, but I need a little more information on exactly what you want to see tested for the first two items.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt
```diff
@@ -165,6 +183,11 @@ internal abstract class GenericScalarStorageEngine<ScalarType> : StorageEngine {
 
         if (clearStore) {
             // We only allow clearing metrics with the "ping" lifetime.
+            val editor = pingLifetimeStorage.edit()
+            dataStores[Lifetime.Ping.ordinal][storeName]?.keys?.forEach { key ->
+                editor.remove("$storeName#$key").apply()
```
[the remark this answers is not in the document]
travis79: Good catch, thanks!  I did move it back out of the loop, I just forgot to remove the .apply() that was inside the loop.

Dexterp37:
> 
> 
> > I think this is still lacking this test coverage: we need to at least cover: missing "ping lifetime" file, corrupted "ping lifetime" file; making sure that the local storage is correctly cleared when sending a ping.
> 
> I don't think there's any way to test for a missing ping lifetime file, since `getSharedPreferences` creates it if it doesn't exist. As for corrupted, how can I tell that from the `SharedPreferences` API? The calls to `getBoolean` or `getString` either work or they don't, I guess I don't understand what you want me to test for. Is there a similar test applied to the User storage that you would like to see implemented for the Ping storage? I will add a test that explicitly tests the ping lifetime storage getting cleared out if possible, but I need a little more information on exactly what you want to see tested for the first two items.

We cleared this up on vidyo: my bad, as discussed, the only test we really need is the one for checking that the lifetime storage gets correctly cleared!

components/service/glean/src/test/java/mozilla/components/service/glean/storages/GenericScalarStorageEngineTest.kt
```diff
@@ -394,4 +419,73 @@ class GenericScalarStorageEngineTest {
...
+                metric1,
+                value = 37
+            )
+
+            val metric2 = GenericMetricType(
+                disabled = false,
+                category = "telemetry",
+                lifetime = Lifetime.Ping,
+                name = "pingLifetimeData",
+                sendInPings = listOf("store1", "store2")
```
Dexterp37: I would also check that data is never cleared from "store2" in every `run` block below.

PR #2249, opened 2019-03-05
Closes #2246: document how to add a new gradle module (with template)

### LINKED ISSUE #2246: Add docs: recommended way to create a new module
As a developer, it is really hard to contribute to components because the code I want to add often requires creating a new gradle module. However, this project is configured in a non-traditional way which makes it hard to figure out and makes the built-in AS tools useless. We should provide documentation to make this process easier.

mcomella:
~~**BEFORE LANDING**: I need to provide an example for what adding a module might look like and I want to reference the commits from PR https://github.com/mozilla-mobile/android-components/pull/2247, which hasn't landed yet.~~

_edit:_ done.

### Pull Request checklist

### DIFF
--- a/docs/contribute/add_a_gradle_module.md
+++ b/docs/contribute/add_a_gradle_module.md
@@ -1,0 +1,45 @@
+---
+layout: page
+title: Add a new Gradle module
+permalink: /contributing/add-a-gradle-module
+---
+
+This document describes how to add a new Gradle module, also known as a Gradle project or subproject. For example, by following this guide, you can add a new module like `:concept-engine`.
+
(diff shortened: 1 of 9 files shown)


mcomella:
@pocmo This should be ready for review.

mcomella:
@pocmo That's fair. Would it make more sense to have an unpublished "template" module that no code gets added to and that developers can copy into a new module? I find it a pain to make sure I remove all the code and everything too. Alternatively, I could just document, "Copy an existing module" (because it's more clear to be explicit than not having any related documentation).

pocmo:
@mcomella I think an unpublished template module could be a good idea! My only concern would be to have gradle do more work for every engineer - but then again we have so many modules that an empty module wouldn't have any measurable negative impact, I guess. :)

PR #2250, opened 2019-03-05
Integrate site permissions storage.

Amejia481:
Depends on https://github.com/mozilla-mobile/android-components/pull/2199
### Pull Request checklist


components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
```diff
@@ -57,6 +64,14 @@ class SitePermissionsFeature(
...
     private val observer = SitePermissionsRequestObserver(sessionManager, feature = this)
 
+    internal val storage by lazy { storageInitializer() }
+    internal val ioCoroutineScope by lazy { CoroutineScope(Dispatchers.IO) }
+
+    @VisibleForTesting
+    internal var storageInitializer = {
+        SitePermissionsStorage(anchorView.context)
+    }
+
```
pocmo: I think we should pass this in as a parameter to the constructor of the feature. The reason for this is that the app probably will need this storage too (since it may build app-specific UI). And I assume we want to avoid that there are multiple instances of this?

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
```diff
@@ -96,32 +111,146 @@ class SitePermissionsFeature(
...
     private fun onContentPermissionGranted(session: Session, grantedPermissions: List<Permission>) {
-        session.contentPermissionRequest.consume {
-            it.grant()
-            // Update the DB
+        session.contentPermissionRequest.consume { request ->
+            request.grant()
+            ioCoroutineScope.launch {
+                storeSitePermissions(request, grantedPermissions, ALLOWED)
+            }
             true
```
pocmo: So far we ignored the checkbox in the mocks. I guess now we would need the "Don't ask again" checkbox to determine whether we would need to save this?

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
```diff
@@ -96,32 +111,146 @@ class SitePermissionsFeature(
...
+        var sitePermissions = storage.findSitePermissionsBy(request.host)
+
+        if (sitePermissions == null) {
+            sitePermissions = request.toSitePermissions(status = status, permissions = permissions)
+            storage.save(sitePermissions)
+        } else {
+            sitePermissions = request.toSitePermissions(status, sitePermissions)
+            storage.update(sitePermissions)
+        }
+    }
```
pocmo: It looks like this needs to be `synchronized` so that we are not updating the same permission concurrently?

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
```diff
@@ -370,13 +498,19 @@ class SitePermissionsFeature(
         return false
     }
 
+    private val PermissionRequest.host get() = uri?.toUri()?.host ?: ""
+    private val Permission.isBackCamera get() = desc?.contains("back") ?: false
+    private val Permission.isFrontCamera get() = desc?.contains("front") ?: false
```
pocmo: This code existed before this PR, but I wonder: Is the description localized? Are there other descriptions? Can't we just use the descriptions here instead of translating them to other strings?

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
Amejia481: Sound reasonable 👍

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
Amejia481: Yes, I agree.
I have some questions about how the checkbox should work. I filled this issue https://github.com/mozilla-mobile/android-components/issues/2287 to discuss about it

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
Amejia481: 👍

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
pocmo: 👍 Let's write them down there and then ping Bram.

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
Amejia481: > This code existed before this PR

I saw what the GV team did in [geckoview_example](https://dxr.mozilla.org/mozilla-central/rev/c2593a3058afdfeaac5c990e18794ee8257afe99/mobile/android/geckoview_example/src/main/java/org/mozilla/geckoview_example/GeckoViewActivity.java#615).

```java
        private String[] normalizeMediaName(final MediaSource[] sources) {
            if (sources == null) {
                return null;
            }

            String[] res = new String[sources.length];
            for (int i = 0; i < sources.length; i++) {
                final int mediaSource = sources[i].source;
                final String name = sources[i].name;
                if (MediaSource.SOURCE_CAMERA == mediaSource) {
                    if (name.toLowerCase(Locale.ENGLISH).contains("front")) {
                        res[i] = getString(R.string.media_front_camera);
                    } else {
                        res[i] = getString(R.string.media_back_camera);
                    }
                } else if (!name.isEmpty()) {
                    res[i] = name;
                } else if (MediaSource.SOURCE_MICROPHONE == mediaSource) {
                    res[i] = getString(R.string.media_microphone);
                } else {
                    res[i] = getString(R.string.media_other);
                }
            }

            return res;
        }
```


> Is the description localized?

I have to investigate because right now this is the only thing that help us to identify between one or another. And if it is localized can break that logic.  

>  Are there other descriptions?

So far, I just saw these two, and one for the microphone. I have to investigate.
In the Fenix, mock-up I saw additional microphone devices but in code I just get one.
image



> Can't we just use the descriptions here instead of translating them to other strings?

I think it won't be ideal for  apps, for instance Fenix has its own strings
image

image

Amejia481:
@pocmo  I will wait for Bram response on https://github.com/mozilla-mobile/android-components/issues/2287 because the logic here could changes

pocmo:
@Amejia481 Yeah, agreed, sounds good.

Amejia481:
@pocmo I updated the PR, including the recommendations, the "Do not ask again on this site" checkbox and #2351 issue.

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissions.kt
```diff
@@ -12,17 +12,22 @@ import mozilla.components.feature.sitepermissions.SitePermissions.Status.NO_DECI
 data class SitePermissions(
     val origin: String,
     val location: Status = NO_DECISION,
+    val doNotAskAgainLocation: Boolean = false,
     val notification: Status = NO_DECISION,
     val microphone: Status = NO_DECISION,
+    val doNotAskAgainMicrophone: Boolean = false,
     val cameraBack: Status = NO_DECISION,
     val cameraFront: Status = NO_DECISION,
+    val doNotAskAgainCamera: Boolean = false,
```
pocmo: I'm confused about the addition of those booleans. Aren't those already covered by the others (e.g. only if the checkbox is checked we need to save anything to the db). Let's chat about that today. :)

Amejia481: I think [this video of how Fennec is using the check box](https://drive.google.com/file/d/1176l1-UUZwitgSX1Hv0z8aNmg6ibhdse/view?usp=sharing), can help to clarify.

From what I could saw on Fennec, the checkbox when is checked remember the last user dissension.

For example: If the user checks the checkbox and allows location next time, the user enters again to this site the location will be auto-granted but if the user doesn't check the checkbox a prompt will be always shown either if the user selects allow or deny.

Amejia481: I think I got what you meant. The idea is to only store the info when the checkbox is checked, otherwise, not store anything always show the prompt. Right?

pocmo: Yes, that's how I understood it.

Amejia481: I updated the PR 👍

Amejia481:
🎉

PR #2266, opened 2019-03-06
Closes #2265: Validating null or empty headers in the download manager

### LINKED ISSUE #2265: Validate no null or empty values are added as headers to the DownloadManager
https://github.com/mozilla-mobile/reference-browser/issues/637

Amejia481:
For some reasons `download.userAgent` was null and we were passing this value as a header to the download manager. The Android download manager doesn't like null or empty headers!

### Pull Request checklist

### DIFF
--- a/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadManager.kt
+++ b/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadManager.kt
@@ -73,15 +73,15 @@ class DownloadManager(
 
         val request = Request(Uri.parse(download.url))
             .setNotificationVisibility(VISIBILITY_VISIBLE_NOTIFY_COMPLETED)
-            .setMimeType(download.contentType)
-            .addRequestHeader("User-Agent", download.userAgent)
 
-        if (cookie.isNotEmpty()) {
-            request.addRequestHeader("Cookie", cookie)
+        if (!download.contentType.isNullOrEmpty()) {
+            request.setMimeType(download.contentType)
         }
 
-        if (refererURL.isNotEmpty()) {
-            request.addRequestHeader("Referer", refererURL)
+        with(request) {
+            addRequestHeaderSafely("User-Agent", download.userAgent)
+            addRequestHeaderSafely("Cookie", cookie)
+            addRequestHeaderSafely("Referer", refererURL)
         }
 
         request.setDestinationInExternalPublicDir(download.destinationDirectory, fileName)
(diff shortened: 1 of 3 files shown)


components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadManager.kt
```diff
@@ -159,3 +159,12 @@ class DownloadManager(
...
 }
+
+internal fun Request.addRequestHeaderSafely(name: String, value: String?): Boolean {
+    return if (value.isNullOrEmpty()) {
+        false
+    } else {
+        addRequestHeader(name, value)
+        true
+    }
+}
```
pocmo: This returns a `Boolean` even though that is not used?

components/feature/downloads/src/test/java/mozilla/components/feature/downloads/DownloadManagerTest.kt
```diff
@@ -99,6 +102,25 @@ class DownloadManagerTest {
...
+        wasAdded = request.addRequestHeaderSafely("User-Agent", null)
+
+        assertFalse(wasAdded)
+
+        val fireFox = "Mozilla/5.0 (Windows NT 5.1; rv:7.0.1) Gecko/20100101 Firefox/7.0.1"
+
+        wasAdded = request.addRequestHeaderSafely("User-Agent", fireFox)
+
+        assertTrue(wasAdded)
+    }
```
pocmo: Instead of using the `Boolean` which seems to only exist for the test. You could call this on a mock request and validate that `addRequestHeader()` was called or not.

docs/changelog.md
```diff
@@ -12,6 +12,9 @@ permalink: /changelog/
 * [Gecko](https://github.com/mozilla-mobile/android-components/blob/master/buildSrc/src/main/java/Gecko.kt)
 * [Configuration](https://github.com/mozilla-mobile/android-components/blob/master/buildSrc/src/main/java/Config.kt)
 
+* **feature-downloads**
+  * Fixing bug #2265. The download notification shows "Unsuccessful download".
```
pocmo: I think this could be a bit more descriptive: What was the issue that was fixed?

Amejia481: Is this better ?
* **feature-downloads**
  * Fixing bug #2265. In some occasions, when trying to download a file, the download failed and the download notification shows "Unsuccessful download".

pocmo: yep 👍

Amejia481: :)

PR #2270, opened 2019-03-06
service-glean: Use HTTP client implementing concept-fetch for upload.

pocmo:
@Dexterp37 This is a PR that modifies `glean` to make use of a HTTP client that implements the interface of `concept-fetch`.

This allows us to unify the HTTP client that an app uses:
* In apps with GeckoView we would want all requests to go through Gecko/Necko (and show up in dev tools, yay) (`browser-engine-gecko-*`)
* In apps that are concerned about the app size we want to go with `HttpUrlConnection` (`lib-fetch-httpurlconnection`)
* And finally in apps that already use `OkHttp` we want to do that for telemetry too (`lib-fetch-okttp`).

Since glean tries to be a "no configuration needed" kind of library I've set the default client to `HttpURLConnectionClient` which means no app will **need** to configure this. I picked `HttpURLConnectionClient` since that doesn't require any third-party dependencies and because it's "only" a wrapper around `HttpURLConnection` it is super small.

What do you think?

### Pull Request checklist

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
@@ -4,6 +4,8 @@
 
 package mozilla.components.service.glean.config
 
+import mozilla.components.concept.fetch.Client
+import mozilla.components.lib.fetch.httpurlconnection.HttpURLConnectionClient
 import mozilla.components.service.glean.BuildConfig
 
 /**
@@ -17,12 +19,14 @@ import mozilla.components.service.glean.BuildConfig
  *           the [serverEndpoint]
  * @property maxEvents the number of events to store before the events ping is sent
  * @property logPings whether to log ping contents to the console.
+ * @property httpClient The HTTP client implementation to use for uploading pings.
  */
 data class Configuration(
     val serverEndpoint: String = "https://incoming.telemetry.mozilla.org",
     val userAgent: String = "Glean/${BuildConfig.LIBRARY_VERSION} (Android)",
-    val connectionTimeout: Int = 10000,
-    val readTimeout: Int = 30000,
+    val connectionTimeout: Long = 10000,
+    val readTimeout: Long = 30000,
     val maxEvents: Int = 500,
-    val logPings: Boolean = false
+    val logPings: Boolean = false,
+    val httpClient: Client = HttpURLConnectionClient()
 )
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/net/HttpPingUploader.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/net/HttpPingUploader.kt
@@ -5,17 +5,19 @@
 package mozilla.components.service.glean.net
 
 import android.support.annotation.VisibleForTesting
+import android.support.annotation.VisibleForTesting.PRIVATE
+import mozilla.components.concept.fetch.Client
+import mozilla.components.concept.fetch.MutableHeaders
+import mozilla.components.concept.fetch.Request
+import mozilla.components.concept.fetch.clientError
+import mozilla.components.concept.fetch.success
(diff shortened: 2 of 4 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/net/HttpPingUploader.kt
```diff
@@ -62,46 +62,56 @@ internal class HttpPingUploader : PingUploader {
...
+        return try {
+            performUpload(config.client, request)
+        } catch (e: IOException) {
+            logger.warn("IOException while uploading ping", e)
+            false
+        }
+    }
 
+    @VisibleForTesting(otherwise = PRIVATE)
+    internal fun buildRequest(path: String, data: String, config: Configuration) = Request(
```
Dexterp37: Oh, wow, I didn't know we could use both `internal` and `VisibleFortesting` one one function. Has this any advantage over using only one or the other?

components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
```diff
@@ -17,12 +19,14 @@ import mozilla.components.service.glean.BuildConfig
...
     val serverEndpoint: String = "https://incoming.telemetry.mozilla.org",
     val userAgent: String = "Glean/${BuildConfig.LIBRARY_VERSION} (Android)",
-    val connectionTimeout: Int = 10000,
-    val readTimeout: Int = 30000,
+    val connectionTimeout: Long = 10000,
+    val readTimeout: Long = 30000,
     val maxEvents: Int = 500,
-    val logPings: Boolean = false
+    val logPings: Boolean = false,
+    val client: Client = HttpURLConnectionClient()
```
Dexterp37: nit: can we call this `httpClient`?

components/service/glean/src/main/java/mozilla/components/service/glean/net/HttpPingUploader.kt
travis79: I'm not sure but I think that the 'internal' would override the annotation, from my understanding of how the @VisibleForTesting annotation works.  The `otherwise` should probably be `INTERNAL` or omitted, since it defaults to `PRIVATE`.

pocmo: `internal` makes this accessible from everywhere in this gradle module. Marking it as `VisibleForTesting(otherwise = PRIVATE)` let's the IDE know that it should treat this (almost) as private. So if I access this somewhere (outside of tests) where this is not considered "private" the IDE will show a warning that I may be doing something wrong:

screenshot 2019-03-06 at 18 28 41

components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
pocmo: Okay, updated. :)

components/service/glean/src/main/java/mozilla/components/service/glean/net/HttpPingUploader.kt
travis79: Ahh, so the annotation just gets enforced by the IDE and not the compiler...

Dexterp37: Great! I'll file a bug for having that in our testing API as well :)

MihaiTabara:
FTR, this is failing because of the changes that landed in bug 1519493. Looking to fix them asap, sorry about the bustage.

MihaiTabara:
Pushed a temp fix in https://github.com/mozilla-mobile/android-components/pull/2282 to unblock this. The real fix is in bug 1533314 which will be tackled in the next days/weeks. The graphs should be unblocked now. Sorry for the bustage!

Issue #2243, opened 2019-03-04
Add engine API for WebExtension messaging


csadilek:
GV is working on API to exchange messages between WebExtensions and the application. We need to expose this API as part of our engine/session.

https://bugzilla.mozilla.org/show_bug.cgi?id=1518843

csadilek:
This is needed to communicate (to the app) that reader mode is available for the current page and that it should be activated/deactivated (to the extension).

We also need it for choosing the font/size.

PR #2273, opened 2019-03-06
Closes #1346: Add ReaderView feature component.

### LINKED ISSUE #1346: Add reader view feature component
Reference Browser User Story: https://github.com/mozilla-mobile/reference-browser/issues/152

For now we only need this for GeckoView. It's unclear whether GV would provide this functionality or whether we would need to do that ourselves. But even if we'd readermodify the current page we'd need some kind of API to mess with the page (inject JS).

csadilek:
**DRAFT:** This brings in the Reader View feature component, implemented as a WebExtension. We need messaging support to finish this (see https://github.com/mozilla-mobile/android-components/issues/2243 and https://bugzilla.mozilla.org/show_bug.cgi?id=1518843), but for now this automatically activates reader view for `blog.mozilla.org`. 

Screenshots from our sample browser:

readerview-demo

Note that readability.js and readability-readerable.js are taken from: https://github.com/mozilla/readability

### DIFF
--- a/components/feature/readerview/src/main/assets/extensions/readerview/moment/moment-2.24.0.js
+++ b/components/feature/readerview/src/main/assets/extensions/readerview/moment/moment-2.24.0.js
@@ -1,0 +1,14765 @@
+;(function (global, factory) {
+    typeof exports === 'object' && typeof module !== 'undefined' ? module.exports = factory() :
+    typeof define === 'function' && define.amd ? define(factory) :
+    global.moment = factory()
+}(this, (function () { 'use strict';
+
+    var hookCallback;
+
+    function hooks () {
+        return hookCallback.apply(null, arguments);
+    }
+
+    // This is done to register the method called with moment()
+    // without creating circular dependencies.
+    function setHookCallback (callback) {
+        hookCallback = callback;
+    }
+
+    function isArray(input) {
+        return input instanceof Array || Object.prototype.toString.call(input) === '[object Array]';
+    }
+
+    function isObject(input) {
+        // IE8 will treat undefined and null as object if it wasn't for
+        // input != null
+        return input != null && Object.prototype.toString.call(input) === '[object Object]';
+    }
+
+    function isObjectEmpty(obj) {
+        if (Object.getOwnPropertyNames) {
+            return (Object.getOwnPropertyNames(obj).length === 0);
+        } else {
+            var k;
+            for (k in obj) {
+                if (obj.hasOwnProperty(k)) {
+                    return false;
+                }
+            }
+            return true;
+        }
+    }
+
+    function isUndefined(input) {
+        return input === void 0;
+    }
+
+    function isNumber(input) {
+        return typeof input === 'number' || Object.prototype.toString.call(input) === '[object Number]';
+    }
+
+    function isDate(input) {
+        return input instanceof Date || Object.prototype.toString.call(input) === '[object Date]';
+    }
+
+    function map(arr, fn) {
+        var res = [], i;
+        for (i = 0; i < arr.length; ++i) {
+            res.push(fn(arr[i], i));
+        }
+        return res;
+    }
+
+    function hasOwnProp(a, b) {
+        return Object.prototype.hasOwnProperty.call(a, b);
+    }
+
+    function extend(a, b) {
+        for (var i in b) {
+            if (hasOwnProp(b, i)) {
+                a[i] = b[i];
+            }
+        }
+
+        if (hasOwnProp(b, 'toString')) {
+            a.toString = b.toString;
+        }
+
+        if (hasOwnProp(b, 'valueOf')) {
+            a.valueOf = b.valueOf;
+        }
+
+        return a;
+    }
+
+    function createUTC (input, format, locale, strict) {
+        return createLocalOrUTC(input, format, locale, strict, true).utc();
+    }
+
+    function defaultParsingFlags() {
+        // We need to deep clone this object.
+        return {
+            empty           : false,
+            unusedTokens    : [],
+            unusedInput     : [],
+            overflow        : -2,
+            charsLeftOver   : 0,
+            nullInput       : false,
+            invalidMonth    : null,
+            invalidFormat   : false,
+            userInvalidated : false,
+            iso             : false,
+            parsedDateParts : [],
+            meridiem        : null,
+            rfc2822         : false,
+            weekdayMismatch : false
+        };
+    }
+
+    function getParsingFlags(m) {
+        if (m._pf == null) {
+            m._pf = defaultParsingFlags();
+        }
+        return m._pf;
+    }
+
+    var some;
+    if (Array.prototype.some) {
+        some = Array.prototype.some;
+    } else {
+        some = function (fun) {
+            var t = Object(this);
+            var len = t.length >>> 0;
+
+            for (var i = 0; i < len; i++) {
+                if (i in t && fun.call(this, t[i], i, t)) {
+                    return true;
+                }
+            }
+
+            return false;
+        };
+    }
+
+    function isValid(m) {
+        if (m._isValid == null) {
+            var flags = getParsingFlags(m);
+            var parsedParts = some.call(flags.parsedDateParts, function (i) {
+                return i != null;
+            });
+            var isNowValid = !isNaN(m._d.getTime()) &&
+                flags.overflow < 0 &&
+                !flags.empty &&
+                !flags.invalidMonth &&
+                !flags.invalidWeekday &&
+                !flags.weekdayMismatch &&
+                !flags.nullInput &&
+                !flags.invalidFormat &&
+                !flags.userInvalidated &&
+                (!flags.meridiem || (flags.meridiem && parsedParts));
+
+            if (m._strict) {
+                isNowValid = isNowValid &&
+                    flags.charsLeftOver === 0 &&
+                    flags.unusedTokens.length === 0 &&
+                    flags.bigHour === undefined;
+            }
+
+            if (Object.isFrozen == null || !Object.isFrozen(m)) {
+                m._isValid = isNowValid;
+            }
+            else {
+                return isNowValid;
+            }
+        }
+        return m._isValid;
+    }
+
+    function createInvalid (flags) {
+        var m = createUTC(NaN);
+        if (flags != null) {
+            extend(getParsingFlags(m), flags);
+        }
+        else {
+            getParsingFlags(m).userInvalidated = true;
+        }
+
+        return m;
+    }
+
+    // Plugins that add properties should also add the key here (null value),
+    // so we can properly clone ourselves.
+    var momentProperties = hooks.momentProperties = [];
+
+    function copyConfig(to, from) {
+        var i, prop, val;
+
+        if (!isUndefined(from._isAMomentObject)) {
+            to._isAMomentObject = from._isAMomentObject;
+        }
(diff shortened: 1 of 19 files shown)


eeejay:
So awesome! Really excited to see this land.

This might have been discussed elsewhere, but we already bundle Firefox's reader mode in GeckoView (mostly by mistake), and their is an official [WebExtensions API](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/toggleReaderMode) for it.

So ideally, this bundled extension could be made much simpler and not duplicate readability and all the functionality. And bugs.

The GeckoView WebExtensions support does not support the `tabs` namespace yet, so that's a blocker.

cc @agi90

csadilek:
> This might have been discussed elsewhere, but we already bundle Firefox's reader mode in GeckoView (mostly by mistake), and their is an official WebExtensions API for it.

@eeejay ah interesting! We did decide to go down this path and use the generic messaging API currently developed in GV, as we want tighter integration with the app e.g. we want to be able to have the font controls in the App UI (not part of content): https://github.com/mozilla-mobile/fenix/issues/190 (see mockups).

Theoretically, you / we can then remove this from GeckoView. Of course, we could re-think, but then we'd have to add reader-view-specific APIs and make it all customizable in GV. Readability is mostly brought in as a third-party lib here, but yes there is some duplication...

agi90:
Opened https://bugzilla.mozilla.org/show_bug.cgi?id=1533156 to remove the code from GeckoView. Thank you @eeejay & @csadilek

components/feature/readerview/src/main/java/mozilla/components/feature/readerview/ReaderViewFeature.kt
```diff
@@ -0,0 +1,46 @@
...
+import mozilla.components.browser.session.SessionManager
+import mozilla.components.concept.engine.Engine
+import mozilla.components.concept.engine.webextension.WebExtension
+import mozilla.components.concept.toolbar.Toolbar
+import mozilla.components.support.base.feature.LifecycleAwareFeature
+import mozilla.components.support.base.log.logger.Logger
+
+class ReaderViewFeature(
+    private val engine: Engine,
+    private val toolbar: Toolbar,
```
pocmo: In the latest Fenix mocks "Reader View" is not a page action anymore but instead in this little drawer above the URL bar. So I assume we will need some kind of callback here so that the app can do something custom. I was actually wondering whether we should just expose this as a field on `Session`? This way we could have two feature classes, one installs the extension, updates the Session etc. And one the deals with the page action (which Fenix wouldn't need/use though!).

components/feature/readerview/src/main/java/mozilla/components/feature/readerview/ReaderViewFeature.kt
```diff
@@ -0,0 +1,46 @@
...
+            engine.installWebExtension(readerViewWebExt,
+                onSuccess = {
+                    Logger.debug("Installed browser-icons extension")
+                    installed = true
+                },
+                onError = { ext, throwable ->
+                    Logger.error("Failed to install extension ${ext.id}", throwable)
+                }
+            )
+        }
```
pocmo: Since this feature has activity/fragment scope and the engine "global" scope: What happens if this feature gets initialized multiple times (e.g. activity re-creation) and we install the extension again? Maybe that's another reason for separate features - a global one (that sets everything up) and a "local" one for messing with the UI?

components/feature/readerview/src/main/java/mozilla/components/feature/readerview/ReaderViewFeature.kt
```diff
@@ -0,0 +1,46 @@
...
+import mozilla.components.concept.engine.webextension.WebExtension
+import mozilla.components.concept.toolbar.Toolbar
+import mozilla.components.support.base.feature.LifecycleAwareFeature
+import mozilla.components.support.base.log.logger.Logger
+
+class ReaderViewFeature(
+    private val engine: Engine,
+    private val toolbar: Toolbar,
+    sessionManager: SessionManager
+) : SelectionAwareSessionObserver(sessionManager), LifecycleAwareFeature {
```
pocmo: nit: Extends `SelectionAwareSessionObserver` but doesn't use that yet? :)

samples/browser/src/main/java/org/mozilla/samples/browser/BrowserFragment.kt
```diff
@@ -145,11 +146,14 @@ class BrowserFragment : Fragment(), BackHandler {
             owner = this,
             view = layout)
 
+        val readerViewFeature = ReaderViewFeature(components.engine, layout.toolbar, components.sessionManager)
```
pocmo: nit: Since this keeps a strong reference to the View which may have a shorter lifetime than the fragment, we may want to wrap this in `ViewBoundFeatureWrapper`.

components/feature/readerview/src/main/java/mozilla/components/feature/readerview/ReaderViewFeature.kt
agi90: yeah the web extension life cycle is tied to the runtime not the session.

components/feature/readerview/src/main/assets/extensions/readerview/moment/moment-2.24.0.js
```diff
@@ -0,0 +1,14765 @@
+;(function (global, factory) {
```
Amejia481: We can use the [minified version](https://momentjs.com/downloads/moment.min.js) to save some bits or minify it our self

components/feature/readerview/src/main/assets/extensions/readerview/readability/readability-0.2.0.js
```diff
@@ -0,0 +1,1800 @@
+/*eslint-env es6:false*/
```
Amejia481: This file could be minified too

components/feature/readerview/src/main/java/mozilla/components/feature/readerview/ReaderViewFeature.kt
csadilek: True, there's already a check that makes sure we only install once, but we can split things up when we bring it all the pieces together.

components/feature/readerview/src/main/java/mozilla/components/feature/readerview/ReaderViewFeature.kt
csadilek: Yes, for now it only registers the observer for the selected session, but there is nothing to observe yet :)

samples/browser/src/main/java/org/mozilla/samples/browser/BrowserFragment.kt
csadilek: Ah good point. Fixed!

components/feature/readerview/src/main/assets/extensions/readerview/moment/moment-2.24.0.js
csadilek: @Amejia481 yes, good idea! For now I want to be able to read / debug the code though. We can land the minified version or swap out in release builds.

components/feature/readerview/src/main/java/mozilla/components/feature/readerview/ReaderViewFeature.kt
csadilek: Good point, this wasn't thought out yet on my part. I've added a TODO to change to a callback (instead of passing in the toolbar), then the app can decide how to handle it. The idea of 2 feature classes sounds good as well, but I would like to try it out once we have messaging. Let's  chat more!

components/feature/readerview/src/main/assets/extensions/readerview/moment/moment-2.24.0.js
csadilek: Filed to investigate how we want to do this for release builds: https://github.com/mozilla-mobile/android-components/issues/2288

components/feature/readerview/src/main/assets/extensions/readerview/readability/readability-0.2.0.js
csadilek: https://github.com/mozilla-mobile/android-components/issues/2288

components/feature/readerview/src/main/assets/extensions/readerview/moment/moment-2.24.0.js
Amejia481: 👍

csadilek:
Now added JavaScript API to:
- set font size and adjust (increase/decrease) in steps (for the plus/minus buttons)
- set font type
- set color scheme (light|sepia|dark)

csadilek:
@pocmo I've updated this to sketch out the feature as discussed, and added TODOs where needed. 

I've also filed follow-ups:
- Engine API for messaging: https://github.com/mozilla-mobile/android-components/issues/2243
- Appearance UI: https://github.com/mozilla-mobile/android-components/issues/2623
- Wiring up messaging between Reader View feature and extension: https://github.com/mozilla-mobile/android-components/issues/2624

This is getting pretty big so I'd like to land it and work on the follow-ups :).

components/feature/readerview/src/main/java/mozilla/components/feature/readerview/ReaderViewFeature.kt
```diff
@@ -0,0 +1,141 @@
...
+        /**
+         * Installs the readerview web extension in the provided engine.
+         *
+         * @param engine a reference to the application's browser engine.
+         */
+        fun install(engine: Engine) {
+            engine.installWebExtension(readerViewWebExt,
+                onSuccess = {
+                    Logger.debug("Installed extension: ${it.id}")
+                    installed = true
```
pocmo: Since this callback is asynchronous, it is still technically possible to call `installWebExtension` twice here..?

components/feature/readerview/src/main/assets/extensions/readerview/readerview.js
```diff
@@ -0,0 +1,247 @@
+
```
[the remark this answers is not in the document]
csadilek: Thanks! Added.

components/feature/readerview/src/main/java/mozilla/components/feature/readerview/ReaderViewFeature.kt
```diff
@@ -0,0 +1,141 @@
...
+ * session.
+ */
+class ReaderViewFeature(
+    private val context: Context,
+    private val engine: Engine,
+    private val sessionManager: SessionManager,
+    private val onReaderViewAvailableChange: OnReaderViewAvailableChange = { }
+) : SelectionAwareSessionObserver(sessionManager), LifecycleAwareFeature, BackHandler {
+
+    private val config = Config(PreferenceManager.getDefaultSharedPreferences(context))
```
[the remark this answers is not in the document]
csadilek: OK. Done!

components/feature/readerview/src/main/java/mozilla/components/feature/readerview/ReaderViewFeature.kt
csadilek: OK, I've fixed this for now, but I think I will try to move the installation state into our `WebExtension` abstraction when we add more more functionality to it (as part of #2243)

csadilek:
Addressed comment and added README.

PR #2276, opened 2019-03-06
Closes #2232: CrashReporter.instance writes not visible across threads.

### LINKED ISSUE #2232: CrashReporter.instance is not thread-safe
We require that apps call `install` in `Application.onCreate`: https://github.com/mozilla-mobile/android-components/blob/master/components/lib/crash/src/main/java/mozilla/components/lib/crash/CrashReporter.kt#L172

The processing of crashes happens on a different thread: https://github.com/mozilla-mobile/android-components/blob/master/components/lib/crash/src/main/java/mozilla/components/lib/crash/handler/CrashHandlerService.kt#L17

We don't guarantee the visibility of `CrashReporter.instance` across threads.

csadilek:
Haven't been able to write meaningful test for this yet. Looking into it.


csadilek:
Added a test similar to the other ones we have for volatile fields. I don't think there's a reliable way to create a test failure for the non-visible write if the annotation is removed, but this way we'd at least fail and know why :).

PR #2278, opened 2019-03-06
Closes #2277 - Add ability to set default search engine

### LINKED ISSUE #2277: Add ability to set default search engine

ekager:
I think since the calls to SearchEngineManager `getDefaultSearchEngine` are tested in `SearchEngineManagerTest`, this PR does need its own tests? 

### Pull Request checklist

### DIFF
--- a/components/browser/search/src/main/java/mozilla/components/browser/search/SearchEngineManager.kt
+++ b/components/browser/search/src/main/java/mozilla/components/browser/search/SearchEngineManager.kt
@@ -8,10 +8,10 @@ import android.content.BroadcastReceiver
 import android.content.Context
 import android.content.Intent
 import android.content.IntentFilter
-import kotlinx.coroutines.async
 import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Deferred
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.async
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
@@ -28,6 +28,7 @@ class SearchEngineManager(
 ) {
     private var deferredSearchEngines: Deferred<List<SearchEngine>>? = null
     private val scope = CoroutineScope(Dispatchers.Default)
+    var defaultSearchEngine: SearchEngine? = null
 
     /**
      * Asynchronously load search engines from providers. Inherits caller's [CoroutineScope.coroutineContext].
@@ -55,8 +56,8 @@ class SearchEngineManager(
     /**
      * Returns the default search engine.
      *
-     * The default engine is the first engine loaded by the first provider passed to the
-     * constructor of SearchEngineManager.
+     * If defaultSearchEngine has not been set, the default engine is the first engine loaded by the
+     * first provider passed to the constructor of SearchEngineManager.
      *
      * Optionally a name can be passed to this method (e.g. from the user's preferences). If
      * a matching search engine was loaded then this search engine will be returned instead.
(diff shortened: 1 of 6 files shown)


ekager:
I'm working on setting a default search engine to actually be used for searches in Fenix, and @sblatz is working on search suggestions which will need one-off searches outside of the default engine. Creating a method to set the default makes sense, but we'll also need a way for the overrides, this was aiming to do both but I can see how passing in the default every time will get annoying.

pocmo:
Ah, okay, that makes sense.

How does this sound:
* For the default we add this to the SearchEngineManager - so that everyone can use the default without needing to know the name in advance.
* For the use case we allow passing in an optional`SearchEngine` object. If provided the use case will use this instead of `getDefaultSearchEngine`.

ekager:
Yeah, that sounds good! I'll work on making those changes

ekager:
So I talked to @sblatz more about this PR, and he says because he'll be able to add and remove providers, we don't think we will need to have the ability for a one-off search with a specific search engine (at least for now!). I've removed the `SearchUseCases` code, and replaced it with just a way to set the default search engine in `SearchEngineManager` (with tests!)! Let me know if this is the right way to do this :)

components/browser/search/src/main/java/mozilla/components/browser/search/SearchEngineManager.kt
```diff
@@ -28,6 +28,7 @@ class SearchEngineManager(
 ) {
     private var deferredSearchEngines: Deferred<List<SearchEngine>>? = null
     private val scope = CoroutineScope(Dispatchers.Default)
+    var defaultSearchEngine: SearchEngine? = null
```
pocmo: I think in `SearchEngineManager` it may be good to still use the name and only have a setter for it (e.g. `setDefault(name: String)`) Otherwise it may be confusing if you look at it from the outside and you have two way to get to a `SearchEngine` - calling `getDefaultSearchEngine` but also `defaultSearchEngine`. But the second one only returns something you have set on the manager manually and not really the default in all cases. :)

components/browser/search/src/main/java/mozilla/components/browser/search/SearchEngineManager.kt
```diff
@@ -66,11 +67,21 @@ class SearchEngineManager(
...
 
+    /**
+     * Returns search engine corresponding to name passed to this method, or null if not found.
+     */
+    @Synchronized
+    fun getSearchEngineByName(context: Context, name: String): SearchEngine? {
+        val searchEngines = getSearchEngines(context)
+
+        return searchEngines.find { it.name == name }
+    }
```
pocmo: Will this be needed after we change the code to allow setting the name directly?

components/feature/search/src/main/java/mozilla/components/feature/search/SearchUseCases.kt
```diff
@@ -37,8 +38,8 @@ class SearchUseCases(
         /**
          * Triggers a search in the currently selected session.
          */
-        override fun invoke(searchTerms: String) {
-            invoke(searchTerms, sessionManager.selectedSession)
+        override fun invoke(searchTerms: String, searchEngine: SearchEngine?) {
+            invoke(searchTerms, sessionManager.selectedSession, searchEngine)
         }
```
pocmo: I wonder if we should add the `searchEngine` parameter to the constructor instead of the `invoke` method: Other components will call `invoke` and not provide the parameter (because they have no knowledge about the default). So instead the app needs to configure the use case to use the right search engine when something calls `invoke` on it.

components/browser/search/src/main/java/mozilla/components/browser/search/SearchEngineManager.kt
ekager: I think this will still be needed in order for @sblatz to pass a non-default search engine to the invoke function?

ekager:
Okay, I created the setDefaultSearchEngine function that takes the engineName as a String, and removed the getEngineByName function. Now TaskCluster is complaining because IntentProcessor uses invoke but does not know about SearchEngine...? If I import it, it's an unused import, do you know how to fix this?

/build/android-components/components/feature/intent/src/main/java/mozilla/components/feature/intent/IntentProcessor.kt: (82, 49): Cannot access class 'mozilla.components.browser.search.SearchEngine'. Check your module classpath for missing or conflicting dependencies

PR #2264, opened 2019-03-06
Issue #651: Add PocketEndpointRaw

mcomella:
This is intended to lay the foundations for accessing the Pocket API: see extended commit summaries for details.

### Pull Request checklist

- [x] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [x] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/android-components/blob/master/docs/changelog.md) or does not need one
- [x] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features

Issue #651, opened 2018-08-23
Add Pocket endpoint component


mcomella:
On Fire TV, we access the Pocket APIs to get a trending video content list: we could componentize this access, as well as any other Pocket APIs other teams use.

We could also consider:
- Wrapping this in a caching layer
- Fetching related resources (e.g. thumbnails from the thumbnail url param)

Benefits:
- Only have to write this code once
- Easy to use one Pocket API if you're using another
- Teams don't need to deal with a REST API: they make a function call and get Kotlin objects back
- If the Pocket API version gets updated, everyone can update for free(ish)

pocmo:
Following the pattern of other components `services-pocket` would be a good home for that. I'll ping the Pocket Android team to see what they think. :)

mcomella:
We have an unsolved problem: **the Pocket API versions each URL** via GET params (e.g. `.../video-recs?version=4` and `.../article-recs?version=2`). Because these APIs are internal to Mozilla, they are liable to change. If the APIs change between versions, we'll likely need to update this module too. However, in a native solution, the updates to this module **are likely breaking changes.**

@pocmo To save time, you may wish to jump to my opinion on the solution before reading through all of the options.

## Background context
- It's unclear how long the Pocket server team will support each API version

## Existing solutions
When REST APIs introduce breaking changes, the major version number is bumped and the client libraries (e.g. for the JVM) will release an entirely new library for the new version (e.g. Yelp API). This is infrequent so is not a problem. This solution doesn't really apply to us.

## Potential solutions
#### 1) Naive solution: do not handle multiple versions, i.e. introduce breaking changes in a-c when changing versions.
- Only known Pocket consumer atm is APT Amazon team so easier to coordinate upgrades

Pros:
- Least work from a-c developer perspective

Cons:
- More potential for a-c breaking changes, forcing work on consumers
- Worst case: the Pocket API is broken for some consumers in a new version so they want to stay on an old a-c. However, old a-c has bugs. They are stuck!
  - Potential solution: temporarily copy entire `:service-pocket` module into a new a-c module until the bug is fixed in the Pocket API

#### 2) Provide a method (and returned data type) for each API URL
e.g. `getGlobalVideoRecommendationsV2(): GlobalVideoRecommendationV2`

Pros:
- All API versions remain available (assuming the server supports them)
- No breaking changes

Cons:
- A lot of redundant code, types, methods, and tests and a lot of work to share code with private functions

[...]

PR #2279, opened 2019-03-07
Issue #651: Add PocketJSONParser

mcomella:
Only review 3 bottom-most commits: the previous commits are part of #2264.

This PR parses the JSON received from the server and converts it into useful objects for the app. The plan is that a public facing medium-level API, `PocketEndpoint`, will call `PocketEndpointRaw` (pulls JSON from the server), put the JSON through the `PocketJSONParser`, and return the parsed objects to the application. I may also add a higher-level API (e.g. that handles periodic backoffs, returns a deferred, etc. - but this would be in a separate class).

One problem to solve (want to discuss during android meeting): how should I handle versioning of the API? Each URL in the API can have an independent version and I'm not sure how to handle this.

### Pull Request checklist


components/service/pocket/src/main/java/mozilla/components/service/pocket/data/PocketGlobalVideoRecommendation.kt
```diff
@@ -0,0 +1,29 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.pocket.data
+
+/**
+ * A recommended video as returned from the Pocket Global Video Recommendation endpoint v2.
+ */
+data class PocketGlobalVideoRecommendation(
```
pocmo: This looks like a class were we would want to document every `@property`? Does the endpoint have an API documentation where we could copy it from? At least a field like `publishedTimestamp: String` in itself doesn't say much.

components/service/pocket/src/main/java/mozilla/components/service/pocket/ext/JSON.kt
```diff
@@ -0,0 +1,28 @@
...
+import org.json.JSONObject
+
+/**
+ * Map over the JSONArray, removing null values after [transform] and ignoring data that throws a [JSONException].
+ * An empty List will be returned if all the values are null or invalid.
+ *
+ * We don't expose this function publicly because it's not generic enough yet: JSONObject should be T. We don't
+ * want to expose this temporarily because then we'll break the public API.
+ */
+internal inline fun <R : Any> JSONArray.mapObjNotNull(transform: (JSONObject) -> R?): List<R> {
```
pocmo: We have a bunch of other `org.json` extensions in `support-ktx`. Maybe we can move this one over too.

components/service/pocket/src/test/java/mozilla/components/service/pocket/ext/JSONKtTest.kt
```diff
@@ -0,0 +1,63 @@
...
+import org.json.JSONArray
+import org.json.JSONObject
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
+class JSONKtTest {
+
```
pocmo: The methods in this test mention `flatMapObj` - but they are testing `mapObjNotNull` :)

components/service/pocket/src/main/java/mozilla/components/service/pocket/ext/JSON.kt
mcomella: @pocmo I chose not to move this over because then it'd have to be a public android-components API. I didn't want it public because it's not generic enough: it should be 
```kotlin
inline fun <T, R : Any> JSONArray.mapNotNull(transform: (T) -> r?): List<R> {`
```

But I had difficulty making this generic enough when I initially wrote it.

mcomella: Actually, I guess I can write something like:
```kotlin
internal inline fun <T, R : Any> JSONArray.mapNotNull(extractFun: JSONArray.(Int) -> T, transform: (T) -> R?): List<R> {
    val transformedResults = mutableListOf<R>()
    for (i in 0 until this.length()) {
        try {
            val transformed = transform(this.extractFun(i))
            if (transformed != null) { transformedResults.add(transformed) }
        } catch (e: JSONException) { /* Do nothing: we skip bad data. */ }
    }

    return transformedResults
}

fun example() {
    JSONArray().mapNotNull(JSONArray::getJSONObject) {
        4
    }
}
```

I'll move it over to the JSON module.

components/service/pocket/src/main/java/mozilla/components/service/pocket/data/PocketGlobalVideoRecommendation.kt
mcomella: > This looks like a class were we would want to document every `@property`?

Good point: I forgot that I was creating a public API. :)

> Does the endpoint have an API documentation where we could copy it from? At least a field like `publishedTimestamp: String` in itself doesn't say much.

Unfortunately, no – these are internal APIs with minimal documentation. There are no details on what each field represents, beyond the occassional incidental comment, and just JSON dumps of examples of the API output.

mcomella: I don't actually know what `publishTimestamp` is, tbh. I didn't think it was worth asking Pocket about unless someone needed that info so I put in:
```kotlin
/**
 * @property publishedTimestamp unknown: please ask for clarification if needed. May be "0".
```

mcomella: I added kdoc to all of the other properties (and added kdoc to the `Authors` field too).

components/service/pocket/src/main/java/mozilla/components/service/pocket/ext/JSON.kt
mcomella: @pocmo Done: you may wish to re-review this because I made the code more generic, adding risk.

components/service/pocket/src/test/java/mozilla/components/service/pocket/ext/JSONKtTest.kt
mcomella: Oi! IDE refactors! Fixed when I moved the code to the new class.

mcomella:
I can't reproduce this TaskCluster test failure. @pocmo Any ideas? I just pushed so I wonder if this is from an old run.

mcomella:
Actually, that was 30min ago so maybe it's not an old run.

mcomella:
I'm able to reproduce with the command run by TC: `./gradlew --no-daemon clean -Pcoverage :service-pocket:assemble :service-pocket:test :service-pocket:lintRelease`

mcomella:
```
java.lang.IllegalAccessError: tried to access method mozilla.components.support.ktx.android.org.json.JSONArrayKt.$jacocoInit()[Z from class mozilla.components.service.pocket.PocketJSONParser
	at mozilla.components.service.pocket.PocketJSONParser.jsonToGlobalVideoRecommendations(PocketJSONParser.kt:26)
	at mozilla.components.service.pocket.PocketJSONParserTest.WHEN parsing global video recommendations with missing fields on all items THEN null is returned(PocketJSONParserTest.kt:164)
```

It looks like this is a jacoco problem.

mcomella:
All these tests fail because my code calls into a method `jacocoInit` on `JSONArrayKt`: my initial guess is that jacoco is adding internal/protected methods on each module and, by calling into the other module, we're violating those visibility modifiers.

components/service/pocket/src/main/java/mozilla/components/service/pocket/ext/JSON.kt
mcomella: btw, another way to define this:
```kotlin
inline fun <T, R : Any> JSONArray.mapNotNull(transform: (T) -> R?): List<R> {
```

And the call site:
```kotlin
JSONArray().mapNotNull<JSONObject, Int> { it.getInt("playCount") }
```

It's theoretically less performant though because the type conversion is all casting.

mcomella:
This issue is caused because the google android gradle plugin instruments methods before they are inlined, which breaks jacoco when testing across module boundaries: https://issuetracker.google.com/issues/109771903 (and jacoco https://github.com/jacoco/jacoco/issues/688 which was closed in favor of the google issue). This is fixed in 3.4.0 (current is 3.3.0, released in January).

The function is inlined for performance (like Collection.map) so the solution is to temporarily uninline it and wait for the version bump.

mcomella:
Filed https://github.com/mozilla-mobile/android-components/issues/2305

mcomella:
@pocmo This should be ready to land.

PR #2280, opened 2019-03-07
Test github checks api.

tomprince:
Experimenting.


pocmo:
Ooooh, that looks interesting :)

PR #2304, opened 2019-03-08
Closes #2255: Adds onBackPressListener to editToolbar

### LINKED ISSUE #2255: Implement better handling of back press when keyboard is showing on search bar
Per recommendation [here](https://github.com/mozilla-mobile/fenix/issues/847), it'd be great if our "hide keyboard" button also called `onBackPressed`.

It seems like we can achieve this by [listening to `KeyEvent.BACK`](https://stackoverflow.com/questions/8080711/android-back-key-event-while-displaying-soft-keyboard)

sblatz:
### Pull Request checklist

### DIFF
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
@@ -468,6 +468,12 @@ class BrowserToolbar @JvmOverloads constructor(
         }
     }
 
+    internal fun onEditCancelled() {
+        if (editToolbar.editListener?.onCancelEditing() != false) {
+            displayMode()
+        }
+    }
+
     private fun updateState(state: State) {
         this.state = state
 
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/edit/EditToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/edit/EditToolbar.kt
@@ -9,6 +9,7 @@ import android.content.Context
 import android.support.v4.content.ContextCompat
 import android.text.InputType
 import android.view.Gravity
+import android.view.KeyEvent
 import android.view.View
 import android.view.ViewGroup
 import android.view.inputmethod.EditorInfo
@@ -60,6 +61,13 @@ class EditToolbar(
         setOnTextChangeListener { text, _ ->
             updateClearViewVisibility(text)
             editListener?.onTextChanged(text)
+        }
+
+        setOnDispatchKeyEventPreImeListener { event ->
+            if (event?.keyCode == KeyEvent.KEYCODE_BACK) {
+                toolbar.onEditCancelled()
+            }
+            false
         }
     }
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -18,6 +18,11 @@ permalink: /changelog/
 * **browser-menu**
    * Added `TwoStateButton` in `BrowserMenuItemToolbar` that will change resources based on the `isInPrimaryState` lambda and added ability to disable the button with optional `disableInSecondaryState` argument.
 
+* **browser-toolbar**
+  * Adds `onCancelEditing` to `onEditListener` in `BrowserToolbar` which is fired when a back button press occurs while the keyboard is displayed.
+    This is especially useful if you want to call `activity.onBackPressed()` to navigate away rather than just dismiss the keyboard.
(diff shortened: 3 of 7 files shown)


components/browser/toolbar/src/test/java/mozilla/components/browser/toolbar/edit/EditToolbarTest.kt
```diff
@@ -84,6 +84,29 @@ class EditToolbarTest {
...
+        val listener = { listenerInvoked = true }
+
+        val toolbar = BrowserToolbar(context)
+        toolbar.editToolbar.urlView.onAttachedToWindow()
+        toolbar.setOnBackPressListener(listener)
+        toolbar.editToolbar.urlView.setText("https://www.mozilla.org")
+
+        assertEquals(listener, toolbar.editToolbar.onBackPressedListener)
+
+        // TODO this crashes as the shadow library does not have KeyEvent.KEYCODE_BACK
```
sblatz: The Shadow version of KeyEvents does not seem to support KEYCODE_BACK. Is there any way around this?

image

sblatz:
Made the changes requested and ready for review!

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/edit/EditToolbar.kt
```diff
@@ -61,6 +62,13 @@ class EditToolbar(
             updateClearViewVisibility(text)
             editListener?.onTextChanged(text)
         }
+
+        setOnDispatchKeyEventPreImeListener { event ->
+            if (event?.keyCode == KeyEvent.KEYCODE_BACK) {
+                editListener?.onCancelEditing()
```
pocmo: This calls the `editListener` instance directly and `onEditCancelled` in `BrowserToolar` never gets called?

sblatz: Fixed, good catch! My previous test I had written would have caught this, but unfortunately there's no way for us to simulate a `KEYCODE_BACK` press 😕

sblatz:
I believe Taskcluster is failing due to something outside of my control. The error it's displaying is [found here](https://taskcluster-artifacts.net/B_N7NHvTQ1-fjdgFRk5RyQ/0/public/logs/live_backing.log)

image

@pocmo any thoughts?

pocmo:
Looks like it passed now? :)

Amejia481:
@sblatz Looks like an network error, they tend to happen from time to time. In case that happens again you could re-trigger the build by updating the title of the PR, you can add a dot "." add the end of the title.

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
```diff
@@ -468,6 +468,12 @@ class BrowserToolbar @JvmOverloads constructor(
         }
     }
 
+    internal fun onEditCancelled() {
+        if (editToolbar.editListener?.onCancelEditing() == true) {
+            displayMode()
+        }
+    }
```
pocmo: Okay, really the last nit... ;)

Should this be:

`if (editToolbar.editListener?.onCancelEditing() != false) {`

Reasoning: We want to switch to display mode if the listener returns `true` OR if there's no listener. I guess now that would only happen if there's a listener and it returns `true`?

sblatz: Fixed :) I was wondering why the one above was doing `!= false` as that just seemed more confusing for no reason. Thanks for the clarification!

docs/changelog.md
```diff
@@ -36,6 +36,11 @@ permalink: /changelog/
 * **concept-fetch**
   * ⚠️ **This is a breaking API change**: the [`Response`](https://mozac.org/api/mozilla.components.concept.fetch/-response/) properties `.success` and `.clientError` were renamed to `.isSuccess` and `isClientError` respectively to match Java conventions.
 
+* **browser-toolbar**
+  * Adds `onCancelEditing` to `onEditListener` in `BrowserToolbar` which is fired when a back button press occurs while the keyboard is displayed.
+    This is especially useful if you want to call `activity.onBackPressed()` to navigate away rather than just dismiss the keyboard.
+    Its return value is used to determine if `displayMode` will switch from edit to view.
+
```
pocmo: Oh, damn. One last thing: This needs to be moved up to 0.47.0 now. And then let's merge :)

PR #2308, opened 2019-03-09
Fixes #1628: Do not display zoom controls when using the Webview

### LINKED ISSUE #1628: System: Disable zoom buttons on page
We should disable these zoom buttons since:
1. they don't work.
2. they're a deprecated way of controlling zoom since we have pinch to zoom.
3. their position seems very off (above the system navigation buttons).

screenshot_20181229-215215

jens1o:
Fixes #1628 

### Pull Request checklist

Does this need a changelog entry?


jens1o:
Could anybody support me? I do not know how to resolve this error I can't reproduce locally.

pengrad:
Maybe `SystemEngineSession.initSettings()` is better place for this

https://github.com/mozilla-mobile/android-components/blob/master/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt#L234

pocmo:
> Maybe `SystemEngineSession.initSettings()` is better place for this
> 
> https://github.com/mozilla-mobile/android-components/blob/master/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt#L234

Yes, this looks like we should do it in this place. :)

jens1o:
Alright, will resolve that tomorrow. Thanks for your help!

pengrad:
Hi @jens1o, sorry that interfering again, I just wanted to contribute this issue too :)
There is a `SystemEngineSessionTest.defaultSettings()` test that directly checks all settings of webview and it fails now.
You can run it locally via `./gradlew :browser-engine-system:testDebugUnitTest`
After fixing it taskcluster will be green and I think this PR will be merged

jens1o:
Thanks for your help! I wonder whether my approach was the greatest solution. Is there any other approach that could be taken without removing this line?

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
```diff
@@ -231,6 +231,8 @@ class SystemEngineSession(
 
             // webViewSettings built-in zoom controls are the only supported ones, so they should be turned on.
             webSettings.builtInZoomControls = true
+            // Disable zoom button on pages #1628
+            webView.settings.displayZoomControls = false
```
jonalmeida: When you're in this code block, you can use `webSettings` to change the `displayZoomControls`.

So it would just be `webSettings.displayZoomControls`.

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
```diff
@@ -231,6 +231,8 @@ class SystemEngineSession(
 
             // webViewSettings built-in zoom controls are the only supported ones, so they should be turned on.
             webSettings.builtInZoomControls = true
+            // Disable zoom button on pages #1628
```
jonalmeida: nit: Space above.

You don't need to add the issue number in the comment.

components/browser/engine-system/src/test/java/mozilla/components/browser/engine/system/SystemEngineSessionTest.kt
```diff
@@ -379,7 +379,6 @@ class SystemEngineSessionTest {
         verify(webViewSettings).userAgentString = "userAgent"
         verify(webViewSettings).mediaPlaybackRequiresUserGesture = false
         verify(webViewSettings).javaScriptCanOpenWindowsAutomatically = true
-        verify(webViewSettings).displayZoomControls = false
```
jonalmeida: We definitely want to keep this in our tests. Maybe there was something else that needed to be altered in the test?

jens1o: Well, without this, it passes fine.

PR #2317, opened 2019-03-11
Closes #2316: Adds BrowserMenuCaption

### LINKED ISSUE #2316: Create a caption menu item
For Fenix we will need a caption menu item in the browser fragment when a custom tab is being used. We should create a menu item for this with a customizable text size for future flexibility.

[See mocks here](https://github.com/mozilla-mobile/fenix/issues/203#issuecomment-463426620)

sblatz:
### Pull Request checklist

### DIFF
--- a/components/browser/menu/src/main/java/mozilla/components/browser/menu/item/SimpleBrowserMenuItem.kt
+++ b/components/browser/menu/src/main/java/mozilla/components/browser/menu/item/SimpleBrowserMenuItem.kt
@@ -18,13 +18,16 @@ private const val NO_ID = -1
  * A simple browser menu item displaying text.
  *
  * @param label The visible label of this menu item.
+ * @param textSize: The size of the label.
+ * @param textColorResource: The color resource to apply to the text.
  * @param listener Callback to be invoked when this menu item is clicked.
  */
 class SimpleBrowserMenuItem(
     private val label: String,
+    private val textSize: Float = NO_ID.toFloat(),
     @ColorRes
     private val textColorResource: Int = NO_ID,
-    private val listener: () -> Unit
+    private val listener: (() -> Unit)? = null
 ) : BrowserMenuItem {
     override var visible: () -> Boolean = { true }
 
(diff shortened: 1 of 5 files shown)


components/browser/menu/src/test/java/mozilla/components/browser/menu/item/BrowserMenuCaptionTest.kt
```diff
@@ -0,0 +1,20 @@
...
+import org.junit.Test
+
+class BrowserMenuCaptionTest {
+
+    @Test
+    fun `browser menu caption uses correct layout`() {
+        val item = BrowserMenuCaption("Powered by Mozilla")
+        assertEquals(R.layout.mozac_browser_menu_item_caption, item.getLayoutResource())
+    }
+}
```
pocmo: Maybe a test to check the text is actually set? :)

sblatz: Ah, I didn't think there was a way to test this, but I looked at `BrowserMenuImageTextTest` for some inspiration. I've added a few more, but please let me know if that's not sufficient!

Amejia481:
@sblatz what is the difference between `BrowserMenuCaption` and `SimpleBrowserMenuItem`?
it's only the text size and the no click listener attach to it, can we add it SimpleBrowserMenuItem`?

sblatz:
@Amejia481 Yes, those are the only differences. I'm happy to add a `textSize` param to `SimpleBrowserMenuItem`, but I also didn't want tapping on it to dismiss the menu, which is why I ended up creating a separate item. Is there another way I can get around this?

Amejia481:
@sblatz we could make, the listener optional and add set the click listener when is not null
 
```kotlin
class SimpleBrowserMenuItem(
        private val label: String,
        @ColorRes
        private val textColorResource: Int = NO_ID,
        private val listener: (() -> Unit)? = null
) : BrowserMenuItem {
    override var visible: () -> Boolean = { true }

    override fun getLayoutResource() = R.layout.mozac_browser_menu_item_simple

    override fun bind(menu: BrowserMenu, view: View) {
        (view as TextView).text = label
        if (textColorResource != NO_ID) {
            view.setTextColor(ContextCompat.getColor(view.context, textColorResource))
        }

        if (listener != null) {
            view.setOnClickListener {
                listener.invoke()
                menu.dismiss()
            }
        } else {
            // This will remove the ripple effect.
            view.background = null
        }
    }
}
```

sblatz:
@Amejia481 seems like a great solution. I'll make that change!

components/browser/menu/src/main/res/layout/mozac_browser_menu_item_caption.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
```
Amejia481: With the new change, I think we are not going to need this.

sblatz: Sorry about that, went to lunch and hadn't finished up my changes. Should be ready for review now!

Amejia481: Don't worry :)

PR #2320, opened 2019-03-12
Closes #651: Add PocketEndpoint!

### LINKED ISSUE #651: Add Pocket endpoint component

mcomella:
Please ignore the first five commits, which are part of PR #2279.

Future development may add:
- An additional method in `PocketEndpoint` to get articles for Pocket Listen
- An additional endpoint (i.e. whole server to contact) to cross-reference audio data for Pocket Listen (if not available in first results)
- A wrapper around `PocketEndpoint`, e.g. `PocketRepository`, that will add higher level behavior like caching and back-off behavior shared between projects (depending on Echo Show implementation)

### Pull Request checklist

### DIFF
--- a/components/service/pocket/src/main/java/mozilla/components/service/pocket/PocketEndpoint.kt
+++ b/components/service/pocket/src/main/java/mozilla/components/service/pocket/PocketEndpoint.kt
@@ -14,7 +14,7 @@ import mozilla.components.service.pocket.net.PocketResponse
 /**
  * Makes requests to the Pocket API and returns the requested data.
  *
- * @see [getInstance] to retrieve an instance.
+ * @see [newInstance] to retrieve an instance.
  */
 class PocketEndpoint internal constructor(
     private val rawEndpoint: PocketEndpointRaw,
@@ -48,7 +48,7 @@ class PocketEndpoint internal constructor(
     companion object {
 
         /**
-         * Returns an instance of [PocketEndpoint].
+         * Returns a new instance of [PocketEndpoint].
          *
          * @param client the HTTP client to use for network requests.
          * @param pocketApiKey the API key for Pocket network requests.
@@ -56,7 +56,7 @@ class PocketEndpoint internal constructor(
          *
          * @throws IllegalArgumentException if the provided API key or user agent is deemed invalid.
          */
-        fun getInstance(client: Client, pocketApiKey: String, userAgent: String): PocketEndpoint {
+        fun newInstance(client: Client, pocketApiKey: String, userAgent: String): PocketEndpoint {
             assertIsValidApiKey(pocketApiKey)
             assertIsValidUserAgent(userAgent)
 
--- a/components/service/pocket/src/main/java/mozilla/components/service/pocket/PocketEndpoint.kt
+++ b/components/service/pocket/src/main/java/mozilla/components/service/pocket/PocketEndpoint.kt
@@ -1,0 +1,75 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.pocket
+
+import android.support.annotation.WorkerThread
+import mozilla.components.concept.fetch.Client
+import mozilla.components.service.pocket.data.PocketGlobalVideoRecommendation
+import mozilla.components.service.pocket.net.PocketResponse
+
+// See @Ignore tests in test class to do end-to-end testing.
+
+/**
+ * Makes requests to the Pocket API and returns the requested data.
+ *
+ * @see [getInstance] to retrieve an instance.
+ */
+class PocketEndpoint internal constructor(
+    private val rawEndpoint: PocketEndpointRaw,
+    private val jsonParser: PocketJSONParser
+) {
+
+    /**
+     * Gets a response, filled with the Pocket global video recommendations from the Pocket API server on success.
+     * This network call is synchronous and the results are not cached. The API version of this call is available in the
+     * source code at [PocketURLs.VALUE_VIDEO_RECS_VERSION].
+     *
+     * If the API returns unexpectedly formatted results, these entries will be omitted and the rest of the items are
+     * returned.
+     *
+     * @return a [PocketResponse.Success] with the Pocket global video recommendations (the list will never be empty)
+     * or, on error, a [PocketResponse.Failure].
+     */
+    @WorkerThread // Synchronous network call.
+    fun getGlobalVideoRecommendations(): PocketResponse<List<PocketGlobalVideoRecommendation>> {
+        val json = rawEndpoint.getGlobalVideoRecommendations()
+        val videoRecs = json?.let { jsonParser.jsonToGlobalVideoRecommendations(it) }
+        return videoRecs.wrapInResponse()
+    }
+
+    private fun <T> List<T>?.wrapInResponse(): PocketResponse<List<T>> = if (isNullOrEmpty()) {
+        PocketResponse.Failure()
+    } else {
+        PocketResponse.Success(this!!)
+    }
+
+    companion object {
+
+        /**
+         * Returns an instance of [PocketEndpoint].
+         *
+         * @param client the HTTP client to use for network requests.
+         * @param pocketApiKey the API key for Pocket network requests.
+         * @param userAgent the user agent for network requests.
+         *
+         * @throws IllegalArgumentException if the provided API key or user agent is deemed invalid.
+         */
+        fun getInstance(client: Client, pocketApiKey: String, userAgent: String): PocketEndpoint {
+            assertIsValidApiKey(pocketApiKey)
+            assertIsValidUserAgent(userAgent)
+
+            val endpoint = PocketEndpointRaw(client, PocketURLs(pocketApiKey), userAgent)
+            return PocketEndpoint(endpoint, PocketJSONParser())
+        }
+    }
+}
+
+private fun assertIsValidApiKey(apiKey: String) {
+    if (apiKey.isBlank()) throw IllegalArgumentException("Expected non-blank API key")
+}
+
+private fun assertIsValidUserAgent(userAgent: String) {
+    if (userAgent.isBlank()) throw java.lang.IllegalArgumentException("Expected non-blank user agent")
+}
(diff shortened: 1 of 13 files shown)


mcomella:
Rebasing to resolve conflicts.

components/service/pocket/src/main/java/mozilla/components/service/pocket/PocketEndpoint.kt
```diff
@@ -0,0 +1,75 @@
...
+     * or, on error, a [PocketResponse.Failure].
+     */
+    @WorkerThread // Synchronous network call.
+    fun getGlobalVideoRecommendations(): PocketResponse<List<PocketGlobalVideoRecommendation>> {
+        val json = rawEndpoint.getGlobalVideoRecommendations()
+        val videoRecs = json?.let { jsonParser.jsonToGlobalVideoRecommendations(it) }
+        return videoRecs.wrapInResponse()
+    }
+
+    private fun <T> List<T>?.wrapInResponse(): PocketResponse<List<T>> = if (isNullOrEmpty()) {
```
pocmo: Over in `PocketParser`we explicitly return null instead of an empty list; but here we are using isNullOrEmpty?

components/service/pocket/src/main/java/mozilla/components/service/pocket/PocketEndpoint.kt
```diff
@@ -0,0 +1,75 @@
...
+        PocketResponse.Failure()
+    } else {
```
pocmo: I wonder if `Failure` should wrap an exception. It would allow the caller to log more meaningful errors. It would also allows you to not map everything to "null" and then to failure and instead just wrap the exception?

components/service/pocket/src/main/java/mozilla/components/service/pocket/PocketEndpoint.kt
```diff
@@ -0,0 +1,75 @@
...
+        /**
+         * Returns an instance of [PocketEndpoint].
+         *
+         * @param client the HTTP client to use for network requests.
+         * @param pocketApiKey the API key for Pocket network requests.
+         * @param userAgent the user agent for network requests.
+         *
+         * @throws IllegalArgumentException if the provided API key or user agent is deemed invalid.
+         */
+        fun getInstance(client: Client, pocketApiKey: String, userAgent: String): PocketEndpoint {
```
pocmo: naming nit: `getInstance()` sounds like the method of a singleton that always returns the same instance. How about. `create()` or `build()` or `createInstance()`?

components/service/pocket/src/main/java/mozilla/components/service/pocket/net/PocketResponse.kt
```diff
@@ -0,0 +1,23 @@
...
+     *
+     * @property data The data returned from the Pocket API.
+     */
+    data class Success<T> internal constructor(val data: T) : PocketResponse<T>()
+
+    /**
+     * A failure response from the Pocket API.
+     */
+    class Failure<T> internal constructor() : PocketResponse<T>()
+}
```
pocmo: This is actually a very generic class and I wonder if we would want such a class in the base component.

components/service/pocket/src/main/java/mozilla/components/service/pocket/PocketEndpoint.kt
mcomella: I wasn't sure how to handle this. I was concerned about someone (unintentionally) changing `PocketJSONParser` to return different values – such as an empty list – so I wanted to code defensively on its consumer, `PocketEndpoint`. However, it's very verbose to have to code defensively all the way up the call hierarchy for contracts we write so I can see why this is a problem too. I wish I could represent this in the type system.

@pocmo Any suggestions on how you would handle this? Maybe I should have `JSONParser` return an empty list instead so I don't have to redundantly handle null and empty.

components/service/pocket/src/main/java/mozilla/components/service/pocket/net/PocketResponse.kt
mcomella: @pocmo It's up to you – if so, let me know! fwiw, Severin suggested returning a wrapping sealed class, rather than null or an empty list directory because:
- RxJava does not support null values so the conusmer would have to wrap anyway
- Returning a sealed class lets us extend the API to add more information later without breaking changes (e.g. adding the exception to the failure).

components/service/pocket/src/main/java/mozilla/components/service/pocket/PocketEndpoint.kt
mcomella: It's possible this getter method actually caches instances under the hood (though I seriously doubt we'll ever do that) so I didn't want to use `create*` or `build` but I can see the problem with `getInstance`. How about, `get`?

components/service/pocket/src/main/java/mozilla/components/service/pocket/PocketEndpoint.kt
mcomella: That's a good question. I programmed it like I was working on a local application (log it ASAP and convert it into an easier error type to deal with than Exception) rather than a public API but it might make sense to expose this data in the API.

That being said, propagating exceptions adds more complexity than just returning an error value like null (i.e. as an API owner, I should catch all known exceptions and convert them into predictable types) and, as a consumer, I doubt I'd do anything other than log it (i.e. I wouldn't handle different errors differently) so I don't think it's worth that complexity right now. We can always add a `cause` property to `Failure` without introducing a breaking change later so I'd vote we stick with what we have now and add the Exception if consumers ask for it.

@pocmo What do you think?

pocmo:
@mcomella I gave you (and the whole APT team) WRITE access. It looks like that wasn't the case before. You should all be able to land patches yourself after approval. :)

components/service/pocket/src/main/java/mozilla/components/service/pocket/PocketEndpoint.kt
pocmo: I guess I would either make PocketJSONParser return a Success/Failure too (and map here if needed) or throw and catch here and then map/wrap that. But tbh I'm okay with any approach you chose. It only read odd to me to first map and empty list to null and then later handle null values and empty lists the same. :)

components/service/pocket/src/main/java/mozilla/components/service/pocket/PocketEndpoint.kt
pocmo: > (i.e. as an API owner, I should catch all known exceptions and convert them into predictable types)

But isn't that what you are doing already? 

> We can always add a cause property to Failure without introducing a breaking change later so I'd vote we stick with what we have now and add the Exception if consumers ask for it.

Alright!

components/service/pocket/src/main/java/mozilla/components/service/pocket/net/PocketResponse.kt
pocmo: Alright. Something to consider later on. I guess in Arrow that would be a [Try](https://arrow-kt.io/docs/arrow/core/try/).

components/service/pocket/src/main/java/mozilla/components/service/pocket/PocketEndpoint.kt
pocmo: `get` is a bit better. But I think I would still expect a Singleton instance from it. :)

> (though I seriously doubt we'll ever do that)

Why choose a name that will work with an unlikely future change though? :)

Anyhow, I'm fine with whatever you decide to pick. :)

components/service/pocket/src/main/java/mozilla/components/service/pocket/PocketEndpoint.kt
mcomella: I agree that wrapping it will provide a more explicit response type but unfortunately, even when wrapping it, we still run into the "the list is empty" issue. I think I'm leaning towards a more generic response wrapping type (like Arrow's Try) that would have a `mapIfSuccessAndNotEmpty` function, such that failures are automatically converted to failures and lists are automatically checked for empty. However, I don't think I understand the problem well enough to come up with the right solution right now so in the interest of time, I'm going to leave it as is. Maybe I'll refactor when I come back for `PocketListenEndpoint`

components/service/pocket/src/main/java/mozilla/components/service/pocket/PocketEndpoint.kt
mcomella: Effective Java seems to imply `getInstance` with parameters doesn't imply singleton: 

> getInstance — Returns an instance that is described by the parameters but cannot be said to have the same value. In the case of a singleton, getInstance takes no parameters and returns the sole instance.

via https://stackoverflow.com/a/22442058

mcomella: So I'm going to leave it as is.

mcomella: > Why choose a name that will work with an unlikely future change though? :)

To be fair, picking a more specific name like `newInstance` or `build` enforces constraints – this will always return a new instance – that maybe we don't want to be confined to.

mcomella: Then again, people's code might assume it returns an newInstance each time so perhaps it's better to name it what it is and create a new method if the behavior changes. I'll do that.

components/service/pocket/src/main/java/mozilla/components/service/pocket/PocketEndpoint.kt
mcomella: > > (i.e. as an API owner, I should catch all known exceptions and convert them into predictable types)
> 
> But isn't that what you are doing already?

😰

mcomella:
Looks like some glean unit tests failed, despite me not touching those files, which I think might be intermittent: I filed https://github.com/mozilla-mobile/android-components/issues/2359 I'll try to re-run TC.

mcomella:
And now it's green. :)

components/service/pocket/src/main/java/mozilla/components/service/pocket/net/PocketResponse.kt
mcomella: @pocmo Woah woah woah: this is in the kotlin stdlib! `Result`: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/index.html

I'll open a PR to change the code over to Result.

PR #2321, opened 2019-03-12
Closes #1382: Add implementation for capture thumbnail of websites  on GeckoEngineView.

### LINKED ISSUE #1382: Add implementation for capture thumbnail of websites on GeckoView
User story: https://github.com/mozilla-mobile/reference-browser/issues/189

There's a issue on Bugzilla for the GeckoView API need it.
https://bugzilla.mozilla.org/show_bug.cgi?id=1462018

Amejia481:
### Pull Request checklist

### DIFF
--- a/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
+++ b/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
@@ -48,7 +48,6 @@ import mozilla.components.concept.engine.EngineView
 import mozilla.components.concept.engine.HitResult
 import mozilla.components.concept.engine.prompt.PromptRequest
 import mozilla.components.concept.engine.request.RequestInterceptor.InterceptionResponse
-import mozilla.components.support.ktx.android.content.isOSOnLowMemory
 import mozilla.components.support.utils.DownloadUtils
 import java.util.Date
 
@@ -157,13 +156,6 @@ class SystemEngineView @JvmOverloads constructor(
                             secure = cert != null,
                             host = cert?.let { Uri.parse(url).host },
                             issuer = cert?.issuedBy?.oName)
-
-                    if (!isLowOnMemory()) {
-                        val thumbnail = session?.captureThumbnail()
-                        if (thumbnail != null) {
-                            onThumbnailChange(thumbnail)
-                        }
-                    }
                 }
             }
         }
@@ -265,11 +257,6 @@ class SystemEngineView @JvmOverloads constructor(
             }
         }
     }
-
-    @VisibleForTesting
-    internal var testLowMemory = false
-
-    private fun isLowOnMemory() = testLowMemory || (context?.isOSOnLowMemory() == true)
 
     @Suppress("ComplexMethod")
     private fun createWebChromeClient() = object : WebChromeClient() {
@@ -592,6 +579,20 @@ class SystemEngineView @JvmOverloads constructor(
     }
 
     override fun canScrollVerticallyDown() = session?.webView?.canScrollVertically(1) ?: false
+
+    override fun captureThumbnail(onFinish: (Bitmap?) -> Unit) {
+        val webView = session?.webView
+
+        val thumbnails = if (webView == null) {
+            null
+        } else {
+            webView.buildDrawingCache()
+            val outBitmap = webView.drawingCache?.let { cache -> Bitmap.createBitmap(cache) }
+            webView.destroyDrawingCache()
+            outBitmap
+        }
+        onFinish(thumbnails)
+    }
 
     private fun resetJSAlertAbuseState() {
         jsAlertCount = 0
--- a/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
+++ b/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
@@ -583,7 +583,7 @@ class SystemEngineView @JvmOverloads constructor(
     override fun captureThumbnail(onFinish: (Bitmap?) -> Unit) {
         val webView = session?.webView
 
-        val thumbnails = if (webView == null) {
+        val thumbnail = if (webView == null) {
             null
         } else {
             webView.buildDrawingCache()
@@ -591,7 +591,7 @@ class SystemEngineView @JvmOverloads constructor(
             webView.destroyDrawingCache()
             outBitmap
         }
-        onFinish(thumbnails)
+        onFinish(thumbnail)
     }
 
     private fun resetJSAlertAbuseState() {
--- a/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineView.kt
+++ b/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineView.kt
@@ -68,7 +68,7 @@ interface EngineView {
     fun canScrollVerticallyDown(): Boolean = true
 
     /**
-     * Take a screenshot of the actual session.
+     * Request a screenshot of the visible portion of the web page currently being rendered.
      * @param onFinish A callback to inform that process of capturing a thumbnail has finished.
      */
     fun captureThumbnail(onFinish: (Bitmap?) -> Unit)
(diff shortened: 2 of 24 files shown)


components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -593,6 +580,20 @@ class SystemEngineView @JvmOverloads constructor(
 
     override fun canScrollVerticallyDown() = session?.webView?.canScrollVertically(1) ?: false
 
+    override fun captureThumbnail(onFinish: (Bitmap?) -> Unit) {
+        val webView = session?.webView
+
+        val thumbnails = if (webView == null) {
```
pocmo: nit: `thumbnails` -> `thumbnail?

components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineView.kt
```diff
@@ -65,6 +66,12 @@ interface EngineView {
     * true if can and false otherwise.
     */
     fun canScrollVerticallyDown(): Boolean = true
+
+    /**
+     * Take a screenshot of the actual session.
+     * @param onFinish A callback to inform that process of capturing a thumbnail has finished.
+     */
```
pocmo: nit: Using the term `session` here may be confusing? Maybe we can borrow from the GeckoView docs: 
https://mozilla.github.io/geckoview/javadoc/mozilla-central/org/mozilla/geckoview/GeckoView.html#capturePixels--

components/feature/session/src/main/java/mozilla/components/feature/session/ThumbnailsFeature.kt
```diff
@@ -0,0 +1,69 @@
...
+import mozilla.components.browser.session.SelectionAwareSessionObserver
+import mozilla.components.browser.session.Session
+import mozilla.components.browser.session.SessionManager
+import mozilla.components.concept.engine.EngineView
+import mozilla.components.support.base.feature.LifecycleAwareFeature
+import mozilla.components.support.ktx.android.content.isOSOnLowMemory
+
+/**
+ * Feature implementation for automatically taking thumbnails of sites.
+ */
```
pocmo: nit: Could be a bit more descriptive: When does it take thumbnails? What does it do with them?

components/feature/session/src/main/java/mozilla/components/feature/session/ThumbnailsFeature.kt
```diff
@@ -0,0 +1,69 @@
...
+    ) : SelectionAwareSessionObserver(sessionManager) {
+
+        override fun onProgress(session: Session, progress: Int) {
+            if (progress == PROGRESS_STOP) {
+                requestPrintScreen(session)
+            }
+        }
+    }
+
+    private fun requestPrintScreen(session: Session) {
```
pocmo: "print screen"?

components/feature/session/src/main/java/mozilla/components/feature/session/ThumbnailsFeature.kt
```diff
@@ -0,0 +1,69 @@
...
+        if (!isLowOnMemory()) {
+            engineView.captureThumbnail {
+                session.thumbnail = it
+            }
+        }
```
pocmo: nit: I assume in a low memory situation we would want to assign `null`? Otherwise we will have a "ghost" image in low memory situations that will never update/change.

components/feature/session/src/main/java/mozilla/components/feature/session/ThumbnailsFeature.kt
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    internal inner class ThumbnailsFeatureRequestObserver(
+        sessionManager: SessionManager
+    ) : SelectionAwareSessionObserver(sessionManager) {
+
+        override fun onProgress(session: Session, progress: Int) {
+            if (progress == PROGRESS_STOP) {
+                requestPrintScreen(session)
+            }
```
pocmo: Instead of "onProgress" you may want to look at `onLoadingStateChanged`.

Interestingly Fennec delays taking the screenshot (500ms) after receiving the state change:
https://searchfox.org/mozilla-central/source/mobile/android/base/java/org/mozilla/gecko/Tab.java#718

components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineView.kt
Amejia481: 👍

components/feature/session/src/main/java/mozilla/components/feature/session/ThumbnailsFeature.kt
Amejia481: 👍

Amejia481: I didn't want to add when it takes the screenshot because I wasn't sure if it is an implementation detail and we may change it.
 
Does this sound better ? :)

```
/**
 * Feature implementation for automatically taking thumbnails of sites.
 * The feature will take a screenshot when the page finishes loading,
 * and will add it to the [Session.thumbnail] property.
 */
```

components/feature/session/src/main/java/mozilla/components/feature/session/ThumbnailsFeature.kt
Amejia481: Sorry 🙃

components/feature/session/src/main/java/mozilla/components/feature/session/ThumbnailsFeature.kt
pocmo: Sounds great :)

components/feature/session/src/main/java/mozilla/components/feature/session/ThumbnailsFeature.kt
Amejia481: When you said "ghost image", did you refer to the previous  `session.thumbnail`?
Because, right now if we are under low memory conditions, we are not requesting a new thumbnail.

```kotlin
        if (!isLowOnMemory()) {
            engineView.captureThumbnail {
                session.thumbnail = it
            }
        }
```

Did you mean `session.thumbnail  = null` on  low memory?

components/feature/session/src/main/java/mozilla/components/feature/session/ThumbnailsFeature.kt
Amejia481: What would be the main difference between them? They are getting called in very similar places
image

https://github.com/mozilla-mobile/android-components/blob/e5cefefbcecf898c07fd149faea20f94015a168d/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt#L367

components/feature/session/src/main/java/mozilla/components/feature/session/ThumbnailsFeature.kt
pocmo: Yep. Otherwise this will keep showing the old image. Better no image than an outdated/wrong one?

components/feature/session/src/main/java/mozilla/components/feature/session/ThumbnailsFeature.kt
pocmo: This is an implementation detail though. There's no guarantee that this will always be the case. For example this is different in `browser-engine-system`. Also there's no guarantee that the loading state will always hit 100% (but it will stop).

components/feature/session/src/main/java/mozilla/components/feature/session/ThumbnailsFeature.kt
Amejia481: Yeah, I didn't add it initially because we are already doing it on `SessionManager`. 

  ```kotlin
  /**
     * Informs this [SessionManager] that the OS is in low memory condition so it
     * can reduce its allocated objects.
     */
    fun onLowMemory() {
        // Removing the all the thumbnails except for the selected session to
        // reduce memory consumption.
        sessions.forEach {
            if (it != selectedSession) {
                it.thumbnail = null
            }
        }
}
```

https://github.com/mozilla-mobile/android-components/blob/024e3de456e3b46e9bf6718db9500ecc52da3d29/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt#L472

components/feature/session/src/main/java/mozilla/components/feature/session/ThumbnailsFeature.kt
Amejia481: 👍

Amejia481:
@pocmo  the PR is update with the suggested changes :)

PR #2328, opened 2019-03-12
[glean] 1533746: Make events survive application restarts

mdboom:
This modifies the events storage engine to use an append-only file on disk to store events so that they survive application shutdowns.

### Pull Request checklist

### DIFF
--- a/components/service/glean/docs/pings/events.md
+++ b/components/service/glean/docs/pings/events.md
@@ -4,10 +4,23 @@
 The events ping's purpose is to transport all of the event metric information.
 
 ## Scheduling
-The `events` ping is normally sent when the application goes into the
-[background](pings.md#defining-background-state). It is also sent when the queue of events exceeds
-`Glean.configuration.maxEvents` (default 500). This happens automatically, with
-no intervention or configuration required by the application.
+
+The `events` ping is sent under the following circumstances:
+
+- Normally, it is sent when the application goes into the
+  [background](pings.md#defining-background-state), if there
+  are any recorded events to send.
+
+- When the queue of events exceeds `Glean.configuration.maxEvents` (default
+  500).
+
+- If there are any unsent events found on disk when starting the application. It
+  would be impossible to coordinate the timestamps across a reboot, so it's best
+  to just collect all events from the previous run into their own ping, and
+  start over.
+
+All of these cases are handled automatically, with no intervention or
+configuration required by the application.
 
 ## Contents
 At the top-level, this ping contains the following keys:
@@ -19,7 +32,7 @@ At the top-level, this ping contains the following keys:
 
 Each entry in the `events` array is an array with the following items:
 
-- [0]: `msSinceStart`: The milliseconds since the start of the process.
+- [0]: `timestamp`: The milliseconds relative to the first event in the ping.
 
 - [1]: `category`: The category of the event, as defined by its location in the
   `metrics.yaml` file.
@@ -27,5 +40,5 @@ Each entry in the `events` array is an array with the following items:
 - [2]: `name`: The name of the event, as definded in the `metrics.yaml` file.
 
 - [3]: `extra` (optional): A mapping of strings to strings providing additional
-  data about the event. The keys are restricted to 40 characters and values in 
+  data about the event. The keys are restricted to 40 characters and values in
   this map will never exceed 100 characters.
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
@@ -92,6 +92,8 @@ open class GleanInternalAPI internal constructor () {
         // API. For this reason we're safe to set `initialized = true` right after it.
         initializeCoreMetrics(applicationContext)
 
+        initialized = true
+
         // Deal with any pending events so we can start recording new ones
         EventsStorageEngine.onReadyToSendPings(applicationContext)
 
@@ -100,7 +102,6 @@ open class GleanInternalAPI internal constructor () {
         // on being dispatched to the API context before any other metric.
         metricsPingScheduler = MetricsPingScheduler(applicationContext)
         metricsPingScheduler.startupCheck()
-        initialized = true
 
         // Other pings might set some other metrics (i.e. the baseline metrics),
         // so we need to be initialized by now.
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
@@ -25,8 +25,10 @@ import mozilla.components.service.glean.storages.PingStorageEngine
 import mozilla.components.service.glean.storages.ExperimentsStorageEngine
 import mozilla.components.service.glean.storages.UuidsStorageEngine
 import mozilla.components.service.glean.storages.DatetimesStorageEngine
+import mozilla.components.service.glean.storages.EventsStorageEngine
 import mozilla.components.service.glean.storages.RecordedExperimentData
 import mozilla.components.service.glean.storages.StringsStorageEngine
+import mozilla.components.service.glean.utils.ensureDirectoryExists
 import mozilla.components.support.base.log.logger.Logger
 
 @Suppress("TooManyFunctions")
@@ -90,6 +92,9 @@ open class GleanInternalAPI internal constructor () {
         // API. For this reason we're safe to set `initialized = true` right after it.
         initializeCoreMetrics(applicationContext)
 
+        // Deal with any pending events so we can start recording new ones
+        EventsStorageEngine.onReadyToSendPings(applicationContext)
+
         // Set up information and scheduling for glean owned pings. Ideally, the "metrics"
         // ping startup check should be performed before any other ping, since it relies
         // on being dispatched to the API context before any other metric.
@@ -101,6 +106,7 @@ open class GleanInternalAPI internal constructor () {
         // so we need to be initialized by now.
         baselinePing = BaselinePing()
 
+        // At this point, all metrics and events can be recorded.
         ProcessLifecycleOwner.get().lifecycle.addObserver(gleanLifecycleObserver)
     }
 
@@ -201,16 +207,7 @@ open class GleanInternalAPI internal constructor () {
 
         val gleanDataDir = File(applicationContext.applicationInfo.dataDir, Glean.GLEAN_DATA_DIR)
 
-        // Make sure the data directory exists and is writable.
-        if (!gleanDataDir.exists() && !gleanDataDir.mkdirs()) {
-            logger.error("Failed to create Glean's data dir ${gleanDataDir.absolutePath}")
-            return
-        }
-
-        if (!gleanDataDir.isDirectory || !gleanDataDir.canWrite()) {
-            logger.error("Glean's data directory is not a writable directory ${gleanDataDir.absolutePath}")
-            return
-        }
+        ensureDirectoryExists(gleanDataDir)
 
         // The first time Glean runs, we set the client id and other internal
         // one-time only metrics.
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
@@ -9,31 +9,57 @@ import android.annotation.SuppressLint
 import android.content.Context
 import android.os.SystemClock
 import android.support.annotation.VisibleForTesting
+import kotlinx.coroutines.Dispatchers as KotlinDispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withTimeout
+import mozilla.components.service.glean.Dispatchers
 import mozilla.components.service.glean.error.ErrorRecording.recordError
 import mozilla.components.service.glean.error.ErrorRecording.ErrorType
 import mozilla.components.service.glean.EventMetricType
 import mozilla.components.service.glean.Lifetime
 import mozilla.components.service.glean.scheduler.PingUploadWorker
+import mozilla.components.service.glean.utils.ensureDirectoryExists
 import mozilla.components.support.base.log.logger.Logger
 import org.json.JSONArray
+import org.json.JSONException
 import org.json.JSONObject
+import java.io.File
+import java.io.IOException
 
 /**
  * This singleton handles the in-memory storage logic for events. It is meant to be used by
  * the Specific Events API and the ping assembling objects.
  *
+ * So that the data survives shutting down of the application, events are stored
+ * in an append-only file on disk, in addition to the store in memory. Each line
+ * of this file records a single event in JSON, exactly as it will be sent in the
+ * ping.  There is one file per store.
+ *
  * This class contains a reference to the Android application Context. While the IDE warns
  * us that this could leak, the application context lives as long as the application and this
  * object. For this reason, we should be safe to suppress the IDE warning.
  */
-@SuppressLint("StaticFieldLeak")
+@SuppressLint("StaticFieldLeak, TooManyFunctions")
 internal object EventsStorageEngine : StorageEngine {
     override lateinit var applicationContext: Context
 
     // Events recorded within the list should be reasonably sorted by timestamp, assuming
     // the sequence of calls to [record] has not been messed with. However, please only
     // trust the recorded timestamp values.
-    private val eventStores: MutableMap<String, MutableList<RecordedEventData>> = mutableMapOf()
+    internal val eventStores: MutableMap<String, MutableList<RecordedEventData>> = mutableMapOf()
+
+    // The storage directory where the append-only events files reside.
+    internal val storageDirectory: File by lazy {
+        val dir = File(
+            applicationContext.applicationInfo.dataDir,
+            "${Glean.GLEAN_DATA_DIR}/events/"
+        )
+        ensureDirectoryExists(dir)
+        dir
+    }
 
     // The timestamp of recorded events is computed relative to this point in time
     // which should roughly match with the process start time. Note that, according
@@ -43,7 +69,73 @@ internal object EventsStorageEngine : StorageEngine {
     // Maximum length of any string value in the extra dictionary, in characters
     internal const val MAX_LENGTH_EXTRA_KEY_VALUE = 100
 
+    // The position of the fields within the JSON payload for each event
+    internal const val TIMESTAMP_FIELD = 0
+    internal const val CATEGORY_FIELD = 1
+    internal const val NAME_FIELD = 2
+    internal const val EXTRA_FIELD = 3
+    internal const val N_FIELDS = 4
+
     private val logger = Logger("glean/EventsStorageEngine")
+
+    // Timeout for waiting on async IO (during testing only)
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    internal const val JOB_TIMEOUT_MS = 250L
+
+    // A lock to prevent simultaneous writing of the event files
+    private val eventFileLock = Any()
+
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    var ioTask: Job? = null
+
+    /**
+     * Initialize events storage. This must be called once on application startup,
+     * e.g. from [Glean.initialize], but after we are ready to send pings, since
+     * this could potentially collect and send pings.
+     *
+     * If there are any events queued on disk, it loads them into memory so
+     * that the memory and disk representations are in sync.
+     *
+     * Secondly, if this is the first time the application has been run since
+     * rebooting, any pings containing events are assembled into pings and cleared
+     * immediately, since their timestamps won't be compatible with the timestamps
+     * we would create during this boot of the device.
+     *
+     * @param context The application context
+     */
+    internal fun onReadyToSendPings(context: Context) {
+        // We want this to run off of the main thread, because it might perform I/O.
+        // However, we don't use the built-in KotlinDispatchers.IO since we need
+        // to make sure this work is done before any other glean API calls, and this
+        // will force them to be queued after this work.
+        Dispatchers.API.launch {
+            // Load events from disk
+            storageDirectory.listFiles()?.forEach { file ->
+                val storeData = eventStores.getOrPut(file.name) { mutableListOf() }
+                file.forEachLine { line ->
+                    try {
+                        val jsonContent = JSONArray(line)
+                        val event = deserializeEvent(jsonContent)
+                        storeData.add(event)
+                    } catch (e: JSONException) {
+                        // pass
+                    }
+                }
+            }
+
+            // TODO Technically, we only need to send out all pending events
+            // if they came from a previous boot (since their timestamps won't match).
+            // However, there are various challenges in detecting when the device has
+            // been booted, so for now we just pay the penalty of a few more
+            // unnecessary pings.
+            if (eventStores.size != 0) {
+                eventStores.keys.forEach { storeName ->
+                    Glean.assembleAndSerializePing(storeName)
+                }
+                PingUploadWorker.enqueueWorker()
+            }
+        }
+    }
 
     /**
      * Record an event in the desired stores.
@@ -84,14 +176,20 @@ internal object EventsStorageEngine : StorageEngine {
             eventKeys
         }
 
-        val msSinceStart = monotonicElapsedMs - startTime
-        val event = RecordedEventData(metricData.category, metricData.name, msSinceStart, truncatedExtraKeys)
+        val event = RecordedEventData(
+            metricData.category,
+            metricData.name,
+            monotonicElapsedMs,
+            truncatedExtraKeys
+        )
+        val jsonEvent = serializeEvent(event).toString()
 
         // Record a copy of the event in all the needed stores.
         synchronized(this) {
             for (storeName in metricData.getStorageNames()) {
                 val storeData = eventStores.getOrPut(storeName) { mutableListOf() }
                 storeData.add(event.copy())
+                writeEventToDisk(storeName, jsonEvent)
                 if (storeData.size == Glean.configuration.maxEvents &&
                     Glean.assembleAndSerializePing(storeName)) {
                     // Queue background worker to upload the newly created ping
@@ -110,13 +208,26 @@ internal object EventsStorageEngine : StorageEngine {
      *
      * @return the list of events recorded in the requested store
      */
-    @Synchronized
     fun getSnapshot(storeName: String, clearStore: Boolean): List<RecordedEventData>? {
-        if (clearStore) {
-            return eventStores.remove(storeName)
+        // Rewrite all of the timestamps to they are relative to the first
+        // timestamp
+        val events = eventStores[storeName]?.let { store ->
+            var firstTimestamp = store[0].timestamp
+            store.map { event ->
+                event.copy(timestamp = event.timestamp - firstTimestamp)
+            }
         }
 
-        return eventStores[storeName]
+        if (clearStore) {
+            ioTask = GlobalScope.launch(KotlinDispatchers.IO) {
+                synchronized(eventFileLock) {
+                    getFile(storeName).delete()
+                }
+            }
+            eventStores.remove(storeName)
+        }
+
+        return events
     }
 
     /**
@@ -131,16 +242,85 @@ internal object EventsStorageEngine : StorageEngine {
         return getSnapshot(storeName, clearStore)?.let { pingEvents ->
             val eventsArray = JSONArray()
             pingEvents.forEach {
-                val eventData = JSONArray()
-                eventData.put(it.msSinceStart)
-                eventData.put(it.category)
-                eventData.put(it.name)
-                if (it.extra != null) {
-                    eventData.put(JSONObject(it.extra))
-                }
-                eventsArray.put(eventData)
+                eventsArray.put(serializeEvent(it))
             }
-            return eventsArray
+            eventsArray
+        }
+    }
+
+    /**
+     * Get the File object in which to store events for the given store.
+     *
+     * @param storeName The name of the store
+     * @return File object to store events
+     */
+    private fun getFile(storeName: String): File {
+        return File(storageDirectory, storeName)
+    }
+
+    /**
+     * Serializes an event to its JSON representation.
+     *
+     * @param category Event category
+     * @param name Event name
+     * @param msSinceStart The ms since the process started
+     * @param extra A `Map<String, String>` of extra values
+     * @return JSONArray representing the event
+     */
+    private fun serializeEvent(event: RecordedEventData): JSONArray {
+        val eventData = JSONArray()
+        eventData.put(event.timestamp)
+        eventData.put(event.category)
+        eventData.put(event.name)
+        if (event.extra != null) {
+            eventData.put(JSONObject(event.extra))
+        }
+        return eventData
+    }
+
+    /**
+     * Deserializes an event in JSON into a RecordedEventData object.
+     *
+     * @param jsonContent The JSONArray containing the data for the event.
+     * @return event data object
+     */
+    private fun deserializeEvent(jsonContent: JSONArray): RecordedEventData {
+        if (jsonContent.length() != N_FIELDS) {
+            throw JSONException("Invalid event")
+        }
+
+        val extra: Map<String, String>? = jsonContent.optJSONObject(EXTRA_FIELD)?.let {
+            val extraValues: MutableMap<String, String> = mutableMapOf()
+            val names = it.names()
+            for (i in 0..(names.length() - 1)) {
+                extraValues[names.getString(i)] = it.getString(names.getString(i))
+            }
+            extraValues
+        }
+
+        return RecordedEventData(
+            jsonContent.getString(CATEGORY_FIELD),
+            jsonContent.getString(NAME_FIELD),
+            jsonContent.getLong(TIMESTAMP_FIELD),
+            extra
+        )
+    }
+
+    /**
+     * Writes an event to a single store on disk.
+     *
+     * @param storeName The name of the store to add the event to.
+     * @param eventContent A string of JSON as returned by [serializeEvent]
+     */
+    internal fun writeEventToDisk(storeName: String, eventContent: String) {
+        ioTask = GlobalScope.launch(KotlinDispatchers.IO) {
+            synchronized(eventFileLock) {
+                try {
+                    getFile(storeName).appendText("${eventContent}\n")
+                } catch (e: IOException) {
+                    logger.warn("IOException while writing event to disk: $e")
+                }
+            }
         }
     }
 
@@ -149,7 +329,29 @@ internal object EventsStorageEngine : StorageEngine {
 
     @VisibleForTesting(otherwise = VisibleForTesting.NONE)
     override fun clearAllStores() {
+        // We need to delete these files on the IO thread to avoid
+        // racing with writing to the file.  However, we want to
+        // block the main thread until this work is done, so the
+        // caller can rely on the files being gone.  This is not a
+        // performance problem since this function is for use in
+        // testing only.
+        synchronized(eventFileLock) {
+            storageDirectory.listFiles()?.forEach {
+                it.delete()
+            }
+        }
         eventStores.clear()
+    }
+
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    internal fun testWaitForWrites(timeout: Long = JOB_TIMEOUT_MS) {
+        ioTask?.let {
+            runBlocking {
+                withTimeout(timeout) {
+                    it.join()
+                }
+            }
+        }
     }
 }
 
@@ -157,7 +359,7 @@ internal object EventsStorageEngine : StorageEngine {
 data class RecordedEventData(
     val category: String,
     val name: String,
-    val msSinceStart: Long,
+    var timestamp: Long,
     val extra: Map<String, String>? = null,
 
     internal val identifier: String = if (category.isEmpty()) { name } else { "$category.$name" }
--- a/components/service/glean/src/test/java/mozilla/components/service/glean/storages/EventsStorageEngineTest.kt
+++ b/components/service/glean/src/test/java/mozilla/components/service/glean/storages/EventsStorageEngineTest.kt
@@ -18,12 +18,14 @@ import mozilla.components.service.glean.resetGlean
 import mozilla.components.service.glean.triggerWorkManager
 import okhttp3.mockwebserver.MockResponse
 import okhttp3.mockwebserver.MockWebServer
+import org.json.JSONObject
 import org.junit.Assert
 import org.junit.Before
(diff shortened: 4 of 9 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
```diff
@@ -133,11 +157,36 @@ internal object EventsStorageEngine : StorageEngine {
      */
     @Synchronized
     fun getSnapshot(storeName: String, clearStore: Boolean): List<RecordedEventData>? {
+        val file = getFile(storeName)
```
Dexterp37: Mh, are we hitting I/O every time we take a snapshot? What if we loaded the content of the file at startup, put that in a different map, e.g. `persistedEventsMap`, and if the map was non empty use it here?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
```diff
@@ -153,15 +202,145 @@ internal object EventsStorageEngine : StorageEngine {
...
+     * assembled immediately.
+     *
+     * @param storeName The name of the store to add the event to.
+     * @param eventContent A string of JSON as returned by [serializeEvent]
+     * @return Returns true if the event buffer filled up with
+     * [Glean.configuration.maxEvents] events
+     */
+    @Synchronized
+    fun writeEventToDisk(storeName: String, eventContent: String): Boolean {
+        var count = eventStoreCounts.get(storeName) ?: getEventCount(storeName)
```
Dexterp37: nit: `count` doesn't seem to ever change - should this be a `val`?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
```diff
@@ -153,15 +202,145 @@ internal object EventsStorageEngine : StorageEngine {
...
+
+        try {
+            getFile(storeName).appendText("${eventContent}\n")
+        } catch (e: IOException) {
+            logger.warn("IOException while writing event to disk: $e")
+            return false
+        }
+
+        var newCount = count + 1
```
Dexterp37: nit: `newCount` doesn't seem to ever change - should this be a `val`?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
```diff
@@ -153,15 +202,145 @@ internal object EventsStorageEngine : StorageEngine {
...
+        eventStoreCounts[storeName] = newCount
+
+        return (
+            newCount == Glean.configuration.maxEvents &&
+            Glean.assembleAndSerializePing(storeName)
```
Dexterp37: Can we use `Glean.sendPings` here instead of calling this and then manually enqueuing the upload worker? I'm not sure why we had it like that before. If there's any specific reason, would you kindly drop a comment in the code?

components/service/glean/src/main/java/mozilla/components/service/glean/utils/FreshBootDetector.kt
```diff
@@ -0,0 +1,118 @@
...
+     * For earlier versions of Android, we can use the Linux kernel's
+     * `/proc/sys/kernel/random/boot_id`.
+     *
+     * Adapted from https://github.com/instacart/truetime-android/pull/93
+     *
+     * @param context An application context
+     */
+    @Suppress("MagicNumber")
+    internal fun getBootId(context: Context): String {
+        if (Build.VERSION.SDK_INT >= 24) {
```
Dexterp37: nit: let's rather have `24` be a constant above and remove the suppression

components/service/glean/src/main/java/mozilla/components/service/glean/utils/FreshBootDetector.kt
```diff
@@ -0,0 +1,118 @@
...
+
+import android.content.Context
+import android.os.Build
+import android.provider.Settings
+import android.support.annotation.VisibleForTesting
+import java.io.File
+import java.io.IOException
+import java.util.UUID
+
+internal object FreshBootDetector {
```
Dexterp37: Can we have somebody from android components kindly review this? Seems very Android specific and I don't feel too comfortable in reviewing this :( @pocmo , any chance you could help ? :)

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
```diff
@@ -153,15 +202,145 @@ internal object EventsStorageEngine : StorageEngine {
...
+        return File(storageDirectory, storeName)
+    }
+
+    /**
+     * Get the number of events currently recorded in the given store on
+     * disk.  This is only needed if the count stored in [eventStoreCount] is
+     * missing.
+     *
+     * @param storeName The name of the store
+     * @return count The number of events in the store.
```
Dexterp37: nit: I think we can remove `count` from here as `@return` doesn't expect it

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
```diff
@@ -153,15 +202,145 @@ internal object EventsStorageEngine : StorageEngine {
...
+
+    /**
+     * Flushes all pings containing any queued events on disk.  Any retained
+     * non-event metrics in those pings will also be sent at this time.
+     */
+    @Synchronized
+    fun flushEvents() {
+        var needToSend = false
+        storageDirectory.listFiles()?.forEach {
+            if (Glean.assembleAndSerializePing(it.name)) {
```
Dexterp37: Mh, I see why we're not calling `sendPings` here. However, is it worth it? Should we simply try to call it and rely on the fact that the `WorkManager` is smart enough to to the right thing with the current enqueued work?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
```diff
@@ -105,20 +128,21 @@ internal object EventsStorageEngine : StorageEngine {
...
         synchronized(this) {
+            var needToSend = false
             for (storeName in metricData.getStorageNames()) {
-                val storeData = eventStores.getOrPut(storeName) { mutableListOf() }
-                storeData.add(event.copy())
-                if (storeData.size == Glean.configuration.maxEvents &&
-                    Glean.assembleAndSerializePing(storeName)) {
-                    // Queue background worker to upload the newly created ping
-                    PingUploadWorker.enqueueWorker()
+                if (writeEventToDisk(storeName, jsonContent)) {
```
Dexterp37: What if we stored things both in the `eventStores` and on disk: we could retain the old behaviour and delegate disk writes to a synchronized queue dispatched on a different thread (so that we don't block the `API` dispatcher for too long). 

Moreover, taking the log for as long as I/O would last sounds a bit too long :(

mdboom:
TBH, I was skeptical about this approach, too, but it's what the existing telemetry library does, and I (without any measurement) sort of thought that the disk cache would be able to deal with this pretty well (it's all sequential accesses, for example).  And I kind of preferred the simplicity of having a single source of truth for the data.  But I'm not opposed to doing something that puts the I/O in a different place.  I think the tradeoff is that rather than paying for I/O on sending every event-containing ping, we'd pay I/O at startup if there were any queued events.

Additionally, is it true that `Dispatcher.API.launch` is blocking?  The docs say "Launches new coroutine without blocking current thread and returns a reference to the coroutine as a Job."  So I took that to mean there's some sort of queueing going on, so while we might block other work in the dispatcher, we're never blocking the main thread.

> At startup, if the files exist, we could load their content in a separate in-memory map, e.g. persistedEventsStore and delete them. Then, when taking the snapshot, we could merge the maps and retain the current snapshotting functionality (+ clearing the files, if they exist). I'm basically suggesting to only use the files when starting the application, for writing events to them or for clearing them.

I suppose since we send the events ping on background, it probably doesn't matter too much -- but we could lose events that way, since the contents of `persistedEventsStore` would end up existing only in  volatile memory.  I think we'd want to load them and not delete them, and the just start adding to the file and in memory representation in tandem.

mdboom:
It would be nice if this way of detecting fresh boot is acceptable.  The other standard way to do this seems to be to handle a `BOOT_COMPLETED` intent, but that would slow the user's boot time.

travis79:
> It would be nice if this way of detecting fresh boot is acceptable. The other standard way to do this seems to be to handle a `BOOT_COMPLETED` intent, but that would slow the user's boot time.

That's what I've read about `BOOT_COMPLETED` as well, which is why I thought your way was clever :).  Especially since we don't need to know _at boot_ just that it was the first time run _since boot_.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
```diff
@@ -151,16 +247,85 @@ internal object EventsStorageEngine : StorageEngine {
...
+
+    /**
+     * Writes an event to a single store on disk.
+     *
+     * @param storeName The name of the store to add the event to.
+     * @param eventContent A string of JSON as returned by [serializeEvent]
+     */
+    @Synchronized
+    @VisibleForTesting
+    fun writeEventToDisk(storeName: String, eventContent: String) {
```
Dexterp37: nit: should this be `internal`?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
```diff
@@ -151,16 +247,85 @@ internal object EventsStorageEngine : StorageEngine {
...
+        )
+    }
+
+    /**
+     * Writes an event to a single store on disk.
+     *
+     * @param storeName The name of the store to add the event to.
+     * @param eventContent A string of JSON as returned by [serializeEvent]
+     */
+    @Synchronized
```
Dexterp37: This is happening on a separate thread and we're not really accessing `eventStore`. Does this need to be `@Synchronized`?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
```diff
@@ -169,15 +334,29 @@ internal object EventsStorageEngine : StorageEngine {
...
     @VisibleForTesting(otherwise = VisibleForTesting.NONE)
     override fun clearAllStores() {
+        storageDirectory.listFiles()?.forEach {
+            it.delete()
+        }
         eventStores.clear()
     }
+
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    fun testWaitForWrites(timeout: Long = 250L) {
```
Dexterp37: nit: should this be `internal`?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
```diff
@@ -151,16 +247,85 @@ internal object EventsStorageEngine : StorageEngine {
...
+        }
+    }
+
+    /**
+     * Get the File object in which to store events for the given store.
+     *
+     * @param storeName The name of the store
+     * @return File object to store events
+     */
+    fun getFile(storeName: String): File {
```
Dexterp37: nit: should this be `private`?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
```diff
@@ -151,16 +247,85 @@ internal object EventsStorageEngine : StorageEngine {
...
+    /**
+     * Serializes an event to its JSON representation.
+     *
+     * @param category Event category
+     * @param name Event name
+     * @param msSinceStart The ms since the process started
+     * @param extra A `Map<String, String>` of extra values
+     * @return JSONArray representing the event
+     */
+    fun serializeEvent(event: RecordedEventData): JSONArray {
```
Dexterp37: nit: should this be `private`?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
```diff
@@ -151,16 +247,85 @@ internal object EventsStorageEngine : StorageEngine {
...
+        return eventData
+    }
+
+    /**
+     * Deserializes an event in JSON into a RecordedEventData object.
+     *
+     * @param jsonContent The JSONArray containing the data for the event.
+     * @return event data object
+     */
+    fun deserializeEvent(jsonContent: JSONArray): RecordedEventData {
```
Dexterp37: nit: should this be `private`?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
```diff
@@ -151,16 +247,85 @@ internal object EventsStorageEngine : StorageEngine {
...
+        if (jsonContent.length() != N_FIELDS) {
+            throw JSONException("Invalid event")
+        }
+
+        val extra: Map<String, String>? = jsonContent.optJSONObject(EXTRA_FIELD)?.let {
+            var extraValues: MutableMap<String, String> = mutableMapOf()
```
Dexterp37: nit: does this need to be a `var`? Does the compiler complain if its a `val`?

components/service/glean/src/test/java/mozilla/components/service/glean/storages/EventsStorageEngineTest.kt
```diff
@@ -350,4 +356,149 @@ class EventsStorageEngineTest {
...
+        )
+        EventsStorageEngine.testWaitForWrites()
+
+        assertEquals(1, EventsStorageEngine.eventStores["store1"]!!.size)
+        EventsStorageEngine.eventStores.clear()
+
+        // Reinitializing Glean should reload existing events on disk
+        Glean.initialized = false
+        FreshBootDetector.cacheIsFreshBoot = false
+        Glean.initialize(ApplicationProvider.getApplicationContext())
```
Dexterp37: Does it make sense to slightly change `resetGlean` to allow for skipping data clearing? This would make it future proof, in case something in the way we reset glean in tests changes in the future.

components/service/glean/src/test/java/mozilla/components/service/glean/storages/EventsStorageEngineTest.kt
```diff
@@ -350,4 +356,149 @@ class EventsStorageEngineTest {
             ) == snapshot.first().extra)
         assertEquals(1, testGetNumRecordedErrors(testEvent, ErrorType.InvalidValue))
     }
+
+    @Test
+    fun `continue recording events from disk`() {
+        val event = EventMetricType(
```
Dexterp37: I would assert that no store file is present at the beginning of this test.

components/service/glean/src/test/java/mozilla/components/service/glean/storages/EventsStorageEngineTest.kt
```diff
@@ -350,4 +356,149 @@ class EventsStorageEngineTest {
...
+            Glean.configuration.copy(
+                serverEndpoint = "http://" + server.hostName + ":" + server.port,
+                logPings = true
+            )
+        )
+
+        // Trigger worker task to upload the pings in the background
+        triggerWorkManager()
+
+        val request = server.takeRequest(20L, TimeUnit.SECONDS)
```
Dexterp37: Let's make sure that the ping we're receiving is of the expected type and that the expected events are being transported. Given that our metric is declaring just `"store1"`, i'm not sure why this test is passing, honestly :|

components/service/glean/src/test/java/mozilla/components/service/glean/storages/EventsStorageEngineTest.kt
```diff
@@ -350,4 +356,149 @@ class EventsStorageEngineTest {
...
+            serverEndpoint = "http://" + server.hostName + ":" + server.port,
+            logPings = true
+        ))
+
+        val event = EventMetricType(
+            disabled = false,
+            category = "telemetry",
+            name = "test_event_clear",
+            lifetime = Lifetime.Ping,
+            sendInPings = listOf("store1"),
```
Dexterp37: Should the `sendInPings` be "default"? We have no `store1` ping.

components/service/glean/src/test/java/mozilla/components/service/glean/storages/EventsStorageEngineTest.kt
```diff
@@ -350,4 +356,149 @@ class EventsStorageEngineTest {
...
+            allowedExtraKeys = listOf("foo")
+        )
+
+        // Record the event in the store, without providing optional arguments.
+        EventsStorageEngine.record(
```
Dexterp37: Should we use `event.record()` here and use the testing API as well?

components/service/glean/src/test/java/mozilla/components/service/glean/storages/EventsStorageEngineTest.kt
```diff
@@ -350,4 +356,149 @@ class EventsStorageEngineTest {
...
+            disabled = false,
+            category = "telemetry",
+            name = "test_event",
+            lifetime = Lifetime.Ping,
+            sendInPings = listOf("store1"),
+            allowedExtraKeys = listOf("foo")
+        )
+
+        // Record the event in the store, without providing optional arguments.
+        EventsStorageEngine.record(
```
Dexterp37: Can we use `event.record()` here?

components/service/glean/src/test/java/mozilla/components/service/glean/storages/EventsStorageEngineTest.kt
```diff
@@ -350,4 +356,149 @@ class EventsStorageEngineTest {
             ) == snapshot.first().extra)
         assertEquals(1, testGetNumRecordedErrors(testEvent, ErrorType.InvalidValue))
     }
+
+    @Test
+    fun `continue recording events from disk`() {
```
Dexterp37: Can we add a test to make sure that store files are correctly deleted when getting a snapshot with `clear = true`?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
```diff
@@ -169,15 +334,29 @@ internal object EventsStorageEngine : StorageEngine {
 
     @VisibleForTesting(otherwise = VisibleForTesting.NONE)
     override fun clearAllStores() {
+        storageDirectory.listFiles()?.forEach {
```
Dexterp37: Does this `forEach` need to run in the `GlobalScope.launch(KotlinDispatchers.IO) ` context? Otherwise we risk racing with writes, right?

components/service/glean/docs/pings/events.md
```diff
@@ -4,10 +4,22 @@
...
 
 ## Scheduling
-The `events` ping is normally sent when the application goes into the
-[background](pings.md#defining-background-state). It is also sent when the queue of events exceeds
-`Glean.configuration.maxEvents` (default 500). This happens automatically, with
-no intervention or configuration required by the application.
+
+The `events` ping is sent under the following circumstances:
+
+- Normally, it is sent when the application goes into the
```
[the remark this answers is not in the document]
mdboom: Good point.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -90,6 +91,8 @@ open class GleanInternalAPI internal constructor () {
         initializeCoreMetrics(applicationContext)
 
         ProcessLifecycleOwner.get().lifecycle.addObserver(gleanLifecycleObserver)
+
+        EventsStorageEngine.initialize(applicationContext)
```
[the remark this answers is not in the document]
mdboom: Trying this, things get a little hairy.

`EventsStorageEngine.initialize` potentially sends pings, so it needs a `PingMaker` and `PingStorageEngine` initialized.  However, `PingMaker` needs a `StorageEngineManager`, so we can't init `EventsStorageEngine` inside of `StorageEngineManager`.  Maybe changing the name of `EventsStorageEngine.initialize` to `onStartup` or something would make this clearer, that it's not really initializing -- it's doing the things it needs to do on startup once the basic glean functionality is available?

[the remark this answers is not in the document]
Dexterp37: I see - yes, let's rename to `onStartup` or `onGleanInitComplete`? We want to be specific about *when* things are expected to happen, so that we don't forget or introduce weirdness in future commits. Let's also add your comment as a code comment near the call site, please.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
```diff
@@ -43,8 +71,58 @@ internal object EventsStorageEngine : StorageEngine {
...
+     *
+     * Secondly, if this is the first time the application has been run since
+     * rebooting, any pings containing events are assembled into pings and cleared
+     * immediately, since their timestamps won't be compatible with the timestamps
+     * we would create during this boot of the device.
+     *
+     * @param context The application context
+     */
+    @Synchronized
+    internal fun initialize(context: Context) {
```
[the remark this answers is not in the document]
mdboom: It definitely makes sense to add test coverage for this.  I think it's already mostly covered, though, by the tests that three new tests that (a) record some events, (b) initialize glean, (c) record more events.  If the queuing happening out-of-order, the old events would appear after the new events.  I think all it hopefully takes to cover this is adding an assertion to make sure the events are in the correct order.

> In order to address the second point, I think it should be sufficient to move the init as early as possible in Glean.initialize and postpone setting initialized = true as late as possible, at least later than this initialize call. That would guarantee that any API call would be enqueued after this.

If `Glean.initialized` is false, metric recording just falls on the floor, it doesn't get queued.  So I actually think we want this flag sent *before* the first asynchronous thing (in this case `EventStorageEngine.initialize`).

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
```diff
@@ -43,8 +71,58 @@ internal object EventsStorageEngine : StorageEngine {
...
+     * that the memory and disk representations are in sync.
+     *
+     * Secondly, if this is the first time the application has been run since
+     * rebooting, any pings containing events are assembled into pings and cleared
+     * immediately, since their timestamps won't be compatible with the timestamps
+     * we would create during this boot of the device.
+     *
+     * @param context The application context
+     */
+    @Synchronized
```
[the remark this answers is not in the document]
mdboom: Good catch.  I think there are some `@Synchronized` things here that are hangovers from the earlier file-only implementation (where I wanted to ensure that there were no simultaneous reads/writes to the file).  There's a bunch that it looks like can be removed...

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
mdboom: Yeah.  My brain hurts jumping languages that have different default visibilities...

components/service/glean/src/test/java/mozilla/components/service/glean/storages/EventsStorageEngineTest.kt
mdboom: That's a good idea.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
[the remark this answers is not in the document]
Dexterp37: > > In order to address the second point, I think it should be sufficient to move the init as early as possible in Glean.initialize and postpone setting initialized = true as late as possible, at least later than this initialize call. That would guarantee that any API call would be enqueued after this.
> 
> If `Glean.initialized` is false, metric recording just falls on the floor, it doesn't get queued. So I actually think we want this flag sent _before_ the first asynchronous thing (in this case `EventStorageEngine.initialize`).

Gah! You're right :)

components/service/glean/src/test/java/mozilla/components/service/glean/storages/EventsStorageEngineTest.kt
mdboom: There's nothing magic about the `events` ping -- event metrics can be on any ping.

But yes, it's a good idea to do some deeper inspection of the ping payload.

components/service/glean/src/test/java/mozilla/components/service/glean/storages/EventsStorageEngineTest.kt
mdboom: Doesn't really matter either way, but most of the tests in this file and others do that...

components/service/glean/src/test/java/mozilla/components/service/glean/storages/EventsStorageEngineTest.kt
mdboom: Good idea.

components/service/glean/src/test/java/mozilla/components/service/glean/storages/EventsStorageEngineTest.kt
mdboom: Good idea.  I'll just add this as part of the existing `clearStores` test.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
mdboom: Yes, potentially.  However, this is should probably also block the main thread until that work is done.  That should be fine since this is a testing-only function.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -82,14 +83,20 @@ open class GleanInternalAPI internal constructor () {
...
+        // At this point, we are ready to record metrics (other than events)
+        // and do things that could potentially send pings.
+        initialized = true
 
+        // Deal with any pending events so we can start recording new ones
+        EventsStorageEngine.onReadyToSendPings(applicationContext)
+
+        // At this point, all metrics and events can be recorded.
         ProcessLifecycleOwner.get().lifecycle.addObserver(gleanLifecycleObserver)
+        initializeCoreMetrics(applicationContext)
```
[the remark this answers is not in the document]
mdboom: Strangely, moving it before `initialized = true` causes 3 tests to fail, though I haven't been able to figure out why.  It works as the very first thing after `initialized = true` though, and that's probably fine as long as we never record events from that function.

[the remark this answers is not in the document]
Dexterp37: Sounds good to me!

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
```diff
@@ -169,15 +338,41 @@ internal object EventsStorageEngine : StorageEngine {
 
     @VisibleForTesting(otherwise = VisibleForTesting.NONE)
     override fun clearAllStores() {
+        // We need to delete these files on the IO thread to avoid
+        // racing with writing to the file.  However, we want to
+        // block the main thread until this work is done, so the
+        // caller can rely on the files being gone.  This is not a
+        // performance problem since this function is for use in
+        // testing only.
```
pocmo: There's no single IO thread. `Dispatchers.IO` uses at least 64 threads. Can this be solved with other synchronization mechanisms?

components/service/glean/src/main/java/mozilla/components/service/glean/utils/FreshBootDetector.kt
pocmo: This is somewhat scary. I have no clue how stable that is and if that works across devices. 🤷‍♂️ 

It would be interesting to collect telemetry about the boot id 😄

components/service/glean/src/main/java/mozilla/components/service/glean/utils/FreshBootDetector.kt
```diff
@@ -0,0 +1,102 @@
...
+     *
+     * For earlier versions of Android, we can use the Linux kernel's
+     * `/proc/sys/kernel/random/boot_id`.
+     *
+     * Adapted from https://github.com/instacart/truetime-android/pull/93
+     *
+     * @param context An application context
+     */
+    internal fun getBootId(context: Context): String {
+        if (Build.VERSION.SDK_INT > Build.VERSION_CODES.O) {
```
pocmo: The linked PR uses `Build.VERSION_CODES.M` but there we are using `O`?

components/service/glean/src/main/java/mozilla/components/service/glean/utils/FreshBootDetector.kt
```diff
@@ -0,0 +1,102 @@
...
+            try {
+                return Settings.Global.getInt(
+                    context.getContentResolver(),
+                    Settings.Global.BOOT_COUNT
+                ).toString()
+            } catch (e: Settings.SettingNotFoundException) {
+                // pass through to old method
```
pocmo: In the linked discussion it sounds like accessing `boot_id` doesn't work anymore on Android M+ ... does it make sense to pass through?

components/service/glean/src/main/java/mozilla/components/service/glean/utils/FreshBootDetector.kt
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        try {
+            val file = File("/proc/sys/kernel/random/boot_id")
+
+            val bootId = StringBuilder()
+
+            file.forEachLine {
+                bootId.append(it)
+            }
```
pocmo: nit: Maybe `readText()` instead of looping over lines?

components/service/glean/src/main/java/mozilla/components/service/glean/utils/FreshBootDetector.kt
```diff
@@ -0,0 +1,102 @@
...
+
+            return bootId.toString()
+        } catch (e: IOException) {
+            // pass through to fallback method
+        }
```
pocmo: Is that the exception it throws on <= M too? Just wondering if it could be a `SecurityException`. That's what a bunch of things started throwing with M+ and runtime permissions.

components/service/glean/src/main/java/mozilla/components/service/glean/utils/FreshBootDetector.kt
```diff
@@ -0,0 +1,102 @@
...
+
+        // If neither of the above methods works, just return
+        // a random id, so it will appear that we are always at
+        // a fresh reboot.
+        return UUID.randomUUID().toString()
```
pocmo: It would be interesting to know how often that happens. This would be much simpler if it would be possible to apply this fallback to all devices. :)

components/service/glean/src/main/java/mozilla/components/service/glean/utils/FreshBootDetector.kt
```diff
@@ -0,0 +1,102 @@
...
+        val prefs = context.getSharedPreferences(
+            this.javaClass.canonicalName, Context.MODE_PRIVATE
+        )
+
+        val currentBootId = getBootId(context)
+        val storedBootId = prefs.getString("bootId", "")
+
+        cacheIsFreshBoot = (currentBootId != storedBootId)
+
+        if (cacheIsFreshBoot!!) {
```
pocmo: nit: Instead of `!!` it makes more sense to do `!!` on the nullable String.

Or use `== true` or `== false`.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/EventsStorageEngine.kt
mdboom: Ah, thanks.  I think using a `synchronized(lock) { }` should work just fine, then.

components/service/glean/src/main/java/mozilla/components/service/glean/utils/FreshBootDetector.kt
mdboom: That might be due to me misunderstanding the testing.  When I had this at `M`, this if clause ran, but then it didn't have the setting. This does contradict the [docs](https://developer.android.com/reference/android/provider/Settings.Global.html#BOOT_COUNT) which it should be available as of `N` (SDK 24).

components/service/glean/src/main/java/mozilla/components/service/glean/utils/FreshBootDetector.kt
mdboom: Yeah, perhaps not.  I tried to find a reference to that `/proc` file being removed or permission-locked, but couldn't find any.  I found a bunch of references to [other parts of proc being locked down](https://issuetracker.google.com/issues/37140047), though.

components/service/glean/src/main/java/mozilla/components/service/glean/utils/FreshBootDetector.kt
mdboom: Sure.

components/service/glean/src/main/java/mozilla/components/service/glean/utils/FreshBootDetector.kt
mdboom: I couldn't actually reproduce the exception.  On `M` it passed, but I didn't try anything newer.

components/service/glean/src/main/java/mozilla/components/service/glean/utils/FreshBootDetector.kt
mdboom: Yes.  The motivation here is that the app may be shut down and restarted many times during the lifetime of a custom ping containing events, and we don't want to be sending extra pings on every process start.  If the events from the last run were during the same boot of the device, their timestamps will be compatible.  If they came from a previous boot, we have to send immediately and "start over" (because relying on realtime clocks to close the gap seemed ever more complex)...

It seems like the standard alternative is to listen to the `BOOT_COMPLETE` intent.  I thought that would be more problematic since it would start the app (at least in the background) on boot.  Might that still be preferred, just because it's a more well-trodden path?

components/service/glean/src/main/java/mozilla/components/service/glean/utils/FreshBootDetector.kt
pocmo: Alright :)

components/service/glean/src/main/java/mozilla/components/service/glean/utils/FreshBootDetector.kt
pocmo: > It seems like the standard alternative is to listen to the BOOT_COMPLETE intent. I thought that would be more problematic since it would start the app (at least in the background) on boot.

That sounds okay-ish since it would only start your broadcast receiver (and initialize the application object). It also feels more like using a defined behavior than accessing something that is not really guaranteed to be accessible?

Anyhow, if we land as-is then let's test this on the different API levels and let's see how it behaves in the reference browser. :)

mdboom: I'll probably try out the `BOOT_COMPLETE` intent approach to see how that shakes out, so we have a point of comparison.

mdboom:
@pocmo: I added a commit that changes this to using a `BroadcastReceiver` to handle `BOOT_COMPLETED` instead.  That intent has been around since API 1, so hopefully there should be good support for that.

Somewhat importantly for "lean data principles", that receiver only notes that there's been a boot.  The handling of that fact (which could lead to sending pings) doesn't happen until that app is explicitly started by the user.

components/service/glean/src/main/AndroidManifest.xml
```diff
@@ -2,9 +2,22 @@
    - License, v. 2.0. If a copy of the MPL was not distributed with this
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="mozilla.components.service.glean" >
+          package="mozilla.components.service.glean" >
+
+    <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED"/>
```
Dexterp37: @pocmo what's your opinion of glean requiring this permission? Is this reasonable for a library to come with such requirements?

components/service/glean/src/main/java/mozilla/components/service/glean/utils/FreshBootDetector.kt
```diff
@@ -4,92 +4,79 @@
...
-import java.io.IOException
-import java.util.UUID
+import mozilla.components.support.base.log.logger.Logger
+
+/*
+ * Receives the BOOT_COMPLETED intent.
+ *
+ * Sets a pref when the device first boots up.
+ */
+class BootCompletedReceiver : BroadcastReceiver() {
```
Dexterp37: @mdboom the new permission requirement shuts me down a bit on this approach. What do you think of approaching this like "always send any pending chunk of data" for now? I really don't want us to block on the boot detection mechanism: we could file a bug, track that as future follow-up work and document the fact that we might generate pings every time we start in `events.md`.

mdboom: One data point: the existing telemetry library [requests this permission](https://github.com/mozilla-mobile/android-components/blob/c8583395fc902aa964e101fa20c652d16c7fc02d/components/service/telemetry/src/main/AndroidManifest.xml#L8), though I'm not sure where it's actually used.

Dexterp37: Ah, I see. Then it's fine for me.

components/service/glean/src/main/AndroidManifest.xml
Dexterp37: Resolving this, as we noticed that the legacy telemetry library uses the same permission.

pocmo:
One last question: In apps we initialize from `Application.onCreate()`. This will happen before the boot receiver is run. Is that an issue? What if there's a (custom?) event added in this early stage?

mdboom:
> One last question: In apps we initialize from Application.onCreate().  This will happen before the boot receiver is run. Is that an issue?  What if there's a (custom?) event added in this early stage?

Indeed, that could be a problem.  Hmmm...  Not sure how to resolve it, though.  Maybe move the initialization to something like `onActivityStarted`?  That's cumbersome, though...

We should be fine as long as `onCreate` doesn't record events, but I'm not sure how to enforce that.

mdboom:
Hmm... actually it's worse than that -- I think the whole idea breaks down if glean is initialized from `Application.onCreate`...  Back to the drawing board...

mdboom:
@Dexterp37, @travis79, @pocmo: I'm giving up on boot detection for now in order to get this into a mergeable state.  I've recorded most of the gory details [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1535639) to pick this up at a later date.

PR #2339, opened 2019-03-13
Closes #2332: Add WebAppManifestParser for creating WebAppManifest from JSON.

### LINKED ISSUE #2332: Parse JSON to create WebAppManifest
GeckoView will give us a `JSONObject` (that is already validated and cleaned up though!) - so we will need to create the `WebAppManifest` object from it.

pocmo:
### Pull Request checklist


components/browser/session/src/main/java/mozilla/components/browser/session/manifest/WebAppManifestParser.kt
```diff
@@ -0,0 +1,152 @@
...
+        "standalone" -> WebAppManifest.DisplayMode.STANDALONE
+        "fullscreen" -> WebAppManifest.DisplayMode.FULLSCREEN
+        "minimal-ui" -> WebAppManifest.DisplayMode.MINIMAL_UI
+        "browser" -> WebAppManifest.DisplayMode.BROWSER
+        else -> WebAppManifest.DisplayMode.BROWSER
+    }
+}
+
+@Suppress("MagicNumber")
+private fun parseColor(color: String?): Int? {
```
Amejia481: nit: You could simplify the function a bit by combining the pre conditions.
```kotlin
    private fun parseColor(color: String?): Int? {
        if (color == null || !color.startsWith("#")) {
            return null
        }
        
        return try {
            Integer.parseInt(color.substring(1), 16)
        } catch (e: NumberFormatException) {
            null
        }
    }
```

components/browser/session/src/test/resources/manifests/spec_typical.json
```diff
@@ -0,0 +1,38 @@
...
+  "icons": [{
+    "src": "icon/lowres.webp",
+    "sizes": "64x64",
+    "type": "image/webp"
+  },{
+    "src": "icon/lowres.png",
+    "sizes": "64x64"
+  }, {
+    "src": "icon/hd_hi",
+    "sizes": "128x128"
```
Amejia481: Just for the sake of having more use cases tested, could we add a sample with an icon with multiple sizes? None of the manifest samples has it. Something like  ` "sizes": "128x128 192x192"`

pocmo:
Okay, updated PR. :)

PR #2353, opened 2019-03-14
1528260: Make expires mandatory

mdboom:
This updates `glean_parser` to a new version that makes the `expires` keyword mandatory.

### Pull Request checklist


mdboom:
The SDK doc has been updated, and the metric parameter docs [here](https://mozilla.github.io/glean_parser/metrics-yaml.html#expires).  It probably should be added to the inline docs in this repo, too, but we don't have any metrics parameters documented here yet, so maybe out-of-scope for this PR and could be done as part of [1529761](https://bugzilla.mozilla.org/show_bug.cgi?id=1529761).

Dexterp37:
> 
> 
> The SDK doc has been updated, and the metric parameter docs [here](https://mozilla.github.io/glean_parser/metrics-yaml.html#expires). It probably should be added to the inline docs in this repo, too, but we don't have any metrics parameters documented here yet, so maybe out-of-scope for this PR and could be done as part of [1529761](https://bugzilla.mozilla.org/show_bug.cgi?id=1529761).

Let's also comment on the bug with what needs to be documented, so that we don't loose track of it.

mdboom:
>  Let's also comment on the bug with what needs to be documented, so that we don't loose track of it.

I don't think we need to -- it's already documented, and updated, just in the current location for metric parameters in `glean_parser`.  It's already noted in the bug that all of that will probably be moved together as a whole.

samples/glean/samples-glean-library/metrics.yaml
```diff
@@ -16,6 +16,7 @@ sample_metrics:
     bugs:
       - 123456789
     data_reviews:
-      - N/A 
+      - N/A
     notification_emails:
       - telemetry-client-dev@mozilla.com
+    expires: never
```
liuche: For the samples, which people will be looking at, we should put something else other than `never` - that shouldn't be a field that is used often, so having it in the example will just lead to confusion
@mdboom

liuche:
Looking at the docs listed in comment https://github.com/mozilla-mobile/android-components/pull/2353#issuecomment-472842200 , we should add a clarification on the `never: This metric never expires` beyond the explanation of what it is used for, but explaining this is for Glean SDK baseline metrics, and should not be used without prior approval. Documentation is meant to be used, not just explaining what things are :)

samples/glean/samples-glean-library/metrics.yaml
mdboom: That's a good point.  However, I'm not sure how to do that without our sample application breaking every six months...  I'll file a bug for this and come back to it, for sure.

PR #2357, opened 2019-03-14
Add Timing Distribution Metric

travis79:
### Pull Request checklist


travis79:
Still pretty rough, I'm working out the kinks in the storage engine and getting ready to write tests.  I originally over-engineered the storage engine and have backed off a bit to simplify it as much as possible.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
```diff
@@ -0,0 +1,243 @@
...
+    // bucket counts are set first.
+    internal val buckets: List<Long> by lazy { getBuckets() }
+
+    /**
+     * Accumulates a sample to the correct bucket
+     *
+     * @param sample Long value representing the sample that is being accumulated
+     */
+    internal fun accumulate(sample: Long) {
+        for (i in buckets.indices) {
```
Dexterp37: Let's briefly mention what's this algorithm  in a comment ;)

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
```diff
@@ -0,0 +1,243 @@
...
+    var overflow: Long = 0,
+    var sum: Long = 0,
+    val timeUnit: TimeUnit = TimeUnit.Millisecond
+) {
+    companion object {
+        // These represent the indexes of the min and max of the timing distribution range
+        const val MIN = 0
+        const val MAX = 1
+
+        // The following are defaults for a simple timing distribution
```
Dexterp37: Let's document how these values were determined and why.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
```diff
@@ -0,0 +1,243 @@
...
+     */
+    private fun getBuckets(): List<Long> {
+        val logMax = Math.log(range[MAX].toDouble())
+        val result: MutableList<Long> = mutableListOf()
+        var current = range[MIN]
+        if (current == 0L) {
+            current = 1L
+        }
+        result.add(current)
+        for (i in 2..bucketCount) {
```
Dexterp37: Let's quicky describe the algorithm below with a comment.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
```diff
@@ -0,0 +1,243 @@
...
+
+@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+data class TimingDistributionData(
+    val category: String,
+    val name: String,
+    val bucketCount: Int = DEFAULT_BUCKET_COUNT,
+    val range: List<Long> = listOf(DEFAULT_RANGE_MIN, DEFAULT_RANGE_MAX),
+    val histogramType: HistogramType = HistogramType.Exponential,
+    val values: MutableMap<String, Long> = mutableMapOf(),
+    var underflow: Long = 0,
```
Dexterp37: What are `underflow` and `overflow` used for?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
```diff
@@ -0,0 +1,243 @@
...
+        metricData: CommonMetricData,
+        sample: Long,
+        timeUnit: TimeUnit
+    ) {
+        val newTD = TimingDistributionData(category = metricData.category, name = metricData.name,
+            timeUnit = timeUnit)
+        newTD.accumulate(sample)
+        super.recordScalar(metricData, newTD, null) { currentValue, newValue ->
+            currentValue?.let {
+                it.accumulate(newValue.sum)
```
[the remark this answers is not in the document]
travis79: I was playing around with this thinking that it had something to do with my multiple-accumulation bug I was working on.  I totally agree that `sample` is much clearer.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
```diff
@@ -0,0 +1,243 @@
...
+            return TimingDistributionData(
+                category = jsonObject.tryGetString("category").orEmpty(),
+                name = jsonObject.tryGetString("name").orEmpty(),
+                bucketCount = jsonObject.tryGetInt("bucketCount") ?: DEFAULT_BUCKET_COUNT,
+                range = try {
+                    jsonObject.getJSONArray("range").toList<Long>()
+                } catch (e: org.json.JSONException) {
+                    listOf(DEFAULT_RANGE_MIN, DEFAULT_RANGE_MAX)
+                },
+                histogramType = HistogramType.values()[jsonObject.tryGetInt("histogramType")
```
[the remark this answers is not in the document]
travis79: I wasn't really intending for it to be a guess, rather I wanted it to default to something sane if it doesn't exist.  Once I move the parsing and validation out of the constructor, as requested, hopefully I can make this more clear.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
```diff
@@ -0,0 +1,243 @@
...
+                    ?: HistogramType.Exponential.ordinal],
+                values = try {
+                    val mapData = jsonObject.getJSONObject("values")
+                    val valueMap: MutableMap<String, Long> = mutableMapOf()
+                    mapData.keys().forEach { key ->
+                        valueMap[key] = mapData.tryGetLong(key) ?: 0L
+                    }
+                    valueMap
+                } catch (e: org.json.JSONException) {
+                    // This should only occur if there isn't a key/value pair stored for "values"
```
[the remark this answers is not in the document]
travis79: It just means that nothing has been accumulated for this metric yet, in which case it shouldn't need to be stored or transmitted.  Let me see if there is something intelligent I can do with this that would bail out of those situations.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
```diff
@@ -127,41 +150,54 @@ data class TimingDistributionData(
...
+            val category = jsonObject.tryGetString("category").orEmpty()
+            val name = jsonObject.tryGetString("name") ?: return null
+            val bucketCount = jsonObject.tryGetInt("bucketCount") ?: return null
+            // If 'range' isn't present, JSONException is thrown
+            val range = try {
+                jsonObject.getJSONArray("range").toList<Long>()
+            } catch (e: org.json.JSONException) {
+                return null
+            }
+            val histogramType = HistogramType.values()[jsonObject.tryGetInt("histogramType")
```
Dexterp37: I'd rework this as:

```
val rawHistogramType = jsonObject.tryGetInt("histogramType") ?: return null
val histogramType = HistogramType.values().getOrNull(rawHistogramType) ?: return null
```

components/service/glean/src/main/java/mozilla/components/service/glean/TimingDistributionMetricType.kt
```diff
@@ -34,11 +34,11 @@ data class TimingDistributionMetricType(
     private val logger = Logger("glean/TimingDistributionMetricType")
 
     // Holds the Job returned from launch{} for awaiting purposes
+    @Volatile
```
Dexterp37: Actually, sorry, let's remove this. This was an attempt fix for the intermittent, but it's not working as intended. Let's settle with what @mdboom finds.

components/service/glean/src/test/java/mozilla/components/service/glean/TimingDistributionMetricTypeTest.kt
```diff
@@ -0,0 +1,139 @@
...
+            name = "string_list_metric",
+            sendInPings = listOf("store1"),
+            timeUnit = TimeUnit.Millisecond
+        )
+        metric.testGetValue()
+    }
+
+    @Test
+    fun `The API saves to secondary pings`() {
+        // Define a timing distribution metric which will be stored in "store1" and "store2"
```
Dexterp37: nit: and "store3" as well :)

components/service/glean/src/main/java/mozilla/components/service/glean/storages/StorageEngineManager.kt
```diff
@@ -31,6 +31,7 @@ internal class StorageEngineManager(
         "string" to StringsStorageEngine,
         "string_list" to StringListsStorageEngine,
         "timespan" to TimespansStorageEngine,
+        "timing_distribution" to TimingDistributionsStorageEngine,
```
Dexterp37: You also need to update `getStorageEngineForMetric` in this file.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
```diff
@@ -0,0 +1,277 @@
...
+     * @param sample the value to accumulate
+     */
+    @Synchronized
+    fun accumulate(
+        metricData: CommonMetricData,
+        sample: Long,
+        timeUnit: TimeUnit
+    ) {
+        if (sample < 0) {
+            ErrorRecording.recordError(
```
Dexterp37: Can we add test coverage for this?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
```diff
@@ -0,0 +1,277 @@
...
+        ))
+    }
+
+    /**
+     * Helper function to generate the list of bucket max values used when accumulating to the
+     * correct buckets.
+     *
+     * @return List containing the bucket limits
+     */
+    private fun getBuckets(): List<Long> {
```
Dexterp37: Can you please add unit test for this specific function?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
```diff
@@ -0,0 +1,277 @@
...
+            }
+        }
+        sum += sample
+    }
+
+    /**
+     * Helper function to build the [TimingDistributionData] into a JSONObject for serialization
+     * purposes.
+     */
+    internal fun toJsonObject(): JSONObject {
```
Dexterp37: Can you please add unit test for this specific function?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
```diff
@@ -0,0 +1,277 @@
...
+    // This is a list of limits for the buckets.  Instantiated lazily to ensure that the range and
+    // bucket counts are set first.
+    internal val buckets: List<Long> by lazy { getBuckets() }
+
+    /**
+     * Accumulates a sample to the correct bucket
+     *
+     * @param sample Long value representing the sample that is being accumulated
+     */
+    internal fun accumulate(sample: Long) {
```
Dexterp37: Can we add test coverage for this specific function, to make sure that the search below is correct? (this is not the same of testing the caller)

components/service/glean/src/test/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngineTest.kt
```diff
@@ -0,0 +1,167 @@
...
+import mozilla.components.service.glean.resetGlean
+import org.junit.Assert.assertEquals
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.ArgumentMatchers
+import org.mockito.Mockito
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
```
Dexterp37: Can we add test coverage for trying to accumulate negative values (check that errors are being reported), values that are too high (they should go in the last bucket)?

components/service/glean/src/test/java/mozilla/components/service/glean/TimingDistributionMetricTypeTest.kt
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    @Test
+    fun `The API must define the expected "default" storage`() {
+        // Define a timing distribution metric which will be stored in "store1"
+        val metric = TimingDistributionMetricType(
+            disabled = false,
+            category = "telemetry",
+            lifetime = Lifetime.Ping,
+            name = "string_list_metric",
```
Dexterp37: nit: this is not a string list

components/service/glean/src/test/java/mozilla/components/service/glean/TimingDistributionMetricTypeTest.kt
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    @Test
+    fun `The API saves to its storage engine`() {
+        // Define a timing distribution metric which will be stored in "store1"
+        val metric = TimingDistributionMetricType(
+            disabled = false,
+            category = "telemetry",
+            lifetime = Lifetime.Ping,
+            name = "string_list_metric",
```
Dexterp37: nit: this is not a string list

components/service/glean/src/test/java/mozilla/components/service/glean/TimingDistributionMetricTypeTest.kt
```diff
@@ -0,0 +1,139 @@
...
+
+    @Test
+    fun `disabled timing distributions must not record data`() {
+        // Define a timing distribution metric which will be stored in "store1"
+        // It's lifetime is set to Lifetime.Ping so it should not record anything.
+        val metric = TimingDistributionMetricType(
+            disabled = true,
+            category = "telemetry",
+            lifetime = Lifetime.Ping,
+            name = "string_list_metric",
```
Dexterp37: nit: this is not a string list

components/service/glean/src/test/java/mozilla/components/service/glean/TimingDistributionMetricTypeTest.kt
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    @Test(expected = NullPointerException::class)
+    fun `testGetValue() throws NullPointerException if nothing is stored`() {
+        // Define a timing distribution metric which will be stored in "store1"
+        val metric = TimingDistributionMetricType(
+            disabled = false,
+            category = "telemetry",
+            lifetime = Lifetime.Ping,
+            name = "string_list_metric",
```
Dexterp37: nit: this is not a string list

components/service/glean/src/test/java/mozilla/components/service/glean/TimingDistributionMetricTypeTest.kt
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    @Test
+    fun `The API saves to secondary pings`() {
+        // Define a timing distribution metric which will be stored in "store1" and "store2"
+        val metric = TimingDistributionMetricType(
+            disabled = false,
+            category = "telemetry",
+            lifetime = Lifetime.Ping,
+            name = "string_list_metric",
```
Dexterp37: nit: this is not a string list

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
```diff
@@ -0,0 +1,243 @@
...
+         */
+        fun fromJsonString(json: String): TimingDistributionData? {
+            val jsonObject: JSONObject
+            try {
+                jsonObject = JSONObject(json)
+            } catch (e: org.json.JSONException) {
+                return null
+            }
+
+            return TimingDistributionData(
```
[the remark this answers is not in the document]
mdboom: Yeah, I think I agree.  Normally I'd say "be liberal in what you accept", but since this is data our own library has written to disk, I think we should be more skeptical and throw data away that looks fishy, and maybe report an error metric.  (We might need a new error metric type for this to indicate "corrupted data" or something).

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
```diff
@@ -0,0 +1,277 @@
...
+    // bucket counts are set first.
+    internal val buckets: List<Long> by lazy { getBuckets() }
+
+    /**
+     * Accumulates a sample to the correct bucket
+     *
+     * @param sample Long value representing the sample that is being accumulated
+     */
+    internal fun accumulate(sample: Long) {
+        for (i in buckets.indices) {
```
mdboom: I'm not sure you need the `buckets` list and to loop through it.  Couldn't you just do something like (pseudo code, because I don't know the specifics of Kotlin math)...

```
val index = Math.floor(Math.log(sample)) as Int
values["$index"] += 1
```

I don't know if natural log or log base 10 is more what people want or are used to.  But my point is just that there should be a mathematical function from value to bucket.  And this also creates the "infinite buckets" thing that was talked about on slack.

travis79: I much prefer this method.  It's going to have overhead Math overhead for every accumulation but probably no more than the linear search.  This may also help with another issue I've been fighting.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
travis79: They were defined in the schema so I added them here, I'll remove them since they aren't currently being used and are listed as 'optional' in the schema definition.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
travis79: As much as I like this approach, in the interest of MVP I'm going to stick with the search approach rather than the calculation approach.  I was struggling to try and account for the min/max range and bucket counts effect on the bucket sizes, so it would be a good iteration when we add infinite buckets feature.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
mdboom: I think they are supposed to be buckets containing out of range values.

For timings specifically (as opposed to a general histogram), they are always positive, so one could argue all negative values should go in `underflow`.  Or we just provide an API that makes that impossible, or we treat those as errors through the error reporting mechanism.

If we use the proposed "inifinite buckets" where we don't really have an upper bound (which works pretty nicely for exponential histograms, probably less so for linear), than `overflow` would be impossible.

tl;dr: I'm fine with removing these, remembering that we'll probably want them back when we implement general-purpose histograms...

components/service/glean/src/main/java/mozilla/components/service/glean/TimingDistributionMetricType.kt
```diff
@@ -0,0 +1,94 @@
...
+    // Holds the Job returned from launch{} for awaiting purposes
+    @Volatile
+    private var ioTask: Job? = null
+
+    /**
+     * Accumulates a sample value to the timing distribution.
+     *
+     * @param sample This is the sample to accumulate
+     */
+    fun accumulate(sample: Long) {
```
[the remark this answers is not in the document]
travis79: I'm okay with that, but let look at that as follow up work.  I'll file a bug if this lands and I think we may want to involve @georgf and get his opinion on the API surface as that's a 'producty' type question.

components/service/glean/docs/metrics/adding-new-metrics.md
```diff
@@ -108,3 +109,54 @@ assertEquals(6, Controls.refreshPressed.testGetValue())
...
+    type: timing_distribution
+    time_unit: millisecond
+    description: >
+      Counts how long each page takes to load into exponential buckets
+    ...
+```
+
+Note that `time_unit` is required in the `metrics.yaml` file to identify the base unit of time
+used to indicate the size of the buckets and the units intended when accumulating and recording
+samples.
```
georgf: Can we avoid mentioning buckets?
E.g. *"`time_unit` is required and allows for values like `millisecond` and ... Which range of values is recorded in detail depends on the `time_unit`, e.g. for `milliseconds` all values greater `60000` are recorded as overflow values.*
(Not sure i got the values right)

components/service/glean/docs/metrics/adding-new-metrics.md
```diff
@@ -108,3 +109,54 @@ assertEquals(6, Controls.refreshPressed.testGetValue())
...
+If you wanted to create a timing distribution to measure page load times, first you need to add 
+an entry for the counter to the `metrics.yaml` file:
+
+```YAML
+pages:
+  page_load:
+    type: timing_distribution
+    time_unit: millisecond
+    description: >
+      Counts how long each page takes to load into exponential buckets
```
georgf: <s>into exponential buckets</s>?

components/service/glean/docs/metrics/adding-new-metrics.md
```diff
@@ -108,3 +109,54 @@ assertEquals(6, Controls.refreshPressed.testGetValue())
...
+Now you can use the event from the applications code:
+```Kotlin
+import org.mozilla.yourApplication.GleanMetrics.Pages
+
+// ...
+pages.pageLoad.accumulate(1L) // Accumulates a sample of 1 millisecond
+pages.pageLoad.accumulate(10L) // Accumulates a sample of 10 milliseconds
+```
+
+At this point, the metric would have a `sum` of `11` and a `values` map of `{ "0" : 1, "8" : 1}`
```
georgf: Maybe don't mention the `values` here. I don't think that adds... value.

components/service/glean/docs/metrics/adding-new-metrics.md
```diff
@@ -108,3 +109,54 @@ assertEquals(6, Controls.refreshPressed.testGetValue())
...
+import org.mozilla.yourApplication.GleanMetrics.Pages
+
+// Was anything recorded?
+assertTrue(pages.pageLoad.testHasValue())
+// Does the sum have the expected value?
+val tdSnapshot = pages.pageLoad.testGetValue()
+assertEquals(11, tdSnapshot?.sum)
+assertEquals(1L, tdSnapshot?.values["0"]) // The bucket the '1L' was accumulated to
+assertEquals(1L, tdSnapshot?.values["8"]) // The bucket the '10L' was accumulated to
+assertNull(tdSnapshot?.values["1"]) // Should be null since nothing was accumulated to this bucket
```
georgf: I think we should avoid test author having to look up the buckets.
E.g. for timings in product, it's hard to test exact values. Instead just checking *N values were recorded* seems a test pattern engineers use.
From [this search](https://searchfox.org/mozilla-central/search?q=gethistogram.%2Bms&case=false&regexp=true&path=test) i found only `sum` and `count` checks:
- [example 1](https://searchfox.org/mozilla-central/rev/7c20ad925005fbad7b8e08813115f1ec7fa1c248/browser/base/content/test/tabs/browser_tabCloseProbes.js#79)
- [example 2](https://searchfox.org/mozilla-central/rev/7c20ad925005fbad7b8e08813115f1ec7fa1c248/dom/webauthn/tests/browser/browser_webauthn_telemetry.js#150)).
- ... etc.

So how about just exposing a `testValueCount`/`testSampleCount` or so?
And just using that for the example?

components/service/glean/docs/metrics/adding-new-metrics.md
```diff
@@ -108,3 +109,54 @@ assertEquals(6, Controls.refreshPressed.testGetValue())
...
 NOTE: Using the testing API requires calling `Glean.enableTestingMode()` first,
 such as from a `@Before` method.
+
+## Timing Distributions
+
+Used to accumulate and store time measurement samples into buckets, which are exponentially sized 
+using a natural log function. 
+
+If you wanted to create a timing distribution to measure page load times, first you need to add 
+an entry for the counter to the `metrics.yaml` file:
```
georgf: <s>the counter</s>?

components/service/glean/docs/metrics/adding-new-metrics.md
```diff
@@ -108,3 +109,54 @@ assertEquals(6, Controls.refreshPressed.testGetValue())
...
+    description: >
+      Counts how long each page takes to load into exponential buckets
+    ...
+```
+
+Note that `time_unit` is required in the `metrics.yaml` file to identify the base unit of time
+used to indicate the size of the buckets and the units intended when accumulating and recording
+samples.
+
+Now you can use the event from the applications code:
```
georgf: <s>the event</s>

components/service/glean/docs/metrics/adding-new-metrics.md
georgf: I can find more examples [with this search](https://searchfox.org/mozilla-central/search?q=_MS%22&case=true&regexp=true&path=test), but the previous pattern holds.

components/service/glean/docs/metrics/adding-new-metrics.md
```diff
@@ -21,6 +21,7 @@ There are different metrics to choose from, depending on what you want to achiev
 * [Events](#Events): These allow recording of e.g. individual occurences of user actions, say every time a view was open and from where.
 * [Counters](#Counters): Used to count how often something happens, say how often a certain button was pressed.
 * [Labeled metrics](#labeled-metrics): Used to record multiple metrics of the same type under different string labels, say every time you want to get a count of different error types in one metric.
+* [Timing Distributions](#Timing-Distributions): Used to accumulate and count time measurements into buckets.
```
georgf: Let's drop the bucket term please? It doesn't seem important to users of the metric here.
s/count time measurements into buckets/count multiple time measurements/

components/service/glean/docs/metrics/adding-new-metrics.md
```diff
@@ -155,3 +154,50 @@ assertTrue(Updater.loadError["not_found"].testHasValue())
...
+```Kotlin
+import org.mozilla.yourApplication.GleanMetrics.Pages
+Glean.enableTestingMode()
+
+// Was anything recorded?
+assertTrue(pages.pageLoad.testHasValue())
+// Does the sum have the expected value?
+val tdSnapshot = pages.pageLoad.testGetValue()
+assertEquals(11, tdSnapshot?.sum)
+assertEquals(2L, tdSnapshot?.count())
```
georgf: Optional: Maybe move this first & add:
```kotlin
// Usually you don't know the exact timing values, but how many should have been recorded.
```

components/service/glean/src/test/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngineTest.kt
```diff
@@ -0,0 +1,340 @@
...
+    @Test
+    fun `getBuckets() correctly populates the buckets property`() {
+        // Hand calculated values using current default range 0 - 60000 and bucket count of 100
+        val testBuckets: List<Long> = listOf(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 17,
+            19, 21, 23, 25, 28, 31, 34, 38, 42, 46, 51, 56, 62, 68, 75, 83, 92, 101, 111, 122, 135,
+            149, 164, 181, 200, 221, 244, 269, 297, 328, 362, 399, 440, 485, 535, 590, 651, 718,
+            792, 874, 964, 1064, 1174, 1295, 1429, 1577, 1740, 1920, 2118, 2337, 2579, 2846, 3140,
+            3464, 3822, 4217, 4653, 5134, 5665, 6250, 6896, 7609, 8395, 9262, 10219, 11275, 12440,
+            13726, 15144, 16709, 18436, 20341, 22443, 24762, 27321, 30144, 33259, 36696, 40488,
+            44672, 49288, 54381, 60000, 60001)
```
georgf: The last bucket with a width of 1 seems odd.
Could this just end with `..., 54381, 6000)` or `..., 54381, 6001)`?

georgf: It might also just be fine to file a follow-up about the bucket fine-tuning for the next iteration.

travis79: That last bucket is actually just the overflow bucket.  It's size of `60001` is just an arbitrary side effect of the algorithm that generates the buckets (which I copied from desktop, so I hope it's right)

georgf: Sure, the overflow is fine. The odd part is the 1-width bucket of `6000`.

If you look at [visualizations](https://mzl.la/2YpTGoK) of histogram data, a 1-sized bucket will be hard to make sense of for analysis, it doesn't fit the exponential bucket pattern.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
travis79: I'll try and explain it, there wasn't any documentation or good comments describing this from the desktop code where it was borrowed from either.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
```diff
@@ -0,0 +1,289 @@
...
+     * bucket where the sample is less than or equal to the bucket value.  This works since the
+     * buckets are sorted in ascending order.  If a mapped value doesn't exist for this bucket yet,
+     * one is created.
+     *
+     * @param sample Long value representing the sample that is being accumulated
+     */
+    internal fun accumulate(sample: Long) {
+        for (i in buckets.indices) {
+            if (sample <= buckets[i] || i == bucketCount - 1) {
+                values["$i"] = (values["$i"] ?: 0) + 1
```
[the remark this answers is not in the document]
travis79: I think I was copying what I saw in the existing `values` for histograms.  The index in the current desktop implementation appears to be a string, or at least surrounded by "" in the JSON body that is transmitted.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
```diff
@@ -0,0 +1,289 @@
...
+ * @param timeUnit the base [TimeUnit] of the bucket values
+ */
+@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+data class TimingDistributionData(
+    val category: String,
+    val name: String,
+    val bucketCount: Int = DEFAULT_BUCKET_COUNT,
+    val range: List<Long> = listOf(DEFAULT_RANGE_MIN, DEFAULT_RANGE_MAX),
+    val histogramType: HistogramType = HistogramType.Exponential,
+    val values: MutableMap<String, Long> = mutableMapOf(),
```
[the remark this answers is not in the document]
travis79: Again, trying to maintain compatibility with the existing `values`, as this object was meant to be as close to the JSON object that would be passed around.  When sent to the pipeline it's expecting a map.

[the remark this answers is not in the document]
travis79: Also, how would I know what the `index` of the bucket was that the accumulator goes with without the map?  Do I keep a separate list of those and try and keep the orders in sync?  I think that map is the appropriate type here, and we don't know that string building overhead is a performance issue here, so I hate to prematurely optimize this.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
```diff
@@ -0,0 +1,289 @@
...
+ * @param values a map containing the bucket index mapped to the accumulated count
+ * @param sum the accumulated sum of all the samples in the timing distribution
+ * @param timeUnit the base [TimeUnit] of the bucket values
+ */
+@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+data class TimingDistributionData(
+    val category: String,
+    val name: String,
+    val bucketCount: Int = DEFAULT_BUCKET_COUNT,
+    val range: List<Long> = listOf(DEFAULT_RANGE_MIN, DEFAULT_RANGE_MAX),
```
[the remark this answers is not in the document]
travis79: Filed a bug to look at this as follow up work

components/service/glean/src/test/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngineTest.kt
travis79: I hesitate to re-design the algo that I was told was used on both the client and pipeline side (the bucket sizing algo), so @Dexterp37 any thoughts here?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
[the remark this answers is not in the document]
Dexterp37: Yes, that is true for the transmitted payload (i.e. serialized). Internally, however, numerical indices are being used.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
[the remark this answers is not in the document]
Dexterp37: As far as I can tell, but please correct me if I'm understanding it the wrong way, buckets are sorted and indices are just natural numbers `>=0` (1, 2, 3, 4, 5...?)? So the position in the array can be used as index?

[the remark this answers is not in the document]
travis79: Yes, but that's not very space conservative, is it?  If I accumulate a value into the last bucket, then I need to fill in all those 0's before it.  So...   if there are 1000 buckets, suddenly I've got a list that's 8KB in size and I only care about 8 bytes of it?  The idea of sparse storage made sense here, for the same reason it makes sense when transmitting the data to the pipeline.

travis79:
@georgf The changes that need approved by you are all in the [adding-new-metrics.md](https://github.com/mozilla-mobile/android-components/pull/2357/files/12b5571f4c993579a15cb4f56e51e626e8a1d632#diff-ac0dc531ccacbebb1bff5a7fc126772e)

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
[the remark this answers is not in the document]
Dexterp37: Ah, I see what you mean. We could still use a `Map<Unsigned Int, Long>` , but still getting rid of strings as indices. Let's file  a bug about that, no need to care about this in this PR :)

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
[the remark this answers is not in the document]
travis79: Looking at this for follow up work, bug filed [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1538745)

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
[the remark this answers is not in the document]
travis79: Looking at this for follow up work, bug filed [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1538745)

components/service/glean/src/test/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngineTest.kt
travis79: Making a [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1538750) to come back to this and discuss it more thoroughly and validate the bucket generation algorithm

components/service/glean/docs/metrics/adding-new-metrics.md
```diff
@@ -155,3 +154,55 @@ assertTrue(Updater.loadError["not_found"].testHasValue())
...
+at this point the metric should have a `sum == 11` and a `count() == 2`:
+```Kotlin
+import org.mozilla.yourApplication.GleanMetrics.Pages
+Glean.enableTestingMode()
+
+// Was anything recorded?
+assertTrue(pages.pageLoad.testHasValue())
+
+// Get snapshot
+val tdSnapshot = pages.pageLoad.testGetValue()
```
georgf: Nit: For the sake of the examples clarity, just `snapshot`?

components/service/glean/docs/metrics/adding-new-metrics.md
```diff
@@ -155,3 +154,55 @@ assertTrue(Updater.loadError["not_found"].testHasValue())
...
+import org.mozilla.yourApplication.GleanMetrics.Pages
+
+// ...
+pages.pageLoad.accumulate(1L) // Accumulates a sample of 1 millisecond
+pages.pageLoad.accumulate(10L) // Accumulates a sample of 10 milliseconds
+```
+
+There are test APIs available too.  For convenience, properties `sum` and `count` are exposed to 
+facilitate validating that data was recorded correctly.  Continuing the `pageLoad` example above,
+at this point the metric should have a `sum == 11` and a `count() == 2`:
```
georgf: Why does `sum` have property syntax & `count()` function syntax.
Please change the API to use the same approach for both.

components/service/glean/docs/metrics/adding-new-metrics.md
```diff
@@ -155,3 +154,55 @@ assertTrue(Updater.loadError["not_found"].testHasValue())
...
+  page_load:
+    type: timing_distribution
+    time_unit: millisecond
+    description: >
+      Counts how long each page takes to load
+    ...
+```
+
+Note that `time_unit` is required and allows for values like `millisecond` and `second`. Which range 
+of values is recorded in detail depends on the time_unit, e.g. for milliseconds all values greater
```
georgf: Nit: missing ticks around `time_unit`

components/service/glean/docs/metrics/adding-new-metrics.md
```diff
@@ -155,3 +154,55 @@ assertTrue(Updater.loadError["not_found"].testHasValue())
...
+Glean.enableTestingMode()
+
+// Was anything recorded?
+assertTrue(pages.pageLoad.testHasValue())
+
+// Get snapshot
+val tdSnapshot = pages.pageLoad.testGetValue()
+
+// Does the sum have the expected value?
+assertEquals(11, tdSnapshot?.sum)
```
georgf: Nit: can we avoid the `?` here?

components/service/glean/docs/metrics/adding-new-metrics.md
travis79: Good point, I just needed to move this to a `get()` in the property so that it was a calculated property instead.

PR #2363, opened 2019-03-15
Closes #2361: add HttpHeaders constants

### LINKED ISSUE #2361: Implement constants for shared HTTP headers
It'd be nice if apps could share a single HTTP header constant definition with values like, "Content-Type" and "Authorization". A list of common HTTP request headers can be found at: https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#Standard_request_fields

@pocmo What's a good place for this? I'd think:
- `support-base`
- `support-utils`
- A new HTTP abstraction module

I'm leaning towards base because it's a "building block". I think it should be separate from `concept-fetch` because, in theory, some consumers may not want to use our abstraction but might want pre-defined headers.

mcomella:
### Pull Request checklist

  - I don't know what to test: they're string constants. To make sure no one accidentally changes them and breaks _all_ the dependent apps, we could add sanity check tests like, "CONTENT_TYPE" includes "Content-Type".


mcomella:
btw, I didn't add move all the existing content-type constants over because it's the end of my day and I ran out of time (it's a lot of work though too, to add the "support-base" dependencies, add the import statement, and replace the text).

components/support/base/src/main/java/mozilla/components/support/base/http/HttpHeaders.kt
```diff
@@ -0,0 +1,19 @@
...
+ * A list of common HTTP request headers can be found at
+ *   https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#Standard_request_fields
+ *
+ * A list of common HTTP response headers can be found at
+ *   https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#Standard_response_fields
+ */
+object HttpHeaders {
+
+    const val CONTENT_TYPE = "Content-Type"
+}
```
pocmo: I like the idea of those constants. But wouldn't we want to move them into `concept-fetch`?

mcomella: I discussed this (briefly) in the issue: https://github.com/mozilla-mobile/android-components/issues/2361#issue-421299227

mcomella:
@pocmo I wasn't sure where to put the headers. Now that there is a `Headers` class already, it felt weird adding another one but with `Http` prefixing it, it sounds more confusing. I opted for `Headers.Common` but let me know if it should move elsewhere.

mcomella:
@pocmo Another consideration we might have: if we want to include values for both common header names but also values (e.g. Content-Type header might have a value of `application/x-www-form-urlencoded`).

mcomella:
I'm not sure how to fit values (e.g. ^) into the current model.

pocmo:
@pocmo Another consideration we might have: if we want to include values for both common header names but also values (e.g. Content-Type header might have a value of application/x-www-form-urlencoded).

🤷‍♂️ 

Maybe on the `Body` since that is what has the type? :)

mcomella:
Glean intermittent failure: rerunning.

mcomella:
> Maybe on the `Body` since that is what has the type? :)

@pocmo `Request.Body`? But this is unrelated to the request: I'm referring to the "Header-Value" i n "Header-Name: Header-Value". Maybe `Headers.Common` and `Headers.Common.Values`?

It could also be `Headers.Names` and `Headers.Values`... 🤔

PR #2370, opened 2019-03-15
Closes #2207: Adding helper to detect if we are running in the main process...

### LINKED ISSUE #2207: Add helper to detect if we are running in the main process
It's ridiculously complicated to do that. Let's add something to `support-utils` for that.

Amejia481:
### Pull Request checklist

### DIFF
--- a/components/support/ktx/src/main/java/mozilla/components/support/ktx/android/content/Context.kt
+++ b/components/support/ktx/src/main/java/mozilla/components/support/ktx/android/content/Context.kt
@@ -13,6 +13,8 @@ import android.content.Intent.EXTRA_SUBJECT
 import android.content.Intent.EXTRA_TEXT
 import android.content.Intent.FLAG_ACTIVITY_NEW_TASK
 import android.content.pm.PackageManager.PERMISSION_GRANTED
+import android.os.Process
+import android.support.annotation.VisibleForTesting
 import android.support.v4.content.ContextCompat.checkSelfPermission
 import mozilla.components.support.base.log.Log
 import mozilla.components.support.ktx.R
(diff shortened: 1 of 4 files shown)


components/support/ktx/src/main/java/mozilla/components/support/ktx/android/content/Context.kt
```diff
@@ -79,3 +80,25 @@ fun Context.share(text: String, subject: String = getString(R.string.mozac_suppo
         false
     }
 }
+
+/**
+ *  Returns true if we are running in the main process false otherwise.
+ */
+fun Context.isMainProcess(): Boolean {
```
Amejia481: @pocmo 
My initial idea was to cache the value as we chatted, but there some [limitations with lazy ](https://youtrack.jetbrains.com/issue/KT-9686) [[2]](https://discuss.kotlinlang.org/t/lazy-extension-properties/2351) that make it tricky to do: 
```
val Context.isMainProcess : Boolean by lazy {
    // I can't access Context@this here.
}
```

components/support/ktx/src/main/java/mozilla/components/support/ktx/android/content/Context.kt
```diff
@@ -79,3 +80,25 @@ fun Context.share(text: String, subject: String = getString(R.string.mozac_suppo
         false
     }
 }
+
```
pocmo: You could add a `var isMainProcess: Boolean? = null` here and only run the block in `isMainProcess()` once (if it is null).

Amejia481: I was a bit afraid to do that, as `var isMainProcess: Boolean? = null` will be static and shared but two processes will have different heaps , and that will be OK. I updated the PR we the change!

Amejia481:
👍

PR #2372, opened 2019-03-15
Closes #2368: SearchSuggestionProvider: Allow setting a limit.

### LINKED ISSUE #2368: SearchSuggestionProvider: Allow setting limit
Currently we just display all the suggestions we get from a search engine. This should be configurable.

pocmo:
### Pull Request checklist

### DIFF
--- a/components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/SearchSuggestionProvider.kt
+++ b/components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/SearchSuggestionProvider.kt
@@ -19,11 +19,18 @@ import java.util.concurrent.TimeUnit
 /**
  * A [AwesomeBar.SuggestionProvider] implementation that provides a suggestion containing search engine suggestions (as
  * chips) from the passed in [SearchEngine].
+ *
+ * @param searchEngine The search engine to request suggestions from.
+ * @param searchUseCase The use case to invoke for searches.
+ * @param fetchClient The HTTP client for requesting suggestions from the search engine.
+ * @param limit The maximum number of suggestions that should be returned.
+ * @param mode Whether to return a single search suggestion (with chips) or one suggestion per item.
  */
 class SearchSuggestionProvider(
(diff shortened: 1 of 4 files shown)


components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/SearchSuggestionProvider.kt
```diff
@@ -106,7 +113,7 @@ class SearchSuggestionProvider(
             chips.add(AwesomeBar.Suggestion.Chip(text))
         }
 
-        result?.forEach { title ->
+        result?.take(limit - chips.size)?. forEach { title ->
```
Amejia481: Could be possible that `chips.size` is bigger than `limit`? If it possible will produce a negative number and cause an error.

pocmo: At this point `chips` either contains no items (0) or the default item (1) - the other items will be added in this loop. The only way it could be bigger is if someone chooses a limit that is 0 (and the default item is already added) or a negative number. Possible - but an error case nevertheless. I guess we could add a check that a limit is a positive number and at least 1. I'll file an issue.

pocmo: At this point `chips` either contains no items (0) or the default item (1) - the other items will be added in this loop. The only way it could be bigger is if someone chooses a limit that is 0 (and the default item is already added) or a negative number. Possible - but an error case nevertheless. I guess we could add a check that a limit is a positive number and at least 1. I'll file an issue.

pocmo: #2432

Amejia481: 👍

PR #2376, opened 2019-03-15
Update A-S dependencies to 0.20.2

grigoryk:
Marking as DRAFT because there are currently problems with publication of 0.20.2, and this PR can't be merged until they're resolved.


grigoryk:
This is blocked on release of 0.2 version of org.mozilla.android gradle plugin, which includes some necessary changes that make it possible to use 0.20.2 appservices libraries. See https://github.com/mozilla-mobile/android-automation-tools/pull/45

grigoryk:
Updated plugin was released, so this is no longer blocked.

grigoryk:
TC tasks keep hitting intermittent failures. :(

PR #2377, opened 2019-03-15
Make the experiments api static       

georgf:
- Make the experiments API a static API
- Make all functions except essential ones internal
- Refactor accordingly

### Pull Request checklist

### DIFF
--- a/components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
+++ b/components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
@@ -6,6 +6,10 @@ package mozilla.components.service.experiments
 
 import android.content.Context
 import mozilla.components.support.base.log.logger.Logger
+import android.support.annotation.VisibleForTesting
+import mozilla.components.concept.fetch.Client
+import mozilla.components.lib.fetch.httpurlconnection.HttpURLConnectionClient
+import java.io.File
 
 /**
  * Entry point of the library
@@ -15,29 +19,96 @@ import mozilla.components.support.base.log.logger.Logger
  * @param valuesProvider provider for the device's values
  */
 @Suppress("TooManyFunctions")
-class Experiments(
-    private val source: ExperimentSource,
-    private val storage: ExperimentStorage,
-    valuesProvider: ValuesProvider = ValuesProvider()
-) {
+open class ExperimentsInternalAPI internal constructor() {
+    private val logger: Logger = Logger(LOG_TAG)
+
     @Volatile private var experimentsResult: ExperimentsSnapshot = ExperimentsSnapshot(listOf(), null)
     private var experimentsLoaded: Boolean = false
-    private val evaluator = ExperimentEvaluator(valuesProvider)
-    private val logger = Logger(LOG_TAG)
+    private var evaluator: ExperimentEvaluator = ExperimentEvaluator(ValuesProvider())
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal var valuesProvider: ValuesProvider = ValuesProvider()
+        set(provider) {
+            field = provider
+            evaluator = ExperimentEvaluator(field)
+        }
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal var source: ExperimentSource? = null
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal var storage: ExperimentStorage? = null
+
+    private var isInitialized = false
+
+    /**
+     * Initialize the experiments library.
+     *
+     * This should only be initialized once by the application.
+     *
+     * @param applicationContext [Context] to access application features, such
+     * as shared preferences.
+     */
+    fun initialize(
+        applicationContext: Context,
+        fetchClient: Client = HttpURLConnectionClient()
+    ) {
+        if (isInitialized) {
+            logger.error("Experiments library should not be initialized multiple times")
+            return
+        }
+
+        experimentsResult = ExperimentsSnapshot(listOf(), null)
+        experimentsLoaded = false
+
+        if (source == null) {
+            source = KintoExperimentSource(
+                EXPERIMENTS_BASE_URL,
+                EXPERIMENTS_BUCKET_NAME,
+                EXPERIMENTS_COLLECTION_NAME,
+                fetchClient
+            )
+        }
+        if (storage == null) {
+            storage = FlatFileExperimentStorage(
+                File(applicationContext.getDir(EXPERIMENTS_DATA_DIR, Context.MODE_PRIVATE),
+                    EXPERIMENTS_JSON_FILENAME)
+            )
+        }
+
+        isInitialized = true
+    }
+
+    /**
+     * Reset the library in tests.
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    internal fun testReset() {
+        experimentsResult = ExperimentsSnapshot(listOf(), null)
+        experimentsLoaded = false
+        evaluator = ExperimentEvaluator(ValuesProvider())
+        valuesProvider = ValuesProvider()
+
+        source = null
+        storage = null
+
+        isInitialized = false
+    }
 
     /**
      * Provides the list of experiments (active or not)
      */
-    val experiments: List<Experiment>
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal val experiments: List<Experiment>
         get() = experimentsResult.experiments.toList()
 
     /**
      * Loads experiments from local storage
      */
     @Synchronized
-    fun loadExperiments() {
-        experimentsResult = storage.retrieve()
-        experimentsLoaded = true
+    internal fun loadExperiments() {
+        storage?.let {
+            experimentsResult = it.retrieve()
+            experimentsLoaded = true
+        } ?: logger.error("No storage specified")
     }
 
     /**
@@ -45,14 +116,17 @@ class Experiments(
      * saves them to local storage
      */
     @Synchronized
-    fun updateExperiments(): Boolean {
+    internal fun updateExperiments(): Boolean {
         if (!experimentsLoaded) {
             loadExperiments()
         }
+
         return try {
-            val serverExperiments = source.getExperiments(experimentsResult)
-            experimentsResult = serverExperiments
-            storage.save(serverExperiments)
+            source?.let {
+                val serverExperiments = it.getExperiments(experimentsResult)
+                experimentsResult = serverExperiments
+                storage?.save(serverExperiments)
+            } ?: logger.error("No source available")
             true
         } catch (e: ExperimentDownloadException) {
             // Keep using the local experiments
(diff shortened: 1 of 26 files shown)


georgf:
I'd like a first look over this. Local tests are passing, but i might violate some practices currently.

This PR has the goal of making the Experiments API static and reducing the public API surface to just what's needed. Tests are adjusted accordingly.
Subsequent PRs will take care of other work.

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
```diff
@@ -15,28 +21,91 @@ import mozilla.components.support.base.log.logger.Logger
...
+        source = aSource
+    }
+
+    /**
+     * Override the experiment storage for testing.
+     *
+     * @param aStorage the experiment storage to use
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun overrideExperimentStorage(aStorage: ExperimentStorage) {
```
Dexterp37: Instead of introducing these two functions, you could also just make `storage` an `internal` and `@VisibleForTesting` variable.

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
```diff
@@ -15,28 +21,91 @@ import mozilla.components.support.base.log.logger.Logger
...
+        storage = aStorage
+    }
+
+    /**
+     * Override the values provider for testing.
+     *
+     * @param valuesProvider the values provider to use
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun overrideValuesProvider(valuesProvider: ValuesProvider) {
```
Dexterp37: As above.

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
```diff
@@ -15,28 +21,91 @@ import mozilla.components.support.base.log.logger.Logger
...
 
     /**
      * Loads experiments from local storage
      */
     @Synchronized
-    fun loadExperiments() {
-        experimentsResult = storage.retrieve()
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun loadExperiments() {
+        experimentsResult = storage!!.retrieve()
```
Dexterp37: The `!!` here throws an exception if `storage` is `null` (I understand this should never happen outside of tests, unless the API is called before `initialize`). If we want the API to not ever throw exception (if that's a concern), we might consider refactoring this to something like:

```
storage?.let {
  experimentsResult = it.retrieve()
  experimentsLoaded = true
} ?: logger.error("No storage specified")
```

Or something like that.

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
```diff
@@ -45,14 +114,15 @@ class Experiments(
...
     @Synchronized
-    fun updateExperiments(): Boolean {
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun updateExperiments(): Boolean {
         if (!experimentsLoaded) {
             loadExperiments()
         }
         return try {
-            val serverExperiments = source.getExperiments(experimentsResult)
+            val serverExperiments = source!!.getExperiments(experimentsResult)
```
Dexterp37: Note that `!!` here will crash if `source = null`, and exception won't be trapped by the `catch` below. You might want to consider doing this:

```
source?.let {
  val serverExperiments = it.getExperiments(experimentsResult)
  experimentResult = serverExperiments
  storage?.save(serverExperiments)
} ?: logger.error("No source available")
```

This way the only throw due to `getExperiments` or, potentially, `save`.

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsTest.kt
```diff
@@ -30,25 +31,36 @@ import java.io.File
...
+
 @RunWith(RobolectricTestRunner::class)
 class ExperimentsTest {
     @Test
     fun loadExperiments() {
         val experimentSource = mock(ExperimentSource::class.java)
         val experimentStorage = mock(ExperimentStorage::class.java)
-        val instance = Experiments(experimentSource, experimentStorage)
-        instance.loadExperiments()
+        resetExperiments(experimentSource, experimentStorage)
```
Dexterp37: I think `resetExperiments` and the related mocks could be moved to a `setup` function called before each test run. See [this](https://github.com/mozilla-mobile/android-components/blob/51ec2c968738b75786ee76c1e645819e4aa5e976/components/service/glean/src/test/java/mozilla/components/service/glean/storages/PingStorageEngineTest.kt#L46-L48). From the tests you could directly access the mocks via `Experiments.source` and `Experiments.storage`. This would make the tests *much* shorter and clearer, imho.

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
```diff
@@ -66,34 +136,35 @@ class Experiments(
...
      *
      * @param context context
-     * @param descriptor descriptor of the experiment to check
+     * @param experimentId the id of the experiment
      * @param block block of code to be executed if the user is part of the experiment
      */
-    fun withExperiment(context: Context, descriptor: ExperimentDescriptor, block: (Experiment) -> Unit) {
-        evaluator.evaluate(context, descriptor, experimentsResult.experiments)?.let { block(it) }
+    fun withExperiment(context: Context, experimentId: String, block: () -> Unit) {
+        evaluator.evaluate(context, ExperimentDescriptor(experimentId), experimentsResult.experiments)?.let { block() }
```
travis79: Should we run this `block()` in it's own thread?  I mean, this could potentially do IO or some other time/resource intensive thing that we might not want to bog down the current thread.

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
```diff
@@ -15,28 +21,91 @@ import mozilla.components.support.base.log.logger.Logger
...
+     * as shared preferences.
+     */
+    fun initialize(
+        applicationContext: Context
+    ) {
+        experimentsResult = ExperimentsSnapshot(listOf(), null)
+        experimentsLoaded = false
+        logger = Logger(LOG_TAG)
+
+        if (source == null) {
```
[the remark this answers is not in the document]
mdboom: I think it's always helpful to behave sanely (if possible) in the face of calling `initialize` twice.  In that case, we should maybe try to do nothing and log a warning if `source != null`.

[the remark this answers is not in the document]
georgf: Currently `initialize()` is called multiple times from tests (and i tried to work around some kinks in testing).
I could use a different pattern though, like a test-only `reset()` function or so?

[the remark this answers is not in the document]
mdboom: That could work.  That's the pattern used in glean.

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
georgf: That's an interesting question.
I'd say this is for the product engineers to decide.
I assume this needs to behave synchronously by default, so, as a product engineer, i can write code as if it was just a logic branch in my function.
If async is needed, they can decide by themselves.

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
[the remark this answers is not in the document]
Dexterp37: Yup. Moreover, that would have the nice side-effect of having a developer friendly error log to explain what's going on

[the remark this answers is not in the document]
georgf: Ok, so, my challenge is:

**(1)**: `source` & `storage` depend on parameters passed to `initialize()`, so i should set them from there for normal library initialization.

**(2)** I want to enable overriding `source` & `storage` in tests while also allowing repeated `initialize()` calls, e.g.:

```kotlin
test() {
  Experiments.source = FooExperimentSource()
  Experiments.initialize()
}
```

**(3)** I don't want to first set `source` & `storage` from `initialize()`, then override them after, because that will trigger actions later that the test does not expect, i.e.:

```kotlin
test() {
  Experiments.initialize()
  // Oops, starts async operation with default `source` already.
  Experiments.source = FooExperimentSource()
}
```

**4** I want to keep `source` & `storage` internal so i can't expose them as optional parameters to `initialize()`?

```kotlin
test() {
  Experiments.initialize(FooExperimentSource())
  // oops, parameter type can't be internal anymore?
}
```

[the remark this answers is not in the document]
georgf: That's what i currently work around with the `null` checks:
```kotlin
test() {
  Experiments.source = FooExperimentSource()
  Experiments.initialize()
  // ok, `source` is non-null, so we won't use the default source.
}

normalProduct() {
  Experiments.initialize()
  // ok, `source` is null, so we will use the default source.
}
```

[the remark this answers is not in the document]
georgf: Am i missing a cleaner option here?

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
travis79: That makes sense... we expect to execute this synchronously, so any async requirements would be the responsibility of the caller and handled withing the `block` closure.  That answers my question.

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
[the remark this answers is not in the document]
travis79: You could create a test only initialize function with the `@VisibleForTesting` annotation that allows you to inject the source and storage but wouldn't be available to the public API, something like this:
```
internal fun initialize(source: ExperimentSource?, storage: ExperimentStorage?) {
     ...
}
```

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsTest.kt
georgf: I nee to set up the `source` & `storage` mocks before the setup functions happen.
I don't think that works out well here?

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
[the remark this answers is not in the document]
Dexterp37: So, another option would be to just use the `ExperimentsInternalAPI` in the tests, so that you can injhect whatever dependency you like for `source` and `storage`. Your code would become something like:

```
class ExperimentsInternalApi(storage, source)
object Experiemnts : ExperimentsInternalApi(new FlatFile(), new KintoSource())
```

In tests you would directly create instances of `ExperimentsInternalApi`

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsTest.kt
Dexterp37: We could set them up in `resetExperiments` as well, since they are always set up the same way, or provide default initializer for arguments in that function if you want to allow for custom ones.

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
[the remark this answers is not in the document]
georgf: Cheers, i will take this to a follow-up bug (as this is a test improvement, separate from reducing the public API surface).

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsTest.kt
georgf: Ah, a default initializer is a good point!
They are not always set up the same way, so i can't always call `resetExperiments()` automatically from `@setup` code.

georgf: I'm taking this to [bug 1536490](https://bugzilla.mozilla.org/show_bug.cgi?id=1536490) as well.

georgf:
I moved some source files into the package `mozilla.components.service.experiments` that were in subpackages before.
I had linting failures from them accessing `internal` functions in the main package and from what i understand the visibility scope is not shared between packages and subpackages anymore.

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
```diff
@@ -15,44 +19,114 @@ import mozilla.components.support.base.log.logger.Logger
...
+        }
+
+        isInitialized = true
+    }
+
+    /**
+     * Reset the library in tests.
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun testReset() {
```
travis79: nit: This could be marked as `@VisibleForTesting(otherwise = VisibleForTesting.NONE)` to narrow down the visibility even further, since this is a test only function.  I couldn't find any 'private' or local class uses of this that would require the visibility of `PRIVATE`

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsTest.kt
```diff
@@ -30,25 +30,37 @@ import java.io.File
 import kotlin.reflect.full.functions
 import kotlin.reflect.jvm.isAccessible
 
+internal fun resetExperiments(
```
travis79: nit: I think this could be `private` since it's not used outside of this file, and could be a function of the test class rather than in global scope.

PR #2378, opened 2019-03-16
Closes #2336: Add PocketListenEndpoint!!

### LINKED ISSUE #2336: Add endpoint for Pocket Listen functionality
We added the :service-pocket module and `PocketEndpoint` API in #651. The Pocket Listen API actually comes from a separate server so we'll need to add another endpoint to represent this server. We'll want to add the endpoint that takes `(id, article_url) -> listen_json_response`.

mcomella:
This PR adds the public `PocketListenEndpoint` class, which closely follows the `PocketEndpoint` class. Since these access two separate servers, I created two separate classes for them.

The code in between essentially:
- Refactors the existing code so it can be shared between the two classes
- Adds a raw endpoint
- Adds a JSONParser
- Adds the public endpoint to combine the two

I wish I could share more test code (because it's essentially doing the same style of tests) but I found it challenging and time consuming to do so.

### Pull Request checklist

### DIFF
--- a/components/service/pocket/src/main/java/mozilla/components/service/pocket/Arguments.kt
+++ b/components/service/pocket/src/main/java/mozilla/components/service/pocket/Arguments.kt
@@ -14,6 +14,6 @@ internal object Arguments {
     }
 
     fun assertIsNotBlank(input: String, exceptionIdentifier: String) {
-        if (input.isBlank()) throw IllegalArgumentException("Expected non-blank $exceptionIdentifier")
+        require(input.isNotBlank()) { "Expected non-blank $exceptionIdentifier" }
     }
 }
--- a/components/service/pocket/src/main/java/mozilla/components/service/pocket/Arguments.kt
+++ b/components/service/pocket/src/main/java/mozilla/components/service/pocket/Arguments.kt
@@ -1,0 +1,19 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.pocket
+
+/**
+ * A collection of shared argument validation functions.
+ */
+internal object Arguments {
+
+    fun assertIsValidUserAgent(userAgent: String) {
+        assertIsNotBlank(userAgent, "user agent")
+    }
+
+    fun assertIsNotBlank(input: String, exceptionIdentifier: String) {
+        if (input.isBlank()) throw IllegalArgumentException("Expected non-blank $exceptionIdentifier")
+    }
+}
--- a/components/service/pocket/src/main/java/mozilla/components/service/pocket/PocketListenEndpointRaw.kt
+++ b/components/service/pocket/src/main/java/mozilla/components/service/pocket/PocketListenEndpointRaw.kt
@@ -1,0 +1,63 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.pocket
+
+import android.net.Uri
+import android.support.annotation.VisibleForTesting
+import android.support.annotation.VisibleForTesting.PRIVATE
+import android.support.annotation.WorkerThread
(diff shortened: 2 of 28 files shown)


components/service/pocket/src/main/java/mozilla/components/service/pocket/Arguments.kt
```diff
@@ -0,0 +1,19 @@
...
+ */
+internal object Arguments {
+
+    fun assertIsValidUserAgent(userAgent: String) {
+        assertIsNotBlank(userAgent, "user agent")
+    }
+
+    fun assertIsNotBlank(input: String, exceptionIdentifier: String) {
+        if (input.isBlank()) throw IllegalArgumentException("Expected non-blank $exceptionIdentifier")
+    }
```
pocmo: nit: Maybe Kotlin's [require](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/require.html) is an option here too?

components/service/pocket/src/main/java/mozilla/components/service/pocket/PocketListenEndpointRaw.kt
```diff
@@ -0,0 +1,63 @@
...
+    @WorkerThread // Synchronous network call.
+    fun getArticleListenMetadata(articleID: Int, articleUrl: String): String? {
+        /** @return "key=value&key2=value2..." for the hard-coded key-value pairs. */
+        fun getRequestBodyString(): String = Uri.Builder()
+            .appendQueryParameter("url", articleUrl)
+            .appendQueryParameter("article_id", articleID.toString())
+            .appendQueryParameter("v", "2")
+            .appendQueryParameter("locale", "en-US")
+            .build()
+            .encodedQuery!! // We added query params so this should be non-null: assert it.
```
pocmo: (File a bug?) This could be another helper on `Request.Body` (We already have `fromString`/`fromFile`).

components/service/pocket/src/main/java/mozilla/components/service/pocket/ext/ConceptFetch.kt
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * @return returns the string contained within the response body for the given [request] or null, on error.
+ */
+@WorkerThread // synchronous network call.
+internal fun Client.fetchBodyOrNull(callingMethod: String, request: Request): String? {
+    val response: Response? = try {
+        fetch(request)
+    } catch (e: IOException) {
+        logger.debug("$callingMethod: network error", e)
```
pocmo: I wonder why you pass in `callingMethod` - won't that be part of the stacktrace already?

components/service/pocket/src/main/java/mozilla/components/service/pocket/Arguments.kt
mcomella: I didn't know this was a thing and it changes my life – thank you. :P

components/service/pocket/src/main/java/mozilla/components/service/pocket/PocketListenEndpointRaw.kt
mcomella: I'll file a bug but I don't know that the "get parameter" formatting is a common request body. 🤔  We can discuss in that issue.

mcomella: Filed https://github.com/mozilla-mobile/android-components/issues/2394

mcomella:
Rebased on master to avoid the glean intermittents.

mcomella:
Looks like TC failed to download deps, rerunning:
```
A problem occurred configuring root project 'android-components'.
> Could not resolve all artifacts for configuration ':classpath'.
   > Could not download kotlin-compiler-embeddable.jar (org.jetbrains.kotlin:kotlin-compiler-embeddable:1.3.21)
      > Could not get resource 'https://jcenter.bintray.com/org/jetbrains/kotlin/kotlin-compiler-embeddable/1.3.21/kotlin-compiler-embeddable-1.3.21.jar'.
         > Connection reset
```

mcomella:
```
> Task :buildSrc:build
Local configuration: absent local.properties; proceeding as normal.
[taskcluster:error] Task timeout after 7200 seconds. Force killing container.
```

Re-running...

mcomella:
Check is successful but now the branch is out of date and I can't merge: re-running.

PR #2389, opened 2019-03-18
Find in page accessibility

eeejay:
This pull request does two things:

### Focus control
Assure engineview is blurred when conducting search.

This is important for TalkBack. When the caret in Gecko moves from a focusable element (like a link) to a non-focusable element (like a text node in a P), the link is blurred and the document gets focus.

TalkBack then moves the accessibility focus to the input focus, and this throws the user off of the search result.

This change depends on Gecko changes, specifically [bug 1535701](https://bugzilla.mozilla.org/show_bug.cgi?id=1535701) and [bug 1536123](https://bugzilla.mozilla.org/show_bug.cgi?id=1536123).

### Announce Results

This lets TalkBack users know how many results there are for a certain string as they type without having to move a11y focus to the results label.

### Pull Request checklist

### DIFF
--- a/components/feature/findinpage/src/main/java/mozilla/components/feature/findinpage/FindInPageFeature.kt
+++ b/components/feature/findinpage/src/main/java/mozilla/components/feature/findinpage/FindInPageFeature.kt
@@ -7,6 +7,7 @@ package mozilla.components.feature.findinpage
 import android.support.annotation.VisibleForTesting
 import mozilla.components.browser.session.Session
 import mozilla.components.browser.session.SessionManager
+import mozilla.components.concept.engine.EngineView
 import mozilla.components.feature.findinpage.internal.FindInPageInteractor
 import mozilla.components.feature.findinpage.internal.FindInPagePresenter
 import mozilla.components.feature.findinpage.view.FindInPageView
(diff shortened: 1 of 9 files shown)


components/feature/findinpage/src/main/java/mozilla/components/feature/findinpage/FindInPageFeature.kt
```diff
@@ -19,10 +20,11 @@ import mozilla.components.support.base.feature.LifecycleAwareFeature
 class FindInPageFeature(
     sessionManager: SessionManager,
     view: FindInPageView,
+    engineView: EngineView,
```
Amejia481: I think this a breaking API change, and we should commenting it out in the changelog :)
In the changelog file there are some examples of previous breaking API change.

image

eeejay: Done!

Amejia481: 👍

pocmo:
It looks like 3 tests failed.

eeejay:
I don't understand what is required of me in codecov/patch.

pocmo:
> I don't understand what is required of me in codecov/patch.

That's fine! Merging now :)

PR #2391, opened 2019-03-18
[glean] WIP: Don't include `.debug` or `.release` in namespace

mdboom:
This is an experiment to fix a bug @boek reported.

For Fenix, Glean is generating code into the package: `org.mozilla.fenix.debug.GleanMetrics`, rather than `org.mozilla.fenix.GleanMetrics`, which means code would need to import glean different depending on whether it's in a debug or release build.

This lobs off `.debug` or `.release` from the `variant.applicationId` if it's there.  An alternative might be to provide a parameter to change this.  I'm happy with either outcome after discussion.

To test this fix, change the `apply from: .../sdk_generator.gradle` line in the application's `build.gradle` to:

`apply from: 'https://raw.githubusercontent.com/mdboom/android-components/glean-namespace-parameter/components/service/glean/scripts/sdk_generator.gradle'`

Cc: @boek

### Pull Request checklist

### DIFF
--- a/components/service/glean/scripts/sdk_generator.gradle
+++ b/components/service/glean/scripts/sdk_generator.gradle
@@ -112,7 +112,13 @@ ext.gleanGenerateMetricsAPI = {
     variant ->
 
     def sourceOutputDir =  "$buildDir/telemetry/src/${variant.name}/kotlin"
-    def fullNamespace = "${variant.applicationId}.GleanMetrics"
+    def applicationId = variant.applicationId
+    if (applicationId.endsWith(".debug")) {
+        applicationId = applicationId.subString[0..-7]
+    } else if (applicationId.endsWith(".release")) {
+        applicationId = applicationId.subString[0..-9]
+    }
+    def fullNamespace = "${applicationId}.GleanMetrics"
     def generateKotlinAPI = task("generateMetricsSourceFor${variant.name.capitalize()}", type: Exec) {
         description = "Generate the Kotlin code for the Metrics API"
 


components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -112,7 +112,13 @@ ext.gleanGenerateMetricsAPI = {
     variant ->
 
     def sourceOutputDir =  "$buildDir/telemetry/src/${variant.name}/kotlin"
-    def fullNamespace = "${variant.applicationId}.GleanMetrics"
+    def applicationId = variant.applicationId
```
Dexterp37: nit: I'd go for something like `filteredApplicationId`. Looks like `applicationId` is already a variable in this context (I only noticed because I mistakenly removed the `def` while testing this locally. Of course the IDE won't tell us :E)

Dexterp37: I would add an extensive comment explaining why this is needed.

components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -112,7 +112,13 @@ ext.gleanGenerateMetricsAPI = {
...
+    if (applicationId.endsWith(".debug")) {
```
Dexterp37: I'm afraid that applications can [manually define](https://developer.android.com/studio/build/application-id#change_the_application_id_for_build_variants) `applicationIdSuffix`(es) in their `build.gradle` file.

Luckily for us, turns out we have a way to get the  [`buildType`](https://android.googlesource.com/platform/tools/build/+/master/gradle/src/main/groovy/com/android/build/gradle/api/BaseVariant.java#85) from the `variant`. We can actually get the `applicationIdSuffix` from there. So this could become, more generally:

```
    def filteredAppId = variant.applicationId
    if (variant.buildType.applicationIdSuffix) {
        filteredAppId = filteredAppId.replaceAll(".$variant.buildType.applicationIdSuffix", "")
    }
```

To prevent future regressions, I would also add a suffix to the glean sample app, like [Fenix does](https://github.com/mozilla-mobile/fenix/blob/72b9d4ebdb3f9d9baeb849b3933efdd6eb915cf7/app/build.gradle#L39-L44).

mdboom:
Some context: @boek noted that we use the `applicationId` to determine what package to put the generated glean metrics in.  If you use something that changes between build variants, it makes it hard to import.  The goal here is not to change the `applicationId` but to change what we use to put on the `package` line at the top of the generated files.

The `applicationId` is also used in the URL when sending pings to the server, but what to do there I think is a separate consideration: whether we want to have separate telemetry for debug and release, to merge it all, or what have you.

Dexterp37:
> 
> 
>     * Isn't that potentially fragile since it secretly changes the application id which may not be expected?

Internally, we use the `variant.applicationId` to figure in which package we need to generate the API at build time. This is not really changing the application id itself.  @pocmo do you have any clue on how to achieve the same with gradle? Maybe there's some other property we're not aware of? This is basically uncharted waters for us :(

>     * Why do we want debug (= dev) builds to send telemetry as if they are a release build?

This doesn't necessarily imply that we'll be sending debug telemetry out (products are expected to handle this), but rather that the API is available in all flavours/variants.

mdboom:
I've created a Bugzilla for this [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1536408).

pocmo:
Oooooh! Got it. I forgot about the package of the generated code.

This is a problem 🤔. Even with this patch. Fenix build variants use different application ids, for example:

* org.mozilla.fenix (Fenix "greenfield" release)
* org.mozilla.fenix.debug (Fenix "greenfield" debug build)
* org.mozilla.firefox (Fenix build for replacing Firefox)
* org.mozilla.firefox.debug (^-- Local debug build)
* org.mozilla.firefox_beta (Fenix build for replacing Firefox Beta)
* org.mozilla.firefox_beta.debug (^-- Local debug build)
* org.mozilla.fennec_aurora (Fenix build for replacing Firefox Nightly)
* org.mozilla.fennec_aurora.debug (^-- Local debug build)

mdboom:
Ok.  Maybe we just need to provide a way for the app to override the package Glean generates into, then, using a parameter or a variable or something (hand waving over the gradle details about how to do that)...

Dexterp37:
> 
> 
> Oooooh! Got it. I forgot about the package of the generated code.
> 
> This is a problem 🤔. Even with this patch. Fenix build variants use different application ids, for example:
> 
>     * org.mozilla.fenix (Fenix "greenfield" release)
> 
>     * org.mozilla.fenix.debug (Fenix "greenfield" debug build)

Ah, I see @pocmo , thanks for letting us know, this is super useful! In general, there are two issues: the first is making sure the API is in the right place so that the application can call it. The other is to make sure data ends up in the right place in the pipeline (I filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1536464) for the latter). This PR aims to solve the first issue. With the current PR the above case would be addressed.
 
>     * org.mozilla.firefox (Fenix build for replacing Firefox)
> 
>     * org.mozilla.firefox.debug (^-- Local debug build)

Interesting. How would this be done technically in gradle? Would the code still live in the same package / directory? Because even if the app id is `org.mozilla.firefox`, we would still need generate code in the `org.mozilla.fenix` package, I think, due to [this](https://github.com/mozilla-mobile/fenix/blob/fb8c82b1bd282aae702c47a8d92727495f5bfd55/app/src/main/java/org/mozilla/fenix/FenixApplication.kt#L5). Right?

pocmo:
> Interesting. How would this be done technically in gradle? Would the code still live in the same package / directory? Because even if the app id is org.mozilla.firefox, we would still need generate code in the org.mozilla.fenix package, I think, due to this. Right?

The code is untouched and lives just where it lives (mostly in org.mozilla.fenix). Only the application id is different. 

See here (search for "applicationId"):
https://github.com/mozilla-mobile/fenix/blob/master/app/build.gradle

We somehow need to make sure that the generated code is always in the same package - regardless of the application id.

This is a similar problem that the Android build faces when generating the `R` and `BuildConfig` classes. They are always in the same place (I assume from the `defaultConfig`). Maybe that's something we can try to do too (not sure how though!)?

components/service/glean/scripts/sdk_generator.gradle
Dexterp37: > We somehow need to make sure that the generated code is always in the same package - regardless of the application id.
> 
> This is a similar problem that the Android build faces when generating the `R` and `BuildConfig` classes. They are always in the same place (I assume from the `defaultConfig`). Maybe that's something we can try to do too (not sure how though!)?

This was the *killer* suggestion! I found:

```
    println("DEBUG: $variant.applicationId")
        Task buildConfig = variant.generateBuildConfig
        def test = buildConfig.getBuildConfigPackageName()
    println("DEBUG: $test")
```

Prints

> DEBUG: org.mozilla.samples.glean.LOL
> DEBUG: org.mozilla.samples.glean

The first is the flavoured id, the second is the package name used in the BuildConfig.java class.

mdboom: Cool.  Whatever we do here will also not need to break in the context of library-defined metrics.  Building the sample app should hopefully catch that, but we'll need to confirm...

pocmo: This is what I get in Fenix (wrapped in `android.applicationVariants.all { ... }`:

```
DEBUG: org.mozilla.fenix
-----
DEBUG: org.mozilla.fennec_aurora
DEBUG: org.mozilla.fenix
-----
DEBUG: org.mozilla.fenix.debug
DEBUG: org.mozilla.fenix
-----
DEBUG: org.mozilla.fenix
DEBUG: org.mozilla.fenix
-----
DEBUG: org.mozilla.firefox.debug
DEBUG: org.mozilla.fenix
-----
DEBUG: org.mozilla.firefox
DEBUG: org.mozilla.fenix
-----
DEBUG: org.mozilla.firefox_beta.debug
DEBUG: org.mozilla.fenix
-----
DEBUG: org.mozilla.firefox_beta
DEBUG: org.mozilla.fenix
-----
DEBUG: org.mozilla.fennec_aurora.debug
DEBUG: org.mozilla.fenix
-----
DEBUG: org.mozilla.fennec_aurora
DEBUG: org.mozilla.fenix
-----
DEBUG: org.mozilla.fenix.debug
DEBUG: org.mozilla.fenix
-----
DEBUG: org.mozilla.fenix
DEBUG: org.mozilla.fenix
-----
DEBUG: org.mozilla.firefox.debug
DEBUG: org.mozilla.fenix
-----
DEBUG: org.mozilla.firefox
DEBUG: org.mozilla.fenix
-----
DEBUG: org.mozilla.firefox_beta.debug
DEBUG: org.mozilla.fenix
-----
DEBUG: org.mozilla.firefox_beta
DEBUG: org.mozilla.fenix
-----
DEBUG: org.mozilla.fennec_aurora.debug
DEBUG: org.mozilla.fenix
-----
DEBUG: org.mozilla.fennec_aurora
DEBUG: org.mozilla.fenix
-----
DEBUG: org.mozilla.fenix.debug
DEBUG: org.mozilla.fenix
-----
DEBUG: org.mozilla.fenix
DEBUG: org.mozilla.fenix
-----
DEBUG: org.mozilla.firefox.debug
DEBUG: org.mozilla.fenix
-----
DEBUG: org.mozilla.firefox
DEBUG: org.mozilla.fenix
-----
DEBUG: org.mozilla.firefox_beta.debug
DEBUG: org.mozilla.fenix
-----
DEBUG: org.mozilla.firefox_beta
DEBUG: org.mozilla.fenix
```

Dexterp37:
Closing this, it's being addressed by #2507.

PR #2403, opened 2019-03-19
[glean] 1535083: Use enums as extra keys on events

mdboom:
This uses enums for the extra keys on events so they are checked at build time rather than run time.

This won't pass CI until https://github.com/mozilla/glean_parser/pull/44 is merged and this PR is updated to use that.

### Pull Request checklist


components/service/glean/src/main/java/mozilla/components/service/glean/EventMetricType.kt
```diff
@@ -54,13 +61,32 @@ data class EventMetricType(
...
+        val extraStrings = extra?.let {
+            // There are two extra "keys" in play here:
+            //   1. The Kotlin enumeration names, in CamelCase
+            //   2. The keys sent in the ping, in snake_case
+            // Here we need to get (2) to send in the ping.  We need to use reflection to get
+            // around limitations in Java generics (since the actual enums for the event are in
+            // a template parameter).
+            if (it.size > 0) {
+                val cls = it.keys.first().javaClass
+                val method = cls.getDeclaredMethod("getValue")
```
Dexterp37: Eugh, I hoped we could make it without it. But we can't :) No problem!

PR #2418, opened 2019-03-19
Fix intermittent Glean failures

mdboom:
This is a cleaned up version of #2406, which has already passed testing three times!!! :tada: 

I tracked down a lot of weirdness down to Jobs being performed that ultimately runs *between* tests or during `resetGlean` when Glean is in an uninitialized state.

The fixes basically boil down to:

- Using a single `ioTask` object for all work scheduled in the `Dispatchers.API` `CoroutineContext`.  This lets us easily wait for all work to finish, regardless of where it came from, importantly from `resetGlean`.

- There was a single test that was intermittently failing because it checks for the pings on disk before they have a chance to be written.  There is a special case `ioTask`/`await` pair added just for the pings-on-disk-storage in the `PingStorageManager` for this.  (It uses the built-in `KotlinDispatchers.IO` context so we can't use our `Dispatchers.API` hack for this).

- `ErrorReporting` was creating its own Jobs to report errors.  This made sense when errors were reported from the `Metric` classes on the main thread, but since that was refactored to be done in the `StorageEngines` this is no longer necessary.  I'm not sure if this is strictly required to fix the failures, but it seems like a good / obvious simplification to make.

- There was a single test in `MetricsPingSchedulerTest.kt` that as a side effect causes subsequent tests to fail.  I haven't gotten to the bottom of this, but I thought it better to get testing back on track and deal with that later.  It is left in as commented out on purpose...

### Pull Request checklist


components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
```diff
@@ -276,7 +274,7 @@ class GleanTest {
     }
 
     @Test
-    fun `Application id sanitazer must correctly filter undesired characters`() {
```
travis79: Oh no!  No more `sanitazer` 🔫

mdboom:
1x tests passed.  Restarting for good luck...

PR #2423, opened 2019-03-20
Closes #1210: Fixes intermittent failing test in SearchEngineManager.

### LINKED ISSUE #1210: Intermittent test failure in SearchEngineManagerTest
The link to the Travis build may expire in the future: https://travis-ci.org/mozilla-mobile/android-components/builds/448033241
```
> Task :browser-search:testReleaseUnitTest
mozilla.components.browser.search.SearchEngineManagerTest > locale update broadcast will trigger reload FAILED
    org.mockito.exceptions.verification.WantedButNotInvoked at SearchEngineManagerTest.kt:117
209 tests completed, 1 failed
```

jonalmeida:
From my investigation, there's a lot going on in this test that can be split up into smaller bits in order to make it easier to unit test without the need for external Android APIs.

1. Registering the broadcast with the intent filter.
2. Broadcast receiver's `onReceive` calling the `load` method.
3. `load` calling the providers `loadSearchEngines`.

Even though the broadcast receiver is in a `runBlocking` block, we see the test fail because the suspend function is hidden instead the receiver so isn't being blocked, thus making it run async regardless. 

My fix is to expose the coroutine context that we use in the SearchEngineManager, so that we can provide it with the `TestCoroutineContext`. The benefit of this, is that it allows us to force the coroutine jobs to be run first with `coroutineContext.triggerActions()` in the `runBlocking` block before verification.

I've written two tests to cover the three parts as individual tests but I've also patched the original one to work. We can keep all three, the new tests, or remove them and rely on the original one.

The `@ObsoleteCoroutinesApi` is required until the `TestCoroutineContext` is moved to a new place in a minor refactor of the coroutine library (see https://github.com/Kotlin/kotlinx.coroutines/issues/541).

### Pull Request checklist

### DIFF
--- a/components/browser/search/build.gradle
+++ b/components/browser/search/build.gradle
@@ -38,6 +38,7 @@ dependencies {
     implementation Dependencies.kotlin_stdlib
     implementation Dependencies.kotlin_coroutines
 
+    testImplementation project(':support-test')
     testImplementation Dependencies.androidx_test_core
     testImplementation Dependencies.testing_junit
     testImplementation Dependencies.testing_robolectric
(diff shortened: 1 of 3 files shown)


jonalmeida:
Task failures are glean related.

PR #2425, opened 2019-03-20
Closes #2407: Throw RuntimeException if GeckoRuntime is shutting down unsolicited.

### LINKED ISSUE #2407: Implement GeckoRuntime.Delegate.onShutdown()
We should implement that callback and crash the app. Otherwise we may end up in a state where the app is still running but Gecko is dead.

pocmo:
### Pull Request checklist


components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
```diff
@@ -32,6 +32,18 @@ class GeckoEngine(
...
 
+    init {
+        runtime.delegate = GeckoRuntime.Delegate {
+            // On shutdown: The runtime is shutting down (possibly because of an unrecoverable error state). We crash
+            // the app here for two reasons:
+            //  - We want to know about those unsolicited shutdowns and fix those issues.
+            //  - We can't recover easily from this situation. Just continuing will leave us with an engine that
+            //    doesn't do anything anymore.
+            @Suppress("TooGenericExceptionThrown")
+            throw RuntimeException("GeckoRuntime is shutting down")
```
Amejia481: Should this change also be included in beta and stable variants?  I noticed that `GeckoRuntime.Delegate` is present there

pocmo: I'll just let it bubble up. After the merge day code moving it will be in beta already. :)

Amejia481: 👍

PR #2429, opened 2019-03-20
Add a synchronous testing mode

mdboom:
This adds a flag to run all `Dispatchers.API` jobs synchronously during testing.

I know there was (understandably) a requirement that using the testing API shouldn't require any special setup.  I think in order to get reproducible tests, we may need to give up on that requirement (at least for now).  I've tried to make that requirement a bit easier on users so that every call to every test method will assert that we're in testing mode, with a helpful message about how to fix things if not.

This should *effectively* get us to the same state as my not serious removal of all async from yesterday and certainly removes a lot of variables around async test running.  There are still asynchronous things that are being tested here -- all this really does is eliminate the front level of async from the equation.

There were a few more `MetricsPingScheduler` tests that failed for me locally that I've commented out here.  As always, the priority is to unblock, and then we'll come back and restore coverage as a second pass.

### Pull Request checklist

### DIFF
--- a/components/service/glean/docs/metrics/adding-new-metrics.md
+++ b/components/service/glean/docs/metrics/adding-new-metrics.md
@@ -62,6 +62,9 @@ val first = snapshot.single()
 assertEquals("login_opened", first.name)
 ```
 
+NOTE: Using the testing API requires calling `Glean.enableTestingMode()` first,
+such as from a `@Before` method.
+
 ## Counters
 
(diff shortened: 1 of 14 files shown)


mdboom:
1 passing test run

mdboom:
2 passing test runs

mdboom:
...and 3 passing test runs.

PR #2441, opened 2019-03-20
Implement glean ping tagging for "debug view".

travis79:
This implements ping tagging using the extra key `debugID` with adb command when launching the `GleanDebugActivity`.  This will add a header to the upload request `X-Debug-ID` matched with the value supplied to the `debugID` flag on the command line.

### Pull Request checklist

### DIFF
--- a/components/service/glean/README.md
+++ b/components/service/glean/README.md
@@ -127,15 +127,23 @@ for the command line switches used to pass the extra keys. These are the current
 |key|type|description|
 |---|----|-----------|
 | logPings | boolean (--ez) | If set to `true`, glean dumps pings to logcat; defaults to `false` |
-| sendPing | string (--es) | Sends the ping with the given name immediately. |
+| sendPing | string (--es) | Sends the ping with the given name immediately |
+| tagPings | string (--es) | Tags all outgoing pings as debug pings to make them available for real-time validation. The value must match the pattern `[a-zA-Z0-9-]{1,20}` |
 
-For example, to direct the glean sample application to dump pings to logcat, and send the "metrics" ping immediately, the following command can be used:
+For example, to direct the glean sample application to (1) dump pings to logcat, (2) tag the ping 
+with the `test-metrics-ping` tag, and (3) send the "metrics" ping immediately, the following command
+can be used:
 
 ```
 adb shell am start -n org.mozilla.samples.glean/mozilla.components.service.glean.debug.GleanDebugActivity \
   --ez logPings true \
-  --es sendPing metrics
+  --es sendPing metrics \
+  --es tagPings test-metrics-ping
 ```
+
+### Important GleanDebugActivity note!
+
+Options that are set using the adb flags are not immediately reset and will persist until the application is closed or manually reset.
 
 ## Contact
 
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
@@ -20,13 +20,20 @@ import mozilla.components.service.glean.BuildConfig
  * @property maxEvents the number of events to store before the events ping is sent
  * @property logPings whether to log ping contents to the console.
  * @property httpClient The HTTP client implementation to use for uploading pings.
+ * @property pingTag String tag to be applied to headers when uploading pings for debug view
  */
 data class Configuration(
-    val serverEndpoint: String = "https://incoming.telemetry.mozilla.org",
+    val serverEndpoint: String = DEFAULT_TELEMETRY_ENDPOINT,
     val userAgent: String = "Glean/${BuildConfig.LIBRARY_VERSION} (Android)",
     val connectionTimeout: Long = 10000,
     val readTimeout: Long = 30000,
     val maxEvents: Int = 500,
     val logPings: Boolean = false,
-    val httpClient: Lazy<Client> = lazy { HttpURLConnectionClient() }
-)
+    val httpClient: Lazy<Client> = lazy { HttpURLConnectionClient() },
+    val pingTag: String? = null
+) {
+    companion object {
+        const val DEFAULT_TELEMETRY_ENDPOINT = "https://incoming.telemetry.mozilla.org"
+        const val DEFAULT_DEBUGVIEW_ENDPOINT = "https://stage.ingestion.nonprod.dataops.mozgcp.net"
+    }
+}
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugActivity.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugActivity.kt
@@ -22,6 +22,16 @@ import mozilla.components.support.base.log.logger.Logger
 class GleanDebugActivity : Activity() {
     private val logger = Logger("glean/GleanDebugActivity")
 
+    companion object {
+        // This is a list of the currently accepted commands
+        const val SEND_PING_EXTRA_KEY = "sendPing"
+        const val LOG_PINGS_EXTRA_KEY = "logPings"
+        const val TAG_DEBUG_VIEW_EXTRA_KEY = "tagPings"
+
+        // Regular expression filter for debugId
+        val pingTagPattern = "[a-zA-Z0-9-]{1,20}".toRegex()
+    }
+
     // IMPORTANT: These activities are unsecured, and may be triggered by
     // any other application on the device, including in release builds.
     // Therefore, care should be taken in selecting what features are
@@ -33,8 +43,21 @@ class GleanDebugActivity : Activity() {
 
         // Enable debugging options and start the application.
         intent.extras?.let {
+            // Check for ping debug view tag to apply to the X-Debug-ID header when uploading the
+            // ping to the endpoint
+            var pingTag: String? = intent.getStringExtra(TAG_DEBUG_VIEW_EXTRA_KEY)
+
+            // Validate the ping tag against the regex pattern
+            pingTag?.let {
+                if (!pingTagPattern.matches(it)) {
+                    logger.error("tagPings value $it does not match accepted pattern $pingTagPattern")
+                    pingTag = null
+                }
+            }
+
             val debugConfig = Glean.configuration.copy(
-                logPings = intent.getBooleanExtra("logPings", Glean.configuration.logPings)
+                logPings = intent.getBooleanExtra(LOG_PINGS_EXTRA_KEY, Glean.configuration.logPings),
+                pingTag = pingTag
             )
 
             // Finally set the default configuration before starting
@@ -42,7 +65,7 @@ class GleanDebugActivity : Activity() {
             logger.info("Setting debug config $debugConfig")
             Glean.configuration = debugConfig
 
-            intent.getStringExtra("sendPing")?.let {
+            intent.getStringExtra(SEND_PING_EXTRA_KEY)?.let {
                 Glean.sendPings(listOf(it))
             }
         }
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/net/HttpPingUploader.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/net/HttpPingUploader.kt
@@ -76,25 +76,42 @@ internal class HttpPingUploader : PingUploader {
     }
 
     @VisibleForTesting(otherwise = PRIVATE)
-    internal fun buildRequest(path: String, data: String, config: Configuration) = Request(
-        url = config.serverEndpoint + path,
-        method = Request.Method.POST,
-        connectTimeout = Pair(config.connectionTimeout, TimeUnit.MILLISECONDS),
-        readTimeout = Pair(config.readTimeout, TimeUnit.MILLISECONDS),
-        headers = MutableHeaders(
+    internal fun buildRequest(path: String, data: String, config: Configuration): Request {
+        val headers = MutableHeaders(
             "Content-Type" to "application/json; charset=utf-8",
             "User-Agent" to config.userAgent,
             "Date" to createDateHeaderValue(),
             // Add headers for supporting the legacy pipeline.
             "X-Client-Type" to "Glean",
             "X-Client-Version" to BuildConfig.LIBRARY_VERSION
-        ),
-        // Make sure we are not sending cookies. Unfortunately, HttpURLConnection doesn't
-        // offer a better API to do that, so we nuke all cookies going to our telemetry
-        // endpoint.
-        cookiePolicy = Request.CookiePolicy.OMIT,
-        body = Request.Body.fromString(data)
-    )
+        )
+
+        var endpoint = config.serverEndpoint
+
+        // If there is a pingTag set, then this header needs to be added in order to flag pings
+        // for "debug view" use.
+        config.pingTag?.let {
+            headers.append("X-Debug-ID", it)
+
+            // NOTE: Tagged pings must be redirected to the GCP endpoint as the AWS endpoint isn't
+            // configured to handle them.  This may pose an issue with testing if a local server
+            // is used to capture pings.
+            endpoint = Configuration.DEFAULT_DEBUGVIEW_ENDPOINT
+        }
+
+        return Request(
+            url = endpoint + path,
+            method = Request.Method.POST,
+            connectTimeout = Pair(config.connectionTimeout, TimeUnit.MILLISECONDS),
+            readTimeout = Pair(config.readTimeout, TimeUnit.MILLISECONDS),
+            headers = headers,
+            // Make sure we are not sending cookies. Unfortunately, HttpURLConnection doesn't
+            // offer a better API to do that, so we nuke all cookies going to our telemetry
+            // endpoint.
+            cookiePolicy = Request.CookiePolicy.OMIT,
+            body = Request.Body.fromString(data)
+        )
+    }
 
     @Throws(IOException::class)
     internal fun performUpload(client: Client, request: Request): Boolean {
--- a/components/service/glean/src/test/java/mozilla/components/service/glean/debug/GleanDebugActivityTest.kt
+++ b/components/service/glean/src/test/java/mozilla/components/service/glean/debug/GleanDebugActivityTest.kt
@@ -21,6 +21,7 @@ import mozilla.components.service.glean.BooleanMetricType
 import mozilla.components.service.glean.Lifetime
 import mozilla.components.service.glean.resetGlean
 import mozilla.components.service.glean.triggerWorkManager
+import mozilla.components.service.glean.TestPingTagClient
 import okhttp3.mockwebserver.MockResponse
 import okhttp3.mockwebserver.MockWebServer
 import org.robolectric.Shadows.shadowOf
@@ -81,7 +82,7 @@ class GleanDebugActivityTest {
         // Set the extra values and start the intent.
         val intent = Intent(ApplicationProvider.getApplicationContext<Context>(),
             GleanDebugActivity::class.java)
-        intent.putExtra("logPings", true)
+        intent.putExtra(GleanDebugActivity.LOG_PINGS_EXTRA_KEY, true)
         val activity = Robolectric.buildActivity(GleanDebugActivity::class.java, intent)
         activity.create().start().resume()
 
@@ -128,9 +129,17 @@ class GleanDebugActivityTest {
         // Set the extra values and start the intent.
         val intent = Intent(ApplicationProvider.getApplicationContext<Context>(),
         GleanDebugActivity::class.java)
-        intent.putExtra("sendPing", "store1")
+        intent.putExtra(GleanDebugActivity.SEND_PING_EXTRA_KEY, "store1")
         val activity = Robolectric.buildActivity(GleanDebugActivity::class.java, intent)
         activity.create().start().resume()
+
+        // Since we reset the serverEndpoint back to the default for untagged pings, we need to
+        // override it here so that the local server we created to intercept the pings will
+        // be the one that the ping is sent to.
+        Glean.configuration = Glean.configuration.copy(
+            serverEndpoint = "http://" + server.hostName + ":" + server.port
+        )
+
         triggerWorkManager()
         val request = server.takeRequest(10L, TimeUnit.SECONDS)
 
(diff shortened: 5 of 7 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugActivity.kt
```diff
@@ -33,16 +40,25 @@ class GleanDebugActivity : Activity() {
 
         // Enable debugging options and start the application.
         intent.extras?.let {
+            var debugId = ""
+
+            // Check for ping debug view tag to apply to the X-Debug-ID header when uploading the
+            // ping to the endpoint
+            if (intent.hasExtra(TAG_DEBUG_VIEW_EXTRA_KEY)) {
+                debugId = intent.getStringExtra(TAG_DEBUG_VIEW_EXTRA_KEY)
```
georgf: Given the note about unsecured above and debug id constraints specced [in the debug view plan](https://docs.google.com/document/d/1w0aJK6AqqGAo_wr-iXMyBsmLYQTMKc5BXuSVkBS-MxU/edit?ts=5c9385d3#heading=h.1l98j8tqs1il)...
- we should sanity-check the data (filter or reject otherwise)
- we should document the constraints in the README

components/service/glean/README.md
```diff
@@ -119,14 +119,16 @@ for the command line switches used to pass the extra keys. These are the current
 |key|type|description|
 |---|----|-----------|
 | logPings | boolean (--ez) | If set to `true`, glean dumps pings to logcat; defaults to `false` |
-| sendPing | string (--es) | Sends the ping with the given name immediately. |
+| sendPing | string (--es) | Sends the ping with the given name immediately |
+| debugID | string (--es) | Tags all pings using the `X-Debug-ID` header combined with the given value |
```
georgf: Nit: Why not follow camel-case consistently here?

components/service/glean/README.md
```diff
@@ -119,14 +119,16 @@ for the command line switches used to pass the extra keys. These are the current
 |key|type|description|
 |---|----|-----------|
 | logPings | boolean (--ez) | If set to `true`, glean dumps pings to logcat; defaults to `false` |
-| sendPing | string (--es) | Sends the ping with the given name immediately. |
+| sendPing | string (--es) | Sends the ping with the given name immediately |
+| debugId | string (--es) | Tags all pings using the `X-Debug-ID` header combined with the given value which must match the pattern `[a-zA-Z0-9-]{1,20}` |
```
georgf: For our users `X-Debug-ID` is not meaningful.
Can we change this to something like...
> Tags all outgoing data as debug pings, which will make them available for real-time validation. The value must ...

components/service/glean/README.md
travis79: After reading the debug view doc again, I renamed this to `tagPings` as seen in the doc (also observing camelCase :)

components/service/glean/src/test/java/mozilla/components/service/glean/debug/GleanDebugActivityTest.kt
```diff
@@ -140,4 +140,102 @@ class GleanDebugActivityTest {
...
         server.shutdown()
     }
+
+    @Test
+    fun `pings are tagged using tagPings`() {
+        val server = MockWebServer()
+        server.enqueue(MockResponse().setBody("OK"))
+
+        resetGlean(config = Glean.configuration.copy(
+            serverEndpoint = "http://" + server.hostName + ":" + server.port
```
georgf: `"http"`?
That means plaintext transmissions.
Are we not using `https` in Glean everywhere?

georgf: Please check with akomar if `https` is availabe on this endpoint.

mdboom: We use `https` in glean everywhere in practice.  This is to communicate with the `MockWebServer` that gets fired up for this specific test.  It looks like it [might be possible](http://square.github.io/okhttp/3.x/mockwebserver/index.html?okhttp3/mockwebserver/MockWebServer.html) to get that server to speak `https` and we could look into doing that.

Are you suggesting the glean should enforce that the `serverEndpoint` is always https even when it's set to a custom value?

components/service/glean/README.md
```diff
@@ -119,14 +119,16 @@ for the command line switches used to pass the extra keys. These are the current
 |key|type|description|
 |---|----|-----------|
 | logPings | boolean (--ez) | If set to `true`, glean dumps pings to logcat; defaults to `false` |
-| sendPing | string (--es) | Sends the ping with the given name immediately. |
+| sendPing | string (--es) | Sends the ping with the given name immediately |
+| tagPings | string (--es) | Tags all pings using the `X-Debug-ID` header combined with the given value which must match the pattern `[a-zA-Z0-9-]{1,20}` |
 
 For example, to direct the glean sample application to dump pings to logcat, and send the "metrics" ping immediately, the following command can be used:
```
mdboom:
```suggestion
For example, to direct the glean sample application to (1) dump pings to logcat, (2) tag the ping with the `test-metrics-ping` tag, and (3) send the "metrics" ping immediately, the following command can be used:
```

components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugActivity.kt
```diff
@@ -33,16 +43,37 @@ class GleanDebugActivity : Activity() {
...
+            pingTag?.let {
+                if (!pingTagPattern.matches(it)) {
+                    logger.error("tagPings value $it does not match accepted pattern $pingTagPattern")
+                    pingTag = null
+                } else {
+                    // Tagged pings are not to be sent to multiple endpoints, but must be redirected to
+                    // the endpoint that can handle them:
+                    // https://bugzilla.mozilla.org/show_bug.cgi?id=1522428#c7
+                    // https://bugzilla.mozilla.org/show_bug.cgi?id=1522428#c8
+                    altUrl = "http://stage.ingestion.nonprod.dataops.mozgcp.net"
```
mdboom: Should this be `https`?

components/service/glean/src/test/java/mozilla/components/service/glean/debug/GleanDebugActivityTest.kt
travis79: Whoops, I think you misunderstand what's going on here.  This is creating a local server in a unit test in order to capture the pings and make sure they look good upon receipt.  We don't want unit test pings to get to the pipeline.

components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugActivity.kt
travis79: I copied verbatim what Arkadiusz posted in the Bugzilla comment

mdboom: Ah -- if that endpoint supports `https` (and it probably does) we should probably use that, though.

For the `MockWebServer`, (elsewhere in the PR) I think it's fine to use `http`.  It's (hopefully) safe to assume that our http client libraries have working `https` support, so I don't think we need to test that...

travis79: I changed it to https and I'll verify that with Arkadiusz

Dexterp37: Yeah, this *MUST* be HTTPS. This must not land with http. Let's check with Arkadiusz, but the endpoint **must** support HTTPS :)

components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugActivity.kt
```diff
@@ -33,16 +43,37 @@ class GleanDebugActivity : Activity() {
...
+            // Check for ping debug view tag to apply to the X-Debug-ID header when uploading the
+            // ping to the endpoint
+            var pingTag: String? = intent.getStringExtra(TAG_DEBUG_VIEW_EXTRA_KEY)
+            var altUrl: String? = null
+
+            // Validate the ping tag against the regex pattern
+            pingTag?.let {
+                if (!pingTagPattern.matches(it)) {
+                    logger.error("tagPings value $it does not match accepted pattern $pingTagPattern")
+                    pingTag = null
```
Dexterp37: We should also reset the `altUrl` to the previously set value. Alternatively, we could change how things are handled in `HttpPingUploader.buildRequest`: if there's a tag, use this endpoint, otherwise, use the one in configuration.

components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugActivity.kt
akkomar: Yes, please talk to us over https :)

components/service/glean/src/main/java/mozilla/components/service/glean/net/HttpPingUploader.kt
```diff
@@ -67,6 +67,10 @@ internal class HttpPingUploader : PingUploader {
 
         val request = buildRequest(path, data, config)
 
+        config.pingTag?.let {
```
Dexterp37: This seems to belong more to `buildRequest`. We should also add a test case for this in `HttpPingUploaderTest.kt` and a comment in the code explaining what's that for.

components/service/glean/src/test/java/mozilla/components/service/glean/debug/GleanDebugActivityTest.kt
Dexterp37: @travis79 can you please file a bug for enforcing *HTTPS* outside of tests?

components/service/glean/src/test/java/mozilla/components/service/glean/debug/GleanDebugActivityTest.kt
```diff
@@ -140,4 +140,102 @@ class GleanDebugActivityTest {
...
+            "Headers must contain X-Debug-ID",
+            "test-debug-ID",
+            request.headers.get("X-Debug-ID")
+        )
+
+        server.shutdown()
+    }
+
+    @Test
+    fun `tagPings filters ID's that don't match the pattern`() {
```
Dexterp37: Let's also check in here (or a separate test) that the URL is correctly reset to the previous value in this case.

components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugActivity.kt
travis79: Unfortunately, if I try and intercept the endpoint in the HttpPingUploader, it's too late to redirect it for tests and I have no way to capture a ping in a test to evaluate the contents.  I tried this originally but realized it made very little difference (functionally) where it got changed and this allowed me to still change the endpoint before WorkManager is triggered.

travis79:
```kotlin
serverEndpoint = altUrl?.let { it } ?: Glean.configuration.serverEndpoint
```
This line should make it so we don't have to reset altUrl to the previous.   If `pingTag` isn't there or if it doesn't match the regex, then altUrl will be null and it will just use the current configuration endpoint.

components/service/glean/src/test/java/mozilla/components/service/glean/debug/GleanDebugActivityTest.kt
travis79: I assert that it already does this.   Since the config is set to the local server endpoint at the beginning of this test, and it doesn't need to be reset in order to capture the ping to see that it doesn't contain the header, this indeed tests that the server endpoint matches the configuration endpoint.  Otherwise the ping would never be received.

components/service/glean/src/test/java/mozilla/components/service/glean/debug/GleanDebugActivityTest.kt
travis79: Bug filed [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1538689)

components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugActivity.kt
Dexterp37: Yes, this is true the first time this is called, but we can't really control what users will use this. Think of this sequence of actions:

- App is started (`config.serverEndpoint = <original server>`)
- Tagging with `X-Debug-ID=test` (`config.serverEndpoint = <tagging server>`)
- Tagging with `X-Debug-ID=£$"(/£$)"£($". The value is not valid ,so `pingTag` gets set to `null`. However `altUrl` is not reset (`config.serverEndpoint = <tagging server>`)

We are now in a mixed condition: if the app continues to be used, even if no tag is applied, data will be sent to the tagging endpoint.

components/service/glean/src/test/java/mozilla/components/service/glean/debug/GleanDebugActivityTest.kt
Dexterp37: Would that cover the case highlighted [here](https://github.com/mozilla-mobile/android-components/pull/2441#discussion_r268679816)?

components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugActivity.kt
travis79: @Dexterp37 Since the ONLY time that the intent extras are read is in onCreate(), and since altUrl is a local variable to that function, I don't think I need to do anything here.  The ONLY way that they could be re-tagged is by recreating the debug activity, which would start up with altUrl = null, and then just use the default endpoint from the configuration.

Basically, it doesn't look like it's possible to send a new tag without tearing down the debug activity and recreating it.

travis79: Per our convo in slack I think I have this addressed now

components/service/glean/src/test/java/mozilla/components/service/glean/debug/GleanDebugActivityTest.kt
travis79: Added coverage to verify the endpoint is reset.  Because of how I reset it I had to touch a couple of other tests as well to keep them working.

components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugActivity.kt
```diff
@@ -22,6 +22,16 @@ import mozilla.components.support.base.log.logger.Logger
...
     private val logger = Logger("glean/GleanDebugActivity")
 
+    companion object {
+        // This is a list of the currently accepted commands
+        const val SEND_PING_EXTRA_KEY = "sendPing"
+        const val LOG_PINGS_EXTRA_KEY = "logPings"
+        const val TAG_DEBUG_VIEW_EXTRA_KEY = "tagPings"
+
+        // Regular expression filter for debugId
+        val pingTagPattern = "[a-zA-Z0-9-]{1,20}".toRegex()
```
Dexterp37: nit: shouldn't this be a `const val PING_TAG_PATTERN =...` as well?

components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugActivity.kt
```diff
@@ -33,16 +43,37 @@ class GleanDebugActivity : Activity() {
 
         // Enable debugging options and start the application.
         intent.extras?.let {
+            // Check for ping debug view tag to apply to the X-Debug-ID header when uploading the
+            // ping to the endpoint
+            var pingTag: String? = intent.getStringExtra(TAG_DEBUG_VIEW_EXTRA_KEY)
+            var altUrl = "https://incoming.telemetry.mozilla.org"
```
Dexterp37: This should go right next to where we set `pingTag = null`. Otherwise, if the application is using a non-default server, it will reset it to the value of `altUrl` to *our* default every time the debug activity is called, even with options other than `tagPings`.

To recap, I think that this should init as `var altUrl: String? = null` and be reset to `altUrl = Configuration().serverEndpoint` after we set `pingTag = null`, below.

travis79: I'll still have to reset the serverEndpoint every time, even when there is no tagPings flag...  Otherwise:
1. User tags ping (server endpoint changed to alt endpoint)
2. User sends new intent with no `tagPings` flag (server endpoint is still the same alt endpoint)

If I don't explicitly set it to the default each time then it never gets reset.

components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugActivity.kt
travis79: `Error: const 'val' has 'Regex'.  Only primitives and String are allowed`

I tried that before, but the error above keeps me from declaring a const val Regex...

components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugActivity.kt
mdboom: That all makes sense -- but I wonder if there's a way to define this constant in one place.  I'm worried that this won't necessarily stay in sync with the one in `Configuration.kt`.

components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugActivity.kt
Dexterp37: Woha, I didn't know about that :)

components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugActivity.kt
Dexterp37: @mdboom I think that using `Configuration().serverEndpoint` works around this, but using some constant is another good take on this!

components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugActivity.kt
travis79: It would work if I left it as a string and only converted it to a Regex when I use it.

Dexterp37: Your call!

components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugActivity.kt
travis79: Indeed, I shall find a nice const home for it :)

travis79: This has pretty much been refactored out of this class to the HttpPingUploader class, but I did think that having endpoints defined as constants was a good idea so I added a companion object with `DEFAULT_AWS_ENDPOINT` and `DEFAULT_GCP_ENDPOINT` constants

components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
```diff
@@ -20,13 +20,20 @@ import mozilla.components.service.glean.BuildConfig
...
     val maxEvents: Int = 500,
     val logPings: Boolean = false,
-    val httpClient: Lazy<Client> = lazy { HttpURLConnectionClient() }
-)
+    val httpClient: Lazy<Client> = lazy { HttpURLConnectionClient() },
+    val pingTag: String? = null
+) {
+    companion object {
+        const val DEFAULT_AWS_ENDPOINT = "https://incoming.telemetry.mozilla.org"
+        const val DEFAULT_GCP_ENDPOINT = "http://stage.ingestion.nonprod.dataops.mozgcp.net/"
```
mdboom: Sorry if I missed the determination: Can this be `https`?

components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
```diff
@@ -20,13 +20,20 @@ import mozilla.components.service.glean.BuildConfig
...
     val readTimeout: Long = 30000,
     val maxEvents: Int = 500,
     val logPings: Boolean = false,
-    val httpClient: Lazy<Client> = lazy { HttpURLConnectionClient() }
-)
+    val httpClient: Lazy<Client> = lazy { HttpURLConnectionClient() },
+    val pingTag: String? = null
+) {
+    companion object {
+        const val DEFAULT_AWS_ENDPOINT = "https://incoming.telemetry.mozilla.org"
```
mdboom: Should we be consistent about the trailing `/`?  Hopefully it doesn't practically matter, but looks a little funny if these are designed to be "interchangeable".

components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
travis79: Ooops, crap, I recopied this from the bugzilla and forgot to change it

components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
travis79: Same problem as above

Dexterp37: nit: `DEFAULT_TELEMETRY_ENDPOINT`

components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
```diff
@@ -20,13 +20,20 @@ import mozilla.components.service.glean.BuildConfig
...
+        const val DEFAULT_GCP_ENDPOINT = "https://stage.ingestion.nonprod.dataops.mozgcp.net"
```
Dexterp37: nit: `DEFAULT_DEBUGVIEW_ENDPOINT`

PR #2470, opened 2019-03-21
Fixes #2469: Bookmarks component for storage and management 

### LINKED ISSUE #2469: Bookmarks component for storage and management
We need an Android Component to provide a stable interface to application-services in order to read, store, modify, and delete bookmarks.

colintheshots:
This is a WIP. I've been testing the code within Fenix to ensure it meets our needs.

I will add tests as soon as I know how to unit test application-services behavior from android-components. I will add documentation to the README and the changelog.md. Currently, this does not include any user facing features requiring accessibility considerations.

### Pull Request checklist


Amejia481:
@colintheshots  when you finish please remember to add this component to the main [README](https://github.com/mozilla-mobile/android-components/blob/master/README.md) file :)

colintheshots:
Sounds good. I'm on it.

colintheshots:
Don't commit this until we have new application-services artifacts to include with the missing classes!

colintheshots:
Seems they just got the new artifacts made, so I'm including them.

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -33,7 +33,7 @@ object Versions {
     const val room = "1.1.1"
     const val paging = "1.0.1"
 
-    const val mozilla_appservices = "0.20.2"
+    const val mozilla_appservices = "0.22.0"
```
grigoryk: We should land this separately and deal with any potential fallback orthogonality to the bookmarks work.

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -23,7 +23,7 @@ object Versions {
     const val lifecycle = "1.1.1"
 
     const val dokka = "0.9.17"
-    const val android_gradle_plugin = "3.2.1"
+    const val android_gradle_plugin = "3.3.2"
```
grigoryk: Why did you need to update the gradle plugin version?

README.md
```diff
@@ -108,6 +108,8 @@ _Combined components to implement feature-specific use cases._
 
 * ⚪ [**Awesomebar**](components/feature/awesomebar/README.md) - A component that connects a [concept-awesomebar](components/concept/awesomebar/README.md) implementation to a [concept-toolbar](components/concept/toolbar/README.md) implementation and provides implementations of various suggestion providers.
 
+* ⚪ [**Bookmarks**](components/feature/bookmarks/README.md) -  A component that adds an interface for accessing, searching, storing, and editing bookmarks.
```
grigoryk: This needs a revision. (same comment as for the changelog)

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesBookmarksStorage.kt
```diff
@@ -0,0 +1,177 @@
...
+
+/**
+ * Implementation of the [BookmarksStorage] which is backed by a Rust Places lib via [PlacesConnection].
+ */
+open class PlacesBookmarksStorage(context: Context) : BookmarksStorage, SyncableStore {
+
+    private val scope by lazy { CoroutineScope(IO) }
+    private val storageDir by lazy { context.filesDir }
+    @VisibleForTesting
+    internal open val places: Connection by lazy {
```
grigoryk: While this will work correctly because `RustPlacesConnection` is a singleton, and because it caches its readers and writers, nothing in the current composition of the two storage layers - history and bookmarks - makes it clear that it's the same db underneath, and the same connection objects should be used. And the fact that you're passing in `storageDir` twice makes this potentially even more confusing.

I would factor out bits that call `RustPlacesConnection.init(...)` into an abstract base class, and extend it in `PlacesBookmarksStorage` and `PlacesHistoryStorage`. `PlacesStorage` seems like a good name for that base class.

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesBookmarksStorage.kt
```diff
@@ -0,0 +1,177 @@
...
+        }
+    }
+
+    /**
+     * Adds a new bookmark tree node to a given node.
+     *
+     * @param newNode The node to add.
+     * @param position The optional position to add it in.
+     * @return The guid of the newly inserted bookmark item.
+     */
```
grigoryk: Here and elsewhere where we end up using a `write` connection - can you please document Sync behaviour in the docstring? For an example, see how `PlacesHistoryStorage` documents sync behaviour. I think for bookmarks a simple note saying that `other clients will see this change if sync` should be enough.

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesBookmarksStorage.kt
```diff
@@ -0,0 +1,177 @@
...
+     * @param newFolder The node to move it to.
+     * @param position The optional position to move it to.
+     * @return The guid of the moved node.
+     */
+    override suspend fun moveNode(nodeToMove: BookmarkTreeNode, newFolder: BookmarkFolder, position: Int?): String {
+        return withContext(scope.coroutineContext) {
+            when (nodeToMove) {
+                is BookmarkFolder -> {
+                    val title = nodeToMove.title
+                    writer.deleteBookmarkNode(nodeToMove.guid)
```
grigoryk: None of these API calls are correct.

You should be instead using `updateBookmark`, and constructing a `BookmarkUpdateInfo` which describes what a move looks like.

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesBookmarksStorage.kt
```diff
@@ -0,0 +1,177 @@
...
+                }
+                else -> throw IllegalStateException("Unsupported bookmark node type for addNode")
+            }
+        }
+    }
+
+    /**
+     * Moves a bookmark tree node to a given node.
+     *
+     * @param newFolder The node to move it to.
```
grigoryk: Incomplete docstring missing `nodeToMove`.

components/concept/storage/src/main/java/mozilla/components/concept/storage/BookmarksStorage.kt
```diff
@@ -0,0 +1,82 @@
...
+import mozilla.appservices.places.BookmarkFolder
+import mozilla.appservices.places.BookmarkItem
+import mozilla.appservices.places.BookmarkRoot
+import mozilla.appservices.places.BookmarkTreeNode
+import mozilla.appservices.places.BookmarkUpdateInfo
+
+interface BookmarksStorage {
+
+    /**
+     * Produces a populated bookmarks tree for the given root.
```
grigoryk: "populated" is an internal term used by the places lib with a very specific meaning (`children` array has real data). In context of an abstract interface, that term doesn't mean anything - especially since you're not defining these data types in this file. Either don't use that term, or expand it into a specific definition of what it means.

components/concept/storage/src/main/java/mozilla/components/concept/storage/BookmarksStorage.kt
```diff
@@ -0,0 +1,82 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.concept.storage
+
+import mozilla.appservices.places.BookmarkFolder
+import mozilla.appservices.places.BookmarkItem
+import mozilla.appservices.places.BookmarkRoot
+import mozilla.appservices.places.BookmarkTreeNode
+import mozilla.appservices.places.BookmarkUpdateInfo
```
grigoryk: This doesn't follow our current pattern of having `concept` components not depend on anything. This class is supposed to define simple interfaces, without imposing any external dependencies or implementations. What you have currently will force an application that only wants to consume an in-memory history storage component to actually depend on the libplaces native libraries!

This is the reason why `concept-sync` and `concept-storage` redefine some of the types exposed by libplaces, and provide conversion points (see the conventional `Types.kt`). You'll have to follow the same pattern here.

components/concept/storage/src/main/java/mozilla/components/concept/storage/BookmarksStorage.kt
```diff
@@ -0,0 +1,82 @@
...
+interface BookmarksStorage {
+
+    /**
+     * Produces a populated bookmarks tree for the given root.
+     *
+     * @param root The bookmark root to obtain.
+     * @param recursive Whether to recurse and obtain all levels of children.
+     * @return The populated root.
+     */
+    suspend fun getTree(root: BookmarkRoot, recursive: Boolean = false): BookmarkTreeNode?
```
grigoryk: The two `getTree` methods are essentially the same, so I'm a little conflicted about the split. I do like the more explicit nature of passing in a specific `BookmarkRoot` though!

components/concept/storage/src/main/java/mozilla/components/concept/storage/BookmarksStorage.kt
```diff
@@ -0,0 +1,82 @@
...
+     */
+    suspend fun deleteNode(node: BookmarkTreeNode): Boolean
+
+    /**
+     * Cleanup all resources used.
+     */
+    fun cleanup()
+
+    companion object {
+        const val DEFAULT_BOOKMARK_SEARCH_LIMIT = 10
```
grigoryk: A concrete implementation of this class may be a better place for constants like these.

components/concept/storage/src/main/java/mozilla/components/concept/storage/BookmarksStorage.kt
```diff
@@ -0,0 +1,82 @@
...
+    suspend fun addNode(target: BookmarkFolder, newNode: BookmarkTreeNode, position: Int? = null): String
+
+    /**
+     * Moves a bookmark tree node to a given node.
+     *
+     * @param newFolder The node to move it to.
+     * @param position The optional position to move it to.
+     * @return The guid of the moved node.
+     */
+    suspend fun moveNode(nodeToMove: BookmarkTreeNode, newFolder: BookmarkFolder, position: Int? = null): String
```
grigoryk: @thomcc what do you think about exposing a "sugar" method like this?

components/concept/storage/src/main/java/mozilla/components/concept/storage/BookmarksStorage.kt
```diff
@@ -0,0 +1,82 @@
...
+     * @param info The info to change in the bookmark.
+     */
+    suspend fun updateNode(node: BookmarkTreeNode, info: BookmarkUpdateInfo)
+
+    /**
+     * Deletes a bookmark node and all of its children, if any.
+     *
+     * @return Whether the bookmark existed or not.
+     */
+    suspend fun deleteNode(node: BookmarkTreeNode): Boolean
```
grigoryk: We might be making it a more complicated API by forcing that an actual `BookmarkTreeNode` is passed in. For the `places` implementation, and for any other reasonable implementation, this doesn't need more than a unique identifier, a `guid`. I'd prefer to see this API operate against minimally necessary primitives like this, I feel like it'll make for a better consumer experience in the long term.

components/concept/storage/src/main/java/mozilla/components/concept/storage/BookmarksStorage.kt
```diff
@@ -0,0 +1,82 @@
...
+     * @return The guid of the moved node.
+     */
+    suspend fun moveNode(nodeToMove: BookmarkTreeNode, newFolder: BookmarkFolder, position: Int? = null): String
+
+    /**
+     * Edits an existing bookmark item.
+     *
+     * @param info The info to change in the bookmark.
+     */
+    suspend fun updateNode(node: BookmarkTreeNode, info: BookmarkUpdateInfo)
```
grigoryk: Same comment here as in `delete` w.r.t. passing in a `node`. It makes for an ambiguous API - the only bit of information used from the `node` will be the `guid`. If users make any other changes there, they will be dropped. All "to-change" bits should live in `info`.

Let's change `node` to `guid`.

components/concept/storage/src/main/java/mozilla/components/concept/storage/BookmarksStorage.kt
```diff
@@ -0,0 +1,82 @@
...
+    suspend fun searchBookmarks(query: String, limit: Int = DEFAULT_BOOKMARK_SEARCH_LIMIT): List<BookmarkItem>
+
+    /**
+     * Adds a new bookmark tree node to a given node.
+     *
+     * @param newNode The node to add.
+     * @param position The optional position to add it in.
+     * @return The guid of the newly inserted bookmark item.
+     */
+    suspend fun addNode(target: BookmarkFolder, newNode: BookmarkTreeNode, position: Int? = null): String
```
grigoryk: s/target/parent. Also, consider if you just want to use `parentGuid` here. I think it'll be much easier down the road from the consumption point of view.

regarding `newNode` - I don't think this will map well to how places operates. `BookmarkTreeNode` requires a `guid`, timestamps, and some other things - it's not application's job to generate those bits of information. And so, that object  type shouldn't be used for creation.

Additionally, the places API specifically provides us with three `create` endpoints, partially to remove potential footguns. I think we should just expose these three `create` methods at the concept level.

components/concept/storage/build.gradle
```diff
@@ -22,6 +22,8 @@ android {
 }
 
 dependencies {
+    api Dependencies.mozilla_places
```
grigoryk: `concept` components should only depend on other concept components, or ideally nothing at all. See my comment in concept-storage/BookmarksStorage

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesBookmarksStorage.kt
```diff
@@ -0,0 +1,177 @@
...
+    /**
+     * Cleans up background work and database connections
+     */
+    override fun cleanup() {
+        scope.coroutineContext.cancelChildren()
+        places.close()
+    }
+
+    /**
+     * Synchronizes bookmarks
```
grigoryk: This should live in the base class.

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesBookmarksStorage.kt
```diff
@@ -0,0 +1,177 @@
...
+import mozilla.components.concept.sync.AuthInfo
+import mozilla.components.concept.sync.SyncStatus
+import mozilla.components.concept.sync.SyncableStore
+
+/**
+ * Implementation of the [BookmarksStorage] which is backed by a Rust Places lib via [PlacesConnection].
+ */
+open class PlacesBookmarksStorage(context: Context) : BookmarksStorage, SyncableStore {
+
+    private val scope by lazy { CoroutineScope(IO) }
```
grigoryk: This should live in the base class. See other comments.

grigoryk:
Let's land appservices bump in https://github.com/mozilla-mobile/android-components/pull/2515.

buildSrc/src/main/java/Dependencies.kt
colintheshots: This is complicated by the fact that the arguments and return values for many of the new bookmarks-related classes use types from application-services. I could create new types, but they would mirror much of the existing data classes.

buildSrc/src/main/java/Dependencies.kt
colintheshots: I thought application-services had upgraded the gradle plugin, but it was just my local build.

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesBookmarksStorage.kt
colintheshots: Good catch! This behavior belongs in a base class.

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesBookmarksStorage.kt
colintheshots: I missed that part implied in the docs that updateBookmark also performed moves. Thanks for catching this.

components/concept/storage/src/main/java/mozilla/components/concept/storage/BookmarksStorage.kt
colintheshots: It's a default argument. An overriding function is not allowed to specify default values. Either I must disallow any default value in concrete implementations, I can specify a val which must be overridden, or I specify a default here. Each has upsides and downsides.

components/concept/storage/src/main/java/mozilla/components/concept/storage/BookmarksStorage.kt
colintheshots: Done

components/concept/storage/src/main/java/mozilla/components/concept/storage/BookmarksStorage.kt
colintheshots: This makes sense. Making the change.

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesBookmarksStorage.kt
colintheshots: Done

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesBookmarksStorage.kt
colintheshots: Done

buildSrc/src/main/java/Dependencies.kt
colintheshots: Being resolved in #2515

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesBookmarksStorage.kt
colintheshots: Fixed

components/concept/storage/src/main/java/mozilla/components/concept/storage/BookmarksStorage.kt
colintheshots: Done

components/concept/storage/src/main/java/mozilla/components/concept/storage/BookmarksStorage.kt
colintheshots: I removed the BookmarkRoot version to make it easier to remove these app-services types from the concept-storage class.

components/concept/storage/src/main/java/mozilla/components/concept/storage/BookmarksStorage.kt
colintheshots: I decided to supply an empty val that must be overridden. That way the concrete implementation always supplies a default.

components/concept/storage/src/main/java/mozilla/components/concept/storage/BookmarksStorage.kt
colintheshots: I took the sugar away. I think it's helpful and maybe there's still a place for it, but it makes less sense when we can only pass primitives and types defined here from primitives. I'll be extra explicit in the docs that updateNode() can be used to move nodes.

components/concept/storage/src/main/java/mozilla/components/concept/storage/BookmarksStorage.kt
colintheshots: Done

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesBookmarksStorage.kt
colintheshots: Done

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesStorage.kt
```diff
@@ -0,0 +1,63 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.cancelChildren
+import kotlinx.coroutines.withContext
+import mozilla.appservices.places.PlacesException
+import mozilla.appservices.places.PlacesReaderConnection
+import mozilla.appservices.places.PlacesWriterConnection
+import mozilla.components.concept.sync.AuthInfo
+import mozilla.components.concept.sync.SyncStatus
+
+open class PlacesStorage(context: Context) {
```
grigoryk: Can't this be `abstract`? I'm not sure we'll ever instantiate this class by itself.

components/concept/storage/src/main/java/mozilla/components/concept/storage/BookmarksStorage.kt
```diff
@@ -0,0 +1,116 @@
...
+     * Adds a new bookmark separator to a given node.
+     *
+     * Sync behavior: will add new separator to remote devices.
+     *
+     * @param parentGuid The parent guid of the new node.
+     * @param newNode The details of the node to add.
+     * @param position The optional position to add the new node or null to append.
+     * @return The guid of the newly inserted bookmark item.
+     */
+    suspend fun addSeparator(parentGuid: String, newNode: BookmarkNode, position: Int? = null): String
```
grigoryk: This doesn't need `newNode`.

components/concept/storage/src/main/java/mozilla/components/concept/storage/BookmarksStorage.kt
```diff
@@ -0,0 +1,116 @@
...
+     * Adds a new bookmark folder to a given node.
+     *
+     * Sync behavior: will add new separator to remote devices.
+     *
+     * @param parentGuid The parent guid of the new node.
+     * @param newNode The details of the node to add.
+     * @param position The optional position to add the new node or null to append.
+     * @return The guid of the newly inserted bookmark item.
+     */
+    suspend fun addFolder(parentGuid: String, newNode: BookmarkNode, position: Int? = null): String
```
grigoryk: You are using a `BookmarkNode` when a `title` would have been sufficient. `BookmarkNode` as you defined it could be a bookmark, a folder or a separator, and it will always have a `guid`. It might even have a `url`. None of those are valid inputs for this API, only `title` and `position` are.

Another way to think about this - `add*` APIs should take a generic description of an item (where, what), not a reified version of it (an object with an assigned `guid`). Doing it the other way is akin to creating a variable and insisting on a memory address, as opposed to asking the memory allocator to give you one for the data you're trying to store.

If you really want to have this API take an object, you'll need to define a `PendingBookmarkFolder`, or a `BookmarkFolder` with an optional `guid`... These options seem worse to me.

components/concept/storage/src/main/java/mozilla/components/concept/storage/BookmarksStorage.kt
```diff
@@ -0,0 +1,116 @@
...
+     * Adds a new bookmark item to a given node.
+     *
+     * Sync behavior: will add new bookmark item to remote devices.
+     *
+     * @param parentGuid The parent guid of the new node.
+     * @param newNode The details of the node to add.
+     * @param position The optional position to add the new node or null to append.
+     * @return The guid of the newly inserted bookmark item.
+     */
+    suspend fun addItem(parentGuid: String, newNode: BookmarkNode, position: Int? = null): String
```
grigoryk: Same comment here as for folder creation. A bookmark is best described by a parentGuid, url, title, and an optional position.

components/concept/storage/src/main/java/mozilla/components/concept/storage/BookmarksStorage.kt
```diff
@@ -0,0 +1,116 @@
...
+     * Sync behavior: will remove bookmark from remote devices.
+     *
+     * @return Whether the bookmark existed or not.
+     */
+    suspend fun deleteNode(guid: String): Boolean
+
+    /**
+     * The default number of bookmark entries to return from searches
+     */
+    val defaultBookmarkSearchLimit: Int
```
grigoryk: Ah, i see what you meant here. Yeah, this is worse than before, my bad! Let's just switch back to what you had before. Defining this as a requirement for interface implementors is pretty weird.

docs/changelog.md
```diff
@@ -70,9 +70,10 @@ permalink: /changelog/
 * **concept-storage**
   * ⚠️ **This is a breaking API change!** for non-component implementations of `HistoryStorage`.
   * `HistoryStorage` got new APIs: `deleteEverything`, `deleteVisitsSince`, `deleteVisitsBetween`, `deleteVisitsFor`, `prune` and `runMaintenance`.
+  * Added `BookmarksStorage` for handling the saving, searching, and management of browser bookmarks.
 
 * **browser-storage-sync**, **browser-storage-memory**
-  * Implementations of `concept-storage`/`HistoryStorage` expose the newly added APIs.
+  * Implementations of `concept-storage`/`HistoryStorage` and `concept-storage`/`BookmarksStorage` expose the newly added APIs.
```
grigoryk: This says that browser-storage-memory also got the implementation, which isn't the case.

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesStorage.kt
```diff
@@ -0,0 +1,63 @@
...
+        RustPlacesConnection
+    }
+
+    internal val writer: PlacesWriterConnection by lazy { places.writer() }
+    internal val reader: PlacesReaderConnection by lazy { places.reader() }
+
+    /**
+     * Internal database maintenance tasks. Ideally this should be called once a day.
+     */
+    suspend fun runMaintenance() {
```
grigoryk: Let's keep `cleanup` and `runMaintanence` at the interface level. You can declare a new concept-storage interface called `Storage` (alongside HistoryStorage and BookmarksStorage), and have these methods live there. Then an abstract `PlacesStorage` will implement `Storage`. 

The reason I think these are better left at the concept level: nothing in the system interacts with a PlacesHistoryStorage, or a PlacesBookmarksStorage. We're always interacting with a _some_ object that implements a HistoryStorage contract, or a BookmarksStorage contract. If `runMaintanence` is hidden within the concrete implementation, we will not be able to actually call these methods from within a `SyncManager`, because it doesn't know about `browser-storage-sync`, it knows about `concept-storage`.

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesStorage.kt
colintheshots: If I make it abstract, detekt fails on "unnecessary abstract class."

components/concept/storage/src/main/java/mozilla/components/concept/storage/BookmarksStorage.kt
colintheshots: Good catch!

components/concept/storage/src/main/java/mozilla/components/concept/storage/BookmarksStorage.kt
colintheshots: Fixed

components/concept/storage/src/main/java/mozilla/components/concept/storage/BookmarksStorage.kt
colintheshots: Fixed

components/concept/storage/src/main/java/mozilla/components/concept/storage/BookmarksStorage.kt
colintheshots: I agree. Going back to the old solution.

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesStorage.kt
grigoryk: That's strange. Even with two different implementations? I'd prefer to tell detekt to go away in this case... Not really a blocker.

docs/changelog.md
colintheshots: Updated

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesStorage.kt
colintheshots: That makes sense. Changes made.

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesStorage.kt
colintheshots: Left this alone, as we agreed due to detekt style suggestions.

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesStorage.kt
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * Cleans up background work and database connections
+     */
+    override fun cleanup() {
+        scope.coroutineContext.cancelChildren()
+        places.close()
+    }
+
+    /**
+     * Synchronizes bookmarks
```
grigoryk: Comment is wrong. It actually doesn't :) And this is a base class, anyway.

components/concept/storage/src/main/java/mozilla/components/concept/storage/Storage.kt
```diff
@@ -0,0 +1,21 @@
...
+
+package mozilla.components.concept.storage
+
+/**
+ * An interface which provides generic operations for storing browser data like history and bookmarks.
+ */
+interface Storage {
+
+    /**
+     * Internal database maintenance tasks. Ideally this should be called once a day.
```
grigoryk: nit: remove last part of this comment; that applies to a specific places impl, not a generic interface.

gradle.properties
```diff
@@ -23,3 +23,4 @@ libVcsUrl=https://github.com/mozilla-mobile/android-components.git
 
 libLicense=MPL-2.0
 libLicenseUrl=https://www.mozilla.org/en-US/MPL/2.0/
+android.enableUnitTestBinaryResources=true
```
grigoryk: This isn't necessary, right?

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -87,6 +87,7 @@ object Dependencies {
     const val mozilla_fxa = "org.mozilla.appservices:fxaclient:${Versions.mozilla_appservices}"
     const val mozilla_sync_logins = "org.mozilla.appservices:logins:${Versions.mozilla_appservices}"
     const val mozilla_places = "org.mozilla.appservices:places:${Versions.mozilla_appservices}"
+    const val mozilla_places_for_unit_tests = "org.mozilla.appservices:places-forUnitTests:${Versions.mozilla_appservices}"
```
grigoryk: You don't need this here, afaik.

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesStorage.kt
colintheshots: Updated

components/concept/storage/src/main/java/mozilla/components/concept/storage/Storage.kt
colintheshots: Done

gradle.properties
colintheshots: I did it because Robolectric was filling the screen with complaints about it missing. It's not strictly necessary though.

buildSrc/src/main/java/Dependencies.kt
colintheshots: Removed.

colintheshots:
Do I need more codecov or should I just merge?

pocmo:
> Do I need more codecov or should I just merge?

Use your own judgement. If it's easily possible then let's try. If the cost is too high or if it's just not feasible then go ahead and merge. :)

colintheshots:
The areas that lost coverage were MetricsPingScheduler and Glean, which I didn't change at any point. I call BS. Landing this.

colintheshots:
Now Glean unit tests are broken.

PR #2481, opened 2019-03-21
Add documentation for Gleans labeled metrics [ci skip]

georgf:
Add a first pass on quick-start documentation for labeled metrics.

### Pull Request checklist


components/service/glean/docs/metrics/adding-new-metrics.md
```diff
@@ -108,3 +108,50 @@ assertEquals(6, Controls.refreshPressed.testGetValue())
...
+```
+
+Now you can use the labeled counter from the applications code:
+```Kotlin
+import org.mozilla.yourApplication.GleanMetrics.Updater
+
+// ...
+Updater.loadError["timeout"].add() // Adds 1 to the "timeout" counter.
+Updater.loadError["not_found"].add(2)
+```
```
mdboom: Should we add:

"If the labels are predefined in the metrics.yaml and a label not in that set it used, the value is applied to a special `__other__` category."

georgf: Hm, i would prefer to keep it out of this document to keep things simple.
This would be important for the longer article for the metric.

mdboom: Ok.

PR #2508, opened 2019-03-25
Update strings from review.

pocmo:
Those are the new, reviewed strings from Michelle.


pocmo:
Okay, lint doesn't like that we have translated strings now that are not used anymore... :)

pocmo:
I disabled the lint checks for `MissingTranslation` and `ExtraTranslation`. With our translation process it is possible that we will have missing translations (just not translated yet) or extra translations (the next import should remove them) temporarily.

components/feature/sitepermissions/src/main/res/values/strings.xml
```diff
@@ -1,43 +1,33 @@
...
     <!-- Option in a dialog for requesting a camera permission, this option will give access to
          back facing camera-->
-    <string name="mozac_feature_sitepermissions_back_facing_camera">Back facing camera</string>
-
+    <string name="mozac_feature_sitepermissions_back_facing_camera2">Rear-facing camera</string>
     <!-- Option in a dialog for requesting a camera permission, this option will give access to
      front facing camera-->
-    <string name="mozac_feature_sitepermissions_selfie_camera">Selfie back camera</string>
-
+    <string name="mozac_feature_sitepermissions_selfie_camera2">Front-facing camera</string>
```
Amejia481: Some of the strings values from site permissions have been updated to more like Fennec style.  I took them from the Fenix mock-ups, now we are changing them, are we going to provide a way for apps to customize them or overridden the strings on the apps will be the preferred alternative?   As we are changing the strings ids the last alternative could be fragile.

pocmo: Those strings are superseding the ones from the mocks. Fenix is supposed to use those updated strings.

Amejia481: 👍

pocmo:
Looks like I missed some tests that used those strings. I am going to update the PR.

PR #2509, opened 2019-03-25
Split some fields out of pings'`ping_info` to `client_info`

Dexterp37:
In addition to this, this moves the following fields from the `baseline` ping payload to the new `client_info`, thus sending them with all the outgoing pings: "os", "os_version", "device_manufacturer", "device_model", "architecture". Fields "locale" and "duration" are retained in the `baseline` ping payload. See [bug 1528787](https://bugzilla.mozilla.org/show_bug.cgi?id=1528787) for more context.

Please note that this additionally changes the internal storage name for internal metrics, thus resetting its content (e.g. client id, first run date).

### Pull Request checklist


components/service/glean/build.gradle
```diff
@@ -15,7 +15,7 @@ apply plugin: 'kotlin-android'
  * This uses a specific version of the schema identified by a git commit hash.
  */
 String GLEAN_PING_SCHEMA_GIT_HASH = "63155ec"
-String GLEAN_PING_SCHEMA_URL = "https://raw.githubusercontent.com/mozilla-services/mozilla-pipeline-schemas/$GLEAN_PING_SCHEMA_GIT_HASH/schemas/glean/baseline/baseline.1.schema.json"
```
Dexterp37: This change will need to be reverted and `GLEAN_PING_SCHEMA_GIT_HASH` will need to be updated once mozilla-services/mozilla-pipeline-schemas#296 is merged.

components/service/glean/metrics.yaml
```diff
@@ -95,29 +111,12 @@ glean.baseline:
...
-    expires: never
-
-glean.internal.metrics:
   client_id:
     type: uuid
     description:
       A UUID uniquely identifying the client.
     send_in_pings:
-      - glean_ping_info
+      - glean_client_info
```
Dexterp37: Please note that changing this will reset the client ids and all the other metrics. I think this should be ok right now, since nobody should be relying on it.

Dexterp37:
> 
> 
> This LGTM. The only thing I might suggest is follow up work to just refactor the need for BaselinePing.kt. Now that all of this stuff was moved back to Glean.kt, there really isn't much point to have an entire class just to initialize a single metric. On the other hand, since the reason these metric initializations were originally moved out of Glean.kt was to clean up all the init code out of the main API file, it may also be valid to move the client_info metric initializations to their own class.

Good point. I filed [bug 1539077](https://bugzilla.mozilla.org/show_bug.cgi?id=1539077) for dealing with this in the future.

PR #2516, opened 2019-03-26
Add the "app_channel" field to glean pings' `client_info`

Dexterp37:
This is slightly different than the [proposed approach](https://bugzilla.mozilla.org/show_bug.cgi?id=1520741#c1). @travis79 suggestion was better, since we have startup pings that will need this field and the only way for them to know the channel soon enough would be to have this bit of information as part of the configuration/initialize. Otherwise, since our outer facing API is async, we risk to not have the channel on time.

### Pull Request checklist


components/service/glean/build.gradle
```diff
@@ -15,7 +15,7 @@ apply plugin: 'kotlin-android'
  * This uses a specific version of the schema identified by a git commit hash.
  */
 String GLEAN_PING_SCHEMA_GIT_HASH = "1933f26"
-String GLEAN_PING_SCHEMA_URL = "https://raw.githubusercontent.com/mozilla-services/mozilla-pipeline-schemas/$GLEAN_PING_SCHEMA_GIT_HASH/schemas/glean/baseline/baseline.1.schema.json"
+String GLEAN_PING_SCHEMA_URL = "https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/5489597276d168a7f275012ff2a83bcb51e233c2/schemas/glean/baseline/baseline.1.schema.json"
```
Dexterp37: This will need to change once mozilla-services/mozilla-pipeline-schemas#297 is merged.

components/service/glean/build.gradle
```diff
@@ -15,7 +15,7 @@ apply plugin: 'kotlin-android'
  * This uses a specific version of the schema identified by a git commit hash.
  */
 String GLEAN_PING_SCHEMA_GIT_HASH = "1933f26"
```
georgf: Is this unused now and should be removed?

components/service/glean/README.md
```diff
@@ -86,6 +86,10 @@ Glean should be initialized as soon as possible, and importantly, before any
 other libraries in the application start using Glean. Library code should never
 call `Glean.initialize`, since it should be called exactly once per application.
 
+**Note**: if the application has the concept of release channels and knows which channel it is on at
+run-time, then it can provide glean with this information by setting it as part of the `Configuration`
+object parameter of the `Glean.initialize` method.
```
georgf: How do i use that? Can you give an example?

components/service/glean/build.gradle
Dexterp37: No, this is a transient thing: once schema changes are merged I'll need to fix this up and revert the line below. See my comment [here](https://github.com/mozilla-mobile/android-components/pull/2516#discussion_r269037330). We need to do this otherwise tests will fail in this repo until schema changes are merged.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -230,6 +230,10 @@ open class GleanInternalAPI internal constructor () {
         StringsStorageEngine.record(GleanInternalMetrics.deviceModel, Build.MODEL)
         StringsStorageEngine.record(GleanInternalMetrics.architecture, Build.SUPPORTED_ABIS[0])
 
+        if (configuration.channel != null) {
```
mdboom: Minor point: I think if you use the `?.let` syntax here, the `orEmpty()` won't be necessary (since Kotlin enforces it's non-null within the block).

Dexterp37: Great point. I think this also fixes the low test coverage :)

components/service/glean/metrics.yaml
```diff
@@ -156,6 +156,21 @@ glean.internal.metrics:
...
+    type: string
+    lifetime: application
+    send_in_pings:
+      - glean_client_info
+    description: |
+      The channel the application is being distributed on.
+    bugs:
+      - 1520741
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1508305#c9
```
Dexterp37: *NOTE TO SELF*: This requires a data review :)

Dexterp37: Requested here: https://bugzilla.mozilla.org/show_bug.cgi?id=1520741#c16

PR #2526, opened 2019-03-26
Make loading page progress bar accessible to TalkBack.

eeejay:
ProgressBars can be made accessible to TalkBack by setting `android:accessibilityLiveRegion`.
They will emit `TYPE_VIEW_SELECTED` events. TalkBack will format those events into percentage
announcements along with a pitch-change earcon. We are not using that feature here for
several reasons:
1. They are dispatched via a 200ms timeout. Since loading a page can be a short process,
   and since we only update the bar a handful of times, these events often never fire and
   they don't give the user a true sense of the progress.
2. The last 100% event is dispatched after the view is hidden. This prevents the event
   from being fired, so the user never gets a "complete" event.
3. Live regions in TalkBack have their role announced, so the user will hear
   "Progress bar, 25%". For a common feature like page load this is very chatty and unintuitive.
4. We can provide custom strings instead of the less useful percentage utterance, but
   TalkBack will not play an earcon if an event has its own text.

For all those reasons, we are going another route here with a "loading" announcement
when the progress bar first appears along with scroll events that have the same
pitch-change earcon in TalkBack (although they are a bit louder). This gives a concise and
consistent feedback to the user that they can depend on.

### Pull Request checklist

### DIFF
--- a/components/browser/toolbar/src/main/res/values/strings.xml
+++ b/components/browser/toolbar/src/main/res/values/strings.xml
@@ -6,4 +6,6 @@
     <!-- Content description (not visible, for screen readers etc.): Description for the overflow menu button in the browser toolbar. -->
     <string name="mozac_browser_toolbar_menu_button">Menu</string>
     <string name="mozac_clear_button_description">Clear</string>
+    <!-- Announcement made by the screen reader when the progress bar is shown and a page is loading -->
+    <string name="mozac_browser_toolbar_progress_loading">Loading</string>
 </resources>
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
@@ -15,6 +15,7 @@ import android.util.TypedValue
 import android.view.Gravity
 import android.view.View
 import android.view.ViewGroup
+import android.view.accessibility.AccessibilityEvent
 import android.widget.ProgressBar
 import mozilla.components.browser.menu.BrowserMenu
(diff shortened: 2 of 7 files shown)


components/browser/toolbar/src/main/res/values/strings.xml
```diff
@@ -6,4 +6,5 @@
     <!-- Content description (not visible, for screen readers etc.): Description for the overflow menu button in the browser toolbar. -->
     <string name="mozac_browser_toolbar_menu_button">Menu</string>
     <string name="mozac_close_button_description">Close</string>
+    <string name="mozac_progress_loading">Loading</string>
```
pocmo: nit: Can you turn this into `mozac_browser_toolbar_progress_loading` ("mozac" + component name + string name)

pocmo: Oh, one more thing: Can you add a comment above this string explaining the context this strings shows up in? Our L10N tools will pick up that comment and show it next to the string to translators. I guess for this string it is worth mentioning that it will be read by talkback.

eeejay: Oops. Apparently I needed to put in a translation for German and Chinese? I grabbed those from Android's core framework.

pocmo: After #2508 that lint check is disabled and will no longer fail the build since it's expected to land code without translations :D

Pike:
Fly-by comment, I see that this patch landed localizations, too. Those don't get synchronized to the l10n process, and won't be seen by the l10n community.

If you're up to contribute localizations for your patches, doing so on Pontoon will be more transparent.

PR #2531, opened 2019-03-26
New Component: lib-push-firebase for FCM push notification support

jonalmeida:
This is the firebase implementation of the `PushService`.

### Pull Request checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,8 @@ _Supporting components with generic helper code._
 
 * ⚪ [**Public Suffix List**](components/lib/publicsuffixlist/README.md) - A library for reading and using the [public suffix list](https://publicsuffix.org/).
 
+* 🔴 [**Push-Firebase**](components/lib/push-firebase/README.md) - A [concept-push](concept/push/README.md) implementation using [Firebase Cloud Messaging](https://firebase.google.com/products/cloud-messaging/).
+
 ## Tooling
 
 * 🔵 [**Fetch-Tests**](components/tooling/fetch-tests/README.md) - A generic test suite for components that implement [concept-fetch](concept/fetch/README.md).
--- a/components/concept/push/src/main/java/mozilla/components/concept/push/PushProcessor.kt
+++ b/components/concept/push/src/main/java/mozilla/components/concept/push/PushProcessor.kt
@@ -30,7 +30,7 @@ interface PushProcessor {
     /**
      * A new push message has been received.
      */
-    fun onMessageReceived(message: PushMessage)
+    fun onMessageReceived(message: EncryptedPushMessage)
 
     /**
      * An error has occurred.
@@ -61,13 +61,40 @@ object PushProvider {
 /**
  * A push message holds the information needed to pass the message on to the appropriate receiver.
  */
-data class PushMessage(val data: String)
+data class EncryptedPushMessage(
+    val channelId: String,
+    val body: String,
+    val encoding: String,
+    val salt: String = "",
+    val cryptoKey: String = "" // diffie–hellman key
+) {
+    companion object {
+        /**
+         * The [salt] and [cryptoKey] are optional as part of the standard for WebPush, so we should default
+         * to empty strings.
+         *
+         * Note: We have use the invoke operator since secondary constructors don't work with nullable primitive types.
+         */
+        operator fun invoke(
+            channelId: String,
+            body: String,
+            encoding: String,
+            salt: String? = null,
+            cryptoKey: String? = null
+        ) = EncryptedPushMessage(channelId, body, encoding, salt ?: "", cryptoKey ?: "")
+    }
+}
 
 /**
- * The different kind of message types that a [PushMessage] can be.
+ * The different kind of message types that a [EncryptedPushMessage] can be:
+ *  - Application Services (e.g. FxA/Send Tab)
+ *  - WebPush messages (see: https://www.w3.org/TR/push-api/)
+ *  - Third party integrated SDKs (e.g. Leanplum)
+ *  - Uncategorized messages.
  */
 enum class PushType {
     Services,
+    WebPush,
     ThirdParty,
     Unknown
 }
--- a/components/concept/push/src/main/java/mozilla/components/concept/push/PushProcessor.kt
+++ b/components/concept/push/src/main/java/mozilla/components/concept/push/PushProcessor.kt
@@ -86,20 +86,6 @@ data class EncryptedPushMessage(
 }
 
 /**
- * The different kind of message types that a [EncryptedPushMessage] can be:
- *  - Application Services (e.g. FxA/Send Tab)
- *  - WebPush messages (see: https://www.w3.org/TR/push-api/)
- *  - Third party integrated SDKs (e.g. Leanplum)
- *  - Uncategorized messages.
- */
-enum class PushType {
-    Services,
-    WebPush,
-    ThirdParty,
-    Unknown
-}
-
-/**
  *  Various error types.
  */
 sealed class Error {
--- a/components/concept/push/src/main/java/mozilla/components/concept/push/PushProcessor.kt
+++ b/components/concept/push/src/main/java/mozilla/components/concept/push/PushProcessor.kt
@@ -6,6 +6,8 @@
 
 package mozilla.components.concept.push
 
+import android.support.annotation.VisibleForTesting
+
 /**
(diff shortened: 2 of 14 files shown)


jonalmeida:
I'd like to add regarding the lack of tests for `AbstractFirebasePushService`: integrating with the FirebaseMessagingService directly poses some restrictions for testing since it internally does some checks on it's singleton that we can't mock easily. I've created a follow issue to address this (see: https://github.com/mozilla-mobile/android-components/issues/2607).

README.md
```diff
@@ -202,6 +202,8 @@ _Supporting components with generic helper code._
 
 * ⚪ [**Public Suffix List**](components/lib/publicsuffixlist/README.md) - A library for reading and using the [public suffix list](https://publicsuffix.org/).
 
+* 🔴 [**Push-Firebase**](components/lib/push-firebase/README.md) - A library for reading and using the [public suffix list](https://publicsuffix.org/).
+
```
pocmo: *A library for reading and using the public suffix list* :)

components/concept/push/src/main/java/mozilla/components/concept/push/PushService.kt
```diff
@@ -32,4 +33,17 @@ interface PushService {
...
+     */
+    fun isAvailable(context: Context): Boolean
+}
+
+/**
+ * Supported push services.
+ */
+enum class ServiceType {
+    FCM,
+    ADM
```
pocmo: This being an enum in the `concept` significantly limits the flexibility of this concept? 🤔

README.md
jonalmeida: Push notifications are magic, they do more things than you expect! 😅 

Fixing.

components/concept/push/src/main/java/mozilla/components/concept/push/PushService.kt
jonalmeida: That's a good point. The `ServiceType` and `PushType` can now be moved further up to the feature since we're integrating the rust component at that level.

components/lib/push-firebase/src/main/java/mozilla/components/lib/push/firebase/AbstractFirebasePushService.kt
```diff
@@ -0,0 +1,61 @@
...
+        }
+    }
+
+    final override fun stop() {
+        FirebaseMessaging.getInstance().isAutoInitEnabled = false
+    }
+
+    final override fun forceRegistrationRenewal() {
+        stop()
+        FirebaseInstanceId.getInstance().deleteInstanceId()
```
grigoryk: This is a blocking function, you should not call it on the main thread.

components/lib/push-firebase/src/main/java/mozilla/components/lib/push/firebase/AbstractFirebasePushService.kt
```diff
@@ -0,0 +1,61 @@
...
+import mozilla.components.concept.push.PushService
+
+abstract class AbstractFirebasePushService : FirebaseMessagingService(), PushService {
+
+    init {
+        start() // Allow the app to choose when to start? This is harder to allow than it looks..
+    }
+
+    final override fun start() {
+        FirebaseMessaging.getInstance().isAutoInitEnabled = true
```
grigoryk: From the docs: `By default, FCM auto-initialization is enabled.` - unless you say otherwise in the manifest. So this seems like a no-op!

components/lib/push-firebase/src/main/java/mozilla/components/lib/push/firebase/AbstractFirebasePushService.kt
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    override fun onNewToken(newToken: String) {
+        PushProvider.requireInstance.onNewToken(newToken)
```
grigoryk: There is nothing that guarantees this method won't be called before the `start` method. Or am I missing something?

components/lib/push-firebase/src/main/java/mozilla/components/lib/push/firebase/AbstractFirebasePushService.kt
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    override fun onMessageReceived(remoteMessage: RemoteMessage?) {
```
grigoryk: Same comment here regarding sequencing of this and `start`.

components/lib/push-firebase/src/main/java/mozilla/components/lib/push/firebase/AbstractFirebasePushService.kt
```diff
@@ -0,0 +1,61 @@
...
+            )
+            PushProvider.requireInstance.onMessageReceived(message)
+        }
+    }
+
+    final override fun stop() {
+        FirebaseMessaging.getInstance().isAutoInitEnabled = false
+    }
+
+    final override fun forceRegistrationRenewal() {
```
grigoryk: It's not very clear to me how this works. Can you add a detailed comment which explains underlying behaviour?

From the docs: "when auto init is enabled, [.... fcm] generates a new token when it is deleted (which prevents deleteInstanceId() from stopping the periodic sending of data)".

So you stop this process first, then call `deleteInstanceId`. From the docs:

"Delete the Instance ID and the data associated with it.

This stops the periodic sending of data to the Firebase backend started when the Instance ID was generated, unless another library that requires InstanceId (like FCM, RemoteConfig or Analytics) is used or it's configured to be executed automatically.

A new Instance ID is generated asynchronously if Firebase Cloud Messaging auto-init is enabled."

So, second paragraph implies that this _may_ be a no-op operation? And last sentence implies that you don't need to switch `isAutoInitEnabled` flag at all! You could have just called a delete method by itself, and it would regenerate stuff (same as re-enabling auto-init after deleting).

That's my interpretation of the docs, I haven't actually used this API. Can you explain how you think it works?

docs/changelog.md
```diff
@@ -21,6 +21,13 @@ permalink: /changelog/
 * **feature-findinpage**
    * Find in Page now emits facts
 
+* **concept-push**
+  * 🆕 Added a new component for supporting push notifications.
```
grigoryk: This PR doesn't add `concept-push`, it modifies it. Or was it not mentioned in previous changelogs?

components/concept/push/src/main/java/mozilla/components/concept/push/PushProcessor.kt
```diff
@@ -6,68 +6,90 @@
...
+    @Volatile
+    private var instance: PushProcessor? = null
+
+    @VisibleForTesting
+    internal fun reset() {
+        instance = null
+    }
+    val requireInstance: PushProcessor
+        get() = instance ?: throw IllegalStateException(
+            "You need to call start() on your Push instance from Application.onCreate()."
```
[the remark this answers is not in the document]
jonalmeida: > I think a better home for the singleton is in the implementation. It's an implementation detail, after all. Conceivably you could have multiple types of push processors running at once, right? Against fcm, against a home-baked push service, whatever. Likely not in practice, but at the concept level you currently both allow it (by virtue of there being an abstract definition of a push processor), and discourage it (by defining a singleton).

Currently, multiple `PushService`'s can run in place but those services we likely don't have control over instantiating (in FCM's case it's init'd but a google plugin the app needs to add). Because of this, we have no way of passing an instance of a `PushProcessor` to the service first.

Using the `CrashReporter` component as an example, having a singleton provider seemed like the only viable solution to ensuring that we get all messages before the service starts.

Screen Shot 2019-04-05 at 5 23 18 PM

This is a communication chart of how each class talks to each other. (It assumes the app developer decided to disable FCM auto-start by default and starts it later by choice).

In the diagram, the `Push.kt` is an implementation of a `PushProcessor` that lives in the feature which receives all messages and tokens from a service, forwards it to the Rust `PushManager` to be decrypted. The message is then forwarded to the relevant listeners via the `MessageRouter`.

components/lib/push-firebase/src/main/java/mozilla/components/lib/push/firebase/AbstractFirebasePushService.kt
jonalmeida: See above.

jonalmeida:
Abandoning this PR for https://github.com/mozilla-mobile/android-components/pull/2694

PR #2532, opened 2019-03-27
Closes #1542: Add QR scanning component

### LINKED ISSUE #1542: Component: QR code scanning
We want to wrap the code needed to do QR code scanning in a component (`lib-qr`).

We currently have two use cases:
* FxA Authentication
* Fenix URL scanning

Fennec currently uses intents to launch an installed barcode scanner. For security reasons we do not want to do that here.

csadilek:
This brings in `feature-qr` as a refactored version of @vladikoff's earlier PR (incl. some fixes, tests and kdoc). It uses zxing and the camera2 API.

What's great is that the scanner works reliably and fast even on the older devices I have tested. What's not so great is that the fragment is quite involved and even after many rounds of refactoring doesn't have ideal code coverage. I can improve coverage further tomorrow (e.g. for `chooseOptimalSize`), but ultimately I think we need a follow-up issue to create an instrumented UI test for it.

I think this is "landable" as a first version now that it works and has basic coverage, but I need another pair of eyes to decide what should happen as a minimum :). @pocmo can you take a look? 

### Pull Request checklist


components/feature/qr/README.md
```diff
@@ -0,0 +1,42 @@
...
+```Groovy
+implementation "org.mozilla.components:feature-qr:{latest-version}"
+```
+
+### Integration
+
+Initializing the feature:
+
+```
+qrFeature = QrFeature(
```
Amejia481: nit: You could add ```kotlin, this makes the code block look a bit nicer

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFeature.kt
```diff
@@ -0,0 +1,77 @@
...
+ * @property onScanResult a callback invoked with the result of the QR scan. The callback will
+ * always be invoked on the main thread.
+ * @property onNeedToRequestPermissions a callback invoked when permissions
+ * need to be requested before a QR scan can be performed. Once the request
+ * is completed, [onPermissionsResult] needs to be invoked.
+ */
+class QrFeature(
+    private val applicationContext: Context,
+    private val fragmentManager: FragmentManager,
+    private val onScanResult: OnScanResult = { },
```
Amejia481: You could add [@MainThread](https://developer.android.com/reference/android/support/annotation/MainThread)

components/feature/qr/lint.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<lint>
+    <issue id="IconMissingDensityFolder" severity="ignore" />
+    <issue id="Overdraw" severity="ignore" />
```
pocmo: Let's suppress `Overdraw` in the file where it's actually happening. In general we want to avoid it and get notified if we add another one. :)

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFeature.kt
```diff
@@ -0,0 +1,77 @@
...
+ * @property applicationContext a reference to the application context.
+ * @property fragmentManager a reference to a [FragmentManager], used to start the [QrFragment].
+ * @property onScanResult a callback invoked with the result of the QR scan. The callback will
+ * always be invoked on the main thread.
+ * @property onNeedToRequestPermissions a callback invoked when permissions
+ * need to be requested before a QR scan can be performed. Once the request
+ * is completed, [onPermissionsResult] needs to be invoked.
+ */
+class QrFeature(
+    private val applicationContext: Context,
```
pocmo: nit: Technically this doesn't need to be the application context right? Since we pass in a fragment manager this seems to have a reference to an activity context already?

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt
```diff
@@ -0,0 +1,557 @@
...
+            return java.lang.Long.signum(lhs.width.toLong() * lhs.height - rhs.width.toLong() * rhs.height)
+        }
+    }
+
+    companion object {
+        fun newInstance(listener: OnScanCompleteListener): QrFragment {
+            val fragment = QrFragment()
+            val args = Bundle().apply {
+                putSerializable(BUNDLE_SCAN_COMPLETE_LISTENER, listener)
+            }
```
pocmo: This part seems problematic to me. The listener is a function which keeps a reference to whatever outside of the feature. Technically this can't really be serialized?

In other similar situations we:
* Create the fragment and set a reference to the feature on the fragment
* Since Android re-creates fragments when the app was destroyed in the background we handle this in the feature: If on initialization there's already a fragment in the fragment manager then we re-set the reference to the new feature.

See `DownloadsFeature` as an example.

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFeature.kt
```diff
@@ -0,0 +1,77 @@
...
+     */
+    fun scan(containerViewId: Int = android.R.id.content): Boolean {
+        this.containerViewId = containerViewId
+
+        val fragment = QrFragment.newInstance(scanCompleteListener)
+
+        return if (applicationContext.isPermissionGranted(CAMERA)) {
+            fragmentManager.beginTransaction()
+                    .replace(containerViewId, fragment)
+                    .addToBackStack(null)
```
pocmo: The backstack is a monster that can cause a bunch of unexpected situations; especially when combined with other transactions that do not use the backstack. I wonder if we should use our `BackHandler` mechanism here and ask the app to forward those events to us?

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFeature.kt
```diff
@@ -0,0 +1,77 @@
...
+     * need to be requested.
+     */
+    fun scan(containerViewId: Int = android.R.id.content): Boolean {
+        this.containerViewId = containerViewId
+
+        val fragment = QrFragment.newInstance(scanCompleteListener)
+
+        return if (applicationContext.isPermissionGranted(CAMERA)) {
+            fragmentManager.beginTransaction()
+                    .replace(containerViewId, fragment)
```
pocmo: Should we just `add` given that we do not know what else is in that container. Or do we expect the app to always pass an empty container to us?

components/feature/qr/lint.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<lint>
+    <issue id="IconMissingDensityFolder" severity="ignore" />
```
pocmo: I assume this is because of `qr_cam_focus.png`? Something's weird about that. There are x2 and x3 versions but they are all in the `hdpi` folder and not in the correct `*dpi` folder? 🤔

components/feature/qr/README.md
csadilek: 👍

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt
csadilek: @pocmo ah, this handles a different case. I should've commented. When the orientation of device changes while scanning, the fragment gets re-created and the listener would be lost otherwise. I tried a few options but only made that case work with this "workaround".

components/feature/qr/lint.xml
csadilek: heh I missed that when cleaning up. Fixing...

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt
pocmo: But that's the same case we handle in `DownloadsFeature`? I can't see how this can work without leaking everything that is referenced from the callback. And I'm confused how this can work when the process gets re-created?

csadilek: I will try this again (re-using the fragment instance instead.)

components/feature/qr/lint.xml
csadilek: Fixed the overdraw issue.

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt
csadilek: OK, this is fixed now!

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFeature.kt
csadilek: Switched to `BackHandler`!

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFeature.kt
csadilek: OK, using `add` now instead of `replace`.

components/feature/qr/lint.xml
csadilek: Fixed!

csadilek:
I've addressed all comments and added a bunch of new tests now.

@pocmo can you take another look, please?

components/feature/qr/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<resources>
+
+    <!-- Content description (not visible, for screen readers etc.): Description of an image view. -->
+    <string name="mozac_qr_scanner">QR scanner</string>
```
pocmo: nit: Following the style of the other strings that would be `mozac_feature_qr_scanner`.

components/feature/qr/src/main/res/values/styles.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<resources>
+    <style name="Theme.Mozac.Qr" parent="Theme.AppCompat.Light.DarkActionBar">
+        <item name="android:background">@android:color/background_dark</item>
+    </style>
+</resources>
```
pocmo: This part is a bit confusing to me: We are creating this theme to set the background on the `RelativeLayout`. There's technically no difference between setting the `background` on the `RelativeLayout` directly.

Is this what caused the `Overdraw` lint warning? In this case we probably can just ignore it inside `fragment_layout.xml` with something like `tools:ignore=".."`.

components/feature/qr/src/main/res/values/strings.xml
csadilek: Thanks, updated!

components/feature/qr/src/main/res/values/styles.xml
csadilek: Yeah, was just trying to satisfy the linter. I've put it on `ignore` now.

vladikoff:
Amazing!!!! So happy to see this land 🤓

PR #2540, opened 2019-03-27
Run automatic experiments updates on library init

georgf:
This is still WIP, but could use some feedback.

In short i do the following here:
- Delete the alternative `JobService`-based scheduler implementation.
- Introduced `ExperimentsUpdater`, which intends to contain the experiments update logic & scheduling. <s>This is based on `WorkManager.enqueueUniquePeriodicWork()`.</s>
- Added `Experiments` integration & initialization to the Glean sample app. This allows us testing app integration.
- Require `Glean` to be initialized first, before `Experiments`.

The rest is refactoring or workarounds resulting from that.

### Pull Request checklist


components/service/experiments/src/main/java/mozilla/components/service/experiments/ExperimentsUpdater.kt
```diff
@@ -0,0 +1,107 @@
...
+import androidx.work.*
+import java.util.concurrent.TimeUnit
+import kotlinx.coroutines.Dispatchers.IO
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.GlobalScope
+import mozilla.components.support.base.log.logger.Logger
+
+class SyncWorker(context: Context, workerParams: WorkerParameters) : androidx.work.Worker(context, workerParams) {
+    override fun doWork(): Result {
+        return if (Experiments.updater!!.updateExperiments()) {
```
georgf: This bit is a blocker for me.
I really want to not call into the static interface, as this is terrible for testing and mutual dependencies.

But how can i call back one specific instance from a `Worker`?
- I can't seem to pass an instance value to it (i see only the notion of int/string/... "in data" through `WorkerParams`.
- Making this an `inner class` of `ExperimentsUpdater` would make it possible to access it's instance-specific variables. However, the `WorkManager` code can't instantiate this because the construct of an `inner class` is not publically (or module wide?) visible, leading to a runtime exception.

georgf:
I'm looking for feedback currently, most specifically for [this part](https://github.com/mozilla-mobile/android-components/pull/2540/files/b737643472f5933eddfa4ecfc3c1d1d798cf9d1b#diff-d7554963efcc18f444711b81dc92abd2R18).
Other feedback is of course welcome though.

components/service/experiments/src/main/java/mozilla/components/service/experiments/ExperimentsUpdater.kt
Dexterp37: >     * Making this an `inner class` of `ExperimentsUpdater` would make it possible to access it's instance-specific variables. However, the `WorkManager` code can't instantiate this because the construct of an `inner class` is not publically (or module wide?) visible, leading to a runtime exception.

Mh, this is interesting. I've been using the `inner class` approach [here](https://github.com/mozilla-mobile/android-components/blob/6e99fd867d018b16e24cfbb2bd8897a774f804bc/components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt#L56) without particular problems. Is the case really different?

components/service/experiments/src/main/java/mozilla/components/service/experiments/ExperimentsUpdater.kt
```diff
@@ -0,0 +1,107 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.experiments
+
+import android.content.Context
+import android.support.annotation.VisibleForTesting
+import androidx.work.*
```
Dexterp37: To prevent Android Studio from doing this:   In the Settings/Preferences dialog (⌘,), select Code Style | Java | Imports.  Make sure that the Use single class import option is enabled.

components/service/experiments/src/main/java/mozilla/components/service/experiments/ExperimentsUpdater.kt
georgf: Hm, ok... let me push that to compare.

components/service/experiments/src/main/java/mozilla/components/service/experiments/ExperimentsUpdater.kt
```diff
@@ -0,0 +1,110 @@
...
+            .setRequiredNetworkType(NetworkType.CONNECTED)
+            .build()
+        /*
+        val syncWork = PeriodicWorkRequest.Builder(SyncWorker::class.java, UPDATE_INTERVAL_HOURS, TimeUnit.HOURS)
+            .addTag(TAG)
+            .setConstraints(constraints)
+            .build()
+        WorkManager.getInstance().enqueueUniquePeriodicWork(TAG, ExistingPeriodicWorkPolicy.KEEP, syncWork)
+        */
+        WorkManager.getInstance().enqueue(OneTimeWorkRequest.from(SyncWorker::class.java))
```
georgf: @Dexterp37  To simplify, let's just use a one-time request, to highlight the issue.

Running this from the Glean sample results in a run-time error:

> 2019-03-27 18:14:28.245 27026-27044/org.mozilla.samples.glean.debug E/WM-WorkerFactory: Could not instantiate mozilla.components.service.experiments.ExperimentsUpdater$SyncWorker
>     java.lang.NoSuchMethodException: <init> [class android.content.Context, class androidx.work.WorkerParameters]
>         at java.lang.Class.getConstructor0(Class.java:2327)
>         at java.lang.Class.getDeclaredConstructor(Class.java:2166)
>         at androidx.work.WorkerFactory.createWorkerWithDefaultFallback(WorkerFactory.java:91)
>         at androidx.work.impl.WorkerWrapper.runWorker(WorkerWrapper.java:233)
>         at androidx.work.impl.WorkerWrapper.run(WorkerWrapper.java:127)
>         at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
>         at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
>         at java.lang.Thread.run(Thread.java:764)
> 2019-03-27 18:14:28.245 27026-27044/org.mozilla.samples.glean.debug E/WM-WorkerWrapper: Could not create Worker mozilla.components.service.experiments.ExperimentsUpdater$SyncWorker
>

georgf: My assumption was... that `SyncWorker`s constructor is not visible to the factory/workmanager code.

Dexterp37: Gah, you're perfectly right, and I just had a closer look at glean's code: it's erroring out there as well.

Dexterp37: Ok, I dug into this a bit more. We can only really pass strings and other basic types to a `Worker` via [`Data`](https://developer.android.com/reference/androidx/work/Data.html) (as @travis79 knows). In order to have more complex injection in place, we'd need to provide the `WorkManager` with a custom [`WorkerFactory`](https://developer.android.com/reference/androidx/work/WorkerFactory). This would allow us to instantiate Workers however we want.

**However**, in order to enable a custom `WorkerFactory` we'd need to disable the automatic `WorkManager` init in the (app? library?) manifest and [manually init](https://developer.android.com/reference/androidx/work/WorkManager.html#initialize(android.content.Context,%20androidx.work.Configuration)) it by passing our custom configuration object with the custom `WorkerFactory`.

**Problem** I'm not sure we can do that from the library, and I'm not sure we should ask consumers to manually disable the `WorkManager` init from the library. These are uncharted waters for me, @pocmo might know better. The fallback option is to do what you didn't want to, calling static APIs.

travis79: Disabling the default WorkManager init may have other consequences if other bits of the client app are using WorkManager.  I'm not sure that's a safe bet as I don't know even if we did it from the library it wouldn't have some effect on the client app.

I think if this is going to work with the way we have it structured now, then we need to see if there's a way to serialize/reconstruct the object from the Worker class.  This worked well with the PingUploadWorker serializing pings and just reading them from disk.  Is there any way to create a data class around the bits you need for the Worker to function, and then use the [`@Parcelize`](https://github.com/Kotlin/KEEP/blob/master/proposals/extensions/android-parcelable.md) functionality to pass it to the Worker?

pocmo: Yeah, the work manager can't create an instance of your class since it is an inner class, which means it can only be created inside `ExperimentsUpdater` ... because it keeps a reference to its outer class - and that's why you can call updateExperiments() in the outer class.

Here's what I suggest to do:
* Create just a normal class that extends Worker.
* Inside `doWork()` you can access your `Experiments` singleton and from there you should get to your `updateExperiments()` (or add a way to get there).

There's no need to pass anything to the worker here...?

georgf: @pocmo I was wondering how to decouple this from the singleton for unit testing.
But maybe i should just live with this for now?

pocmo: @georgf Yeah, singletons are always a pain in tests. It's one of the reasons why in other components we leave creating the singleton to the app (with the help of a DI framework if the app chooses so).

For tests in other situations we tried to hide the singleton part behind a method that we can mock in a test (e.g. if we can't mock System.currentTimeMillis() then we hide this behind a local now() method that we can mock).

georgf: Interesting, do you have a link to such a function and its mocks in tests handy? :)

travis79: @georgf I can point you at an example like he's talking about: [this](https://github.com/mozilla-mobile/android-components/blob/761f43462120913f2d9e0e00c8d73e37dd9a214d/components/service/telemetry/src/main/java/org/mozilla/telemetry/schedule/jobscheduler/TelemetryJobService.java#L139)

It's mocked in the `testDailyLimitIsEnforced()` test [here](https://github.com/mozilla-mobile/android-components/blob/210751ad2b02639cbfc729b0e96a0c2dc204e846/components/service/telemetry/src/test/java/org/mozilla/telemetry/schedule/jobscheduler/TelemetryJobServiceTest.java#L72)

The Kotlin version should be pretty close to that.  I think we have some examples of mocking functions using the `when` syntax in service-glean already, such as [here](https://github.com/mozilla-mobile/android-components/blob/1c017bf3d87b4ed738bd5c9a2f809cc5ce2b0ee8/components/service/glean/src/test/java/mozilla/components/service/glean/TestUtil.kt#L155)

georgf: Cheers!

georgf:
This is ready for a review pass.

An overview of what's in this PR:
- Integrate this library into the Glean sample app, to have an integration test we can run.
- On library init (`Experiments.initialize()`)
  - automatically load experiments from storage
  - automatically start async updating of the experiments list from the server (from Kinto, triggered by `ExperimentsUpdater`)
- It does not contain periodic updating of experiments yet (to reduce the scope here, i broke this out to [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1541973)).
- Flatten the `ExperimentSource` and `ExperimentStorage` hierarchies. They just have one implementation now and are not configurable.
- Restructure the testing/mocking approach & clean up `ExperimentsTest.kt` significantly.
  - Most importantly, use `ExperimentsInternalAPI` testable using instance & don't use the singleton in tests.
- Start depending on Glean. I could break that part out if that helps with review.

components/service/experiments/README.md
```diff
@@ -12,137 +12,44 @@ Use Gradle to download the library from [maven.mozilla.org](https://maven.mozill
...
+            applicationContext,
+            configuration // This is optional, e.g. for overriding the fetch client.
         )
     }
 }
 ```
 
-#### Using Kinto as experiment source
-Experiments includes a default source implementation for a Kinto backend, which you can use like this:
+Note that this library depends on the Glean library, which has to be updated first. See the [Glean README](../glean/README.md) for more details.
```
Dexterp37: What do you mean by glean needs to be "updated first"?

components/service/experiments/README.md
```diff
@@ -155,67 +62,8 @@ Experiments allows you to specify the following filters:
...
-        // save snapshot to disk
-    }
-
-    override fun retrieve(): ExperimentsSnapshot {
-        // load snapshot from disk
-        return snapshot
-    }
-}
-```
+- *(TBD: region: custom region, different from the one from the default locale (like a GeoIP, or something similar).)*
```
Dexterp37: nit: is there any bug filed for these TBD that you can link here?

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
```diff
@@ -87,12 +93,29 @@ open class ExperimentsInternalAPI internal constructor() {
...
-        source = null
+        updater?.let {
+            it.source = null
+        }
         storage = null
 
         isInitialized = false
     }
 
+    internal fun getGlean(): GleanInternalAPI {
```
Dexterp37: nit: is it useful to annotate this with `VisibleForTesting` as well?

samples/glean/src/main/java/org/mozilla/samples/glean/GleanApplication.kt
```diff
@@ -22,6 +23,9 @@ class GleanApplication : Application() {
         // Initialize the Glean library. Ideally, this is the first thing that
         // must be done right after enabling logging.
         Glean.initialize(applicationContext)
+        // Initialize the Experiments library right afterwards. Experiments can
+        // not be activated before this, so it's important to do this early.
+        Experiments.initialize(applicationContext)
```
Dexterp37: Let's also update the sample `README.md` file to mention that we're also showcasing the experiments library here.

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
```diff
@@ -47,34 +48,39 @@ open class ExperimentsInternalAPI internal constructor() {
...
-        fetchClient: Client = HttpURLConnectionClient()
+        configuration: Configuration = Configuration()
     ) {
         if (isInitialized) {
             logger.error("Experiments library should not be initialized multiple times")
             return
         }
 
+        // Any code below might trigger recording into Glean, so make sure Glean is initialized.
+        if (!getGlean().isInitialized()) {
```
Dexterp37: Do we want to really stop init in this case? Glean doesn't crash if a recording is attempted before init is complete: we'll simply discard, if it's *too* early. Maybe printing the error is enough?

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
```diff
@@ -47,34 +48,39 @@ open class ExperimentsInternalAPI internal constructor() {
...
+        // is available.
+        loadExperiments()
+
+        // We now have the last known experiment state loaded for product code
+        // that needs to check it early in startup.
+        // Next we need to update the experiments list from the server async,
+        // without blocking the app launch and schedule future periodic
+        // updates.
+        updater = getExperimentsUpdater(applicationContext)
+        updater!!.initialize(configuration)
```
Dexterp37: nit: note that this will throw a null-pointer exception in the very unlikely case that `updater` is null at this point. In other places below we're using `updater?.let {}`.

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
```diff
@@ -87,12 +93,29 @@ open class ExperimentsInternalAPI internal constructor() {
...
         storage = null
 
         isInitialized = false
     }
 
+    internal fun getGlean(): GleanInternalAPI {
+        return Glean
+    }
+
+    internal open fun getExperimentsUpdater(context: Context): ExperimentsUpdater {
```
Dexterp37: Why is this function `open`? I can't seem to find a place in this PR in which it is overriden. Are you planning a follow-up PR for using this?

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
```diff
@@ -87,12 +93,29 @@ open class ExperimentsInternalAPI internal constructor() {
...
+        return ExperimentsUpdater(context, this)
+    }
+
+    internal open fun getExperimentsStorage(context: Context): FlatFileExperimentStorage {
```
Dexterp37: Why is this function `open`? I can't seem to find a place in this PR in which it is overriden. Are you planning a follow-up PR for using this?

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
```diff
@@ -47,34 +48,39 @@ open class ExperimentsInternalAPI internal constructor() {
      * @param applicationContext [Context] to access application features, such
      * as shared preferences.
      */
-    fun initialize(
+    open fun initialize(
```
Dexterp37: Why is this function `open`? I can't seem to find a place in this PR in which it is overriden. Are you planning a follow-up PR for using this?

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
```diff
@@ -116,23 +139,11 @@ open class ExperimentsInternalAPI internal constructor() {
      * saves them to local storage
      */
     @Synchronized
-    internal fun updateExperiments(): Boolean {
-        if (!experimentsLoaded) {
-            loadExperiments()
-        }
+    internal fun onExperimentsUpdated(serverState: ExperimentsSnapshot) {
+        assert(experimentsLoaded)
```
Dexterp37: `assert` in Kotlin is [only enabled](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/assert.html) if "runtime assertions have been enabled on the JVM using the -ea JVM option.". This is usually the case in tests, as far as I can tell. If you want this code to crash with an useful message at runtime, you can use [`check`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/check.html).

components/service/experiments/README.md
georgf: That should be "initialized".

components/service/experiments/README.md
georgf: Not at this point, it's an open question/conversation in the planning doc.

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
georgf: Good point.

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
georgf: Good point on this, those are left-overs.

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
georgf: This is what i expect here. The intent is to have issues show up earlier and more clear in testing.

I don't know if that's the best pattern though, so i'm open to feedback on this.

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
georgf: With the larger testing refactoring, `updater` & `storage` can actually just be `lateinit`. Changing this.

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
georgf: Not doing this would make Glean an optional dependency.
What i want to make sure is that all experimentation data is recorded, to make semantics clear.

This could technically initialize here, but i want to enforce Glean as a strict dependency to reduce potential for errors.

Dexterp37: I see now, makes sense then!

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
Dexterp37: If it's testing, then this is fine as it is. I might suggest adding an assertion message as, sometimes, we're unable to get the file/line of the assertions on CI. I'm not sure why that happens, but it happens.

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
```diff
@@ -111,6 +101,21 @@ open class ExperimentsInternalAPI internal constructor() {
         isInitialized = false
     }
 
+    internal fun getGlean(): GleanInternalAPI {
+        return Glean
+    }
+
+    internal open fun getExperimentsUpdater(context: Context): ExperimentsUpdater {
```
travis79: If this function isn't used outside of this class I suggest adding the `@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)` annotation.

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
```diff
@@ -111,6 +101,21 @@ open class ExperimentsInternalAPI internal constructor() {
...
+        return ExperimentsUpdater(context, this)
+    }
+
+    internal open fun getExperimentsStorage(context: Context): FlatFileExperimentStorage {
```
travis79: Same nit as above with `@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)` annotation.

components/service/experiments/src/main/java/mozilla/components/service/experiments/ExperimentsUpdater.kt
```diff
@@ -40,44 +23,22 @@ internal class ExperimentsUpdater(
...
-        val syncWork = PeriodicWorkRequest.Builder(SyncWorker::class.java, UPDATE_INTERVAL_HOURS, TimeUnit.HOURS)
-            .addTag(TAG)
-            .setConstraints(constraints)
-            .build()
-        WorkManager.getInstance().enqueueUniquePeriodicWork(TAG, ExistingPeriodicWorkPolicy.KEEP, syncWork)
-        */
-        //WorkManager.getInstance().enqueue(OneTimeWorkRequest.from(SyncWorker::class.java))
-    }
-
-    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
```
travis79: nit, unless you are planning on using this outside of this class, you could leave the `@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)` annotation.

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
georgf: Good point, ok.

PR #2542, opened 2019-03-27
1539480: Copy the user-lifetime metrics from old to new location

mdboom:
If this isn't the first run, it looks for the two user-lifetime metrics in
the old location of `glean_ping_info` and copies them to `glean_client_info`.
It is not problematic that the old values remain in `glean_ping_info`, since
the contents of `ping_info` are hard-coded and don't include these values.

### Pull Request checklist


components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -218,6 +219,26 @@ open class GleanInternalAPI internal constructor () {
             val uuid = UUID.randomUUID()
             UuidsStorageEngine.record(GleanInternalMetrics.clientId, uuid)
             DatetimesStorageEngine.set(GleanInternalMetrics.firstRunDate)
+        } else {
+            // ////////////////////////////////////////////////////////////
+            // BACKWARD COMPATIBILITY HACK
```
Dexterp37: nit: I would point out that this is due to bug 1528787

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -218,6 +219,26 @@ open class GleanInternalAPI internal constructor () {
...
+            // 1539480: This moves the two user-lifetime metrics, client_id, and
+            // first_run_date from their old locations in the glean_ping_info
+            // store to the glean_client_info store.  Failure to do so means
+            // that the client_id won't be sent in the ping, and the ping will
+            // then be rejected by the pipeline.
+            val uuidSnapshot = UuidsStorageEngine.getSnapshot("glean_ping_info", false)
```
Dexterp37: Can we add test coverage for this?

PR #2546, opened 2019-03-28
Closes #175: BrowserToolbar: Add support for "edit actions"..

### LINKED ISSUE #175: Toolbar: Add support for editing actions
Structure of the EditToolbar:
```

 +---------------------------------+---------+------+
 |                 url             | actions | exit |
 +---------------------------------+---------+------+
```

Some components may want to add actions that show up editing mode.

Examples:
* Barcode scanner
* Microphone input

pocmo:
### Pull Request checklist

### DIFF
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
@@ -399,6 +399,7 @@ class BrowserToolbar @JvmOverloads constructor(
      */
     fun invalidateActions() {
         displayToolbar.invalidateActions()
+        editToolbar.invalidateActions()
     }
 
     /**
@@ -431,6 +432,13 @@ class BrowserToolbar @JvmOverloads constructor(
      */
     override fun addNavigationAction(action: Toolbar.Action) {
         displayToolbar.addNavigationAction(action)
+    }
+
+    /**
+     * Adds an action to be displayed on the right of the URL in edit mode.
+     */
+    override fun addEditAction(action: Toolbar.Action) {
+        editToolbar.addEditAction(action)
     }
 
     /**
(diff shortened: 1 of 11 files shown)


components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/SubToolbar.kt
```diff
@@ -0,0 +1,88 @@
...
+import android.view.View
+import mozilla.components.browser.toolbar.display.DisplayToolbar
+import mozilla.components.browser.toolbar.edit.EditToolbar
+import android.view.ViewGroup
+import mozilla.components.concept.toolbar.Toolbar
+
+/**
+ * Abstract base class for shared code between [DisplayToolbar] and [EditToolbar].
+ */
+internal abstract class SubToolbar(
```
csadilek: nit: This name confused me a bit i.e. why is it a `Sub` toolbar if `Display` and `Edit` toolbar extend it? Maybe just `AbstractToolbar`?

pocmo: I took `SubToolbar` since it's only extended by the two "sub" components and not the toolbar itself.

```
            +----------------+
            | BrowserToolbar |
            +--------+-------+
                     +
             +-------+-------+
             |               |
   +---------v------+ +-------v--------+
   | DisplayToolbar | |   EditToolbar  |
   +----------------+ +----------------+
```

I feel like `AbstractToolbar` sounds like it's used by the `BrowserToolbar` which implements `Toolbar`. Confusing :D

pocmo: Actually nothing in `SubToolbar` requires any state or requires this to be an abstract class. I think I'm going to move the things in it to be just functions.. 🤔

pocmo:
@csadilek Updated the PR: This now uses (extension) functions and no shared abstract class anymore.

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/SubToolbar.kt
csadilek: Ah ok! I didn't get that :). The new version looks great though!

pocmo:
Ah, in `feature-toolbar` we have some tests that create a fake toolbar. I'll need to add the new methods to them. That has happened more than once in the past..

PR #2561, opened 2019-03-29
Fix #2344: 1535258: Fix dependency on Miniconda install on build dir.

### LINKED ISSUE #2344: :service-glean:Bootstrap_CONDA_'Miniconda3' task does not execute in isolation
STR:
- Run `./gradlew clean :service-glean:Bootstrap_CONDA_\'Miniconda3\'`

Expected: task completes successfully
Actual: task fails with:

```
Local configuration: loaded local.properties
Local configuration: application-services substitution path missing. Specify it via 'substitutions.application-services.dir' setting.

> Task :service-glean:Bootstrap_CONDA_'Miniconda3' FAILED
Downloading Miniconda3-4.5.11-MacOSX-x86_64.sh

* What went wrong:
Execution failed for task ':service-glean:Bootstrap_CONDA_'Miniconda3''.
> java.io.FileNotFoundException: /Users/mcomella/dev/moz/android-components/components/service/glean/build/Miniconda3-4.5.11-MacOSX-x86_64.sh (No such file or directory)
```

Speculation: it might be assuming other tasks create the build directory hierarchy before it runs, in which case it should not assume that or should declare other tasks as a dependency.

CC @Dexterp37 on glean's behalf

mdboom:
This properly creates a dependency from the Miniconda install task to a task to create glean's build dir.  Unfortunately the `gradle-python-envs` plugin from JetBrains that we use for this doesn't allow the specifying a "work directory" to download the installer, otherwise I would have preferred that.

@mcomella : Please confirm that this addresses your original use case.

### Pull Request checklist


components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -100,6 +90,21 @@ task createBuildDir {
...
+envs {
+    bootstrapDirectory = GLEAN_BOOTSTRAP_DIR
+    pipInstallOptions = "--trusted-host pypi.python.org --no-cache-dir"
+
+    // Setup a miniconda environment. conda is used because it works
+    // non-interactively on Windows, unlike the other options
+    conda "Miniconda3", "Miniconda3-${MINICONDA_VERSION}", "64", ["glean_parser==${GLEAN_PARSER_VERSION}"]
+}
+
+tasks.whenTaskAdded { task ->
```
Dexterp37: Can you add a comment here explaining why this is required?

mdboom: Sure.  I added detail to the already-existing comment above about why JetBrain's suggested solution isn't sufficient.

mcomella:
> @mcomella : Please confirm that this addresses your original use case.

I confirmed that the STR I suggested in #2344 works as recommended.

However, I noticed `./gradlew compiledebugkotlin --parallel` is still broken:
```
> Task :samples-glean-library:createBuildDir
Creating build directory:/Users/mcomella/dev/moz/android-components/samples/glean/samples-glean-library/build

> Task :service-glean:createBuildDir
Creating build directory:/Users/mcomella/dev/moz/android-components/components/service/glean/build

> Task :samples-glean:Bootstrap_CONDA_'Miniconda3'
Downloading Miniconda3-4.5.11-MacOSX-x86_64.sh
Bootstraping to /Users/mcomella/.gradle/glean/bootstrap-4.5.11/Miniconda3
ERROR: File or directory already exists: '/Users/mcomella/.gradle/glean/bootstrap-4.5.11/Miniconda3'
If you want to update an existing installation, use the -u option.

> Task :samples-glean:Bootstrap_CONDA_'Miniconda3' FAILED

> Task :samples-glean-library:Bootstrap_CONDA_'Miniconda3'
Downloading Miniconda3-4.5.11-MacOSX-x86_64.sh
Bootstraping to /Users/mcomella/.gradle/glean/bootstrap-4.5.11/Miniconda3
ERROR: File or directory already exists: '/Users/mcomella/.gradle/glean/bootstrap-4.5.11/Miniconda3'
If you want to update an existing installation, use the -u option.

> Task :samples-glean-library:Bootstrap_CONDA_'Miniconda3' FAILED

> Task :service-glean:Bootstrap_CONDA_'Miniconda3'
```

This is a race condition so, in theory, this may not always fail. It looks like the bootstrap task is assigned to multiple modules and, I assume, isn't designed to run in parallel. While running in parallel isn't a requirement for a-c (at least I assume so since someone disabled it), it'd be nice to not break it. Please file a follow-up bug or fix it here. I don't know what the proper Gradle way to fix this is but one way would be to define the bootstrap task to the root module and have each submodule tasks depend on this task: I think that'd guarantee it only runs once and not in parallel.

mdboom:
@mcomella: I'm able to reproduce the problem with `./gradlew compiledebugkotlin --parallel`.

It does seem to work for rebuilds, however, so if your goal is just to have faster rebuilds, you should be ok as long as the first build is non-parallel.

Obviously, ideally we'd want this to work on first build, too -- I'll look into what you suggest.

PR #2523, opened 2019-03-26
[glean] 1536792: Reinstate metrics ping scheduler tests.

mdboom:
This is a draft PR while I debug and re-examine the `MetricsPingScheduler` tests that were turned off last week.

### Pull Request checklist

- [ ] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/android-components/blob/master/docs/changelog.md) or does not need one
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features

PR #2587, opened 2019-04-01
[glean] Move `MetricsPingWorker` out of `MetricsPingScheduler`

Dexterp37:
The `MetricsPingWorker` was an inner class of the metrics ping scheduler. However, `WorkManager` was unable to instantiate it. This solves the issue by moving the worker out of the scheduler class, removing the inner qualifier. This additionally restores some tests that were previously disabled.

In order to still be able to call the functions from `MetricsPingWorker`, the worker fetches the active instance of the metrics ping scheduler through the `Glean` singleton. This is ugly, other options are documented within the code.

Please note that I've attempted again to add test coverage for the `WorkManager`, but this is still failing. The test mostly comes from Google's [official guidelines](https://developer.android.com/topic/libraries/architecture/workmanager/how-to/testing)

```
    @Test
    @Throws(Exception::class)
    fun `MetricsPingWorker can get scheduled and executed`() {
        // This is required so that `Glean.metricsPingScheduler` contains a valid value.
        resetGlean()

        val testDriver = WorkManagerTestInitHelper.getTestDriver()
        val workManager = WorkManager.getInstance()
        val mps = Glean.metricsPingScheduler
        val (request, operation) = mps.enqueueWorker(workManager, 2000)

        // Wait for the result of the worker to come in. This will implicitly
        // wait for the worker to finish execution. See
        // https://developer.android.com/topic/libraries/architecture/workmanager/how-to/testing
        operation.result.get()
        testDriver.setInitialDelayMet(request.id)

        // Get the WorkInfo for the executed worker.
        val workInfo = workManager.getWorkInfoById(request.id).get()
        assertEquals(WorkInfo.State.SUCCEEDED, workInfo.state)
  }
```

Still, `request.id` won't be found. I think this might have to do with the fact that we're using the Robolectric test runner.

### Pull Request checklist


Dexterp37:
@mdboom @travis79 what's your take on the current approach? Any alternative suggestion?

Dexterp37:
Interkittens spin-wheel: 1 successful - 0 failed

mdboom:
Just wanted to link this to [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1536792) and this PR #2523, which are also (currently failing) attempts to reinstate these tests.  I think this is a better and further along approach -- let's work here and then close those bugs and PRs as appropriate...

mdboom:
I think this approach is probably the best of not-great options.  We certainly can go with this for now if it helps us pass more tests ;)

PR #2588, opened 2019-04-01
Issue #2571: Add option to recover from crashed sessions...

### LINKED ISSUE #2571: Allow apps to recover crashed sessions by restoring last state
With #2569 and #2568 the app knows when a `Session` (tab) crashed. Since the state refactoring (#2497) we always have a reference to the last state. We can use that to allow the app to restore this last state after crash.

pocmo:
Closes #2571. This is the last piece for better crash handling support.

### Pull Request checklist

### DIFF
--- a/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
+++ b/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
@@ -245,6 +245,13 @@ abstract class EngineSession(
     abstract fun exitFullScreenMode()
 
     /**
+     * Tries to recover from a crash by restoring the last know state.
+     *
+     * Returns true if a last known state was restored, otherwise false.
+     */
+    abstract fun recoverFromCrash(): Boolean
+
+    /**
      * Close the session. This may free underlying objects. Call this when you are finished using
      * this session.
      */
--- a/components/feature/session/src/main/java/mozilla/components/feature/session/SessionUseCases.kt
+++ b/components/feature/session/src/main/java/mozilla/components/feature/session/SessionUseCases.kt
@@ -174,6 +174,22 @@ class SessionUseCases(
         }
     }
 
+    /**
+     * Tries to recover from a crash by restoring the last know state.
+     */
+    class CrashRecoveryUseCase internal constructor(
+        private val sessionManager: SessionManager
+    ) {
+        fun invoke(session: Session): Boolean {
+            val recovered = sessionManager.getOrCreateEngineSession(session)
+                .recoverFromCrash()
+
+            session.crashed = false
+
+            return recovered
+        }
+    }
+
     val loadUrl: DefaultLoadUrlUseCase by lazy { DefaultLoadUrlUseCase(sessionManager, onNoSession) }
     val loadData: LoadDataUseCase by lazy { LoadDataUseCase(sessionManager, onNoSession) }
     val reload: ReloadUrlUseCase by lazy { ReloadUrlUseCase(sessionManager) }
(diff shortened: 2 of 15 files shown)


components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
```diff
@@ -244,6 +244,11 @@ abstract class EngineSession(
      */
     abstract fun exitFullScreenMode()
 
+    /**
+     * Tries to recover from a crash by restoring the last know state.
+     */
+    abstract fun recoverFromCrash()
```
Amejia481: Question: As the comment implies  **Tries to recover**, that means that in some cases it won't be able to recover from a crash, should we indicate to the consumer when we are unable to recover like returning a Boolean? This way the consumer can handle the situation properly.

Amejia481: Or  `session.crashed` is the way to communicate it ?

components/feature/session/src/main/java/mozilla/components/feature/session/SessionUseCases.kt
```diff
@@ -174,6 +174,20 @@ class SessionUseCases(
...
+     * Tries to recover from a crash by restoring the last know state.
+     */
+    class CrashRecoveryUseCase internal constructor(
+        private val sessionManager: SessionManager
+    ) {
+        fun invoke(session: Session) {
+            sessionManager.getOrCreateEngineSession(session)
+                .recoverFromCrash()
+
+            session.crashed = false
```
Amejia481: What happens when we are unable to recover? Are we communicating a wrong message?

components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
pocmo: We could return a `Boolean` indicating whether we had a state and whether we tried to restore that. I'll add that.

components/feature/session/src/main/java/mozilla/components/feature/session/SessionUseCases.kt
pocmo: I'll add the `Boolean` above.

Here in this case we tried to recover and that's all we could do. This doesn't mean the Session was fully restored - but we handled the crash. And then Session is operational and you can continue to load sites.

components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
Amejia481: 👍

pocmo:
PR updated. :)

components/feature/session/src/main/java/mozilla/components/feature/session/SessionUseCases.kt
Amejia481: 👍

Amejia481:
That was fast  "_"

Amejia481:
Some errors on TC

```
> Task :browser-session:compileDebugUnitTestKotlin FAILED
e: /build/android-components/components/browser/session/src/test/java/mozilla/components/browser/session/engine/EngineObserverTest.kt: (36, 29): Object is not abstract and does not implement abstract base class member public abstract fun recoverFromCrash(): Boolean defined in mozilla.components.concept.engine.EngineSession
e: /build/android-components/components/browser/session/src/test/java/mozilla/components/browser/session/engine/EngineObserverTest.kt: (86, 29): Object is not abstract and does not implement abstract base class member public abstract fun recoverFromCrash(): Boolean defined in mozilla.components.concept.engine.EngineSession
e: /build/android-components/components/browser/session/src/test/java/mozilla/components/browser/session/engine/EngineObserverTest.kt: (125, 29): Object is not abstract and does not implement abstract base class member public abstract fun recoverFromCrash(): Boolean defined in mozilla.components.concept.engine.EngineSession
```

pocmo:
Updated once more :)

pocmo:
Ah, one more test in `concept-engine`.

PR #2594, opened 2019-04-01
[glean] 1513616: Provide an API to clear user lifetime metrics

mdboom:
### Pull Request checklist

### DIFF
--- a/components/service/glean/README.md
+++ b/components/service/glean/README.md
@@ -95,6 +95,18 @@ object parameter of the `Glean.initialize` method. For example:
 Glean.initialize(applicationContext, Configuration(channel = "beta"))
 ```
 
+### Clearing user-lifetime metrics
+
+"User-lifetime metrics" are metrics that are persistent for the lifetime of the
+user profile. Your app may need to reset these metrics when the profile is
+cleared, or, depending on policy, when telemetry is disabled.
+
+To clear user-lifetime metrics, call:
+
+```
+Glean.clearUserLifetimeMetrics()
+```
+
 ### Adding new metrics
 
 All metrics that your application collects must be defined in a `metrics.yaml`
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericStorageEngine.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericStorageEngine.kt
@@ -37,6 +37,7 @@ internal typealias MetricsCombiner<T> = (currentValue: T?, newValue: T) -> T
  * A base class for common metric storage functionality. This allows sharing the common
  * store managing and lifetime behaviours.
  */
+@Suppress("TooManyFunctions")
 internal abstract class GenericStorageEngine<MetricType> : StorageEngine {
     override lateinit var applicationContext: Context
 
@@ -288,6 +289,11 @@ internal abstract class GenericStorageEngine<MetricType> : StorageEngine {
             }
         }
         editor?.apply()
+    }
+
+    override fun clearUserLifetimeMetrics() {
+        userLifetimeStorage.edit().clear().apply()
+        dataStores[Lifetime.User.ordinal].clear()
     }
 
     @VisibleForTesting(otherwise = VisibleForTesting.NONE)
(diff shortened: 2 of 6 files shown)


components/service/glean/README.md
```diff
@@ -95,6 +95,18 @@ object parameter of the `Glean.initialize` method. For example:
...
+"User-lifetime metrics" are metrics that are persistent for the lifetime of the
+user profile. Your app may need to reset these metrics when the profile is
+cleared, or, depending on policy, when telemetry is disabled.
+
+To clear user-lifetime metrics, call:
+
+```
+Glean.clearUserLifetimeMetrics()
+```
+
```
travis79: Do we also want to add a warning label here, as in `THIS IS IRREVERSIBLE`?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/StorageEngine.kt
```diff
@@ -36,6 +36,13 @@ internal interface StorageEngine {
      */
     fun getIdentifiersInStores(stores: List<String>): Sequence<String> = sequence {}
 
+    /**
+    * Clear all user lifetime metrics
+    */
+    fun clearUserLifetimeMetrics() {
+        // intentionally blank for stores that don't inherit from GenericStorageEngine
```
travis79: Nice!

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericStorageEngine.kt
```diff
@@ -290,6 +291,11 @@ internal abstract class GenericStorageEngine<MetricType> : StorageEngine {
         editor?.apply()
     }
 
+    override fun clearUserLifetimeMetrics() {
+        userLifetimeStorage.edit().clear().apply()
+        dataStores[Lifetime.User.ordinal].clear()
```
travis79: Should we also have a way to clear or reset the firstRunFlag()?  My worry is that once things like the client_id are cleared, it will be difficult to re-populate those values without somehow resetting the first-run flag.

components/service/glean/README.md
mdboom: Good point.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericStorageEngine.kt
travis79: I kinda like the idea of using the client_id as a first run flag...  and we just don't set it if glean is disabled to keep it from regenerating.

components/service/glean/README.md
```diff
@@ -95,6 +95,18 @@ object parameter of the `Glean.initialize` method. For example:
 Glean.initialize(applicationContext, Configuration(channel = "beta"))
 ```
 
+### Clearing user-lifetime metrics
+
+"User-lifetime metrics" are metrics that are persistent for the lifetime of the
+user profile. Your app may need to reset these metrics when the profile is
+cleared, or, depending on policy, when telemetry is disabled.
+
+To clear user-lifetime metrics, call:
```
Dexterp37: I think this requires a bit of design or discussion: this is a product-level feature that *might* happen automatically when turning telemetry off, without requiring any additional call from users and thus reducing our API surface.

You might want to check in with @georgf about this.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericStorageEngine.kt
Dexterp37: This is a very good point!

I think I like the idea of ditching the file based "first run detector".

components/service/glean/README.md
georgf: What are the use-cases when we would expect this to happen?

I would like exposing the resetting of user lifetime metrics outside of Glean, this is easy to use incorrectly.
I'd rather see us expose functions that have a clear purpose and intent.

mdboom:
Closing as an incorrect solution due to recent discussion in the bug.

Issue #2563, opened 2019-03-29
Crash in QR scanner when camera no longer available


csadilek:
This can happen temporarily and needs to be caught / logged. Right now it brings down the app.

```
 android.hardware.camera2.CameraAccessException: CAMERA_DISCONNECTED (2): submitRequestList:137: Camera device no longer alive
        at android.hardware.camera2.CameraManager.throwAsPublicException(CameraManager.java:660)
        at android.hardware.camera2.impl.ICameraDeviceUserWrapper.submitRequestList(ICameraDeviceUserWrapper.java:86)
        at android.hardware.camera2.impl.CameraDeviceImpl.submitCaptureRequest(CameraDeviceImpl.java:899)
        at android.hardware.camera2.impl.CameraDeviceImpl.setRepeatingRequest(CameraDeviceImpl.java:938)
        at android.hardware.camera2.impl.CameraCaptureSessionImpl.setRepeatingRequest(CameraCaptureSessionImpl.java:243)
        at mozilla.components.feature.qr.QrFragment$createCameraPreviewSession$$inlined$let$lambda$1.onConfigured(QrFragment.kt:428)
        at java.lang.reflect.Method.invoke(Native Method)
        at android.hardware.camera2.dispatch.InvokeDispatcher.dispatch(InvokeDispatcher.java:39)
        at android.hardware.camera2.dispatch.HandlerDispatcher$1.run(HandlerDispatcher.java:65)
        at android.os.Handler.handleCallback(Handler.java:751)
        at android.os.Handler.dispatchMessage(Handler.java:95)
        at android.os.Looper.loop(Looper.java:154)
        at android.os.HandlerThread.run(HandlerThread.java:61)
     Caused by: android.os.ServiceSpecificException: submitRequestList:137: Camera device no longer alive
        at android.os.Parcel.readException(Parcel.java:1698)
        at android.os.Parcel.readException(Parcel.java:1637)
        at android.hardware.camera2.ICameraDeviceUser$Stub$Proxy.submitRequestList(ICameraDeviceUser.java:334)
        at android.hardware.camera2.impl.ICameraDeviceUserWrapper.submitRequestList(ICameraDeviceUserWrapper.java:84)
```

PR #2611, opened 2019-04-03
Issue #2563: Handle CameraAccessException when opening camera

csadilek:
This is a follow-up to #2563 to also catch and log `CameraAccessException` when opening the camera. This also resolves https://github.com/mozilla-mobile/reference-browser/issues/698 as well, but I couldn't reproduce it. In either case it's better to log this error than to throw and crash the app.


csadilek:
Marking this as blocked. Want to play around with this more to see if I can reproduce the r-b issue.

PR #2617, opened 2019-04-03
[ci skip] 1540134: Doc that the sample application id changes w/ variant

mdboom:
### Pull Request checklist

### DIFF
--- a/components/service/glean/README.md
+++ b/components/service/glean/README.md
@@ -132,13 +132,14 @@ Glean exports the [`GleanDebugActivity`](src/main/java/mozilla/components/servic
 that can be used to toggle debugging features on or off. Users can invoke this special activity, at
 run-time, using the following [`adb`](https://developer.android.com/studio/command-line/adb) command:
 
-`adb shell am start -n [package.name]/mozilla.components.service.glean.debug.GleanDebugActivity [extra keys]`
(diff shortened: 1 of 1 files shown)


components/service/glean/README.md
```diff
@@ -134,18 +134,24 @@ run-time, using the following [`adb`](https://developer.android.com/studio/comma
...
-|key|type|description|
-|---|----|-----------|
-| logPings | boolean (--ez) | If set to `true`, glean dumps pings to logcat; defaults to `false` |
-| sendPing | string (--es) | Sends the ping with the given name immediately |
-| tagPings | string (--es) | Tags all outgoing pings as debug pings to make them available for real-time validation. The value must match the pattern `[a-zA-Z0-9-]{1,20}` |
-
-For example, to direct the glean sample application to (1) dump pings to logcat, (2) tag the ping 
+In the above:
+
+- `[package.name]` is the product's package name as exported in the manifest
```
Dexterp37: I think this is called "application id" in the [Android docs](https://developer.android.com/studio/build/application-id), so we can rename this to `[application.id]`.  By doing so, we can simply state that the full application id is required, including any suffix for the built flavour specified in the gradle build script.

mdboom: That makes sense.  I still like the idea of including what it is for the sample app here, though.

PR #2618, opened 2019-04-03
Glean: Limit timing distribution's range property to exactly 2 elements

travis79:
This is a fix for [1538697](https://bugzilla.mozilla.org/show_bug.cgi?id=1538697)

### Pull Request checklist


travis79:
> Looks good to me.
> 
> It doesn't look like there's anything testing the error cases (range.length() != 2, range is missing, etc.) That's fairly paranoid since this is data we generate ourselves, though.

Good point!  @Dexterp37 wouldn't let me get away without testing every nook and cranny so let me add some...  there's already a test that checks the deserialization that I can add a test case to.

travis79:
I added to the test to cover all the branches I could from the `fromJsonString()` function

PR #2619, opened 2019-04-03
Closes #2392 - Add ability for progress bar to be displayed on top of…

### LINKED ISSUE #2392: Add ability for progress bar to be laid out at top of toolbar
In Fenix some users (https://github.com/mozilla-mobile/fenix/issues/1049) are complaining that with a bottom toolbar and some phones that do not have the Android nav bar, it is difficult to see the progress bar at the very bottom of the display. Adding a flag for the progress bar to be displayed at the top of the toolbar may be helpful.

ekager:
… toolbar

Hi @pocmo this has UX now :) https://github.com/mozilla-mobile/fenix/issues/1049#issuecomment-479178216
I tested it in Fenix and I don't think we need a drawable attr here, I am able to override the progress bar style and set the drawable, so just the ability for the top layout for now. 

### Pull Request checklist

### DIFF
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
@@ -27,6 +27,7 @@ import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
 import mozilla.components.browser.menu.BrowserMenuBuilder
 import mozilla.components.browser.toolbar.display.DisplayToolbar
+import mozilla.components.browser.toolbar.display.DisplayToolbar.Companion.BOTTOM_PROGRESS_BAR
 import mozilla.components.browser.toolbar.edit.EditToolbar
 import mozilla.components.concept.toolbar.AutocompleteDelegate
 import mozilla.components.concept.toolbar.AutocompleteResult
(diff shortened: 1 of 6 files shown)


pocmo:
> Hi @pocmo this has UX now :) mozilla-mobile/fenix#1049 (comment)

Nice :D

components/browser/toolbar/src/main/res/values/attrs_browser_toolbar.xml
```diff
@@ -15,5 +15,6 @@
     <attr name="browserToolbarSuggestionForegroundColor" format="color" />
     <attr name="browserToolbarClearColor" format="color"/>
     <attr name="browserToolbarFadingEdgeSize" format="dimension" />
+    <attr name="topProgressBar" format="boolean" />
```
pocmo: I wonder if we could make this more flexible by not using a boolean but a "gravity" flag here, like in Android layouts.

- https://developer.android.com/reference/android/view/Gravity
- https://stackoverflow.com/questions/9531177/correct-format-for-androidgravity-in-attrs-xml

So that I could set this to something like: `browserToolbarProgressBarGravity=TOP/BOTTOM`

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
```diff
@@ -305,6 +314,10 @@ class BrowserToolbar @JvmOverloads constructor(
     init {
         context.obtainStyledAttributes(attrs, R.styleable.BrowserToolbar, defStyleAttr, 0).run {
             attrs?.let {
+                progressBarGravity = getInt(
+                    R.styleable.BrowserToolbar_browserToolbarProgressBarGravity,
+                    0
```
pocmo: nit: Let's use the constant (`BOTTOM_PROGRESS_BAR`) here instead of `0`.

Issue #2620, opened 2019-04-03
Site security icon color is not set immediately


sblatz:
When we set the `securityIconColor` on the `BrowserToolbar`, the color resource for the `securityIcon` in `DisplayToolbar` is not changed until `setSiteSecurity` is called. We should change this so that color setting is applied immediately.

PR #2625, opened 2019-04-03
Closes #2615: Adds titleView to DisplayToolbar.

### LINKED ISSUE #2615: Add a title view as optional text on the DisplayToolbar
In Fenix we will want the option to display the title of a website in custom tabs. Because of this, we'll need to create a `titleView` that is optionally shown in the `DisplayToolbar`

See [here](https://github.com/mozilla-mobile/fenix/issues/203#issuecomment-463426620) for more info.

sblatz:
### Pull Request checklist

### DIFF
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
@@ -193,6 +193,15 @@ class BrowserToolbar @JvmOverloads constructor(
         }
 
     /**
+     * Sets the colour of the text for title displayed in the toolbar.
+     */
+    var titleColor: Int
+        get() = displayToolbar.urlView.currentTextColor
+        set(value) {
+            displayToolbar.titleView.setTextColor(value)
+        }
+
+    /**
      * Sets the colour of the text for the URL/search term displayed in the toolbar.
      */
     var textColor: Int
@@ -200,6 +209,15 @@ class BrowserToolbar @JvmOverloads constructor(
         set(value) {
             displayToolbar.urlView.setTextColor(value)
             editToolbar.urlView.setTextColor(value)
+        }
+
+    /**
+     * Sets the size of the text for the title displayed in the toolbar.
+     */
+    var titleTextSize: Float
+        get() = displayToolbar.titleView.textSize
+        set(value) {
+            displayToolbar.titleView.textSize = value
         }
 
     /**
@@ -295,6 +313,13 @@ class BrowserToolbar @JvmOverloads constructor(
     private var state: State = State.DISPLAY
     private var searchTerms: String = ""
     private var urlCommitListener: ((String) -> Boolean)? = null
+
+    override var title: String = ""
+        set(value) {
+            displayToolbar.updateTitle(value)
+
+            field = value
+        }
 
     override var url: CharSequence = ""
         set(value) {
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
@@ -10,6 +10,7 @@ import android.support.v4.content.ContextCompat
 import android.support.v7.widget.AppCompatImageButton
 import android.support.v7.widget.AppCompatImageView
 import android.support.v7.widget.AppCompatTextView
+import android.text.TextUtils
 import android.util.TypedValue
 import android.view.Gravity
 import android.view.View
@@ -66,7 +67,7 @@ import mozilla.components.ui.icons.R.drawable.mozac_ic_lock
  *
  */
 @SuppressLint("ViewConstructor") // This view is only instantiated in code
-@Suppress("LargeClass")
+@Suppress("LargeClass", "TooManyFunctions")
 internal class DisplayToolbar(
     context: Context,
     val toolbar: BrowserToolbar
@@ -87,6 +88,16 @@ internal class DisplayToolbar(
         // with a value or null text behind the icon can be selectable.
         // https://github.com/mozilla-mobile/reference-browser/issues/448
         setOnClickListener(null)
+    }
+
+    internal val titleView = AppCompatTextView(context).apply {
+        id = R.id.mozac_browser_toolbar_title_view
+        gravity = Gravity.CENTER_VERTICAL
+        textSize = URL_TEXT_SIZE
+        visibility = View.GONE
+        ellipsize = TextUtils.TruncateAt.END
+
+        setSingleLine(true)
     }
 
     internal val urlView = AppCompatTextView(context).apply {
@@ -195,9 +206,18 @@ internal class DisplayToolbar(
 
     init {
         addView(siteSecurityIconView)
+        addView(titleView)
         addView(urlView)
         addView(menuView)
         addView(progressView)
+    }
+
+    /**
+     * Updates the title to be displayed.
+     */
+    fun updateTitle(title: String) {
+        titleView.text = title
+        titleView.visibility = if (title.isEmpty()) View.GONE else View.VISIBLE
     }
 
     /**
@@ -306,7 +326,8 @@ internal class DisplayToolbar(
         val height = MeasureSpec.getSize(heightMeasureSpec)
 
         val fixedHeightSpec = MeasureSpec.makeMeasureSpec(height, MeasureSpec.EXACTLY)
-
+        val halfFixedHeightSpec = MeasureSpec
+                .makeMeasureSpec(height / MEASURED_HEIGHT_DENOMINATOR, MeasureSpec.EXACTLY)
         setMeasuredDimension(width, height)
 
         // The icon and menu fill the whole height and have a square shape
@@ -325,7 +346,15 @@ internal class DisplayToolbar(
         val urlWidth = (width - iconSize - browserActionsWidth - pageActionsWidth -
             menuWidth - navigationActionsWidth - 2 * urlBoxMargin)
         val urlWidthSpec = MeasureSpec.makeMeasureSpec(urlWidth, MeasureSpec.EXACTLY)
-        urlView.measure(urlWidthSpec, fixedHeightSpec)
+
+        if (titleView.isVisible()) {
+            // With a title view, the url and title split the rest of the space vertically
+            titleView.measure(urlWidthSpec, halfFixedHeightSpec)
+            urlView.measure(urlWidthSpec, halfFixedHeightSpec)
+        } else {
+            // With no title view, the url view takes up the rest of the space
+            urlView.measure(urlWidthSpec, fixedHeightSpec)
+        }
 
         val progressHeightSpec = MeasureSpec.makeMeasureSpec(resources.pxToDp(PROGRESS_BAR_HEIGHT_DP),
                 MeasureSpec.EXACTLY)
@@ -427,10 +456,32 @@ internal class DisplayToolbar(
         //   | navigation  | icon | url       [ page    ] | browser  | menu |
         //   |   actions   |      |           [ actions ] | actions  |      |
         //   +-------------+------+-----------------------+----------+------+
-
         val iconWidth = if (siteSecurityIconView.isVisible()) siteSecurityIconView.measuredWidth else 0
         val urlLeft = navigationActionsWidth + iconWidth + urlBoxMargin
-        urlView.layout(urlLeft, 0, urlLeft + urlView.measuredWidth, measuredHeight)
+
+        // If the titleView is visible, it will appear above the URL:
+        //   +-------------+------+-----------------------+----------+------+
+        //   | navigation  | icon |  title       [ page    ] | browser  | menu |
+        //   |   actions   |      |  url         [ actions ] | actions  |      |
+        //   +-------------+------+-----------------------+----------+------+
+        if (titleView.isVisible()) {
+            val totalTextHeights = urlView.measuredHeight + titleView.measuredHeight
+            val totalAvailablePadding = height - totalTextHeights
+            val padding = totalAvailablePadding / MEASURED_HEIGHT_DENOMINATOR
+
+            titleView.layout(
+                    urlLeft,
+                    padding,
+                    urlLeft + titleView.measuredWidth,
+                    padding + titleView.measuredHeight)
+            urlView.layout(
+                    urlLeft,
+                    padding + titleView.measuredHeight,
+                    urlLeft + urlView.measuredWidth,
+                    padding + titleView.measuredHeight + urlView.measuredHeight)
+        } else {
+            urlView.layout(urlLeft, 0, urlLeft + urlView.measuredWidth, measuredHeight)
+        }
 
         // The progress bar by default is going to be drawn at the bottom of the toolbar, top if defined:
         progressView.layout(
@@ -449,6 +500,7 @@ internal class DisplayToolbar(
     }
 
     companion object {
+        internal const val MEASURED_HEIGHT_DENOMINATOR = 2
         internal const val URL_FADING_EDGE_SIZE_DP = 24
 
         const val BOTTOM_PROGRESS_BAR = 0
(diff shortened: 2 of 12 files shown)


components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
```diff
@@ -23,6 +23,7 @@ import mozilla.components.concept.toolbar.Toolbar
 import mozilla.components.support.base.feature.BackHandler
 import mozilla.components.support.base.feature.LifecycleAwareFeature
 import mozilla.components.support.ktx.android.content.share
+import java.net.URL
```
jonalmeida: `android.net.Uri` or the extension function for `String.toUri` is probably better.

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
```diff
@@ -10,6 +10,7 @@ import mozilla.components.browser.session.Session
 import mozilla.components.browser.session.SessionManager
 import mozilla.components.concept.toolbar.Toolbar
 import mozilla.components.feature.toolbar.internal.URLRenderer
+import java.net.URL
```
jonalmeida: `android.net.Uri` or the extension function for `String.toUri` here as well.

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
```diff
@@ -64,16 +65,39 @@ class ToolbarPresenter(
...
     }
 
     override fun onUrlChanged(session: Session, url: String) {
-        renderer.post(url)
+        renderer.post(getURLToPost(session, url))
         toolbar.setSearchTerms(session.searchTerms)
     }
 
+    private fun getURLToPost(session: Session, url: String): String {
+        return if (session.isCustomTabSession()) {
```
jonalmeida: I'm not a fan of adding custom tab checks around the codebase since it increases our complexity in multiple components at once. How about always passing the url in, and when `displayTitle` is enabled, parse the url for the host in the CustomTabsToolbarFeature itself which does the global "isCustomTab" check in one place.

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
```diff
@@ -101,6 +101,12 @@ class BrowserToolbar @JvmOverloads constructor(
             }
         }
 
+    override var displayTitle: Boolean
+        get() = displayToolbar.titleView.isVisible()
+        set(value) {
+            displayToolbar.titleView.visibility = if (value) View.VISIBLE else View.GONE
+        }
```
pocmo: Does this need to be a separate flag or can this be controlled by whether a title is set or not? Not every page has a title so I assume if there's no title then we do not want to display this view?

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
```diff
@@ -89,6 +89,15 @@ internal class DisplayToolbar(
         setOnClickListener(null)
     }
 
+    internal val titleView = AppCompatTextView(context).apply {
+        id = R.id.mozac_browser_toolbar_title_view
+        gravity = Gravity.CENTER_VERTICAL
+        textSize = URL_TEXT_SIZE
+        visibility = View.GONE
+
+        setSingleLine(true)
```
pocmo: Do we want to define here what happens if the text is longer than the view's width?
-> https://developer.android.com/reference/android/widget/TextView.html#setEllipsize(android.text.TextUtils.TruncateAt)

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
```diff
@@ -65,6 +66,12 @@ class CustomTabsToolbarFeature(
             if (config.showShareMenuItem) addShareButton(session)
             // Add menu items
             if (config.menuItems.isNotEmpty()) addMenuItems(config.menuItems)
+
+            toolbar.url = URL(session.url).host
```
pocmo: Not every URL has a host. Not every URL is http(s). :)

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
```diff
@@ -64,16 +65,39 @@ class ToolbarPresenter(
...
+        val urlToPost = if (session != null) {
+            getURLToPost(session, session.url)
+        } else {
+            ""
+        }
+
+        // Set the title to be the url as we don't have a title until the page loads
+        session?.let { onTitleChanged(it, it.url) }
+
+        renderer.post(urlToPost)
```
pocmo: This logic shouldn't live here. The consuming app can already provide a `UrlRenderConfiguration` to define how the URL should be rendered. If we want a "host" only mode (Btw. is it really the host? Or the "registrable domain"?).

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
```diff
@@ -176,7 +185,13 @@ internal class DisplayToolbar(
 
     private val defaultColor = ContextCompat.getColor(context, R.color.photonWhite)
 
+    private var currentSiteSecurity = SiteSecurity.INSECURE
+
     internal var securityIconColor = Pair(defaultColor, defaultColor)
+        set(value) {
+            field = value
+            setSiteSecurity(currentSiteSecurity)
+        }
```
pocmo: How is this change related? What's the background?

We try really hard to not have any state here and keep that in the `Session` only.

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
```diff
@@ -65,6 +66,12 @@ class CustomTabsToolbarFeature(
...
+            toolbar.displayTitle = true
```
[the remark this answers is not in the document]
sblatz: Our UX asks for this to be [displayed universally](https://github.com/mozilla-mobile/fenix/issues/203#issuecomment-463426620) and doesn't make a mention of letting this be customizable. Are we sure we want to be setting it based on this flag?

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
sblatz: Good point, I assumed every URL had a title.

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
sblatz: I initially filed [a separate issue](https://github.com/mozilla-mobile/android-components/issues/2620) for that change, but wasn't sure if that was necessary. I'm happy to handle it in a separate PR :)

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
sblatz: Our [mocks](https://github.com/mozilla-mobile/fenix/issues/203#issuecomment-463426620) used the word "host" and that seemed to give me the right string (aka "google.com"). How could I get the "registrable domain"?

pocmo: See `UrlRenderer` this is already used when coloring the domain. That's also a good example - if you look at desktop Firefox, the highlighted part of the URL is the "registrable domain".

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
sblatz: Updated this, though I wonder why the `url` doesn't also have this set?

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
sblatz: Removed both of these usages :)

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
sblatz: If I go with [Sebastian's recommendation here](https://github.com/mozilla-mobile/android-components/pull/2625#discussion_r272075518) how should I handle this? I agree that constantly checking for `isCustomTab` could be annoying.

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
pocmo: The url sets a fading edge in the `init {}` block.

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
jonalmeida: This can still be done within the `CustomTabsToolbarFeature` using a session observer to check when the title is received in that session.

The [`WindowFeature`](https://github.com/mozilla-mobile/android-components/blob/master/components/feature/session/src/main/java/mozilla/components/feature/session/WindowFeature.kt#L19-L34) has a nice example of how to make an internal session observer in a feature class. However, since you want to focus on just one session, you wouldn't want to inherit from `SelectionAwareSessionObserver`, but provide the `sessionId` of the CT to it and make changes.

There might also be the case where the title is already received so you would have to check the `session.title` before deciding to not show the title entirely.

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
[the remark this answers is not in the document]
jonalmeida: I've reached out to Bram offline to see if this was an oversight or if we want to intentionally not follow the Custom Tab spec for showing titles. I'll follow up here.

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
sblatz: It looks like `CustomToolbarFeature` uses the same `oolbarPresenter` as `ToolbarFeature`. Because of this, I'm struggling to use a config that displays only the registrable domain. When the config is set to `showOnlyRegistrableDomain` for example, it will show in both custom and normal tabs, which is not the behavior we want.

Any recommendations on a clean way to solve this?

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
```diff
@@ -411,7 +437,21 @@ internal class DisplayToolbar(
 
         val iconWidth = if (siteSecurityIconView.isVisible()) siteSecurityIconView.measuredWidth else 0
         val urlLeft = navigationActionsWidth + iconWidth + urlBoxMargin
-        urlView.layout(urlLeft, 0, urlLeft + urlView.measuredWidth, measuredHeight)
+
```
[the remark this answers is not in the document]
sblatz: Added a few more regarding layout!

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
sblatz: I think my plan is to do something like mentioned in my discussion with Sebastian [below](https://github.com/mozilla-mobile/android-components/pull/2625#discussion_r272077771). Does that sound good to you? Just not sure what solution will work with the config being the same between the two bars.

sblatz:
@jonalmeida, @pocmo: I believe I addressed all concerns raised. Please let me know if there's any other issues you see!

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
```diff
@@ -310,7 +330,14 @@ internal class DisplayToolbar(
         val urlWidth = (width - iconSize - browserActionsWidth - pageActionsWidth -
             menuWidth - navigationActionsWidth - 2 * urlBoxMargin)
         val urlWidthSpec = MeasureSpec.makeMeasureSpec(urlWidth, MeasureSpec.EXACTLY)
-        urlView.measure(urlWidthSpec, fixedHeightSpec)
+
+        titleView.measure(urlWidthSpec, fixedHeightSpec / MEASURED_HEIGHT_DENOMINATOR)
+
+        if (titleView.isVisible()) {
```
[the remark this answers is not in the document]
sblatz: 👍 moved the titleView measured inside the if check

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
```diff
@@ -64,7 +64,12 @@ class ToolbarPresenter(
         val session = sessionId?.let { sessionManager.findSessionById(sessionId) }
             ?: sessionManager.selectedSession
 
-        renderer.post(session?.url ?: "")
+        // Set the title to be the url as we don't have a title until the page loads
+        session?.let {
+            onTitleChanged(it, it.url)
```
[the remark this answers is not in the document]
sblatz: How should I handle a session **not** having a title until page load then? On desktop, we set the title to be the url until our observer gets notified. Is doing so not acceptable here?

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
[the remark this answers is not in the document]
jonalmeida: We can land it this way for now and loop back when we hear back from @brampitoyo

sblatz:
I have fixed all the issues raised. Unfortunately, now my tests are failing. For some reason whenever I call `updateTitle` the measuredHeight of both `urlTextViewRect` and `titleViewRect` are set to 0 (though their widths are correct.

@pocmo is there any chance you could take a look at my most recent commit to see if there is something you can figure out? I have tested this in sample browser and Fenix and the feature works as expected, so I think it's just a discrepancy in the test.

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
```diff
@@ -317,7 +337,15 @@ internal class DisplayToolbar(
         val urlWidth = (width - iconSize - browserActionsWidth - pageActionsWidth -
             menuWidth - navigationActionsWidth - 2 * urlBoxMargin)
         val urlWidthSpec = MeasureSpec.makeMeasureSpec(urlWidth, MeasureSpec.EXACTLY)
-        urlView.measure(urlWidthSpec, fixedHeightSpec)
+
+        if (titleView.isVisible()) {
+            // With a title view, the url and title split the rest of the space vertically
+            titleView.measure(urlWidthSpec, fixedHeightSpec / MEASURED_HEIGHT_DENOMINATOR)
+            urlView.measure(urlWidthSpec, fixedHeightSpec / MEASURED_HEIGHT_DENOMINATOR)
```
pocmo: I haven't fully debugged it yet. But this is the part that looks wrong: `fixedHeightSpec` is a MeasureSpec (packed in an int) and not a height you can divide:
https://developer.android.com/reference/android/view/View.MeasureSpec

Instead you need to create a new spec (see where `fixedHeightSpec` gets created):
```Kotlin
 MeasureSpec.makeMeasureSpec(height / MEASURED_HEIGHT_DENOMINATOR /** Divide here! **/, MeasureSpec.EXACTLY)
```

PR #2627, opened 2019-04-03
1538745: Make TimingDistributionData.values use Int as key

travis79:
This reduces the overhead of having to do string interpolation from Int to String every time `accumulate` is called and only has to convert this to a map with String keys when serializing.

### Pull Request checklist


components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
```diff
@@ -248,13 +248,20 @@ data class TimingDistributionData(
...
+        }
+
         return JSONObject(mapOf(
             "category" to category,
             "name" to name,
             "bucketCount" to bucketCount,
             "range" to JSONArray(range),
             "histogramType" to histogramType.ordinal,
-            "values" to values,
+            "values" to jsonValues,
```
Dexterp37: nit: I think this could become `"values" to values.mapKeys {"${it.key}"}`

PR #2629, opened 2019-04-04
Restrict the exposed `Configuration` fields

Dexterp37:
We don't expect consuming applications to change some of the options we expose. A few of them are only meant to be changed by the `GleanDebugActivity` (internally) and some of them should only be changed by tests. This PR makes the primary constructor of the `Configuration` object internal and introduces a secondary, public constructor that only exposes things that can be changed.

@pocmo do you have any suggestion about alternative approaches to this? For example, in the mid-future we might want to be able to expose `serverEndpoint` again for pointing instrumented/integration tests to a local server instance, while still not allowing other apps to change it.

### Pull Request checklist


Dexterp37:
>  Do we instead want to use a regular class (non-data class) instead and have those fields be truly `internal`

That would require a bit more mangling through the rest of glean because we rely on some features of the data classes. It's a good idea, and that could be helpful when we might need to do that to point glean to local servers for integration tests. @mdboom would you be ok with me deferring this until is needed in that bug?

mdboom:
> @mdboom would you be ok with me deferring this until is needed in that bug?

Sure, I'm probably just a little touchy about unintentional uses of the "public" API today.

PR #2652, opened 2019-04-05
1536792:  Reinstate disabled MetricsPingSchedulerTests.

travis79:
In order to accomplish this, a window in `Glean.initialize()` where metrics could be recorded before init was complete needed to be closed.  In order to do this, a new `initializing` flag/state was added to allow for uploading of pings while still preventing recording of metrics during initialization.

### Pull Request checklist

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
@@ -46,8 +46,19 @@ open class GleanInternalAPI internal constructor () {
 
     private val gleanLifecycleObserver by lazy { GleanLifecycleObserver() }
 
-    // `internal` so this can be modified for testing
-    internal var initialized = false
+    internal enum class GleanInitStates {
+        // This state represents glean as totally uninitialized
+        UNINITIALIZED,
+        // This state represents glean initializing, where pings can be triggered by internal code
+        // but external callers of the specific metrics API are not allowed to record
+        INITIALIZING,
+        // This state represents glean in normal run-mode state
+        INITIALIZED
+    }
+
+    // Represents glean's init state
+    internal var initState = GleanInitStates.UNINITIALIZED
+
     private var uploadEnabled = true
 
     // The application id detected by glean to be used as part of the submission
@@ -82,7 +93,7 @@ open class GleanInternalAPI internal constructor () {
         applicationContext: Context,
         configuration: Configuration = Configuration()
     ) {
-        if (isInitialized()) {
+        if (isInitializing() || isInitialized()) {
             logger.error("Glean should not be initialized multiple times")
             return
         }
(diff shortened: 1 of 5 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -48,6 +48,10 @@ open class GleanInternalAPI internal constructor () {
 
     // `internal` so this can be modified for testing
     internal var initialized = false
+    // This is used to control metrics recording while still allowing sending of pings during glean
+    // initialization
+    internal var initializing = false
```
mdboom: Should we collapse this into a single enumerated value?  There's potential for introducing a new bug at some point in the future where we're in an undefined state (both `initializing` and `initialized` are `true`).

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -97,8 +101,10 @@ open class GleanInternalAPI internal constructor () {
         // API. For this reason we're safe to set `initialized = true` right after it.
         initializeCoreMetrics(applicationContext)
 
-        // This must be set before anything that might trigger the sending of pings.
-        initialized = true
+        // This must be set before anything that might trigger the sending of pings.  We still don't
```
mdboom: Should we also guide against `initializing` already true at the top of this function?

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
travis79: `initializing` should only ever be used internal to the `initialize()` function (and tests) and there were only two states...  but I see the point behind this and made the change.  We may end up with more states in the future.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
travis79: Good point.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -47,7 +47,18 @@ open class GleanInternalAPI internal constructor () {
     private val gleanLifecycleObserver by lazy { GleanLifecycleObserver() }
 
     // `internal` so this can be modified for testing
-    internal var initialized = false
+    internal enum class GleanInitStates {
+        // This state represents glean as totally uninitialized
+        UNINITIALIZED,
+        // This state represents glean initializing, where we can send pings but not record metrics
```
Dexterp37: nit: "... initializing , where pings can be triggered by internal code but external callers of the specific metrics API are not allowed to record".

travis79:
2 x green's in a row...

travis79:
Unrelated failure in `samples-browser` on this CI run...

travis79:
2 x all green...

travis79:
3x

travis79:
Closing this to revisit it later... this has likely suffered from bit rot by now.

PR #2659, opened 2019-04-05
Glean: improve comment regarding bucketing to be more clear about overflow

travis79:
### Pull Request checklist


components/service/glean/src/test/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngineTest.kt
```diff
@@ -226,7 +226,10 @@ class TimingDistributionsStorageEngineTest {
 
     @Test
     fun `getBuckets() correctly populates the buckets property`() {
-        // Hand calculated values using current default range 0 - 60000 and bucket count of 100
+        // Hand calculated values using current default range 0 - 60000 and bucket count of 100.
```
Dexterp37: I think that, in addition to code comments, this should be documented in some user-facing document. I think either [here](https://github.com/mozilla-mobile/android-components/blob/master/components/service/glean/docs/metrics/adding-new-metrics.md#Timing-Distributions) (maybe in a separate subsection) or, better, in its own metric-specific details doc. @georgf , thoughts?

components/service/glean/src/test/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngineTest.kt
```diff
@@ -226,7 +226,10 @@ class TimingDistributionsStorageEngineTest {
...
+        // NOTE: The final bucket here of 60001 is a product of the bucketing algorithm and
```
Dexterp37: nit: I would focus more on the purpose of the bucket rather than how it originated. I would (please push back or phrase it better!) say that regardless of the bucket size/width, the last bucket holds samples beyond the max range.

components/service/glean/src/test/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngineTest.kt
travis79: Since we don't really expose the bucket sizes to the public API, is it still a good idea to document it in the API docs?  Right now the API just asks "give me a time to record".  While the test API functions do return an object that does have the bucket property accessible, one of the things that @georgf asked me to change in the docs for the timing distribution was the comparing the count of a particular bucket, instead pushing to demonstrate just checking the sum and the total count.

mdboom: I sort of feel like the contents of `TimingDistributionData` are sort of an implementation detail.  Ideally, I'd like to see the test functions grow to include methods that get specific things out of it so we could change the implementation (which roughly corresponds to the JSON payload) at some point in the future without breaking tests.

However, I'm not necessarily suggesting we do that, since I think it's a lot of work for very little gain, but I do look at it as an implementation detail, and we should at least limit the documented examples to things that are less likely to change in the future.

components/service/glean/src/test/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngineTest.kt
mdboom: :+1:

PR #2662, opened 2019-04-05
Fixes #2658 - Adds the ability to long click the urlView

### LINKED ISSUE #2658: Provide interface to add an OnLongCLickListener to DisplayToolbar
In Fenix we want to longpress to copy the current URL

boek:
### Pull Request checklist


components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
```diff
@@ -284,6 +291,20 @@ internal class DisplayToolbar(
         invalidateActions(navigationActions + pageActions + browserActions)
     }
 
+    /**
+     * Set a LongClickListner to the urlView of the toolbar.
+     */
+    private fun setOnUrlLongClickListener(handler: (() -> Unit)? = null) {
+        if (handler == null) {
+            urlView.setOnLongClickListener(null)
```
jonalmeida: Do we need to set the click listener to null? Can we not just do the else case always? 🤔 

Maybe we could even make the listener non-null as well.

boek: @jonalmeida We probably could. While writing this I tested and the longpress always has the haptic feedback, so I thought there might be a case someone wants to fully remove it.

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
```diff
@@ -163,6 +163,13 @@ class BrowserToolbar @JvmOverloads constructor(
         get() = displayToolbar.onUrlClicked
         set(value) { displayToolbar.onUrlClicked = value }
 
+    /**
+     * Sets a lambda that will be invoked whenever the URL in display mode was long clicked.
+     */
+    var onUrlLongClicked: (() -> Unit)?
+        get() = displayToolbar.onUrlLongClicked
+        set(value) { displayToolbar.onUrlLongClicked = value }
```
pocmo: I wonder whether this needs to be a field or could just be a setter like other listeners in here? It would avoid some of the indirection and you could added the OnLongClickListener on the view here directly.

boek: I originally made it a field to be consistent with the click listener, I'll rip it out :)

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
jonalmeida: Oh, interesting. I didn't realize the system hooks into that and does that when a listener is attached.

boek:
Ok I'll try that :)

PR #2670, opened 2019-04-08
FxA device & send tab integration

grigoryk:
Depends on https://github.com/mozilla/application-services/pull/676.

This still needs a light coat of polish, but should be good for a feedback round.

As a preparatory commit, this PR changes how account state is persisted. It's now callback-driven, ensuring we don't miss out on internal state mutations (such as would happen during device registration and events processing).

For the main show, the PR:
- describes various device-related functionality - registering, renaming, device events (such as Send Tab)
- adds a module-internal `PollingDeviceManager`, responsible for event polling and maintaining a cache of the current and all other devices
- FirefoxAccount has been expanded to provide relevant functionality, and it can now be observed for events targeting the current device (e.g. tabs received)
- FxaAccountManager was modified to include account lifecycle tie-ins with the device registration stuff

I expect consuming applications to interact with these changes as follows:
- to display information about current and/or other devices, query synchronous `currentDevice` and `otherDevices` methods, while also subscribing to changes via `DeviceConstellationObserver`
- to enable device event processing (i.e. to receive tabs), including periodic polling and fxa device registration, pass in an `EventsProcessor` instance to the FxaAccountManager. Account manager now also takes an initial device name.
- to rename current device, use appropriate methods exposed on FirefoxAccount
- to process events received and decrypted via Push, pass the plaintext payload to `processRawEvent` to an instance of FirefoxAccount obtained from the account manager. This will cause all of the currently registered events observers to be triggered.

Currently no extra work is performed if the application doesn't specify an EventsProcessor instance when initializing an FxaAccountManager. This means that it's possible to opt-out of FxA device registration; we may want to change that before shipping, I'm not sure. Currently there is no way to declare current device's capabilities, and so it's not possible to register a device while also opting out of being a Send Tab target. That'll need to change in the fxaclient library if we want to start splitting the two concepts.

I went back and forth on the complexity of this, and in the end tried to introduce minimal overhead on top of primitives exposed by the library.

### Pull Request checklist

### DIFF
--- a/components/concept/sync/src/main/java/mozilla/components/concept/sync/DeviceEvents.kt
+++ b/components/concept/sync/src/main/java/mozilla/components/concept/sync/DeviceEvents.kt
@@ -1,0 +1,31 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.concept.sync
+
+/**
+ * Allows monitoring events targeted at the current device.
+ */
+interface DeviceEventsObserver {
+    fun onEvents(events: List<DeviceEvent>)
+}
+
+/**
+ * Incoming device events, targeted at the current device.
+ */
+sealed class DeviceEvent {
+    class TabReceived(val from: Device?, val entries: List<TabData>) : DeviceEvent()
+}
+
+/**
+ * Outgoing device events, targeted at other devices.
+ */
+sealed class DeviceEventOutgoing {
+    class SendTab(val title: String, val url: String) : DeviceEventOutgoing()
+}
+
+data class TabData(
+    val title: String,
+    val url: String
+)
--- a/components/concept/sync/src/main/java/mozilla/components/concept/sync/Devices.kt
+++ b/components/concept/sync/src/main/java/mozilla/components/concept/sync/Devices.kt
@@ -1,0 +1,155 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+@file:SuppressWarnings("TooManyFunctions")
+package mozilla.components.concept.sync
+
+import androidx.lifecycle.LifecycleOwner
+import kotlinx.coroutines.Deferred
+import mozilla.components.support.base.observer.Observable
+
+/**
+ * Describes available interactions with the current device and other devices associated with an [OAuthAccount].
+ */
+interface DeviceConstellation : Observable<DeviceEventsObserver> {
+    /**
+     * Register current device in the associated [DeviceConstellation].
+     *
+     * @param name An initial name for the current device. This may be changed via [setDeviceNameAsync].
+     * @param type Type of the current device. This can't be changed.
+     * @param capabilities A list of capabilities that the current device claims to have.
+     * @return A [Deferred] that will be resolved once initialization is complete.
+     */
+    fun initDeviceAsync(
+        name: String,
+        type: DeviceType = DeviceType.MOBILE,
+        capabilities: List<DeviceCapability>
+    ): Deferred<Unit>
+
+    /**
+     * Ensure that all initialized [DeviceCapability], such as [DeviceCapability.SEND_TAB], are configured.
+     * This may involve backend service registration, or other work involving network/disc access.
+     * @return A [Deferred] that will be resolved once operation is complete.
+     */
+    fun ensureCapabilitiesAsync(): Deferred<Unit>
+
+    /**
+     * Current state of the constellation. May be missing if state was never queried.
+     * @return [ConstellationState] describes current and other known devices in the constellation.
+     */
+    fun state(): ConstellationState?
+
+    /**
+     * Allows monitoring state of the device constellation via [DeviceConstellationObserver].
+     * Use this to be notified of changes to the current device or other devices.
+     */
+    fun registerDeviceObserver(observer: DeviceConstellationObserver, owner: LifecycleOwner, autoPause: Boolean)
+
+    /**
+     * Get all devices in the constellation.
+     * @return A list of all devices in the constellation.
+     */
+    fun fetchAllDevicesAsync(): Deferred<List<Device>>
+
+    /**
+     * Set name of the current device.
+     * @param name New device name.
+     * @return A [Deferred] that will be resolved once operation is complete.
+     */
+    fun setDeviceNameAsync(name: String): Deferred<Unit>
+
+    /**
+     * Set a [DevicePushSubscription] for the current device.
+     * @param subscription A new [DevicePushSubscription].
+     * @return A [Deferred] that will be resolved once operation is complete.
+     */
+    fun setDevicePushSubscriptionAsync(subscription: DevicePushSubscription): Deferred<Unit>
+
+    /**
+     * Send an event to a specified device.
+     * @param targetDeviceId A device ID of the recipient.
+     * @param outgoingEvent An event to send.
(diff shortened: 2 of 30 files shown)


grigoryk:
Reference browser integration https://github.com/mozilla-mobile/reference-browser/pull/704

grigoryk:
This basically works and I'll stop tweaking it. It needs to land soon :-)
There are some library issues, but those aren't blocking and could be dealt with separately from this work.

@pocmo @csadilek could ya'll do a review pass on this?

grigoryk:
I'll call this out right away: I'm not a fan of this API. The "device constellation" thing feels forced. But, putting everything onto OAuthAccount feels wrong as well and doesn't give us a nice path for implementations that don't care about devices stuff... At this point, our current generic accounts abstraction is starting to have a bit of friction with the FxA account.

Underneath, at the library level, everything is under the accounts object. I feel that the few affordances I've added in shape of various observers make it a little nicer. Judge for yourself - the `sample-sync` app provides an example of how this is all put together.

grigoryk:
`samples-sync` app now lets you:
- view current device information
- view all other devices and their detailed information
- receive tabs (from desktop, other instances of the sample app, etc) - press `Refresh Device` after sending it
- send a hard-coded sample app to all devices that support Send Tab

components/concept/sync/src/main/java/mozilla/components/concept/sync/DeviceEvents.kt
```diff
@@ -0,0 +1,31 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.concept.sync
+
+/**
+ * Allows monitoring events targeted at the current device.
+ */
+interface DeviceEventsObserver {
+    fun onTabsReceived(tabs: List<TabData>, from: Device?)
```
csadilek: Shouldn't the `DeviceEventsObserver` observe `DeviceEvent`s. Seemed strange to have `onTabsReceveived` and also `DeviceEnvent.TabsReceived`.

components/concept/sync/src/main/java/mozilla/components/concept/sync/Devices.kt
```diff
@@ -0,0 +1,147 @@
...
+import androidx.lifecycle.LifecycleOwner
+import kotlinx.coroutines.Deferred
+import mozilla.components.support.base.observer.Observable
+
+/**
+ * Describes available interactions with the current device and other devices associated with an [OAuthAccount].
+ */
+interface DeviceConstellation : OAuthAccount, Observable<DeviceEventsObserver> {
+    /**
+     * Make sure there is an available a device record in the [DeviceConstellation] for the current device.
```
csadilek: nit: Makes sure there is a device record available ....

components/concept/sync/src/main/java/mozilla/components/concept/sync/Devices.kt
```diff
@@ -0,0 +1,147 @@
...
+     * @param name An initial name for the current device. This may be changed via [setDeviceNameAsync].
+     * @param type Type of the current device. This can't be changed.
+     * @param capabilities A list of capabilities that the current device claims to have.
```
csadilek: nit: `@return`

components/concept/sync/src/main/java/mozilla/components/concept/sync/Devices.kt
```diff
@@ -0,0 +1,147 @@
...
+     */
+    fun initDeviceAsync(
+        name: String,
+        type: DeviceType = DeviceType.MOBILE,
+        capabilities: List<DeviceCapability>
+    ): Deferred<Unit>
+
+    /**
+     * Ensure that all initialized capabilities are configured.
```
csadilek: nit: would be nice to elaborate a bit here. What are capabilities? Probably just link to `[DeviceCapability]`

components/concept/sync/src/main/java/mozilla/components/concept/sync/Devices.kt
```diff
@@ -0,0 +1,147 @@
...
+package mozilla.components.concept.sync
+
+import androidx.lifecycle.LifecycleOwner
+import kotlinx.coroutines.Deferred
+import mozilla.components.support.base.observer.Observable
+
+/**
+ * Describes available interactions with the current device and other devices associated with an [OAuthAccount].
+ */
+interface DeviceConstellation : OAuthAccount, Observable<DeviceEventsObserver> {
```
csadilek: This relationship feels wrong to me. Why is a device constellation a `OAuthAccount`? I think you mentioned that in your comment also.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FirefoxAccount.kt
```diff
@@ -4,28 +4,45 @@
...
 typealias PersistCallback = mozilla.appservices.fxaclient.FirefoxAccount.PersistCallback
 
 /**
  * FirefoxAccount represents the authentication state of a client.
  */
 @Suppress("TooManyFunctions")
-class FirefoxAccount internal constructor(private val inner: InternalFxAcct) : OAuthAccount {
+class FirefoxAccount internal constructor(
+    private val inner: InternalFxAcct
+) : DeviceConstellation, Observable<DeviceEventsObserver> by ObserverRegistry() {
```
csadilek: I think here it's even clearer that there's something wrong with the `is a` relationship. Why is a `FirefoxAccount` a `DeviceConstellation`? I would definitely switch that to `has a` and use composition.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
```diff
@@ -101,6 +66,8 @@ private interface OAuthObserver {
     fun onError(error: FxaException)
 }
```
csadilek: Can we add some docs here why this class exists relative to Device.

components/concept/sync/src/main/java/mozilla/components/concept/sync/DeviceEvents.kt
grigoryk: Ha, yeah. That is pretty weird. Changing.

grigoryk: So, the intention here was to have two layers of event processing; an internal layer - for "system" events, like "please wipe your history storage", and for events that are handled downstream, such as "you got a new tab".

But, I didn't quite realize that vision anyway, and it's a future requirement. So let's keep this simple for now.

grigoryk: Better example would be forcing logout remotely ("pls logout!"), or informing of new devices in the constellation. We already have some support for the latter.

components/concept/sync/src/main/java/mozilla/components/concept/sync/Devices.kt
```diff
@@ -0,0 +1,139 @@
...
+        type: DeviceType = DeviceType.MOBILE,
+        capabilities: List<DeviceCapability>
+    ): Deferred<Unit>
+
+    /**
+     * Ensure that all initialized capabilities are configured.
+     */
+    fun ensureCapabilitiesAsync(): Deferred<Unit>
+
+    fun state(): ConstellationState?
```
csadilek: nit: kdocs :)

grigoryk:
What's not covered in tests is the `FxaDeviceConstellation`, but it just passes through calls to `FirefoxAccount`, for the most part. Anything with actual logic is covered.

Once https://github.com/mozilla-mobile/android-components/pull/2943 establishes the pattern for native stack testing, and once we'll be able to configure rust networking stack in tests, we'll be able to test more of this in a meaningful way.

PR #2676, opened 2019-04-08
Closes #2674: Adds facts to custom tabs

### LINKED ISSUE #2674: Add facts to custom tabs

sblatz:
### Pull Request checklist

**Note: I did not add a test for emitting close fact or action fact in `CustomTabsToolbarFeature` because those should be covered by `clicking on navigation action will execute listener of the action` inside `DisplayToolbarTest`**

### DIFF
--- a/components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenuBuilder.kt
+++ b/components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenuBuilder.kt
@@ -10,13 +10,14 @@ import android.content.Context
  * Helper class for building browser menus.
  *
  * @param items List of BrowserMenuItem objects to compose the menu from.
+ * @param extras Map of extra values that are added to emitted facts
  */
 class BrowserMenuBuilder(
-    val items: List<BrowserMenuItem>
+    val items: List<BrowserMenuItem>,
+    val extras: Map<String, Any> = emptyMap()
 ) {
     fun build(context: Context): BrowserMenu {
         val adapter = BrowserMenuAdapter(context, items)
-
         return BrowserMenu(adapter)
     }
 }
--- a/components/browser/toolbar/README.md
+++ b/components/browser/toolbar/README.md
@@ -34,8 +34,10 @@ This component emits the following [Facts](../../support/base/README.md#Facts):
 
 | Action | Item    | Extras         | Description                        |
 |--------|---------|----------------|------------------------------------|
-| CLICK  | menu    |                | The user opened the overflow menu. |
+| CLICK  | menu    | `menuExtras`   | The user opened the overflow menu. |
 | COMMIT | toolbar | `commitExtras` | The user has edited the URL.       |
+
+`menuExtras` are additional extras set on the `BrowserMenuBuilder` passed to the `BrowserToolbar` (see [browser-menu](../menu/README.md)).
 
 #### `commitExtras`
 
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
@@ -137,7 +137,7 @@ internal class DisplayToolbar(
                 anchor = this,
                 orientation = BrowserMenu.determineMenuOrientation(toolbar))
 
-            emitOpenMenuFact()
+            emitOpenMenuFact(menuBuilder?.extras)
         }
     }
 
--- a/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
+++ b/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
@@ -100,7 +100,10 @@ class CustomTabsToolbarFeature(
         }.also {
             val button = Toolbar.ActionButton(
                 it, context.getString(R.string.mozac_feature_customtabs_exit_button)
-            ) { closeListener.invoke() }
+            ) {
+                emitCloseFact()
+                closeListener.invoke()
+            }
             toolbar.addNavigationAction(button)
         }
     }
@@ -111,7 +114,10 @@ class CustomTabsToolbarFeature(
             val button = Toolbar.ActionButton(
                 BitmapDrawable(context.resources, config.icon),
                 config.description
-            ) { config.pendingIntent.send() }
+            ) {
+                emitActionButtonFact()
+                config.pendingIntent.send()
+            }
 
             toolbar.addBrowserAction(button)
         }
@@ -137,8 +143,10 @@ class CustomTabsToolbarFeature(
             SimpleBrowserMenuItem(it.name) { it.pendingIntent.send() }
         }.also { items ->
             val combinedItems = menuBuilder?.let { builder -> builder.items + items } ?: items
-
-            toolbar.setMenuBuilder(BrowserMenuBuilder(combinedItems))
+            val combinedExtras = menuBuilder?.let {
+                    builder -> builder.extras + Pair("customTab", true)
+            }
+            toolbar.setMenuBuilder(BrowserMenuBuilder(combinedItems, combinedExtras ?: emptyMap()))
         }
     }
 
--- a/components/browser/menu/README.md
+++ b/components/browser/menu/README.md
@@ -75,7 +75,6 @@ To customize the menu you could use separate properties 1 or full access to the 
 1) If you just want to change a specify property, just add one these dimen items to your ``dimes.xml`` file.
 
 ```xml
-
     <!--Menu Item -->
        <!--Change the text_size for ALL menu items NOT only for the BrowserMenuImageText -->
         <dimen name="mozac_browser_menu_item_text_size" tools:ignore="UnusedResources">16sp</dimen>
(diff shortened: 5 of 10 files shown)


components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenuBuilder.kt
```diff
@@ -12,7 +12,8 @@ import android.content.Context
  * @param items List of BrowserMenuItem objects to compose the menu from.
  */
 class BrowserMenuBuilder(
-    val items: List<BrowserMenuItem>
+    val items: List<BrowserMenuItem>,
+    val isCustomTabMenu: Boolean? = false
```
pocmo: I think this is too specific. If apps use menus in different situations then we can't add booleans here for all possible menus that an app may use. To make this more generic we could allow passing in an additional map of values that will be added as extras. Would that work?

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
```diff
@@ -92,7 +92,16 @@ class CustomTabsToolbarFeature(
...
             val button = Toolbar.ActionButton(
                 it, context.getString(R.string.mozac_feature_customtabs_exit_button)
-            ) { closeListener.invoke() }
+            ) {
+                val selectedId = sessionManager.selectedSession?.id
+                for (i in 0..sessionManager.sessions.size) {
+                    if (sessionManager.sessions[i].id == selectedId) {
+                        emitCloseFact(i, sessionManager.sessions.size)
+                    }
+                }
```
pocmo: FYI: There's also a `findSessionById()` method.

I wonder what this does though. It emits a close fact here, but it doesn't actually close this session here. Is this the right place?

sblatz: This should get called right before `closeListener.invoke()` is called right beneath it, no? We care that the "close" button is pressed, not that the session is closed.

Also, I need to know the **index** of the session, which is why I chose to use a for loop rather than `findSessionById()`.

components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenuBuilder.kt
sblatz: Agreed, updated to consider this :)

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
pocmo: I'm not really sure what you are trying to achieve here though: This is a custom tab that we are closing: It is not `selectedSession` and its also not part of `sessions`. Why do those values matter here? I can understand that they are interesting in the context of an actual tab getting closed. :)

components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenuBuilder.kt
pocmo: Like that! :) Can you add a KDoc for the parameter explaining what those extras are used for? :)

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
sblatz: I raised these concerns [in the original ticket](https://github.com/mozilla-mobile/fenix/issues/977)!

sblatz: @pocmo I updated the PR to remove this, as product decided that detail wasn't necessary. Thanks for bringing this up 😄

sblatz:

1) **What questions will you answer with this data?**
- How often quick bar is being opened/closed
- Time between opened and closed (Amplitude)
- What are the most clicked actions?

2) **Why does Mozilla need to answer these questions?  Are there benefits for users? Do we need this information to address product or business requirements?**

* To measure how useful these features are
* To see if the quick action sheet is "discoverable" enough based on its adoption

3) **What alternative methods did you consider to answer these questions? Why were they not sufficient?**

* N/A (These are baseline metrics)

4) **Can current instrumentation answer these questions?**

* Currently no, as these are some of the first metrics we're recording

5) **List all proposed measurements and indicate the category of data collection for each measurement, using the Firefox [data c](https://wiki.mozilla.org/Firefox/Data_Collection)[ollection ](https://wiki.mozilla.org/Firefox/Data_Collection)[categories](https://wiki.mozilla.org/Firefox/Data_Collection) on the found on the Mozilla wiki.**

* All data is Category 2.

6) **How long will this data be collected?**

Until 03/01/2020

7) **What populations will you measure?**

* All release, beta, and nightly users with telemetry enabled.

8) **Please provide a general description of how you will analyze this data.**

* Glean / Amplitude

9) **Where do you intend to share the results of your analysis?**

* Only on Glean, Amplitude and with mobile teams.

sblatz:
@boek for data review!

boek:
@sblatz we should do the data review where we record the metrics.yml :)

components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenuBuilder.kt
```diff
@@ -10,13 +10,14 @@ import android.content.Context
  * Helper class for building browser menus.
  *
  * @param items List of BrowserMenuItem objects to compose the menu from.
+ * @param extras Map of extra values that can be provided (e.g. 'customTab')
```
pocmo: nit: Let's mention that they will be added to emitted `Fact`s? :)

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
```diff
@@ -137,7 +137,9 @@ internal class DisplayToolbar(
                 anchor = this,
                 orientation = BrowserMenu.determineMenuOrientation(toolbar))
 
-            emitOpenMenuFact()
+            val isCustom = (menuBuilder?.extras?.get("customTab")) as? Boolean
+
+            emitOpenMenuFact(isCustom == true)
```
pocmo: No need to do anything special for "customTab". Let's just pass along the whole map and then add it as metadata in `emitOpenMenuFact`.

components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenuBuilder.kt
```diff
@@ -10,13 +10,14 @@ import android.content.Context
...
  */
 class BrowserMenuBuilder(
-    val items: List<BrowserMenuItem>
+    val items: List<BrowserMenuItem>,
+    val extras: Map<String, Any>? = null
```
pocmo: nit: If you turn this into `emptyMap()` then you do not need to deal with nullability.

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
```diff
@@ -138,7 +144,10 @@ class CustomTabsToolbarFeature(
         }.also { items ->
             val combinedItems = menuBuilder?.let { builder -> builder.items + items } ?: items
 
-            toolbar.setMenuBuilder(BrowserMenuBuilder(combinedItems))
+            val extras = mapOf(
+                    "customTab" to true
+            )
+            toolbar.setMenuBuilder(BrowserMenuBuilder(combinedItems, extras))
```
pocmo: We may end up overriding extras the app has provided.

* Do we need to do anything here? If we allow the app to specific extras then it can attach the flag for us already.
* If we always want to provide this flag ourselves from the feature then let's combine the extras with the already existing ones (like we do for menu items). :)

components/browser/toolbar/README.md
```diff
@@ -34,9 +34,15 @@ This component emits the following [Facts](../../support/base/README.md#Facts):
...
-| CLICK  | menu    |                | The user opened the overflow menu. |
+| CLICK  | menu    | `menuExtras`   | The user opened the overflow menu. |
 | COMMIT | toolbar | `commitExtras` | The user has edited the URL.       |
 
+#### `menuExtras`
+
+| Key          | Type    | Value                                 |
+|--------------|---------|---------------------------------------|
+| custom_tab   | Boolean | Whether the menu is from a custom tab |
+
```
pocmo: I think we can remove this here now since that is not really added by the `browser-toolbar` component anymore.

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
sblatz: We don't actually need any of the app's `extras` so I've just chosen to rename this variable so we're not overriding it.

sblatz:
@pocmo resolved the issues you mentioned 😄

pocmo:
@sblatz I pushed a small update to this PR. All extras from `BrowserMenuBuilder` will be added to the emitted `Fact`. CustomTabsToolbarFeature injects `customTab: true` into that map (in addition to whatever is already there). Will that work for you? If yes then I think this is ready to land now. :)

PR #2683, opened 2019-04-08
Upgrade coroutines to fix main thread hang on init

colintheshots:
### Pull Request checklist

### DIFF
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -5,7 +5,7 @@
 // Synchronized version numbers for dependencies used by (some) modules
 object Versions {
     const val kotlin = "1.3.21"
-    const val coroutines = "1.1.1"
+    const val coroutines = "1.2.0-alpha-2"
 
     const val androidx_test = "1.1.0"
     const val androidx_runner = "1.1.0"


buildSrc/src/main/java/Dependencies.kt
```diff
@@ -5,7 +5,7 @@
 // Synchronized version numbers for dependencies used by (some) modules
 object Versions {
     const val kotlin = "1.3.21"
-    const val coroutines = "1.1.1"
+    const val coroutines = "1.2.0-alpha-2"
```
sarentz: I see this is using an `alpha-2` version. That sounds a bit scary, but if this change improves cold startup a lot, then I would like to ask if we can assess the risk of using an alpha version. Also, assuming this is temporary and that more alphas or beta or maybe a release version of the `coroutines` library will happen in the coming months.

jonalmeida: It might make sense to see if this change has the improvements closer to a stable release. I would be hesitant to merge an alpha release of a dependency.

colintheshots: I can only guess based on history. It seems likely we'll get a stable version in the next few weeks. This improvement is important for our MVP metrics, but we could possibly wait.

jonalmeida:
There's a PR (https://github.com/mozilla-mobile/android-components/pull/2739) to investigate the test failures that is being worked on. Closing.

PR #2694, opened 2019-04-09
Issue #2530: Add Autopush support components

jonalmeida:
### Pull Request checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,6 +210,8 @@ _Supporting components with generic helper code._
 
 * ⚪ [**Public Suffix List**](components/lib/publicsuffixlist/README.md) - A library for reading and using the [public suffix list](https://publicsuffix.org/).
 
+* 🔴 [**Push-Firebase**](components/lib/push-firebase/README.md) - A [concept-push](concept/push/README.md) implementation using [Firebase Cloud Messaging](https://firebase.google.com/products/cloud-messaging/).
+
 ## Tooling
 
 * 🔵 [**Fetch-Tests**](components/tooling/fetch-tests/README.md) - A generic test suite for components that implement [concept-fetch](concept/fetch/README.md).
--- a/README.md
+++ b/README.md
@@ -142,6 +142,8 @@ _Combined components to implement feature-specific use cases._
 
 * ⚪ [**Prompts**](components/feature/prompts/README.md) - A component that will handle all the common prompt dialogs from web content.
 
+* ⚪ [**Push**](components/feature/push/README.md) - A component that provides Autopush messages with help from a supported push service.
+
 * ⚪ [**Find In Page**](components/feature/findinpage/README.md) - A component that provides an UI widget for [find in page functionality](https://support.mozilla.org/en-US/kb/search-contents-current-page-text-or-links).
 
 * 🔴 [**Site Permissions**](components/feature/sitepermissions/README.md) - A feature for showing site permission request prompts.
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -47,6 +47,11 @@ object Versions {
         const val work = "2.0.0"
         const val arch_core_testing = "2.0.0"
     }
+
+    object Firebase {
+        const val core = "16.0.1"
+        const val messaging = "17.3.4"
+    }
 }
 
 // Synchronized dependencies used by (some) modules
@@ -112,4 +117,7 @@ object Dependencies {
     const val thirdparty_zxing = "com.google.zxing:core:${Versions.zxing}"
     const val thirdparty_jna = "net.java.dev.jna:jna:${Versions.jna}@jar"
     const val thirdparty_disklrucache = "com.jakewharton:disklrucache:${Versions.disklrucache}"
+
+    const val firebase_core = "com.google.firebase:firebase-core:${Versions.Firebase.core}"
+    const val firebase_messaging = "com.google.firebase:firebase-messaging:${Versions.Firebase.messaging}"
 }
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -107,6 +107,7 @@ object Dependencies {
     const val mozilla_sync_logins = "org.mozilla.appservices:logins:${Versions.mozilla_appservices}"
     const val mozilla_places = "org.mozilla.appservices:places:${Versions.mozilla_appservices}"
     const val mozilla_places_forUnitTests = "org.mozilla.appservices:places-forUnitTests:${Versions.mozilla_appservices}"
+    const val mozilla_push = "org.mozilla.appservices:push:${Versions.mozilla_appservices}"
     const val mozilla_rustlog = "org.mozilla.appservices:rustlog:${Versions.mozilla_appservices}"
     const val mozilla_servo_arm = "org.mozilla.servoview:servoview-armv7:${Versions.servo}"
     const val mozilla_servo_x86 = "org.mozilla.servoview:servoview-x86:${Versions.servo}"
--- a/components/concept/push/src/main/java/mozilla/components/concept/push/Bus.kt
+++ b/components/concept/push/src/main/java/mozilla/components/concept/push/Bus.kt
@@ -6,6 +6,8 @@
 
 package mozilla.components.concept.push
 
+import androidx.lifecycle.LifecycleOwner
+
 /**
  * Interface for a basic bus that is implemented by MessageBus so that classes can observe particular event types.
  *
@@ -36,6 +38,16 @@ interface Bus<T, M> {
      * @param observer The observer to be notified for the type.
      */
     fun register(type: T, observer: Bus.Observer<T, M>)
+
+    /**
+     * Registers an observer to get notified about events.
+     *
+     * @param observer The observer to be notified for the type.
+     * @param owner the lifecycle owner the provided observer is bound to.
+     * @param autoPause whether or not the observer should automatically be
+     * paused/resumed with the bound lifecycle.
+     */
+    fun register(type: T, observer: Bus.Observer<T, M>, owner: LifecycleOwner, autoPause: Boolean)
 
     /**
      * Unregisters an observer to stop getting notified about events.
--- a/components/concept/push/src/main/java/mozilla/components/concept/push/PushProcessor.kt
+++ b/components/concept/push/src/main/java/mozilla/components/concept/push/PushProcessor.kt
@@ -6,53 +6,58 @@
 
 package mozilla.components.concept.push
 
-import androidx.annotation.CallSuper
+import androidx.annotation.VisibleForTesting
 
 /**
  * A push notification processor that handles registration and new messages from the [PushService] provided.
  * Starting Push in the Application's onCreate is recommended.
  */
-abstract class PushProcessor {
+interface PushProcessor {
 
     /**
-     * Initialize the PushProcessor before it
+     * Start the push processor and any service associated.
      */
-    @CallSuper
-    open fun initialize() {
-        instance = this
-    }
+    fun initialize()
 
     /**
-     * Start the push processor and starts any service associated.
+     * Removes all push subscriptions from the device.
      */
-    abstract fun start()
-
-    /**
-     * Stop the push service and stops any service associated.
-     */
-    abstract fun stop()
+    fun shutdown()
 
     /**
      * A new registration token has been received.
      */
-    abstract fun onNewToken(newToken: String)
+    fun onNewToken(newToken: String)
 
     /**
      * A new push message has been received.
      */
-    abstract fun onMessageReceived(message: PushMessage)
+    fun onMessageReceived(message: EncryptedPushMessage)
 
     /**
      * An error has occurred.
      */
-    abstract fun onError(error: Error)
+    fun onError(error: PushError)
 
     companion object {
+        /**
+         * Initialize and installs the PushProcessor into the application.
+         * This needs to be called in the application's onCreate before a push service has started.
+         */
+        fun install(processor: PushProcessor) {
+            instance = processor
+        }
+
         @Volatile
-        var instance: PushProcessor? = null
+        private var instance: PushProcessor? = null
+
+        @VisibleForTesting
+        internal fun reset() {
+            instance = null
+        }
         val requireInstance: PushProcessor
             get() = instance ?: throw IllegalStateException(
-                "You need to call start() on your Push instance from Application.onCreate()."
+                "You need to call initialize() on your Push instance from Application.onCreate()."
             )
     }
 }
@@ -60,20 +65,37 @@ abstract class PushProcessor {
 /**
  * A push message holds the information needed to pass the message on to the appropriate receiver.
  */
-data class PushMessage(val data: String)
-
-/**
- * The different kinds of push messages.
- */
-enum class PushType {
-    Services,
-    ThirdParty
+data class EncryptedPushMessage(
+    val channelId: String,
+    val body: String,
+    val encoding: String,
+    val salt: String = "",
+    val cryptoKey: String = "" // diffie–hellman key
+) {
+    companion object {
+        /**
+         * The [salt] and [cryptoKey] are optional as part of the standard for WebPush, so we should default
+         * to empty strings.
+         *
+         * Note: We have use the invoke operator since secondary constructors don't work with nullable primitive types.
+         */
+        operator fun invoke(
+            channelId: String,
+            body: String,
+            encoding: String,
+            salt: String? = null,
+            cryptoKey: String? = null
+        ) = EncryptedPushMessage(channelId, body, encoding, salt ?: "", cryptoKey ?: "")
+    }
 }
 
 /**
  *  Various error types.
  */
-sealed class Error {
-    data class RegistrationError(val desc: String) : Error()
-    data class NetworkError(val desc: String) : Error()
+sealed class PushError(open val desc: String) {
+    data class Registration(override val desc: String) : PushError(desc)
+    data class Network(override val desc: String) : PushError(desc)
+    data class Rust(override val desc: String) : PushError(desc)
+    data class MalformedMessage(override val desc: String) : PushError(desc)
+    data class ServiceUnavailable(override val desc: String) : PushError(desc)
 }
--- a/components/feature/push/README.md
+++ b/components/feature/push/README.md
@@ -1,0 +1,69 @@
+# [Android Components](../../../README.md) > Feature > Push
+
+A compoennt that implements push notifications with a supported push service.
+
+## Usage
+
+Add a supported push service for providing the encrypted messages (for example Firebase Cloud Messaging via `lib-push-firebase`):
+```kotlin
+class FirebasePush : AbstractFirebasePushService()
+```
+
+Create a push configuration with the project info and also place the required service's API keys in the project directory:
+
+```kotlin
+PushConfig(
+  senderId = "push-test-f408f",
+  serverHost = "push.service.mozilla.com",
+  serviceType = ServiceType.FCM,
+  protocol = Protocol.HTTPS
+)
+```
+
+We can then start the AutoPushFeature to get the subscription info and decrypted push message:
+```kolin
+val service = FirebasePush()
+
+val feature = AutoPushFeature(
+  context = context,
+  service = pushService,
+  config = config
+)
+
+// To start the feature and the service.
+feature.initialize()
+
+// To stop the feature and the service.
+feature.shutdown()
+
+// To receive the subscription info for all the subscription changes.
+feature.registerForSubscriptions(object : PushSubscriptionObserver {
+  override fun onSubscriptionAvailable(subscription: AutoPushSubscription) {
+    // handle subscription info here.
+  }
+}, lifecycle, false)
+
+// Force request all subscription information.
+feature.subscribeAll()
+
+// To receive a message for a specific push message type.
+feature.registerForPushMessages(PushType.Services, object: Bus.Observer<PushType, String> {
+  override fun onEvent(type: PushType, message: String) {
+    // Handle decrypted message here.
+  }
+}, lifecycle, false)
+```
+
+### Setting up the dependency
+
+Use Gradle to download the library from [maven.mozilla.org](https://maven.mozilla.org/) ([Setup repository](../../../README.md#maven-repository)):
+
+```Groovy
+implementation "org.mozilla.components:feature-push:{latest-version}"
+```
+
+## License
+
+    This Source Code Form is subject to the terms of the Mozilla Public
+    License, v. 2.0. If a copy of the MPL was not distributed with this
+    file, You can obtain one at http://mozilla.org/MPL/2.0/
--- a/components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
+++ b/components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
@@ -1,0 +1,327 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ *  License, v. 2.0. If a copy of the MPL was not distributed with this
+ *  file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package mozilla.components.feature.push
+
+import android.content.Context
+import android.content.SharedPreferences
+import androidx.lifecycle.LifecycleOwner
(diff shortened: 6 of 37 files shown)


jonalmeida:
Some comments from @grigoryk after an offline conversation:

- Don't surface the errors for firebase. This isn't very useful for clients of the push feature to do anything.
- Surface the subscription info to subscribers.
- After some sleuthing in fennec, third party push subscribers only need the FCM registration token in order to do their thing.
  - Remove `ThirdParty` from `PushType`.
  - Surface a way for subscribers to read the registration token **and** listen to token updates.

README.md
```diff
@@ -210,6 +210,8 @@ _Supporting components with generic helper code._
 
 * ⚪ [**Public Suffix List**](components/lib/publicsuffixlist/README.md) - A library for reading and using the [public suffix list](https://publicsuffix.org/).
 
+* 🔴 [**Push-Firebase**](components/lib/push-firebase/README.md) - A [concept-push](concept/push/README.md) implementation using [Firebase Cloud Messaging](https://firebase.google.com/products/cloud-messaging/).
+
```
grigoryk: Add Feature-Push as well to this list.

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -28,6 +28,9 @@ object Versions {
 
     const val material = "1.0.0"
 
+    const val firebase_core = "16.0.1"
```
grigoryk: why are these different versions? I'm assuming that's necessary for something, but it's a little surprising. Can you add a comment clarifying this?

components/concept/push/src/main/java/mozilla/components/concept/push/Bus.kt
```diff
@@ -37,6 +39,13 @@ interface Bus<T, M> {
      */
     fun register(type: T, observer: Bus.Observer<T, M>)
 
+    /**
+     * Registers an observer to get notified about events.
+     *
+     * @param observer The observer to be notified for the type.
+     */
+    fun register(type: T, observer: Bus.Observer<T, M>, owner: LifecycleOwner, autoPause: Boolean)
```
grigoryk: nit: You _could_ document other params as well ;)

components/concept/push/src/main/java/mozilla/components/concept/push/PushProcessor.kt
```diff
@@ -6,74 +6,95 @@
...
 
     /**
      * An error has occurred.
      */
-    abstract fun onError(error: Error)
+    fun onError(error: PushError)
 
     companion object {
+        /**
+         * Initialize the PushProcessor that needs to be called before
```
grigoryk: unfinished comment?

components/concept/push/src/main/java/mozilla/components/concept/push/PushProcessor.kt
```diff
@@ -6,74 +6,95 @@
...
+    }
 }
 
 /**
  *  Various error types.
  */
-sealed class Error {
-    data class RegistrationError(val desc: String) : Error()
-    data class NetworkError(val desc: String) : Error()
+sealed class PushError(open val desc: String) {
```
grigoryk: nit: you could just define `desc` on member classes, without the overrides

components/feature/push/README.md
```diff
@@ -0,0 +1,21 @@
+# [Android Components](../../../README.md) > Feature > Push
+
+A compoennt that implements push notifications with a supported push service.
+
+## Usage
+
+TBD
```
grigoryk: nit: remove this or add something here, or file a follow-up issue and link it here

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
```diff
@@ -0,0 +1,325 @@
...
+import mozilla.components.support.base.log.logger.Logger
+import mozilla.components.support.base.observer.Observable
+import mozilla.components.support.base.observer.ObserverRegistry
+import java.io.File
+import java.util.UUID
+import java.util.concurrent.Executors
+import kotlin.coroutines.CoroutineContext
+
+/**
+ * A implementation of a [PushProcessor] that lives as a singleton. It receives messages from a service and
```
grigoryk: nit: it only "lives" in a singleton if your consumers make it so. this class doesn't enforce it

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
```diff
@@ -0,0 +1,325 @@
...
+ * forwards them to be decrypted and routed. Should be created with the PushProvider.
```
grigoryk: nit: PushProvider doesn't create this class

components/concept/push/src/main/java/mozilla/components/concept/push/PushProcessor.kt
```diff
@@ -6,74 +6,95 @@
...
      * An error has occurred.
      */
-    abstract fun onError(error: Error)
+    fun onError(error: PushError)
 
     companion object {
+        /**
+         * Initialize the PushProcessor that needs to be called before
+         */
+        fun install(processor: PushProcessor) {
```
grigoryk: I'm not really sure there's value in having this singletone stuff live at the concept level. I'd expect to see it in `feature-push`.

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
```diff
@@ -0,0 +1,325 @@
...
+ * Subscription information for each registered [PushType] can be retrieved asynchronously:
+ *
+ * <code>
+ *    feature.registerForSubscriptions(object: Bus.Observer<PushType, String> {
+ *        override fun onSubscriptionAvailable(subscription: AutoPushSubscription) { }
+ *    })
+ *    feature.subscribeForType(PushType.Services)
+ * </code>
+ *
+ * Push messages for each registered [PushType] can also be retrieved asynchronously:
```
grigoryk: "can be listened for..." is probably a better way to phrase that. `retrieved` implies historical data, e.g. messages received in the past.

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
```diff
@@ -0,0 +1,325 @@
...
+ * <code>
+ *     class Application {
+ *          override fun onCreate() {
+ *              val feature = AutoPushFeature(context, service, configuration)
+ *              PushProvider.install(push)
+ *          }
+ *     }
+ * </code>
+ *
+ * Subscription information for each registered [PushType] can be retrieved asynchronously:
```
grigoryk: `retrieved` is probably not the right word. see below

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
```diff
@@ -0,0 +1,325 @@
...
+        socketProtocol = config.protocol,
+        serviceType = config.serviceType,
+        databasePath = File(context.filesDir, DB_NAME).canonicalPath
+    )
+) : PushProcessor {
+
+    private val logger = Logger("AutoPushFeature")
+    private val subscriptionObservers: Observable<PushSubscriptionObserver> = ObserverRegistry()
+    private val messageObserverBus: Bus<PushType, String> = MessageBus()
+    private val prefToken = preferences(context).getString(PREF_TOKEN, null)
```
grigoryk: what is this? add a comment

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
```diff
@@ -0,0 +1,325 @@
...
+
+    internal var job: Job = SupervisorJob()
+    private val scope = CoroutineScope(coroutineContext) + job
+
+    init {
+        // If we have a token, initialize the rust component first.
+        prefToken?.let { token ->
+            runBlocking(coroutineContext) {
```
grigoryk: do you actually need to block on this? Won't `scope.launch { }` work?

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
```diff
@@ -0,0 +1,325 @@
...
+        type: PushType,
+        observer: Bus.Observer<PushType, String>,
+        owner: LifecycleOwner,
+        autoPause: Boolean
+    ) = messageObserverBus.register(type, observer, owner, autoPause)
+
+    /**
+     * Un-subscribes from all push message channels. This should only be done on an account logout or app data deletion.
+     */
+    override fun shutdown() {
```
grigoryk: nit: group `overrides` together at the top of the class

grigoryk: Rename this method. `unsubcribeAll` is a better name. Current name implies you call this during a regular app lifecycle, and that this method is oppose to `initialize`.

components/concept/push/src/main/java/mozilla/components/concept/push/PushProcessor.kt
```diff
@@ -6,74 +6,95 @@
...
-     */
-    abstract fun start()
+    fun initialize()
 
     /**
-     * Stop the push service and stops any service associated.
+     * Removes all push subscriptions from the device.
      */
-    abstract fun stop()
+    fun shutdown()
```
grigoryk: rename this (`unsubscribeAll`?), current name implies that this is opposite to `initialize` and is part of the app lifecycle.

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
```diff
@@ -0,0 +1,325 @@
...
+        service.stop()
+
+        DeliveryManager.with(connection) {
+            scope.launch {
+                // TODO replace with unsubscribeAll API when available
```
grigoryk: oh hey, look at that - it's already going to use `unsubscribeAll` underneath :)

components/concept/push/src/main/java/mozilla/components/concept/push/PushProcessor.kt
grigoryk: Also, I think you want this to return `Deferred<Unit>`, so that consumers can block (if they want to) on the results of this. E.g. if we want to ensure that we've done with this operation before doing other cleanup.

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
```diff
@@ -0,0 +1,325 @@
...
+    /**
+     * New registration tokens are received and sent to the Autopush server which also performs subscriptions for
+     * each push type and notifies the subscribers.
+     */
+    override fun onNewToken(newToken: String) {
+        scope.launchAndTry {
+            connection.updateToken(newToken)
+
+            // Subscribe all only if this is the first time.
+            if (prefToken.isNullOrEmpty()) {
```
grigoryk: who can modify `prefToken`? is this thread-safe?

buildSrc/src/main/java/Dependencies.kt
jonalmeida: The core library is separate from the messaging one, as we add more firebase features, each one has their own dependency version. I'm not sure what the value of adding a comment to this would be. How about a "firebase" object similar to how we've done it for AndroidX?

components/concept/push/src/main/java/mozilla/components/concept/push/PushProcessor.kt
jonalmeida: If you have a when expression, you would need to cast it to a specific type to access the description then (see AutoPushFeature#onError which is one use case).

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
jonalmeida: Good point. I've changed this to "should live" so that it's clear, since this instance doesn't need to be a singleton, but _an_ instance does.

components/concept/push/src/main/java/mozilla/components/concept/push/PushProcessor.kt
jonalmeida: Every time I ask someone, I keep moving it to the location they prefer until the next review. :)

I think this is fine at the concept level since it's a necessity for a PushProcessor to be installed into the lifecycle of the Application regardless of the feature that implements it.

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
jonalmeida: If we have a token stored locally, we need to make sure the push API is init'd with that first. If we register a new token, and then the old one messages would never be delivered. This is a small cost since initialization of the Push API is just instantiation.

components/concept/push/src/main/java/mozilla/components/concept/push/PushProcessor.kt
jonalmeida: > rename this (unsubscribeAll?), current name implies that this is opposite to initialize and is part of the app lifecycle.

Since we're stopping the push service as well, we figured that this should be called something more dramatic. I've updated the KDocs, but did we still want to change it to "unsubscribeAll" since it does more than that.

> Also, I think you want this to return Deferred<Unit>, so that consumers can block (if they want to) on the results of this. E.g. if we want to ensure that we've done with this operation before doing other cleanup.

I'll look this up to verify, but I don't think I can since I'm cancelling the job right after.

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
jonalmeida: See above comment.

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
jonalmeida: Only the feature modifies it. SharedPrefs are thread safe (not process safe).

components/lib/push-firebase/src/main/java/mozilla/components/lib/push/firebase/AbstractFirebasePushService.kt
```diff
@@ -0,0 +1,77 @@
...
+        }
+    }
+
+    /**
+     * Stops the Firebase messaging service.
+     */
+    final override fun stop() {
+        FirebaseMessaging.getInstance().apply {
+            isAutoInitEnabled = false
+            stopSelf()
```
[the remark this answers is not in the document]
jonalmeida: This is an Android API: https://developer.android.com/reference/android/app/Service.html#stopSelf()

components/lib/push-firebase/src/main/java/mozilla/components/lib/push/firebase/AbstractFirebasePushService.kt
```diff
@@ -0,0 +1,77 @@
...
+import mozilla.components.concept.push.PushService
+
+abstract class AbstractFirebasePushService : FirebaseMessagingService(), PushService {
+
+    /**
+     * Initializes Firebase and starts the messaging service if not already started.
+     */
+    override fun start(context: Context) {
+        FirebaseApp.initializeApp(context)
+        FirebaseMessaging.getInstance().isAutoInitEnabled = true
```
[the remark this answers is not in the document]
jonalmeida: Without this, if we stop the service, the next initialization will start the service automatically.

In the previous instance when we spoke about it, I was using it incorrectly to "stop" the service before deleting the token - this is no longer a valid option.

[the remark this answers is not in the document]
jonalmeida: I've added API comments for start/stop.

components/lib/push-firebase/src/main/java/mozilla/components/lib/push/firebase/AbstractFirebasePushService.kt
```diff
@@ -0,0 +1,77 @@
...
+            }
+        }
+    }
+
+    /**
+     * Stops the Firebase messaging service.
+     */
+    final override fun stop() {
+        FirebaseMessaging.getInstance().apply {
+            isAutoInitEnabled = false
```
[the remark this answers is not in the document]
jonalmeida: See above comments at 'start' that apply here too.

components/lib/push-firebase/src/main/java/mozilla/components/lib/push/firebase/AbstractFirebasePushService.kt
```diff
@@ -0,0 +1,77 @@
...
+            stopSelf()
+        }
+    }
+}
+
+/**
+ * A helper method that allows you to check if the device is supported by Google Play Services in order to
+ * receive push notifications from Firebase Cloud Messaging.
+ */
+fun isPlayServicesAvailable(context: Context): Boolean {
```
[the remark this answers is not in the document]
jonalmeida: Part of the previous review I had with Christian was that, if we can avoid the calling app from touching implementation-level APIs, we should.

If this a debatable topic, I'll remove it to move the review along.

components/lib/push-firebase/src/main/java/mozilla/components/lib/push/firebase/AbstractFirebasePushService.kt
```diff
@@ -0,0 +1,77 @@
...
+            try {
+                val message = EncryptedPushMessage(
+                    channelId = it.data.getValue("chid"),
+                    body = it.data.getValue("body"),
+                    encoding = it.data.getValue("con"),
+                    salt = it.data["enc"],
+                    cryptoKey = it.data["cryptokey"]
+                )
+                PushProcessor.requireInstance.onMessageReceived(message)
+            } catch (e: NoSuchElementException) {
```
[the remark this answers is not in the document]
jonalmeida: If I move the `onMessageReceived` to below the try block I won't have a message to send (compiler error).

components/feature/push/src/main/java/mozilla/components/feature/push/MessageBus.kt
```diff
@@ -0,0 +1,76 @@
...
+import androidx.lifecycle.LifecycleOwner
+import mozilla.components.concept.push.Bus
+import mozilla.components.support.base.observer.ObserverRegistry
+
+typealias Registry<T, M> = ObserverRegistry<Bus.Observer<T, M>>
+
+/**
+ * An implementation of [Bus] where the event type is restricted to an enum.
+ */
+class MessageBus<T : Enum<T>, M> : Bus<T, M> {
```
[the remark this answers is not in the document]
jonalmeida: The bus was supposed to live in the `support-base` to be multi-purpose. We wanted to refrain from that until we find a better place for it so that `support-base` doesn't become a dumping ground for utility functions. Sebastian recommended we keep it in the feature for now until we know where to move it.

Removing this would be a bigger change. Would doing this in a follow-up work?

components/feature/push/src/main/java/mozilla/components/feature/push/Connection.kt
```diff
@@ -0,0 +1,121 @@
...
+internal class RustPushConnection(
+    private val databasePath: String,
+    private val senderId: String,
+    private val serverHost: String,
+    private val socketProtocol: Protocol,
+    private val serviceType: ServiceType
+) : PushConnection {
+
+    @VisibleForTesting
+    internal var api: PushAPI? = null
```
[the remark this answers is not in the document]
jonalmeida: It's set in the updateToken. It's mutable because the PushManager needs a registration token to be instantiated the first time, after that we can call the PushManager's updateToken.

The purpose of this wrapper is to take out that requirement since it would need every call to the PushManager to be checked first. It also helps us by removing the need to add a-s unit-testing dependencies since we only need to test that the API contract is being kept without needing to call their code.

components/feature/push/src/main/java/mozilla/components/feature/push/Connection.kt
```diff
@@ -0,0 +1,121 @@
...
+
+    override suspend fun subscribe(channelId: String, scope: String): SubscriptionResponse = synchronized(this) {
+        val pushApi = api
```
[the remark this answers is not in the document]
jonalmeida: This is another case of multiple folks to reviewing this change: it was recommended in order to remove the `!!` at the caller since the API is nullable (see above comments for more details on this).

The val ensures the same binding is used when calling the method's below.

components/feature/push/src/main/java/mozilla/components/feature/push/Connection.kt
```diff
@@ -0,0 +1,121 @@
...
+        check(pushApi != null) { "Rust API is not initiated; updateToken hasn't been called yet." }
+        return pushApi.subscribe(channelId, scope)
+    }
+
+    override suspend fun unsubscribe(channelId: String): Boolean = synchronized(this) {
+        val pushApi = api
+        check(pushApi != null) { "Rust API is not initiated; updateToken hasn't been called yet." }
```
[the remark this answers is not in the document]
jonalmeida: See above comment.

components/feature/push/src/main/java/mozilla/components/feature/push/Connection.kt
```diff
@@ -0,0 +1,121 @@
...
+        channelId: String,
+        body: String,
+        encoding: String = "",
+        salt: String = "",
+        cryptoKey: String = ""
+    ): ByteArray
+    fun isInitialized(): Boolean
+}
+
+internal class RustPushConnection(
```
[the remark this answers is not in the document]
jonalmeida: The purpose of this wrapper is to take out the requirement of needing a registration token in order to init the PushManager since it would need every call to the PushManager to be checked first. By having this wrapper, hold onto the other attributes, and init the PushManager when we get the registration token later, we can reduce the complexity that would fall into the feature.

Yes it does help a lot with testing - we don't need to add a-s unit-testing dependencies since we only need to test that the API contract is being kept without needing to call their code.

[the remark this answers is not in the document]
jonalmeida: I can add a follow-up issue to discuss this there if that helps.

components/feature/push/src/main/java/mozilla/components/feature/push/Connection.kt
```diff
@@ -0,0 +1,121 @@
...
+    private val senderId: String,
+    private val serverHost: String,
+    private val socketProtocol: Protocol,
+    private val serviceType: ServiceType
+) : PushConnection {
+
+    @VisibleForTesting
+    internal var api: PushAPI? = null
+
+    override suspend fun subscribe(channelId: String, scope: String): SubscriptionResponse = synchronized(this) {
```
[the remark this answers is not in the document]
jonalmeida: Nice, I didn't know about that annotation! I can also removing the `runBlocking` from the feature since the synchronized also covers it's use case.

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
jonalmeida: See below comment. Since access to the API is synchronized, this runBlocking is no longer needed.

components/concept/push/src/main/java/mozilla/components/concept/push/PushProcessor.kt
grigoryk: Sounds good. You own this, it's your call ;)

components/feature/push/src/main/java/mozilla/components/feature/push/MessageBus.kt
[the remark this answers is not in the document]
grigoryk: Sure. Try not to loose track of this.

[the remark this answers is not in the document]
jonalmeida: Filed https://github.com/mozilla-mobile/android-components/issues/3171

components/feature/push/src/main/java/mozilla/components/feature/push/Connection.kt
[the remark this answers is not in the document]
jonalmeida: Filed https://github.com/mozilla-mobile/android-components/issues/3172

PR #2703, opened 2019-04-10
Expose automaticLanguageAdjustment, and automatically detect the locale.

jhugman:
This PR adds a setting to the `concept-engine` to track the locale the device is set to and hint to the engine to change the `Accept-Language` headers sent to server.

Fixes [fenix#942][1]. 

[1]: https://github.com/mozilla-mobile/fenix/issues/942

### Pull Request checklist

~- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features~

### DIFF
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
@@ -6,6 +6,7 @@ package mozilla.components.browser.engine.gecko
 
 import android.content.Context
 import android.util.AttributeSet
+import mozilla.components.browser.engine.gecko.integration.LocaleSettingUpdater
 import mozilla.components.browser.engine.gecko.mediaquery.from
 import mozilla.components.browser.engine.gecko.mediaquery.toGeckoValue
 import mozilla.components.concept.engine.Engine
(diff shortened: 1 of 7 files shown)


components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
```diff
@@ -114,6 +117,13 @@ class GeckoEngine(
             get() = runtime.settings.automaticFontSizeAdjustment
             set(value) { runtime.settings.automaticFontSizeAdjustment = value }
 
+        override var automaticLanguageAdjustment: Boolean
```
csadilek: @jhugman Do we still need to initialize this field based on the default setting? To make sure the setter is called and the locale updated if the default value is `true`, similar to: https://github.com/mozilla-mobile/android-components/blob/b0824f92891b095ff50eec9234f565a41631aa9a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt#L159

jhugman: Good catch!

I've taken the source of truth for if the option is working in the getter by looking at the `localeUpdater` directly.

csadilek:
@jhugman If we want higher coverage and make codecov happy we'd have to add the setting to https://github.com/mozilla-mobile/android-components/blob/master/components/concept/engine/src/test/java/mozilla/components/concept/engine/SettingsTest.kt

But just a nit in this case :). Ready to land.

csadilek:
✨

Issue #2708, opened 2019-04-10
Create Bookmark Suggestion Provider


ekager:
We will need an `AwesomeBar.SuggestionProvider` implementation that provides suggestions based on the bookmark storage.

jonalmeida:
Closing manually since the PR is merged.

PR #2709, opened 2019-04-10
For #2708 - Adds BookmarksStorageSuggestionProvider

ekager:
### Pull Request checklist

### DIFF
--- a/components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/BookmarksStorageSuggestionProvider.kt
+++ b/components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/BookmarksStorageSuggestionProvider.kt
@@ -1,0 +1,53 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.feature.awesomebar.provider
+
+import mozilla.components.browser.icons.BrowserIcons
+import mozilla.components.concept.awesomebar.AwesomeBar
+import mozilla.components.concept.storage.BookmarkNode
+import mozilla.components.concept.storage.BookmarksStorage
+import mozilla.components.feature.awesomebar.internal.loadLambda
+import mozilla.components.feature.session.SessionUseCases
+import java.util.UUID
+
+private const val BOOKMARKS_SUGGESTION_LIMIT = 20
+
+/**
+ * A [AwesomeBar.SuggestionProvider] implementation that provides suggestions based on the bookmarks
+ * stored in the [BookmarksStorage].
+ */
+class BookmarksStorageSuggestionProvider(
+    private val bookmarksStorage: BookmarksStorage,
+    private val loadUrlUseCase: SessionUseCases.LoadUrlUseCase,
+    private val icons: BrowserIcons? = null
+) : AwesomeBar.SuggestionProvider {
+
+    override val id: String = UUID.randomUUID().toString()
+
+    override suspend fun onInputChanged(text: String): List<AwesomeBar.Suggestion> {
+        if (text.isEmpty()) {
+            return emptyList()
+        }
+
+        val suggestions = bookmarksStorage.searchBookmarks(text, BOOKMARKS_SUGGESTION_LIMIT)
+        return suggestions.filter { it.url != null }.distinctBy { it.url }.sortedBy { it.guid }
+            .into()
+    }
+
+    /**
+    * Expects list of BookmarkNode to be specifically of bookmarks (e.g. nodes with a url).
+    */
+    private fun List<BookmarkNode>.into(): List<AwesomeBar.Suggestion> {
+        return this.map {
+            AwesomeBar.Suggestion(
+                provider = this@BookmarksStorageSuggestionProvider,
+                icon = icons.loadLambda(it.url!!),
+                title = it.title,
+                description = it.url,
+                onSuggestionClicked = { loadUrlUseCase.invoke(it.url!!) }
+            )
+        }
+    }
+}
--- a/components/feature/awesomebar/src/test/java/mozilla/components/feature/awesomebar/provider/BookmarksStorageSuggestionProviderTest.kt
+++ b/components/feature/awesomebar/src/test/java/mozilla/components/feature/awesomebar/provider/BookmarksStorageSuggestionProviderTest.kt
@@ -1,0 +1,156 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.feature.awesomebar.provider
+
+import kotlinx.coroutines.runBlocking
+import mozilla.components.concept.storage.BookmarkInfo
+import mozilla.components.concept.storage.BookmarkNode
+import mozilla.components.concept.storage.BookmarkNodeType
+import mozilla.components.concept.storage.BookmarksStorage
+import mozilla.components.support.test.mock
+import mozilla.components.support.utils.StorageUtils.levenshteinDistance
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import java.util.UUID
+
+@RunWith(RobolectricTestRunner::class)
+class BookmarksStorageSuggestionProviderTest {
+    val bookmarks = testableBookmarksStorage()
+    private val newItem = BookmarkNode(
+        BookmarkNodeType.ITEM, "123", "456", null,
+        "Mozilla", "http://www.mozilla.org", null
+    )
+
+    @Test
+    fun `Provider returns empty list when text is empty`() = runBlocking {
+        val provider = BookmarksStorageSuggestionProvider(mock(), mock())
+
+        val suggestions = provider.onInputChanged("")
+        assertTrue(suggestions.isEmpty())
+    }
+
+    @Test
+    fun `Provider returns suggestions from configured bookmarks storage`() = runBlocking {
+        val provider = BookmarksStorageSuggestionProvider(bookmarks, mock())
+
+        bookmarks.addItem("Mobile", newItem.url!!, newItem.title!!, null)
+
+        val suggestions = provider.onInputChanged("moz")
+        assertEquals(1, suggestions.size)
+        assertEquals("http://www.mozilla.org", suggestions[0].description)
+    }
+
+    @Test
+    fun `Provider does not return duplicate suggestions`() = runBlocking {
+        val provider = BookmarksStorageSuggestionProvider(bookmarks, mock())
+
+        for (i in 1..20) {
+            bookmarks.addItem("Mobile", newItem.url!!, newItem.title!!, null)
+        }
+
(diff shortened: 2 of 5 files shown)


components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesStorage.kt
```diff
@@ -27,7 +27,7 @@ open class PlacesStorage(context: Context) : Storage, SyncableStore {
     private val storageDir by lazy { context.filesDir }
 
     @VisibleForTesting
-    internal open val places: Connection by lazy {
+    open val places: Connection by lazy {
```
ekager: I'm not sure if I should do this but I can't override in my test otherwise IIUC?

ekager:
So I needed some way to test using `BookmarksStorage`, but bookmarks doesn't have the same non-sync class that I could use for testing like `InMemoryHistoryStorage`. I mocked up a fake class based on `InMemoryHistoryStorage` with the methods I need for testing purposes, but let me know if this doesn't work for ya'll and if you have other ideas.

components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/BookmarksStorageSuggestionProvider.kt
```diff
@@ -0,0 +1,49 @@
...
+) : AwesomeBar.SuggestionProvider {
+
+    override val id: String = UUID.randomUUID().toString()
+
+    override suspend fun onInputChanged(text: String): List<AwesomeBar.Suggestion> {
+        if (text.isEmpty()) {
+            return emptyList()
+        }
+
+        return bookmarksStorage.searchBookmarks(text, BOOKMARKS_SUGGESTION_LIMIT).into()
```
pocmo: FYI: Last week we applied a small patch to the provider here to make sure the list is sorted and does not contain any duplicates: https://github.com/mozilla-mobile/android-components/blob/master/components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/HistoryStorageSuggestionProvider.kt#L37

components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/BookmarksStorageSuggestionProvider.kt
```diff
@@ -0,0 +1,49 @@
...
+
+    override val id: String = UUID.randomUUID().toString()
+
+    override suspend fun onInputChanged(text: String): List<AwesomeBar.Suggestion> {
+        if (text.isEmpty()) {
+            return emptyList()
+        }
+
+        val suggestions = bookmarksStorage.searchBookmarks(text, BOOKMARKS_SUGGESTION_LIMIT)
+        return suggestions.distinctBy { it.url }.into()
```
grigoryk: :+1: it's possible we'll have duplicate URLs returned, so good thing we're de-duping here. Going forward that's the expectation from the providers (or you'll see runtime exceptions thrown a layer above, as you might have already).

components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/BookmarksStorageSuggestionProvider.kt
```diff
@@ -0,0 +1,49 @@
...
+ * A [AwesomeBar.SuggestionProvider] implementation that provides suggestions based on the bookmarks
+ * stored in the [BookmarksStorage].
+ */
+class BookmarksStorageSuggestionProvider(
+    private val bookmarksStorage: BookmarksStorage,
+    private val loadUrlUseCase: SessionUseCases.LoadUrlUseCase,
+    private val icons: BrowserIcons? = null
+) : AwesomeBar.SuggestionProvider {
+
+    override val id: String = UUID.randomUUID().toString()
```
grigoryk: :+1: This should be fine as long as we don't swap out our bookmark providers mid-flight. That is, we use this ID later on to form a type of a guid, so it needs to be stable.

components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/BookmarksStorageSuggestionProvider.kt
grigoryk: Do you know in which order we get back the results? The order, ideally, needs to be stable. You can add a further `sortBy { it.guid }` here to make sure. Or, confirm with the API that the order _is_ stable across calls... But best to make sure on our end.

components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/BookmarksStorageSuggestionProvider.kt
```diff
@@ -0,0 +1,49 @@
...
+
+        val suggestions = bookmarksStorage.searchBookmarks(text, BOOKMARKS_SUGGESTION_LIMIT)
+        return suggestions.distinctBy { it.url }.into()
+    }
+
+    private fun List<BookmarkNode>.into(): List<AwesomeBar.Suggestion> {
+        return this.map {
+            AwesomeBar.Suggestion(
+                provider = this@BookmarksStorageSuggestionProvider,
+                icon = icons.loadLambda(it.url!!),
```
grigoryk: The underlying API doesn't guarantee that there's a `url`, which is why you're casting here. From their docs: "@return A list of bookmarks where either the URL or the title contain a word (e.g. space separated item) from the query."

This reads to me like you may get back a folder. and you probably don't want to throw if the `url` isn't there. This is something you can double-check above at the `onInputChanged` level.

components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/BookmarksStorageSuggestionProvider.kt
```diff
@@ -0,0 +1,49 @@
...
+                title = it.title,
+                description = it.url,
+                onSuggestionClicked = { loadUrlUseCase.invoke(it.url!!) }
```
grigoryk: nit: you don't need subsequent `!!`, it's auto-cast to a non-nullable after the first one.

components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/BookmarksStorageSuggestionProvider.kt
grigoryk: From api docs: "The order of the results is unspecified." - so yeah, sort this.

components/feature/awesomebar/src/test/java/mozilla/components/feature/awesomebar/provider/BookmarksStorageSuggestionProviderTest.kt
```diff
@@ -0,0 +1,194 @@
...
+            bookmarks.addItem(BookmarkRoot.Mobile.id, newItem.url!!, newItem.title!!, null)
+        }
+
+        val suggestions = provider.onInputChanged("moz")
+        assertEquals(1, suggestions.size)
+    }
+
+    @Test
+    fun `Provider limits number of returned unique suggestions`() = runBlocking {
+        val provider = BookmarksStorageSuggestionProvider(bookmarks, mock())
```
grigoryk: Perhaps you can just add an InMemoryBookmarkStorage to `browser-storage-memory`? You already did some of that work here. Up to you :)

components/feature/awesomebar/src/test/java/mozilla/components/feature/awesomebar/provider/BookmarksStorageSuggestionProviderTest.kt
```diff
@@ -0,0 +1,194 @@
...
+        override suspend fun runMaintenance() {
+            TODO("Not needed for the test")
+        }
+
+        override fun cleanup() {
+            TODO("Not needed for the test")
+        }
+
+        // Borrowed from https://gist.github.com/ademar111190/34d3de41308389a0d0d8
+        private fun levenshteinDistance(a: String, b: String): Int {
```
grigoryk: `support-utils` is a good place for this.

docs/changelog.md
```diff
@@ -26,6 +26,9 @@ permalink: /changelog/
 * **feature-findinpage**
    * Find in Page now emits facts
 
+* **feature-awesomebar**
```
grigoryk: Wrong name.

components/feature/awesomebar/build.gradle
```diff
@@ -43,6 +43,7 @@ dependencies {
 
     testImplementation project(':support-test')
     testImplementation project(':browser-storage-memory')
+    testImplementation project(':browser-storage-sync')
```
grigoryk: You don't need this dependency, or at least I can't find how you're using it.

components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/BookmarksStorageSuggestionProvider.kt
grigoryk: cc @thomcc if you feel like changing the API :)

docs/changelog.md
ekager: Which name should it be? Isn't it in `mozilla.components.feature.awesomebar.provider`?

grigoryk: Ah! You're right! My bad, I got wrapped up thinking about storage layers :)

components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/BookmarksStorageSuggestionProvider.kt
ekager: I added two filters one here and one in `onInputChanged` to make sure that the URLs aren't `null`, but I don't think map is smart enough to not let me have the `!!` in both places...

components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/BookmarksStorageSuggestionProvider.kt
```diff
@@ -0,0 +1,50 @@
...
+            return emptyList()
+        }
+
+        val suggestions = bookmarksStorage.searchBookmarks(text, BOOKMARKS_SUGGESTION_LIMIT)
+        return suggestions.filter { it.url != null }.distinctBy { it.url }.sortedBy { it.guid }
+            .into()
+    }
+
+    private fun List<BookmarkNode>.into(): List<AwesomeBar.Suggestion> {
+        return this.filter { it.url != null }.map {
```
grigoryk: You already do this filtering above; i'd drop this filter but add a comment saying that `into` method expects the list to be specifically for bookmarks (e.g. nodes with a `url`). This way we'll do a little bit less work overall! i.e. filter things.

components/support/utils/src/main/java/mozilla/components/support/utils/StorageUtils.kt
```diff
@@ -0,0 +1,42 @@
+package mozilla.components.support.utils
```
grigoryk: License block.

components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/BookmarksStorageSuggestionProvider.kt
thomcc: Yep, totally willing to change how that function works to better suit usage.

PR #2713, opened 2019-04-11
1541493: Add thread-safe timing API

mdboom:
This adds the thread-safe timing API that doesn't require passing any (extra) objects around.

One thing to note -- I originally envisioned the `TimingManager` being a member variable of `TimespanMetricType`, which would avoid the need for using the metric name in the `Map`.  Unfortunately that doesn't work for labeled metrics where the metrics are created "on-the-fly" -- they end up getting a new map each time so you can't "stop" and timing interval.

Using a `WeakHashMap` might be a premature optimization, but I also felt like saying "here, hand us your big singleton carrying everything on the page, and we *promise* to not hold on to it" felt a little too risky.  Alternatively, we can require integers in the API and put the onus on the user to provide us with an int using `object.hashCode`.

### Pull Request checklist

### DIFF
--- a/components/service/glean/docs/metrics/adding-new-metrics.md
+++ b/components/service/glean/docs/metrics/adding-new-metrics.md
@@ -329,18 +329,22 @@ rendering:
     ...
 ```
 
-Now you can use the timespan from the application's code:
+Now you can use the timespan from the application's code. Each time interval that
+the timespan metric records must be associated with an object provided by the user.
+This is so that intervals can be measured concurrently.  For example, to measure page
+load time on a number of tabs that are loading at the same time, each time interval
+should be associated with an object that uniquely represents each tab.
 
 ```Kotlin
-import org.mozilla.yourApplication.GleanMetrics.Rendering
+import org.mozilla.yourApplication.GleanMetrics.Pages
 
-Rendering.imageDecodeTime.start() // start the timer
-try {
-   // ... decode an image ...
-} catch (e: Exception) {
-   Rendering.imageDecodeTime.cancel() // cancel the timer -- nothing is recorded
+fun onPageStart(e: Event) {
+    Pages.pageLoad.start(e.target)
 }
-Rendering.imageDecodeTime.stopAndSum() // stop the timer
+
+fun onPageLoaded(e: Event) {
+    Pages.pageLoad.stopAndSum(e.target)
+}
 ```
 
 The time reported in the telemetry ping will be the sum of all of these
@@ -349,13 +353,13 @@ timespans recorded during the lifetime of the ping.
 There are test APIs available too:
 
 ```Kotlin
-import org.mozilla.yourApplication.GleanMetrics.Rendering
+import org.mozilla.yourApplication.GleanMetrics.Pages
 Glean.enableTestingMode()
 
 // Was anything recorded?
-assertTrue(Rendering.imageDecodeTime.testHasValue())
+assertTrue(Pages.pageLoad.testHasValue())
 // Does the timer have the expected value
-assertTrue(Rendering.imageDecodeTime.testGetValue() > 0)
+assertTrue(Pages.pageLoad.testGetValue() > 0)
 ```
 
 ## Timing Distributions
@@ -388,15 +392,22 @@ pages:
 ```
 
 
-Now that the timing distribution is defined in `metrics.yaml` you can use the metric to record data
-in the application code:
+Now you can use the timing distribution from the application's code. Each time interval that
+the timing distribution metric records must be associated with an object provided by the user.
+This is so that intervals can be measured concurrently.  For example, to measure page
+load time on a number of tabs that are loading at the same time, each time interval
+should be associated with an object that uniquely represents each tab.
 
 ```Kotlin
 import org.mozilla.yourApplication.GleanMetrics.Pages
 
-// ...
-pages.pageLoad.accumulate(1L) // Accumulates a sample of 1 millisecond
-pages.pageLoad.accumulate(10L) // Accumulates a sample of 10 milliseconds
+fun onPageStart(e: Event) {
+    Pages.pageLoad.start(e.target)
+}
+
+fun onPageLoaded(e: Event) {
+    Pages.pageLoad.stopAndAccumulate(e.target)
+}
 ```
 
 There are test APIs available too.  For convenience, properties `sum` and `count` are exposed to 
--- a/components/service/glean/docs/metrics/adding-new-metrics.md
+++ b/components/service/glean/docs/metrics/adding-new-metrics.md
@@ -316,34 +316,40 @@ of 500 nanoseconds will be truncated to 0 microseconds.
 
 ### Usage
 
-Say you're adding a new timespan for the time spent decoding images. First you need
+Say you're adding a new timespan for the time spent logging into the app. First you need
 to add an entry for the counter to the `metrics.yaml` file:
 
 ```YAML
-rendering:
-  image_decode_time:
+auth:
+  login_time:
     type: timespan
     description: >
-      Measures the time spent decoding images.
-    time_unit: microseconds
+      Measures the time spent logging in.
+    time_unit: milliseconds
     ...
 ```
 
-Now you can use the timespan from the application's code. Each time interval that
-the timespan metric records must be associated with an object provided by the user.
-This is so that intervals can be measured concurrently.  For example, to measure page
-load time on a number of tabs that are loading at the same time, each time interval
-should be associated with an object that uniquely represents each tab.
+Now you can use the timespan from the application's code. Each time interval
+that the timespan metric records must be associated with an object provided by
+the user. This is so that intervals can be measured concurrently. In our example
+using login time, this might be an object representing the login UI page.
 
 ```Kotlin
-import org.mozilla.yourApplication.GleanMetrics.Pages
+import org.mozilla.yourApplication.GleanMetrics.Auth
 
-fun onPageStart(e: Event) {
-    Pages.pageLoad.start(e.target)
+fun onShowLogin(e: Event) {
+    Auth.loginTime.start(e.target)
+    // ...
 }
 
-fun onPageLoaded(e: Event) {
-    Pages.pageLoad.stopAndSum(e.target)
+fun onLogin(e: Event) {
+    Auth.loginTime.stop(e.target)
+    // ...
+}
+
+fun onLoginCancel(e: Event) {
+    Auth.loginTime.cancel(e.tart)
+    // ...
 }
 ```
 
@@ -353,13 +359,13 @@ timespans recorded during the lifetime of the ping.
 There are test APIs available too:
 
 ```Kotlin
-import org.mozilla.yourApplication.GleanMetrics.Pages
+import org.mozilla.yourApplication.GleanMetrics.Auth
 Glean.enableTestingMode()
 
 // Was anything recorded?
-assertTrue(Pages.pageLoad.testHasValue())
+assertTrue(Auth.loginTime.testHasValue())
 // Does the timer have the expected value
-assertTrue(Pages.pageLoad.testGetValue() > 0)
+assertTrue(Auth.loginTime.testGetValue() > 0)
 ```
 
 ## Timing Distributions
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/private/TimespanMetricType.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/private/TimespanMetricType.kt
@@ -37,16 +37,16 @@ data class TimespanMetricType(
      * called with no corresponding [stopAndSum]): in that case the original
      * start time will be preserved.
      *
-     * @param anyObject The object to associate with this timing.  This allows
+     * @param timerId The object to associate with this timing.  This allows
      * for concurrent timing of events associated with different objects to the
      * same timespan metric.
      */
-    fun start(anyObject: Any) {
+    fun start(timerId: Any) {
         if (!shouldRecord(logger)) {
             return
         }
 
-        TimingManager.start(this, anyObject)
+        TimingManager.start(this, timerId)
     }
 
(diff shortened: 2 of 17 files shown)


components/service/glean/docs/metrics/adding-new-metrics.md
```diff
@@ -329,18 +329,22 @@ rendering:
...
+the timespan metric records must be associated with an object provided by the user.
+This is so that intervals can be measured concurrently.  For example, to measure page
+load time on a number of tabs that are loading at the same time, each time interval
+should be associated with an object that uniquely represents each tab.
 
 ```Kotlin
-import org.mozilla.yourApplication.GleanMetrics.Rendering
+import org.mozilla.yourApplication.GleanMetrics.Pages
+
+fun onPageStart(e: Event) {
```
Dexterp37: nit: I would use a different example, this is identical to the timing distribution one and could cause confusion. What about measuring the "startup time"?

components/service/glean/src/main/java/mozilla/components/service/glean/private/TimespanMetricType.kt
```diff
@@ -30,39 +32,58 @@ data class TimespanMetricType(
...
     /**
-     * Start tracking time for the provided metric. This records an error if it’s
-     * already tracking time (i.e. start was already called with no corresponding
-     * [stopAndSum]): in that case the original start time will be preserved.
+     * Start tracking time for the provided metric and associated object. This
+     * records an error if it’s already tracking time (i.e. start was already
+     * called with no corresponding [stopAndSum]): in that case the original
+     * start time will be preserved.
+     *
+     * @param anyObject The object to associate with this timing.  This allows
```
Dexterp37: nit: I would be a bit more explicit with the name of the parameter here. Maybe something like `timerId` instead of `anyObject`

components/service/glean/src/main/java/mozilla/components/service/glean/private/TimespanMetricType.kt
```diff
@@ -30,39 +32,58 @@ data class TimespanMetricType(
...
+     * Stop tracking time for the provided metric and associated object. Add the
+     * elapsed time to the time currently stored in the metric. This will record
+     * an error if no [start] was called.
+     *
+     * @param anyObject The object to associate with this timing.  This allows
+     * for concurrent timing of events associated with different objects to the
+     * same timespan metric.
      */
-    fun stopAndSum() {
+    fun stopAndSum(anyObject: Any) {
```
Dexterp37: nit: I would be a bit more explicit with the name of the parameter here. Maybe something like `timerId` instead of `anyObject`

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
```diff
@@ -76,19 +77,21 @@ internal open class TimingDistributionsStorageEngineImplementation(
...
         }
 
+        val sampleInUnit = getAdjustedTime(timeUnit, sample)
+
         // Since the custom combiner closure captures this value, we need to just create a dummy
         // value here that won't be used by the combine function, and create a fresh
         // TimingDistributionData for each value that doesn't have an existing current value.
         val dummy = TimingDistributionData(category = metricData.category, name = metricData.name,
             timeUnit = timeUnit)
         super.recordMetric(metricData, dummy, null) { currentValue, newValue ->
```
Dexterp37: nit: since we're here, can we fix the `newValue` warning? If this isn't used, let's change it to `_`

components/service/glean/src/main/java/mozilla/components/service/glean/private/TimespanMetricType.kt
```diff
@@ -30,39 +32,58 @@ data class TimespanMetricType(
...
+     * same timespan metric.
      */
-    fun start() {
+    fun start(anyObject: Any) {
         if (!shouldRecord(logger)) {
             return
         }
 
-        TimespansStorageEngine.start(this)
+        TimingManager.start(this, anyObject)
```
Dexterp37: Please note that `TimingManager` is now shared across all the different metric types. Consider this example:

```
fun onPageLoaded(target: Event) {
  Pages.postLoadOperationTimespan.start(target)

  // some post load op happens

  // Accumulate the time
  Pages.onLoadDistribution.stopAndSum(target) // This will use postLoadOperationTimespan start time
  Pages.postLoadOperationTimespan.stopAndSum(target) // This will be broken?
}
```

I think the `TimingManager` should distinguish between the different metric types. This can either happen by having the manager be an instance within each relevant storage engine (thus making the storage engine depend on the timing manager) or by having the timing manager know about the "concept" of storage engines, having a structure like `Map[Engine, Map[Id, Time]]`. I'd personally go with the former, which gives us one less `object` to care about. But I have no strong opinion on this.

components/service/glean/src/main/java/mozilla/components/service/glean/private/TimingDistributionMetricType.kt
```diff
@@ -33,26 +34,65 @@ data class TimingDistributionMetricType(
...
+     * called with no corresponding [stopAndAccumulate]): in that case the original
+     * start time will be preserved.
      *
-     * @param sample This is the sample to accumulate
+     * @param anyObject The object to associate with this timing.  This allows
+     * for concurrent timing of events associated with different objects to the
+     * same timespan metric.
      */
-    fun accumulate(sample: Long) {
+    fun start(anyObject: Any) {
```
Dexterp37: nit: I would be a bit more explicit with the name of the parameter here. Maybe something like `timerId` instead of `anyObject`

components/service/glean/src/main/java/mozilla/components/service/glean/private/TimingDistributionMetricType.kt
```diff
@@ -33,26 +34,65 @@ data class TimingDistributionMetricType(
...
+    /**
+     * Stop tracking time for the provided metric and associated object. Add a
+     * count to the corresponding bucket in the timing distribution.
+     * This will record an error if no [start] was called.
+     *
+     * @param anyObject The object to associate with this timing.  This allows
+     * for concurrent timing of events associated with different objects to the
+     * same timespan metric.
+     */
+    fun stopAndAccumulate(anyObject: Any) {
```
Dexterp37: nit: I would be a bit more explicit with the name of the parameter here. Maybe something like `timerId` instead of `anyObject`

components/service/glean/src/main/java/mozilla/components/service/glean/timing/TimingManager.kt
```diff
@@ -0,0 +1,107 @@
...
+    internal var getElapsedNanos = { SystemClock.elapsedRealtimeNanos() }
+
+    /**
+     * Start tracking time associated with the provided object. This records an
+     * error if it’s already tracking time (i.e. start was already called with
+     * no corresponding [stop]): in that case, the original start time will be
+     * preserved.
+     *
+     * @param metricData The metric managing the timing. Used for error reporting.
+     * @param anyObject The object to associate with this timing.
```
Dexterp37: nit: same concern wrt `anyObject`

components/service/glean/src/main/java/mozilla/components/service/glean/timing/TimingManager.kt
```diff
@@ -0,0 +1,107 @@
...
+     */
+    public fun start(metricData: CommonMetricData, anyObject: Any) {
+        val startTime = getElapsedNanos()
+        val metricName = metricData.identifier
+
+        uncommittedStartTimes[metricName]?.let { metricTimings ->
```
Dexterp37: Should this access be `synchronized` as well?

components/service/glean/src/main/java/mozilla/components/service/glean/timing/TimingManager.kt
```diff
@@ -0,0 +1,107 @@
...
+            uncommittedStartTimes.getOrPut(metricName, { WeakHashMap<Any, Long>() })[anyObject] = startTime
+        }
+    }
+
+    /**
+     * Stop tracking time for the associated object. This will record
+     * an error if no [start] was called.
+     *
+     * @param metricData The metric managing the timing. Used for error reporting.
+     * @param anyObject The object to associate with this timing.
```
Dexterp37: nit: same concern wrt `anyObject`

components/service/glean/src/main/java/mozilla/components/service/glean/timing/TimingManager.kt
```diff
@@ -0,0 +1,107 @@
...
+                )
+                null
+            }
+        }
+    }
+
+    /**
+     * Abort a previous [start] call. No error is recorded if no [start] was called.
+     *
+     * @param anyObject The object to associate with this timing.
```
Dexterp37: nit: same concern wrt `anyObject`

components/service/glean/src/test/java/mozilla/components/service/glean/TestUtil.kt
```diff
@@ -118,6 +120,8 @@ internal fun resetGlean(
     config: Configuration = Configuration(),
     clearStores: Boolean = true
 ) {
+    TimingManager.getElapsedNanos = { SystemClock.elapsedRealtimeNanos() }
```
Dexterp37: Do we need this? Isn't this the same of TimingManager.kt:37?

components/service/glean/src/test/java/mozilla/components/service/glean/storages/TimespansStorageEngineTest.kt
```diff
@@ -116,109 +96,31 @@ class TimespansStorageEngineTest {
...
+            listOf("store1"),
+            timeUnit = TimeUnit.Nanosecond
+        )
 
         // Record the time for the first chunk.
-        doReturn(0.toLong()).`when`(spiedEngine).getElapsedNanos()
-        spiedEngine.start(metric)
-        doReturn(expectedChunkNanos).`when`(spiedEngine).getElapsedNanos()
-        spiedEngine.stopAndSum(metric, TimeUnit.Nanosecond)
+        TimingManager.getElapsedNanos = { 0 }
```
Dexterp37: Oh, I now see why you reset `TimingManager.getElapsedNanos` in `resetGlean`. Should we instead put this in the `@Before` above? This function should only really be overridden in timing related test files, not all of them.

components/service/glean/docs/metrics/adding-new-metrics.md
georgf: +1, ideally an example that makes sense to only measure once per app run.
"startup time" doesn't have a "cancel" scenario... maybe something like "logging into the app" or "load app data from disk"?

georgf: Also update around line 319 and 323 here.

components/service/glean/docs/metrics/adding-new-metrics.md
```diff
@@ -329,18 +329,22 @@ rendering:
...
+    Pages.pageLoad.start(e.target)
```
georgf: Do we have to require the parameter even for timespans?
I can see concurrency issues with the timing distributions, but i wonder if we can avoid the parameter for the timespan API.

components/service/glean/src/test/java/mozilla/components/service/glean/TestUtil.kt
mdboom: This resets the timer to something "real" after it may have been overridden by a test.

components/service/glean/src/test/java/mozilla/components/service/glean/storages/TimespansStorageEngineTest.kt
mdboom: Sure -- I can move it there.  In that case, I think it should be in `@After`, though.

components/service/glean/docs/metrics/adding-new-metrics.md
mdboom: I don't think there's a requirement that timespans are used once per application run, since they sum together a bunch of timespans.  It could be used to measure the total time spent performing a task, which might be performed on multiple threads, even concurrently.  I think it's safest (fewest foot-guns) to require the parameter, since

components/service/glean/src/main/java/mozilla/components/service/glean/private/TimespanMetricType.kt
mdboom: I'm not sure this is an issue.  The current implementation uses `Map[MetricName, Map[Id, Time]]`.  Since metric names are globally unique (you can't have a timer and a timing distribution with the same name), this should be ok.  I'm not opposed to having separate `TimingManager` instances for each storage engine though as insurance against this.

Dexterp37: Oh, good point. Not needed!

components/service/glean/src/test/java/mozilla/components/service/glean/TestUtil.kt
Dexterp37: This shouldn't be needed for *all the tests*, right? Maybe we should just be careful about this in the specific test files (see below)

components/service/glean/docs/metrics/adding-new-metrics.md
mdboom: Good suggestion.  I'll make the timer example be "logging into the app".

components/service/glean/src/main/java/mozilla/components/service/glean/timing/TimingManager.kt
mdboom: I was thinking since it's read-only maybe not, but I think you're right...

Also, this raises a concern about blocking on these.  I think I need to rework the async here so these updates happen off the main thread...

PR #2691, opened 2019-04-09
[glean] 1540725: Make labeled metrics invididual metric types.

mdboom:
Depends on https://github.com/mozilla/glean_parser/pull/56

### Pull Request checklist

- [x] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [x] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/android-components/blob/master/docs/changelog.md) or does not need one
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features

PR #2716, opened 2019-04-11
[glean] 15443487: Don't use Kotlin introspection

mdboom:
Requires https://github.com/mozilla/glean_parser/pull/57 and https://github.com/mozilla-mobile/android-components/pull/2691 to be merged first.

This is a workaround to not require Kotlin introspection.

We currently use introspection to get the string value of the extra key from the enum value. The idea here is to not try to get the value directly from the enum, but just store a List of the strings with each event metric and use the enum's ordinal to look up the string in that List.

### Pull Request checklist


docs/changelog.md
```diff
@@ -30,6 +30,10 @@ permalink: /changelog/
      intended for public use, part of the glean API has been renamed from
      `mozilla.components.service.glean.metrics` to
      `mozilla.components.service.glean.private`.
+   * ⚠️ **This is a breaking API change**: Labeled metrics are now their own
```
Dexterp37: I guess this (as with some other changes) come from the fact that you're based off some other local branch?

components/service/glean/src/main/java/mozilla/components/service/glean/private/EventMetricType.kt
```diff
@@ -62,14 +63,10 @@ data class EventMetricType<ExtraKeysEnum : Enum<ExtraKeysEnum>>(
...
-            // Here we need to get (2) to send in the ping.  We need to use reflection to get
-            // around limitations in Java generics (since the actual enums for the event are in
-            // a template parameter).
+            // Here we need to get (2) to send in the ping.
             if (it.size > 0) {
-                val cls = it.keys.first().javaClass
-                val method = cls.getDeclaredMethod("getValue")
                 it.mapKeys { entry ->
-                    method.invoke(entry.key).toString()
+                    allowedExtraKeys[entry.key.ordinal]
```
Dexterp37: This is highly unlikely (it would mean a bug on our end), but what if `allowedExtraKeys` is empty? Would this be a NPE? Should we deal with that?

docs/changelog.md
mdboom: Yeah -- I still need to get the labeled stuff merged first (hence the WIP on this).  That should clear out before I finally merge.

components/service/glean/src/main/java/mozilla/components/service/glean/private/EventMetricType.kt
mdboom: That's a good point.  At a minimum we should avoid the NPE, but I don't know if we have to do anything much more sophisticated.

Dexterp37: Yes, nothing fancy. I think using [`elemetnAtOrElse`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/element-at-or-else.html) and printing an error should be enough.

mdboom:
This change will require an immediate snapshot again to avoid breaking Fenix :(  Would love to have a better solution for that...

mdboom:
@pocmo: This is ready to merge, but it's another one where we'll want to merge-and-snapshot-right-away.

PR #2717, opened 2019-04-11
Update CODEOWNERS so anyone can review the changelog

mdboom:
I don't have a lot of experience with `CODEOWNERS`, but I *think* this will let anyone on any mozilla-mobile team review PRs involving the changelog.

### Pull Request checklist


CODEOWNERS
```diff
@@ -19,6 +19,9 @@
 # the repo. Unless a later match takes precedence.
 * @mozilla-mobile/act
 
+# Changelog is reviewable by anyone on any team
+/docs/changelog.md @mozilla-mobile
```
mitchhentges: Hmm, I'm not sure if this works organization-wide (even when just assigning reviewers directly, it's not possible to add an "organization" as a reviewer), but there's only one way to find out :)

Note that there's two repercussions to having an entity as a CODEOWNER:
1. They're able to provide the :heavy_check_mark: for the change
2. They're emailed since they've been assigned as a reviewer

If this _does_ work, this means that everybody in `@mozilla-mobile` will be emailed every time this `changelog` is changed, IIRC :thinking:

mdboom: > Hmm, I'm not sure if this works organization-wide (even when just assigning reviewers directly, it's not possible to add an "organization" as a reviewer), but there's only one way to find out :)

CODEOWNERS [isn't well documented](https://help.github.com/articles/about-codeowners/), but all of the examples there use organizations with out the `/team` suffix, so that gives me some hope this would work.

> They're emailed since they've been assigned as a reviewer

Is that true?  For example, I'm in the `@mozilla-mobile/telemetry` team and don't get e-mails for everything in our part of the repo -- only when someone explicitly requests my review.  But that may be an notification preferences thing somewhere...  We certainly don't want everyone spammed by this, though, so I'm fine to reject this PR if that's the case.

mitchhentges: I think that in the `CODEOWNERS` examples without the `/team` suffix, they're referring to a user, not an organization? Unsure.

Weird if you're not getting emails for `@mozilla-mobile/telemetry`. Maybe it's a per-org setting, in which case this can/should be merged and individuals can ensure that they're not getting review requests for org-wide? Maybe I turned mine on :thinking: 

Either way, I think I'm :+1: on this, but I'll leave it for someone in ACT to stamp with :heavy_check_mark:

PR #2719, opened 2019-04-11
Closes #2718: Updates web error strings

### LINKED ISSUE #2718: Update web error strings
We want new strings based on the copy [supplied here](https://docs.google.com/spreadsheets/d/1SIAm3XZz5uGOm39rAR1hcCje7XYOfqqfSkxnXWttBeU/edit#gid=1453201767).

sblatz:
### Pull Request checklist

### DIFF
--- a/components/browser/errorpages/src/main/java/mozilla/components/browser/errorpages/ErrorPages.kt
+++ b/components/browser/errorpages/src/main/java/mozilla/components/browser/errorpages/ErrorPages.kt
@@ -28,6 +28,7 @@ object ErrorPages {
             .bufferedReader()
             .use { it.readText() }
             .replace("%pageTitle%", context.getString(R.string.mozac_browser_errorpages_page_title))
+            .replace("%backButton%", context.getString(R.string.mozac_browser_errorpages_page_go_back))
             .replace("%button%", context.getString(R.string.mozac_browser_errorpages_page_refresh))
             .replace("%messageShort%", context.getString(errorType.titleRes))
             .replace("%messageLong%", context.getString(errorType.messageRes, uri))
--- a/components/browser/errorpages/src/main/res/values/strings.xml
+++ b/components/browser/errorpages/src/main/res/values/strings.xml
@@ -7,19 +7,30 @@
     <string name="mozac_browser_errorpages_page_title">Problem loading page</string>
 
     <!-- The button that appears at the bottom of an error page. -->
-    <string name="mozac_browser_errorpages_page_refresh">Try Again</string>
+    <string name="mozac_browser_errorpages_page_refresh">Try again</string>
+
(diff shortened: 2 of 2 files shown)


jonalmeida:
@Delphine since you went through the strings with me the first time (with encoding), did you want to have a look at these updates as well?

jonalmeida:
> Sure, I can help.
> However, could you give me some background on why we are updating all these strings now? We have already exposed them to localization, and most of the work is already done.
> This is going to be a lot of "double" work for our localizers who have already gone through these strings, that were considered as final when we got them.
> Please let me know. Thanks
> (also CC @Pike so he's in the loop)

I believe these were updates that @mheubusch was working on and can weigh in.

Delphine:
Thanks. Some questions/comments:
1. Are you sure these strings are final? it seems like this is a WIP, correct? If so, can you try working on a non-localizable file until they are considered as final - and only then expose them on master? The idea is that what's on master goes to localization and should be final

2. We are a bit confused by some strings in this file, in particular the ones where they contain placeholders for branding and URL - as well as at least one string that seems empty ("mozac_browser_errorpages_unknown_socket_type_message"). 
Can you please show us screenshots of these strings in use, so we can double check what's going on and how/where they appear in the UI?
Are these coming from[ this file](https://docs.google.com/spreadsheets/d/1SIAm3XZz5uGOm39rAR1hcCje7XYOfqqfSkxnXWttBeU/edit#gid=1453201767)? If so please let us know what cells they show up on

3. Apostrophes: we use curly apostrophes and quotes in strings (not needed in localization comments)

4. Most of these strings are going to need new strings IDs, since they change entirely. The only time I see here where you won't need to update the string ID is when the you're just changing from small caps to caps or vice versa (see https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_content_best_practices#Changing_existing_strings for more info)

Spoke with Michelle offline, and it's probably best to not expose these strings right now. There's also still some work to do on them. Please let us know in the meantime about point 1 and 2 above though!
thanks

pocmo:
> Are you sure these strings are final? it seems like this is a WIP, correct? If so, can you try working on a non-localizable file until they are considered as final - and only then expose them on master?

I suggest we do that in the Fenix repo if we want to experiment with those strings: Fenix can override strings from AC by using the same string ids (e.g. in a non-localizable file). Once we are confident that the strings are final (and we want to use the across all our products) we can land them here in AC (with new ids!).

mheubusch:
So, no need to add strings separately for MVP - please use the existing strings provided by Android Components that are currently in use.

sblatz:
Since we're wanting to hold off on these strings for now, I will close this PR but keep my branch around in case we want them in the future.

PR #2721, opened 2019-04-12
Guarantee that pending glean pings are fully written to disk

Dexterp37:
Glean pings are currently directly written to their final location. Some other thread might be scanning that location for uploading pings and might see partially written files.
This PR writes pings to the cache directory and, after they are fully written, moves them to their final destination.

**Note**: This additionally fixes a weird log line and removes the warning in the files

@pocmo is there a better, Android friendly way to do the same? We looked at  `AtomicFile` but the docs state it should not be used for this. For the moving part, to ensure atomicity, we should have used [`Files.move`](https://developer.android.com/reference/java/nio/file/Files.html#move(java.nio.file.Path,%20java.nio.file.Path,%20java.nio.file.CopyOption...)), but it's API 26+ only :(

### Pull Request checklist


Dexterp37:
I manually tested this:

> D/glean/PingMaker: Collecting baseline
> D/glean/PingMaker: Collecting events
> D/glean/PingStorageEngine: Storing ping 55a8c497-5619-4c37-b3b2-bb95e5c3d641 at /data/user/0/org.mozilla.samples.glean.debug/glean_data/pending_pings/55a8c497-5619-4c37-b3b2-bb95e5c3d641
> D/glean/PingStorageEngine: Processing persisted pings at /data/user/0/org.mozilla.samples.glean.debug/glean_data/pending_pings
> D/glean/PingStorageEngine: Processing ping: 55a8c497-5619-4c37-b3b2-bb95e5c3d641
> D/glean/HttpPingUploader: Submitting ping to: https://incoming.telemetry.mozilla.org/submit/org-mozilla-samples-glean-debug/baseline/1/55a8c497-5619-4c37-b3b2-bb95e5c3d641
> D/glean/HttpPingUploader: Ping successfully sent (200)
> D/glean/PingStorageEngine: 55a8c497-5619-4c37-b3b2-bb95e5c3d641 was deleted: true
> I/WM-WorkerWrapper: Worker result SUCCESS for Work [ id=aa8909fa-86e3-431b-a81d-26f07641b68b, tags={ mozac_service_glean_ping_upload_worker, mozilla.components.service.glean.scheduler.PingUploadWorker } ]

Dexterp37:
> 
> 
> Question: What happens to the cache directory if we crashed while writing the pings to it and before we move them to the processing directory? Is it persistent from run to run or is it temporary?

In that case, we'd lose the ping. The cache directory isn't necessarily cleared, but if we crash after writing and before moving, we have no way to get the data back. I don't think this is a big deal, we would have the same problem if we crashed while writing at the final destination.

mdboom:
I think, before this change, if we crashed while writing, we might have a truncated file being sent to the uploader.  With this change, at least we shouldn't ever have truncated files being sent.  But, yes, AFAICT the OS is free to clear out the cache directory at any time behind our backs, even while the app is running.  But I presume not files that are actively open and being written to.

pocmo:
> @pocmo is there a better, Android friendly way to do the same? 

Two thoughts:

* This seems to be only an issue if you read/write concurrently. Could you synchronize on that? Or maybe even use the same thread for that? If you always only *write* or *read* a file then this should not be a problem, right?

* Looking at the internals of AtomicFile, they seem to do a sync after the write.. that may be important here too?
https://android.googlesource.com/platform/frameworks/support/+/androidx-master-dev/core/src/main/java/androidx/core/util/AtomicFile.java#212

Dexterp37:
> 
> 
> > @pocmo is there a better, Android friendly way to do the same?
> 
> Two thoughts:
> 
>     * This seems to be only an issue if you read/write concurrently. Could you synchronize on that? Or maybe even use the same thread for that? If you always only _write_ or _read_ a file then this should not be a problem, right?

Yes, this is where we expect to see this problem. Glean consumers can trigger pings at any given time, thus writing pings to disk. The upload worker runs in its own thread, managed by the `WorkManager`, and scans the directory for files to upload.

We do our best to make sure the worker is only triggered after `sendPings` has finished I/O, but multiple consecutive calls to that function might result in writes to the disk while another thread is scanning.

>     * Looking at the internals of AtomicFile, they seem to do a sync after the write.. that may be important here too?
>       https://android.googlesource.com/platform/frameworks/support/+/androidx-master-dev/core/src/main/java/androidx/core/util/AtomicFile.java#212

Ah! You are right! I thought that was the behaviour of `flush` (which we are calling), but by carefully [reading the docs](https://docs.oracle.com/javase/8/docs/api/java/io/Writer.html#flush--), looks like it isn't. There's no guarantee that flush... flushes to disk. We need to use `sync` :D

Dexterp37:
> > ```
> > * Looking at the internals of AtomicFile, they seem to do a sync after the write.. that may be important here too?
> >   https://android.googlesource.com/platform/frameworks/support/+/androidx-master-dev/core/src/main/java/androidx/core/util/AtomicFile.java#212
> > ```
> 
> Ah! You are right! I thought that was the behaviour of `flush` (which we are calling), but by carefully [reading the docs](https://docs.oracle.com/javase/8/docs/api/java/io/Writer.html#flush--), looks like it isn't. There's no guarantee that flush... flushes to disk. We need to use `sync` :D

Thinking of this again, I think we should be still ok without it, with the current approach: we're writing to the cache directory, and we're moving the file as soon as the writer is closed.

PR #2723, opened 2019-04-12
Regenerate the client id / first run date if needed

Dexterp37:
Old builds of products using glean, which were already executed once, once updated, might not correctly report the client id and the first run date. Since these are likely low number of developer builds, we simply regenerate these fields.

### Pull Request checklist


components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -216,20 +216,41 @@ open class GleanInternalAPI internal constructor () {
...
+            if (legacyClientId != null) {
+                UuidsStorageEngine.record(GleanInternalMetrics.clientId, legacyClientId)
+            } else {
+                // Apparently, this is a very old build that was already run
+                // at least once and has a different name for the underlying
+                // UUID storage engine persistence (i.e. UuidStorageEngine.xml
+                // vs using the full class name with the namespace).
+                // This should be mostly devs and super early adopters, so just regenerate
+                // the data.
+                UuidsStorageEngine.record(GleanInternalMetrics.clientId, UUID.randomUUID())
```
mdboom: Random thought -- we could record an error on the metric here so we could see how often this happens.  If it shows up again in the future, it might help us detect mistakes...

travis79: Ooh, I like this idea!

Dexterp37: Yeah, that might be a good idea. The thing is that, in contrast to desktop, we have no alerting on this _for now_. It it would be up to humans (that's us!) to manually check in that for now. I'd rather have a monitor or alerting on schema validation errors (that caught this!!!!), which is a lower hanging fruit with the tooling that's available right now. Additionally, I'm planning on prioritizing the integration test.

Dexterp37:
> 
> 
> r+ (Do we have/should we file) a bug to look at this again in 6 months to remove the cruft (after it's hopefully no longer needed)?

I filed this https://bugzilla.mozilla.org/show_bug.cgi?id=1543979

Issue #2740, opened 2019-04-15
Update detekt to latest version


pocmo:
We are using: 1.0.0-RC11
Latest version is: 1.0.0-RC14

Testing with the latest version I get a lot of `LongMethod` warnings and some `LargeClass` warnings. 

We may want to less strict configurations for those rules (at least for LongMethod`?).

```
        LongMethod - 62/20 - [applyAutocompleteResult] at /Users/skaspari/Projects/Mozilla/android-components/components/ui/autocomplete/src/main/java/mozilla/components/ui/autocomplete/InlineAutocompleteEditText.kt:386:5
        LongMethod - 27/20 - [afterTextChanged] at /Users/skaspari/Projects/Mozilla/android-components/components/ui/autocomplete/src/main/java/mozilla/components/ui/autocomplete/InlineAutocompleteEditText.kt:584:9
        LongMethod - 36/20 - [createDoorhanger] at /Users/skaspari/Projects/Mozilla/android-components/components/ui/doorhanger/src/main/java/mozilla/components/ui/doorhanger/DoorhangerPrompt.kt:38:5
        LongMethod - 26/20 - [createView] at /Users/skaspari/Projects/Mozilla/android-components/components/ui/doorhanger/src/main/java/mozilla/components/ui/doorhanger/DoorhangerPrompt.kt:161:1
        LongMethod - 29/20 - [createBoxAnimatorSet] at /Users/skaspari/Projects/Mozilla/android-components/components/ui/tabcounter/src/main/java/mozilla/components/ui/tabcounter/TabCounter.kt:125:5
        LongMethod - 21/20 - [createBarAnimatorSet] at /Users/skaspari/Projects/Mozilla/android-components/components/ui/tabcounter/src/main/java/mozilla/components/ui/tabcounter/TabCounter.kt:174:5
        LongMethod - 21/20 - [setCountWithAnimation] at /Users/skaspari/Projects/Mozilla/android-components/components/ui/tabcounter/src/main/java/mozilla/components/ui/tabcounter/TabCounter.kt:66:5
        LargeClass - 159/150 - [TabCounter] at /Users/skaspari/Projects/Mozilla/android-components/components/ui/tabcounter/src/main/java/mozilla/components/ui/tabcounter/TabCounter.kt:22:1
[...]

PR #2741, opened 2019-04-15
Update ktlint to 0.31.0.

pocmo:
For detekt I filed a new issue: #2740. There's more to fix for the latest detekt version and right now I don't want to go through that.


pocmo:
Nice, now it passed.

PR #2749, opened 2019-04-15
Add periodic updates to service-experiments updater.

travis79:
Use WorkManager along with a PeriodicWorkRequest to update experiments every 6 hours.

### Pull Request checklist

### DIFF
--- a/components/service/experiments/src/main/java/mozilla/components/service/experiments/ExperimentsUpdater.kt
+++ b/components/service/experiments/src/main/java/mozilla/components/service/experiments/ExperimentsUpdater.kt
@@ -6,11 +6,32 @@ package mozilla.components.service.experiments
 
 import android.content.Context
 import android.support.annotation.VisibleForTesting
-import kotlinx.coroutines.Dispatchers.IO
-import kotlinx.coroutines.launch
-import kotlinx.coroutines.GlobalScope
-import kotlinx.coroutines.Job
+import androidx.work.Constraints
+import androidx.work.ExistingPeriodicWorkPolicy
+import androidx.work.NetworkType
+import androidx.work.PeriodicWorkRequest
+import androidx.work.WorkManager
+import androidx.work.Worker
+import androidx.work.WorkerParameters
 import mozilla.components.support.base.log.logger.Logger
+import java.util.concurrent.TimeUnit
+
+/**
+ * This class represents a [Worker] object that will fetch updates to the experiments.
+ */
+internal class ExperimentsUpdaterWorker(
+    context: Context,
+    workerParameters: WorkerParameters
+) : Worker(context, workerParameters) {
+
+    override fun doWork(): Result {
+        return if (Experiments.updater.updateExperiments()) {
+            Result.success()
+        } else {
+            Result.failure()
+        }
+    }
+}
 
 /**
  * Class used to schedule sync of experiment
@@ -25,23 +46,60 @@ internal class ExperimentsUpdater(
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
     internal lateinit var source: KintoExperimentSource
 
-    @Volatile private var updateJob: Job? = null
-
     internal fun initialize(configuration: Configuration) {
         source = getExperimentSource(configuration)
 
-        updateJob = GlobalScope.launch(IO) {
-            // Update the experiments list from the server async, without blocking
-            // the app launch.
-            updateExperiments()
-        }
+        // Schedule the periodic experiment updates
+        scheduleUpdates()
     }
 
-    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
-    internal suspend fun testWaitForUpdate() {
-        updateJob!!.join()
+    /**
+     * Schedule update with the default constraints, but only if the periodic worker is not already
+     * scheduled.
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun scheduleUpdates() {
+        WorkManager.getInstance().enqueueUniquePeriodicWork(
+            TAG,
+            // We rely on REPLACE behavior to run immediately and then schedule the next update per
+            // the specified interval.  KEEP behavior does not run immediately and it is desired to
+            // have fresh experiments from the server on startup.
+            ExistingPeriodicWorkPolicy.REPLACE,
+            getWorkRequest())
     }
 
+    /**
+     * Builds the periodic work request that will control the scheduling of the worker and provide
+     * the tag value so that the worker can be interacted with later.
+     *
+     * @return [PeriodicWorkRequest]
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun getWorkRequest(): PeriodicWorkRequest {
+        return PeriodicWorkRequest.Builder(
+            ExperimentsUpdaterWorker::class.java, UPDATE_INTERVAL_HOURS, TimeUnit.HOURS)
+            .addTag(TAG)
+            .setConstraints(getWorkConstraints())
+            .build()
+    }
+
+    /**
+     * Builds the [Constraints] for when the worker is allowed to run.
+     *
+     * @return [Constraints]
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun getWorkConstraints(): Constraints {
+        return Constraints.Builder()
+            .setRequiredNetworkType(NetworkType.CONNECTED)
+            .build()
+    }
+
+    /**
+     * Returns the experiment source for which to update the experiments
+     *
+     * @return [KintoExperimentSource]
+     */
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
     internal fun getExperimentSource(configuration: Configuration): KintoExperimentSource {
         return KintoExperimentSource(
@@ -55,6 +113,8 @@ internal class ExperimentsUpdater(
     /**
      * Requests new experiments from the server and
      * saves them to local storage
+     *
+     * @return true if the experiments were updated, false if there was an [ExperimentDownloadException]
      */
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
     @Synchronized
@@ -74,8 +134,10 @@ internal class ExperimentsUpdater(
     companion object {
         private const val LOG_TAG = "experiments"
 
-        private const val TAG = "mozilla.components.service.experiments"
-        private const val UPDATE_INTERVAL_HOURS = 6L
+        @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+        internal const val TAG = "mozilla.components.service.experiments"
+        @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+        internal const val UPDATE_INTERVAL_HOURS = 6L
 
         // Where to access the experiments data in Kinto.
         // Live example: https://firefox.settings.services.mozilla.com/v1/buckets/fennec/collections/experiments/records
--- a/components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsTest.kt
+++ b/components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsTest.kt
@@ -9,7 +9,10 @@ import android.content.SharedPreferences
 import android.content.pm.PackageInfo
 import android.content.pm.PackageManager
 import androidx.test.core.app.ApplicationProvider
+import androidx.work.testing.WorkManagerTestInitHelper
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
 import mozilla.components.support.test.any
 import org.junit.Assert.assertEquals
@@ -19,7 +22,6 @@ import org.junit.Assert.assertTrue
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.robolectric.RobolectricTestRunner
-import org.robolectric.RuntimeEnvironment
 import java.io.File
 import mozilla.components.service.glean.GleanInternalAPI
 import mozilla.components.support.test.eq
@@ -65,8 +67,12 @@ class ExperimentsTest {
         storage: FlatFileExperimentStorage = getDefaultExperimentStorageMock(),
         valuesProvider: ValuesProvider = object : ValuesProvider() {
             override fun getClientId(context: Context): String = EXAMPLE_CLIENT_ID
-        }
+        },
+        updateAsync: Boolean = false
     ) {
+        // Initialize WorkManager (early) for instrumentation tests.
+        WorkManagerTestInitHelper.initializeTestWorkManager(context)
+
         configuration = Configuration()
         experiments = spy(ExperimentsInternalAPI())
         experiments.valuesProvider = valuesProvider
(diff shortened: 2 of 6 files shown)


components/service/experiments/src/main/java/mozilla/components/service/experiments/ExperimentsUpdater.kt
```diff
@@ -84,3 +123,23 @@ internal class ExperimentsUpdater(
...
+    workerParameters: WorkerParameters,
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal var updater: ExperimentsUpdater = Experiments.updater
+) : Worker(context, workerParameters) {
+
+    override fun doWork(): Result {
+        return if (updater.updateExperiments()) {
+            Result.success()
+        } else {
+            Result.failure()
```
georgf: What happens on failure?
- Do we get automatic retry from work manager?
- Or do we need a follow-up bug for retry behavior?

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsUpdaterTest.kt
```diff
@@ -0,0 +1,90 @@
...
+    fun `ExperimentsUpdater initialize schedules ExperimentsUpdateWorker in WorkManager`() {
+        assertFalse("ExperimentsUpdateWorker should not be scheduled yet",
+            isWorkScheduled(ExperimentsUpdater.TAG))
+        experimentsUpdater.initialize(configuration)
+        assertTrue("initialize must schedule ExperimentsUpdateWorker",
+            isWorkScheduled(ExperimentsUpdater.TAG))
+    }
+
+    @Test
+    fun `ExperimentsUpdateWorker doWork returns SUCCESS`() {
```
georgf: What about test coverage for the failure case?

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsUpdaterTest.kt
```diff
@@ -0,0 +1,90 @@
...
+        resetExperiments()
+    }
+
+    @Test
+    fun `ExperimentsUpdater initialize schedules ExperimentsUpdateWorker in WorkManager`() {
+        assertFalse("ExperimentsUpdateWorker should not be scheduled yet",
+            isWorkScheduled(ExperimentsUpdater.TAG))
+        experimentsUpdater.initialize(configuration)
+        assertTrue("initialize must schedule ExperimentsUpdateWorker",
+            isWorkScheduled(ExperimentsUpdater.TAG))
```
georgf: Should we assert the right parameters being set?

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsTest.kt
```diff
@@ -77,14 +81,21 @@ class ExperimentsTest {
         `when`(experimentsUpdater.getExperimentSource(configuration)).thenReturn(experimentSource)
     }
 
+    private fun mockUpdaterWorkManagerFunctionality() {
```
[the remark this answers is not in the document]
travis79: I was nervous about running everything async, but I like the idea of adding _limited_ async testing since it feels like it would be easier to manage/control a test or two running async rather than everything.

components/service/experiments/src/main/java/mozilla/components/service/experiments/ExperimentsUpdater.kt
```diff
@@ -25,23 +29,55 @@ internal class ExperimentsUpdater(
...
     internal fun initialize(configuration: Configuration) {
         source = getExperimentSource(configuration)
 
-        updateJob = GlobalScope.launch(IO) {
-            // Update the experiments list from the server async, without blocking
-            // the app launch.
-            updateExperiments()
-        }
+        // Schedule the periodic experiment updates
+        scheduleUpdates()
```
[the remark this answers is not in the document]
travis79: No, it should fire immediately for the first execution when the `.enqueueUniquePeriodicWork` function is called.  This does appear to be the case from my manual testing.

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsTest.kt
```diff
@@ -77,14 +81,21 @@ class ExperimentsTest {
...
+        // Since WorkManager doesn't support working with robolectric, need to mock running the task
+        // but doing it synchronously to avoid a lot of headaches, to guarantee we don't need to
+        // await anything.
+        `when`(experimentsUpdater.scheduleUpdates()).then {
```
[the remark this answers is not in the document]
travis79: I believe it could, I will refactor and see.  There were several tests that didn't require it (i.e. they pass without it) which is why I split it out like this.

components/service/experiments/src/main/java/mozilla/components/service/experiments/ExperimentsUpdater.kt
```diff
@@ -25,23 +29,55 @@ internal class ExperimentsUpdater(
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
     internal lateinit var source: KintoExperimentSource
 
-    @Volatile private var updateJob: Job? = null
```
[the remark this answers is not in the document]
travis79: Per the previous comments, I will add async executions back to specific test(s) in order to test this.  Using WorkManager now, it may be that it has to be faked and awaited a little differently than before.

components/service/experiments/src/main/java/mozilla/components/service/experiments/ExperimentsUpdater.kt
```diff
@@ -84,3 +123,23 @@ internal class ExperimentsUpdater(
...
+
+/**
+ * This class represents a [Worker] object that periodically updates the experiments.
+ * @param updater used so that the updater can be injected from tests.
+ */
+internal class ExperimentsUpdaterWorker(
+    context: Context,
+    workerParameters: WorkerParameters,
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal var updater: ExperimentsUpdater = Experiments.updater
```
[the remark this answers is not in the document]
travis79: I didn't intend for it to get passed under normal circumstances, it should resort to the default of using the singleton's updater.  I added this in order to inject an `ExperimentsUpdater` from a test so that I could initialize it and manipulate it when unit testing the `ExperimentsUpdateWorker` class.

components/service/experiments/src/main/java/mozilla/components/service/experiments/ExperimentsUpdater.kt
travis79: No, this is an oversight on my part.  I _intended_ to return `Result.retry()` here.

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsUpdaterTest.kt
travis79: It's actually hard to test that bit because of the call chain.  Let me take a stab at mocking the updater function to return a false and see if I can get coverage for that line.

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsUpdaterTest.kt
travis79: We absolutely can, that just seemed more like testing WorkManager than testing the scheduler, but I guess just checking that it was scheduled is sort of doing the same thing.  I'll add asserts for the default parameters just to be thorough.

components/service/experiments/src/main/java/mozilla/components/service/experiments/ExperimentsUpdater.kt
travis79: Actually, what I think I was thinking is that `Result.failure()` would just give up this attempt and it would try again at the next periodic scheduled interval.

travis79: I was planning on setting up a manual test today using samples-glean to run the update on the minimum scheduled interval (15 minutes :( ) in order to make sure the periodic part was working and to measure the drift.  I'll try it with both `.failure()` and `.retry()` in order to see what the difference is, but I suspect that `retry` will retry immediately and `failure` will run at the next scheduled interval.

components/service/experiments/src/main/java/mozilla/components/service/experiments/ExperimentsUpdater.kt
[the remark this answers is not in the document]
travis79: It looks like since I've added this, my manual test in samples-glean has broken because of it.  Looks like I'll have to find another way around this...

components/service/experiments/src/main/java/mozilla/components/service/experiments/ExperimentsUpdater.kt
```diff
@@ -25,23 +29,55 @@ internal class ExperimentsUpdater(
...
 
+    /**
+     * Builds the periodic work request that will control the scheduling of the worker and provide
+     * the tag value so that the worker can be interacted with later.
+     *
+     * @return [PeriodicWorkRequest]
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun getWorkRequest(): PeriodicWorkRequest {
+        return PeriodicWorkRequest.Builder(
```
georgf: Per our conversation, this also can trigger when in the background.
This has some implications for how experiments run, we should file a follow-up bug and talk through that.

components/service/experiments/src/main/java/mozilla/components/service/experiments/ExperimentsUpdater.kt
georgf: We should retry on failure and probably quickly initially.
Maybe we can find out what the normandy behavior is and align with that?

travis79: Normandy client behavior is reported to be to fail and retry in 6 hours, with the possibility that this may change in the future.  Since this is easy to iterate on, I'm going to verify that returning `Result.failure()` does what we expect (i.e. matches current Normandy client behavior) and call this good for now.

components/service/experiments/src/main/java/mozilla/components/service/experiments/ExperimentsUpdater.kt
travis79: Follow up bug [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1544931)

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsUpdaterTest.kt
travis79: added tests to cover all the parameters I could get at.

travis79:
Review feedback has been addressed, ready for another look.

components/service/experiments/src/main/java/mozilla/components/service/experiments/ExperimentsUpdater.kt
georgf: Per IRC conversation, we currently don't guarantee immediate updates on app launch.

We can change this to use .REPLACE, then we get:
- An immediate update on app launch.
- A predictable schedule (it resets to update every 6h from app launch again)

travis79:
Final changes based on feedback around `KEEP` and `REPLACE` polices has been completed and pushed.  I also had to fix a couple of merge conflicts after the multi-branching PR landed, which are also included in the latest push.

components/service/experiments/src/main/java/mozilla/components/service/experiments/ExperimentsUpdater.kt
```diff
@@ -88,3 +133,21 @@ internal class ExperimentsUpdater(
         private const val EXPERIMENTS_COLLECTION_NAME = "mobile-experiments"
     }
 }
+
+/**
+ * This class represents a [Worker] object that periodically updates the experiments.
+ * @param updater used so that the updater can be injected from tests.
+ */
+internal class ExperimentsUpdaterWorker(
```
georgf: Is it a typical style to have this at the bottom?

From the desktop code and other projects i'm used to having helpers & utils before the main class.
That helps with actually discovering them, i wouldn't think of looking for other classes after this files main class.

Unless there is a good reason, let's maybe move this one up to before `class ExperimentsUpdater`?

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsTest.kt
```diff
@@ -80,16 +86,29 @@ class ExperimentsTest {
         `when`(experiments.getExperimentsUpdater(context)).thenReturn(experimentsUpdater)
         experimentSource = source
         `when`(experimentsUpdater.getExperimentSource(configuration)).thenReturn(experimentSource)
+
+        // Since WorkManager doesn't support working with robolectric, need to mock running the task
```
georgf: Super nit: "we need to"?

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsTest.kt
```diff
@@ -80,16 +86,29 @@ class ExperimentsTest {
...
+        // with a way to execute synchronously to avoid a lot of headaches, but also supporting
```
georgf: Super nit: "... headaches. For some test cases we also ..."?
The full sentence was hard to parse for me.

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsTest.kt
```diff
@@ -802,15 +793,31 @@ class ExperimentsTest {
...
         file.writer().use {
             it.write("""{"experiment":[""")
         }
 
         val experimentStorage = FlatFileExperimentStorage(file)
 
-        resetExperiments(experimentSource, experimentStorage)
+        // Because we interact with experimentsUpdater in resetExperiments in order to setup mocking
+        // of the update process, we need to use a scoped down version of the reset here to pass the
+        // verify zero interactions assertion
```
georgf: Can we use `verifyNoMoreInteractions()`?

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsTest.kt
```diff
@@ -99,7 +118,24 @@ class ExperimentsTest {
 
     @Test
     fun `initialize experiments before glean`() {
-        resetExperiments()
+        // Because we interact with experimentsUpdater in resetExperiments in order to setup mocking
+        // of the update process, we need to use a scoped down version of the reset here to pass the
+        // verify zero interactions assertion
```
georgf: Can't we use `verifyNoMoreInteractions()` here instead?

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsUpdaterTest.kt
```diff
@@ -0,0 +1,138 @@
...
+        val workInfo = getWorkInfoByTag(ExperimentsUpdater.TAG)
+        assertNotNull("workInfo must not be null", workInfo)
+        assertEquals("workInfo id must match request id", workRequest.id, workInfo?.id)
+    }
+
+    @Test
+    fun `ExperimentsUpdateWorker doWork returns SUCCESS`() {
+        experiments.initialize(context, configuration)
+        experimentsUpdater.initialize(configuration)
+        // Replace the updater in the singleton with the one built for the test
```
georgf: Super nit: A few of the comments in this file (and possibly elsewhere?) are missing trailing ".".

components/service/experiments/src/main/java/mozilla/components/service/experiments/ExperimentsUpdater.kt
travis79: It's fairly common in mobile to have extensions and stuff at the end of the file, but I see your point.  Maybe it would be _best_ if I just put it in it's very own file?

travis79: Nevermind, I remember now why I did it in the same file...  due to visibility warnings/issues of how Kotlin treats `private` as rather `file private` than `class private`.  The worker needed access to call the updateExperiments() function.   I'll just move it to the top as you suggested.

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsTest.kt
travis79: No, it turns out I couldn't get it to work with `verifyNoMoreInteractions` as there the interaction appeared to happening in a way to make this fail too.  I can test for the 1 interaction in resetExperiments that I know about, so I changed it to check for that and no more interactions.

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsTest.kt
travis79: I cannot use `verifyNoMoreInteractions` here, this comment was relevant to the other place I had to `resetExperiments` minus the updater mocking and I just didn't update it to reflect the use-case.  I updated the comment to try and explain what I was seeing and why.

PR #2751, opened 2019-04-16
Closes #2623: Add customizable UI for ReaderView

### LINKED ISSUE #2623: Create customizable ReaderView appearance UI
This is to control the font size, font type and color scheme. The JavaScript and Kotlin APIs are available now.

At least the colors and fonts of this UI need to be customizable, but the general layout should be:

Screen Shot 2019-04-03 at 14 36 39

The UI should disappear when scrolling or clicking/tapping on the article.

jonalmeida:
### Pull Request checklist


components/feature/readerview/src/main/java/mozilla/components/feature/readerview/ReaderViewFeature.kt
```diff
@@ -35,10 +38,13 @@ class ReaderViewFeature(
     private val context: Context,
     private val engine: Engine,
     private val sessionManager: SessionManager,
+    controlsView: ReaderViewControlsView,
     private val onReaderViewAvailableChange: OnReaderViewAvailableChange = { }
 ) : SelectionAwareSessionObserver(sessionManager), LifecycleAwareFeature, BackHandler {
 
     private val config = Config(context.getSharedPreferences("mozac_feature_reader_view", Context.MODE_PRIVATE))
+    internal val presenter = ReaderViewControlsPresenter(controlsView, config)
+    internal val interactor = ReaderViewControlsInteractor(controlsView, config)
```
csadilek: nit: maybe we should rename these to `controlsPresenter` and `controlsInteractor` as well, since they are only for the controls UI?

components/feature/readerview/src/main/java/mozilla/components/feature/readerview/view/ReaderViewControlsBar.kt
```diff
@@ -0,0 +1,152 @@
...
+            val colorSchemeChoice = when (checkedId) {
+                R.id.mozac_feature_readerview_color_dark -> ColorScheme.DARK
+                R.id.mozac_feature_readerview_color_sepia -> ColorScheme.SEPIA
+                R.id.mozac_feature_readerview_color_light -> ColorScheme.LIGHT
+                else -> ColorScheme.DARK
+            }
+            listener?.onColorSchemeChanged(colorSchemeChoice)
+        }
+        fontIncrementButton = applyClickListener(R.id.mozac_feature_readerview_font_size_increase) {
+            listener?.onFontSizeChanged(FontSize.INCREASE)
```
csadilek: nit: don't think we really need the FontSize enum anymore. The listener could just have `onIncreaseFontSize` and `onDecreaseFontSize`.

jonalmeida:
> I am not sure if we should try and set a theme in the sample app right away to make sure we can customize the colors / controls, but we can do that all in a follow-up.

I'll file a follow up. That's a lot more trivial to add, and merging this in let's us work async when the messaging API lands.

My Android styling/theming foo is weak so there's an element of trial-and-error.

components/feature/readerview/src/main/java/mozilla/components/feature/readerview/view/ReaderViewControlsBar.kt
jonalmeida: Ha, good point!

PR #2752, opened 2019-04-16

Pike:
State: mozilla-l10n/android-l10n@ce0bbdb9470479dba0f3775414204eb6d547a884

### Pull Request checklist


pocmo:
Android Lint is complaining about:

support-ktx:

```
  /build/android-components/components/support/ktx/src/main/res/values-ko/strings.xml:3: Error: Replace "..." with ellipsis character (…, &#8230;) ? [TypographyEllipsis]
      <string name="mozac_support_ktx_menu_share_with">공유...</string>
```

feature-findinpage:

```
  /build/android-components/components/feature/findinpage/src/main/res/values-es/strings.xml:22: Error: "pagina" is a common misspelling; did you mean "página" ? [Typos]
      <string name="mozac_feature_findindpage_dismiss">Descartar buscar en pagina</string>
```

@Pike What do you think? The `TypographyEllipsis` check may be something we could add to `compare-locales` if we agree with it? Regarding the `Typos` check: 🤷‍♂️

Pike:
I'm generally curious on how to get these results. The ellipsis one is a bit tough, there's some data on cldr that suggests that it's OK for strings to have '...' (three dots) in some languages, http://www.unicode.org/cldr/charts/32/by_type/miscellaneous.linguistic_elements.html#5d9dd084fd590ee, notably for American Spanish. So also in a locale we care about.

The typos one is interesting, I wonder if we can upstream that into pontoon. I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1544695 on that.

That said, I don't think either of those checks has enough bang to block us from moving forward. Is there a way to get visibility into those and still be able to merge the string landings?

CC @callek, @Delphine, @flodolo

FWIW, the concrete cases that triggered the issues here should be fixed now, I fixed those in pontoon and updated the export.

PR #2756, opened 2019-04-16
glean: 1526343: Add public sendPings API

mdboom:
### Pull Request checklist

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
@@ -346,11 +346,14 @@ open class GleanInternalAPI internal constructor () {
      * this function. If the name of a Glean-managed ping is passed in, an
      * error is logged to logcat.
      *
-     * Both the ping collection and ping uploading happens asyncronously.
+     * While the collection of metrics into pings happens synchronously, the
+     * ping queuing and ping uploading happens asyncronously.
+     * There are no guarantees that this will happen immediately.
+     *
      * If the ping currently contains no content, it will not be sent.
      *
      * @param pingNames List of pings to send.
-     * @return true if any pings were actually sent.
+     * @return true if any pings had content and were queued for uploading
      */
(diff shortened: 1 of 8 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -343,7 +374,7 @@ open class GleanInternalAPI internal constructor () {
      * @param pingNames List of pings to send.
      * @return true if any pings were actually sent.
      */
-    internal fun sendPings(pingNames: List<String>): Boolean {
+    internal fun sendPingsInternal(pingNames: List<String>): Boolean {
```
Dexterp37: nit: this is not strictly related to this patch, but if consumers will start attempting to send empty pings using the `sendPings` API they will receive no feedback if the payload is empty and nothing happens. I'd suggest adding a debug log line [here](https://github.com/mozilla-mobile/android-components/blob/0dfbad94ab623d922d1f9b4d41be6bc594e80f63/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt#L392) to just say "Nothing to collect for ping: `pingname`".

Dexterp37:
> 
> 
> I'd like to see this documented in the docs as part of this. I don't think it's worth leaving that as technical debt and would illustrate how we intend this to be used and point out what it's not supposed to do, like send glean internal pings.

Oh, great catch!

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
mdboom: That's not a bad idea.  This information is currently in the return value, too.  Do you think maybe that's not necessary?  None of the uses internally use the return value right now.

Dexterp37: Mh, I'm not sure if that's needed or not. I might have a use case for the `activation` ping, so let's keep it around for now :)

PR #2774, opened 2019-04-17
Add ExperimentsDebugActivity to service-experiments.

travis79:
This adds the first command, "updateExperiments" that will force the ExperimentsUpdater to update immediately.

### Pull Request checklist

### DIFF
--- a/components/service/experiments/src/main/java/mozilla/components/service/experiments/debug/ExperimentsDebugActivity.kt
+++ b/components/service/experiments/src/main/java/mozilla/components/service/experiments/debug/ExperimentsDebugActivity.kt
@@ -1,0 +1,46 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.experiments.debug
+
+import android.app.Activity
+import android.os.Bundle
+import android.support.annotation.VisibleForTesting
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.launch
+import mozilla.components.service.experiments.Experiments
+import mozilla.components.support.base.log.logger.Logger
+
+class ExperimentsDebugActivity : Activity() {
+    private val logger = Logger(LOG_TAG)
+
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    internal var updateJob: Job? = null
+
+    companion object {
+        private const val LOG_TAG = "ExperimentsDebugActivity"
+
+        // This is a list of the currently accepted commands
+        const val UPDATE_EXPERIMENTS_EXTRA_KEY = "updateExperiments"
+    }
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        logger.debug("onCreate")
+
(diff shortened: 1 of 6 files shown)


travis79:
@georgf This is ready for a first look.  It currently has only the command to force experiments updates implemented.  I'm fine with adding the other commands to this PR but I thought it would be good to start getting feedback on it early so that it would be less to review without the more complicated commands in place.

components/service/experiments/src/main/java/mozilla/components/service/experiments/debug/ExperimentsDebugActivity.kt
```diff
@@ -0,0 +1,39 @@
...
+package mozilla.components.service.experiments.debug
+
+import android.app.Activity
+import android.os.Bundle
+import android.support.annotation.VisibleForTesting
+import mozilla.components.service.experiments.Experiments
+import mozilla.components.support.base.log.logger.Logger
+
+class ExperimentsDebugActivity : Activity() {
+    private val logger = Logger(this.javaClass.canonicalName)
```
georgf: This doesn't quite match the logger setup elsewhere in this library.
Is this a better practice to have? If so let's have a follow-up bug on standardizing on this. If not let's use the `TAG` approach we use elsewhere.

components/service/experiments/src/main/java/mozilla/components/service/experiments/debug/ExperimentsDebugActivity.kt
```diff
@@ -0,0 +1,39 @@
...
+
+    companion object {
+        // This is a list of the currently accepted commands
+        const val UPDATE_EXPERIMENTS_EXTRA_KEY = "updateExperiments"
+    }
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        logger.debug("ExperimentsDebugActivity onCreate")
```
Dexterp37: nit: this doesn't need `EperimentsDEbugActivity` since this should be in the Logger name, as Georg suggests.

components/service/experiments/src/main/java/mozilla/components/service/experiments/debug/ExperimentsDebugActivity.kt
```diff
@@ -0,0 +1,39 @@
...
+import android.os.Bundle
+import android.support.annotation.VisibleForTesting
+import mozilla.components.service.experiments.Experiments
+import mozilla.components.support.base.log.logger.Logger
+
+class ExperimentsDebugActivity : Activity() {
+    private val logger = Logger(this.javaClass.canonicalName)
+
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    internal var testExperimentsUpdated = false
```
Dexterp37: nit: I don't personally like introducing these kind of variables to make the test work, but I have no concrete suggestion on how to make this better, so ... yeah :D

travis79: I didn't like it either, but I couldn't come up with a better way at the time either.

travis79:
@georgf and @Dexterp37, I believe I have addressed the changes you have requested and this is ready for another look.

components/service/experiments/src/main/java/mozilla/components/service/experiments/debug/ExperimentsDebugActivity.kt
travis79: I managed to come up with a better way to resolve this since I ran into a problem with the debug activity attempting to do networking on the main thread.

PR #2783, opened 2019-04-18
Closes #2771: Only clear Session.searchTerms after request initiated by the user.

### LINKED ISSUE #2771: Clear session searchTerm onLoadRequest
Currently we're clearing the searchTerms onLocationChange, which means that the session only knows about its search terms in the brief moment between onURLCommitted and onLocationChange.

Instead let's copy the implementation we used for focus where we cleared session searchTerms `onLoadRequest` **if and only if** the user interacted (remember, the boolean for this user interaction is incorrectly named on the GV side!)

pocmo:
### Pull Request checklist

### DIFF
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -302,7 +302,19 @@ class GeckoEngineSession(
                     is InterceptionResponse.Url -> loadUrl(url)
                 }
             }
-            return GeckoResult.fromValue(if (response != null) AllowOrDeny.DENY else AllowOrDeny.ALLOW)
+
+            return if (response != null) {
+                GeckoResult.fromValue(AllowOrDeny.DENY)
+            } else {
+                notifyObservers {
+                    // Unlike the name LoadRequest.isRedirect may imply this flag is not about http redirects. The flag
+                    // is "true if and only if the request was triggered by user interaction."
+                    // See: https://bugzilla.mozilla.org/show_bug.cgi?id=1545170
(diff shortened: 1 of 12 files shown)


components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -302,7 +302,14 @@ class GeckoEngineSession(
                     is InterceptionResponse.Url -> loadUrl(url)
                 }
             }
-            return GeckoResult.fromValue(if (response != null) AllowOrDeny.DENY else AllowOrDeny.ALLOW)
+
+            return if (response != null) {
+                GeckoResult.fromValue(AllowOrDeny.DENY)
+            } else {
+                notifyObservers { onLoadRequest(request.isRedirect) }
```
csadilek: @pocmo this part I don't understand. Why does `request.isRedirect` mean `triggeredByUserInteraction`.

pocmo: Yeah, I didn't understand this either. But I verified that that's what it actually does (and it says so in the javadoc) and that's how it's used in Focus too. And then filed this yesterday:
https://bugzilla.mozilla.org/show_bug.cgi?id=1545170

pocmo: I'll add this as a comment here (with link to the bug)

csadilek: 👍

csadilek:
@pocmo ah this one has a failing test now.

PR #2786, opened 2019-04-18
Make sure metric recording happens in the expected order

Dexterp37:
This changes glean internals so that recording right before a call to `sendPings` puts the data in that ping. In order to sanely allow this, `sendPings` now returns a `Deferred<Boolean>` which can be used to check if the ping was correctly taken care of.
This additionally introduces a function to wait for `WorkManager` jobs to be queued (in tests) and an `async` facility in Dispatchers.kt

**Expected behaviour**

- Callers should be able to "fire and forget" `sendPings`;
- Taking a snapshot of the recorded metrics, up to the call to `sendPings`, happens through `assembleAndSerializePing` which should be executed in the context of `Dispatchers.API` (so, off-the-main thread).
- Ping persistence on the disk is already happening through Kotlin's `Dispatchers.IO` (by `PingStorageEngine`). `sendPingsInternal` is async, so we're not blocking the caller with that, even though we call `joinAll`.

**Notes**

- We're now always printing a debug log message if there's no data for the ping, even for pings we own; this will really help for locally testing; moreover, the log line was not printing the ping name correctly.
- `Dispatchers.enableTestingMode` has been changed to `Dispatchers.setTestingMode` so that we don't have to access Dispatcher's internals for toggling this mode off.

### Pull Request checklist


Dexterp37:
@pocmo would you kindly take a sanity check look at `sendPingsInternalAsync`? I'm concerned, as usual, about my usage of coroutines and scopes. From manual runs, this seems to behave as expected. Tests don't complain. But I'd still like another set of eyes on this, if possible. It would be highly appreciated!

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -448,7 +446,7 @@ open class GleanInternalAPI internal constructor () {
     @VisibleForTesting(otherwise = VisibleForTesting.NONE)
     fun enableTestingMode() {
         @Suppress("EXPERIMENTAL_API_USAGE")
-        Dispatchers.API.enableTestingMode()
+        Dispatchers.API.setTestingMode(enabled = true)
```
mdboom: Would it be useful to provide a way for third-parties to also temporarily disable testing mode?

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -381,45 +385,39 @@ open class GleanInternalAPI internal constructor () {
      * @param pingNames List of pings to send.
      * @return true if any pings had content and were queued for uploading
      */
-    internal fun sendPingsInternal(pingNames: List<String>): Boolean {
+    internal fun sendPingsInternalAsync(pingNames: List<String>): Deferred<Boolean> {
```
travis79: I think I agree with @mdboom on the fact that we never use the result of this function.  If I remember correctly, it was originally used to determine whether or not to enqueue the Worker in WorkManager but a lot of how that works has changed since then.  Since we wouldn't have a return value anymore, much of the `Deferred` stuff could be done away with and all of the `@Suppress("DeferredResultUnused")` linter suppressions as well.  Unless you see a use-case for using the returned Boolean in some way?

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -355,7 +359,7 @@ open class GleanInternalAPI internal constructor () {
      * @param pingNames List of pings to send.
      * @return true if any pings had content and were queued for uploading
      */
-    public fun sendPings(pingNames: List<String>): Boolean {
+    public fun sendPings(pingNames: List<String>): Deferred<Boolean> {
```
travis79: `public` is redundant here since Kotlin uses it by default

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
travis79: If we really wanted to return something and add functionality to the public API, I think I'd much rather see a completion handler that ran after the pings were _actually_ uploaded, rather than returning a Boolean indicating that "yeah, I plan on sending some pings later...".  Otherwise, I'd vote to remove the return handler and Deferred stuff as it adds complexity.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
Dexterp37: I think it would: I filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1545759).

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
Dexterp37: Good point on removing the returned `Boolean`.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -379,47 +380,33 @@ open class GleanInternalAPI internal constructor () {
      * If the ping currently contains no content, it will not be sent.
      *
      * @param pingNames List of pings to send.
-     * @return true if any pings had content and were queued for uploading
      */
-    internal fun sendPingsInternal(pingNames: List<String>): Boolean {
+    internal fun sendPingsInternal(pingNames: List<String>) = Dispatchers.API.launch {
```
mdboom: TIL: That's quite an interesting bit of Kotlin shorthand.

PR #2792, opened 2019-04-18
glean 1529226: Provide a way to specify whether a ping should send client_id

mdboom:
Requires https://github.com/mozilla/glean_parser/pull/58 to be merged first.

### Pull Request checklist


components/service/glean/pings.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+$schema: moz://mozilla.org/schemas/glean/pings/1-0-0
+
+baseline:
+  description: >
+    This ping is intended to provide metrics that are managed by the library
+    itself, and not explicitly set by the application or included in the
+    application's `metrics.yaml` file.
+    The `baseline` ping is automatically sent when the application is moved to
+    the background.
```
mdboom: This information is duplicated from `docs/pings/baseline.md`.  We should probably decide how much info we want to include here vs. there, etc...

components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -161,6 +161,7 @@ ext.gleanGenerateMetricsAPI = {
         args "-s"
         args "namespace=$fullNamespace"
         args "$projectDir/metrics.yaml"
+        args "$projectDir/pings.yaml"
```
mdboom: We pass both of these files to `glean_parser`, even though only one or the other might exist.  `glean_parser` is smart enough not to care if one of them is missing.  (Though I guess it would be weird to have `pings` without `metrics`, having `metrics` without `pings` is completely legit.)

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -336,52 +329,24 @@ open class GleanInternalAPI internal constructor () {
      */
     fun handleBackgroundEvent() {
         // Schedule the baseline and event pings
-        sendPingsInternal(listOf(BASELINE_STORE_NAME, "events"))
+        sendPings(listOf(Pings.baseline, Pings.events))
```
mdboom: We could do `Pings.baseline.send()` here, but that will log a debug message if the ping is empty when it's queued.  That's the only difference between the "public" API and this one.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -336,52 +329,24 @@ open class GleanInternalAPI internal constructor () {
...
      * If the ping currently contains no content, it will not be sent.
      *
-     * @param pingNames List of pings to send.
+     * @param pings List of pings to send.
+     * @param includeClientInfoOverride If not-null, overrides the `includeClientInfo`
+     * value specified in the [PingType] objects in `pings`.
      * @return true if any pings had content and were queued for uploading
      */
-    internal fun sendPingsInternal(pingNames: List<String>): Boolean {
+    internal fun sendPings(pings: List<PingType>, includeClientInfoOverride: Boolean? = null): Boolean {
```
mdboom: This includes the optional `includeClientInfoOverride` parameter, since I started the implementation before seeing that @georgf said we could safely ignore that for now.  Let me know if it should be left in or removed for now.

components/service/glean/src/test/java/mozilla/components/service/glean/debug/GleanDebugActivityTest.kt
```diff
@@ -120,7 +120,7 @@ class GleanDebugActivityTest {
             category = "telemetry",
             lifetime = Lifetime.Application,
             name = "boolean_metric",
-            sendInPings = listOf("store1")
+            sendInPings = listOf("metrics")
```
mdboom: After this PR, sending completely arbitrary pings that don't have an entry in a `pings.yaml` is an error, so needed to rename these to a "known" ping.

components/service/glean/docs/pings/custom.md
```diff
@@ -3,9 +3,30 @@
...
+  [`client_info`](pings.md#The-client_info-section)) section.
+- `disabled` (optional): When `true`, sending of the ping is disabled.
+
+For example, to define a custom ping specifically for search information:
+
+```YAML
+search:
+  description: >
+    A ping to record search data.
+  include_client_info: false
```
Dexterp37: We don't want to exclude the full `client_info`, as this also contains informations about the app version, channel and other metadata which are not PII. As decided [here](https://docs.google.com/document/d/1dGiUmR63Kd1oqoL7eQzEana0gadPffvq00LEith3uCY/edit#), we only want to allow excluding the `client_id`.

components/service/glean/docs/pings/custom.md
```diff
@@ -32,9 +53,12 @@ type, you can add the special value `default` to `send_in_pings`:
...
-call this method at the schedule and cadence that is appropriate for the ping.
+To send a custom ping, call the `send` method on the `PingType` object that
+glean generated for your ping.
+
+For example, to send the custom ping defined above:
 
 ```kotlin
-Glean.sendPings(listOf("search"))
+import org.mozilla.yourApplication.GleanMetrics.Pings
+Pings.search.send()
```
Dexterp37: Does this buy us much more over the `Glean.sendPing` API? I'd avoid going through the hassle of generating new files (potential binary overhead?) to only wrap `Glean.sendPings(listOf("search"))`. Moreover, I think this is more of a product direction question that should involve @georgf (and also FYI @badboy )

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
Dexterp37: I think it's ok to print a message even though our pings are empty. I'm actually making this change in another PR: that would really, really help diagnosing if things go wrong with our pings as well.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
Dexterp37: Let's remove this for now, we decided to slightly change the `activation` ping, so this requirement doesn't need to be satisfied any more.

components/service/glean/docs/pings/custom.md
mdboom: We need metadata about whether a particular ping should contain `client_id` stored somewhere.  There are two cases for this:
- When sending events that have been queued from a previous run at app startup.  If we don't have the ping metadata, we don't know whether any events in a custom ping should include a client_id in that ping.
- When using the debug activity, so we don't require the tester to get that right.
The first pass at implementing this included `includeClientId` only as a parameter to `sendPings`, which wasn't good enough because it could support these use cases.

I think this will buy us a lot in the future as well as pings grow more options and functionality.

I wouldn't worry about binary overhead.  This should always be a very small number of small data classes.

Dexterp37: Good points. I think I'm hooked into this now :)

travis79: I like how this exposes the `.send` function through the API of the ping instead of the Glean singleton and doesn't rely on the user to know the String name of the ping.

components/service/glean/src/test/java/mozilla/components/service/glean/debug/GleanDebugActivityTest.kt
travis79: I actually like this, it's an extra protection against sending something invalid.

mdboom: Yeah.  It also makes it easier for new pings to require data review (you can just watch `pings.yaml`), which in turn will give the pipeline folks a heads up about new incoming docTypes.  Clients can't just start throwing a bunch of new stuff at the pipeline...

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -409,20 +371,47 @@ open class GleanInternalAPI internal constructor () {
...
+     * @param pingNames List of ping names to send.
+     * @return true if any pings had content and were queued for uploading
+     */
+    internal fun sendPingsByName(pingNames: List<String>): Job? {
+        val pings = pingNames.mapNotNull { pingName ->
+            PingType.pingRegistry.get(pingName)?.let {
+                it
+            } ?: run {
+                logger.error("Attempted to send unknown ping '$pingName'")
+                null
```
Dexterp37: When this happens, we'll have `null` in `pings`. Should we filter for `null`s again?

components/service/glean/src/main/java/mozilla/components/service/glean/private/PingType.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    companion object {
+        private val logger = Logger("glean/PingType")
+        internal val pingRegistry: MutableMap<String, PingType> = mutableMapOf()
+    }
+
+    /**
+     * Send the ping.
+     *
+     * While the collection of metrics into pings happens synchronously, the
```
Dexterp37: nit: mh, I'd rephrase this a bit. After the last change to `sendPings`, collection happens asynchronously as well. What about.. "The ping is assembled as soon as possible, but upload is not guaranteed to happen immediately as it depends on the upload policies."

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -336,52 +329,21 @@ open class GleanInternalAPI internal constructor () {
...
      *
      * While the collection of metrics into pings happens synchronously, the
      * ping queuing and ping uploading happens asyncronously.
      * There are no guarantees that this will happen immediately.
      *
-     *
      * If the ping currently contains no content, it will not be sent.
      *
-     * @param pingNames List of pings to send.
+     * @param pings List of pings to send.
```
Dexterp37: I was missing this before, can you please also add a `@returns` here? we're returning a `Job` anyway.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
mdboom: I don't think so, because of the use of [mapNotNull](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/map-not-null.html).

Dexterp37: Gah, sorry, I misread the docs there.

mdboom:
@pocmo: This is another PR that will need to be coordinated with a snapshot.

PR #2806, opened 2019-04-19
Implement ExperimentsDebugActivity command to switch between Kinto instances.

travis79:
Using the ExperimentsDebugActivity, implement the "setKintoInstance" command to switch between "dev", "staging", or "prod" instances. 

### Pull Request checklist


travis79:
@georgf I wanted to get this started, I know I'll be gone next week so feel free to run with this if you need it landed next week.

components/service/experiments/src/main/java/mozilla/components/service/experiments/debug/ExperimentsDebugActivity.kt
```diff
@@ -38,6 +41,21 @@ class ExperimentsDebugActivity : Activity() {
             }
         }
 
+        if (intent.hasExtra(SET_KINTO_INSTANCE_EXTRA_KEY)) {
+            val instance = intent.getStringExtra(SET_KINTO_INSTANCE_EXTRA_KEY)
+
+            val config: Configuration? = when (instance) {
+                "dev" -> Configuration(kintoEndpoint = ExperimentsUpdater.KINTO_ENDPOINT_DEV)
```
Dexterp37: Note that `Configuration()` will drop any custom value that was previously provided as part of the configuration (e.g. the http client), and fall back to the defaults. What you really want here is to do `Configuration.copy(kintoEndpoint = ...)`

georgf:
One note here: i think we'll need to persist the kinto instance choice between restarts.
Otherwise it will be hard to test behavior across restarts (e.g. start experiment, stop app, delete experiment on server, restart app, verify experiment is stopped).

components/service/experiments/src/main/java/mozilla/components/service/experiments/debug/ExperimentsDebugActivity.kt
travis79: Thanks, I will fix this as I get back to work on this PR!

travis79:
> One note here: i think we'll need to persist the kinto instance choice between restarts.
> Otherwise it will be hard to test behavior across restarts (e.g. start experiment, stop app, delete experiment on server, restart app, verify experiment is stopped).

That makes sense, I'll add persistence to it, but we may need to handle it a little differently than the other configuration items then.

travis79:
On second thought, in thinking about persistence...   do we really need to persist this when the app can just be restarted using the same kinto instance?  It seems to me like this would make sense when being run from a test as this is intended to be.  If we allow for persistence, then it seems possible to accidentally override the kinto instance to something unintended if someone didn't reset it back to the proper one.  I guess what I'm trying to say is, I would rather have the extra step be in launching the app to get it changed rather than trying to remember to reset the instance back to production every time.

georgf:
> On second thought, in thinking about persistence... do we really need to persist this when the app can just be restarted using the debug activity to set it to the correct instance? It seems to me like this would make sense when being run from a test as this is intended to be. If we allow for persistence, then it seems possible to accidentally override the kinto instance to something unintended if someone didn't reset it back to the proper one. I guess what I'm trying to say is, I would rather have the extra step be in launching the app to get it changed rather than trying to remember to reset the instance back to production every time.

That's a good point. For now, let's simply require launching the app with the debug command to change kinto endpoints.
We can iterate later on this if needed.

components/service/experiments/README.md
```diff
@@ -73,12 +73,15 @@ In the above:
     |key|type|description|
     |---|----|-----------|
     | updateExperiments | boolean (--ez) | forces the experiments updater to run and fetch experiments immediately |
+    | setKintoInstance | string (--es) | sets the Kinto instance to the specified instance ("dev", "staging", "prod" only) |
 
-For example, to direct a release build of the Glean sample application to update experiments immediately, the following command
-can be used:
+For example, to direct a release build of the glean sample application to (1) update experiments immediately and
```
Dexterp37: nit: we settled on `Glean` rather than `glean` :)

travis79: oops ;)

travis79: merged rolled right over it...

PR #2818, opened 2019-04-21
Closes #2817: QR Feature cannot scan with custom containerViewId

### LINKED ISSUE #2817: QR Feature cannot scan with custom containerViewId after Permission granted
## Steps to reproduce
1. In a client app like Fenix, launch the QR feature scan(id: Int) method without CAMERA permissions granted.
2. Grant permissions when prompted.

### Expected behavior
The scan fragment appears after granting permissions.

### Actual behavior
The scan fragment does not appear because the QR feature uses the default view ID to scan after permissions are granted.

### Device information

* Android device: Pixel 3XL running Android Q
* Fenix version: Latest master 1.0.1917

colintheshots:
### Pull Request checklist


colintheshots:
I think you're right that that's the actual issue. It looked like the QrFeature would require the correct view ID, but that was not the cause of the problem. Closing.

PR #2835, opened 2019-04-24
Fixes 2812 - Unify string url detection to use Fennec patterns.

### LINKED ISSUE #2812: Re-evaluate url parsing regex
We thought #2790 would be good enough to pass our url detection regex but this doesn't seem to be the case. We should re-evaluate (ASAP) how we can update our regex check and/or revert to our old version if we need more time for this.

The [gruber v2 or stephenhay regex](https://mathiasbynens.be/demo/url-regex) might be good enough for us to use for now.

jhugman:
Fixes #2812 Re-evaluate url parsing regex 

This PR unifies string parsing of URL like strings into a `URLStringUtils` class, which uses the Fennec regular expressions for same. 

It also deprecates (but doesn't remove) string extension methods.

### Pull Request checklist

- ~[ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features~

### DIFF
--- a/components/support/utils/src/test/java/mozilla/components/support/utils/URLStringUtilsTest.kt
+++ b/components/support/utils/src/test/java/mozilla/components/support/utils/URLStringUtilsTest.kt
@@ -1,0 +1,76 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.support.utils
+
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
+import org.junit.Assert.assertNotEquals
+import org.junit.Assert.assertFalse
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+import mozilla.components.support.utils.URLStringUtils.isURLLike
+import mozilla.components.support.utils.URLStringUtils.toNormalizedURL
+import mozilla.components.support.utils.URLStringUtils.isSearchTerm
+
+@RunWith(RobolectricTestRunner::class)
+class URLStringUtilsTest {
+
+    @Test
+    fun toNormalizedURL() {
+        val expectedUrl = "http://mozilla.org"
+        assertEquals(expectedUrl, toNormalizedURL("http://mozilla.org"))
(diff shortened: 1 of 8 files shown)


jonalmeida:
I can see a couple usages for `isUrl` in Fenix as well, so we may want to put up a PR or notify them that this is going to break once it's merged.

components/support/utils/src/test/java/mozilla/components/support/utils/URLStringUtilsTest.kt
```diff
@@ -0,0 +1,71 @@
...
+        assertTrue(isURLLike("https://mozilla.org"))
+        assertTrue(isURLLike("file://somefile.txt"))
+
+        assertTrue(isURLLike("http://mozilla"))
+        assertTrue(isURLLike("http://192.168.255.255"))
+        assertTrue(isURLLike("192.167.255.255"))
+        assertTrue(isURLLike("about:crashcontent"))
+        assertTrue(isURLLike(" about:crashcontent "))
+        assertTrue(isURLLike("sample:about "))
+
```
jonalmeida: Can we add one more assertion for URLs with hypens that we didn't catch in our tests from the reported Fenix bug: https://github.com/mozilla-mobile/fenix/issues/1821

`https://mozilla-mobile.org` or something.

components/support/utils/src/test/java/mozilla/components/support/utils/URLStringUtilsTest.kt
```diff
@@ -0,0 +1,71 @@
...
+        assertTrue(isURLLike("http://mozilla.org"))
+        assertTrue(isURLLike("https://mozilla.org"))
+        assertTrue(isURLLike("file://somefile.txt"))
+
+        assertTrue(isURLLike("http://mozilla"))
+        assertTrue(isURLLike("http://192.168.255.255"))
+        assertTrue(isURLLike("192.167.255.255"))
+        assertTrue(isURLLike("about:crashcontent"))
+        assertTrue(isURLLike(" about:crashcontent "))
+        assertTrue(isURLLike("sample:about "))
```
jonalmeida: Can we add one more assertion for URLs with hypens that we didn't catch in our tests from the reported Fenix bug: https://github.com/mozilla-mobile/fenix/issues/1821

`https://mozilla-mobile.org` or something.

Issue #2624, opened 2019-04-03
ReaderView: Wire up messaging between feature and web extension


csadilek:
This depends on #2243.

PR #2836, opened 2019-04-24
Closes #2243: Add engine API for WebExtension messaging

### LINKED ISSUE #2243: Add engine API for WebExtension messaging

csadilek:
This brings in the web extension messaging functionality (+ tests and Kdoc). 

I want to commit a working example with this so I added a simply bidirectional message exchange to our stub reader view feature. Once this PR lands, work on reader view will continue in https://github.com/mozilla-mobile/android-components/issues/2624 and these simple example messages will replaced with the actual reader view messages.

As we now need the engine to return a `WebExtension` instance (abstracting over the concrete engine-specific implementation), there is a slight API change to install extensions by `id` and `url` and pass along the installed web extension in `onSuccess`.

### DIFF
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
@@ -6,6 +6,7 @@ package mozilla.components.browser.engine.gecko
 
 import android.content.Context
 import android.util.AttributeSet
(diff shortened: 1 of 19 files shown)


ligi:
Thanks so much for this PR! 
I am a bit puzzled on this part of the documentation though:

https://github.com/mozilla-mobile/android-components/pull/2836/files#diff-343723a0cb380f664aedf0dfed5c1c2fR42

PR #2838, opened 2019-04-24
1513616: Clear metrics when telemetry is disabled.

mdboom:
This follows the requirements @georgf listed in the bug:

The expectation on Desktop is, that on disabling Telemetry (code):

-    That we set the client id to a known value.
-    That we discard all pending pings.
-    We stop all uploading of pings.
-    That we reset all metrics (Desktop only resets some actually?). (Some exceptions might apply for Glean, like not resetting say install date?)
-    That we send a last optout ping, which contains no client id  **this is not implemented in this PR -- saving for a follow-up**

On re-enabling Telemetry, we:

-    Set the client id to a new UUID (this i think is a policy requirement).
-    Start freshly recording metrics again (implementation dependent: either reset one more time or stop discarding recorded values).

### Pull Request checklist


components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -140,6 +157,32 @@ open class GleanInternalAPI internal constructor () {
         return uploadEnabled
     }
 
+    /**
+     * Clear any pending metrics when telemetry is disabled.
+     */
+    fun clearMetrics() = Dispatchers.API.launch {
```
Dexterp37: Let's make this private: we don't want it to be part of the public api surface

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -140,6 +157,32 @@ open class GleanInternalAPI internal constructor () {
...
+        val existingFirstRunDate = DatetimesStorageEngine.getSnapshot(
+            firstRunDateMetric.sendInPings[0], false)?.get(firstRunDateMetric.identifier)?.let {
+            parseISOTimeString(it)
+        }
+
+        pingStorageEngine.clearPendingPings()
+        storageEngineManager.clearAllStores()
+
+        // Store a "dummy" KNOWN_CLIENT_ID in clientId.  This will make it easier to detect pings
+        // that were sent unintentionally when uploading is disabled.
```
Dexterp37: nit: can we slightly rephrase this to "This will make it easier to detect if pings are unintentionally sent after upload is disabled"

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -140,6 +157,32 @@ open class GleanInternalAPI internal constructor () {
...
+        // firstRunDate. Here, we store its value so we can restore it after clearing
+        // the metrics.
+        val firstRunDateMetric = GleanInternalMetrics.firstRunDate
+        val existingFirstRunDate = DatetimesStorageEngine.getSnapshot(
+            firstRunDateMetric.sendInPings[0], false)?.get(firstRunDateMetric.identifier)?.let {
+            parseISOTimeString(it)
+        }
+
+        pingStorageEngine.clearPendingPings()
+        storageEngineManager.clearAllStores()
```
Dexterp37: Please note that this doesn't clear experiments: https://github.com/mozilla-mobile/android-components/blob/81e5463966ed6115c4707350059fa859574ecbbf/components/service/glean/src/main/java/mozilla/components/service/glean/storages/StorageEngineManager.kt#L26

docs/changelog.md
```diff
@@ -29,6 +29,8 @@ permalink: /changelog/
      information.
    * A method for sending metrics on custom pings has been added. See
      `docs/pings/custom.md` for more information.
+   * Disabling telemetry through `setUploadEnabled` now clears all metrics
```
Dexterp37: We should make this behaviour explicit in the docs

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -130,7 +137,17 @@ open class GleanInternalAPI internal constructor () {
      */
```
Dexterp37: Let's update the function docs here to mention the new behaviour.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
mdboom: Good catch.  Do we want to clear experiments?  I'm a bit concerned about weird interactions with the experiments library if we do, particularly around *restoring* the right state if we turn uploading back on.

mdboom: Cc: @georgf ?

Dexterp37: I *think* that it shouldn't matter: this is only tagging for glean pings. If we turn upload off, then experiments or not, we don't send anything.

docs/changelog.md
```diff
@@ -29,6 +29,10 @@ permalink: /changelog/
...
+     immediately.
+   * ⚠️ **This is a breaking API change**: `Glean.setUploadEnabled` now requires
+     an application context to be passed in.
```
mdboom: This is unfortunate, but I haven't been able to find a better solution.  The release linter says we can't store a reference to the applicationContext in the Glean singleton since it would leak the application (which is what I did in the initial version of this PR).

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
mdboom: I think I'll add a comment for now saying that this doesn't clear experiments on purpose.

docs/changelog.md
Dexterp37: This is ok, that's what we do in other singletons as well. The `applicationContext` is a special context that lives throughout the whole application, as our `Glean` object. They have the same lifetime, so we don't need to care :)

components/service/glean/README.md
```diff
@@ -79,7 +79,7 @@ class SampleApplication : Application() {
         // Call setUploadEnabled first, since Glean.initialize
         // might send pings if there are any metrics queued up
         // from a previous run.
-        Glean.setUploadEnabled(Settings.isTelemetryEnabled)
+        Glean.setUploadEnabled(Settings.isTelemetryEnabled, applicationContext)
```
Dexterp37: I think it's ok to revert back to the previous version of the API: we store the applicationContext in other singletons. See [here](https://github.com/mozilla-mobile/android-components/blob/01837eee6cb00b698daafe96636f41c45cd88b34/components/service/glean/src/main/java/mozilla/components/service/glean/storages/UuidsStorageEngine.kt#L17-L21).

components/service/glean/README.md
```diff
@@ -102,6 +102,14 @@ object parameter of the `Glean.initialize` method. For example:
 Glean.initialize(applicationContext, Configuration(channel = "beta"))
 ```
 
+### Enabling and disabling metrics
+
+`Glean.setUploadEnabled()` should be called in response to the user enabling or
+disabling telemetry.
+
+When going from enabled to disabled, all pending events, metrics and pings are
+cleared. When re-enabling, core glean metrics will be recomputed at that time.
```
Dexterp37: nit: Glean

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -128,14 +138,29 @@ open class GleanInternalAPI internal constructor () {
      *
      * Metric collection is enabled by default.
      *
-     * When disabled, metrics aren't recorded at all and no data
+     * When uploading is disabled, metrics aren't recorded at all and no data
      * is uploaded.
      *
+     * When disabling, all pending metrics, events and queued pings are cleared.
+     *
+     * When enabling, the core glean metrics are recreated.
```
Dexterp37: nit: Glean

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -145,6 +170,38 @@ open class GleanInternalAPI internal constructor () {
...
+        // There is only one metric that we want to survive after clearing all metrics:
+        // firstRunDate. Here, we store its value so we can restore it after clearing
+        // the metrics.
+        val firstRunDateMetric = GleanInternalMetrics.firstRunDate
+        val existingFirstRunDate = DatetimesStorageEngine.getSnapshot(
+            firstRunDateMetric.sendInPings[0], false)?.get(firstRunDateMetric.identifier)?.let {
+            parseISOTimeString(it)
+        }
+
+        pingStorageEngine.clearPendingPings()
```
Dexterp37: It just occurred to me that we also need to cancel the workmanager upload job, probably before clearing the ping storage. What do you think, @mdboom ?

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -62,6 +64,12 @@ open class GleanInternalAPI internal constructor () {
 
     internal lateinit var pingStorageEngine: PingStorageEngine
 
+    companion object {
+        internal val KNOWN_CLIENT_ID = UUID.nameUUIDFromBytes(
+            "c0ffeec0-ffee-c0ff-eec0-ffeec0ffeec0".toByteArray()
```
badboy: (Drive-by comment) oh noes. 

(* good job, let's make sure we never see this)

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
mdboom: That's a good point.  

(1) There's probably a slight chance of a race condition that pings will get sent right before we do this (but sending and deleting pings is at least synchronized so hopefully that doesn't cause an I/O error reading a file that was just deleted etc.).

(2) The WorkManager will keep waking up, finding no pings, and doing nothing.  Which is suboptimal, but doesn't violate the privacy expectations, at least.

Anyway, it's worth figuring out how to turn on/off the workmanager as part of the `uploadEnabled` flag.

> 
> That's a good point.
> 
> (1) There's probably a slight chance of a race condition that pings will get sent right before we do this (but sending and deleting pings is at least synchronized so hopefully that doesn't cause an I/O error reading a file that was just deleted etc.).

Yes. I think we could still live with the I/O error, if it happens, in this case.

> (2) The WorkManager will keep waking up, finding no pings, and doing nothing. Which is suboptimal, but doesn't violate the privacy expectations, at least.

Would it? We bail out attempting to send pings if upload is disabled. [See here](https://github.com/mozilla-mobile/android-components/blob/786e1a30d32e3b89c54f057523cd8b1616c8c50d/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt#L358).

> Anyway, it's worth figuring out how to turn on/off the workmanager as part of the `uploadEnabled` flag.

I agree. I think this could be some method on the PingWorker that [cancels the worker](https://developer.android.com/topic/libraries/architecture/workmanager/how-to/cancel-stop-work#stopping).

Dexterp37: @mdboom I'm also happy to take this to a follow-up PR, doesn't need to block this.

mdboom: It was easy enough to add, but I'm a little stumped by how to test it.

Dexterp37: What about, for now, simply asserting `Assert.assertFalse("A scheduled PingUploadWorker must exist", isWorkScheduled(PingUploadWorker.PING_WORKER_TAG))`

mdboom: That makes sense.

mdboom: This is turning into a confusing mess of async WorkManager stuff and NPEs, so I'm going to pass on this for now and make a new bug to deal with this down the road.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -137,14 +147,29 @@ open class GleanInternalAPI internal constructor () {
...
      */
     fun setUploadEnabled(enabled: Boolean) {
         logger.info("Metrics enabled: $enabled")
+        val origUploadEnabled = uploadEnabled
+        // Change the value of uploadEnabled before initializeCoreMetrics so
+        // that the core metrics are actually recorded in the stores.
         uploadEnabled = enabled
+        if (isInitialized()) {
+            if (origUploadEnabled == true && enabled == false) {
+                clearMetrics()
```
Dexterp37: @mdboom Mh, hang on. I just realized we might inadvertently delete all the stored data when not wanted. We encourage calling `setUploadEnabled` *before* the init. Could this happen to remove all the local data (including the client id) if things are flipped at that time?

mdboom: If we call `setUploadEnabled` before `init`, nothing happens other than setting the flag.  This is in part because if it goes from `false` to `true`, it would try to regenerate the application metrics, but since Glean is uninitialized it wouldn't do anything.  And in general, this guard is there because doing anything too heavy with Glean before it's initialized is probably undefined behavior and full of NPEs etc.  So nothing will ever happen on startup if (as we encourage) `setUploadEnabled` is called before `initialize`.

I think there's potential for confusion here, though, since the enabled state isn't persisted.  It always starts at `true` before calling `setUploadEnabled`, but it might actually have already been set to `false` in a previous run.  I'm trying to work through whether that really matters in practice -- I think not...

mdboom: This has been updated so that `initialize` always handles the state switch once upon startup, and added a couple of tests around startup.  I think it's sorted now so we should always be in a correct state.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -154,6 +172,55 @@ open class GleanInternalAPI internal constructor () {
...
+
+    /**
+     * Clear any pending metrics when telemetry is disabled.
+     */
+    @Suppress("EXPERIMENTAL_API_USAGE")
+    private fun clearMetrics() = Dispatchers.API.launch {
+        // There is only one metric that we want to survive after clearing all metrics:
+        // firstRunDate. Here, we store its value so we can restore it after clearing
+        // the metrics.
+        val firstRunDateMetric = GleanInternalMetrics.firstRunDate
```
Dexterp37: Note: this is not clearing sequence numbers, as we discussed yesterday

mdboom:
This has been updated to clear sequence numbers when disabling uploading.  (In addition to the refactoring to avoid "invalid" states during startup).  This should now be ready for another review.

Cc: @georgf -- @Dexterp37 called you out explicitly as probably being interested in this one.

Dexterp37:
@mdboom this is a bit weird? Looks like there's too many commits on this PR?

mdboom:
@Dexterp37 : Sorry, I blame it on jetlag :)   Should be fixed now.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
mdboom: It should be now.

docs/changelog.md
```diff
@@ -100,6 +100,8 @@ permalink: /changelog/
    * ⚠️ **This is a breaking API change**: Custom pings must be explicitly
      registered with Glean at startup time. See
      `components/service/glean/docs/pings/custom.md` for more information.
+   * Disabling telemetry through `setUploadEnabled` now clears all metrics
```
Dexterp37: Let's call out that this includes the 'client_id' but not the 'first_run_date'

components/service/glean/src/test/java/mozilla/components/service/glean/private/EventMetricTypeTest.kt
```diff
@@ -197,11 +202,14 @@ class EventMetricTypeTest {
         Glean.setUploadEnabled(false)
         assertEquals(false, Glean.getUploadEnabled())
         eventMetric.record(mapOf(testNameKeys.testName to "event2"))
-        val snapshot2 = eventMetric.testGetValue()
-        assertEquals(1, snapshot2.size)
+        try {
```
Dexterp37: Let's use the assertion with exception [utility](https://kotlinlang.org/api/latest/kotlin.test/kotlin.test/assert-fails-with.html) here.

mdboom: Good catch.  Done.

docs/changelog.md
mdboom: Done.  Also had to move this to a new section of the changelog...

components/service/glean/src/test/java/mozilla/components/service/glean/private/EventMetricTypeTest.kt
mdboom: That library doesn't seem to be available to us in `a-c`, unfortunately.

Dexterp37: Ouch, no problem then!

mdboom:
Comment: Add a comment about startup / generating application lifetime metrics and how that works here now.

components/service/glean/README.md
```diff
@@ -109,6 +109,15 @@ object parameter of the `Glean.initialize` method. For example:
...
 ```
 
+### Enabling and disabling metrics
+
+`Glean.setUploadEnabled()` should be called in response to the user enabling or
+disabling telemetry.  This method should also be called at least once prior
+to calling `Glean.initialized()`.
+
+When going from enabled to disabled, all pending events, metrics and pings are
+cleared. When re-enabling, core Glean metrics will be recomputed at that time.
```
badboy: Do we need to mention the one metric we keep (first run date)?

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -154,6 +172,56 @@ open class GleanInternalAPI internal constructor () {
         return uploadEnabled
     }
 
+    /**
+     * Handles the changing of state when uploadEnabled changes.
```
badboy: I wonder if we should note here that it should only be called if the flag actually changes.

(I realize this is internal stuff, I'm just thinking about future-us having to deal with it)

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -154,6 +172,56 @@ open class GleanInternalAPI internal constructor () {
...
+     *
+     * When disabling, all pending metrics, events and queued pings are cleared.
+     *
+     * When enabling, the core Glean metrics are recreated.
+     *
+     * If Glean is not initialized, this is a no-op.
```
badboy: That is not true. 
Neither `initializeCoreMetrics` nor `clearMetrics` have this check.
It's only call site (for now) has the check.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -154,6 +172,56 @@ open class GleanInternalAPI internal constructor () {
...
+        } else {
+            clearMetrics()
+        }
+    }
+
+    /**
+     * Clear any pending metrics when telemetry is disabled.
+     */
+    @Suppress("EXPERIMENTAL_API_USAGE")
+    private fun clearMetrics() = Dispatchers.API.launch {
```
badboy: Personally I find this style of "inlining" the Dispatcher call a bit weird.
IMO it hides that it is async.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -283,13 +351,21 @@ open class GleanInternalAPI internal constructor () {
...
+        if (existingClientId == null || existingClientId == KNOWN_CLIENT_ID) {
             val uuid = UUID.randomUUID()
-            UuidsStorageEngine.record(GleanInternalMetrics.clientId, uuid)
-            DatetimesStorageEngine.set(GleanInternalMetrics.firstRunDate)
-        } else {
-            fixLegacyPingInfoMetrics()
+            UuidsStorageEngine.record(clientIdMetric, uuid)
+        }
+
+        val firstRunDateMetric = GleanInternalMetrics.firstRunDate
```
badboy: Hah, so no more `FileFirstRunDetector`. I like that.

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
```diff
@@ -513,4 +501,112 @@ class GleanTest {
...
+        assertFalse(stringMetric.testHasValue())
+        stringMetric.set("TEST VALUE")
+        assertTrue(stringMetric.testHasValue())
+    }
+
+    @Test
+    fun `Core metrics should be cleared and restored when disabling and enabling uploading`() {
+        assertTrue(GleanInternalMetrics.os.testHasValue())
+
+        Glean.pingMaker.getPingSeq("custom")
```
badboy: Should we test that this is a non-zero value?

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
```diff
@@ -513,4 +501,112 @@ class GleanTest {
...
+
+        Glean.setUploadEnabled(false)
+        assertEquals(0, Glean.pingStorageEngine.testGetNumPendingPings())
+
+        Glean.setUploadEnabled(true)
+        assertEquals(0, Glean.pingStorageEngine.testGetNumPendingPings())
+    }
+
+    @Test
+    fun `clientId is set to known id when setting disabled from a cold start`() {
```
badboy: Can we move the `clientId is managed correctly when disabling and enabling metrics` test closer to these 2? They are testing related things, so having them close by makes sense IMO.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
mdboom: Yeah, I think we just remove this comment, then.  It's a private method, and all callers of this method ensure that invariant.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
mdboom: Sure, it takes some getting used to.  I'd suggest moving that into its own bug, though, since this pattern is used in a number of places already.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
mdboom: Yep.  :bomb:

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
mdboom: Could be zero the first time...

mdboom: Maybe we should send a ping first, then we could do that.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
badboy: Ah yeah, no need to handle it here.

mdboom:
@Dexterp37 
> @mdboom what about the internal document (markdown, in-tree) that highlights the expected behaviour of this? (for @georgf to double check as well)

There are docs [here](https://github.com/mozilla-mobile/android-components/pull/2838/files#diff-604f80987fbf3998dcc3cc21cb5dd8afR112), or did you mean somewhere else?

components/service/glean/docs/internal.md
```diff
@@ -0,0 +1,25 @@
+# Internal behavior
+
+## Clearing metrics when disabling/enabling Glean
+
+When disabling metrics (`Glean.setUploadEnabled(false)`), metrics are also
```
Dexterp37: nit: "When disabling upload"

components/service/glean/docs/internal.md
```diff
@@ -0,0 +1,25 @@
...
+  
+When re-enabling metrics:
+
+- `first_run_date` is left as-is. (It should remain a correct date of first run
+  of the application, unaffected by disabling/enabling Glean).
+
+- `client_id` is regenerated. It has no connection to the `client_id` used prior
+  to disabling Glean.
+  
+- Application lifetime metrics are regenerated from scratch so that they will
```
Dexterp37: I'd also point out that the product is required to re-set any metric that it requires, since they are cleared as well.

components/service/glean/README.md
```diff
@@ -109,6 +109,16 @@ object parameter of the `Glean.initialize` method. For example:
 Glean.initialize(applicationContext, Configuration(channel = "beta"))
 ```
 
+### Enabling and disabling metrics
+
+`Glean.setUploadEnabled()` should be called in response to the user enabling or
+disabling telemetry.  This method should also be called at least once prior
+to calling `Glean.initialize()`.
```
georgf: Post-Fenix-MVP thought:
Can we have a follow-up bug on considering to make this setting sticky, so the "should be called at least once prior" requirement falls away?
I don't remember exactly where our discussions went on this, but from the embedding products side, this seems like a footgun (or "just one more thing to remember") that we might be able to hide away.

components/service/glean/docs/internal.md
```diff
@@ -0,0 +1,27 @@
...
+- `client_id` is set to the special value
+  `"c0ffeec0-ffee-c0ff-eec0-ffeec0ffeec0"`. This should make it possible to
+  detect the error case when metrics are sent from a client that has been disabled.
+  
+When re-enabling metrics:
+
+- `first_run_date` is left as-is. (It should remain a correct date of first run
+  of the application, unaffected by disabling/enabling Glean).
+
+- `client_id` is regenerated. It has no connection to the `client_id` used prior
```
georgf: Nit: "`client_id` is set to a new UUID" or so? I.e. call out clearly that it gets a new UUID value, "regenerated" is not very clear to me on whats happening.

components/service/glean/docs/internal.md
```diff
@@ -0,0 +1,27 @@
...
+  to disabling Glean.
+  
+- Application lifetime metrics owned by Glean are regenerated from scratch so
+  that they will appear in subsequent pings. This is the same process that
+  happens during every startup of the application when Glean is enabled. The
+  application is also responsible for setting any application lifetime metrics
+  that it manages at this time.
```
georgf: Request for a follow-up bug:

If metrics are regenerated / start recording mid-application-lifecycle (after the user happened to enable), this might lead to weirdly looking data.

Could we tell when this happened vs. a new profile? Do we set any metric that allows us detecting this state?

components/service/glean/docs/internal.md
```diff
@@ -0,0 +1,27 @@
...
+  
+When re-enabling metrics:
+
+- `first_run_date` is left as-is. (It should remain a correct date of first run
+  of the application, unaffected by disabling/enabling Glean).
+
+- `client_id` is regenerated. It has no connection to the `client_id` used prior
+  to disabling Glean.
+  
+- Application lifetime metrics owned by Glean are regenerated from scratch so
```
georgf: What about the other metric lifetimes?

components/service/glean/README.md
mdboom: Sure.  I filed [this](https://bugzilla.mozilla.org/show_bug.cgi?id=1555328)

components/service/glean/docs/internal.md
mdboom: I don't think we do -- maybe if there's a ping with a new client id long after its first_run_date?  But that seems pretty brittle.  Anyway, I filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1555335)

components/service/glean/docs/internal.md
mdboom: Good point.  Updated the text here.

components/service/glean/docs/internal.md
mdboom: They would just naturally be set next time they are recorded (as per normal operation of glean from a fresh restart).  Worth adding a sentence about that, though.

PR #2856, opened 2019-04-26
glean: 1547273: Display error from GleanDebugActivity when glean isn't initialized

mdboom:
### Pull Request checklist

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugActivity.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugActivity.kt
@@ -41,6 +41,14 @@ class GleanDebugActivity : Activity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
+        if (!Glean.isInitialized()) {
+            logger.error(
+                "Glean is not initialized. " +
+                "It may be disabled by the application."
+            )
+            return
+        }
+
         // Enable debugging options and start the application.
         intent.extras?.let {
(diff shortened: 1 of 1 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugActivity.kt
```diff
@@ -41,6 +41,14 @@ class GleanDebugActivity : Activity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
+        if (!Glean.isInitialized()) {
```
Dexterp37: Does this work even if the application is started by the intent?

mdboom: It seems to.  i.e. I tested by killing the app in the application manager and then running the `adb` command (on a version of `sample-glean` with `Glean.initialize` commented out).

Dexterp37: I wonder if this is racing with the init from the main activity :\

mdboom: Well, looking into this sent me down the rabbit hole where I found #2863 which is much higher priority.  Let's get that merged then come back to this.

mdboom: This seems to work now that #2863 is implemented, so ready for another review.

PR #2858, opened 2019-04-26
Closes reference-browser#623 - Select custom tabs on add to enable menu items

jhugman:
Fixes [#623][1].

The intent processor now selects the custom tab as it is loaded. This will allow the toolbar buttons to update with respect to the session state (`hasBack`, `hasForward`, `isLoading` etc), and direct the actions to currently selected session.

- ~[ ] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/android-components/blob/master/docs/changelog.md) or does not need one~
- ~[ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features~

[1]: https://github.com/mozilla-mobile/reference-browser/issues/623

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/tab/CustomTabConfig.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/tab/CustomTabConfig.kt
@@ -97,18 +97,22 @@ class CustomTabConfig internal constructor(
                 null
             }
 
-            val closeButtonIcon = run {
-                val icon = intent.getParcelableExtra(EXTRA_CLOSE_BUTTON_ICON) as? Bitmap
-                val density = displayMetrics?.density ?: 1f
-                if (icon != null &&
-                    icon.width / density <= MAX_CLOSE_BUTTON_SIZE_DP &&
-                    icon.height / density <= MAX_CLOSE_BUTTON_SIZE_DP
-                ) {
+            val closeButtonIcon = if (intent.hasExtra(EXTRA_CLOSE_BUTTON_ICON)) {
+                val bitmap = intent.getParcelableExtra(EXTRA_CLOSE_BUTTON_ICON) as? Bitmap
+                bitmap?.let { icon ->
+                    val density = displayMetrics?.density ?: 1f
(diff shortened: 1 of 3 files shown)


components/browser/session/src/main/java/mozilla/components/browser/session/tab/CustomTabConfig.kt
```diff
@@ -97,18 +97,22 @@ class CustomTabConfig internal constructor(
...
+                        null
+                    }
+                }?.also {
                     options.add(CLOSE_BUTTON_OPTION)
-                    icon
-                } else {
-                    null
                 }
+            } else {
+                null
```
jhugman: This hunk tidies an error message unrelated to this PR, but found in testing: `getParceableExtra` was making a big mess but not failing the tests when the close button was not being specified (hardly an exceptional case).

components/feature/intent/src/main/java/mozilla/components/feature/intent/IntentProcessor.kt
```diff
@@ -51,7 +51,7 @@ class IntentProcessor(
                     val displayMetrics = context.resources.displayMetrics
                     this.customTabConfig = CustomTabConfig.createFromIntent(safeIntent, displayMetrics)
                 }
-                sessionManager.add(session)
+                sessionManager.add(session, selected = true)
```
pocmo: While this may work in this case this solution breaks a "hidden" contract: The selected session is the "selected tab" in the browser and therefore **never** a custom tab. We failed to make this an explicit API in `SessionManager` - hopefully we can do that better in `browser-state` (e.g. we want to avoid mixing "normal" and custom tab sessions in the same list again).

pocmo: (The side effect of this patch will be that the custom tab will show up as the active tab in the browser too. And if we will have two activities (brower & custom tab) render the same session GV will even crash).

jhugman:
Fixed in [fenix/#2229][1].

[1]: https://github.com/mozilla-mobile/fenix/pull/2229

PR #2861, opened 2019-04-26
Closes #1385: Determine and expose reader-ability state

### LINKED ISSUE #1385: Expose state: "Is this page viewable in reader view?"
Issue #1346 is about showing a page in reader view. However not all pages can be watched in reader view. The engine would need to tell us and we would need to forward that information through the `Session` to the UI.

csadilek:
This closes #1385 and also gets us halfway to #2624 .

### DIFF
--- a/components/browser/session/src/test/java/mozilla/components/browser/session/SessionTest.kt
+++ b/components/browser/session/src/test/java/mozilla/components/browser/session/SessionTest.kt
@@ -609,6 +609,7 @@ class SessionTest {
         defaultObserver.onMediaAdded(session, emptyList(), mock())
         defaultObserver.onMediaRemoved(session, emptyList(), mock())
         defaultObserver.onIconChanged(session, mock())
+        defaultObserver.onReaderableStateUpdated(session, true)
     }
 
     @Test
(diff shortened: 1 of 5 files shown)


components/browser/session/src/test/java/mozilla/components/browser/session/SessionTest.kt
```diff
@@ -932,4 +933,28 @@ class SessionTest {
...
         assertEquals(bitmapMock, notifiedIcon)
     }
+
+    @Test
+    fun `observer is notified when readerable state updated`() {
+        val observer = mock(Session.Observer::class.java)
+
+        val session = Session("https://www.mozilla.org")
+        session.register(observer)
+
```
jonalmeida: Coverage didn't complain about asserting the default state? 🙃

components/feature/readerview/src/main/assets/extensions/readerview/readerview.js
```diff
@@ -5,9 +5,24 @@
 /* Avoid adding ID selector rules in this style sheet, since they could
  * inadvertently match elements in the article content. */
 
+const supportedProtocols = ["http:", "https:"];
+const blockedHosts = ["amazon.com", "github.com", "mail.google.com", "pinterest.com", "reddit.com", "twitter.com", "youtube.com"];
```
jonalmeida: Reviewing this, it seems like a random selection of choices with no explanation of why we're doing this. Maybe a comment to say this a Fennec list of approved blacklist or something of that sort?

csadilek: K, I will add some comment where this comes from.

PR #2863, opened 2019-04-26
glean: 1547330: Make sure ping metadata is loaded.

mdboom:
I discovered that following #2792, sending pings using the `GleanDebugActivity` wouldn't work, unless the app has already sent a ping on its own.

The reason for this is that sending pings by name (as opposed to using `PingType::send()`) relies on having the ping type registered in `PingType.pingRegistry`.  This registry is updated when each of the `PingType` instances are created in `PingType::init`.  However, those instances don't get created until the generated `Pings` object that they live in is instantiated.  It does not seem possible in Kotlin to create a static initializer that is always called on module import prior to the first access of the object.  Therefore, we must require that each library or application that defines a `pings.yaml` file explicitly instantiates the generated `Pings` object from its startup code.  (Glean can do this for its own pings from `Glean.initialize`, but it can't do it for the generated code for Glean-dependent applications and libraries).

Cc: @linacambridge (just FYI: I know you've been working with custom pings).

### Pull Request checklist


linacambridge:
Oh, this is good to know, thanks for the heads-up!

components/service/glean/README.md
```diff
@@ -71,11 +71,16 @@ excellent place to perform this operation is within the `onCreate` method of the
...
 ```Kotlin
+import org.mozilla.yourApplication.GleanMetrics.Pings
+
 class SampleApplication : Application() {
 
     override fun onCreate() {
         super.onCreate()
 
+        // OPTIONAL: Initialize your application's custom pings, if it has any.
+        @Suppress("UNUSED_VARIABLE") val _pings = Pings
```
Dexterp37: I don't think this is an acceptable requirement.

What if, instead, we generated a `PingRegistry` at build time, instead of populating it at run-time when we hit the constructor?

mdboom: I haven't been able to find any way to make this happen, given the requirement of having multiple `pings.yaml` files (for the app, libraries and glean itself).  You could generate a ping registry for *one* of those files, but you would still have to do something like this to merge them into a central registry.  The problem is that Kotlin doesn't have non-lazy static initialization, as far as I can tell.  (It's not a bad design choice given how problematic those things are in C++ for example).

travis79: Just brainstorming here, would it be possible to have the parser write a file-class-function that contained these `init` lines, that we could then call from `Glean.init()`?

mdboom: A workable alternative (that removes the need for the @Suppress) might be:

`Glean.registerCustomPings(Pings)`

(and this method itself would be a no-op ;)

travis79: I haven't really looked at how the parser is handling this yet, so I may be on a tangent here.

mdboom: > Just brainstorming here, would it be possible to have the parser write a file-class-function that contained these init lines, that we could then call from Glean.init()?

No, since this is generated code, `Glean.initialize()` doesn't know about it.

Maybe you could stick the package name in a `buildConfig` variable and use introspection to get at it.  But (a) introspection is problematic and (b) that wouldn't work for libraries...

mdboom: > I haven't really looked at how the parser is handling this yet, so I may be on a tangent here.

Doesn't matter -- we could extend the generator to generate whatever we need.  I'm still not sure what that buys us since we couldn't run that code implicitly, without doing *something* like this.

Dexterp37: @mdboom Is this a problem only for the debug activity?

Dexterp37: Looks like we could try to use *object expressions* in this case to work around the problem, [see here](https://kotlinlang.org/docs/tutorials/kotlin-for-py/objects-and-companion-objects.html#object-expressions).

mdboom:
@pocmo: I'd appreciate an extra set of eyes on this if possible.  Basically the context is:

1) Any Glean-using app or library has metadata about the pings it sends in `pings.yaml`, importantly whether that ping includes a client_id
2) There are cases where Glean needs to send pings on its own (if there are events left on disk when Glean starts up, for example), so we need to find this metadata by name
3) We use a singleton `pingRegistry` for this, but since Kotlin/JVM doesn't seem to have static initializers, there doesn't seem to be an *implicit* way for the ping metadata to be added to the ping registry at startup
4) The solution is to be explicit about this and require apps and libraries to do this in their startup code.  It's not a huge burden, but we always like to require as little as possible from "Glean users"

Just curious if there's an alternative solution you can think of that we may have missed.

components/service/glean/README.md
```diff
@@ -71,11 +71,16 @@ excellent place to perform this operation is within the `onCreate` method of the
...
 
 ```Kotlin
+import org.mozilla.yourApplication.GleanMetrics.Pings
+
 class SampleApplication : Application() {
 
     override fun onCreate() {
         super.onCreate()
 
+        // OPTIONAL: Initialize your application's custom pings, if it has any.
```
travis79: nit: The wording of this comment seems a little unclear to me.  Maybe this could be worded more like:
```
// If you have custom pings in your application, you must register them using the following command. 
// This command should be omitted for applications not using custom pings.
```

mdboom: Good point.  Thanks.

PR #2864, opened 2019-04-26
Fixes #2769: Allows disabling of autocomplete text

### LINKED ISSUE #2769: Autocomplete suggestion gets triggered on search term
## STR
1. Type "Mo" then backspace to remove the autocomplete suggestion
2. Press "go" button
3. Tap the URL bar to go back to the search view

## Expected
1. URL bar displays "Mo"

## Actual
1. URL bar displays "Mozilla.org" (the autocompleted suggestion based on the search query).

sblatz:
### Pull Request checklist


components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/edit/EditToolbar.kt
```diff
@@ -116,8 +116,12 @@ internal class EditToolbar(
      * Updates the URL. This should only be called if the toolbar is not in editing mode. Otherwise
      * this might override the URL the user is currently typing.
      */
-    fun updateUrl(url: String) {
+    fun updateUrl(url: String, shouldAutoComplete: Boolean = false) {
+        val previousEnabledState = urlView.isEnabled
+
+        urlView.isEnabled = shouldAutoComplete
         urlView.setText(url)
+        urlView.isEnabled = previousEnabledState
```
pocmo: nit: It feels like we could move this even one level lower into `InlineAutoCompleteText`. Maybe by adding an optional `shouldAutoComplete` parameter to `setText()`?

components/ui/autocomplete/src/main/java/mozilla/components/ui/autocomplete/InlineAutocompleteEditText.kt
```diff
@@ -137,6 +137,8 @@ open class InlineAutocompleteEditText @JvmOverloads constructor(
     private var autoCompleteSpans: List<Any>? = null
     // Do not process autocomplete result
     private var discardAutoCompleteResult: Boolean = false
+    // Controls whether or not suggestions are shown
+    var enableAutocomplete: Boolean = true
```
pocmo: It looks like you are not using this and instead rely on `isEnabled`?

sblatz: Ah, good catch. This was leftover from a previous implementation.

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/edit/EditToolbar.kt
sblatz: I moved it down a level, but that required me modifying the override function to not be overriden since we can't add parameters to the EditText interface (Android native). I'm unsure if this has unexpected results.

pocmo:
Moved changelog entry to 0.52.0-SNAPSHOT and landed in 365282612db6eea7b3c8315b20bf8187a1a215fe

PR #2865, opened 2019-04-26
History tracking delegate improvements

grigoryk:
Changes:
- added a scheme blacklist to HistoryDelegate
- added `VisitType` to `HistoryTrackingDelegate`'d public API, forcing consumers to generate a visit type
- mapped available GV flags to our internal visit types. Same for SystemEngine.

### Pull Request checklist

### DIFF
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -22,6 +22,7 @@ import mozilla.components.concept.engine.Settings
 import mozilla.components.concept.engine.history.HistoryTrackingDelegate
 import mozilla.components.concept.engine.request.RequestInterceptor
 import mozilla.components.concept.engine.request.RequestInterceptor.InterceptionResponse
+import mozilla.components.concept.storage.VisitType
 import mozilla.components.support.ktx.android.util.Base64
 import mozilla.components.support.ktx.kotlin.isEmail
 import mozilla.components.support.ktx.kotlin.isGeoLocation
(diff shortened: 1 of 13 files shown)


grigoryk:
Need to test this locally. I'll post an update once that's done.

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -395,17 +397,34 @@ class GeckoEngineSession(
...
-                // Don't track visits in private mode, redirects, or error
-                // pages, even if they're top-level visits.
+                // Don't track visits in private mode, or error pages, even if they're top-level visits.
                 return GeckoResult.fromValue(false)
             }
 
+            val isReload = lastVisitedURL?.let { it == url } ?: false
+
+            val visitType = if (isReload) {
+                VisitType.RELOAD
```
pocmo: Did you ask if this is a flag/type GV could tell us about?

components/feature/session/src/main/java/mozilla/components/feature/session/HistoryDelegate.kt
```diff
@@ -32,4 +36,33 @@ class HistoryDelegate(private val historyStorage: HistoryStorage) : HistoryTrack
...
+        // Allow about about:reader uris. They are of the form:
+        // about:reader?url=http://some.interesting.page/to/read.html
+        if (uri.startsWith("about:reader")) {
+            return true
+        }
+
+        val schemasToIgnore = listOf(
+            "about", "imap", "news", "mailbox", "moz-anno", "moz-extension",
+            "view-source", "chrome", "resource", "data", "javascript", "blob"
+        )
```
pocmo: I was wondering if an "allowlist only" would be a better choice.. but seeing that this is from Fennec this seems to be fine. :)

grigoryk: Not sure how well a whitelist will work here. It seems more flexible to forbid certain select schemes than to allow certain schemes.

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
grigoryk: Yeah; they're open to revising/expanding these flags (cc @linacambridge), but let's not block on that.

grigoryk:
@pocmo now with more tests. Let's do another pass? :)

grigoryk:
Looks good in local testing.

PR #2875, opened 2019-04-29
1547330: Minimal fix for sending built-in pings from GleanDebugActivity

mdboom:
This is a subset of #2683 to just fix the issue that built-in Glean pings can't be sent using `GleanDebugActivity`, while we work out the broader question of how to best support custom pings, etc.

### Pull Request checklist


components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -85,6 +85,8 @@ open class GleanInternalAPI internal constructor () {
             return
         }
 
+        @Suppress("UNUSED_VARIABLE") var pings = Pings
```
travis79: Let's add a brief comment here explaining what and why we are doing this here.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -85,6 +85,11 @@ open class GleanInternalAPI internal constructor () {
             return
         }
 
+        // Explicitly instantiate the Pings object so the built-in pings will be
+        // added to the ping registry.  This allows the GleanDebugActivity to look
+        // up pings by name.
+        @Suppress("UNUSED_VARIABLE") var pings = Pings
```
Dexterp37: @mdboom late nit: this could have been a `val`. Otherwise, we get a warning in the IDE :)

PR #2876, opened 2019-04-29
Closes #2706: Do not remember site permissions in private browsing.

### LINKED ISSUE #2706: Do not remember site permissions in private browsing
Related issue: https://github.com/mozilla-mobile/fenix/issues/1309

Amejia481:
### Pull Request checklist

### DIFF
--- a/components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
+++ b/components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
@@ -68,7 +68,11 @@ class SitePermissionsFeature(
 ) : LifecycleAwareFeature {
 
     private val observer = SitePermissionsRequestObserver(sessionManager, feature = this)
-    internal val ioCoroutineScope by lazy { CoroutineScope(Dispatchers.IO) }
+    internal val ioCoroutineScope by lazy { coroutineScopeInitializer() }
+
+    internal var coroutineScopeInitializer = {
+        CoroutineScope(Dispatchers.IO)
+    }
 
     override fun start() {
         observer.observeSelected()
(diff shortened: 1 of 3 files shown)


components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
```diff
@@ -120,18 +120,23 @@ class SitePermissionsFeature(
...
 
     @Synchronized internal fun storeSitePermissions(
+        session: Session,
         request: PermissionRequest,
         permissions: List<Permission> = request.permissions,
         status: SitePermissions.Status
     ) {
+        if (session.private) {
+            return
+        }
```
pocmo: I wonder if we should do that check one level higher where we also check `shouldStore` .. there's probably no need to launch the coroutine?

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
```diff
@@ -176,8 +181,12 @@ class SitePermissionsFeature(
             return null
         }
 
-        val permissionFromStorage = withContext(ioCoroutineScope.coroutineContext) {
-            storage.findSitePermissionsBy(request.host)
+        val permissionFromStorage = if (session.private) {
+            null
+        } else {
```
pocmo: Is the opposite really true too? On desktop permissions granted in normal browsing seem to be granted in private browsing automatically too.

Amejia481: You right, I will remove this one!

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
Amejia481: Yeah, that could be much better. I will add the coroutine launch in `storeSitePermissions` function!

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
```diff
@@ -68,14 +69,19 @@ class SitePermissionsFeature(
...
+        CoroutineScope(Dispatchers.IO)
+    }
 
     override fun start() {
         observer.observeSelected()
     }
 
     override fun stop() {
         observer.stop()
+        ioCoroutineScope.cancel()
```
Amejia481: @pocmo I also added a this to cancel all the coroutines when the feature is stopped

Amejia481:
@pocmo PR updated :)

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
pocmo: Doesn't this mean the feature won't work anymore on the next start? Or can you continue to use the scope after cancelling?

Amejia481: Good point, I will do some research!

Amejia481: I will address that in https://github.com/mozilla-mobile/android-components/issues/2889. I will remove this part from this PR

pocmo: 👍

PR #2878, opened 2019-04-30
Closes #2624: ReaderView: Wire up messaging between feature and web ext.

### LINKED ISSUE #2624: ReaderView: Wire up messaging between feature and web extension

csadilek:
This is the first functioning version of ReaderView including appearance controls, integrated in our sample browser.

readerview


csadilek:
OK, added tests and changelog. This is ready now.

components/feature/readerview/src/main/assets/extensions/readerview/manifest.json
```diff
@@ -7,7 +7,7 @@
       "matches": ["*://*/*"],
       "js": ["readability/readability-0.2.0.js", "readability/readability-readerable-0.2.0.js", "moment/moment-2.24.0.js", "readerview.js"],
       "css": ["readerview.css"],
-      "run_at": "document_end"
+      "run_at": "document_idle"
```
jonalmeida: This seems like the right thing to do but I've noticed in some testing yesterday that pages don't always re-enter reader view on app startup (maybe because something is still running?)

STR:
1. Visit an article on `blog.mozilla.org`.
2. Enable reader mode.
3. Swipe away the app from the app drawer.
4. Re-open and observe the state of the page.

You see the reader view controls button and the reader view icon is enabled but the page isn't actually in reader view.

csadilek: @jonalmeida ah, but this is unrelated. This happens when the app closes before the session state was persisted. You can reproduce this by simply loading a new URL and quickly closing the app too.

PR #2890, opened 2019-04-30
glean: 1547967: Support additional ping metadata

mdboom:
This PR won't work until https://github.com/mozilla/glean_parser/pull/61 is merged.

This adds the additional ping metadata to the built-in pings.  It also implements the `disabled` parameter for pings.

### Pull Request checklist

### DIFF
--- a/components/service/glean/docs/pings/custom.md
+++ b/components/service/glean/docs/pings/custom.md
@@ -10,9 +10,13 @@ directory alongside your app's `metrics.yaml` file.
 
 Each ping has the following parameters:
 
-- `description` (required): A human-readable description of the ping.
 - `include_client_id` (required): A boolean indicating whether to include the
   `client_id` in the [`client_info` section](pings.md#The-client_info-section)).
+
+In addition to these parameters, pings also support the parameters related to
+data review and expiration defined in [common metric
+parameters](../metrics/adding-new-metrics.md#common-metric-parameters):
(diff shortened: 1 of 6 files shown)


mdboom:
> I think we should go ahead and add the new fields to the custom.md documentation while we are at it.

Good point.  Added.

components/service/glean/docs/pings/custom.md
```diff
@@ -25,6 +29,14 @@ search:
   description: >
     A ping to record search data.
   include_client_id: false
+  notification_emails:
+    - CHANGE-ME@test-only.com
+  bugs:
+    - 123456789
+  data_reviews:
+    - http://test-only.com/path/to/data-review
```
badboy: Use `example.com` in examples: https://tools.ietf.org/html/rfc2606#section-3

mdboom: Good point.  We have a bunch of other instances of these that I've changed as well.

mdboom:
I've ripped out the expiration / disabling on pings, which can definitely be addressed later, in order to move this forward.

mdboom:
A few things to get this across the finish line:

- We should determine what bugs we want to link to for the built-in pings.

- Merging this will require the usual glean_parser coordination **plus** the custom activation ping metadata in Fenix will need to be added.

PR #2909, opened 2019-05-02
Using DialogFragment for site permissions prompts

Amejia481:
@pocmo I was doing some tests using `DialogFragment` instead of `DoorhangerPrompt`. I want hear what do you think about this implementation before switching completely.

One use case that we need to support is changing the site permissions prompts position. If we are in normal browsing is should be at the bottom covering the toolbar and for custom tab should be at the top.

image

[An example of switching prompts position between normal browsing and custom tabs](https://drive.google.com/file/d/1Rnq4rsL6REES26tyUfHH1XoOP-mqQavd/view?usp=sharing)

Some advantages of `DialogFragment` as it a `Fragment` is can survive configuration changes.

I didn't change the internal implementation of `DoorhangerPrompt` to use `DialogFragment`:

* I don't know if we want to use it as is now for other purposes.
* For Site permissions goal creating a custom`DialogFragment` is just fine, I think. It seems over-killing to convert `DoorhangerPrompt` to a `DialogFragment` and support restoring all the dynamic `controlGroups` and `buttons` in case the Fragment is destroyed by the system.

### DIFF
--- a/components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsDialogFragment.kt
+++ b/components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsDialogFragment.kt
@@ -1,0 +1,141 @@
+package mozilla.components.feature.sitepermissions
+
+import android.annotation.SuppressLint
+import android.app.Dialog
+import android.graphics.Color
+import android.os.Bundle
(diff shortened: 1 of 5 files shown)


pocmo:
@Amejia481 Looks good and your reasons make absolutely sense to me. Let's try to build that with a dialog fragment directly and after that we can think of deprecating the doorhanger component.

Amejia481:
👍

PR #2913, opened 2019-05-02
WIP: glean: Avoid returning duplicate sequence numbers

mdboom:
### Pull Request checklist

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/ping/PingMaker.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/ping/PingMaker.kt
@@ -39,12 +39,17 @@ internal class PingMaker(
      * @param pingName The name of the ping
      * @return sequence number
      */
+    @Synchronized
     private fun getPingSeq(pingName: String): Int {
+        // This is synchronized to avoid returning the same seq number twice.
+
         sharedPreferences?.let {
             val key = "${pingName}_seq"
             val currentValue = it.getInt(key, 0)
             val editor = it.edit()
             editor.putInt(key, currentValue + 1)
+            // This writes to disk asynchronously, but updates the in-memory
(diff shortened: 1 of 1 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/ping/PingMaker.kt
```diff
@@ -39,12 +39,17 @@ internal class PingMaker(
      * @param pingName The name of the ping
      * @return sequence number
      */
+    @Synchronized
```
Dexterp37: This makes sense to me, in theory. However, I'm struggling to figure out a case in which that would happen :\

mdboom: Yeah, ditto.  I'm grasping at straws, honestly.  Also, it introduces an interkitten in Taskcluster that I can't reproduce locally...

Dexterp37: @travis79 @mdboom .. any idea?

travis79: Maybe if we we hit the events limit, went over by 1 or more events, and then immediately went to background.  All other paths through here seem pretty isolated/sync.   That's the only situation I can come up with that might have us two separate execution threads in this same bit of code near the same time.

mdboom: Yeah.  I can't imagine anything else.  But that seems really unlikely to explain the high number of the dupes found in chutten's analysis.  (Plus, I'm not sure any of the dupes involve the event ping).  I also don't think we're getting a ton of events from fenix, but it's probably worth checking to see if we have any event pings with around 500 events in them.

PR #2919, opened 2019-05-03
Make Glean mock webserver reply 200 to all requests

Dexterp37:
`MockWebServer` provides a `QueuedDispatcher` which requires the consumer to enqueue a response for every single received request. If no response is found for a queued request, the test times out in an unclear way. This changes the tests to use a mockweb server that always replies 200: if the specific test receives an unexpected ping, it will fail due to the unexpected content, rather than timing out in obscure ways.

### Pull Request checklist


Dexterp37:
Well, at least tests are failing due to non-Glean problems.

Dexterp37:
> 
> 
> Yeah, LGTM.
> 
> I know we don't have any now, but there may be tests in the future where we want to confirm we get some different response from the server -- I guess in that cause we just use the old pattern?

Yes, in that specific case we'll go the old way!

PR #2922, opened 2019-05-03
Closes #1740: Add support for authentication alerts for SystemEngineView

### LINKED ISSUE #1740: Add support for authentication alerts for SystemEngineView
Related https://github.com/mozilla-mobile/android-components/issues/1470

pengrad:
### Pull Request checklist

For #1740

I tried to make this dialog same in Gecko.
Texts from here: https://github.com/mozilla/gecko/blob/central/toolkit/locales/en-US/chrome/global/commonDialogs.properties#L28
URL formatting and realm trimming from https://github.com/mozilla/gecko/blob/central/mobile/android/components/geckoview/GeckoViewPrompt.js#L736

Title is empty to use default one ("Sign in") which is used in Gecko also.
Other params are default as WebView supports only Basic auth or we can't get that info (such as Method.HOST, Level.NONE)


components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -273,6 +275,42 @@ class SystemEngineView @JvmOverloads constructor(
...
+            } else {
+                context.getString(R.string.mozac_browser_engine_system_auth_message, trimmedRealm, formattedUrl)
+            }
+
+            session.notifyObservers {
+                onPromptRequest(
+                    PromptRequest.Authentication(
+                        "",
+                        message,
+                        "",
```
Amejia481: You could get the user and password from `Webview` by using [WebViewDatabase.getHttpAuthUsernamePassword](https://developer.android.com/reference/android/webkit/WebViewDatabase.html#getHttpAuthUsernamePassword(java.lang.String,%20java.lang.String)) or [webView.getHttpAuthUsernamePassword](https://developer.android.com/reference/android/webkit/WebView#getHttpAuthUsernamePassword(java.lang.String,%2520java.lang.String)).

```kotlin
        private fun WebView.getAuthCredentials(host: String, realm: String): Pair<String, String> {

            val credentials = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
                val webViewDB = WebViewDatabase.getInstance(context)
                webViewDB.getHttpAuthUsernamePassword(host, realm)
            } else {
                this.getHttpAuthUsernamePassword(host, realm)
            }

            var credentialsPair = "" to ""

            if (!credentials.isNullOrEmpty() && credentials.size == 2) {

                val user = if (credentials[0] != null) credentials[0] else ""
                val pass = if (credentials[1] != null) credentials[1] else ""

                credentialsPair = user to pass
            }
            return credentialsPair
        }
```

pengrad:
@Amejia481 I've added credentials retrieval and tests for it. Could you please take a look again?

Amejia481:
Awesome @pengrad. Sure I'll take a look!

Amejia481:
Thanks @pengrad 👍

PR #2943, opened 2019-05-08
Update history storage unit tests to exercise native (Rust) appservices stack

grigoryk:
Let's start changing our unit tests that run against appservices libs to actually exercise the native stack whenever feasible.


grigoryk:
@pocmo you'll be happy to see that this _increases_ test coverage ;)

rfk:
> I hate that we still need these @aar/@jar workarounds but love that this is working for you

@ncalexan do we have any theories about how to make this situation better over time?

ncalexan:
> > I hate that we still need these @aar/@jar workarounds but love that this is working for you
> 
> @ncalexan do we have any theories about how to make this situation better over time?

Yes: we could repackage our own versions and call them different names.

The advantage of repackaging is that we can strip `.so` files that shouldn't ever be there -- I think we have mips deps that are garbage in the upstream ones.

The advantage of different names is that we avoid the @extension thing altogether, which is a rather crufty bit of Gradle that doesn't work that well.

Assuming that's license-compatible, it's not "hard" per se, it's Just More Work.

PR #2955, opened 2019-05-09
Closes #1974: Destroy FxA device record after logging out

### LINKED ISSUE #1974: Destroy FxA device after logging out
Part of FxA device management. Fennec implementation - https://dxr.mozilla.org/mozilla-central/source/mobile/android/services/src/main/java/org/mozilla/gecko/background/fxa/FxAccountClient20.java#826

grigoryk:
This destroys the FxA device record for the current device on logout.

Tests on the constellation manager are lacking because mockito really doesn't want to mock fxaclient/FirefoxAccount, it doesn't have an interface it implements, and even though we're able to use the real object + native stack in tests now (see https://github.com/mozilla-mobile/android-components/pull/2943), we can't do it in this case yet because tests involving the network aren't a supported use case.

### Pull Request checklist

### DIFF
--- a/components/concept/sync/src/main/java/mozilla/components/concept/sync/Devices.kt
+++ b/components/concept/sync/src/main/java/mozilla/components/concept/sync/Devices.kt
@@ -27,6 +27,15 @@ interface DeviceConstellation : Observable<DeviceEventsObserver> {
     ): Deferred<Unit>
 
     /**
+     * Destroy current device record.
+     * Use this when device record is no longer relevant, e.g. while logging out. On success, other
+     * devices will no longer see the current device in their device lists.
+     *
+     * @return A boolean flag indicating if the operation succeeded.
+     */
+    fun destroyCurrentDeviceAsync(): Deferred<Boolean>
+
+    /**
      * Ensure that all initialized [DeviceCapability], such as [DeviceCapability.SEND_TAB], are configured.
      * This may involve backend service registration, or other work involving network/disc access.
      * @return A [Deferred] that will be resolved once operation is complete.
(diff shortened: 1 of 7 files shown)


components/concept/sync/src/main/java/mozilla/components/concept/sync/Devices.kt
```diff
@@ -26,6 +26,8 @@ interface DeviceConstellation : Observable<DeviceEventsObserver> {
         capabilities: List<DeviceCapability>
     ): Deferred<Unit>
 
+    fun destroyCurrentDeviceAsync(): Deferred<Boolean>
```
csadilek: KDocs missing! :)

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/Utils.kt
```diff
@@ -0,0 +1,23 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.fxa
+
+/**
+ * Runs a provided lambda, and if that throws non-panic FxA exception, runs the second lambda.
+ */
+fun <T> maybeExceptional(block: () -> T, onFailure: (e: FxaException) -> T): T {
```
csadilek: nice!

nit: I'd find `handleError` or `withErrorHandling` clearer here.

components/concept/sync/src/main/java/mozilla/components/concept/sync/Devices.kt
grigoryk: Damn! Thanks :)

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/Utils.kt
grigoryk: ah, yeah, that's clearer. naming things is hard :)

PR #2956, opened 2019-05-09
Fixed issue on SitePermissionsFeature that doesn't allow its usage on custom tabs

Amejia481:
`SitePermissionsFeature` was using the selected session `onPermissionsResult`, instead of checking the custom session id. This didn't allow its usage on custom tabs. Related issue https://github.com/mozilla-mobile/fenix/issues/2343

### Pull Request checklist

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
@@ -224,21 +224,24 @@ fun SessionManager.runWithSession(
 }
 
 /**
- * Tries to find a session with the provided session ID or uses them selectedSession and runs the block if found.
+ * Tries to find a session with the provided session ID or uses the selected session and runs the block if found.
  *
  * @return True if the session was found and run successfully.
  */
-inline fun SessionManager.runWithSessionIdOrSelected(
+fun SessionManager.runWithSessionIdOrSelected(
     sessionId: String?,
-    block: SessionManager.(Session) -> Boolean
+    block: SessionManager.(Session) -> Unit
 ): Boolean {
+    var blockWasRun = false
     sessionId?.let {
         findSessionById(sessionId)?.let { session ->
-            return block(session)
+            block(session)
+            blockWasRun = true
         }
     }
     selectedSession?.let {
-        return block(it)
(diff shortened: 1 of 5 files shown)


components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
```diff
@@ -213,3 +213,23 @@ fun SessionManager.runWithSession(
     }
     return false
 }
+
+/**
+ * Tries to find a session with the provided session ID or uses them selectedSession and runs the block if found.
+ *
+ * @return True if the session was found and run successfully.
```
pocmo: It says it returns `True` if the session was found and the block run. But it returns what the block returns. So if the block returns `False` then even if the session was found and the block was run it returned false? Maybe that's correct and it's just the wording of this doc that is confusing?

Amejia481: You are right. If the session was found, I was delegating to the lambda function to determine if the block run successfully. But the consumer just wants to know if the block was executed or not. I'll update it to reflect that 👍

components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
```diff
@@ -222,3 +222,26 @@ fun SessionManager.runWithSession(
...
+        findSessionById(sessionId)?.let { session ->
+            block(session)
+            blockWasRun = true
+        }
+    }
+    selectedSession?.let {
+        block(it)
+        blockWasRun = true
+    }
+    return blockWasRun
```
pocmo: @Amejia481 Wait. Now that you are not returning early.. are you not running the block twice if there's a session id and a selected session?

Amejia481: Fixing it!

Amejia481: Thanks for the catch!

Amejia481: https://github.com/mozilla-mobile/android-components/pull/2970

PR #2958, opened 2019-05-09
Have the sync sample show the entire bookmarks tree

mhammond:
This enhances @grigoryk's sync sample so that it shows the entire bookmarks tree. It's very crude but seems OK for a sample for now.


grigoryk:
This is quite valuable to see, but I've specifically limited the list to the top roots so that it doesn't overwhelm the rest of the UI. I normally use this against my personal account which, for better or worse, has a lot of bookmarks.

Especially with the devices list, we'll at least need to re-order them, and ideally move this full list into a separate activity. Any chance you can do that, @mhammond? It should be pretty straightforward.

grigoryk:
Oh! I didn't notice that you've added scrolling. Maybe that's good enough then

grigoryk:
Can you appease ktlint and detekt? You can run them locally via `./gradlew ktlint` and `./gradlew detekt`

grigoryk:
@mhammond ktlint still failed :(

PR #2959, opened 2019-05-09
Fix experiments library issues from smoke test.

georgf:
We now have server-side Kinto setups we can test against. Per [bug 1544751](https://bugzilla.mozilla.org/show_bug.cgi?id=1544751) i've tested the library against some test experiments in Kinto.
This PR contains fixes for the issues i found in these smoke tests.

This is still WIP:
- Some test failures are not fixed yet.
- Some of the fixes may need explicit test coverage.

### Pull Request checklist

### DIFF
--- a/components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
+++ b/components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
@@ -4,6 +4,7 @@
 
 package mozilla.components.service.experiments
 
+import android.annotation.SuppressLint
 import android.content.Context
 import mozilla.components.support.base.log.logger.Logger
 import mozilla.components.service.glean.Glean
@@ -39,13 +40,17 @@ open class ExperimentsInternalAPI internal constructor() {
 
     internal lateinit var configuration: Configuration
 
+    private lateinit var context: Context
+
     /**
      * Initialize the experiments library.
      *
      * This should only be initialized once by the application.
      *
      * @param applicationContext [Context] to access application features, such
-     * as shared preferences.
+     * as shared preferences.  As we cannot enforce through the compiler that the context pass to
+     * the initialize function is a applicationContext, there could potentially be a memory leak
+     * if the initializing application doesn't comply.
      */
     fun initialize(
         applicationContext: Context,
@@ -66,17 +71,22 @@ open class ExperimentsInternalAPI internal constructor() {
 
         experimentsResult = ExperimentsSnapshot(listOf(), null)
         experimentsLoaded = false
-
+        context = applicationContext
         storage = getExperimentsStorage(applicationContext)
 
         isInitialized = true
 
-        // Load cached experiments from storage. After this, experiments status
-        // is available.
+        // Load cached experiments from storage. After this, experiments status is available.
         loadExperiments()
 
-        // Load active experiment from cache, if any.
+        // Load the active experiment from cache, if any.
         activeExperiment = loadActiveExperiment(applicationContext, experiments)
+
+        // If no active experiment was loaded from cache, check the cached experiments list for any
+        // that should be launched now.
+        if (activeExperiment == null) {
+            findAndStartActiveExperiment()
+        }
 
         // We now have the last known experiment state loaded for product code
         // that needs to check it early in startup.
@@ -122,20 +132,6 @@ open class ExperimentsInternalAPI internal constructor() {
         experimentsLoaded = true
     }
 
-    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
-    internal fun loadActiveExperiment(
-        context: Context,
-        experiments: List<Experiment>
-    ): ActiveExperiment? {
-        val activeExperiment = ActiveExperiment.load(context, experiments)
-        logger.info(activeExperiment?.let
-            { """Loaded active experiment - id="${it.experiment.id}", branch="${it.branch}"""" }
-            ?: "No active experiment"
-        )
-
-        return activeExperiment
-    }
-
     /**
      * Requests new experiments from the server and
      * saves them to local storage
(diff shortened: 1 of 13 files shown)


georgf:
@travis79, i'm requesting a first review pass from you this week to check especially:
- If the overall changes make sense.
- If there are any concerns, clear issues.
- If there are any changes that need explicit test coverage.

samples/glean/src/main/res/layout/activity_main.xml
```diff
@@ -64,4 +64,24 @@
...
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:text="Check experiments" />
+
+    <TextView
+        android:id="@+id/textViewExperimentStatus"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:background="@android:color/white"
+        android:text="Experiment not active"
```
travis79: nit:  Extract string resource to strings.xml

samples/glean/src/main/res/layout/activity_main.xml
```diff
@@ -64,4 +64,24 @@
         android:layout_marginTop="5dp"
         android:text="@string/send_ping_info" />
 
+    <Button
+        android:id="@+id/buttonCheckExperiments"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:text="Check experiments" />
```
travis79: Another place where you aren't following best practices and putting the label "Check experiments" in strings.xml and using `@strings/check_experiments` or whatever you named it.

samples/glean/src/main/res/layout/activity_main.xml
```diff
@@ -64,4 +64,24 @@
...
+        android:layout_height="wrap_content"
+        android:background="@android:color/white"
+        android:text="Experiment not active"
+        android:textStyle="italic" />
+
+    <TextView
+        android:id="@+id/textView3"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:text='The CHECK EXPERIMENTS button checks the experiments status of the "test-color" experiment and sets the color accordingly.' />
```
travis79: nit:  Extract string resource to strings.xml

travis79:
I added a commit to fix the `SignatureVerifierTests` but on the `ExperimentsTests` that are failing, for whatever reason it's not setting the `activeExperiment` correctly anymore.  I've done some debugging around the failed tests to try and determine why it's not setting it correctly but I haven't found anything yet.  I'll continue this in the morning with fresh eyes.

travis79:
@pocmo: @georgf  indicated that you have some context on the experiments library and I've taken over the sheparding of this PR and was hoping you could give it a review.  Thanks!

georgf:
> @pocmo: @georgf indicated that you have some context on the experiments library and I've taken over the sheparding of this PR and was hoping you could give it a review. Thanks!

For what it's worth, i can review logical changes here, if they match the design we've specd.
Another review look from the Android engineering side would be great (as the other two people on the team with Android experience are on PTO).

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsTest.kt
```diff
@@ -486,75 +414,76 @@ class ExperimentsTest {
                     Experiment.Branch("branch3", 1)
                 )
             )
-        ))
-
+        )
         // This should force choosing branch 2 of the experiment on update.
         branchDiceRoll = 1
-        experimentsUpdater.updateExperiments()
+        experiments.onExperimentsUpdated(ExperimentsSnapshot(experimentsList, null))
```
georgf: If we're removing the server mocking here, i assume we're testing that integration path (updater -> experiments -> correct result) in another unit test?

travis79: I think that the `second update of experiments` test covers this integration path correctly.  I ended up removing some of the mocking here and calling the `onExperimentsUpdated()` directly to try and simplify an already complex test, and because trying to rely on the update while using the `mockContext` fails because the WorkManager requires an actual `context` to run and it doesn't work out well to mix an match mock and real contexts.

georgf: Fair enough on the simplifying overly complex tests. As long as we cover the path elsewhere i'm fine with it.

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
```diff
@@ -314,6 +364,7 @@ open class ExperimentsInternalAPI internal constructor() {
     }
 }
 
+@SuppressLint("StaticFieldLeak")
```
Amejia481: As we can't not enforce by the compiler, that the `context` pass to `initialize` function is a `applicationContext` I think could be a good idea at least, warm the consumers in the docs, about possible memory leaks if the don't comply.

``` kotlin
    /**
     * Initialize the experiments library.
     *
     * This should only be initialized once by the application.
     *
     * as shared preferences. No passing an applicationContext here could cause memory leaks.
     */
    fun initialize(
        applicationContext: Context,
    ){}
```

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
```diff
@@ -146,35 +140,100 @@ open class ExperimentsInternalAPI internal constructor() {
...
      * @param experimentId the id of the experiment
      * @param block block of code to be executed if the user is part of the experiment
      */
-    fun withExperiment(context: Context, experimentId: String, block: (branch: String) -> Unit) {
-        val activeExperiment = evaluator.evaluate(
-            context,
-            ExperimentDescriptor(experimentId), experimentsResult.experiments
-        )
-        activeExperiment?.let { block(it.branch) }
+    fun withExperiment(experimentId: String, block: (branch: String) -> Unit) {
```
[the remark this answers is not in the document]
travis79: If we make this inline, it then exposes some of the internal variables through the public API.  I'll file a follow-up bug to investigate if we can reduce the memory overhead here and other places, but it looks like it would require more refactoring to accomplish.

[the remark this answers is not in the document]
Amejia481: 👍

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
travis79: Expanded the comment to reflect this, thanks!

PR #2966, opened 2019-05-09
Change service-experiments tests to use a fake Kinto endpoint

travis79:
This is to prevent accidentally downloading experiments from the server and affecting test results.

### Pull Request checklist

### DIFF
--- a/components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsTest.kt
+++ b/components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsTest.kt
@@ -73,7 +73,9 @@ class ExperimentsTest {
         // Initialize WorkManager (early) for instrumentation tests.
         WorkManagerTestInitHelper.initializeTestWorkManager(context)
 
-        configuration = Configuration()
+        // Setting the endpoint to a non-existent one to prevent actual experiments from being
+        // downloaded to tests.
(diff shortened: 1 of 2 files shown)


components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsTest.kt
```diff
@@ -73,7 +73,9 @@ class ExperimentsTest {
         // Initialize WorkManager (early) for instrumentation tests.
         WorkManagerTestInitHelper.initializeTestWorkManager(context)
 
-        configuration = Configuration()
+        // Setting the endpoint to a non-existent one to prevent actual experiments from being
+        // downloaded to tests.
+        configuration = Configuration(kintoEndpoint = "https://example.com")
```
georgf: This domain actually exists and could lead to retrieving unexpected data or to unexpected delays.

Please change this to [`https://example.invalid`](https://serverfault.com/questions/846522/is-there-an-example-domain-which-is-guaranteed-not-to-exist)

travis79: Changed to `.invalid`

travis79:
> Question for a follow-up bug:
> Is there a way to default all tests for the library to that dummy domain?

Follow-up bug filed [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1550732)

PR #2967, opened 2019-05-09
Closes #2962: Reader View styles can be affected by css on page [ci skip]

### LINKED ISSUE #2962: Reader View css styles can be affected by other css on the page
This is similar to #2800 where our CSS isn't properly isolated but in this case it's styles on the page that affect reader view.

csadilek:


jonalmeida:
Tested this out as well! ✅

Issue #2988, opened 2019-05-11
(Advice needed) Uncheck "Remember decision" on permision request dialog. 


dector:
There is issue [#fenix/2403](https://github.com/mozilla-mobile/fenix/issues/2403) in Fenix project to uncheck "Remember decision" checkbox by default.

![Request permission dialog](https://user-images.githubusercontent.com/265872/57565463-0b7df100-73c7-11e9-8a0e-a775c074abf1.jpg)

However, current implementation has hard-coded value that is not configurable.
[Layout](https://github.com/mozilla-mobile/android-components/blob/master/components/feature/sitepermissions/src/main/res/layout/mozac_site_permissions_prompt.xml#L54) | [Fragment](https://github.com/mozilla-mobile/android-components/blob/master/components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsDialogFragment.kt#L154)

I see two ways of fixing this:

**a)** _(Breaking change. Simple solution.)_ Just change value in layout to "more-privacy-by-default" (unchecked).
**b)** _(Non-breaking change. More code.)_ Leave checked value by default. Add ability to configure default checked state in [SitePermissionsFeature](https://github.com/mozilla-mobile/android-components/blob/master/components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt#L63) from outside.

Please advise. I'm assuming that option **b** is a way to go for now.

pocmo:
CC: @Amejia481

jan-ivar:
What does "breaking" means in this case?

> Leave checked value by default. 

On the surface this does not sound like it would solve https://github.com/mozilla-mobile/fenix/issues/2403, or am I mistaken?

mheubusch:
IANAEngineering, and may not be understanding  this issue, but I would  think we should always strive to fix the issue directly and not rely on a workaround in the case of user control and privacy.  Like the bug in 2403 mentions, the default setting for camera, microphone, location permissions is "Always Ask" so we **should not have** the Remember this decision checked by default, even if we have different code in components to reverse that decision, in the   case that  a-c changes and someone is not aware  of the dependency  and Fenix reverts to the check becoming the  default.

Amejia481:
@dector thanks for taking this. 

[...]

PR #2989, opened 2019-05-11
For #2988. Added configuration for "Remember choice" checkbox in Permissions dialog

dector:
Please read #2988 and make diagonal-check solution in this PR. Is this a way-to-go solution?
Or should we go with breaking change instead (I suppose that it'll be nice to have ability to configure default behavior)?

### Pull Request checklist

### DIFF
--- a/components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsDialogFragment.kt
+++ b/components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsDialogFragment.kt
@@ -21,54 +21,59 @@ import android.widget.Button
 import android.widget.LinearLayout.LayoutParams
 import androidx.appcompat.app.AppCompatDialogFragment
 import androidx.core.content.ContextCompat
+import kotlin.properties.Delegates
 
 internal const val KEY_SESSION_ID = "KEY_SESSION_ID"
 internal const val KEY_TITLE = "KEY_TITLE"
-private const val KEY_USER_CHECK_BOX = "KEY_USER_CHECK_BOX"
 private const val KEY_DIALOG_GRAVITY = "KEY_DIALOG_GRAVITY"
 private const val KEY_DIALOG_WIDTH_MATCH_PARENT = "KEY_DIALOG_WIDTH_MATCH_PARENT"
 private const val KEY_TITLE_ICON = "KEY_TITLE_ICON"
 private const val KEY_POSITIVE_BUTTON_BACKGROUND_COLOR = "KEY_POSITIVE_BUTTON_BACKGROUND_COLOR"
 private const val KEY_POSITIVE_BUTTON_TEXT_COLOR = "KEY_POSITIVE_BUTTON_TEXT_COLOR"
-private const val KEY_SHOULD_INCLUDE_CHECKBOX = "KEY_SHOULD_INCLUDE_CHECKBOX"
+private const val KEY_SHOULD_SHOW_DO_NOT_ASK_AGAIN_CHECKBOX = "KEY_SHOULD_SHOW_DO_NOT_ASK_AGAIN_CHECKBOX"
+private const val KEY_SHOULD_PRESELECT_DO_NOT_ASK_AGAIN_CHECKBOX = "KEY_SHOULD_PRESELECT_DO_NOT_ASK_AGAIN_CHECKBOX"
 private const val KEY_IS_NOTIFICATION_REQUEST = "KEY_IS_NOTIFICATION_REQUEST"
 private const val DEFAULT_VALUE = Int.MAX_VALUE
 
 internal class SitePermissionsDialogFragment : AppCompatDialogFragment() {
 
-    internal var feature: SitePermissionsFeature? = null
-
-    internal val sessionId: String get() = safeArguments.getString(KEY_SESSION_ID, "")
-
-    internal val title: String get() = safeArguments.getString(KEY_TITLE, "")
-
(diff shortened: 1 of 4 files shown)


dector:
Ooops. Ment to make is draft PR. :unamused:

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsDialogFragment.kt
```diff
@@ -178,6 +185,7 @@ internal class SitePermissionsDialogFragment : AppCompatDialogFragment() {
                 putString(KEY_TITLE, title)
                 putInt(KEY_TITLE_ICON, titleIcon)
                 putBoolean(KEY_SHOULD_INCLUDE_CHECKBOX, shouldIncludeCheckBox)
+                putBoolean(KEY_SHOULD_SELECT_REMEMBER_CHOICE, shouldSelectRememberChoice)
                 putBoolean(KEY_USER_CHECK_BOX, shouldIncludeCheckBox)
```
Amejia481: We have to update the`KEY_USER_CHECK_BOX (userSelectionCheckBox)` with the value of `shouldSelectRememberChoice` because this is the property that stores the user decision. Right now it's using `shouldIncludeCheckBox` value if the user hits the positive button his/her decision will be store even thought the checkbox is not unchecked.

```kotlin
        positiveButton.setOnClickListener {
            feature?.onPositiveButtonPress(sessionId, userSelectionCheckBox)
            dismiss()
      }
```

Amejia481: `shouldIncludeCheckBox`  only gets updated when the user hit checkbox.
  
```kotlin
  containerView.findViewById<CheckBox>(R.id.do_not_ask_again).apply {
            visibility = VISIBLE
            isChecked = checked
            setOnCheckedChangeListener { _, isChecked ->
                userSelectionCheckBox = isChecked
            }
        }
```

components/feature/sitepermissions/src/test/java/mozilla/components/feature/sitepermissions/SitePermissionsDialogFragmentTest.kt
```diff
@@ -53,6 +53,30 @@ class SitePermissionsDialogFragmentTest {
...
+                )
+        )
+
+        doReturn(context).`when`(fragment).requireContext()
+        val dialog = fragment.onCreateDialog(null)
+        dialog.show()
+
+        val checkBox = dialog.findViewById<CheckBox>(R.id.do_not_ask_again)
+
+        assertTrue(checkBox.isVisible())
```
Amejia481: I think we also have to check that `userSelectionCheckBox` gets initialize with the right value.

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsDialogFragment.kt
```diff
@@ -167,7 +173,8 @@ internal class SitePermissionsDialogFragment : AppCompatDialogFragment() {
             titleIcon: Int,
             feature: SitePermissionsFeature,
             shouldIncludeCheckBox: Boolean,
-            isNotificationRequest: Boolean = false
+            isNotificationRequest: Boolean = false,
+            shouldSelectRememberChoice: Boolean = true
```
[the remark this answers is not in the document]
dector: I agree.

Sorry, I'm not aware with internal engineering processes. But I suppose that the combination of two facts:
  - this component is in under-production (1.0) state;
  - you are ready to take responsibility for this change :smile:;
is enough in this case. Will fix it.

components/feature/sitepermissions/src/test/java/mozilla/components/feature/sitepermissions/SitePermissionsDialogFragmentTest.kt
dector: Good point.

dector:
@Amejia481 done. Changes:
  - Rebased with `master`.
  - Added few more tests.
  - Documented public API.
  - Made `dialogConfig` nullable and `shouldPreselectDoNotAskAgain=false` by default.
  - Logically grouped arguments in the top of `SitePermissionsDialogFragment`.
  - Initializing `userSelectionCheckBox` in lateinit-way (as it's not argument but state).

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsDialogFragment.kt
Amejia481: We don't need anymore `KEY_USER_CHECK_BOX`

dector:
> When we use 1) it ends up creating additional objects for just storing a primitive boolean. I think it doesn't worth the cost.
> Maybe this talk from I/O maybe interesting [Kotlin Under the Hood: Understand the Internals](https://www.youtube.com/watch?v=Ta5wBJsC39s) :)

Yep, I'm aware of hidden cost for delegation. Used it is here for few reasons:
  - `lateinit` isn't available for primitives.
  - Defining this variable with default value is more error-prone because it can be used without proper initialization (breaking fail-fast principle). [A]
  - [improvements in latest ART runtime](https://www.youtube.com/watch?v=Zc4JP8kNGmQ) soon be quite widespread. (Chet and Romain are really awesome speakers!).

However in the end of the day I think that I'm agree with your concern/compromise here. Base components should be as fast/efficient as possible and all ambiguity described in [A] should be heavily covered with tests.

Thanks a lot for review! :) :balloon:

Amejia481:
You are more than welcome :) , keep up with the good job!

Issue #2042, opened 2019-02-13
browser-icons: Create disk cache


pocmo:
There are two things we want to cache:

* The mapping from URL to icon URL (if loaded from URL)
* The icon URL to the actual image file.

Fennec stores PNGs. In this new component we may be able to use WEBP since that seems to be supported on our minSdkVersion+.

PR #3001, opened 2019-05-13
Issue #2042: browser-icons: Add disk cache for icons.

pocmo:
This caches the list of resources and the image on disk. There are two improvements over Fennec:
* We store all icon resource URLs now
* We store icons as WEBP which are much smaller than PNGs

This introduces a new third party dependency: "Jake Wharton's" DiskLruCache. It's a standalone version of the DiskLruCache of AOSP. I would like to not have this dependency but it makes no sense to re-implement that and this library is used A LOT: 
https://www.appbrain.com/stats/libraries/details/disklrucache/disk-lru-cache

We could think of moving this dependency into our tree - it's not big. That would at least prevent us from depending on pulling it from jcenter.

### Pull Request checklist


components/browser/icons/src/main/java/mozilla/components/browser/icons/extension/IconMessage.kt
```diff
@@ -0,0 +1,129 @@
...
+    "apple-touch-icon image_src" to IconRequest.Resource.Type.APPLE_TOUCH_ICON,
+    "apple-touch-icon-precomposed" to IconRequest.Resource.Type.APPLE_TOUCH_ICON,
+    "og:image" to IconRequest.Resource.Type.OPENGRAPH,
+    "og:image:url" to IconRequest.Resource.Type.OPENGRAPH,
+    "og:image:secure_url" to IconRequest.Resource.Type.OPENGRAPH,
+    "twitter:image" to IconRequest.Resource.Type.TWITTER,
+    "msapplication-TileImage" to IconRequest.Resource.Type.MICROSOFT_TILE
+)
+
+private fun Map<String, IconRequest.Resource.Type>.reverseLookup(type: IconRequest.Resource.Type): String {
```
Amejia481: Should we return a list here. I noticed on `typeMap` there are multiple keys for one given value, for instance:

`IconRequest.Resource.Type.OPENGRAPH` has ` ["og:image", "og:image:url", "og:image:secure_url"]`

`IconRequest.Resource.Type.APPLE_TOUCH_ICON` has `["apple-touch-icon","apple-touch-icon precomposed","apple-touch-icon image_src"]`

pocmo: The actual value used is not important in this case as long as when parsing it's deserialized to `IconRequest.Resource.Type.OPENGRAPH` again.

pocmo:
Updated the failing test. Now it should pass.

components/browser/icons/src/main/java/mozilla/components/browser/icons/extension/IconMessage.kt
Amejia481: 👍

PR #3014, opened 2019-05-14
Make sure to always initialize the GleanDebugActivity.kt

Dexterp37:
This changes the `launchMode` of the activity to `"singleTask"`, which according to the [docs here](https://developer.android.com/guide/topics/manifest/activity-element),
goes through `onNewIntent` if the activity was already there, rather than reporting `"Warning: Activity not started, intent has been delivered to currently running top-most instance.".`

**What was the problem**: sometimes, while launching the activity from `adb` and the app was already started, we would not go through [`onCreate`](https://github.com/mozilla-mobile/android-components/blob/89ac9428b55cd628afd6e559fde3ce54bd4cee05/components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugActivity.kt#L42), and nothing would happen. Other times, the above warning would be printed. Most of the times, it would go through the onCreate even if the app was open.

**Note**: I'm not entirely sure why the problem was there in the first place. I only happen to be randomly reproducing it *sometimes*. Now I can't randomly reproduce it anymore, but I'm not really confident this is the right fix. If anyone could shed more light on this, it would be great. 

### Pull Request checklist

### DIFF
--- a/components/service/glean/src/main/AndroidManifest.xml
+++ b/components/service/glean/src/main/AndroidManifest.xml
@@ -5,6 +5,7 @@
     package="mozilla.components.service.glean" >
     <application>
         <activity android:name=".debug.GleanDebugActivity"
-            android:exported="true" />
+            android:exported="true"
+            android:launchMode="singleTask" />
     </application>
 </manifest>
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugActivity.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/debug/GleanDebugActivity.kt
@@ -5,6 +5,7 @@
 package mozilla.components.service.glean.debug
 
 import android.app.Activity
+import android.content.Intent
 import android.os.Bundle
 import mozilla.components.service.glean.Glean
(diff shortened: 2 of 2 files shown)


dector:
I'm not fully aware about context of this activity. Are there some cases when it's need to be packaged in the release application?

If it's only here for `debug`/`dev` purposes - we can make it's just another launch activity with special text/icon (to launch it easier). Or setup `IntentService` + `BroadcastReceiver` for launching it.

Dexterp37:
@dector thanks for chiming in!

> I'm not fully aware about context of this activity. Are there some cases when it's need to be packaged in the release application?

This should always be packaged with the release application: it's a tool for QA to perform validation on the telemetry data. The ideal workflow:

- QA engineer downloads the released app from the play store;
- executes the test plan;
- makes sure the data shows up at the end of the pipeline.

dector:
@Dexterp37 Thanks, got it! I think your fix will work. :)

Dexterp37:
Closing this PR as we found that this was really a problem with our docs rather than a bug in the code itself!

Issue #170, opened 2018-05-16
Menu: Add mechanism for dynamically update items


pocmo:
Some menu items will need to update dynamically while they are being shown:

Examples:
* Stop/Reload: In Focus while loading a page there's a "stop" button. If the page has finished loading we show a "reload" button.
* Forward: Whether this button is disabled or not depends on the current state. This state can change while the item is displayed. See #169

NotWoods:
I'm interested in working on this! @csadilek @pocmo

NotWoods:
Closed by #3022

PR #3022, opened 2019-05-15
Closes #169, #170: Dynamically update items

### LINKED ISSUE #169: Menu: Add mechanism to mark items as disabled
Not sure yet whether this is something the item does on its own or whether the menu needs to know about disabled items.

NotWoods:
Adds the ability to dynamically update BrowserMenuItemToolbar buttons by updating an observer. Additionally makes the single state buttons' disabled state controlled through a lambda, similar to the TwoStateButton.

I'm looking for feedback on the API design while I work on integrating the new menu features with other components.

### Pull Request checklist


pocmo:
I had some problems following the code since there's quite some back and forth going on: The ItemToolbar registers to the Button. If I call update() on the Button the ItemToolbar will be notified and then invoke updateView on the Button again with the matching View for that Button. That took a while to understand from looking at the diff. :)

I don't have any specific change requests, just some ideas to think about:

* I wonder if ObserverRegistry/Observable make it harder to follow here. All the Button would need is a dynamically set reference to its parent toolbar - not a dynamic observer mechanism. Would that be simpler?

* For BrowserToolbar we use a simpler approach where we can call invalidate on the toolbar - which will then update the views of all toolbar actions. Similar how Android handles menu items. What makes it simpler is that you invoke it on the top item (the toolbar) and it goes down to the actions and updates the views. In the current implementation here you start at the bottom (the actions) and need to go up to the ItemToolbar (via notifyObservers) and then down again (via updateView()). For BrowserToolbar the amount of views is small enough (and updating a view with the same value is cheap) that this is still performant enough.

* I would like to see this used for the reload/stop button example mentioned in #170.

PR #3037, opened 2019-05-16
Closes #2591: HttpIconLoader: Do not try to load resources that have failed recently.

### LINKED ISSUE #2591: HttpIconLoader: Do not try to download icons that have failed recently
The version of `HttpIconLoader` that is about to land always performs a request to try to download the icon.

The version in Fennec keeps a "failure cache" and will not re-try downloading from an URL that failed before for a certain amount of time.

pocmo:
### Pull Request checklist


components/browser/icons/src/main/java/mozilla/components/browser/icons/loader/HttpIconLoader.kt
```diff
@@ -42,13 +48,50 @@ class HttpIconLoader(
...
+     */
+    fun rememberFailure(url: String) {
+        cache.put(url, now())
     }
+
+    /**
+     * Has loading from this [url] failed previously and recently?
+     */
+    fun hasFailedRecently(url: String) =
+        cache.get(url)?.let { failedAt ->
```
csadilek: One thing I missed in my first pass is that `HttpIconLoader` runs on 3 worker threads. `LruCache` is thread-safe but we'd need to synchronize if we want to run multiple operations atomically. For the use case here I think that is fine though. The consequence is that we could remove a url that is no longer there (no effect) or remove one that failed more recently in case we have mutliple thread running for the same url.

pocmo: Good catch! I agree, in this case we may not need atomic operations.

PR #3042, opened 2019-05-17
#3038 Improving workaround for uploading files

Amejia481:
### Pull Request checklist

### DIFF
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/prompt/GeckoPromptDelegate.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/prompt/GeckoPromptDelegate.kt
@@ -413,15 +413,19 @@ internal class GeckoPromptDelegate(private val geckoEngineSession: GeckoEngineSe
 
     private fun Uri.toFileUri(context: Context): Uri {
         val uri = this
-        val temporalFile = java.io.File(context.cacheDir, uri.getFileName(context))
+        val cacheUploadDirectory = java.io.File(context.cacheDir, "/uploads")
+
+        if (!cacheUploadDirectory.exists()) {
+            cacheUploadDirectory.mkdir()
+        }
+
+        val temporalFile = java.io.File(cacheUploadDirectory, uri.getFileName(context))
         try {
-            val inStream = context.contentResolver.openInputStream(uri) as FileInputStream
-            val outStream = FileOutputStream(temporalFile)
-            val inChannel = inStream.channel
-            val outChannel = outStream.channel
-            inChannel.transferTo(0, inChannel.size(), outChannel)
-            inStream.close()
-            outStream.close()
+            (context.contentResolver.openInputStream(uri) as FileInputStream).use { inStream ->
+                FileOutputStream(temporalFile).use { outStream ->
+                    inStream.copyTo(outStream)
+                }
+            }
         } catch (e: IOException) {
             val logger = Logger("GeckoPromptDelegate")
             logger.warn("Could not convert uri to file uri", e)


components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/prompt/GeckoPromptDelegate.kt
```diff
@@ -413,18 +414,27 @@ internal class GeckoPromptDelegate(private val geckoEngineSession: GeckoEngineSe
...
             val inChannel = inStream.channel
             val outChannel = outStream.channel
             inChannel.transferTo(0, inChannel.size(), outChannel)
-            inStream.close()
-            outStream.close()
         } catch (e: IOException) {
             val logger = Logger("GeckoPromptDelegate")
             logger.warn("Could not convert uri to file uri", e)
+        } finally {
+            inStream?.close()
```
dector: Funny part about `Closeable` is that it [can throw exception](https://docs.oracle.com/javase/7/docs/api/java/io/InputStream.html#close()) as well. :smile: 

What do you think about using [`use`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/use.html) instead (it's analogue to `try-with-resources` in Java)?

Amejia481: 👍

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/prompt/GeckoPromptDelegate.kt
```diff
@@ -413,18 +411,21 @@ internal class GeckoPromptDelegate(private val geckoEngineSession: GeckoEngineSe
...
-            logger.warn("Could not convert uri to file uri", e)
+        val cacheUploadDirectory = java.io.File(context.cacheDir, "/uploads")
+
+        if (!cacheUploadDirectory.exists()) {
+            cacheUploadDirectory.mkdir()
+        }
+
+        val temporalFile = java.io.File(cacheUploadDirectory, uri.getFileName(context))
+        (context.contentResolver.openInputStream(uri) as FileInputStream).use { inStream ->
+            inStream.use {
```
dector: Seems like `inStream` is in use for the second time.
I suppose only two `use` declarations requires (as we have two streams).
Also there still need to catch exception in case if [it'll be thrown](https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/jvm/src/kotlin/io/Closeable.kt#L27).

I thought about something like:
```
try {
    inStream.use {
        outStream.use {
		    // ...
        }
    }
} catch (/*...*/) {
    // ...
}
```

Amejia481: Sorry, I pushed by mistake, I'm update it locally.

Amejia481: Updated :) !

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/prompt/GeckoPromptDelegate.kt
```diff
@@ -413,15 +413,21 @@ internal class GeckoPromptDelegate(private val geckoEngineSession: GeckoEngineSe
...
-            val inChannel = inStream.channel
-            val outChannel = outStream.channel
-            inChannel.transferTo(0, inChannel.size(), outChannel)
-            inStream.close()
-            outStream.close()
+            (context.contentResolver.openInputStream(uri) as FileInputStream).use { inStream ->
+                FileOutputStream(temporalFile).use { outStream ->
+                    val inChannel = inStream.channel
+                    val outChannel = outStream.channel
+                    inChannel.transferTo(0, inChannel.size(), outChannel)
```
pocmo: nit: Kotlin also has [copyTo](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/java.io.-input-stream/copy-to.html) which you can use directly on the streams without having to deal with channels.

Amejia481: Awesome 👍

Issue #2346, opened 2019-03-13
Enable kotlin warningsAsErrors on remaining modules that its disabled on


mcomella:
In #54, I enabled kotlin warnings as errors on modules that already had no errors. For each disabled module, we should:
- Fix the warnings
- Remove it from the list of modules with warnings as errors disabled

When we empty the list, don't forget to enable the check to selectively enable warnings as errors.

mcomella:
To fix a module locally, you should remove it from the list of suppressed modules in KotlinCompiler.kt, compile kotlin for the production code, the unit test code, and the android test code, e.g. for all modules:
```sh
./gradlew compileDebugKotlin compileDebugUnitTest compileDebugAndroidTest
```

Then fix all the errors.

mcomella:
@pocmo fwiw, as a downstream consumer, I'd prefer this to happen sooner than later: there are issues the compiler can catch _right now_, before they have a chance to crash our app in production – it'd be great to enforce this at least on the most popular modules like `browser-session` and `browser-engine-*`.

pocmo:
Landed fixes for:
* browser-session
* concept-engine
* browser-engine-gecko-nightly

Amejia481:
@pocmo Is there a way to pull [this list](https://github.com/mozilla-mobile/android-components/blob/c3cae262766d1f288bab81ac6f9db9ffd5d33dbd/buildSrc/src/main/java/KotlinCompiler.kt#L15) from [.buildconfig.yml](https://github.com/mozilla-mobile/android-components/blob/master/.buildconfig.yml)?

pocmo:
@Amejia481 Yeah, would be an option. But I don't think it's worth changing the code to a configuration flag in .buildconfig.yml since the goal is for this list to no longer exist. :)

mcomella:
> Landed fixes for:

🙌 Thanks for working on this so quickly! :)

pocmo:
Additionally off the list now:
* `browser-engine-gecko-beta` (Thanks merge day!)
* `support-base`
* `support-ktx`
* `lib-crash`

dector:
Taken:
```
browser-domains
support-utils
tooling-lint
ui-autocomplete
ui-progress
ui-tabcounter
```

dector:
Fixes for all modules are on review. :)

pocmo:
This is finally done. Thank you for all your hard work @dector! 💯

PR #3043, opened 2019-05-17
[★☆☆] For #2346. Add `testContext` providing for tests.

dector:
### Issue #2346

Delegate for providing application context into unit tests.

### Complexity
Easy (★☆☆)

### Motivation

Increase code readability.

_Because of relatively big amount of changed in context of this issue I've decided to commit this dependency as a separate PR. Will use it in other commits for this issue._

### Pull Request checklist

### DIFF
--- a/components/support/test/build.gradle
+++ b/components/support/test/build.gradle
@@ -40,6 +40,8 @@ dependencies {
     }
     testImplementation Dependencies.testing_maven_ant_tasks
 
+    implementation Dependencies.androidx_test_core
+
     testImplementation Dependencies.androidx_core
     testImplementation project(':support-ktx')
 }
--- a/components/support/test/src/main/java/mozilla/components/support/test/robolectric/Extensions.kt
+++ b/components/support/test/src/main/java/mozilla/components/support/test/robolectric/Extensions.kt
@@ -1,0 +1,10 @@
+package mozilla.components.support.test.robolectric
+
+import android.content.Context
+import androidx.test.core.app.ApplicationProvider
+
+/**
+ * Provides application context for test purposes
+ */
+inline val testContext: Context
+    get() = ApplicationProvider.getApplicationContext()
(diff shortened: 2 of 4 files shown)


components/support/test/src/test/java/mozilla/components/support/test/robolectric/DelegatesTest.kt
```diff
@@ -0,0 +1,24 @@
...
+import androidx.test.core.app.ApplicationProvider
+import org.junit.Assert
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
+class DelegatesTest {
+
+    private val context by providedAppContext()
```
pocmo: I was thinking about something similar some time ago. My idea was to add a `context` extension property to `Any` so that it can be used in tests without declaring it every time. What do you think?

dector: Sounds interesting and I like the idea but I'm affraid that implicit `val Any.context` can be misleading sometimes. There are cases when we need/want to mock methods in `Context` (e.g. mocking resources manager). But we can make it more explicit with `fun Any.context()`. Method seems like explicit way to say "provide me context" and it will not intersect with _possible_ local `context` variable.

pocmo: Yeah, we can even make it more descriptive and name it `testContext` or similar.

dector: Awesome! I like it. :+1:

dector: Done.

pocmo:
The commit message may need an update now :)

dector:
Oh, my bad.

dector:
CI check needs restart due to `[taskcluster:error] Pulling docker image has failed.`

dector:
Please check latest changes:
  - Moved changelog entry to `0.54.0` since `0.53.0` was released.
  - Made `testContext` as extension property. Name speaks for itself so it should be safe implement it as you've suggested.

pocmo:
Some other changes landed in-between and now we need to fix a changelog conflict here :)

dector:
Got it. Give me 5 minutes.

pocmo:
Thank you!

dector:
Now my hands are untied. :D

PR #3060, opened 2019-05-21
Update GleanDebugActivity docs regarding `sendPing` flag [ci skip]

travis79:
Since the `sendPing` flag does not allow for things like `duration` to be calculated and added to the ping since the ping is sent before application dependent things like `duration` are started, the documentation needed to be updated to explicitly indicate this was the actual behavior of the `GleanDebugActivity`.

### Pull Request checklist

### DIFF
--- a/components/service/glean/README.md
+++ b/components/service/glean/README.md
@@ -180,9 +180,13 @@ adb shell am start -n org.mozilla.samples.glean/mozilla.components.service.glean
   --es tagPings test-metrics-ping
 ```
 
-### Important GleanDebugActivity note!
+### Important GleanDebugActivity notes!
 
-Options that are set using the adb flags are not immediately reset and will persist until the application is closed or manually reset.
+- Options that are set using the adb flags are not immediately reset and will persist until the application is closed or manually reset.
+
+- There are a couple different ways in which to send pings through the GleanDebugActivity.
+    1. You can use the `GleanDebugActivity` in order to tag pings and trigger them manually using the UI.  This should always produce a ping with all required fields.
+    2. You can use the `GleanDebugActivity` to tag _and_ send pings.  This has the side effect of potentially sending a ping which does not include all fields because `sendPings` triggers pings to be sent before certain application behaviors can occur which would record that information.  For example, `duration` is not calculated or included in a baseline ping sent with `sendPing` because it forces the ping to be sent before the `duration` metric has been recorded.
 
 ## Data documentation
 


components/service/glean/README.md
```diff
@@ -180,9 +180,11 @@ adb shell am start -n org.mozilla.samples.glean/mozilla.components.service.glean
   --es tagPings test-metrics-ping
 ```
 
-### Important GleanDebugActivity note!
+### Important GleanDebugActivity notes!
 
-Options that are set using the adb flags are not immediately reset and will persist until the application is closed or manually reset.
+- Options that are set using the adb flags are not immediately reset and will persist until the application is closed or manually reset.
+
+- Pings sent using the `sendPing` flag are **NOT** expected to follow the normal application behavior.  For example, `duration` is not calculated or included in a ping sent with `sendPing` because it forces the ping to be sent before the duration is started.
```
mdboom: Are there other (known) things that are different?  Maybe we should call them all out?  I feel like this paragraph sort of assumes the reader knows what normal application behavior is.  Maybe better to just say: "The following fields will have different data than under normal circumstances..."

Dexterp37: I agree with you, Mike, this assumes reader knows about "what's the expected behaviour". I think we should tackle it differently: we can point out the 2 main scenarios that can be triggered using this. I've highlighted the [two ways of testing](https://bugzilla.mozilla.org/show_bug.cgi?id=1549982#c2) in the bug, maybe it's worth rephrasing them and adding them here? Then we could say that depending on the ping type and its behaviour, data *may* be missing if using the debug activity for tagging and sending pings. Then we could call out, for example, that duration will be missing for reason XYZ when triggering the 'baseline' ping. What do you think?

travis79: `duration` was the only thing I found so far that was obviously missing.  If this was first run, then maybe other things might be missing as well?

Dexterp37: > 
> 
> `duration` was the only thing I found so far that was obviously missing. If this was first run, then maybe other things might be missing as well?

In [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1552798), we had a similar problem: the `metrics` ping was being triggered right at startup, with no data potentially available.

travis79: I'd say it bears a quick experiment to see if I can find any missing fields, but I think it may be possible if the `initializeCoreMetrics` hasn't been called yet.

travis79: Nevermind, it's been a long week away.  `GleanDebugActivity` checks that, so it shouldn't be possible to run the Intent without it being initialized.

Dexterp37: > 
> 
> I'd say it bears a quick experiment to see if I can find any missing fields, but I think it may be possible if the `initializeCoreMetrics` hasn't been called yet.

I would personally say that this doesn't matter *too much*: the different testing modes will be still valid with custom pings as well, so here we might want to be as general as possible (still calling out one example for things we own). No need to spend too much time on it, IMHO ;)

travis79: Okay, I've attempted to reword this to capture what @Dexterp37 left in the bugzilla comment.  Please take another look as you have time.

PR #3069, opened 2019-05-21
Closes #3026: Add fontSizeFactor and fontInflationEnabled settings

### LINKED ISSUE #3026: Surface Text Size Related Settings from GV Runtime Settings
For https://github.com/mozilla-mobile/fenix/issues/1286 I will need `setFontSizeFactor` surfaced so I can dynamically change this value.

csadilek:
/cc @ekager, just a heads-up that `automaticFontSizeAdjustment` needs to be set to false to use `fontSizeFactor` (we already expose that setting with default=true).


ekager:
Thank you! ♥

PR #3083, opened 2019-05-22
For #2346. Enable kotlin warningsAsErrors for `browser-domains` module.

dector:
### Issue #2346

  - Refactored (a bit) `ShippedDomainsProvider` and `CustomDomainsProvider`. Using
composition instead of inheritance. Was needed to cover with unit tests functionality in this classes. `Domains` loading already has tests, so `ShippedDomainsProvider` and `CustomDomainsProvider` can be represented as a building function if needed.
  - Add tests for `BaseDomainAutocompleteProvider`.

### Complexity

Easy+ (★☆☆)

### Pull Request checklist


docs/changelog.md
```diff
@@ -12,6 +12,10 @@ permalink: /changelog/
 * [Gecko](https://github.com/mozilla-mobile/android-components/blob/master/buildSrc/src/main/java/Gecko.kt)
 * [Configuration](https://github.com/mozilla-mobile/android-components/blob/master/buildSrc/src/main/java/Config.kt)
 
+* **browser-domains**
+  * New domain autocomplete providers `ShippedDomainsProvider` and `CustomDomainsProvider` that
+    should be used instead of deprecated `DomainAutoCompleteProvider`.
```
pocmo: nit: This needs to be moved to the latest version now. :)

components/browser/domains/src/main/java/mozilla/components/browser/domains/autocomplete/Providers.kt
```diff
@@ -23,40 +23,40 @@ enum class DomainList(val listName: String) {
...
-    DomainAutocompleteProvider {
-    internal val scope = CoroutineScope(Dispatchers.IO)
-    // We compute 'domains' on the worker thread; make sure it's immediately visible on the UI thread.
-    @Volatile
+open class BaseDomainAutocompleteProvider(
+    private val list: DomainList,
+    private val domainsLoader: DomainsLoader
+) : DomainAutocompleteProvider, CoroutineScope by CoroutineScope(Dispatchers.IO) {
+
     var domains: List<Domain> = emptyList()
```
pocmo: Did you intentionally remove the `@Volatile` annotation here? With reading/writing the list from different threads it looks like we still need it?

docs/changelog.md
dector: It is in `0.54.0-SNAPSHOT`. :)

components/browser/domains/src/main/java/mozilla/components/browser/domains/autocomplete/Providers.kt
dector: Yes, I've removed it intentionally but I was wrong doing this. :)
Will return it back.

dector:
Moved changelog entry to 0.55.0

Issue #1292, opened 2018-11-07
Sample apps: Disable "release" build types


pocmo:
We did that for the sample browser but not for the other sample apps. There's no need to build a debug and a release type for all of them.

Shubhamurkade:
Hi @pocmo, Can I take this up?

pocmo:
@Shubhamurkade Sure! Let us know if you have any questions.

Shubhamurkade:
Thank you!

PR #3089, opened 2019-05-22
#1292: Disable release build types for sample apps

Shubhamurkade:
Modified the following files to disable release builds for various sample apps:
1. /samples/crash/build.gradle
2. /samples/dataprotect/build.gradle
3. /samples/firefox-accounts/build.gradle
4./samples/sync-logins/build.gradle
5. /samples/sync/build.gradle
6./samples/toolbar/build.gradle

### DIFF
--- a/samples/crash/build.gradle
+++ b/samples/crash/build.gradle
@@ -25,6 +25,14 @@ android {
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
         }
     }
+
+    variantFilter { variant ->
+        if (variant.buildType.name == "release") {
+            // This is a sample app that we are not releasing. Save some time and do not build
+            // release versions.
+            setIgnore(true)
+        }
+    }
 }
 
 
--- a/samples/dataprotect/build.gradle
+++ b/samples/dataprotect/build.gradle
@@ -21,6 +21,14 @@ android {
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
         }
     }
+
+    variantFilter { variant ->
+        if (variant.buildType.name == "release") {
+            // This is a sample app that we are not releasing. Save some time and do not build
+            // release versions.
+            setIgnore(true)
+        }
+    }
 }
 
 dependencies {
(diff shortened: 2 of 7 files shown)


jonalmeida:
Looks like there are build failures for the sample apps:
```
* What went wrong:
Task 'lintRelease' not found in project ':samples-crash'.
```

From the [`decision_task.py`](https://github.com/mozilla-mobile/android-components/blob/master/automation/taskcluster/decision_task.py#L116-L117), we're running `lintRelease` for all tasks so we don't lint all variants.

We probably need to override this to check for the release variant otherwise, use `lintDebug`? This might be a @pocmo question.

Shubhamurkade:
> Looks like there are build failures for the sample apps:
> 
> ```
> FAILURE: Build failed with an exception.
> * What went wrong:
> Task 'lintRelease' not found in project ':samples-crash'.
> ```
> 
> From the [`decision_task.py`](https://github.com/mozilla-mobile/android-components/blob/master/automation/taskcluster/decision_task.py#L116-L117), we're running `lintRelease` for all tasks so we don't lint all variants.
> 
> We probably need to override this to check for the release variant otherwise, use `lintDebug`? This might be a @pocmo question.

Hi @pocmo. Can you help with this query?

samples/crash/build.gradle
```diff
@@ -25,6 +25,14 @@ android {
...
         }
     }
+
+    variantFilter { variant ->
+        if (variant.buildType.name == "release") {
+            // This is a sample app that we are not releasing. Save some time and do not build
+            // release versions.
+            setIgnore(true)
+        }
+    }
```
pocmo: Instead of adding this to all the module `build.gradle` files it would be nice to do that in the root `build.gradle` for **app** modules only - there's already a block that configures all modules.

pocmo:
> We probably need to override this to check for the release variant otherwise, use lintDebug? This might be a @pocmo question.

Ah, mh. Tricky. `lintDebug` may be an option - however we picked `lintRelease` specifically because we want to lint what we release. If lint has really become faster (what they promised) then we could consider going back to just `lint`.

PR #3090, opened 2019-05-22
Expand labeled metrics to allow for dotted snake-case labels.

travis79:
This should **NOT** be merged until [this](https://github.com/mozilla/glean_parser/pull/69) and [this](https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/335) have landed

### Pull Request checklist

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/private/LabeledMetricType.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/private/LabeledMetricType.kt
@@ -37,7 +37,24 @@ data class LabeledMetricType<T>(
     companion object {
         private const val MAX_LABELS = 16
         private const val OTHER_LABEL = "__other__"
-        private val labelRegex = Regex("^[a-z_][a-z0-9_]{0,29}$")
+
+        // This regex is used for matching against labels and should allow for dots and/or
+        // snake_case.
+        private val labelRegex = Regex("^[a-z_][a-z0-9_]{0,29}(\\.[a-z0-9_]{0,29})*$")
+        // Some examples of good and bad labels:
+        //
+        // Good:
+        //   this.is.fine
+        //   this_is_fine_too
+        //   this.is_still_fine
+        //   thisisfine
+        //   this.is_fine.2
+        // Bad:
+        //   this.is.not_fine_due_tu_the_length_being_too_long_i_thing.i.guess
+        //   1.not_fine
+        //   this.is-not-fine
+        //   this.$isnotfine
+
         private const val MAX_LABEL_LENGTH = 30
     }
 
@@ -102,7 +119,7 @@ data class LabeledMetricType<T>(
                     recordError(
                         this,
                         ErrorType.InvalidLabel,
-                        "label must be snake_case, got '$label'",
+                        "label must be dotted snake_case, got '$label'",
                         logger
                     )
                     return OTHER_LABEL
--- a/components/service/glean/scripts/sdk_generator.gradle
+++ b/components/service/glean/scripts/sdk_generator.gradle
@@ -20,7 +20,7 @@ import groovy.json.JsonOutput
 // so that it will be shared between all libraries that use Glean.  This is
 // important because it is approximately 300MB in installed size.
 
-String GLEAN_PARSER_VERSION = "0.27.1"
+String GLEAN_PARSER_VERSION = "0.28.1"
 // The version of Miniconda is explicitly specified.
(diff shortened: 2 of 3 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/private/LabeledMetricType.kt
```diff
@@ -37,7 +37,7 @@ data class LabeledMetricType<T>(
     companion object {
         private const val MAX_LABELS = 16
         private const val OTHER_LABEL = "__other__"
-        private val labelRegex = Regex("^[a-z_][a-z0-9_]{0,29}$")
+        private val labelRegex = Regex("^[a-z_][a-z0-9_]{0,29}(\\\\.[a-z0-9_]{0,29})*\$")
```
mdboom: That very last `\` seems fishy to me.  It's escaping the `$` which *maybe* we need to do since `$` has special meaning in Kotlin strings, but it wasn't escaped before...

Maybe to be sure about this and other reasons, we should add a test [around here](https://github.com/mozilla-mobile/android-components/blob/master/components/service/glean/src/test/java/mozilla/components/service/glean/private/LabeledMetricTypeTest.kt#L256) that sets a label with some dots in it and confirms it gets recorded?

travis79: The IDE actually inserted that in there when I copied the string verbatim from the glean_parser changes I just made.  I trusted it knew what it was doing and left it there.  I know that the `$` can denote an interpolated string so I'm guessing that is why it was escaped.  I was planning on adding some test cases around this already, which is why I left this as a Draft PR while I was working on the tests.

travis79: Turns out that _all_ of the escaping that the IDE so _thoughtfully_ added for me was incorrect.  I had to remove the escaped `$` as well as the escaped `\\` in order to get this to work.  So literal of `"^[a-z_][a-z0-9_]{0,29}(\\.[a-z0-9_]{0,29})*$"` was what it needed.

mdboom: :musical_note: Sometimes help is the kind of help, we all could do without :musical_note:

components/service/glean/src/main/java/mozilla/components/service/glean/private/LabeledMetricType.kt
```diff
@@ -37,7 +37,7 @@ data class LabeledMetricType<T>(
     companion object {
         private const val MAX_LABELS = 16
         private const val OTHER_LABEL = "__other__"
-        private val labelRegex = Regex("^[a-z_][a-z0-9_]{0,29}(\\\\.[a-z0-9_]{0,29})*\$")
+        private val labelRegex = Regex("^[a-z_][a-z0-9_]{0,29}(\\.[a-z0-9_]{0,29})*$")
```
Dexterp37: nit: can we add a comment above this with a couple of examples of valid labels?

components/service/glean/src/main/java/mozilla/components/service/glean/private/LabeledMetricType.kt
badboy: This regex makes it unbounded. There's no length limit enforced anymore at all. `*` will _always_ match.

mdboom: Yes, but the total length limit is enforced elsewhere in code.  We still have the benefit that any one section has a max of 29 characters.

travis79:
> Does this require changing the parser's version so that [this](https://github.com/mozilla/glean_parser/pull/69) is pulled in? If so, before merging this, can we test that everything works ok with the new parsers?

Since this technically just expands what is accepted by the regex, the previous more restricted version of the label names should all pass as a subset of this new regex, which is why I'm not too worried about trying to synchronize the releases.  I would like to land them fairly closely since the parser would technically allow a label name that the client wouldn't, so yes, I was planning on waiting to get the updated parser version and roll it in with this.

I am also waiting on confirmation that the pipeline-schema changes are deployed before I let this and the glean_parser changes land to ensure that we can accept data before anyone runs with this change and potentially loses any data.

travis79:
@mdboom Would you care to take another look when you get a chance.  This is still blocked on your review.  Thanks!

PR #3092, opened 2019-05-22
Record Sync telemetry in Glean pings

linacambridge:
This PR adds two Glean pings to the Storage-Sync browser component:
one for history, and one for bookmarks. Both pings record the same set
of base metrics, including incoming and outgoing counts, sync duration,
the hashed FxA UID, and, most importantly, the failure reason if the
store fails to sync. The bookmarks ping records additional validation
data.

The Glean schema is a flattened version of the Sync ping that's
currently sent on Firefox Desktop and Firefox for iOS. Instead of
sending a single ping with multiple syncs, each having multiple
engines, we send one Glean ping per engine per sync.

The pings are recorded directly in the component, so any app that
consumes it should get pings for free.

This depends on a not-yet-released version of a-s, so the tests won't pass until then.

### Pull Request checklist


components/service/glean/src/main/java/mozilla/components/service/glean/error/ErrorRecording.kt
```diff
@@ -68,7 +68,7 @@ object ErrorRecording {
 
         logger.warn("${metricData.identifier}: $message")
 
-        // There are two reasons for using `CountersStorageEngine.record` below
+        // There are two reasons for using `CountersStorageEngine.add` below
```
travis79: Let's keep the discussion in one place, I started a review [here](https://github.com/mozilla-mobile/fenix/pull/2749#discussion_r286678555)

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -358,7 +358,7 @@ open class GleanInternalAPI internal constructor () {
      * @param pings List of pings to send.
      * @return The async Job performing the work of assembling the ping
      */
-    internal fun sendPings(pings: List<PingType>) = Dispatchers.API.launch {
+    fun sendPings(pings: List<PingType>) = Dispatchers.API.launch {
```
Dexterp37: Why is this change needed?

components/service/glean/src/main/java/mozilla/components/service/glean/private/CounterMetricType.kt
```diff
@@ -28,6 +28,26 @@ data class CounterMetricType(
 
     private val logger = Logger("glean/CounterMetricType")
 
+    /**
+     * ...
+     *
+     * @param value This is the new value for the counter.
+     */
+    fun set(value: Int) {
```
Dexterp37: As discussed [in the other PR](https://github.com/mozilla-mobile/fenix/pull/2749#discussion_r286814929), I don't think we need to change this. We can rely on the metrics being cleared after pings are sent :) If that doesn't happen, it's a bug on our end!

components/service/glean/src/main/java/mozilla/components/service/glean/storages/CountersStorageEngine.kt
```diff
@@ -42,13 +42,36 @@ internal open class CountersStorageEngineImplementation(
     }
 
     /**
-     * Record a string in the desired stores.
+     * Set the counter to the given value, overwriting any existing value.
+     *
+     * @param metricData object with metric settings
+     * @param value the integer value to set.
+     */
+    fun set(
```
Dexterp37: I think this can be dropped as well :)

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
linacambridge: Reverted, it was before I figured out how to send a custom ping. 😄

components/browser/storage-sync/pings.yaml
```diff
@@ -0,0 +1,24 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+$schema: moz://mozilla.org/schemas/glean/pings/1-0-0
+
+history_sync:
+  description: >
```
Dexterp37: heads-up: please note that once [this](https://github.com/mozilla/glean_parser/pull/70) is merged, a `data_reviews` field will be required for each ping as well. You can be future proof by adding this already, so you won't need to change it later.

linacambridge:
OK, I think this is ready for a first round of review! Thanks for all your help with this, @Dexterp37! I can't assign reviewers, so paging @travis79 and @liuche, too. ☎️

For more context, the a-s PR is mozilla/application-services#1112. The existing Sync ping is also documented [here](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/data/sync-ping.html); this pulls those fields into Glean pings.

components/service/glean/src/main/java/mozilla/components/service/glean/error/ErrorRecording.kt
linacambridge: I moved the pings into a-c, so that PR can be closed. Thanks for all the feedback there!

components/browser/storage-sync/metrics.yaml
```diff
@@ -0,0 +1,257 @@
...
+    send_in_pings:
+    - bookmarks_sync
+    bugs:
+    - 3092
+    data_reviews:
+    - https://github.com/mozilla-mobile/android-components/pull/3092
+    notification_emails:
+    - sync-core@mozilla.com
+    expires: never
+  remote_tree_problems:
```
linacambridge: This is another case where recording an arbitrary timestamp via `NumberMetricType` would be nifty—in addition to recording validation problems, we'd also like to record how long validation, and each phase of the merge, took.

That's not something we can do with timespans today, because the Rust code measures that internally. Calling into Glean via the FFI from Rust definitely wouldn't be worth it for this one case.

We don't need it for the Fenix MVP; `startedAt`, `finishedAt`, and Desktop + iOS telemetry via the Sync ping is enough for now. But it's something we'd like to collect eventually. 😄

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/Connection.kt
```diff
@@ -22,14 +23,14 @@ const val DB_NAME = "places.sqlite"
  *
  * Writer is always the same, as guaranteed by [PlacesApi].
  */
-interface Connection : Closeable {
+internal interface Connection : Closeable {
```
linacambridge: `sync15` is an `implementation` dependency, so I don't think I can export this. Since this will likely change with the sync manager, I opted to make the interface `internal` instead of exposing a public type wrapping `SyncTelemetryPing`.

components/browser/storage-sync/metrics.yaml
```diff
@@ -0,0 +1,257 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+$schema: moz://mozilla.org/schemas/glean/metrics/1-0-0
+
+# The `history_sync` and `bookmarks_sync` metrics use the same structure, but, since are
+# sent in separate pings, after each sync. This means we submit one ping per
+# store per sync.
```
linacambridge: It's unfortunate that `history_sync` and `bookmarks_sync` duplicate all the base fields. I thought of having a separate `sync` category, and changing `send_in_pings` to send in both, then send one ping after the other...but the order of individual store (bookmarks and history) syncs within a `StorageSync::sync` operation [isn't guaranteed](https://mozilla.slack.com/archives/CA4JKKLSG/p1559338609037400), and I'm worried that'll interleave counts for both stores.

components/browser/storage-sync/src/test/java/mozilla/components/browser/storage/sync/PlacesHistoryStorageTest.kt
```diff
@@ -485,12 +486,13 @@ class PlacesHistoryStorageTest {
                 return mock()
             }
 
-            override fun syncHistory(syncInfo: SyncAuthInfo) {
+            override fun syncHistory(syncInfo: SyncAuthInfo): SyncTelemetryPing? {
```
linacambridge: You can tell I didn't add tests yet, because this won't compile. 😅

components/browser/storage-sync/metrics.yaml
Dexterp37: I wonder if the timeline allows you to wait for the potential Rust API to come out? (no ETA for it yet, but I suspect somewhen 2020).

components/browser/storage-sync/metrics.yaml
```diff
@@ -0,0 +1,257 @@
...
+    bugs:
+    - 3092
+    data_reviews:
+    - https://github.com/mozilla-mobile/android-components/pull/3092
+    notification_emails:
+    - sync-core@mozilla.com
+    expires: never
+  remote_tree_problems:
+    type: labeled_counter
+    version: 3
```
Dexterp37: nit: any reason you're starting from 3 here?

components/browser/storage-sync/src/test/java/mozilla/components/browser/storage/sync/PlacesHistoryStorageTest.kt
Dexterp37: Have you considered writing unit tests for your metrics? Glean offers a [testing API](https://github.com/mozilla-mobile/android-components/blob/master/components/service/glean/docs/metrics/testing-metrics.md), which would make your life easier for validating assumptions on the metrics you're landing

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesBookmarksStorage.kt
```diff
@@ -178,11 +182,63 @@ open class PlacesBookmarksStorage(context: Context) : PlacesStorage(context), Bo
...
+        sync.engines.forEach eachEngine@{ engine ->
+            if (engine.name != "bookmarks") {
+                return@eachEngine
+            }
+            BookmarksSync.apply {
+                val base = BaseGleanSyncPing.fromEngineInfo(ping.uid, engine)
+                uid.set(base.uid)
+                startedAt.set(base.startedAt)
+                finishedAt.set(base.finishedAt)
+                if (base.applied > 0) {
```
Dexterp37: nit: I'd drop a comment here specifying that, since all these metrics have `ping` lifetime, they get cleared after the ping is generated, so it's ok to simply add to them.

components/browser/storage-sync/metrics.yaml
Dexterp37: We filed a bug about adding an API to record arbitrary timespans [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1556341).

components/browser/storage-sync/metrics.yaml
linacambridge: Hmm, so if we have a `metrics.yaml` that looks like this:

```yaml
sync:
  outgoing:
    type: labeled_counter
    labels:
    - uploaded
    - failed_to_upload
  send_in_pings:
  - history_sync
  - bookmarks_sync
```

And we do this:

```kotlin
GleanMetrics.Sync.outgoing["uploaded"].add(historyPingFromRust.uploaded)
GleanMetrics.Pings.historySync.send()
```

And this, potentially before `Pings.historySync.send()` is called:

```kotlin
GleanMetrics.Sync.outgoing["uploaded"].add(bookmarksPingFromRust.uploaded)
GleanMetrics.Pings.bookmarksSync.send()
```

How does Glean know that the second call to `Sync["uploaded"].add()` is meant for bookmarks, not history, if we haven't sent the history ping yet?

components/browser/storage-sync/metrics.yaml
linacambridge: Whoops, I meant for this to be 2, not 3. 😄 The reason is that Desktop (and iOS) will collect validation telemetry v2 in the Sync ping as of [bug 1552621](https://bugzilla.mozilla.org/show_bug.cgi?id=1552621)...but, thinking about it some more, this is a completely different ping, so there's no reason to make the Glean version match Desktop's.

components/browser/storage-sync/metrics.yaml
travis79: Hi Lina, jumping in here since @Dexterp37 is already past his EOD.

`history_sync` and `bookmarks_sync` would each have a copy of the same `outgoing` counter.  So in the case you mentioned, both of the stores/pings would increment whenever the `add()` method is called.  In essence, you cannot specifically `add()` to one or the other.  You could end up in situations where the counts were different, depending on when the pings were sent, but you should never be able to add to one and not the other.

In this case, I think what you may want to do is separate them at the metrics level something like this:

```yaml
sync.history:
  outgoing:
    type: labeled_counter
    labels:
    - uploaded
    - failed_to_upload
    send_in_pings:
    - history_sync
sync.bookmarks:
  outgoing:
    type: labeled_counter
    labels:
    - uploaded
    - failed_to_upload
    send_in_pings:
    - bookmarks_sync
```

This keeps the counters separate since they are now treated as unique metrics, and they are then able to be sent in separate pings.  The only difference is now you have two separate metrics that are no longer dependent, and the way you call it changes slightly (notice the additional `.history` in the object hierarchy):
```Kotlin
GleanMetrics.Sync.history.outgoing["uploaded"].add(historyPingFromRust.uploaded)
```

linacambridge: Thanks @travis79! Sounds like I'll still need to have the same duplication as in this patch, just nested under `sync`, instead of making them separate top-level metrics.

components/browser/storage-sync/metrics.yaml
linacambridge: Thanks! I think the pings we get from Desktop and iOS with the information should be enough to tide us over until the Rust Glean API is ready.

components/browser/storage-sync/metrics.yaml
Dexterp37: > 
> 
> Thanks @travis79! Sounds like I'll still need to have the same duplication as in this patch, just nested under `sync`, instead of making them separate top-level metrics.

Yes, sorry about the noise, I misunderstood the problem and looks like Travis' solution or your current solution are the correct ones.

components/browser/storage-sync/metrics.yaml
Dexterp37: Yup :) Versioning it's useful on our hand in case you need to change the labels , but since this is a different ping/format, you can simply remove so we can assume the default, for now!

linacambridge: Will do, thanks!

components/browser/storage-sync/src/test/java/mozilla/components/browser/storage/sync/PlacesHistoryStorageTest.kt
linacambridge: With help from @travis79 and @mdboom, I got tests working using a mock server. Unfortunately, I couldn't use `testGetValue` because I'd need to change the interface to allow intercepting the ping before it was submitted. 😞

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesHistoryStorage.kt
```diff
@@ -165,11 +171,69 @@ open class PlacesHistoryStorage(context: Context) : PlacesStorage(context), Hist
...
+                    outgoing["failed_to_upload"].add(base.failedToUpload)
+                }
+                if (base.outgoingBatches > 0) {
+                    outgoingBatches.add(base.outgoingBatches)
+                }
+                base.failureReason?.let {
+                    recordFailureReason(it)
+                }
+            }
+            Pings.historySync.send()
```
Dexterp37: After reading the discussion in Slack, I realized that we shouldn't need to trigger the ping at all. Coming back to this code, I think there's an alternative solution: what if refactored `Pings.historySync.send()` to a separate, `internal` function, something like this:

```
internal fun submitHystorySyncPing() {
  Pings.historySync.send()
}
```

Then in tests, since you're testing each engine in isolation, you could probably do something like:

```
// TODO: mock submitHystorySyncPing

// Ensure that the ping submission code was called. 
verify(storage).submitHystorySyncPing()

// TODO: use the testing api
```

Alternatively, you could also change `MockingPlacesHistoryStorage` to accept a callback function and call that instead of `submitHystorySyncPing`.

components/browser/storage-sync/src/test/java/mozilla/components/browser/storage/sync/PlacesBookmarksStorageTest.kt
```diff
@@ -263,5 +276,108 @@ class PlacesBookmarksStorageTest {
         verify(writer, times(1)).deleteBookmarkNode(newSeparator.guid)
     }
 
+    @Test
+    fun `build and submit telemetry pings on success`() = runBlocking {
+        val pingServer = getMockWebServer()
```
Dexterp37: I think we don't really need to do this dance, see my comments above. The only requirement I see, is that we have the possibility of calling the testing API before the ping is actually sent. This could be achieved without touching the Glean internals, by slightly refactoring the implementation. What do you think, @linacambridge ?

components/service/glean/src/test/java/mozilla/components/service/glean/TestUtil.kt
```diff
@@ -122,25 +124,11 @@ internal fun resetGlean(
...
-        // The experiments storage engine needs to be cleared manually as it's not listed
-        // in the `StorageEngineManager`.
-        ExperimentsStorageEngine.clearAllStores()
-    }
-
-    // Init Glean.
-    Glean.initialized = false
-    Glean.setUploadEnabled(true)
-    Glean.initialize(context, config)
+    TestUtils.resetGlean(context, config, clearStores)
```
Dexterp37: Please, let's not do this if possible :)

components/service/glean/src/main/java/mozilla/components/service/glean/utils/TestUtils.kt
```diff
@@ -0,0 +1,93 @@
...
+
+object TestUtils {
+    /**
+     * Resets the Glean state and trigger init again.
+     *WorkManagerTestInitHelper
+     * @param context the application context to init Glean with
+     * @param config the [Configuration] to init Glean with
+     * @param clearStores if true, clear the contents of all stores
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
```
Dexterp37: I'd rather not increase our API surface, if possible. Looks like your use-case could be satisfied by slightly refactoring the code, as described above. I'm happy to have a meeting to further discuss this! I filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1556985) for providing an example. Once again, thank you very much for helping us on this.

travis79:
> Thank you Lina for following up on this. I think I have an actionable proposal to make testing code a bit simpler and not rely on Glean SDK exposing more than required. Let me know what you think about that (I couldn't test it locally: is this PR requiring some other change somewhere else?).

After looking at @Dexterp37's suggestion, I have to say that I agree with him that mocking would be much better solution for what you are trying to do.

components/service/glean/src/main/java/mozilla/components/service/glean/utils/TestUtils.kt
mdboom: I think this function in particular might need to be public-for-testing though.   Being able to clear the stores between tests feels pretty required...

linacambridge: In the latest patch, I factored out the logic to reset Glean into a test-only `Glean.reset` function, and changed `resetGlean` in the Glean and Sync tests to call it after enabling testing and initializing the test work manager.

components/service/glean/src/test/java/mozilla/components/service/glean/TestUtil.kt
linacambridge: I think we have to, both for initialization and clearing storage between the Sync and history tests. AFAICT, that's the only additional function we need Glean to expose, though.

components/browser/storage-sync/src/test/java/mozilla/components/browser/storage/sync/PlacesBookmarksStorageTest.kt
linacambridge: Thanks, that does simplify things a lot! Since `Connection` is an internal interface, anyway (though it wasn't before; hopefully, @grigoryk is fine with that), I added separate `assemble{History, Bookmarks}Ping` and `send{History, Bookmarks}Ping` methods, with default implementations. The `RustPlacesConnection` interface actually implements the latter. That felt a little easier than futzing around with Mockito spies for `PlacesApi` and `RustPlacesConnection`.

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesHistoryStorage.kt
linacambridge: Thanks, I went with a slightly modified variation of your suggestion! 😄

components/browser/storage-sync/metrics.yaml
linacambridge: Edit: @mdboom talked about this on Slack, and, after thinking about it, we'd really like to have a way to record raw timespans for the MVP, or soon after, so we can measure the performance of our storage layer.

linacambridge:
> I couldn't test it locally: is this PR requiring some other change somewhere else?

This depends on a yet-to-be-published version of `application-services`...I think you'll have to build and publish a-s to your local Maven repo if you want to test locally now. Once we cut an a-s release with telemetry support, I'll drop b713ef3e142dcb89897f9674f629852cb20ea166, and we can point directly to that.

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/Connection.kt
```diff
@@ -22,14 +29,126 @@ const val DB_NAME = "places.sqlite"
...
     // Until we get a real SyncManager in application-services libraries, we'll have to live with this
     // strange split that doesn't quite map all that well to our internal storage model.
     fun syncHistory(syncInfo: SyncAuthInfo)
     fun syncBookmarks(syncInfo: SyncAuthInfo)
+
+    // These are implemented as default methods on `Connection` instead of
+    // `RustPlacesConnection` to make testing easier.
+    @Suppress("ComplexMethod", "NestedBlockDepth")
+    fun assembleHistoryPing(ping: SyncTelemetryPing) {
+        if (!Glean.getUploadEnabled()) {
```
Dexterp37: Are you doing this for performance reasons? Because, on our end, accumulating to our metrics is a no-op if upload is disabled. I'd consider dropping this for now and just doing that if there's any performance issue.

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/Connection.kt
```diff
@@ -22,14 +29,126 @@ const val DB_NAME = "places.sqlite"
...
+    }
+    fun sendHistoryPing() {}
+
+    // This function is almost identical to `recordHistoryPing`, with additional
+    // reporting for validation problems. Unfortunately, since the
+    // `BookmarksSync` and `HistorySync` metrics are two separate objects, we
+    // can't factor this out into a generic function.
+    @Suppress("ComplexMethod", "NestedBlockDepth")
+    fun assembleBookmarksPing(ping: SyncTelemetryPing) {
+        if (!Glean.getUploadEnabled()) {
```
Dexterp37: As above, I'd consider removing this check.

components/browser/storage-sync/src/test/java/mozilla/components/browser/storage/sync/PlacesBookmarksStorageTest.kt
```diff
@@ -263,5 +275,108 @@ class PlacesBookmarksStorageTest {
...
+                            Assert.assertFalse(incoming["failed_to_apply"].testHasValue())
+                            Assert.assertFalse(incoming["reconciled"].testHasValue())
+                            Assert.assertEquals(10, outgoing["uploaded"].testGetValue())
+                            Assert.assertEquals(5, outgoing["failed_to_upload"].testGetValue())
+                            Assert.assertEquals(1, outgoingBatches.testGetValue())
+                        }
+                    }
+                    else -> Assert.fail()
+                }
+                Pings.bookmarksSync.send()
```
Dexterp37: By doing this you already clear the stores, so no need to do that in the `resetGlean` :)

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesHistoryStorage.kt
Dexterp37: Great! Glad it worked out!

components/service/glean/src/test/java/mozilla/components/service/glean/TestUtil.kt
Dexterp37: Mh, why do we need that? They are different pings, right? I'm happy to meet today to talk through this!

linacambridge: I forgot I still have to send the ping after checking the values in `send{History, Bookmarks}Ping`. 🤦‍♀ With that fixed, the tests work! 😅

components/browser/storage-sync/metrics.yaml
linacambridge: Thanks!

components/browser/storage-sync/src/test/java/mozilla/components/browser/storage/sync/TestUtil.kt
```diff
@@ -0,0 +1,36 @@
...
+    config: Configuration = Configuration(),
+    clearStores: Boolean = true
+) {
+    Glean.enableTestingMode()
+
+    // We're using the WorkManager in a bunch of places, and Glean will crash
+    // in tests without this line. Let's simply put it here.
+    WorkManagerTestInitHelper.initializeTestWorkManager(context)
+
+    Glean.reset(context, config, clearStores)
```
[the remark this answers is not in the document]
linacambridge: Done!

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -519,6 +519,37 @@ open class GleanInternalAPI internal constructor () {
...
     }
+
+    /**
+     * Resets the Glean state and trigger init again.
+     *
+     * @param context the application context to init Glean with
+     * @param config the [Configuration] to init Glean with
+     * @param clearStores if true, clear the contents of all stores
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
```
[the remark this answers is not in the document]
linacambridge: Ahh, got it. I thought clearing the stores between tests was encouraged, but turns out I was just using the API wrong! 😆

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/Connection.kt
linacambridge: Yep, it was just to save on unnecessary work, but it's a bit silly. Dropped!

components/browser/storage-sync/src/test/java/mozilla/components/browser/storage/sync/TestUtil.kt
```diff
@@ -0,0 +1,36 @@
...
+import java.util.Date
+
+/**
+ * Enable testing mode, initialize [WorkManagerTestInitHelper] for testing,
+ * and initialize Glean. This should be called before each test that sends
+ * Glean telemetry pings.
+ *
+ * @param context the application context to init Glean with
+ * @param config the [Configuration] to init Glean with
+ * @param clearStores if true, clear the contents of all stores
```
Dexterp37: nit: this is no longer present

components/browser/storage-sync/src/test/java/mozilla/components/browser/storage/sync/TestUtil.kt
```diff
@@ -0,0 +1,36 @@
...
+ */
+internal fun resetGlean(
+    context: Context = ApplicationProvider.getApplicationContext(),
+    config: Configuration = Configuration()
```
Dexterp37: nit: configuration is not strictly required for your tests

components/browser/storage-sync/src/test/java/mozilla/components/browser/storage/sync/TestUtil.kt
```diff
@@ -0,0 +1,36 @@
...
+
+import android.content.Context
+import androidx.test.core.app.ApplicationProvider
+import androidx.work.testing.WorkManagerTestInitHelper
+import mozilla.components.service.glean.Glean
+import mozilla.components.service.glean.config.Configuration
+import java.util.Date
+
+/**
+ * Enable testing mode, initialize [WorkManagerTestInitHelper] for testing,
```
Dexterp37: nit: "This should be called before each test that sends Glean telemetry pings" it's not strictly true: this must be called at least *once* before any test sending or recording glean metrics can work :)

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/Types.kt
```diff
@@ -66,3 +70,51 @@ internal fun mozilla.appservices.places.VisitInfo.into(): VisitInfo {
...
+                it.failed + it.newFailed
+            } ?: 0
+            val (uploaded, failedToUpload) = info.outgoing.fold(Pair(0, 0)) { totals, batch ->
+                val (uploaded, failedToUpload) = totals
+                Pair(uploaded + batch.sent, failedToUpload + batch.failed)
+            }
+            return BaseGleanSyncPing(
+                uid = uid,
+                startedAt = Date(info.at.toLong() * MILLIS_PER_SEC),
+                // Glean intentionally doesn't support recording arbitrary
```
Dexterp37: nit: it might be useful for your team to have a reference to the issue here, https://bugzilla.mozilla.org/show_bug.cgi?id=1556341

linacambridge:

We would like to measure the performance and correctness of our Rust sync implementation. This includes collecting the time taken to sync each data type (currently history and bookmarks), incoming and outgoing record counts, any errors that occur (reporting sanitized error messages in a `string` field), and, for bookmarks, tree structure problem counts.

With the exception of the error string, which does not contain PII, we're submitting timings and counts only.

2) Why does Mozilla need to answer these questions?  Are there benefits for users? Do we need this information to address product or business requirements?

We need to understand how our new Sync implementation behaves in the wild. The [Sync ping](...) in Firefox Desktop and Firefox for iOS already exists, and has been extremely valuable in identifying and diagnosing Sync issues.

This pull request collects the same information as the Sync ping, but ports its structure to Glean.

Server-side metrics are not sufficient to understand Sync performance (especially for each step of an engine sync), given that the bulk of the work happens on clients. Validation data can only be collected on the client, since Sync records are encrypted and opaque to the server. The Sync ping for Desktop and iOS provides some stats about Desktop, but, since all three still use different Sync implementations, can't be extrapolated to Fenix.

Android Components consumers, including Fenix, do not currently report any Sync telemetry.

5) List all proposed measurements and indicate the category of data collection for each measurement, using the Firefox [data collection categories](https://wiki.mozilla.org/Firefox/Data_Collection) on the Mozilla wiki.

<table>
  <tr>
    <td>Tracking Bug #</td>
  </tr>
  <tr>
    <td>Timings, counts, and failure reasons for history syncs</td>
    <td>Interaction data</td>
    <td>#3092</td>
  </tr>
  <tr>
    <td>Timings, counts, and failure reasons for bookmark syncs</td>
    <td>Interaction data</td>
    <td>#3092</td>
  </tr>
</table>

I want to permanently monitor this data. (Lina Cambridge)

All Sync users with Sync enabled.

The data is _not_ correlated to the `client_id`; instead, we send a hash of the user's Firefox account ID (`uid`). This does not expose new identifiers, as these are already submitted in the Sync ping on other platforms.

* Which release channels?

All.

* Which countries?

All.

* Which locales?

All.

No.

Users can opt-out by disabling telemetry, or signing out of Sync.

We will create queries in re:dash to monitor the engines, adding to our existing sync engine error/success (https://sql.telemetry.mozilla.org/dashboard/sync-leif-status-dashboard-wip) dashboards and engine error analysis notebook for Desktop (https://gist.github.com/mhammond/66684669e1478d65bd60446cf150c244).

See above.

11) Is there a third-party tool (i.e. not Telemetry) that you are proposing to use for this data collection?

No.

grigoryk:
Not sure why github isn't letting me land this - " Merging can be performed automatically once the requested changes are addressed. ", but AFAIK it's all good to go (and I don't see any unresolved change requests).

@pocmo could you land this, please, with your super powers?

pocmo:
🛬

linacambridge:
Thanks for all your help getting this working! ❤️

PR #3099, opened 2019-05-23
Add a Glean CrashReporterService to lib-crash

travis79:
This includes two metrics defined in a metrics.yaml inside of lib-crash that instruments a count of each type (uncaught exceptions and native code crashes).

### Pull Request checklist


components/lib/crash/metrics.yaml
```diff
@@ -0,0 +1,33 @@
...
+  uncaught_exception:
+    type: counter
+    description: >
+      Counts the number of uncaught exceptions raised by the application.
+    bugs:
+      - 1522461
+    data_reviews:
+      - TBD
+    notification_emails:
+      - glean-team@mozilla.com
```
Dexterp37: I'm not sure this should be us. @georgf , thoughts?

components/lib/crash/metrics.yaml
```diff
@@ -0,0 +1,33 @@
...
+  native_code_crash:
+    type: counter
+    description: >
+      Counts the number of native code crashes in the application.
+    bugs:
+      - 1522461
+    data_reviews:
+      - TBD
+    notification_emails:
+      - glean-team@mozilla.com
```
Dexterp37: As above!

components/lib/crash/metrics.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+crash_metrics:
+  uncaught_exception:
+    type: counter
+    description: >
+      Counts the number of uncaught exceptions raised by the application.
+    bugs:
+      - 1522461
+    data_reviews:
+      - TBD
```
Dexterp37: Comment as a reminder to not merge this w\o data review :)

components/lib/crash/metrics.yaml
```diff
@@ -0,0 +1,33 @@
...
+      - glean-team@mozilla.com
+    expires: never
+  native_code_crash:
+    type: counter
+    description: >
+      Counts the number of native code crashes in the application.
+    bugs:
+      - 1522461
+    data_reviews:
+      - TBD
```
Dexterp37: Comment as a reminder to not merge this w\o data review :)

components/lib/crash/metrics.yaml
```diff
@@ -0,0 +1,33 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
Dexterp37: @mdboom does this file need to added to the probe-scraper too?

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanService.kt
```diff
@@ -0,0 +1,21 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.lib.crash.service
+
+import mozilla.components.lib.crash.Crash
+import mozilla.components.lib.crash.GleanMetrics.CrashMetrics
+
+/**
+ * A [CrashReporterService] implementation for recording metrics with Glean
```
Dexterp37: Let's expand a bit more about what this is actually doing.

components/lib/crash/metrics.yaml
```diff
@@ -0,0 +1,33 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# This file defines the metrics that are recorded by glean telemetry. They are
+# automatically converted to Kotlin code at build time using the `glean_parser`
+# PyPI package.
+
+$schema: moz://mozilla.org/schemas/glean/metrics/1-0-0
+
+crash_metrics:
```
Dexterp37: Do we want these metrics to be in the `metrics` ping?

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanService.kt
```diff
@@ -0,0 +1,21 @@
...
+ */
+class GleanService : CrashReporterService {
```
Dexterp37: This is lacking a doc entry in the [README FILE](https://github.com/mozilla-mobile/android-components/blob/5254d1a9d26210cef568963c03e99034ad9946f9/components/lib/crash/README.md)

components/lib/crash/src/test/java/mozilla/components/lib/crash/service/GleanServiceTest.kt
```diff
@@ -0,0 +1,63 @@
...
+import org.junit.Test
+import org.junit.Before
+import org.junit.runner.RunWith
+import org.mockito.Mockito.spy
+import org.mockito.Mockito.verify
+import org.mockito.Mockito.verifyNoMoreInteractions
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
+class GleanServiceTest {
```
Dexterp37: I'd add a test to simulate a crash/recording before Glean is initialized, to make sure this doesn't act weird.

components/lib/crash/src/test/java/mozilla/components/lib/crash/service/GleanServiceTest.kt
```diff
@@ -0,0 +1,63 @@
...
+    private val context: Context
+        get() = ApplicationProvider.getApplicationContext()
+
+    @Before
+    fun setupGleanServiceTest() {
```
Dexterp37: You can call this manually at the beginning of each test, so that you can have a test with "no glean initialzied" (the first test). I'd put a big comment there to mention that test ordering is important in this case.

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanService.kt
```diff
@@ -0,0 +1,21 @@
...
+    override fun report(crash: Crash.UncaughtExceptionCrash) {
+        CrashMetrics.uncaughtException.add()
+    }
+
+    override fun report(crash: Crash.NativeCodeCrash) {
+        CrashMetrics.nativeCodeCrash.add()
+    }
```
pocmo: I think there's a potential problem here: The crash reporter is running in a separate process (since the app process may have died). Glean is not initialized in this other process and probably shouldn't be in other's than the main app process.

components/lib/crash/build.gradle
```diff
@@ -33,6 +37,8 @@ dependencies {
 
     implementation project(':support-base')
 
+    implementation project(':service-glean')
```
pocmo: I wonder how we can prevent having a hard-dependency on Glean for every user of this library. I can think of two options:
* 1) We reverse the dependency and GleanService lives in Glean. However then Glean has a dependency on lib-crash - does this make more sense?
* 2) We turn this dependency into a compileOnly dependency like we did with Sentry below: An app using Glean will already depend on it and therefore the dependency gets resolved - but for everyone else it does not and they do not need to pay the cost.

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanService.kt
pocmo: naming nit: Something like `GleanCrashReporterService` may be more descriptive.

components/lib/crash/build.gradle
travis79: The other way I was considering going with this was to implement the Glean CrashReporterService in the application rather than the library, but I'll definitely look into doing in the service-glean.  We really didn't want a hard dependency on the lib-crash either so I understand not wanting the hard dependency the other way.

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanService.kt
travis79: Will test to see how this works out..  thanks for the heads up.

components/lib/crash/build.gradle
Dexterp37: Can we try instead to move on with option (2)?

travis79: Yes, absolutely!  I meant to say that was the route I would attempt at first, but I guess I didn't finish my thought in the previous comment.

components/lib/crash/metrics.yaml
travis79: Good question, I assume this would be a good thing to have in `baseline` or `glean_client_info`?

components/lib/crash/src/test/java/mozilla/components/lib/crash/service/GleanServiceTest.kt
travis79: It doesn't act weird or crash or anything, but it doesn't record anything either.  Basically, it looks like we won't be able to record any crashes that occur before Glean is initialized without thinking around the problem a bit.

components/lib/crash/metrics.yaml
Dexterp37: I'm not sure about that. `baseline` is for things common to all products. Not all products might use `lib-crash`, so it feels this should go in `metrics`. @georgf , thoughts?

components/lib/crash/src/test/java/mozilla/components/lib/crash/service/GleanServiceTest.kt
Dexterp37: We don't need to record before Glean is initialized I think, but we don't want to crash either.

mdboom: Intuitively, it feels like for MVP it's good enough to only record crashes the occur after `Glean.initialize`...

travis79: So a simple test to show that lib-crash attempting to record before Glean is initialized doesn't crash and records no data, reasonable enough.

travis79: That way if we change it to where we can record data before Glean is initialized, we can just modify the test a bit to show the collection.

components/lib/crash/metrics.yaml
georgf: This is a measure owned by Glean, as such it should go into its default ping - the `metrics` ping - unless there are good arguments otherwise.

components/lib/crash/metrics.yaml
georgf: We are doing the initial work, but we shouldn't own this.
Sebastian or Stefan would probably know which email should go there.

components/lib/crash/metrics.yaml
```diff
@@ -0,0 +1,33 @@
...
+  uncaught_exception:
```
georgf: What about using a single [labeled counter](https://github.com/mozilla-mobile/android-components/blob/master/components/service/glean/docs/metrics/adding-new-metrics.md#labeled-metrics) here?
With the labels `uncaught_exception`, `native_code_crash`, `other` or so.

components/lib/crash/metrics.yaml
travis79: Sounds good to me, I think that makes sense.

components/lib/crash/metrics.yaml
travis79: Sounds good to me.

@pocmo or @st3fan, would you mind helping to answer our question about who should/will own crash counts in lib-crash?  Basically we just need a point of contact email to insert here for ownership purposes.  Thanks!

components/lib/crash/metrics.yaml
travis79: lib-crash only records the two types of crashes so there is no `other`.  It seemed simple enough to instrument as two plain counters rather than a single labeled one.  If we wanted to use a labeled counter for this, I would almost rather use the 'labeled' bit to record the exception that was raised rather than which type of crash it was, that seems like it would be more useful to me, but I am open to suggestions here.

components/lib/crash/metrics.yaml
pocmo: For now that's probably `android-components-team@mozilla.com` - does it make sense to add both teams AC+Glean?

georgf: I think we should only add actual owners of the metric to this field.
While we're helping out with this, our team wouldn't be the owner of it going forward.

components/lib/crash/metrics.yaml
georgf: Ok, my only strong opinion here is lets stick with simple counts first.

For collecting more rich data, the question is what data in which shape is useful to the Android engineering teams, which requires a conversation we should have separately later.

A labeled counter here would match the semantics of "these two counts belong together, counting to versions of the same thing".
That matches what on desktop we use `categorical` histograms for, see e.g. [this](https://mzl.la/2KcDEKg). It has more than two labels, but this shows the concept well. You know you will only record one of them for each occurence, that you can sum them and that you can graph them together.

travis79: Okay, you have convinced me :)

components/lib/crash/metrics.yaml
mdboom: Yes, it will need to be added.

travis79:
This is ready for a thorough look, especially by @pocmo to be sure I'm not violating Android or lib-crash too badly ;)

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
```diff
@@ -0,0 +1,93 @@
...
+        // prevent the situation that arises because the integrating app's Application will be
+        // re-created when prompting to report the crash, and Glean is not initialized there since
+        // it's not technically the main process.
+        if (context.isMainProcess() && file.exists() && file.canRead()) {
+            // Because lib-crash runs in a separate process when report() is called, we need to do
+            // actual recording here because the GleanCrashReporterService constructor is called and
+            // runs in the process where Glean is actually initialized.
+            //
+            // NOTE:  It is important to make sure to initialize Glean before instantiating and
+            // registering the CrashReporter!
```
chutten: The CrashReporter or the GleanCrashReporterService? 

Also: is it possible to assert this?

Also also: if it is this important we should probably make a note of it in libcrash's integration doc

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
```diff
@@ -0,0 +1,93 @@
...
+            val lines = file.readLines()
+
+            // It's possible that there was more than one crash recorded in the file so process each
+            // line.
+            lines.forEach { line ->
+                if (line == UNCAUGHT_EXCEPTION_KEY) {
```
chutten: There's no Kotlin string enum class kinda thing is there? It doesn't seem likely that there'll ever be more than these two possible families of crash (though if this key becomes useful, expect layer violation errors later as things like shutdown crashes and hangs on exit get added despite being subclasses of uncaught exceptions or native crashes)

(from conversation on IRC sounds like there isn't. Putting this in the review just in case it leads somewhere)

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * A [CrashReporterService] implementation for recording metrics with Glean.  The purpose of this
+ * crash reporter is to collect crash count metrics by capturing [Crash.UncaughtExceptionCrash] and
+ * [Crash.NativeCodeCrash] events and record to the respective
+ * [mozilla.components.service.glean.private.CounterMetricType].
+ */
+class GleanCrashReporterService(val context: Context) : CrashReporterService {
+    companion object {
+        // These keys correspond to the labels found for crashCount metric in metrics.yaml
+        const val CRASH_FILE_NAME = "glean_crash_counts"
```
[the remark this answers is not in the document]
chutten: (in fact we might want that in documentation someplace so that apps know what they're signing themselves up for)

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
```diff
@@ -0,0 +1,93 @@
...
+        const val UNCAUGHT_EXCEPTION_KEY = "uncaught_exception"
+        const val NATIVE_CODE_CRASH_KEY = "native_code_crash"
+    }
+
+    private val logger = Logger("glean/GleanCrashReporterService")
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal val file = File(context.filesDir, CRASH_FILE_NAME)
```
mdboom: Glean itself uses `context.application.dataDir`.  I think the latter is immune to moves of the app to/from the SD card so is a more stable choice.  If there's a good reason for `filesDir`, maybe add a documentation note?

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
mdboom: Yeah -- I think there is a use case of an app using `lib-crash` but not `Glean`.  I guess in that case Glean isn't initialized and nothing is recorded, no harm, no foul?  (Since we don't have that use case in Fenix and time is short, maybe just file a bug to investigate that?)

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
travis79: Upgraded to a `when` statement to be more idomatic of Kotlin ;)

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
travis79: Added the Note to the README.md

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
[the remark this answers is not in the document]
travis79: Perfectly reasonable request

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanService.kt
travis79: Finally found a solution that didn't break Glean or lib-crash 😄

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
```diff
@@ -0,0 +1,109 @@
...
+        if (!file.exists()) {
+            file.createNewFile()
+        }
+
+        // Add a line representing the crash that was received
+        if (file.canWrite()) {
+            file.appendText(NATIVE_CODE_CRASH_KEY + "\n")
+        }
+
+        val test = Test.One
```
Dexterp37: That's probably undesired, right? :P

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
```diff
@@ -0,0 +1,109 @@
...
+
+        test.name
+    }
+}
+
+enum class Test {
```
Dexterp37: I think we can remove this.

components/lib/crash/src/test/java/mozilla/components/lib/crash/service/GleanCrashReporterServiceTest.kt
```diff
@@ -0,0 +1,112 @@
...
+import org.mockito.Mockito.verifyNoMoreInteractions
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
+class GleanCrashReporterServiceTest {
+    private val context: Context
+        get() = ApplicationProvider.getApplicationContext()
+
+    private fun setupGleanCrashReporterServiceTest() {
+        WorkManagerTestInitHelper.initializeTestWorkManager(
```
Dexterp37: Eugh, it hurts me so much we need to do this in every Glean consumer :(

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -88,6 +89,10 @@ open class GleanInternalAPI internal constructor () {
         applicationContext: Context,
         configuration: Configuration = Configuration()
     ) {
+        // In certain situations Glean.initialize may be called from a process other than the main
+        // process.  In this case we want initialize to be a no-op and just return.
+        if (!applicationContext.isMainProcess()) { return }
```
Dexterp37: Can we print a log line when this happens? (Is logging ok to call off the main process? I think so, but maybe @pocmo can confirm that)

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -88,6 +89,10 @@ open class GleanInternalAPI internal constructor () {
         applicationContext: Context,
         configuration: Configuration = Configuration()
     ) {
+        // In certain situations Glean.initialize may be called from a process other than the main
```
Dexterp37: Can we please file a bug about clarifying the story with Glean and multi process? If somebody tries to record stuff off the main process, thanks to this it will complain about Glean not being initialized. We might want to send a different message or something. Let's take this to a bug :)

components/lib/crash/src/test/java/mozilla/components/lib/crash/service/GleanCrashReporterServiceTest.kt
```diff
@@ -0,0 +1,112 @@
...
+            ApplicationProvider.getApplicationContext())
+
+        Glean.enableTestingMode()
+        Glean.initialize(context)
+    }
+
+    @Test
+    fun `GleanCrashReporterService records native code crashes`() {
```
Dexterp37: If we do the refactoring above, we could do real unit testing, in isolation. We could test that:

* the function that writes crashes:
  - correctly writes from the expected processes;
  - works if the file doesn't exist yet;
  - doesn't crash if it can't write to the file;
  - the data is written in the expected format.

* the function that reads crashes:
  - only works from the main process;
  - correctly parses the data;
  - doesn't crash if the data is corrupted;
  - properly asserts that Glean is initialized.

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        // Add a line representing the crash that was received
+        if (file.canWrite()) {
+            file.appendText(UNCAUGHT_EXCEPTION_KEY + "\n")
+        }
+    }
+
+    override fun report(crash: Crash.NativeCodeCrash) {
+        // Persist the crash in a file so that it can be recorded on the next application start. We
```
Dexterp37: Are the `report` calls thread safe? Can it ever happen to have them at the same time? Can it ever happen for them to be called from different processes? What if we try to write, at the same time, from different processes?

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
```diff
@@ -0,0 +1,109 @@
...
+            // line.
+            lines.forEach { line ->
+                when (line) {
+                    UNCAUGHT_EXCEPTION_KEY -> CrashMetrics.crashCount[UNCAUGHT_EXCEPTION_KEY].add()
+                    NATIVE_CODE_CRASH_KEY -> CrashMetrics.crashCount[NATIVE_CODE_CRASH_KEY].add()
+                }
+            }
+
+            // Clear persisted counts by deleting the file
+            file.delete()
```
Dexterp37: Can this be hit while we're writing something?

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
```diff
@@ -0,0 +1,109 @@
...
+            //
+            // NOTE:  It is important to make sure to initialize Glean before instantiating and
+            // registering the CrashReporter!
+            val lines = file.readLines()
+
+            // It's possible that there was more than one crash recorded in the file so process each
+            // line.
+            lines.forEach { line ->
+                when (line) {
+                    UNCAUGHT_EXCEPTION_KEY -> CrashMetrics.crashCount[UNCAUGHT_EXCEPTION_KEY].add()
```
Dexterp37: nit: micro-premature-optimization: what if we accumulate the counts to a local variable and then call `add(theCount)`? This would minimize the locking tango on the Glean side.

components/lib/crash/src/test/java/mozilla/components/lib/crash/service/GleanCrashReporterServiceTest.kt
```diff
@@ -0,0 +1,112 @@
...
+        }
+    }
+
+    @Test
+    fun `GleanCrashReporterService is a no-op if Glean is not initialized`() {
+        // We are not calling `setupGleanCrashReporterServiceTest()` here on purpose in order to
+        // make sure that the CrashReporterService does not crash when interacting with Glean before
+        // it is initialized.
+
+        val service = spy(GleanCrashReporterService(context))
```
Dexterp37: can we `assertFalse(Glean.isInitialized())` here?

components/lib/crash/src/test/java/mozilla/components/lib/crash/service/GleanCrashReporterServiceTest.kt
```diff
@@ -0,0 +1,112 @@
...
+            assertTrue("Glean must record a value",
+                CrashMetrics.crashCount[GleanCrashReporterService.UNCAUGHT_EXCEPTION_KEY].testHasValue())
+            assertEquals("Glean must record correct value",
+                1,
+                CrashMetrics.crashCount[GleanCrashReporterService.UNCAUGHT_EXCEPTION_KEY].testGetValue())
+        }
+    }
+
+    @Test
+    fun `GleanCrashReporterService is a no-op if Glean is not initialized`() {
```
Dexterp37: Can we add a test that records both multiple native crashes and uncaught exceptions?

components/lib/crash/metrics.yaml
travis79: There's not much risk until someone actually integrates it into another app.  This is just another service that can handle crashes and until Fenix or some other app integrates it, it's harmless.  At this point in the game, I wasn't really thinking it would make it into Fenix, especially since this still requires data review, and would still need added to Fenix and pass review there.

Dexterp37: Right, perfect then :)

components/lib/crash/README.md
```diff
@@ -82,13 +82,29 @@ Add a `MozillaSocorroService` instance to your `CrashReporter` in order to uploa
...
+
+```Kotlin
+CrashReporter(
+    services = listOf(
+		GleanCrashReporterService()
+    )
+).install(applicationContext)
+```
+
+⚠️ Note: In order for Glean to be able to record to metrics, it **MUST** be initialized before instantiating the `GleanCrashReporterService`
```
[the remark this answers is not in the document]
travis79: The normal workflow for lib-crash using application is to instantiate the service and then register it with `CrashHandler`.  In the case of `GleanCrashReporterService`, you must call `Glean.initialize()` before instantiating the object.

components/lib/crash/metrics.yaml
```diff
@@ -0,0 +1,26 @@
...
+# PyPI package.
+
+$schema: moz://mozilla.org/schemas/glean/metrics/1-0-0
+
+crash_metrics:
+  crash_count:
+    type: labeled_counter
+    description: >
+      Counts the number of crashes that occur in the application.  This measures only the counts of
+      each crash in association with the labeled type of the crash.
```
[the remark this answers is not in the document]
travis79: The labels correspond to the types of crashes handled and documented by lib-crash

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
```diff
@@ -0,0 +1,109 @@
...
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal val file = File(context.applicationInfo.dataDir, CRASH_FILE_NAME)
+
+    init {
+        // We only want to record things on the main process because that is the only one in which
+        // Glean is properly initialized.  Checking to see if we are on the main process here will
+        // prevent the situation that arises because the integrating app's Application will be
+        // re-created when prompting to report the crash, and Glean is not initialized there since
+        // it's not technically the main process.
+        if (context.isMainProcess() && file.exists() && file.canRead()) {
```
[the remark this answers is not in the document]
travis79: Will do

components/lib/crash/metrics.yaml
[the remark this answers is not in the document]
Dexterp37: It's worth mentioning it here then :)

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
```diff
@@ -0,0 +1,109 @@
...
+            // Because lib-crash runs in a separate process when report() is called, we need to do
+            // actual recording here because the GleanCrashReporterService constructor is called and
+            // runs in the process where Glean is actually initialized.
+            //
+            // NOTE:  It is important to make sure to initialize Glean before instantiating and
```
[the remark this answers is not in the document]
travis79: Are we sure we need to call `Glean.isInitialized()` here?  This won't crash or anything if it's not, it just won't be able to record the data (which is a no-op inside of Glean if it's not initialized).

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
travis79: There really should only be one crash in the file.  It would have to be a weird case if there were more than one because typically:
1. The app would crash
2. lib-crash would record the line in the file
3. User restarts app and the crash is recorded into Glean.

It's possible, but I found it very unlikely to have more than a single crash in the file.

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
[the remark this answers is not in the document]
Dexterp37: Good point. I'd still advise to print a long line though. If somebody goes through the effort of setting up the crash service for Glean, they might want to know that they're not getting data out due to Glean not being initialized.

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
Dexterp37: Is this true, in general? @pocmo might know better, but I gues we'll probably have multiple content processes in the future. In that case, the content process will crash, or more of them will crash, and we'll have more than one line in here.

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
travis79: Yeah, oops...

components/lib/crash/src/test/java/mozilla/components/lib/crash/service/GleanCrashReporterServiceTest.kt
travis79: Totally agree.  I wonder if we could create another testAPI function that would perform this dirty work.  Maybe inside of the `enabledTestingMode` function?

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
travis79: I think it's probably okay, I'll add it and test.

components/lib/crash/src/test/java/mozilla/components/lib/crash/service/GleanCrashReporterServiceTest.kt
Dexterp37: That would be weird, as we'd need to depend on the workmanager testing package in a non testing environment :( Anyway, there's [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1552114) for it, let's brainstorm there!

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
travis79: [Bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1558181) filed

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
travis79: That's all implementation details of lib-crash...   I don't think you could have an app crash for 2 reasons at the same time.  @pocmo indicated that the report() functions are called synchronously and from lib-crash's own process, so very unlikely that they would be called at the same time, or be called from another process.

Dexterp37: Great! Thanks for digging into that. Do you mind adding that as a comment somewhere within this file, for future reference?

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
travis79: no, if we crash in this process, it would be suspended/terminated when the lib-crash process was writing.  I suppose it could be possible to crash while writing if something very weird happened, but that would likely result in a startup-crash loop.

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
travis79: Don't confuse threads and processes here...   multi-process in android is much different than having content running in a coroutine thread

Dexterp37: I really mean processes, not threads. On Firefox Desktop we have multiple content processes.

@pocmo , do we support/is it on the roadmap to have multiple content processes in Fennec/a-c?

components/lib/crash/README.md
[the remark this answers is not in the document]
travis79: I tried to clarify this a bit.

components/lib/crash/metrics.yaml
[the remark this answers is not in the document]
travis79: Done

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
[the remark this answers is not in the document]
travis79: I added the check for `isInitialized()` as well as more explicit error handling.

components/lib/crash/src/test/java/mozilla/components/lib/crash/service/GleanCrashReporterServiceTest.kt
travis79: I'm okay with adding the additional test coverage, the only problem being with testing around processes.  As far as I know, it's not possible to test multiple processes for background things like this.  I did find [Multiprocess Espresso](https://developer.android.com/training/testing/espresso/multiprocess) but it looks like it's purpose is for testing UI interactions across multiple processes.

travis79:
@liuche I added the data review request form to Bugzilla [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1553935#c3).  Please let me know if you have any questions.

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
```diff
@@ -0,0 +1,166 @@
...
+            Glean.isInitialized() &&
+            checkFileConditions()) {
+
+            // Parse the persisted crashes
+            parseCrashFile()
+
+            // Clear persisted counts by deleting the file
+            file.delete()
+        } else {
+            if (!context.isMainProcess()) {
```
Dexterp37: nit: I think this could be changed to be slightly more readable. We could re-arrange the `init` like this:

```
init {
   if (!context.isMainProcess()) {
       logger.error("GleanCrashReporterService init off of main process")
       return
   }

   if (!Glean.isInitialized()) {
       logger.error("Glean is not initialized")
       return
   }

  if (!checkFileConditions()) {
     // The function already logs internally about error conditions. 
     return
  }
  
  // Parse the persisted crashes
  parseCrashFile()

  ....
}
```

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    /**
+     * Checks the file conditions to ensure it can be opened and read.
+     *
+     * @return True if the file exists and is able to be read, otherwise false
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun checkFileConditions(): Boolean {
+        var fileIsValid = true
```
Dexterp37: nit: instead of having this `var` here, can't we simply return `false` in the `if` braces below and always return true after them?

components/lib/crash/src/test/java/mozilla/components/lib/crash/service/GleanCrashReporterServiceTest.kt
Dexterp37: @travis79 I think the tests above are still missing. I think it's ok to not have the process specific tests, but the others that I called out should really be there.

components/lib/crash/README.md
```diff
@@ -82,13 +82,29 @@ Add a `MozillaSocorroService` instance to your `CrashReporter` in order to uploa
...
 ).install(applicationContext)
 ```
 
 ⚠️ Note: Currently only native code crashes get uploaded to Socorro. Socorro has limited support for "uncaught exception" crashes too, but it is recommended to use a more elaborate solution like Sentry for that.
 
+### Sending crash reports to Glean
+
+[Glean](https://docs.telemetry.mozilla.org/concepts/glean/glean.html) is a new way to collect telemetry by Mozilla.
+
+Add a `GleanCrashReporterService` instance to your `CrashReporter` in order to record crashes in Glean:
```
georgf: I'm missing documentation here on:
- What kind of information is recorded into the Glean SDK.
- How the data flows from the crash process to get recorded into the Glean SDK.

Is this documented anywhere else (outside of code)?

travis79: No, nothing outside of code.  I was planning on adding a section to the documentation explaining this but with the focus on getting experiments done first has kept me from getting back to this PR.

components/lib/crash/src/test/java/mozilla/components/lib/crash/service/GleanCrashReporterServiceTest.kt
travis79: > If we do the refactoring above, we could do real unit testing, in isolation. We could test that:
> 
>     * the function that writes crashes:
>       
>       * correctly writes from the expected processes;
I believe that this is covered by the first 3 tests, checking that it records and can read back the 2 types of crashes as well as being able to process multiple crashes from a single file
>       * works if the file doesn't exist yet;
Each of the first 3 tests check to see that no file exists before starting.
>       * doesn't crash if it can't write to the file;
I added a test for this
>       * the data is written in the expected format.
Each of the first 3 tests check for content of the file.
> 
>     * the function that reads crashes:
>       
>       * only works from the main process;
No way to easily test multi-process stuff in unit tests, can file a bug for follow up if you agree
>       * correctly parses the data;
The first 3 tests check this by verifying data through the Glean TestAPI
>       * doesn't crash if the data is corrupted;
I added a test for this
>       * properly asserts that Glean is initialized.
I added a log message to indicate when the lib-crash interaction that requires the main process happens off of the main process.

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
travis79: `return` not allowed from `init`...  I suppose I could extract the check to a separate function if you are worried about readability here.

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
travis79: Yeah, but we can just put the return before the `if` to be even more Kotlinish

components/lib/crash/README.md
travis79: I added to the `README.md`, let me know if we want to add anything.  If we want to document this elsewhere, I'll file a follow-up bug.

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
jonalmeida: > `return` not allowed from `init`...

I ran (pun intended) into this issue too and found that you can wrap things in a `run` block so you can use early returns. See [TabsToolbarFeature](https://github.com/mozilla-mobile/android-components/blob/master/components/feature/tabs/src/main/java/mozilla/components/feature/tabs/toolbar/TabsToolbarFeature.kt#L26).

_carries on driving by_ 😄

components/lib/crash/README.md
```diff
@@ -82,13 +82,33 @@ Add a `MozillaSocorroService` instance to your `CrashReporter` in order to uploa
...
 ).install(applicationContext)
 ```
 
 ⚠️ Note: Currently only native code crashes get uploaded to Socorro. Socorro has limited support for "uncaught exception" crashes too, but it is recommended to use a more elaborate solution like Sentry for that.
 
+### Sending crash reports to Glean
+
+[Glean](https://docs.telemetry.mozilla.org/concepts/glean/glean.html) is a new way to collect telemetry by Mozilla. 
+This will record crash counts as a labelled counter with each label corresponding to a specific type of crash (native_code_crash, and unhandled_exception, currently).
+Due to the fact that Glean requires to be recorded to in the main process and lib-crash runs in a separate process when it runs to handle the crash,
```
mdboom: nit: Just found this a little hard to parse on the first read.

```suggestion
Due to the fact that Glean must be recorded to in the main process and lib-crash runs in a separate process when it runs to handle the crash, 
```

components/lib/crash/README.md
```diff
@@ -82,13 +82,33 @@ Add a `MozillaSocorroService` instance to your `CrashReporter` in order to uploa
...
+
+```Kotlin
+CrashReporter(
+    services = listOf(
+		GleanCrashReporterService()
+    )
+).install(applicationContext)
+```
+
+⚠️ Note: In order for Glean to be able to record to metrics, it **MUST** be initialized within the application before instantiating the `GleanCrashReporterService` and registering it with the `CrashReporter`
```
mdboom:
```suggestion
⚠️ Note: In order for Glean to be able to record to metrics, it **MUST** be initialized within the application before instantiating the `GleanCrashReporterService` and registering it with the `CrashReporter`.
```

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
```diff
@@ -0,0 +1,163 @@
...
+            checkFileConditions()) {
+
+            // Parse the persisted crashes
+            parseCrashFile()
+
+            // Clear persisted counts by deleting the file
+            file.delete()
+        } else {
+            if (!context.isMainProcess()) {
+                logger.error("GleanCrashReporterService initialized off of main process")
```
pocmo: This probably shouldn't be an `error` since that will definitely happen from the crash handler service and prompt process.

travis79:
> I'm a bit worried about this single file that we write/read from different processes. But as you noted that is unlikely to happen concurrently - although possible I guess -> something else may have restarted the app process already!
> 
I agree, but I'm going to hold off doing this for the short term. If it looks like this is happening and causing problems, then I'll follow this up with a mutex/lock on the file reading and writing.

> It feels a bit weird that `GleanCrashReporterService` does work from `init()`. If we ever create those services lazily (which may makes sense) then this wouldn't work anymore?
I am not 100% happy with doing it this way either.  The other solution that was discussed was to try and write this to a place where Glean could just read for it on initialize, rather than doing the work in the init, but then the metrics.yaml has to exist in Glean for it to record to the metrics and I felt it muddied the waters between lib-crash and Glean too much.

components/lib/crash/README.md
```diff
@@ -82,13 +82,33 @@ Add a `MozillaSocorroService` instance to your `CrashReporter` in order to uploa
...
     )
 ).install(applicationContext)
 ```
 
 ⚠️ Note: Currently only native code crashes get uploaded to Socorro. Socorro has limited support for "uncaught exception" crashes too, but it is recommended to use a more elaborate solution like Sentry for that.
 
+### Sending crash reports to Glean
+
+[Glean](https://docs.telemetry.mozilla.org/concepts/glean/glean.html) is a new way to collect telemetry by Mozilla. 
+This will record crash counts as a labelled counter with each label corresponding to a specific type of crash (native_code_crash, and unhandled_exception, currently).
```
Dexterp37: readability nits: `native_code_crash`, `unhandled_exception` (wrap them in backticks)

components/lib/crash/README.md
```diff
@@ -82,13 +82,33 @@ Add a `MozillaSocorroService` instance to your `CrashReporter` in order to uploa
...
+Due to the fact that Glean must be recorded to in the main process and lib-crash runs in a separate process when it runs to handle the crash, 
+lib-crash persists the data in a file format and then reads and records the data from the main process when the application is next run since the GleanCrashReporterService
```
Dexterp37: readability nit: wrap GleanCrashReporterService in backticks.

components/lib/crash/metrics.yaml
```diff
@@ -0,0 +1,27 @@
...
+      Counts the number of crashes that occur in the application.  This measures only the counts of
+      each crash in association with the labeled type of the crash. The labels correspond to the
+      types of crashes handled by lib-crash.
+    labels:
+      - uncaught_exception
+      - native_code_crash
+    bugs:
+      - 1553935
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1553935#c3
```
Dexterp37: I noticed that the data-review on this is still pending, could you check in with @liuche about it?
I suspect that application using lib-crash + Glean are still required to ask for data review. Let's also clear this with @liuche before merging because, if it does, we need to explicitly state that in the README file.

components/lib/crash/README.md
Dexterp37: I'd add a sentence somewhere in here stating "The list of collected metrics is available in the `[metrics.yaml file](metrics.yaml)`" (basically, I'd link to the file.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -288,6 +288,20 @@ open class GleanInternalAPI internal constructor () {
         return ExperimentsStorageEngine.getSnapshot().getValue(experimentId)
     }
 
+    /**
+     * Provides a way to clear all stores and serialized pings from a test, and set Glean back to an
+     * uninitialized state. When outside libraries use the testAPI, it may be difficult to reset the
+     * metrics and Glean to provide a blank slate to test from.
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    fun testResetGlean() {
```
Dexterp37: What is the problem you're trying to solve here? Which test is failing?

```
// Start of the test
val initialValue = theMetric.testGetValue()

// Do something test-specific

val finalValue = testMetric.testGetValue()

// Check that it stored the right thing
assertEqual(expectedValue, finalValue - initialValue)
```

Could the above work? I'd really try to hold off from this change for now, as this requires a broader discussion on how to do that (= proposal). Maybe we don't want to expose an API like that, maybe we just want to expose a JUnit rule that does that automatically.

travis79: The problem is that `Glean` singleton is maintaining state between tests.  I can run each individual test by itself and not see this, but if I run the entire suite of tests for this module then it maintains some state between tests within the `Glean` singleton.  This is throwing off the clean state from which to start a test from.

I considered just measuring the change from initial value, but unfortunately there is a test in there that checks that experiments does not crash if Glean is not initialized, and unless you know a way to control the order that the tests are run in, I cannot guarantee that test gets ran prior to any test that initializes Glean.  That is significant because there is currently no way to _de-init_ Glean.

> 
> The problem is that `Glean` singleton is maintaining state between tests. I can run each individual test by itself and not see this, but if I run the entire suite of tests for this module then it maintains some state between tests within the `Glean` singleton. This is throwing off the clean state from which to start a test from.

I mean, what's the specific test that's failing when you run the full suite of tests?

> I considered just measuring the change from initial value, but unfortunately there is a test in there that checks that experiments does not crash if Glean is not initialized, and unless you know a way to control the order that the tests are run in, I cannot guarantee that test gets ran prior to any test that initializes Glean. That is significant because there is currently no way to _de-init_ Glean.

You would need to get the initial value within each test, and check at the end of each test. That should work, unless I'm missing something here.

travis79: Basically every test that runs after the first test that initializes Glean and relies on a value from the Glean testAPI.  In this case, it's 3 tests.  If I rename them so that the order of the tests changes, then the specific tests that fail may change.

I _still need_ to have a way to run a test _before_ Glean is initialized, to guarantee that interacting with an experiment when Glean is not initialized does not crash.  Unless you know of a way to make the tests run in a particular order every time, I cannot guarantee that test gets ran before a test that initializes Glean.  Once Glean is initialized in the context of testing the whole suite of tests, there is no way to get `Glean.isInitialized()` to return false, so I cannot guarantee that Glean is not initialized for the test that it _has_ to be uninitialized for because I cannot control the order that the tests are run in.

travis79: On top of that, having interdependent tests just seems _wrong_...

components/lib/crash/src/test/java/mozilla/components/lib/crash/service/GleanCrashReporterServiceTest.kt
```diff
@@ -0,0 +1,237 @@
...
+    @Test
+    fun `GleanCrashReporterService records native code crashes`() {
+        var initialValue = 0
+        setupGleanCrashReporterServiceTest()
+
+        run {
+            // Because of how Glean is implemented, it can potentially persist information between
+            // tests or even between test classes, so we compensate by capturing the initial value
+            // to compare to.
+            if (CrashMetrics.crashCount[GleanCrashReporterService.NATIVE_CODE_CRASH_KEY].testHasValue()) {
```
Dexterp37: nit: not super important, at your discretion here and in the other locations :)

```
initialvalue = try {
  CrashMetrics.crashCount[GleanCrashReporterService.NATIVE_CODE_CRASH_KEY].testGetValue()
} catch(..) { 0 }
```

travis79: This is good!

pocmo:
For risk assessment: The first Fenix RC 1.1 build will be cut next Thursday. It will ship with AC 4.0.0 which is currently in development and will be released next Tuesday. If you want to minimize risk then land after the 4.0.0 release on Tuesday - then it will not affect Fenix 1.1 RC 1 and not AC 4.0.x dot releases if we need to ship one for further RC builds. But I leave the decision to you. :)

travis79:
> For risk assessment: The first Fenix RC 1.1 build will be cut next Thursday. It will ship with AC 4.0.0 which is currently in development and will be released next Tuesday. If you want to minimize risk then land after the 4.0.0 release on Tuesday - then it will not affect Fenix 1.1 RC 1 and not AC 4.0.x dot releases if we need to ship one for further RC builds. But I leave the decision to you. :)

Thanks @pocmo!  I'll discuss that in telemetry planning meeting.  

Since this would still require Fenix to add and register a `GleanCrashReporterService` before it should be affected by the changes, I think it's pretty low risk to merge this in AC 4.0.0.  But, since I am still waiting on a data review on this (which I've now pushed towards chutten since liuche seems quite busy these days), it may be that this isn't possible to land before then anyways.

components/lib/crash/README.md
```diff
@@ -82,13 +82,33 @@ Add a `MozillaSocorroService` instance to your `CrashReporter` in order to uploa
...
     services = listOf(
-		MozillaSocorroService(applicationContext, "your app name"),
+		MozillaSocorroService(applicationContext, "your app name")
     )
 ).install(applicationContext)
 ```
 
 ⚠️ Note: Currently only native code crashes get uploaded to Socorro. Socorro has limited support for "uncaught exception" crashes too, but it is recommended to use a more elaborate solution like Sentry for that.
 
+### Sending crash reports to Glean
```
chutten: Is this a good location to put a reminder that applications using the GleanCrashReporterService are required to undergo Data Collection Review for the crash counts they will be collecting?

travis79: I think so, this should be the launch point for someone wishing to use the `GleanCrashReporterService` in their app, and they would hopefully see a note about requesting review for their particular app/population.

travis79: Added an additional note:
⚠️ Note: Applications using the `GleanCrashReporterService` are required to undergo Data Collection Review for the crash counts that they will be collecting.

components/lib/crash/metrics.yaml
travis79: Note added to README.md

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
```diff
@@ -0,0 +1,168 @@
...
+     * a newline character.
+     *
+     * Example:
+     *
+     * <--Beginning of file-->
+     * uncaught_exception\n
+     * uncaught_exception\n
+     * native_code_crash\n
+     * uncaught_exception\n
+     * <--End of file-->
```
pocmo: Nothing blocking here, but what would be the strategy to expand this in the future? Switch to a different file with new semantics? Or introduce some versioning to know which format this is in?

travis79: We are moving to Reference Browser as soon as this lands!  As far as expanding on this, I think we would probably switch to a different file with different semantics, as there will likely be more metrics than just the counts that we want to capture.  That's one of the next big things on my radar is to expand crash metrics.

PR #3122, opened 2019-05-25
Fixes #3121. Store observers in registry in `Set` instead of `List`.

### LINKED ISSUE #3121: ObserverRegistry can store more than one reference to the same observer.
### Problem

[ObserverRegistry](https://github.com/mozilla-mobile/android-components/blob/51ef74db39a20fc76d62775db0f2039a6d0959bd/components/support/base/src/main/java/mozilla/components/support/base/observer/ObserverRegistry.kt#L19) stores registered observers in `List`. This may lead to situation when one observer can be subscribed multiple times but only first entry will be removed on [unregister request](https://github.com/mozilla-mobile/android-components/blob/51ef74db39a20fc76d62775db0f2039a6d0959bd/components/support/base/src/main/java/mozilla/components/support/base/observer/ObserverRegistry.kt#L65).

### Fix

Using `Set` instead. :)

dector:
### Pull Request checklist

### DIFF
--- a/components/support/base/src/main/java/mozilla/components/support/base/observer/ObserverRegistry.kt
+++ b/components/support/base/src/main/java/mozilla/components/support/base/observer/ObserverRegistry.kt
@@ -4,6 +4,7 @@
 
 package mozilla.components.support.base.observer
 
+import android.annotation.SuppressLint
 import android.view.View
 import androidx.lifecycle.GenericLifecycleObserver
 import androidx.lifecycle.Lifecycle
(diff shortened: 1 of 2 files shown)


components/support/base/src/main/java/mozilla/components/support/base/observer/ObserverRegistry.kt
```diff
@@ -117,6 +130,7 @@ class ObserverRegistry<T> : Observable<T> {
     /**
      * GenericLifecycleObserver implementation to bind an observer to a Lifecycle.
      */
+    @SuppressLint("RestrictedApi")
```
NotWoods: Is this suppress needed? I can't replicate the lint error.

dector: Probably yes. :) `GenericLifecycleObserver` (the one we are extending here) has restricted scope [LIBRARY](https://android.googlesource.com/platform/frameworks/support/+/refs/heads/master/lifecycle/common/src/main/java/androidx/lifecycle/GenericLifecycleObserver.java#25).

pocmo: We should not suppress this and fix that - Let's file a follow-up issue. :)

NotWoods: Follow-up: #3140

dector: @pocmo @NotWoods Thanks. :+1:

PR #3123, opened 2019-05-25
For #2346. Enable kotlin warningsAsErrors for `samples-toolbar` module.

dector:
### Issue #2346

Super-Trivial changes.

### Complexity
Extremelly-trivial (☆☆☆)

### Pull Request checklist

### DIFF
--- a/samples/toolbar/src/main/java/org/mozilla/samples/toolbar/SampleToolbarHelpers.kt
+++ b/samples/toolbar/src/main/java/org/mozilla/samples/toolbar/SampleToolbarHelpers.kt
@@ -28,7 +28,9 @@ enum class ToolbarConfiguration(val label: String) {
 }
 
 @Suppress("MagicNumber")
-class ConfigurationAdapter(val configuration: ToolbarConfiguration) : RecyclerView.Adapter<ConfigurationViewHolder>() {
+class ConfigurationAdapter(
+    private val configuration: ToolbarConfiguration
+) : RecyclerView.Adapter<ConfigurationViewHolder>() {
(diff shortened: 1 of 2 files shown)


samples/toolbar/src/main/java/org/mozilla/samples/toolbar/SampleToolbarHelpers.kt
```diff
@@ -126,7 +126,7 @@ class UrlBoxProgressView(
         progressDrawable.setBounds(0, 0, w, h)
     }
 
-    override fun onDraw(canvas: Canvas?) {
+    override fun onDraw(canvas: Canvas) {
```
dector: I haven't found evidences that `Canvas` can be null when `onDraw()` is called. So I assume that it should be hadled as `@NonNull` and null-check here will be redundant.

samples/toolbar/src/main/java/org/mozilla/samples/toolbar/SampleToolbarHelpers.kt
```diff
@@ -27,7 +27,7 @@ enum class ToolbarConfiguration(val label: String) {
     PRIVATE_MODE("Private Mode")
 }
 
-@Suppress("MagicNumber")
+@Suppress("MagicNumber", "MemberVisibilityCanBePrivate")
```
pocmo: I wonder why we suppress `MemberVisibilityCanBePrivate` here and can't make the members it's complaining about private?

dector: I think you are right. We can do it for `sample-*` packages.

PR #3140, opened 2019-05-27
Use public LifecycleObserver instead of internal

NotWoods:
Removed used of [internal GenericLifecycleObserver](https://github.com/mozilla-mobile/android-components/pull/3122#discussion_r287835670)

### Pull Request checklist

### DIFF
--- a/components/support/base/src/main/java/mozilla/components/support/base/observer/ObserverRegistry.kt
+++ b/components/support/base/src/main/java/mozilla/components/support/base/observer/ObserverRegistry.kt
@@ -4,11 +4,15 @@
 
 package mozilla.components.support.base.observer
 
-import android.annotation.SuppressLint
 import android.view.View
-import androidx.lifecycle.GenericLifecycleObserver
-import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.Lifecycle.Event.ON_DESTROY
+import androidx.lifecycle.Lifecycle.Event.ON_PAUSE
+import androidx.lifecycle.Lifecycle.Event.ON_RESUME
+import androidx.lifecycle.Lifecycle.State.DESTROYED
+import androidx.lifecycle.Lifecycle.State.RESUMED
+import androidx.lifecycle.LifecycleObserver
 import androidx.lifecycle.LifecycleOwner
+import androidx.lifecycle.OnLifecycleEvent
 import java.util.Collections
 import java.util.WeakHashMap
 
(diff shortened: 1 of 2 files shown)


components/support/base/src/main/java/mozilla/components/support/base/observer/ObserverRegistry.kt
```diff
@@ -35,17 +39,18 @@ class ObserverRegistry<T> : Observable<T> {
     }
 
     override fun register(observer: T, owner: LifecycleOwner, autoPause: Boolean) {
-        if (owner.lifecycle.currentState == Lifecycle.State.DESTROYED) {
+        // Don't register if the owner is already destroyed
+        if (!owner.lifecycle.currentState.isAtLeast(CREATED)) {
             return
```
pocmo: Is `currentState` something we can't access anymore? This is still what `LiveData` does internally:
https://android.googlesource.com/platform/frameworks/support/+/androidx-master-dev/lifecycle/livedata-core/src/main/java/androidx/lifecycle/LiveData.java#190

NotWoods: We can still access it, but `isAtLeast` covers the `INITIALIZED` state too. If `LiveData` just uses equals internally I could switch back.

pocmo: I wonder if there can be a situation where we pass in a `LifecycleOwner` that is not CREATED yet and then we would never register? 🤔

NotWoods: [During onCreate](https://stackoverflow.com/questions/45459036/lifecycle-state-always-initialized), before the `ON_CREATE` event is triggered. However, it's probably more sensible to allow the observer to be registered during that time.

components/support/base/src/main/java/mozilla/components/support/base/observer/ObserverRegistry.kt
```diff
@@ -35,17 +39,18 @@ class ObserverRegistry<T> : Observable<T> {
...
-        val lifecycleObserver = LifecycleBoundObserver(
-                owner,
-                registry = this,
-                observer = observer,
-                autoPause = autoPause)
+        val lifecycleObserver = if (autoPause) {
+            AutoPauseLifecycleBoundObserver(owner, registry = this, observer = observer)
+        } else {
+            LifecycleBoundObserver(owner, registry = this, observer = observer)
+        }
```
pocmo: If we specify `autoPause` then we do not have the destroy behavior here anymore, or am I missing something?

NotWoods: `AutoPauseLifecycleBoundObserver` extends `LifecycleBoundObserver`

pocmo: Oops, missed that, somehow I thought it extended `LifecycleObserver`.

PR #3142, opened 2019-05-27
Closes #3137: Add helper for for providing unique and stable noticiation ids.

### LINKED ISSUE #3137: Add helper to base component for negotiating notification ids
Notification ids are `Int`s. With different components and the app showing notifications, it's likely that there will be conflicts and the components/app overriding/cancelling each other's notifications.

To solve this issue we could add a helper to `support-base` that hands out notification ids to components/code that wants to show a notification.

The component/app would provide a `String` tag for the notification and the helper would return a unique `Int`. Internally the helper stores this assignment and maybe makes use of a LRU cache to eventually release unused ids.

pocmo:
### Pull Request checklist

### DIFF
--- a/components/support/base/src/main/java/mozilla/components/support/base/ids/NotificationIds.kt
+++ b/components/support/base/src/main/java/mozilla/components/support/base/ids/NotificationIds.kt
@@ -1,0 +1,75 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.support.base.ids
+
+import android.app.Notification
+import android.app.NotificationManager
+import android.content.Context
+import androidx.annotation.VisibleForTesting
+import androidx.core.app.NotificationManagerCompat
+
+// If the tag is not used again in one week then clear the id
+private const val NOTIFICATION_ID_LIFETIME: Long = 1000 * 60 * 60 * 24 * 7
+
+// We start at 10000 instead of 0 to avoid conflicts with app code using random low numbers.
+private const val ID_OFFSET = 10000
+
+private const val PREFERENCE_FILE = "notification_ids"
+
+/**
+ * Helper for component and app code to use unique notification ids without conflicts.
+ */
+object NotificationIds {
+    private val ids = SharedIds(PREFERENCE_FILE, NOTIFICATION_ID_LIFETIME, ID_OFFSET)
+
+    /**
+     * Get a unique notification ID for the provided unique tag.
+     */
+    fun getIdForTag(context: Context, tag: String): Int = ids.getIdForTag(context, tag)
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal var now: () -> Long
+        get() = ids.now
+        set(value) { ids.now = value }
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun clear(context: Context) = ids.clear(context)
+}
+
+/**
+ * Post a notification to be shown in the status bar.
+ *
+ * Uses a unique [String] tag instead of an [Int] id like [NotificationManager.notify].
+ */
+fun NotificationManager.notify(context: Context, tag: String, notification: Notification) {
+    notify(NotificationIds.getIdForTag(context, tag), notification)
+}
+
+/**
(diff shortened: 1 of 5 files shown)


components/support/base/src/main/java/mozilla/components/support/base/ids/NotificationIds.kt
```diff
@@ -0,0 +1,75 @@
...
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun clear(context: Context) = ids.clear(context)
+}
+
+/**
+ * Post a notification to be shown in the status bar.
+ *
+ * Uses a unique [String] tag instead of an [Int] id like [NotificationManager.notify].
+ */
+fun NotificationManager.notify(context: Context, tag: String, notification: Notification) {
```
Amejia481: Could be helpful to also mention the life span of the unique id?

docs/changelog.md
```diff
@@ -52,6 +52,23 @@ permalink: /changelog/
...
+
+  // Extension methods for showing and cancelling notifications
+  NotificationManagerCompat
+      .from(context)
+      .notify(context, "mozac.my.feature", notification)
+
+  NotificationManagerCompat
+      .from(context)
+      .cancel(context, "mozac.my.feature")
+  ```
```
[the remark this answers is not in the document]
pocmo: If it needs two separate ids, yes. Of course it should pick better names for its tag.

components/support/base/src/main/java/mozilla/components/support/base/ids/NotificationIds.kt
pocmo: I consider that an implementation detail that may change.

docs/changelog.md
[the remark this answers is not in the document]

components/support/base/src/main/java/mozilla/components/support/base/ids/SharedIds.kt
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    private fun preferences(context: Context): SharedPreferences {
+        return context.getSharedPreferences(fileName, Context.MODE_PRIVATE)
+    }
+
+    /**
+     * Remove all expired notification ids.
+     */
+    private fun compact(preferences: SharedPreferences, editor: SharedPreferences.Editor) {
```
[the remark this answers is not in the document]
pocmo: Updated!

components/support/base/src/main/java/mozilla/components/support/base/ids/SharedIds.kt
```diff
@@ -0,0 +1,89 @@
...
+        preferences.all.entries.filter { entry ->
+            entry.key.startsWith(KEY_LAST_USED_PREFIX)
+        }.filter { entry ->
+            val lastUsed = entry.value as Long
+            lastUsed < now() - idLifeTime
+        }.forEach { entry ->
+            val tag = entry.key.substring(KEY_LAST_USED_PREFIX.length)
+            editor.remove(tagToKey(tag))
+            editor.remove(tagToLastUsedKey(tag))
```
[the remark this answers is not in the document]
pocmo: Updated!

components/support/base/src/main/java/mozilla/components/support/base/ids/NotificationIds.kt
Amejia481: Then should we also return the id? Because as consumer ,I could assumed that the unique id is not going to change for this tag.

pocmo: If you want the ID then you can use `NotificationIds` yourself. Those are extension methods since you do not need to care about them when you just want to `notify()` and `cancel()`. The ids doesn't change for a tag (unless expired..).

Amejia481: What I meant, was that we are not communicating that the ids can get expired  neither here nor on `NotificationIds.getIdForTag`

pocmo: I don't think the consumer necessarily needs to care. This is mostly to not have an indefinitely growing file in the app directory. If a component needs to show a notification again one month later then it's fine if it gets a new id assigned. It's only critical when updating/cancelling/overriding/conflicting with an existing notification?

Amejia481: 👍

Issue #2724, opened 2019-04-12
Fxa state machine must integrate with Auth errors


grigoryk:

vladikoff:
from mtg: "How much detail do we show in the errors?"

@ryanfeeley "Mark wrote about:sync-logs for detailed errors, in Fenix we should only show hard errors when you change your password or you have to reconnect to sync"

vladikoff:
cc @mhammond @grigoryk let us know if you need to speak with @ryanfeeley about this

PR #3157, opened 2019-05-28
Fix changelog from issue #2724 [ci skip]

grigoryk:


grigoryk:
Landed changelog in the wrong version.

PR #3163, opened 2019-05-29
For #2346. Enable kotlin warningsAsErrors for `feature-prompts` module.

dector:
### Issue #2346

Trivial changes in tests.

### Complexity

Easy (★☆☆)

### Pull Request checklist

### DIFF
--- a/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt
+++ b/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt
@@ -153,6 +153,7 @@ class PromptFeature(
      * @param grantResults The grant results for the corresponding permissions
      * @see [onNeedToRequestPermissions].
      */
+    @Suppress("UNUSED_PARAMETER")
     fun onPermissionsResult(permissions: Array<String>, grantResults: IntArray) {
         if (grantResults.isNotEmpty() && grantResults.all { it == PackageManager.PERMISSION_GRANTED }) {
             onPermissionsGranted()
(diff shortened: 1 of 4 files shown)


components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt
```diff
@@ -149,6 +149,7 @@ class PromptFeature(
      * @param grantResults The grant results for the corresponding permissions
      * @see [onNeedToRequestPermissions].
      */
+    @Suppress("UNUSED_PARAMETER")
```
dector: `permissions` is unused. Seems like misleading implementation. Should I raise issue on it?

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/TimePickerDialogFragment.kt
```diff
@@ -110,7 +110,7 @@ internal class TimePickerDialogFragment : PromptDialogFragment(), DatePicker.OnD
         return view
     }
 
-    @SuppressWarnings("deprecation")
+    @Suppress("DEPRECATION")
```
dector: There is no way to set upper bound for required API, unfortunately.

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt
NotWoods: I believe the intent is to have a similar function signature to [`onRequestPermissionsResult`](https://developer.android.com/reference/android/support/v4/app/ActivityCompat.OnRequestPermissionsResultCallback).

dector: I think you are right. However parameter that is present in public API but absent in actual implementation can be potential cause of problems in the future. :)

PR #3178, opened 2019-05-29
Closes #3146: Refresh devices after updating device name

### LINKED ISSUE #3146: SendTab: setDeviceNameAsync does not get cached on the device constellation
When we update the device name, other devices can see it, but querying for it on the device constellation API doesn't show these changes (this affects the UI being out-of-date).

See for more details: https://github.com/mozilla-mobile/fenix/issues/2877#issuecomment-496256561

jonalmeida:
Writing tests for this right now isn't straight-forward so I'm differing this to another (currently unknown) time.

### Pull Request checklist

### DIFF
--- a/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FxaDeviceConstellation.kt
+++ b/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FxaDeviceConstellation.kt
@@ -138,9 +138,13 @@ class FxaDeviceConstellation(
 
     override fun setDeviceNameAsync(name: String): Deferred<Boolean> {
         return scope.async {
-            handleFxaExceptions(logger, "changing device name") {
+            val rename = handleFxaExceptions(logger, "changing device name") {
                 account.setDeviceDisplayName(name)
             }
+            // See the latest device (name) changes after changing it.
+            val refreshDevices = deviceManager.refreshDevicesAsync().await()
+
+            rename && refreshDevices
         }
     }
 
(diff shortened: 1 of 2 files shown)


components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FxaDeviceConstellation.kt
```diff
@@ -138,9 +138,13 @@ class FxaDeviceConstellation(
...
     override fun setDeviceNameAsync(name: String): Deferred<Boolean> {
         return scope.async {
-            handleFxaExceptions(logger, "changing device name") {
+            val rename = handleFxaExceptions(logger, "changing device name") {
                 account.setDeviceDisplayName(name)
             }
+            // See the latest device (name) changes after changing it.
+            val refreshDevices = deviceManager.refreshDevicesAsync().await()
+
+            rename && refreshDevices
```
grigoryk: this may be an overkill: a failing `refreshDevices` doesn't mean that we failed to update the name. But, let's roll with this.

grigoryk: It's also interesting to consider what you do when device names misalign. `name` was 'foo`, but after refreshing our device record from the server, we saw `bar`. Is that a failure? Maybe not necessarily, but it's worth logging at least!

So yeah - let's check if our names align, and log an `info` success and a `warn` on `name!=newName` (probably without logging actual device names).

jonalmeida: > It's also interesting to consider what you do when device names misalign. `name` was 'foo`, but after refreshing our device record from the server, we saw `bar`. Is that a failure?

Our observer should be called so we update the UI to show what we get from the API.

> So yeah - let's check if our names align, and log an `info` success and a `warn` on `name!=newName` (probably without logging actual device names).

Good call for logging this case. It _shouldn't_ happen though.

jonalmeida: Ah, the logging would have to be done in Fenix since that's where we observe the device constellation changes.

PR #3187, opened 2019-05-30
For fenix/#1306. Add `isInUrlEditMode` property to Toolbar. 

dector:
### Issue [fenix/#1366](https://github.com/mozilla-mobile/fenix/issues/1366)

Companion PR [fenix/#2908](https://github.com/mozilla-mobile/fenix/pull/2908)

### Changes

First commit:
  - Added Kotlin-friendly `fail()` method for tests. It returns `Nothing` thus can be used everywhere where `Any` expected.

Second commit:
  - Added `Toolbar.isInUrlEditMode` property with tests.

P.S. You can take a look at companion PR to get context for this changes.

### Complexity

Simple (★☆☆)

### Pull Request checklist

### DIFF
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
@@ -64,7 +64,7 @@ private const val AUTOCOMPLETE_QUERY_THREADS = 3
  *  +----------------+ +----------------+
  * ```
  */
-@Suppress("TooManyFunctions")
+@Suppress("TooManyFunctions", "LargeClass")
 class BrowserToolbar @JvmOverloads constructor(
     context: Context,
     attrs: AttributeSet? = null,
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
@@ -254,6 +254,9 @@ class BrowserToolbar @JvmOverloads constructor(
             editToolbar.urlView.typeface = value
         }
 
+    override val isInUrlEditMode: Boolean
+        get() = state == State.EDIT
+
     /**
      * Sets a listener to be invoked when focus of the URL input view (in edit mode) changed.
      */
--- a/components/support/test/src/main/java/mozilla/components/support/test/Assert.kt
+++ b/components/support/test/src/main/java/mozilla/components/support/test/Assert.kt
@@ -1,0 +1,21 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+@file:JvmName("AssertUtils")
+
+package mozilla.components.support.test
+
+/**
+ * Throw assertion error.
+ *
+ * @param message optional message for error
+ */
+@JvmOverloads
+fun fail(message: String? = null): Nothing {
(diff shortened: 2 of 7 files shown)


components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
```diff
@@ -254,6 +254,9 @@ class BrowserToolbar @JvmOverloads constructor(
             editToolbar.urlView.typeface = value
         }
 
+    override val isInUrlEditMode: Boolean
```
dector: KDoc for overrided properties/methods are inherited by default.

components/support/test/src/main/java/mozilla/components/support/test/Assert.kt
```diff
@@ -0,0 +1,21 @@
...
+package mozilla.components.support.test
+
+/**
+ * Throw assertion error.
+ *
+ * @param message optional message for error
+ */
+@JvmOverloads
+fun fail(message: String? = null): Nothing {
+    message ?: throw AssertionError()
```
NotWoods: I'd prefer to use either an `if` statement or `.let`, so that the `AssertionError` with a message is thrown next to the not-null check.
```suggestion
    message?.let { throw AssertionError(it) } ?: throw AssertionError()
```
```suggestion
    throw if (message != null) AssertionError(message) else AssertionError()
```

dector: Okay, I'll use `if-else` statement here.

dector:
Thanks for the suggestion, @pocmo . However, I'm affraid that keeping this behavior inside the component is more prefferable as it:
a) doesn't break incapsulation;
b) doesn't intruduce more entropy.

The reason for this changes is because "we don't want to send signal about mode change if we are already in this mode". Seems like the part of the logic for UI component itself. Are there some reasons we want to allow clients to change mode of the toolbar component *but don't allow* them to know which mode is used now?

docs/changelog.md
```diff
@@ -26,6 +26,9 @@ permalink: /changelog/
   * Added `onWebAppManifestLoaded` to `EngineSession`, called when the engine finds a web app manifest.
   * Added `WebAppManifest` from browser-session.
 
+* **concept-toolbar**
+  * Added `isInEditMode()` to `Toolbar` and `BrowserToolbar`.
```
dector: Typo: `isInUrlEditMode()` should be used.

pocmo:
@dector 

The reasons why I would like to avoid exposing this state:
* Minor: Low API surface - Only exposing things if really needed.
* Minor: The switching between those two states is an implementation detail of `browser-toolbar` and I would like to not add this as a requirement to all `concept-toolbar` implementations. We already added `displayMode() and editMode()` but I would like to not extend that and rather scale back.
* Major: I think our components for UI should have at best no state (or as minimal as possible). The app / other code is authoritative and has state and tells the toolbar to display sth. of that state. Whether or not the toolbar should display the URL or make it editable should be the responsibility of the app. The app should model that in its state if it needs to know about this.

> The reason for this changes is because "we don't want to send signal about mode change if we are already in this mode". 

This makes sense. And I wonder if we can model this inside the component. For example if the app triggers `editMode()` then the code shouldn't do anything if it's already in `editMode`- can this be handled internally in the component in this case?

dector:
Thanks, @pocmo . Really appreciated for your attention and extended answer.

I've got your point. I'm not fully agree in some details, howerver. :smiley_cat: (e.g. existing component is not pure functions and it already encapsulates some functionality and knowledge about its state, treating it outside the component seems bit error-prone breaking abstraction process for me). Though, seems  that discussion will take too much time than we have for the related Fenix issue. :)

Probably, this PR can be closed. I'll suggest existing fix for Fenix. Should I open new one for changes in tests (`fail()` method)?

Issue #2364, opened 2019-03-15
AwesomeBar: Two different ViewHolders have the same stable ID


pocmo:
From the reference browser:

> Two different ViewHolders have the same stable ID. Stable IDs in your adapter MUST BE unique and SHOULD NOT change. ViewHolder 1:ViewHolder{959c237 position=6 id=1262243701, oldPos=-1, pLpos:-1 not recyclable(1)} View Holder 2:ViewHolder{39ea416 position=2 id=1262243701, oldPos=-1, pLpos:-1 not recyclable(1)} mozilla.components.browser.awesomebar.BrowserAwesomeBar{d37fdf7 VFED..... ......ID 0,0-1080,1101 #7f080020 app:id/awesomeBar}, adapter:mozilla.components.browser.awesomebar.SuggestionsAdapter@4429464, layout:androidx.recyclerview.widget.LinearLayoutManager@2e7d6cd, context:org.mozilla.reference.browser.BrowserActivity@ad8263b

* https://sentry.prod.mozaws.net/operations/reference-browser/issues/5329251/
* https://sentry.prod.mozaws.net/operations/reference-browser/issues/5327619/

pocmo:
@grigoryk Collisions :(

grigoryk:
Ah! Just filed a duplicate #2501 for this. Yup, collisions.

PR #3194, opened 2019-05-30
Issue #2364: getItemId is not synchronized

csadilek:
@pocmo digging further into remaining possible causes of https://github.com/mozilla-mobile/android-components/issues/2364, I just saw that we have all but one method synchronized in `SuggestionsAdapter`, which doesn't seem right.

If `removeProviders` or `allProviders` (and therefore removeSuggestions or addSuggestions) are ever not called on the main thread this would cause trouble. I haven't been able to observe any of these not being called on the main thread, but since we're synchronizing all other methods, we should add the same guard here?


csadilek:
@pocmo amended commit as the same applies to the nested `getUniqueSuggestionId` which should behave the same as all other `synchronized` methods in this class.

PR #3204, opened 2019-05-31
Don't say 'metrics ping already collected' in fresh installs

Dexterp37:
On fresh installs of products using glean, when no date is available for when the 'metrics' ping was sent, don't provide "today" as the default date: no ping would be sent otherwise, and it will delegate sending to the next day. This adds the missing scenario to the ping docs.

@chutten would you kindly review the additions to the `metrics.md` file? 

### Pull Request checklist

### DIFF
--- a/components/service/glean/docs/pings/metrics.md
+++ b/components/service/glean/docs/pings/metrics.md
@@ -10,17 +10,22 @@ length of this window.
 
 ## Scheduling
 The desired behaviour is to collect the ping at the first available opportunity after 4AM local
-time on a new calendar day. This breaks down into two scenarios:
+time on a new calendar day. This breaks down into three scenarios:
 
-1. the application was just started (after a crash or a long inactivity period);
-2. the application was open and the 4AM due time was hit.
+1. the application was just installed;
+2. the application was just started (after a crash or a long inactivity period);
+3. the application was open and the 4AM due time was hit;
 
-In the first case, if the `metrics` ping was already collected on the current calendar day, a new
+In the first case, since the application was just installed, if the due time for the current calendar
+day has passed, a `metrics` ping is immediately generated and scheduled for sending. Otherwise, if the
+due time for the current calendar day has not passed, a ping collection is scheduled for that time.
+
+In the second case, if the `metrics` ping was already collected on the current calendar day, a new
 collection will be scheduled for the next calendar day, at 4AM. If no collection happened yet,
 and the due time for the current calendar day has passed, a `metrics` ping is immediately generated
 and scheduled for sending.
 
-In the second case, similarly to the previous case, if the `metrics` ping was already collected on the
+In the third case, similarly to the previous case, if the `metrics` ping was already collected on the
 current calendar day when we hit 4AM, then a new collection is scheduled for the next calendar day.
 Otherwise, the `metrics` is immediately collected and scheduled for sending.
 
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
@@ -158,7 +158,7 @@ internal class MetricsPingScheduler(val applicationContext: Context) {
      */
     fun startupCheck() {
         val now = getCalendarInstance()
-        val lastSentDate = getLastCollectedDate(now.time)
+        val lastSentDate = getLastCollectedDate()
 
         // We expect to cover 3 cases here:
         //
@@ -169,7 +169,7 @@ internal class MetricsPingScheduler(val applicationContext: Context) {
         // (3) - the ping was NOT collected the current calendar day, but we still have
         //       some time to the due time; schedule for sending the current calendar day.
 
-        val alreadySentToday = DateUtils.isToday(lastSentDate.time)
+        val alreadySentToday = (lastSentDate != null) && DateUtils.isToday(lastSentDate.time)
         when {
             alreadySentToday -> {
                 // The metrics ping was already sent today. Schedule it for the next
@@ -220,19 +220,22 @@ internal class MetricsPingScheduler(val applicationContext: Context) {
     /**
      * Get the date the metrics ping was last collected.
      *
-     * @param defaultDate the [Date] object to return if no valid last sent date can be found
-     * @return a [Date] object representing the date the metrics ping was last collected
+     * @return a [Date] object representing the date the metrics ping was last collected, or
+     *         null if no metrics ping was previously collected.
      */
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
-    internal fun getLastCollectedDate(defaultDate: Date = Date()): Date {
+    internal fun getLastCollectedDate(): Date? {
         val loadedDate = try {
             sharedPreferences.getString(LAST_METRICS_PING_SENT_DATETIME, null)
         } catch (e: ClassCastException) {
-            logger.error("MetricsPingScheduler last stored ping time was not valid")
             null
         }
 
-        return loadedDate?.let { parseISOTimeString(it) } ?: defaultDate
+        if (loadedDate == null) {
+            logger.error("MetricsPingScheduler last stored ping time was not valid")
+        }
+
+        return loadedDate?.let { parseISOTimeString(it) }
     }
 
     /**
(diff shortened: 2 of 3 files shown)


components/service/glean/src/test/java/mozilla/components/service/glean/scheduler/MetricsPingSchedulerTest.kt
```diff
@@ -356,6 +354,60 @@ class MetricsPingSchedulerTest {
         verify(mpsSpy, never()).collectPingAndReschedule(kotlinFriendlyAny<Calendar>())
     }
 
+    @Test
+    fun `startupCheck must correctly handle fresh installs (before due time)`() {
+        // Set the current system time to a known datetime: after 4am local.
```
chutten: Isn't this before 4am local?

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
```diff
@@ -220,19 +220,22 @@ internal class MetricsPingScheduler(val applicationContext: Context) {
...
+    internal fun getLastCollectedDate(): Date? {
         val loadedDate = try {
             sharedPreferences.getString(LAST_METRICS_PING_SENT_DATETIME, null)
         } catch (e: ClassCastException) {
-            logger.error("MetricsPingScheduler last stored ping time was not valid")
             null
         }
 
-        return loadedDate?.let { parseISOTimeString(it) } ?: defaultDate
+        if (loadedDate == null) {
```
travis79: nit: I think this could be more Kotlin idiomatic if you used the Elvis operator off of your `val loadedDate = try`.  Something like this:
```Kotlin
        val loadedDate = try {
            sharedPreferences.getString(LAST_METRICS_PING_SENT_DATETIME, null)
        } catch (e: ClassCastException) {
            null
        } ?: logger.error("MetricsPingScheduler last stored ping time was not valid")
```

components/service/glean/src/test/java/mozilla/components/service/glean/scheduler/MetricsPingSchedulerTest.kt
Dexterp37: Right :| Copy pasta!

components/service/glean/docs/pings/metrics.md
```diff
@@ -10,17 +10,22 @@ length of this window.
...
+time on a new calendar day. This breaks down into three scenarios:
 
-1. the application was just started (after a crash or a long inactivity period);
-2. the application was open and the 4AM due time was hit.
+1. the application was just installed;
+2. the application was just started (after a crash or a long inactivity period);
+3. the application was open and the 4AM due time was hit;
 
-In the first case, if the `metrics` ping was already collected on the current calendar day, a new
+In the first case, since the application was just installed, if the due time for the current calendar
```
travis79: At first I thought this was a little weird to send a metrics ping immediately on a fresh install since there would likely be nothing to send, but I realized that it helps to fill in the missing metrics pings we saw from the initial analysis and some metrics do get set at startup.

components/service/glean/src/test/java/mozilla/components/service/glean/scheduler/MetricsPingSchedulerTest.kt
```diff
@@ -153,33 +153,31 @@ class MetricsPingSchedulerTest {
...
 
     @Test
-    fun `getLastCollectedDate must report the default when no stored date is available`() {
+    fun `getLastCollectedDate must report null when no stored date is available`() {
         val mps = MetricsPingScheduler(ApplicationProvider.getApplicationContext<Context>())
         mps.sharedPreferences.edit().clear().apply()
 
-        val expectedDate = Date()
-        assertEquals(expectedDate, mps.getLastCollectedDate(expectedDate))
+        assertNull(mps.getLastCollectedDate())
```
travis79: nit:  assertion messages are handy and have helped us debug tests :)

components/service/glean/src/test/java/mozilla/components/service/glean/scheduler/MetricsPingSchedulerTest.kt
```diff
@@ -153,33 +153,31 @@ class MetricsPingSchedulerTest {
...
         mps.sharedPreferences
             .edit()
             .putLong(MetricsPingScheduler.LAST_METRICS_PING_SENT_DATETIME, 123L)
             .apply()
 
-        // Wrong key type should trigger the default date.
-        val expectedDate = Date()
-        assertEquals(expectedDate, mps.getLastCollectedDate(expectedDate))
+        // Wrong key type should trigger returning null.
+        assertNull(mps.getLastCollectedDate())
```
travis79: nit: assertion message

components/service/glean/src/test/java/mozilla/components/service/glean/scheduler/MetricsPingSchedulerTest.kt
```diff
@@ -153,33 +153,31 @@ class MetricsPingSchedulerTest {
...
 
-        // Wrong date format string should trigger the default date.
+        // Wrong date format string should trigger returning null.
         mps.sharedPreferences
             .edit()
             .putString(MetricsPingScheduler.LAST_METRICS_PING_SENT_DATETIME, "not-an-ISO-date")
             .apply()
 
-        assertEquals(expectedDate, mps.getLastCollectedDate(expectedDate))
+        assertNull(mps.getLastCollectedDate())
```
travis79: nit: assertion message

components/service/glean/src/test/java/mozilla/components/service/glean/scheduler/MetricsPingSchedulerTest.kt
```diff
@@ -190,7 +188,7 @@ class MetricsPingSchedulerTest {
 
         // Wrong key type should trigger the default date.
         val expectedDate = parseISOTimeString(testDate)!!
-        assertEquals(expectedDate, mps.getLastCollectedDate(expectedDate))
+        assertEquals(expectedDate, mps.getLastCollectedDate())
```
travis79: nit: assertion message

components/service/glean/src/test/java/mozilla/components/service/glean/scheduler/MetricsPingSchedulerTest.kt
```diff
@@ -356,6 +354,60 @@ class MetricsPingSchedulerTest {
...
+
+        // Make sure to return the fake date when requested.
+        doReturn(fakeNow).`when`(mpsSpy).getCalendarInstance()
+
+        // Trigger the startup check.
+        mpsSpy.startupCheck()
+
+        // And that we're storing the current date (this only reports the date, not the time).
+        fakeNow.set(Calendar.HOUR_OF_DAY, 0)
+        assertEquals(fakeNow.time, mpsSpy.getLastCollectedDate())
```
travis79: nit: assertion message

Dexterp37:
> This means that there are still some "sessions" for which we won't have a "metrics" ping: those that start and finish between midnight and 4am local. But for all others there will be a "metrics" ping.

Yes, that's true, probably complementary to this specific bug though. I'm honestly not sure how to approach this, we might want to loop in DS for that.

> What will be in that initial "metrics" ping? Just the client_info and ping_info? Is it worth sending only for initial "sessions" that happen after 4am local? (Is it worth sending at all?)

I'm not sure about that either. We're definitely not seeing clients show up, which could be weird from an analysis standpoint. The current logic was due to a bug in the code, so was not right either :/

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
Dexterp37: This looked nice, I just tried it, the compiler complains as `loadedDate` never becomes null :)

travis79: Oh well, worth a shot ;)

travis79: Could you still call the logger inside of the catch, and return the `null` like `return@try null`

PR #3209, opened 2019-05-31
Issue #2364: Make sure suggestions are removed before new ones are added

csadilek:
@pocmo I found another cause of our duplicate viewholder ID crash. I was able to reproduce this one pretty reliably in sample browser:  

The `SessionSuggestionProvider` has `shouldClearSuggestions=true`, but occasionally its (old) suggestion weren't cleared before new ones were added, which also bypasses our unique ID check. We only check for new suggestions being unique and make sure they get stable unique IDs, assuming the old ones have been deleted by this point.

I think the reason this crash came back in the past week or so is this change: https://github.com/mozilla-mobile/android-components/commit/8e2b62363d136ff23ebc51cd69d3322912fd0711

This commit moved `optionallyClearSuggestions` to after we launched the coroutine, which made this more likely. I was able to verify that `optionallyClearSuggestions` and `addSuggestions` can execute in an order that results in duplicate suggestions e.g. it was possible for two `onInputChanged` events to trigger two successive `optionallyClearSuggestion` calls before the two coroutines (to produce suggestions) are launched. Then those coroutines launched and we ended up with duplicated suggestions for the `SessionSuggestionProvider`.

This PR fixes that and also removes an unneeded context switch. The additional nested launch isn't needed as we're already on the `CoroutineScope(Dispatchers.Main)`.

I am not seeing any problems/crashes anymore with this fix, but I will continue testing. Let me know what you think.

### DIFF
--- a/components/browser/awesomebar/src/main/java/mozilla/components/browser/awesomebar/BrowserAwesomeBar.kt
+++ b/components/browser/awesomebar/src/main/java/mozilla/components/browser/awesomebar/BrowserAwesomeBar.kt
@@ -43,7 +43,7 @@ class BrowserAwesomeBar @JvmOverloads constructor(
     defStyleAttr: Int = 0
 ) : RecyclerView(context, attrs, defStyleAttr), AwesomeBar {
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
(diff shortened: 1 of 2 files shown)


csadilek:
Added a unit test and updated description to be a little clearer (hopefully). I can no longer reproduce the problem with the fix.

PR #3215, opened 2019-05-31
 Closes #3195: Ignore UrlParseFailed exceptions coming out of places

### LINKED ISSUE #3195: Catch UrlParseFailed places exceptions in PlacesStorage & friends
Specifically, do this for `noteObservation` calls. We don't want to crash if we fail to record invalid urls into our history store.

Consider similar treatment for others `places` calls as well.

grigoryk:
### Pull Request checklist


components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesStorage.kt
```diff
@@ -48,4 +52,15 @@ abstract class PlacesStorage(context: Context) : Storage, SyncableStore {
         scope.coroutineContext.cancelChildren()
         places.close()
     }
+
+    /**
+     * Runs [block] described by [operation], ignoring and logging any thrown [UrlParseFailed] exceptions.
+     */
+    protected fun ignoreUrlExceptions(operation: String, block: () -> Unit) {
```
jonalmeida: nit: `inline` would probably help performance if we hit this case often. No biggie though.

PR #3219, opened 2019-06-03
Have AsyncLoginsStorage.sync() return the sync ping 

mhammond:
Lockwise wants the sync telemetry ping ASAP. I opened https://github.com/mozilla/application-services/pull/1229 so the rust components expose it. This patch just passes it through to lockwise - at least I *think* it does - I failed to get lockwise for Android building with these changes, but I'll open a PR there with as far as I got. It duplicates some stuff from @linacambridge's #3092 but they are trivial to resolve.

Not sure if there's more to do here, but I'm marking it as a draft until we get lockwise working against it.

cc @sashei, @linuxwolf @grigoryk


mhammond:
I've updated this now that @linacambridge's patch has landed. Sadly, I think I'm still doing something wrong as I can convince lockbox to see the SyncTelemetryPing object. I'll get back to this after whistler, but pushing it now just incase I look at it *at* whistler :)

pocmo:
@mhammond Is this something that we still want to get in? :)

mhammond:
Thom has incorporated these changes into #3806

PR #3242, opened 2019-06-04
Gradle: zipMavenArtifacts wait for publish to be finished.

pocmo:


automation/taskcluster/decision_task.py
```diff
@@ -215,7 +215,7 @@ def _get_release_gradle_tasks(module_name, is_snapshot):
         for gradle_task in ('assemble', 'test', 'lint')
     )
 
-    return gradle_tasks + ' ' + module_name + ':publish' + ' zipMavenArtifacts'
+    return gradle_tasks + ' ' + module_name + ':publish' + ' ' + module_name + ':zipMavenArtifacts'
```
mitchhentges: Is `:publish` a valid gradle command? When a command is prefixed with a colon, doesn't that mean that it expects a submodule name, e.g.: `:module:publish`? Or does `:publish` mean to publish the root project (so, publish all subprojects)?

If it's the first case, won't this command fail?
If it's the second case, then that means that every separate build will do `publish` for all projects, causing us to do `<number of projects> * <number of projects>` of work

pocmo: I didn't modify the publish command, it's still `module_name + ':publish`. I added the module prefix to `zipMavenArtifacts` artifacts since we only want to run it for that module: `module_name + ':zipMavenArtifacts'`

mitchhentges: Derp, nice!
Does `:zipMavenArtifacts` know how to run with the scope of a single project?
Or does `dependsOn publish` mean that it depends on the root `publish` task?

pocmo: I think (or guess, I'm not really sure, gradle is mysterious), that we already define it in the project scope since we include it from publish.gradle into every module's build.gradle - I can run it and also see it listed when running `module:tasks`. And therefore I assumed that `dependsOn publish` means it depends on its local publish task.

I tested this by listing `zipMavenArtifacts` before `publish` locally:
> ./gradlew service-sync-logins:clean && ./gradlew service-sync-logins:zipMavenArtifacts service-sync-logins:publish

This reliably didn't create a zip. With `dependsOn` added the zip existed. It could still have been random luck but after running it a couple of times I think it does what I think it does. I would love to feel more confident though. :)

pocmo: > and also see it listed when running module:tasks

Correction: This is only true for the `publish` task. It does appear though if I add it to a group:

```
    task zipMavenArtifacts(type: Zip) {
        dependsOn publish
        group = "Publishing"
        [..]
```

mitchhentges: Makes sense: I bet it would show up in `./gradlew :module:tasks --all`, though I think it makes sense to add to the "Publishing" group.
I'm :+1: on this PR either way

PR #3246, opened 2019-06-04
Add setRawNanos API to TimespanMetricType

mdboom:
### Pull Request checklist

### DIFF
--- a/components/service/glean/docs/metrics/adding-new-metrics.md
+++ b/components/service/glean/docs/metrics/adding-new-metrics.md
@@ -427,10 +427,14 @@ The time reported in the telemetry ping will be the sum of all of these
 timespans recorded during the lifetime of the ping.
 
 If the above `start`/`stopAndSum`/`cancel` API isn't practical to use from your
-application, you may use the alternative lower-level API `setRawNanos`. Unlike
-the above, this sets the value of the timespan directly, and does not sum it to
-the existing value, so use with caution if the span of the ping the metric is
-sent in may include multiple timings.
+application, such as using a third-party library that provides raw timespan
+values, you may use the alternative lower-level API `sumRawNanos` and `setRawNanos`.
+
+`sumRawNanos` adds the timespan to any existing timespans already recorded since the last time the ping was sent.
+
+`setRawNanos` directly sets the timespan value, erasing any timespans recorded since the last time the ping was sent.  
+Care should be taken using this if the ping lifetime might contain more than one timespan measurement.  
+To be safe, `setRawNanos` should generally be followed by sending a custom ping containing the timespan. 
 
 ```Kotlin
 import org.mozilla.yourApplication.GleanMetrics.Auth
--- a/components/service/glean/docs/metrics/adding-new-metrics.md
+++ b/components/service/glean/docs/metrics/adding-new-metrics.md
@@ -434,7 +434,6 @@ values, you may use the alternative lower-level API `sumRawNanos` and `setRawNan
 
 `setRawNanos` directly sets the timespan value, erasing any timespans recorded since the last time the ping was sent.  
 Care should be taken using this if the ping lifetime might contain more than one timespan measurement.  
-To be safe, `setRawNanos` should generally be followed by sending a custom ping containing the timespan. 
 
 ```Kotlin
 import org.mozilla.yourApplication.GleanMetrics.Auth
--- a/components/service/glean/docs/metrics/adding-new-metrics.md
+++ b/components/service/glean/docs/metrics/adding-new-metrics.md
@@ -426,13 +426,11 @@ fun onLoginCancel(e: Event) {
 The time reported in the telemetry ping will be the sum of all of these
 timespans recorded during the lifetime of the ping.
 
-If the above `start`/`stopAndSum`/`cancel` API isn't practical to use from your
-application, such as using a third-party library that provides raw timespan
-values, you may use the alternative lower-level API `sumRawNanos` and `setRawNanos`.
+If the above `start()`/`stopAndSum()`/`cancel()` API isn't practical to use from your application, such as using a third-party library that provides raw timespan values, you may use the alternative lower-level API `sumRawNanos()` and `setRawNanos()`.
 
-`sumRawNanos` adds the timespan to any existing timespans already recorded since the last time the ping was sent.
+`sumRawNanos()` adds the timespan to any existing timespans already recorded since the last time the ping was sent.
 
-`setRawNanos` directly sets the timespan value, erasing any timespans recorded since the last time the ping was sent.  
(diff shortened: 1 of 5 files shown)


components/service/glean/docs/metrics/adding-new-metrics.md
```diff
@@ -343,19 +343,34 @@ fun onShowLogin(e: Event) {
...
-    Auth.loginTime.cancel(e.tart)
+    Auth.loginTime.cancel(e.target)
     // ...
 }
 ```
 
 The time reported in the telemetry ping will be the sum of all of these
 timespans recorded during the lifetime of the ping.
 
+If the above `start`/`stopAndSum`/`cancel` API isn't practical to use from your
```
Dexterp37: I would be very explicit here: "If the above `start`/`stopAndSum`/`cancel` API can't be used due to restrictions (e.g. using 3rd party library that is providing raw timespan values), you may use..."

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimespansStorageEngine.kt
```diff
@@ -121,6 +121,30 @@ internal open class TimespansStorageEngineImplementation(
...
+        timeUnit: TimeUnit,
+        elapsedNanos: Long
+    ) {
+        // Look for the start time: if it's there, commit the timespan.
+        val timespanName = metricData.identifier
+
+        // Store the time unit: we'll need it when snapshotting.
+        timeUnitsMap[timespanName] = timeUnit
+
+        super.recordMetric(metricData, elapsedNanos, timeUnit) { currentValue, newValue ->
```
Dexterp37: I think something will complain about `currentValue` not being used. Please replace it with `_`.

components/service/glean/src/test/java/mozilla/components/service/glean/private/TimespanMetricTypeTest.kt
```diff
@@ -144,4 +144,25 @@ class TimespanMetricTypeTest {
         assertTrue(metric.testGetValue("store2") >= 0)
         assertEquals(1, testGetNumRecordedErrors(metric, ErrorType.InvalidValue))
     }
+
+    @Test
+    fun `test setRawNanos`() {
```
Dexterp37: Can you add a separate test to check that when using both APIs things don't go wrong? Looks like it shouldn't, but I'd rather be sure about it.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimespansStorageEngine.kt
```diff
@@ -121,6 +121,30 @@ internal open class TimespansStorageEngineImplementation(
         }
     }
 
+    /**
+     * Set the elapsed time explicitly.
```
Dexterp37: What happens if we mistakenly use this API and the usual API on the same metric? Will it break? Should we take care of that or at least document that the APIs are meant to be mutually exclusive?

components/service/glean/docs/metrics/adding-new-metrics.md
```diff
@@ -343,19 +343,34 @@ fun onShowLogin(e: Event) {
...
+application, you may use the alternative lower-level API `setRawNanos`. Unlike
+the above, this sets the value of the timespan directly, and does not sum it to
+the existing value, so use with caution if the span of the ping the metric is
+sent in may include multiple timings.
```
georgf: This difference in behavior between the two ways to use the APIs is confusing.
We should follow up soon to have one consistent behavior here ([see this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1556798)) or make it consistent in this PR.

In general, if we have to call out inconsistent behavior like that, for me that's a flag for "does the API or design need changing?".

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimespansStorageEngine.kt
mdboom: It won't "break".  Using the `set` API after the `sum` API will override the value.  Using the `sum` API after the `set` API will add to the value.  Do we want / need to provide stricter assurances?  (I agree on making the documentation language stronger on this point).

components/service/glean/docs/metrics/adding-new-metrics.md
mdboom: As explained in the bug, I **strongly** think it would be wrong to change the existing API.  I'd be fine with changing `setRawNanos` to `addRawNanos` to be consistent.  I also think documenting that `setRawNanos` is only really safe if the ping is sent immediately after recording the metric would be a good idea.

Dexterp37: I'm fine with the additions here, the APIs are now convergent. @mdboom can you expand a bit on why `setRawNanos` is only safe if the ping is sent after recording? Is it because adding could change the value? If so, I'd rather phrase it differently, like "when 'setting' make sure no other operation takes places that modifies the value of the metric" :) From the usage seen so far, I suspect that's not too much of a big risk, but I agree it's worth documenting.

mdboom: Maybe using the word "safe" is confusing.  What I mean is -- if you use `setRawNanos` and then `setRawNanos` again before sending, that first measurement is totally gone.  Maybe someone would intend to do that, but it seems more likely that that would be a bug.  If you `setRawNanos` and then send the ping right away, you avoid that problem.  (It looks like the sync telemetry work is likely to do that, so at least in that instance it's doing the right thing).  I'm also fine to remove the "To be safe..." sentence altogether if it just adds confusion rather than clarifies.

Dexterp37: Oh, I see. Yes, I'd rather remove the "To be safe.." then.

georgf: For this issue, changing the API is a risk, so please treat my points as "follow-up discussion".
We have an active discussion on the semantics on bugzilla, we can continue there.

components/service/glean/docs/metrics/adding-new-metrics.md
```diff
@@ -343,19 +343,37 @@ fun onShowLogin(e: Event) {
...
     // ...
 }
 ```
 
 The time reported in the telemetry ping will be the sum of all of these
 timespans recorded during the lifetime of the ping.
 
+If the above `start`/`stopAndSum`/`cancel` API isn't practical to use from your
+application, such as using a third-party library that provides raw timespan
+values, you may use the alternative lower-level API `sumRawNanos` and `setRawNanos`.
```
georgf: Nit: `sumRawNanos()`, `setRawNanos()` to make clear those are methods/functions?
Ditto below.

Issue #3248, opened 2019-06-04
Emit fact when pressing custom tab share action button


sblatz:
Currently we're only emitting a fact for other action buttons, but share buttons need them too.

PR #3249, opened 2019-06-04
For #3248: Adds emit fact to custom tab share button

sblatz:
### Pull Request checklist


components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
```diff
@@ -146,7 +146,10 @@ class CustomTabsToolbarFeature(
         }
 
         val button = Toolbar.ActionButton(drawable, context.getString(R.string.mozac_feature_customtabs_share_link)) {
-            val listener = shareListener ?: { context.share(session.url) }
+            val listener = shareListener ?: {
+                emitActionButtonFact()
```
jonalmeida: This will emit the fact _only_ if we use the default share sheet and not if there is a custom one. Is that what was intended?

sblatz: Whoops. Good catch. Should be emitted any time--updated!

PR #3255, opened 2019-06-04
Fixes #3254: fixes memory leak in ObserverRegistry

### LINKED ISSUE #3254: Memory leak in ObserverRegistry
`ObserverRegistry` stores observers in a `WeakHashMap` and does not remove values explicitly, relying on the weak references for cleanup.  

Buried deep in the `WeakHashMap` javadoc is a warning that the “value objects in a `WeakHashMap` are held by ordinary strong references. Thus care should be taken to ensure that value objects do not strongly refer to their own keys, either directly or indirectly, since that will prevent the keys from being discarded.”

Values of `ObserverRegistry` maps (`LifecycleBoundObserver` and `ViewBoundObserver`) both hold strong references to their observers, preventing deallocation.  This is leaking memory on FFTV, which can be shown by comparing a heap dump from master (which depends on A-C 54) with one that depends on [this draft PR](https://github.com/mozilla-mobile/android-components/pull/3255).  It seems likely that this affects most if not all projects that depend upon A-C.

Baron-Severin:
### Pull Request checklist

### DIFF
--- a/components/support/base/src/main/java/mozilla/components/support/base/observer/ObserverRegistry.kt
+++ b/components/support/base/src/main/java/mozilla/components/support/base/observer/ObserverRegistry.kt
@@ -84,8 +84,12 @@ class ObserverRegistry<T> : Observable<T> {
             pausedObservers.remove(observer)
         }
 
+        // Unregister observers
         lifecycleObservers[observer]?.remove()
         viewObservers[observer]?.remove()
+        // Remove observers from map
+        lifecycleObservers.remove(observer)
+        viewObservers.remove(observer)
     }
 
     override fun unregisterObservers() {
@@ -95,6 +99,8 @@ class ObserverRegistry<T> : Observable<T> {
             }
             observers.clear()
             pausedObservers.clear()
+            lifecycleObservers.clear()
+            viewObservers.clear()
         }
     }
 
--- a/components/support/base/src/test/java/mozilla/components/support/base/observer/ObserverRegistryTest.kt
+++ b/components/support/base/src/test/java/mozilla/components/support/base/observer/ObserverRegistryTest.kt
@@ -182,12 +182,19 @@ class ObserverRegistryTest {
     @Test
     fun `unregisterObservers unregisters all observers`() {
         val registry = ObserverRegistry<TestObserver>()
+        val activity = Robolectric.buildActivity(Activity::class.java).create().get()
+        val view = View(testContext)
 
         val observer1 = TestObserver()
         val observer2 = TestObserver()
+        val observer3 = TestObserver()
+        val observer4 = TestObserver()
 
         registry.register(observer1)
         registry.register(observer2)
+        registry.register(observer3, MockedLifecycleOwner(Lifecycle.State.CREATED))
+        registry.register(observer4, view)
+        activity.windowManager.addView(view, WindowManager.LayoutParams(100, 100))
 
         assertFalse(observer1.notified)
         assertFalse(observer2.notified)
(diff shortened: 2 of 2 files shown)


components/support/base/src/main/java/mozilla/components/support/base/observer/ObserverRegistry.kt
```diff
@@ -22,8 +23,8 @@ import java.util.WeakHashMap
  */
 class ObserverRegistry<T> : Observable<T> {
     private val observers = mutableSetOf<T>()
-    private val lifecycleObservers = WeakHashMap<T, LifecycleBoundObserver<T>>()
-    private val viewObservers = WeakHashMap<T, ViewBoundObserver<T>>()
+    private val lifecycleObservers = WeakHashMap<T, WeakReference<LifecycleBoundObserver<T>>>()
+    private val viewObservers = WeakHashMap<T, WeakReference<ViewBoundObserver<T>>>()
```
pocmo: Do we really need a `WeakReference` here?

Wouldn't it be enough (and more explicit) to remove the entries in `unregister()`?

pocmo: And the explicit "remove" may be easier to test too :)

Baron-Severin:
> Great find!
> 
> For this to leak it seems like you also need app code to keep a strong reference to the observer (since that is the weak key). Is that the case in your app? Are you keeping an observer around? I wonder if there are follow-up leaks (of other things) in this case.
> 
> CC @csadilek since we were debugging a leak recently too and that may have contributed to it.

Thanks!

@pocmo We're not holding any strong refs to the observer, and when I build FFTV with this modified code our memory leak disappears.  It seems that the strongly referenced entry value holds a reference to its key (`protected val observer: T`), preventing it from being collected.

I'll make your suggested change, write some tests, and open this for review!

pocmo:
>  It seems that the strongly referenced entry value holds a reference to its key (protected val observer: T), preventing it from being collected.

Oh, yes, @csadilek mentioned that too today. With that we'll always keep the values :D

components/support/base/src/main/java/mozilla/components/support/base/observer/ObserverRegistry.kt
```diff
@@ -95,6 +99,8 @@ class ObserverRegistry<T> : Observable<T> {
             }
             observers.clear()
             pausedObservers.clear()
+            lifecycleObservers.clear()
+            viewObservers.clear()
```
pocmo: It looks like there's a second bug hiding here: First we go through all observers and remove a lifecycle observer if it exists, but we never do that for an potentially existing view observer. I'll file a follow-up issue for that. Maybe we should play safe and just call `unregister()` for every observer here. Anyhow, different issue!

components/support/base/src/main/java/mozilla/components/support/base/observer/ObserverRegistry.kt
```diff
@@ -84,8 +84,12 @@ class ObserverRegistry<T> : Observable<T> {
             pausedObservers.remove(observer)
         }
 
+        // Unregister observers
         lifecycleObservers[observer]?.remove()
         viewObservers[observer]?.remove()
+        // Remove observers from map
+        lifecycleObservers.remove(observer)
+        viewObservers.remove(observer)
```
csadilek: I think we'd need to move all these 4 four calls up into the synchronized block as the method claims to be thread-safe, but that was also already a problem before and can be a follow-up (currently only invoked on the main thread afaik).

csadilek: Follow-up: https://github.com/mozilla-mobile/android-components/issues/3277

components/support/base/src/main/java/mozilla/components/support/base/observer/ObserverRegistry.kt
csadilek: Follow-up: https://github.com/mozilla-mobile/android-components/issues/3278

Issue #2041, opened 2019-02-13
browser-icons: Create processor for resizing images


pocmo:
Loaded icons can have arbitrary sizes. We want a `Processor` in the pipeline that can resize them to the desired size. See code in Fennec. There's some logic to avoid scaling up images to much and making them blurry.

PR #3263, opened 2019-06-05
Read web app manifest in browser-icons

NotWoods:
Updates browser-icons so it also checks the Web App Manifest along with other sources. This will be used for PWAs later too.

Large icons are now scaled rather than thrown out, since [some sites](https://proxx.app/) only supply large icons.

### Pull Request checklist

### DIFF
--- a/components/browser/icons/src/main/java/mozilla/components/browser/icons/pipeline/IconResourceComparator.kt
+++ b/components/browser/icons/src/main/java/mozilla/components/browser/icons/pipeline/IconResourceComparator.kt
@@ -5,7 +5,6 @@
 package mozilla.components.browser.icons.pipeline
 
 import mozilla.components.browser.icons.IconRequest
-import kotlin.math.min
 
 /**
  * This [Comparator] implementations compares [IconRequest.Resource] objects to determine which icon to try to load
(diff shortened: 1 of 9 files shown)


components/browser/icons/src/main/java/mozilla/components/browser/icons/decoder/AndroidIconDecoder.kt
```diff
@@ -35,19 +36,27 @@ class AndroidIconDecoder : IconDecoder {
...
                 bitmap.width * (1.0f / maxScaleFactor) > maxSize ||
                     bitmap.height * (1.0f / maxScaleFactor) > maxSize -> {
-                    logger.debug("BitmapFactory returned way too large image")
-                    null
+                    val scale = maxSize.toFloat() / max(bitmap.width, bitmap.height)
+                    createScaledBitmap(
+                        bitmap,
+                        (bitmap.width * scale).toInt(),
+                        (bitmap.height * scale).toInt()
+                    )
```
pocmo: Please let's not scale here. This is planned to be added as an `IconProcessor` in #2041.

components/browser/icons/src/main/java/mozilla/components/browser/icons/pipeline/IconResourceComparator.kt
```diff
@@ -41,9 +41,11 @@ internal object IconResourceComparator : Comparator<IconRequest.Resource> {
     }
 }
 
-@Suppress("MagicNumber")
+@Suppress("MagicNumber", "ComplexMethod")
 private fun IconRequest.Resource.Type.rank(): Int {
     return when (this) {
+        IconRequest.Resource.Type.MANIFEST_MASKABLE_ICON -> 25
+        IconRequest.Resource.Type.MANIFEST_ICON -> 20
```
pocmo: This needs further discussion. I'm not convinced that we should use the manifest icon everywhere in the app UI. Is that a good choice?

NotWoods:
I'll wait until after we pin the A-C version in Fenix to merge this.

components/browser/icons/src/main/java/mozilla/components/browser/icons/extension/IconMessageHandler.kt
```diff
@@ -30,7 +30,16 @@ internal class IconMessageHandler(
...
+            val messageRequest = message.toIconRequest()
+            val manifestRequest = session.webAppManifest?.toIconRequest()
+            val request = if (messageRequest != null && manifestRequest != null) {
+                // Merge the request resources
+                messageRequest.copy(resources = messageRequest.resources + manifestRequest.resources)
+            } else {
+                // Use the not-null request, if any
+                messageRequest ?: manifestRequest
+            }
+            loadRequest(request)
```
pocmo: This part I'm not sure about:
* It mixes the manifest icon into the list and we will try to load it and use it in the app UI. Do we actually need/want to do that? We can still load the manifest icon once we actually need it for an app launcher icon.
* There's now some race here: There's no guarantee that the manifest will be set on the `Session` by the time we receive this message from the icons add-on.

NotWoods: I think we do want to mix in the manifest icons. The manifest represents replacing apple icons with a standard spec, so if we want to display high res apple icons we should want to display high res manifest icons.

NotWoods: I don't have an idea yet how to deal with the race condition.

pocmo: Can we split this into a separate issue and look at that isolated (using manifest icons during normal icon load for a URL)? :)

NotWoods: Filed #3642

components/concept/engine/src/main/java/mozilla/components/concept/engine/manifest/Size.kt
```diff
@@ -15,6 +18,11 @@ data class Size(
     val width: Int,
     val height: Int
 ) {
+
+    val maxLength get() = max(width, height)
+
+    val minLength get() = min(width, height)
```
pocmo: nit: kdoc? :)

PR #3266, opened 2019-06-05
 Closes #2916: Improve BrowserMenu popup orientation..

### LINKED ISSUE #2916: Dynamically set menu direction in MenuBuilder
When I set the the browser menu direction of a `MenuBuilder` to be `UP` it is not respected. It seems like only `DOWN` (the default param) is ever set.

Amejia481:
### Pull Request checklist


components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenu.kt
```diff
@@ -90,12 +90,20 @@ class BrowserMenu internal constructor(
 
     enum class Orientation {
         UP,
-        DOWN
+        DOWN,
+        // Chooses the best orientation for a popup menu. It will appear below the anchor view if there is room,
+        // or above it if there is not.
+        DYNAMIC
```
pocmo: If you turn this comment into KDoc then it will show up here:
https://mozac.org/api/mozilla.components.browser.menu/-browser-menu/-orientation/

components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenu.kt
```diff
@@ -105,3 +113,28 @@ private fun determineVerticalOffset(orientation: BrowserMenu.Orientation, view:
...
+            BrowserMenu.Orientation.UP
+        } else {
+            BrowserMenu.Orientation.DOWN
+        }
+    }
+
+private val View.isCloseToTheBottomOfTheScreen: Boolean
+    get() {
+        val screenHeight = context.resources.displayMetrics.heightPixels
+        return screenLocationY > (screenHeight / 2)
```
pocmo: Does using `heightPixels` still work when using split screen mode? Since the other value is "location in window" I wonder if you need the height of the "window" here.

components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenu.kt
Amejia481: 👍

sblatz:
I'm in agreement with @jhugman. I think my biggest concern is that this doesn't actually check the height of the menu to be drawn, and thus is not guaranteed to fit in the direction selected.

components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenu.kt
```diff
@@ -27,34 +26,33 @@ class BrowserMenu internal constructor(
     private var currentPopup: PopupWindow? = null
     private var menuList: RecyclerView? = null
 
+    @Suppress("UNUSED_PARAMETER")
```
Amejia481: Used to avoid compilation errors while testing. It will be removed.

Amejia481:
@jhugman, @sblatz  and @pocmo  thank for your comments, they made me realize that maybe we really don't care in which direction the menu is shown as long as is shown accordingly to the available space on the screen.

I found a more clear alternative and I would like to know what do you think about it.

 If we provide the `height` of the `content` to the `PopupWindow` and use the `showAsDropDown` function without passing any parameters, it can calculate by itself the proper popup dimension and orientation without any interference from our side, as the documentations suggests.

[PopupWindow#showAsDropDown(android.view.View)](https://developer.android.com/reference/android/widget/PopupWindow#showAsDropDown(android.view.View))

> Display the content view in a popup window anchored to the bottom-left corner of the anchor view. If there is not enough room on screen to show the popup in its entirety, this method tries to find a parent scroll view to scroll. If no parent scroll view can be scrolled, the bottom-left corner of the popup is pinned at the top left corner of the anchor view.

I adapted the PR to use this approuch,  and tested thoroughly on Fenix. This fixes all the issues that we have related to it mozilla-mobile/fenix#2162, mozilla-mobile/fenix#986, mozilla-mobile/fenix#2825 and also I tested using **split screen mode**, and worked well too. Please take a look at [this demo video](https://drive.google.com/file/d/13NQjPjrEJSMVIILtIs0Nda5q47M2lXXr/view?usp=sharing) and feel free of playing locally with this branch.

components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenu.kt
```diff
@@ -27,34 +26,33 @@ class BrowserMenu internal constructor(
...
     @SuppressLint("InflateParams")
     fun show(anchor: View, orientation: Orientation = Orientation.DOWN): PopupWindow {
```
Amejia481: If you agree with this approach, I don't think we will need to provide an `orientation` parameter anymore. As the popup will be responsible for laying out the orientation and dimension. We can delete everything related to it.

components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenu.kt
Amejia481: I tested using `heightPixels` with the split screen mode and it returns the actual height of the windows. For example, I changed the app size and it was getting updated as I was increasing or reducing the app size.

pocmo:
@Amejia481 The offsets existed on purpose to draw the menu **on top of** the menu button and not below or above. This gets more apparent if you look at a menu shown for the toolbar.

pocmo:
@NotWoods also just landed some animations in e6df84c330e483cddcfed76997d29da4231c281a that depend on UP/DOWN. :)

jhugman:
I think at this stage, let's check if we can draw/animate the menu in the direction specified, and fallback to the opposite if not. If not the opposite, then go the preferred way with a scroll.

This would mean the API doesn't change, just that it is slightly more robust (which I think was the intent of this issue).

Amejia481:
PR updated following the feedback!

components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenu.kt
```diff
@@ -98,14 +97,101 @@ class BrowserMenu internal constructor(
...
+    val spec = View.MeasureSpec.makeMeasureSpec(0, View.MeasureSpec.UNSPECIFIED)
+    containerView.measure(spec, spec)
+
+    val topAndBottomPair = getMaxAvailableHeightToTopAndBottom(anchor)
+    val availableHeightToTop = topAndBottomPair.first
+    val availableHeightToBottom = topAndBottomPair.second
+    val containerHeight = containerView.measuredHeight
+
+    val fitsUp = availableHeightToTop >= containerHeight
+    val fitsDown = availableHeightToBottom >= containerHeight
```
jhugman: 😻

PR #3295, opened 2019-06-06
Closes #2041 - Add ResizingProcessor

### LINKED ISSUE #2041: browser-icons: Create processor for resizing images

NotWoods:
Ports processors from Fennec to A-C.

Additionally updated behaviour so that a generated bitmap is used in place of a null bitmap in BrowserIcons.

### Pull Request checklist


components/browser/icons/src/main/java/mozilla/components/browser/icons/processor/ColorProcessor.kt
```diff
@@ -0,0 +1,87 @@
...
+import androidx.annotation.ColorInt
+import androidx.palette.graphics.Palette
+import mozilla.components.browser.icons.Icon
+import mozilla.components.browser.icons.IconRequest
+import kotlin.math.max
+
+/**
+ * [IconProcessor] implementation to extract the dominant color from the icon.
+ */
+class ColorProcessor : IconProcessor {
```
pocmo: It turned out that Fenix didn't need that and therefore we never implemented that. I wonder if we should split this PR into two and maybe hold back on landing this part?

components/browser/icons/src/main/java/mozilla/components/browser/icons/processor/ResizingProcessor.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    companion object {
+        /**
+         * This is the largest factor we'll scale up an image by: the goal is an image
+         * that both fills the top site space well but does not have extreme resizing
+         * artifacts. This number was chosen anecdotally by comparing variously-sized
+         * favicons across devices to see which factor(s) looked the best. bug 1398970
+         * is filed to take a more comprehensive approach to favicons.
+         */
+        private const val MAX_SCALE_FACTOR = 3
```
pocmo: This introduces a new constant; although we already have `MAXIMUM_SCALE_FACTOR` in BrowserIcons which we pass to the decoders.

components/browser/icons/src/main/java/mozilla/components/browser/icons/processor/ResizingProcessor.kt
```diff
@@ -0,0 +1,71 @@
...
+import android.graphics.Bitmap
+import androidx.annotation.VisibleForTesting
+import mozilla.components.browser.icons.Icon
+import mozilla.components.browser.icons.IconRequest
+import mozilla.components.support.ktx.android.content.res.pxToDp
+
+/**
+ * [IconProcessor] implementation for resizing the loaded icon based on the target size.
+ */
+class ResizingProcessor(
```
pocmo: You haven't added this to the list of processors in `BrowserIcons` - was this intentional?

components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt
```diff
@@ -112,7 +112,9 @@ class BrowserIcons(
         }
 
         // (3) Finally process the icon.
-        return process(context, processors, request, resource, icon)
+        return process(context, processors, request, resource, icon).run {
+            if (bitmap != null) this else generator.generate(context, request)
+        }
```
pocmo: Did you see the bitmap being null here? Just curious since the previous code should already have returned an `Icon` with bitmap.

pocmo: Looking more at it almost looks like we can make `Icon.bitmap` not nullable.

NotWoods: `ResizingProcessor` returns a null bitmap when the icon is too small to indicate that a generated icon should be used instead. We can opt to have a special variant of `Icon` with a nullable bitmap or let process return a `Icon?`.

components/browser/icons/src/main/java/mozilla/components/browser/icons/processor/ResizingProcessor.kt
NotWoods: This was intentional since you asked [not to introduce new icon behaviour](https://github.com/mozilla-mobile/android-components/pull/3263#pullrequestreview-246069317) previously.

components/browser/icons/src/main/java/mozilla/components/browser/icons/processor/ResizingProcessor.kt
NotWoods: I had to recompute `targetSize` as well. Should `process` be changed to take these variable as parameters, or alternatively should we add new fields to `IconRequest`? Could also just make `MAXIMUM_SCALE_FACTOR` internal instead of private.

components/browser/icons/src/main/java/mozilla/components/browser/icons/processor/ColorProcessor.kt
NotWoods: I can split this out, I'll need it for adaptive launcher icons.

components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt
NotWoods: After playing around with it, I think letting process return an `Icon?` is the best way to do this. I updated `Icon.bitmap` to be not-null.

components/browser/icons/src/main/java/mozilla/components/browser/icons/processor/ResizingProcessor.kt
pocmo: For now I'd just pass the to `process()` and not add them to the public API of `IconRequest`.

PR #3296, opened 2019-06-06
For mozilla-mobile/fenix#3186: Support passing a Sentry environment

colintheshots:
### Pull Request checklist


components/lib/crash/src/main/java/mozilla/components/lib/crash/service/SentryService.kt
```diff
@@ -35,6 +37,7 @@ class SentryService(
             .build()
             .toString(),
         clientFactory).apply {
+        environment = env
```
pocmo: nitpicking level over 9000: If this is named `environment` in `SentryClient` maybe we should name it that too for consistency.

colintheshots:
Everything appears to be working, but a random unit test failure. I think TaskCluster just needs to re-run.

PR #3298, opened 2019-06-06
For #3214: Sets url and title text size to use dp if at max

sblatz:
Without this, the text size gets set way too small

### Pull Request checklist

### DIFF
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
@@ -219,7 +219,14 @@ class BrowserToolbar @JvmOverloads constructor(
         get() = displayToolbar.titleView.textSize
         set(value) {
             val newSize = minOf(value, MAX_TITLE_SIZE)
-            displayToolbar.titleView.textSize = newSize
+            // We only want the text size to scale smaller (with sp) if it's below the max size
+            val unitType = if (newSize == MAX_TITLE_SIZE) {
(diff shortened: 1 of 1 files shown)


jonalmeida:
There's something else happening here: the tests in the previous PR and this one should be failing during assertions. When you use `setTextSize` it uses 'sp' are the unit type, but when you call `getTextSize` it gives you the value in raw pixels. See: docs for [get](https://developer.android.com/reference/android/widget/TextView#getTextSize()) and [set](https://developer.android.com/reference/android/widget/TextView.html#setTextSize(float)).

Maybe this is related to why the initial fix didn't work?

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
```diff
@@ -219,7 +219,14 @@ class BrowserToolbar @JvmOverloads constructor(
         get() = displayToolbar.titleView.textSize
         set(value) {
             val newSize = minOf(value, MAX_TITLE_SIZE)
-            displayToolbar.titleView.textSize = newSize
+            // We only want the text size to scale smaller (with sp) if it's below the max size
+            val unitType = if (newSize == MAX_TITLE_SIZE) {
+                TypedValue.COMPLEX_UNIT_DIP
+            } else {
+                TypedValue.COMPLEX_UNIT_SP
+            }
```
pocmo: Something is not right here (even though the result may look okay on your device).

First we compare values (`minOf`), then later on we use different units -> The comparison isn't comparing the same thing. This seems wrong. Do we need to have both values converted to actual pixels before we can compare? Or am I missing something?

PR #3317, opened 2019-06-07
Fixes #162 - Replace Drawable.toBitmap with core-ktx

### LINKED ISSUE #162: Remove Drawable.toBitmap in favor of official KTX version
https://developer.android.com/reference/kotlin/androidx/core/graphics/drawable/package-summary#(android.graphics.drawable.Drawable).toBitmap(kotlin.Int,+kotlin.Int,+android.graphics.Bitmap.Config)

Unclear what the differences in implementation could be.

NotWoods:

`toBitmap()` in core-ktx is identical, except that it also lets you create a Bitmap of a different size.
Also replaced `systemService` with `getSystemService`.

### Pull Request checklist


components/support/ktx/src/main/java/mozilla/components/support/ktx/android/org/json/JSONObject.kt
```diff
@@ -11,34 +11,19 @@ import java.util.TreeMap
...
  * Returns the value mapped by {@code key} if it exists, and
  * if the value returned is not null. If it's null, it returns null
  */
-fun JSONObject.tryGetInt(key: String): Int? {
-    if (!isNull(key)) {
-        return getInt(key)
-    }
-    return null
-}
+fun JSONObject.tryGetInt(key: String): Int? = if (isNull(key)) null else getInt(key)
```
jhugman: Nit: This will still throw a `JSONException` if the value is non-null and cannot be coerced into an `Int`.

If `org.json.JSON` is visible, you may be better off doing: 

```kotlin
fun JSONObject.tryGetInt(key: String): Int? = JSON.toInteger(opt(key))
```

NotWoods: Looks like `org.json.JSON` is not visible. I'm going to leave the current implementation since its what we had before.

PR #3319, opened 2019-06-07
Replace String.toUri with core-ktx

NotWoods:

`toUri()` in core-ktx is identical.

### Pull Request checklist


jonalmeida:
Did you want to alias/remove our extension function as well?

https://github.com/mozilla-mobile/android-components/blob/master/components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlin/String.kt#L55

components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlin/String.kt
```diff
@@ -42,18 +42,13 @@ fun String.isGeoLocation() = re.geoish.matches(this)
...
         formatter.parse(this)
     } else {
         Date()
     }
 }
 
-/**
- * Converts a [String] to a [Uri] object.
- */
-fun String.toUri() = Uri.parse(this)
```
NotWoods: The A-C extension function was removed

docs/changelog.md
```diff
@@ -175,6 +175,7 @@ permalink: /changelog/
 
 * **support-ktx**
   * Added `Context.hasCamera()` to check if the device has a camera.
+  * Removed `String.toUri()` in favour of Android Core KTX.
```
dector: Should it be marked as Breaking change (as updating to new version will lead to compile-time errors)?

dector: Oh, and seems like it's need to be moved to `0.56.0` version.

Amejia481: > Should it be marked as Breaking change (as updating to new version will lead to compile-time errors)?

Yes, also we normally use this format to be more noticeable, for consumers:
⚠️ **This is a breaking change:**

components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlin/String.kt
jonalmeida: Oops thanks, I didn't catch that!

pocmo: Let's deprecate first and allow for less API breakage. Whistler would be a good time to discuss a process around deprecating and removing functionality. It's something that have come up more often recently, but where we do not have any process guidelines.

NotWoods: I was told by @csadilek that deprecating also causes API breakage because lint will fail. Should I hold off on this refactoring until another time?

pocmo: Oh yeah, lint is super annoying. In other projects with strict rules failing lint is fine; however it breaks our own code and tests if we still use the old method anywhere, but I think we do not since you migrated that?

dector: Lint suppression is much easier than applying changes to code (I need to find what I should use instead and apply it around the project - it can be extremely painful).

Luckily, Kotlin has deprecation on steroids:

```kotlin
@Deprecated("Use `newImplementation()` instead", ReplaceWith("newImplementation(args)", "mozilla.x"))
fun oldImplementationWith(args) {
}
```

that will allow clients to change code in the whole project with two clicks. :)

components/support/ktx/src/test/java/mozilla/components/support/ktx/kotlin/StringTest.kt
```diff
@@ -84,30 +83,6 @@ class StringTest {
         assertTrue("Geo:1,-1".isGeoLocation())
     }
 
-    @Test
-    fun toUri() {
```
pocmo: Nit: As long as the extension method exists (even though `@Deprecated`), we should keep the test too.

PR #3330, opened 2019-06-10
Add check to ensure Glean is only initialized on the main process.

travis79:
Trying to initialize Glean on a process other than the main process results in an error when Glean tries to use WorkManager because WorkManager is only initialized on the main process.  This checks whether initialize is being called by the main process and returns if it isn't.

### Pull Request checklist


components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -88,6 +89,13 @@ open class GleanInternalAPI internal constructor () {
         applicationContext: Context,
         configuration: Configuration = Configuration()
     ) {
+        // In certain situations Glean.initialize may be called from a process other than the main
+        // process.  In this case we want initialize to be a no-op and just return.
+        if (!applicationContext.isMainProcess()) {
+            logger.error("Attempted to initialize Glean on a process other than the main process")
```
Dexterp37: @travis79 can you make sure that it's ok to call the `Logger` on other processes?

travis79: After experimenting with this, it appears to log as intended.

dector:
Just wondering if runtime exception should be thrown as well?
If Glean component is required to be initialized in main thread - doing this in another thread seems like configuration error. :thinking:

dector:
Awesome, thanks!

travis79:
@dector I've filed an investigation bug [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1558181) in case you are interested in following it.  It's sort of a broad bug for now to cover a full investigation on the pros and cons and other pitfalls of using Glean in a multi-process application.  It will likely result in more specific tasks as we dig into it.

dector:
@travis79 Great, thank you! I've created follow-up issue in AC project: #3342 .

Issue #2605, opened 2019-04-02
Resources.pxToDp is actually dpToPx


mcomella:
`Resources.pxToDp` converts a value in DIP to its pixels values: the naming is incorrect (though the kdoc correctly explain what this method does).

It looks like I accidentally suggested the wrong name in #754 and it got implemented.

We should do this sooner, rather than later, so we can introduce a real `pxToDp` method later (for consistency) :).

NotWoods:
I'll do this refactor soon after the codebase settles down

PR #3335, opened 2019-06-10
Reduce references to Context in browser-icons

NotWoods:
### Pull Request checklist

### DIFF
--- a/components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt
+++ b/components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt
@@ -36,7 +36,7 @@ import mozilla.components.browser.session.utils.AllSessionsObserver
 import mozilla.components.concept.engine.Engine
 import mozilla.components.concept.fetch.Client
 import mozilla.components.support.base.log.logger.Logger
-import mozilla.components.support.ktx.android.content.res.pxToDp
+import mozilla.components.support.ktx.android.util.pxToDp
 import java.util.concurrent.Executors
 
 private const val MAXIMUM_SIZE_DP = 64
@@ -57,7 +57,7 @@ internal val sharedDiskCache = DiskCache()
 class BrowserIcons(
     private val context: Context,
     private val httpClient: Client,
-    private val generator: IconGenerator = DefaultIconGenerator(context),
+    private val generator: IconGenerator = DefaultIconGenerator(context.resources.displayMetrics),
     private val preparers: List<IconPreprarer> = listOf(
         MemoryIconPreparer(sharedMemoryCache),
         DiskIconPreparer(sharedDiskCache)
@@ -79,7 +79,7 @@ class BrowserIcons(
     jobDispatcher: CoroutineDispatcher = Executors.newFixedThreadPool(THREADS).asCoroutineDispatcher()
 ) {
     private val logger = Logger("BrowserIcons")
-    private val maximumSize = context.resources.pxToDp(MAXIMUM_SIZE_DP)
+    private val maximumSize = context.resources.displayMetrics.pxToDp(MAXIMUM_SIZE_DP)
     private val scope = CoroutineScope(jobDispatcher)
 
     /**
@@ -92,7 +92,7 @@ class BrowserIcons(
     }
 
     private fun loadIconInternal(initialRequest: IconRequest): Icon {
-        val targetSize = context.resources.pxToDp(initialRequest.size.value)
+        val targetSize = context.resources.displayMetrics.pxToDp(initialRequest.size.value)
 
         // (1) First prepare the request.
         val request = prepare(context, preparers, initialRequest)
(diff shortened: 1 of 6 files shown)


components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt
```diff
@@ -79,7 +79,7 @@ class BrowserIcons(
     jobDispatcher: CoroutineDispatcher = Executors.newFixedThreadPool(THREADS).asCoroutineDispatcher()
 ) {
     private val logger = Logger("BrowserIcons")
-    private val maximumSize = context.resources.pxToDp(MAXIMUM_SIZE_DP)
+    private val maximumSize = context.resources.displayMetrics.pxToDp(MAXIMUM_SIZE_DP)
```
jhugman: This code doesn't look right. I understand that this is a fairly mechanical refactor. 

The function `pxToDp` would appear to take a `px` value and spit out a `dp` value. This line appears to give that function a `dp`.

I think `git blame` and ask whoever wrote this line originally.

components/support/ktx/src/main/java/mozilla/components/support/ktx/android/content/res/Resources.kt
```diff
@@ -1,10 +1,13 @@
...
 import android.content.res.Resources
-import android.util.TypedValue
+import mozilla.components.support.ktx.android.util.pxToDp
 
 /**
  * Converts a value in density independent pixels (pxToDp) to the actual pixel values for the display.
  */
-fun Resources.pxToDp(pixels: Int) = TypedValue.applyDimension(
-        TypedValue.COMPLEX_UNIT_DIP, pixels.toFloat(), displayMetrics).toInt()
+fun Resources.pxToDp(pixels: Int) = displayMetrics.pxToDp(pixels)
```
jhugman: Here is the problem. The documentation and the method name do not match. Which one is correct?

components/support/ktx/src/main/java/mozilla/components/support/ktx/android/util/DisplayMetrics.kt
```diff
@@ -0,0 +1,14 @@
...
+package mozilla.components.support.ktx.android.util
+
+import android.util.DisplayMetrics
+import android.util.TypedValue
+
+/**
+ * Converts a value in density independent pixels (pxToDp) to the actual pixel values for the display.
+ */
+fun DisplayMetrics.pxToDp(pixels: Int) = TypedValue.applyDimension(
+    TypedValue.COMPLEX_UNIT_DIP, pixels.toFloat(), this).toInt()
```
jhugman: The `TypeValue.applyDimension` docs suggest that this the method name should be `dpToPx`.

We should file an issue and fix this. 

[Units are hard][1].

[1]: http://edition.cnn.com/TECH/space/9909/30/mars.metric.02/

components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt
NotWoods: #754 is where it was decided on. I'll ask about the name.

NotWoods:
I was going to file a follow-up issue, but [turns out it already exists](https://github.com/mozilla-mobile/android-components/issues/2605).

components/browser/icons/src/main/java/mozilla/components/browser/icons/utils/DiskCache.kt
```diff
@@ -61,11 +61,11 @@ internal class DiskCache :
         return emptyList()
     }
 
-    internal fun putResources(context: Context, request: IconRequest) {
+    internal fun putResources(cacheDir: File, request: IconRequest) {
```
pocmo: As mentioned in the other PR: I do not think I agree with such a refactoring: `Context` is needed to lookup a  directory. Passing in a `File` is way more generic and requires the caller to know which File to pass in (which is meant to be an implementation detail). This doesn't protect against a caller providing wrong/other directories.

NotWoods: For private/internal functions, is that an issue?

pocmo: I agree that it's less an issue than a public API - but I expect many people to work on that code over time... :)

components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt
dector: May I suggest API like:

```kotlin
private val maximumSize = MAXIMUM_SIZE_DP.pxToDpWith(context)
```

?

NotWoods:
Closing as many of the changes landed in other PRs

PR #3336, opened 2019-06-10
Reduce references to Context in feature-tab-collections

NotWoods:
### Pull Request checklist

### DIFF
--- a/components/feature/tab-collections/src/main/java/mozilla/components/feature/tab/collections/TabCollectionStorage.kt
+++ b/components/feature/tab-collections/src/main/java/mozilla/components/feature/tab/collections/TabCollectionStorage.kt
@@ -17,14 +17,16 @@ import mozilla.components.feature.tab.collections.db.TabCollectionDatabase
 import mozilla.components.feature.tab.collections.db.TabCollectionEntity
 import mozilla.components.feature.tab.collections.db.TabEntity
 import mozilla.components.support.ktx.java.io.truncateDirectory
+import java.io.File
 import java.util.UUID
 
 /**
  * A storage implementation that saves snapshots of tabs / sessions in named collections.
  */
 class TabCollectionStorage(
-    private val context: Context,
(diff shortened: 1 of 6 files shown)


components/feature/tab-collections/src/main/java/mozilla/components/feature/tab/collections/TabCollectionStorage.kt
```diff
@@ -23,9 +23,10 @@ import java.util.UUID
  * A storage implementation that saves snapshots of tabs / sessions in named collections.
  */
 class TabCollectionStorage(
-    private val context: Context,
+    context: Context,
     private val sessionManager: SessionManager
 ) {
+    private val filesDir = context.filesDir
```
jhugman: Nit: prefer this as an optional constructor param. 

i.e. 

```kotlin
class TabCollectionStorage(
    context: Context,
    private val filesDir: String = context.filesDir,
    private val sessionManager: SesionManager
) {
```

components/feature/tab-collections/src/main/java/mozilla/components/feature/tab/collections/db/TabEntity.kt
```diff
@@ -49,13 +48,13 @@ internal data class TabEntity(
     @ColumnInfo(name = "created_at")
     var createdAt: Long
 ) {
-    internal fun getStateFile(context: Context): AtomicFile {
-        return AtomicFile(File(getStateDirectory(context), stateFile))
+    internal fun getStateFile(filesDir: File): AtomicFile {
+        return AtomicFile(File(getStateDirectory(filesDir), stateFile))
```
pocmo: I'm not sure about this refactoring. The API was meant to give you a specific file and the context was needed to lookup a directory. Now this is part of the public API and every caller needs to know which directory to pass in - and its possible to pass in other/wrong directories and files.

PR #3339, opened 2019-06-10
Closes #2043 - Add ColorProcessor from Fennec

### LINKED ISSUE #2043: browser-icons: Create processor for extracting primary color
Often in the UI we need a color matching the icon. For example in the past that has been used as background for too small icons.

Implement a `Processor` that uses the `palette` androidx library to extract a color.

NotWoods:
Ports processor from Fennec to A-C.

### Pull Request checklist


components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt
```diff
@@ -135,16 +137,11 @@ class BrowserIcons(
...
 }
 
-private fun prepare(context: Context, preparers: List<IconPreprarer>, request: IconRequest): IconRequest {
-    var preparedRequest: IconRequest = request
-
-    preparers.forEach { preparer ->
-        preparedRequest = preparer.prepare(context, preparedRequest)
+private fun prepare(context: Context, preparers: List<IconPreprarer>, request: IconRequest): IconRequest =
+    preparers.fold(request) { preparedRequest, preparer ->
+        preparer.prepare(context, preparedRequest)
```
jhugman: 😍

components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt
```diff
@@ -192,12 +189,7 @@ private fun process(
...
     resource: IconRequest.Resource?,
     icon: Icon
-): Icon {
-    var processedIcon = icon
-
-    processors.forEach { processor ->
-        processedIcon = processor.process(context, request, resource, processedIcon)
+): Icon =
+    processors.fold(icon) { processedIcon, processor ->
+        processor.process(context, request, resource, processedIcon)
```
jhugman: Yiss.

components/browser/icons/src/main/java/mozilla/components/browser/icons/processor/ColorProcessor.kt
```diff
@@ -0,0 +1,87 @@
...
+         * @param length The length of the array to be traversed. Must be smaller than, or equal to
+         * the total length of the array.
+         * @return The colour contained within the array, or null if colours vary.
+         */
+        @ColorInt
+        private fun getEdgeColorFromSingleDimension(edge: IntArray, length: Int): Int? {
+            @ColorInt val color = edge[0]
+            val allSameColor = edge.asSequence().drop(1).take(length - 1).all { it == color }
+            return if (allSameColor) color else null
+        }
```
jhugman: Nit: put these `private fun` functions into the processor itself, rather than the companion object.

components/browser/icons/src/main/java/mozilla/components/browser/icons/processor/ColorProcessor.kt
```diff
@@ -0,0 +1,87 @@
...
+ * [IconProcessor] implementation to extract the dominant color from the icon.
+ */
+class ColorProcessor : IconProcessor {
+
+    override fun process(context: Context, request: IconRequest, resource: IconRequest.Resource?, icon: Icon): Icon {
+        if (icon.bitmap == null || icon.color != null) return icon
+
+        getEdgeColor(icon.bitmap)?.let { edgeColor ->
+            return icon.copy(color = edgeColor)
+        }
```
jhugman: I'm confused why we're using this, given that we're using `Palette` now?

NotWoods: I ported this existing code from Fennec, but I agree it makes more sense to just always use `Palette`.

pocmo: I remember this was added for activity stream - where we used the color as background color for the tile on which we draw the icon. If we pick a color that is used along the edges then the icon "merges" with the background.

components/browser/icons/src/test/java/mozilla/components/browser/icons/processor/ColorProcessorTest.kt
```diff
@@ -0,0 +1,53 @@
+package mozilla.components.browser.icons.processor
```
pocmo: nit: license header

components/browser/icons/src/main/java/mozilla/components/browser/icons/processor/ColorProcessor.kt
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * [IconProcessor] implementation to extract the dominant color from the icon.
+ */
+class ColorProcessor : IconProcessor {
+
+    override fun process(context: Context, request: IconRequest, resource: IconRequest.Resource?, icon: Icon): Icon {
+        if (icon.bitmap == null || icon.color != null) return icon
+
+        val dominantColor = Palette.from(icon.bitmap).generate().getDominantColor(DEFAULT_COLOR)
```
pocmo: A default color worked fine when this was app code. Now as a component I think we should just return no color so that the app can pick a good default for its use case.

components/browser/icons/src/main/java/mozilla/components/browser/icons/processor/ColorProcessor.kt
```diff
@@ -0,0 +1,33 @@
...
+        return icon.copy(color = dominantColor and DOMINANT_COLOR_MASK)
+    }
+
+    companion object {
+        /**
+         * No color, here we use photon color
+         */
+        internal val DEFAULT_COLOR = rgb(177, 177, 179)
+
+        private const val DOMINANT_COLOR_MASK = 0x7FFFFFFF
```
pocmo: Here I also wonder if we should leave that up to the app since the app may be opinionated on how it wants to use the color.

components/browser/icons/src/main/java/mozilla/components/browser/icons/processor/ColorProcessor.kt
NotWoods: Sounds good, I'll update it so that the icon isn't modified if no colour is found.

PR #3343, opened 2019-06-10
Remove GCP ping redirection for tagged pings.

travis79:
This removes the redirection to GCP endpoint for all tagged pings, unless the ping tag begins with "gcp-".


travis79:
@Dexterp37 I know that this doesn't have any documentation additions yet, putting this up for discussion.

components/service/glean/src/main/java/mozilla/components/service/glean/net/HttpPingUploader.kt
```diff
@@ -93,10 +93,12 @@ internal class HttpPingUploader : PingUploader {
...
             headers.append("X-Debug-ID", it)
 
-            // NOTE: Tagged pings must be redirected to the GCP endpoint as the AWS endpoint isn't
-            // configured to handle them.  This may pose an issue with testing if a local server
-            // is used to capture pings.
-            endpoint = Configuration.DEFAULT_DEBUGVIEW_ENDPOINT
+            // NOTE: Tagged pings can be redirected to the GCP endpoint by prefixing the tag with
+            // `gcp-.  This may pose an issue with testing if a local server is used to capture
+            // pings.
+            if (it.startsWith("gcp-")) {
```
Dexterp37: Let's not overload the tag with specific meaning. If that's something we really need, let's add a separate option to the debug activity.

travis79: This was really just a workaround due to Arkadiusz asking for the ability to configure the endpoint back to GCP if there were problems.  I would definitely like to remove the redirection altogether and just rely on the configuration endpoint.

travis79:
> This looks good to me now, thanks. Let's not merge this before 24th of June and, before doing that, let's check back with @pocmo and @akkomar . Can you please file a bug about doing the same for glean-core? :)

Follow-up bug for glean-core filed [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1561269)

PR #3346, opened 2019-06-10
Fix for intermittent test failures in Glean

travis79:
Due to upgrading WorkManager dependency, testing of WorkManager now works so all of the tests depending on our workaround in `triggerWorkManager` needed a fix to prevent intermittent failures.  This required us to use the WorkManager TestDriver to trigger the conditions and now we test WorkManager as it is supposed to be tested.

### Pull Request checklist


travis79:
> One small point which you should feel free to ignore because this is an internal API and it's more important to get the fix in: I'm not a huge fan of "Pair" because "first" and "second" isn't really obvious, and I'd rather use a small class with named members.

Totally agree, refactored out the `Pair` and pushed over the top of it.

PR #3350, opened 2019-06-10
Closes #3344: Add getProvidedDefaultSearchEngine to SearchEngineManager to get the provided default search engine.

### LINKED ISSUE #3344: Provide API to get providedDefault from SearchEngineManager
Fenix has a requirement to show the default (not selected) search engine at the top of the list.

I'm using `getDefaultSearchEngine` with the paramater `"."` to force the behavior that I need in [SearchEngineManager.kt](https://github.com/mozilla-mobile/android-components/blob/master/components/browser/search/src/main/java/mozilla/components/browser/search/SearchEngineManager.kt)`. But we should get a more explicit API for this behavior

gabrielluong:
Fixes #3344

### Pull Request checklist

### DIFF
--- a/components/browser/search/src/main/java/mozilla/components/browser/search/SearchEngineManager.kt
+++ b/components/browser/search/src/main/java/mozilla/components/browser/search/SearchEngineManager.kt
@@ -91,12 +91,22 @@ class SearchEngineManager(
     @Synchronized
     fun getDefaultSearchEngine(context: Context, name: String = EMPTY): SearchEngine {
         val searchEngineList = getSearchEngineList(context)
-        val providedDefault = searchEngineList.default ?: searchEngineList.list[0]
+        val providedDefault = getProvidedDefaultSearchEngine(context)
 
         return when (name) {
             EMPTY -> defaultSearchEngine ?: providedDefault
             else -> searchEngineList.list.find { it.name == name } ?: providedDefault
         }
+    }
+
+    /**
+     * Returns the provided default search engine or the first search engine if the default
+     * is not set.
+     */
+    @Synchronized
+    fun getProvidedDefaultSearchEngine(context: Context): SearchEngine {
+        val searchEngineList = getSearchEngineList(context)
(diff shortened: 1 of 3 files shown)


components/browser/search/src/main/java/mozilla/components/browser/search/SearchEngineManager.kt
```diff
@@ -99,6 +99,15 @@ class SearchEngineManager(
         }
     }
 
+    /**
+     * Returns the provided default search engine if the user has no preference.
```
pocmo: documentation nit: "if the user has no preference" is not fully true: This will always provide the default even if the user has selected a different one.

components/browser/search/src/main/java/mozilla/components/browser/search/SearchEngineManager.kt
```diff
@@ -99,6 +99,15 @@ class SearchEngineManager(
...
+     */
+    @Synchronized
+    fun getProvidedDefaultSearchEngine(context: Context): SearchEngine {
+        val searchEngineList = getSearchEngineList(context)
+        return searchEngineList.default ?: searchEngineList.list[0]
+    }
```
pocmo: nit: We could use that method from `getDefaultSearchEngine()` now to avoid duplicating this logic.

gabrielluong: If we use getDefaultSearchEngine() in this function, we would have to provide a name that is not empty and doesn't match any of the search engine such as ".". Should we continue with such a change? 

Alternatively, we could call `getProvidedDefaultSearchEngine` in the providedDefault in `getDefaultSearchEngine` on line 94 to avoid some duplication in logic. What do you think?

pocmo: I meant the reverse: `getDefaultSearchEngine()` should call your new `getProvidedDefaultSearchEngine()` method for the default instead of duplicating this part.

PR #3379, opened 2019-06-11
Fix intermittent test in Glean.

travis79:
The `schedulePingCollection must correctly append a work request to the WorkManager` test creates it’s own metrics ping scheduler, but then `doWork()` for the worker references the one in the singleton  `Glean.metricsPingScheduler`.  Corrected by replacing the singleton's copy with the test one.


components/service/glean/src/test/java/mozilla/components/service/glean/scheduler/MetricsPingSchedulerTest.kt
```diff
@@ -426,13 +426,15 @@ class MetricsPingSchedulerTest {
 
     @Test
     fun `schedulePingCollection must correctly append a work request to the WorkManager`() {
-        val mps = MetricsPingScheduler(ApplicationProvider.getApplicationContext<Context>())
+        // Replacing the singleton's metricsPingScheduler here since doWork() refers to it when
+        // the worker runs, otherwise we can get a lateinit property is not initialized error.
+        Glean.metricsPingScheduler = MetricsPingScheduler(ApplicationProvider.getApplicationContext<Context>())
```
Dexterp37: Good catch. Is this the only test that requires it?

travis79: I believe so, every other call to `schedulePingCollection` is really just using `verify` to make sure it gets called or doesn't get called.  This is the only one where we get into WorkManager to ensure it's there, which is what was enqueuing the worker and causing the issue.

PR #3380, opened 2019-06-11
Closes #2382 - Save PWA manifest and shortcuts

### LINKED ISSUE #2382: Save and load manifests of installed web apps

NotWoods:
🎉 

Adds support for basic PWAs by creating shortcuts for web apps, then reloading their manifests from `Room`. 

### Pull Request checklist


components/feature/pwa/src/main/java/mozilla/components/feature/pwa/ManifestStorage.kt
```diff
@@ -0,0 +1,42 @@
...
+
+package mozilla.components.feature.pwa
+
+import android.content.Context
+import android.content.Context.MODE_PRIVATE
+import androidx.core.content.edit
+import mozilla.components.concept.engine.manifest.WebAppManifest
+import mozilla.components.concept.engine.manifest.WebAppManifestParser
+
+class ManifestStorage(context: Context) {
```
pocmo: could this be `internal` for now?

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/ManifestStorage.kt
```diff
@@ -0,0 +1,42 @@
...
+                is WebAppManifestParser.Result.Success -> result.manifest
+                is WebAppManifestParser.Result.Failure -> null
+            }
+        }
+    }
+
+    fun saveManifest(manifest: WebAppManifest) {
+        preferences.edit {
+            val json = parser.serialize(manifest).toString()
+            putString(manifest.startUrl, json)
```
pocmo: I wonder if we should *not* use the startUrl here and instead generate a UUID for this installation. This ID we will also pass to the launcher intent.
The startUrl may change over time and we probably do not want to hard-code it in the intent. And it may not really be a unique identifier. 🤔

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/WebAppShortcutManager.kt
```diff
@@ -0,0 +1,86 @@
...
+import android.content.pm.ShortcutManager
+import android.os.Build.VERSION.SDK_INT
+import android.os.Build.VERSION_CODES
+import androidx.core.content.getSystemService
+import androidx.core.content.pm.ShortcutInfoCompat
+import androidx.core.content.pm.ShortcutManagerCompat
+import androidx.core.net.toUri
+import mozilla.components.concept.engine.manifest.WebAppManifest
+
+class WebAppShortcutManager(
```
pocmo: Can this be `internal` for now?

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/WebAppShortcutManager.kt
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+    /**
+     * Uninstalls a set of PWAs from the user's device by disabling their
+     * shortcuts and removing the associated manifest data.
+     *
+     * @param startUrls List of manifest startUrls to remove.
+     * @param disabledMessage Message to display when a disable shortcut is tapped.
+     */
+    fun uninstallShortcuts(context: Context, startUrls: List<String>, disabledMessage: String? = null) {
```
pocmo: Is that needed? 🤔

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/ManifestStorage.kt
```diff
@@ -0,0 +1,42 @@
...
+package mozilla.components.feature.pwa
+
+import android.content.Context
+import android.content.Context.MODE_PRIVATE
+import androidx.core.content.edit
+import mozilla.components.concept.engine.manifest.WebAppManifest
+import mozilla.components.concept.engine.manifest.WebAppManifestParser
+
+class ManifestStorage(context: Context) {
+    private val preferences = context.getSharedPreferences(PREFERENCE_NAME, MODE_PRIVATE)
```
pocmo: Instead of using `SharedPreferences` we may want to write manifests to a flat file.
* A SharedPreferences **write** always writes *everything* to disk as XML
* And a SharedPreferences **read** always reads *everything* into memory and will keep it there - even if not referenced.

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/WebAppShortcutManager.kt
NotWoods: This is here in case we want to expose some UI to delete the manifest.

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/ManifestStorage.kt
NotWoods: If we don't hard-code the startUrl in the intent, then there is no way to launch a normal browser session if we lose the manifest data. Do other browsers let a PWA migrate to a different URL?

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/WebAppLauncherActivity.kt
```diff
@@ -22,34 +26,48 @@ import mozilla.components.support.ktx.kotlin.toUri
  */
 class WebAppLauncherActivity : AppCompatActivity() {
     private val logger = Logger("WebAppLauncherActivity")
+    private lateinit var storage: ManifestStorage
+    private val activityJob = Job()
```
dector: _Not a big issue for not, but may be important for future modifications._

I suggest using `SupervisorJob` here (to prevent cancelling all children if one of them failed).

Or event better (I think), something like:
```kotlin
class WebAppLauncherActivity : AppCompatActivity(), CoroutineScope by MainScope() {

    override fun onDestroy() {
        coroutineContext.cancel()
        super.onDestroy()
    }
}
```

components/feature/pwa/src/test/java/mozilla/components/feature/pwa/ManifestStorageTest.kt
```diff
@@ -0,0 +1,94 @@
...
+        val manifest = WebAppManifest(name = "Firefox", startUrl = "https://firefox.com")
+        val json = WebAppManifestParser().serialize(manifest)
+
+        val stream = mock<FileOutputStream>()
+        val file = storage.mockFile("https://firefox.com")
+        `when`(file.startWrite()).thenReturn(stream)
+
+        storage.saveManifest(manifest)
+        verify(stream).write(json.toString().toByteArray())
+        Unit
```
dector: Seems like good place to use `runBlockingTest {}`. :)

components/feature/pwa/src/test/java/mozilla/components/feature/pwa/ManifestStorageTest.kt
```diff
@@ -0,0 +1,94 @@
...
+@RunWith(AndroidJUnit4::class)
+class ManifestStorageTest {
+    @Test
+    fun `load returns null if entry does not exist or is invalid`() = runBlocking {
+        val storage = spy(ManifestStorage(testContext))
+        assertNull(storage.loadManifest("https://example.com"))
+
+        val mozFile = storage.mockFile("https://mozilla.org")
+        val fireFile = storage.mockFile("https://firefox.com")
+        `when`(mozFile.readFully()).thenReturn("[]".toByteArray())
```
dector: nit: What do you think about using `whenever()` here? Backtick-less notation is bit more clear as for me.

components/feature/pwa/src/test/java/mozilla/components/feature/pwa/WebAppShortcutManagerTest.kt
```diff
@@ -0,0 +1,208 @@
...
+import org.mockito.Mockito.doReturn
+import org.mockito.Mockito.never
+import org.mockito.Mockito.spy
+import org.mockito.Mockito.verify
+import org.robolectric.util.ReflectionHelpers
+
+@ExperimentalCoroutinesApi
+@RunWith(AndroidJUnit4::class)
+class WebAppShortcutManagerTest {
+    private lateinit var context: Context
```
dector: nit: We can start using `@Mock` annotations in tests. E.g.

```kotlin
@Mock
lateinit var context: Context

@Before
fun setUp() {
    initMocks(this)
}
```

components/feature/pwa/src/test/java/mozilla/components/feature/pwa/WebAppShortcutManagerTest.kt
```diff
@@ -0,0 +1,208 @@
...
+        setSdkInt(Build.VERSION_CODES.N_MR1)
+        val domains = listOf("https://mozilla.org", "https://firefox.com")
+        val message = "Can't touch this - its uninstalled."
+        WebAppShortcutManager(storage).uninstallShortcuts(context, domains, message)
+
+        verify(shortcutManager).disableShortcuts(domains, message)
+    }
+
+    private fun setSdkInt(sdk: Int) {
+        ReflectionHelpers.setStaticField(Build.VERSION::class.java, "SDK_INT", sdk)
```
dector: nit: More sounding parameter name (e.g. `sdkVersion`) can be used.

dector: nit: Kotlin field refence make it even more awesome. 

```kotlin
private fun setSdkInt(sdkVersion: Int) {
    setStaticField(Build.VERSION::SDK_INT.javaField, sdkVersion)
}
```

components/feature/pwa/src/test/java/mozilla/components/feature/pwa/ManifestStorageTest.kt
NotWoods: Sadly `runBlockingTest` doesn't work with switching to `IO` for now.

components/feature/pwa/src/test/java/mozilla/components/feature/pwa/WebAppShortcutManagerTest.kt
NotWoods: Didn't know about this, thanks! 😃

components/feature/pwa/src/test/java/mozilla/components/feature/pwa/ManifestStorageTest.kt
dector: Do you mean `UncompletedCoroutinesError` in the end? I suppose we can pass `coroutineContext`/`dispatcher` into contructor to control execution context in tests.

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/WebAppLauncherActivity.kt
NotWoods: TIL, thanks!

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/WebAppShortcutManager.kt
```diff
@@ -0,0 +1,86 @@
+package mozilla.components.feature.pwa
```
pocmo: nit: license header

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/db/ManifestConverter.kt
```diff
@@ -0,0 +1,26 @@
...
+package mozilla.components.feature.pwa.db
+
+import androidx.room.TypeConverter
+import mozilla.components.concept.engine.manifest.WebAppManifest
+import mozilla.components.concept.engine.manifest.WebAppManifestParser
+
+/**
+ * Converts a web app manifest to and from JSON strings
+ */
+class ManifestConverter {
```
pocmo: could this be internal?

PR #3416, opened 2019-06-12
Closes #3341: Allow restoring Tab/TabCollection without reusing session ids.

### LINKED ISSUE #3341: Allow "restoreWithNewId" on a session
In Fenix, we'd like the ability to restore a session with a newly created sessionId so that, when restoring a session from a collection, it won't clash with existing id's. 

See https://github.com/mozilla-mobile/fenix/issues/3215 for more information, and https://github.com/mozilla-mobile/fenix/issues/3280 for issues caused by duplicate id's.

pocmo:
### Pull Request checklist

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/ext/AtomicFile.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/ext/AtomicFile.kt
@@ -47,7 +47,8 @@ fun AtomicFile.writeSnapshot(
  */
 fun AtomicFile.readSnapshotItem(
     engine: Engine,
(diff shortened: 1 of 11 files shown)


pocmo:
> Looks good! I was wondering if we should handle the case in SessionManager also e.g. throw when a duplicate ID is added because then the system is in an invalid state. Should we file a follow-up to address that later or wait until browser-state?

Yeah, I was wondering that too! Since we override `equals()` of `Session` to use the ID, we probably just want to check if the *same* `Session` is already in `SessionManager`. Makes a lot of sense. Let's land that after MVP though - there may be surprises. :)

pocmo:
Filed #3437

PR #3431, opened 2019-06-13
Improve `QrFeatureTest` readability.

dector:
Just cosmetic test structure improvements. :)

### Complexity

Easy (★☆☆)

### Pull Request checklist

### DIFF
--- a/components/feature/qr/src/test/java/mozilla/components/feature/qr/QrFeatureTest.kt
+++ b/components/feature/qr/src/test/java/mozilla/components/feature/qr/QrFeatureTest.kt
@@ -4,7 +4,7 @@
 
 package mozilla.components.feature.qr
 
-import android.Manifest
+import android.Manifest.permission.CAMERA
 import android.content.pm.PackageManager
 import androidx.fragment.app.FragmentManager
 import androidx.fragment.app.FragmentTransaction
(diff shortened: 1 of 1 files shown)


components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFeature.kt
```diff
@@ -37,7 +37,7 @@ class QrFeature(
 ) : LifecycleAwareFeature, BackHandler {
     private var containerViewId: Int = 0
 
-    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    @VisibleForTesting
```
pocmo: Why did you remove the parameter here? Because it's the default?

components/feature/qr/src/test/java/mozilla/ext/TestCallback.kt
```diff
@@ -0,0 +1,14 @@
+package mozilla.ext
```
pocmo: nit: license

components/feature/qr/src/test/java/mozilla/ext/TestCallbackTest.kt
```diff
@@ -0,0 +1,48 @@
+package mozilla.ext
```
pocmo: nit: license

components/feature/qr/src/test/java/mozilla/ext/TestCallback.kt
dector: Yep!

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFeature.kt
dector: Yep!

components/feature/qr/src/test/java/mozilla/ext/TestCallback.kt
dector: It'll be nice to add lint rule for this in the future. :)

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFeature.kt
pocmo: I often still use "private" in order to:
* Make it more readable without having to know or lookup the default
* Make it clear what my intent is - as opposed to "I forgot or didn't know there's a parameter"

What do you think?

dector: Tbh, I prefer simple `@VisibleForTesting` in this case. Private-by-default is the most strict behavior, so we'll not expose part of the private API accidentally. And sources looks more readable - properties are parsed visually with less effort.

Anyway, as a code owner you always can insist and I'll change it back. :sweat_smile: 

BTW. What do you think about project-level code style document? I can prepare some preview version that can be extended when new questions will arise.

pocmo:
> As @pocmo has started this review I will only comment rather than approve.

I also did a quick drive-by and flagged @csadilek for the actual review. Feel free to take that over. :)

dector:
@csadilek

As for me, it's good idea to avoid `var`s in tests (well, not only in tests :) ) as non-encapsulated mutable variables requires more cognitive resources to understand. So the motivation behind the `TestCallback` was to encapsulate mutation side-effect and provide simple checks (e.g. `.isInvoked` etc).

Sure, it's good idea to move it into `:support-test` module.

But I'm not sure that I've got your idea why we'll need to create it's subclass for `QrFeatureTest`. Can you explain it a bit, please?

However I totally agree that verification with `TestCallback` is not good enough and requires further improvements. There was another way I was considering: using Mockito to create/verify that callback was called (instead of custom `TestCallback`). Maybe I should get back to it - now it look for me bit more clear solution. What do you think?

Was:
```kotlin
assertEquals(1, permissionsCallback.invocationsCount)
permissionsCallback.registry.first().let { request ->
    assertEquals(1, request.size)
    assertEquals(CAMERA, request.first())
}
```

Will be:
```kotlin
verify(permissionsCallback)(arrayOf(CAMERA))
```

I think that I like Mockito-version more. If you like it as well - I'll remove `TestCallback` as use `mock()` instead. The only issue I see here is this double invocation: `verify()()` (Unfortunately Mockito don't recognize `.invoke()` as correct verification). But it's minor.

And finaly, thanks for raising this concern! :+1:

dector:
BTW. I've used [mockito-kotlin bindings](https://github.com/nhaarman/mockito-kotlin) before. It has quite nice dsl. We can use it as well. What do you think?

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFeature.kt
dector: @pocmo What kind of practice I should consider as a project guideline for now?

csadilek:
@dector 

> As for me, it's good idea to avoid `var`s in tests (well, not only in tests :) ) as non-encapsulated mutable variables requires more cognitive resources to understand.

Yes, I definitely agree with that!

> Sure, it's good idea to move it into :support-test module. But I'm not sure that I've got your idea why we'll need to create it's subclass for QrFeatureTest. Can you explain it a bit, please?

I just meant that `val permissionCallback = PermissionCallback()` is easier to read than `val permissionCallback = TestCallback<Array<String>>()`.

> There was another way I was considering: using Mockito to create/verify that callback was called (instead of custom TestCallback). Maybe I should get back to it - now it look for me bit more clear solution. What do you think?

Yes, that looks perfect to me. Thank you!

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFeature.kt
csadilek: We likely aren't consistent here, but I would let it in as well.

dector: :ok:

csadilek:
Thanks @dector.

PR #3436, opened 2019-06-14

Pike:
State: mozilla-l10n/android-l10n@60fcf55878eadd7f19bbe27bf78f9817810f3d10

Last import for Fenix, CC @pocmo @Delphine.

@pocmo, do you want to cherry pick that onto the relbranch, or would you rather have a dedicated PR for that?


pocmo:
> @pocmo, do you want to cherry pick that onto the relbranch, or would you rather have a dedicated PR for that?

I'll cherry-pick that!

Issue #3274, opened 2019-06-05
Add `setFocus` to BrowserToolbar


sblatz:
Per Fenix's issue: https://github.com/mozilla-mobile/fenix/issues/3175 we want a separate function (outside of [BrowserToolbar.editMode()](https://github.com/mozilla-mobile/android-components/blob/master/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt#L514-L523)) that has the ability to just set the focus of the editToolbar.

pocmo:
@dector Now I wonder if calling `editMode()` should actually only focus if the toolbar is already in "edit mode" :)

dector:
@pocmo May be like this but I'm bit worried that `editMode()` will violate SRP. :) It's inner logic is quite complex to understand already.

pocmo:
@dector Yes, true, although right now it does a state transition even if it is already in that state.

dector:
@pocmo I totally agree that `editMode()` should do smart state-transition instead.

sblatz:
I tried working on this for a bit. Unfortunately it's not a quick fix since calling `focus()` highlights all of the text in the urlView.

pocmo:
@sblatz In `EditToolbar` we set `setSelectAllOnFocus(true)` on `urlView`. :)

pocmo:
We could try removing that call and calling `selectAll()` ourselves when switching to `editMode`. However I do not know if that has any other side effects.

sblatz:
@pocmo Yes, I've tried setting it to false, but doing this:

<img width="634" alt="image" src="https://user-images.githubusercontent.com/4400286/59522551-c9dbdc80-8e83-11e9-8d6b-dac9e28556c6.png">

Doesn't work since `focus()` does not happen immediately

sblatz:
> We could try removing that call and calling selectAll() ourselves when switching to editMode. However I do not know if that has any other side effects.

This is a great suggestion. I will test this :)

PR #3447, opened 2019-06-14
For #3274: Adds editModeFocus to BrowserToolbar

sblatz:
I ended up implementing `editModeFocus`. This function is necessary because in Fenix, we use calling `editMode` (while already set to `state editing` to clear the search field or properly set it to the search terms). Because of that, having a separate function helps us explicitly focus the the editToolbar.urlView.

@pocmo let me know if you have any concerns with this.

### Pull Request checklist

### DIFF
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
@@ -504,6 +504,13 @@ class BrowserToolbar @JvmOverloads constructor(
     }
 
     /**
+     * Focuses the editToolbar if already in edit mode
+     */
+    fun focus() {
+        editToolbar.focus()
+    }
+
+    /**
      * Switches to URL editing mode.
      */
     override fun editMode() {
@@ -512,10 +519,9 @@ class BrowserToolbar @JvmOverloads constructor(
         val shouldAutoComplete = searchTerms.isEmpty()
 
         editToolbar.updateUrl(urlValue.toString(), shouldAutoComplete)
-
         updateState(State.EDIT)
-
         editToolbar.focus()
+        editToolbar.urlView.selectAll()
     }
 
     /**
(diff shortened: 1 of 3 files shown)


pocmo:
> while already set to state editing to clear the search field or properly set it to the search terms

This sounds like we should figure out what you want to do here and create a proper API for that - because it sounds a bit like we need this focus() method because you are using editMode() in an undocumented way already :)

sblatz:
@pocmo I'm unclear how we should proceed with this. Could we sync up on Monday?

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
```diff
@@ -503,6 +503,13 @@ class BrowserToolbar @JvmOverloads constructor(
         editToolbar.addEditAction(action)
     }
 
+    /**
+     * Focuses the editToolbar if already in edit mode
```
pocmo: nit: `editToolbar` is an implementation detail and not visible from the "outside". We should describe what the method does without mentioning it. :)

components/browser/toolbar/src/test/java/mozilla/components/browser/toolbar/BrowserToolbarTest.kt
```diff
@@ -70,6 +70,19 @@ class BrowserToolbarTest {
         assertTrue(toolbar.editToolbar.visibility == View.VISIBLE)
     }
 
+    @Test
+    fun `calling editModeFocus() focuses the editToolbar`() {
```
pocmo: nit: `editModeFocus`-> `focus`

docs/changelog.md
```diff
@@ -12,6 +12,10 @@ permalink: /changelog/
 * [Gecko](https://github.com/mozilla-mobile/android-components/blob/master/buildSrc/src/main/java/Gecko.kt)
 * [Configuration](https://github.com/mozilla-mobile/android-components/blob/master/buildSrc/src/main/java/Config.kt)
 
+
+* **browser-toolbar**
+  * Adds `focus()` which provides a hook for calling `editMode.focus()` to focus the edit mode `urlView`
```
pocmo: nit: Like the kdoc I think this should be more descriptive, explaining what it does (more from a user's point of view) than talking about internals like `urlView`.

PR #3452, opened 2019-06-14
Move ui-icons fill to resource

NotWoods:
Ctrl-R "#FFFFFF" -> "@color/mozac_ui_icons_fill"


jonalmeida:
I'd also check to see if consuming apps like Fenix work fine with this chance. There are some places where these icons are mutated, passed colour resources, etc.

I don't think our tests would catch for such things so it might require a manual check (especially for night mode?).

dector:
@jonalmeida New color resource is just created so I don't expect some issue here.

PR #3453, opened 2019-06-14
Fix null checks before Android Q

NotWoods:
SDK 29 adds many more `@Nullable` and `@NotNull` checks, which causes parts of our code to break since they assume nullability one way or another. I pulled some of the fixes I needed to get compling to work into their own branch here.

### DIFF
--- a/components/browser/domains/src/main/java/mozilla/components/browser/domains/Domains.kt
+++ b/components/browser/domains/src/main/java/mozilla/components/browser/domains/Domains.kt
@@ -5,6 +5,8 @@
 package mozilla.components.browser.domains
 
 import android.content.Context
+import android.os.Build.VERSION.SDK_INT
+import android.os.Build.VERSION_CODES
 import android.os.LocaleList
 import android.text.TextUtils
 import java.io.IOException
@@ -46,9 +48,9 @@ object Domains {
         val availableDomains = LinkedHashSet<String>()
         val assetManager = context.assets
         val domains = try {
-            assetManager.list("domains")
+            assetManager.list("domains") ?: emptyArray<String>()
         } catch (e: IOException) {
-            arrayOf<String>()
+            emptyArray<String>()
         }
         availableDomains.addAll(domains)
         return availableDomains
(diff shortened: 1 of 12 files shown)


components/browser/domains/src/main/java/mozilla/components/browser/domains/Domains.kt
```diff
@@ -46,9 +48,9 @@ object Domains {
         val availableDomains = LinkedHashSet<String>()
         val assetManager = context.assets
         val domains = try {
-            assetManager.list("domains")
+            assetManager.list("domains") ?: emptyArray<String>()
```
dector: nit: Alternatively can be something like

```kotlin
val domains: Array<String> = try {
    assetManager.list("domains")
} catch (e: IOException) {
    null
} ?: emptyArray()
```

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/MultiButtonDialogFragment.kt
```diff
@@ -29,11 +29,11 @@ internal class MultiButtonDialogFragment : PromptDialogFragment() {
      */
     internal val hasShownManyDialogs: Boolean by lazy { safeArguments.getBoolean(KEY_MANY_ALERTS) }
 
-    internal val positiveButtonTitle: String by lazy { safeArguments.getString(KEY_POSITIVE_BUTTON_TITLE) }
+    internal val positiveButtonTitle: String? by lazy { safeArguments.getString(KEY_POSITIVE_BUTTON_TITLE) }
```
dector: Just wondering if it's possible to keep this values as non-nullable with:

```kotlin
by lazy { safeArguments.getString(KEY_POSITIVE_BUTTON_TITLE) ?: "" }
```

components/support/utils/src/main/java/mozilla/components/support/utils/SafeIntent.kt
```diff
@@ -77,7 +77,7 @@ class SafeIntent(val unsafe: Intent) {
 
     fun <T : Parcelable> getParcelableArrayListExtra(name: String): ArrayList<T>? {
         return safeAccess {
-            val value: ArrayList<T> = unsafe.getParcelableArrayListExtra(name)
+            val value: ArrayList<T>? = unsafe.getParcelableArrayListExtra(name)
```
dector: nit: can it be simplified to

```kotlin
fun <T : Parcelable> getParcelableArrayListExtra(name: String): ArrayList<T>? =
    safeAccess { unsafe.getParcelableArrayListExtra(name) }
```

dector:
Cool! Do we have an (epic) issue for raising target/compile sdk up to 29?

NotWoods:
new meta issue: #3481

pocmo:
There seem to be some test failures and a merge conflict.

PR #3485, opened 2019-06-19
Closes #3424 - Check if session is installable

### LINKED ISSUE #3424: Define PWA installability
Currently we are primarily focused on the existence of a manifest. Browsers often use other additional criteria to define if a pwa is "installable". For example Chrome's criteria are listed here:
https://developers.google.com/web/fundamentals/app-install-banners/#criteria

* What additional criteria should we use to define an app as installable?
* Does Gecko(View) do any checks before handing the manifest to us; e.g. does a manifest received through Gecko(View) mean it's already "installable"?

NotWoods:
Adds `Session.isInstallable()` to check if session represents installable web app.

In addition to installablity criteria outlined in the design doc, we filter out badge icons since they aren't meant to be used as launcher icons.

### Pull Request checklist

### DIFF
--- a/components/concept/engine/src/main/java/mozilla/components/concept/engine/manifest/WebAppManifestParser.kt
+++ b/components/concept/engine/src/main/java/mozilla/components/concept/engine/manifest/WebAppManifestParser.kt
@@ -5,6 +5,7 @@
 package mozilla.components.concept.engine.manifest
 
 import mozilla.components.support.ktx.android.org.json.asSequence
+import mozilla.components.support.ktx.android.org.json.tryGetString
 import org.json.JSONException
 import org.json.JSONObject
 
@@ -37,9 +38,13 @@ class WebAppManifestParser {
      */
     fun parse(json: JSONObject): Result {
         return try {
+            val shortName = json.tryGetString("short_name")
+            val name = json.tryGetString("name") ?: shortName
+                ?: return Result.Failure(JSONException("Missing manifest name"))
+
             Result.Success(WebAppManifest(
-                name = json.getString("name"),
-                shortName = json.optString("short_name", null),
+                name = name,
+                shortName = shortName,
                 startUrl = json.getString("start_url"),
                 display = getDisplayMode(json),
                 backgroundColor = parseColor(json.optString("background_color", null)),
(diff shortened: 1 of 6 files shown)


components/feature/pwa/src/main/java/mozilla/components/feature/pwa/ext/Session.kt
```diff
@@ -0,0 +1,31 @@
...
+
+private const val MIN_INSTALLABLE_ICON_SIZE = 192
+
+/**
+ * Checks if the current session represents an installable web app.
+ *
+ * Websites are installable if:
+ * - The site is served over HTTPS
+ * - The site has a valid manifest with a name or short_name
+ * - The icons array in the manifest contains an icon of at least 192x192
```
Amejia481: I think having a [start_url](https://developers.google.com/web/fundamentals/web-app-manifest/#start-url) property should also be also a requirement to be installable. What do you think?

NotWoods: `start_url` and `name` must be set if `webAppManifest` is not null. They aren't nullable so the manifest won't be parsed without them.

Amejia481: Got you! Can they be empty String?  Or are we preventing that with an exception?

NotWoods: We aren't currently checking that, but we should. The spec actually defines that name can be omitted, so I'm going to update this PR with some changes to the parser.

Amejia481: 👍

components/concept/engine/src/main/java/mozilla/components/concept/engine/manifest/WebAppManifestParser.kt
```diff
@@ -37,9 +38,13 @@ class WebAppManifestParser {
      */
     fun parse(json: JSONObject): Result {
         return try {
+            val shortName = json.tryGetString("short_name")
+            val name = json.tryGetString("name") ?: shortName
+                ?: return Result.Failure(JSONException("Missing manifest name"))
```
Amejia481: nit: We could add a test to verify this :)

PR #3497, opened 2019-06-20

Pike:
State: mozilla-l10n/android-l10n@4c73402212e4766cf050b520318cb56961c78e2b

CC @Delphine


Delphine:
This patch applies also to the release branch, and we don't know the severity of these fixes. We know it's late but if you can take these into release it may be worth it.

PR #3502, opened 2019-06-21
browser-state: Move state handling to generic lib-state component.

pocmo:
### Pull Request checklist


components/lib/state/src/main/java/mozilla/components/lib/state/Store.kt
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * @param initialState The initial state until a dispatched [Action] creates a new state.
+ * @param reducer A function that gets the current [State] and [Action] passed in and will return a new [State].
+ */
+open class Store<S : State, A: Action>(
+    initialState: S,
+    private val reducer: Reducer<S, A>
+) {
+    private val storeContext = Executors.newSingleThreadExecutor().asCoroutineDispatcher()
+    private val storeScope = CoroutineScope(storeContext)
```
NotWoods: Should `Store` use `CoroutineScope` as a delegate instead?

components/lib/state/src/main/java/mozilla/components/lib/state/Store.kt
```diff
@@ -0,0 +1,108 @@
...
+    private val subscriptions: MutableList<Subscription<S, A>> = mutableListOf()
```
NotWoods: Use a `MutableSet` instead since the subscriptions are removed by value, not index

components/lib/state/src/main/java/mozilla/components/lib/state/ext/StoreExtensions.kt
```diff
@@ -60,10 +62,10 @@ private class LifecycleBoundObserver(
 /**
  * View.OnAttachStateChangeListener implementation to bind an observer to a View.
  */
-private class ViewBoundObserver(
+private class ViewBoundObserver<S : State, A: Action>(
```
NotWoods: Since `LifecycleBoundObserver` and `ViewBoundObserver` are both bindings, perhaps they should be called `Binding` instead?

components/lib/state/src/main/java/mozilla/components/lib/state/ext/StoreExtensions.kt
```diff
@@ -45,10 +47,10 @@ fun BrowserStore.observe(
...
-private class LifecycleBoundObserver(
+private class LifecycleBoundObserver<S : State, A: Action>(
     private val owner: LifecycleOwner,
-    private val subscription: BrowserStore.Subscription
-) : LifecycleObserver, BrowserStore.Subscription.Binding {
+    private val subscription: Store.Subscription<S, A>
+) : LifecycleObserver, Store.Subscription.Binding {
     @OnLifecycleEvent(Lifecycle.Event.ON_DESTROY)
     override fun unbind() {
         subscription.unsubscribe()
```
NotWoods: This looks like infinite recursion

components/lib/state/src/main/java/mozilla/components/lib/state/helpers/Helpers.kt
```diff
@@ -0,0 +1,41 @@
...
+ * [then] only if the value has changed from the last mapped value.
+ *
+ * @param onMainThread Whether or not the [then] function should be executed on the main thread.
+ * @param map A function that maps [State] to the value we want to observe for changes.
+ * @param then Function that gets invoked whenever the mapped value changed.
+ */
+@Suppress("FunctionNaming")
+fun <S : State, T> onlyIfChanged(
+    onMainThread: Boolean = false,
+    map: (S) -> T?,
```
NotWoods: It think it makes more sense to extract mapping to its own helper.

components/lib/state/src/main/java/mozilla/components/lib/state/Store.kt
```diff
@@ -0,0 +1,108 @@
...
+ * [Observer]s.
+ *
+ * @param initialState The initial state until a dispatched [Action] creates a new state.
+ * @param reducer A function that gets the current [State] and [Action] passed in and will return a new [State].
+ */
+open class Store<S : State, A: Action>(
+    initialState: S,
+    private val reducer: Reducer<S, A>
+) {
+    private val storeContext = Executors.newSingleThreadExecutor().asCoroutineDispatcher()
```
grigoryk: not to bike-shed, but this is a dispatcher, not context, right?

components/lib/state/src/main/java/mozilla/components/lib/state/Store.kt
```diff
@@ -0,0 +1,108 @@
...
+     * on the [Subscription] will unregister the observer.
+     */
+    class Subscription<S : State, A : Action> internal constructor(
+        internal val observer: Observer<S>,
+        private val store: Store<S, A>
+    ) {
+        var binding: Binding? = null
+
+        fun unsubscribe() {
+            store.removeSubscription(this)
```
grigoryk: This may be a no-op, if the store doesn't have this `subscription` anymore, right? Currently the only way to remove a subscription from the list is via this method, but perhaps that should either be documented, or somehow enforced?

components/lib/state/src/main/java/mozilla/components/lib/state/Store.kt
```diff
@@ -0,0 +1,108 @@
...
+
+        synchronized(subscriptions) {
+            subscriptions.add(subscription)
+        }
+
+        if (receiveInitialState) {
+            observer.invoke(currentState)
+        }
+
+        return subscription
```
grigoryk: It's possible for the `subscription` to outlive the `store`, right? Lifetimes of these two aren't currently coordinated at all, once you return the `subscription`. And subscription has a reference to the `store`, which is opaque to users of the subscription. I don't think this sort of a class would actually compile in Rust, where this stuff is enforced.

NotWoods: This is similar to how redux handles subscriptions, but using a `WeakRef` for the store might be better.

NotWoods:
Can supplying a binding be required? I think defaulting to using bindings is better and a "ForeverBoundObservable" class could be used for cases where it's needed.

pocmo:
Updated the PR with multiple small changes and the following bigger changes:
* I removed the `receiveInitialState` flag. It looks like we will always want to get the current state passed to the observer. If there will be a reason to not do that then we can add it later on. But for now removing this makes the code simpler.
* I renamed the default `observe` method to `observeManually` and added a `@CheckResult` annotation, which will cause a warning if the returned `Subscription` is not used by calling code. This hopefully makes it more obvious that you are on your own if you use that method and makes it less likely to use it accidentally. The extension methods are still called `observe` (with an `observeForever` added) and do not return a `Subscription` anymore since that is managed for you.
* The `Store` reference in `Subscription` is now a weak reference only.

pocmo:
Landing to unblock the follow-up work.

PR #3520, opened 2019-06-24
Closes #3519 - Gives AwesomeBar Bookmark Suggestions an ID

### LINKED ISSUE #3519: Bookmark suggestions flicker when text is typed
The suggestions need an ID

STR:
1. Bookmark a site (ie twitter.com)
2. Clear history
3. Type "tw" then "twi" 

The bookmark suggestion will flicker between them

ekager:
The guid here should be a safe ID to use?
 
### Pull Request checklist

### DIFF
--- a/components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/BookmarksStorageSuggestionProvider.kt
+++ b/components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/BookmarksStorageSuggestionProvider.kt
@@ -43,6 +43,7 @@ class BookmarksStorageSuggestionProvider(
         return this.map {
             AwesomeBar.Suggestion(
                 provider = this@BookmarksStorageSuggestionProvider,
+                id = it.guid,
                 icon = icons.loadLambda(it.url!!),
                 title = it.title,
                 description = it.url,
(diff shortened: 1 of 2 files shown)


grigoryk:
Using a `guid` as a unique identifier for a bookmark should be fine here! r+.

But to respond to @csadilek 's comment we actually do something subtly different for history, because the concepts don't quite line up as neatly.

When querying bookmarks, you're essentially just querying `moz_bookmarks`.
When querying history, you're first querying `moz_origins`, then `moz_places`, maybe `moz_historyvisits` (see https://github.com/mozilla/application-services/blob/f6ca45a4e45d19aaeec467d81478bffda158cc1c/components/places/src/api/matcher.rs#L25).

Of these, only `moz_places` will have an actual `guid`.
If our history query will return a result out of `moz_origins`, that will be a "synthetic" piece of data from places' point of view (it's generated, not part of sync data; think of it as a "sql view"), and so it won't have a `guid`. It will have a standard incrementing integer "id", which is probably good enough here - but that's not exposed at the API boundary.

`places` exposes a history query interface, which operates over a `SearchResult` - see https://github.com/mozilla/application-services/blob/f6ca45a4e45d19aaeec467d81478bffda158cc1c/components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt#L835-L841 which deliberately omits a `guid`. The closest thing we have is a `url`, and that's what we end up using in HistoryStorageSuggestionProvider.kt#L48 linked above. This conversion happens in `PlacesHistoryStorage@getSuggestions`.

The contract that HistoryStorage exposes via its version of `SearchResult` which has an `id` field and is built on top of the one in places is that the `id` is unique. In turn, we rely on the places lib to not return duplicated URLs when we're querying it for history matches.

csadilek:
> The guid here should be a safe ID to use?

Let's have @grigoryk confirm that though before merging :).

grigoryk:
Note that `guid` or not, we're still depending on `searchBookmarks` to not return duplicated BookmarkNodes. It seems like a bug if it will, so I wouldn't worry about it. Worst case, we'll uncover that bug.

ekager:
Thank you both for the very thorough version of "LGTM" ;)

PR #3523, opened 2019-06-25
Restructure (reduced) BrowserState and add tests.

pocmo:
* This re-structures `BrowserState` to use the new structure we talked about and agreed on in Whistler.
* It reduces the `ContentState` to only contain the properties we care about at this time. Other items (for `Session` parity) will be added in later iterations.
* Adds unit tests to bring the coverage up to 100%.

After this lands I'll file follow-up issues for the missing pieces and steps we want to take next.


pocmo:
This PR contains some noise: I noticed that the classes used the `mozilla.components.browser.session` package and I moved them to `mozilla.components.browser.state`. Together with other changes git thinks some of the files are new.

components/browser/state/src/main/java/mozilla/components/browser/state/state/CustomTabSessionState.kt
```diff
@@ -0,0 +1,15 @@
...
+
+package mozilla.components.browser.state.state
+
+import java.util.UUID
+
+/**
+ * Value type that represents the state of a Custom Tab.
+ */
+data class CustomTabSessionState(
+    override val id: String = UUID.randomUUID().toString(),
```
NotWoods: Could a default ID function be added to encapsulate this? `UUID.randomUUID().toString()` is used in a couple different places.

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * [Action] implementation related to [BrowserState].
+ */
+sealed class BrowserAction : Action
+
+/**
+ * [BrowserAction] implementations related to updating the list of [TabSessionState] inside [BrowserState].
+ */
+sealed class TabsListAction : BrowserAction() {
```
csadilek: nit: `TabListAction` instead of plural `TabsListAction`, same as `CustomTabListAction`?

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -0,0 +1,89 @@
...
+     * Removes an existing [CustomTabSessionState] to [BrowserState.customTabs].
+     */
+    data class RemoveCustomTabAction(val tabId: String) : CustomTabListAction()
+}
+
+/**
+ * [BrowserAction] implementations related to updating the [ContentState] of a single [SessionState] inside
+ * [BrowserState].
+ */
+sealed class ContentAction : BrowserAction() {
```
csadilek: 👍 this looks really nice now!

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/CustomTabListReducer.kt
```diff
@@ -0,0 +1,25 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.state.reducer
+
+import mozilla.components.browser.state.action.CustomTabListAction
+import mozilla.components.browser.state.selector.findCustomTab
+import mozilla.components.browser.state.state.BrowserState
+
+internal object CustomTabListReducer {
```
csadilek: nit: should we add kdoc here similar to the other reducers?

components/browser/state/src/main/java/mozilla/components/browser/state/selector/Selectors.kt
```diff
@@ -0,0 +1,64 @@
...
+    get() = if (selectedTabId == null) {
+        null
+    } else {
+        tabs.firstOrNull { it.id == selectedTabId }
+    }
+
+/**
+ * Finds and returns the tab with the given id. Returns null if no matching tab could be
+ * found.
+ */
```
csadilek: nit: `@param` and `@return`

components/browser/state/src/main/java/mozilla/components/browser/state/selector/Selectors.kt
```diff
@@ -0,0 +1,64 @@
...
+fun BrowserState.findTab(tabId: String): TabSessionState? {
+    return tabs.firstOrNull { it.id == tabId }
+}
+
+/**
+ * Finds and returns the Custom Tab with the given id. Returns null if no matching tab could be
+ * found.
+ */
```
csadilek: nit: `@param` and `@return`

components/browser/state/src/main/java/mozilla/components/browser/state/selector/Selectors.kt
```diff
@@ -0,0 +1,64 @@
...
+fun BrowserState.findCustomTab(tabId: String): CustomTabSessionState? {
+    return customTabs.firstOrNull { it.id == tabId }
+}
+
+/**
+ * Finds and returns the tab with the given id or the selected tab if no id was provided (null). Returns null
+ * if no matching tab could be found or if no selected tab exists.
+ */
```
csadilek: nit: `@param` and `@return`

components/browser/state/src/main/java/mozilla/components/browser/state/state/CustomTabSessionState.kt
csadilek: Hmm, I think I would leave this as is. We're using UUIDs in various modules all over A-C. We won't gain much in terms readability or code reduction from moving this to `support-utils` (or similar). I'd also rather evaluate new situations and check if a UUID is really needed than having everyone just call `uniqueID` because it's convenient.

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/CustomTabListReducer.kt
```diff
@@ -0,0 +1,25 @@
...
+    fun reduce(state: BrowserState, action: CustomTabListAction): BrowserState {
+        return when (action) {
+            is CustomTabListAction.AddCustomTabAction -> state.copy(customTabs = state.customTabs + action.tab)
+            is CustomTabListAction.RemoveCustomTabAction -> {
+                val tab = state.findCustomTab(action.tabId)
```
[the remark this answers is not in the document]
pocmo: :+1:

components/browser/state/src/main/java/mozilla/components/browser/state/selector/Selectors.kt
```diff
@@ -0,0 +1,64 @@
...
+
+// Extension functions for querying and dissecting [BrowserState]
+
+/**
+ * The currently selected tab if there's one.
+ *
+ * Only one [TabSessionState] can be selected at a given time.
+ */
+val BrowserState.selectedTab: TabSessionState?
+    get() = if (selectedTabId == null) {
```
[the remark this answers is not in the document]
pocmo: Makes sense. `findTab()` didn't exist when I wrote `selectedTab`. :)

Issue #2047, opened 2019-02-13
browser-icons: Create Preparer using tippy-top list


pocmo:
We do have a mapping from domain to "great icon" URLs (for top 200 and top 2000).

See: https://github.com/mozilla/tippy-top-sites

PR #3536, opened 2019-06-25
Issue #2047: browser-icons: Ship with tippy-top-sites top 200 list for icon resource lookup..

pocmo:
This is the fallback list that desktop / activity stream uses too. There's a (Kinto-backed) service nowadays; but for now I'm just using the static list.

### Pull Request checklist


components/browser/icons/src/main/java/mozilla/components/browser/icons/preparer/TippyTopIconPreparer.kt
```diff
@@ -0,0 +1,78 @@
...
+    }
+}
+
+private fun parseList(assetManager: AssetManager): Map<String, String> {
+    val map = mutableMapOf<String, String>()
+
+    try {
+        JSONArray(assetManager.open(LIST_FILE_PATH).bufferedReader().readText())
+            .asSequence()
+            .forEach { entry ->
```
NotWoods: Instead of `forEach` and a `MutableMap`, you could use `flatMap` and produce pairs.
```kotlin
.asSequence()
.flatMap {
  val json = entry as JSONObject
  val domains = json.getJSONArray("domains")
  val iconUrl = json.getString("image_url")
  
  domains.asSequence().map { domain -> Pair(domain.toString(), iconUrl) }
}
.toMap()
```

components/browser/icons/src/main/assets/mozac.browser.icons/icons-top200.json
```diff
@@ -0,0 +1,677 @@
+[
+    {
```
Amejia481: I'm not quite sure how often this list is going to be updated, should be we have some automation similar to **the Public Suffix List**?

pocmo: The team said in Whistler that this happens very rarely. Also looking at the commits of the repo (42) there doesn't seem to be much activity:
https://github.com/mozilla/tippy-top-sites/commits/master

There's a hosted version of the list that I would like to switch to in the future. With that we wouldn't need to ship the list ourselves - and also not update it.

components/browser/icons/src/main/java/mozilla/components/browser/icons/preparer/TippyTopIconPreparer.kt
pocmo: :+1: Updated the PR.

components/browser/icons/src/main/assets/mozac.browser.icons/icons-top200.json
Amejia481: 👍

Issue #3529, opened 2019-06-25
browser-state: Dispatch actions from SessionManager


pocmo:
To re-create the same list of sessions during the migration phase, we want to dispatch actions from `SessionManager` to an optional `BrowserStore`.
At the same time we should add some "parity tests" to make sure the state in `LegacySessionManager` and `BrowserState` look the same.

pocmo:
Closing this: I added separate issues for the other things we need to synchronize.

PR #3543, opened 2019-06-25
Issue #3529: Dispatch actions from SessionManager for add/remove/select..

pocmo:
Those are the humble beginnings of #3529 and the `browser-state` migration code. With that we start to synchronize adding, removing and selecting sessions/tabs.


pocmo:
> with one suggestion to add a syncDispatch to Store:

As an internal extension function only in `browser-session`, right? No one else should ever do that. :)

PR #3546, opened 2019-06-25
Increase default experiments bucket count from 100 to 1000

travis79:
See Bugzilla [1556749](https://bugzilla.mozilla.org/show_bug.cgi?id=1556749) for more info

### Pull Request checklist


components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentEvaluatorTest.kt
```diff
@@ -482,25 +482,25 @@ class ExperimentEvaluatorTest {
         }
 
         distribution
-                .groupingBy { it }
-                .eachCount()
-                .forEach {
-                    Assert.assertTrue(it.value in 0..25)
-                }
+            .groupingBy { it / 10 }
```
georgf: Doesn't this mean we miss a first check here?
Maybe something like:
`.groupingBy{ it }.eachCount().forEach{ Assert.assertTrue(it.value in 0..2) }`

travis79: Experimentally, the range that the value falls in is more like`0..9`.  I set up a loop and ran the test 1000 times to arrive at that value.  I was a little unsure as to the method used to derive the low and high bounds on the range but `0..2` seems to fail off the high bound frequently.

georgf: I'm not sure what the exact value to me here, i just saw that it looks like we might lose some potential test coverage.

PR #3547, opened 2019-06-25
Store browser-icons constants in res

NotWoods:
Moves constants like colors and dp dimens into corresponding `res/` XML files.

### DIFF
--- a/components/browser/icons/src/main/java/mozilla/components/browser/icons/generator/DefaultIconGenerator.kt
+++ b/components/browser/icons/src/main/java/mozilla/components/browser/icons/generator/DefaultIconGenerator.kt
@@ -5,47 +5,50 @@
 package mozilla.components.browser.icons.generator
 
 import android.content.Context
+import android.content.res.Resources
 import android.graphics.Bitmap
+import android.graphics.Bitmap.Config.ARGB_8888
 import android.graphics.Canvas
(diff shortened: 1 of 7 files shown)


components/browser/icons/src/main/java/mozilla/components/browser/icons/generator/DefaultIconGenerator.kt
```diff
@@ -5,85 +5,97 @@
...
+    internal fun pickColor(resources: Resources, url: String): Int {
+        return resources.obtainTypedArray(backgroundColorsRes).use { backgroundColors ->
+            if (url.isEmpty()) {
+                backgroundColors.getColor(0, 0)
+            } else {
+                val snippet = getRepresentativeSnippet(url)
+                val index = Math.abs(snippet.hashCode() % backgroundColors.length())
+
+                backgroundColors.getColor(index, 0)
+            }
```
pocmo: lint is complaining here:

> Error: This TypedArray should be recycled after use with #recycle()

may be a false positive since I assume that is `use` from androidx doing?

NotWoods: It's a false positive. Somehow this is fine in other places but lint thinks its a problem here.

Issue #3481, opened 2019-06-18
Meta: compileSdkVersion 29 (Android Q)


NotWoods:
We need to bump `compileSdkVersion` to use some new Android Q APIs.

This is necessary for a [Gecko-based Download Manager](https://github.com/mozilla-mobile/android-components/issues/1968) that supports Scoped Storage.

PR #3551, opened 2019-06-25
Initial implementation of FetchDownloadManager

NotWoods:
The new download manager implementation here using concept-fetch for downloads, and informs the android system once the download is done.

Some parts of the new download manager are blocked by #3481 

Screenshot_20190705-153638

### DIFF
--- a/components/concept/fetch/src/main/java/mozilla/components/concept/fetch/Headers.kt
+++ b/components/concept/fetch/src/main/java/mozilla/components/concept/fetch/Headers.kt
@@ -4,8 +4,6 @@
 
 package mozilla.components.concept.fetch
 
-import java.lang.IllegalArgumentException
-
 /**
  * A collection of HTTP [Headers] (immutable) of a [Request] or [Response].
  */
@@ -52,7 +50,11 @@ interface Headers : Iterable<Header> {
      * @see [Headers.Values]
      */
     object Names {
+        const val CONTENT_DISPOSITION = "Content-Disposition"
+        const val CONTENT_LENGTH = "Content-Length"
         const val CONTENT_TYPE = "Content-Type"
+        const val COOKIE = "Cookie"
+        const val REFERRER = "Referer"
         const val USER_AGENT = "User-Agent"
     }
 
@@ -83,12 +85,13 @@ data class Header(
 /**
  * A collection of HTTP [Headers] (mutable) of a [Request] or [Response].
  */
-class MutableHeaders(
-    vararg pairs: Pair<String, String>
-) : Headers, MutableIterable<Header> {
-    private val headers: MutableList<Header> = pairs.map {
-            (name, value) -> Header(name, value)
-    }.toMutableList()
+class MutableHeaders(headers: List<Header>) : Headers, MutableIterable<Header> {
+
+    private val headers = headers.toMutableList()
+
+    constructor(vararg pairs: Pair<String, String>) : this(
+        pairs.map { (name, value) -> Header(name, value) }.toMutableList()
+    )
 
     /**
      * Gets the [Header] at the specified [index].
@@ -151,4 +154,8 @@ class MutableHeaders(
 
         return append(name, value)
     }
+
+    override fun equals(other: Any?) = other is MutableHeaders && headers == other.headers
+
+    override fun hashCode() = headers.hashCode()
 }
--- a/components/feature/downloads/src/main/AndroidManifest.xml
+++ b/components/feature/downloads/src/main/AndroidManifest.xml
@@ -7,4 +7,5 @@
     package="mozilla.components.feature.downloads">
 
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
 </manifest>
--- a/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
+++ b/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
@@ -1,0 +1,246 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.feature.downloads
+
+import android.annotation.TargetApi
+import android.app.DownloadManager.ACTION_DOWNLOAD_COMPLETE
+import android.app.DownloadManager.EXTRA_DOWNLOAD_ID
+import android.app.Notification
+import android.app.NotificationChannel
+import android.app.NotificationManager
+import android.content.Context
+import android.content.Intent
+import android.net.Uri
+import android.os.Build
+import android.os.Build.VERSION.SDK_INT
+import android.os.Environment
+import android.os.IBinder
+import androidx.core.app.NotificationCompat
+import androidx.core.content.getSystemService
+import androidx.core.net.toUri
+import androidx.localbroadcastmanager.content.LocalBroadcastManager
+import kotlinx.coroutines.Dispatchers.IO
+import kotlinx.coroutines.withContext
+import mozilla.components.browser.session.Download
+import mozilla.components.concept.fetch.Client
+import mozilla.components.concept.fetch.Header
+import mozilla.components.concept.fetch.Headers.Names.CONTENT_DISPOSITION
+import mozilla.components.concept.fetch.Headers.Names.CONTENT_LENGTH
+import mozilla.components.concept.fetch.Headers.Names.CONTENT_TYPE
+import mozilla.components.concept.fetch.Headers.Names.REFERRER
+import mozilla.components.concept.fetch.MutableHeaders
+import mozilla.components.concept.fetch.Request
+import mozilla.components.concept.fetch.Response
+import mozilla.components.feature.downloads.ext.getDownloadExtra
+import mozilla.components.feature.downloads.manager.SystemDownloadManager
+import mozilla.components.feature.downloads.manager.getFileName
+import mozilla.components.support.base.ids.NotificationIds
+import java.io.File
+import java.io.FileOutputStream
+import java.io.OutputStream
+
+/**
+ * Service that performs downloads through a fetch [Client] rather than through the native
+ * Android download manager.
+ *
+ * To use this service, you must create a subclass in your application and it to the manifest.
+ *
+ * @param broadcastManager Override the [LocalBroadcastManager] instance.
+ */
+abstract class AbstractFetchDownloadService(
+    broadcastManager: LocalBroadcastManager? = null
+) : CoroutineService() {
+
+    protected abstract val httpClient: Client
+    private val broadcastManager = broadcastManager ?: LocalBroadcastManager.getInstance(this)
+
+    override fun onCreate() {
+        startForeground(
+            NotificationIds.getIdForTag(this, ONGOING_DOWNLOAD_NOTIFICATION_TAG),
+            buildNotification()
+        )
+        super.onCreate()
+    }
+
+    override fun onBind(intent: Intent?): IBinder? = null
+
+    override suspend fun onStartCommand(intent: Intent?, flags: Int) {
+        val download = intent?.getDownloadExtra() ?: return
+        performDownload(download)
+
+        val downloadID = intent.getLongExtra(EXTRA_DOWNLOAD_ID, -1)
+        sendDownloadCompleteBroadcast(downloadID)
+    }
+
+    private suspend fun performDownload(download: Download) = withContext(IO) {
+        val headers = listOf(
+            CONTENT_TYPE to download.contentType,
+            CONTENT_LENGTH to download.contentLength?.toString(),
+            REFERRER to download.referrerUrl
+        ).mapNotNull { (name, value) ->
+            if (value.isNullOrBlank()) null else Header(name, value)
+        }
+
+        val request = Request(download.url, headers = MutableHeaders(headers))
+
+        val response = httpClient.fetch(request)
+        val filename = download.getFileName(response.headers[CONTENT_DISPOSITION])
+
+        response.body.useStream { inStream ->
+            useFileStream(download, response, filename) { outStream ->
+                inStream.copyTo(outStream)
+            }
+        }
+    }
+
+    /**
+     * Informs [mozilla.components.feature.downloads.manager.FetchDownloadManager] that a download
+     * has been completed.
+     */
+    private fun sendDownloadCompleteBroadcast(downloadID: Long) {
+        val intent = Intent(ACTION_DOWNLOAD_COMPLETE)
+        intent.putExtra(EXTRA_DOWNLOAD_ID, downloadID)
+        broadcastManager.sendBroadcast(intent)
+    }
+
+    /**
+     * Creates an output stream on the local filesystem, then informs the system that a download
+     * is complete after [block] is run.
+     *
+     * Encapsulates different behaviour depending on the SDK version.
+     */
+    @Suppress("LongMethod")
+    internal fun useFileStream(
+        download: Download,
+        response: Response,
+        filename: String,
+        block: (OutputStream) -> Unit
+    ) {
+        /*if (SDK_INT >= Build.VERSION_CODES.Q) {
+            val values = ContentValues().apply {
+                put(MediaStore.Downloads.DISPLAY_NAME, filename)
+                put(MediaStore.Downloads.MIME_TYPE, download.contentType)
+                put(MediaStore.Downloads.SIZE, download.contentLength)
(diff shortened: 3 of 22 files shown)


components/feature/downloads/src/main/java/mozilla/components/feature/downloads/manager/FetchDownloadManager.kt
```diff
@@ -0,0 +1,168 @@
...
+            safeAppend("Cookie", cookie)
+            safeAppend("Referer", refererUrl)
+        }
+        val request = Request(url = download.url, headers = headers)
+
+        return startDownload(download, request, refererUrl.toUri())
+    }
+
+    private suspend fun startDownload(download: Download, request: Request, refererUrl: Uri) = withContext(IO) {
+        val response = httpClient.fetch(request)
```
pocmo: Ideally we wouldn't perform the download here immediately. Instead we should hand it off to a foreground(!) service, primarily to (a) download independently from the app task being still there (b) continuing to download without getting killed. Using a service should also make it easier to use a richer "ongoing download" notification later on - that allows the user to pause/cancel etc.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/manager/FetchDownloadManager.kt
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        if (!applicationContext.isPermissionGranted(INTERNET, WRITE_EXTERNAL_STORAGE)) {
+            throw SecurityException("You must be granted INTERNET and WRITE_EXTERNAL_STORAGE permissions")
+        }
+
+        val headers = MutableHeaders().apply {
+            safeAppend(CONTENT_TYPE, download.contentType)
+            safeAppend("User-Agent", download.userAgent)
+            safeAppend("Cookie", cookie)
```
pocmo: `Cookie` and `User-Agent` primarily exist for the WebView flow and for Android's Download Manager:
* We can omit `User-Agent` since Gecko(View) already uses its own (We only need to tell Android's Download Manager to use one that matches the browser)
* We can set a `cookiePolicy` on the request to let Gecko(View) use whatever cookies it has for this request. This again only exists to pass of cookies from WebView to Android's Download Manager. This should never be set by Gecko(View) and in the worst case we are overriding the Cookie header here with an empty string.

Amejia481: I think we should keep the `Cookie` and `User-Agent` headers as the consumers of `FetchDownloadManager` could pass whatever children of `Client` not necessarily `GeckoViewFetchClient`. In the specific case of `GeckoViewFetchClient`, we could check if the `cookie` are empty in that case Gecko(View) could use whatever cookies it has if the `cookiePolicy` specified that.

samples/browser/src/main/java/org/mozilla/samples/browser/BrowserFragment.kt
```diff
@@ -117,15 +120,22 @@ class BrowserFragment : Fragment(), BackHandler {
...
         downloadsFeature.set(
             feature = DownloadsFeature(
-                requireContext(),
+                requireContext().applicationContext,
                 sessionManager = components.sessionManager,
                 fragmentManager = childFragmentManager,
+                onDownloadCompleted = { download ->
+                    Logger.debug("Download done. $download")
+                },
+                /*downloadManager = FetchDownloadManager(
```
Amejia481: Is there any specific reason why we couldn't use `FetchDownloadManager` here?

NotWoods: I didn't want to switch yet until the new download manager is more mature.

Amejia481: 👍

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -0,0 +1,231 @@
...
+import mozilla.components.concept.fetch.Response
+import mozilla.components.feature.downloads.ext.getDownloadExtra
+import mozilla.components.feature.downloads.manager.SystemDownloadManager
+import mozilla.components.feature.downloads.manager.getFileName
+import mozilla.components.support.base.ids.NotificationIds
+import mozilla.components.support.utils.NotificationUtils
+import java.io.File
+import java.io.FileOutputStream
+import java.io.OutputStream
+
```
csadilek: nit: Add some KDocs :)

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/JobService.kt
```diff
@@ -0,0 +1,57 @@
...
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.launch
+import java.util.concurrent.Executors
+
+/**
+ * Service that runs suspend functions in parallel.
+ * When all jobs are completed, the service is stopped automatically.
+ */
+abstract class JobService(
+    jobDispatcher: CoroutineDispatcher = Executors.newFixedThreadPool(THREADS).asCoroutineDispatcher()
```
csadilek: Should this just use the IO dispatcher? /cc @pocmo

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/manager/FetchDownloadManager.kt
```diff
@@ -0,0 +1,90 @@
...
+import android.util.LongSparseArray
+import androidx.annotation.RequiresPermission
+import androidx.core.util.isEmpty
+import androidx.core.util.set
+import androidx.localbroadcastmanager.content.LocalBroadcastManager
+import mozilla.components.browser.session.Download
+import mozilla.components.feature.downloads.AbstractFetchDownloadService
+import mozilla.components.feature.downloads.ext.putDownloadExtra
+import kotlin.random.Random
+
```
csadilek: nit: kdocs.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/manager/FetchDownloadManager.kt
```diff
@@ -0,0 +1,90 @@
+package mozilla.components.feature.downloads.manager
```
csadilek: license header.

samples/browser/src/main/java/org/mozilla/samples/browser/BrowserFragment.kt
```diff
@@ -109,6 +112,8 @@ class BrowserFragment : Fragment(), BackHandler {
 
         TabsToolbarFeature(layout.toolbar, components.sessionManager, sessionId, ::showTabs)
 
+        val fetchClient = HttpURLConnectionClient()
```
csadilek: Let's move this to `DefaultComponents`

samples/browser/src/main/java/org/mozilla/samples/browser/downloads/DownloadService.kt
```diff
@@ -0,0 +1,12 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.samples.browser.downloads
+
+import mozilla.components.feature.downloads.AbstractFetchDownloadService
+import mozilla.components.lib.fetch.httpurlconnection.HttpURLConnectionClient
+
+class DownloadService : AbstractFetchDownloadService() {
+    override val httpClient = HttpURLConnectionClient()
```
csadilek: Can we get this from the context/components?

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/JobService.kt
pocmo: Yeah, for now that probably makes sense. Eventually I could see us use a separate one if we want to have more control about the number of concurrent downloads?

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/JobService.kt
```diff
@@ -0,0 +1,57 @@
...
+import kotlinx.coroutines.asCoroutineDispatcher
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.launch
+import java.util.concurrent.Executors
+
+/**
+ * Service that runs suspend functions in parallel.
+ * When all jobs are completed, the service is stopped automatically.
+ */
+abstract class JobService(
```
pocmo: nit: Let's try to find a different name for that. [JobService](https://developer.android.com/reference/android/app/job/JobService) is already used in the JobScheduler API and at first I thought you'd use it here. :)

components/concept/fetch/src/main/java/mozilla/components/concept/fetch/Headers.kt
```diff
@@ -52,7 +50,10 @@ interface Headers : Iterable<Header> {
      * @see [Headers.Values]
      */
     object Names {
+        const val CONTENT_DISPOSITION = "Content-Disposition"
+        const val CONTENT_LENGTH = "Content-Length"
         const val CONTENT_TYPE = "Content-Type"
+        const val REFERRER = "Referer"
```
pocmo: Heh, looks like you isolated the misspelling here. :)

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -0,0 +1,231 @@
...
+}
+
+internal object DownloadNotificationChannel {
+    private const val NOTIFICATION_CHANNEL_ID = "Downloads"
+
+    /**
+     * Check if notifications have been blocked for this channel or its group.
+     */
+    fun isBlocked(context: Context) =
+        NotificationUtils.isNotificationBlocked(context, NOTIFICATION_CHANNEL_ID)
```
pocmo: I can't find where this is actually used? 🤔 - Do we even need to check for that?

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -0,0 +1,231 @@
...
+            )
+        }
+    }
+
+    private fun buildNotification(): Notification {
+        val channelId = DownloadNotificationChannel.ensureChannelExists(this)
+
+        return NotificationCompat.Builder(this, channelId)
+            .setSmallIcon(android.R.drawable.stat_sys_download)
+            .setContentTitle("Downloading...")
```
pocmo: nit: Needs to move to strings.xml (and check whether that file is listed in l10n.toml.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -0,0 +1,231 @@
...
+                showNotification
+            )
+        }
+    }
+
+    private fun buildNotification(): Notification {
+        val channelId = DownloadNotificationChannel.ensureChannelExists(this)
+
+        return NotificationCompat.Builder(this, channelId)
+            .setSmallIcon(android.R.drawable.stat_sys_download)
```
pocmo: I wonder if we have a Firefox icon for this (which will avoid that this is a random icon the manufacturer decided to use). For example there's a download icon here: https://design.firefox.com/icons/viewer/#

samples/browser/src/main/java/org/mozilla/samples/browser/downloads/DownloadService.kt
Amejia481: I'm not quite sure if we want that every app have to create their own service to use this functionality, Should AC provide a default and avoid extra steps for the consumers?

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/JobService.kt
Amejia481: This looks pretty similar to [IntentService](https://developer.android.com/reference/android/app/IntentService)

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/manager/FetchDownloadManager.kt
```diff
@@ -0,0 +1,90 @@
...
+) : BroadcastReceiver(), DownloadManager {
+
+    private val queuedDownloads = LongSparseArray<Download>()
+    private var isSubscribedReceiver = false
+
+    override val permissions = arrayOf(INTERNET, WRITE_EXTERNAL_STORAGE, FOREGROUND_SERVICE)
+
+    @RequiresPermission(allOf = [INTERNET, WRITE_EXTERNAL_STORAGE, FOREGROUND_SERVICE])
+    override fun download(download: Download, cookie: String): Long? {
+        if (!download.isScheme(listOf("http", "https"))) {
```
Amejia481: They're some validations that I noticed that are duplicated between the Android download manger and this one. Maybe in the feature we may want to share validations.

samples/browser/src/main/AndroidManifest.xml
```diff
@@ -67,6 +67,8 @@
             </intent-filter>
         </service>
 
+        <service android:name=".downloads.DownloadService" />
```
Amejia481: Could this be internal to ac?

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
NotWoods: I was initially using this for the completion notification, falling back to native if the channel is blocked. I ended up not implementing a download complete notification for now, so I can pull this code out into its own PR.

components/feature/downloads/src/main/AndroidManifest.xml
```diff
@@ -7,4 +7,6 @@
     package="mozilla.components.feature.downloads">
 
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
-</manifest>
\ No newline at end of file
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
+    <uses-permission android:name="android.permission.DOWNLOAD_WITHOUT_NOTIFICATION" />
```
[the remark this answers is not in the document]
NotWoods: I was testing a custom download completed notification which required this permission. I'll pull it out for now.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/manager/FetchDownloadManager.kt
NotWoods: This validation will change in the future as GeckoView gets support for blob URLs (and possibly other schemes). I tried to share most of the validation code and allow for a custom scheme list.

samples/browser/src/main/java/org/mozilla/samples/browser/downloads/DownloadService.kt
pocmo: I think it's okay to force the app to provide the dependency here. The app team should make a conscious decision. The only possible default is `HttpURLConnectionClient` and that's the opposite of what we'd recommend for Mozilla apps.

NotWoods:
>This looks pretty similar to IntentService

`IntentService` only supports one task at a time, so we couldn't have simultaneous downloads.

>I'm not quite sure if we want that every app have to create their own service to use this functionality, Should AC provide a default and avoid extra steps for the consumers?

I don't think AC should provide a default, since that would force every app using AC to include the default in addition to whatever client they actually use. It also makes using two different clients more likely since a developer could forget to override a default.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
NotWoods: Since this PR is already pretty large, I'll have a follow up for a custom icon. Using the native icon means that we'll blend in anyways.

pocmo: Sounds good!

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -0,0 +1,246 @@
...
+    /**
+     * Build the notification to be displayed while the service is active.
+     */
+    private fun buildNotification(): Notification {
+        val channelId = ensureChannelExists(this)
+
+        return NotificationCompat.Builder(this, channelId)
+            .setSmallIcon(android.R.drawable.stat_sys_download)
+            .setContentTitle("Downloading...")
+            .setContentText("Download in progress")
```
pocmo: You created strings in `strings.xml`, but you do not use them here. :)

PR #3552, opened 2019-06-25
Closes #3533: browser-state: Dispatch actions from Session

### LINKED ISSUE #3533: browser-state: Dispatch actions from Session
To keep a `Session` in sync with its `TabSessionState` / `CustomTabSessionState`.

csadilek:
Decided to keep the tests separate (not amend the existing ones) because it's easier to read this way. We had some inconsistent formatting here as well, which I fixed.

### DIFF
--- a/components/browser/session/src/test/java/mozilla/components/browser/session/SessionTest.kt
+++ b/components/browser/session/src/test/java/mozilla/components/browser/session/SessionTest.kt
@@ -12,7 +12,10 @@ import kotlinx.coroutines.runBlocking
 import mozilla.components.browser.session.Session.Source
 import mozilla.components.browser.session.engine.request.LoadRequestMetadata
 import mozilla.components.browser.session.engine.request.LoadRequestOption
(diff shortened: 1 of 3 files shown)


components/browser/session/src/test/java/mozilla/components/browser/session/SessionTest.kt
```diff
@@ -91,6 +94,19 @@ class SessionTest {
...
+        val store: BrowserStore = mock()
+        `when`(store.dispatch(any())).thenReturn(mock())
+
+        val session = Session("https://www.mozilla.org")
+        session.store = store
+        session.url = "http://www.firefox.com"
+
+        verify(store).dispatch(ContentAction.UpdateUrlAction(session.id, session.url))
+        verifyNoMoreInteractions(store)
+    }
```
pocmo: Personally I'd prefer more of a blackbox test here that verifies that an actual `Session`<-> `(Custom)TabSessionState` pair is kept in sync - instead of verifying an implementation detail (which action is dispatched). But then again this test is good to have too.

PR #3553, opened 2019-06-25
Use interface with features that need permissions

NotWoods:
This fixes the need to suppress unused parameters in a couple of places, and enforces consistency in the future.


pocmo:
Nice! Looks great. Those interface will make it easier to make those implementations consistent and reduce some of the boilerplate that may be required in the future.

Taskcluster has some complaints though! :)

Issue #3526, opened 2019-06-25
browser-state: Update selected session if selected tab is removed


pocmo:
For `browser-session` parity.

PR #3554, opened 2019-06-25
Closes #3525: browser-state: Auto-select first tab

### LINKED ISSUE #3525: browser-state: Auto-select first tab
For `browser-session` parity.

csadilek:
This one depends a bit on how we'll implement https://github.com/mozilla-mobile/android-components/issues/3526. Decided to just check for `selectedTabId == null` as that's true initially (empty state) and it would be set to null when the last tab is removed (similar to our `NO_SELECTION` checks in `LegacySessionManager`).

### DIFF
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TabListReducer.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TabListReducer.kt
@@ -15,9 +15,16 @@ internal object TabListReducer {
      */
     fun reduce(state: BrowserState, action: TabListAction): BrowserState {
         return when (action) {
-            is TabListAction.AddTabAction -> state.copy(
-                tabs = state.tabs + action.tab,
-                selectedTabId = if (action.select) action.tab.id else state.selectedTabId)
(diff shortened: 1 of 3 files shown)


NotWoods:
Instead of doing this check in the reducer, just change the `selectedTab` getter

dector:
@NotWoods I'm not sure if it's good idea to move presentation logic into state model.

csadilek:
> Instead of doing this check in the reducer, just change the `selectedTab` getter

@NotWoods interesting. While that would work, the advantage I see with the current approach is that we'll have all state changes and decisions in one place (the reducer). This will make things easier to maintain and debug esp. when we add more actions that affect `selectedTabId` e.g. `RemoteTabAction` and `RemoveAllTabsAction`. What advantage do you see moving this to the state?

pocmo:
> Instead of doing this check in the reducer, just change the selectedTab getter

This is not explicit though (you will *have* to go through the getter and looking at / debugging the state hides that completely from you. Also note that this would require a stable order, I guess? Apps like Fennec allow you to re-order tabs.

Issue #3567, opened 2019-06-26
BrowserStore/SessionManager: Make result of "nearby" tab selection match


pocmo:
We use the same oscillating algorithm in both. However since `SessionManager` has custom tabs in the same list the result can vary.

The following unit test fails currently:

```kotlin
    @Test
    fun `Remove session will not select custom tab`() {
        val session1 = Session(id = "session1", initialUrl = "https://www.firefox.com").apply {
            customTabConfig = mock() }
        val session2 = Session(id = "session2", initialUrl = "https://developer.mozilla.org/en-US/")
        val session3 = Session(id = "session3", initialUrl = "https://www.mozilla.org/en-US/internet-health/").apply {
            customTabConfig = mock() }
        val session4 = Session(id = "session4", initialUrl = "https://www.mozilla.org/en-US/technology/")
        val session5 = Session(id = "session5", initialUrl = "https://example.org/555").apply {
            customTabConfig = mock() }
        val session6 = Session(id = "session6", initialUrl = "https://example.org/Hello").apply {
            customTabConfig = mock() }
        val session7 = Session(id = "session7", initialUrl = "https://example.org/World").apply {
            customTabConfig = mock() }
        val session8 = Session(id = "session8", initialUrl = "https://example.org/JustTestingThings").apply {
            customTabConfig = mock() }
        val session9 = Session(id = "session9", initialUrl = "https://example.org/NoCustomTab")

        val state = BrowserState()
        val store = BrowserStore(state)

        val manager = SessionManager(engine = mock(), store = store)
        manager.add(session1)
        manager.add(session2)
        manager.add(session3)
        manager.add(session4)
        manager.add(session5)
        manager.add(session6)
        manager.add(session7)
        manager.add(session8)
        manager.add(session9)

        manager.select(session4)
        assertEquals(session4, manager.selectedSession)
        assertEquals("session4", store.state.selectedTabId)

        //---------------------------------------------------------
        // The following assertion fails:
        // -> SessionManager selects tab 2
        // -> BrowserStore selects tab 9
        //---------------------------------------------------------
        manager.remove(session4)
        assertEquals(session2, manager.selectedSession)
        assertEquals("session2", store.state.selectedTabId)

        manager.remove(session2)
        assertEquals(session9, manager.selectedSession)
        assertEquals("session9", store.state.selectedTabId)

        manager.remove(session9)
        assertNull(manager.selectedSession)
        assertNull(store.state.selectedTabId)
    }
```

[...]

PR #3570, opened 2019-06-26
Closes #3526: Update selected tab id after selected tab is removed.

### LINKED ISSUE #3526: browser-state: Update selected session if selected tab is removed

pocmo:
Follow-up #3567

### DIFF
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TabListReducer.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TabListReducer.kt
@@ -8,6 +8,7 @@ import mozilla.components.browser.state.action.TabListAction
 import mozilla.components.browser.state.selector.findTab
 import mozilla.components.browser.state.state.BrowserState
 import mozilla.components.browser.state.state.TabSessionState
+import kotlin.math.max
 
 internal object TabListReducer {
     /**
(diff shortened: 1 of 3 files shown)


components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TabListReducer.kt
```diff
@@ -26,16 +27,93 @@ internal object TabListReducer {
...
+            }
 
             is TabListAction.RemoveTabAction -> {
                 val tab = state.findTab(action.tabId)
+
                 if (tab == null) {
                     state
                 } else {
-                    state.copy(tabs = state.tabs - tab)
+                    val previousIndex = state.tabs.indexOf(tab)
```
NotWoods: nit: Move this computation inside the if then branch
```kt
val updatedSelection = if (state.selectedTabId == tab.id) {
    val previousIndex = state.tabs.indexOf(tab)
    findNewSelectedTabId(updatedTabList, tab.content.private, previousIndex)
} else ...
```

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TabListReducer.kt
```diff
@@ -26,16 +27,93 @@ internal object TabListReducer {
...
+    tabs: List<TabSessionState>,
+    isPrivate: Boolean,
+    previousIndex: Int
+): String? {
+    if (tabs.isEmpty()) {
+        // There's no tab left to select.
+        return null
+    }
+
+    val predicate: (TabSessionState) -> Boolean = { tab -> tab.content.private == isPrivate }
```
NotWoods: Can you omit the type if you do this?
```suggestion
    val predicate = { tab: TabSessionState -> tab.content.private == isPrivate }
```

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TabListReducer.kt
```diff
@@ -26,16 +27,93 @@ internal object TabListReducer {
...
+    val maxSteps = max(tabs.lastIndex - index, index)
+    if (maxSteps < 0) {
+        return null
+    }
+
+    // Try tabs oscillating near the index.
+    for (steps in 1..maxSteps) {
+        listOf(index - steps, index + steps).forEach { current ->
+            if (current >= 0 &&
+                current <= tabs.lastIndex &&
```
NotWoods: nit: `current in 0..tabs.lastIndex`

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TabListReducer.kt
pocmo: Yeah, that works. But do you think it's more readable? If I omit the type then Android Studio will show it itself and there's actually more text on the screen after the change :D

NotWoods: I didn't realize Android Studio does that, feel free to keep it as you originally wrote it.

PR #3572, opened 2019-06-26
Closes #3561: Add default (empty) state to BrowserStore

### LINKED ISSUE #3561: BrowserStore: Set default BrowserState
Currently you need to provide a `BrowserState` when creating a `BrowserStore`. Since we restore asynchronously it may make sense to just use an empty `BrowserState()` by default.

csadilek:


csadilek:
> What do you think about a "sanity check" test that a BrowserStore() has no tabs and no selected tab id?

Sounds good. I will add a test.

csadilek:
Updated.

PR #3579, opened 2019-06-27
Sync access token caching, API simplification

grigoryk:
I've started pulling on one little thread, and ended up with a few more changes than initially anticipated.

Raison d'être for this PR - introducing access token caching for Sync.
- Some background on the issue: Rust FirefoxAccount object maintains an in-memory cache of access tokens, keyed by 'scope'. During every sync, we "rehydrate" an instance of FirefoxAccount, starting with a fresh cache. We then obtain an access token from it to sync; this performs a network request (since the internal cache is empty), which is quite costly at scale for our services. This creates a situation when we may overwhelm our own servers with a large enough, actively syncing user base.
- This PR adds a caching layer for sync authInfo objects. Sync workers no longer interact with the account directly, and instead look into the cache to obtain authentication info necessary for syncing. No more "talk to the FxA server before every sync".
Account manager is responsible for keeping the cache up-to-date, and resetting it when necessary. Cache is currently updated: on startup (but only if access token has expired), on authentication, and when we recover from auth problems.

And this is where the "thread pulling" begins! In order to "own" the access token for sync, account manager needs to be aware of the "sync scope".
Before, we just relied on the application to specify that scope. Instead, I've changed account manager's constructor to take a SyncConfig object which allows consuming application to configure how sync should behave (enabled at all?, periodic syncing enabled? how often to sync? which stores should be synced?).
Ownership of the "sync manager" moved down the stack, from the application layer into the account manager.

Application is now expected to interact with sync only via AccountManager's `sync` method, which exposes an internal SyncManager instance (if sync is enabled).

Above changes were a good reason to move support classes from feature-sync and into services-firefox-account. Note that since "sync" is part of our "storage" modules, this change doesn't mean that you need to take an extra native dependency on your classpath simply if you need to use FxA. Thanks to concept-sync, actual "Firefox Sync" machinery (within libplaces) is still fully decoupled from FxA. `feature-sync` has been removed entirely.

Since we're churning the public API anyway, I took the chance to introduce a few more simplifications at the API layer:
- 'SyncManager' interface was removed, since we're not expecting to have multiple implementations of it
- 'Config' was renamed to 'ServerConfig'
- 'DeviceTuple' was renamed to 'DeviceConfig'
- account manager grew a new public API, 'setSyncConfig', which allows application to re-configure how it wants sync to behave
- 'AuthInfo' was renamed to 'SyncAuthInfo', and a bunch of cleanup happened in that area
- 'AccountObservable'@'onError' method was removed. The only error that could have been passed into it (unable to restore account) wasn't actionable by the application anyway, and none of the integrations did anything with that call

Documentation of public APIs and classes was improved.

### Pull Request checklist


samples/sync/src/main/java/org/mozilla/samples/sync/MainActivity.kt
```diff
@@ -139,7 +139,7 @@ class MainActivity :
 
         // NB: ObserverRegistry takes care of unregistering this observer when appropriate, and
         // cleaning up any internal references to 'observer' and 'owner'.
-        syncManager.register(syncObserver, owner = this, autoPause = true)
+        accountManager.syncManager?.register(syncObserver, owner = this, autoPause = true)
```
grigoryk: Hmmm, currently `syncManager` is only available when we have an authenticated account. However, this will be called before that! meaning that we won't actually ever register for sync events here.

grigoryk:
@csadilek I think I like the API as it stands now. I've simplified it a bit further since we last chatted about it.

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/Types.kt
```diff
@@ -5,20 +5,20 @@
...
 import mozilla.appservices.sync15.FailureReason
 import mozilla.components.concept.storage.VisitInfo
 import mozilla.components.concept.storage.VisitType
-import mozilla.components.concept.sync.AuthInfo
 
 // We have type definitions at the concept level, and "external" types defined within Places.
 // In practice these two types are largely the same, and this file is the conversion point.
 
 /**
  * Conversion from a generic AuthInfo type into a type 'places' lib uses at the interface boundary.
```
csadilek: nit: update comment?

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/Types.kt
```diff
@@ -5,20 +5,20 @@
...
  */
-internal fun AuthInfo.into(): SyncAuthInfo {
+internal fun mozilla.components.concept.sync.SyncAuthInfo.into(): SyncAuthInfo {
```
csadilek: 👍 I like this change...made the code easier to read overall. Maybe we can find a different name later, because `into` is still a bit hard to grasp e.g. `unwrap()` or `toNative()`?

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/Config.kt
```diff
@@ -4,4 +4,40 @@
...
+import mozilla.components.concept.sync.DeviceType
+import mozilla.components.service.fxa.manager.FxaAccountManager
+import mozilla.components.service.fxa.sync.GlobalSyncableStoreProvider
+
+typealias ServerConfig = mozilla.appservices.fxaclient.Config
+
+/**
+ * Configuration for the current device.
+ *
+ * @param name An initial name to use for the device record which will be created during authentication.
```
csadilek: nit: These are all `@property`.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/Config.kt
```diff
@@ -4,4 +4,40 @@
...
+data class DeviceConfig(
+    val name: String,
+    val type: DeviceType,
+    val capabilities: Set<DeviceCapability>
+)
+
+/**
+ * Configuration for sync.
+ *
+ * @param syncableStores A set of store names to sync, exposed via [GlobalSyncableStoreProvider].
```
csadilek: nit: `@property`

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/SyncAuthInfoCache.kt
```diff
@@ -0,0 +1,80 @@
...
+        val o = JSONObject()
+        o.put(KEY_KID, this.kid)
+        o.put(KEY_FXA_ACCESS_TOKEN, this.fxaAccessToken)
+        o.put(KEY_FXA_ACCESS_TOKEN_EXPIRES_AT, this.fxaAccessTokenExpiresAt)
+        o.put(KEY_SYNC_KEY, this.syncKey)
+        o.put(KEY_TOKEN_SERVER_URL, this.tokenServerUrl)
+        // JSONObject swallows any JSONExceptions thrown during 'toString', and simply returns 'null'.
+        val asString: String? = o.toString()
+        if (asString == null) {
+            logger.error("Failed to stringify")
```
csadilek: In this case do we really want to write `null` to the cache? It seems fine as it will remove any older value...or should we not write to the cache and throw an exception instead so we know that something is badly broken? :)

samples/sync-logins/src/main/java/org/mozilla/samples/sync/logins/MainActivity.kt
```diff
@@ -60,10 +57,9 @@ class MainActivity : AppCompatActivity(), LoginFragment.OnLoginCompleteListener,
     private val accountManager by lazy {
         FxaAccountManager(
                 applicationContext,
-                Config.release(CLIENT_ID, REDIRECT_URL),
-                arrayOf("profile", "https://identity.mozilla.com/apps/oldsync"),
-                DeviceTuple("A-C Logins Sync Sample", DeviceType.MOBILE, listOf()),
-                syncManager
+                ServerConfig.release(CLIENT_ID, REDIRECT_URL),
+                DeviceConfig("A-C Logins Sync Sample", DeviceType.MOBILE, setOf()),
+                SyncConfig(setOf("logins"))
```
csadilek: 👍 this looks much better now!

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/Types.kt
grigoryk: Sure, happy to explore other names. I've used `into` as a direct carry-over from Rust, actually.
https://doc.rust-lang.org/std/convert/trait.Into.html

pocmo: In Kotlin those are usually `to*()` methods.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/SyncAuthInfoCache.kt
grigoryk: Let's throw for now, and see if we hit that at all :) Then we can consider fixing this if necessary.

grigoryk:
This is ready to land. Waiting for green CI.

Issue #3480, opened 2019-06-18
Port ColorPicker from Fennec and add Advanced mode


NotWoods:
Fennec has a nice picker for `<input type="color">` that we can port over to A-C. However, that picker (and the current A-C picker) are limited to only selecting predefined colors. 

An advanced mode for the Fennec color picker was designed but seemingly not implemented, and we should implement that mode too.

![](https://blog.mozilla.org/ux/files/2013/12/mob_cpicker.png)

https://blog.mozilla.org/ux/2013/12/a-new-colour-picker-on-firefox-for-android/

NotWoods:
We should also add a `contentDescription` to the colours for screen readers. @thesillypeanut wrote code for this in Firefox Desktop DevTools: https://phabricator.services.mozilla.com/D35980. The patch can be ported to the color picker code in Kotlin.

pocmo:
@antlam You may recognize this color picker :)

antlam:
@pocmo HAH! ❤️ 

glad to be of service!

PR #3588, opened 2019-06-27
Issue #3480 - Port Fennec basic color picker

NotWoods:
Redesigns the color picker based on the design in Fennec. This PR only includes the basic list implementation, future updates will add content descriptions for screen readers and an advanced picker mode.

### Pull Request checklist


components/feature/prompts/src/main/java/mozilla/components/feature/prompts/ColorPickerDialogFragment.kt
```diff
@@ -35,137 +33,91 @@ internal class ColorPickerDialogFragment : PromptDialogFragment() {
...
-
-        override fun getItemCount(): Int = defaultColors.size
-
-        override fun onCreateViewHolder(parent: ViewGroup, type: Int): RecyclerView.ViewHolder {
-            val view = inflater.inflate(R.layout.mozac_feature_prompts_color_item, parent, false)
-            return ColorViewHolder(view)
+        // Set up recycler view
+        listAdapter = BasicColorAdapter(this::onColorChange)
+        view.findViewById<RecyclerView>(R.id.recyclerView).apply {
+            layoutManager = LinearLayoutManager(context, RecyclerView.VERTICAL, false)
```
Amejia481: The initial selected color is not visible as is the last element of the list, something that not happens on Fennec. You could test with this [site](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color).

Screenshot_20190628-120530

I think could add `stackFromEnd = true` to the `layoutManager` that will fix that.

PR #3589, opened 2019-06-27
Experiments: add matcher for min/max versions

travis79:
- Add experiments matcher functionality for min and max versions
- Add VersionString utility class to aid in comparing versions
- Update tests to for coverage of matchers

### Pull Request checklist


components/service/experiments/README.md
```diff
@@ -178,12 +178,14 @@ The experiments records in Kinto contain the following properties:
...
+| match.locale_language     |        |          | The default locales language.                                                                                                                   | "eng|zho|deu"                                                  |
+| match.app_display_version | regex  |          | The application version.                                                                                                                        | `"1.0.0"`                                                      |
+| match.app_min_version     | string |          | The minimum application version, inclusive.                                                                                                     | `"1.0.0"`                                                      |
+| match.app_max_version     | string |          | The maximum application version, inclusive.                                                                                                     | `"1.0.0"`                                                      |
```
georgf: Nit: It would be nice to use different versions for these, e.g. "1.0.1" & "1.4.2", to highlight how those could be used.

components/service/experiments/src/main/java/mozilla/components/service/experiments/util/VersionString.kt
```diff
@@ -0,0 +1,73 @@
+package mozilla.components.service.experiments.util
+
+import java.lang.IllegalArgumentException
+import kotlin.math.max
+
+internal class VersionString(
```
georgf: Isn't there a lot of overlap to [`EngineVersion`](https://github.com/mozilla-mobile/android-components/blob/master/components/concept/engine/src/main/java/mozilla/components/concept/engine/utils/EngineVersion.kt) here?
We shouldn't duplicate logic if we can avoid it. Why not share this as single utility code somewhere?

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentEvaluatorTest.kt
```diff
@@ -251,6 +251,64 @@ class ExperimentEvaluatorTest {
...
+
+        packageInfo.versionName = "1.1.18"
+        assertNotNull(evaluator.evaluate(context, ExperimentDescriptor("testexperiment"), listOf(experiment), 20))
+
+        packageInfo.versionName = "0.1.18"
+        assertNull(evaluator.evaluate(context, ExperimentDescriptor("testexperiment"), listOf(experiment), 20))
+    }
+
+    @Test
+    fun `evaluate appMaxVersion`() {
```
georgf: Please also add a test for evaluating both min & max version together, both matching and non-matching.
Please also test on the edges of the possible value spaces, e.g. for `evaluate appMinVersion` with min=1.0.0, test for version being: 0.9.9, 1.0.0, 1.0.1, somewhere further away like 1.1.18.

components/service/experiments/src/main/java/mozilla/components/service/experiments/util/VersionString.kt
travis79: I'll file a follow-up bug and discuss it with the A-C team

travis79: I was worried about creating a cross-dependency right now

travis79: Filed the bug to follow-up [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1562241).  I don't think that this blocks or hurts us to proceed with this functionality living in service-experiments for now.

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentEvaluatorTest.kt
travis79: I expanded the test cases to better cover edge cases and close cases.

components/service/experiments/src/main/java/mozilla/components/service/experiments/util/VersionString.kt
```diff
@@ -0,0 +1,73 @@
...
+    override fun compareTo(other: VersionString): Int {
+        if (!version.matches(VERSION_REGEX)) {
+            throw IllegalArgumentException("Unexpected format in VersionString")
+        }
+        if (!other.version.matches(VERSION_REGEX)) {
+            throw IllegalArgumentException("Unexpected format in VersionString")
+        }
+        val thisParts = version.split(".")
+        val otherParts = other.version.split(".")
+        val length = max(thisParts.count(), otherParts.count())
```
georgf: Your code supports versions with different element counts, but i don't see test coverage for that scenario.
Let's add that before landing.

components/service/experiments/src/main/java/mozilla/components/service/experiments/util/VersionString.kt
```diff
@@ -0,0 +1,73 @@
...
+        for (i in 0..length) {
+            val thisPart = if (i < thisParts.count()) {
```
georgf: Non-blocking feedback: filling the `*Parts` lists up with `0`s before the loop would probably make for more readable code.

travis79: Added a comment for now, to better explain what's going on.

components/service/experiments/src/main/java/mozilla/components/service/experiments/util/VersionString.kt
travis79: No problem, added a handful of tests for both shorter and longer element counts

Issue #3557, opened 2019-06-26
browser-state: Add restore behavior


pocmo:
Needed for save and restore of the "browsing session". We need to make sure to keep the same format. Maybe we can turn `Snapshot` into something more generic that can be used from/by `SessionManager` as well as `BrowserStore`?

pocmo:
I'll transform this into being only about adding `restore` and making `SessionManager.restore()` sync the state with `BrowserStore`. That's all we need right now.

Issue #3432, opened 2019-06-13
SessionManager: restore and intent processor race condition - tab added randomly at start/end of list.


sblatz:
It'd be great if we had timestamps on session creation (or most recently used) so we could sort them accordingly in Fenix.

Would help us address: https://github.com/mozilla-mobile/fenix/issues/3346

@jonalmeida and I chatted about this and he has more info :)

jonalmeida:
To clarify the current use case, when a session restore is taking place asynchronously, opening a new session from a cold boot puts the session at the top of the list of session and then follows with the restored tabs. This list is presented in a reverse order so puts the new session at the end of the collections list.

pocmo:
It may be time to get UX back into this. Fenix has already diverged on how browsers traditionally sort tabs. Changing the order will fix the linked issue but very likely break other situations (e.g. child tabs opened next to their parent). We should look at those use cases and scenarios and define the behavior together with the UX team. AC currently just uses the same behavior as our other and third-party browsers.

That said, there's definitely an issue here with concurrent restore and incoming intent. And that's something we should fix - and that will fix https://github.com/mozilla-mobile/fenix/issues/3346 too without having to reorder.

pocmo:
Two potential solutions I can think of:
* Delaying adding a `Session` coming from an Intent after restore has happened (Makes loading that tab slower though?)
* Adding restored tabs to the beginning of the list (as if they were restored first, shouldn't regress loading the incoming tab).

sblatz:
I think your second solution sounds reasonable. Will that cause sorting issues when we use the "open collection" feature (which will restore previous sessions with unique ID's)?

pocmo:
Yes, it will. Either we add a flag for this behavior - or change the way how we add tabs from a collection (a new "add multiple tabs" API vs. "restore") - which may make sense since we do not "restore" anymore but create new sessions (ids) now.

pocmo:
@sblatz I implemented the following changes:
*  LANDED: (browser-session)`SessionManager.restore()` will always add tabs at the beginning of the list so that restoring can happen asynchronously / concurrently and it will always look like those tabs were the first.
* IN REVIEW: (feature-tab-collections) `Tab.restore()` now returns a `Session` (instead of a `Snapshot`) and `TabCollection.restore()/restoreSubset()` now returns a `List<Session>` (instead of a `Snapshot`)
[...]

PR #3590, opened 2019-06-27
Align and synchronize restore behavior between SessionManager and BrowserStore

pocmo:
Issue #3557 and issue #3432.

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
@@ -4,6 +4,7 @@
 
 package mozilla.components.browser.session
 
+import androidx.lifecycle.Transformations.map
 import mozilla.components.browser.session.ext.syncDispatch
 import mozilla.components.browser.session.ext.toCustomTabSessionState
 import mozilla.components.browser.session.ext.toTabSessionState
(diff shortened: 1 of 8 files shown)


components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
```diff
@@ -122,7 +122,32 @@ class SessionManager(
      * @param snapshot A [Snapshot] which may be produced by [createSnapshot].
      * @param updateSelection Whether the selected session should be updated from the restored snapshot.
      */
-    fun restore(snapshot: Snapshot, updateSelection: Boolean = true) = delegate.restore(snapshot, updateSelection)
+    fun restore(snapshot: Snapshot, updateSelection: Boolean = true) {
+        delegate.restore(snapshot, updateSelection)
+
+        val tabs = snapshot.sessions.mapNotNull { item ->
```
csadilek: nit: filter then map would be easier to read (but that's subjective) :)

components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
```diff
@@ -122,7 +122,32 @@ class SessionManager(
...
+            } else {
+                // A restored snapshot should not contain any custom tab so we should be able to safely ignore
+                // them here.
+                null
+            }
+        }
+
+        val selectedTabId = if (updateSelection && snapshot.selectedSessionIndex != NO_SELECTION) {
+            val index = snapshot.selectedSessionIndex
+            if (index in 0..snapshot.sessions.lastIndex) {
```
csadilek: nit: index <= lastIndex?

components/browser/session/src/test/java/mozilla/components/browser/session/SessionManagerMigrationTest.kt
```diff
@@ -200,4 +200,100 @@ class SessionManagerMigrationTest {
         assertNull(manager.selectedSession)
         assertNull(store.state.selectedTabId)
     }
+
+    @Test
+    fun `Restoring snapshot with invalid index`() {
+        val store = BrowserStore(BrowserState())
```
csadilek: nit: don't need to provide empty state.

components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
pocmo: Not sure about this comment. We also need to check >= 0 (at least no selection is -1).

csadilek: Hm not sure what I wanted to say there :).

>We also need to check >= 0 (at least no selection is -1).
Yeah, makes sense.

PR #3596, opened 2019-06-28
Closes #3559: Add actions for removing tabs to browser-state

### LINKED ISSUE #3559: browser-state: Sync SessionManager.removeAll()/removeSessions()
We need to synchronize calls to `removeSessions()` and `removeAll()`.

I wonder if we can get rid of `removeAll()` - removing all sessions including custom tabs seems like something we wouldn't ever want? We use it in tests currently only.

csadilek:
Decided to implement all cases here since we also have `TabsUseCases.RemoveAllTabsOfTypeUseCase` so having independent action will come in handy:

- `TabsListAction.RemoveAllTabsAction`: Remove all tabs (private and non-private; maps to current `removeSessions` in SessionManager)
- `TabsListAction.RemoveAllPrivateTabsAction`: Remove all private tabs
- `TabsListAction.RemoveAllNormalTabsAction`: Remove all non-private (normal) tabs
- `CustomTabListAction.RemoveAllCustomTabsAction`: Remove all custom tabs

We probably want to deprecate `removeAll` in `SessionManager`, as it's not needed (except in tests), but I'd vote for doing that separately (in a follow-up). For now I've made it work by dispatching both `RemoveAllTabsAction` and `RemoveAllCustomTabsAction`.

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
@@ -187,6 +187,9 @@ class SessionManager(
      */
     fun removeSessions() {
         delegate.removeSessions()
+        store?.syncDispatch(
+            TabListAction.RemoveAllTabsAction
+        )
     }
 
     /**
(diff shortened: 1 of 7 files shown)


pocmo:
> We probably want to deprecate removeAll in SessionManager, as it's not needed (except in tests), but I'd vote for doing that separately (in a follow-up). For now I've made it work by dispatching both RemoveAllTabsAction and RemoveAllCustomTabsAction.

Fully agree with that! :+1:

PR #3597, opened 2019-06-28
Closes #3521: Expose GeckoView load flags for loadUrl calls

### LINKED ISSUE #3521: Expose GeckoView load flags for loadUrl calls
We currently don't provide any load flags when calling `GeckoSession.loadUri`. We should expose this API so that we can provide the various supported flags esp. `LOAD_FLAGS_EXTERNAL`.

csadilek:
This allows specifying Gecko load flags (similar to other Gecko flags e.g. for clear data and tracking protection) and should also keep the API backward compatible.

### DIFF
--- a/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
+++ b/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
@@ -143,9 +143,9 @@ abstract class EngineSession(
             const val SAFE_BROWSING_ALL =
                 SAFE_BROWSING_MALWARE + SAFE_BROWSING_UNWANTED + SAFE_BROWSING_HARMFUL + SAFE_BROWSING_PHISHING
 
-            internal const val RECOMMENDED: Int = AD + ANALYTICS + SOCIAL + TEST + SAFE_BROWSING_ALL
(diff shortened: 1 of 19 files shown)


components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
```diff
@@ -206,10 +206,43 @@ abstract class EngineSession(
...
+        companion object {
+            const val NONE: Int = 0
+            const val BYPASS_CACHE: Int = 1 shl 0
+            const val BYPASS_PROXY: Int = 1 shl 1
+            const val EXTERNAL: Int = 1 shl 2
+            const val ALLOW_POPUPS: Int = 1 shl 3
+            const val BYPASS_CLASSIFIER: Int = 1 shl 4
+            internal const val ALL = BYPASS_CACHE + BYPASS_PROXY + EXTERNAL + ALLOW_POPUPS + BYPASS_CLASSIFIER
+
+            fun all() = LoadUrlFlags(ALL)
```
pocmo: Is `ALL` something that would ever make sense to actually use?

pocmo:
I'll fix the conflicts and land.

Issue #1514, opened 2018-11-29
Browser toolbar: Follow app bar specs


pocmo:
https://material.io/design/components/app-bars-top.html#specs

In this case I mean the icon specs and specifically for the menu button. In our implementation there's a noticeable large gap between the menu and browser actions.

jonalmeida:
Bram has also provided us with a design layout for Custom Tabs:
![](https://user-images.githubusercontent.com/51007/53133685-dee0e700-35d8-11e9-96d8-09c2be0bbdd4.png)

Amejia481:
Related https://github.com/mozilla-mobile/android-components/issues/4503

rocketsroger:
Related https://github.com/mozilla-mobile/android-components/issues/4442

sblatz:
(I believe) related: https://github.com/mozilla-mobile/fenix/issues/5671

pocmo:
@sblatz Good to know that it is already broken. I was confused why it is broken in my refactoring branch. :D

Will fix this as part of this bug here.

theres-waldo:
Is there a corresponding patch for `reference-browser` that allows it to build against this change?

PR #3610, opened 2019-07-01
For #1514: Better follow app bar specs

sblatz:
### Before

Screen Shot 2019-07-01 at 10 38 45 AM

### After
Screen Shot 2019-07-01 at 10 43 27 AM

### Pull Request checklist

### DIFF
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
@@ -16,7 +16,6 @@ import androidx.appcompat.widget.AppCompatImageView
 import androidx.appcompat.widget.AppCompatTextView
 import androidx.core.content.ContextCompat
 import androidx.core.view.isVisible
-import androidx.core.view.setPadding
 import mozilla.components.browser.menu.BrowserMenuBuilder
 import mozilla.components.browser.toolbar.BrowserToolbar
 import mozilla.components.browser.toolbar.R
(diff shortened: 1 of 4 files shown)


components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
```diff
@@ -356,13 +362,12 @@ internal class DisplayToolbar(
         val navigationActionsWidth = navigationActions
             .asSequence()
             .mapNotNull { it.view }
-            .fold(0) { usedWidth, view ->
-                val viewLeft = usedWidth
+            .fold(0) { _, view ->
```
NotWoods: This loop looks like it's supposed to sum together the width of all the actions, so the `usedWidth` argument shouldn't be discarded.

sblatz:
I was able to update most of the tests to pass, but I'm struggling with getting these tests working:

* url box position is enclosing icon, url and page actions
* url box position is enclosing icon, url and page actions

There's a mismatch in the widths 😕

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
sblatz: Ah, my bad. Fixed!

sblatz:
@pocmo let me know if you want to pair on this at all :)

pocmo:
Yeah, that may be a good idea. :)

A quick update: I did test this a bit today and it seems like the resizing does break some of the custom touch backgrounds (since the buttons do not seem to be a square anymore) used by Firefox for Echo Show.

pocmo:
Closing for now. Let's get back to that in one of the next sprints. There are a bunch of toolbar improvements that we could group together and address in one sprint.

Issue #1095, opened 2018-10-18
JEXL: Create sub types of AstNode


pocmo:
AstNode is currently quite complex because it needs to represent all possible nodes. It would be nicer if we could create separate classes for every type.

* https://github.com/mozilla-mobile/android-components/blob/master/components/lib/jexl/src/main/java/mozilla/components/lib/jexl/ast/AstNode.kt#L13
* https://github.com/mozilla-mobile/android-components/blob/master/components/lib/jexl/src/main/java/mozilla/components/lib/jexl/ast/AstType.kt

PR #3617, opened 2019-07-02
For #1095. Create subtypes for JEXL AstNode.

dector:
Sealed classes seems like perfect solution for this issue.

### Pull Request checklist


pocmo:
Did a quick drive-by. Looks great so far! Looking forward to removing the properties from the base class. :)

dector:
@pocmo Good. I'll continue refactoring then.

Unfortunately, simplifying `AstNode` will lead to many changes in parser as it's used heavily (e.g. as a partially-constructed object).

dector:
Closing for now to not to trigger CI on each push.

Will squash/reorganize and reopen when there will be something more to discuss.

dector:
Finished splitting to classes. Changed `toString()` implementation (I hope it's easier to follow now).

@pocmo it would be nice to hear you thoughts on it. The proposal is kinda-ready. I'll squash commits later (it's easier to follow/revert some changes with unsquashed ones).

pocmo:
@dector We are currently cleaning up the list of PRs and I wonder what the state of this one here is? :)

dector:
Hey. :)
I'm on vacation right now and will be back in few days. I'll continue
working on this soon.

pocmo:
Enjoy your vacation! :)

dector:
Thanks! I'm back now. Will rebase PR on master and prepare it for review till Tuesday I suppose.

dector:
Squashed all commits. But I really want to add more real-life tests for parser. Is it possible to find them somewhere? :smiley_cat:

pocmo:
Most tests and examples I took from the JavaScript and Python implementations/tests:
* https://github.com/mozilla/mozjexl
* https://github.com/mozilla/pyjexl

I guess you could also look at the coverage report and if there are uncovered sections write a test that uses exactly this part of the language. :)

dector:
I was wondering why expressions like `!false` can't be evaluated. But checked on master branch - seems it works "as designed". :smile: 

I think we are ready to go with this.

components/lib/jexl/src/main/java/mozilla/components/lib/jexl/parser/StateMachine.kt
```diff
@@ -141,7 +144,10 @@ internal val stateMachine: Map<State, StateMap> = mapOf(
     ),
     State.ARGUMENT_VALUE to StateMap(
         subHandler = { parser, node ->
-            parser.cursor!!.arguments.add(node!!)
+            val cursor = parser.cursor!!
+            if (cursor is Transformation) {
+                cursor.arguments.add(node!!)
+            }
```
dector: @pocmo Maybe I should throw exception here if node is incorrect? What do you think?

dector: Or even make cast that will throw exception.

components/lib/jexl/src/main/java/mozilla/components/lib/jexl/parser/Parser.kt
```diff
@@ -108,7 +113,11 @@ internal class Parser(
         if (cursor == null) {
             tree = node
         } else {
-            cursor!!.right = node
+            cursor?.let { cursor ->
+                if (cursor is BranchNode) {
```
dector: And cast here as well.

pocmo:
> I was wondering why expressions like !false can't be evaluated. But checked on master branch - seems it works "as designed". 😄

Can you file this? Sounds like this *should* work. May be worth checking against mozjexl.

components/lib/jexl/src/main/java/mozilla/components/lib/jexl/parser/StateMachine.kt
pocmo: Yes, that sounds good - especially if this keeps the tests green. :)

components/lib/jexl/src/main/java/mozilla/components/lib/jexl/parser/Parser.kt
pocmo: 👍

components/lib/jexl/src/main/java/mozilla/components/lib/jexl/parser/StateMachine.kt
dector: :heavy_check_mark:

components/lib/jexl/src/main/java/mozilla/components/lib/jexl/parser/Parser.kt
dector: :heavy_check_mark:

dector:
@pocmo.

There is no evaluator handler for `AstType.UNARY_EXPRESSION` now.
Will check with `mozjexl` and create follow up issue if result will be different.

dector:
Follow-up issue: #3996

dector:
@pocmo any updates on this?
Maybe I've got something wrong and something is expected from me? :)

PR #3628, opened 2019-07-02
Closes #3584: Api for supporting language switch.

### LINKED ISSUE #3584: Create an API for supporting language switching

Amejia481:
Initial support for  language switching. The Fennec and friends (Focus, FFTV, etc) implementation made intensive use of deprecated apis like [updateConfiguration](https://developer.android.com/reference/android/content/res/Resources#updateConfiguration(android.content.res.Configuration,%2520android.util.DisplayMetrics)) and others. The idea of this new API is to focus on bringing some none deprecated pieces from Fennec and combine them with this new api.

### Pull Request checklist


components/support/base/src/main/java/mozilla/components/support/base/android/locale/Extensions.kt
```diff
@@ -0,0 +1,22 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.support.base.android.locale
```
pocmo: Let's move this into a separate component! :)

Amejia481: 👍

Amejia481:
@pocmo  we still need to have the base app/activity classes because when we set a new language a [new Context is created](https://developer.android.com/reference/android/content/Context#createConfigurationContext(android.content.res.Configuration)) containing all the new localized resources and we have to provide the new context via [attachBaseContext](https://developer.android.com/reference/android/content/ContextWrapper.html#attachBaseContext(android.content.Context)).

Amejia481:
The consumer have to call `Activity#recreate()` after calling `setNewLocale`, but it will dependents on which approach the consumer is using for handling navigation, if it using the single activity, the user just have to call `Activity#recreate()` if not, it will be a bit tricky more below.

```kotlin
 /**
     * Change the system defined locale to the indicated in the [language] parameter.
     * This new [language] will be stored and will be the new current locale returned by [getCurrentLocale].
     *
     * After calling this function, to visualize the locale changes you have to make sure all your visible activities
     * get recreated. If your app is using the single activity approach, this will be trivial just call
     * [AppCompatActivity.recreate]. On the other hand, if you have multiple activity this could be tricky, one
     * alternative could be restarting your application process see https://github.com/JakeWharton/ProcessPhoenix
     * @return A new Context object for whose resources are adjusted to match the new [language].
     */
    fun setNewLocale(context: Context, language: String): Context {
        Storage.save(context, language)
        return updateResources(context)
    }

```

components/support/base/src/main/java/mozilla/components/support/base/android/locale/Extensions.kt
Amejia481: Moved to `support-locale`.

PR #3631, opened 2019-07-03
Closes #3630: Bump application-services to 0.31.3

### LINKED ISSUE #3630: Update application-services to v0.31.3 as part of android-components v0.56.6
The application-services team have identified a number of low-risk, high-value fixes that we'd like to include in an upcoming Fenix point-release, and have backported them to a new point-release of the v0.31 branch that is currently shipping in Fenix.  They are:

* [#1314](https://github.com/mozilla/application-services/pull/1314) - Ensure that Fenix registers its FxA device record with `{ type: "mobile" }` rather than `{ type: "desktop" }`. This will ensure that it shows up with the correct icon in the send-tab device list, FxA "devices and apps" view, etc.
  * Why uplift? The current version of Fenix is incorrectly writing `{ type: "desktop" }` into its device record on the server, and we don't have a plan for automatically correcting that data, so want to limit it to as few users as possible.

- [#1342](https://github.com/mozilla/application-services/pull/1342) - Ensure that bookmark sync doesn't fail if a bookmark or query is missing or has an invalid URL. This will allow users who have corrupt bookmarks data uploaded by other clients to still sync successfully on Fenix.
  * Why uplift? This seems to be a surprisingly common error and is fatal to bookmark syncing, so we want to unstick these users.

- [#1353](https://github.com/mozilla/application-services/pull/1353) - Update `smallvec` dependency to incorporate a fix for a potential use-after-free bug. We don't have any reason to believe it's exploitable, but [the fix was trivial](https://github.com/servo/rust-smallvec/compare/6510c426157ed8739a7757fa1cc16401a28d66a2...eea8cc489f1018ffa0c725c05e2e0ee153417b31) so it seems worth erring on the side of caution.
  * Why uplift? Because it's memory-unsafety and I can't guarantee that it won't lead to some undesirable behavior.

I've built and tested the `sample-sync` app with this new version and it seems to work successfully. I will also build and test with the Fenix release branch for completeness.

Please consider including this update in one of the upcoming android-components releases that will be cut to support a Fenix point-release.

rfk:
For your consideration, as described in https://github.com/mozilla-mobile/android-components/issues/3630

I targeted this PR at the `releases/0.56` branch but I don't really know if that's the right approach, please correct as appropriate. /cc @grigoryk.

### DIFF
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -25,7 +25,7 @@ object Versions {
     const val jna = "5.2.0"
     const val disklrucache = "2.0.2"
 
-    const val mozilla_appservices = "0.31.2"
+    const val mozilla_appservices = "0.31.3"
     const val servo = "0.0.1.20181017.aa95911"
 
     const val material = "1.0.0"


pocmo:
> I targeted this PR at the releases/0.56 branch but I don't really know if that's the right approach

Usually it's enough to just target `master` and we will take care of uplifting for a dot release. But since this PR already exist we can keep it and then cherry-pick into master. :)

Leaving it to @grigoryk or @csadilek to approve.

pocmo:
Since Fenix 1.0.1 will go out without an AC update, there won't be a 056.6 for now. Let's get this into `master` for now. :)

jonalmeida:
We've [tagged Fenix 1.0.1](https://github.com/mozilla-mobile/fenix/releases/tag/v1.0.1) so we might be able to merge this now.

rfk:
I've updated the PR to target master.

PR #3635, opened 2019-07-03

Pike:
State: mozilla-l10n/android-l10n@63a524065c58cffbc884fa383025773ca4bb9108

### DIFF
--- a/components/lib/crash/src/main/res/values-de/strings.xml
+++ b/components/lib/crash/src/main/res/values-de/strings.xml
@@ -11,4 +11,10 @@
 
     <!-- Label of the button closing the crash reporter dialog and restarting the app. -->
     <string name="mozac_lib_crash_dialog_button_restart">%1$s neu starten</string>
+
+    <!-- Name of the "notification channel" used for displaying a notification when the app crashed and we can't show a prompt. See https://developer.android.com/training/notify-user/channels -->
+    <string name="mozac_lib_crash_channel">Abstürze</string>
+
+    <string name="mozac_lib_crash_notification_action_report">Bericht</string>
+
 </resources>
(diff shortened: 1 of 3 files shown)


components/lib/crash/src/main/res/values-de/strings.xml
```diff
@@ -11,4 +11,10 @@
 
     <!-- Label of the button closing the crash reporter dialog and restarting the app. -->
     <string name="mozac_lib_crash_dialog_button_restart">%1$s neu starten</string>
+
+    <!-- Name of the "notification channel" used for displaying a notification when the app crashed and we can't show a prompt. See https://developer.android.com/training/notify-user/channels -->
+    <string name="mozac_lib_crash_channel">Abstürze</string>
+
+    <string name="mozac_lib_crash_notification_action_report">Bericht</string>
```
pocmo: Oh, this translations tells me I should add a better translation note. This is the label for a notification button that will "report" the crash.

pocmo: What's actually the best approach to deal with that? I can add the translation comment - but that won't notify the translator. :)

Pike: @Delphine , any suggestions?

I looked at https://transvision.mozfr.org/string/?entity=android_l10n/mozilla-mobile/android-components/components/lib/crash/src/main/res/values/strings.xml:mozac_lib_crash_notification_action_report&repo=android_l10n and it's not looking like the Chinese got this, either. But that's just according to google translate.

Delphine: @pocmo You can update the comment and once it's done, I'll contact the localizers directly so they are aware 🙂

pocmo: Alright! https://github.com/mozilla-mobile/android-components/issues/3643

pocmo: Added the comment:
https://github.com/mozilla-mobile/android-components/commit/0d401c670448af7571fd37ba4b33640a808ceff7

I'm going to merge this PR as-is since that string is not used yet. :)

PR #3646, opened 2019-07-03
1563177: TimingDistribution should use snake_case for its fields

mdboom:
This fixes the data format so it matches the schema, and also adds a unit test to check against the schema.

Cc: @badboy to flag that these changes should make it over to glean-core.

### Pull Request checklist

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
@@ -172,7 +172,7 @@ data class TimingDistributionData(
             // something is wrong and we should return null.
             val category = jsonObject.tryGetString("category").orEmpty()
             val name = jsonObject.tryGetString("name") ?: return null
-            val bucketCount = jsonObject.tryGetInt("bucketCount") ?: return null
+            val bucketCount = jsonObject.tryGetInt("bucket_count") ?: return null
             // If 'range' isn't present, JSONException is thrown
             val range = try {
                 val array = jsonObject.getJSONArray("range")
@@ -190,8 +190,13 @@ data class TimingDistributionData(
             } catch (e: org.json.JSONException) {
                 return null
             }
-            val rawHistogramType = jsonObject.tryGetInt("histogramType") ?: return null
-            val histogramType = HistogramType.values().getOrNull(rawHistogramType) ?: return null
+            val rawHistogramType = jsonObject.tryGetString("histogram_type") ?: return null
+            var histogramType: HistogramType
+            try {
+                histogramType = HistogramType.valueOf(rawHistogramType.capitalize())
+            } catch (e: IllegalArgumentException) {
+                return null
+            }
             // Attempt to parse the values map, if it fails then something is wrong and we need to
             // return null.
             val values = try {
@@ -206,8 +211,13 @@ data class TimingDistributionData(
                 return null
             }
             val sum = jsonObject.tryGetLong("sum") ?: return null
-            val timeUnit = TimeUnit.values()[jsonObject.tryGetInt("timeUnit")
-                ?: return null]
+            val rawTimeUnit = jsonObject.tryGetString("time_unit") ?: return null
+            var timeUnit: TimeUnit
+            try {
+                timeUnit = TimeUnit.valueOf(rawTimeUnit.capitalize())
+            } catch (e: IllegalArgumentException) {
+                return null
+            }
 
             return TimingDistributionData(
                 category = category,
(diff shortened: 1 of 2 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
```diff
@@ -190,8 +190,13 @@ data class TimingDistributionData(
             } catch (e: org.json.JSONException) {
                 return null
             }
-            val rawHistogramType = jsonObject.tryGetInt("histogramType") ?: return null
-            val histogramType = HistogramType.values().getOrNull(rawHistogramType) ?: return null
+            val rawHistogramType = jsonObject.tryGetString("histogram_type") ?: return null
+            var histogramType: HistogramType
+            try {
```
Dexterp37: nit: this could be more idiomatic by doing

`val histogramType = try {
 ...
} catch (...) {
  return null
}`

the catch here is that we should check if `return` here is returning from the scope (and assigning the value to the variable) or returning from the function. Maybe we should double-check this  anyway as I see this pattern already used for `range`, above.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
```diff
@@ -206,8 +211,13 @@ data class TimingDistributionData(
                 return null
             }
             val sum = jsonObject.tryGetLong("sum") ?: return null
-            val timeUnit = TimeUnit.values()[jsonObject.tryGetInt("timeUnit")
-                ?: return null]
+            val rawTimeUnit = jsonObject.tryGetString("time_unit") ?: return null
+            var timeUnit: TimeUnit
```
Dexterp37: As above, we could do without a `var` and assign the return value of the `try` block to `timeUnit`.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
```diff
@@ -262,12 +272,12 @@ data class TimingDistributionData(
         return JSONObject(mapOf(
             "category" to category,
             "name" to name,
-            "bucketCount" to bucketCount,
+            "bucket_count" to bucketCount,
             "range" to JSONArray(arrayOf(rangeMin, rangeMax)),
-            "histogramType" to histogramType.ordinal,
+            "histogram_type" to histogramType.toString().toLowerCase(),
```
Dexterp37: Great catch. Gah, I missed all these things in the review a while back. Shame on me.

components/service/glean/src/test/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngineTest.kt
```diff
@@ -128,6 +135,39 @@ class TimingDistributionsStorageEngineTest {
             snapshot["telemetry.test_timing_distribution"]?.toJsonObject().toString())
     }
 
+    @Test
+    fun `serializer should serialize timing distribution that matches schema`() {
+        val ping1 = PingType("store1", true)
+
+        val server = getMockWebServer()
```
Dexterp37: Instead of setting up the whole ping machinery here and generating a ping, what if we had a test in GleanTest.kt that adds every single metric type and checks the schema ? (can be done as a follow-up). Alternatively, what about we call in the `StorageEngineManager` and get the resulting payload here?

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
badboy: `return` always returns from the function scope.

components/service/glean/src/test/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngineTest.kt
```diff
@@ -128,6 +135,39 @@ class TimingDistributionsStorageEngineTest {
             snapshot["telemetry.test_timing_distribution"]?.toJsonObject().toString())
```
Dexterp37: Let's also change the case in [this test](https://github.com/mozilla-mobile/android-components/blob/b49cd72c5c27ad335fd1885a358071abc2c70214/components/service/glean/src/test/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngineTest.kt#L89-L105). It is supposed to fail, but let's make sure it fails for the right reasons :)

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
mdboom: Got it.  I'll definitely make the syntactic simplification.  Confirmed that return returns from the function, and thus bails on the whole thing if there's a syntax error in deserialization (which I think is what we want, since this is data that we generated ourselves, so if there's a problem, it would be pretty unlikely / serious).

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
mdboom: No worries.  We're "eventually consistent"

PR #3651, opened 2019-07-04
Closes #3621: sample browser: Add simple UI test that loads a website.

### LINKED ISSUE #3621: Add simple UI test for samples-browser
Just a sanity check test:

* Spin up http server
* Load website
* Make sure website is loaded

pocmo:
This introduces a very simple UI tests that launches the sample browser, loads a website from a local webserver and verifies that the web content is visible.

Once this is in I want to add it to the tests that run on Firebase and want to explore how we can use this to have this and more sanity checks to make sure we never release broken components (in addition to our unit tests).


samples/browser/src/androidTest/java/org/mozilla/samples/browser/rules/WebserverRule.kt
```diff
@@ -0,0 +1,65 @@
...
+import okhttp3.mockwebserver.MockWebServer
+import okhttp3.mockwebserver.RecordedRequest
+import org.junit.rules.TestWatcher
+import org.junit.runner.Description
+import java.io.IOException
+
+/**
+ * A [TestWatcher] junit rule that will serve content from assets in the test package.
+ */
+class WebserverRule : TestWatcher() {
```
Amejia481: Could this rule be usable for other apps? If so we can put it in a shared component.

pocmo: Ha, yes, this could live in `support-android-test`. As part of working on this I thought we could come up with a instrumentation component that makes it nice to test apps that use components. For example code that does things like "Type something into the toolbar" is basically the same across all apps that use our toolbar - so we could provide a nice "top level" test API for that.

Amejia481: 💯 !

PR #3659, opened 2019-07-04
Closes #3527: browser-state: Add parent tab functionality

### LINKED ISSUE #3527: browser-state: Add "parent" tab functionality
For `browser-session` parity.

csadilek:
Port the parent session/tab handling functionality to browser-state.

### DIFF
--- a/components/browser/state/build.gradle
+++ b/components/browser/state/build.gradle
@@ -28,14 +28,15 @@ dependencies {
     implementation project(':support-utils')
     implementation project(':support-ktx')
 
-    implementation Dependencies.kotlin_stdlib
-    implementation Dependencies.kotlin_coroutines
     implementation Dependencies.androidx_browser
+    implementation Dependencies.kotlin_coroutines
+    implementation Dependencies.kotlin_stdlib
 
     testImplementation project(':support-test')
+    testImplementation Dependencies.androidx_test_junit
     testImplementation Dependencies.testing_junit
-    testImplementation Dependencies.testing_robolectric
     testImplementation Dependencies.testing_mockito
+    testImplementation Dependencies.testing_robolectric
 }
 
 apply from: '../../../publish.gradle'
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
@@ -16,27 +16,40 @@ import mozilla.components.lib.state.Action
  * [Action] implementation related to [BrowserState].
  */
 sealed class BrowserAction : Action
+
 /**
  * [BrowserAction] implementations related to updating the list of [TabSessionState] inside [BrowserState].
  */
 sealed class TabListAction : BrowserAction() {
     /**
      * Adds a new [TabSessionState] to the list.
+     *
+     * @property tab the [TabSessionState] to add
+     * @property select whether or not to the tab should be selected.
      */
     data class AddTabAction(val tab: TabSessionState, val select: Boolean = false) : TabListAction()
 
     /**
      * Marks the [TabSessionState] with the given [tabId] as selected tab.
+     *
+     * @property tabId the ID of the tab to select.
      */
     data class SelectTabAction(val tabId: String) : TabListAction()
 
     /**
      * Removes the [TabSessionState] with the given [tabId] from the list of sessions.
+     *
+     * @property tabId the ID of the tab to remove.
+     * @property selectParentIfExists whether or not a parent tab should be
+     * selected if one exists, defaults to true.
      */
-    data class RemoveTabAction(val tabId: String) : TabListAction()
+    data class RemoveTabAction(val tabId: String, val selectParentIfExists: Boolean = true) : TabListAction()
 
     /**
      * Restores state from a (partial) previous state.
+     *
+     * @property tabs the [TabSessionState]s to restore.
+     * @property selectedTabId the ID of the tab to select.
      */
     data class RestoreAction(val tabs: List<TabSessionState>, val selectedTabId: String? = null) : TabListAction()
 
@@ -62,11 +75,15 @@ sealed class TabListAction : BrowserAction() {
 sealed class CustomTabListAction : BrowserAction() {
     /**
      * Adds a new [CustomTabSessionState] to [BrowserState.customTabs].
+     *
+     * @property tab the [CustomTabSessionState] to add.
      */
     data class AddCustomTabAction(val tab: CustomTabSessionState) : CustomTabListAction()
 
     /**
      * Removes an existing [CustomTabSessionState] to [BrowserState.customTabs].
+     *
+     * @property tabId the ID of the custom tab to remove.
      */
     data class RemoveCustomTabAction(val tabId: String) : CustomTabListAction()
 
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TabListReducer.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TabListReducer.kt
@@ -18,8 +18,22 @@ internal object TabListReducer {
     fun reduce(state: BrowserState, action: TabListAction): BrowserState {
         return when (action) {
             is TabListAction.AddTabAction -> {
+
+                val updatedTabList = if (action.tab.parentId != null) {
+                    val parentIndex = state.tabs.indexOfFirst { it.id == action.tab.parentId }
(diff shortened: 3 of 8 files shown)


components/browser/state/build.gradle
```diff
@@ -36,6 +36,7 @@ dependencies {
     testImplementation Dependencies.testing_junit
     testImplementation Dependencies.testing_robolectric
     testImplementation Dependencies.testing_mockito
+    testImplementation Dependencies.androidx_test_junit
```
NotWoods: nit: Move this to line 36 so it's in alphabetical order

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TabListReducer.kt
```diff
@@ -18,8 +18,23 @@ internal object TabListReducer {
     fun reduce(state: BrowserState, action: TabListAction): BrowserState {
         return when (action) {
             is TabListAction.AddTabAction -> {
+
+                val updatedTabList = if (action.parent != null) {
```
NotWoods: Since the logic for adding a tab from a parent is fairly different, should it be its own action?

components/browser/state/build.gradle
csadilek: OK, re-ordered.

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TabListReducer.kt
csadilek: I am not sure. They are quite similar, minus the one if statement and I like to keep the API symmetric AddTab/RemoveTab vs. AddTab/AddChildTab/RemoveTab. When removing, the caller shouldn't need to worry about whether a tab is a parent/child tab or not.

csadilek: Happy to change though if we feel it's getting too complicated? Can also be refactored later though...

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -16,27 +16,46 @@ import mozilla.components.lib.state.Action
...
+     * @property select whether or not to the tab should be selected.
+     * @property parent the parent [TabSessionState] of the provided tab. The
+     * tab will be added next to its parent, if provided.
      */
-    data class AddTabAction(val tab: TabSessionState, val select: Boolean = false) : TabListAction()
+    data class AddTabAction(
+        val tab: TabSessionState,
+        val select: Boolean = false,
+        val parent: TabSessionState? = null
+    ) : TabListAction()
```
pocmo: I wonder if we even need to add the `parent` parameter here - since the `TabSessionState` can already contain a parent id. We could just use that (if it is set) and make sure that this is a valid parent - what do you think?

The higher-level use case for adding a tab would hide most of that anyways, I assume.

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -16,27 +16,46 @@ import mozilla.components.lib.state.Action
...
 
     /**
      * Removes the [TabSessionState] with the given [tabId] from the list of sessions.
+     *
+     * @property tabId the ID of the tab to remove.
+     * @property selectParentIfExists whether or not a parent tab should be
+     * selected if one exists.
      */
-    data class RemoveTabAction(val tabId: String) : TabListAction()
+    data class RemoveTabAction(val tabId: String, val selectParentIfExists: Boolean = false) : TabListAction()
```
pocmo: I wonder why we decided to use `false` as default in `SessionManager` (maybe to not change the default behavior?). It seems like `true` is the default you'd expect in a browser.

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TabListReducer.kt
```diff
@@ -18,8 +18,23 @@ internal object TabListReducer {
...
+                    val parentIndex = state.tabs.indexOfFirst { it.id == action.parent.id }
+                    if (parentIndex == -1) {
+                        throw IllegalArgumentException("The parent does not exist")
+                    }
+
+                    // Add the child tab next to its parent
+                    val child = action.tab.copy(parentId = action.parent.id)
+                    val childIndex = parentIndex + 1
+                    state.tabs.subList(0, childIndex) + child + state.tabs.subList(childIndex, state.tabs.size)
```
pocmo: nit: Another option may be to use `toMutableList()` and `add()` with an index?

components/browser/state/src/main/java/mozilla/components/browser/state/state/TabSessionState.kt
```diff
@@ -11,20 +11,24 @@ import java.util.UUID
  *
  * @property id the ID of this tab and session.
  * @property content the [ContentState] of this tab.
+ * @property parentId the parent ID of this tab.
```
pocmo: nit: this reads a bit mechanical and doesn't really say what a parent id is. Maybe something like "The ID of the parent tab or null if this tab has no parent"?

components/browser/state/src/test/java/mozilla/components/browser/state/store/BrowserStoreExceptionTest.kt
```diff
@@ -0,0 +1,40 @@
...
+
+            store.dispatch(TabListAction.AddTabAction(child, parent = parent)).joinBlocking()
+
+            // Wait for the main looper to process the re-thrown exception.
+            ShadowLooper.idleMainLooper()
+        } catch (e: IllegalStateException) {
+            val cause = e.cause
+            if (cause != null) {
+                throw cause
+            }
```
pocmo: This part makes me wonder if we should create a `StoreException` just to make sure we are catching the right thing in our tests. Related: I used `RuntimException` before and my tests passed. But it turned out it was a RuntimeException from the tests runner (I wasn't using Robolectric and therefore there was no main looper).

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
csadilek: Yeah, makes sense. I think we wanted to preserve the default behaviour at the time but we don't need to do this now. Updated.

components/browser/state/src/main/java/mozilla/components/browser/state/state/TabSessionState.kt
csadilek: Ah yes. I've added more details here now i.e. what is a parent tab. Good catch. :)

components/browser/state/src/test/java/mozilla/components/browser/state/store/BrowserStoreExceptionTest.kt
csadilek: OK, let's chat our lib-state to store to throw a `StoreException`. I will amend a commit here with that change.

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TabListReducer.kt
csadilek: True, but I'd prefer if we sticked to immutable types in the reducer. A local mutable list wouldn't be a problem, of course, so it's mostly readability. Personally, I find it easier to read to stick to immutable types when we're copying state everywhere, but don't feel strongly :).

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
csadilek: Interesting. I was wondering about that too and thought it will make things easier for the callers. However, I just looked through the existing call sites and it makes not difference at all. I'll remove it from the action and just keep the parentId in the state.

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TabListReducer.kt
csadilek: Updated.

components/browser/state/src/test/java/mozilla/components/browser/state/store/BrowserStoreExceptionTest.kt
csadilek: Updated.

csadilek:
OK PR is ready for another review :).

PR #3661, opened 2019-07-05
Closes #3567: BrowserStore/SessionManager: Make result of "nearby" tab selection match.

### LINKED ISSUE #3567: BrowserStore/SessionManager: Make result of "nearby" tab selection match

pocmo:
This is a bit crazy. But `browser-session` is going away eventually....

### Pull Request checklist

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/LegacySessionManager.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/LegacySessionManager.kt
@@ -12,6 +12,7 @@ import mozilla.components.concept.engine.EngineSessionState
 import mozilla.components.support.base.observer.Observable
 import mozilla.components.support.base.observer.ObserverRegistry
 import kotlin.math.max
+import kotlin.math.min
 
 /**
  * This class provides access to a centralized registry of all active sessions.
@@ -398,20 +399,49 @@ class LegacySessionManager(
         return findNearbySession(index) { !it.isCustomTabSession() }
     }
 
+    /**
+     * Tries to find a [Session] that is near [index] and satisfies the [predicate].
+     *
+     * Its implementation is synchronized with the behavior in `TabListReducer` of the `browser-state` component.
+     */
+    @Suppress("LongMethod") // Yes, this method is pretty long. I hope we can delete it soon.
     private fun findNearbySession(index: Int, predicate: (Session) -> Boolean): Int {
-        val maxSteps = max(values.lastIndex - index, index)
+        // Okay, this is a bit stupid and complex. This implementation intends to implement the same behavior we use in
+        // BrowserStore - which is operating on a list without custom tabs. Since LegacySessionManager uses a list with
+        // custom tabs mixed in, we need to shift the index around a bit.
 
-        if (maxSteps <= 0) {
-            return NO_SELECTION
+        // First we need to determine the number of custom tabs that are *before* the index we removed from. This is
+        // later needed to calculate the index we would have removed from in a list without custom tabs.
+        var numberOfCustomTabsBeforeRemovedIndex = 0
+        for (i in 0..min(index - 1, values.lastIndex)) {
+            if (values[i].isCustomTabSession()) {
+                numberOfCustomTabsBeforeRemovedIndex++
+            }
         }
 
-        // Try sessions oscillating near the index.
-        for (steps in 1..maxSteps) {
-            listOf(index - steps, index + steps).forEach { current ->
+        // Now calculate the index and list without custom tabs included.
+        val indexWithoutCustomTabs = index - numberOfCustomTabsBeforeRemovedIndex
+        val sessionsWithoutCustomTabs = sessions
+
+        // Code run before this one would have re-used the selected index if it is still a valid index that satisfies
+        // the predicate - however that list contained custom tabs. Now without custom tabs we may be able to re-use
+        // the exiting index. Let's check if we can:
+        if (indexWithoutCustomTabs >= 0 &&
+            indexWithoutCustomTabs <= sessionsWithoutCustomTabs.lastIndex &&
+            predicate(sessionsWithoutCustomTabs[indexWithoutCustomTabs])) {
+            return values.indexOf(sessionsWithoutCustomTabs[indexWithoutCustomTabs])
+        }
+
+        // Finally try sessions (without custom tabs) oscillating near the index.
+        val range = 1..max(sessionsWithoutCustomTabs.lastIndex - indexWithoutCustomTabs, indexWithoutCustomTabs)
+        for (steps in range) {
+            listOf(indexWithoutCustomTabs - steps, indexWithoutCustomTabs + steps).forEach { current ->
                 if (current >= 0 &&
-                    current <= values.lastIndex &&
-                    predicate(values[current])) {
-                    return current
+                    current <= sessionsWithoutCustomTabs.lastIndex &&
+                    predicate(sessionsWithoutCustomTabs[current])) {
+                    // Now that we have a match we need to re-calculate the index in the list that includes custom tabs.
+                    val session = sessionsWithoutCustomTabs[current]
+                    return values.indexOf(session)
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


components/browser/session/src/test/java/mozilla/components/browser/session/SessionManagerMigrationTest.kt
```diff
@@ -339,4 +339,116 @@ class SessionManagerMigrationTest {
...
+     * - Selects a random tab
+     * - Randomly removes tabs
+     * - Verifies that the selected tabs in BrowserStore and SessionManager match
+     *
+     * The `println` statements have been commented out to not spam the CI log. However they are helpful in case someone
+     * needs to actually debug a problem here.
+     */
+    @Test
+    fun `Remove session from randomized set will keep store and manager in sync`() {
+        repeat(10000) {
```
jhugman: 😻

components/browser/session/src/main/java/mozilla/components/browser/session/LegacySessionManager.kt
```diff
@@ -398,20 +399,44 @@ class LegacySessionManager(
...
+    @Suppress("LongMethod") // Yes, this method is pretty long. I hope we can delete it soon.
     private fun findNearbySession(index: Int, predicate: (Session) -> Boolean): Int {
-        val maxSteps = max(values.lastIndex - index, index)
+        // Okay, this is a bit stupid and complex. This implementation intends to implement the same behavior we use in
+        // BrowserStore - which is operating on a list without custom tabs. Since LegacySessionManager uses a list with
+        // custom tabs mixed in, we need to shift the index around a bit.
+
+        // First we need to determine the number of custom tabs that are *before* the index we removed from. This is
+        // later needed to calculate the index we would have removed from in a list without custom tabs.
+        var numberOfCustomTabsBeforeRemovedIndex = 0
```
csadilek: nit: could write as `sublist.filter.size` but not sure that's much easier to read.

components/browser/session/src/main/java/mozilla/components/browser/session/LegacySessionManager.kt
```diff
@@ -398,20 +399,44 @@ class LegacySessionManager(
         return findNearbySession(index) { !it.isCustomTabSession() }
     }
 
+    @Suppress("LongMethod") // Yes, this method is pretty long. I hope we can delete it soon.
     private fun findNearbySession(index: Int, predicate: (Session) -> Boolean): Int {
```
jhugman: This needs a doc comment.

components/browser/session/src/main/java/mozilla/components/browser/session/LegacySessionManager.kt
```diff
@@ -398,20 +399,44 @@ class LegacySessionManager(
...
+            listOf(indexWithoutCustomTabs - steps, indexWithoutCustomTabs + steps).forEach { current ->
                 if (current >= 0 &&
-                    current <= values.lastIndex &&
-                    predicate(values[current])) {
-                    return current
+                    current <= sessionsWithoutCustomTabs.lastIndex &&
+                    predicate(sessionsWithoutCustomTabs[current])) {
+                    // Now that we have a match we need to re-calculate the index in the list that includes custom tabs.
+                    val session = sessionsWithoutCustomTabs[current]
+                    return values.indexOf(session)
```
jhugman: I'm a bit confused how this re-write helps. Is `predicate` necessary?

This `predicate` is only specified twice: once where it's checking we're not in a custom tab, and the other to check that we're in `private` browsing or not.

jhugman: I wonder if this code could be clearer if it were moar functional.

components/browser/session/src/test/java/mozilla/components/browser/session/SessionManagerMigrationTest.kt
```diff
@@ -339,4 +339,116 @@ class SessionManagerMigrationTest {
...
+        manager.add(session6)
+        manager.add(session7)
+        manager.add(session8)
+        manager.add(session9)
+
+        manager.select(session4)
+        assertEquals(session4, manager.selectedSession)
+        assertEquals("session4", store.state.selectedTabId)
+
+        println("Remove session 4")
```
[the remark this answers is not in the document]
pocmo: Oh, I can remove those now. That was only for debugging.

components/browser/session/src/main/java/mozilla/components/browser/session/LegacySessionManager.kt
pocmo: Tried `sublist.filter.size` and I like it! Also with that this can be `val` instead of `var`.

It doesn't like an invalid range though (e.g. `fromIndex(0) > toIndex(-1)`)

However with `take(max(0, ....))` that works. It's not easier to read though. :D

pocmo: Oh, mh, now I have a test failure. I'll change it back. Don't want to look at that code any longer... :)

components/browser/session/src/main/java/mozilla/components/browser/session/LegacySessionManager.kt
pocmo: The goal of the rewrite was to have the same behavior here as in `TabListReducer` of `browser-state` so that the new test (see #3567) passes.

I agree that this could be rewritten in a way more sane way. But I tried to write as little code as possible since this is the old code that is going away and I didn't want to refactor all that. So yeah, this legacy code is now a bit stupid and does definitely not optimized.

The same code in `TabListReducer` is thankfully much simpler:
https://github.com/mozilla-mobile/android-components/blob/master/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TabListReducer.kt#L171

PR #3665, opened 2019-07-05
Issue #3481 - Use setPrimaryClip 

NotWoods:
Android Q changes setPrimaryClip to be NotNull, but getPrimaryClip remains Nullable.
Kotlin no longer exposes a setter as a result.

### Pull Request checklist

### DIFF
--- a/components/concept/engine/src/main/java/mozilla/components/concept/engine/manifest/WebAppManifestParser.kt
+++ b/components/concept/engine/src/main/java/mozilla/components/concept/engine/manifest/WebAppManifestParser.kt
@@ -125,7 +125,7 @@ private fun parseIcons(json: JSONObject): List<WebAppManifest.Icon> {
             WebAppManifest.Icon(
                 src = obj.getString("src"),
                 sizes = parseIconSizes(obj),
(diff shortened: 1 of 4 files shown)


components/concept/engine/src/main/java/mozilla/components/concept/engine/manifest/WebAppManifestParser.kt
```diff
@@ -125,7 +125,7 @@ private fun parseIcons(json: JSONObject): List<WebAppManifest.Icon> {
             WebAppManifest.Icon(
                 src = obj.getString("src"),
                 sizes = parseIconSizes(obj),
-                type = obj.optString("type", null),
+                type = obj.tryGetString("type"),
```
jhugman: These `tryGetString` changes belong in an earlier PR no?

Land them here if you have already landed an early transformation.

NotWoods: I already landed `tryGetString` previously in #3453, I just missed a few (or perhaps they slipped in).

PR #3667, opened 2019-07-05
Closes #3481 - Increase compileSdkVersion to 29

### LINKED ISSUE #3481: Meta: compileSdkVersion 29 (Android Q)

NotWoods:


buildSrc/src/main/java/Config.kt
```diff
@@ -11,7 +11,7 @@ class Config(val componentsVersion: String) {
         const val componentsGroupId = "org.mozilla.components"
 
         // Synchronized build configuration for all modules
-        const val compileSdkVersion = 28
+        const val compileSdkVersion = 29
```
pocmo: We should mention this in the changelog! :)

NotWoods:
Blocked by #3670 and #3671

components/lib/crash/src/main/java/mozilla/components/lib/crash/Crash.kt
```diff
@@ -87,12 +87,12 @@ sealed class Crash {
 
     companion object {
         fun fromIntent(intent: Intent): Crash {
-            val bundle = intent.getBundleExtra(INTENT_CRASH)
+            val bundle: Bundle? = intent.getBundleExtra(INTENT_CRASH)
```
pocmo: Let's use `!!` up here because if it's null continuing doesn't make sense.

components/service/experiments/src/main/java/mozilla/components/service/experiments/debug/ExperimentsDebugActivity.kt
```diff
@@ -78,8 +78,8 @@ class ExperimentsDebugActivity : Activity() {
             }
         } else if (intent.hasExtra(OVERRIDE_EXPERIMENT_EXTRA_KEY) &&
             intent.hasExtra(OVERRIDE_BRANCH_EXTRA_KEY)) {
-            val experiment = intent.getStringExtra(OVERRIDE_EXPERIMENT_EXTRA_KEY)
-            val branch = intent.getStringExtra(OVERRIDE_BRANCH_EXTRA_KEY)
+            val experiment = intent.getStringExtra(OVERRIDE_EXPERIMENT_EXTRA_KEY).orEmpty()
+            val branch = intent.getStringExtra(OVERRIDE_BRANCH_EXTRA_KEY).orEmpty()
             logger.info("Override to the following experiment/branch:$experiment/$branch")
             Experiments.setOverride(applicationContext, experiment, true, branch)
```
pocmo: If any of those is empty we should just bail and not set any overrides.

Dexterp37:
@travis79 would you mind checking the changes to the Experiments, here?

pocmo:
@NotWoods How about merging this tomorrow after the 3.0.0 release? This way we have one more cycle of SNAPSHOTS to test that this has no side effects we didn't think of.

NotWoods:
@pocmo Sounds good, I'm happy to hold off on this change.

PR #3670, opened 2019-07-05
Issue #3481 - Support compling to Q in Glean

NotWoods:


Dexterp37:
@NotWoods , before merging this, a question: does bumping the `compileSdkVersion` to 29 suddenly turn on a "warning as errors" flag? I tried to check if all your changes are required, locally, but I only get errors in `GenericScalarEngine` and `TimespanStoregeEngine`. The rest are warnings that *still* make the build fail. Is that the reason why you addressed them in this PR?

pocmo:
@Dexterp37 AC has enabled `warnings as errors` already for a while. Those are probably all new warnings and errors.
https://github.com/mozilla-mobile/android-components/blob/master/build.gradle#L50

Dexterp37:
> 
> 
> @Dexterp37 AC has enabled `warnings as errors` already for a while. Those are probably all new warnings and errors.
> https://github.com/mozilla-mobile/android-components/blob/master/build.gradle#L50

Oh, thanks, this makes sense then :) Thanks!

Issue #3642, opened 2019-07-03
Check web app manifest for tab icons


NotWoods:
We current prioritize `apple-touch-icon` as one of the best icon sources. `apple-touch-icon` is traditionally used as a home screen icon when a website is added to the home screen in iOS. The `icons` field in the Web App Manifest serves to provide a standard way to declare icons for the home screen just like `apple-touch-icon`. 

We should read the icons field in the web app manifest when loading icons for the session and prioritize them above apple icons, since they represent the standard.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-2988)

csadilek:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1793163

Change performed by the [Move to Bugzilla add-on](https://addons.mozilla.org/en-US/firefox/addon/move-to-bugzilla/).

PR #3674, opened 2019-07-05
Closes #3207 - Generate PWA launcher icons

### LINKED ISSUE #3207: PWA: Generate launcher icons from WebAppManifest
We'll need to create icons for Progressive Web Apps to display on the launcher.

There are a few different icon types that need to be supported:

## Adaptive icons from icon masks
[Icon masks](https://www.w3.org/TR/appmanifest/#icon-masks) were recently added to spec and let website owners define an image with padding so the platform can crop it. Adaptive icons have a few possible shapes depending on vendor/user settings.

Icon masks define only a single layer so we'll probably need to put the entire icon on the background layer and leave the foreground layer empty.

The icon masks in spec don't quite span the entire adaptive icon area. Spec suggests that the extra space can be filled with a black border, but I'd like to fill the area using colours from the edges of the mask image.

These icons have a size of 108dp but a safe-zone of 72dp or 66dp. The raw icon should be 90dp or 82dp (80% safe zone) and then the extra space will be filled in.

- [**Examples**](https://www.w3.org/TR/appmanifest/#examples-of-masks)

## Adaptive icons from standard icons
Some web apps don't specify a maskable icon so we'll need to fallback to taking the icon and putting some background behind it. AFAIK Chrome puts a white background behind these icons. However, I think it would look better to use the theme colour as the background, similar to how splash screens are generated.

## Legacy icons for pre-Oreo devices
On devices without adaptive icon support, just use the standard icon defined in the manifest.

These icons are displayed at 48dp.

## No icon defined/no manifest defined
Fall back to some generic icon or favicon.

NotWoods:
The [issue](https://github.com/mozilla-mobile/android-components/issues/3207) describes a lot about the different icons.

- On Oreo+, use a maskable icon if possible (adjusted to correct size)
- On Oreo+, if the icon is not maskable, add lots of padding and use the icon color as a background
- On Nougat-, don't do anything to non-maskable icons.
- On Nougat-, maskable icons are automatically reformatted by the support library.

I'm currently leaning to moving the `AdaptiveIconProcessor` into feature-pwa and using the manifest theme color as the background instead.

### Pull Request checklist

### DIFF
--- a/components/browser/icons/src/main/java/mozilla/components/browser/icons/extension/WebAppManifest.kt
+++ b/components/browser/icons/src/main/java/mozilla/components/browser/icons/extension/WebAppManifest.kt
@@ -4,13 +4,21 @@
 
 package mozilla.components.browser.icons.extension
 
+import android.os.Build
+import android.os.Build.VERSION.SDK_INT
 import mozilla.components.browser.icons.IconRequest
+import mozilla.components.browser.icons.IconRequest.Resource.Type.MANIFEST_ICON
+import mozilla.components.browser.icons.IconRequest.Size.LAUNCHER
+import mozilla.components.browser.icons.IconRequest.Size.LAUNCHER_ADAPTIVE
 import mozilla.components.concept.engine.manifest.WebAppManifest
 import mozilla.components.concept.engine.manifest.WebAppManifest.Icon.Purpose
 
+/**
+ * Creates an [IconRequest] for retrieving the icon specified in the manifest.
+ */
 fun WebAppManifest.toIconRequest() = IconRequest(
     url = startUrl,
-    size = IconRequest.Size.LAUNCHER,
+    size = if (SDK_INT >= Build.VERSION_CODES.O) LAUNCHER_ADAPTIVE else LAUNCHER,
     resources = icons.mapNotNull { it.toIconResource() }
 )
 
@@ -19,7 +27,7 @@ private fun WebAppManifest.Icon.toIconResource(): IconRequest.Resource? {
 
     return IconRequest.Resource(
         url = src,
-        type = IconRequest.Resource.Type.MANIFEST_ICON,
+        type = MANIFEST_ICON,
         sizes = sizes,
         mimeType = type,
         maskable = Purpose.MASKABLE in purpose
--- a/components/browser/icons/src/main/java/mozilla/components/browser/icons/processor/ResizingProcessor.kt
+++ b/components/browser/icons/src/main/java/mozilla/components/browser/icons/processor/ResizingProcessor.kt
@@ -50,7 +50,6 @@ class ResizingProcessor(
             }
         }
 
-        originalBitmap.recycle()
         return icon.copy(bitmap = resizedBitmap ?: return null)
     }
 
--- a/components/feature/pwa/src/main/java/mozilla/components/feature/pwa/WebAppShortcutManager.kt
+++ b/components/feature/pwa/src/main/java/mozilla/components/feature/pwa/WebAppShortcutManager.kt
@@ -9,19 +9,48 @@ import android.content.Intent
 import android.content.pm.ShortcutManager
 import android.os.Build.VERSION.SDK_INT
 import android.os.Build.VERSION_CODES
+import androidx.annotation.VisibleForTesting
 import androidx.core.content.getSystemService
 import androidx.core.content.pm.ShortcutInfoCompat
 import androidx.core.content.pm.ShortcutManagerCompat
 import androidx.core.graphics.drawable.IconCompat
 import androidx.core.net.toUri
+import mozilla.components.browser.icons.BrowserIcons
+import mozilla.components.browser.icons.decoder.AndroidIconDecoder
+import mozilla.components.browser.icons.decoder.ICOIconDecoder
+import mozilla.components.browser.icons.extension.toIconRequest
+import mozilla.components.browser.icons.generator.DefaultIconGenerator
+import mozilla.components.browser.icons.loader.DataUriIconLoader
+import mozilla.components.browser.icons.loader.DiskIconLoader
+import mozilla.components.browser.icons.loader.HttpIconLoader
+import mozilla.components.browser.icons.loader.MemoryIconLoader
+import mozilla.components.browser.icons.preparer.DiskIconPreparer
+import mozilla.components.browser.icons.preparer.MemoryIconPreparer
+import mozilla.components.browser.icons.preparer.TippyTopIconPreparer
+import mozilla.components.browser.icons.processor.AdaptiveIconProcessor
+import mozilla.components.browser.icons.processor.ColorProcessor
+import mozilla.components.browser.icons.processor.DiskIconProcessor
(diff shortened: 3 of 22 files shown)


components/browser/icons/src/main/java/mozilla/components/browser/icons/extension/WebAppManifest.kt
```diff
@@ -4,13 +4,21 @@
...
 import mozilla.components.concept.engine.manifest.WebAppManifest
 import mozilla.components.concept.engine.manifest.WebAppManifest.Icon.Purpose
 
+/**
+ * Creates an [IconRequest] for retrieving the icon specified in the manifest.
+ */
 fun WebAppManifest.toIconRequest() = IconRequest(
     url = startUrl,
-    size = IconRequest.Size.LAUNCHER,
+    size = if (SDK_INT >= O) LAUNCHER_ADAPTIVE else LAUNCHER,
```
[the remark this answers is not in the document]
pocmo: Oh, lol, that's the letter `O`. I only noticed when later in the code this showed up again. I wonder if `Build.VERSION_CODES.O` is just easier to read/understand.

components/browser/icons/src/main/java/mozilla/components/browser/icons/utils/DiskCache.kt
```diff
@@ -33,7 +33,7 @@ private const val WEBP_QUALITY = 90
 /**
  * Caching bitmaps and resource URLs on disk.
  */
-internal class DiskCache :
+class DiskCache :
```
pocmo: If we need to open up the visibility here then we should rename this (and `MemoryCache`). For a public API that is way to generic. `IconDiskCache` and `IconMemoryCache` may be enough. In this case the whole interface shenanigans  (`DiskIconLoader.LoaderDiskCache, DiskIconPreparer.PreparerDiskCache, DiskIconProcessor.ProcessorDiskCache`) may be not needed anymore too and only complicate things.

components/browser/icons/src/main/java/mozilla/components/browser/icons/processor/ResizingProcessor.kt
```diff
@@ -50,7 +50,6 @@ class ResizingProcessor(
             }
         }
 
-        originalBitmap.recycle()
```
Amejia481: 🤔  This seems important, I wonder why are we removing this line?

components/browser/icons/src/main/java/mozilla/components/browser/icons/extension/WebAppManifest.kt
```diff
@@ -19,9 +27,9 @@ private fun WebAppManifest.Icon.toIconResource(): IconRequest.Resource? {
 
     return IconRequest.Resource(
         url = src,
-        type = IconRequest.Resource.Type.MANIFEST_ICON,
+        type = MANIFEST_ICON,
         sizes = sizes,
         mimeType = type,
-        maskable = Purpose.MASKABLE in purpose
+        maskable = Purpose.MASKABLE in purpose || "maskable" in src
```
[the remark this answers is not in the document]
NotWoods: Oops, I forgot to remove this. I'm waiting for [GeckoView](https://bugzilla.mozilla.org/show_bug.cgi?id=1559412) to not strip out the "purpose" flag and this was my workaround for the time being.

components/browser/icons/src/main/java/mozilla/components/browser/icons/processor/ResizingProcessor.kt
NotWoods: Recycling the bitmap breaks the cache ☹️

Amejia481: 👍

NotWoods:
The new `BrowserIcons` instance uses a separate cache so it shouldn't conflict with the main instance. The new PWA instance uses the new processors I've written (resizing, palette, etc.) so it's not easy to share them even if the adaptive icon processor is adjusted.

pocmo:
> The new BrowserIcons instance uses a separate cache so it shouldn't conflict with the main instance.

Isn't that going to be problematic for #3642?

I wonder if for installed apps we wouldn't want a cache but an actual storage (that doesn't evict; also see #2046).

NotWoods:
Using a separate cache for PWAs won't affect code that just loads the icons. We would just sometimes have the same icon cached in two places.

Storing the icons might make more sense for PWAs.

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/WebAppShortcutManager.kt
```diff
@@ -9,17 +9,42 @@ import android.content.Intent
...
 import mozilla.components.concept.engine.manifest.WebAppManifest
+import mozilla.components.concept.fetch.Client
 import mozilla.components.feature.pwa.ext.installableManifest
 
+private val pwaIconMemoryCache = IconMemoryCache()
+private val pwaIconDiskCache = IconDiskCache()
+
 class WebAppShortcutManager(
     private val storage: ManifestStorage,
+    private val icons: BrowserIcons,
```
pocmo: I think we should construct this internally and only ask for context/client. Otherwise there's a high chance that you get a wrongly configured `BrowserIcons` instance - even though you want exactly what comes out of `webAppIcons()`.

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/WebAppShortcutManager.kt
```diff
@@ -119,3 +156,40 @@ class WebAppShortcutManager(
...
     }
 }
+
+/**
+ * Creates custom version of [BrowserIcons] for loading web app icons.
+ *
+ * This version has its own cache to avoid affecting tab icons.
+ */
+@Suppress("LongMethod")
+fun webAppIcons(
```
pocmo: I'd prefer if we can make this internal to avoid apps using this to construct multiple `BrowserIcons` instances.

I'm still not convinced that we should have multiple, differently configured `BrowserIcons` instances. But for now that may be a good compromise.

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/WebAppShortcutManager.kt
```diff
@@ -119,3 +156,40 @@ class WebAppShortcutManager(
...
+    decoders = listOf(
+        AndroidIconDecoder(ignoreSize = true),
+        ICOIconDecoder()
+    ),
+    processors = listOf(
+        MemoryIconProcessor(pwaIconMemoryCache),
+        ResizingProcessor(),
+        DiskIconProcessor(pwaIconDiskCache),
+        ColorProcessor(),
+        AdaptiveIconProcessor()
```
pocmo: I think in a follow-up we should look at the storage of the icons - I feel like for installed apps we want to explicitly store the icon and not rely on a cache.

NotWoods: Filed #3799

Issue #1340, opened 2018-11-13
Expose "session history" in Session


pocmo:
Reference browser user story: https://github.com/mozilla-mobile/reference-browser/issues/69

To be able to render the "history" of a single `Session` we will need to expose this list in the `Session` itself.

Unclear is where this data should come from.

* WebView has [copyBackForwardList()](https://developer.android.com/reference/android/webkit/WebView#copyBackForwardList()).
* GeckoView doesn't seem to have anything like that yet.

pocmo:
https://bugzilla.mozilla.org/show_bug.cgi?id=1509110

pocmo:
After discussing this with the GV team the consensus is that GV will provide APIs for:
* Getting the history list of a GeckoSession.
* Getting the current index (in the list) the session is at
* Navigating to a specific index (this is something we'd implement in a different issue)

pocmo:
https://github.com/mozilla-mobile/firefox-tv/issues/1558

cpeterso:
GV's History Stack API has landed in mozilla-central and should be in tomorrow's GV 68 Nightly build.

https://bugzilla.mozilla.org/show_bug.cgi?id=1509110

bifleming:
@vesta0 this is still in the 'blocked' column on teh A/C sprint board... should it be moved? But is it post-MVP? we should take everything that is Post-MVP off of that board I think.

jonalmeida:
This would be needed for supporting the tab history for sync'd tabs.

jonalmeida:
This seems mostly for "read" access to the history stack, but we would need a way to add to it.

PR #3678, opened 2019-07-07
Issue mozilla-mobile/fenix#1048 - Add historyList to Session

NotWoods:
Adds `historyList` property to Session, so that we can later use it to render a list of pages when the back/forward buttons are held.

### Pull Request checklist

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
@@ -13,12 +13,13 @@ import mozilla.components.browser.session.ext.toSecurityInfoState
 import mozilla.components.browser.session.tab.CustomTabConfig
 import mozilla.components.browser.state.action.ContentAction.UpdateLoadingStateAction
 import mozilla.components.browser.state.action.ContentAction.UpdateProgressAction
-import mozilla.components.browser.state.action.ContentAction.UpdateSecurityInfo
 import mozilla.components.browser.state.action.ContentAction.UpdateSearchTermsAction
+import mozilla.components.browser.state.action.ContentAction.UpdateSecurityInfo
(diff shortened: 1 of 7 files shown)


components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
```diff
@@ -218,6 +220,13 @@ class Session(
         notifyObservers(old, new) { onNavigationStateChanged(this@Session, canGoBack, new) }
     }
 
+    /**
+     * Local session history, with the current page selected.
+     */
+    var historyList: List<HistoryItem> by Delegates.observable(emptyList()) { _, old, new ->
+        notifyObservers(old, new) { onHistoryStateChanged(this@Session, new) }
+    }
```
pocmo: naming: nit: I would suggest to only name this `history`. In addition to that I wonder if a rich state object (a bit like GeckoView's HistoryList) would be nicer to use: We can later attach things like canGoForward/canGoBack all to this `SessionHistory` object. That's at least how I would do it for `browser-state`.

components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
```diff
@@ -34,6 +34,7 @@ abstract class EngineSession(
         fun onProgress(progress: Int) = Unit
         fun onLoadingStateChange(loading: Boolean) = Unit
         fun onNavigationStateChange(canGoBack: Boolean? = null, canGoForward: Boolean? = null) = Unit
+        fun onHistoryStateChange(historyList: List<HistoryItem>) = Unit
```
pocmo: I would be interesting to prototype this API for `browser-engine-system` too:
* Fire TV wants to use it (See #1340)
* The API is a bit different and implementing both makes sure we do not build something that only works for GeckoView.

PR #3681, opened 2019-07-07
Closes #3578: Api for excluding the selectedSession on the SessionSuggestionProvider.

### LINKED ISSUE #3578: SessionSuggestionProvider shows current selected session, does nothing
See https://github.com/mozilla-mobile/fenix/issues/3695

Amejia481:
### Pull Request checklist

### DIFF
--- a/components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/SessionSuggestionProvider.kt
+++ b/components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/SessionSuggestionProvider.kt
@@ -5,6 +5,7 @@
 package mozilla.components.feature.awesomebar.provider
 
 import mozilla.components.browser.icons.BrowserIcons
+import mozilla.components.browser.session.Session
 import mozilla.components.browser.session.SessionManager
 import mozilla.components.concept.awesomebar.AwesomeBar
 import mozilla.components.feature.awesomebar.internal.loadLambda
@@ -18,7 +19,8 @@ import java.util.UUID
 class SessionSuggestionProvider(
     private val sessionManager: SessionManager,
     private val selectTabUseCase: TabsUseCases.SelectTabUseCase,
-    private val icons: BrowserIcons? = null
+    private val icons: BrowserIcons? = null,
+    private val excludeSelectedSession: Boolean = false
 ) : AwesomeBar.SuggestionProvider {
     override val id: String = UUID.randomUUID().toString()
 
@@ -30,8 +32,7 @@ class SessionSuggestionProvider(
         val suggestions = mutableListOf<AwesomeBar.Suggestion>()
 
         sessionManager.sessions.forEach { session ->
-            if ((session.url.contains(text, ignoreCase = true) ||
-                    session.title.contains(text, ignoreCase = true)) && !session.private
+            if (session.contains(text) && !session.private && shouldIncludeSelectedSession(session)
             ) {
                 suggestions.add(
                     AwesomeBar.Suggestion(
@@ -48,4 +49,17 @@ class SessionSuggestionProvider(
 
         return suggestions
     }
+
+    private fun Session.contains(text: String) =
+        (url.contains(text, ignoreCase = true) || title.contains(text, ignoreCase = true))
+
+    private fun shouldIncludeSelectedSession(session: Session): Boolean {
+        return if (excludeSelectedSession) {
+            !isSelectedSession(session)
+        } else {
+            true
+        }
+    }
+
+    private fun isSelectedSession(session: Session) = sessionManager.selectedSession?.equals(session) ?: false
 }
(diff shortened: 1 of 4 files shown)


components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/SessionSuggestionProvider.kt
```diff
@@ -48,4 +49,17 @@ class SessionSuggestionProvider(
 
         return suggestions
     }
+
+    private fun Session.contains(text: String) =
+        (url.contains(text, ignoreCase = true) || title.contains(text, ignoreCase = true))
+
+    private fun shouldIncludeSelectedSession(session: Session): Boolean {
+        return if (excludeSelectedSession) {
+            !(isSelectedSession(session) ?: false)
```
pocmo: nit: this is pretty hard to read. `isSelectedSession` probably should just return a non-null boolean.

Amejia481: Updated 👍

docs/changelog.md
```diff
@@ -12,6 +12,12 @@ permalink: /changelog/
 * [Gecko](https://github.com/mozilla-mobile/android-components/blob/master/buildSrc/src/main/java/Gecko.kt)
 * [Configuration](https://github.com/mozilla-mobile/android-components/blob/master/buildSrc/src/main/java/Config.kt)
 
+* **feature-tab**
+  * ⚠️ **This is a breaking change**: Now `TabsUseCases.SelectTabUseCase` is an interface, if you want to use is previous behavior you could keep using  `TabsUseCases.selectTab` or us `TabsUseCases.DefaultSelectTabUseCase`
+
+* **feature-awesomebar**
+  * ⚠️ **This is a breaking change**: `SessionSuggestionProvider` now have a new parameter `excludeSelectedSession`, to ignore the selected session on the suggestions.
```
csadilek: nit: This shouldn't be a breaking change as `excludeSelectedSession` defaults to false, which is the same as before?

Amejia481: Thanks for the catch, updated!

ekager:
I may be confused but from my understanding this still won't fix Fenix's problem. Clicking the currently selected session will never do anything in Fenix because this lives in SearchFragment. I think we need a custom use case here to be able to hook into this onSuggestionClicked action and route the user to BrowserFragment.

Amejia481:
> I may be confused but from my understanding this still won't fix Fenix's problem. Clicking the currently selected session will never do anything in Fenix because this lives in SearchFragment. I think we need a custom use case here to be able to hook into this onSuggestionClicked action and route the user to BrowserFragment.

Now that `TabsUseCases.SelectTabUseCase` is an interface you can provide your own implementation use case, using the same pattern as you are doing with other suggestion providers like [loadUrlUseCase](https://github.com/mozilla-mobile/fenix/blob/867067537a7dccf3d5f17dfbf3730f076cbc4bc2/app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarUIView.kt#L72), [searchUseCase](https://github.com/mozilla-mobile/fenix/blob/867067537a7dccf3d5f17dfbf3730f076cbc4bc2/app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarUIView.kt#L78) and [shortcutSearchUseCase](https://github.com/mozilla-mobile/fenix/blob/867067537a7dccf3d5f17dfbf3730f076cbc4bc2/app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarUIView.kt#L84) and get hooked when the item is selected.

Something like: 

```kotlin
    private val selectTapUseCase = object : TabsUseCases.SelectTabUseCase {
        override fun invoke(session: Session) {
            actionEmitter.onNext(AwesomeBarAction.URLTapped(session.url))
        }
    }
```

and when you init the `sessionProvider`:

```kotlin

            sessionProvider =
                SessionSuggestionProvider(
                    components.core.sessionManager,
                    selectTapUseCase,
                    components.core.icons,
                    excludeSelectedSession = true
                )
```

Additionally with the new property `excludeSelectedSession`,  now you can control if the selected session should be included in the suggestion list.

For a full example take a look [here](https://github.com/Amejia481/fenix/commit/0182f30f61ba7e07b78b821dc4b96e03e588aa86). Please play with it locally, and let me know if there is any point of improvements :)

ekager:
Okay! I think this will work for us then! I think UX still needs to decide if/when you should see your current session, but it will be useful to have this available :)

docs/changelog.md
```diff
@@ -56,6 +56,12 @@ permalink: /changelog/
 * **feature-media**
   * Added `MediaNotificationFeature` - a feature implementation to show an ongoing notification (keeping the app process alive) while web content is playing media.
 
+* **feature-tab**
+  * ⚠️ **This is a breaking change**: Now `TabsUseCases.SelectTabUseCase` is an interface, if you want to use is previous behavior you could keep using  `TabsUseCases.selectTab` or use `TabsUseCases.DefaultSelectTabUseCase`
```
csadilek: nit: want to rely on its...

Amejia481: Thanks for the catch, updated!

PR #3683, opened 2019-07-08
Deprecated View.isGone, View.isInvisible

NotWoods:

### Pull Request checklist

### DIFF
--- a/components/support/ktx/src/main/java/mozilla/components/support/ktx/android/view/View.kt
+++ b/components/support/ktx/src/main/java/mozilla/components/support/ktx/android/view/View.kt
@@ -30,8 +30,8 @@ val View.isLTR: Boolean
 
 /**
  * Returns true if this view's visibility is set to View.VISIBLE.
- * @deprecated Use Android KTX instead.
  */
+@Deprecated("Use Android KTX instead", ReplaceWith("isVisible", "androidx.core.view.isVisible"))
 fun View.isVisible(): Boolean {
     return visibility == View.VISIBLE
 }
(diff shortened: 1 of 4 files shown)


components/support/ktx/src/main/java/mozilla/components/support/ktx/android/view/View.kt
```diff
@@ -38,13 +38,15 @@ fun View.isVisible(): Boolean {
...
  */
 fun View.isGone(): Boolean {
     return visibility == View.GONE
 }
 
 /**
  * Returns true if this view's visibility is set to View.INVISIBLE.
+ * @deprecated Use Android KTX instead.
  */
 fun View.isInvisible(): Boolean {
```
Amejia481: I think we could at ` @Deprecated` [annotation](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-deprecated/index.html) on this function.

components/support/ktx/src/main/java/mozilla/components/support/ktx/android/view/View.kt
```diff
@@ -38,13 +38,15 @@ fun View.isVisible(): Boolean {
 
 /**
  * Returns true if this view's visibility is set to View.GONE.
+ * @deprecated Use Android KTX instead.
  */
 fun View.isGone(): Boolean {
```
Amejia481: I think we could at ` @Deprecated` [annotation](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-deprecated/index.html) on this function.

NotWoods:
Using `@Deprecated` will break builds for fenix if it uses the functions, so I used a soft KDocs comment instead for now.

PR #3686, opened 2019-07-08
Ensure experiments record enrollment/unenrollment in Glean

travis79:
The enrollment call to `Glean.setExperimentActive` was already here, just adding the unenrollment call to `Glean.setExperimentInactive`.

### Pull Request checklist

### DIFF
--- a/components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsTest.kt
+++ b/components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsTest.kt
@@ -13,6 +13,7 @@ import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.work.testing.WorkManagerTestInitHelper
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.runBlocking
+import mozilla.components.service.glean.Glean
 import mozilla.components.support.test.any
 import mozilla.components.support.test.eq
 import mozilla.components.support.test.mock
@@ -79,6 +80,7 @@ class ExperimentsTest {
         `when`(prefsEditor.putString(anyString(), anyString())).thenReturn(prefsEditor)
         `when`(sharedPrefs.edit()).thenReturn(prefsEditor)
         `when`(sharedPrefs.getBoolean(anyString(), anyBoolean())).thenAnswer { invocation -> invocation.arguments[1] as Boolean }
+        `when`(prefsEditor.clear()).thenReturn(prefsEditor)
         `when`(context.getSharedPreferences(anyString(), anyInt())).thenReturn(sharedPrefs)
 
         val packageInfo: PackageInfo = mock()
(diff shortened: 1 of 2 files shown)


travis79:
> I see that enrollment is [already recorded](https://github.com/mozilla-mobile/android-components/blob/ced4ade265d8ae32d7e685f4b90019481a7c6f64/components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt#L183). Should this have test coverage using the experiment's test API? (maybe file a follow-up bug)

I think it should be here on this PR. I actually forgot we had test API functionality around the experiments or I would have tried to put it in already.  Let me push a commit with a couple of tests around this.

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsTest.kt
```diff
@@ -322,6 +324,23 @@ class ExperimentsTest {
         assertEquals("second-id", experiments.activeExperiment!!.experiment.id)
     }
 
+    @Test
+    fun `experiment enrollment recorded in Glean`() {
```
Dexterp37: Since we're doing this, I'd also add a test for making sure we correctly unenroll.

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsTest.kt
```diff
@@ -322,6 +324,23 @@ class ExperimentsTest {
...
+        resetExperiments()
+        experiments.initialize(mockContext, configuration)
+        experiments.experimentsResult = ExperimentsSnapshot(experimentsList, null)
+        experiments.findAndStartActiveExperiment()
```
Dexterp37: I'd add `assertFalse(Glean.testIsExperimentActive("second-id"))` before this, just to make sure we're in a consistent state.

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsTest.kt
travis79: This test actually does this, in the call to `onExperimentsUpdated` where the empty list is passed in, unenrollment occurs and then the test does the `assertFalse(Glean.testIsExperimentActive("second-id"))` to verify.  I admit that the name of the test was misleading so I renamed the test to better indicate that it tests both enrollment and unenrollment.

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsTest.kt
travis79: 👍

components/service/experiments/src/test/java/mozilla/components/service/experiments/ExperimentsTest.kt
Dexterp37: Ok, let's drop a comment before that line to call that out.

travis79: Already assumed you would ask so I commented the whole test to explain its steps :)

Dexterp37: Oh my, now you're scaring me ;) Great job!

PR #3694, opened 2019-07-08
Add glide-style icon loading function

NotWoods:
A lot of our code involves setting up coroutine scopes for the sole purpose of loading an icon from `BrowserIcons`, which already has its own scope. I've written a new helper that emulates the `Glide.load(url).into(view)` call which doesn't need a parent coroutine scope.

Loading is cancelled if the image view is detached, and loading for an image view is restarted if the function is called again with the same image view.

### Pull Request checklist

### DIFF
--- a/components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt
+++ b/components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt
@@ -6,11 +6,19 @@ package mozilla.components.browser.icons
 
 import android.content.Context
 import android.graphics.Bitmap
+import android.graphics.drawable.Drawable
+import android.widget.ImageView
+import androidx.annotation.MainThread
+import androidx.annotation.WorkerThread
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Deferred
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.asCoroutineDispatcher
 import kotlinx.coroutines.async
+import kotlinx.coroutines.launch
 import mozilla.components.browser.icons.decoder.AndroidIconDecoder
 import mozilla.components.browser.icons.decoder.ICOIconDecoder
 import mozilla.components.browser.icons.decoder.IconDecoder
@@ -30,6 +38,7 @@ import mozilla.components.browser.icons.preparer.TippyTopIconPreparer
 import mozilla.components.browser.icons.processor.DiskIconProcessor
 import mozilla.components.browser.icons.processor.IconProcessor
 import mozilla.components.browser.icons.processor.MemoryIconProcessor
+import mozilla.components.browser.icons.utils.CancelOnDetach
 import mozilla.components.browser.icons.utils.DiskCache
 import mozilla.components.browser.icons.utils.MemoryCache
 import mozilla.components.browser.session.SessionManager
@@ -37,6 +46,7 @@ import mozilla.components.browser.session.utils.AllSessionsObserver
 import mozilla.components.concept.engine.Engine
 import mozilla.components.concept.fetch.Client
 import mozilla.components.support.base.log.logger.Logger
+import java.lang.ref.WeakReference
 import java.util.concurrent.Executors
 
 private const val MAXIMUM_SCALE_FACTOR = 2.0f
@@ -91,6 +101,7 @@ class BrowserIcons(
         }
     }
 
+    @WorkerThread
     private fun loadIconInternal(initialRequest: IconRequest): Icon {
         val desiredSize = DesiredSize(
             targetSize = context.resources.getDimensionPixelSize(initialRequest.size.dimen),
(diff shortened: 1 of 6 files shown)


pocmo:
Nice. I recently wanted to have that in AC too. :D Will look through that PR asap.

components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt
```diff
@@ -138,6 +143,34 @@ class BrowserIcons(
...
+        view.setTag(R.id.mozac_browser_icons_tag_job, deferredIcon)
+        val onAttachStateChangeListener = CancelOnDetach(deferredIcon).also {
+            view.addOnAttachStateChangeListener(it)
+        }
+
+        try {
+            val icon = deferredIcon.await()
+            view.setImageBitmap(icon.bitmap)
+        } catch (e: CancellationException) {
+            view.setImageDrawable(null)
```
pocmo: I wonder if we really want to set the drawable to `null` here? 🤔

In the scenario of a RecyclerView that would be the job of the adapter. In other scenarios inside the app the app may actually set a placeholder image that we would temporarily clear here until a new icon is loaded.

components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt
```diff
@@ -138,6 +143,34 @@ class BrowserIcons(
                 Logger.error("Could not install browser-icons extension", throwable)
             })
     }
+
+    /**
+     * Loads an icon asynchronously using [BrowserIcons] and then displays it in the [ImageView].
+     * If the view is detached from the window before loading is completed, then loading is cancelled.
+     */
+    fun loadIntoView(view: ImageView, request: IconRequest) = scope.async(Dispatchers.Main) {
```
pocmo: Since we are using the scope of `BrowserIcons` here, I wonder whether we could temporarily leak the view (and then its context/activity) by still having the coroutine around. We cancel when the View gets detached, but the cancellation itself may not happen immediately...? 🤔

NotWoods: I'm not sure what the best way to proceed is here. 

After looking at the Kotlin docs, cancellation seems to be immediate as long as the function is in a suspended state. We could adjust `loadIcon` to check if the coroutine was cancelled.

We could store the view in a `WeakReference` and that might be enough to avoid memory leaks.  

I saw that Picasso implements LifecycleObserver, so you could optionally add it as an observer in the fragment and automatically cancel all jobs on destroy. Would that be a good pattern to follow?

components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt
NotWoods: I'll add parameters for a placeholder and failure drawable and default to null.

components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt
pocmo: > After looking at the Kotlin docs, cancellation seems to be immediate as long as the function is in a suspended state. We could adjust loadIcon to check if the coroutine was cancelled.

The biggest issue here is probably the network request that may not really be cancelable as of now.

> We could store the view in a WeakReference and that might be enough to avoid memory leaks.

I think that would be a great idea.

> I saw that Picasso implements LifecycleObserver, so you could optionally add it as an observer in the fragment and automatically cancel all jobs on destroy. Would that be a good pattern to follow?

Oh, that's interesting.

We kind of should cover this with the attached state listener though; a destroy of the fragment should detach the views too.

And I assume we'd still run into the "cancel does not happen immediately" problem?

A `WeakReference` for the View sounds like the best option imo.

components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt
```diff
@@ -110,9 +120,7 @@ class BrowserIcons(
             is IconLoader.Result.BitmapResult -> Icon(result.bitmap, source = result.source)
 
             is IconLoader.Result.BytesResult ->
-                decode(result.bytes, decoders, desiredSize)?.let { bitmap ->
-                    Icon(bitmap, source = result.source)
-                } ?: return generator.generate(context, request)
+                decode(result.bytes, decoders, desiredSize)?.let { Icon(it, source = result.source) }
```
pocmo: I do not understand how this change is related and why we make it: Why would we want to continue to run through the processors with a `null` icon?

components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt
```diff
@@ -138,6 +146,57 @@ class BrowserIcons(
...
+     * @param request Load icon for this given [IconRequest].
+     * @param placeholder [Drawable] to display while icon is loading.
+     * @param error [Drawable] to display if loading fails.
+     */
+    fun loadIntoView(
+        view: ImageView,
+        request: IconRequest,
+        placeholder: Drawable? = null,
+        error: Drawable? = null
+    ) = scope.async(Dispatchers.Main) {
```
pocmo: Why is this using `async` instead of `launch`? It looks like we do not actually return anything from here. Even for the tests returning an actual `Job` from `launch()` may make more sense than an `Deferred<Unit>`.

components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt
NotWoods: I think I made this change because the linter complained at one point. `process` returns immediately if the icon is null so the effect is the same. I'll remove these changes.

PR #3697, opened 2019-07-08
Closes #3573: Adds SendTabUseCases in a component 

### LINKED ISSUE #3573: Add UseCases for common FxA actions
Potential list:
 - SendTabToDevice
 - SendTabToAll
 - ???

jonalmeida:
This PR adds a new component `feature-sendtab` with use cases that make sending a tab easier. Providing an AccountManager is the only dependency, and if the account is in the correct state, tabs will be sent.

An additional helper extension was also added to `support-ktx` for retrieving the cartesian product of two collections.

### Pull Request checklist

### DIFF
--- a/components/feature/sendtab/src/main/java/mozilla/components/feature/sendtab/SendTabUseCases.kt
+++ b/components/feature/sendtab/src/main/java/mozilla/components/feature/sendtab/SendTabUseCases.kt
@@ -1,0 +1,176 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ *  License, v. 2.0. If a copy of the MPL was not distributed with this
+ *  file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package mozilla.components.feature.sendtab
+
+import androidx.annotation.VisibleForTesting
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Deferred
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.async
+import kotlinx.coroutines.plus
+import mozilla.components.concept.sync.Device
+import mozilla.components.concept.sync.DeviceCapability
+import mozilla.components.concept.sync.DeviceConstellation
+import mozilla.components.concept.sync.DeviceEventOutgoing.SendTab
+import mozilla.components.concept.sync.TabData
+import mozilla.components.service.fxa.manager.FxaAccountManager
+import mozilla.components.support.ktx.kotlin.crossProduct
+import kotlin.coroutines.CoroutineContext
+
+/**
+ * Contains use cases for sending tabs to devices related to the firefox-accounts.
+ *
+ * @param accountManager The AccountManager on which we want to retrieve our devices.
+ * @param coroutineContext The Coroutine Context on which we want to do the actual sending.
+ * By default, we want to do this on the IO dispatcher since it involves making network requests to
+ * the Sync servers.
+ */
+class SendTabUseCases(
+    accountManager: FxaAccountManager,
+    coroutineContext: CoroutineContext = Dispatchers.IO
+) {
+    private var job: Job = SupervisorJob()
+    private val scope = CoroutineScope(coroutineContext) + job
+
+    class SendToDeviceUseCase internal constructor(
+        private val accountManager: FxaAccountManager,
+        private val scope: CoroutineScope
+    ) {
+        /**
+         * Sends the tab to provided deviceId if possible.
+         *
+         * @param deviceId The deviceId to send the tab.
+         * @param tab The tab to send.
+         * @return a deferred boolean if the result was successful or not.
+         */
+        operator fun invoke(deviceId: String, tab: TabData) =
+            scope.async { send(deviceId, tab) }
+
+        /**
+         * Sends the tabs to provided deviceId if possible.
+         *
+         * @param deviceId The deviceId to send the tab.
+         * @param tabs The list of tabs to send.
+         * @return a deferred boolean as true if the combined result was successful or not.
+         */
+        operator fun invoke(deviceId: String, tabs: List<TabData>): Deferred<Boolean> {
+            return scope.async {
+                tabs.map { tab ->
+                    send(deviceId, tab)
+                }.fold(true) { acc, result ->
+                    acc and result
+                }
+            }
+        }
+
+        private suspend fun send(deviceId: String, tab: TabData): Boolean {
+            filterSendTabDevices(accountManager) { constellation, devices ->
+                val device = devices.firstOrNull {
+                    it.id == deviceId
+                }
+                device?.let {
+                    return constellation.sendEventToDeviceAsync(
+                        device.id,
+                        SendTab(tab.title, tab.url)
+                    ).await()
+                }
+            }
+
+            return false
+        }
+    }
+
+    class SendToAllUseCase internal constructor(
+        private val accountManager: FxaAccountManager,
+        private val scope: CoroutineScope
+    ) {
+
+        /**
+         * Sends the tab to all send-tab compatible devices.
+         *
+         * @param tab The tab to send.
+         * @return a deferred boolean as true if the combined result was successful or not.
+         */
+        operator fun invoke(tab: TabData): Deferred<Boolean> {
+            return scope.async {
+                sendToAll { devices ->
+                    devices.map { device ->
+                        device to tab
+                    }
+                }
+            }
+        }
+
+        /**
+         * Sends the tabs to all the send-tab compatible devices.
+         *
+         * @param tabs a collection of tabs to send.
+         * @return a deferred boolean as true if the combined result was successful or not.
+         */
+        operator fun invoke(tabs: Collection<TabData>): Deferred<Boolean> {
+            return scope.async {
+                sendToAll { devices ->
+                    devices.crossProduct(tabs) { device, tab ->
+                        device to tab
+                    }
+                }
+            }
+        }
+
+        private suspend inline fun sendToAll(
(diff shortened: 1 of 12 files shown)


components/feature/sendtab/src/test/java/mozilla/components/feature/sendtab/SendTabUseCasesTest.kt
```diff
@@ -0,0 +1,254 @@
...
+class SendTabUseCasesTest {
+
+    private val manager: FxaAccountManager = mock()
+    private val account: OAuthAccount = mock()
+    private val constellation: DeviceConstellation = mock()
+    private val state: ConstellationState = mock()
+
+    @Before
+    fun setup() {
+        `when`(manager.authenticatedAccount()).thenReturn(account)
```
Amejia481: nit: A nice way will be [whenever](https://github.com/mozilla-mobile/android-components/blob/6341e10cde7bc25f6655e3c3cd84dc6efea9fd72/components/support/test/src/main/java/mozilla/components/support/test/Mock.kt#L30) :)

components/feature/sendtab/src/main/java/mozilla/components/feature/sendtab/SendTabUseCases.kt
```diff
@@ -0,0 +1,109 @@
...
+            }
+        }
+    }
+
+    val sendToDevice: SendToDeviceUseCase by lazy { SendToDeviceUseCase(accountManager) }
+    val sendToAll: SendToAllUseCase by lazy { SendToAllUseCase(accountManager) }
+}
+
+@VisibleForTesting
+internal fun filterSendTabDevices(
```
Amejia481: Could this function be `inline`?

jonalmeida: This was inline but coverage doesn't understand inline functions. I'll make it inline regardless and we can force-merge this since we know it is actually green.

components/feature/sendtab/src/test/java/mozilla/components/feature/sendtab/SendTabUseCasesTest.kt
jonalmeida: I don't follow the value of the alias:
```kotlin
whenever(manager.authenticatedAccount()).thenReturn(account)
`when`(manager.authenticatedAccount()).thenReturn(account)
```

They both do the same thing but nothing is really gained from the alias. 🤔

components/feature/sendtab/src/test/java/mozilla/components/feature/sendtab/SendTabUseCasesTest.kt
```diff
@@ -0,0 +1,254 @@
...
+import org.mockito.Mockito.times
+import org.mockito.Mockito.verify
+import java.util.UUID
+
+class SendTabUseCasesTest {
+
+    private val manager: FxaAccountManager = mock()
+    private val account: OAuthAccount = mock()
+    private val constellation: DeviceConstellation = mock()
+    private val state: ConstellationState = mock()
```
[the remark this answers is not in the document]
jonalmeida: I was curious about this. When I verify the constellation in each of my tests, the mock already seems to be reset as expected so I assumed that that was the assurance that was provided by Mockito.

I'll update this to use the annotations regardless to keep it consistent. :)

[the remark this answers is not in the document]
jonalmeida: Looks like we have that safety according to the JUnit docs for `@Test`:

> To run the method, JUnit first constructs a fresh instance of the class then invokes the annotated method.

https://junit.org/junit4/javadoc/latest/org/junit/Test.html

components/feature/sendtab/src/main/java/mozilla/components/feature/sendtab/SendTabUseCases.kt
```diff
@@ -0,0 +1,109 @@
...
+    accountManager: FxaAccountManager,
+    block: (DeviceConstellation, Collection<Device>) -> Unit
+) {
+    val constellation = accountManager.authenticatedAccount()?.deviceConstellation() ?: return
+
+    constellation.state()?.let { state ->
+        state.otherDevices.filter {
+            it.capabilities.contains(DeviceCapability.SEND_TAB)
+        }.let { device ->
```
grigoryk: `devices`

components/feature/sendtab/src/main/java/mozilla/components/feature/sendtab/SendTabUseCases.kt
```diff
@@ -0,0 +1,109 @@
...
+         * @param tabs The list of tabs to send.
+         */
+        operator fun invoke(deviceId: String, tabs: List<TabData>) {
+            filterSendTabDevices(accountManager) { constellation, devices ->
+                val device = devices.firstOrNull { it.id == deviceId }
+
+                device?.let {
+                    tabs.forEach { tab ->
+                        constellation.sendEventToDeviceAsync(device.id, SendTab(tab.title, tab.url))
+                    }
```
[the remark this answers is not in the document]
jonalmeida: The original thought was to check the account/device exists before trying to iterate through all the tabs and getting back multiple errors. That might be an over-optimization for now and we can change that when we have a sendAll API from a-s.

components/feature/sendtab/src/main/java/mozilla/components/feature/sendtab/SendTabUseCases.kt
```diff
@@ -0,0 +1,106 @@
...
+import mozilla.components.concept.sync.DeviceConstellation
+import mozilla.components.concept.sync.DeviceEventOutgoing.SendTab
+import mozilla.components.concept.sync.TabData
+import mozilla.components.service.fxa.manager.FxaAccountManager
+import mozilla.components.support.ktx.kotlin.crossProduct
+
+/**
+ * Contains use cases for sending tabs to devices related to the firefox-accounts.
+ */
+class SendTabUseCases(accountManager: FxaAccountManager) {
```
grigoryk: These methods currently swallow any errors that may occur.

While there will be some non-actionable errors, I'd expect applications to at least handle network problems by display a "try again" toast, for example. In either case, I think we shouldn't pretend we succeeded when we either didn't do anything or actually failed. Especially if we failed due to a transient network error, that user may be able to correct themselves easily (e.g. by disconnecting from a bad wi-fi).

`sendEventToDeviceAsync` returns a Deferred<Boolean>, where `false` means "it didn't work". It's not a very rich error reporting API, but it's what you want to expose for a basic "try again later" UI experience.

components/feature/sendtab/src/test/java/mozilla/components/feature/sendtab/SendTabUseCasesTest.kt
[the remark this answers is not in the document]
Amejia481: Awesome, I didn't know that, it's nice to know :)

components/feature/sendtab/src/test/java/mozilla/components/feature/sendtab/SendTabUseCasesTest.kt
Amejia481: Yes you're right, It's a bit easier to write because you can use the IDE auto complete, but a part from that it's just aesthetics

components/feature/sendtab/src/main/java/mozilla/components/feature/sendtab/SendTabUseCases.kt
Amejia481: 👍

components/feature/sendtab/src/main/java/mozilla/components/feature/sendtab/SendTabUseCases.kt
jonalmeida: That's a fair point. Thanks for the catch! What I've done is to make the use cases into suspend functions since we have to combine the result of sending multiple tabs together before we can return it to the app in case of a failure.

components/feature/sendtab/src/main/java/mozilla/components/feature/sendtab/SendTabUseCases.kt
```diff
@@ -0,0 +1,138 @@
...
+         */
+        suspend operator fun invoke(deviceId: String, tabs: List<TabData>): Boolean {
+            return tabs.map { tab ->
+                sendAsync(deviceId, tab)
+            }.fold(true) { acc, result ->
+                acc and result
+            }
+        }
+
+        private suspend fun sendAsync(deviceId: String, tab: TabData): Boolean {
```
grigoryk: Here and elsewhere:
The convention for methods with the `Async` suffix is that they return a Deferred value. How about just `send`? It's already a suspend function.

jonalmeida: Fixed! I've left the `Async` suffix on only the use case bindings since those are the only ones that need it now.

jonalmeida:
> I think you should either:
> 
>     * provide guidance on why these methods are suspend, and which dispatcher is appropriate for performing that work (e.g. maybe Dispatchers.IO or any other non-main thread since we'll do networking here), or...
> 
>     * own the execution within the use-case, making these methods real "async" methods that return Deferred. This way your consumers don't need to worry about picking the right execution context, and we're not giving them a foot-gun to run network calls on the main thread.
> 
> 
> WDYT? I'm leaning toward the second option, it seems less error prone.

I think we can do both!

This was the dilema I had as well. For single tab sending this isn't an issue, but for any combination of 'send all', we need to reduce all the results of the tab sending to one value and returning that value in a non-suspendable state seemed difficult.

Looking back at this now, I think we can add a default dispatcher for guidance as you suggested and also own the dispatching as well by returning a `Deferred<Boolean>` making it easier for the caller to decide if they want to await for the result or not. 🎉

PR #3704, opened 2019-07-09
Closes #3666: Expose useWideViewPort for SystemEngine

### LINKED ISSUE #3666: Expose useWideViewPort web setting
FFTV needs to override the WebSetting [`useWideViewPort`](https://developer.android.com/reference/android/webkit/WebSettings.html#setUseWideViewPort(boolean)), but it is currently not exposed.

psymoon:
### Pull Request checklist

### DIFF
--- a/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
+++ b/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
@@ -319,6 +319,7 @@ class SystemEngineSession(
                 webFontsEnabled = it.webFontsEnabled
                 displayZoomControls = it.displayZoomControls
                 loadWithOverviewMode = it.loadWithOverviewMode
+                useWideViewPort = it.useWideViewPort
                 trackingProtectionPolicy = it.trackingProtectionPolicy
                 historyTrackingDelegate = it.historyTrackingDelegate
                 requestInterceptor = it.requestInterceptor
--- a/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
+++ b/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
@@ -49,6 +49,7 @@ class SystemEngineSession(
     @Volatile internal var trackingProtectionPolicy: TrackingProtectionPolicy? = null
     @Volatile internal var webFontsEnabled = true
     @Volatile internal var currentUrl = ""
+    @Volatile internal var useWideViewPort: Boolean? = null
     @Volatile internal var fullScreenCallback: WebChromeClient.CustomViewCallback? = null
 
     // This is public for FFTV which needs access to the WebView instance. We can mark it internal once
@@ -272,6 +273,11 @@ class SystemEngineSession(
         webSettings.savePassword = false
     }
 
+    private fun setUseWideViewPort(settings: WebSettings, useWideViewPort: Boolean?) {
+        this.useWideViewPort = useWideViewPort
+        useWideViewPort?.let { settings.useWideViewPort = it }
+    }
+
     private fun initSettings(webView: WebView, s: WebSettings) {
         internalSettings = object : Settings() {
             override var javascriptEnabled by WebSetting(s::getJavaScriptEnabled, s::setJavaScriptEnabled)
@@ -281,7 +287,9 @@ class SystemEngineSession(
             override var userAgentString by WebSetting(s::getUserAgentString, s::setUserAgentString)
             override var displayZoomControls by WebSetting(s::getDisplayZoomControls, s::setDisplayZoomControls)
             override var loadWithOverviewMode by WebSetting(s::getLoadWithOverviewMode, s::setLoadWithOverviewMode)
-            override var useWideViewPort by WebSetting(s::getUseWideViewPort, s::setUseWideViewPort)
+            override var useWideViewPort: Boolean?
+                get() = this@SystemEngineSession.useWideViewPort
+                set(value) = setUseWideViewPort(s, value)
             override var supportMultipleWindows by WebSetting(s::supportMultipleWindows, s::setSupportMultipleWindows)
             override var allowFileAccessFromFileURLs by WebSetting(
                     s::getAllowFileAccessFromFileURLs, s::setAllowFileAccessFromFileURLs)
@@ -344,7 +352,7 @@ class SystemEngineSession(
     override fun toggleDesktopMode(enable: Boolean, reload: Boolean) {
         val webSettings = webView.settings
         webSettings.userAgentString = toggleDesktopUA(webSettings.userAgentString, enable)
-        webSettings.useWideViewPort = enable
+        webSettings.useWideViewPort = settings.useWideViewPort ?: enable
 
         notifyObservers { onDesktopModeChange(enable) }
 
--- a/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
+++ b/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
@@ -281,6 +281,7 @@ class SystemEngineSession(
             override var userAgentString by WebSetting(s::getUserAgentString, s::setUserAgentString)
             override var displayZoomControls by WebSetting(s::getDisplayZoomControls, s::setDisplayZoomControls)
             override var loadWithOverviewMode by WebSetting(s::getLoadWithOverviewMode, s::setLoadWithOverviewMode)
+            override var useWideViewPort by WebSetting(s::getUseWideViewPort, s::setUseWideViewPort)
             override var supportMultipleWindows by WebSetting(s::supportMultipleWindows, s::setSupportMultipleWindows)
             override var allowFileAccessFromFileURLs by WebSetting(
                     s::getAllowFileAccessFromFileURLs, s::setAllowFileAccessFromFileURLs)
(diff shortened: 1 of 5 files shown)


components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
```diff
@@ -343,6 +345,7 @@ class SystemEngineSession(
         val webSettings = webView.settings
         webSettings.userAgentString = toggleDesktopUA(webSettings.userAgentString, enable)
         webSettings.useWideViewPort = enable
+        defaultSettings?.useWideViewPort = enable
```
csadilek: Hm won't this change the default for all sessions if desktop mode is toggled? Is this needed / if so can you add a comment?

components/browser/engine-system/src/test/java/mozilla/components/browser/engine/system/SystemEngineSessionTest.kt
```diff
@@ -633,7 +637,8 @@ class SystemEngineSessionTest {
     @Test
     fun desktopMode() {
```
csadilek: There's one more test that needs an update: `withProvidedDefaultSettings` to test that an engine default can be specified for `useWideViewPort`.

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
csadilek: In other words, I don't think we should be writing default settings within a single session. So, wondering if you really need that.

pocmo: I'm concerned about the interaction between `useWideViewPort` and `toggleDesktopMode` here.

That desktop mode internally flips `useWideViewPort` is an implementation detail and ideally shouldn't be leaked to the consumer. Or in other words I think we shouldn't let the consuming app deal with figuring out how to not override itself.

As a consumer I would expect the following behavior:
- If I do not specify `useWideViewPort` then desktop mode will toggle it on/off.
- If I do specify `useWideViewPort = false` then desktop mode will need to toggle it on/off
- If I do specify `useWideViewPort = true` then I expect it to be always on and desktop mode will never toggle it off.

What do you think? (CC @csadilek)

This will probably require the addition of a boolean flag (instead of only writing/reading from WebSettings) so that we can figure out whether this WebSetting was flipped deliberately by `useWideViewPort` (and needs to stay that way) or by something else like desktop mode.

psymoon: Makes sense to me

csadilek: @pocmo @psymoon yes, that sounds good! I think the user supplied default setting (which is for all sessions) shouldn't change by toggling desktop mode. It should just always reflect what the user configured. Of course, when "asking" a specific session for `session.setttings.useWideViewPort` the result could be different than the default if the session is in desktop mode or it was configured directly. We can add some session-internal state to decide if we have to toggle `useWideViewPort` when `toggleDesktopMode` is invoked, as @pocmo suggested, based on the value of `settings.useWideViewPort` before desktop mode is toggled.

psymoon: Doesn't the desktop mode also require non-mobile UA? At least in FFTV, `useWideViewPort = true` doesn't guarantee desktop mode 🤔 . I agree with your concern though - maybe @dnarcese can provide additional context on the consumer behaviour.

pocmo: Yeah, currently `toggleDesktopMode` does enable widemode AND changes the UA.

dnarcese: >    -  If I do not specify useWideViewPort then desktop mode will toggle it on/off.
>    - If I do specify useWideViewPort = false then desktop mode will need to toggle it on (and off again)
>    - If I do specify useWideViewPort = true then I expect it to be always on and desktop mode will never toggle it off.

I agree with this behaviour.

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
```diff
@@ -272,6 +273,11 @@ class SystemEngineSession(
         webSettings.savePassword = false
     }
 
+    private fun setUseWideViewPort(s: WebSettings, useWideViewPort: Boolean?) {
+        this@SystemEngineSession.useWideViewPort = useWideViewPort
```
pocmo: nit: I can't see the whole code in this review mode - but can't that be just `this.useWideViewPort = useWideViewPort`?

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
```diff
@@ -272,6 +273,11 @@ class SystemEngineSession(
         webSettings.savePassword = false
     }
 
+    private fun setUseWideViewPort(s: WebSettings, useWideViewPort: Boolean?) {
```
pocmo: nit: `s` is not really a good name - no need to save letters here :)

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
```diff
@@ -342,7 +352,7 @@ class SystemEngineSession(
     override fun toggleDesktopMode(enable: Boolean, reload: Boolean) {
         val webSettings = webView.settings
         webSettings.userAgentString = toggleDesktopUA(webSettings.userAgentString, enable)
-        webSettings.useWideViewPort = enable
+        webSettings.useWideViewPort = settings.useWideViewPort ?: enable
```
pocmo: My assumption here is that for `enable = true` we will always need to enable `useWideViewPort` (otherwise desktop mode is just broken). Only for `enable = false` we need to either fall back to the consumer's choice  or `false`.

components/browser/engine-system/src/test/java/mozilla/components/browser/engine/system/SystemEngineSessionTest.kt
```diff
@@ -659,6 +665,22 @@ class SystemEngineSessionTest {
...
+        val webView = mock<WebView>()
+        val webViewSettings = mock<WebSettings>()
+
+        engineSession.webView = webView
+        whenever(webView.settings).thenReturn(webViewSettings)
+        whenever(webViewSettings.userAgentString).thenReturn(userAgentMobile)
+
+        engineSession.toggleDesktopMode(false)
+        verify(webViewSettings).useWideViewPort = true
+    }
```
pocmo: Let's make sure we have a test case for all three cases:
- toggle desktop mode on and off without modifying useWideView (we should have that already?)
- toggle desktop mode on and off with useWideView=true
- toggle desktop mode on and off with useWideView=false

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
psymoon: This also implies for `toggleDesktopUA` and `notifyObserver` condition right?

pocmo: Yes, the UA will continue to toggle like before :)

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
```diff
@@ -49,6 +49,7 @@ class SystemEngineSession(
     @Volatile internal var trackingProtectionPolicy: TrackingProtectionPolicy? = null
     @Volatile internal var webFontsEnabled = true
     @Volatile internal var currentUrl = ""
+    @Volatile internal var useWideViewPort: Boolean? = null
```
pocmo: Reading through that again I wonder if we should leave a short comment here why we have this flag (mentioning toggleDesktopMode)?

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
```diff
@@ -338,13 +348,18 @@ class SystemEngineSession(
 
     /**
      * See [EngineSession.toggleDesktopMode]
+     *
+     * Precondition:
+     * If settings.useWideViewPort = true, then desktopMode is always on (will never be toggled)
+     * If settings.useWideViewPort = false or null, then desktopMode can be toggled on/off
```
pocmo: Love the idea to add a comment here explaining side effects. But isn't the comment a bit misleading? You can still continue to use toggleDesktopMode() regardless of what you've set useWideViewPort to?

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
```diff
@@ -338,13 +348,18 @@ class SystemEngineSession(
...
     override fun toggleDesktopMode(enable: Boolean, reload: Boolean) {
         val webSettings = webView.settings
-        webSettings.userAgentString = toggleDesktopUA(webSettings.userAgentString, enable)
-        webSettings.useWideViewPort = enable
+        val toggle = if (settings.useWideViewPort == true) true else enable
+        webSettings.userAgentString = toggleDesktopUA(webSettings.userAgentString, toggle)
+        webSettings.useWideViewPort = toggle
 
-        notifyObservers { onDesktopModeChange(enable) }
+        notifyObservers { onDesktopModeChange(toggle) }
```
pocmo: Here we still want to pass in `enable` because we enabled desktop mode regardless of `useWideViewPort` (we do more than that, e.g. right now changing the UA).

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
psymoon: Ah okay, I misunderstood the toggle logic

pocmo:
Oh, before landing: Can you mention this in docs/changelog.md? :)

PR #3706, opened 2019-07-09
Fixes #3705 - Fix conflicting authorities

### LINKED ISSUE #3705: INSTALL_FAILED_CONFLICTING_PROVIDER when using feature-prompts from two apps.
Yesterday we landed code in `feature-prompts` that introduces a content provider with the authority `mozilla.components.feature.prompts.fileprovider`. This prevents multiple apps with that feature/authority to be installed (INSTALL_FAILED_CONFLICTING_PROVIDER).

NotWoods:
(Oops)

Removes hardcoded authority in feature-prompts so that two apps that use the library can be installed at the same time.

### Pull Request checklist

### DIFF
--- a/components/feature/prompts/src/main/AndroidManifest.xml
+++ b/components/feature/prompts/src/main/AndroidManifest.xml
@@ -5,7 +5,7 @@
     package="mozilla.components.feature.prompts">
     <application>
         <provider
-            android:authorities="mozilla.components.feature.prompts.fileprovider"
+            android:authorities="${applicationId}.fileprovider"
             android:name="androidx.core.content.FileProvider"
             android:exported="false"
             android:grantUriPermissions="true">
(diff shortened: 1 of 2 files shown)


components/feature/prompts/src/main/AndroidManifest.xml
```diff
@@ -5,7 +5,7 @@
     package="mozilla.components.feature.prompts">
     <application>
         <provider
-            android:authorities="mozilla.components.feature.prompts.fileprovider"
+            android:authorities="${applicationId}.fileprovider"
```
pocmo: Let's get this merged to fix the immediate problem for the app teams. However I wonder if we should still have some "android component" uniqueness in the name instead of just `fileprovider` which may be used by an app too (causing a conflict).

PR #3708, opened 2019-07-09
Clean up toolbar feature code

NotWoods:
Some minor clean up for feature-toolbar.

- Instead of having a list of `Any` in `ToolbarAutocompleteFeature`, just have two similar for loops. Overall it's less code and doesn't deal with casting.
- Simplify `URLRenderer.updateUrl` and keep the order consistent between the when statement and the enum.

### DIFF
--- a/components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarAutocompleteFeature.kt
+++ b/components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarAutocompleteFeature.kt
@@ -17,18 +17,20 @@ import mozilla.components.concept.toolbar.Toolbar
 class ToolbarAutocompleteFeature(val toolbar: Toolbar) {
     private val historyProviders: MutableList<HistoryStorage> = mutableListOf()
     private val domainProviders: MutableList<DomainAutocompleteProvider> = mutableListOf()
-    private val allProviders = { historyProviders.asSequence() + domainProviders }
 
     init {
-        toolbar.setAutocompleteListener { value, delegate ->
-            for (provider in allProviders()) {
-                val result = provider.autocomplete(value)
-                if (result != null) {
-                    delegate.applyAutocompleteResult(result)
-                    return@setAutocompleteListener
-                }
+        toolbar.setAutocompleteListener { query, delegate ->
+            val historyResults = historyProviders.asSequence()
+                .mapNotNull { provider -> provider.getAutocompleteSuggestion(query)?.into() }
+            val domainResults = domainProviders.asSequence()
+                .mapNotNull { provider -> provider.getAutocompleteSuggestion(query)?.into() }
+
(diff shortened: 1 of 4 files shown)


components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarAutocompleteFeature.kt
```diff
@@ -40,32 +42,10 @@ class ToolbarAutocompleteFeature(val toolbar: Toolbar) {
         domainProviders.add(provider)
     }
 
-    // The rest is a little bit of boilerplate to tie things together.
-    // We terminate our dependency chains at the 'feature' level. This is an alternative to creating
-    // an "AutocompleteProvider" concept. This pattern allows a storage component, for example, to
```
jonalmeida: From this comment here, I'm not sure if the intent was to write an `AutocompleteProvider` concept eventually or this was the alternative to it.

jonalmeida: HistoryStorage may have been built concurrently while the AwesomeBar was being built so that may have been the reasoning for the different APIs when it comes to retrieving the autocomplete results.

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarAutocompleteFeature.kt
```diff
@@ -17,18 +17,20 @@ import mozilla.components.concept.toolbar.Toolbar
...
-                if (result != null) {
-                    delegate.applyAutocompleteResult(result)
-                    return@setAutocompleteListener
-                }
+        toolbar.setAutocompleteListener { query, delegate ->
+            val historyResults = historyProviders.asSequence()
+                .mapNotNull { provider -> provider.getAutocompleteSuggestion(query)?.into() }
+            val domainResults = domainProviders.asSequence()
+                .mapNotNull { provider -> provider.getAutocompleteSuggestion(query)?.into() }
+
```
jonalmeida: Fewer early returns! @pocmo  😛

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/internal/URLRenderer.kt
```diff
@@ -58,45 +60,30 @@ internal class URLRenderer(
...
-    }
-
-    private suspend fun setRegistrableDomainUrl(url: String) {
-        val host = url.toUri().host
-
-        if (!host.isNullOrEmpty()) {
-            toolbar.url = getRegistrableDomain(host) ?: url
-        } else {
-            toolbar.url = url
+        toolbar.url = when (configuration.renderStyle) {
```
jonalmeida: Thank you for clean up! I would knight you if I could ⚔️

PR #3709, opened 2019-07-09
Break apart different BrowserFragment versions

NotWoods:
@csadilek and I discussed this morning how to separate the different entry points for `BrowserFragment` (normal browsing, custom tabs, PWAs). We ended up settling on keeping the flat list of features structure, with 3 variant classes that append a few additional features depending on what entry point they're used for.

Whiteboard showing separation of features

This PR also includes two other changes:
- Pass `GoBackUseCase` directly to `SessionFeature`
- Fix `BrowserMenu` so that "Add to homescreen" doesn't show up in custom tabs.


samples/browser/src/main/java/org/mozilla/samples/browser/CustomTabBrowserFragment.kt
```diff
@@ -0,0 +1,46 @@
...
+                components.menuBuilder,
+                closeListener = { activity?.finish() }),
+            owner = this,
+            view = layout)
+
+        return layout
+    }
+
+    override fun onBackPressed(): Boolean =
+        super.onBackPressed() || customTabsToolbarFeature.onBackPressed()
```
jonalmeida: You probably want to evaluate the feature's `onBackPressed` first before super. Same goes for the `WebAppBroserFragment.kt`.

NotWoods: This matches the original behaviour where the custom tabs feature was the last to be evaluated. Should I change it?

pocmo: I think it's correct - just looks confusing. I was confused by it too. In this case `super.onBackPressed()` does not call through to the system (like in an activity), but to `BaseBrowserFragment` which will go through the other features first.

NotWoods: I updated this method with a comment to describe what's happening and clear up the ambiguity.

jonalmeida: Gotcha, I see it now. Thanks for clarifying!

PR #3714, opened 2019-07-09
Closes #3703: Makes menu respect width of items

### LINKED ISSUE #3703: FNX2-16751 ⁃ Menu should have a width based on the width of its elements
Currently the menu is a fixed dimension of 250dp. Let's make it constrained based on the width of its children instead.

[Relevant Fenix issue](https://github.com/mozilla-mobile/fenix/issues/1871)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FNX2-16751)

sblatz:
### Pull Request checklist


components/browser/menu/src/main/java/mozilla/components/browser/menu/item/BrowserMenuImageText.kt
```diff
@@ -23,6 +24,19 @@ internal fun ImageView.setTintResource(@ColorRes tintColorResource: Int) {
     }
 }
 
+internal fun TextView.tintedDrawable(@DrawableRes imageResource: Int, @ColorRes tintColorResource: Int): Drawable {
+    val draw = context.getDrawable(imageResource)!!
+
+    if (tintColorResource != NO_ID) {
+        DrawableCompat.setTint(
```
NotWoods: nit: could just use `draw.setTint` since we're only supporting Lollipop and above.

components/browser/menu/src/main/java/mozilla/components/browser/menu/item/BrowserMenuImageText.kt
```diff
@@ -56,25 +70,24 @@ open class BrowserMenuImageText(
...
+        with(textView) {
+            text = label
+            setColorResource(textColorResource)
 
-    private fun bindImage(view: View) {
-        val imageView = view.findViewById<AppCompatImageView>(R.id.image)
-        with(imageView) {
-            setImageResource(imageResource)
-            setTintResource(iconTintColorResource)
+            setCompoundDrawablesWithIntrinsicBounds(
```
NotWoods: nit: I just added a KTX function for this 😃 
```suggestion
            putCompoundDrawablesWithIntrinsicBounds(start = tintedDrawable(imageResource, iconTintColorResource))
```

components/browser/menu/src/main/res/layout/mozac_browser_menu_item_checkbox.xml
```diff
@@ -2,13 +2,33 @@
...
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    style="@android:style/TextAppearance.Material.Menu"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:background="?selectableItemBackground"
     android:layout_width="match_parent"
-    android:layout_height="@dimen/mozac_browser_menu_item_container_layout_height"
-    android:button="@null"
-    android:drawableEnd="?android:attr/listChoiceIndicatorMultiple"
-    android:drawablePadding="12dp"
+    android:layout_height="48dp"
```
NotWoods: Does it make sense to use a minimum height instead and expand if the text overflows? I'm not sure what the native menu does here.

Alternatively, maybe `@dimen/mozac_browser_menu_item_container_layout_height` should be used?

components/browser/menu/src/main/res/layout/mozac_browser_menu_item_checkbox.xml
```diff
@@ -2,13 +2,33 @@
...
+
+    <TextView
+        android:id="@+id/text"
+        style="@style/Mozac.Browser.Menu.Item.Text"
+        android:background="@null"
+        android:clickable="false"
+        android:layout_width="wrap_content"
+        android:checkMark="@null"
+        android:layout_height="@dimen/mozac_browser_menu_item_container_layout_height"
+        android:gravity="start|center_vertical"
```
NotWoods: Is gravity necessary since the parent has it set already?

components/browser/menu/src/main/res/layout/mozac_browser_menu_item_image_text.xml
```diff
@@ -2,29 +2,12 @@
...
-            style="@style/Mozac.Browser.Menu.Item.ImageText.Label"
-            android:clickable="false"
-            android:background="@android:color/transparent"
-            android:gravity="center_vertical"
-            tools:text="Item"
-            android:importantForAccessibility="no"/>
-</LinearLayout>
\ No newline at end of file
+<TextView xmlns:android="http://schemas.android.com/apk/res/android"
+    android:id="@+id/imageText"
```
NotWoods: Would be nice to use `tools:drawableStart` and `tools:text` so the design view has some sample data.

components/browser/menu/src/main/res/layout/mozac_browser_menu_item_checkbox.xml
sblatz: It seems necessary--when I remove it, it's no longer centered vertically. I can remove the "Start" though. 

image

components/browser/menu/src/test/java/mozilla/components/browser/menu/item/BrowserMenuHighlightableItemTest.kt
```diff
@@ -59,14 +59,16 @@ class BrowserMenuHighlightableItemTest {
...
 
-        val imageView = view.findViewById<AppCompatImageView>(R.id.image)
+        val image = textView.compoundDrawables[0]
+
         val highlightImageView = view.findViewById<AppCompatImageView>(R.id.highlight_image)
 
-        assertNotNull(imageView.drawable)
-        assertNotNull(imageView.imageTintList)
+        assertNotNull(image)
+        // TODO: Not sure how to verify iconTintColorResource:
```
sblatz: Any ideas on how to best test this? I'm having trouble since both color filter and compoundDrawableTintList are null

sblatz:
I believe I addressed all concerns raised. Still need some pointers on the tests as mentioned above, but should be ready for another review!

components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenu.kt
```diff
@@ -29,7 +29,8 @@ import mozilla.components.support.ktx.android.view.isRTL
  * A popup menu composed of BrowserMenuItem objects.
  */
 class BrowserMenu internal constructor(
-    private val adapter: BrowserMenuAdapter
+    private val adapter: BrowserMenuAdapter,
+    private val extras: Map<String, Any>? = null
```
pocmo: I wonder why this is a cryptic `extras` map and not just a parameter for the desired width? 🤔

components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenu.kt
```diff
@@ -61,6 +62,12 @@ class BrowserMenu internal constructor(
             adapter = this@BrowserMenu.adapter
         }
 
+        val recyclerView = view.findViewById<RecyclerView>(R.id.mozac_browser_menu_recyclerView)
+        recyclerView.minimumWidth = when (extras?.get("minimum_width")) {
+            null -> WindowManager.LayoutParams.WRAP_CONTENT
+            else -> extras["minimum_width"] as Int
+        }
```
pocmo: nit: We already find and manipulate the RecyclerView a couple of lines above (`menuList`). :)

components/browser/menu/src/main/java/mozilla/components/browser/menu/item/BrowserMenuCheckbox.kt
```diff
@@ -14,9 +20,36 @@ import mozilla.components.browser.menu.R
...
  */
 class BrowserMenuCheckbox(
-    label: String,
-    initialState: () -> Boolean = { false },
-    listener: (Boolean) -> Unit
-) : BrowserMenuCompoundButton(label, initialState, listener) {
+    private val label: String,
+    private val initialState: () -> Boolean = { false },
+    private val listener: (Boolean) -> Unit
+) : BrowserMenuItem {
```
pocmo: Can you help me understand why we copy (and change?) the implementation here and can't rely on `BrowserMenuCompoundButton` anymore?

CC @jonalmeida, who I think created this base class.

components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenu.kt
sblatz: 🤦‍♂ whoops.

components/browser/menu/src/main/java/mozilla/components/browser/menu/item/BrowserMenuCheckbox.kt
sblatz: `BrowserMenuCheckbox` is now a `LinearLayout` at the parent level instead of an `AppCompatCheckbox`, because of this it cannot be cast `BrowserMenuCompoundButton` without crashing. We also want to add custom `onClick` behavior.

jonalmeida: Sorry, I may have missed something, I don't follow why compound buttons need to have a LinearLayout to alter the width. Couldn't we just change the `layout_width` in `mozac_browser_menu_item_switch` and `mozac_browser_menu_item_checkbox`?

pocmo:
@sblatz There seems to be a test failure here

components/browser/menu/src/main/java/mozilla/components/browser/menu/item/BrowserMenuCheckbox.kt
sblatz: @jonalmeida unfortunately, no.

@NotWoods and I tried to do this, but it seems there's a bug with the way we were using `AppCompatCheckbox` that made it so doing `layout_width` was not respected (the text would get clipped). I believe this is because we were using `drawableEnd` and setting the actual button itself to `null` which is a bit of a hacky solution to begin with.

components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenu.kt
sblatz: Resolved with a separate param :)

sblatz:
> Are all those changes related to the menu width? 🤔

Yep!

sblatz:
As @jonalmeida was testing this, he noticed that the wrapping still wasn't working properly with BrowserMenuImageText. It seems like wrap_content just [does not play nicely with compound drawables](https://stackoverflow.com/questions/47031749/textview-drawablestart-wrap-content-doesnt-work-well). I'm going to close this PR, as I am focused on other things right now, but I'm going to update the ticket with some of the information I found when investigating.

PR #3727, opened 2019-07-10
Bump version of application-services libraries to 0.34.0

thomcc:
Only real change here is that we return results in a slightly different order. Note that the order was never guaranteed, so this shouldn't matter.

CC @grigoryk

### DIFF
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -25,7 +25,7 @@ object Versions {
     const val jna = "5.2.0"
     const val disklrucache = "2.0.2"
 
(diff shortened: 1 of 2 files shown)


grigoryk:
Isn't it nice that those test actually exercise the native stack :)

thomcc:
> Actual work here would be in ensuring this doesn't blow up in some interesting way or regress some behaviour.

From slack: This actually has gotten manual QA in fenix already: https://groups.google.com/a/mozilla.com/d/msg/application-services/yMkTBBmqyCY/R2EDbe_2BQAJ (not sure how public that link is, but the important bit is "Recommendation from QA:  Ship it.")

grigoryk:
That's really great! :ship:

PR #3730, opened 2019-07-10
Add helper for one-time OnGlobalLayoutListener

NotWoods:
This is a pattern used in a few places in A-C and Fenix, so I think it makes sense to put it in a helper to improve readability. 

### Pull Request checklist

### DIFF
--- a/components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenu.kt
+++ b/components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenu.kt
@@ -12,11 +12,9 @@ import android.os.Build
 import android.view.Gravity
 import android.view.LayoutInflater
 import android.view.View
-import android.view.ViewTreeObserver
 import android.view.WindowManager
 import android.widget.PopupWindow
 import androidx.annotation.VisibleForTesting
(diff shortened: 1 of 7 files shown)


components/support/ktx/src/main/java/mozilla/components/support/ktx/android/view/ViewTreeObserver.kt
```diff
@@ -0,0 +1,21 @@
...
+package mozilla.components.support.ktx.android.view
+
+import android.view.ViewTreeObserver
+import android.view.ViewTreeObserver.OnGlobalLayoutListener
+
+/**
+ * Registers a one-time callback to be invoked when the global layout state
+ * or the visibility of views within the view tree changes.
+ */
+fun ViewTreeObserver.onNextGlobalLayout(callback: () -> Unit) {
```
Amejia481: Could we inline this function?

what do you think about?

```kotlin
inline fun ViewTreeObserver.onNextGlobalLayout( crossinline callback : () -> Unit) {
    var listener: OnGlobalLayoutListener? = null
    listener = OnGlobalLayoutListener {
        removeOnGlobalLayoutListener(listener)
        callback()
    }
    addOnGlobalLayoutListener(listener)
}
```

PR #3760, opened 2019-07-13
GeckoEngineSession now receives an elapsed time for page load

hawkinsw:
This patch brings ac up-to-date with changes to GV nightly
that expose a document request's elapsed load time. This
patch receives the elapsed time from GV and then calls
observers to notify them.

### DIFF
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -396,12 +396,13 @@ class GeckoEngineSession(
             }
         }
 
-        override fun onPageStop(session: GeckoSession, success: Boolean) {
+        override fun onPageStop(session: GeckoSession, success: Boolean, elapsed: Int) {
             // by the time we reach here, any new request will come from web content.
             // If it comes from the chrome, loadUrl(url) or loadData(string) will set it to
             // false.
             requestFromWebContent = true
             notifyObservers {
+                onElapsedLoadTimeMS(elapsed)
                 onProgress(PROGRESS_STOP)
                 onLoadingStateChange(false)
             }
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
@@ -67,6 +67,7 @@ class Session(
         fun onUrlChanged(session: Session, url: String) = Unit
         fun onTitleChanged(session: Session, title: String) = Unit
         fun onProgress(session: Session, progress: Int) = Unit
+        fun onElapsedLoadTimeMS(session: Session, progress: Int) = Unit
         fun onLoadingStateChanged(session: Session, loading: Boolean) = Unit
         fun onNavigationStateChanged(session: Session, canGoBack: Boolean, canGoForward: Boolean) = Unit
         fun onLoadRequest(
@@ -195,6 +196,13 @@ class Session(
         if (notifyObservers(old, new) { onProgress(this@Session, new) }) {
             store?.syncDispatch(UpdateProgressAction(id, new))
         }
+    }
+
+    /**
+     * The elapsed time of the load.
+     */
+    var elapsedLoadTimeMS: Int by Delegates.observable(-1) {
+        _, old, new -> notifyObservers(old, new) { onElapsedLoadTimeMS(this@Session, new) }
     }
 
     /**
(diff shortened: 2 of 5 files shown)


hawkinsw:
@bdekoz

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -396,12 +396,13 @@ class GeckoEngineSession(
...
         }
 
-        override fun onPageStop(session: GeckoSession, success: Boolean) {
+        override fun onPageStop(session: GeckoSession, success: Boolean, elapsed: Int) {
             // by the time we reach here, any new request will come from web content.
             // If it comes from the chrome, loadUrl(url) or loadData(string) will set it to
             // false.
             requestFromWebContent = true
             notifyObservers {
+                onElapsedTimeMS(elapsed)
```
pocmo: Naming nit: Let's not encode the unit (`ms`) into the method name, also this should probably say something about `load` in the name? 🤔

components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
```diff
@@ -197,6 +198,13 @@ class Session(
         }
     }
 
+    /**
+     * The elapsed time of the load.
+     */
+    var elapsedTimeMS: Int by Delegates.observable(0) {
+        _, old, new -> notifyObservers(old, new) { onElapsedTimeMS(this@Session, new) }
+    }
```
pocmo: This part I do not understand without context: Why do we want to expose this on `Session`? :)

Also: Same naming nit as for the callback method.

components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
```diff
@@ -197,6 +198,13 @@ class Session(
         }
     }
 
+    /**
+     * The elapsed time of the load.
+     */
+    var elapsedTimeMS: Int by Delegates.observable(0) {
```
pocmo: I wonder if we should pick something like `-1` as a default to indicate that this is not an actual value.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
hawkinsw: I did not include the unit in the original underlying code for this and submitted it for review. Ehsan requested that I encode the unit in the name and I actually thought that was a good idea. I definitely agree that we should rename the method to include something about `load` and I will do that. I propose `onElapsedLoadTimeMS`?

components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
hawkinsw: As far as I understand it, this is _the_ path from the underlying browser engine to the upper-level components. That is why I am exposing this on `Session`. The `Session` that receives these notifications can correlate that with the page that was loaded and associate the incoming data with the appropriate page load for generating the proper metrics. At least that was the goal. I am assuming that, given your comment, there is another path for surfacing this data that you would prefer that I use? I am definitely deferring to your expertise and was just trying to do _the right thing_ as far as I could grok the code!

components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
hawkinsw: This is definitely not a bad idea! I will make that change! Thank you for the suggestion!

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
hawkinsw: Check the new version of the patch!

components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
hawkinsw: Resolved in the latest version of the patch!

hawkinsw:
> Can you help us understand when and why this was added? Without a use case / reason it's hard to review this patch accordingly. Ideally we'd have an issue filed here. :)
Added an issue! See #3774

hawkinsw:
I am closing this because we had to refactor some code at a lower level that is going to mean we need to have a slightly different set of methods added to ac. I will reopen a PR to address #3774 with some different code as soon as it's ready.

PR #3767, opened 2019-07-15
 Don't overwrite existing value/running timer with raw API

badboy:
This adds tests to verify the behavior [we discussed in bug 1562859](https://bugzilla.mozilla.org/show_bug.cgi?id=1562859).

### Pull Request checklist


components/service/glean/src/main/java/mozilla/components/service/glean/private/TimespanMetricType.kt
```diff
@@ -121,6 +121,16 @@ data class TimespanMetricType(
             return
         }
 
+        if (timerId != null) {
```
travis79: The function description comment may need to be edited now.  I'm not sure the "Care should be taken..." bit really applies with this protection added.  Maybe replace it with a note about the new (correct) behavior?

Issue #2747, opened 2019-04-15
Add a better API for Custom Tabs


jonalmeida:
We currently rely on the `sesionId` being passed to us via the `IntentProcessor`. This can be confusing and we should probably come up with a better API.

jonalmeida:
I think it's fine keeping this closed. The feature makes more sense now with a BrowserStore.

PR #3776, opened 2019-07-15
Issue #2747 - Split intent processors 

NotWoods:
This PR separates the `IntentProcessor` class into different classes with a shared interface. In the future we plan to add intent processors for progressive web apps and private mode, in addition to the existing standard browsing and custom tab intent processors. 

I've kept the legacy version to preserve backwards compatibility. 

### Pull Request checklist

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/intent/IntentExtensions.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/intent/IntentExtensions.kt
@@ -1,0 +1,40 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.session.intent
+
+import android.content.Intent
+import mozilla.components.support.utils.SafeIntent
+
+const val EXTRA_SESSION_ID = "activeSessionId"
+
+/**
+ * Retrieves [mozilla.components.browser.session.Session] ID from the intent.
+ *
+ * @return The session ID previously added with [putSessionId],
+ * or null if no ID was found.
+ */
+fun Intent.getSessionId(): String? = getStringExtra(EXTRA_SESSION_ID)
+
+/**
+ * Retrieves [mozilla.components.browser.session.Session] ID from the intent.
+ *
+ * @return The session ID previously added with [putSessionId],
+ * or null if no ID was found.
+ */
+fun SafeIntent.getSessionId(): String? = getStringExtra(EXTRA_SESSION_ID)
+
+/**
+ * Add [mozilla.components.browser.session.Session] ID to the intent.
+ *
+ * @param sessionId The session ID data value.
+ *
+ * @return Returns the same Intent object, for chaining multiple calls
+ * into a single statement.
+ *
+ * @see [getSessionId]
+ */
+fun Intent.putSessionId(sessionId: String?): Intent {
+    return putExtra(EXTRA_SESSION_ID, sessionId)
+}
--- a/components/feature/customtabs/src/test/java/mozilla/components/feature/customtabs/AbstractCustomTabsServiceTest.kt
+++ b/components/feature/customtabs/src/test/java/mozilla/components/feature/customtabs/AbstractCustomTabsServiceTest.kt
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-package mozilla.components.browser.session.tab
+package mozilla.components.feature.customtabs
 
 import android.net.Uri
 import android.os.Bundle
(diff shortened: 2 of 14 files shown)


pocmo:
Mh, from the issue ("This can be confusing") and PR ("does a little too much") I'm not really sure what the issue and solution is.

NotWoods:
Updated description. I discussed this change with @jonalmeida offline but neglected to elaborate in the description.

components/feature/intent/src/main/java/mozilla/components/feature/intent/IntentProcessor.kt
```diff
@@ -39,71 +36,28 @@ class IntentProcessor(
...
-    }
-
-    private fun createSession(url: String, private: Boolean = false, source: Source): Session {
-        return if (openNewTab) {
-            Session(url, private, source).also { sessionManager.add(it, selected = true) }
-        } else {
-            sessionManager.selectedSession ?: Session(url, private, source)
-        }
-    }
+) : mozilla.components.browser.session.intent.IntentProcessor {
```
jonalmeida: nit: We probably want to rename the interface or the implementation so they don't have the same name. Maybe the interface could be `IntentProcessHandler` since we have multiple of these handlers?

components/feature/customtabs/src/test/java/mozilla/components/feature/customtabs/AbstractCustomTabsServiceTest.kt
```diff
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-package mozilla.components.browser.session.tab
+package mozilla.components.feature.customtabs
```
jonalmeida: Nice!

components/feature/intent/src/main/java/mozilla/components/feature/intent/IntentProcessor.kt
```diff
@@ -39,71 +36,28 @@ class IntentProcessor(
     private val useDefaultHandlers: Boolean = true,
     private val openNewTab: Boolean = true,
     private val isPrivate: Boolean = false
```
jonalmeida: Looking at the way Fenix uses [the intent processor](https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/components/Utilities.kt#L28-L34) they create two separate ones for private/regular tabs, so that would mean separate processors for custom tabs/PWAs in private/regular states as well.

For this reason I originally thought it would be nice to separate out the `private` flag to be a separate processor, but I like this direction a lot. I wonder if @pocmo is aware of any pitfalls from having two instances of an IntentProcessor with each having their own CT/PWA processors.

components/feature/intent/src/main/java/mozilla/components/feature/intent/IntentProcessor.kt
```diff
@@ -141,6 +95,6 @@ class IntentProcessor(
     }
 
     companion object {
-        const val ACTIVE_SESSION_ID = "activeSessionId"
+        const val ACTIVE_SESSION_ID = EXTRA_SESSION_ID
```
jonalmeida: Do we still need this `ACTIVE_SESSION_ID`?

components/browser/session/src/main/java/mozilla/components/browser/session/intent/IntentExtensions.kt
```diff
@@ -0,0 +1,33 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
jonalmeida: 👏 👏 👏

components/browser/session/src/main/java/mozilla/components/browser/session/intent/IntentExtensions.kt
```diff
@@ -0,0 +1,33 @@
...
+const val EXTRA_SESSION_ID = "activeSessionId"
+
+/**
+ * Retrieves [mozilla.components.browser.session.Session] ID from the intent.
+ *
+ * @return The session ID previously added with [putSessionId],
+ * or null if no ID was found.
+ */
+fun Intent.getSessionId() {
+    getStringExtra(EXTRA_SESSION_ID)
```
jonalmeida: Needs a `return` or use the `=` expression right?

components/feature/intent/src/main/java/mozilla/components/feature/intent/IntentProcessor.kt
NotWoods: Long term I want to update Fenix to use two `TabIntentProcessors` + other processors and entirely get rid of the fat `IntentProcessor`.

components/feature/intent/src/main/java/mozilla/components/feature/intent/IntentProcessor.kt
NotWoods: Yes, until we can remove references to it in Fenix.

components/browser/session/src/main/java/mozilla/components/browser/session/intent/IntentExtensions.kt
NotWoods: Oops! Thanks

components/feature/intent/src/main/java/mozilla/components/feature/intent/IntentProcessor.kt
jonalmeida: Oh perfect! That's along the similar lines that I was thinking. ➕ 💯

NotWoods:
I started working on `WebAppIntentProcessor` in another branch and decided to use suspend functions for `process` instead. 

Additionally added a `matches` field to cover the common use case of adding custom tab/pwa specific activities to the intent after `process` is called.

PR #3777, opened 2019-07-15
Closes #3528: Validate (initial) Store state

### LINKED ISSUE #3528: browser-state: BrowserStore: Validate initial state
Since you can construct and pass in a custom `BrowserState` to `BrowserStore` we should do some basic validation that the state is valid; e.g. if `selectedTabId` is set then this tab should exist; no duplicates (#3524).

csadilek:
Added this to `Store` directly so all stores can inherit it if they need state validation. Also figured it's useful to not just validate the initial state but also the state after every action (after the reducer has run). We have precondition checks, but this way, if we ever miss a check, we'd still validate and fail...

Might be better to land this after the release tomorrow though if we want this?


components/lib/state/src/main/java/mozilla/components/lib/state/Store.kt
```diff
@@ -65,6 +65,17 @@ open class Store<S : State, A : Action>(
...
+    /**
+     * Validates the provided or current [State]. This method will be executed on every
+     * state change and must be overridden by subtypes (concrete [Store]
+     * implementations) to have an effect. By default (if not overridden), all
+     * states are considered valid.
+     *
+     * @return the provided state, if valid.
+     * @throws [StateValidationFailedException] if the state is invalid.
+     */
+    open fun validateState(state: S = currentState) = state
```
pocmo: Are you sure this is something we are going to need? I can see the need for validating an initial `BrowserState` since we are not in control of that and can guard against this in tests.
But for the reduced state - Isn't that what our unit tests cover and wouldn't we end up running test code in production here. Or asked differently: Why do this after every reduce and not in a test?

csadilek: @pocmo This isn't meant to replace tests, of course :). It just guarantees that we fail if we're in an inconsistent state (have a bug), which then needs a fix and a test. It's like an integrity task running in a db or similar.

It may be excessive? I can remove but I guess we can keep the generic `validateState` that subtypes need to provide.

pocmo: With the store pattern I think there's only a limited set of inputs, which means we can cover all of them in a test:
- We can and want to cover all actions/reducers
- We can't really test all input states - but at least those pieces of the state that the reducer actually looks at.

For the bug part specially for browser-state (I thought I filed an issue for that, but I can't find it right now): I definitely was thinking about having a validate check in `SessionManager` to make sure `SessionManager` and `BrowserStore` are roughly in sync. Here I definitely would want to catch issues early on.

> I can remove but I guess we can keep the generic validateState that subtypes need to provide

I feel like we should be very conservative adding public API to the store since we are still in an early stage and those things could make it hard to iterate. But this one specifically is not that critical maybe.

csadilek: @pocmo OK updated. I removed the generic validation after `reduce` and just kept the initial state validation for browser-store. I do still think it's a useful feature to have to be able to validate the store/state all the time, but it should probably be configurable and we can add it in a follow-up if we want it.

csadilek:
I've added a commit here to fix https://github.com/mozilla-mobile/android-components/issues/3794 as it's related and was exposed by the previous validation logic (after reduce.).

csadilek:
@pocmo removed the commit to address the selection issue. So, I think, this ready now :).

PR #3783, opened 2019-07-16
Migrate feature-toolbar to use browser-state.

### LINKED ISSUE #3530: Migrate feature-toolbar to use browser-state
In Whistler we decided that this would be a good target for a migration since the scope is quite small and since it wouldn't affect anything else much.

pocmo:
Closes #3530.

### Pull Request checklist

### DIFF
--- a/components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
+++ b/components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
@@ -5,13 +5,19 @@
 package mozilla.components.feature.toolbar
 
 import androidx.annotation.VisibleForTesting
-import mozilla.components.browser.session.SelectionAwareSessionObserver
-import mozilla.components.browser.session.Session
-import mozilla.components.browser.session.SessionManager
-import mozilla.components.concept.engine.content.blocking.Tracker
+import androidx.annotation.VisibleForTesting.PRIVATE
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.flow.map
+import mozilla.components.browser.state.selector.findCustomTabOrSelectedTab
+import mozilla.components.browser.state.state.SessionState
+import mozilla.components.browser.state.store.BrowserStore
 import mozilla.components.concept.toolbar.Toolbar
-import mozilla.components.feature.toolbar.internal.URLRenderer
 import mozilla.components.concept.toolbar.Toolbar.SiteTrackingProtection
+import mozilla.components.feature.toolbar.internal.URLRenderer
+import mozilla.components.lib.state.ext.flowScoped
+import mozilla.components.support.ktx.kotlinx.coroutines.flow.ifChanged
 
 /**
  * Presenter implementation for a toolbar implementation in order to update the toolbar whenever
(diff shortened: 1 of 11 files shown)


components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
```diff
@@ -18,77 +23,59 @@ import mozilla.components.feature.toolbar.internal.URLRenderer
...
 
     @VisibleForTesting
     internal var renderer = URLRenderer(toolbar, urlRenderConfiguration)
 
+    private var subscription: Store.Subscription<BrowserState, BrowserAction>? = null
+
     /**
      * Start presenter: Display data in toolbar.
      */
     fun start() {
```
NotWoods: Should this implement `LifecycleAwareFeature` if `start` and `stop` are used?

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
```diff
@@ -18,77 +23,59 @@ import mozilla.components.feature.toolbar.internal.URLRenderer
...
+        renderer.post(tab.content.url)
 
-        renderer.post(session?.url ?: "")
+        toolbar.setSearchTerms(tab.content.searchTerms)
+        toolbar.displayProgress(tab.content.progress)
 
-        toolbar.setSearchTerms(session?.searchTerms ?: "")
-        toolbar.displayProgress(session?.progress ?: 0)
-        updateToolbarSecurity(session?.securityInfo ?: Session.SecurityInfo())
+        when (tab.content.securityInfo.secure) {
```
NotWoods: Should just be an if statement.
```suggestion
        toolbar.siteSecure = if (tab.content.securityInfo.secure) SECURE else INSECURE
```

components/lib/state/src/main/java/mozilla/components/lib/state/helpers/Helpers.kt
```diff
@@ -8,34 +8,43 @@ import mozilla.components.lib.state.Observer
...
+ * callback [then] only if the value has changed from the last mapped value. If rhe mapped value
+ * changed to `null` then the optional [otherwise] callback gets invoked.
  *
  * @param onMainThread Whether or not the [then] function should be executed on the main thread.
  * @param map A function that maps [State] to the value we want to observe for changes.
  * @param then Function that gets invoked whenever the mapped value changed.
+ * @param otherwise Optional function that gets invoked whenever the mapped value is `null`.
  */
 @Suppress("FunctionNaming")
 fun <S : State, T> onlyIfChanged(
```
NotWoods: I feel like this function is getting too complex. We should have separate `onlyIfChanged` and `map` functions that can be composed.

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
pocmo: The actual feature does and delegates to its sub components like the presenter and interactor.

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
pocmo: :+1:

components/lib/state/src/main/java/mozilla/components/lib/state/helpers/Helpers.kt
pocmo: I agree. @csadilek and I wanted to come back to this function at some point and think of better approaches. We haven't filed anything for that though - so I did that now:
https://github.com/mozilla-mobile/android-components/issues/3786

components/lib/state/src/main/java/mozilla/components/lib/state/helpers/Helpers.kt
```diff
@@ -8,34 +8,43 @@ import mozilla.components.lib.state.Observer
 import mozilla.components.lib.state.State
 
 /**
- * Creates an [Observer] that will map the received [State] to [T] (using [map]) and will invoke the callback
- * [then] only if the value has changed from the last mapped value.
+ * Creates an [Observer] that will map the received [State] to [T] (using [map]) and will invoke the
+ * callback [then] only if the value has changed from the last mapped value. If rhe mapped value
```
csadilek: nit: typo "the mapped value".

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
```diff
@@ -18,77 +23,60 @@ import mozilla.components.feature.toolbar.internal.URLRenderer
...
+    private val store: BrowserStore,
+    private val customTabId: String? = null,
     urlRenderConfiguration: ToolbarFeature.UrlRenderConfiguration? = null
-) : SelectionAwareSessionObserver(sessionManager) {
+) {
 
     @VisibleForTesting
     internal var renderer = URLRenderer(toolbar, urlRenderConfiguration)
 
+    private var subscription: Store.Subscription<BrowserState, BrowserAction>? = null
```
csadilek: Should we add a typealias `BrowserStoreSubscription`. Wdyt?

pocmo:
Adding blocked label: Needs update.

pocmo:
@csadilek I wonder if we should still land this. `onlyIfChanged` does exactly what we tried to avoid: Launch multiple coroutines using the main dispatcher. Technically we can't really guarantee that they will happen in order.

Maybe we should look at getting rid of `onlyIfChanged` first. With our new channel API we may be able to come up with something better. I prototyped and commented a bit in #3786.

csadilek:
> Maybe we should look at getting rid of `onlyIfChanged` first. With our new channel API we may be able to come up with something better. I prototyped and commented a bit in #3786.

@pocmo ah yes, I agree! Let's land https://github.com/mozilla-mobile/android-components/issues/3786 which will improve integration.

pocmo:
@csadilek Updated the PR. `feature-toolbar` is now using the channel-based API and I re-implemented `onlyIfChanged()` as extension method on the channel `consumeEachIfChanged()`. What do you think about this approach? It's not ready for landing yet though, some tests and docs are missing.

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
```diff
@@ -18,77 +25,62 @@ import mozilla.components.feature.toolbar.internal.URLRenderer
...
-    override fun stop() {
-        super.stop()
-
-        renderer.stop()
+        scope.launch {
+            store.channel()
+                .map { state -> state.findCustomTabOrSelectedTab(customTabId) }
+                .consumeEachIfChanged(
+                    then = { render(it) },
+                    otherwise = { clear() }
```
csadilek: I think `otherwise` doesn't read right here. It sounds like `if changed to do this and otherwise that`, but it's really covering the null case, right? Should we rename this to `onNull` or `ifNull`, or add a separate method for handling the null case?

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
```diff
@@ -18,77 +25,62 @@ import mozilla.components.feature.toolbar.internal.URLRenderer
...
-    override fun onSessionSelected(session: Session) {
-        super.onSessionSelected(session)
-        initializeView()
-    }
-
-    override fun onAllSessionsRemoved() {
-        initializeView()
-    }
+    fun stop() {
+        scope.cancel()
```
csadilek: This will cancel the scope and its job and make it unusable, if I understand correctly. So, `onStart` wouldn't work anymore (we wouldn't be able to "launch" on the scope anymore). So far, we always had to create a new job because of this: https://github.com/mozilla-mobile/android-components/blob/master/samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/MainActivity.kt#L59

pocmo:
@csadilek This PR is back and updated! :)

* Uses the new `Flow` extensions of `Store`
* Uses the new `SiteTracking` things that have been added in the meantime.

pocmo:
bors r=csadilek

pocmo:
bors retry

pocmo:
bors retry

pocmo:
bors retry

PR #3792, opened 2019-07-16
Fixes #1884 - Move intent code to feature-customtabs

### LINKED ISSUE #1884: Revert CustomTabsConfig workaround Jetifier
Context: https://github.com/mozilla-mobile/reference-browser/issues/487

We've found that we need to introduce a small hack in order to stop the jetifier tool from re-writing CustomTabsIntent strings.

Once `1.0.0-beta03` of `jetifier-processor` is published *and* we add it to our classpath, as mentioned in the google tracker [issue 119183822](https://issuetracker.google.com/issues/119183822), we can then go about removing this hack.

NotWoods:
- Make `CustomTabConfig`'s constructor public.
- Move the `customTabConfigFromIntent` function out of browser-session.

For PWAs, we want to manually build `CustomTabConfig` instances so PWAs can have their own custom tabs. Since the constructor is now public, I'm moving the config-from-intent function into feature-customtabs instead of browser-session. This consolidates the custom tab code into the feature rather than having parts of it across many components. 

### Pull Request checklist

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/tab/CustomTabConfig.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/tab/CustomTabConfig.kt
@@ -8,30 +8,60 @@ import android.app.PendingIntent
 import android.graphics.Bitmap
 import android.os.Bundle
 import android.os.Parcelable
-import androidx.browser.customtabs.CustomTabsIntent
 import android.util.DisplayMetrics
 import androidx.annotation.ColorInt
-
+import androidx.browser.customtabs.CustomTabsIntent
 import mozilla.components.support.utils.SafeBundle
 import mozilla.components.support.utils.SafeIntent
 import java.util.ArrayList
-import java.util.Collections
 import java.util.UUID
 
 /**
- * Holds configuration data for a Custom Tab. Use [createFromIntent] to
- * create instances.
+ * Holds configuration data for a Custom Tab.
+ *
+ * @property toolbarColor Background color for the toolbar.
+ * @property closeButtonIcon Custom icon of the back button on the toolbar.
+ * @property enableUrlbarHiding Enables the toolbar to hide as the user scrolls down on the page.
+ * @property actionButtonConfig Custom action button on the toolbar.
+ * @property showShareMenuItem Specifies whether a default share button will be shown in the menu.
+ * @property menuItems Custom overflow menu items.
+ * @property exitAnimations Bundle containing custom exit animations for the tab.
+ * @property titleVisible Whether the title should be shown in the custom tab.
  */
-class CustomTabConfig internal constructor(
+data class CustomTabConfig(
     val id: String,
     @ColorInt val toolbarColor: Int?,
     val closeButtonIcon: Bitmap?,
-    val disableUrlbarHiding: Boolean,
+    val enableUrlbarHiding: Boolean,
     val actionButtonConfig: CustomTabActionButtonConfig?,
     val showShareMenuItem: Boolean,
-    val menuItems: List<CustomTabMenuItem>,
-    val options: List<String>
+    val menuItems: List<CustomTabMenuItem> = emptyList(),
+    val exitAnimations: Bundle? = null,
+    val titleVisible: Boolean = false
 ) {
+    inline val disableUrlbarHiding
+        get() = !enableUrlbarHiding
+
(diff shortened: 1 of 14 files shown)


pocmo:
> For PWAs, we want to manually build CustomTabConfig instances so PWAs can have their own custom tabs.

Are you sure about this? Wouldn't it be more stable to dispatch a custom tab intent on the app itself? This will go through all the regular custom tabs flow and avoids you needing to know how to actually launch into this custom tab. Also think of things like additional things the app may do (e.g. telemetry for custom tabs that come in).

NotWoods:
We're just using the custom tab config to set up `CustomTabToolbarFeature`, so I'm trying to avoid the extra step of constructing an intent just to make a data class.

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabIntentProcessor.kt
```diff
@@ -22,12 +21,11 @@ import mozilla.components.support.utils.SafeIntent
 class CustomTabIntentProcessor(
     private val sessionManager: SessionManager,
     private val loadUrlUseCase: SessionUseCases.DefaultLoadUrlUseCase,
-    private val displayMetrics: DisplayMetrics
+    private val resources: Resources
 ) : IntentProcessor {
 
     override fun matches(intent: Intent): Boolean {
-        val safeIntent = SafeIntent(intent)
```
jonalmeida: We should keep this since activities can make direct calls to `matches` and we should use a `SafeIntent` for that.

components/support/utils/src/main/java/mozilla/components/support/utils/SafeBundle.kt
```diff
@@ -15,31 +15,32 @@ import mozilla.components.support.base.log.logger.Logger
...
+
+    fun getString(name: String): String? =
+        safeAccess { getString(name) }
+
+    fun <T : Parcelable> getParcelable(name: String): T? =
+        safeAccess { getParcelable<T>(name) }
 
     @SuppressWarnings("TooGenericExceptionCaught")
-    fun <T : Parcelable> getParcelable(name: String): T? {
+    private fun <T> safeAccess(default: T? = null, block: Bundle.() -> T): T? {
```
jonalmeida: ⭐️ ⭐️ 

First one is for making a nice reusable function. Second is for making it private because that `T` is too powerful.

Also, tests? 🙃

components/browser/session/src/main/java/mozilla/components/browser/session/tab/CustomTabConfig.kt
```diff
@@ -8,30 +8,65 @@ import android.app.PendingIntent
...
+    val enableUrlbarHiding: Boolean,
     val actionButtonConfig: CustomTabActionButtonConfig?,
     val showShareMenuItem: Boolean,
     val menuItems: List<CustomTabMenuItem>,
-    val options: List<String>
+    options: List<String>? = null,
+    val exitAnimations: Bundle? = null,
+    val titleVisible: Boolean = false
 ) {
+    inline val disableUrlbarHiding
```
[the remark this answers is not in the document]
NotWoods: This was just to avoid breaking the API. I'll remove it.

docs/changelog.md
```diff
@@ -17,6 +17,18 @@ permalink: /changelog/
 * **browser-icons**
   * Added `BrowserIcons.loadIntoView` to automatically load an icon into an `ImageView`.
 
+* **browser-session**
+  * Added `IntentProcessor` interface to represent a class that processes intents to create sessions.
+  * Deprecated `CustomTabConfig.isCustomTabIntent` and `CustomTabConfig.createFromIntent`. Use `isCustomTabIntent` and `createFromCustomTabIntent` in feature-customtabs instead.
```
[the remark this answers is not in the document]
NotWoods: As far as I know, we don't have any deprecation strategy yet.

components/support/utils/src/main/java/mozilla/components/support/utils/SafeBundle.kt
NotWoods: Updated with tests

NotWoods:
Let's merge mozilla-mobile/fenix#4220 before this PR so the `@Deprecated` annotations don't break fenix.

PR #3798, opened 2019-07-17

Pike:
State: mozilla-l10n/android-l10n@e521a2aab2cc7d45f47523f7c620cb5c4228ff89

This also adds new locales to the config and the first localizations for Swedish.

CC @Delphine


pocmo:
Uplifting this to the release branch for 4.0.1 now.

PR #3801, opened 2019-07-17
Enable the use of `-` for labels in the Glean SDK

Dexterp37:
This is part of the change required to support a long term solution for mozilla-mobile/fenix#4082

This requires mozilla-services/mozilla-pipeline-schemas#356 to be merged.

**Important**: mozilla/glean_parser#73 is not blocking this, as it only deals with 'static' labels. A new version of the glean_parser must be cut and referenced in the build script for Glean, but that will happen **as a separate PR to NOT be uplifted**.

### Pull Request checklist

### After merge


mdboom:
This will also need an update to `mozilla-pipeline-schemas` [here](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/00ed704cad528b45bfbe567595e8c63168fe936a/templates/include/glean/labeled_group.1.schema.json#L4)

components/service/glean/src/main/java/mozilla/components/service/glean/private/LabeledMetricType.kt
```diff
@@ -38,9 +38,9 @@ data class LabeledMetricType<T>(
         private const val MAX_LABELS = 16
         private const val OTHER_LABEL = "__other__"
 
-        // This regex is used for matching against labels and should allow for dots and/or
-        // snake_case.
-        private val labelRegex = Regex("^[a-z_][a-z0-9_]{0,29}(\\.[a-z0-9_]{0,29})*$")
+        // This regex is used for matching against labels and should allow for dots,
+        // snake_case and/or hyphens.
```
travis79: One minor nit here, it's calling out allowing `_` by using the term "snake_case", and allowing `-` by saying "hyphens".  It also doesn't call out that the starting character is restricted to a letter or an underscore.  It might be a little more clear if we reworded this to something like:

```
// This regex is used for matching against labels and should allow for dots, underscores,
// and/or hyphens.  Labels are also limited to starting with either a letter or an
// underscore character.
```

PR #3816, opened 2019-07-18
Closes #2293 - Open custom tabs when leaving scope

### LINKED ISSUE #2292: PWA: Support scope handling

### LINKED ISSUE #2293: PWA: Open custom tab for links that leave scope
Depends on #2292

NotWoods:
Refactors PWA code so that it uses a BrowserFragment subclass. When leaving the scope defined in the web app manifest, the custom tab toolbar is displayed. When returning, the toolbar is shown again.

Also closes #2292

### Pull Request checklist

### After merge


components/concept/engine/src/main/java/mozilla/components/concept/engine/manifest/WebAppManifestParser.kt
```diff
@@ -90,7 +92,11 @@ class WebAppManifestParser {
     }
 }
 
-private const val HEX_RADIX = 16
+fun WebAppManifestParser.Result.get(): WebAppManifest? = when (this) {
+    is WebAppManifestParser.Result.Success -> manifest
+    is WebAppManifestParser.Result.Failure -> null
+}
+
```
pocmo: nit: Public function, consider adding kdoc.

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/ext/Bundle.kt
```diff
@@ -0,0 +1,23 @@
...
+
+fun Bundle.putWebAppManifest(webAppManifest: WebAppManifest) {
+    val json = WebAppManifestParser().serialize(webAppManifest)
+    putString(EXTRA_WEB_APP_MANIFEST, json.toString())
+}
+
+fun Bundle.getWebAppManifest(): WebAppManifest? {
+    return getString(EXTRA_WEB_APP_MANIFEST)?.let { json ->
+        WebAppManifestParser().parse(json).get()
+    }
```
pocmo: nit: Public function, consider adding kdoc.

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/ext/Bundle.kt
```diff
@@ -0,0 +1,23 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.feature.pwa.ext
+
+import android.os.Bundle
+import mozilla.components.concept.engine.manifest.WebAppManifest
+import mozilla.components.concept.engine.manifest.WebAppManifestParser
+import mozilla.components.concept.engine.manifest.get
+
+const val EXTRA_WEB_APP_MANIFEST = "mozilla.components.feature.pwa.EXTRA_WEB_APP_MANIFEST"
```
pocmo: Can this be internal to avoid that being visible everywhere?

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/ext/Intent.kt
```diff
@@ -0,0 +1,18 @@
...
+import mozilla.components.concept.engine.manifest.WebAppManifestParser
+
+fun Intent.putWebAppManifest(webAppManifest: WebAppManifest) {
+    val json = WebAppManifestParser().serialize(webAppManifest)
+    putExtra(EXTRA_WEB_APP_MANIFEST, json.toString())
+}
+
+fun Intent.getWebAppManifest(): WebAppManifest? {
+    return extras?.getWebAppManifest()
+}
```
pocmo: nit: Public function, consider adding kdoc.

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/feature/WebAppToolbarFeature.kt
```diff
@@ -0,0 +1,56 @@
...
+    private val manifest: WebAppManifest?
+) : Session.Observer, LifecycleAwareFeature {
+
+    init {
+        toolbar.isGone = true
+    }
+
+    override fun onUrlChanged(session: Session, url: String) {
+        toolbar.isGone = isInScope(url.toUri())
+    }
```
pocmo: This setup seems to be fragile to me. I wouldn't want this to be an actual full-featured browser fragment where we just hide the toolbar and then show it again. I think we should *actually* switch to a custom tab.

A browser may have more UI that the component does not know about and the component can't control; e.g. the little drawer above the toolbar in Fenix.

You also may want the freedom to configure Custom Tabs / PWAs differently and *not* use the same fragment configuration for both.

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/intent/WebAppIntentProcessor.kt
```diff
@@ -0,0 +1,53 @@
...
+        val url = intent.toSafeIntent().dataString
+
+        return if (!url.isNullOrEmpty() && matches(intent)) {
+            val webAppManifest = storage.loadManifest(url) ?: return false
+
+            val session = Session(url, private = false, source = Source.HOME_SCREEN)
+            session.webAppManifest = webAppManifest
+            session.customTabConfig = webAppManifest.toCustomTabConfig()
+
+            sessionManager.add(session)
```
pocmo: By adding the custom tab config here we are piggy backing on the fact that custom tabs are excluded from most of the browser UI. I think we should keep them as completely separate entities that may be handled differently and that may have different lifetimes. But I also agree that we shouldn't start dealing with that in `browser-session` - but definitely in `browser-state`.

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/feature/WebAppActivityFeature.kt
```diff
@@ -0,0 +1,63 @@
...
+    private val manifest: WebAppManifest?
+) : LifecycleObserver {
+
+    private val scope = MainScope()
+
+    @OnLifecycleEvent(Lifecycle.Event.ON_RESUME)
+    fun onResume() {
+        if (manifest?.display == WebAppManifest.DisplayMode.FULLSCREEN) {
+            activity.enterToImmersiveMode()
+        }
```
pocmo: Isn't immersive mode something you can exit? That's something we would not want to do for a fullscreen web app, right?

NotWoods: I don't think you can turn off immersive mode as a user, just pull up the navigation bar so you can leave the app. I'll double check.

NotWoods:
Updated so that custom tabs and PWAs use the same fragment. The hide toolbar feature has been adjusted to allow for multiple trusted scopes, to be compatible with TWAs later (which can specify multiple scopes).

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/feature/WebAppHideToolbarFeature.kt
```diff
@@ -0,0 +1,68 @@
...
+ * @param sessionId ID of the custom tab session.
+ * @param trustedScopes Scopes to hide the toolbar at.
+ * Scopes correspond to [WebAppManifest.scope]. They can be a path (PWA) or just an origin (TWA).
+ */
+class WebAppHideToolbarFeature(
+    private val sessionManager: SessionManager,
+    private val toolbar: View,
+    private val sessionId: String,
+    private val trustedScopes: List<Uri>
+) : Session.Observer, LifecycleAwareFeature {
```
pocmo: nit: In general I think it's cleaner to implement the observer interfaces internally in a feature - like we do in some others, but not all unfortunately. The main reason for that is that it won't pollute the public API of the feature with observer methods that make no sense to the developer using the feature in an app and that no other code should call anyways.

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/intent/WebAppIntentProcessor.kt
```diff
@@ -0,0 +1,53 @@
...
+import mozilla.components.browser.session.intent.IntentProcessor
+import mozilla.components.browser.session.intent.putSessionId
+import mozilla.components.concept.engine.EngineSession
+import mozilla.components.feature.pwa.ManifestStorage
+import mozilla.components.feature.pwa.ext.putWebAppManifest
+import mozilla.components.feature.pwa.ext.toCustomTabConfig
+import mozilla.components.feature.session.SessionUseCases
+import mozilla.components.support.utils.toSafeIntent
+
+class WebAppIntentProcessor(
```
pocmo: nit: Maybe add some kdoc to this class.

samples/browser/src/main/java/org/mozilla/samples/browser/ExternalAppBrowserActivity.kt
```diff
@@ -0,0 +1,30 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.samples.browser
+
+import androidx.core.net.toUri
+import androidx.fragment.app.Fragment
+import mozilla.components.concept.engine.manifest.WebAppManifest
+import mozilla.components.feature.pwa.ext.getWebAppManifest
+
+class ExternalAppBrowserActivity : BrowserActivity() {
```
pocmo: nit: This class deserves a comment explaining a bit what an "external app browser" is :)

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/feature/WebAppHideToolbarFeature.kt
NotWoods: I'll keep that in mind in the future. In this case I think its fine since we only have one method from the session observer, and it represents the core functionality of the feature.

pocmo: But you will still inherit all the methods with default implementation :)

PR #3821, opened 2019-07-18
Add a way to record metrics to Glean when Glean is not initialized

travis79:
This is what I was thinking as a possible solution to this [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1566917)

### Pull Request checklist

### After merge

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Dispatchers.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Dispatchers.kt
@@ -11,6 +11,8 @@ import kotlinx.coroutines.ObsoleteCoroutinesApi
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.newSingleThreadContext
 import kotlinx.coroutines.runBlocking
+import mozilla.components.support.base.log.logger.Logger
+import java.util.concurrent.ConcurrentLinkedQueue
 
 @ObsoleteCoroutinesApi
 internal object Dispatchers {
@@ -18,24 +20,44 @@ internal object Dispatchers {
         // When true, jobs will be run synchronously
         internal var testingMode = false
 
+        // When true, jobs will be queued and not ran until triggered by calling
+        // flushQueuedInitialTasks()
+        @Volatile
+        private var queueInitialTasks = true
+
+        // Use a [ConcurrentLinkedQueue] to take advantage of it's thread safety and no locking
+        internal val taskQueue: ConcurrentLinkedQueue<() -> Unit> = ConcurrentLinkedQueue()
+
+        private val logger = Logger("glean/Dispatchers")
+
+        companion object {
+            // This value was chosen in order to allow several tasks to be queued for execution but
+            // still be conservative of memory. This queue size is important for cases where
+            // setUploadEnabled(false) is not called so that we don't continue to queue tasks and
+            // waste memory.
+            const val MAX_QUEUE_SIZE = 100
+        }
+
         /**
-         * Launch a block of work asyncronously.
+         * Launch a block of work asynchronously.
          *
-         * If [setTestingMode] has enabled testing mode, the work will run
-         * synchronously.
+         * If [queueInitialTasks] is true, then the work will be queued and executed when
+         * [flushQueuedInitialTasks] is called.
          *
-         * @return [Job], or null if run synchronously.
+         * If [setTestingMode] has enabled testing mode, the work will run synchronously.  This is
+         * true regardless of whether [queueInitialTasks] is true or not.
+         *
+         * @return [Job], or null if queued or run synchronously.
          */
         fun launch(
             block: suspend CoroutineScope.() -> Unit
         ): Job? {
-            return if (testingMode) {
-                runBlocking {
-                    block()
+            return when {
+                queueInitialTasks -> {
+                    addTaskToQueue(block)
+                    null
                 }
-                null
-            } else {
-                coroutineScope.launch(block = block)
+                else -> executeTask(block)
             }
         }
 
@@ -63,6 +85,79 @@ internal object Dispatchers {
         @VisibleForTesting(otherwise = VisibleForTesting.NONE)
         fun setTestingMode(enabled: Boolean) {
             testingMode = enabled
+        }
+
+        /**
+         * Enable queueing mode, which causes tasks to be queued until launched by calling
+         * [flushQueuedInitialTasks].
+         *
+         * @param enabled whether or not to enable the testing mode
+         */
+        @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+        fun setTaskQueueing(enabled: Boolean) {
+            queueInitialTasks = enabled
+        }
+
+        /**
+         * Stops queueing tasks and processes any tasks in the queue. Since [queueInitialTasks] is
+         * set to false prior to processing the queue, newly launched tasks should be executed
+         * on the couroutine scope rather than added to the queue.
+         */
+        internal fun flushQueuedInitialTasks() {
+            // Setting this to false first should cause any new tasks to just be executed (see
+            // launch() above) making it safer to process the queue.
+            //
+            // NOTE: This has the potential for causing a task to execute out of order in certain
+            // situations. If a library or thread that runs before init happens to record
+            // between when the queueInitialTasks is set to false and the taskQueue finishing
+            // launching, then that task could be executed out of the queued order. See the bugzilla
+            // bug tracking this for more info: https://bugzilla.mozilla.org/show_bug.cgi?id=1568503
+            queueInitialTasks = false
+            taskQueue.forEach { task ->
+                task.invoke()
+            }
+            taskQueue.clear()
+        }
+
+        /**
+         * Helper function to add task to queue as either a synchronous or asynchronous operation,
+         * depending on whether [testingMode] is true.
+         */
+        private fun addTaskToQueue(block: suspend CoroutineScope.() -> Unit) {
+            if (taskQueue.size >= MAX_QUEUE_SIZE) {
+                logger.error("Exceeded maximum queue size, discarding task")
+                return
+            }
+
+            if (testingMode) {
+                logger.info("Task queued for execution in test mode")
+                taskQueue.add {
+                    runBlocking {
+                        block()
+                    }
+                }
+            } else {
+                logger.info("Task queued for execution and delayed until flushed")
+                taskQueue.add {
+                    coroutineScope.launch(block = block)
+                }
+            }
+        }
+
+        /**
+         * Helper function to execute the task as either an synchronous or asynchronous operation,
+         * depending on whether [testingMode] is true.
+         */
+        private fun executeTask(block: suspend CoroutineScope.() -> Unit): Job? {
+            return when {
+                testingMode -> {
+                    runBlocking {
+                        block()
+                    }
+                    null
+                }
+                else -> coroutineScope.launch(block = block)
+            }
         }
     }
 
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
@@ -85,6 +85,7 @@ open class GleanInternalAPI internal constructor () {
      * as shared preferences
      * @param configuration A Glean [Configuration] object with global settings.
      */
+    @Suppress("LongMethod")
     fun initialize(
         applicationContext: Context,
         configuration: Configuration = Configuration()
@@ -133,6 +134,10 @@ open class GleanInternalAPI internal constructor () {
         if (!Dispatchers.API.testingMode) {
             metricsPingScheduler.startupCheck()
         }
+
+        // Signal Dispatcher that init is complete
+        @Suppress("EXPERIMENTAL_API_USAGE")
+        Dispatchers.API.flushQueuedInitialTasks()
 
         // At this point, all metrics and events can be recorded.
         ProcessLifecycleOwner.get().lifecycle.addObserver(gleanLifecycleObserver)
--- a/components/service/glean/src/test/java/mozilla/components/service/glean/DispatchersTest.kt
+++ b/components/service/glean/src/test/java/mozilla/components/service/glean/DispatchersTest.kt
@@ -11,17 +11,17 @@ import org.junit.Assert.assertEquals
 import org.junit.Assert.assertNotSame
 import org.junit.Assert.assertSame
 
+@Suppress("EXPERIMENTAL_API_USAGE")
 class DispatchersTest {
 
     @Test
     fun `API scope runs off the main thread`() {
         val mainThread = Thread.currentThread()
         var threadCanary = false
-        @Suppress("EXPERIMENTAL_API_USAGE")
         Dispatchers.API.setTestingMode(false)
+        Dispatchers.API.setTaskQueueing(false)
 
         runBlocking {
-            @Suppress("EXPERIMENTAL_API_USAGE")
             Dispatchers.API.launch {
                 assertNotSame(mainThread, Thread.currentThread())
                 // Use the canary bool to make sure this is getting called before
@@ -31,9 +31,33 @@ class DispatchersTest {
             }!!.join()
         }
 
-        @Suppress("EXPERIMENTAL_API_USAGE")
         Dispatchers.API.setTestingMode(true)
         assertEquals(true, threadCanary)
         assertSame(mainThread, Thread.currentThread())
+    }
+
+    @Test
+    fun `launch() correctly adds tests to queue if queueTasks is true`() {
+        var threadCanary = 0
+
+        Dispatchers.API.setTestingMode(true)
+        Dispatchers.API.setTaskQueueing(true)
+
+        // Add 3 tasks to queue each one setting threadCanary to true to indicate if any task has ran
+        repeat(3) {
+            Dispatchers.API.launch {
+                threadCanary += 1
+            }
+        }
+
+        assertEquals("Task queue contains the correct number of tasks",
+            3, Dispatchers.API.taskQueue.size)
+        assertEquals("Tasks have not run while in queue", 0, threadCanary)
+
+        // Now trigger execution to ensure the tasks fired
+        Dispatchers.API.flushQueuedInitialTasks()
+
+        assertEquals("Tasks have executed", 3, threadCanary)
+        assertEquals("Task queue is cleared", 0, Dispatchers.API.taskQueue.size)
     }
 }
--- a/components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
+++ b/components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
@@ -15,6 +15,7 @@ import kotlinx.coroutines.runBlocking
 import mozilla.components.service.glean.GleanMetrics.GleanInternalMetrics
 import mozilla.components.service.glean.GleanMetrics.Pings
 import mozilla.components.service.glean.config.Configuration
+import mozilla.components.service.glean.private.CounterMetricType
 import mozilla.components.service.glean.private.DatetimeMetricType
 import mozilla.components.service.glean.private.EventMetricType
 import mozilla.components.service.glean.private.Lifetime
@@ -206,6 +207,7 @@ class GleanTest {
             sendInPings = listOf("store1")
         )
         Glean.initialized = false
+        Dispatchers.API.testingMode = false
         stringMetric.set("foo")
         assertNull(
             "Metrics should not be recorded if Glean is not initialized",
@@ -213,6 +215,38 @@ class GleanTest {
         )
 
         Glean.initialized = true
+    }
+
+    @Test
+    fun `queued recorded metrics correctly record during init`() {
+        val counterMetric = CounterMetricType(
+            disabled = false,
+            category = "telemetry",
+            lifetime = Lifetime.Application,
+            name = "counter_metric",
+            sendInPings = listOf("store1")
+        )
+
+        // First set to a pre-init state
+        Glean.initialized = false
+        Dispatchers.API.setTaskQueueing(true)
+
+        // This will queue 3 tasks that will add to the metric value once Glean is initialized
+        for (i in 0..2) {
+            counterMetric.add()
+        }
+
+        // Ensure that no value has been stored yet since the tasks have only been queued and not
+        // executed yet
+        assertFalse("No value must be stored", counterMetric.testHasValue())
+
+        // Calling resetGlean here will cause Glean to be initialized and should cause the queued
+        // tasks recording metrics to execute
+        resetGlean(clearStores = false)
+
+        // Verify that the callback was executed by testing for the correct value
+        assertTrue("Value must exist", counterMetric.testHasValue())
+        assertEquals("Value must match", 3, counterMetric.testGetValue())
     }
 
     @Test
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/private/CommonMetricData.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/private/CommonMetricData.kt
@@ -39,12 +39,6 @@ interface CommonMetricData {
     val identifier: String get() = if (category.isEmpty()) { name } else { "$category.$name" }
 
     fun shouldRecord(logger: Logger): Boolean {
-        // Don't record metrics if we aren't initialized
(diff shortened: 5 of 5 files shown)


components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
```diff
@@ -52,21 +52,12 @@ class GleanCrashReporterService(
...
             if (!checkFileConditions()) {
                 // checkFileConditions() internally logs error conditions
                 return@run
             }
 
-            // Parse the persisted crashes
             parseCrashFile()
-
-            // Clear persisted counts by deleting the file
-            file.delete()
```
Dexterp37: Is there a specific reason why you're moving this?

components/service/glean/src/main/java/mozilla/components/service/glean/Dispatchers.kt
```diff
@@ -34,6 +34,12 @@ internal object Dispatchers {
                     block()
                 }
                 null
+            } else if (!Glean.isInitialized()) {
```
Dexterp37: I like this approach! However, I feel like this should not depend on the `Glean`. Instead, we could:

- have the list of queued coroutines as an implementation detail of the `Dispatchers` object;
- provide a flag for changing the behaviour (as we do for `testingMode`; this would also make sure we don't need to flip the behaviour of other Glean tests as we could only flip the behaviour on for a specific test;
- control the Dispatcher's behaviour *within* `Glean.initialize` (so that the init logic is only in one place)
- we should somehow make sure this is thread-safe (enqueuing and dispatching the queue should not stumble upon each other);

components/service/glean/src/main/java/mozilla/components/service/glean/Dispatchers.kt
```diff
@@ -34,6 +34,12 @@ internal object Dispatchers {
...
+                coroutineScope.launch {
+                    Glean.queueTaskForInit {
```
Dexterp37: We should do two things here, IMHO: 

- log that the call is being dalayed, due to Glean not being initialized;
- add an upper bound to the number of possible enqueued calls.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -65,6 +65,8 @@ open class GleanInternalAPI internal constructor () {
 
     internal lateinit var pingStorageEngine: PingStorageEngine
 
+    internal val tasksQueuedBeforeInit: MutableList<() -> Unit> = mutableListOf()
```
Dexterp37: I wonder if there's any lock-free data structure, in Kotlin, to provide thread-safe access to this.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -118,6 +120,12 @@ open class GleanInternalAPI internal constructor () {
         // This must be set before anything that might trigger the sending of pings.
         initialized = true
 
+        // Process tasks that were queued before init and then clear the list
+        for (task in tasksQueuedBeforeInit) {
```
Dexterp37: This should probably be moved to `Dispatchers`, maybe `Dispatchers.flushDelayedCalls` or something like that.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -157,6 +165,22 @@ open class GleanInternalAPI internal constructor () {
...
 
+    /**
+     * Register a closure that will be called once Glean has been initialized.  If Glean is already
+     * initialized, then the callback will be executed immediately. This can be useful if using
+     * Glean from a library where Glean may not be initialized when the library needs to record
+     * metrics or interact with a part of the Glean API that requires it to be initialized.
+     *
+     * @param callback A function/closure that takes no parameters and returns nothing (Unit).
+     */
+    internal fun queueTaskForInit(callback: () -> Unit) {
```
Dexterp37: I think that if we move the logic to `Dispatchers`, this might not be needed anymore.

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
travis79: No, it was an artifact of the first commit before I made the queueing system internal to glean.  I was passing the `parseCrashFile()` function to Glean to execute as part of it's init and had moved the `file.delete()` there to ensure it wasn't deleted before it was read.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
travis79: [ConcurrentLinkedQueue](https://developer.android.com/reference/java/util/concurrent/ConcurrentLinkedQueue.html) is probably the best bet

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
travis79: Totally agree

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
travis79: This was just an artifact of starting this out with this function as part of the public API.  If we move the queue to Dispatchers.kt, then this is not needed here at all.

components/service/glean/src/main/java/mozilla/components/service/glean/Dispatchers.kt
```diff
@@ -64,6 +94,20 @@ internal object Dispatchers {
...
             testingMode = enabled
         }
+
+        /**
+         * Handle initialization complete and process any queued tasks
+         */
+        fun onInitComplete() {
+            // Setting this to false first should cause any new tasks to just be executed (see
+            // launch() above) making it safer to process the queue.
+            queueTasks = false
```
Dexterp37: If some task gets dispatched after this point but before the loop below, when would they be executed?

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -118,6 +119,9 @@ open class GleanInternalAPI internal constructor () {
         // This must be set before anything that might trigger the sending of pings.
         initialized = true
 
+        // Signal Dispatcher that init is complete
+        Dispatchers.API.onInitComplete()
```
Dexterp37: This should probably be at the end of this function, right before the comment for the `ProcessLifecycleOwner.get()...`. Before that point no metric is recorded anyway, they're just delayed in a less controlled way.

Your change here is honestly a *great* improvement for the Glean SDK on Android: we can now control dispatching things instead of relying on the fact that we're using a thread pool with a single consumer. We might even expand that pool, if needed! This wasn't possible before as we relied on the single thread to have more control on the init behaviour and make sure that nothing got recorded before metrics/events pings were generated ;)

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
```diff
@@ -215,6 +218,50 @@ class GleanTest {
...
+            category = "telemetry",
+            lifetime = Lifetime.Application,
+            name = "counter_metric",
+            sendInPings = listOf("store1")
+        )
+
+        // First set to a pre-init state
+        Glean.initialized = false
+        Dispatchers.API.queueTasks = true
+        Dispatchers.API.testingMode = false
```
Dexterp37: Why do we need `testingMode` to be false here? 

**Update** Ok I re-read the code and found out :) Left a comment above!

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
```diff
@@ -215,6 +218,50 @@ class GleanTest {
...
+
+        // This will queue 3 tasks that will add to the metric value once Glean is initialized
+        for (i in 0..2) {
+            counterMetric.add()
+        }
+
+        // Enable testing mode and ensure that no value has been stored yet
+        Glean.enableTestingMode()
```
Dexterp37: Can't we have testing mode enabled at the beginning of the test?

components/service/glean/src/main/java/mozilla/components/service/glean/Dispatchers.kt
```diff
@@ -11,31 +11,61 @@ import kotlinx.coroutines.ObsoleteCoroutinesApi
...
          */
+        @Suppress("LongMethod")
         fun launch(
             block: suspend CoroutineScope.() -> Unit
         ): Job? {
-            return if (testingMode) {
-                runBlocking {
-                    block()
+            return when {
+                testingMode -> {
```
Dexterp37: What if we wrap this somehow and make sure that enqueued tasks are also `testingMode` aware? This will make it more clearer in tests too (see my comments below)

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
```diff
@@ -215,6 +218,50 @@ class GleanTest {
...
+        assertFalse("No value must be stored", counterMetric.testHasValue())
+
+        // Calling resetGlean here will cause Glean to be initialized and should cause the queued
+        // tasks recording metrics to execute
+        resetGlean(clearStores = false)
+
+        // Wait for the queued tasks to run by checking the taskQueue.size  Using a timeout just in
+        // case something hangs up so the test won't stall.
+        runBlocking {
```
Dexterp37: If we make the queued tasks `testMode` aware, then this becomes much simpler and we won't require timeouts and checking `Dispatchers` internals at all.

components/service/glean/src/main/java/mozilla/components/service/glean/Dispatchers.kt
travis79: Yes, if launch() was called after queueTask=false and before the loop, then it would just be executed as normal.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -129,11 +130,13 @@ open class GleanInternalAPI internal constructor () {
         // If Glean is being initialized from a test, then we want to skip running the
         // startupCheck() as it can cause some intermittent test failures due to multi-
         // threaded race conditions.
-        @Suppress("EXPERIMENTAL_API_USAGE")
         if (!Dispatchers.API.testingMode) {
             metricsPingScheduler.startupCheck()
         }
 
+        // Signal Dispatcher that init is complete
+        Dispatchers.API.flushQueuedInitialTasks()
```
mdboom: So, I understand the underlying data structure is threadsafe and lock free, but what happens if we are iterating over it to clear it and another thread is adding to it?  Would those events fall on the floor?  Should we add a lock to synchronize against that?

mdboom: I guess that wouldn't be possible, since Glean.initialized is set to true above, so maybe it's ok.

Trying to think about how one could write a test for this that wouldn't be flaky...

travis79: The actual diversion from launch going into the queue is stopped before it processes the queue at all.  That way, all calls to launch _after_ `queueInitialTasks` is set to false would just be processed normally.  

So, by the time we start processing the queue, coroutines should be executing like normal run mode and it shouldn't be possible to add any more to the queue because the `queueInitialTasks` flag is no longer true.

mdboom: That makes sense.  Thanks for bearing with me.  Seems like a risky change wrt multithreading, but I don't have any specific thing to point at...

travis79: As for setting up a way to test this?  Let me know if you come up with something as I couldn't come up with a reliable way to test the edge case of adding while the queue is processing, aside from putting some very long running tasks in the queue (but again, no way to guarantee that the queue was still processing when adding new tasks)

travis79: I was also concerned about how this would affect the multi-threading balance we have going for us, but after testing this more and more, I think using the threadsafe collection and the way that it diverts the tasks to just running when it gets ready to process the queue makes me feel pretty confident about it.  The only thing I can see that might happen is that tasks that get added while it's processing the queue could get recorded before tasks that were added to the queue causing something to potentially get recorded out of chronological order.

mdboom: > The only thing I can see that might happen is that tasks that get added while it's processing the queue could get recorded before tasks that were added to the queue causing something to potentially get recorded out of chronological order.

Trying to think of where this might matter.  I'm optimistic it doesn't matter much...

- Timespans (if `stop` and `start`) were unordered -- however, `start` is actually synchronous, so I think that's a non-issue.
- String Lists -- that's just an argument for moving to String Sets which we're already planning
- Events shouldn't matter, since everything is timestamped, we can always sort by the timestamp if it matters for a particular analysis.  (I think the assumption already is that the *could* be out-of-order).
- Explicit calls to `sendPing` -- the ping may not include recent recordings that otherwise would be in the ping.  But I guess they'll show up in the next ping, so ultimately not an issue...?

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
```diff
@@ -52,11 +51,6 @@ class GleanCrashReporterService(
                 return@run
             }
 
-            if (!Glean.isInitialized()) {
```
Dexterp37: This change is out of scope for this PR. Let's take it out and have it as a follow-up PR.

components/service/glean/src/main/java/mozilla/components/service/glean/Dispatchers.kt
```diff
@@ -64,6 +85,61 @@ internal object Dispatchers {
...
+        private fun addTaskToQueue(block: suspend CoroutineScope.() -> Unit) {
+            if (testingMode) {
+                logger.info("Task queued for execution in test mode")
+                taskQueue.add {
+                    runBlocking {
+                        block()
+                    }
+                }
+            } else {
+                if (taskQueue.size >= MAX_QUEUE_SIZE) {
```
Dexterp37: I think this should be the first check of this function: the cap should be there regardless of the mode we're in. This would simplify analysis in case there's any bug around this.

```
if (taskQueue.size >= MAX_QUEUE_SIZE) {
  logger.error("Exceeded maximum queue size, discarding task")
  return
}

if (testingMode) {
    logger.info(...)
    taskQueue.add {
      block()
     }
} else {
    logger.info("Task queued for execution and delayed until flushed")
    taskQueue.add {
      coroutineScope.launch(block = block)
    }
}
```

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -129,11 +130,13 @@ open class GleanInternalAPI internal constructor () {
         // If Glean is being initialized from a test, then we want to skip running the
         // startupCheck() as it can cause some intermittent test failures due to multi-
         // threaded race conditions.
-        @Suppress("EXPERIMENTAL_API_USAGE")
```
Dexterp37: Let's leave this here and add another suppression: this would make it easier to remove/reconsider them if needed.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
Dexterp37: I see your concern, it's a good one. At the very least we should call this out with an extensive comment in `Dspatchers.flushQueuedInitialTasks`.

Synchronizing this would make us pay the cost of synchronization even *after* init. I'm afraid I don't have a good solution to offer for this off the top of my head. Can we also file a bug about this?

components/service/glean/src/test/java/mozilla/components/service/glean/DispatchersTest.kt
```diff
@@ -8,20 +8,22 @@ import kotlinx.coroutines.runBlocking
...
+@Suppress("EXPERIMENTAL_API_USAGE")
 class DispatchersTest {
 
     @Test
     fun `API scope runs off the main thread`() {
         val mainThread = Thread.currentThread()
         var threadCanary = false
-        @Suppress("EXPERIMENTAL_API_USAGE")
         Dispatchers.API.setTestingMode(false)
+        Dispatchers.API.queueInitialTasks = false
```
Dexterp37: Mh, should we have an internal test only method for setting the initial task flag, as we do for the testing mode? at the very least, we'd have a comparable API surface.

samples/crash/src/main/java/org/mozilla/samples/crash/CrashApplication.kt
```diff
@@ -46,6 +41,10 @@ class CrashApplication : Application() {
             nonFatalCrashIntent = createNonFatalPendingIntent(this),
             enabled = true
         ).install(this)
+
+        // Initialize Glean for recording by the GleanCrashReporterService.  
+        Glean.setUploadEnabled(true)
```
Dexterp37: Let's take this to a separate PR.

components/service/glean/src/main/java/mozilla/components/service/glean/Dispatchers.kt
```diff
@@ -11,31 +11,52 @@ import kotlinx.coroutines.ObsoleteCoroutinesApi
...
+        internal var queueInitialTasks = true
+
+        // Use a [ConcurrentLinkedQueue] to take advantage of it's thread safety and no locking
+        internal val taskQueue: ConcurrentLinkedQueue<() -> Unit> = ConcurrentLinkedQueue()
+
+        private val logger = Logger("glean/Dispatchers")
+
+        companion object {
+            // This value was chosen in order to allow several tasks to be queued for execution but
+            // still be conservative of memory.  This queue size is important for cases where
```
Dexterp37: nit: double whitespace between "...memory." and "This queue..."

components/service/glean/src/main/java/mozilla/components/service/glean/Dispatchers.kt
```diff
@@ -11,31 +11,52 @@ import kotlinx.coroutines.ObsoleteCoroutinesApi
...
+            // setUploadEnabled(false) is not called so that we don't continue to queue tasks and
+            // waste memory
```
Dexterp37: nit: missing trailing dot

components/service/glean/src/main/java/mozilla/components/service/glean/Dispatchers.kt
```diff
@@ -11,31 +11,52 @@ import kotlinx.coroutines.ObsoleteCoroutinesApi
...
+            const val MAX_QUEUE_SIZE = 100
+        }
+
         /**
-         * Launch a block of work asyncronously.
+         * Launch a block of work asynchronously.
+         *
+         * If [queueInitialTasks] is true, then the work will be queued until triggered.
```
Dexterp37: nit: "... queued and executed when [flushQueuedInitialTasks] is called."

components/service/glean/src/main/java/mozilla/components/service/glean/Dispatchers.kt
travis79: I'll go ahead and remove the extra space, but I will also leave you with this article ;)   

https://www.theatlantic.com/science/archive/2018/05/two-spaces-after-a-period/559304/

components/service/glean/src/main/java/mozilla/components/service/glean/Dispatchers.kt
```diff
@@ -64,6 +100,76 @@ internal object Dispatchers {
...
             testingMode = enabled
         }
+
+        /**
+         * Enable testing mode, which makes all of the Glean public API synchronous.
+         *
+         * @param enabled whether or not to enable the testing mode
+         */
+        @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+        fun enableQueueing(enabled: Boolean) {
```
Dexterp37: nit: let's be consistent with `setTestingMode` and call this `setTaskQueueing`

components/service/glean/src/main/java/mozilla/components/service/glean/Dispatchers.kt
```diff
@@ -64,6 +100,76 @@ internal object Dispatchers {
         fun setTestingMode(enabled: Boolean) {
             testingMode = enabled
         }
+
+        /**
+         * Enable testing mode, which makes all of the Glean public API synchronous.
+         *
+         * @param enabled whether or not to enable the testing mode
```
Dexterp37: Copy-paste comment out of sync with what the function does ;)

components/service/glean/src/test/java/mozilla/components/service/glean/GleanTest.kt
```diff
@@ -215,6 +222,81 @@ class GleanTest {
...
+        resetGlean(clearStores = false)
+
+        // Verify that the callback was executed by testing for the correct value
+        assertTrue("Value must exist", counterMetric.testHasValue())
+        assertEquals("Value must match", 3, counterMetric.testGetValue())
+    }
+
+    @Test
+    fun `queued tasks are executed in correct order`() {
+        val stringListMetric = StringListMetricType(
```
Dexterp37: Are you sure the test `Before` isn't interacting with this test?

components/service/glean/src/main/java/mozilla/components/service/glean/Dispatchers.kt
```diff
@@ -64,6 +100,76 @@ internal object Dispatchers {
...
+        /**
+         * Stops queueing tasks and processes any tasks in the queue. Since [queueInitialTasks] is
+         * set to false prior to processing the queue, newly launched tasks should be executed
+         * on the couroutine scope rather than added to the queue.
+         */
+        fun flushQueuedInitialTasks() {
+            queueSemaphore.acquire(2)
+            queueInitialTasks = false
+            taskQueue.forEach { task ->
+                runBlocking {
```
Dexterp37: What's this `runBlocking` doing here? I don't think it's needed nor desirable.

components/service/glean/src/main/java/mozilla/components/service/glean/Dispatchers.kt
```diff
@@ -64,6 +100,76 @@ internal object Dispatchers {
...
+            queueInitialTasks = enabled
+        }
+
+        /**
+         * Stops queueing tasks and processes any tasks in the queue. Since [queueInitialTasks] is
+         * set to false prior to processing the queue, newly launched tasks should be executed
+         * on the couroutine scope rather than added to the queue.
+         */
+        fun flushQueuedInitialTasks() {
+            queueSemaphore.acquire(2)
```
Dexterp37: What if we have 3-level nesting? We'd deadlock again. I think a semaphore might not be a good solution here. We should really go back to the drawing board and think about this thoroughly. Moreover, what's up with the solution of sending a "flipping task"?

travis79: Semaphore idea was just to illustrate the problem more than provide a solution.  I've backed out those changes per our discussion.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
travis79: I filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1568503) to follow up on this issue.

components/service/glean/src/main/java/mozilla/components/service/glean/Dispatchers.kt
```diff
@@ -64,6 +86,73 @@ internal object Dispatchers {
...
+        }
+
+        /**
+         * Stops queueing tasks and processes any tasks in the queue. Since [queueInitialTasks] is
+         * set to false prior to processing the queue, newly launched tasks should be executed
+         * on the couroutine scope rather than added to the queue.
+         */
+        internal fun flushQueuedInitialTasks() {
+            // Setting this to false first should cause any new tasks to just be executed (see
+            // launch() above) making it safer to process the queue.
```
Dexterp37: nit: please extend this comment to mention that flipping this could cause some tasks to be executed out of order (explain the reason) and also mention the bug that its looking to fix this.

PR #3829, opened 2019-07-19
Introduce an `HistogramBase` interface in the GleanSDK

Dexterp37:
The context for this is bug 1566368: this is needed in order for the glean parsers to generate an API to route batched streaming telemetry calls to the Glean SDK. It additionally:

* Changes the error reporting mechanism to allow for reporting more than 1 error of the same type.
* Adds support for accumulating multiple samples in `TimingDistributionStorageEngine`

### Pull Request checklist

### After merge

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/error/ErrorRecording.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/error/ErrorRecording.kt
@@ -41,12 +41,15 @@ object ErrorRecording {
      *     ping.  It does not need to include the metric name, as that is automatically
      *     prepended to the message.
      * @param logger The [Logger] instance to display the warning.
+     * @param numErrors The optional number of errors to report for this [ErrorType].
      */
+    @Suppress("LongMethod")
     internal fun recordError(
         metricData: CommonMetricData,
         errorType: ErrorType,
         message: String,
-        logger: Logger
+        logger: Logger,
+        numErrors: Int? = null
     ) {
         val errorName = GLEAN_ERROR_NAMES[errorType]!!
 
@@ -82,7 +85,7 @@ object ErrorRecording {
         // in the __other__ category.
         CountersStorageEngine.record(
             errorMetric,
-            amount = 1
+            amount = numErrors ?: 1
         )
     }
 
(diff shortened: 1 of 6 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/error/ErrorRecording.kt
```diff
@@ -41,12 +41,15 @@ object ErrorRecording {
      *     ping.  It does not need to include the metric name, as that is automatically
      *     prepended to the message.
      * @param logger The [Logger] instance to display the warning.
+     * @param numErrors The optional number of errors to report for this [ErrorType].
      */
+    @Suppress("LongMethod")
```
travis79: Wow, adding that little bit pushed it over the `LongMethod` limit?

Dexterp37: Yes :| That was a bit surprising to me too, but... meh :|

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
```diff
@@ -97,6 +98,54 @@ internal open class TimingDistributionsStorageEngineImplementation(
...
+                currentValue?.let {
+                    it.accumulate(sample)
+                    it
+                } ?: let {
+                    val newTD = TimingDistributionData(category = metricData.category, name = metricData.name,
+                        timeUnit = timeUnit)
+                    newTD.accumulate(sample)
+                    return@let newTD
+                }
+            }
```
mdboom: Maybe reduce the code duplication between this and `accumulate` by just calling this from accumulate with an array of length 1?  I'm concerned about these two code paths becoming out-of-sync in the future.

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
```diff
@@ -124,7 +173,8 @@ internal open class TimingDistributionsStorageEngineImplementation(
  * @param category of the metric
  * @param name of the metric
  * @param bucketCount total number of buckets
- * @param range an array always containing 2 elements: the minimum and maximum bucket values
+ * @param rangeMin the minimum value that can be represented
+ * @param rangeMin the maximum value that can be represented
```
mdboom:
```suggestion
 * @param rangeMax the maximum value that can be represented
```

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
Dexterp37: Good point. I changed this but I had to leave the negative value check in place. See the comment in the code.

Issue #3775, opened 2019-07-15
lib-push-amazon: Integrate & implement with concept-push and ADM SDK


psymoon:
- [x] Copy the amazon-device-messaging-1.0.1.jar file into `lib-push-amazon/libs` folder
- [x] Add an `AbstractAmazonPushService.kt` that implements `PushService`. This should be similar to [`AbstractFirebasePushService.kt`](https://github.com/mozilla-mobile/android-components/blob/master/components/lib/push-firebase/src/main/java/mozilla/components/lib/push/firebase/AbstractFirebasePushService.kt)
- [x] [Publish lib-push-adm component](https://github.com/mozilla-mobile/android-components/issues/3568)

Meta: https://github.com/mozilla-mobile/android-components/issues/1918

PR #3852, opened 2019-07-22
#3775: Integrate ADM push services

psymoon:
### Pull Request checklist

### After merge


components/lib/push-amazon/src/main/java/mozilla/components/lib/push/amazon/AbstractAmazonPushService.kt
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    override fun onRegistrationError(errorId: String?) {
+        errorId?.let { PushProcessor.requireInstance.onError(
+                PushError.MalformedMessage("registration failed: $errorId")
+            )
+        }
+    }
+
+    override fun onUnregistered(registrationId: String?) {
```
psymoon: @jonalmeida Do we have anything in `PushProcessor` or `AutopushFeature` to handle unregistration?

jonalmeida: From the API docs:
> Called on successful unregistration. This method may be called in response to your app calling startUnregister() or if ADM has unregistered the app for some reason (**typically because the device has lost its association with a user's Amazon account**).

This is interesting! While there isn't anything to handle the unregistration, we could surface the error with `PushProcessor.onError(PushError.Registration("account no longer associated with device"))`.

However, [there are situations](https://github.com/mozilla-mobile/android-components/issues/3859) where we need to force an unregistration, so we should make sure to only propagate the error in that case where we didn't explicitly call `startUnregister`.

components/lib/push-amazon/src/main/java/mozilla/components/lib/push/amazon/AbstractAmazonPushService.kt
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PROTECTED)
+    public override fun onRegistered(registrationId: String?) {
+        registrationId?.let { PushProcessor.requireInstance.onNewToken(it) }
+    }
+
+    override fun onRegistrationError(errorId: String?) {
+        errorId?.let { PushProcessor.requireInstance.onError(
+                PushError.MalformedMessage("registration failed: $errorId")
```
jonalmeida: Probably `PushError.Registration` instead of `MalformedMessage`?

components/lib/push-amazon/src/main/java/mozilla/components/lib/push/amazon/AbstractAmazonPushService.kt
```diff
@@ -0,0 +1,81 @@
...
+
+    override fun start(context: Context) {
+        adm = ADM(context)
+        if (adm.registrationId == null) {
+            adm.startRegister()
+        }
+    }
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PROTECTED)
+    public override fun onRegistered(registrationId: String?) {
```
jonalmeida: Is this a valid nullable string or is it the Java interoperability that causes it? It seems odd that we would get this callback with a null registration ID.

components/lib/push-amazon/src/main/java/mozilla/components/lib/push/amazon/AbstractAmazonPushService.kt
```diff
@@ -0,0 +1,81 @@
...
+                PushProcessor.requireInstance.onError(
+                    PushError.MalformedMessage("parsing encrypted message failed: $e")
+                )
+            }
+        }
+    }
+
+    override fun stop() {
+        if (adm.registrationId != null) {
+            adm.startUnregister()
```
jonalmeida: We should stop the service not unregister it. Can we tell the underlying service to stop running with `stopSelf` and/or any other things that need to be stopped?

components/lib/push-amazon/src/main/java/mozilla/components/lib/push/amazon/AbstractAmazonPushService.kt
```diff
@@ -0,0 +1,81 @@
...
+    override fun onRegistrationError(errorId: String?) {
+        errorId?.let { PushProcessor.requireInstance.onError(
+                PushError.MalformedMessage("registration failed: $errorId")
+            )
+        }
+    }
+
+    override fun onUnregistered(registrationId: String?) {
+        registrationId?.let {
+            PushProcessor.requireInstance.shutdown()
```
jonalmeida: This is a recursive call in the `AutopushFeature` since it calls `PushService#stop` which calls `adm.startUnregister` that invokes this callback.

components/lib/push-amazon/build.gradle
```diff
@@ -24,7 +24,11 @@ android {
 }
 
 dependencies {
+    compileOnly files('libs/amazon-device-messaging-1.0.1.jar')
```
jonalmeida: Since `compileOnly` is deprecated for `implementation` and `api`, do neither of those work?

.gitignore
```diff
@@ -62,5 +62,4 @@ jacoco.exec
 results/
 test_artifacts/
 google-cloud-sdk/
-*.jar
```
jonalmeida: @pocmo we talked about this briefly, but I think I misunderstood. Wouldn't we want to force add only the ADM jar and keep this gitignore filter here so we don't mistakenly commit other jars that shouldn't be here?

pocmo: Never force add anything. Fix the gitignore filter instead. This is way to broad. Whether we want to add a jar or not always depends on the case.
This was added for ignoring some UI test jars, and the fix in Fenix is here - let's do the same:
https://github.com/mozilla-mobile/fenix/pull/4088/commits/fe2d589171a67d44150c89a288f6331b166a95a8

components/lib/push-amazon/build.gradle
psymoon: `compileOnly` is to replace deprecated `provided`; different from runtime available `implementation` or `api`

psymoon: https://blog.gradle.org/introducing-compile-only-dependencies

components/lib/push-amazon/src/main/java/mozilla/components/lib/push/amazon/AbstractAmazonPushService.kt
psymoon: Ah, that's a good one. I will check on all the nullable string params

components/lib/push-amazon/src/main/java/mozilla/components/lib/push/amazon/AbstractAmazonPushService.kt
psymoon: Works for me

components/lib/push-amazon/build.gradle
jonalmeida: Ah my bad, I was still thinking about `compile`. :)

components/lib/push-amazon/src/main/java/mozilla/components/lib/push/amazon/AbstractAmazonPushService.kt
```diff
@@ -0,0 +1,74 @@
...
+
+    override fun onRegistrationError(errorId: String) {
+        PushError.Registration("registration failed: $errorId")
+    }
+
+    override fun onUnregistered(registrationId: String) {
+        PushError.Registration("account no longer associated with device")
+    }
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PROTECTED)
```
jonalmeida: Do we need these annotations for public overrides?

psymoon: It's protected fun by super, abstracts are intended to extended; only making it public for mocking test visibility. We could leave it public of course, of course, what do you think?

jonalmeida: Ah gotcha, no this is fine then. Thanks!

PR #3855, opened 2019-07-22
Remove service-experiments requirement that Glean be initialized

travis79:
Since Glean no longer requires it's consumers to wait for it to be initialized before recording metrics, service-experiments no longer needs to care about Glean's state.


travis79:
Waiting on #3821 before this can be merged

PR #3867, opened 2019-07-23
Extract the Glean 'metrics.yaml' file from the GV AAR in engine-gecko-nightly

Dexterp37:
This stack adds the following mechanics to support GeckoView metrics to the Glean SDK and engine-gecko-nightly:

- introduces an option to make `sdk_generator.gradle` look for `metrics.yaml` in AAR files (**note**: this doesn't restrict fetching the files to the geckoview AAR - could this be a problem?);
- ensures the `metrics.yaml` from AAR is available for generating the Kotlin APIs;
- changes the script so that the generation task runs even if no local file is available, as long as the AAR option is enabled;
- it integrates Glean to engine-gecko-nightly as a compileOnly dependency.

**How does this work?**

When the `allowMetricsFromAAR` option is enabled, the Glean SDK build script registers an artifact transformation class (i.e. `GleanMetricsYamlTransform`) that, as per the `registerTransform` block, is applied on an "exploded AAR" (unzipped). It looks for the 'metrics.yaml' in the root dir of the exploded AAR. That file is fed as input to the `generateMetricsSourceFor*` within the `doFirst` block of that task.

**How did I test this?**

- Look up the cached GeckoView AAR file available locally (e.g. `C:\Users\Alessio\.gradle\caches\modules-2\files-2.1\org.mozilla.geckoview\geckoview-nightly\70.0.20190722093634\b4070cbe52a9ae4b8f245dae0bf419564c3cc3d7\geckoview-nightly-70.0.20190722093634.aar` on my machine);
- add a sample `metrics.yaml` file to its root (for example, the one attached [here](https://github.com/mozilla-mobile/android-components/files/3423095/metrics-original.zip))
- Build engine-gecko-nightly
- A `GeckoviewGfx.kt` file should be available in `android-components\components\browser\engine-gecko-nightly\build\generated\source\glean\debug\kotlin`

### Pull Request checklist

### After merge


components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -126,6 +127,50 @@ clean {
...
+//
+// Note that this mechanism only applies to `module` dependencies (i.e., AAR files downloaded from
+// Maven) and not to `project` dependencies in the same root project or substituted as part of a
+// Gradle composite build.
+class GleanMetricsYamlTransform extends ArtifactTransform {
+    List<File> transform(File file) {
+        println("GleanMetricsYamlTransform was called")
+        def f = new File(file, "metrics.yaml")
+        if (f.exists()) {
+            return f
```
ncalexan: This should be a `List<File>`, I think.

components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -126,6 +127,50 @@ clean {
...
+        }
+        return null
```
ncalexan: And this an empty list, I expect.

components/browser/engine-gecko-nightly/build.gradle
```diff
@@ -26,6 +30,11 @@ android {
     }
 }
 
+// Set configuration for the Glean parses to extract metrics.yaml
```
mdboom:
```suggestion
// Set configuration for the Glean parser to extract metrics.yaml
```

components/browser/engine-gecko-nightly/build.gradle
```diff
@@ -46,11 +55,19 @@ dependencies {
     testImplementation project(':support-test')
     testImplementation project(':tooling-fetch-tests')
 
+
```
mdboom: Remove extra blank line?

components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -126,6 +127,50 @@ clean {
...
+        def f = new File(file, "metrics.yaml")
+        if (f.exists()) {
+            println("Glean SDK - extracting ${f.path} from an AAR archive")
+            return [f]
+        }
+        return []
+    }
+}
+
+if (project.ext.has("allowMetricsFromAAR")) {
```
mdboom: Maybe instead of `allow`, say `use`?  Since if this is true, it will pull the `metrics.yaml` from the AAR and *not* use a local one.  "allow" implies (at least to me) that it might use both.

components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -161,7 +218,11 @@ ext.gleanGenerateMetricsAPI = {
         args "$sourceOutputDir"
         args "-s"
         args "namespace=$fullNamespace"
-        args "$projectDir/metrics.yaml"
+        // Only hardcode the path to the 'metrics.yaml' file if we're not trying to fetch
+        // it from the AAR files.
+        if (!project.ext.has("allowMetricsFromAAR")) {
+            args "$projectDir/metrics.yaml"
+        }
```
mdboom: Is this strictly necessary?  I think `glean_parser` silently ignores missing files.  Just in the interest of simplifying things.

Dexterp37: Good point. I was torn on this, but we might actually want this behaviour for the GeckoView environment.

components/service/glean/scripts/sdk_generator.gradle
Dexterp37: Would this still be the case if I leave the local 'metrics.yaml' around as per the next comment?

mdboom: Good point.  Maybe `include`?  But don't feel as strongly -- leaving it at `allow` would be ok, too.

Dexterp37: Ok, I left it as `allow` for now. I've also updated the commit message to explain what's going on! Let me know if it's still unclear!

components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -126,6 +127,50 @@ clean {
...
+// https://android.googlesource.com/platform/tools/base/+/studio-master-dev/build-system/gradle-core/src/main/java/com/android/build/gradle/internal/VariantManager.java#592.
+//
+// Note that this mechanism only applies to `module` dependencies (i.e., AAR files downloaded from
+// Maven) and not to `project` dependencies in the same root project or substituted as part of a
+// Gradle composite build.
+class GleanMetricsYamlTransform extends ArtifactTransform {
+    List<File> transform(File file) {
+        def f = new File(file, "metrics.yaml")
+        if (f.exists()) {
+            println("Glean SDK - extracting ${f.path} from an AAR archive")
```
ncalexan: You'll get nicer output if you get a `logger` from a Gradle `Project` instance -- `logger.lifecycle` is often what you want for "always on" output, or `logger.info` or `logger.debug`.  There are examples of this in patches I posted, if you care to do it.

components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -161,6 +218,8 @@ ext.gleanGenerateMetricsAPI = {
         args "$sourceOutputDir"
         args "-s"
         args "namespace=$fullNamespace"
+        // We intentionally don't let out the local metrics.yaml, even if another one will
+        // be extracted from the AAR file. The parser will just ignore this one, if not present.
         args "$projectDir/metrics.yaml"
```
ncalexan: Include these two as `inputs file(...)`, so that Gradle dependencies are correct.

components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -142,6 +187,18 @@ ext.gleanGenerateMetricsAPI = {
...
+        if (project.ext.has("allowMetricsFromAAR")) {
+            // This is sufficiently lazy to be valid at configuration time.  See the model at
+            // https://github.com/google/protobuf-gradle-plugin/blob/6d99a421c8d15710045e4e8d31a3af6cb0cc3b70/src/main/groovy/com/google/protobuf/gradle/ProtobufPlugin.groovy#L270-L277
+            inputs.files variant.compileConfiguration.incoming.artifactView {
+                attributes {
+                    it.attribute(ArtifactAttributes.ARTIFACT_FORMAT, 'glean-metrics-yaml')
+                }
+            }.files
+
+            outputs.dir sourceOutputDir
```
ncalexan: This should be outside the `if`, for Gradle dependencies to work.

components/browser/engine-gecko-nightly/build.gradle
```diff
@@ -46,11 +55,18 @@ dependencies {
     testImplementation project(':support-test')
     testImplementation project(':tooling-fetch-tests')
 
+    // We only compile against Glean. It's up to the app to add those dependencies
```
ncalexan: These are required at compile time 'cuz the generated code uses the Glean SDK, right?

I don't feel strongly about this but consider making this a regular `implementation` dependency; folks who don't actually consume the generated Glean definitions should get that stuff stripped by r8/PG.

components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -142,6 +187,18 @@ ext.gleanGenerateMetricsAPI = {
     def generateKotlinAPI = task("generateMetricsSourceFor${variant.name.capitalize()}", type: Exec) {
         description = "Generate the Kotlin code for the Metrics API"
 
+        if (project.ext.has("allowMetricsFromAAR")) {
+            // This is sufficiently lazy to be valid at configuration time.  See the model at
+            // https://github.com/google/protobuf-gradle-plugin/blob/6d99a421c8d15710045e4e8d31a3af6cb0cc3b70/src/main/groovy/com/google/protobuf/gradle/ProtobufPlugin.groovy#L270-L277
+            inputs.files variant.compileConfiguration.incoming.artifactView {
```
ncalexan: Just FYI, if you did want to restrict to only `geckoview-*` dependencies, I think you'd probably do it here, filtering by dependency name or group or something.

components/service/glean/scripts/sdk_generator.gradle
Dexterp37: The problem with it is that the files *can* be there, but it's not mandatory. If I add that directive the build will fail if the files are not there. We can have the following situations:

- no `metrics.yaml` at all (that's ok, Glean will only report the baseline ping with its core metrics);
- a local `metrics.yaml` and no `pings.yaml` (Glean will report the new metrics through its [built-in pings](https://mozilla.github.io/glean/book/user/pings/index.html));
- a local `metrics.yaml` and a local `pings.yaml` (Glean will report the new metrics as defined in the pings file);
- local files + AAR  `metrics.yaml` file (Glean will process all the 'metrics.yaml' and do whatever is needed)

components/browser/engine-gecko-nightly/build.gradle
Dexterp37: This was a requirement by the a-c team (cc @pocmo) to not have an hard dependency on Glean at runtime, unless the product consuming 'engine-gecko' is also using Glean (which is not always the case). We're using this pattern in `lib-crash` as well. We might consider revisiting this in the future :)

components/service/glean/scripts/sdk_generator.gradle
Dexterp37: Thanks for mentioning that. I thought about doing this, but I really don't want to hardcode too many GeckoView specific things in the Glean SDK. If we stumble upon the situation of multiple AARs exporting metrics.yaml files, this won't break compilation, but will generate code that will end up being stripped away by r8/PG. Worst case, we will end up introducing another configuration option to restrict this. I'd leave this as it is for now, for the current use-case.

components/service/glean/scripts/sdk_generator.gradle
Dexterp37: Oh, thanks, good catch.

components/service/glean/scripts/sdk_generator.gradle
ncalexan: > The problem with it is that the files _can_ be there, but it's not mandatory. If I add that directive the build will fail if the files are not there.

Huh, I thought missing inputs were accepted by Gradle, but apparently not: see https://devsoap.com/optional-input-files-for-gradle/.

Alternatively, if you put these files in `glean/`, then you could add the directory as input instead, which does the right thing generally...

But this can be for future expansion.

Dexterp37: > 
> 
> > The problem with it is that the files _can_ be there, but it's not mandatory. If I add that directive the build will fail if the files are not there.
> 
> Huh, I thought missing inputs were accepted by Gradle, but apparently not: see https://devsoap.com/optional-input-files-for-gradle/.

I was quite surprised about this as well :(

> Alternatively, if you put these files in `glean/`, then you could add the directory as input instead, which does the right thing generally...

Yes, I'm afraid that doing this change right now would require changing all the consumers of Glean :) I'd open that can of worms some other day ;)

PR #3882, opened 2019-07-24
Initial draft SyncEngineStore

Amejia481:
@grigoryk  I would like to know what do you think about  using a `SyncEngineStore` for keeping the
state of which engines the users want to be able to sync from. In the future, I will also like to have a component that will  persist/ restore the state for the `SyncEngineStore` (ideally using shared pref).

```kotlin
    val engineStore = SyncEngineStore()

    // Activating Bookmarks
    engineStore.dispatch(EnableSyncEngineAction(SyncAbleEngine.Bookmarks))

    // Activating History
    engineStore.dispatch(EnableSyncEngineAction(SyncAbleEngine.History))

    // Disabling History
    engineStore.dispatch(DisableSyncEngineAction(SyncAbleEngine.Bookmarks))

    // Disabling all engines
    engineStore.dispatch(DisableAllSyncEnginesAction)

    // Apps and other components can react to changes on the enabledEngines
    engineStore.observe(view){ newState ->
        newState.enabledEngines
    }
```

### DIFF
--- a/components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/engine/SyncEngineState.kt
+++ b/components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/engine/SyncEngineState.kt
@@ -1,0 +1,11 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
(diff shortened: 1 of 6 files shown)


Amejia481:
Thanks for your interest, I was taking with @grigoryk on Friday about this new approach. We decided for the sake of simplicity, for the moment we are going to continue using the actual infrastructure that the sync components offer.

Amejia481:
Thanks for the references I will take a look sounds like an exciting topic to read about :)

Issue #3757, opened 2019-07-12
Implement a standard repository configuration across Android projects


mcomella:
Gradle introduced their dependency -> repository matching feature that can improve the security of our apps (see quoted text below) and performance of Gradle configuration. There are problems with expecting each team to implement & maintain this individually:
- They may implement it incorrectly
- They need to keep it up-to-date if the requirements change
- It takes time for each team to implement it

Ideally, we could share a solution across all of our projects so we only have to keep it up-to-date in one spot: we could leverage [our gradle plugin](https://github.com/mozilla-mobile/android-automation-tools/tree/master/gradle-plugin) to do so.

Copied from https://groups.google.com/a/mozilla.com/d/msg/mobile-all/CJinm9Tg4E0/veP0RiWvBwAJ:

> You can improve the security of fetching your dependencies by using a feature Gradle released to [match dependencies to specific repositories](https://docs.gradle.org/current/userguide/declaring_repositories.html#sec::matching_repositories_to_dependencies). In practice, the minimum security implementation looks something like:
```
maven {
    url "https://maven.mozilla.org/maven2"
}
google()
jcenter {
    content {
        excludeGroupByRegex "mozilla.components...
        excludeGroupByRegix "com.google...
    }
}
```
> which will prevent us from accidentally downloading malicious dependencies from jcenter that are, e.g.:
> - newer versions of our dependencies
> - commonly typo'd versions of our dependencies
> 
> Here's [a PR](https://github.com/mozilla-mobile/firefox-tv/pull/2374) for a complete implementation on FFTV that we've been using for over a month: note that it also uses this feature to decrease Gradle configuration time. I think most consumer apps should be able to copy it verbatim.
> 
> N.B.: this API is considering Incubating in Gradle and is subject to breaking changes.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-3107)

csadilek:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1794688

Change performed by the [Move to Bugzilla add-on](https://addons.mozilla.org/en-US/firefox/addon/move-to-bugzilla/).

PR #3884, opened 2019-07-24
Issue #3757: Use more strict repository configuration.

pocmo:
@mcomella Can you glance over this too - since you did something similar for `firefox-tv`?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,19 +24,63 @@ plugins {
 
 allprojects {
     repositories {
-        google()
-        jcenter()
-
         maven {
             url "https://maven.mozilla.org/maven2"
+            content {
+                // We want to fetch all our known Mozilla groups from our own repository.
+                includeGroup GroupMatching.GROUP_MOZILLA_APPSERVICES
+                includeGroup GroupMatching.GROUP_MOZILLA_COMPONENTS
+                includeGroup GroupMatching.GROUP_MOZILLA_GECKOVIEW
+                excludeGroup GroupMatching.GROUP_MOZILLA_SERVOVIEW
+            }
         }
 
-        ////////////////////////////////////////////////////////////////////////////////////////////
-        // Servo
-        ////////////////////////////////////////////////////////////////////////////////////////////
+        google() {
+            content {
+                // We want to fetch all AndroidX, Google and Firebase groups from Google's repository.
+                includeGroupByRegex GroupMatching.REGEX_ANDROIDX
+                includeGroupByRegex GroupMatching.REGEX_ANDROID
+                includeGroupByRegex GroupMatching.REGEX_GOOGLE_ANDROID
+                includeGroup GroupMatching.GROUP_GOOGLE_FIREBASE
+
+                // We never want to fetch Mozilla groups from Google.
+                excludeGroupByRegex GroupMatching.REGEX_MOZILLA
+                excludeGroup GroupMatching.GROUP_MOZILLA_SERVOVIEW
+            }
+        }
+
+        mavenCentral() {
+            // We list Maven Central before jcenter since we prefer pulling artifacts from there if available. Maven
+            // Central has a more strict vetting process when adding new artifacts.
+
+            content {
+                // We never want to fetch Mozilla, AndroidX, Google or Firebase groups from Maven Central.
+                excludeGroupByRegex GroupMatching.REGEX_MOZILLA
+                excludeGroupByRegex GroupMatching.REGEX_ANDROIDX
+                excludeGroupByRegex GroupMatching.REGEX_ANDROID
+                includeGroupByRegex GroupMatching.REGEX_GOOGLE_ANDROID
+                excludeGroup GroupMatching.GROUP_GOOGLE_FIREBASE
+                excludeGroup GroupMatching.GROUP_MOZILLA_SERVOVIEW
+            }
+        }
+
+        jcenter() {
+            content {
+                // We never want to fetch Mozilla, AndroidX, Google or Firebase groups from jcenter.
+                excludeGroupByRegex GroupMatching.REGEX_MOZILLA
+                excludeGroupByRegex GroupMatching.REGEX_ANDROIDX
+                excludeGroupByRegex GroupMatching.REGEX_ANDROID
+                excludeGroupByRegex GroupMatching.REGEX_GOOGLE_ANDROID
+                excludeGroup GroupMatching.GROUP_GOOGLE_FIREBASE
+                excludeGroup GroupMatching.GROUP_MOZILLA_SERVOVIEW
+            }
+        }
 
         maven {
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -24,19 +24,63 @@ plugins {
 
 allprojects {
     repositories {
-        google()
-        jcenter()
-
         maven {
             url "https://maven.mozilla.org/maven2"
+            content {
+                // We want to fetch all our known Mozilla groups from here.
```
mcomella: nit: I find these comments redundant to the code.

build.gradle
```diff
@@ -24,19 +24,63 @@ plugins {
...
+                // We never want to fetch Mozilla, AndroidX, Google or Firebase groups from Maven Central.
+                excludeGroupByRegex GroupMatching.REGEX_MOZILA
+                excludeGroupByRegex GroupMatching.REGEX_ANDROIDX
+                excludeGroupByRegex GroupMatching.REGEX_ANDROID
+                excludeGroup GroupMatching.GROUP_GOOGLE_FIREBASE
+
+                // We are not excluding REGEX_GOOGLE_ANDROID here since there's one group that is not getting published
+                // on maven.google.com so far (but on Maven Central, also jcenter but we prefer Maven Central):
+                // com.google.android.apps.common.testing.accessibility.framework
+                // See: https://github.com/google/Accessibility-Test-Framework-for-Android/issues/14
```
mcomella: This exposes you to vulnerabilities on com.google.android dependencies. I'd prefer if you explicitly excluded this dependency from the matching like we do on FFTV: https://searchfox.org/mozilla-mobile/source/firefox-tv/buildSrc/src/main/java/org/mozilla/gradle/Dependencies.kt#28 Though I understand the trade-off of complexity.

build.gradle
```diff
@@ -24,19 +24,63 @@ plugins {
...
+                // We never want to fetch Mozilla, AndroidX, Google or Firebase groups from jcenter.
+                excludeGroupByRegex GroupMatching.REGEX_MOZILA
+                excludeGroupByRegex GroupMatching.REGEX_ANDROIDX
+                excludeGroupByRegex GroupMatching.REGEX_ANDROID
+                excludeGroupByRegex GroupMatching.REGEX_GOOGLE_ANDROID
+                excludeGroup GroupMatching.GROUP_GOOGLE_FIREBASE
+            }
+        }
 
         maven {
```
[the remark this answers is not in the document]
pocmo: I think the original reason why we moved this to the bottom was that we didn't want to hit `download.servo.org` early and for a lot of dependency requests. With the exclusion rules - does the order for *this* repository matter?

pocmo:
Pushed an updated version.

build.gradle
mcomella: Yes, I think they should be escaped and yes, I think it should be framework. Nice catch.

build.gradle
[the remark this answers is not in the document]
mcomella: With the current code, I don't think the order matters: we will only hit `download.servo.org` for the appropriate requests and those dependencies cannot be downloaded from elsewhere. However, moving it upward is defensive coding: if someone later changes the repositories above later such that SERVOVIEW is not excluded, we could be vulnerable to using the wrong dependency.

Note that if you move it up, you should add the exclusions too, though it's not strictly necessary (since `includeGroup...` means it'll only look at those specific dependencies).

Then again, this is getting sufficiently complicated that it'll be hard to change correctly later: perhaps just using `includeGroup`s without `excludeGroup`s but with an explicit comment like, "If this changes, be very careful because you could add a vulnerability!" or something is better than being explicit with our includeGroups and excludeGroups. @pocmo What do you think?

build.gradle
```diff
@@ -24,19 +24,63 @@ plugins {
...
+        mavenCentral() {
+            // We list Maven Central before jcenter since we prefer pulling artifacts from there if available. Maven
+            // Central has a more strict vetting process when adding new artifacts.
+
+            content {
+                // We never want to fetch Mozilla, AndroidX, Google or Firebase groups from Maven Central.
+                excludeGroupByRegex GroupMatching.REGEX_MOZILLA
+                excludeGroupByRegex GroupMatching.REGEX_ANDROIDX
+                excludeGroupByRegex GroupMatching.REGEX_ANDROID
+                includeGroupByRegex GroupMatching.REGEX_GOOGLE_ANDROID
```
mcomella: nit: you should explain why this is included. Note that the comment above is outdated too.

build.gradle
```diff
@@ -24,19 +24,63 @@ plugins {
...
+                excludeGroup GroupMatching.GROUP_GOOGLE_FIREBASE
+                excludeGroup GroupMatching.GROUP_MOZILLA_SERVOVIEW
+            }
+        }
+
+        jcenter() {
```
mcomella: Do any artifacts actually get pulled from jcenter? Perhaps we can remove this.

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -132,3 +132,22 @@ object Dependencies {
...
+ * Used for matching dependencies to specific repositories.
+ */
+object GroupMatching {
+    const val REGEX_ANDROIDX = "androidx\\..*"
+    const val REGEX_MOZILLA = "org\\.mozilla\\..*"
+    const val REGEX_ANDROID = "com\\.android\\..*"
+
+    // Excluding the accessibility framework since it is not getting published to maven.google.com
+    // See https://github.com/google/Accessibility-Test-Framework-for-Android/issues/14
+    const val REGEX_GOOGLE_ANDROID = "com\\.google\\.android\\.(?!apps\\.common\\.testing\\.accessibility\\.framework).*"
```
mcomella: The trailing . should be escaped. :)

build.gradle
```diff
@@ -24,19 +24,63 @@ plugins {
 
 allprojects {
     repositories {
```
mcomella: The following files also have `repositories` declarations: in theory, we should cover them too:
- build.gradle (this file, the `buildscript` block)
- settings.gradle
- buildSrc/build.gradle.kts

build.gradle
mcomella: fwiw, we could probably share these declarations with our android gradle plugin...

mcomella: I'm not sure it's worth the time though.

buildSrc/src/main/java/Dependencies.kt
mcomella: Nevermind, no it shouldn't. Regex x_x

build.gradle
dector: Seems like it's true. Can't find concrete descriptions of changes for Bintray verification since [the last incident](https://status.bintray.com/incidents/w4dfr0rpznkt).

PR #3901, opened 2019-07-25
Closes #3741: Introduce engine startup method

### LINKED ISSUE #3741: Introduce engine startup / warmup method
We need a method for applications to invoke in onCreate so that we can make sure

- the Gecko runtime/engine is intialized on the main thread
- have a hook to run init/warmup logic

See: https://github.com/mozilla-mobile/android-components/blob/master/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/AbstractCustomTabsService.kt#L34

csadilek:
@pocmo putting this up as a draft so we can start pairing/discussing...

There's no code we have to run right now, but we want the lazy engine init blocks to execute in `Application.onCreate` so we can guarantee we're on the main thread. Alternative naming suggestions welcome :) e.g `ensureStarted` or `ensureInitialized`?

### DIFF
--- a/components/concept/engine/src/main/java/mozilla/components/concept/engine/Engine.kt
+++ b/components/concept/engine/src/main/java/mozilla/components/concept/engine/Engine.kt
@@ -6,6 +6,7 @@ package mozilla.components.concept.engine
 
 import android.content.Context
 import android.util.AttributeSet
+import androidx.annotation.MainThread
 import mozilla.components.concept.engine.utils.EngineVersion
 import mozilla.components.concept.engine.webextension.WebExtension
 import org.json.JSONObject
@@ -50,6 +51,16 @@ interface Engine {
 
         override fun hashCode() = types
     }
+
+    /**
+     * Makes sure all required engine initialization logic is executed. The
+     * details are specific to individual implementations, but the following must be true:
+     *
+     * - The engine must be operational after this method was called successfully
+     * - Calling this method on an engine that is already initialized has no effect
+     */
+    @MainThread
+    fun warmUp() = Unit
 
     /**
      * Creates a new view for rendering web content.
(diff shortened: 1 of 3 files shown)


components/concept/engine/src/main/java/mozilla/components/concept/engine/Engine.kt
```diff
@@ -51,6 +51,15 @@ interface Engine {
...
     }
 
+    /**
+     * Starts up this engine and makes sure all required initialization logic is executed. The
+     * details are specific to individual implementations, but the following must be true:
+     *
+     * - The engine must be operational after this method was called successfully
+     * - Calling this method on an engine that is already started has no effect
+     */
+    fun start() = Unit
```
pocmo: nit: Following `AbstractCustomTabsService` I think I'd just call it `warmup()`.

components/concept/engine/src/main/java/mozilla/components/concept/engine/Engine.kt
```diff
@@ -51,6 +51,15 @@ interface Engine {
...
+
```
pocmo: I wonder if we should override this in`GeckoEngine` implementations and do a thread/process check? 🤔 - But maybe that's more something for the creation of the object.

components/concept/engine/src/main/java/mozilla/components/concept/engine/Engine.kt
csadilek: OK, changed!

components/concept/engine/src/main/java/mozilla/components/concept/engine/Engine.kt
csadilek: Yeah, I agree it would be have to when we create the instance, but not sure. `Gecko` will already crash if we're on the wrong thread. Wdyt?

csadilek:
OK, ready for review / landing when we're happy.

components/concept/engine/src/main/java/mozilla/components/concept/engine/Engine.kt
pocmo: Oh yeah, right. It only didn't for the process, I think. But that may be too late anyways here.

PR #3904, opened 2019-07-25
Closes #3903 - Add feature-notifications component

### LINKED ISSUE #3903: Add feature-notifications component
This will be the home for the Web Notifications API implementation.

NotWoods:
A new empty component for the Web Notifications API.


.buildconfig.yml
```diff
@@ -60,6 +60,10 @@ projects:
     path: components/feature/media
     description: 'Feature component for website media related features.'
     publish: true
+  feature-notifications:
```
pocmo: naming nit: let's maybe be more explicit and pick `feature-web-notifications` or `feature-webnotifications` since just "notifications" can mean a lot and could be confusing since there are other components showing notifications.

NotWoods: Updated!

Issue #2295, opened 2019-03-07
Implement GeckoWebNotificationDelegate


pocmo:
Waiting for GeckoView API.

cpeterso:
GV bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1533057

cpeterso:
The Web Push API landed in GV 70 Nightly on 2019-08-21:

https://bugzilla.mozilla.org/show_bug.cgi?id=1533057#c11

jonalmeida:
Looking at [the API docs](https://mozilla.github.io/geckoview/javadoc/mozilla-central/org/mozilla/geckoview/WebNotificationDelegate.html), this is an engine delegate as opposed to an engine session one which we usually have.

I imagine we can follow a similar pattern and have engine observers that are connected to a notification manager that handles showing the android OS notification/dismissal.

For `WebNotificationDelegate#onCloseNotification`, aside from notifying observers, I'm not sure what we should do with it or apps could do with it. 🤔

pocmo:
> For WebNotificationDelegate#onCloseNotification, aside from notifying observers, I'm not sure what we should do with it or apps could do with it. 🤔

Remove the notification if it is still showing? 🤔

jonalmeida:
> Remove the notification if it is still showing? 🤔

I believe this maps to the Web API [`onClose`](https://developer.mozilla.org/en-US/docs/Web/API/Notification/onclose) which I imagine would be invoked when we call [`close`](https://developer.mozilla.org/en-US/docs/Web/API/Notification/close) (GV is [`dismiss`](https://mozilla.github.io/geckoview/javadoc/mozilla-central/org/mozilla/geckoview/WebNotification.html#dismiss--)) on it.

If that's true, then this would be for other observers to know a user interacted and dismissed the notification. 

It is still true, that we may also want to also ensure that the no notification is showing if we have to programatically close all notifications for currently unknown reasons.

**TODO:** observe all interactions possible with WebNotifications in Desktop and Chrome Android to compare.

pocmo:
I assume its triggered by this:
https://developer.mozilla.org/en-US/docs/Web/API/Notification/close

And we need to know that we need to cancel the notification if JS decides to close it?

PR #3949, opened 2019-07-29
Issue #2295 - Convert WebNotification to Notification

NotWoods:
Adds a new data class for [web notifications](https://developer.mozilla.org/en-US/docs/Web/API/Notification), and a method to convert it to an Android notification (not to actually send the notification, yet).

### Pull Request checklist

### After merge

### DIFF
--- a/components/concept/engine/src/main/java/mozilla/components/concept/engine/webnotifications/WebNotification.kt
+++ b/components/concept/engine/src/main/java/mozilla/components/concept/engine/webnotifications/WebNotification.kt
@@ -22,7 +22,6 @@ package mozilla.components.concept.engine.webnotifications
  * notification was clicked.
  * @property onClose Callback called when the notification is dismissed.
  */
-@Suppress("Unused")
 data class WebNotification(
     val origin: String,
     val title: String? = null,
--- a/components/feature/webnotifications/build.gradle
+++ b/components/feature/webnotifications/build.gradle
@@ -22,14 +22,20 @@ android {
     }
 }
 
-
 dependencies {
+    implementation project(':browser-icons')
+    implementation project(':concept-engine')
     implementation project(':support-ktx')
+
+    implementation Dependencies.androidx_core_ktx
+    implementation Dependencies.kotlin_stdlib
+    implementation Dependencies.kotlin_coroutines
 
     testImplementation project(':support-test')
 
     testImplementation Dependencies.androidx_test_core
     testImplementation Dependencies.androidx_test_junit
+    testImplementation Dependencies.testing_coroutines
     testImplementation Dependencies.testing_robolectric
     testImplementation Dependencies.testing_mockito
 }
--- a/components/feature/webnotifications/src/main/java/mozilla/components/feature/notifications/Placeholder.kt
+++ b/components/feature/webnotifications/src/main/java/mozilla/components/feature/notifications/Placeholder.kt
@@ -1,7 +1,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
-package mozilla.components.feature.webnotifications
-
-class Placeholder
--- a/components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/NativeNotificationBridge.kt
+++ b/components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/NativeNotificationBridge.kt
@@ -1,0 +1,78 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.feature.webnotifications
+
+import android.app.Notification
+import android.content.Context
+import android.graphics.Bitmap
+import android.net.Uri
+import android.os.Build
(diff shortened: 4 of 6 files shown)


components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotification.kt
```diff
@@ -0,0 +1,60 @@
...
+ * @property timestamp Time when the notification was created.
+ * @property renotify Preference flag that indicates the end user should be alerted after the show
+ * steps have run with a new notification that has the same tag as an existing notification.
+ * @property silent Preference flag that indicates no sounds or vibrations should be made.
+ * @property actions Action buttons to display below the notification.
+ * @property onClick Callback called with the selected action, or null if the main body of the
+ * notification was clicked.
+ * @property onClose Callback called when the notification is dismissed.
+ */
+data class WebNotification(
```
pocmo: That probably will need to live in `concept-engine`? 🤔 

I wonder if it is a bit early to write this code here. The GV API doesn't even exist yet. Looking at the WIP patch for GV it doesn't have half of the fields here.

NotWoods: The intention was to support all the fields in Android Components and let GeckoView incrementally build out support for everything here. I can wait for this to land until the GV side is ready, but I've been working with Alvina from the GV team on this patch.

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/SystemNotificationBridge.kt
```diff
@@ -0,0 +1,78 @@
...
+        notification: WebNotification,
+        context: Context,
+        channelId: String
+    ): Notification {
+        val builder = if (SDK_INT >= Build.VERSION_CODES.O) {
+            Notification.Builder(context, channelId)
+        } else {
+            @Suppress("Deprecation")
+            Notification.Builder(context)
+        }
```
pocmo: nit: `NotificationCompat.Builder()` can deal with that for you.

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/SystemNotificationBridge.kt
```diff
@@ -0,0 +1,78 @@
...
+            builder.setVibrate(vibrate)
+
+            timestamp?.let {
+                builder.setShowWhen(true).setWhen(it)
+            }
+
+            if (silent) {
+                @Suppress("Deprecation")
+                builder.setDefaults(0)
+            }
```
pocmo: I think properties like silent and vibrate will be ignored if a notification channel is used. Probably nothing we can do here though. Still makes sense to set the behavior for older versions.

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/SystemNotificationBridge.kt
```diff
@@ -0,0 +1,78 @@
...
+import android.net.Uri
+import android.os.Build
+import android.os.Build.VERSION.SDK_INT
+import mozilla.components.browser.icons.BrowserIcons
+import mozilla.components.browser.icons.Icon.Source
+import mozilla.components.browser.icons.IconRequest
+import mozilla.components.browser.icons.IconRequest.Size
+import mozilla.components.concept.engine.webnotifications.WebNotification
+
+internal class SystemNotificationBridge(
```
jonalmeida: Maybe call this `NativeNotificationBridge` since we use `System` to denote the system engine? We have to also add a default implementation of the abstraction for the system engine.

components/concept/engine/src/main/java/mozilla/components/concept/engine/webnotifications/WebNotification.kt
```diff
@@ -0,0 +1,39 @@
...
+ */
+data class WebNotification(
+    val origin: String,
+    val title: String? = null,
+    val body: String? = null,
+    val tag: String? = null,
+    val iconUrl: Uri? = null,
+    val vibrate: LongArray = longArrayOf(),
+    val timestamp: Long? = null,
+    val requireInteraction: Boolean = false,
```
jonalmeida: Did we also want to add `textDirection` and `lang`? I'm not sure how valuable they would be to us though since that would probably be dictated by the system language setting instead.

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/SystemNotificationBridge.kt
NotWoods: Unfortunately notification compat doesn't have silent/vibrate :(

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/SystemNotificationBridge.kt
NotWoods: Yeah, we can at least support older devices here.

components/concept/engine/src/main/java/mozilla/components/concept/engine/webnotifications/WebNotification.kt
NotWoods: I chose to omit those fields since we can't use them.

pocmo:
> Land it (after fixing CI)!

In that case avoid using `bors r+` to not slow down the queue with a build that is doomed to fail. :)

NotWoods:
bors r=jonalmeida

PR #3953, opened 2019-07-29
Closes #1503: Add support for window requests to engine-gecko.

### LINKED ISSUE #1503: Add support for window requests to engine-gecko
We now have the API and the implementation for engine-system (added by https://github.com/mozilla-mobile/android-components/pull/1495).

For GeckoView, we're waiting for https://bugzilla.mozilla.org/show_bug.cgi?id=1510314. Once that's resolved, we can send `onOpenWindowRequest` / `onCloseWindowRequest` notifications from engine-gecko.

We also need to make a decision which setting to use enable/disable popup blocking (e.g. `javascriptCanOpenWindowsAutomatically`).

csadilek:
This brings in window request support for `engine-gecko-nightly`. We can now open new tabs for `target="_blank"` links and `window.open` calls.  🎉 

The trick here was to have an API that works with both `engine-system` and `engine-gecko`, but luckily I only needed a small API change and neither R-B nor Fenix use the feature yet.

@pocmo there's one interesting problem here that we could solve with lib-state. I had to use `MainScope.launch()` when notifying observers of the window request. This is because (for engine-gecko) processing the window request will itself have a side-effect and change the observers list. The side-effect is that a new session is selected causing an observer (the crash observer) to get unregistered (in `GeckoEngineView.render`). This affects the observers we're currently iterating over causing a `ConcurrentModificationException`. By "launching", we can complete the current iteration over existing observers and avoid the problem.

### DIFF
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -8,10 +8,13 @@ import android.annotation.SuppressLint
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
+import kotlinx.coroutines.MainScope
+import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
 import mozilla.components.browser.engine.gecko.media.GeckoMediaDelegate
 import mozilla.components.browser.engine.gecko.permission.GeckoPermissionRequest
 import mozilla.components.browser.engine.gecko.prompt.GeckoPromptDelegate
+import mozilla.components.browser.engine.gecko.window.GeckoWindowRequest
 import mozilla.components.browser.errorpages.ErrorType
 import mozilla.components.concept.engine.EngineSession
 import mozilla.components.concept.engine.EngineSessionState
@@ -58,7 +61,8 @@ class GeckoEngineSession(
             .build()
         GeckoSession(settings)
     },
-    private val context: CoroutineContext = Dispatchers.IO
+    private val context: CoroutineContext = Dispatchers.IO,
+    openGeckoSession: Boolean = true
 ) : CoroutineScope, EngineSession() {
 
     internal lateinit var geckoSession: GeckoSession
@@ -90,7 +94,7 @@ class GeckoEngineSession(
         get() = context + job
 
     init {
-        createGeckoSession()
+        createGeckoSession(shouldOpen = openGeckoSession)
     }
 
     /**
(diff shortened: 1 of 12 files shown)


components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -307,13 +311,6 @@ class GeckoEngineSession(
             session: GeckoSession,
             request: NavigationDelegate.LoadRequest
         ): GeckoResult<AllowOrDeny> {
-            // TODO use onNewSession and create window request:
-            // https://github.com/mozilla-mobile/android-components/issues/1503
-            if (request.target == GeckoSession.NavigationDelegate.TARGET_WINDOW_NEW) {
-                geckoSession.loadUri(request.uri)
-                return GeckoResult.fromValue(AllowOrDeny.DENY)
-            }
```
pocmo: I wonder if we should return ALLOW here instead of removing the block. We didn't execute the following block for `TARGET_WINDOW_NEW` before and I wonder if it makes sense to do it now - especially since the following code (like interceptors) do not know about the target? 🤔

components/browser/engine-gecko-nightly/src/test/java/mozilla/components/browser/engine/gecko/window/GeckoWindowRequestTest.kt
```diff
@@ -0,0 +1,19 @@
+package mozilla.components.browser.engine.gecko.window
```
pocmo: nit: license header

docs/changelog.md
```diff
@@ -69,6 +70,13 @@ permalink: /changelog/
 * **support-ktx**
   * Added `Collection.crossProduct` to retrieve the cartesian product of two `Collections`.
 
+* **feature-session**
+  * ⚠️ **This is a breaking change**:
+  * The `WindowFeature` no longer needs and engine can now be created using just: 
+  ```kotlin
+     val windowFeature = WindowFeature(components.sessionManager)
+  ```
```
pocmo: Needs to be moved to 7.0.0 now :)

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
csadilek: Yeah, good call. Updated and re-tested.

components/browser/engine-gecko-nightly/src/test/java/mozilla/components/browser/engine/gecko/window/GeckoWindowRequestTest.kt
csadilek: Fixed.

docs/changelog.md
csadilek: ok, moved :).

Issue #3952, opened 2019-07-29
TWA: Verify Digital Asset Links


NotWoods:
Used to verify that a Trusted Web Activity can hide the toolbar. Can also be used for the `validateRelationship` in `CustomTabsService`.

- https://developers.google.com/digital-asset-links/
- https://developer.android.com/training/app-links/verify-site-associations

NotWoods:
Blocked by #3930

We need to store the package name of the application creating the custom tab/trusted web activity in order to lookup the signature later.

PR #3954, opened 2019-07-30
Issue #3952 - Add Digital Asset Links verification

NotWoods:
Largely based on Google's asset links code. This will be used for Trusted Web Activities to determine if the toolbar should be visible.

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/verify/DigitalAssetLinksHandler.kt
+++ b/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/verify/DigitalAssetLinksHandler.kt
@@ -1,0 +1,91 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.feature.customtabs.verify
+
+import android.webkit.URLUtil
+import androidx.annotation.VisibleForTesting
+import androidx.core.net.toUri
+import mozilla.components.concept.fetch.Client
+import mozilla.components.concept.fetch.Request
+import mozilla.components.concept.fetch.Response
+import org.json.JSONException
+import org.json.JSONObject
+import java.io.IOException
+import java.util.concurrent.TimeUnit
+
+/**
+ * Digital Asset Links APIs allows any caller to check pre declared
+ * relationships between two assets which can be either web domains
+ * or native applications. This requests checks for a specific
+ * relationship declared by a web site with an Android application
+ */
+internal class DigitalAssetLinksHandler(
+    private val httpClient: Client
+) {
+
+    fun checkDigitalAssetLinkRelationship(
+        webDomain: String,
+        packageName: String,
+        fingerprint: String?,
+        relationship: String
+    ): Boolean {
+        val requestUrl = getUrlForCheckingRelationship(webDomain, packageName, fingerprint, relationship)
+        val response = fetch(requestUrl) ?: return false
+
+        val responseBody = response.use { response.body.string() }
+        return try {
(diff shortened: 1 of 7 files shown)


components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/verify/DigitalAssetLinksHandler.kt
```diff
@@ -0,0 +1,74 @@
...
+    fun checkDigitalAssetLinkRelationship(
+        webDomain: String,
+        packageName: String,
+        fingerprint: String?,
+        relationship: String
+    ): Boolean {
+        val requestUrl = getUrlForCheckingRelationship(webDomain, packageName, fingerprint, relationship)
+        if (!URLUtil.isValidUrl(requestUrl)) return false
+
+        val response = httpClient.fetch(Request(requestUrl))
```
pocmo: Should we define some timeouts here so that this doesn't potentially read/connect forever?

pocmo: We need to catch the IOException here! This is an easy mistake that I made too since Kotlin does not enforce catching exceptions. I almost wonder whether we should change the return type to something like [Try](https://arrow-kt.io/docs/arrow/core/try/) to enforce thinking about this case.

pocmo: I recommend using `httpClient.fetch(Request(requestUrl)).use { ... }` here to always make sure you are closing the response/body.

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/verify/DigitalAssetLinksHandler.kt
```diff
@@ -0,0 +1,74 @@
...
+        if (response.status != HTTP_OK) return false
```
pocmo: Do you explicitly only care about 200? Otherwise you can also use `Response.isSuccess`.

NotWoods: We only explicitly care about 200 here.

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/verify/DigitalAssetLinksHandler.kt
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+
+    private fun fetch(requestUrl: String): Response? {
+        if (!URLUtil.isValidUrl(requestUrl)) return null
+        return try {
+            val response = httpClient.fetch(
+                Request(
+                    requestUrl,
+                    connectTimeout = TIMEOUT
```
pocmo: No `readTimeout`?

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/verify/OriginVerifier.kt
```diff
@@ -0,0 +1,162 @@
...
+
+        val input = ByteArrayInputStream(signature.toByteArray())
+        return try {
+            val certificate = CertificateFactory.getInstance("X509").generateCertificate(input) as X509Certificate
+            byteArrayToHexString(MessageDigest.getInstance("SHA256").digest(certificate.encoded))
+        } catch (e: CertificateEncodingException) {
+            // Certificate type X509 encoding failed
+            null
+        } catch (e: CertificateException) {
+            // This shouldn't happen.
```
pocmo: If you are really sure that should not happen then I suggest throwing `AssertionError("Should not happen", e)` to verify your assumptions.

PR #3969, opened 2019-07-31
 Closes #3964: Add onAuthenticatedLogin to AccountObserver

### LINKED ISSUE #3734: Add empty default implementation for AccountObserver
This would not require account observers to implement the entire interface if they don't want to.

For example, in push, we care only about `onAuthenticated` and `onLoggedOut`, but we have to implement the others with no-op.

The change should be simple enough as adding `= Unit` for all interface methods.

### LINKED ISSUE #3964: Expose when an account is newly signed in vs re-authenticated to observers
For `AccountManager` observers, we see that a cold boot of an app fires the observer's `onAuthenticated`. It would be useful to have know the difference between a newly signed in account for symmetry with `AccountObserver.onLoggedOut`.

@grigoryk mentioned this state already exists within the AccountManager, so we would need to expose this to the observers.

jonalmeida:
WIP to see if I'm on the right track.

Also fixes #3734: Add default impl for AccountObserver
### Pull Request checklist

### After merge


jonalmeida:
> I think a simpler approach would be to add a flag - or an emum? - to `onAuthenticated`.
> Something like this, but probably with a different flag name: `onAuthenticated(account: OAuthAccount, newAccount: Boolean)`
> 
> This way, consumers of this observer that don't care about this distinction can simply avoid the extra parameter. Otherwise, they'd need to have more boilerplate - an extra method to override, which they don't care about.

Good point! I was wondering mostly if the change that needs to take place is in the right state. If so, I'll update the tests as well.

jonalmeida:
> Otherwise, they'd need to have more boilerplate - an extra method to override, which they don't care about.

Forgot to mention that we've added default implementations on the AccountObserver as well so that you don't have to force all observers to implement a new method.

jonalmeida:
> You are touching code files here - [ci skip] is dangerous here :P

Whoops, this was when it was a draft, thanks for catching!

components/concept/sync/src/main/java/mozilla/components/concept/sync/OAuthAccount.kt
```diff
@@ -56,24 +56,24 @@ interface AccountObserver {
...
      */
-    fun onLoggedOut()
+    fun onLoggedOut() = Unit
 
     /**
      * Account was successfully authenticated.
      * @param account An authenticated instance of a [OAuthAccount].
      */
-    fun onAuthenticated(account: OAuthAccount)
+    fun onAuthenticated(account: OAuthAccount, newAccount: Boolean = false) = Unit
```
grigoryk: This is a breaking change for consumers that implement AccountObserver, right? Let's document it.

grigoryk: Also: missing `@param newAccount` docs.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
```diff
@@ -561,7 +561,7 @@ open class FxaAccountManager(
 
                         maybeUpdateSyncAuthInfoCache()
 
-                        notifyObservers { onAuthenticated(account) }
+                        notifyObservers { onAuthenticated(account, true) }
```
grigoryk: Also in `SignedInShareableAccount` we get a new account.

components/concept/sync/src/main/java/mozilla/components/concept/sync/OAuthAccount.kt
grigoryk: Also: I'd prefer you omit a default `false` value, so that we're forced to change every invocation of `onAuthenticated`. This way it's super obvious we're not missing any calls, and at every call site you immediately see what kind of notification we'll fire. Default params are a footgun.

jonalmeida: With a default param, this shoudn't be a breaking change, but I forgot to add it to the changelog!

jonalmeida: Ignore, I saw your comment below.

PR #3993, opened 2019-08-01
1570647 - Do not send 'metrics' ping if the app was not used

travis79:
This attempts to address the issue around sending of metrics pings when the app hasn't been used for that day.

### Pull Request checklist

### After merge

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
@@ -8,6 +8,10 @@ import android.content.Context
 import android.content.SharedPreferences
 import androidx.annotation.VisibleForTesting
 import android.text.format.DateUtils
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.LifecycleObserver
+import androidx.lifecycle.OnLifecycleEvent
+import androidx.lifecycle.ProcessLifecycleOwner
 import androidx.work.ExistingWorkPolicy
 import androidx.work.OneTimeWorkRequestBuilder
 import androidx.work.Worker
@@ -31,8 +35,12 @@ import java.util.concurrent.TimeUnit as AndroidTimeUnit
  * - ping is overdue (due time already passed) for the current calendar day;
  * - ping is soon to be sent in the current calendar day;
  * - ping was already sent, and must be scheduled for the next calendar day.
+ *
+ * The scheduler also makes use of the [LifecycleObserver] in order to correctly schedule
+ * the [MetricsPingWorker]
  */
-internal class MetricsPingScheduler(val applicationContext: Context) {
+@Suppress("TooManyFunctions")
+internal class MetricsPingScheduler(val applicationContext: Context) : LifecycleObserver {
     private val logger = Logger("glean/MetricsPingScheduler")
     internal val sharedPreferences: SharedPreferences by lazy {
         applicationContext.getSharedPreferences(this.javaClass.canonicalName, Context.MODE_PRIVATE)
@@ -41,6 +49,12 @@ internal class MetricsPingScheduler(val applicationContext: Context) {
     companion object {
         const val LAST_METRICS_PING_SENT_DATETIME = "last_metrics_ping_iso_datetime"
         const val DUE_HOUR_OF_THE_DAY = 4
+        @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+        internal var isInForeground = false
+    }
+
+    init {
+        ProcessLifecycleOwner.get().lifecycle.addObserver(this)
     }
 
     /**
@@ -157,7 +171,7 @@ internal class MetricsPingScheduler(val applicationContext: Context) {
      * Performs startup checks to decide when to schedule the next metrics ping
      * collection.
      */
-    fun startupCheck() {
+    fun schedule() {
         val now = getCalendarInstance()
         val lastSentDate = getLastCollectedDate()
 
@@ -218,8 +232,14 @@ internal class MetricsPingScheduler(val applicationContext: Context) {
         // Update the collection date: we don't really care if we have data or not, let's
         // always update the sent date.
         updateSentDate(getISOTimeString(now, truncateTo = TimeUnit.Day))
-        // Reschedule the collection.
-        schedulePingCollection(now, sendTheNextCalendarDay = true)
+
+        // Reschedule the collection if we are in the foreground so that any metrics collected after
+        // this are sent in the next window.  If we are in the background, then we may stay there
+        // until the app is killed so we shouldn't reschedule unless the app is foregrounded again
+        // (see GleanLifecycleObserver).
+        if (isInForeground) {
+            schedulePingCollection(now, sendTheNextCalendarDay = true)
+        }
     }
 
     /**
@@ -262,6 +282,32 @@ internal class MetricsPingScheduler(val applicationContext: Context) {
      */
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
     internal fun getCalendarInstance(): Calendar = Calendar.getInstance()
+
+    /**
+     * Update flag to show we are no longer in the foreground.
+     */
+    @OnLifecycleEvent(Lifecycle.Event.ON_STOP)
+    fun onEnterBackground() {
+        isInForeground = false
+    }
+
+    /**
+     * Update the flag to indicate we are moving to the foreground, and if Glean is initialized we
+     * will check to see if the metrics ping needs scheduled for collection.
+     */
+    @OnLifecycleEvent(Lifecycle.Event.ON_START)
+    fun onEnterForeground() {
+        isInForeground = true
+
+        // We check for the metrics ping schedule here because the app could have been in the
+        // background and resumed in which case Glean would already be initialized but we still need
+        // to perform the check to determine whether or not to collect and schedule the metrics ping.
+        // If this is the first ON_START event since the app was launched, Glean wouldn't be
+        // initialized yet.
+        if (Glean.isInitialized()) {
+            schedule()
+        }
+    }
 }
 
 /**
--- a/components/service/glean/src/test/java/mozilla/components/service/glean/scheduler/MetricsPingSchedulerTest.kt
+++ b/components/service/glean/src/test/java/mozilla/components/service/glean/scheduler/MetricsPingSchedulerTest.kt
@@ -4,21 +4,20 @@
 
 package mozilla.components.service.glean.scheduler
 
-import android.content.Context
 import android.os.SystemClock
 import androidx.test.core.app.ApplicationProvider
 import androidx.work.testing.WorkManagerTestInitHelper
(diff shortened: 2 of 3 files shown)


travis79:
This is silly but the test I added (`startupCheck happens when returning from background when Glean is already initialized`) runs perfect in debug but fails in release, so any ideas on why that might be would be greatly appreciated.

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/GleanLifecycleObserver.kt
```diff
@@ -40,5 +40,14 @@ internal class GleanLifecycleObserver : LifecycleObserver {
...
         // time the app goes to background.
         GleanBaseline.duration.start()
+
+        // We check for the metrics ping schedule here because the app could have been in the
+        // background and resumed in which case Glean would already be initialized but we still need
+        // to perform the check to determine whether or not to collect and schedule the metrics ping.
+        // If this is the first ON_START event since the app was launched, Glean wouldn't be
+        // initialized yet.
+        if (Glean.isInitialized()) {
+            Glean.metricsPingScheduler.startupCheck()
```
mdboom: The risk of this change might be that when it enters foreground multiple times during the same 24 hour period, the ping is scheduled for sending twice (at the same 4AM instant the next morning).  I think the fact that we use `WorkManager.getInstance().enqueueUniqueWork()` makes that impossible, but I wonder if there's a way to write a test for that.

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
```diff
@@ -218,8 +219,26 @@ internal class MetricsPingScheduler(val applicationContext: Context) {
...
+    /**
+     * Helper function to determine whether the app is running in the background or not.
+     *
+     * @return [Boolean] true if running in background, or false if in foreground.
+     */
+    private fun isApplicationInForeground(): Boolean {
+        val procInfo = ActivityManager.RunningAppProcessInfo()
+        ActivityManager.getMyMemoryState(procInfo)
+
+        return procInfo.importance == ActivityManager.RunningAppProcessInfo.IMPORTANCE_FOREGROUND
```
mdboom: This is an **amazing** API from Android.

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/GleanLifecycleObserver.kt
travis79: I could also check to see if a `Worker` is already enqueued, which I didn't really think to do either.  I agree we should be okay since we do use the `enqueueUniqueWork` but a test and/or a check wouldn't hurt.

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
travis79: 😄 	I hope you are being sarcastic because just looking at the docs for this made me feel wrong, from weird `out` parameter calling structure to the methods and riddled with deprecated `Int` constants representing the importances (all with arbitrary mysterious values). 

The positive side is that it seems to work quite well.

travis79: It definitely smells of very early single digit Android API

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/GleanLifecycleObserver.kt
Dexterp37: nit: I think we need to rename `startupCheck` since this is no longer something that's only called at startup. What about just `schedule` or `check`?

Dexterp37: One test could be to use the workmanager testing api to enqueue two unique work units and check if both are there. But that would only test the WorkManager API, and I bet that such a test is already there, for them

components/service/glean/src/test/java/mozilla/components/service/glean/scheduler/MetricsPingSchedulerTest.kt
```diff
@@ -440,6 +444,34 @@ class MetricsPingSchedulerTest {
...
     }
 
+    @Test
+    fun `startupCheck happens when returning from background when Glean is already initialized`() {
+        // For faking calling the lifecycle observer.
+        val lifecycleRegistry = LifecycleRegistry(mock(LifecycleOwner::class.java))
+        val gleanLifecycleObserver = GleanLifecycleObserver()
+        lifecycleRegistry.addObserver(gleanLifecycleObserver)
+
+        // Set the last sent date to yesterday.
```
Dexterp37: nit: is this the wrong comment?

components/service/glean/src/test/java/mozilla/components/service/glean/scheduler/MetricsPingSchedulerTest.kt
```diff
@@ -440,6 +444,34 @@ class MetricsPingSchedulerTest {
...
+
+        // Simulate going to background.
+        lifecycleRegistry.handleLifecycleEvent(Lifecycle.Event.ON_STOP)
+
+        // Simulate returning to the foreground.
+        lifecycleRegistry.handleLifecycleEvent(Lifecycle.Event.ON_START)
+
+        // Verify that we're performing the startup check. Using 'timeout' to allow time for
+        // the lifecycle event to process.
+        verify(mpsSpy, timeout(1000).times(1)).startupCheck()
```
Dexterp37: How was this timeout picked? Let's use an higher one, maybe 5 s is enough to account for slow CI, we had bugs before about that and had [to bump to 5s](https://github.com/mozilla-mobile/android-components/blob/dcf534f57024dfae161e4d47541e590db23a8c7b/components/service/glean/src/test/java/mozilla/components/service/glean/TestUtil.kt#L185)

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
Dexterp37: Mh, looks like this usage of their API has been [actively discouraged by Google devs themselves](https://groups.google.com/forum/#!msg/android-developers/zH-2bovZSLg/L2YM8Z1N-HwJ). Looks like the only/proper solution to this is to check for `ON_START`.

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/GleanLifecycleObserver.kt
travis79: Sounds good to me!

components/service/glean/src/test/java/mozilla/components/service/glean/scheduler/MetricsPingSchedulerTest.kt
travis79: Yes it is...  pesky things, I thought I put out a trap for them. ;)

components/service/glean/src/test/java/mozilla/components/service/glean/scheduler/MetricsPingSchedulerTest.kt
travis79: It could actually be much less than this. I think 100ms worked pretty consistently so I figured 10x that for CI, but this is a cool function and bails out early if it can verify before the timeout (at least according to the docs) so I'll go ahead and up this as you request.

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
pocmo: Also: The importance of the process can change, right? At process start it may be something lower but later become FOREGROUND? Also how does that work with other, specifically higher, levels, e.g. IMPORTANCE_FOREGROUND_SERVICE - the app can have foreground UI and a foreground service in the same process (e.g. that happens when media is playing).

travis79: In light of your concerns, I've decided to switch to using a flag that is set in the `GleanLifecycleObserver` when it handles the lifecycle events.

components/service/glean/src/test/java/mozilla/components/service/glean/scheduler/MetricsPingSchedulerTest.kt
```diff
@@ -54,18 +58,13 @@ class MetricsPingSchedulerTest {
...
 
-    @Before
-    fun setup() {
-        WorkManagerTestInitHelper.initializeTestWorkManager(
-            ApplicationProvider.getApplicationContext())
-
-        Glean.enableTestingMode()
-    }
+    @get:Rule
+    val gleanRule = GleanTestRule(ApplicationProvider.getApplicationContext())
```
Dexterp37: I think this might end up causing a few more intermittents: we init the scheduler in this rule as well. This is the reason why this test wasn't using the `GleanTestRule` in the first place. I'd recommend adding a test-only `testReset` method to `MetricsPingScheduler` that clears whatever SharedPrefs it uses and then call it at the beginning of the relevant tests.

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
```diff
@@ -224,23 +223,11 @@ internal class MetricsPingScheduler(val applicationContext: Context) {
         // this are sent in the next window.  If we are in the background, then we may stay there
         // until the app is killed so we shouldn't reschedule unless the app is foregrounded again
         // (see GleanLifecycleObserver).
-        if (isApplicationInForeground()) {
+        if (GleanLifecycleObserver.isInForeground) {
```
Dexterp37: Instead of making `MetricsPingScheduler` tightly coupled to `GleanLifecycleObserver`, what about it provides a `setForegroundStatus(isForeground: boolean)` and we call it from `GleanLifecycleObserver`? Another option would be to have another observer embedded in the scheduler, to make it completely stand-alone. But I have no strong opinion on this last bit.

Moreover, by providing the method, you might be able to write tests more easily.

components/service/glean/src/test/java/mozilla/components/service/glean/scheduler/MetricsPingSchedulerTest.kt
travis79: Just one and it's not intermittent, it's always on.  I'm refactoring this now to create a method specific reset as you suggest.

travis79:
> This looks good to me. When porting this to glean-core, we should document the new specifications (we don't send if we were not in foreground in the measurement window).
> 
> Let's get a review from Sebastian and Mike as well. Please also take this for a manual test drive.

Sounds good to me, working on the manual test drive today and looks like you already flagged @mdboom and @pocmo

PR #4002, opened 2019-08-02
Test the 'baseline' ping in the glean-sample app

Dexterp37:
This adds an instrumented test that sends the app to background, triggers a ping and then validates
the received data.

**What does this do?**

- Tweak the `GleanTestRule` to allow for passing a `Configuration`;
- Introduces a custom test runner that spins up an HTTP server;
- Runs the sample-app and moves it to background to trigger pings;

**What doesn't it do?**

- It doesn't perform schema validation. It really should, but I would be copying over most of Glean's `testUtils.kt`. Maybe as a follow-up PR to find a better way to share the code?
- It doesn't work without mocking/force-triggering the WorkManager. This is weird, but I found no way other than this. It even requires a sleep. Eugh.

### Pull Request checklist

### After merge


components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
```diff
@@ -46,13 +46,14 @@ data class Configuration internal constructor(
     // constructor and only initialized with a proper default when calling the primary
     // constructor from the secondary, public one, below.
     constructor(
+        serverEndpoint: String = DEFAULT_TELEMETRY_ENDPOINT,
```
travis79: I knew that this would eventually happen, but it's still a little scary to open up the server endpoint to the public constructor.

Dexterp37: Yup, but think of this: the only way to change this would be to recompile the app. At that point, if recompilation is required, then anything can happen.

As long as this is not a runtime setting, we should be fine.

components/service/glean/src/main/java/mozilla/components/service/glean/testing/GleanTestRule.kt
```diff
@@ -24,18 +24,27 @@ import org.junit.runner.Description
...
-    val context: Context
+    val context: Context,
+    val retainServerEndpoint: Boolean = false
 ) : TestWatcher() {
     override fun starting(description: Description?) {
         // We're using the WorkManager in a bunch of places, and Glean will crash
         // in tests without this line. Let's simply put it here.
         WorkManagerTestInitHelper.initializeTestWorkManager(context)
 
+        val config: Configuration = if (retainServerEndpoint && Glean.isInitialized()) {
```
Dexterp37: @mdboom @travis79 this looks like an hack to me, but I really couldn't find a better way around this. `Glean` is a singleton, so it gets reset by the test rule. However, the `GleanTestRunner` only initializes once. So we need to find a way for the test rule to not overwrite the url, if requested.

travis79: Is there a way that you could just pass the `Configuration` object into the test rule? Maybe add a function that you could call?  I admit I don't know exactly what all is possible here so I need to research `Rules` a little more.

Dexterp37: That's a great idea! I did not think of that, it actually simplified things quite a lot. This now looks less like an hack. Thanks @travis79 !

components/service/glean/src/main/java/mozilla/components/service/glean/testing/GleanTestRule.kt
```diff
@@ -24,9 +24,13 @@ import org.junit.runner.Description
...
  * val gleanRule = GleanTestRule(ApplicationProvider.getApplicationContext())
  * ```
+ *
+ * @param context the application context
+ * @param configToUse an optional [Configuration] to initialize the Glean SDK with
  */
 class GleanTestRule(
-    val context: Context
+    val context: Context,
+    val configToUse: Configuration = Configuration()
```
badboy: I guess this doesn't need public documentation [in the book](https://mozilla.github.io/glean/book/user/testing-metrics.html), as it shouldn't concern the majority of users, right?

samples/glean/src/androidTest/java/org/mozilla/samples/glean/GleanTestRunner.kt
```diff
@@ -0,0 +1,89 @@
...
+import okhttp3.mockwebserver.Dispatcher
+import okhttp3.mockwebserver.MockResponse
+import okhttp3.mockwebserver.MockWebServer
+import okhttp3.mockwebserver.RecordedRequest
+
+/**
+ * Create a mock webserver that accepts all requests.
+ * @return a [MockWebServer] instance
+ */
+private fun getMockWebServer(): MockWebServer {
```
badboy: I was confused by this name, as the docs say it creates one and the next method has the same signature as well. Care to rename this into `createMockWebServer` or `instantiateMockWebServer` or something more descriptive?

samples/glean/src/androidTest/java/org/mozilla/samples/glean/GleanTestRunner.kt
```diff
@@ -0,0 +1,89 @@
...
+import androidx.annotation.VisibleForTesting
+import androidx.test.platform.app.InstrumentationRegistry
+import androidx.test.runner.AndroidJUnitRunner
+import okhttp3.mockwebserver.Dispatcher
+import okhttp3.mockwebserver.MockResponse
+import okhttp3.mockwebserver.MockWebServer
+import okhttp3.mockwebserver.RecordedRequest
+
+/**
+ * Create a mock webserver that accepts all requests.
```
badboy:
```suggestion
 * Create a mock webserver that accepts all requests and replies with "OK".
```

samples/glean/src/androidTest/java/org/mozilla/samples/glean/GleanTestRunner.kt
```diff
@@ -0,0 +1,89 @@
...
+@VisibleForTesting(otherwise = VisibleForTesting.NONE)
+class GleanTestRunner : AndroidJUnitRunner() {
+    // Add a lazy ping server to the app runner. This is only initialized once
+    // since the `Application` object is re-used.
+    internal val pingServer: MockWebServer by lazy { getMockWebServer() }
+    internal var pingServerAddress: String? = null
+
+    init {
+        // We need to start the server off the main process, otherwise
+        // Android will throw an exception. Spawning a thread and joining
```
badboy: What exception is thrown?

samples/glean/src/androidTest/java/org/mozilla/samples/glean/GleanTestRunner.kt
```diff
@@ -0,0 +1,89 @@
...
+        // seems fine.
```
badboy: this is fine

(Sorry, I had to)

samples/glean/src/androidTest/java/org/mozilla/samples/glean/pings/BaselinePingTest.kt
```diff
@@ -0,0 +1,134 @@
...
+        pingName: String,
+        maxAttempts: Int = 3
+    ): JSONObject?
+    {
+        val server = getPingServer()
+
+        var attempts = 0
+        do {
+            attempts += 1
+            val request = server.takeRequest(20L, TimeUnit.SECONDS)
```
badboy: Should we really wait that long? In tests we expect far lower times, don't we?

samples/glean/src/androidTest/java/org/mozilla/samples/glean/pings/BaselinePingTest.kt
```diff
@@ -0,0 +1,134 @@
...
+            }
+        } while (attempts < maxAttempts)
+
+        return null
+    }
+
+    /**
+     * Sadly, the WorkManager still requires us to manually trigger the upload job.
+     * This function goes through all the active jobs by Glean (there should only be
+     * one!) and triggers them.
```
badboy: Would there be any benefit in asserting that it's just one?

samples/glean/src/androidTest/java/org/mozilla/samples/glean/pings/BaselinePingTest.kt
```diff
@@ -0,0 +1,134 @@
...
+    @Test
+    fun validateBaselinePing() {
+        // Wait for the app to be idle/ready.
+        InstrumentationRegistry.getInstrumentation().waitForIdleSync()
+        val device = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
+        device.waitForIdle()
+
+        // Wait for 1 second: this should guarantee we have some valid duration in the
+        // ping.
+        Thread.sleep(1000)
```
badboy: :/ (but I get it)

components/service/glean/src/main/java/mozilla/components/service/glean/testing/GleanTestRule.kt
Dexterp37: Correct!

samples/glean/src/androidTest/java/org/mozilla/samples/glean/GleanTestRunner.kt
Dexterp37: `NetworkOnMainThreadException`, updated the comment.

samples/glean/src/androidTest/java/org/mozilla/samples/glean/pings/BaselinePingTest.kt
Dexterp37: We shouldn't, we really expect way lower times. However, we have seen intermittents before on a-c with higher times, so we settled on this number (that we're using elsewhere too).

samples/glean/src/androidTest/java/org/mozilla/samples/glean/pings/BaselinePingTest.kt
Dexterp37: I don't think so, since we're explicitly using a `WorkManager` API that should take care of that (`enqueueUniqueWork`). We'd be testing that API :)

samples/glean/src/androidTest/java/org/mozilla/samples/glean/pings/BaselinePingTest.kt
Dexterp37: Yeah :( The other approach would have been to use internal functions, which I've tried hard to avoid here.

Dexterp37:
> 
> 
> I share your concerns about the sleeps, but maybe when working at this really high integration level, there's not much better we could do.

Yes, I agree with you. The only sleep that really concerns me is the one from the `WorkManager`, but maybe that will go away if we bump the version.

> I don't think it's super critical to check the schema here, since our other unit tests do that, but it would be nice to maybe file a bug (if there isn't one already) to come back around to that.

I agree. I filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1572649) for tracking this.

Issue #3173, opened 2019-05-29
Support invalidating Push subscriptions on an interval


jonalmeida:
We need to periodically check that the push subscriptions need to be invalidated with `RustPushConnection#verifyConnection`. This then needs to update the subscribers when a subscription has be re-validated.

PR #4010, opened 2019-08-02
 Closes #3769: Add SendTabFeature with push support 

### LINKED ISSUE #3769: Add SendTabFeature to feature-sendtab
A lot of the code [in this PR][0] can be converted to a SendTabFeature in our new feature-sendtab component.

[0]: https://github.com/mozilla-mobile/fenix/pull/4041

jonalmeida:
Send Tab Feature, so beautiful, so many tests! 😭 Merge immediately.

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/sendtab/src/main/java/mozilla/components/feature/sendtab/SendTabFeature.kt
+++ b/components/feature/sendtab/src/main/java/mozilla/components/feature/sendtab/SendTabFeature.kt
@@ -1,0 +1,140 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ *  License, v. 2.0. If a copy of the MPL was not distributed with this
+ *  file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package mozilla.components.feature.sendtab
+
+import androidx.lifecycle.LifecycleOwner
+import androidx.lifecycle.ProcessLifecycleOwner
+import mozilla.components.concept.push.Bus
+import mozilla.components.concept.push.PushService
+import mozilla.components.concept.sync.AccountObserver as SyncAccountObserver
+import mozilla.components.concept.sync.Device
+import mozilla.components.concept.sync.DeviceConstellation
+import mozilla.components.concept.sync.DeviceEvent
+import mozilla.components.concept.sync.DeviceEventsObserver
+import mozilla.components.concept.sync.DevicePushSubscription
+import mozilla.components.concept.sync.OAuthAccount
+import mozilla.components.concept.sync.TabData
+import mozilla.components.feature.push.AutoPushFeature
+import mozilla.components.feature.push.AutoPushSubscription
+import mozilla.components.feature.push.PushSubscriptionObserver
+import mozilla.components.feature.push.PushType
+import mozilla.components.service.fxa.manager.FxaAccountManager
+import mozilla.components.support.base.log.logger.Logger
+
+/**
+ * A feature that uses the [FxaAccountManager] to send and receive tabs with optional push support
+ * for receiving tabs from the [AutoPushFeature] and a [PushService].
+ *
+ * If the push components are not used, the feature can still function while tabs would only be
+ * received when refreshing the device state.
+ *
+ * @param accountManager Firefox account manager.
+ * @param pushFeature The [AutoPushFeature] if that is setup for observing push events.
+ * @param owner Android lifecycle owner for the observers. Defaults to the [ProcessLifecycleOwner]
+ * so that we can always observe events throughout the application lifecycle.
+ * @param autoPause whether or not the observer should automatically be
+ * paused/resumed with the bound lifecycle.
+ * @param onTabsReceived the callback invoked with new tab(s) are received.
+ */
+class SendTabFeature(
+    accountManager: FxaAccountManager,
+    pushFeature: AutoPushFeature? = null,
+    owner: LifecycleOwner = ProcessLifecycleOwner.get(),
+    autoPause: Boolean = false,
+    onTabsReceived: (Device?, List<TabData>) -> Unit
+) {
+    init {
+        val accountObserver = AccountObserver(pushFeature)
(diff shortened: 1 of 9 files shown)


components/feature/sendtab/src/main/java/mozilla/components/feature/sendtab/SendTabFeature.kt
```diff
@@ -0,0 +1,150 @@
...
+ * so that we can always observe events throughout the application lifecycle.
+ * @param autoPause whether or not the observer should automatically be
+ * paused/resumed with the bound lifecycle.
+ * @param onTabsReceived the callback invoked with new tab(s) are received.
+ */
+class SendTabFeature(
+    context: Context,
+    accountManager: FxaAccountManager,
+    pushFeature: AutoPushFeature? = null,
+    pushService: PushService? = null,
```
grigoryk: IIUC, you use `pushService` below as part of the `AccountObserver`. But, `AutoPushFeature` already has a `shutdown` and `initialize` methods for logging out and starting, which do what you want. Why not just delegate this work to the `pushFeature`, and remove `pushService` as a parameter here?

components/feature/sendtab/src/main/java/mozilla/components/feature/sendtab/SendTabFeature.kt
```diff
@@ -0,0 +1,150 @@
...
+
+        // Always observe the account for device events.
+        accountManager.registerForDeviceEvents(deviceObserver, owner, autoPause)
+
+        pushFeature?.apply {
+            registerForPushMessages(PushType.Services, pushObserver, owner, autoPause)
+            registerForSubscriptions(pushObserver, owner, autoPause)
+
+            // observe the account only if we have the push feature (service is optional)
+            accountManager.register(accountObserver, owner, autoPause)
```
grigoryk: In the parameters to this class, both `pushFeature` and `pushService` are optional. However, here you're linking them together, since `accountObserver` is using `pushService`.. It's very easy to use the current API incorrectly, I think.

Another vote for dropping `pushService` from the parameters!

components/feature/sendtab/src/main/java/mozilla/components/feature/sendtab/SendTabFeature.kt
```diff
@@ -0,0 +1,150 @@
...
+    private val context: Context,
+    private val feature: AutoPushFeature?,
+    private val service: PushService?
+) : SyncAccountObserver {
+    private val logger = Logger("AccountObserver")
+
+    override fun onAuthenticated(account: OAuthAccount, newAccount: Boolean) {
+        service?.start(context)
+
+        if (newAccount) {
```
grigoryk: Can you add a comment describing why we only subscribe once per account lifecycle?

components/feature/sendtab/src/main/java/mozilla/components/feature/sendtab/SendTabFeature.kt
jonalmeida: The `pushService` definitely doesn't need to be there. I added in the docs that this service can be passed in if you want the service to be managed by the feature to be turned on/off when an account is signed in/out.

> Why not just delegate this work to the pushFeature, and remove pushService as a parameter here?

Correct, `AutoPushFeature.initialize` does start the service and we could just use the feature to do that, however the `AutoPushFeature.shutdown` doesn't just shutdown the service, but also unsubscribes all the subscriptions. If we want to remove the optional service from being managed by the feature, I don't think it's necessary to make the AutoPushFeature even start/stop the service.

components/feature/sendtab/src/main/java/mozilla/components/feature/sendtab/SendTabFeature.kt
jonalmeida: The AccountObserver is only added if the feature is available, but the service is part of the nice-to-have part of this feature (see above comment).

jonalmeida:
> But more importantly: let's think about the near-future. Once we start adding more PushTypes (new device connected, password changed), how will we surface those to the application?

Those should be a separate feature that has it's own account manager.
 
> I see value in separating send tab into its own feature, but at the same time it really is just a special case of a broader system.
>
> Did you consider something like feature-account-events which is basically the same as this but can later expand to take handler functions for other events we'd like to surface?

I agree, it is just a "special event" that we regard as special because it's a feature that we care more for (that's also why we call them component "features"!). Having this will let apps that only want this one defining event to work.

> Another nudge in the same direction of generalizing: the SendTabFeature as written doesn't actually allow sending tabs. And this makes sense in the context of what you're doing - tying together push, accounts and event processing.

Good point, maybe this could be better named since the `SendTabUseCases` are what do the sending, but 'Send Tab' is also treated more like it's own product.

jonalmeida:
I'd also like to land this as a product (and android component) feature is to allow TV to easily pop in a smaller code snippet rather than the giant spaghetti we have in Fenix's `BackgroundServices.kt` 😅

grigoryk:
> Those should be a separate feature that has it's own account manager.

What feature will be responsible for updating push registration information on the device record?
If you want to keep "send tab" separate, this bit of functionality would need to live elsewhere.

I'm happy to address these concerns in the future though, once we get there.

jonalmeida:
> > Those should be a separate feature that has it's own account manager.
> 
> What feature will be responsible for updating push registration information on the device record?
> If you want to keep "send tab" separate, this bit of functionality would need to live elsewhere.
> 
> I'm happy to address these concerns in the future though, once we get there.

That's work that should go into the AutopushFeature (which is still on my plate in https://github.com/mozilla-mobile/android-components/issues/3173) since no individual product feature needs to know about that.

grigoryk:
Great. I would land that piece first - or together with this patch. Having that separate from day one would let you have a nicer/simpler sendtab feature API. But you do you, if you want to land this as-is - just make sure to explain your future plans in the comment by the code the handles updating push registration.

Issue #4009, opened 2019-08-02
Do not run Kotlin warningsAsError for debug builds


jonalmeida:
Feels like a productivity loss when you're building locally and debugging code, but the compiler fails (without explanation why) because of an unused variable.

Is it possible to add this as something to be part of a pre-push hook?

dector:
I can suggest using something like `local.properties` to control this behaviour.

PR #4011, opened 2019-08-03
For #4009. Added ability to disable `warningsAsError` for local builds.

dector:
To disable `warningsAsError` for local environment add this line into `local.properties`:
```
mozilla.build.kotlin.warnings.as.error=false
```

What do you think?

[ ] Force `warningsAsError` for `detekt` and `ktlint` tasks / in pre-commit hook.

### Pull Request checklist

### After merge

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,6 +22,8 @@ plugins {
     id("io.gitlab.arturbosch.detekt").version("1.0.0-RC11")
 }
 
+apply plugin: plugins.LocalPropertiesLoader
+
 allprojects {
     repositories {
         google()
@@ -46,8 +48,15 @@ subprojects {
     apply plugin: 'org.jetbrains.dokka-android'
 
     // Enable Kotlin warnings as errors for all modules
-    tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
-        kotlinOptions.allWarningsAsErrors = true
+    def enableWarningsAsErrors = rootProject.ext.getProperties()
+            .get('mozilla.build.kotlin.warnings.as.error', true)
+            .toBoolean()
+    if (enableWarningsAsErrors) {
+        tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
+            kotlinOptions.allWarningsAsErrors = true
+        }
+    } else {
+        println("Warning: `kotlinOptions.allWarningsAsErrors` is not enabled for module ':${project.name}'.")
     }
 
     // Enforce that all (transitive) dependencies are using the same support library version as we do.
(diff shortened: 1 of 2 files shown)


dector:
Is this solution looks like a way to go?

Amejia481:
I like the idea, `warningsAsError` while your developing is pretty annoying!  

> Force warningsAsError for detekt and ktlint tasks / in pre-commit hook.

👍 

@NotWoods @csadilek @jonalmeida @grigoryk  @pocmo @royang51   what do you think?

build.gradle
```diff
@@ -46,8 +48,15 @@ subprojects {
...
-        kotlinOptions.allWarningsAsErrors = true
+    def enableWarningsAsErrors = rootProject.ext.getProperties()
+            .get('mozilla.build.kotlin.warnings.as.error', true)
+            .toBoolean()
+    if (enableWarningsAsErrors) {
+        tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
+            kotlinOptions.allWarningsAsErrors = true
+        }
+    } else {
+        println("Warning: `kotlinOptions.allWarningsAsErrors` is not enabled for module ':${project.name}'.")
```
jonalmeida: `$(project.name)` would resolve to the root project `android-components` right? Is that intentional or did you mean for it to be on each component?

dector: `$project` will be resolved relative to `subproject {}`.
To use root project we need `$rootProject` or `$project.rootProject`. :)

dector:
@jonalmeida Thanks! I'll take a look at `settings.gradle`.

build.gradle
jonalmeida: Oh yes, this is in `allprojects`, thanks!

grigoryk:
I found the warnings helpful, myself. They're annoying - yes, but I'd need to resolve them sooner rather than later anyway, so for me personally it's helpful to be hit with them right away.

I think as-is, without push hooks, this guarantees that folks will simply hit these warnings in TC instead of locally, making the feedback loop longer and wasting TC resources.

jonalmeida:
I'm mostly indifferent so I don't mind commenting out those lines myself manually.

dector:
To sum up:
  - we want an ability easy way to turn off `warningsAsError` while hacking/prototyping; `(1)`
  - we want automatic check (can be used with `detekt` call) to happen before the push with `warningsAsError` enabled. `(2)`

Solutions for `(1)`:
  - adding configuration key to `local.properties` file;
  - _???_.

Solutions for `(2)`:
  - adding pre-push git hook;
  - _???_.

I still don't have "perfect" solution for integrating this ability into workflow smoothly enough but what about minimizing CI usage by doing all heavy checks (e.g. tests) only if `./gradlew detekt` is finished successfully?

cc: @pocmo @Amejia481 @grigoryk @jonalmeida

pocmo:
For now let's punt on this. It doesn't seem worth making our setup more complex at the moment.

I agree with grisha. It still seems more helpful than annoying. And you can still disable them temporarily locally if it gets totally in your way.

dector:
Ah. I was going to write plugin for AS for this. :grin:

Issue #3664, opened 2019-07-05
BrowserIcons: Configure loader differently for private sessions


pocmo:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-3008)

mauryat:
@pocmo curious what exactly the task entails.

pocmo:
@mauryat In private mode we do not save any browsing history. We should also not save icons for private sessions to disk.

mauryat:
Thanks @pocmo ! I'd like to work on this if nobody is taking it.

pocmo:
Sounds good. Do you have an idea how you'd want to expose this? One idea I had was to add a `private` flag to the `IconRequest` (default: `false`?) and use that internally. `IconMessageHandler` is the place where we start a load for a visited website. Here we also have access to the `Session` and could set the flag.

I didn't verify if that is a feasible approach or if there is a better one. Feel free to try other things. :)

mauryat:
Hmm, that's a helpful tip! I was trying to initialize a separate instance of BrowserIcons with an isPrivate flag, but now realize I'd be needlessly creating another heavy-weight instance. Let me try your suggestions.

kbrosnan:
Is this done? @pocmo

csadilek:
Landed back in https://github.com/mozilla-mobile/android-components/pull/4012.

PR #4012, opened 2019-08-03
#3664 Configure Icon loader for private sessions.

mauryat:
Icons being loaded in private sessions will not be cached on the disk.

### Pull Request checklist

### After merge


components/browser/icons/src/main/java/mozilla/components/browser/icons/IconRequest.kt
```diff
@@ -20,7 +20,8 @@ data class IconRequest(
     val url: String,
     val size: Size = Size.DEFAULT,
     val resources: List<Resource> = emptyList(),
-    @ColorInt val color: Int? = null
+    @ColorInt val color: Int? = null,
+    val isPrivate: Boolean = false
```
mauryat: @pocmo general question: What's the policy on API data model changes? For instance, in this IconRequest object, if a new mandatory uninitialized property is added (in this specific scenario if I had omitted the false value for isPrivate), then all callers of the constructor IconRequest() (even in other modules) would have to change. My understanding of reusable components makes me think that data model changes are forbidden. But, wanted your thoughts on it.

components/browser/icons/src/test/java/mozilla/components/browser/icons/processor/DiskIconProcessorTest.kt
```diff
@@ -83,4 +84,40 @@ class DiskIconProcessorTest {
 
         verify(cache, never()).put(any(), any(), any(), eq(icon))
     }
+
+    @Test
+    fun `Icon loaded in private mode is not saved in cache`() {
+        val icon: Icon = mock()
+
+        /* Can be Source.INLINE as well. To ensure that the icon is eligible for caching on the disk. */
+        `when`(icon.source).thenReturn(Icon.Source.DOWNLOAD)
```
mauryat: This is kinda ugly, but was the only option. Ideally, if Mockito allowed mocking static variables, then we could have mocked Icon.shouldCacheOnDisk extension property.

components/browser/icons/src/main/java/mozilla/components/browser/icons/IconRequest.kt
pocmo: They are not forbidden but yeah we try to avoid them if possible, either by adding a default value, like you did or by adding the new API and deprecating the old API. If we can't avoid it then we make the breaking change and explicitly mention that in the cchangelog.

components/browser/icons/src/test/java/mozilla/components/browser/icons/processor/DiskIconProcessorTest.kt
NotWoods: Instead of mocking `Icon` you could just mock `Bitmap` so you don't have to deal with the getters.

```kt
val icon = Icon(mock, source = Icon.Source.DOWNLOAD)
```

pocmo: @mauryat Do you want to try that? Other than that the PR looks good to me.

docs/changelog.md
```diff
@@ -108,6 +108,7 @@ permalink: /changelog/
 
 * **browser-icons**
   * Handles low-memory scenarios by reducing memory footprint.
+  * Ensures icons are not cached on the disk in private sessions.
```
pocmo: Can you move this changelog entry up to 10.0.0-SNAPSHOT (You'll need to rebase this branch)? :)

PR #4028, opened 2019-08-05
1564989: Implement a functional mapping for timing distributions

mdboom:
See also [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1565253) for more context.

This rips out the existing bucketing replacing it with a simple logarithmic function.

The payload (and on-disk representation) also need much less information, there aren't really any "parameters" to the histogram.

### Pull Request checklist

### After merge


components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
```diff
@@ -60,47 +63,40 @@ internal open class TimingDistributionsStorageEngineImplementation(
      *
      * @param metricData the metric information for the timing distribution
      * @param sample the value to accumulate, in nanoseconds
-     * @param timeUnit the [TimeUnit] the sample will be converted to
      */
     @Synchronized
     fun accumulate(
         metricData: CommonMetricData,
-        sample: Long,
-        timeUnit: TimeUnit
```
Dexterp37: Can you describe a bit what happens to the `timeUnit` here? Are we assuming we'll only receive nanoseconds from the API? For the specific case of GeckoView, this might be impossible and might require a double conversions: data in Gecko might have its own time unit (usually milliseconds).

mdboom: Yes -- everything is always nanoseconds now.  The point is to not have to care up front.

We should probably move any unit conversion required for GV to a GV-only path.  I think we would only have to convert once -- from ms or us into ns.

mdboom: I've updated this so it should be compatible with the plans for GV.  For calling `accumulateSamples` only -- the incoming samples are expected to be in the given `timeUnit` and will be converted to `nanoseconds` for storage in the map.

Dexterp37: Thanks for that. I have one last high-level concern: does this mean we'll always be sending the highest possible resolution to the pipeline (i.e. always nanos)?

components/service/glean/build.gradle
```diff
@@ -14,7 +14,7 @@ apply plugin: 'kotlin-android'
  * created during unit testing.
  * This uses a specific version of the schema identified by a git commit hash.
  */
-String GLEAN_PING_SCHEMA_GIT_HASH = "64b852c"
+String GLEAN_PING_SCHEMA_GIT_HASH = "73c5a65"
```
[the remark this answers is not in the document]
mdboom: Yes, it's here: https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/374

components/service/glean/src/main/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngine.kt
mdboom: Yes, it's always nanos.  But of course, as the values get higher, the buckets get larger, so the resolution decreases.

Dexterp37: This changes the semantics of this metric type a bit, though. [In our initial design](https://docs.google.com/document/u/1/d/186GI_2ljdqWqWTBk5P3QjHD3BwpMyd8kRayRAuz1-dk/edit), we discussed about truncating the precision of the data sent out: `day` time unit would only send days out, not nanoseconds.

Since we have no current usage, this should be ok, however this kind of defeats the purpose of `time_unit`. What do you think?

travis79: Isn't the purpose of truncation to limit the amount of data that might become personally identifiable in some way?  I liked the initial design's idea of truncation but if it makes it easier to implement the bucketing this way, I don't have any strong feelings against it aside from making sure we are considering the original intent.

mdboom: Yes -- at this point `time_unit` is only used to deal with the legacy values from GeckoView (and possibly as a hint for displaying metrics in a dashboard someday).

Thinking about truncation is important for timespans, when you get an exact single value back.  For distributions, you only ever get an approximate value anyway (the value of the bucket the sample falls in, which are never unique to a particular user).  And the primary purpose of this change is to take the burden off of the developer of needing to guess what the resolution needs to be up front (which is the source of a lot of unusable data).

IMHO, the intent of time unit in timing distributions evolved in the implementation (before the idea of functional bucketing) to essentially be a single knob that would be a proxy for range (what order of magnitude should we put the buckets in) -- it wasn't really about truncation in anything but the earliest draft.  We probably should have updated the document at that point.

All that said, there is one corner case where there it is possibly too leaky.  In the functional mapping the, buckets for < 32ns are 1:1, so that is the one part of the mapping where we get values at the accuracy of the timer itself (which is nanoseconds).  We could, out of an abundance of caution, add an `sample = (sample < 32) ? 32 : sample` to address this.

Another important change to note is this has a maximum time of 10 minutes, so that the maximum buckets that ever need to be tracked is 313.  This is based on the idea that as timings get that long, they get decreasingly useful as distributions, because the number of samples is so low.  (Again, we have timespans available for things we want to time that are on the order of days).  We should confirm that restriction is sufficient (and of course that will be documented when the documentation gets written.

Dexterp37: I would say it's ok to live with the fact that buckets < 32ns are 1:1. Thanks for the explanation. That would make a great addition to the class documentation and the docs themselves!

mdboom: Great.  I'll try to add some of this to the comments, here, and of course the not-yet-written docs that will go in glean-core.

mdboom:
@Dexterp37: Yeah, will squash before merging (just wanted to make it easier to review my updates).

Also, my plan is to get this approved, then do the glean-core port, then merge both "simultaneously" so that the docs won't be out of date with reality.

Issue #3663, opened 2019-07-05
Investigate Session.icon clearing


pocmo:
When navigating to a new URL we clear the icon on `Session`. But not always is that an actual navigation that will re-trigger the content script.

For example when loading twitch.tv then we start on:
* https://m.twitch.tv/?desktop-redirect=true

We load an icon for that page and then the URL changes to:
*  https://m.twitch.tv/

After that we clear the icon. The content script is not re-triggered. The result is that we do not have an icon for this page.

See also #3662 - should we ignore query parameters?

pocmo:
Follow-up from https://github.com/mozilla-mobile/android-components/issues/3294

mauryat:
Can I work on this (after I'm done with #3664)?

pocmo:
Yes, absolutely. :)

Also filed #4007 which may be interesting too!

mauryat:
Thank you! Will check the other one out too!

PR #4045, opened 2019-08-07
Issue #3663 Clear session icon if URL host changes

mauryat:
Earlier, session icon was cleared whenever URL chagned. But, that resulted in no icon in the scenario of same-host URL redirects. Now, we are clearing session icon only if the scheme and host part of the URL changes.

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/engine/EngineObserver.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/engine/EngineObserver.kt
@@ -6,6 +6,7 @@ package mozilla.components.browser.session.engine
 
 import android.graphics.Bitmap
 import android.os.Environment
+import androidx.core.net.toUri
 import mozilla.components.browser.session.Download
 import mozilla.components.browser.session.Session
 import mozilla.components.browser.session.engine.request.LoadRequestMetadata
@@ -33,6 +34,9 @@ internal class EngineObserver(
     override fun onLocationChange(url: String) {
         if (session.url != url) {
             session.title = ""
+        }
+
+        if (!isHostEquals(session.url, url)) {
             session.icon = null
         }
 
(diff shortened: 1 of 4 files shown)


mauryat:
@pocmo

components/browser/session/src/main/java/mozilla/components/browser/session/engine/EngineObserver.kt
```diff
@@ -49,6 +50,13 @@ internal class EngineObserver(
         }
     }
 
+    private fun isHostEquals(sessionUrl: String, newUrl: String): Boolean {
+        val sessionUri = URI(sessionUrl)
+        val newUri = URI(newUrl)
+
+        return sessionUri.scheme == newUri.scheme && sessionUri.host == newUri.host
```
pocmo: nit: There's an `toUri()`extension method in core-ktx afaik - which will use `Uri` that seems to be more forgiving than `URI`.

components/browser/session/src/main/java/mozilla/components/browser/session/engine/EngineObserver.kt
```diff
@@ -31,7 +32,7 @@ internal class EngineObserver(
 ) : EngineSession.Observer {
 
     override fun onLocationChange(url: String) {
-        if (session.url != url) {
+        if (!isHostEquals(session.url, url)) {
             session.title = ""
             session.icon = null
```
pocmo: I think we should only use that workaround for `icon` and not the title.

mauryat: Makes sense. I'll change that.

components/browser/session/src/main/java/mozilla/components/browser/session/engine/EngineObserver.kt
mauryat: The deal with this extension method is that it calls the Uri.parse() method of the android.net package. So, many of our unit tests would have to be changed to run as instrumented unit tests. Just wanted to check with you before making that change. Would probably require us to strip some methods from the unit test file and move them to a new instrumented test file running under the AndroidJunit4 test runner. Could end up being a more involved change.

pocmo: Using the other runner is enough; they do not need to become instrumented tests running on devices/emulators. We run those unit tests with Robolectric locally. They only reason why we avoid that runner if possible is that it is a bit slower to startup (a couple of seconds) compared to plain unit tests.

PR #4054, opened 2019-08-07
Update SafeIntent to match Android SDK 29

jonalmeida:
There was an API change in the (unreleased) SDK 29 that we compile against, so we need to match the new nullable Android API call.

### Pull Request checklist

### After merge

### DIFF
--- a/components/support/utils/src/main/java/mozilla/components/support/utils/SafeIntent.kt
+++ b/components/support/utils/src/main/java/mozilla/components/support/utils/SafeIntent.kt
@@ -65,7 +65,7 @@ class SafeIntent(val unsafe: Intent) {
     }
 
     fun <T : Parcelable> getParcelableExtra(name: String): T? = safeAccess {
-        unsafe.getParcelableExtra(name) as T
+        unsafe.getParcelableExtra(name) as T?
     }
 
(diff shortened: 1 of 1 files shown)


jonalmeida:
There are more errors that don't make sense to me:
```
/Users/jalmeida/src/android-components/components/support/ktx/src/main/java/mozilla/components/support/ktx/android/org/json/JSONObject.kt: (47, 13): Overload resolution ambiguity:
@Nullable public open fun get(@Nullable key: String): Any? defined in java.util.TreeMap
@Nullable public open fun get(@Nullable key: String!): Any? defined in java.util.TreeMap
```

IIUC, the compiler believes there are two `get` methods on the  Collection class (via the TreeMap), so it fails since it can't infer which one to use. It seems like this a bug in the compiler. 🤔

jonalmeida:
From some minimal sleuthing: since the source for SDK 29 isn't available on androidxref.com I compared the decompiled code to SDK 28 and you can see that there are now annotations on the collection classes.

SDK 28:
Screen Shot 2019-08-08 at 10 59 27 PM

SDK 29:
Screen Shot 2019-08-08 at 10 25 09 PM

Issue #3439, opened 2019-06-14
Add "report" action to crash notification


pocmo:
Follow-up from #3097.

When showing a "crash notification" we could add a "report" action to it to immediately report the crash without the user having to click on the notification to open the crash reporter first.

royang51:
I'm working on this

PR #4087, opened 2019-08-09
For #3439 - Add "report" action to crash notification

rocketsroger:
### Pull Request checklist

### After merge

### DIFF
--- a/components/lib/crash/src/main/AndroidManifest.xml
+++ b/components/lib/crash/src/main/AndroidManifest.xml
@@ -6,6 +6,7 @@
 
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
 
     <application>
         <activity android:name=".prompt.CrashReporterActivity"
@@ -16,6 +17,9 @@
 
         <service android:name=".handler.CrashHandlerService"
             android:process=":mozilla.components.lib.crash.CrashHandler"
+            android:exported="false" />
+
+        <service android:name=".service.SendCrashReportService"
             android:exported="false" />
     </application>
 
--- a/components/lib/crash/src/main/java/mozilla/components/lib/crash/notification/CrashNotification.kt
+++ b/components/lib/crash/src/main/java/mozilla/components/lib/crash/notification/CrashNotification.kt
@@ -15,11 +15,13 @@ import mozilla.components.lib.crash.Crash
 import mozilla.components.lib.crash.CrashReporter
 import mozilla.components.lib.crash.R
 import mozilla.components.lib.crash.prompt.CrashPrompt
+import mozilla.components.lib.crash.service.SendCrashReportService
 import mozilla.components.support.base.ids.notify
 
-private const val NOTIFICATION_CHANNEL_ID = "Crashes"
-private const val NOTIFICATION_TAG = "mozac.lib.crash.CRASH"
 private const val NOTIFICATION_SDK_LEVEL = 29 // On Android Q+ we show a notification instead of a prompt
+
+internal const val NOTIFICATION_CHANNEL_ID = "Crashes"
+internal const val NOTIFICATION_TAG = "mozac.lib.crash.CRASH"
 
 internal class CrashNotification(
     private val context: Context,
@@ -28,10 +30,20 @@ internal class CrashNotification(
 ) {
     fun show() {
         val pendingIntent = PendingIntent.getActivity(
-            context, 0, CrashPrompt.createIntent(context, crash), 0
+                context, 0, CrashPrompt.createIntent(context, crash), 0
         )
 
-        val channel = ensureChannelExists()
+        val reportPendingIntent = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            PendingIntent.getForegroundService(
+                    context, 0, SendCrashReportService.createReportIntent(context, crash), 0
+            )
+        } else {
+            PendingIntent.getService(
+                    context, 0, SendCrashReportService.createReportIntent(context, crash), 0
+            )
+        }
+
+        val channel = ensureChannelExists(context)
 
         val notification = NotificationCompat.Builder(context, channel)
             .setContentTitle(context.getString(R.string.mozac_lib_crash_dialog_title, configuration.appName))
@@ -39,29 +51,13 @@ internal class CrashNotification(
             .setPriority(NotificationCompat.PRIORITY_DEFAULT)
             .setCategory(NotificationCompat.CATEGORY_ERROR)
             .setContentIntent(pendingIntent)
+            .addAction(R.drawable.mozac_lib_crash_notification, context.getString(
+                    R.string.mozac_lib_crash_notification_action_report), reportPendingIntent)
             .setAutoCancel(true)
             .build()
 
         NotificationManagerCompat.from(context)
             .notify(context, NOTIFICATION_TAG, notification)
-    }
-
-    private fun ensureChannelExists(): String {
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-            val notificationManager: NotificationManager = context.getSystemService(
-                Context.NOTIFICATION_SERVICE
-            ) as NotificationManager
-
-            val channel = NotificationChannel(
-                NOTIFICATION_CHANNEL_ID,
-                context.getString(R.string.mozac_lib_crash_channel),
-                NotificationManager.IMPORTANCE_DEFAULT
-            )
-
-            notificationManager.createNotificationChannel(channel)
-        }
-
-        return NOTIFICATION_CHANNEL_ID
     }
 
     companion object {
@@ -81,6 +77,24 @@ internal class CrashNotification(
                 // This is a fatal native crash. We may not be able to launch an activity from here.
                 else -> crash is Crash.NativeCodeCrash && crash.isFatal
             }
+        }
+
+        fun ensureChannelExists(context: Context): String {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                val notificationManager: NotificationManager = context.getSystemService(
+                        Context.NOTIFICATION_SERVICE
+                ) as NotificationManager
+
+                val channel = NotificationChannel(
+                        NOTIFICATION_CHANNEL_ID,
+                        context.getString(R.string.mozac_lib_crash_channel),
+                        NotificationManager.IMPORTANCE_DEFAULT
+                )
+
+                notificationManager.createNotificationChannel(channel)
+            }
+
+            return NOTIFICATION_CHANNEL_ID
         }
     }
 }
--- a/components/lib/crash/src/main/java/mozilla/components/lib/crash/service/SendCrashReportService.kt
+++ b/components/lib/crash/src/main/java/mozilla/components/lib/crash/service/SendCrashReportService.kt
@@ -1,0 +1,91 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.lib.crash.service
+
+import android.app.Service
+import android.content.Context
+import android.content.Intent
+import android.os.Build
+import android.os.IBinder
+import androidx.annotation.VisibleForTesting
+import androidx.core.app.NotificationCompat
+import androidx.core.app.NotificationManagerCompat
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+import mozilla.components.lib.crash.Crash
+import mozilla.components.lib.crash.CrashReporter
+import mozilla.components.lib.crash.R
+import mozilla.components.lib.crash.notification.CrashNotification
+import mozilla.components.lib.crash.notification.NOTIFICATION_TAG
+import mozilla.components.support.base.ids.NotificationIds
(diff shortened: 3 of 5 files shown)


components/lib/crash/src/main/java/mozilla/components/lib/crash/handler/CrashHandlerService.kt
```diff
@@ -29,10 +31,18 @@ class CrashHandlerService : IntentService(WORKER_THREAD_NAME) {
 
         intent.extras?.let { extras ->
             val crash = Crash.NativeCodeCrash.fromBundle(extras)
+            val skipPrompt = extras.getBoolean(Crash.INTENT_EXTRA_SKIP_PROMPT, false)
+
+            if (skipPrompt) {
+                val notificationManager = getSystemService(Context.NOTIFICATION_SERVICE)
+                        as NotificationManager
+                notificationManager.cancelAll()
```
pocmo: Cancelling all notification is a bit aggressive here.

components/lib/crash/src/main/java/mozilla/components/lib/crash/handler/CrashHandlerService.kt
```diff
@@ -29,10 +31,18 @@ class CrashHandlerService : IntentService(WORKER_THREAD_NAME) {
...
+                sendBroadcast(Intent(Intent.ACTION_CLOSE_SYSTEM_DIALOGS))
```
pocmo: What's the reason for this broadcast here?  🤔

components/lib/crash/src/main/java/mozilla/components/lib/crash/handler/CrashHandlerService.kt
```diff
@@ -29,10 +31,18 @@ class CrashHandlerService : IntentService(WORKER_THREAD_NAME) {
 
         intent.extras?.let { extras ->
             val crash = Crash.NativeCodeCrash.fromBundle(extras)
+            val skipPrompt = extras.getBoolean(Crash.INTENT_EXTRA_SKIP_PROMPT, false)
+
+            if (skipPrompt) {
```
pocmo: I'm concerned about reusing the `CrashHandlerService` class here: It's main purpose is to receive crashes from GeckoView (hence the `Crash.NativeCodeCrash.fromBundle()` above) and forward it to the `CrashReporter` instance.

With this patch it also receives crashes from the notification (which may not be `Crash.NativeCodeCrash`!). Since we kill this process immediately I wonder if we'd even be able to report any crashes (which may happen asynchronously).

We may want to consider using `CrashReporter.submitReport()` (like `CrashReporterActivity` does) and wait for it to completely. Not sure from where though? Maybe a foreground service? Or a workmanager worker (which would still need a service to schedule it?)? A workmanager worker would have the benefit that we could fix some of our other lib-crash issues (e.g. CrashReporterActivity blocking, CrashReporterActivity trying only once, ..). What do you think?

components/lib/crash/src/main/java/mozilla/components/lib/crash/handler/CrashHandlerService.kt
royang51: Once the report button is clicked, it retracts the notification drawer.  I will agree if you think this is not needed.

components/lib/crash/src/main/java/mozilla/components/lib/crash/handler/CrashHandlerService.kt
royang51: I'll look into if I can just cancel the crash notification.  Thanks,

components/lib/crash/src/main/java/mozilla/components/lib/crash/handler/CrashHandlerService.kt
royang51: I see, I've misunderstood the role of CrashHandlerService then.  I'll look into both the foreground service and the worker idea.  Thanks,

components/lib/crash/src/main/AndroidManifest.xml
```diff
@@ -17,6 +18,10 @@
         <service android:name=".handler.CrashHandlerService"
             android:process=":mozilla.components.lib.crash.CrashHandler"
             android:exported="false" />
+
+        <service android:name=".service.SendCrashReportService"
+            android:process=":mozilla.components.lib.crash.CrashHandler"
```
pocmo: nit: Let's not define a process here. The `CrashHandlerService´ needs that to escape from the crashing process. The service for sending does not have this constraint.

components/lib/crash/src/main/java/mozilla/components/lib/crash/notification/CrashNotification.kt
```diff
@@ -28,42 +31,36 @@ internal class CrashNotification(
...
-        val channel = ensureChannelExists()
+        val reportPendingIntent = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            PendingIntent.getForegroundService(
+                    context, REPORT_REQUEST_CODE, CrashPrompt.createReportIntent(context, crash), 0
+            )
+        } else {
+            PendingIntent.getService(
+                    context, REPORT_REQUEST_CODE, CrashPrompt.createReportIntent(context, crash), 0
+            )
+        }
```
pocmo: Are the request codes needed? We never seem to not read them in this setup.

components/lib/crash/src/main/java/mozilla/components/lib/crash/notification/CrashNotification.kt
```diff
@@ -28,42 +31,36 @@ internal class CrashNotification(
...
 
-        val channel = ensureChannelExists()
+        val reportPendingIntent = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            PendingIntent.getForegroundService(
+                    context, REPORT_REQUEST_CODE, CrashPrompt.createReportIntent(context, crash), 0
+            )
+        } else {
+            PendingIntent.getService(
+                    context, REPORT_REQUEST_CODE, CrashPrompt.createReportIntent(context, crash), 0
+            )
```
pocmo: nit: We could consider moving the creation of such a pending intent behind a helper. Maybe even in `support-utils`.

components/lib/crash/src/main/java/mozilla/components/lib/crash/prompt/CrashPrompt.kt
```diff
@@ -29,6 +30,13 @@ internal class CrashPrompt(
             return intent
         }
 
+        fun createReportIntent(context: Context, crash: Crash): Intent {
+            val intent = Intent(context, SendCrashReportService::class.java)
+            crash.fillIn(intent)
+
+            return intent
+        }
```
pocmo: nit: I wonder if we should move this somewhere else (maybe to `SendCrashReportService`?) since this is now not related `CrashPrompt`.

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/SendCrashReportService.kt
```diff
@@ -0,0 +1,81 @@
...
+                    .setContentTitle(getString(R.string.mozac_lib_crash_dialog_checkbox,
+                            crashReporter.promptConfiguration.organizationName))
+                    .setSmallIcon(R.drawable.mozac_lib_crash_notification)
+                    .setPriority(NotificationCompat.PRIORITY_DEFAULT)
+                    .setCategory(NotificationCompat.CATEGORY_ERROR)
+                    .setAutoCancel(true)
+                    .setProgress(0, 0, true)
+                    .build()
+
+            startForeground(NOTIFICATION_ID, notification)
```
pocmo: To avoid a notification ID conflict take a look at our `NotificationIds` helper in `support-base`. It can give you a unique ID based on a `String` tag.

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/SendCrashReportService.kt
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    private var reporterCoroutineContext: CoroutineContext = EmptyCoroutineContext
+
+    override fun onStartCommand(intent: Intent, flags: Int, startId: Int): Int {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            val channel = CrashNotification.ensureChannelExists(this)
+            val notification = NotificationCompat.Builder(this, channel)
+                    .setContentTitle(getString(R.string.mozac_lib_crash_dialog_checkbox,
+                            crashReporter.promptConfiguration.organizationName))
```
pocmo: Re-using the checkbox string may work in English but maybe not work in other translations. It's probably safer to create a separate string.

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/SendCrashReportService.kt
```diff
@@ -0,0 +1,81 @@
...
+        intent.extras?.let { extras ->
+            val crash = Crash.NativeCodeCrash.fromBundle(extras)
+            NotificationManagerCompat.from(this).cancel(this, NOTIFICATION_TAG)
+
+            sendCrashReport(crash) {
+                stopSelf()
+            }
+        } ?: logger.error("Received intent with null extras")
+
+        return START_STICKY
```
pocmo: Do we really want STICKY here? My assumption is that we never want Android to re-launch this service on its own and return NOT_STICKY here? 🤔

components/lib/crash/src/main/AndroidManifest.xml
royang51: agreed.  Thanks,

components/lib/crash/src/main/java/mozilla/components/lib/crash/notification/CrashNotification.kt
royang51: Agreed.  I'm launching a different receiver with the PendingIntent so I don't need to have different request code anymore.

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/SendCrashReportService.kt
royang51: Agreed.  I'll create new string

components/lib/crash/src/main/java/mozilla/components/lib/crash/prompt/CrashPrompt.kt
royang51: Yes, makes sense

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/SendCrashReportService.kt
royang51: Done.

components/lib/crash/src/main/java/mozilla/components/lib/crash/notification/CrashNotification.kt
royang51: Had some issues moving this to support-util.  Couldn't find a clean way because of the Crash class dependency.

rocketsroger: opened #4113 to track this.

components/lib/crash/src/main/java/mozilla/components/lib/crash/notification/CrashNotification.kt
```diff
@@ -15,55 +15,51 @@ import mozilla.components.lib.crash.Crash
...
+
 internal class CrashNotification(
     private val context: Context,
     private val crash: Crash,
     private val configuration: CrashReporter.PromptConfiguration
 ) {
     fun show() {
         val pendingIntent = PendingIntent.getActivity(
-            context, 0, CrashPrompt.createIntent(context, crash), 0
+                context, 0, CrashPrompt.createIntent(context, crash), 0
```
pocmo: No need to change that in the commit. But this indentation seems to come from the default AS formatting that is a bit different than what ktlint does.

You can get Android Studio to follow ktlint (on Mac) like this:
```
brew install ktlint
ktlint --apply-to-idea-project --android
```

components/lib/crash/src/main/res/values/strings.xml
```diff
@@ -21,4 +21,6 @@
     <!-- Label of a notification action/button that will send the crash report to Mozilla. -->
     <string name="mozac_lib_crash_notification_action_report">Report</string>
 
+    <!-- Label of notification showing that the crash report service is running. -->
+    <string name="mozac_lib_send_crash_report_in_progress">Sending crash report to %1$s</string>
```
pocmo: 👍 for adding the comment here. Those comments will show up when our translators translate the strings in Pontoon: https://pontoon.mozilla.org/projects/android-l10n/

Please mention in the comment what `%1$s` will be replaced with in the UI. This will help the translators to understand the context better. :)

rocketsroger: Good catch thanks.

components/lib/crash/src/main/java/mozilla/components/lib/crash/notification/CrashNotification.kt
rocketsroger: After applying and restarting android studio.  It still wants to do double intent for this line.  I'll look into why.  Thanks for the ktlint apply commands.

rocketsroger:
> Something looks not right here. There are now 29 commits in this PR.

Yes, something went wrong when I reorder git history on my branch.  Let me take a look.  Thanks,

pocmo:
Looks good now. Can you re-force-push your branch (or rebase) so that we do not get all those merge commits merged in? I just disabled the "update branch" button so that it's now gone :D

rocketsroger:
> Looks good now. Can you re-force-push your branch (or rebase) so that we do not get all those merge commits merged in? I just disabled the "update branch" button so that it's now gone :D

Done.  Only 1 commit now for this change.  Please let me know if its not clean.  Thanks,

PR #4089, opened 2019-08-09
Closes #3923: Create async API for SearchEngineManager

### LINKED ISSUE #3923: Replace kotlin_coroutines_test usages with testing_coroutines
In `Dependencies.kt` we seem to have a duplicate references for the `kotlinx-coroutine-test` library.

One called [`kotlin_coroutines_test`](https://github.com/mozilla-mobile/android-components/blob/master/buildSrc/src/main/java/Dependencies.kt#L66) and [`testing_coroutines`](https://github.com/mozilla-mobile/android-components/blob/master/buildSrc/src/main/java/Dependencies.kt#L73).

We should replace [all references](https://searchfox.org/mozilla-mobile/search?path=&q=kotlin_coroutines_test) to `kotlin_coroutines_test` since the testing one follows the naming convention we already use.

csadilek:
@pocmo I think this is basically what we want. Keeping the old methods and adding  async (suspending) alternatives. Just opening as draft as there were too many 🔥 🔥 this week. So, I'll take another look on Monday.

### DIFF
--- a/components/browser/search/src/main/java/mozilla/components/browser/search/SearchEngineManager.kt
+++ b/components/browser/search/src/main/java/mozilla/components/browser/search/SearchEngineManager.kt
@@ -24,6 +24,7 @@ import kotlin.coroutines.CoroutineContext
 /**
  * This class provides access to a centralized registry of search engines.
  */
+@Suppress("TooManyFunctions")
 class SearchEngineManager(
     private val providers: List<SearchEngineProvider> = listOf(
             AssetsSearchEngineProvider(LocaleSearchLocalizationProvider())),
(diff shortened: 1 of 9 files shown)


csadilek:
@pocmo ok this is ready for review. 

As discussed, we won't deprecate the synchronous API. I've also decided to not break existing API. I think it makes sense to support both cases in `addSearchProvider`. Either the app already knows/has a `searchEngine` or it wants to load the default search engine using the `SearchEngineManager`. In our browsers, we can now use the new API to lazily load the default search engine.

There's no change in behaviour for the existing API, just refactorings i.e. we would have already returned an empty list for search providers that don't support suggestions. This logic just moved into `SearchSuggestionClient`, which no longer throws if created directly (for a provider that doesn't support it), but nobody is using this class directly, so it should be the same behaviour as before.

Issue #1200, opened 2018-10-29
Let bot do: Rebase -> All Green? -> Land


pocmo:
It's too time consuming to manually update a PR, check back later and merge or update PR again. It would be much nicer if we could do something like:
* Add [needs landing] label
* Bot comes along and:
  * Checks out PR
  * Rebases PR
  * Builds, runs tests etc.
* If all green: Bot merges PR
* If red: Bot removes label and comments

There are some existing solutions:
* Servo uses `bors`: https://github.com/graydon/bors
* @jonalmeida mentioned Homu: https://github.com/barosl/homu

pocmo:
Another option would be to do that from a taskcluster task that is triggered by a hook. We may want to have a build cache first, so that we can build rebased PRs pretty quickly: #638

jonalmeida:
I'm not particular to either one, but homu might work better with a taskcluster as our custom CI. I'll ask around in the Rust #infra channel if anyone has done this already.

jonalmeida:
> bors can integrate without too much of a problem with any CI that posts either commit statuses or checks

So we should be good.

> well, our bot account is called bors, but the software behind it is actually homu

Turns out their the same thing!

jonalmeida:
There was a recommendation to use bors-ng which has taskcluster support already: https://github.com/bors-ng/bors-ng/blob/45de72ba106e0cabe7fddc4152c09e9c6adef83f/bors.toml#L1

Servo's bors config: https://github.com/servo/saltfs/blob/2a56d14c47d637dabe55c2f83a3088d35896cfc9/homu/files/cfg.toml#L286-L287

> This is kind of soft-blocked on utilizing the cache API: We need fast builds to make this work without causing more pain.

Seems fair. I was curious about the tools required so I did some basic perusing. 😃

pocmo:
I played with bors-ng a bit in a different repository. This looks great and I think we should try it! I'll add this issue to the triage list. :)

jonalmeida:
https://i.giphy.com/media/5GoVLqeAOo6PK/giphy.webp

pocmo:
Closing: Bors is running now. Let's see how it is going to be. :)

PR #4099, opened 2019-08-12
Issue #1200: Disable taskcluster checks for bors tmp branches.

pocmo:
Those branches are created temporarily and then deleted by bors. Running taskcluster tasks for them is a waste of energy and also slows down other tasks that are waiting for the resources.


pocmo:
bors r=Amejia481

pocmo:
bors retry

PR #4101, opened 2019-08-12
1571422: Implement memory distribution metric

mdboom:
This adds a new memory distribution metric, which uses the identical functional
bucketing from the timing distribution metric, except the units for conversion
are different and it has a different maximum (1TB).

In addition to the obvious additions, there is some refactoring going on here:

- The distribution data classes from timing dist. and custom dist. were moved
  to their own package and renamed `FunctionalHistogram` and
  `PrecomputedHistogram` respectively, so that they could be shared between
  different metric types.  This prompted some moving around of tests.

- It was discovered that these classes stored the name and category of the
  metric, but they really don't need to.  All we need is a way to round-trip
  the bucket positions, values and sum.

- `HistogramBase` was renamed to `HistogramMetricBase`, since it's not the base
  of these new histogram data classes, but of metrics that use a histogram.

---
<!- Text above this line will be added to the commit once "bors" merges this PR -->

### Pull Request checklist

### After merge


components/service/glean/src/main/java/mozilla/components/service/glean/utils/MemoryUtils.kt
```diff
@@ -0,0 +1,25 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.service.glean.utils
+
+import mozilla.components.service.glean.private.MemoryUnit
+
+/**
+ * Convenience method to convert a memory size in a different unit to bytes.
+ *
+ * @param memoryUnit the unit the value is in
```
travis79: nit: use `[MemoryUnit]` to link it to the class to view possible values

components/service/glean/src/test/java/mozilla/components/service/glean/histogram/FunctionalHistogramTest.kt
```diff
@@ -0,0 +1,68 @@
...
+import androidx.test.core.app.ApplicationProvider
+import mozilla.components.service.glean.testing.GleanTestRule
+import org.junit.Assert.assertEquals
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import java.lang.Math.pow
+
+@RunWith(RobolectricTestRunner::class)
```
travis79: Can this be done with AndroidJUnit test runner?  I know that the rest of Glean still has a lot of Robolectric but the rest of android-components has already switched over.

components/service/glean/src/test/java/mozilla/components/service/glean/histogram/PrecomputedHistogramTest.kt
```diff
@@ -0,0 +1,137 @@
...
+import androidx.test.core.app.ApplicationProvider
+import mozilla.components.service.glean.private.HistogramType
+import mozilla.components.service.glean.testing.GleanTestRule
+import org.junit.Assert.assertEquals
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
```
travis79: Can this be done with AndroidJUnit test runner? I know that the rest of Glean still has a lot of Robolectric but the rest of android-components has already switched over.

components/service/glean/src/test/java/mozilla/components/service/glean/histogram/PrecomputedHistogramTest.kt
```diff
@@ -0,0 +1,137 @@
...
+        // Accumulate some samples to populate sum and values properties
+        tdd.accumulate(1L)
+        tdd.accumulate(2L)
+        tdd.accumulate(3L)
+
+        // Convert to JSON object using toJsonObject()
+        val jsonTdd = tdd.toJsonObject()
+
+        // Verify properties
+        assertEquals("JSON bucket count must match Custom Distribution bucket count",
```
travis79: Is the `Custom Distribution` the correct term here?

components/service/glean/src/test/java/mozilla/components/service/glean/private/MemoryDistributionMetricTypeTest.kt
```diff
@@ -0,0 +1,196 @@
...
+import org.junit.Assert.assertFalse
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import java.lang.NullPointerException
+
+@ObsoleteCoroutinesApi
+@ExperimentalCoroutinesApi
+@RunWith(RobolectricTestRunner::class)
```
travis79: Can this be done with AndroidJUnit test runner? I know that the rest of Glean still has a lot of Robolectric but the rest of android-components has already switched over.

components/service/glean/src/test/java/mozilla/components/service/glean/storages/MemoryDistributionsStorageEngineTest.kt
```diff
@@ -0,0 +1,279 @@
...
+import org.junit.Assert.assertTrue
+import org.junit.Assert.assertEquals
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.ArgumentMatchers
+import org.mockito.Mockito
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
```
travis79: Can this be done with AndroidJUnit test runner? I know that the rest of Glean still has a lot of Robolectric but the rest of android-components has already switched over.

components/service/glean/src/test/java/mozilla/components/service/glean/storages/MemoryDistributionsStorageEngineTest.kt
```diff
@@ -0,0 +1,279 @@
...
+        val td = FunctionalHistogram()
+
+        val persistedSample = mapOf(
+            "store1#telemetry.invalid_string" to "invalid_string",
+            "store1#telemetry.invalid_bool" to false,
+            "store1#telemetry.null" to null,
+            "store1#telemetry.invalid_int" to -1,
+            "store1#telemetry.invalid_list" to listOf("1", "2", "3"),
+            "store1#telemetry.invalid_int_list" to "[1,2,3]",
+            "store1#telemetry.invalid_td_values" to "{\"values\":{\"0\": \"nope\"},\"sum\":0}",
```
travis79: nit: The `td` in `invalid_td_values` here probably referred to timing distribution...

components/service/glean/src/test/java/mozilla/components/service/glean/storages/MemoryDistributionsStorageEngineTest.kt
```diff
@@ -0,0 +1,279 @@
...
+            "store1#telemetry.invalid_td_sum" to "{\"values\":{},\"sum\":\"nope\"}",
```
travis79: Same nit with `td` as above...

components/service/glean/src/test/java/mozilla/components/service/glean/storages/MemoryDistributionsStorageEngineTest.kt
```diff
@@ -0,0 +1,279 @@
...
+                clearStore = true
+            )
+            assertEquals(1, snapshot!!.size)
+            assertEquals(1L, snapshot["telemetry.test_memory_distribution"]?.values!![1024])
+        }
+    }
+
+    @Test
+    fun `deserializer should correctly parse memory distributions`() {
+        val td = FunctionalHistogram()
```
travis79: nit: `td` probably refers to timing distribution.

components/service/glean/src/test/java/mozilla/components/service/glean/storages/MemoryDistributionsStorageEngineTest.kt
```diff
@@ -0,0 +1,279 @@
...
+                category = "telemetry",
+                lifetime = Lifetime.User,
+                name = "test_memory_distribution",
+                sendInPings = listOf("store1", "store2"),
+                memoryUnit = MemoryUnit.Kilobyte
+            )
+
+            // Using the FunctionalHistogram object here to easily turn the object into JSON
+            // for comparison purposes.
+            val td = FunctionalHistogram()
```
travis79: nit: `td` referring to timing distribution?

components/service/glean/src/test/java/mozilla/components/service/glean/storages/MemoryDistributionsStorageEngineTest.kt
```diff
@@ -0,0 +1,279 @@
...
+            )
+        }
+
+        // Create a new instance of storage engine to verify serialization to storage rather than
+        // to the cache
+        run {
+            val storageEngine = MemoryDistributionsStorageEngineImplementation()
+            storageEngine.applicationContext = ApplicationProvider.getApplicationContext()
+
+            val td = FunctionalHistogram()
```
travis79: nit: `td` -> timing distribution?

components/service/glean/src/test/java/mozilla/components/service/glean/storages/TimingDistributionsStorageEngineTest.kt
```diff
@@ -26,7 +27,6 @@ import org.junit.runner.RunWith
 import org.mockito.ArgumentMatchers
 import org.mockito.Mockito
 import org.robolectric.RobolectricTestRunner
-import java.lang.Math.pow
 
 @RunWith(RobolectricTestRunner::class)
```
travis79: Can this be ran with `AndroidJUnit` test runner (see other comments)

components/service/glean/src/main/java/mozilla/components/service/glean/histogram/FunctionalHistogram.kt
```diff
@@ -0,0 +1,174 @@
...
+ * and deserialize data to the correct format for transport and storage, as well
+ * as performing the calculations to determine the correct bucket for each sample.
+ *
+ * The bucket index of a given sample is determined with the following function:
+ *
+ *     i = ⌊n log₂(𝑥)⌋
+ *
+ * In other words, there are n buckets for each power of 2 magnitude.
+ *
+ * The value of 8 for n was determined experimentally based on existing data to have sufficient
```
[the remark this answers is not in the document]
mdboom: Yeah, we can link to this bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1565253

components/service/glean/src/main/java/mozilla/components/service/glean/histogram/PrecomputedHistogram.kt
```diff
@@ -0,0 +1,274 @@
...
+
+    /**
+     * Helper function to build the [PrecomputedHistogram] into a JSONObject for sending in the
+     * ping payload. Compared to [toJsonObject] which is designed for lossless roundtripping:
+     *
+     *   - this does not include the bucketing parameters
+     *   - all buckets [0, max) are inserted into values
+     */
+    internal fun toJsonPayloadObject(): JSONObject {
+        // Include all buckets [0, max), where max is the maximum bucket with
```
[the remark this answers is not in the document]
mdboom: Yes -- square bracket indicates inclusive, curved indicates exclusive.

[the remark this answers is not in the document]
travis79: TIL 😄

components/service/glean/src/main/java/mozilla/components/service/glean/private/MemoryDistributionMetricType.kt
```diff
@@ -0,0 +1,112 @@
...
+import mozilla.components.service.glean.Dispatchers
+import mozilla.components.service.glean.histogram.FunctionalHistogram
+import mozilla.components.service.glean.storages.MemoryDistributionsStorageEngine
+import mozilla.components.support.base.log.logger.Logger
+
+/**
+ * This implements the developer facing API for recording memory distribution metrics.
+ *
+ * To prevent the number of buckets from being unbounded, values larger than 1 TB
+ * are truncated to 1 TB.
```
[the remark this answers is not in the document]
mdboom: No -- desktop has explicitly an set min/max/bucket_count.  Here we're trying to do something "parameterless" so that data is less likely to end up in an under or overflow bucket.

We will need to confirm that 1TB is sufficient -- it is for the histograms on the GeckoView shortlist.

components/service/glean/src/test/java/mozilla/components/service/glean/histogram/PrecomputedHistogramTest.kt
mdboom: Yeah -- should probably be "precomputed histogram"

mdboom:
@travis79: It looks like nothing in Glean is using the Android JUnit runner yet, and it's going to require some build reconfiguration (it doesn't just work to import it).  So maybe we take that as a follow-on bug?

travis79:
> @travis79: It looks like nothing in Glean is using the Android JUnit runner yet, and it's going to require some build reconfiguration (it doesn't just work to import it). So maybe we take that as a follow-on bug?

I'm fine with that, I was just thinking that if we did anything new it may be better to go with AndroidJUnit, but follow-up work is fine.

components/service/glean/src/main/java/mozilla/components/service/glean/histogram/FunctionalHistogram.kt
```diff
@@ -0,0 +1,185 @@
...
+            "values" to values.mapKeys { "${it.key}" },
+            "sum" to sum
+        ))
+    }
+
+    /**
+     * Helper function to build the [FunctionalHistogram] into a JSONObject for sending in the
+     * ping payload.
+     *
+     * @return The histogram as JSON for a ping payload
```
Dexterp37: nit `as [JSONObject] for..`

components/service/glean/src/main/java/mozilla/components/service/glean/histogram/FunctionalHistogram.kt
```diff
@@ -0,0 +1,185 @@
...
+ * and deserialize data to the correct format for transport and storage, as well
+ * as performing the calculations to determine the correct bucket for each sample.
+ *
+ * The bucket index of a given sample is determined with the following function:
+ *
+ *     i = ⌊n log₂(𝑥)⌋
+ *
+ * In other words, there are n buckets for each power of 2 magnitude.
+ *
+ * The value of 8 for n was determined experimentally based on existing data to have sufficient
```
Dexterp37: Shouldn't the coefficient be 8 for timing and 16 for memory? See [this comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1565253#c5)

components/service/glean/src/main/java/mozilla/components/service/glean/histogram/FunctionalHistogram.kt
```diff
@@ -0,0 +1,185 @@
...
+    // map from bucket limits to accumulated values
+    val values: MutableMap<Long, Long> = mutableMapOf(),
+    var sum: Long = 0
+) {
+    companion object {
+        // The base of the logarithm used to determine bucketing
+        internal const val LOG_BASE = 2.0
+
+        // The buckets per each order of magnitude of the logarithm.
+        internal const val BUCKETS_PER_MAGNITUDE = 8.0
```
Dexterp37: Does this need to be configurable?

components/service/glean/src/main/java/mozilla/components/service/glean/histogram/FunctionalHistogram.kt
```diff
@@ -0,0 +1,185 @@
...
+         */
+        internal fun bucketIndexToBucketMinimum(bucketIndex: Long): Long {
+            return pow(EXPONENT, bucketIndex.toDouble()).toLong()
+        }
+
+        /**
+         * Maps a sample to the minimum value of the bucket it belongs in.
+         *
+         * @param sample The sample value
+         @ @return the minimum value of the bucket the sample belongs in
```
Dexterp37: nit: `@ @return` should be `* @return`

components/service/glean/src/main/java/mozilla/components/service/glean/histogram/PrecomputedHistogram.kt
```diff
@@ -0,0 +1,276 @@
...
+ * storage, as well as including helper functions to calculate the bucket sizes.
+ *
+ * @param rangeMin the minimum value that can be represented
+ * @param rangeMax the maximum value that can be represented
+ * @param bucketCount total number of buckets
+ * @param histogramType the [HistogramType] representing the bucket layout
+ * @param values a map containing the bucket index mapped to the accumulated count
+ * @param sum the accumulated sum of all the samples in the custom distribution
+ */
+data class PrecomputedHistogram(
```
Dexterp37: Bikeshedding: what if we'll need to precompute `bucketIndexToBucketMinimum` for the `FunctionalHistogram` due to performance reasons? Would it make the latter pre-computed too?

components/service/glean/src/main/java/mozilla/components/service/glean/private/HistogramMetricBase.kt
```diff
@@ -8,7 +8,7 @@ package mozilla.components.service.glean.private
  * A common interface to be implemented by all the histogram-like metric types
  * supported by the Glean SDK.
  */
-interface HistogramBase {
+interface HistogramMetricBase {
```
Dexterp37: This would require a change in the Glean parsers, as this name is used in the generated GV adapter code. Can we leave this as it is or change the parsers accordingly?

components/service/glean/src/main/java/mozilla/components/service/glean/histogram/FunctionalHistogram.kt
mdboom: Oh, thanks for catching that -- I hadn't seen Corey's response...

This will require some retooling to make parameterize that...

components/service/glean/src/main/java/mozilla/components/service/glean/histogram/PrecomputedHistogram.kt
mdboom: In that case, I would use the `PrecomputedHistogram` implementation and add a new `histogram_type` enum to handle the different bucketing (probably with some feature so we only have a single static buckets list, since it will be much larger than the other ones).

I'm skeptical that the binary search will beat out the log calculation, though, since the former is memory bound...  But all that is conjecture without benchmarking, of course.

components/service/glean/src/main/java/mozilla/components/service/glean/private/HistogramMetricBase.kt
mdboom: Good catch.  I'll update this over in https://github.com/mozilla/glean_parser/pull/85

components/service/glean/src/main/java/mozilla/components/service/glean/histogram/FunctionalHistogram.kt
mdboom: Yeah, I guess it does.  I've updated this PR so it's a parameter...

components/service/glean/src/main/java/mozilla/components/service/glean/histogram/FunctionalHistogram.kt
Dexterp37: nit: "The value of 8" needs update, since that is now parametrized

PR #4106, opened 2019-08-13
fixes #1444 exposed title in HitResult and set image title as fragment title 

### LINKED ISSUE #1444: HitResult: Can we expose the title?
For images it would be nice to get the value of the title tag. Famous example is XKCD.

For GeckoView we may need an API.

mcarare:
related Fenix issues : mozilla-mobile/fenix#4684 mozilla-mobile/fenix#580
added title in class Image & added title param in all usages. made it optional, with default null value, to keep WebView-based implementation of EngineView unchanged (SystemEngineView.kt)
added title param in handleLongClick from 'element.title', as it is exposed by element information from gecko
in ContextMenuCandidate added case for IMAGE. returning src string when title is null, blank or starting with http

---

### Pull Request checklist

### After merge


pocmo:
nit: Please do not mention Fenix issues in the commit title since GitHub does not understand that and then links the PR to unrelated AC issues. Use the commit message title to describe the change you did to the code (not the bug). You change the PR title here a bunch of times; but that doesn't change the commit title. :)

components/feature/contextmenu/src/main/java/mozilla/components/feature/contextmenu/ContextMenuCandidate.kt
```diff
@@ -262,6 +262,8 @@ private fun HitResult.isLink(): Boolean =
 internal fun HitResult.getLink(): String = when (this) {
     is HitResult.UNKNOWN -> src
     is HitResult.IMAGE_SRC -> uri
+    is HitResult.IMAGE ->
+        if (title.isNullOrBlank() || title?.startsWith("http", true) == true) src else title.toString()
```
jonalmeida: `|| title?.startsWith("http", true) == true` I don't think we need this part right? If the title is null or blank that should be enough.

jonalmeida:
> nit: Please do not mention Fenix issues in the commit title since GitHub does not understand that and then links the PR to unrelated AC issues. Use the commit message title to describe the change you did to the code (not the bug). You change the PR title here a bunch of times; but that doesn't change the commit title. :)

You can squash your changes into one as well! It looks like you have three commits all part of the same change that can be one.

Thanks for fixes!

components/feature/contextmenu/src/main/java/mozilla/components/feature/contextmenu/ContextMenuCandidate.kt
mcarare: There was a discussion on https://github.com/mozilla-mobile/fenix/issues/580 if we should consider not showing titles that could hide the real src of the image. If you think it is not necessary, we can change that.

mcarare:
I did not add tests for the "http" issue. Please let me know if I should change the getLink code or add a test.

components/feature/contextmenu/src/main/java/mozilla/components/feature/contextmenu/ContextMenuCandidate.kt
mcarare: I did not add tests for the "http" issue. Please let me know if I should change the getLink code or add a test.

pocmo: Yeah, `title?.startsWith("http", true)` looks fishy. This is true for titles that are not URLs.

mcarare: The concern was to not allow images to hide real url by using a title.
https://www.kevinbrosnan.net/testcases/absurdly-long-title-text.html
If you test it on Chrome, you can see they they also exchange title for src when the title starts with "http"

jonalmeida: > The concern was to not allow images to hide real url by using a title.
> https://www.kevinbrosnan.net/testcases/absurdly-long-title-text.html
> If you test it on Chrome, you can see that they also exchange title for src when the title starts with "http"

Would `title.isUrl()` work in that case?

jonalmeida: For what it's worth, the test page spoofing doesn't work on r-b so I'm not sure if this is needed.

mcarare: I guess we could use another way to check if it is an url or it poses as one. I haven't checked the code on reference browser, but in fenix, without that check it shows the title with http.

mcarare:
@jonalmeida does this need additional work, or could it be landed? Thank you!

jonalmeida:
@mcarare Ah sorry, I think I missed the notification for this when the PR was updated. Could you fix the placement for the changelog entry? I think would be great then. 👌

jonalmeida:
Thank you!

The TC failure looks intermittent: 
```

* What went wrong:
Execution failed for task ':feature-tab-collections:jacocoDebug'.
> There was a failure while executing work items
   > A failure occurred while executing com.android.build.gradle.internal.tasks.JacocoTaskDelegate$JacocoWorkerAction
      > java.lang.ExceptionInInitializerError (no error message)

```

PR #4129, opened 2019-08-14
Improve/simplify sync and device APIs

### LINKED ISSUE #3284: Fenix syncs twice immediately after startup
The Fenix logs show that immediately after startup there are 2 syncs scheduled. They are correctly serialized, but as soon as the first completes the second starts. This is likely to make "warming up" of the browser slower than it needs to be.

One comes [from android components, here](https://searchfox.org/mozilla-mobile/rev/2316858e5e401a9c303e381e0c3b697afb1af28d/android-components/components/feature/sync/src/main/java/mozilla/components/feature/sync/BackgroundSyncManager.kt#209) and the other comes [from Fenix, here](https://searchfox.org/mozilla-mobile/rev/2316858e5e401a9c303e381e0c3b697afb1af28d/fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt#112).

The a-c one is responding to onAuthenticated, so I suspect it does a "sync now" for the initial sync as an account is connected, but because Fenix also schedules a "startup" sync we end up getting called twice when starting when already connected. So I'm not sure where the fix for this should be, but I opened the bug in this repo and I'm confident one of the fine people here will get the bug to its rightful home :)

cc @grigoryk

grigoryk:
This patch series aims to improve the API surface which exposes sync and device management.

## First patch:
Refresh device constellation and polls for events during account init (restore, login, signup, recovery, etc). This allows API consumers to worry somewhat less about having to do this manually.
Refresh is performed only for devices that have SEND_TAB capability enabled.

In the future we'll have "system events" that we'd like to poll as soon as possible, which this enables.

## Second patch:
Introduces a `debounce` parameter that can be passed into `syncNowAsync`. When `debounce=true`, sync request will be ignored if sync already ran recently. With this change in place, applications can simply invoke `syncNowAsync(debounce = true)` in `onResume`, without having to worry about running sync twice in a row in case of startup (sync is executed during account manager initialization).

This change lets us trivially fix #3284. 

## Third patch:
Simplifies internal implementation of `FxaDeviceConstellation`, removing `PollingDeviceManager` entirely. Public API is now both simpler and provides applications with more control: polling for device events is now separate from refreshing constellation state. Lots of new tests added.

---

### Pull Request checklist

### After merge


grigoryk:
Marking as WIP, still need to finish up new constellation tests.

grigoryk:
Reviewing by parts should be much easier.
@jonalmeida part 1 and part 3 should be close to your interests. Similar for @Amejia481 - part 2. But of course feel free to review all of it :)

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
```diff
@@ -753,6 +747,21 @@ open class FxaAccountManager(
...
 
+    private suspend fun postAuthenticated(newAccount: Boolean) {
+        // Before any sync workers have a chance to access it, make sure our SyncAuthInfo cache is hot.
+        maybeUpdateSyncAuthInfoCache()
+
+        // Notify our internal (sync) and external (app logic) observers.
+        notifyObservers { onAuthenticated(account, newAccount) }
+
+        // If device supports SEND_TAB...
+        if (deviceConfig.capabilities.contains(DeviceCapability.SEND_TAB)) {
```
jonalmeida: I'm wonder if there are event categories that we would want to consider when to refresh and poll for. I see why we want to treat SEND_TAB in a special way, but I guess I'm curious about what happens tomorrow when we those new events coming in? Would it make sense then to have a sort of whitelist of event types that we would want check for?

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/sync/WorkManagerSyncManager.kt
```diff
@@ -135,7 +135,7 @@ class WorkManagerSyncDispatcher(
         return isSyncActive
     }
 
-    override fun syncNow(startup: Boolean) {
+    override fun syncNow(startup: Boolean, debounce: Boolean) {
         logger.debug("Immediate sync requested, startup = $startup")
         val delayMs = if (startup) {
             // Startup delay is there to avoid SQLITE_BUSY crashes, since we currently do a poor job
```
jonalmeida: Unrelated nit: we can remove this FIXME.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
grigoryk: Possibly! I think for now, given how things are, polling based on capabilities makes sense. Our public API is that applications declare which capabilities they have, not which events they support (which, in retrospect, sounds like a pretty good way to model this as well!).
Once we start expanding "capabilities" list, we can also expand these checks.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/sync/WorkManagerSyncManager.kt
grigoryk: Heh, I hope so!

components/service/firefox-accounts/README.md
```diff
@@ -63,6 +63,10 @@ accountManager.register(accountObserver, owner = this, autoPause = true)
 accountManager.registerForSyncEvents(syncObserver, owner = this, autoPause = true)
 
 // Observe incoming device events (e.g. SEND_TAB events from other devices).
+// Note that since device is configured with a SEND_TAB capability, device constellation will be
```
Amejia481: nit: Note that since **the** device

components/service/firefox-accounts/src/test/java/mozilla/components/service/fxa/FxaAccountManagerTest.kt
```diff
@@ -398,13 +398,15 @@ class FxaAccountManagerTest {
...
 
         // Can trigger syncs.
         manager.syncNowAsync().await()
-        verify(latestSyncManager!!.dispatcher.inner, times(2)).syncNow(anyBoolean())
+        verify(latestSyncManager!!.dispatcher.inner, times(2)).syncNow(startup = false, debounce = false)
         manager.syncNowAsync(startup = true).await()
-        verify(latestSyncManager!!.dispatcher.inner, times(3)).syncNow(anyBoolean())
+        verify(latestSyncManager!!.dispatcher.inner, times(1)).syncNow(startup = true, debounce = false)
+        manager.syncNowAsync(startup = true, debounce = true).await()
+        verify(latestSyncManager!!.dispatcher.inner, times(1)).syncNow(startup = true, debounce = true)
```
Amejia481: @grigoryk Just to confirm, is this verifying #3284? Or do we need an extra test for it?

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
jonalmeida: If capabilities are similar to a super-set of event types, then that could work too!

components/service/firefox-accounts/src/test/java/mozilla/components/service/fxa/FxaAccountManagerTest.kt
grigoryk: We don't have tests covering dispatcher/worker behaviour right now, unfortunately. So this is _partially_ verifying that, but not quite :)

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
grigoryk: That's how I've been thinking about them _for now_, but I'm not sure it'll hold over time. We'll see, anyway.

components/service/firefox-accounts/src/test/java/mozilla/components/service/fxa/FxaAccountManagerTest.kt
Amejia481: 👍

grigoryk:
bors r=Amejia481,jonalmeida

PR #4138, opened 2019-08-16
fixes #4118 added String function for conversion to date for minDateString and maxDateString of unknown date formats

### LINKED ISSUE #4118: ParseException Unparseable date when clicking on datepicker widgets
Related Fenix issue https://github.com/mozilla-mobile/fenix/issues/4685

The problem is that we consider minDateString and maxDateString as having the same format as initialDateString, and that is not always the case.

Also, in cases when the minDateString and maxDateString are empty we set them as current date (keeping us from selecting other dates than the current date), which IMO is contrary to the purpose of not setting min or max (being able to select any date)

@Amejia481 how do you think we should to handle these issues?

mcarare:
minDateString and maxDateString are not always in same format as initialDate, so we need to convert them to date trying multiple formats, catching parseException and having a null value ass fallback. Other date formats can be supported by adding them to the array in the future.
Function is using original toDate() function so it keeps returning null for empty or null strings and also considers Locale.
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlin/String.kt
+++ b/components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlin/String.kt
@@ -6,6 +6,7 @@ package mozilla.components.support.ktx.kotlin
 
 import mozilla.components.support.utils.URLStringUtils
 import java.security.MessageDigest
+import java.text.ParseException
 import java.text.SimpleDateFormat
 import java.util.Date
 import java.util.Locale
@@ -59,4 +60,29 @@ fun String.sha1(): String {
     return digest.joinToString(separator = "", transform = { byte ->
         String(charArrayOf(characters[byte.toInt() shr 4 and 0x0f], characters[byte.toInt() and 0x0f]))
     })
+}
+
+/**
+ * Tries to convert a [String] to a [Date] using a list of [possibleFormats].
+ * @param possibleFormats one ore more possible format.
+ * @return a [Date] object with the values in the provided in this string,
+ * if the conversion is not possible null will be returned.
+ */
+fun String.toDate(
+    vararg possibleFormats: String = arrayOf(
+            "yyyy-MM-dd'T'HH:mm",
+            "yyyy-MM-dd",
+            "yyyy-'W'ww",
+            "yyyy-MM",
+            "HH:mm"
+    )
+): Date? {
+    possibleFormats.forEach {
+        try {
+            return this.toDate(it)
+        } catch (pe: ParseException) {
+            // move to next possible format
+        }
+    }
+    return null
 }
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/prompt/GeckoPromptDelegate.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/prompt/GeckoPromptDelegate.kt
@@ -380,8 +380,8 @@ internal class GeckoPromptDelegate(private val geckoEngineSession: GeckoEngineSe
         geckoCallback: TextCallback
     ) {
         val initialDate = initialDateString.toDate(format)
-        val minDate = if (minDateString.isNullOrEmpty()) null else minDateString.toDate(format)
-        val maxDate = if (maxDateString.isNullOrEmpty()) null else maxDateString.toDate(format)
+        val minDate = if (minDateString.isNullOrEmpty()) null else minDateString.toDate()
+        val maxDate = if (maxDateString.isNullOrEmpty()) null else maxDateString.toDate()
         val onSelect: (Date) -> Unit = {
             val stringDate = it.toString(format)
             geckoCallback.confirm(stringDate)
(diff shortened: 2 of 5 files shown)


components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlin/String.kt
```diff
@@ -67,3 +68,19 @@ fun String.sha1(): String {
         String(charArrayOf(characters[byte.toInt() shr 4 and 0x0f], characters[byte.toInt() and 0x0f]))
     })
 }
+
+/**
+ * Converts String to date, trying multiple formats
+ */
+fun String.getDateFromString(): Date? {
+
+    val possibleDateFormats = arrayOf("yyyy-MM-dd'T'HH:mm", "yyyy-MM-dd", "yyyy-'W'ww", "yyyy-MM", "HH:mm")
```
Amejia481: I'm not quite sure if trying to guess the format could be the right approach, because this could lead us to unexpected behaviors, in some cases, we may be doing the right thing and others not. As the date was not in the correct format since the beginning, we really don't know what was the initial intention of the developer we may pick a wrong format trying to recovering. 

This is hard question to answer, I checked for some references (they are different between each other):

**Android:**
Fennec and Chrome do not try to guess the format allowing the user to pick whatever date.
The site is unusable if you don't pick a date between today and a month after, the dates are never picked.

**Desktop:**
FireFox and Chrome they try to guess the format allowing the user to pick dates from `minDate` as today  until `maxDate` a month after today.

Amejia481: 🤔  I'm not sure which approach we follow here desktop or mobile(Android)

mcarare: IMO, on Android, Chrome and Fennec don't do a very good job, ignoring minDate and maxDate, letting user select invalid dates.

On desktop, making minDate today and maxDate random în the future ignores the website's developer constraints.

We can also limit the formats we can expect, by removing the ones that could generate errors from the list, but this way we can account for the values send by the website.

Amejia481: In the guessing we are also doing wrong for instance the `min` and `max` date they look like `yyyy-MM-dd` but the real developer intention was `yyyy-MM-dd'T'HH:mm`. On Fennec and chrome for Android try to select today but with a time before `16:16`(04:16PM)

"2019-08-16 16:16" -> ✅  The date is set
**Clear the input**
"2019-08-16 16:15" -> 🔴   The date is not set

mcarare: Yes, that is why I put the format with time first in the list. The main usage for the min limit is to prevent selecting a time before now() on searches for booking, routes, etc, cause that would render them obsolete.

Amejia481: Yes, but what we are getting from the website as `minDate` is `2019-08-16` and that will fall under `yyyy-MM-dd` allowing users to select any time before `16:16` that was the developer intention as we can see because they are clearing out the input.

This is why I'm not completely sure if guessing is the right approach because. In guessing, we are not completely sure if we are doing the right thing.

Before leaning to one approach or another I would like to hear the third opinion @pocmo what do you think?

Amejia481: @mcarare  I'm trying to find what FireFox desktop is doing because they are guessing the full date including the time. At least this way we are going to have an aligned behavior across platforms, otherwise we can use the approach that you suggested above 👍

Amejia481: I couldn't find what exactly desktop ([The farder that I got was](https://dxr.mozilla.org/mozilla-central/source/dom/html/HTMLInputElement.cpp#1721)) was doing to guess the date format. But I think your approach it is good alternative :) 👍 

One last, suggestion: I think `String.getDateFromString()` could be improved if we rename to
`String.toDate` that will allow the consumers to change the `possibleFormats`.

```kotlin
/**
 * Tries to convert a [String] to a [Date] using a list of [possibleFormats].
 * @param possibleFormats one ore more possible format.
 * @return a [Date] object with the values in the provided in this string,
 * if the conversion is not possible null will be returned.
 */
fun String.toDate(
    vararg possibleFormats: String = arrayOf(
        "yyyy-MM-dd'T'HH:mm",
        "yyyy-MM-dd",
        "yyyy-'W'ww",
        "yyyy-MM",
        "HH:mm"
    )
): Date? {
    possibleFormats.forEach {
        try {
            return this.toDate(it)
        } catch (pe: ParseException) {
            // move to next possible format
        }
    }
    return null
}
```

What do you think?

mcarare: I think this is a good approach! Will you make these changes, or should I?

Amejia481: I can do the change :) 
Thanks!

PR #4150, opened 2019-08-19

Pike:
State: mozilla-l10n/android-l10n@413e05dba427ef8cbf9e5272ae5cc3b3ccdf91d3

In addition to the regular updates, this also add new locales, in support of Fenix and Lockwise: Bulgarian, Danish, Hungarian, Norwegian (both variants), Tamil, Turkish, Vietnamese.

CC @Delphine

### DIFF
--- a/components/browser/engine-system/src/main/res/values-nb-rNO/strings.xml
+++ b/components/browser/engine-system/src/main/res/values-nb-rNO/strings.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <!-- Text for the title of an alert dialog displayed by a web page. %1$s will be replaced with the URL of the current page (displaying the dialog). -->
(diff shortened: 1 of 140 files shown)


pocmo:
We could either check if this is a typo and needs fixing or disable the lint check 🤔

Pike:
I can see that both ways. Over to @Delphine to reach out.

Delphine:
Reached out to Havar, waiting to hear back!

Delphine:
Localizers should have updated this in all instances on Pontoon (it is in fact a mispelling, although sometimes accepted, we should thus avoid using it). 
So we should be good to go!

pocmo:
Thank you! In this case let's keep the `Typos` check for now.

@Pike Can you craft a new import? :)

Pike:
> @Pike Can you craft a new import? :)

Just pushed an update.

PR #4152, opened 2019-08-19
Consume new `substitute-local-geckoview.gradle` script from Bug 1533465.  Fixes #4068.

### LINKED ISSUE #4068: Explicitly set GeckoView dependency to have classifier `@aar`
For technical reasons, Gradle module substitution can't change the "classifier type" of substituted modules.  Also for technical reasons, the "default" type of modules is `@jar`, but the Android-Gradle plugin has internal work-arounds to make it find AARs.  Together, this means that module substitution isn't straightforward: substituting `org.mozilla.geckoview:geckoview-nightly` for `org.mozilla.geckoview:geckoview-default` will find `@jar` type modules, not `@aar`.

It is possible to work around this: consumers can always find some way to make `browser-engine-gecko-*` _not_ include the GeckoView dependency, and then include it (or substitute it) themselves.  But it would be easier to do that -- for example, using the approaches of [Bug 1533465](https://bugzilla.mozilla.org/show_bug.cgi?id=1533465) -- if `browser-engine-gecko-*` set the `@aar` classifier itself.  Concretely, that means using:
```
implementation "org.mozilla.geckoview:geckoview-nightly:$VERSION@aar"
```

I am not aware of any repercussions from setting the classifier explicitly, but it's possible there are some that I'm not aware of.

ncalexan:
@pocmo this works for me locally.  The upstream script doesn't use `allprojects` (figured consumers can do that), but I'm not sure you really want to do this for all projects when it will impact ~50 projects.  Consider doing that or, or else plumb the single line into the relevant projecst (~7, it looked like to me).

### DIFF
--- a/samples/browser/build.gradle
+++ b/samples/browser/build.gradle
@@ -182,3 +182,11 @@ dependencies {
     androidTestImplementation Dependencies.androidx_espresso_core
     androidTestImplementation Dependencies.testing_mockwebserver
 }
+
+if (gradle.hasProperty('localProperties.dependencySubstitutions.geckoviewTopsrcdir')) {
+    if (gradle.hasProperty('localProperties.dependencySubstitutions.geckoviewTopobjdir')) {
+        ext.topobjdir = gradle."localProperties.dependencySubstitutions.geckoviewTopobjdir"
+    }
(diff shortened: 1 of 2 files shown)


pocmo:
(On previous run a taskcluster task was hanging indefinitely)

bors r+

PR #4162, opened 2019-08-20
Closes #3637: Add SearchLocalizationProvider implementation that uses MozillaLocationService

### LINKED ISSUE #3637: Implement SearchEngineProvider using MLS
Also see: #1720, https://github.com/mozilla-mobile/fenix/issues/3869

In Fennec we do a region lookup via MLS. Let's port this to AC so that we can use it in Fenix. Our updated `list.json` requires a region and using the locale for that is not that helpful.

pocmo:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/search/src/main/java/mozilla/components/browser/search/provider/localization/SearchLocalizationProvider.kt
+++ b/components/browser/search/src/main/java/mozilla/components/browser/search/provider/localization/SearchLocalizationProvider.kt
@@ -8,39 +8,6 @@ package mozilla.components.browser.search.provider.localization
  * Class providing language, country and optionally region (actual location) of the user/device to
  * customize the search experience.
  */
-abstract class SearchLocalizationProvider {
-    /**
-     * ISO 639 alpha-2 or alpha-3 language code, or registered language subtags up to 8 alpha letters
-     * (for future enhancements).
-     *
-     * Example: "en" (English), "ja" (Japanese), "kok" (Konkani)
-     */
-    abstract val language: String
-
-    /**
-     * ISO 3166 alpha-2 country code or UN M.49 numeric-3 area code.
-     *
-     * Example: "US" (United States), "FR" (France), "029" (Caribbean)
-     */
-    abstract val country: String
-
-    /**
-     * Optional actual location (region only) of the user/device. ISO 3166 alpha-2 country code or
-     * UN M.49 numeric-3 area code.
-     *
-     * This value is used to customize the experience for users that use a languageTag that doesn't match
-     * their geographic region. For example a user in Russia with "en-US" languageTag might get offered
-     * an English version of Yandex if the region is set to "RU".
-     *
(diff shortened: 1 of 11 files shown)


components/browser/search/src/test/java/mozilla/components/browser/search/provider/AssetsSearchEngineProviderTest.kt
```diff
@@ -20,10 +21,8 @@ class AssetsSearchEngineProviderTest {
 
     @Test
     fun `Load search engines for en-US from assets`() = runBlocking {
-        val localizationProvider = object : SearchLocalizationProvider() {
-            override val country: String = "US"
-            override val language = "en"
-            override val region: String? = null
+        val localizationProvider = object : SearchLocalizationProvider {
+            override suspend fun determineRegion() = SearchLocalization("en", "US", "US")
```
Amejia481: nit: As different provider repeat on the file, you may want to consider having them as properties like `usProvider` or `germanyProvider` :)

components/browser/search/src/main/java/mozilla/components/browser/search/provider/localization/SearchLocalizationProvider.kt
```diff
@@ -8,39 +8,6 @@ package mozilla.components.browser.search.provider.localization
  * Class providing language, country and optionally region (actual location) of the user/device to
```
Amejia481: nit: As we removed the class and added an interface the docs are now outdated. Is this breaking change? If so, is it worthy to be mentioned on the changelog?

pocmo: Yeah, makes sense. I'll update that.

components/browser/search/src/test/java/mozilla/components/browser/search/provider/AssetsSearchEngineProviderTest.kt
pocmo: Yeah, it's an option. I also like if test methods do not have overlapping dependencies and can be modified independently. Since this code existed before I'm keeping it for now. :)

pocmo:
bors r=Amejia481

components/browser/search/src/main/java/mozilla/components/browser/search/provider/localization/SearchLocalizationProvider.kt
Amejia481: 👍

components/browser/search/src/test/java/mozilla/components/browser/search/provider/AssetsSearchEngineProviderTest.kt
Amejia481: 👍

PR #4168, opened 2019-08-20
Make MenuButton class public

sblatz:
Making MenuButton public so it can be used in Fenix's buttons :)

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/MenuButton.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/MenuButton.kt
@@ -21,7 +21,7 @@ import mozilla.components.browser.toolbar.R
 import mozilla.components.browser.toolbar.facts.emitOpenMenuFact
 
 @Suppress("ViewConstructor") // This view is only instantiated in code
-internal class MenuButton(
+class MenuButton(
     context: Context,
     private val parent: View
 ) : FrameLayout(context) {


components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/MenuButton.kt
```diff
@@ -21,7 +21,7 @@ import mozilla.components.browser.toolbar.R
 import mozilla.components.browser.toolbar.facts.emitOpenMenuFact
 
 @Suppress("ViewConstructor") // This view is only instantiated in code
-internal class MenuButton(
+class MenuButton(
```
NotWoods: We should change the constructor if the class is public.

`parent` can probably be replaced with `view.getParent() as? View` in `determineOrientation`. If `getParent` isn't a view some default orientation could be used.

sblatz: @NotWoods to clarify: are you wanting `parent` to become `view`? We don't have a reference to view in here.

NotWoods: I meant the button itself as the `view`. 

Now that I think about it more, `parent` doesn't actually correspond to `View.getParent()` so that wouldn't work properly 😅 

Maybe just change parent to be a var so it doesn't get set in the constructor.

```kt
class MenuButton @JvmOverloads constructor(
    context: Context,
    attrs: AttributeSet? = null,
    defStyle: Int = 0
) : FrameLayout(context, attrs, defStyle) {

    var parentToolbar: View? = null
    ...

    init {
        ...
            menu?.show(
                anchor = this,
                orientation = if (parentToolbar != null) BrowserMenu.determineMenuOrientation(parentToolbar) else DOWN,
                endOfMenuAlwaysVisible = endAlwaysVisible
            ) { menu = null }
        ...
    }
    ...
}
```

PR #4223, opened 2019-08-27
Closes #3359: Document deprecation process.

### LINKED ISSUE #3359: Define deprecation process
Related: #3281

We do not have a process about how we deprecate code and when we remove it.

pocmo:
First iteration of document and defining our deprecation process.


pocmo:
> How this page is going to be discoverable? Should it be linked from another main page like contributing or home page?

Oops, yeah, it should!

pocmo:
bors r=Amejia481

PR #4229, opened 2019-08-27
WIP: 1541820: Implement a Gradle plugin for Glean.

mdboom:
This replaces the Gradle script that Glean currently uses to generate the API for metrics with a proper Gradle plugin.

There are some known issues remaining, for which I'm hoping to get some help from some Gradle experts. :)

1. This plugin needs to be used both by external tools (e.g. Fenix) and internally by other components within android-components.  The usual `project(...)` referencing doesn't seem to work for plugins, and there doesn't seem to be any pointers in the docs about making a plugin available in both contexts.  The workaround here is to include the plugin source as a symlink in `buildSrc` for internal use, but I suspect that's unlikely to work on Windows.

2. For bonus points, I'd like to remove the requirement that users of this plugin must also include the JetBrains Python plugin.  It's an indirect dependency of the Glean plugin, but shouldn't have to be a direct dependency of Glean-using projects.  But I can't figure out how to properly specify the dependency of my plugin on another plugin.

For reference, you can see the changes required to use a local build of this plugin with Fenix [here](https://github.com/mozilla-mobile/fenix/compare/master...mdboom:glean-gradle-plugin?expand=1).

---

### Pull Request checklist

### After merge


components/service/glean-gradle-plugin/build.gradle
```diff
@@ -0,0 +1,44 @@
...
+                    license {
+                        name = libLicense
+                        url = libLicenseUrl
+                    }
+                }
+
+                developers {
+                    developer {
+                        name = 'Mozilla Android Components Team'
+                        email = 'android-components@lists.mozilla.org'
```
Dexterp37: I think we should either add more than 1 `developer` section or just put the Glean team here.

components/service/glean-gradle-plugin/src/main/groovy/mozilla/components/service/glean-gradle-plugin/GleanGradlePlugin.groovy
```diff
@@ -0,0 +1,287 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import org.apache.tools.ant.taskdefs.condition.Os
```
Dexterp37: I wish we could get rid of Groovy and just use Kotlin, darn :|

components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -2,6 +2,19 @@
...
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+/**********************************************************************
+
+  THIS FILE IS OBSOLETE AND UNMAINTAINED.
+
+  The functionality in this file has moved to a proper Gradle plugin
+  in components/service/glean-gradle-plugin.
+
+  However, this file must remain on the master branch indefinitely so that
```
Dexterp37: Gah :(

components/service/glean-gradle-plugin/src/main/groovy/mozilla/components/service/glean-gradle-plugin/GleanGradlePlugin.groovy
```diff
@@ -0,0 +1,287 @@
...
+    List<File> transform(File file) {
+        def f = new File(file, "metrics.yaml")
+        if (f.exists()) {
+            return [f]
+        }
+        return []
+    }
+}
+
+class GleanPlugin implements Plugin<Project> {
```
Dexterp37: This really looks like an almost copy-paste of the current script. Is that right?

components/service/glean-gradle-plugin/build.gradle
Dexterp37: I see, good point. Let's not touch it then.

components/service/glean-gradle-plugin/src/main/groovy/mozilla/components/service/glean-gradle-plugin/GleanGradlePlugin.groovy
Dexterp37: Totally makes sense! Let's file a bug after this lands!

pocmo:
> @pocmo: And a question for you. It makes things convenient to keep versions in sync etc. to include this in android-components and use its publishing cycle for this plugin. Do you have any objections to this plugin living in android-components, even as glean in a-c itself is likely to become a wrapper around a library implemented elsewhere in the future?

* Wouldn't consumers of "pure glean without AC" need such a plugin too?
* Keeping the versions in sync makes sense to me.
* I have no objections, but we have no mechanism to publish a plugin yet. That's something we'd have to figure out first. This has come up for other things too. So this would be a good reason to finally look into how we could do that.

.buildconfig.yml
```diff
@@ -232,6 +232,10 @@ projects:
     path: components/service/glean
     description: 'A client-side telemetry SDK for collecting metrics and sending them to the Mozilla telemetry service'
     publish: true
+  service-glean-gradle-plugin:
```
pocmo: namespace nit: `service` in AC relates to a "web service" or anything talking to servers. Technically I think this would fit better into our `tooling` category that hosts things that are not really components themselves (e.g. `tooling-lint` for custom lint checks). However we could consider creating a `plugins` category because as said before there is some interest in providing other plugins!

components/service/glean-gradle-plugin/build.gradle
```diff
@@ -0,0 +1,44 @@
...
+                artifactId = archivesBaseName
+                description = project.ext.description
+                version = config.componentsVersion + (project.hasProperty('snapshot') ? '-SNAPSHOT' : '')
+
+                licenses {
+                    license {
+                        name = libLicense
+                        url = libLicenseUrl
+                    }
+                }
```
pocmo: I wonder if there's a way to merge this partially with the existing `publish.gradle`. 🤔

Dexterp37:
> 
> 
> > @pocmo: And a question for you. It makes things convenient to keep versions in sync etc. to include this in android-components and use its publishing cycle for this plugin. Do you have any objections to this plugin living in android-components, even as glean in a-c itself is likely to become a wrapper around a library implemented elsewhere in the future?
> 
>     * Wouldn't consumers of "pure glean without AC" need such a plugin too?

Yes, they would.

>     * Keeping the versions in sync makes sense to me.

@mdboom @pocmo if we were to add this to the mozilla/glean repo, would it be reasonable to have a constant in the `Dependencies.kt` file here to make sure things are in sync? Is there any specific scenario that would be tricky otherwise?

>     * I have no objections, but we have no mechanism to publish a plugin yet. That's something we'd have to figure out first. This has come up for other things too. So this would be a good reason to finally look into how we could do that.

@ncalexan might have insights for that, given that they publish the gradle-rust plugin

pocmo:
> Ah, this is where my Gradle naïveté probably shows: how is publishing a plugin different from any other library? I was able to publish it to my local maven repo and use it for Fenix from there without issue...

I believe you that the gradle part is working. It's our automation pipeline that I am not sure about. But we'll see.. :)

ncalexan:
> This replaces the Gradle script that Glean currently uses to generate the API for metrics with a proper Gradle plugin.
> 
> There are some known issues remaining, for which I'm hoping to get some help from some Gradle experts. :)
> 
>     1. This plugin needs to be used both by external tools (e.g. Fenix) and internally by other components within android-components.  The usual `project(...)` referencing doesn't seem to work for plugins, and there doesn't seem to be any pointers in the docs about making a plugin available in both contexts.  The workaround here is to include the plugin source as a symlink in `buildSrc` for internal use, but I suspect that's unlikely to work on Windows.

This is a known tricky thing.  `GradleKit` solves this, IIRC.  The simplest way to do this is to use Gradle composite builds to substitute the plugin into the consumer.  See https://github.com/mozilla/rust-android-gradle/blob/master/samples/app/settings.gradle and related things.  I should be able to get this working for/with you without difficulty.

>     2. For bonus points, I'd like to remove the requirement that users of this plugin must also include the JetBrains Python plugin.  It's an indirect dependency of the Glean plugin, but shouldn't have to be a direct dependency of Glean-using projects.  But I can't figure out how to 
properly specify the dependency of my plugin on another plugin.

Generally, you do this symbolically.  See [Google's protobuf Gradle plugin](https://github.com/google/protobuf-gradle-plugin/blob/74997dceac6eef0095826067a32f24f11b2c051d/src/main/groovy/com/google/protobuf/gradle/ProtobufPlugin.groovy#L86-L105) for a detailed example.  (I've learned a lot from that plugin!)

It can be difficult to avoid dependencies in the chain here, but my guess is you'll be able to do it with a little effort.

ncalexan:
> > > @pocmo: And a question for you. It makes things convenient to keep versions in sync etc. to include this in android-components and use its publishing cycle for this plugin. Do you have any objections to this plugin living in android-components, even as glean in a-c itself is likely to become a wrapper around a library implemented elsewhere in the future?
> > 
> > 
> > ```
> > * Wouldn't consumers of "pure glean without AC" need such a plugin too?
> > ```
> 
> Yes, they would.
> 
> > ```
> > * Keeping the versions in sync makes sense to me.
> > ```
> 
> @mdboom @pocmo if we were to add this to the mozilla/glean repo, would it be reasonable to have a constant in the `Dependencies.kt` file here to make sure things are in sync? Is there any specific scenario that would be tricky otherwise?
> 
> > ```
> > * I have no objections, but we have no mechanism to publish a plugin yet. That's something we'd have to figure out first. This has come up for other things too. So this would be a good reason to finally look into how we could do that.
> > ```
> 
> @ncalexan might have insights for that, given that they publish the gradle-rust plugin

Yes, we do, and it's an embarrassing manual process that I have no automation for.  It's not possible to have multiple people "own" a plugin on plugins.gradle.org, AFAIK.  (People ask for it periodically.)

I'd prefer to publish to `maven.mozilla.org` and teach folks to consume like a regular dependency; the plugin magic is 100% Not Worth It.

buildSrc/build.gradle.kts
```diff
@@ -4,13 +4,25 @@
...
 }
 
 repositories {
     jcenter()
 }
 
+gradlePlugin {
+    plugins {
+        create("gleanPlugins") {
+            id = "mozilla.components.service.glean-gradle-plugin"
```
ncalexan: Strongly consider decoupling the plugin ID from `android-components`.  You probably want this to have an independent name at _some_ time.

buildSrc/src/main/groovy/GleanGradlePlugin.groovy
```diff
@@ -0,0 +1 @@
+../../../../components/service/glean-gradle-plugin/src/main/groovy/mozilla/components/service/glean-gradle-plugin/GleanGradlePlugin.groovy
```
ncalexan: See my comment about `settings.gradle` and composite builds.

components/service/glean-gradle-plugin/src/main/groovy/mozilla/components/service/glean-gradle-plugin/GleanGradlePlugin.groovy
```diff
@@ -0,0 +1,287 @@
...
+    String GLEAN_PARSER_VERSION = "1.3.0"
+    // The version of Miniconda is explicitly specified.
+    // Miniconda3-4.5.12 is known to not work on Windows.
+    String MINICONDA_VERSION = "4.5.11"
+
+    /* This script runs a given Python module as a "main" module, like
+     * `python -m module`. However, it first checks that the installed
+     * package is at the desired version, and if not, upgrades it using `pip`.
+     */
+    String runPythonScript = """
```
ncalexan: As follow-up, push this into a Java resource so it's a file on disk.

components/service/glean-gradle-plugin/src/main/groovy/mozilla/components/service/glean-gradle-plugin/GleanGradlePlugin.groovy
```diff
@@ -0,0 +1,287 @@
...
+                // https://github.com/GoogleCloudPlatform/endpoints-framework-gradle-plugin/commit/2f2b91476fb1c6647791e2c6fe531a47615a1e85.
+                // The added directory doesn't appear in the paths listed by the
+                // `sourceSets` task, for reasons unknown.
+                variant.registerJavaGeneratingTask(generateKotlinAPI, new File(sourceOutputDir))
+            }
+        }
+    }
+
+    void apply(Project project) {
+        // This installs a Miniconda3 environment into the gradle user home directory
```
ncalexan: This feels like the wrong time to do this.  Plugin application is very early.  You want to do this as part of some task, surely -- otherwise I can't run `./gradlew app:tasks` without installing Python, etc.

components/service/glean-gradle-plugin/src/main/groovy/mozilla/components/service/glean-gradle-plugin/GleanGradlePlugin.groovy
```diff
@@ -0,0 +1,287 @@
...
+            "glean/bootstrap-${MINICONDA_VERSION}"
+        )
+        File condaDir = new File(
+            condaBootstrapDir,
+            "Miniconda3"
+        )
+
+        // Define the path to the Miniconda3 installation as a buildConfig variable so
+        // that tests that need to run JSON through our JSON schema can do so.
+        project.android {
```
ncalexan: This needs to ensure that your plugin depends on the Android plugin, which in turn can be tricky to arrange 'cuz the consuming project's Android plugin version can be different than a-c's.  I never found a great solution to this: see https://github.com/mozilla/application-services/pull/921#pullrequestreview-223342221 and the linked project.  This will definitely require more attention before landing, from me or from testers.

ncalexan: You may find that this Just Works with Groovy and not with Kotlin.  Groovy is dynamic/at run-time; Kotlin is not.  Try to convert to Kotlin and you'll see the issues; I suggest you address them up front to avoid painting yourself into a corner.  (I painted myself into a corner.)

components/service/glean/scripts/sdk_generator.gradle
ncalexan: Can we print a deprecation warning, like the Android plugin (and many parts of Gradle itself) do?

components/service/glean-gradle-plugin/src/main/groovy/mozilla/components/service/glean-gradle-plugin/GleanGradlePlugin.groovy
mdboom: That's a good point.  I'll see if I can move this inside of `generateMetricsAPI` (where it is definitely needed).

components/service/glean-gradle-plugin/src/main/groovy/mozilla/components/service/glean-gradle-plugin/GleanGradlePlugin.groovy
mdboom: It looks like this buildConfig variable is only needed by Glean itself, not by products using Glean, so it doesn't have to go in the plugin.  That neatly sidesteps this problem.

components/service/glean-gradle-plugin/src/main/groovy/mozilla/components/service/glean-gradle-plugin/GleanGradlePlugin.groovy
mdboom: It looks like this is actually happening at the right time.  This is just setting up tasks (the most important of which is actually defined by JetBrain's Python Gradle plugin) and installs it into the work tree with `project.preBuild.finalizedBy("build_envs")` below.  I think `preBuild` is the right thing to hang off of, since we do need to have the generated code ready *before* build proper.

I tested by removing the Python environment (in `~/.gradle/glean`) and simple Gradle tasks like `dependencies` work without installing the whole Python environment.

So I'd say we just need to clarify this comment...

components/service/glean-gradle-plugin/src/main/groovy/mozilla/components/service/glean-gradle-plugin/GleanGradlePlugin.groovy
```diff
@@ -0,0 +1,287 @@
...
+        project.envs {
+            bootstrapDirectory = condaBootstrapDir
+            pipInstallOptions = "--trusted-host pypi.python.org --no-cache-dir"
+
+            // Setup a miniconda environment. conda is used because it works
+            // non-interactively on Windows, unlike the other options
+            conda "Miniconda3", "Miniconda3-${MINICONDA_VERSION}", "64", ["glean_parser==${GLEAN_PARSER_VERSION}"]
+        }
+
+        project.tasks.whenTaskAdded { task ->
```
ncalexan: For later, this whole task-before-task would be easier if you added a `doFirst` to the actual `Bootstrap_CONDA` task.  But this will work too.

components/service/glean-gradle-plugin/src/main/groovy/mozilla/components/service/glean-gradle-plugin/GleanGradlePlugin.groovy
ncalexan: OK, I see that the paths can be computed without actually doing any work.  Roll on.

components/service/glean-gradle-plugin/src/main/groovy/mozilla/components/service/glean-gradle-plugin/GleanGradlePlugin.groovy
mdboom: We had tried that in an earlier revision (of the script this plugin is based on).  The problem is that the `Bootstrap_CONDA` task isn't created until later, so we can't just call `doFirst` on it at this point.

buildSrc/build.gradle.kts
mdboom: We did some brainstorming on this today. I think we want to keep this in the android-components namespace because it's the version of the plugin that will evolve in sync with the a-c version of Glean. We will soon have the non-a-c implementation with its own plugin, which may need to evolve in a different direction. If the plugins had the same namespace, I think we're more likely to have a version mismatch between the plugin and the library. (All of this overlap should be very short-lived so I'm more concerned with the possibility of version mismatch between the library and the plugin than I am with the maintenance burden of a fork).

tl;dr: I think it's a feature, not a bug, that this is in the components namespace since it is so closely coupled with the library it supports.

components/service/glean-gradle-plugin/build.gradle
mdboom: Looking into this closer, I think it's fine to update this to say `glean-team@mozilla.com`.

components/service/glean/scripts/sdk_generator.gradle
mdboom: Good idea.

mdboom:
@ncalexan wrote:
> I'd prefer to publish to maven.mozilla.org and teach folks to consume like a regular dependency; the plugin magic is 100% Not Worth It.

I'm on board with this.  I think it just means the syntax for using the plugin is a bit more verbose, right?  Our users will probably just copy and paste something we provide anyway :)  I think the benefits of publishing everything in a unified way outweigh that downside.

mdboom:
@ncalexan wrote:
> I think your Python setup isn't quite ready for production, but maybe I'm not understanding what actually happens when.

If it's not, I'd like to make it so ;)

Basically, we use a third-party plugin [gradle-python-envs](https://github.com/JetBrains/gradle-python-envs) to install a Python environment -- Miniconda3 specifically.  It puts it in `~/.gradle/glean` so that it is shared between every library/component/application that uses Glean.  This task is assigned to `preBuild.finalizedBy`.

The code generation itself happens in a task that is attached to each variant using `variant.registerJavaGeneratingTask`, which IIUC is designed for the exact purpose of inserting a code generation step.

That task calls out to Python and our glean_parser library to do the code generation, dropping the `.kt` files in `build/generated`.  One wrinkle there is that the version of `glean_parser` is checked, and if it's not the one we need for this version of android-components, we update it over the network at that moment, and then fork another process so we can reload the new install to actually do the code generation.  This allows for revision hopping (e.g. git bisecting) to work correctly.

We'd considered using some other technology to do this, but glean_parser is also used by the pipeline to do various things with the metrics information, and that infrastructure is all Python.

mdboom:
A summary of the remaining issues after the first round of review and reworking:

1. I think consensus is that we publish this as a regular maven package, and not as a gradle plugin.  My understanding is that the downside is that it's a few more lines to include the plugin in third party code.  The upshot is that it publishes in the same way as all other android-components, including if publishing to a local maven repo, which will be helpful for development of Glean. @ncalexan: Can you confirm my understanding is correct?

2. There's a little code duplication in the publishing configuration between `components/service/glean-gradle-plugin/build.gradle` and the publication config used for all other android-components in `publish.gradle`.  I couldn't figure out a way around that -- if there's a simple way, let me know, otherwise I don't consider it a huge problem.

3. This plugin needs to work both externally and internally within android-components. @ncalexan wrote:

> This is a known tricky thing. GradleKit solves this, IIRC. The simplest way to do this is to use Gradle composite builds to substitute the plugin into the consumer. See https://github.com/mozilla/rust-android-gradle/blob/master/samples/app/settings.gradle and related things. I should be able to get this working for/with you without difficulty.

I couldn't get this to work, but I think I'm just missing some key knowledge about how Gradle works.

4. I tried to make this plugin *depend on* the Python plugin.  I tried the approach @ncalexan linked to from the Google protobuf extension, but that seems to only ensure that a plugin is applied *after* another one.  It doesn't actually cause the application of that other plugin.  So it doesn't resolve the issue of requiring that all users load this plugin *and* the python plugin directly as well.  However, since all that does is reduce the copy-and-paste for our users, I'm happy to defer solving that for now.

Thanks for everyone's patience while I learn Gradle! :)

components/tooling/glean-gradle-plugin/build.gradle
```diff
@@ -0,0 +1,60 @@
+plugins {
```
pocmo: nit: license header

ncalexan:
> A summary of the remaining issues after the first round of review and reworking:
> 
>     1. I think consensus is that we publish this as a regular maven package, and not as a gradle plugin.  My understanding is that the downside is that it's a few more lines to include the plugin in third party code.  The upshot is that it publishes in the same way as all other android-components, including if publishing to a local maven repo, which will be helpful for development of Glean. @ncalexan: Can you confirm my understanding is correct?

Yes, as far as I know it is possible to apply plugins as "regular" Maven packages.  I have certainly done it to override bits and pieces, for example.  What does not work (or at least, not work smoothly out of the box) is the new `apply plugin: ...` syntax.  That uses some Maven artifact rewriting and some assumptions deep in Gradle, last I checked.

>     2. There's a little code duplication in the publishing configuration between `components/service/glean-gradle-plugin/build.gradle` and the publication config used for all other android-components in `publish.gradle`.  I couldn't figure out a way around that -- if there's a simple way, let me know, otherwise I don't consider it a huge problem.
> 
>     3. This plugin needs to work both externally and internally within android-components. @ncalexan wrote:
> 
> 
> > This is a known tricky thing. GradleKit solves this, IIRC. The simplest way to do this is to use Gradle composite builds to substitute the plugin into the consumer. See https://github.com/mozilla/rust-android-gradle/blob/master/samples/app/settings.gradle and related things. I should be able to get this working for/with you without difficulty.
> 
> I couldn't get this to work, but I think I'm just missing some key knowledge about how Gradle works.

OK.  I will look at this in a moment.

>     1. I tried to make this plugin _depend on_ the Python plugin.  I tried the approach @ncalexan linked to from the Google protobuf extension, but that seems to only ensure that a plugin is applied _after_ another one.  It doesn't actually cause the application of that other plugin.  So it doesn't resolve the issue of requiring that all users load this plugin _and_ the python plugin directly as well.  However, since all that does is reduce the copy-and-paste for our users, I'm happy to defer solving that for now.

Hmm, I see.  In general this is [possible](https://stackoverflow.com/a/45652773) but I have found it to be very tricky.  The issue is that the easiest way to do this has the Glean plugin pin a specific version of the Python plugin _at Glean plugin build time_.  (With effort it's possible to work-around this: reflection, etc.)  But the actual Python plugin version will be determined by Gradle at _Glean plugin application time_, like any other dependency (Gradle and Maven are all about late binding).

So in practice it's a headache.  Now, with something like the Android-Gradle plugin, version clashes are a nightmare.  But depending on what you want to do with the Python plugin this might not be a problem at all.  I will take a quick look and see if I understand enough to have an opinion.  If not we can, of course, defer.
 
> Thanks for everyone's patience while I learn Gradle! :)

components/service/glean/build.gradle
```diff
@@ -72,3 +74,17 @@ apply from: '../../../publish.gradle'
 ext.configurePublish(config.componentsGroupId, archivesBaseName, project.ext.description)
 
 apply plugin: 'mozilla.components.tooling.glean-gradle-plugin'
+
+pluginManager.withPlugin('mozilla.components.tools.glean-gradle-plugin') {
```
ncalexan: This will always be available 'cuz you apply the plugin yourself, I see.  Was going to ask what happened if the plugin wasn't around.   I expect a build failure; you agree?

On an unrelated note: baking paths like this into the _source_ that will get shipped in Fenix makes reproducible builds impossible.  Why on earth do you need this _at runtime_?

mdboom: Ok.  I'll look into one of these other options.  FWIW, master currently has this issue as well.

components/tooling/glean-gradle-plugin/src/main/groovy/mozilla/components/tooling/glean-gradle-plugin/GleanGradlePlugin.groovy
```diff
@@ -175,6 +181,63 @@ subprocess.check_call([
         }
     }
 
+    private Task generateGleanMetricsDocs
+
+    void generateMarkdownDocs(Project project, File condaDir) {
```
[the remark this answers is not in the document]
mdboom: The Glean sources are generated into the build directory (since they aren't intended to be included in source control), and IIUC, there is a separate build subdirectory for each variant, so we're sort of forced into this.

The markdown docs are in the source tree and intended to be checked into source control.

So, feels to me like there's just an inherent inconsistency we have to live with, but maybe I'm not understanding...

components/tooling/glean-gradle-plugin/src/main/groovy/mozilla/components/tooling/glean-gradle-plugin/GleanGradlePlugin.groovy
```diff
@@ -0,0 +1,363 @@
...
+        File condaBootstrapDir = new File(
+            project.getGradle().gradleUserHomeDir,
+            "glean/bootstrap-${MINICONDA_VERSION}"
+        )
+        File condaDir = new File(
+            condaBootstrapDir,
+            "Miniconda3"
+        )
+
+        // Define the path to the Miniconda3 installation as a buildConfig variable so
```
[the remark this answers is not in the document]
mdboom: Thanks for catching.  Just an oversight.

components/tooling/glean-gradle-plugin/src/main/groovy/mozilla/components/tooling/glean-gradle-plugin/GleanGradlePlugin.groovy
```diff
@@ -0,0 +1,363 @@
...
+                !project.file(project.buildDir).exists()
+            }
+            doLast {
+                project.logger.lifecycle("Creating build directory:" + project.buildDir.getPath())
+                project.buildDir.mkdir()
+            }
+        }
+
+        // Configure the Python environments
+        project.envs {
```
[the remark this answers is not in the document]
mdboom: Yes, I believe so.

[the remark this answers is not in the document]
mdboom: Yes, I believe so.

mdboom:
@ncalexan wrote:
> I think you need to add metrics.yaml and events.yaml as optional inputs, 

Good point.  Indeed, this doesn't retrigger an update of the generated code.

> In particular, your solution allows apply plugin: to work locally in a-c, which is very valuable.

Yes, but because it relies on a symlink, I'm concerned about it not working on Windows. @Dexterp37 : could I ask you kindly to test this on Windows?  Start with a fully `git clean`'d repo, and delete your `GRADLE_HOME` directory, and build `samples-glean` and see if it works.

> Now, with something like the Android-Gradle plugin, version clashes are a nightmare. But depending on what you want to do with the Python plugin this might not be a problem at all. 

It's unlikely we'll need to update the Python plugin at all (except if there were a critical security reason, I guess).  It *works*, and we can update the actual Python version we get from it by changing a config variable, without needing to update the plugin.  Does that help?  Like I said, though, this automatic dependency on the Python plugin is just a "nice to have", not a necessity if it proves to be too hard.

@Dexterp37 wrote:
> Please, before merging this, make sure (one last time!) that this is in sync with the latest version of sdk_generator.gradle. 

Yep.  I've been tracking that.  Just updated it again.

> Do you have a tentative PR for using this in the mozilla/glean repo?

No -- there's so many moving parts, I thought it best to get this merged and then tackle that.

components/tooling/glean-gradle-plugin/src/main/groovy/mozilla/components/tooling/glean-gradle-plugin/GleanGradlePlugin.groovy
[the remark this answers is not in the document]
ncalexan: It is possible to have multiple variants consume the same generated sources -- it's the equivalent of `src/main` vs `src/{debug,release}`, say.  But it's not the path of least resistance.

Since this really is just a nit, let it ride.

Dexterp37:
> > In particular, your solution allows apply plugin: to work locally in a-c, which is very valuable.
> 
> Yes, but because it relies on a symlink, I'm concerned about it not working on Windows. @Dexterp37 : could I ask you kindly to test this on Windows? Start with a fully `git clean`'d repo, and delete your `GRADLE_HOME` directory, and build `samples-glean` and see if it works.

I just tried that, it works. It generated the kotlin source for all the libraries. I have no build error.

mdboom:
We've decided to implement this in the new cross-platform glean-core implementation and not include it as part of Android Components.  See https://github.com/mozilla/glean/pull/421

PR #4242, opened 2019-08-28
Allow generating docs for Glean metrics

Dexterp37:
The docs are written, at build-time, to '$project/docs/metrics.md'. This additionally
enables automatic doc generation for the following components: lib-crash, storage-sync.

Bonus: add a note to engine-gecko-nightly's readme to explicitly require a data-review for products using it.

@pocmo - r? for the changes to lib-crash
@linacambridge - r? for the changes to browser-sync 

**Note**: whenever new metrics are added, the build system will automatically update the docs so that the dev can simply commit the changes.

---

### Pull Request checklist

### After merge


components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -273,6 +275,56 @@ ext.gleanGenerateMetricsAPI = {
...
+    }
+
+    // Attach the docs generation task to the code generation task, to make sure
+    // we generate docs, if we're asked for it.
+    tasks.whenTaskAdded { task ->
+        if (task.name.startsWith(gleanTaskNamePrefix)) {
+            task.dependsOn(gleanGenerateMetricsDocs)
+        }
+    }
+}
```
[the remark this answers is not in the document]
Dexterp37: This is a great point. I had the same idea and attempted to implement that first. Well, I failed :( The only workaround I found was to have `preBuild.finalizedBy(gleanGenerateMetricsDocs)`. However, I'm not sure `preBuild` is late enough. Unfortunately, I could not find a "postBuild". Thoughts?

Dexterp37:
> 
> 
> I guess we aren't documenting the Glean internal metrics with this for reasons?

That's a good point. I thought about doing that, but then I realized that we'd have duplicated docs (see [here](https://mozilla.github.io/glean/book/user/pings/index.html)). Maybe this is not a concern and would be better to have that autogenerated as well? Maybe I can file a follow-up and do that once we move to glean-core.

components/service/glean/scripts/sdk_generator.gradle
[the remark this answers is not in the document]
Dexterp37: > 
> 
> `preBuild` might not be late enough because the Python environment may not yet exist? I guess you could always test that theory by deleting `~/.gradle/glean` and then rebuilding a clean checkout.

I just confirmed. Unfortunately `preBuild` is too early :'(

> Beyond that, I don't know if I have any great ideas. We could try something like just attaching it to the first variant rather than `android.applicationVariants.all`...

Unfortunately that wouldn't work as well :( The *first* variant isn't necessarily the currently active variant in the IDE :(

> But like I said, if finding a solution here is too cumbersome, I'm 💯 on just leaving this as-is.

I think I will do this :( sorry :(

Dexterp37:
> 
> 
> I like the idea of a follow-up for core glean metrics. Will probably require a few small tweaks to the glean_parser to handle the special case of `ping_info` and `client_info` things as well. But no urgency in that they are already manually documented quite well and they arent changing much these days.

Filed bug https://bugzilla.mozilla.org/show_bug.cgi?id=1577182

Dexterp37:
@pocmo , tricky question: we really need to enable this in engine-gecko-* as well. However, the metrics.yaml might get updated on the GV side so, each time the GV dependency is updated, that PR will/should automatically update the docs too. Is that fine for you?

Dexterp37:
> 
> 
> > @pocmo , tricky question: we really need to enable this in engine-gecko-* as well. However, the metrics.yaml might get updated on the GV side so, each time the GV dependency is updated, that PR will/should automatically update the docs too. Is that fine for you?
> 
> Would it make more sense for the `metrics.md` to live in m-c next to the `metrics.yaml` in gecko instead? I don't fully understand the legal requirement of where these files need to live, so just throwing that out there.

It would make sense, in general, since the file lives there. However, we're not yet vendoring the glean_parser in m-c (that might have its own problems, I'm not even sure we have python 3.7 on m-c :D ). Moreover, products using Glean and including engine-gecko-* will transmit the metrics so, at the very least, the component README should link or advertise the metrics it collects. GV itself does not send any metric from the metrics.yaml file.

pocmo:
> @pocmo , tricky question: we really need to enable this in engine-gecko-* as well. However, the metrics.yaml might get updated on the GV side so, each time the GV dependency is updated, that PR will/should automatically update the docs too. Is that fine for you?

That is doable, I think.

A build of the component is enough to generate the new file?

The task that updates GeckoView dependencies runs the following script:
https://github.com/mozilla-mobile/android-components/blob/master/automation/taskcluster/action/update_geckoview.sh

I creates the commit before building/testing in order to see if there are actually changes. You could modify this script to add the doc file and amend the commit before pushing to a branch.

Dexterp37:
> 
> 
> > @pocmo , tricky question: we really need to enable this in engine-gecko-* as well. However, the metrics.yaml might get updated on the GV side so, each time the GV dependency is updated, that PR will/should automatically update the docs too. Is that fine for you?
> 
> That is doable, I think.
> 
> A build of the component is enough to generate the new file?

Yes, it is!

> The task that updates GeckoView dependencies runs the following script:
> https://github.com/mozilla-mobile/android-components/blob/master/automation/taskcluster/action/update_geckoview.sh
> 
> I creates the commit before building/testing in order to see if there are actually changes. You could modify this script to add the doc file and amend the commit before pushing to a branch.

Thanks! I'll take that as a separate, focused PR!

components/browser/storage-sync/README.md
```diff
@@ -2,6 +2,11 @@
 
 A syncable implementation of `concept-storage` backed by [application-services' Places lib](https://github.com/mozilla/application-services).
 
+## Before using this component
+Products sending telemetry and using this component *must request* a data-review following [this process](https://wiki.mozilla.org/Firefox/Data_Collection).
```
linacambridge: I wonder if it's worth making this configurable—say, with a Boolean that we check in `RustPlacesConnection::send{History, Bookmarks}Ping`—so that applications that don't need telemetry won't send it. (We'll need to revisit this when we start sending telemetry for storage, not just Sync). I think requiring data review to use this component with Glean is fine now, though!

components/browser/storage-sync/docs/metrics.md
```diff
@@ -0,0 +1,48 @@
...
+
+
+## bookmarks_sync
+A ping sent for every bookmarks sync. It doesn't include the `client_id` because it reports a hashed version of the user's Firefox Account ID.
+
+The following metrics are added to the ping:
+
+| Name | Type | Description | Data reviews | Extras | Expiration |
+| --- | --- | --- | --- | --- | --- |
+| bookmarks_sync.failure_reason |[labeled_string](https://mozilla.github.io/glean/book/user/metrics/labeled_strings.html) |Records bookmark sync failure reasons.  |[1](https://github.com/mozilla-mobile/android-components/pull/3092)||never |
```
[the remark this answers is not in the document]
Dexterp37: Great catch! This is indeed a bug, we should document labels. Since this is done by the glean_parser fixing this requires to release a new version. Once we do so, changes will be automatically picked up by all the repos. I've filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1578345) for fixing labels.

components/browser/storage-sync/README.md
Dexterp37: In general, if this component is used in an application that doesn't have Glean enabled, nothing will get collected (so no need for special casing or other options). However, if the product uses Glean then this component will automatically contribute with its own metrics. We might consider adding more flexibility here if we stumble upon the case of a component supporting Glean, being embedded in a product using Glean, with the product sending telemetry but not wanting the component to send it.

components/lib/crash/docs/metrics.md
```diff
@@ -0,0 +1,25 @@
+<!-- AUTOGENERATED BY glean_parser.  DO NOT EDIT. -->
+
+# Metrics
+This document enumerates the metrics collected by this project.
```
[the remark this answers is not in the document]
Dexterp37: Ouch, right :) Let me file a bug about that too. [Filed this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1578347).

components/lib/crash/README.md
```diff
@@ -92,7 +92,8 @@ CrashReporter(
 ### Sending crash reports to Glean
 
 [Glean](https://docs.telemetry.mozilla.org/concepts/glean/glean.html) is a new way to collect telemetry by Mozilla. 
-This will record crash counts as a labelled counter with each label corresponding to a specific type of crash (`native_code_crash`, and `unhandled_exception`, currently). The list of collected metrics is available in the [metrics.yaml file](metrics.yaml). 
+This will record crash counts as a labelled counter with each label corresponding to a specific type of crash (`native_code_crash`, and `unhandled_exception`, currently).
```
mdboom:
```suggestion
This will record crash counts as a labeled counter with each label corresponding to a specific type of crash (`native_code_crash`, and `unhandled_exception`, currently).
```

(Unfortunate) U.S. English

components/service/glean/scripts/sdk_generator.gradle
```diff
@@ -273,6 +275,54 @@ ext.gleanGenerateMetricsAPI = {
...
+
+        outputs.dir "$projectDir/docs"
+        workingDir rootDir
+
+        // Note that the command line is OS dependant: on linux/mac is Miniconda3/bin/python.
+        if (Os.isFamily(Os.FAMILY_WINDOWS)) {
+            commandLine new File(GLEAN_MINICONDA_DIR, "python")
+        } else {
+            commandLine new File(GLEAN_MINICONDA_DIR, "bin/python")
+        }
```
mdboom: (Maybe) pull this out into a function to reduce code duplication with the Kotlin generator...

components/lib/crash/README.md
Dexterp37: Ouch. Thanks for spotting that!

mdboom: No problem -- it trips me up even as a native English speaker.

components/service/glean/scripts/sdk_generator.gradle
Dexterp37: Done!

Dexterp37:
bors retry

PR #4258, opened 2019-08-29
`Store` extensions to observe `State` using Kotlin's `Flow` API

### LINKED ISSUE #3786: browser-state: Explore better ways to express what onlyIfChanged does
In `lib-state` we currently have an `onlyIfChanged()` helper. It does its job but we were never satisfied how complex it looked like to use. We wanted to look at other options.

pocmo:
For #3530. Fixes #3786.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlinx/coroutines/flow/Flow.kt
+++ b/components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlinx/coroutines/flow/Flow.kt
@@ -1,0 +1,31 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.support.ktx.kotlinx.coroutines.flow
+
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.filter
+
+/**
+ * Returns a [Flow] containing only values of the original [Flow] that have changed compared to
+ * the value emitted before them.
+ *
+ * Example:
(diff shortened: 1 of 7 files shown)


components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlinx/coroutines/flow/Flow.kt
```diff
@@ -0,0 +1,34 @@
...
+ * Original Flow: A, B, B, C, A, A, A, D, A
+ * Returned Flow: A, B, C, A, D, A
+ */
+fun <T> Flow<T>.ifChanged(): Flow<T> {
+    val lock = Any()
+    var observedValueOnce = false
+    var lastValue: T? = null
+
+    return filter { value ->
+        synchronized(lock) {
```
csadilek: I think I need a comment to understand why we need `synchronized` here :). In the docs it says "By default, flows are _sequential_ and all flow operations are executed sequentially in the same coroutine, with an exception for a few operations specifically designed to introduce concurrency into flow execution such as [buffer] and [flatMapMerge]. See their documentation for details.".

pocmo: My thinking here was that `observedValueOnce` and `lastValue` are getting created on the calling thread and then read inside the lambda running on some coroutine and that  this could cause a visibility issue. But I think you are right, since they are only read and written inside the coroutine lambda and if this happens in the same coroutine then I think we could strip that. 🤔

pocmo: Tests still pass without it. Updated the PR and removed the lock.

pocmo:
bors r=csadilek

PR #4259, opened 2019-08-29
Closes #3544: Add custom tab config to browser-state

### LINKED ISSUE #3544: Add missing properties to CustomTabSessionState
Right now `CustomTabSessionState` is just a clone of `TabSessionState`. Let's add the missing properties (see `CustomTabConfig` in `browser-session`).

csadilek:
Instead of copying/duplicating `CustomTabConfig`, I decided to move the existing one to browser-state. This works as browser-session already depends on browser-state. `CustomTabConfig` is not used by applications directly, so this doesn't break API either.

We shouldn't need a separate action to update this config, as the config is provided when the custom tab is created and then doesn't change.

I've also added some missing tests.

### DIFF
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/state/CustomTabConfig.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/state/CustomTabConfig.kt
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-package mozilla.components.browser.session.tab
+package mozilla.components.browser.state.state
 
 import android.app.PendingIntent
 import android.graphics.Bitmap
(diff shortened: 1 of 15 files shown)


components/browser/state/src/main/java/mozilla/components/browser/state/state/CustomTabConfig.kt
```diff
@@ -26,8 +28,8 @@ import androidx.browser.customtabs.CustomTabsSessionToken
  * @property sessionToken The token associated with the custom tab.
  */
 data class CustomTabConfig(
-    val id: String,
```
pocmo: nit: Not 100% sold on the name since it differs from our other `*State` classes here. But that's something quick to change - especially when this is not used by apps as you said.

components/browser/state/src/main/java/mozilla/components/browser/state/state/CustomTabSessionState.kt
```diff
@@ -12,16 +12,23 @@ import java.util.UUID
...
-    override val trackingProtection: TrackingProtectionState = TrackingProtectionState()
+    override val trackingProtection: TrackingProtectionState = TrackingProtectionState(),
+    val config: CustomTabConfig
 ) : SessionState
 
-internal fun createCustomTab(url: String, id: String = UUID.randomUUID().toString()): CustomTabSessionState {
+internal fun createCustomTab(
+    url: String,
+    id: String = UUID.randomUUID().toString(),
+    config: CustomTabConfig = CustomTabConfig()
```
pocmo: Eventually we may want to enforce providing a config (I'm going to make that helper visible in a different PR) but for now that's good enough (and probably means we do not have to change all the tests!).

Issue #3027, opened 2019-05-16
Improve Suggestions of HistoryStorageSuggestionProvider


ekager:
Awesome Bar Suggestions still contain many redirects/useless URLs and we should introduce some filtering/excluding here

See: 
https://github.com/mozilla-mobile/fenix/issues/2542

## Expected behavior

I should see history entries for https://www.mozilla.org/en-US/firefox and
https://www.mozilla.org/en-US/firefox/new

## Actual behavior

I get four history entries:
https://www.mozilla.org/firefox
https://www.mozilla.org/en-US/firefox
https://www.mozilla.org/firefox/new
https://www.mozilla.org/en-US/firefox/new

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-2448)

bifleming:
@ekager Janet is requesting a little more info on this bug...specifically with regard to 'useless URLs'.

ekager:
Please direct Janet to the linked Fenix issue above for more details

sblatz:
Fenix issue: https://github.com/mozilla-mobile/fenix/issues/2542

thomcc:
I'm a bit surprised these are showing up, and will have to do some digging. Over time does this issue remain? I'd expect hte redirects to have a very low frecency score once you have a decent amount of history...

vesta0:
Vesta will discuss this with A-S and UX

jdragojevic:
[a-s PR 1715](https://github.com/mozilla/application-services/pull/1715)  Exclude hidden URLs from syncing and views and
[a-c PR 4268](https://github.com/mozilla-mobile/android-components/pull/4268) Pass redirect source and target flags to history tracking delegates should resolve this.

cc @linacambridge @bifleming @vesta0

linacambridge:
To close the loop, #4268 no longer syncs or shows redirect sources in the history list view (#1331, even if they aren't consecutive), but they'll still be shown in the awesomebar (https://github.com/mozilla-mobile/android-components/pull/4268#issuecomment-531947764).

csadilek:
Tested this now and this has since been fixed. Closing.

PR #4268, opened 2019-08-30
Pass redirect source and target flags to history tracking delegates

linacambridge:
There's some confusion in `GeckoEngineSession` about redirect flags.
The `VISIT_REDIRECT_SOURCE` and `VISIT_REDIRECT_SOURCE_PERMANENT` flags
that we get from GeckoView's history delegate are for the redirect
_source_, not the visit type. They indicate if the URL passed to
`onVisited` is redirecting _to_ another URL, most likely because the
server returned an HTTP 3xy status code with a `Location` header.
Rust Places decides whether to mark the URL as hidden based on
these flags.

`VISIT_REDIRECT_{PERMANENT, TEMPORARY}`, however, indicate if the
URL passed to `onVisited` is the _target_ of a redirect (in other
words, the page that's _in_ the `Location` header). These get
translated into `VisitType` flags, which Rust Places stores as the
visit transition type. These two flags don't affect whether a URL
is hidden.

Note that, in a redirect chain, the middle links are both sources and
targets. For example, in "mozilla.org" -> "www.mozilla.org" ->
"www.mozilla.org/en-US", "www.mozilla.org" is both a redirect target
(since "mozilla.org" redirected to it), and a source (it redirected
to "www.mozilla.org/en-US").

See mozilla-mobile/fenix#3526.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -25,6 +25,8 @@ import mozilla.components.concept.engine.history.HistoryTrackingDelegate
 import mozilla.components.concept.engine.manifest.WebAppManifestParser
 import mozilla.components.concept.engine.request.RequestInterceptor
 import mozilla.components.concept.engine.request.RequestInterceptor.InterceptionResponse
+import mozilla.components.concept.storage.PageVisit
+import mozilla.components.concept.storage.RedirectSource
 import mozilla.components.concept.storage.VisitType
 import mozilla.components.support.ktx.android.util.Base64
 import mozilla.components.support.ktx.kotlin.isEmail
@@ -487,13 +489,30 @@ class GeckoEngineSession(
             val visitType = if (isReload) {
                 VisitType.RELOAD
             } else {
-                if (flags and GeckoSession.HistoryDelegate.VISIT_REDIRECT_SOURCE_PERMANENT != 0) {
+                // Note the difference between `VISIT_REDIRECT_PERMANENT`,
+                // `VISIT_REDIRECT_TEMPORARY`, `VISIT_REDIRECT_SOURCE`, and
+                // `VISIT_REDIRECT_SOURCE_PERMANENT`.
+                //
+                // The former two indicate if the visited page is the *target*
+                // of a redirect; that is, another page redirected to it.
+                //
+                // The latter two indicate if the visited page is the *source*
+                // of a redirect: it's redirecting to another page, because the
+                // server returned an HTTP 3xy status code.
+                if (flags and GeckoSession.HistoryDelegate.VISIT_REDIRECT_PERMANENT != 0) {
                     VisitType.REDIRECT_PERMANENT
-                } else if (flags and GeckoSession.HistoryDelegate.VISIT_REDIRECT_SOURCE != 0) {
+                } else if (flags and GeckoSession.HistoryDelegate.VISIT_REDIRECT_TEMPORARY != 0) {
                     VisitType.REDIRECT_TEMPORARY
                 } else {
                     VisitType.LINK
                 }
+            }
+            val redirectSource = when {
+                flags and GeckoSession.HistoryDelegate.VISIT_REDIRECT_SOURCE_PERMANENT != 0 ->
+                    RedirectSource.PERMANENT
+                flags and GeckoSession.HistoryDelegate.VISIT_REDIRECT_SOURCE != 0 ->
+                    RedirectSource.TEMPORARY
+                else -> RedirectSource.NOT_A_SOURCE
             }
 
             val delegate = settings.historyTrackingDelegate ?: return GeckoResult.fromValue(false)
@@ -504,7 +523,7 @@ class GeckoEngineSession(
             }
 
             return launchGeckoResult {
-                delegate.onVisited(url, visitType)
+                delegate.onVisited(url, PageVisit(visitType, redirectSource))
                 true
             }
         }
--- a/components/concept/storage/src/main/java/mozilla/components/concept/storage/HistoryStorage.kt
+++ b/components/concept/storage/src/main/java/mozilla/components/concept/storage/HistoryStorage.kt
@@ -12,9 +12,9 @@ interface HistoryStorage : Storage {
     /**
      * Records a visit to a page.
      * @param uri of the page which was visited.
-     * @param visitType type of the visit, see [VisitType].
+     * @param visit Information about the visit; see [PageVisit].
      */
-    suspend fun recordVisit(uri: String, visitType: VisitType)
+    suspend fun recordVisit(uri: String, visit: PageVisit)
 
     /**
      * Records an observation about a page.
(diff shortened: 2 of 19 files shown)


linacambridge:
This isn't complete yet—I think we'll also need a way to filter out hidden URLs in views.

linacambridge:
> I think we'll also need a way to filter out hidden URLs in views.

After thinking more about this, let's do the filtering in a-s for now (mozilla/application-services#1715). We can always change a-c or Fenix to exclude hidden on a per-view basis as we need, and that minimizes the changes that Fenix will need to make (just bump the a-c and a-s versions).

See also mozilla/application-services#1492.

linacambridge:
@grigoryk Please review at your convenience!

pocmo:
Will fix this https://github.com/mozilla-mobile/fenix/issues/2542

linacambridge:
> Will this fix this? mozilla-mobile/fenix#2542

Nope, that's different; we intentionally [don't filter out hidden visits from the awesomebar](https://docs.google.com/document/d/1er0UfQBp6ylnqqyjP4Ijr7nbh6abivDdf-axhTznZMM/edit). I'd be curious to hear if they persist once a profile accumulates more history, like @thomcc mentioned in https://github.com/mozilla-mobile/android-components/issues/3027#issuecomment-509765006.

I see the same thing in a fresh Desktop profile: visiting https://www.mozilla.org/firefox/new (which redirects to https://www.mozilla.org/en-US/firefox/new/), then typing "mozilla firefox" into the awesomebar shows this:

Screenshot of redirect source and target in Firefox 71 awesomebar

Doing the same in my regular profile doesn't, because I've visited and bookmarked waaay more Mozilla and Firefox sites.

But that suggestion algorithm was written with Desktop in mind, not mobile. It sounds like it's worth diverging here, because suggestions for redirects are much lower value here.

jdragojevic:
A little nudge here: @linacambridge does this need more work, or @grigoryk does this need more review?

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -467,14 +469,31 @@ class GeckoEngineSession(
...
+                // `VISIT_REDIRECT_TEMPORARY`, `VISIT_REDIRECT_SOURCE`, and
+                // `VISIT_REDIRECT_SOURCE_PERMANENT`.
+                //
+                // The former two indicate if the visited page is the *target*
+                // of a redirect; that is, another page redirected to it.
+                //
+                // The latter two indicate if the visited page is the *source*
+                // of a redirect: it's redirecting to another page, because the
+                // server returned an HTTP 3xy status code.
+                if (flags and GeckoSession.HistoryDelegate.VISIT_REDIRECT_PERMANENT != 0) {
```
rocketsroger: nit: I'm assuming these check works correctly as 
((flags and GeckoSession.HistoryDelegate.VISIT_REDIRECT_PERMANENT) != 0)
But I can't confirm that looking at Kotlin language docs.  Can we explicitly put the brackets around so its clear for readers?

components/concept/storage/src/main/java/mozilla/components/concept/storage/HistoryStorage.kt
```diff
@@ -14,7 +14,7 @@ interface HistoryStorage : Storage {
      * @param uri of the page which was visited.
      * @param visitType type of the visit, see [VisitType].
```
rocketsroger: Update comments with updated param.  But please see comment below first.

components/concept/storage/src/main/java/mozilla/components/concept/storage/HistoryStorage.kt
```diff
@@ -14,7 +14,7 @@ interface HistoryStorage : Storage {
...
      */
-    suspend fun recordVisit(uri: String, visitType: VisitType)
+    suspend fun recordVisit(uri: String, visit: PageVisit)
```
rocketsroger: nit: I like adding the 'RedirectSource' enum.  But I'm  not sure with the decision of adding a new data class.  Looks like we can get away with just adding a extra parameter for recordVisit().  I think just adding a extra parameter makes it easier to read.

docs/changelog.md
```diff
@@ -14,6 +14,8 @@ permalink: /changelog/
 
 * **feature-pwa**
   * Adds `WebAppUseCases.isInstallable` to check if the current session can be installed as a Progressive Web App.
+* **browser-engine-gecko-nightly**, **browser-engine-gecko-beta** and **browser-engine-gecko**
```
rocketsroger: Extra empty line needed

components/concept/storage/src/main/java/mozilla/components/concept/storage/HistoryStorage.kt
linacambridge: I thought about adding the source as an extra parameter, but decided against it because we might also want to [track](https://bugzilla.mozilla.org/show_bug.cgi?id=468710) and eventually [sync](https://bugzilla.mozilla.org/show_bug.cgi?id=623667) referrer chains, which we don't currently do. Four params might get unwieldy—but maybe it's premature to worry about that now.

linacambridge:
Thanks, @rocketsroger! 🎊 Would you like me to go ahead and flatten `PageVisit` into `visitType` and `redirectSource` arguments, or keep it as is?

components/concept/storage/src/main/java/mozilla/components/concept/storage/HistoryStorage.kt
rocketsroger: I see. if there's a reason then I'm good with it.

rocketsroger:
> Thanks, @rocketsroger! 🎊 Would you like me to go ahead and flatten `PageVisit` into `visitType` and `redirectSource` arguments, or keep it as is?

I'm good with keeping it as is.  Since there's a good reason for it.  Please update the parameter and also move the change in changlog.md to the latest version.  I think this is is good to go except the small changes.

docs/changelog.md
rocketsroger: @linacambridge Please move this to the latest version.

linacambridge:
@rocketsroger Done!

docs/changelog.md
```diff
@@ -35,6 +35,7 @@ permalink: /changelog/
     })
   }
   ```
+  * ⚠️ **This is a breaking change**: Redirect source and target flags are now passed to history tracking delegates. As part of this change, `HistoryTrackingDelegate.onVisited()` receives a new `PageVisit` data class as its second argument, specifying the `visitType` and `redirectSource`. For more details, please see [PR #4268](https://github.com/mozilla-mobile/android-components/pull/4268).
```
rocketsroger: @linacambridge looks like you're missing the first line and accidentally add the '+'.  This is causing the ktlint issue.

linacambridge: Hmm, not sure what you mean. Could you push a follow-up commit or a GitHub suggested change with the style you'd like, please?

rocketsroger: My mistake, I thought this was missing the title.  After expanding, I see that you're putting this behind a existing **engine** change.  Once the tests pass I'll approve.  Thanks!

Issue #4284, opened 2019-09-02
Add Downloads to browser-state


pocmo:
Add `Download` state to `browser-state` and synchronize it with `browser-session`.

Needed for #4282.

Issue #4282, opened 2019-09-02
Migrate feature-downloads to use browser-state


pocmo:

csadilek:
Just adding a note here to also refactor how we handle download IDs in our FetchDownloadService, see: https://github.com/mozilla-mobile/android-components/pull/4315#discussion_r321037512

PR #4315, opened 2019-09-04
For #4284, 4285: HitResult and Download consumables for browser-state

csadilek:
@pocmo As discussed, there are 3 parts/commits here:

- Support for adding listeners to Consumables
- Add HitResult to browser-state (and sync from browser-session)
- Sync downloads between browser-session and browser-state

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/Download.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/Download.kt
@@ -5,6 +5,8 @@
 package mozilla.components.browser.session
 
 import android.os.Environment
+import mozilla.components.browser.state.state.content.DownloadState
+import java.util.UUID
 
 /**
  * Value type that represents a Download.
@@ -16,6 +18,7 @@ import android.os.Environment
  * @property userAgent The user agent to be used for the download.
  * @property destinationDirectory The matching destination directory for this type of download.
  * @property referrerUrl The site that linked to this download.
+ * @property id a unique ID of this download.
  */
 data class Download(
     val url: String,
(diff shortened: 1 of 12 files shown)


components/browser/session/src/main/java/mozilla/components/browser/session/Download.kt
```diff
@@ -24,5 +27,17 @@ data class Download(
     val contentLength: Long? = null,
     val userAgent: String? = null,
     val destinationDirectory: String = Environment.DIRECTORY_DOWNLOADS,
-    val referrerUrl: String? = null
+    val referrerUrl: String? = null,
+    val id: String = UUID.randomUUID().toString()
```
csadilek: Now that we add an ID here, we should refactor feature-downloads to use this ID (it generates a unique long right now). I'll take another look tomorrow, but we could also do that as a follow-up.

components/support/base/src/main/java/mozilla/components/support/base/observer/Consumable.kt
```diff
@@ -52,6 +64,22 @@ class Consumable<T> private constructor(
...
+    /**
+     * Adds a listener to be invoked when this [Consumable] is consumed.
+     *
+     * @param listener the listener to add.
+     */
+    @Synchronized
+    fun onConsume(listener: ConsumableListener) {
+        listeners.add(listener)
+    }
+
```
pocmo: Should we add those two methods to the changelog? The Firefox for Fire TV team uses `Consumable` in their code. Maybe those could help them too.

components/browser/session/src/main/java/mozilla/components/browser/session/Download.kt
pocmo: Yeah, let's do it in a follow-up when migrating the feature (#4282).

Issue #4113, opened 2019-08-13
Add Pending Intent Creation Helper


rocketsroger:
Follow up for #3439.
Moving the creation of pending intents behind a Helper  If possible it should be in support-utils.

PR #4329, opened 2019-09-05
 For #4113. Add extension method for creating service `PendingIntent`. 

dector:
---

Moved pending intent creation for crash service to `support-utils`.

### Pull Request checklist

### After merge

### DIFF
--- a/components/support/utils/src/main/java/mozilla/components/support/utils/intents.kt
+++ b/components/support/utils/src/main/java/mozilla/components/support/utils/intents.kt
@@ -16,8 +16,8 @@ import android.os.Build
  *
  * @param context an [Intent] to start a service.
  */
-@JvmName("createPendingIntentForLaunchService")
-fun Intent.asPendingIntentForLaunchService(context: Context): PendingIntent =
+@JvmName("createForegroundServicePendingIntent")
+fun Intent.asForegroundServicePendingIntent(context: Context): PendingIntent =
     if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
         PendingIntent.getForegroundService(context, 0, this, 0)
     } else {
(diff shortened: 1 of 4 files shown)


components/support/utils/src/main/java/mozilla/components/support/utils/intents.kt
```diff
@@ -0,0 +1,25 @@
...
+ * Create a [PendingIntent] instance to run a certain service described with the [Intent].
+ *
+ * This method will allow you to launch a service that will be able to overpass
+ * [background service limitations](https://developer.android.com/about/versions/oreo/background#services)
+ * introduced in Android Oreo.
+ *
+ * @param context an [Intent] to start a service.
+ */
+@JvmName("createPendingIntentForLaunchService")
+fun Intent.asPendingIntentForLaunchService(context: Context): PendingIntent =
```
pocmo: nit: I think the name should mention that it is a **foreground** service!

dector: Naming ideas are welcomed. I don't think that we can use "foreground" here because it's decieving for older android versions. Technically this is just a **service-that-can-start-activities-from-background**.

pocmo: I think using foreground makes still sense here. It will gracefully downgrade on services before Q - similar to `ContextCompat.startForegroundService()`.

rocketsroger: This is a tough one.  Both makes sense.  I would lean towards adding "foreground" because most users of this API will be supporting O+ devices and would need to be aware that this will start a foreground service.

dector: Deal. I'll change the name, thanks. :)

dector: :heavy_check_mark: Done.

PR #4330, opened 2019-09-05
Closes #3558: browser-state: Add engine session state and sync w/ browser-session

### LINKED ISSUE #3558: browser-state: Link to something like EngineSessionHolder
`Session.engineSessionHolder` keeps a reference to the `EngineSession` or `EngineSessionState`. We need to mirror this to the state in `browser-state`. This will be needed for every `SessionState`. Maybe we want to create a sub-object for that, like we did for `ContentState`?

csadilek:
Interestingly, we could live without an action for unlinking, as that only happens when the session is removed and since our new session state is immutable there's nothing to update, but for consistency/readability I've introduced it, plus there may be more use cases later.

There's some special handling here for `restore`, as we first have to dispatch the restore action (which happens in the outer manager) before dispatching the link action (otherwise we won't find the session). Otherwise, all as discussed. I've named the object passed down to the legacy session manager "EngineSessionLinker", but better name suggestions welcome :)

### DIFF
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/EngineStateReducer.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/EngineStateReducer.kt
@@ -1,0 +1,53 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.state.reducer
+
+import mozilla.components.browser.state.action.EngineAction
+import mozilla.components.browser.state.state.BrowserState
+import mozilla.components.browser.state.state.EngineState
+import mozilla.components.browser.state.state.SessionState
+
+internal object EngineStateReducer {
+
+    /**
+     * [EngineAction] Reducer function for modifying a specific [EngineState]
+     * of a [SessionState].
+     */
+    fun reduce(state: BrowserState, action: EngineAction): BrowserState = when (action) {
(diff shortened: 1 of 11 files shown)


components/browser/state/src/main/java/mozilla/components/browser/state/reducer/EngineStateReducer.kt
```diff
@@ -0,0 +1,49 @@
+package mozilla.components.browser.state.reducer
```
pocmo: nit: license header.

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/EngineStateReducer.kt
```diff
@@ -0,0 +1,49 @@
...
+                current
+            }
+        },
+            customTabs = customTabs.map { current ->
+                if (current.id == tabId) {
+                    current.copy(engineState = update.invoke(current.engineState))
+                } else {
+                    current
+                }
+            }
```
pocmo: nit: somehow `tabs` and `customTabs` have a different indentation.

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/EngineStateReducer.kt
csadilek: Thanks, fixed!

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/EngineStateReducer.kt
csadilek: fixed!

csadilek:
bors r=pocmo

PR #4401, opened 2019-09-13

Pike:
State: mozilla-l10n/android-l10n@e78d5868c745d7ab696b35e534183bd97a04c67c

### DIFF
--- a/components/browser/toolbar/src/main/res/values-de/strings.xml
+++ b/components/browser/toolbar/src/main/res/values-de/strings.xml
@@ -6,9 +6,9 @@
     <!-- Content description: For the tracking protection toolbar icon, it is set when the site has tracking protection enabled, but none trackers have been blocked or detected. -->
     <string name="mozac_browser_toolbar_content_description_tracking_protection_on_no_trackers_blocked">Schutz vor Aktivitätenverfolgung ist an</string>
     <!-- Content description: For the tracking protection toolbar icon, it is set when the site has tracking protection enabled, and trackers have been blocked or detected.-->
-    <string name="mozac_browser_toolbar_content_description_tracking_protection_on_trackers_blocked">Schutz vor Aktivitätenverfolgung ist aktiv und blockiert Tracker</string>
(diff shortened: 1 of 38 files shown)


components/browser/toolbar/src/main/res/values-de/strings.xml
```diff
@@ -6,9 +6,9 @@
     <!-- Content description: For the tracking protection toolbar icon, it is set when the site has tracking protection enabled, but none trackers have been blocked or detected. -->
     <string name="mozac_browser_toolbar_content_description_tracking_protection_on_no_trackers_blocked">Schutz vor Aktivitätenverfolgung ist an</string>
     <!-- Content description: For the tracking protection toolbar icon, it is set when the site has tracking protection enabled, and trackers have been blocked or detected.-->
-    <string name="mozac_browser_toolbar_content_description_tracking_protection_on_trackers_blocked">Schutz vor Aktivitätenverfolgung ist aktiv und blockiert Tracker</string>
+    <string name="mozac_browser_toolbar_content_description_tracking_protection_on_trackers_blocked1">Der Tracking-Schutz hat Tracker blockiert</string>
```
cadeyrn: With this change the wording no longer consistent with other Firefox products…

pocmo: @Pike @Delphine What's your recommendation here?

pocmo: CC @MichaelKohler

Pike: Localization reviews shouldn't happen in these PRs, these are really just for exporting the data from one repo to the other.

The goal for these PRs is to automatically merge once tests pass, which is actively worked on, so there's no point in establishing other process in the meantime.

pocmo: Merging 👍

cadeyrn: > Localization reviews shouldn't happen in these PRs

@Pike What is the alternative? It's not possible to discuss translations on Pontoon at all. And since it's not the first time that there is a need to discuss a translation it needs a place to have this kind of conversation.

PR #4409, opened 2019-09-13
Add WebCompat Addon as feature-webcompat

denschub:
This imports the sources for the WebCompat System Addon, hosted at https://github.com/mozilla/webcompat-addon, as an Android Component, in preparation for shipping it in Fenix, see https://github.com/mozilla-mobile/fenix/issues/1174.

---

Now, this looks a bit larger than it actually is. The sources inside `components/feature/webcompat/src/main/assets/extensions/webcompat/` are imported from our GitHub and are, with the exception of some fresh interventions, already merged into mozilla-central and/or already released to the Firefox Desktop and Fennec population. This does not mean you *can't* review them, it just means you don't *have to*. :)

For everything else, please keep in mind that my Kotlin and A-C knowledge-levels are pretty much at 0, so pick apart everything. There are some points I wanted to have additional clarification on:

* I'm a bit torn if this is a `feature` or a `service`. Right now, there is no real user-visible impact besides websites getting fixed, but there will be one. From my understanding of the existing code, `service`s are mainly used for shipping background tasks and helpers for other things. However, I'm happy to move things around!
* This PR has no tests (yet). I'm unsure what exactly to test here, given the extension is only registering a webextension. The WebExt-registering has test coverage in a-c, and our system addon has test-coverage in our repository. Do you suggest just adding a test that checks if the extension is loaded correctly, or do you want something more advanced?
* I used `samples-browser` as a test vehicle, but did not include that change in this PR. I don't know if you actually want this inside the sample browser, given the addon can and will change websites in a way that's maybe unexpected by developers. However, we as the webcompat team, would really like an easy test method for development and manual tests, so the sample browser would be somewhat convenient. What do you suggest here, should I add it to the sample-browser, keep testing with my local patch, or spin another sample that just loads the WebCompat feature?

Also, although this is working for me, I'm sure I missed some things here. Sorry in advance. :)

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/webcompat/src/main/java/mozilla/components/feature/webcompat/WebCompatFeature.kt
+++ b/components/feature/webcompat/src/main/java/mozilla/components/feature/webcompat/WebCompatFeature.kt
@@ -1,0 +1,29 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.feature.webcompat
+
+import mozilla.components.concept.engine.Engine
+import mozilla.components.support.base.log.logger.Logger
+
+/**
+ * Feature to enable website-hotfixing via the Web Compatibility System-Addon.
+ */
+object WebCompatFeature {
+    private val logger = Logger("mozac-webcompat")
+
+    internal const val WEBCOMPAT_EXTENSION_ID = "webcompat@mozilla.com"
+    internal const val WEBCOMPAT_EXTENSION_URL = "resource://android/assets/extensions/webcompat/"
+
+    fun install(engine: Engine) {
+        engine.installWebExtension(WEBCOMPAT_EXTENSION_ID, WEBCOMPAT_EXTENSION_URL,
+            onSuccess = {
+                logger.debug("Installed WebCompat webextension: ${it.id}")
+            },
+            onError = { ext, throwable ->
+                logger.error("Failed to install WebCompat webextension: $ext", throwable)
+            }
+        )
+    }
+}
--- a/.buildconfig.yml
+++ b/.buildconfig.yml
@@ -120,6 +120,10 @@ projects:
     path: components/feature/qr
     description: 'A feature that provides functionality for scanning QR codes.'
     publish: true
+  feature-webcompat:
+    path: components/feature/webcompat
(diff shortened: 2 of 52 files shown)


components/feature/webcompat/src/main/assets/extensions/webcompat/moz.build
```diff
@@ -0,0 +1,78 @@
+# -*- Mode: python; indent-tabs-mode: nil; tab-width: 40 -*-
+# vim: set filetype=python:
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
```
pocmo: Is this moz.build needed here? :)

components/feature/webcompat/src/main/java/mozilla/components/feature/webcompat/WebCompatFeature.kt
```diff
@@ -0,0 +1,50 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.feature.webcompat
+
+import mozilla.components.concept.engine.Engine
+import mozilla.components.concept.engine.webextension.WebExtension
+import mozilla.components.support.base.feature.LifecycleAwareFeature
+import mozilla.components.support.base.log.logger.Logger
+
+class WebCompatFeature(
```
pocmo: nit: A little bit of kdoc here would be helpful to describe what this feature does.

components/feature/webcompat/src/main/java/mozilla/components/feature/webcompat/WebCompatFeature.kt
```diff
@@ -0,0 +1,50 @@
...
+    private val engine: Engine
+) : LifecycleAwareFeature {
+
+    override fun start() {
```
pocmo: This feature could be simplified a lot. Since this component has no native UI at all, you could turn this into an `object` instead of a class. And this object would only need an install method.

Where we create the `Engine` we could then also install your feature directly, once. So you wouldn't need that logic here. :)

denschub:
I addressed all your feedback! I haven't even thought about just installing the extension directly when we create the Engine, thanks for that! :)

pocmo:
bors rety

pocmo:
bors retry

PR #4434, opened 2019-09-16
Closes #4425 Remove Padding When Both URL and Title is Visible In Bro…

### LINKED ISSUE #4425: Custom Tab URL and Website Name Text is Too Small
When using Custom Tab where there is both the website name and the URL.  The text in the URL bar is too small after auto adjusting.  This needs to be fixed so the texts should take more room in the URL bar.

rocketsroger:
…wserToolBar

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/display/DisplayToolbar.kt
@@ -388,8 +388,8 @@ internal class DisplayToolbar(
         val height = MeasureSpec.getSize(heightMeasureSpec)
 
         val fixedHeightSpec = MeasureSpec.makeMeasureSpec(height, MeasureSpec.EXACTLY)
-        val thirdFixedHeightSpec = MeasureSpec
-                .makeMeasureSpec(height / MEASURED_HEIGHT_THIRD_DENOMINATOR, MeasureSpec.EXACTLY)
+        val halfFixedHeightSpec = MeasureSpec
+                .makeMeasureSpec(height / MEASURED_HEIGHT_DENOMINATOR, MeasureSpec.EXACTLY)
         setMeasuredDimension(width, height)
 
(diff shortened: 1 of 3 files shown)


rocketsroger:
Due to the padding between both the URL and Title, It caused the text to shrink till its 1/3 the toolbar size.  I've remove the padding between the URL and Title, since TextViews have default padding already.

Amejia481:
@jonalmeida Do you think the actual padding between the url and the title is too big? Or this was part of the original design?   

**The actual fix**
image

**Chrome**
image

jonalmeida:
> @jonalmeida Do you think the padding between the url and the title is too big? Or this was part of the original design?

Correct, the padding seems a little too much. It should look like the original that we had before:
Screen Shot 2019-09-17 at 4 44 21 PM

rocketsroger:
Figured out it wasn't the padding.  There is actually no padding between the two textviews.  The text was set to 15sp and that was how big the text are.  I'm looking into how best fix this problem.  
1. If we add padding to push these two textviews closer, we end up having issue when user set the display size larger.  
2. If we remove the padding, then the textviews are taking the same amount of space but now the distance between the textviews are too much.

rocketsroger:
Gravity is the solution.  I've made the Title textview gravity bottom and URL textview gravity top.  If only URL then gravity vertical center.  This allows us to remove the padding while still keep the textviews closer together when text is small.

jonalmeida:
Excellent! Keep in mind, these are the UX specs that we're (eventually) trying to get close to for handheld devices. Our toolbar needs to work with tablets and tv screens (Echo Show) as well so we're taking each implementation bit in small steps. 👌

rocketsroger:
Agreed.  Maybe we need to allow the user of BrowserBar to be able to specify the text size?

jonalmeida:
> Agreed. Maybe we need to allow the user of BrowserBar to be able to specify the text size?

Sounds like follow-up work.

rocketsroger:
Sure, I'll create a issue to track it.  Thanks,

PR #4454, opened 2019-09-18
Enable calling the Glean SDK General API from Java

Dexterp37:
This changes the API in a few places:

- adds `@JvmOverloads` to public methods in Glean that  have default arguments;
- adds `@JvmOverloads` to the Configuration object's  public constructor.
- moves the `channel` parameter of the configuration object  earlier in the function signature, to make it more convenient  to call.

@pocmo does these changes make sense, or am I introducing footguns that might be obvious to you? I tested it by building locally and pointing FirefoxReality to that build, this seems to work. No compilation error, at least.

@mdboom for a sanity check :)

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
@@ -91,6 +91,7 @@ open class GleanInternalAPI internal constructor () {
      * as shared preferences
      * @param configuration A Glean [Configuration] object with global settings.
      */
+    @JvmOverloads
     fun initialize(
         applicationContext: Context,
         configuration: Configuration = Configuration()
@@ -268,6 +269,7 @@ open class GleanInternalAPI internal constructor () {
      * @param branch The experiment branch (maximum 30 bytes)
      * @param extra Optional metadata to output with the ping
      */
+    @JvmOverloads
     fun setExperimentActive(
         experimentId: String,
         branch: String,
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
@@ -14,6 +14,8 @@ import mozilla.components.service.glean.net.PingUploader
  *
  * @property serverEndpoint the server pings are sent to. Please note that this is
  *           is only meant to be changed for tests.
+ * @property channel the release channel the application is on, if known. This will be
+ *           sent along with all the pings, in the `client_info` section.
  * @property userAgent the user agent used when sending pings, only to be used internally.
  * @property maxEvents the number of events to store before the events ping is sent
  * @property logPings whether to log ping contents to the console. This is only meant to be used
@@ -21,10 +23,10 @@ import mozilla.components.service.glean.net.PingUploader
  * @property httpClient The HTTP client implementation to use for uploading pings.
  * @property pingTag String tag to be applied to headers when uploading pings for debug view.
  *           This is only meant to be used internally by the `GleanDebugActivity`.
- * @property channel the release channel the application is on, if known. This will be
- *           sent along with all the pings, in the `client_info` section. */
+ */
 data class Configuration internal constructor(
     val serverEndpoint: String,
+    val channel: String? = null,
     val userAgent: String = DEFAULT_USER_AGENT,
     val maxEvents: Int = DEFAULT_MAX_EVENTS,
     val logPings: Boolean = DEFAULT_LOG_PINGS,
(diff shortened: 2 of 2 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
```diff
@@ -14,45 +14,47 @@ import mozilla.components.service.glean.net.PingUploader
...
     // constructor and only initialized with a proper default when calling the primary
     // constructor from the secondary, public one, below.
+    @JvmOverloads
     constructor(
         serverEndpoint: String = DEFAULT_TELEMETRY_ENDPOINT,
+        channel: String? = null,
         maxEvents: Int = DEFAULT_MAX_EVENTS,
-        httpClient: PingUploader = ConceptFetchHttpUploader(lazy { HttpURLConnectionClient() }),
-        channel: String? = null
+        httpClient: PingUploader = ConceptFetchHttpUploader(lazy { HttpURLConnectionClient() })
```
badboy: what's the reason for the change of order here?

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -91,6 +91,7 @@ open class GleanInternalAPI internal constructor () {
      * as shared preferences
      * @param configuration A Glean [Configuration] object with global settings.
      */
+    @JvmOverloads
```
badboy: The JVM Overlords. 🙈 (sorry, had to make that joke, nothing to see here, move on!)

Dexterp37: Oh my. But that's a good one :D

components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
Dexterp37: The `JvmOverloads` is basically generating as many function as the number of defaults. So we'll have:

```java
Configuration(server)
Configuration(server, maxEvents)
Configuration(server, maxEvents, httpClient)
Configuration(server, maxEvents, httpClient, channel)
```

So, with the previous ordering, in order to set the channel, we'd need to provide *all* the arguments, including the `httpClient`. This means we'd need to Javify `ConceptFetchHttpUploader` as well. If we did that, to set the channel, we'd need to:

```java
Configuration("someserver", 200, new ConceptFetchHttpUpdate(new Kotlin.lazy(.....)), "release")
```

vs

```kotlin
Configuration().copy(channel="release)
```

in kotlin. We could probably consider making the `channel` the first parameter, since it's the one that would most likely change.

pocmo:
> @pocmo does these changes make sense, or am I introducing footguns that might be obvious to you? I tested it by building locally and pointing FirefoxReality to that build, this seems to work. No compilation error, at least.

No, that is fine. For everything else in AC we do not guarantee/maintain any Java compatibility though.

Dexterp37:
> 
> 
> > @pocmo does these changes make sense, or am I introducing footguns that might be obvious to you? I tested it by building locally and pointing FirefoxReality to that build, this seems to work. No compilation error, at least.
> 
> No, that is fine. For everything else in AC we do not guarantee/maintain any Java compatibility though.

Thanks for checking. That's a good point. In this specific case, given that we deprecated *service-telemetry*, I think we should. `FirefoxReality` is planning on using sync, and sync is using Glean for telemetry :)

pocmo:
> That's a good point. In this specific case, given that we deprecated service-telemetry, I think we should. FirefoxReality is planning on using sync, and sync is using Glean for telemetry :)

If Glean would not have been developed we would have migrated `service-telemetry` to Kotlin too. :)

Firefox Reality already contains some Kotlin code - so I do not think its super critical. But I can understand that Glean may want broader support and it also is very minimal what's actually needed here.

components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
mdboom: Just to confirm: hopefully that doesn't break the API for existing (Kotlin) usages?

Dexterp37: Not that I found, no. We use named arguments in [Fenix](https://github.com/mozilla-mobile/fenix/blob/8a9340fd9d3ab4e5d1f849cd1de8bd3a2bde4a78/app/src/main/java/org/mozilla/fenix/components/metrics/GleanMetricsService.kt#L402)/reference-browser. So we should be fine.

Dexterp37:
> 
> 
> > That's a good point. In this specific case, given that we deprecated service-telemetry, I think we should. FirefoxReality is planning on using sync, and sync is using Glean for telemetry :)
> 
> If Glean would not have been developed we would have migrated `service-telemetry` to Kotlin too. :)

Eheh, I see :)

> Firefox Reality already contains some Kotlin code - so I do not think its super critical. But I can understand that Glean may want broader support and it also is very minimal what's actually needed here.

Yes, I expected many more changes. Looks like very trivial ones, at least for the General API.

Dexterp37:
bors=pocmo,mdboom,badboy
bors r+

Dexterp37:
bors retry

PR #4459, opened 2019-09-19
Closes #4458: Use cached registration token from ADM if available

### LINKED ISSUE #4458: Clearing app data does not reset the registration ID in ADM
When you clear app data, the assumption was that the app's registration token for the push service should be removed as well.

This is however not the case since the ADM instance is local to the device and not to the app, we can receive a cached version.

Usually this wouldn't matter, but it seems like we were never handling that case.

jonalmeida:
Since the ADM instance does not live in the app itself, clearing app
data doesn't remove the registration ID for it. This means that we never
initialize the native PushManager with the cached code.

Our fix is straight-forward: if we have a cached registration ID, invoke
the `PushProcessor#onNewToken` and follow the regular flow.

---

### Pull Request checklist

### After merge


jonalmeida:
Reference Fire TV issue: https://github.com/mozilla-mobile/firefox-tv/issues/2867

jonalmeida:
bors r=Amejia481

Amejia481:
bors retry

Issue #4241, opened 2019-08-28
Import from Fennec session storage (Open tabs)


pocmo:
To be able to keep "open tabs" when updating Fennec to Fenix we'd need to read from Fennec's Session Storage once.

pocmo:
* https://searchfox.org/mozilla-central/source/mobile/android/base/java/org/mozilla/gecko/SessionParser.java
* https://searchfox.org/mozilla-central/source/mobile/android/base/java/org/mozilla/gecko/GeckoApp.java#210
* https://searchfox.org/mozilla-central/source/mobile/android/base/java/org/mozilla/gecko/home/RecentTabsAdapter.java#168
* https://searchfox.org/mozilla-central/source/mobile/android/geckoview/src/main/java/org/mozilla/gecko/GeckoProfile.java
* https://searchfox.org/mozilla-central/source/mobile/android/geckoview/src/main/java/org/mozilla/gecko/GeckoProfile.java#662-753

pocmo:
First iteration done. Will file a separate issue for integration in Fenix.

buttercookie42:
Am I  reading that commit correctly that this only loads the currently open page and discards all the rest of the data? The full contents of the `entries` array (i.e. that particular tab's session history), as well as `scrolldata` and `formdata` (where present) should basically still be compatible with GeckoView's session saving/restoring APIs, although a little rearranging of the JSON might be required (e.g. I think the `entries` array is now called `history`).

PR #4462, opened 2019-09-19
Add support-migration component for landing Fennec migration code.

pocmo:
For #4241 and other migration code. Using this component we can isolate the migration code from the other components and also share some logic (e.g. for finding the profile).

In this first iteration I added the code for detecting Fennec's default profile (#4463).

---

### Pull Request checklist

### After merge


components/support/migration/src/main/java/mozilla/components/support/migration/FennecProfile.kt
```diff
@@ -0,0 +1,158 @@
...
+}
+
+@Suppress("ComplexMethod")
+private fun findProfiles(
+    mozillaDirectory: File,
+    fileName: String = "profiles.ini"
+): List<FennecProfile> {
+    // Implementation note: Fennec saves information about its profiles in mozilla/profiles.ini.
+    // A Fennec installation can have multiple profiles and the profile names always contain a
+    // salt causing them to be different for every installation.
```
Amejia481: 🤔  I'm not quite sure what salt means in this context?

.buildconfig.yml
```diff
@@ -248,6 +248,10 @@ projects:
     path: components/support/ktx
     description: 'A set of Kotlin extensions.'
     publish: true
+  support-migration:
```
Amejia481: nit: I'm not sure if we should add the `fennec` to the name of the component like `support-fennec-migration` as the `support-migration` name is pretty broad term and everything in this component is related to the `fennec-migration`

components/support/migration/src/main/java/mozilla/components/support/migration/FennecProfile.kt
pocmo: https://en.wikipedia.org/wiki/Salt_(cryptography)

In this situation it means the folder for a profile named "arturo" will not be just  `arturo` but something like `10aaayu4-arturo`. `10aaayu4` will be the salt in this case.

.buildconfig.yml
pocmo: Technically it can be used to migrate other forks like the TOR browser to an AC-based app too. :)

components/support/migration/src/main/java/mozilla/components/support/migration/FennecProfile.kt
Amejia481: Awesome, thanks for the link! I didn't know that salt was useful outside the food business XD

.buildconfig.yml
Amejia481: Great 👍

pocmo:
bors r=Amejia481

PR #4477, opened 2019-09-19
1581556: Enforce that Glean.initialize is only called on the MainThread

mdboom:
Android `LifecycleObserver` s must be registered on the main thread [otherwise there is a race condition that may cause them to not be registered and/or unregister other observers](https://issuetracker.google.com/issues/138953075#comment2).  Also see [the implementation](https://android.googlesource.com/platform/frameworks/support/+/androidx-master-dev/lifecycle/lifecycle-runtime/src/main/java/androidx/lifecycle/LifecycleRegistry.java#187) that has a non-atomic push/operate/pop sequence of events.

This bug is the source of all of the significant Fenix Sentry issues involving Glean:

- https://sentry.prod.mozaws.net/operations/fenix/issues/5856320/?query=is:unresolved%20glean
- https://sentry.prod.mozaws.net/operations/fenix/issues/5863576/?query=is:unresolved%20glean
- https://sentry.prod.mozaws.net/operations/fenix/issues/5837548/?query=is:unresolved%20glean
- https://sentry.prod.mozaws.net/operations/fenix/issues/5831193/?query=is:unresolved%20glean

Beyond theoretical, this is quite likely the source of [Fenix reporting 50% more clients on metrics ping than baseline ping](https://bugzilla.mozilla.org/show_bug.cgi?id=1581556).  Glean has two lifecycle observers.  Failing to register one of them would disable the baseline ping.  Failing to register the other would send metrics pings more than once between runs of the application.  The combination of these two is *probably* enough to explain the problem, though it's hard to compare the scale of Sentry events to the scale of the problems in the data.

@pocmo: Can you please confirm this is the appropriate way to dispatch something **back** to the main thread from a coroutine thread?  Also, there are other instances of `addObserver` that may also need this treatment, but that largely depends on how that code is being used from the application, I suppose.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
@@ -8,6 +8,7 @@ import android.annotation.SuppressLint
 import android.content.Context
 import android.content.pm.PackageManager
 import android.os.Build
+import androidx.annotation.MainThread
 import androidx.annotation.VisibleForTesting
 import androidx.lifecycle.ProcessLifecycleOwner
 import kotlinx.coroutines.Job
@@ -92,6 +93,7 @@ open class GleanInternalAPI internal constructor () {
      * @param configuration A Glean [Configuration] object with global settings.
      */
     @JvmOverloads
+    @MainThread
     fun initialize(
         applicationContext: Context,
         configuration: Configuration = Configuration()
@@ -146,6 +148,9 @@ open class GleanInternalAPI internal constructor () {
         Dispatchers.API.flushQueuedInitialTasks()
 
         // At this point, all metrics and events can be recorded.
+        // This should only be called from the main thread. This is enforced by
+        // the @MainThread decorator on this method.
+        // See https://bugzilla.mozilla.org/show_bug.cgi?id=1581556
         ProcessLifecycleOwner.get().lifecycle.addObserver(gleanLifecycleObserver)
     }
 
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
@@ -39,6 +39,7 @@ import mozilla.components.service.glean.utils.getLocaleTag
 import mozilla.components.service.glean.utils.parseISOTimeString
 import mozilla.components.support.base.log.logger.Logger
 import mozilla.components.support.ktx.android.content.isMainProcess
+import mozilla.components.support.utils.ThreadUtils
 
 @Suppress("TooManyFunctions", "LargeClass")
 open class GleanInternalAPI internal constructor () {
@@ -100,6 +101,12 @@ open class GleanInternalAPI internal constructor () {
         applicationContext: Context,
         configuration: Configuration = Configuration()
     ) {
+        // Glean initialization must be called on the main thread, or lifecycle
+        // registration may fail. This is also enforced at build time by the
+        // @MainThread decorator, but this run time check is also performed to
+        // be extra certain.
+        ThreadUtils.assertOnUiThread()
+
         // In certain situations Glean.initialize may be called from a process other than the main
         // process.  In this case we want initialize to be a no-op and just return.
         if (!applicationContext.isMainProcess()) {
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
@@ -88,6 +88,8 @@ open class GleanInternalAPI internal constructor () {
      * A LifecycleObserver will be added to send pings when the application goes
      * into the background.
      *
+     * This method must be called from the main thread.
+     *
      * @param applicationContext [Context] to access application features, such
      * as shared preferences
      * @param configuration A Glean [Configuration] object with global settings.
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
@@ -24,6 +24,7 @@ import mozilla.components.support.base.log.logger.Logger
 import mozilla.components.service.glean.utils.getISOTimeString
 import mozilla.components.service.glean.utils.parseISOTimeString
 import mozilla.components.service.glean.private.TimeUnit
+import mozilla.components.support.utils.ThreadUtils
 import java.util.Calendar
 import java.util.Date
 import java.util.concurrent.TimeUnit as AndroidTimeUnit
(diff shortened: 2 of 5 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
```diff
@@ -61,7 +64,9 @@ internal class MetricsPingScheduler(val applicationContext: Context) : Lifecycle
     }
 
     init {
-        ProcessLifecycleOwner.get().lifecycle.addObserver(this)
+        CoroutineScope(KotlinxDispatchers.Main).launch {
```
travis79: Same nit about [MainScope](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-main-scope.html) as above.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -145,8 +148,9 @@ open class GleanInternalAPI internal constructor () {
         @Suppress("EXPERIMENTAL_API_USAGE")
         Dispatchers.API.flushQueuedInitialTasks()
 
-        // At this point, all metrics and events can be recorded.
-        ProcessLifecycleOwner.get().lifecycle.addObserver(gleanLifecycleObserver)
+        CoroutineScope(KotlinxDispatchers.Main).launch {
```
travis79: Any reason here you aren't using [MainScope](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-main-scope.html)?

As in:
```Kotlin
MainScope().launch {
    ...
}
```

It seems a little clearer/cleaner...

mdboom: Ah -- I just didn't know about it.  I agree, that's clearer.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -145,8 +147,10 @@ open class GleanInternalAPI internal constructor () {
         @Suppress("EXPERIMENTAL_API_USAGE")
         Dispatchers.API.flushQueuedInitialTasks()
 
-        // At this point, all metrics and events can be recorded.
-        ProcessLifecycleOwner.get().lifecycle.addObserver(gleanLifecycleObserver)
+        MainScope().launch {
```
Dexterp37: By looking at the docs and the [rationale for this](https://github.com/Kotlin/kotlinx.coroutines/issues/829), looks like we should declare `private val mainScope = MainScope()` in the `GleanInternalAPI` class and use `mainScope.launch` here. This follows the same pattern we've used for `CoroutineScope` in the dispatcher.

I don't know if that's *strictly* required, but I wouldn't want this to have surprising behaviours due to its short lived existence.

components/service/glean/src/main/java/mozilla/components/service/glean/scheduler/MetricsPingScheduler.kt
```diff
@@ -61,7 +63,9 @@ internal class MetricsPingScheduler(val applicationContext: Context) : Lifecycle
     }
 
     init {
-        ProcessLifecycleOwner.get().lifecycle.addObserver(this)
+        MainScope().launch {
```
Dexterp37: As above.

docs/changelog.md
```diff
@@ -26,6 +26,9 @@ permalink: /changelog/
 
 * **support-webextensions**
   * 🆕 New component containing building blocks for features implemented as web extensions.
+  
+* **service-glean**
+  * Glean now registers its lifecycle observers on the main thread.
```
Dexterp37: For future reference, any chance you could link to the bug that started it all (bug 1581556) and to the [google issue](https://issuetracker.google.com/issues/138953075#comment2) that explicitly says this is expected as registration must happen on the main thread?

Maybe the bug number is fine, since you stated all the details in the PR message and we can always trace back to it.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
pocmo: tbh I'm suprised and confused why Fenix is calling this on a different thread in the first place (since all other apps call it right from a main thread lifecycle callback of the `Application`). Is that really needed? Do you want to support that? Another route you could take is marking this with `@MainThread` or even use `ThreadUtils.assertOnUiThread()` to enforce it.

I fear that if you allow/support multiple threads that you may run into even more issues - specifically around [visibility](https://www.logicbig.com/tutorials/core-java-tutorial/java-multi-threading/visibility-volatile.html). One example is `initialized` in `Glean`: It's written in `initialize()` and not `synchronized` or `volatile`. If you set it to `true` from the thread that calls `initialize()` then a different thread calling `isInitialized()` may not see that value (yet) and still read `false`.

Dexterp37: I *think* Fenix is doing this for performance reasons but, honestly, I don't know :)

That's a good point with respect to the other implications. The `initialize` method might trigger I/O (off the main thread already!) so, thinking of this again, I'm not too sure this is strictly required. @mdboom we should probably talk to the Fenix people and check with them: maybe we can get along with enforcing init on the main thread.

pocmo: If there are performance reasons (are there?) then we'd want to "upstream" those optimizations into Glean, I assume. :)

> 
> If there are performance reasons (are there?) then we'd want to "upstream" those optimizations into Glean, I assume. :)

Yes, you're completely right. We'll check with them before moving on with this. Thanks for the feedback!

travis79: Can you please explain your concerns about the `MainScope().launch` lifetime.  I read the issue you linked as the "rationale" and it is just in support of creating the MainScope() factory method, I don't see anything talking about lifetime issues.  

I see this same pattern being used in other places in Android Studio.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -146,6 +150,9 @@ open class GleanInternalAPI internal constructor () {
         Dispatchers.API.flushQueuedInitialTasks()
 
         // At this point, all metrics and events can be recorded.
+        // This should only be called from the main thread. This is enforced by
+        // the @MainThread decorator on this method.
```
Dexterp37: Does `@MainThread` break the build if called off the main thread? Or assert? Do we need to add `ThreadUtils.assertOnUiThread`? Moreover, can we add simple test coverage for this?

mdboom: It breaks the build.  Though, possibly only when called from Kotlin -- I'll need to confirm that.  I'm not sure how to write a test for a build failure.

It probably wouldn't hurt to add the assert as well.

Dexterp37: Ah, ok, if it breaks the build is fine. Don't mind the test.

mdboom: On further investigation, it seems to only break the build on some cases, in others, it's just a warning.  So I think doing both build time and run time checks seems safer.  And we can write a test for the run time check :)

Dexterp37: Great, thanks for checking that!

Issue #4421, opened 2019-09-16
Remove download property from Session


pocmo:
Now that `feature-downloads` and `feature-contextmenu` is using `browser-state` we should be able to get rid of the `download` property in `Session` completely. Whoever needs this value can read it from `browser-state`. To my knowledge there's no third-party reader, so it should be pretty safe to get rid of. We could `deprecate` it initially.

PR #4506, opened 2019-09-23
Issue #4421: Remove download property from Session.

pocmo:
---

### Pull Request checklist

### After merge


pocmo:
Blocked: Need to update changelog and prepare Fenix / RB pull requests.

components/feature/contextmenu/src/test/java/mozilla/components/feature/contextmenu/ContextMenuCandidateTest.kt
```diff
@@ -372,6 +374,10 @@ class ContextMenuCandidateTest {
             HitResult.IMAGE_SRC("https://www.mozilla.org/media/img/logos/firefox/logo-quantum.9c5e96634f92.png",
                 "https://firefox.com"))
 
+        // Dispatching a random unrelated action to block on it in order to wait for all other
+        // dispatched actions to have completed.
+        store.dispatch(SystemAction.LowMemoryAction).joinBlocking()
```
pocmo: This is a pattern I started to need now that we have code that calls code that dispatches an `Action`. In those situations I can't `join` on anything because that is transparent to the caller. Of course dispatching `LowMemoryAction` here is a bit stupid - but it does the job. I wonder if we can find a better pattern for testing. What do you think, @csadilek?

csadilek: Ah makes sense. How about an extension method on `Store` that we use in tests e.g. `processAllActions` that internally dispatches a NOOP action and joins on it?

pocmo: That may be an option. I wonder if we could build this in a generic way for all `Store` variants. I would love to not make this available to app code so that it is only usable in tests. But that may only work with a separate module and that's not worth it. Or maybe a compileOnly dependency in support-test? 🤔 🤷‍♂

pocmo: Filed #4529

pocmo:
bors r=csadilek

Issue #3859, opened 2019-07-23
Update push registration when notified by DeviceConstellation


jonalmeida:
There can be instances where the push endpoint in a subscription is invalid. We can only know about this, when another device tries to send an event to us and it fails, leading to the `subscriptionExpired` (a.k.a `pushEndpointExpired`) flag is set.

The next time we pull the device constellation from FxA, we need to tell the push feature to renew it's registration token that will also update autopush, thus generating new endpoints.

Depends on:
- [x] #3769 - Add SendTabFeature to feature-sendtab.

Bug for historical context: https://bugzilla.mozilla.org/show_bug.cgi?id=1359279

jonalmeida:
I think I'm going to pull this out of our current sprint and make it dependant on #3769. If we want to do this now, we need to do it in Fenix's `BackgroundServices.kt` component group which is pretty massive already.

jonalmeida:
This "feature" needs to depend on push and the device constellation. So when we have both, then we can do the dance. This means:
 - Observe the account manager if the push feature is non-null to know when to add the `DeviceConstellationObserver`.
   - This is a pattern that I've had to do in a few places. I wonder if it makes sense to have the device constellation directly on the account manager to observe when it's available (cc: @grigoryk for thoughts).
 - When we get events, check our own device in the constellation state, and if the `subscriptionExpired` flag is true, tell the `AutoPushFeature#forceRegistrationRenewal`.
   - It seems excessive to check on every new event if the flag is set. We could add a periodic check to flip a `checkNow` flag.

Other notes:
- In the `ConstellationState` the `currentDevice` (us) is nullable. I'm not sure how that could be possible if the sync returned back results and we know that we exist! ("I sync, therefore I am.") cc: @grigoryk 
- Another issue is where should this live:
   - My immediate thoughts are that this is a feature of FxA and the `FxaAccountManager` does a lot already, so in `service-firefox-accounts` we could add a `FxaSupportFeature` that depends on push?

PR #4507, opened 2019-09-23
Issue #3859: Add FxA support feature to update push registration

jonalmeida:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FxaPushSupportFeature.kt
+++ b/components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FxaPushSupportFeature.kt
@@ -1,0 +1,191 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.feature.accounts
+
+import android.content.Context
+import androidx.annotation.VisibleForTesting
+import androidx.lifecycle.LifecycleOwner
+import androidx.lifecycle.ProcessLifecycleOwner
+import mozilla.components.concept.push.PushProcessor
+import mozilla.components.concept.sync.AuthType
+import mozilla.components.concept.sync.ConstellationState
+import mozilla.components.concept.sync.DeviceConstellationObserver
+import mozilla.components.concept.sync.OAuthAccount
+import mozilla.components.concept.sync.AccountObserver as SyncAccountObserver
+import mozilla.components.service.fxa.manager.FxaAccountManager
+import mozilla.components.service.fxa.manager.ext.withConstellation
+import mozilla.components.support.base.log.logger.Logger
+import mozilla.components.support.base.utils.SharedPreferencesCache
+import org.json.JSONObject
+
+internal const val PREFERENCE_NAME = "mozac_feature_accounts"
+internal const val LAST_VERIFIED = "last_verified_push_subscription"
+
+/**
+ * A feature used for supporting FxA and push integration where needed. One of the main functions is when FxA notifies
+ * the device during a sync, that it's unable to reach the device via push messaging; triggering a push
+ * registration renewal.
+ *
+ * @param context The application Android context.
+ * @param accountManager The FxaAccountManager.
+ * @param push The push processor instance that needs to be notified.
+ * @param owner the lifecycle owner for the observer. Defaults to [ProcessLifecycleOwner].
+ * @param autoPause whether to stop notifying the observer during onPause lifecycle events.
+ * Defaults to false so that observers are always notified.
+ */
+class FxaPushSupportFeature(
+    context: Context,
+    accountManager: FxaAccountManager,
+    push: PushProcessor,
+    owner: LifecycleOwner = ProcessLifecycleOwner.get(),
+    autoPause: Boolean = false
+) {
+    init {
+        val constellationObserver = ConstellationObserver(context, push)
+
+        val accountObserver = AccountObserver(context, accountManager, constellationObserver, owner, autoPause)
+
+        accountManager.register(accountObserver)
+    }
+}
+
+/**
+ * An [FxaAccountManager] observer to know when an account has been added, so we can begin observing the device
+ * constellation.
+ */
+internal class AccountObserver(
+    private val context: Context,
+    private val accountManager: FxaAccountManager,
+    private val observer: DeviceConstellationObserver,
+    private val lifecycleOwner: LifecycleOwner,
+    private val autoPause: Boolean
+) : SyncAccountObserver {
+    private val logger = Logger("AccountObserver")
+
+    override fun onAuthenticated(account: OAuthAccount, authType: AuthType) {
+        accountManager.withConstellation { constellation ->
+            constellation.registerDeviceObserver(observer, lifecycleOwner, autoPause)
+        }
+    }
+
+    override fun onLoggedOut() {
+        // Delete renewal pref.
+        preference(context).edit().remove(LAST_VERIFIED).apply()
+    }
+}
+
+/**
+ * A DeviceConstellation observer to know when we should notify the push feature to begin the registration renewal.
+ */
+internal class ConstellationObserver(
+    context: Context,
+    private val push: PushProcessor,
+    private val verifier: VerificationDelegate = VerificationDelegate(context)
+) : DeviceConstellationObserver {
+
+    private val logger = Logger("ConstellationObserver")
+
+    override fun onDevicesUpdate(constellation: ConstellationState) {
+        val updateSubscription = constellation.currentDevice?.subscriptionExpired ?: false
+
+        // If our subscription has not expired, we do nothing.
+        // If our last check was recent (see: PERIODIC_INTERVAL_MILLISECONDS), we do nothing.
+        if (!updateSubscription || !verifier.allowedToRenew()) {
+            return
+        }
+
+        logger.warn("We have been notified that our push subscription has expired; renewing registration.")
+
+        push.renewRegistration()
+
+        verifier.increment()
+    }
+}
+
+/**
+ * A helper that rate limits how often we should notify our servers to renew push registration.
+ *
+ * Implementation notes: This saves the timestamp of our renewal and the number of times we have renewed our
+ * registration within the [PERIODIC_INTERVAL_MILLISECONDS] interval of time.
+ */
+internal class VerificationDelegate(context: Context) : SharedPreferencesCache<VerificationState>(context) {
+    override val logger: Logger = Logger("VerificationDelegate")
+    override val cacheKey: String = LAST_VERIFIED
+    override val cacheName: String = PREFERENCE_NAME
+
+    override fun VerificationState.toJSON() =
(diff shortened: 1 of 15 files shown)


components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FxaSupportFeature.kt
```diff
@@ -0,0 +1,109 @@
...
+    private val logger = Logger("ConstellationObserver")
+    private var lastVerified: Long
+        get() = preference(context).getLong(LAST_VERIFIED, System.currentTimeMillis())
+        set(value) = preference(context).edit().putLong(LAST_VERIFIED, value).apply()
+
+    override fun onDevicesUpdate(constellation: ConstellationState) {
+        val updateSubscription = constellation.currentDevice?.subscriptionExpired ?: false
+
+        // If our subscription has not expired, we do nothing.
+        if (updateSubscription) {
```
grigoryk: reverse this

components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FxaSupportFeature.kt
```diff
@@ -0,0 +1,109 @@
...
+            return
+        }
+
+        // If our last check was recent (see: PERIODIC_INTERVAL_MILLISECONDS), we do nothing.
+        if (shouldVerifyNow().not()) {
```
grigoryk: if we're told by our own device record that we have a bad subscription, we should `renewRegistration` regardless of the timestamps.

components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FxaSupportFeature.kt
```diff
@@ -0,0 +1,109 @@
...
+import mozilla.components.concept.sync.OAuthAccount
+import mozilla.components.concept.sync.AccountObserver as SyncAccountObserver
+import mozilla.components.service.fxa.manager.FxaAccountManager
+import mozilla.components.support.base.log.logger.Logger
+
+internal const val PREFERENCE_NAME = "mozac_feature_accounts"
+internal const val LAST_VERIFIED = "last_verified_push_subscription"
+internal const val PERIODIC_INTERVAL_MILLISECONDS = 24 * 60 * 60 * 1000L // 24 hours
+
+class FxaSupportFeature(
```
grigoryk: `FxaPushIntegration`?

components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FxaSupportFeature.kt
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        logger.warn("We have been notified that our push subscription has expired; renewing registration.")
+
+        push.renewRegistration()
+
+        lastVerified = System.currentTimeMillis()
+    }
+
+    private fun shouldVerifyNow(): Boolean {
```
grigoryk: This should live outside of `ConstellationObserver`; it shouldn't be used by it (see comment above), but rather by the "verifyConnection" startup logic.

jonalmeida: Wouldn't that be a different timer? On the rust native side, calling `verifyConnection()` updates subscriptions as well which is separate from requesting a new registration + updating subscriptions.

jonalmeida: Spoke about it offline, this can stay here since verifyConnection does just subscription updates and not registration token.

grigoryk: What I mean is that you don't even need this timer for `ConstellationObserver`. You're not invoking a check here (like in verifyConnection), you're reacting to data received. If your device record tells you you need new subscription, a token refresh+device record update flow is triggered, regardless of timers.

We should also add a daily maximum count for how many times we're allowed to trigger this flow. Backend bugs happen, so if `subscriptionExpired` flag is faulty, we don't want to DDOS ourselves/our clients.

components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FxaSupportFeature.kt
jonalmeida: Oops!

components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FxaSupportFeature.kt
jonalmeida: Ooh I see now. That's an excellent point! I was trying to avoid hitting the server too often with  this catch, but your suggestion is a much better one.

components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FxaSupportFeature.kt
jonalmeida: I made this into `FxaPushSupportFeature` since we use the term 'Integration' in apps when we need to integrate two features on that side of the fence.

components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FxaPushSupportFeature.kt
```diff
@@ -0,0 +1,184 @@
...
+    private val accountManager: FxaAccountManager,
+    private val observer: DeviceConstellationObserver,
+    private val lifecycleOwner: LifecycleOwner,
+    private val autoPause: Boolean
+) : SyncAccountObserver {
+    private val logger = Logger("AccountObserver")
+
+    override fun onAuthenticated(account: OAuthAccount, authType: AuthType) {
+        // Register observer when we have an existing account.
+        if (authType != AuthType.Existing) {
```
grigoryk: Remove this check. You always want to monitor device state.

components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FxaPushSupportFeature.kt
```diff
@@ -0,0 +1,184 @@
...
+    }
+}
+
+/**
+ * A helper that rate limits how often we should notify our servers to renew push registration.
+ *
+ * Implementation notes: This saves the timestamp of our renewal and the number of times we have renewed our
+ * registration within the [PERIODIC_INTERVAL_MILLISECONDS] interval of time.
+ */
+internal class VerificationDelegate(private val context: Context) {
```
grigoryk: Let's use `SharedPreferencesCache` here as a backing storage.

components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FxaPushSupportFeature.kt
```diff
@@ -0,0 +1,184 @@
...
+import mozilla.components.concept.sync.OAuthAccount
+import mozilla.components.concept.sync.AccountObserver as SyncAccountObserver
+import mozilla.components.service.fxa.manager.FxaAccountManager
+import mozilla.components.service.fxa.manager.ext.withConstellation
+import mozilla.components.support.base.log.logger.Logger
+
+internal const val PREFERENCE_NAME = "mozac_feature_accounts"
+internal const val LAST_VERIFIED = "last_verified_push_subscription"
+internal const val PERIODIC_INTERVAL_MILLISECONDS = 24 * 60 * 60 * 1000L // 24 hours
+internal const val MAX_REQUEST_IN_INTERVAL = 500 // 500 requests in 24 hours
```
grigoryk: move relevant constants to the verification delegate

components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FxaPushSupportFeature.kt
```diff
@@ -0,0 +1,184 @@
...
+        return verify
+    }
+
+    /**
+     * Should be called whenever a successful invocation has taken place and we want to record it.
+     */
+    fun increment() {
+        val count = innerCount + 1
+
+        save("$innerTimestamp;$count")
```
grigoryk: I think you want to update `innerTimestamp` to the current time here.

I'm thinking of this as a rolling time window.
```
-----------------------------------------------------------------> time
   |                           |                         |
 last checked                  window                   now
                                 |------------------------|
                                               |
                                        "interval"
```
where `window = currentTime - PERIODIC_INTERVAL_MILLISECONDS`
and `last checked = innerTimestamp`

So when you "increment", you need to bump the "last checked" timestamp. Otherwise, your "inInterval" check may lie to you, since it may think that "last checked" happened longer ago than it actually did.

components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FxaPushSupportFeature.kt
```diff
@@ -0,0 +1,184 @@
...
+    init {
+        val (timestamp, count) = get()
+        innerTimestamp = timestamp
+        innerCount = count
+    }
+
+    /**
+     * Checks whether we're within our rate limiting constraints.
+     */
+    fun shouldVerifyNow(): Boolean {
```
grigoryk: Rename this to `allowedToRenew`.

components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FxaPushSupportFeature.kt
```diff
@@ -0,0 +1,184 @@
...
+        val inRateAcceptance = innerCount <= MAX_REQUEST_IN_INTERVAL
+        val verify = inInterval && inRateAcceptance
+
+        // If it's been PERIODIC_INTERVAL_MILLISECONDS since we last checked, we can reset
+        // out rate limiter and verify now.
+        if (!inInterval) {
+            reset()
+            return true
+        }
+
```
grigoryk:
```
fun isAllowedToRenew() {
  withinTimeFrame = System.currentTimeMillis() - PERIODIC_INTERVAL_MILLISECONDS
  if (lastChecked outside withinTimeFrame) {
    // all good! we haven't checked for a while.
    withinIntervalCounter = 0
    return true
  }
  return withinIntervalCounter < MAX_ALLOWED_CHECK_COUNT
}

fun increment() {
  lastChecked = System.currentTimeMillis()
  withinIntervalCounter++
}
```

components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FxaPushSupportFeature.kt
jonalmeida: Talked offline: see below issue.

components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FxaPushSupportFeature.kt
jonalmeida: Spoke offline: This change ended up looking identical to what we currently had, so I decided to change the variable names since that might have been what was confusing.

jonalmeida:
bors r=grigoryk

jonalmeida:
bors retry

jonalmeida:
bors retry

PR #4519, opened 2019-09-24
Closes #4514: Do not restore parent tab ID for collections

### LINKED ISSUE #4514: Do not save/restore parent tab ID for collections
Tabs in a collection do not have a parent/child relationship and keeping the parent ID can cause problems when opening tabs from collections e.g. the parent could've been removed. See: https://github.com/mozilla-mobile/fenix/issues/5486

csadilek:


csadilek:
bors r=pocmo

PR #4543, opened 2019-09-26
Issue #4282: Remove browser-session dependency from feature-downloads

csadilek:
Just some cleanup we missed? in #4282


csadilek:
bors retry

PR #4552, opened 2019-09-26
Support OPT_OUT_FROM_CUSTOM_TABS

NotWoods:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabIntentProcessor.kt
+++ b/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabIntentProcessor.kt
@@ -7,6 +7,7 @@ package mozilla.components.feature.customtabs
 import android.content.Intent
 import android.content.Intent.ACTION_VIEW
 import android.content.res.Resources
+import androidx.browser.customtabs.CustomTabsIntent.shouldAlwaysUseBrowserUI
 import mozilla.components.browser.session.Session
 import mozilla.components.browser.session.SessionManager
 import mozilla.components.concept.engine.EngineSession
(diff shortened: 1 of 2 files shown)


components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabIntentProcessor.kt
```diff
@@ -34,7 +35,7 @@ class CustomTabIntentProcessor(
         val safeIntent = SafeIntent(intent)
         val url = safeIntent.dataString
 
-        return if (!url.isNullOrEmpty() && matches(intent)) {
+        return if (!url.isNullOrEmpty() && matches(intent) && !shouldAlwaysUseBrowserUI(intent)) {
```
pocmo: How does a user actually opt-out of custom tabs? Is that up to the app? I'm confused why the app wouldn't just send a regular VIEW intent in that case? 🤔

components/feature/customtabs/src/test/java/mozilla/components/feature/customtabs/CustomTabIntentProcessorTest.kt
```diff
@@ -72,9 +76,52 @@ class CustomTabIntentProcessorTest {
...
     @Suppress("UNCHECKED_CAST")
     private fun <T> anySession(): T {
         any<T>()
         return null as T
     }
+
+    companion object {
+        private const val EXTRA_USER_OPT_OUT_FROM_CUSTOM_TABS =
+            "android.support.customtabs.extra.user_opt_out"
+    }
```
pocmo: It's somehow weird that this is the only `EXTRA_*` constant in `CustomTabsIntent` that is private. Should we file a bug to make that public? 🤔

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabIntentProcessor.kt
jonalmeida: I agree, this is a strange flag to have. Are they trying to provide the equivalent of 'Open in browser' as part of the config? Is it supposed to be a user-option when the calling app doesn't let you switch out of a custom tab? Had to tell what the intention is.

I don't think we need to support this since all (most) of our browsers support the functionality that this provides, but maybe that should be up to the browser app as well. 🤷‍♂

NotWoods:
Closing since it's questionable if we want this feature.

Issue #3264, opened 2019-06-05
Support tracking protection exception


pocmo:
We want to support features where the user can add exceptions for specific sites ("Disable tracking protection on YouTube"). Currently Fenix has its own implementation of this using a request interceptor that updates the tracking protection policy.
https://github.com/mozilla-mobile/fenix/blob/94881f824190054d8c2395e2740c3a075dcce047/app/src/main/java/org/mozilla/fenix/AppRequestInterceptor.kt#L26-L42

We think that functionality could be provided by AC for all apps.

Additionally I filed a GeckoView bug to get the opinion from the GV team. Maybe there's the opportunity to improve the implementation if Gecko knows about exceptions.
https://bugzilla.mozilla.org/show_bug.cgi?id=1557009

gabrielluong:
I would like to work on this when its respective GV bug lands.

Amejia481:
The GeckoView team [agreed to provide an api for this,](https://bugzilla.mozilla.org/show_bug.cgi?id=1557009#c5) we are waiting until that to work on the ac part

liuche:
Looks like this will be in GV's Sept sprint, and will get uplifted to GV 70 beta.

Amejia481:
This can be broken down in two pieces:
- [ ] Adding API for for interacting with the tracking protection exceptions
- [ ] Add verification in the `ToolbarPresenter` to verify if a session is on the exception list, if so update the tracking protection state to `OFF_FOR_A_SITE`

cpeterso:
The GV bug was fixed in GV 70: https://bugzilla.mozilla.org/show_bug.cgi?id=1557009#c33

PR #4577, opened 2019-09-29
For issue #3264 Display the OFF_FOR_A_SITE tracking protection state.

Amejia481:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
+++ b/components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
@@ -74,6 +74,7 @@ class ToolbarPresenter(
         }
 
         toolbar.siteTrackingProtection = when {
+            tab.trackingProtection.ignoredOnTrackingProtection -> SiteTrackingProtection.OFF_FOR_A_SITE
             tab.trackingProtection.enabled && tab.trackingProtection.blockedTrackers.isNotEmpty() ->
                 SiteTrackingProtection.ON_TRACKERS_BLOCKED
 
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -197,6 +197,21 @@ class GeckoEngineSession(
         notifyObservers { onTrackerBlockingEnabledChange(false) }
     }
 
+    /**
+     * Indicates if this [EngineSession] should be ignored the tracking protection policies.
+     * @param onResult A callback to inform if this [EngineSession] is in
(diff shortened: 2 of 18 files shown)


components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
```diff
@@ -73,7 +73,24 @@ class ToolbarPresenter(
             Toolbar.SiteSecurity.INSECURE
         }
 
+        val session = tab.engineState.engineSession
+        if (session != null) {
+            try {
+                session.isIgnoredForTrackingProtection { offForThisSession ->
+                    setTrackingProtectionState(tab, offForThisSession)
+                }
```
pocmo: Having a callback here is problematic. Is that executed asynchronously? If so then there's no guarantee that at the time of executing the callback we are still "presenting" this state. :(

Amejia481: > Is that executed asynchronously?

Unfortunately yes, we are making the [request to geckoview ](https://mozilla.github.io/geckoview/javadoc/mozilla-central/org/mozilla/geckoview/ContentBlockingController.html#checkException-org.mozilla.geckoview.GeckoSession-) and they are returning us a `GeckoResult`. Should I make this call blocking?

```kotlin
    override fun isIgnoredForTrackingProtection(onResult: (Boolean) -> Unit) {
        runtime.contentBlockingController.checkException(geckoSession).accept {
            if (it != null) {
                onResult(it)
            } else {
                onResult(false)
            }
        }
    }
```

pocmo: Is there a good time where we could do this call and add it to the state? E.g. do we know when this changes (if it goes through our API) so that we can know when to update the state? Ideally the presenter only renders the state and has no side effects.

Amejia481: >  E.g. do we know when this changes.

Yes, it goes through our API `TrackingProtectionUseCases.AddExceptionUseCase` https://github.com/mozilla-mobile/android-components/blob/b9c4b125b006532ad788560c8cad522d9a4fb9cd/components/feature/session/src/main/java/mozilla/components/feature/session/TrackingProtectionUseCases.kt#L40


I can add an additional property to `TrackingProtectionState` to indicate that is off for this site and change it when is removed `TrackingProtectionUseCases.RemoveExceptionUseCase`

https://github.com/mozilla-mobile/android-components/blob/b9c4b125b006532ad788560c8cad522d9a4fb9cd/components/feature/session/src/main/java/mozilla/components/feature/session/TrackingProtectionUseCases.kt#L61

Amejia481: I updated the pr

Issue #4583, opened 2019-09-30
Engine Icon in Awesomebar Item is cut off in RTL


mcarare:
Fenix Issue : https://github.com/mozilla-mobile/fenix/issues/5522

mcarare:
Verified as fixed: https://github.com/mozilla-mobile/fenix/issues/5522#issuecomment-546322119

PR #4584, opened 2019-09-30
For #4583 Avoid icon engine in suggestion being cut off in RTL

mcarare:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/awesomebar/src/main/res/layout/mozac_browser_awesomebar_item_generic.xml
+++ b/components/browser/awesomebar/src/main/res/layout/mozac_browser_awesomebar_item_generic.xml
@@ -10,6 +10,7 @@
     android:background="?android:attr/selectableItemBackground"
     android:clickable="true"
     android:focusable="true"
+    android:clipToPadding="false"
     android:paddingStart="20dp"
     android:paddingEnd="16dp">
 


components/browser/awesomebar/src/main/res/layout/mozac_browser_awesomebar_item_generic.xml
```diff
@@ -10,6 +10,7 @@
     android:background="?android:attr/selectableItemBackground"
     android:clickable="true"
     android:focusable="true"
+    android:clipToPadding="false"
```
pocmo: I'm confused. Doesn't this mean we draw the icon over the padding we defined? Is that really the right fix here?

mcarare: I could swap padding with margin. Is that better?

mcarare: There are other options that I have taken into consideration that have some tradeoffs:
1. Swap padding with margin - no visual difference, except talkback rectangle will be over the icon in both RTL and LTR
2. Change either paddingStart or paddingEnd to have the same value - the padding will change in both RTL and LTR

I thought that having a smaller padding (because we draw the icon over the padding) just in RTL was the one with the smallest visual change. If you think that another solution is more acceptable, please let me know, I will be happy to try it. TY!

pocmo:
bors try

mcarare:
@pocmo do you want me to find another fix, or this has just slipped through the cracks? :)

PR #4595, opened 2019-10-01
1585021: Fix ordering of events at startup to prevent negative timestamps

mdboom:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1585021 for context.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
@@ -156,7 +156,7 @@ open class GleanInternalAPI internal constructor () {
             metricsPingScheduler.schedule()
         }
 
-        // Flush any tasks that where queued prior to initalization.
+        // Flush any tasks that were queued prior to initalization.
         @Suppress("EXPERIMENTAL_API_USAGE")
         Dispatchers.API.flushQueuedInitialTasks()
 
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
@@ -504,9 +504,9 @@ open class GleanInternalAPI internal constructor () {
     /**
      * Send a list of pings.
      *
-     * The ping content is assembled as soon as possible, but upload is not
-     * guaranteed to happen immediately, as that depends on the upload
-     * policies.
+     * The ping content is assembled synchronously, but serialization to disk
+     * happens in a deferred coroutine, and upload happens later as that depends
+     * on the upload policies.
      *
      * If the ping currently contains no content, it will not be sent.
      *
@@ -550,9 +550,9 @@ open class GleanInternalAPI internal constructor () {
      * Each ping will be looked up in the known instances of [PingType]. If the
      * ping isn't known, an error is logged and the ping isn't queued for uploading.
      *
-     * The ping content is assembled as soon as possible, but upload is not
-     * guaranteed to happen immediately, as that depends on the upload
-     * policies.
+     * The ping content is assembled synchronously, but serialization to disk
+     * happens in a deferred coroutine, and upload happens later as that depends
+     * on the upload policies.
      *
      * If the ping currently contains no content, it will not be sent.
      *
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
@@ -140,6 +140,13 @@ open class GleanInternalAPI internal constructor () {
         // This must be set before anything that might trigger the sending of pings.
         initialized = true
 
+        // There are now dispatcher tasks that we need to perform *before* any tasks
+        // that may have been queued prior to initialization.  Therefore, we turn off
+        // task queueing now, so we can set those early tasks up before actually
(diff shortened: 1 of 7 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -140,6 +140,13 @@ open class GleanInternalAPI internal constructor () {
         // This must be set before anything that might trigger the sending of pings.
         initialized = true
 
+        // There are now dispatcher tasks that we need to perform *before* any tasks
+        // that may have been queued prior to initialization.  Therefore, we turn off
+        // task queueing now, so we can set those early tasks up before actually
+        // flushing the queue below.
+        @Suppress("EXPERIMENTAL_API_USAGE")
+        Dispatchers.API.setTaskQueueing(false)
```
travis79: I think that this may be dangerous as I think it opens up a case where we can get the tasks executed out of order, especially if there is enough time between this and flushing of the queued tasks.  

Since this is the _only_ place that `onReadyToSendPings` is called, maybe it would be better to change the implementation of that function to not use `Dispatchers.API.launch`. Instead create a unique coroutine scope for the purpose of `onReadyToSendPings` or perhaps use `Dispachers.IO`.  The only other coroutines that I could find that get triggered as a result of `onReadyToSendPings`are the ones that serialize the pings to disk if `sendPingsByName` is called. Fortunately, these coroutines aren't using our `API` couroutine scope but `Dispatchers.IO`.  That means that neither they, nor a refactored `onReadyToSendPings` would be subject to the queueing.

mdboom: > I think it opens up a case where we can get the tasks executed out of order, especially if there is enough time between this and flushing of the queued tasks.

That's a good point.  If other tasks get called between this line and the flushing, they will get run *before* everything in the queue.

If `onReadyToSendPings` is called on its own coroutine scope, then there is no guarantee that it will run before any event recording tasks already in the queue.

`onReadyToSendPings` calls `sendPingsByName` which is `sendPings` under the hood which *does* use `Dispatchers.API` for ping collection.  It's only the serialization and writing to disk that happens on the IO thread.  This ping collection needs to happen before any pre-queued tasks, or we risk co-mingling events from two different boot timers in the same ping.

So let me go back to the drawing board.  Obviously flipping the global switch on whether things go to the "real" queue or the secondary task queue that we manage is the wrong approach.  I would just add an API to say "queue this with the real dispatcher now and never put in the task queue", but that doesn't handle the secondary things that that also queues (the `sendPing` calls).  <sigh>

mdboom:
I've refactored this to do this without tweaking the global `queueInitialTasks` flag.

I noticed that `sendPings`/`sendPingsByName` is almost always called from secondarily from a coroutine, so there's no need for it to create a new one (other than the internal bit where it has to wait to launch the work manager).  Instead, I moved the coroutine creation for these to the outer (public) edges, which is basically how all the metrics already work.  This means that the "load and send events on startup" task can call `sendPingsByName` without launching a new task, and the ping collection will be guaranteed to happen before the pre-initialization-queued tasks.

It's a few more changes than before, but I think this is more robust to the global state change problem @travis79 highlighted before.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -535,10 +535,12 @@ open class GleanInternalAPI internal constructor () {
...
         // the WorkManager upload job.
         if (pingSerializationTasks.any()) {
-            // Await the serialization tasks. Once the serialization tasks have all completed,
-            // we can then safely enqueue the PingUploadWorker.
-            pingSerializationTasks.joinAll()
-            PingUploadWorker.enqueueWorker()
+            Dispatchers.API.launch {
+                // Await the serialization tasks. Once the serialization tasks have all completed,
+                // we can then safely enqueue the PingUploadWorker.
+                pingSerializationTasks.joinAll()
```
mdboom: I feel like this join could be removed given this refactor, but also vaguely feel like we tried that already...

mdboom: Oh never mind -- it's because the jobs being joined on are on the IO thread, and we join them in the single dispatcher thread.  Seems necessary.

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -504,25 +504,25 @@ open class GleanInternalAPI internal constructor () {
...
+     * on the upload policies.
      *
      * If the ping currently contains no content, it will not be sent.
      *
      * @param pings List of pings to send.
      * @return The async Job performing the work of assembling the ping
      */
     @Suppress("EXPERIMENTAL_API_USAGE")
-    internal fun sendPings(pings: List<PingType>) = Dispatchers.API.launch {
+    internal fun sendPings(pings: List<PingType>) {
```
Dexterp37: This turned out to be structured _exactly_as on [desktop](https://searchfox.org/mozilla-central/rev/f372e8a46ef7659ef61be9938ec2a3ea34d343c6/toolkit/components/telemetry/app/TelemetryController.jsm#502). I think this is a good call.

PR #4600, opened 2019-10-02
 Closes #4289: Migrate feature-prompts to use browser-state 

### LINKED ISSUE #4289: Migrate feature-prompts to use browser-state

csadilek:
This migrates our feature-prompts module to browser-state. 

@pocmo I ended up introducing `Flow.ifAnyChanged` here because returning a `Pair` or other data class instance from `ifChanged` wouldn't have worked. We're comparing for reference equality in `ifChanged` so returning a new `Pair` would always indicate change and forward the value. 

The `ifChanged` block would have to get quite complicated. We'd have to return a different instance only if changed but then we're doing all the work in the feature module and the API doesn't help us. With `ifAnyChanged` this looks quite nice and we can refine the API more later.

Most other changes are in the unit tests. I also did a round of manual testing, thanks to @Amejia481 for sending me lots of links to test :). 

I'll put up PRs for R-B and Fenix, once this one is approved.

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
@@ -15,14 +15,12 @@ import mozilla.components.browser.session.ext.toTabSessionState
 import mozilla.components.browser.state.action.ContentAction.AddFindResultAction
 import mozilla.components.browser.state.action.ContentAction.ClearFindResultsAction
 import mozilla.components.browser.state.action.ContentAction.ConsumeHitResultAction
-import mozilla.components.browser.state.action.ContentAction.ConsumePromptRequestAction
 import mozilla.components.browser.state.action.ContentAction.RemoveIconAction
 import mozilla.components.browser.state.action.ContentAction.RemoveThumbnailAction
 import mozilla.components.browser.state.action.ContentAction.UpdateHitResultAction
 import mozilla.components.browser.state.action.ContentAction.UpdateIconAction
 import mozilla.components.browser.state.action.ContentAction.UpdateLoadingStateAction
 import mozilla.components.browser.state.action.ContentAction.UpdateProgressAction
-import mozilla.components.browser.state.action.ContentAction.UpdatePromptRequestAction
 import mozilla.components.browser.state.action.ContentAction.UpdateSearchTermsAction
 import mozilla.components.browser.state.action.ContentAction.UpdateSecurityInfoAction
 import mozilla.components.browser.state.action.ContentAction.UpdateThumbnailAction
@@ -39,7 +37,6 @@ import mozilla.components.concept.engine.manifest.WebAppManifest
 import mozilla.components.concept.engine.media.Media
 import mozilla.components.concept.engine.media.RecordingDevice
 import mozilla.components.concept.engine.permission.PermissionRequest
-import mozilla.components.concept.engine.prompt.PromptRequest
 import mozilla.components.concept.engine.window.WindowRequest
 import mozilla.components.support.base.observer.Consumable
 import mozilla.components.support.base.observer.Observable
@@ -103,7 +100,6 @@ class Session(
         fun onThumbnailChanged(session: Session, bitmap: Bitmap?) = Unit
         fun onContentPermissionRequested(session: Session, permissionRequest: PermissionRequest): Boolean = false
         fun onAppPermissionRequested(session: Session, permissionRequest: PermissionRequest): Boolean = false
-        fun onPromptRequested(session: Session, promptRequest: PromptRequest): Boolean = false
         fun onOpenWindowRequested(session: Session, windowRequest: WindowRequest): Boolean = false
         fun onCloseWindowRequested(session: Session, windowRequest: WindowRequest): Boolean = false
         fun onMediaRemoved(session: Session, media: List<Media>, removed: Media) = Unit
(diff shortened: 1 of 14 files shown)


components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt
```diff
@@ -433,37 +458,23 @@ class PromptFeature(
...
-internal fun SessionManager.consumePromptFrom(
+internal fun BrowserStore.consumePromptFrom(
     sessionId: String?,
-    consumer: (PromptRequest) -> Boolean
+    consume: (PromptRequest) -> Unit
 ) {
-    runWithSessionIdOrSelected(sessionId) {
-        it.promptRequest.consume(consumer)
+    if (sessionId == null) {
+        state.selectedTab
```
csadilek: Note: We have to do this as `findCustomTabOrSelectedTab` is not good enough here. We're calling this method with session IDs that can be either for custom tab or regular tabs. If `findCustomTabOrSelectedTab` is called with an ID it always assumes it is for a custom tab.

csadilek:
Added changelog entry and opened R-B PR: https://github.com/mozilla-mobile/reference-browser/pull/918

csadilek:
Made `ifAnyChanged` more efficient by using a sequence so we don't have to process all items if we already found a change.

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt
```diff
@@ -126,7 +137,23 @@ class PromptFeature(
...
+                }
+                .collect { state ->
+                    state?.content?.let {
+                        if (it.promptRequest != activePromptRequest) {
+                            onPromptRequested(state)
+                        } else if (!it.loading) {
+                            promptAbuserDetector.resetJSAlertAbuseState()
+                        }
+                        activePromptRequest = it.promptRequest
+                    }
```
pocmo: The `ifAnyChanged()` is fancy. But it makes the collect more complicated to digest because you now need to `if else` check what actually changed. It doesn't need to block landing this - but we should consider thinking about alternatives eventually.

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt
```diff
@@ -433,37 +458,23 @@ class PromptFeature(
...
-        it.promptRequest.consume(consumer)
+    if (sessionId == null) {
+        state.selectedTab
+    } else {
+        state.findTabOrCustomTab(sessionId)
+    }?.let { tab ->
+        tab.content.promptRequest?.let {
+            consume(it)
+            dispatch(ContentAction.ConsumePromptRequestAction(tab.id))
+        }
```
pocmo: Don't we still need to consume from `Session(Manager)` since otherwise both are not in sync anymore? That's where we used UseCases in other situations.

At least unless we do not remove it from `Session`, which we could do too since I'm pretty sure no one else except this feature reads prompt requests. Otherwise this would be a good follow-up.

In this current version I fail to understand why we try to get the tab first and then dispatch with its id when we have that id already passed to this method?

components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlinx/coroutines/flow/Flow.kt
```diff
@@ -46,3 +46,35 @@ fun <T, R> Flow<T>.ifChanged(transform: (T) -> R): Flow<T> {
...
     }
 }
+
+/**
+ * Returns a [Flow] containing only values of the original [Flow] where the result array
+ * of calling [transform] contains at least one different value.
+ *
+ * Example:
+ * ```
+ * Block: x -> x[0]  // Map to first character of input
```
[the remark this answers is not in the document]
pocmo: That changed already since I wrote that :D

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt
csadilek: Yes, agreed! I think we need API that allows handling individual changes, but this would be a bigger change. Hope we can work on that later this month!

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt
csadilek: @pocmo 

> In this current version I fail to understand why we try to get the tab first and then dispatch with its id when we have that id already passed to this method?

That's because we need to get to the prompt request to consume it. So, we need to lookup the "tab" to get to its `content.promptRequest`.


>At least unless we do not remove it from Session, which we could do too since I'm pretty sure no one else except this feature reads prompt requests. Otherwise this would be a good follow-up.

Ah, you're right! I've added a commit here to remove the property from browser session. This way we don't have to keep it in sync and nobody else uses it.

components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
```diff
@@ -449,24 +445,6 @@ class Session(
             !request.consumeBy(consumers)
     }
 
-    /**
-     * [Consumable] State for a prompt request from web content.
-     */
-    var promptRequest: Consumable<PromptRequest> by Delegates.vetoable(Consumable.empty()) { _, _, request ->
```
pocmo: Bye bye 👋

docs/changelog.md
```diff
@@ -12,6 +12,9 @@ permalink: /changelog/
 * [Gecko](https://github.com/mozilla-mobile/android-components/blob/master/buildSrc/src/main/java/Gecko.kt)
 * [Configuration](https://github.com/mozilla-mobile/android-components/blob/master/buildSrc/src/main/java/Config.kt)
 
+* **feature-prompts**
+  *  ⚠️ **This is a breaking change**: The `feature-prompts` component has been migrated to `browser-state` from `browser-session`. Therefore creating a `PromptFeature` requires a `BrowserStore` instance (instead of a `SessionManager` instance).
+
 # 15.0.0
```
pocmo: Should we call out the removal of `promptRequest`? Technically it's a breaking change. :)

csadilek: ok, added.

csadilek:
OK, did another round of testing.

bors r=pocmo

PR #4615, opened 2019-10-02
Closes #4379 - Upgrade WorkManager to v2.2.0

### LINKED ISSUE #4379: Update AndroidX workmanager dependency
I'm currently updating a bunch of AndroidX libraries. I reverted the workmanager update since `WorkManager.getInstnace()` is now deprecated and the new version requires a `Context` parameter. This had a ripple effect in `glean` where I needed to pass down a `Context` through a long chain. So I file this as a separate issue to fix later.

R4md4c:
---
Upgraded the library into v2.2.0 and removed the deprecated access of WorkManager and now it expects a context to be able to get the instance of the WM.

Also starting from v2.1.0, the dependency on lifecycle_extensions was removed and I had to declare them explicitly to continue using ProcessLifecycleOwner.

### Pull Request checklist

### After merge

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
@@ -40,12 +40,13 @@ import mozilla.components.service.glean.utils.parseISOTimeString
 import mozilla.components.support.base.log.logger.Logger
 import mozilla.components.support.ktx.android.content.isMainProcess
 import mozilla.components.support.utils.ThreadUtils
+import kotlin.properties.Delegates
 
 @Suppress("TooManyFunctions", "LargeClass")
 open class GleanInternalAPI internal constructor () {
     private val logger = Logger("glean/Glean")
 
-    private var applicationContext: Context? = null
+    private var applicationContext: Context by Delegates.notNull()
 
     // Include our singletons of StorageEngineManager and PingMaker
     private lateinit var storageEngineManager: StorageEngineManager
@@ -231,7 +232,7 @@ open class GleanInternalAPI internal constructor () {
      * will perform the migration one more time.
      */
     private fun resetMigration() {
-        val migrationPrefs = applicationContext?.getSharedPreferences(
+        val migrationPrefs = applicationContext.getSharedPreferences(
             MIGRATION_PREFS_FILE,
             Context.MODE_PRIVATE
         )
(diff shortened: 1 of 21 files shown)


R4md4c:
The task group `Android Components - Module :browser-session assembleAndTestAndLintReleaseAll   ` is failing even though I tried to run the same command [1] in my local machine and it is passing, any idea what might be wrong?

[1] `./gradlew clean :browser-session:assemble :browser-session:assembleAndroidTest :browser-session:test :browser-session:lintRelease`

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -40,12 +40,13 @@ import mozilla.components.service.glean.utils.parseISOTimeString
...
 import mozilla.components.support.ktx.android.content.isMainProcess
 import mozilla.components.support.utils.ThreadUtils
+import kotlin.properties.Delegates
 
 @Suppress("TooManyFunctions", "LargeClass")
 open class GleanInternalAPI internal constructor () {
     private val logger = Logger("glean/Glean")
 
-    private var applicationContext: Context? = null
+    private var applicationContext: Context by Delegates.notNull()
```
jonalmeida: `lateinit var` instead?

R4md4c:
@jonalmeida Thanks for your review, I added a Changelog item, but I'm not sure if I've covered all the public API changes that were in this PR.

jonalmeida:
Still needs a review from someone in @mozilla-mobile/telemetry since most of the changes are coming from there.

travis79:
> Still needs a review from someone in @mozilla-mobile/telemetry since most of the changes are coming from there.

Looking now!

components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
Dexterp37: Just to make sure I understand: why is a change required on this line in the first place?

R4md4c: We needed to cache the context to be able to use it in `cancelPingWorkers()`

R4md4c: I could have used an `applicationContext?.let` but I thought if someone has forgotten to call `initialize` and then `cancelPingWorkers ` was called it would've skipped doing the actual work silently wasn't flagging an exception that applicationContext was not initialized.

mdboom:
Users of Glean are free to never call `Glean.initialize` if they aren't collecting telemetry (as Fenix does).  So it isn't really an error to not have set `applicationContext`.  However, I don't think `cancelWorkManager` would ever get called in that case either, so I think we're ok either way.

R4md4c:
Since the `lateinit` change brought controversy, I think it would be a good idea to revert it, and use simply `applicationContext!!` as it achieve the same purpose when compared to making it `lateinit`

PR #4626, opened 2019-10-03
Closes #4469: Add engine API for browserAction support

### LINKED ISSUE #4469: Add engine API for browserAction support
GeckoView issue is tracked here: https://bugzilla.mozilla.org/show_bug.cgi?id=1530402

We will need to hook up a delegate on the web extension object and integrate with our toolbar.

Amejia481:
@csadilek  I want to verify if I'm going towards the right direction.

When the GeckoView api is provided my intention is to notify `onBrowserActionChange` everytime the `ActionDelegate` gets called. 

---

### Pull Request checklist

### After merge


components/browser/session/src/main/java/mozilla/components/browser/session/engine/EngineObserver.kt
```diff
@@ -205,6 +207,15 @@ internal class EngineObserver(
         media.unregisterObservers()
     }
 
+    override fun onBrowserActionChange(action: BrowserAction) {
+        store?.dispatch(
+            WebExtensionAction.UpdateBrowserAction(
```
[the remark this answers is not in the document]
Amejia481: Addressed!

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
```diff
@@ -83,6 +89,48 @@ class ToolbarPresenter(
...
+        contentDescription = title, // TODO: Should we show the title on the UI or just the image?
+        imageDrawable = icon,
+        background = badgeBackgroundColor
+        // TODO: TBD  action.badgeText?
+    ) {
+        onClick()
+    }
+
+    private fun observeWebExtensions() {
+        webExtensionScope = store.flowScoped { flow ->
```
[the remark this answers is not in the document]
Amejia481: Addressed!

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -241,6 +244,12 @@ sealed class TrackingProtectionAction : BrowserAction() {
     data class ClearTrackersAction(val tabId: String) : TrackingProtectionAction()
 }
 
+sealed class WebExtensionAction : BrowserAction() {
+    data class InstallWebExtension(val extension: WebExtensionState) : WebExtensionAction()
+    data class UpdateBrowserAction(val extensionId: String, val browserAction: WebExtensionBrowserAction) : WebExtensionAction()
+    data class ConsumeBrowserAction(val extensionId: String) : WebExtensionAction()
```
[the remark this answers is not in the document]
Amejia481: Addressed!

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -246,6 +247,61 @@ sealed class TrackingProtectionAction : BrowserAction() {
...
+/**
+ * [BrowserAction] implementations related to updating BrowserState.extensions] and TabState.extensions.
+ */
+sealed class WebExtensionAction : BrowserAction() {
+
+    /**
+     * [BrowserAction] implementations related to updating the
+     * [mozilla.components.browser.state.state.BrowserState.extensions]
+     */
+    sealed class BrowserState : WebExtensionAction() {
```
csadilek: We could flatten this into just three actions, but it depends on how you want to implement the reducer. If we keep the action groups / nested groups , let's *not* call them "*State" though. That's pretty confusing to read, as the reducer applies actions to state :)

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/WebExtensionReducer.kt
```diff
@@ -0,0 +1,136 @@
...
+}
+
+private object ExtensionTabStateReducer {
+    /**
+     * [WebExtensionAction] Reducer function for modifying a specific [WebExtensionState] in
+     * [BrowserState]
+     */
+    fun reduce(state: BrowserState, action: WebExtensionAction.TabState): BrowserState {
+        return when (action) {
+            is WebExtensionAction.TabState.InstallWebExtension -> {
```
csadilek: As discussed, we don't need this action. When we get a session-specific override, we can add it to the list if it's not already there.

components/browser/state/src/main/java/mozilla/components/browser/state/state/SessionState.kt
```diff
@@ -11,10 +11,12 @@ package mozilla.components.browser.state.state
...
  * @property trackingProtection the [TrackingProtectionState] of this session.
  * @property engineState the [EngineState] of this session.
+ * @property extensions the list of extensions this [SessionState] has.
  */
 interface SessionState {
     val id: String
     val content: ContentState
     val trackingProtection: TrackingProtectionState
     val engineState: EngineState
+    val extensions: List<WebExtensionState> get() = emptyList()
```
csadilek: nit: let's call the val `extensionState`.

csadilek: Don't need the `get()`?

components/browser/state/src/main/java/mozilla/components/browser/state/state/SessionState.kt
```diff
@@ -11,10 +11,12 @@ package mozilla.components.browser.state.state
  * @property content the [ContentState] of this session.
  * @property trackingProtection the [TrackingProtectionState] of this session.
  * @property engineState the [EngineState] of this session.
+ * @property extensions the list of extensions this [SessionState] has.
```
csadilek: nit: extensionState web extension states specific to this [SessionState].

components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
```diff
@@ -59,6 +60,10 @@ abstract class EngineSession(
         fun onCloseWindowRequest(windowRequest: WindowRequest) = Unit
         fun onMediaAdded(media: Media) = Unit
         fun onMediaRemoved(media: Media) = Unit
+        /**
```
csadilek: nit: Not sure we need kdocs here, but if then we should describe it differently.

This isn't to "add a browser action", but to change/override it for this session e.g. changing the badgeTest.

components/concept/engine/src/main/java/mozilla/components/concept/engine/webextension/WebExtension.kt
```diff
@@ -81,6 +82,29 @@ abstract class WebExtension(
...
+     * @property title The title of the web extension to be visible in the user interface.
+     * @property enabled Indicate if the browser action should be enabled or disabled.
+     * @property icon The image for this browser icon.
+     * @property uri The url to get the HTML document representing the internal user interface of the extension.
+     * @property badgeText The browser action's badge text.
+     * @property badgeTextColor The browser action's badge text color.
+     * @property badgeBackgroundColor The browser action's badge background color.
+     * @property onClick A callback to be executed when this browser action is clicked.
+     */
+    data class BrowserAction(
```
csadilek: Let's move this out to a top-level type again. If you're worried about the name clash we can use `import as`.

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -246,6 +247,61 @@ sealed class TrackingProtectionAction : BrowserAction() {
...
+        /**
+         * Installs a given [extension] to the default
+         * [mozilla.components.browser.state.state.BrowserState.extensions].
+         */
+        data class InstallWebExtension(val extension: WebExtensionState) :
```
csadilek: Now that we our delegate we can also dispatch this action.

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -246,6 +247,61 @@ sealed class TrackingProtectionAction : BrowserAction() {
...
+     * [BrowserAction] implementations related to updating TabState.extensions
+     */
+    sealed class TabState : WebExtensionAction() {
+
+        /**
+         * Overrides the default value of an extension given by
+         * [mozilla.components.browser.state.state.BrowserState.extensions] and replace it
+         * with the provided [extension] on a specific [TabSessionState.extensions].
+         */
+        data class InstallWebExtension(
```
csadilek: As discussed, this one we don't need anymore :).

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
```diff
@@ -42,48 +43,77 @@ class ToolbarPresenter(
...
+        } else {
+            state.extensions
+        }
 
-            else -> SiteTrackingProtection.OFF_GLOBALLY
+        extensions.forEach { extension ->
+            extension.browserAction?.let { extensionAction ->
+                val toolbarAction = convertToToolbarAction(extensionAction)
+                toolbar.addBrowserAction(toolbarAction)
+            }
```
csadilek: I think we'll have to call `invalidateActions` on the BrowserToolbar after this so it redraws?

components/browser/state/src/main/java/mozilla/components/browser/state/state/SessionState.kt
Amejia481: Without it the compiler get mad on us :(
image

components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
Amejia481: Yes, we need the comments, since our last update to detekt.yml file  https://github.com/mozilla-mobile/android-components/commit/f82a69d9f1027cb9a75fc4d29dcf73bd2132a8e0 we get an [UndocumentedPublicFunction](https://arturbosch.github.io/detekt/comments.html#undocumentedpublicfunction) if we don't do that.

components/browser/state/src/main/java/mozilla/components/browser/state/state/SessionState.kt
csadilek: Ah yes, but we don't need to set it to an `emptyList` in the interface. We initialize everything else in the impls `TabSessionState`, right?

components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
csadilek: 👍 we're getting stricter and stricter :)

components/browser/state/src/main/java/mozilla/components/browser/state/state/SessionState.kt
Amejia481: It was a bit convenient to do it there as `CustomTabSessionState` also inherit from  `SessionState` and  I didn't have to add it also on `CustomTabSessionState`.

I added it on `TabSessionState` to have a convenient parameter on the constructor, for the `createTab` function without it, I had to create `TabSessionState` object first and them set its `extensionState`.
  
image

**To keep consistency** between both `TabSessionState`and `CustomTabSessionState`. I will remove the initialization from the interface and initialized in both the child classes with  `emptyList`. Thanks for highlighting it!

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
Amejia481: Good catch! 
Unfortunately, `invalidateActions` is not part of the [Toolbar interface](https://github.com/mozilla-mobile/android-components/blob/master/components/concept/toolbar/src/main/java/mozilla/components/concept/toolbar/Toolbar.kt) is just defined on [BrowserToolbar](https://github.com/mozilla-mobile/android-components/blob/708f41ac193939316037dbbab3750d74095a181d/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt#L583). We can't have access to it without doing an explicit cast `(toolbar as BrowserToolbar).invalidateActions()` and It won't be a good idea to introduce concrete classes. I'm not sure if I should added it as @pocmo is doing some drastic modifications on the toolbar  on https://github.com/mozilla-mobile/android-components/issues/1514 and could cause some headaches for him. What do you thing @pocmo?

csadilek: Another thing we have to consider here is that the SessionState can update multiple times. So, right now we'd be adding the BrowserAction multiple times as well.

components/browser/state/src/main/java/mozilla/components/browser/state/state/SessionState.kt
csadilek: 👍

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
Amejia481: Yeap, for this reason I filed https://github.com/mozilla-mobile/android-components/issues/4651. I will add comment pointing to this issue.

Amejia481: I think that right now we are re-rendering the whole toolbar every time we have changes on `BrowserAction`s, as we are observing changes on both `tabs` or `extensions`.

```kotlin
    fun start() {
        renderer.start()

        scope = store.flowScoped { flow ->
            flow.ifAnyChanged { arrayOf(it.findCustomTabOrSelectedTab(customTabId), it.extensions) }
                .collect { state ->
                    render(state)
                }
        }
    }
```

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
Amejia481: Dispatched :)

Amejia481:
@csadilek  I think I addressed all the concerns please let me know what you think

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
pocmo: I think adding `invalidateActions()` to the `Toolbar` interface makes sense. It already knows about adding actions, so it should know about invalidating them too.

components/browser/state/src/main/java/mozilla/components/browser/state/state/CustomTabSessionState.kt
```diff
@@ -19,7 +19,8 @@ data class CustomTabSessionState(
     override val content: ContentState,
     override val trackingProtection: TrackingProtectionState = TrackingProtectionState(),
     val config: CustomTabConfig,
-    override val engineState: EngineState = EngineState()
+    override val engineState: EngineState = EngineState(),
+    override val extensionState: List<WebExtensionState> = emptyList()
```
pocmo: I feel like the global and the tab `WebExtensionState` should actually be two separate objects. It's not unlikely that we will have more global state properties that will not make sense on the tab-local state. What do you think?

components/browser/state/src/main/java/mozilla/components/browser/state/state/BrowserState.kt
```diff
@@ -12,10 +12,13 @@ import mozilla.components.lib.state.State
...
+ * The extensions here represent the default values for all [BrowserState.extensions].
+ * and can be overridden per [SessionState.extensions].
  */
 data class BrowserState(
     val tabs: List<TabSessionState> = emptyList(),
     val selectedTabId: String? = null,
-
-    val customTabs: List<CustomTabSessionState> = emptyList()
+    val customTabs: List<CustomTabSessionState> = emptyList(),
+    val extensions: List<WebExtensionState> = emptyList()
```
pocmo: It seems like we are always traversing this list to find a specific extension. So I was wondering if a `Map` from id to state may make more sense here? 🤔

components/browser/state/src/main/java/mozilla/components/browser/state/state/CustomTabSessionState.kt
Amejia481: > It's not unlikely that we will have more global state properties that will not make sense on the tab-local state. What do you think?

To be sure, did you mean  with **It's not unlikely** == **it likely**? :)
Do you have any examples on mind?

components/browser/state/src/main/java/mozilla/components/browser/state/state/BrowserState.kt
Amejia481: Yes, agree it makes senses 👍

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
Amejia481: I will include it as part of this pr

components/browser/state/src/main/java/mozilla/components/browser/state/state/BrowserState.kt
Amejia481: Fixed!

components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
```diff
@@ -59,6 +60,10 @@ abstract class EngineSession(
...
+         * Event to notifies that a web extension browser action has changed.
```
csadilek: nit: to notify ...

components/browser/state/src/main/java/mozilla/components/browser/state/state/CustomTabSessionState.kt
Amejia481: Let's approach this later as right now we don't have all the details about which properties only make sense on tab-local vs global one.

components/concept/engine/src/main/java/mozilla/components/concept/engine/webextension/BrowserAction.kt
```diff
@@ -0,0 +1,30 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.concept.engine.webextension
+
+import android.graphics.drawable.Drawable
+
+/**
+ * Value type that represents the state of a browser action within a [WebExtension].
+ *
+ * @property title The title of the web extension to be visible in the user interface.
```
csadilek: This is the title of the browser action https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/browserAction/setTitle not the extension.

components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
```diff
@@ -43,9 +45,10 @@ object WebExtensionSupport {
             }
 
             override fun onInstalled(webExtension: WebExtension) {
-                // TODO see https://github.com/mozilla-mobile/android-components/pull/4626/
-                // store.dispatch(WebExtensionAction.InstallWebExtension(webExtension))
+                store.dispatch(WebExtensionAction.InstallWebExtension(webExtension.toState()))
```
csadilek: 👍

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -246,6 +248,37 @@ sealed class TrackingProtectionAction : BrowserAction() {
...
+ * [TabSessionState.extensionState].
+ */
+sealed class WebExtensionAction : BrowserAction() {
+    /**
+     * Installs a given [extension] to the default [BrowserState.extensions].
+     */
+    data class InstallWebExtension(val extension: WebExtensionState) :
+        WebExtensionAction()
+
+    /**
```
csadilek: nit: Just: Updates the browser action of the given extension.

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -246,6 +248,37 @@ sealed class TrackingProtectionAction : BrowserAction() {
...
+     * Updates a browser action of a given [extensionId] with the provided [browserAction],
+     * on the default [BrowserState.extensions].
+     */
+    data class UpdateGlobalBrowserAction(
```
csadilek: nit: would remove "global"

components/browser/state/src/main/java/mozilla/components/browser/state/state/CustomTabSessionState.kt
```diff
@@ -13,13 +13,17 @@ import java.util.UUID
  * @property content the [ContentState] of this custom tab.
  * @property trackingProtection the [TrackingProtectionState] of this custom tab.
  * @property config the [CustomTabConfig] used to create this custom tab.
+ * @property extensionState a maps of web extension ids and extensions that contains all installed web
```
csadilek: It doesn't contain all web extensions, only contains state when there's is a session/tab specific override.

nit: a map[s] of webextensions...

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
```diff
@@ -42,48 +44,80 @@ class ToolbarPresenter(
...
+            renderWebExtensionActions(state, tab)
+        } else {
+            clear()
+        }
+    }
+
+    @VisibleForTesting(otherwise = PRIVATE)
+    internal fun renderWebExtensionActions(state: BrowserState, tab: SessionState) {
+        val extensionsMap = tab.extensionState.toMutableMap()
+        extensionsMap.putAll(state.extensions)
```
csadilek: Doesn't this do the inverse of what you want? Start with the session specific ones and then override with the global ones?

components/feature/toolbar/src/test/java/mozilla/components/feature/toolbar/ToolbarPresenterTest.kt
```diff
@@ -101,6 +106,110 @@ class ToolbarPresenterTest {
...
+    @Test
+    fun `render overridden web extension action from tabs`() {
+        val mockedBrowserButton = mock<WebExtensionToolbarAction>()
+        val mockedBrowserAction = mock<WebExtensionBrowserAction>()
+        val toolbar: Toolbar = mock()
+        val extensions: Map<String, WebExtensionState> = mapOf(
+            "id" to WebExtensionState("id", "url", mockedBrowserAction)
+        )
+        val store = spy(
+            BrowserStore(
```
csadilek: If we want to test that this overrides the BrowserAction for the Tab, we should also add a global one to the store so we can verify we added the right one (see my comment above on the impl.)

components/browser/state/src/main/java/mozilla/components/browser/state/state/CustomTabSessionState.kt
Amejia481: Thanks for the catch I really meant the other way around

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
Amejia481: Thanks for the catch I really meant the other way around. I updated the pr and I added a test to verified the expected behavior.

components/feature/toolbar/src/test/java/mozilla/components/feature/toolbar/ToolbarPresenterTest.kt
Amejia481: Test added!

Amejia481:
I finished the manual testing and works well!
bors r=csadilek,pocmo

PR #4627, opened 2019-10-03
mozilla-mobile/fenix#5334 - Extendable CustomTabToolbarFeature

NotWoods:
Let the CustomTabToolbarFeature be extended so that apps can customize the toolbar. This will be used for private mode custom tabs in Fenix.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
+++ b/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
@@ -20,6 +20,7 @@ import mozilla.components.browser.session.Session
 import mozilla.components.browser.session.SessionManager
 import mozilla.components.browser.session.runWithSession
 import mozilla.components.browser.state.state.CustomTabActionButtonConfig
+import mozilla.components.browser.state.state.CustomTabConfig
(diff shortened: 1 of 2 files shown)


components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
```diff
@@ -49,50 +50,52 @@ class CustomTabsToolbarFeature(
...
-    internal fun initialize(session: Session): Boolean {
-        session.customTabConfig?.let { config ->
-            // Don't allow clickable toolbar so a custom tab can't switch to edit mode.
-            toolbar.onUrlClicked = { false }
-
-            // If it's available, hold on to the readable colour for other assets.
-            config.toolbarColor?.let {
-                readableColor = getReadableTextColor(it)
-            }
+    open fun initialize(session: Session, config: CustomTabConfig) {
```
Amejia481: If we are going to go with this approach could we add this annotation [@CallSuper](https://developer.android.com/reference/androidx/annotation/CallSuper.html). As consumers could forget calling super after overridden, as their main purpose could be just changing one thing and they could end up  loss the parent implementation.

NotWoods:
Closing in favour of #4700. I think the new color scheme API is a better way to handle this.

PR #4654, opened 2019-10-04
For #4653: Do not display suggestion until icon is ready

sblatz:
---

The update behavior here is *much* cleaner than we had previously:

new behavior

### Pull Request checklist

- [ ] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/android-components/blob/master/docs/changelog.md) or does not need one
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features

### After merge
- [ ] **Milestone**: Make sure issues closed by this pull request are added to the [milestone](https://github.com/mozilla-mobile/android-components/milestones) of the version currently in development.
- [ ] **Breaking Changes**: If this is a breaking change, please push a draft PR on [Reference Browser](https://github.com/mozilla-mobile/reference-browser) to address the breaking issues.

Issue #4653, opened 2019-10-04
Do not render a suggestion until its icon is ready to serve


sblatz:
Updating this behavior to match desktop will provide us with two fixes:

1) We will not have icons "flicker." Currently every time a suggestion is shown its icon gets re-rendered since we have to throw out the old one and re-render the new one. If we actually keep the old icon around and just compare it to the new one passed in, we don't have to null it out and can only set it to the new icon when it's actually different.

2) We will not serve "half rendered" suggestions. Currently it's possible for a suggestion to have text and a description with no icon which will look odd to a user. Again, desktop chooses to WAIT until a suggestion is fully ready before displaying it. It makes sense to wait until the icon is ready to be displayed so we don't show a suggestion in a half rendered state.

sblatz:
I am in the process of fixing the "waterfall" animation on Fenix (seen in [this ticket](https://github.com/mozilla-mobile/fenix/issues/4190)).

I fixed this by making it so we don't add the remove/providers every time, but we still got a "flickering" on the search icons since they would get nulled out [here](https://github.com/mozilla-mobile/android-components/blob/master/components/browser/awesomebar/src/main/java/mozilla/components/browser/awesomebar/layout/DefaultSuggestionViewHolder.kt#L144) every time the view holder got recycled, which would cause them to be redrawn.

The new behavior can be seen in this demo:
![new behavior](https://user-images.githubusercontent.com/4400286/66235311-a04df780-e6a4-11e9-8292-0679812bc823.gif)

PR #4655, opened 2019-10-04
 For #4653: Do not display suggestion until icon is ready

sblatz:
---

@csadilek and I discussed this for quite a bit on Friday. Our takeaways were that we should likely make a follow up PR in the future to do a "timeout icon" where if a network request takes too long to fetch the icon, it just serves a default one instead (after 500ms or so) so that way the user is not stuck waiting for suggestions.

One flaw with this solution is that every suggestion in each type of provider is blocked on the other.
For example if I have 3 websites with different favicons from the HistoryProvider come up, each icon will have to load before ANY of the suggestions are displayed. We could also fix this in a follow up PR but I don't think it's a big issue for now and this is still a huge improvement over the bad visual bugs we have now.

This PR is identical to #4654, but uses "passing in bitmaps" as the solution instead. Created two PR's for easy comparison of solutions :)

### Pull Request checklist

### After merge


components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/BookmarksStorageSuggestionProvider.kt
```diff
@@ -44,7 +45,8 @@ class BookmarksStorageSuggestionProvider(
             AwesomeBar.Suggestion(
                 provider = this@BookmarksStorageSuggestionProvider,
                 id = it.guid,
-                icon = icons.loadLambda(it.url!!),
+                // We can runBlocking here to get the icon since we are already on an IO thread
+                icon = runBlocking { icons?.loadIcon(IconRequest(it.url!!))?.await()?.bitmap },
```
csadilek: I like this solution better than blocking inside `IconLoader` as it's not blocking the main thread. This is blocking the provider thread though until the icon is loaded, meaning that we won't get any suggestions from this provider if an icon is slow to load. So, a slow loading icon will delay displaying *all* suggestions from this provider. This is mitigated by our icon cache and we could add a timeout (upper bound) when fetching icons and fall back to a default icon to address this. It seems like a path forward (make sure the suggestion is ready, has an icon, before it's being displayed), but wondering what @pocmo thinks about it :).

csadilek: @sblatz however, since this is only called from a `suspend` function, we can remove the `runBlocking` by either moving the `loadIcon` out of `into`, or simply making into `suspend` as well. This will have the same consequence of only creating the suggestion once we have the icon.

sblatz:
I got final approval on this from @pocmo this morning :)

components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/HistoryStorageSuggestionProvider.kt
```diff
@@ -41,16 +41,17 @@ class HistoryStorageSuggestionProvider(
...
+        return this.zip(iconRequests) { result, icon ->
             AwesomeBar.Suggestion(
-                provider = this@HistoryStorageSuggestionProvider,
-                id = it.id,
-                icon = icons.loadLambda(it.url),
-                title = it.title,
-                description = it.url,
-                score = it.score,
-                onSuggestionClicked = { loadUrlUseCase.invoke(it.url) }
+                    provider = this@HistoryStorageSuggestionProvider,
```
csadilek: nit: formatting.

components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/BookmarksStorageSuggestionProvider.kt
```diff
@@ -39,15 +39,17 @@ class BookmarksStorageSuggestionProvider(
...
+
+        return this.zip(iconRequests) { result, icon ->
             AwesomeBar.Suggestion(
-                provider = this@BookmarksStorageSuggestionProvider,
-                id = it.guid,
-                icon = icons.loadLambda(it.url!!),
-                title = it.title,
-                description = it.url,
-                onSuggestionClicked = { loadUrlUseCase.invoke(it.url!!) }
+                    provider = this@BookmarksStorageSuggestionProvider,
```
csadilek: nit: formatting 4 spaces, not 8.

components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/provider/SessionSuggestionProvider.kt
```diff
@@ -30,28 +32,28 @@ class SessionSuggestionProvider(
...
-                    AwesomeBar.Suggestion(
-                        provider = this,
-                        id = session.id,
-                        title = session.title,
-                        description = session.url,
-                        icon = icons.loadLambda(session.url),
-                        onSuggestionClicked = { selectTabUseCase(session) }
-                    )
+                        AwesomeBar.Suggestion(
+                                provider = this,
```
csadilek: nit:formatting.

pocmo:
bors r-

There's a ktlint failure. I'll push a fix.

PR #4657, opened 2019-10-04
Closes #2887: Allow reporting exceptions without crashing

### LINKED ISSUE #2887: lib-crash: Allow reporting exceptions without crashing
In some situations you need to catch a certain type of exception (famous example: `JSONException`). In those situations often you can continue without crashing the app. In the case of an `JSONException` the cause is often that the JSON was different than what you expected. In those situations it would be nice to not only silently continue running but also report this exception (like a crash) via `lib-crash`. With that this error is visible to us and we can fix our code.

rocketsroger:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
+++ b/components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
@@ -14,8 +14,8 @@ import java.io.File
 
 /**
  * A [CrashReporterService] implementation for recording metrics with Glean.  The purpose of this
- * crash reporter is to collect crash count metrics by capturing [Crash.UncaughtExceptionCrash] and
- * [Crash.NativeCodeCrash] events and record to the respective
+ * crash reporter is to collect crash count metrics by capturing [Crash.UncaughtExceptionCrash],
+ * [Throwable] and [Crash.NativeCodeCrash] events and record to the respective
  * [mozilla.components.service.glean.private.CounterMetricType].
  */
 class GleanCrashReporterService(
@@ -34,6 +34,7 @@ class GleanCrashReporterService(
         // These keys correspond to the labels found for crashCount metric in metrics.yaml as well
         // as the persisted crashes in the crash count file (see above comment)
         const val UNCAUGHT_EXCEPTION_KEY = "uncaught_exception"
+        const val CAUGHT_EXCEPTION_KEY = "caught_exception"
         const val NATIVE_CODE_CRASH_KEY = "native_code_crash"
     }
 
@@ -89,8 +90,8 @@ class GleanCrashReporterService(
 
     /**
      * Parses the crashes collected in the persisted crash file.  The format of this file is simple,
-     * each line may contain either [UNCAUGHT_EXCEPTION_KEY], or [NATIVE_CODE_CRASH_KEY] followed by
-     * a newline character.
+     * each line may contain [UNCAUGHT_EXCEPTION_KEY], [CAUGHT_EXCEPTION_KEY], or [NATIVE_CODE_CRASH_KEY]
+     * followed by a newline character.
      *
      * Example:
      *
@@ -99,6 +100,7 @@ class GleanCrashReporterService(
      * uncaught_exception\n
      * native_code_crash\n
      * uncaught_exception\n
+     * caught_exception\n
      * <--End of file-->
      *
      * It is unlikely that there will be more than one crash in a file, but not impossible.  This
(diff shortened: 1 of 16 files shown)


components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
```diff
@@ -34,6 +34,7 @@ class GleanCrashReporterService(
         // These keys correspond to the labels found for crashCount metric in metrics.yaml as well
         // as the persisted crashes in the crash count file (see above comment)
         const val UNCAUGHT_EXCEPTION_KEY = "uncaught_exception"
+        const val CAUGHT_EXCEPTION_KEY = "caught_exception"
```
travis79: Please also update the label [here](https://github.com/mozilla-mobile/android-components/blob/6c6dc3ce553b1032c5d3e183fb61201e4be39ed2/components/lib/crash/metrics.yaml#L19) in the metrics.yaml to match.  You may need to do a clean and then build for the changes to the metrics.yaml to be picked up by the Glean_Parser.  I don't think that there are any pipeline changes to make as a result of this, but I'm looking into that just to be sure.

travis79: Nope, looks like the pipeline treats the labels as arbitrary string keys so nothing to change beyond this.

rocketsroger: Thanks for catching this.

rocketsroger:
bors r=travis79

pocmo:
bors r-

Let's get an AC review too. I'll do that quickly. :)

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/MozillaSocorroService.kt
```diff
@@ -28,8 +28,14 @@ import java.util.concurrent.TimeUnit
...
 
-private val defaultServerUrl = "https://crash-reports.mozilla.com/submit?" +
+private const val DEFAULT_SERVER_URL = "https://crash-reports.mozilla.com/submit?" +
         "id=${BuildConfig.MOZ_APP_ID}&version=${BuildConfig.MOZILLA_VERSION}&${BuildConfig.MOZ_APP_BUILDID}"
+
+@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+internal const val CAUGHT_EXCEPTION_PREFIX = "Not a crash"
+
+@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+internal const val UNCAUGHT_EXCEPTION_PREFIX = "Fatal crash"
```
pocmo: Are those for distinguishing both types on Crashstats? I wonder if we should use some kind of prefix like on mailing lists, e.g. `[CRASH]`. Not sure about the other though. :) `[INFO]`, `[REPORT]`, ..

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/SentryService.kt
```diff
@@ -104,3 +118,11 @@ private fun createMessage(crash: Crash.NativeCodeCrash): String {
 
     return "NativeCodeCrash(fatal=$fatal, minidumpSuccess=$minidumpSuccess)"
 }
+
+private fun createMessage(crash: Crash.UncaughtExceptionCrash): String {
+    return "Fatal crash: ${crash.throwable.message})"
+}
```
pocmo: Won't that mess with our existing crashes since they now start to have a different message? Maybe we should only add a prefix for the new type of "non-crashes"? 🤔

rocketsroger: Agreed.  Sure, I'll remove the prefix for crashes.  Only do it for non-crashes

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/MozillaSocorroService.kt
rocketsroger: I'll use the [INFO] tag.  Assuming we won't put a prefix for the uncaught exception since we don't want to affect uncaught exceptions.

rocketsroger:
bors = travis79, pocmo

rocketsroger:
bors r=[travis79, pocmo]

rocketsroger:
bors r=[travis79, pocmo]

PR #4658, opened 2019-10-04
Auto-publication workflow support

grigoryk:
This requires a Fenix counterpart: https://github.com/mozilla-mobile/fenix/pull/5801

Initially, I went down the "make composite builds work well with a-c" rabbithole, and while they generally worked, experience in Android Studio was horrible. It would get _very_ confused and broken. So instead, I've settled for automating a publication workflow.

This patch enabled support for an auto-publication workflow for android-components.
It automates a common pattern seen in local development:

Old way:
- after every change in a-c, publish it locally with a unique version (bumping it manually)
- manually modify Fenix to consume a custom version of a-c from a mavenLocal repository
- this involves modifying Dependencies.kt for each new a-c version, fixing a few things in `app/build.gradle`, and specifying a new repository in `build.gradle`. 

New way:
- set a flag in fenix's local.properties to enable auto-publication
- run Fenix builds after making changes to a-c. Changes in a-c will be automatically picked up.
- no other changes are necessary to any Fenix files other than a single flag in local.properties!
- manually bumping android-components is no longer needed!

### DIFF
--- a/publish.gradle
+++ b/publish.gradle
@@ -8,6 +8,11 @@ def libVcsUrl = properties.libVcsUrl
 def libLicense = properties.libLicense
 def libLicenseUrl = properties.libLicenseUrl
 
+def getTimestamp() {
+    def date = new Date()
+    return date.format('yyyyMMddHHmmss')
+}
+
 ext.configurePublish = { groupIdArg, artifactIdArg, descriptionArg ->
     apply plugin: 'digital.wup.android-maven-publish'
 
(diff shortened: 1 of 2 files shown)


grigoryk:
For posterity:
The reason i've added the substitute-local-ac.gradle script to a-c side is so that a-c repo is in control of how its modules are processed; it's a bit academic, since the actual logic just looks at the group, of a module, and so could easily live in fenix as well... but, this lets Fenix (or another consuming application) not care about details of which modules in a-c need to be changes for auto-publication stuff. for example, if a-c adds a module that's not in group org.mozilla.components, fenix won't care or need to be changed - we'll just need to adjust that a-c script.

This also follows the pattern we already have established for substituting GeckoView dependencies with artifacts from local `mozilla-central`.
(e.g. in mozilla-central there's a substitute-geckoview-default.gradle script which knows how to accomplish this, and fenix simply applies it)

grigoryk:
Also, two follow-ups from this:
- move most of the gradle code out of the Fenix PR and into the `substitute-local-ac.gradle` script in a-c
- figure out how to make the `does a-c need to be published?` check zero-cost. Currently, we compile a-c and look if that actually ran any tasks, and this adds overhead. Ideally, figuring that out should have very little overhead. This will allow Fenix builds without any a-c changes to be as fast as regular fenix builds, allowing us to just recommend people to always use the `local.properties` flag.

PR #4674, opened 2019-10-07
1586881: Fix high error rate on glean.baseline.duration

mdboom:
There is a race condition in how the TimingManager is used here.
When stopping a timer, we dispatch a task to actually record the value in the
timespan, and only reset the TimingManager in that async task after the
recording, which requires I/O, is complete.

If the timespan is started in the meantime, an error will be recorded, rather
than safely starting a new timespan measurement.

This change could be considered on the principle (followed elsewhere), that
we do timing management and measurement synchronously, and only recording
asynchronously.

---

### Pull Request checklist

### After merge


mdboom:
I'm marking this as "draft" because I'm stumped as to how to write a test for it -- the testing infrastructure runs tasks synchronously, and this is sort of dependent on running some things asynchronously.

Dexterp37:
> 
> 
> I'm marking this as "draft" because I'm stumped as to how to write a test for it -- the testing infrastructure runs tasks synchronously, and this is sort of dependent on running some things asynchronously.

Yes, testing this would be a bit tricky. I think the fix here makes sense, I would merge this anyway.

PR #4684, opened 2019-10-08
Closes #4679 - lower download notification importance for Android O+

### LINKED ISSUE #4679: DL finish notification has too high priority
Android 9

"download finished" notifications play sound, and this is annoying. 

This notification should have lower priority and stay silent.

pt2121:
---

I think  [IMPORTANCE_LOW](https://developer.android.com/reference/android/app/NotificationManager.html#IMPORTANCE_LOW) is the right one but I am not sure if this requires ux-feedback.
https://material.io/design/platform-guidance/android-notifications.html#settings

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadNotification.kt
+++ b/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadNotification.kt
@@ -28,7 +28,8 @@ import kotlin.random.Random
 @Suppress("TooManyFunctions")
 internal object DownloadNotification {
 
-    private const val NOTIFICATION_CHANNEL_ID = "Downloads"
+    private const val NOTIFICATION_CHANNEL_ID = "mozac.feature.downloads.generic"
+    private const val LEGACY_NOTIFICATION_CHANNEL_ID = "Downloads"
 
     internal const val EXTRA_DOWNLOAD_ID = "downloadId"
 
@@ -49,6 +50,7 @@ internal object DownloadNotification {
             .setOngoing(true)
             .addAction(getPauseAction(context, downloadState.id))
             .addAction(getCancelAction(context, downloadState.id))
+            .setPriority(NotificationCompat.PRIORITY_LOW)
             .build()
     }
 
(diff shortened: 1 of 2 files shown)


components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadNotification.kt
```diff
@@ -110,7 +110,7 @@ internal object DownloadNotification {
             val channel = NotificationChannel(
                 NOTIFICATION_CHANNEL_ID,
                 context.getString(R.string.mozac_feature_downloads_notification_channel),
-                NotificationManager.IMPORTANCE_DEFAULT
+                NotificationManager.IMPORTANCE_LOW
```
pocmo: Note that you cannot retroactively change the importance of a channel. If we'd really want to make this change for all existing users then we'd have to create a new channel id and remove the old channel.

mcarare: Couldn't we just remove the sound for the channel? (`.setSound(null,null)`)
This way we avoid creating a new channel, and also avoid issues with a lower importance (notification on channel with `IMPORTANCE_LOW` will not appear if the status bar already is filled with higher importance notifications.)

NotWoods: I think in this case let's create a new channel ID. I'm fine with the low importance for download notifications.

mcarare: FYI: There are some reports that in Android Q even the `IMPORTANCE_LOW` priority plays a sound. I haven't tested that, though...

pt2121: Thanks all for the inputs. Updated - mostly used https://github.com/mozilla-mobile/android-components/commit/31a0c2fb6b7e891fd42a40f2d4285e31408521e4 as a reference.

pt2121:
Updated my code. Can anyone take another look when having a chance?

pt2121:
Done!

PR #4688, opened 2019-10-08
Set version to 17.0.0.

pocmo:


pocmo:
bors r=MickeyMoz,MozLando

🙃

PR #4693, opened 2019-10-09
Fennec data migration scaffolding

grigoryk:
This is going to be a rolling PR for a little bit, until this is ready to land.
Companion Fenix PR: https://github.com/mozilla-mobile/fenix/pull/5878

Workflow for testing it:
- fetch an ESR68 build from TC: https://tools.taskcluster.net/index/gecko.v2.mozilla-esr68.signed-nightly.nightly.latest.mobile
- re-sign it using a local debug keystore:
  - `zip -d target.apk "META-INF/*"`
  - `~/Code/mozilla-central/mobile/android/debug_sign_tool.py target.apk`
- enable auto-publication workflow for a-c in fenix
- assemble fennecProduction version of Fenix (built against this branch, see auto-publication)
  - `./gradlew assembleFennecProduction`
- sign it using keys used to sign fennec esr apk (otherwise we won't be able to overwrite one with another):
  - `~/Code/mozilla-central/mobile/android/debug_sign_tool.py app/build/outputs/apk/geckoBeta/fennecProduction/app-geckoBeta-x86-fennecProduction-unsigned.apk`
- install fennec esr apk build
- generate some profile data for migration (sign-into fxa, browse around, etc..)
- install fenix build: `adb install -r app/build/outputs/apk/geckoBeta/fennecProduction/app-geckoBeta-x86-fennecProduction-unsigned.apk`

TODO:
- currently needs https://github.com/mozilla/application-services/pull/1942 to work
- add separate migration versioning for history and bookmarks; currently the whole thing is versioned
- add migration tests
  - in fennec, we have a bunch of "schema migration" tests that operate over older versions of browser.db
  - this is a good approach here as well - we can run migration code in tests against a few test db files, that exercise various history/bookmarks configurations and try multiple schema versions
- improve the workflow around this, it's quite tedious currently


components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigration.kt
```diff
@@ -0,0 +1,216 @@
...
+        return JSONObject().also { result ->
+            result.put("list", JSONArray().also { list ->
+                this.forEach {
+                    list.put(JSONObject().also { run ->
+                        run.put("migration", it.key.name)
+                        run.put("version", it.value.version)
+                        run.put("success", it.value.success)
+                    })
+                }
+            })
```
pocmo: random nit: I don't feel like the fancy use of `also` chains here makes this actually easier to read than just sequential code. :)

components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigration.kt
```diff
@@ -0,0 +1,216 @@
...
+        migrations.forEach { versionedMigration ->
+            logger.debug("Migrating $versionedMigration")
+
+            val migrationResult = when (versionedMigration.first) {
+                Migration.HISTORY -> migrateHistory()
+                Migration.BOOKMARKS -> migrateBookmarks()
+                Migration.OPEN_TABS -> migrateOpenTabs()
+            }
+
+            results[versionedMigration.first] = versionedMigration.second to migrationResult
```
pocmo: nit: The usage of `first` and `second` and a map with a pair makes it hard to follow this code.

components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigration.kt
```diff
@@ -0,0 +1,216 @@
...
+            MigrationResult.Failure(e)
+        }
+    }
+
+    private fun migrateOpenTabs(): MigrationResult {
+        if (profile == null) {
+            return MigrationResult.Failure(IllegalStateException("Missing Profile path"))
+        }
+        return try {
+            FennecSessionMigration.migrate(File(profile.path))
```
pocmo: What's your plan for the result here? I would like the migration code to take care of taking the result and injecting this into the sessionmanager/store.

components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigration.kt
```diff
@@ -0,0 +1,216 @@
...
+ * Result of running a set of [Migration]s. Maps a [Migration] to a tuple of (migration version, [MigrationResult]).
+ */
+typealias MigrationsResult = Map<Migration, Pair<Int, MigrationResult>>
+
+// Describes ordering and versioning of migrations.
+private val migrations = listOf(
+    Migration.HISTORY to 1,
+    Migration.BOOKMARKS to 1,
+    Migration.OPEN_TABS to 1
+)
```
pocmo: I wonder if we can make this more transparent for `MigrationHistoryStore`. So that the store doesn't know about what migrations it actually runs. Our migration code would implement some `Migration` interface (or whatever)  and we register those with the migration store/manager thingy here. This allows us to more flexible "compose" a migration instead of hardcoding everything in the store. What do you think? I think it would be quite nice if Fenix (or another Fennec fork) could be in control what gets migrated this way (and we can land things in AC without immediately affecting Fenix with the next snapshot).

components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigration.kt
grigoryk: \/\o/\/ I kind of like it, it looks nested :)

components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigration.kt
grigoryk: As-is, that logic would need to live in Fenix. If we can configure `FennecMigration` with the sessionStore, then `migrateOpenTabs` can do the injection.

components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigration.kt
grigoryk: I've updated `FennecMigration` constructor to take in a list of migrations and their versions, allowing Fenix to configure run whatever migrations it needs. I think that achieves what you describe, right?

This was also necessary to make things testable.

components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigration.kt
grigoryk: Ah yeah, pairs kind of suck :(

grigoryk:
@pocmo ended up rewriting most of this, partially because I couldn't get the type system to co-operate, and partially because a builder pattern just feels much better for this thing.

grigoryk:
Testing this in Fenix, I'm running into an issue with open tabs migration. But not the migration itself, but the fact that I need to initialize `sessionManager` within application's `onCreate`. This seems to lead to a crash with a very unclear stacktrace.

I'm also seeing that `onCreate` is being executed twice on startup, in very close succession. Perhaps that's playing a role here as well.

History and bookmarks migrations sort of work, but tend to crash on first run hitting https://github.com/mozilla/application-services/issues/1965. On second run, migrated data is there.

grigoryk:
It appears that I forgot to wrap my migration code in a `isMainProcess` check, and so I'd end up initialize Gecko multiple times.

grigoryk:
... and, as expected at this point, missing `isMainProcess` check was the exact reason I was hitting mozilla/application-services#1965! So, it actually all works now.

PR #4695, opened 2019-10-09
Closes #4681: Support observing engine state changes for web extensions

### LINKED ISSUE #4681: Provide a mechanism to observe engine state changes (for web extensions)
We currently have our `EngineObserver,` which observes changes to `EngineSession`s, but we have no way of observing engine changes that are independent of sessions.

This is relevant for web extensions where we need to:
- observe that an extension has been installed/uninstalled or enabled/disabled
- observe that an extension wants to open a new tab
- observe that an extension wants to add/update a global browser or page action
- ...

A solution here should only be based on `browser-state`, as we're moving away from `browser-session`. Ultimately, we'll be able to pass a store instance to engine, but that's not possible right now, because we'd introduce a dependency cycle. It seems the simplest solution for now is to add a `feature-webextensions` or a similar component that bridges the gap and makes the required engine state changes observable via the store.

csadilek:


components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
```diff
@@ -0,0 +1,49 @@
...
+import mozilla.components.concept.engine.EngineSession
+import mozilla.components.concept.engine.webextension.WebExtension
+import mozilla.components.concept.engine.webextension.WebExtensionDelegate
+
+/**
+ * Provides functionality to make sure web extension related events in the
+ * [Engine] are reflected in the browser and session states by dispatching
+ * the corresponding actions to the [BrowserStore].
+ */
+object WebExtensionSupport {
```
pocmo: I almost feel like this should go into a `feature-webextensions` since this is going to grow and actually provide a `feature`. :)

components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
```diff
@@ -0,0 +1,49 @@
...
+     */
+    fun initialize(
+        engine: Engine,
+        store: BrowserStore,
+        onNewTabOverride: ((WebExtension?, EngineSession, String) -> Unit)? = null
+    ) {
+        engine.registerWebExtensionDelegate(object : WebExtensionDelegate {
+            override fun onNewTab(webExtension: WebExtension?, url: String, engineSession: EngineSession) {
+                onNewTabOverride?.invoke(webExtension, engineSession, url)
+                        ?: store.dispatch(TabListAction.AddTabAction(createTab(url)))
```
pocmo: `onNewTabOverride` override exists so that you do not need to deal with `SessionManager` here? What's the long term goal for this; will we remove this at some point?

components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
csadilek: @pocmo yes I was also thinking about that and going back and forth on `feature-webextensions`. I think I'd like to land this to unblock other work and get back to it in the coming weeks.

csadilek: Main reason I decided against introducing a separate `feature` is that if we ultimately (after the migration) end up adding the store to the engine, then we don't need this anymore.

components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
csadilek: @pocmo yes, exactly. I will add a comment to clarify.

components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
pocmo: I was thinking that the code that will add browser actions will have to live in that feature eventually too. :)
But, yeah, definitely doesn't need to block landing this.

csadilek:
bors r=pocmo, Amejia481

csadilek:
bors r=pocmo

csadilek:
bors r=pocmo,Amejia481

csadilek:
bors r=pocmo,Amejia481

csadilek:
bors retry

csadilek:
bors r=pocmo

csadilek:
bors r=pocmo

csadilek:
bors retry

PR #4697, opened 2019-10-09
Update glean_parser version to 1.9.3

brizental:
The new version adds functionaly to show labels from Labeled types on the Extras column in the Markdown template.


Dexterp37:
bors=mdboom,dexterp37
bors r+

PR #4701, opened 2019-10-09
Closes #4378 - Refactor AbstractFetchDownloadService

### LINKED ISSUE #4378: Refactor AbstractFetchDownloadService to not use LocalBroadcastManager
We are using `LocalBroadcastManager` in `AbstractFetchDownloadService`. `LocalBroadcastManager` has been deprecated:

> LocalBroadcastManager is an application-wide event bus and embraces layer violations in your app: any component may listen events from any other. You can replace usage of LocalBroadcastManager with other implementation of observable pattern, depending on your usecase suitable options may be LiveData or reactive streams. 

In our case we already have the `DownloadsFeature` and could potentially extend that to support notifying the app of completed downloads.

R4md4c:
---
I've removed the dependency on `LocalBroadcastManager` and added a field in the `BrowserState` to maintain the completedDownloads ids. 

When the download completes I dispatch an action through the `BrowserStore` which has a subscription in `FetchDownloadManager` which in turn invokes `onDownloadCompleted`.

I'm not sure if the modeling of the `BrowserState` is optimal or not, I'm open for better ideas however.

EDIT: Added more explanation about what I have done. 

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
@@ -308,7 +308,22 @@ sealed class EngineAction : BrowserAction() {
  */
 sealed class DownloadAction : BrowserAction() {
     /**
-     * Adds the [downloadId] into the set of completed downloads inside the [BrowserState].
+     * Adds the [download] to the queued downloads action.
      */
-    data class DownloadCompletedAction(val downloadId: Long) : DownloadAction()
+    data class AddDownloadAction(val download: DownloadState): DownloadAction()
+
+    /**
+     * Marks the [download] as started.
+     */
+    data class DownloadStartedAction(val download: DownloadState): DownloadAction()
+
+    /**
+     * Marks the [download] as completed.
+     */
+    data class DownloadCompletedAction(val download: DownloadState) : DownloadAction()
+
+    /**
+     * Marks the [download] as failure
+     */
+    data class DownloadFailedAction(val download: DownloadState): DownloadAction()
 }
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
@@ -302,3 +302,13 @@ sealed class EngineAction : BrowserAction() {
         val engineSessionState: EngineSessionState
     ) : EngineAction()
 }
+
+/**
+ * [BrowserAction] implementations relating to update the completed DownloadState inside [BrowserState]
+ */
+sealed class DownloadAction : BrowserAction() {
+    /**
+     * Adds the [downloadId] into the set of completed downloads inside the [BrowserState].
+     */
+    data class DownloadCompletedAction(val downloadId: Long) : DownloadAction()
+}
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/DownloadStateReducer.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/DownloadStateReducer.kt
@@ -1,0 +1,19 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.state.reducer
+
+import mozilla.components.browser.state.action.DownloadAction
+import mozilla.components.browser.state.state.BrowserState
+
+internal object DownloadStateReducer {
+
+    /**
+     * [DownloadAction] Reducer function for modifying [BrowserState.completedDownloads].
+     */
+    fun reduce(state: BrowserState, action: DownloadAction): BrowserState = when (action) {
+        is DownloadAction.DownloadCompletedAction ->
+            state.copy(completedDownloads = state.completedDownloads + action.downloadId)
+    }
+}
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/DownloadStateReducer.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/DownloadStateReducer.kt
@@ -6,14 +6,43 @@ package mozilla.components.browser.state.reducer
 
 import mozilla.components.browser.state.action.DownloadAction
 import mozilla.components.browser.state.state.BrowserState
+import mozilla.components.browser.state.state.content.DownloadState
+import mozilla.components.browser.state.state.content.DownloadStatus
 
 internal object DownloadStateReducer {
 
     /**
-     * [DownloadAction] Reducer function for modifying [BrowserState.completedDownloads].
+     * [DownloadAction] Reducer function for modifying [BrowserState.queuedDownloads].
      */
     fun reduce(state: BrowserState, action: DownloadAction): BrowserState = when (action) {
-        is DownloadAction.DownloadCompletedAction ->
-            state.copy(completedDownloads = state.completedDownloads + action.downloadId)
+        is DownloadAction.DownloadStartedAction -> {
+            state.changeDownloadStatusOf(downloadState = action.download, newState = DownloadStatus.RUNNING)
+        }
+        is DownloadAction.AddDownloadAction -> {
+            state.changeDownloadStatusOf(downloadState = action.download, newState = DownloadStatus.QUEUED)
+        }
+        is DownloadAction.DownloadCompletedAction -> {
+            state.changeDownloadStatusOf(downloadState = action.download, newState = DownloadStatus.COMPLETED)
+        }
+        is DownloadAction.DownloadFailedAction -> {
+            state.changeDownloadStatusOf(downloadState = action.download, newState = DownloadStatus.FAILED)
+        }
     }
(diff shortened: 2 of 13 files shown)


components/browser/state/src/main/java/mozilla/components/browser/state/state/BrowserState.kt
```diff
@@ -11,11 +11,14 @@ import mozilla.components.lib.state.State
  *
  * @property tabs the list of open tabs, defaults to an empty list.
  * @property selectedTabId the ID of the currently selected (active) tab.
+ * @property completedDownloads a set of ids of the completed downloads.
  * @property customTabs the list of custom tabs, defaults to an empty list.
  */
 data class BrowserState(
     val tabs: List<TabSessionState> = emptyList(),
     val selectedTabId: String? = null,
```
rocketsroger: nit: remove extra lines

rocketsroger: nit: remove extra empty lines

components/browser/state/src/main/java/mozilla/components/browser/state/state/BrowserState.kt
```diff
@@ -11,11 +11,14 @@ import mozilla.components.lib.state.State
...
 
+    val completedDownloads: Set<Long> = emptySet(),
+
```
rocketsroger: nit: same as above

components/browser/state/src/main/java/mozilla/components/browser/state/state/BrowserState.kt
```diff
@@ -11,11 +11,14 @@ import mozilla.components.lib.state.State
...
  * @property tabs the list of open tabs, defaults to an empty list.
  * @property selectedTabId the ID of the currently selected (active) tab.
+ * @property completedDownloads a set of ids of the completed downloads.
  * @property customTabs the list of custom tabs, defaults to an empty list.
  */
 data class BrowserState(
     val tabs: List<TabSessionState> = emptyList(),
     val selectedTabId: String? = null,
 
+    val completedDownloads: Set<Long> = emptySet(),
```
rocketsroger: Is completed download a browser wide value or a tab specific value?  Will we ever need the tab to show that its downloading/finished downloaded?  This might require more conversation.

components/browser/state/src/main/java/mozilla/components/browser/state/state/BrowserState.kt
rocketsroger: nit: not sure why we had a empty line here.  Let's remove this line and also the extra empty line you added.

components/browser/state/src/test/java/mozilla/components/browser/state/action/DownloadActionTest.kt
```diff
@@ -0,0 +1,25 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.state.action
+
+import mozilla.components.browser.state.store.BrowserStore
+import mozilla.components.support.test.ext.joinBlocking
+import org.junit.Assert
```
rocketsroger: nit: import org.junit.Assert.assertEquals will make the test (arguable marginal) easier to read.

components/browser/state/src/main/java/mozilla/components/browser/state/state/BrowserState.kt
R4md4c: The current behavior in `AndroidDownloadManager` and `AbstractFetchDownloadService` doesn't seem to contain any tab information, so I think it should go to the BrowserState, however I might be not seeing the big picture, so I'll leave the judgement to people who has more knowledge than me in the code base. 
Let me know what do you think please.

rocketsroger: I agree, this one will need @pocmo to think about.

R4md4c:
Hello @pocmo , Any input please regarding @rocketsroger's comment?

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/manager/FetchDownloadManager.kt
```diff
@@ -81,29 +84,31 @@ class FetchDownloadManager<T : AbstractFetchDownloadService>(
...
+     * Finds the completed download in the browser state, then removes it
      */
-    override fun onReceive(context: Context, intent: Intent) {
+    private fun onReceiveCompletedDownloads(completedDownloads: Set<Long>) {
         if (queuedDownloads.isEmpty()) unregisterListeners()
 
-        val downloadID = intent.getLongExtra(EXTRA_DOWNLOAD_ID, -1)
+        val downloadID = completedDownloads
+                .find { queuedDownloads.indexOfKey(it) >= 0 } ?: return
+
```
pocmo: We process a download id from the set here but we never remove it from the set. I think we should add a "consume" action to remove a processed id from the store again.

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -302,3 +302,13 @@ sealed class EngineAction : BrowserAction() {
...
 }
+
+/**
+ * [BrowserAction] implementations relating to update the completed DownloadState inside [BrowserState]
+ */
+sealed class DownloadAction : BrowserAction() {
+    /**
+     * Adds the [downloadId] into the set of completed downloads inside the [BrowserState].
+     */
+    data class DownloadCompletedAction(val downloadId: Long) : DownloadAction()
```
pocmo: I wonder whether we could put the whole `Download` object into this action. But I assume this does not work in the case where we pass the download to Android's Download Manager (except if we keep the download object in the download manager wrapper?).

R4md4c:
@csadilek 
I think that's a good idea, but then IMO we would need to introduce actions like `DownloadStarted` and `DownloadFinished` to be able to maintain the queued download, right?

My only concern is that do we need to introduce that change in a "Refactor AbstractFetchDownloadService" ticket?

pocmo:
@csadilek I like that idea!

@R4md4c Maybe we could model this with an enum state inside the `Download` object? That would be quite nice in general if an app wants to builder UI based on the state of the Download (without needing to look at various places inside the state object).

R4md4c:
@pocmo @csadilek Can you please take a look on my approach of the `BrowserState` and `DownloadState`? in [1] and [2] respectively?

I've implemented that statuses taking into account that the users of the `BrowserStore` may display a UI of the current `DownloadState`s in the store.

Please note that the last commit is missing tests and other stuff, I thought about committing and validating my approach first before going further with the complete implementation.

[1] https://github.com/R4md4c/android-components/blob/f2e6fcbcecc8bbf7547672d1cc935f052eddb853/components/browser/state/src/main/java/mozilla/components/browser/state/state/BrowserState.kt#L24

[2] https://github.com/R4md4c/android-components/blob/f2e6fcbcecc8bbf7547672d1cc935f052eddb853/components/browser/state/src/main/java/mozilla/components/browser/state/state/content/DownloadState.kt#L22

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/DownloadStateReducer.kt
```diff
@@ -0,0 +1,48 @@
...
+        is DownloadAction.AddDownloadAction -> {
+            state.changeDownloadStatusOf(downloadState = action.download, newState = DownloadStatus.QUEUED)
+        }
+        is DownloadAction.DownloadCompletedAction -> {
+            state.changeDownloadStatusOf(downloadState = action.download, newState = DownloadStatus.COMPLETED)
+        }
+        is DownloadAction.DownloadFailedAction -> {
+            state.changeDownloadStatusOf(downloadState = action.download, newState = DownloadStatus.FAILED)
+        }
+    }
```
pocmo: I still think we also need an action to "consume" a download here once we no longer need it in the state?

components/browser/state/src/main/java/mozilla/components/browser/state/state/BrowserState.kt
```diff
@@ -19,6 +21,7 @@ import mozilla.components.lib.state.State
 data class BrowserState(
     val tabs: List<TabSessionState> = emptyList(),
     val selectedTabId: String? = null,
+    val queuedDownloads: List<DownloadState> = emptyList(),
```
pocmo: Now that we keep all state in here... can this be moved back to the individual tabs? Afaik now we do not need to have this as a global state anymore?

R4md4c:
@pocmo To be able to continue working on this ticket, I'd like to have some questions answered.

1. You mentioned that `queuedDownloads` needs to be per `TabSessionState`, however there is `downloadState` which is inside `ContentState`, would it still be needed if we've added queuedDownloads to the `TabSessionState`? or better put the `queuedDownloads` inside the `ContentState`?
2. Since we've introduced the DownloadAction, what is the purpose then of `ContentAction.UpdateDownloadAction` and `ContentAction.ConsumeDownloadAction` and would it make sense to keep them since DownloadActions can be used to provide the same reduction to the `ContentState`?

PR #4716, opened 2019-10-11
Fix snapshots: (publish.gradle) Do not override "timestamp" property.

pocmo:
This broke our snapshots since `getTimestamp()` overrides the `timestamp` property that our scripts pass in when building snapshots.


pocmo:
bors r=Amejia481,JohanLorenzo

PR #4731, opened 2019-10-14
Update jacoco to version 0.8.5.

pocmo:
A new version of jacoco was released and there's at least one improvement for Kotlin code:
> Branches added by the Kotlin compiler for functions with default arguments and containing arguments of type long or double should be filtered out during generation of report (GitHub #908).

Let's see if this works in automation and how that affects our coverage.


pocmo:
bors r=csadilek

PR #4739, opened 2019-10-15
Closes #3720: Incorrect selection notification when session removed

### LINKED ISSUE #3720: SessionManager.remove() with the selected session doesn't notify Observer onSessionSelected
recalculateSelectionIndex is supposed to return true if the selection was updated, however it performs the check by seeing if the new selection index is the same as before.  
In the case were we delete the selected session, i.e. at index=0. both the newly selected index and the previously selected index will be 0 (since recalculateSelectionIndex selects a new index only if we removed a session to the left of the selectedIndex).

```kotlin
var newSelectedIndex = when {
//.... cases for selectParentIfExists or no more sessions
        indexToRemove < selectedIndex -> selectedIndex - 1
        selectedIndex == values.size -> selectedIndex - 1
        else -> selectedIndex
}
//....
val selectionUpdated = newSelectedIndex != selectedIndex
```
Which in turn makes remove() avoid notifying the observers:
```kotlin
if (selectionUpdated && selectedIndex != NO_SELECTION) {
    notifyObservers { onSessionSelected(selectedSessionOrThrow) }
}
```
This can be solved by changing recalculateSelectionIndex
```kotlin
val selectionUpdated = newSelectedIndex != selectedIndex
```  
To:  
```kotlin
val selectionUpdated = indexToRemove == selectedIndex ||
     newSelectedIndex != selectedIndex
```
That is, the selection was updated if the selected index was updated OR if the item removed was stored in the selected index.

csadilek:
This fixes two issues: 

- The issue described in #3720 where removing the selected session does not cause a `onSessionSelected` notification

- Sending an unnecessary `onSessionSelected` when removing a session to the left of the selected session. In that case the index changed but it's still the same session that's selected and observers should not need to get notified that a session was selected

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/LegacySessionManager.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/LegacySessionManager.kt
@@ -309,15 +309,17 @@ class LegacySessionManager(
             return
         }
 
+        val selectedBeforeRemove = selectedSession
+
         values.removeAt(indexToRemove)
 
         unlink(session)
 
-        val selectionUpdated = recalculateSelectionIndex(
-                indexToRemove,
-                selectParentIfExists,
-                session.private,
-                session.parentId
+        recalculateSelectionIndex(
+            indexToRemove,
+            selectParentIfExists,
+            session.private,
+            session.parentId
         )
 
         values.filter { it.parentId == session.id }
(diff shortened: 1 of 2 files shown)


csadilek:
Thanks for the review @rocketsroger , I'd like have @pocmo double-check as well before landing :).

csadilek:
bors r=pocmo,rocketsroger

PR #4740, opened 2019-10-15
For #4758: Remove min API of M for dataprotect

ekager:
---

Trying to remove the min SDK of 23 dependency to use these for Fenix so looking for feedback because I don't know much about the history here :)

### Pull Request checklist

### After merge

### DIFF
--- a/components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/Keystore.kt
+++ b/components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/Keystore.kt
@@ -5,37 +5,80 @@
 package mozilla.components.lib.dataprotect
 
 import android.annotation.TargetApi
+import android.content.Context
 import android.os.Build
+import android.os.Build.VERSION_CODES.M
 import android.security.keystore.KeyGenParameterSpec
 import android.security.keystore.KeyProperties
+import java.io.FileNotFoundException
+import java.math.BigInteger
 import java.security.GeneralSecurityException
 import java.security.InvalidKeyException
 import java.security.Key
+import java.security.KeyPair
+import java.security.KeyPairGenerator
 import java.security.KeyStore
+import java.security.KeyStoreException
+import java.security.NoSuchAlgorithmException
+import java.security.PrivateKey
+import java.security.SecureRandom
+import java.security.UnrecoverableKeyException
+import java.util.Calendar
 import javax.crypto.Cipher
 import javax.crypto.KeyGenerator
 import javax.crypto.SecretKey
 import javax.crypto.spec.GCMParameterSpec
+import javax.security.auth.x500.X500Principal
 
-private const val KEYSTORE_TYPE = "AndroidKeyStore"
+private const val PROVIDER_ANDROID = "AndroidKeyStore"
 private const val ENCRYPTED_VERSION = 0x02
+internal const val PROVIDER_BOUNCY_CASTLE = "BC"
 
-internal const val CIPHER_ALG = KeyProperties.KEY_ALGORITHM_AES
-internal const val CIPHER_MOD = KeyProperties.BLOCK_MODE_GCM
-internal const val CIPHER_PAD = KeyProperties.ENCRYPTION_PADDING_NONE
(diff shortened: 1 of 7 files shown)


components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/Keystore.kt
```diff
@@ -82,15 +94,24 @@ open class KeyStoreWrapper {
...
-        val gen = KeyGenerator.getInstance(CIPHER_ALG, KEYSTORE_TYPE)
-        gen.init(spec)
-        return gen.generateKey()
+            val gen = KeyGenerator.getInstance(CIPHER_ALG, KEYSTORE_TYPE)
+            gen.init(spec)
+            return gen.generateKey()
+        } else {
+            val spec = SecureRandom()
+            val gen = KeyGenerator.getInstance(CIPHER_ALG, KEYSTORE_TYPE)
+            gen.init(spec)
```
grigoryk: You probably want to use `init (AlgorithmParameterSpec params, SecureRandom random)` and specify key size, block mode, encryption padding.

ekager:
@linuxwolf from your experience building this do you have any recommendations on how I could test if this does function on older devices?

linuxwolf:
> @linuxwolf from your experience building this do you have any recommendations on how I could test if this does function on older devices?

I think it should be testable in a simulator running an old enough version of Android, using the [dataprotect sample app](https://github.com/mozilla-mobile/android-components/tree/master/samples/dataprotect).  I don't know offhand what changes will need to be made to get it built for an older version of Android, unfortunately.

grigoryk:
@linuxwolf what about using `RSA/ECB/NoPadding` or `RSA/ECB/PKCS1Padding`? These are API18+. And we'd use `AES` for newer devices. While I'm not sure how RSA compares to AES, it's probably better than nothing?

linuxwolf:
> @linuxwolf what about using `RSA/ECB/NoPadding` or `RSA/ECB/PKCS1Padding`? These are API18+. And we'd use `AES` for newer devices. While I'm not sure how RSA compares to AES, it's probably better than nothing?

RSA has very hard limits on content, so is not appropriate for arbitrary data.  It's very possible to use `RSA/ECB/PKCS1Padding` as a KEK (key encryption key) the protects the real CEK (content encryption key).  It'll take some care to not break newer and more efficient platforms, though.

@ekager and I have a meeting on Monday to discuss.

ekager:
You're right that this won't currently work on 21/22. Is the ugly workaround something like the gist mentioned here? 
https://github.com/google/tink/issues/226#issue-460242558

linuxwolf:
> You're right that this won't currently work on 21/22.
😞 
> Is the ugly workaround something like the gist mentioned here?
> [google/tink#226 (comment)](https://github.com/google/tink/issues/226#issue-460242558)

That gist is just about it, yes.  I think it can be slightly simplified (e.g., don't need X509 `CN` or `Serial`), but still about that shape.

ekager:
Wanted to get this working before we chatted Monday 😁 

Using the sample, this works locally for me on API 22 and API 29. I tried to maintain as much of the original API so we wouldn't impact downstream consumers (aka Lockwise?), but let's chat more Monday if we can clean this up a bit more or there are any major concerns about changing things in this way.

ekager:
So I was looking into key-loss issues and was asking around in ASG and apparently even though this would "work" on <23 there are major reported crashes and the official Google response to those bugs is just "don't use KeyStore for <23". Sooooo we should probably just ditch this approach and go with something less "good" but less "crashy" like storing a generated key in shared preferences.

> This is one of the issues we saw in the wild after releasing code that uses KeyStore on API < 23 (we were not using setEncryptionRequired(true)): https://issuetracker.google.com/issues/37051017
> 
> The official response to that bug is Do not use the old keystore APIs.  They're broken and deprecated.  Use the new ones added in Marshmallow.. :dumpsterfire: https://issuetracker.google.com/issues/37051017#comment18

grigoryk:
That's unfortunate. Thank you for researching this @ekager !!

PR #4742, opened 2019-10-16
Closes #4151: Add enum for SitePermissions types

### LINKED ISSUE #4151: Add enum for SitePermissions types
Site permissions are currently defined using property names.
```kt
    val location: Status = NO_DECISION,
    val notification: Status = NO_DECISION,
    val microphone: Status = NO_DECISION,
    val camera: Status = NO_DECISION,
    val bluetooth: Status = NO_DECISION,
    val localStorage: Status = NO_DECISION,
```
This gets difficult to work with quickly when dealing with SitePermissionsRules (which also has named properties) or serializing the permissions to SharedPreferences.

Fenix has the `PhoneFeature` enum to deal with this but it would make more sense to uplift the functionality to Android Components.

We already have the `Permission` sealed class so methods to work with it could be exposed.

calvin-li:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissions.kt
+++ b/components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissions.kt
@@ -4,14 +4,15 @@
 
 package mozilla.components.feature.sitepermissions
 
-import android.os.Parcel
 import android.os.Parcelable
+import kotlinx.android.parcel.Parcelize
 import mozilla.components.feature.sitepermissions.SitePermissions.Status.NO_DECISION
-import mozilla.components.feature.sitepermissions.db.StatusConverter
+import mozilla.components.feature.sitepermissions.SitePermissionsStorage.Permission
 
 /**
  * A site permissions and its state.
  */
+@Parcelize
 data class SitePermissions(
     val origin: String,
     val location: Status = NO_DECISION,
@@ -22,19 +23,6 @@ data class SitePermissions(
     val localStorage: Status = NO_DECISION,
     val savedAt: Long
 ) : Parcelable {
-
-    constructor(parcel: Parcel) :
-        this(
-            requireNotNull(parcel.readString()),
-            requireNotNull(converter.toStatus(parcel.readInt())),
-            requireNotNull(converter.toStatus(parcel.readInt())),
-            requireNotNull(converter.toStatus(parcel.readInt())),
-            requireNotNull(converter.toStatus(parcel.readInt())),
-            requireNotNull(converter.toStatus(parcel.readInt())),
-            requireNotNull(converter.toStatus(parcel.readInt())),
-            parcel.readLong()
-        )
-
     enum class Status(
         internal val id: Int
     ) {
@@ -50,30 +38,17 @@ data class SitePermissions(
         }
     }
 
-    override fun writeToParcel(parcel: Parcel, flags: Int) {
-        parcel.writeString(origin)
-        parcel.writeInt(converter.toInt(location))
-        parcel.writeInt(converter.toInt(notification))
-        parcel.writeInt(converter.toInt(microphone))
-        parcel.writeInt(converter.toInt(camera))
-        parcel.writeInt(converter.toInt(bluetooth))
-        parcel.writeInt(converter.toInt(localStorage))
-        parcel.writeLong(savedAt)
-    }
-
-    override fun describeContents(): Int {
-        return 0
-    }
-
-    companion object CREATOR : Parcelable.Creator<SitePermissions> {
-        override fun createFromParcel(parcel: Parcel): SitePermissions {
-            return SitePermissions(parcel)
+    /**
+     * Gets the current status for a [Permission] type
+     */
+    operator fun get(permissionType: Permission): Status {
+        return when (permissionType) {
+            Permission.MICROPHONE -> microphone
+            Permission.BLUETOOTH -> bluetooth
+            Permission.CAMERA -> camera
+            Permission.LOCAL_STORAGE -> localStorage
+            Permission.NOTIFICATION -> notification
+            Permission.LOCATION -> location
         }
-
-        override fun newArray(size: Int): Array<SitePermissions?> {
-            return arrayOfNulls(size)
-        }
-
-        private val converter = StatusConverter()
     }
 }
--- a/components/feature/sitepermissions/build.gradle
+++ b/components/feature/sitepermissions/build.gradle
@@ -5,6 +5,7 @@
 apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-kapt'
+apply plugin: 'kotlin-android-extensions'
 
 android {
     compileSdkVersion config.compileSdkVersion
--- a/components/feature/sitepermissions/src/test/java/mozilla/components/feature/sitepermissions/SitePermissionsTest.kt
+++ b/components/feature/sitepermissions/src/test/java/mozilla/components/feature/sitepermissions/SitePermissionsTest.kt
@@ -1,0 +1,47 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.feature.sitepermissions
+
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import mozilla.components.feature.sitepermissions.SitePermissions.Status.NO_DECISION
+import mozilla.components.feature.sitepermissions.SitePermissions.Status.ALLOWED
+import mozilla.components.feature.sitepermissions.SitePermissions.Status.BLOCKED
+import mozilla.components.feature.sitepermissions.SitePermissionsStorage.Permission.NOTIFICATION
+import mozilla.components.feature.sitepermissions.SitePermissionsStorage.Permission.LOCAL_STORAGE
+import mozilla.components.feature.sitepermissions.SitePermissionsStorage.Permission.LOCATION
+import mozilla.components.feature.sitepermissions.SitePermissionsStorage.Permission.MICROPHONE
+import mozilla.components.feature.sitepermissions.SitePermissionsStorage.Permission.BLUETOOTH
+import mozilla.components.feature.sitepermissions.SitePermissionsStorage.Permission.CAMERA
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class SitePermissionsTest {
+    @Test
+    fun `Tests get() against direct property access`() {
+        var sitePermissions = SitePermissions(origin = "mozilla.dev", savedAt = 0)
+
+        assertEquals(NO_DECISION, sitePermissions[NOTIFICATION])
+        assertEquals(NO_DECISION, sitePermissions[LOCATION])
+        assertEquals(NO_DECISION, sitePermissions[LOCAL_STORAGE])
+        assertEquals(NO_DECISION, sitePermissions[MICROPHONE])
+        assertEquals(NO_DECISION, sitePermissions[BLUETOOTH])
+        assertEquals(NO_DECISION, sitePermissions[CAMERA])
+
+        sitePermissions = sitePermissions.copy(
+            location = ALLOWED,
+            notification = BLOCKED,
+            microphone = NO_DECISION
+        )
+
+        assertEquals(BLOCKED, sitePermissions[NOTIFICATION])
+        assertEquals(ALLOWED, sitePermissions[LOCATION])
+        assertEquals(NO_DECISION, sitePermissions[MICROPHONE])
+        assertEquals(NO_DECISION, sitePermissions[BLUETOOTH])
+        assertEquals(NO_DECISION, sitePermissions[CAMERA])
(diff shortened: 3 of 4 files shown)


components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissions.kt
```diff
@@ -7,31 +7,64 @@ package mozilla.components.feature.sitepermissions
...
-    val notification: Status = NO_DECISION,
-    val microphone: Status = NO_DECISION,
-    val camera: Status = NO_DECISION,
-    val bluetooth: Status = NO_DECISION,
-    val localStorage: Status = NO_DECISION,
+    val permissions: Map<Permission, Status> = emptyMap(),
     val savedAt: Long
 ) : Parcelable {
 
+    // TODO: Remove, and replace all usages
```
NotWoods: Adding `@Deprecated` here is fine, so that you don't need to replace every single usage right now.

calvin-li: I tried that, but it won't compile because warnings are treated as errors.

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissions.kt
```diff
@@ -76,4 +109,21 @@ data class SitePermissions(
 
         private val converter = StatusConverter()
     }
+
+    internal operator fun get(permission: Permission): Status {
```
Amejia481: This can be useful externally (it can be marked as public).

Amejia481: This is similar to what Fenix is doing 
https://github.com/mozilla-mobile/fenix/blob/885e81699e0e945048c9aebe6a455aea4032ceac/app/src/main/java/org/mozilla/fenix/settings/PhoneFeature.kt#L103

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissions.kt
Amejia481: It happens because we are using this constructor internally.

https://github.com/mozilla-mobile/android-components/blob/0e3b8ad582c974993d4babdeee2230065c07e2cc/components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/db/SitePermissionsEntity.kt#L59

calvin-li: Ok I'll go ahead and deprecate this, then remove any internal usages. Just wanted to make sure I was headed in the right direction in general

pre-push
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+./gradlew --daemon ktlint detekt
```
Amejia481: Not sure if we want commit this file.

calvin-li: Yeah, removed

Amejia481:
It's looking good, please let's squash the commit. I will do a round of manual tests :)

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissions.kt
```diff
@@ -7,31 +7,78 @@ package mozilla.components.feature.sitepermissions
...
  */
 data class SitePermissions(
     val origin: String,
-    val location: Status = NO_DECISION,
-    val notification: Status = NO_DECISION,
-    val microphone: Status = NO_DECISION,
-    val camera: Status = NO_DECISION,
-    val bluetooth: Status = NO_DECISION,
-    val localStorage: Status = NO_DECISION,
+    val permissions: Map<Permission, Status> = emptyMap(),
```
Amejia481: I was thinking that we could find **shorter path** that lead of to **less code to maintain**,  **less code to modify** wand the same end result.

What about if instead of deprecating the constructor we just add a `get` operator  that relies on the existing properties? 

```kotlin
    /**
     * Gets the current status for a [Permission] type
     */
    operator fun get(permissionType: Permission): Status {
        return when(permissionType){
            MICROPHONE -> microphone
            BLUETOOTH -> bluetooth
            CAMERA -> camera
            LOCAL_STORAGE -> localStorage
            NOTIFICATION -> notification
            LOCATION -> location
        }
    }
```
This way we can do:

```kotlin
    val sitePermission = SitePermissions(
            origin = "mozilla.dev",
            savedAt = 0
    )
    sitePermission[NOTIFICATION]
```

Without having to maintain our own `equals` and `hashCode` functions as they come for free form the `data class` and at the same time we minimize the changes on the consumers end.

image


@calvin-li what do you think?

calvin-li: So, I've never used data classes before and didn't realize they had Equals() built-in. Looks like we could just use the default implementation then? I can try it later when I'm at home. 

We should already have the `site[NOTIFICATION]` behavior, if you're concerned about having to change usages in the rest of the code we can just not deprecate the old constructor?

Amejia481: > So, I've never used data classes before and didn't realize they had Equals() built-in. Looks like we could just use the default implementation then? I can try it later when I'm at home.

Yes, [these are the nice things about data classes](https://kotlinlang.org/docs/reference/data-classes.html) :)

> We should already have the site[NOTIFICATION] behavior, if you're concerned about having to change usages in the rest of the code we can just not deprecate the old constructor?

Yes, I think we could just leave the `SitePermissions` as it was before and add the `get` operator.

Amejia481:
I pushed some small changes that I think could simplify the existing api and reduce the changes on the consumer apps. @calvin-li @NotWoods what do you think?  @calvin-li  feel free to push the previous changes if you feel this ones are missing something.

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissions.kt
calvin-li: @NotWoods can clairfy, but I one of the reasons the issue was opened in the first place was the difficulty serializing the class. Adding `get` like this wouldn't help. This also means we would still have to update the `SitePermissions` class every time a new `Permission` type is added, which IMO is another good reason to use the enums.

Also, how would the current changes (bba4cce) make the api easier to use? Does it offer anything over leaving the class as-is, and is it enough to justify going through the testing/review process? Maybe we shouldn't be doing anything here at all.

Finally, in the future it would be nice to have these kinds of discussions on the first draft of the pull request, before all the work for testing, updating usages elsewhere, etc are already done

calvin-li:
> 
> 
> I pushed some small changes that I think could simplify the existing api and reduce the changes on the consumer apps. @calvin-li @NotWoods what do you think? @calvin-li feel free to push the previous changes back if you feel this ones are missing something.

I replied to your earlier review. We could also change `SitePermissions` to use `Map<Permission, Status>`, but also not deprecate the old constructor and just keep it in use if changing it everywhere is too much. Would also make the pull request much smaller. Either way, we should decide on this before moving forward with any new commits.

Amejia481:
It was very improper of me to push my commit without getting to an agreement first. Sorry for that. I brought back your commit.

I pushed my commit on my personal repo discuss the two approaches https://github.com/Amejia481/android-components/commit/3f6753f54911522744df10f72e8c3426c98ca070.

Amejia481:
After taking a closer look after manual testing, I'm not convince that we really need a map (Sorry I should spotted that earlier). If the goal it's just an easier way to access the properties on `SitePermissions` we could do it easy by adding an `get` operator like https://github.com/Amejia481/android-components/commit/3f6753f54911522744df10f72e8c3426c98ca070 with less modifications and less code to maintain. 

> This gets difficult to work with quickly when dealing with SitePermissionsRules (which also has named properties) or serializing the permissions to SharedPreferences.

@NotWoods  could you clarify what is the main goal?

calvin-li:
I don't mind you pushing your commit over mine; it's a better way to show the changes you're proposing than trying to describe them through text. What I do mind is the wasted time from doing the follow-up commits. Anyways, let's just figure out the best option here, and try to avoid this situation in the future.

Also, is using `get` really easier than just using the properties directly? It's `sitePermissions.notifications` vs `sitePermissions[NOTIFICATIONS]`, and you need an extra import statement for the second option, one for each permission type.

Amejia481:
> Also, is using get really easier than just using the properties directly? It's sitePermissions.notifications vs sitePermissions[NOTIFICATIONS], and you need an extra import statement for the second option, one for each permission type.

It could depends what you are trying to achieve. But it's also nice to have both alternatives. I would like to @NotWoods to help us to clarify the initial issue to better identify the best option.

NotWoods:
The goal is to be able to use getters with arbitrary preferences because we have a lot of switch maps/when blocks in Fenix's code. 

For example:
```kt
    private fun getStatus(sitePermissions: SitePermissions?): SitePermissions.Status? {
        sitePermissions ?: return null
        return when (this) {
            CAMERA -> sitePermissions.camera
            LOCATION -> sitePermissions.location
            MICROPHONE -> sitePermissions.microphone
            NOTIFICATION -> sitePermissions.notification
            AUTOPLAY -> SitePermissions.Status.NO_DECISION // No support from GV or A-C yet
        }
    }
```

Ideally this can be replaced with the new getter.

```kt
    private fun getStatus(sitePermissions: SitePermissions?): SitePermissions.Status? {
        return sitePermissions[this]
    }
```

Then we can start passing `Permission` around instead of `PhoneFeature` and move some of that code into Android Components.

Amejia481:
Sorry @calvin-li for all the back and forth, we really appreciate all your effort and time you put in this patch ❤️. I should noticed it since the beginning, it was completely my mistake, I'm really sorry for that, I will take it into consideration in the future. I think now we have to go with the simplest solution, as I believe is https://github.com/Amejia481/android-components/commit/3f6753f54911522744df10f72e8c3426c98ca070. I'm not sure if you want to bring this changes in or do you prefer that I do it?

calvin-li:
I'll add the changes to this PR. I think there are a couple of things we should add to your commit as well, particularly updating the serialize/deserialize code. I'll push a new commit today or tomorrow

Amejia481:
Thanks @calvin-li !

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissions.kt
```diff
@@ -51,20 +58,33 @@ data class SitePermissions(
...
     override fun writeToParcel(parcel: Parcel, flags: Int) {
+        Permission.values()
         parcel.writeString(origin)
-        parcel.writeInt(converter.toInt(location))
-        parcel.writeInt(converter.toInt(notification))
-        parcel.writeInt(converter.toInt(microphone))
-        parcel.writeInt(converter.toInt(camera))
-        parcel.writeInt(converter.toInt(bluetooth))
-        parcel.writeInt(converter.toInt(localStorage))
+        Permission.values().forEach {
```
Amejia481: Relying on the position of the constants could not be a good idea as changes on the `enum` could cause silent bugs.
We can completely outsource the parcel work to  the [Kotlin @Parcelize annotation](https://kotlinlang.org/docs/tutorials/android-plugin.html#parcelable-implementations-generator) and remove all the custom implementations :) . You can find an usage of  `@Parcelize` annotation in the source [here](https://github.com/mozilla-mobile/android-components/blob/75d6f13cfe436b8a9397141e8f9898e9553a373e/components/lib/crash/src/main/java/mozilla/components/lib/crash/Breadcrumb.kt#L15).

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissions.kt
```diff
@@ -51,20 +58,33 @@ data class SitePermissions(
     }
 
     override fun writeToParcel(parcel: Parcel, flags: Int) {
+        Permission.values()
```
Amejia481: nit: I this was checked in by mistake.

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissions.kt
calvin-li: Yeah, `@Parcelize` makes the code much simpler. I've updated the branch with it

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsStorage.kt
```diff
@@ -150,7 +150,8 @@ class SitePermissionsStorage(
         }
     }
 
+    @Suppress("MagicNumber")
```
Amejia481: I don't think the annotation is needed any more. As we don't need the reordering I think the could exclude this file changes.

Amejia481:
Great job @calvin-li 🥇!

One last thing before merging. Let's squash the commits and improve the commit name to something more meaningful to the actual change.   

For example:
`Add an get operator to the SitePermissions class.`

calvin-li:
Ok, everything should be ready for merging now

Amejia481:
bors r=Amejia481

Amejia481:
bors retry

PR #4750, opened 2019-10-16
Update glean_parser version to 1.9.4

brizental:
Because the new version changes Labeled types labes from sets to lists, I also needed to make some changes to the glean component's code. Nothing major, but since I am new to the project I am not sure that this doesn't break anything.

Tests pass and build is successfull, but I think it would be nice if someone with more experience in the project took a closer look :)

---

### Pull Request checklist

### After merge


components/service/glean/src/main/java/mozilla/components/service/glean/private/LabeledMetricType.kt
```diff
@@ -29,7 +29,7 @@ data class LabeledMetricType<T>(
     override val name: String,
     override val sendInPings: List<String>,
     val subMetric: T,
-    val labels: Set<String>? = null
+    val labels: List<String>? = null
```
Dexterp37: This is changing the metric types themselves and has some implications. Let's leave it as it is :)

components/service/glean/src/test/java/mozilla/components/service/glean/private/LabeledMetricTypeTest.kt
```diff
@@ -123,7 +123,7 @@ class LabeledMetricTypeTest {
             name = "labeled_counter_metric",
             sendInPings = listOf("metrics"),
             subMetric = counterMetric,
-            labels = setOf("foo", "bar", "baz")
+            labels = listOf("foo", "bar", "baz")
```
Dexterp37: Same, let's remove this change.

components/service/glean/src/test/java/mozilla/components/service/glean/private/LabeledMetricTypeTest.kt
```diff
@@ -577,7 +577,7 @@ class LabeledMetricTypeTest {
...
```
Dexterp37: Same, let's remove this change.

components/service/glean/src/main/java/mozilla/components/service/glean/private/LabeledMetricType.kt
brizental: The problem is, after que glean_parser change, if I don't change this, it breaks.

Dexterp37: That's weird, since we did not change that part in [mozilla/glean](https://github.com/mozilla/glean/blob/d27e4ce99e230cb3f8d744e6770540d4d45dda67/glean-core/android/src/main/java/mozilla/telemetry/glean/private/LabeledMetricType.kt#L30) and it works just fine. How is it breaking? What's the error?

brizental: Yes, I also thought is was weird... I will investigate and come back to it.

brizental: This is the error:
```
> Task :browser-engine-gecko-nightly:compileDebugKotlin FAILED
e: /Users/beatrizmachado/Documents/android-components/components/browser/engine-gecko-nightly/build/generated/source/glean/debug/kotlin/GfxContentFrameTime.kt: (126, 9): Type inference failed: constructor LabeledMetricType<T>(disabled: Boolean, category: String, lifetime: Lifetime, name: String, sendInPings: List<String>, subMetric: T, labels: Set<String>? = ...)
cannot be applied to
(String,String,CounterMetricType,Boolean,Lifetime,List<String>,List<String>)

e: /Users/beatrizmachado/Documents/android-components/components/browser/engine-gecko-nightly/build/generated/source/glean/debug/kotlin/GfxContentFrameTime.kt: (133, 22): Type inference failed. Expected type mismatch: inferred type is List<String> but Set<String>? was expected

```

Dexterp37: Interesting. Looks like mozilla/glean_parser#126 had the unexpected consequence of changing the generated code as well. We should fix that, cut a new release of glean_parser and then come back at this, I'm afraid.

brizental: Just to check that I understood this correctly. I want the new changes I added to glean_parser to only affect the markdown template and not the rest of the generated code. Is that right?

Dexterp37: Exactly. We still want the label order to be the same, but we don't want to generate different code.

brizental: Okay, that would required that I change the solution I did before. I can't simply turn [this](https://github.com/mozilla/glean_parser/blob/2e2d34f4e3a4eac4d217112658a04a02a589d0d6/glean_parser/metrics.py#L306) back into a set, because the moment we do that we loose the order... After giving this a little thought, here are some possible approches:

1. Undo that change and just order alphabetically in the md template. That fixes the random unnecessary changes in the md template, but doesn't attend to one of our needs that is keeping the order from `metrics.yaml`.
2. Turn it into a set when generating the code, but that is just bad and not sustainable IMO.
3. Add a new `ordered_labels` prop to the `Labeled` class that is a list, and keep the `labels` as a set, thus keeping both functionalities we want. 

I think I should go with 3, but I wanted to make sure that it makes sense or if you know a better way to do it.

mdboom: Yeah -- 3 seems tractable to me.  :+1:

brizental: I am going to close this PR in favor of the next one then.

PR #4759, opened 2019-10-16
Download updates

sblatz:
---

### Pull Request checklist

### After merge


sblatz:
Opening this up to review. I have a lot of "TODO" comments that are mostly questions I have for whoever reviews :)

components/browser/state/src/main/java/mozilla/components/browser/state/state/content/DownloadState.kt
```diff
@@ -27,5 +29,5 @@ data class DownloadState(
     val destinationDirectory: String = Environment.DIRECTORY_DOWNLOADS,
     val referrerUrl: String? = null,
     val skipConfirmation: Boolean = false,
-    val id: String = UUID.randomUUID().toString()
+    val id: Long = Random.nextLong()
```
NotWoods: Why long instead of UUID?

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -54,55 +62,175 @@ abstract class AbstractFetchDownloadService : CoroutineService() {
...
-        )
-        super.onCreate()
+    private var listOfDownloadJobs = mutableMapOf<Long, DownloadJobState>()
+
+    // TODO: Can this be simplified/refactored?
+    data class DownloadJobState(
+        var job: Job? = null,
+        var state: DownloadState,
+        var currentBytesCopied: Long = 0,
+        var isPaused: Boolean = false,
```
NotWoods: Maybe use an enum for ACTIVE/PAUSED/CANCELLED

sblatz: This is a great idea :)

components/feature/downloads/src/main/res/layout/mozac_downloads_prompt.xml
```diff
@@ -0,0 +1,86 @@
+<RelativeLayout
```
NotWoods: I think we should use ConstraintLayout and avoid RelativeLayout

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/SimpleDownloadDialogFragment.kt
```diff
@@ -3,16 +3,32 @@
...
-import mozilla.components.feature.downloads.R.string.mozac_feature_downloads_dialog_cancel
+import androidx.core.content.ContextCompat
+import kotlinx.android.synthetic.main.mozac_downloads_prompt.*
+import kotlinx.android.synthetic.main.mozac_downloads_prompt.view.*
+import kotlinx.android.synthetic.main.mozac_downloads_prompt.view.download_button
 import mozilla.components.feature.downloads.R.string.mozac_feature_downloads_dialog_download
-import mozilla.components.feature.downloads.R.string.mozac_feature_downloads_dialog_title
+import mozilla.components.feature.downloads.R.string.mozac_feature_downloads_dialog_title2
+
+private const val KEY_POSITIVE_BUTTON_BACKGROUND_COLOR = "KEY_POSITIVE_BUTTON_BACKGROUND_COLOR"
```
NotWoods: I'd prefer to stick these in the companion object since that already exists

components/browser/state/src/main/java/mozilla/components/browser/state/state/content/DownloadState.kt
sblatz: Fixed :)

components/feature/downloads/src/main/res/layout/mozac_downloads_prompt.xml
sblatz: This was mostly pattern-matched from `mozac_site_permissions_prompt.xml` which uses a RelativeLayout. May I ask why it's important to use ConstraintLayout instead? 

If we change this one to that we should likely change both :)

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/ext/DownloadState.kt
```diff
@@ -35,6 +35,11 @@ internal fun DownloadState.withResponse(headers: Headers, stream: InputStream?):
         contentType = headers[CONTENT_TYPE]
     }
 
+    var contentLength = this.contentLength
```
NotWoods: This appears to duplicate line 50: `contentLength ?: headers[CONTENT_LENGTH]?.toLongOrNull()`

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/SimpleDownloadDialogFragment.kt
```diff
@@ -63,41 +138,55 @@ class SimpleDownloadDialogFragment : DownloadDialogFragment() {
...
 
+                putInt(KEY_TITLE_TEXT, dialogTitleText)
+
                 putBoolean(KEY_CANCELABLE, cancelable)
+
+                promptsStyling?.apply {
+                    putInt(KEY_DIALOG_GRAVITY, gravity)
+                    putBoolean(KEY_DIALOG_WIDTH_MATCH_PARENT, shouldWidthMatchParent)
+
+                    positiveButtonBackgroundColor?.apply {
```
NotWoods: nit: use `let` since you're using the value

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt
```diff
@@ -192,4 +203,13 @@ class DownloadsFeature(
         val download = state.content.download ?: return
         block(Pair(state, download))
     }
+
+    data class PromptsStyling(
+        val gravity: Int,
+        val shouldWidthMatchParent: Boolean = false,
+        @ColorRes
+        val positiveButtonBackgroundColor: Int? = null,
```
NotWoods: Maybe use `DEFAULT_VALUE` so you don't need to deal with null

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadNotification.kt
```diff
@@ -118,4 +152,48 @@ internal object DownloadNotification {
...
+        ).build()
+    }
+
+    private fun createPendingIntent(context: Context, action: String, downloadStateId: Long): PendingIntent {
+        val intent = Intent(action)
+        intent.setPackage(context.applicationContext.packageName)
+
+        val bundleExtra = Bundle()
+        bundleExtra.putLong(EXTRA_DOWNLOAD_ID, downloadStateId)
+        intent.putExtras(bundleExtra)
```
NotWoods: nit: Use `putLongExtra` for a one-liner
```suggestion
        intent.putLongExtra(EXTRA_DOWNLOAD_ID, downloadStateId)
```

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadNotification.kt
```diff
@@ -14,37 +14,69 @@ import android.content.Intent
...
 internal object DownloadNotification {
 
     private const val NOTIFICATION_CHANNEL_ID = "Downloads"
+    internal const val EXTRA_DOWNLOAD_ID = "downloadId"
 
     /**
      * Build the notification to be displayed while the download service is active.
      */
-    fun createOngoingDownloadNotification(context: Context): Notification {
+    fun createOngoingDownloadNotification(context: Context, downloadState: DownloadState?): Notification {
```
NotWoods: Does it make sense to show a notification for null downloads? 🤔

components/feature/downloads/src/main/res/layout/mozac_downloads_prompt.xml
sblatz: Yeah, it doesn't look like anything in AC uses ConstraintLayout, so I'd prefer to stick to RelativeLayout for now and do that as a holistic transition in the future :)

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadNotification.kt
sblatz: Done!

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/ext/DownloadState.kt
sblatz: Fixed :)

sblatz:
I have done all the cleanup I believe is necessary and addressed @NotWoods's comments. 

Am I able to get another look on this? @NotWoods / @csadilek :)

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -18,89 +22,250 @@ import android.os.ParcelFileDescriptor
...
+@Suppress("TooManyFunctions", "LargeClass")
+abstract class AbstractFetchDownloadService : Service() {
 
     protected abstract val httpClient: Client
     @VisibleForTesting
     internal val broadcastManager by lazy { LocalBroadcastManager.getInstance(this) }
     @VisibleForTesting
     internal val context: Context get() = this
 
-    override fun onCreate() {
```
NotWoods: `onCreate` and `startForeground` are needed for Android to count this as a foreground service.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -18,89 +22,250 @@ import android.os.ParcelFileDescriptor
...
-        startForeground(
-            NotificationIds.getIdForTag(context, ONGOING_DOWNLOAD_NOTIFICATION_TAG),
-            DownloadNotification.createOngoingDownloadNotification(context)
-        )
-        super.onCreate()
+    private var listOfDownloadJobs = mutableMapOf<Long, DownloadJobState>()
```
NotWoods: nit: `mapOfDownloadJobs` or just `downloadJobs`

samples/browser/src/main/AndroidManifest.xml
```diff
@@ -84,6 +84,16 @@
         </service>
 
         <service android:name=".downloads.DownloadService" />
+        <provider
```
NotWoods: Perhaps the provider should be set in the component instead? See feature-prompts for an example.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -18,89 +22,250 @@ import android.os.ParcelFileDescriptor
...
+
+    private data class DownloadJobState(
+        var job: Job? = null,
```
NotWoods: Maybe use immutable data classes and `.copy` instead?

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -18,89 +22,250 @@ import android.os.ParcelFileDescriptor
...
+
+    override fun onDestroy() {
+        super.onDestroy()
+        listOfDownloadJobs.values.forEach {
+            it.job?.cancel()
         }
+    }
+
+    private fun startDownloadJob(download: DownloadState, isResuming: Boolean): Job {
+        return CoroutineScope(IO).launch {
```
NotWoods: If you have a scope for the service, you could just `cancel` that rather than having to loop through jobs in `onDestroy`

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -18,89 +22,250 @@ import android.os.ParcelFileDescriptor
...
+    }
+
+    override fun onDestroy() {
+        super.onDestroy()
+        listOfDownloadJobs.values.forEach {
+            it.job?.cancel()
         }
+    }
+
+    private fun startDownloadJob(download: DownloadState, isResuming: Boolean): Job {
```
NotWoods: I think it would be nicer if this was a suspend function and `.launch` was put inside `onStartCommand`.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt
```diff
@@ -46,6 +47,7 @@ import mozilla.components.support.ktx.kotlinx.coroutines.flow.ifChanged
  * @property useCases [DownloadsUseCases] instance for consuming processed downloads.
  * @property fragmentManager a reference to a [FragmentManager]. If a fragment
  * manager is provided, a dialog will be shown before every download.
+ * @property: TODO
```
NotWoods: nit

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt
```diff
@@ -58,8 +60,11 @@ class DownloadsFeature(
     private val downloadManager: DownloadManager = AndroidDownloadManager(applicationContext),
     private val customTabId: String? = null,
     private val fragmentManager: FragmentManager? = null,
+    private val promptsStyling: PromptsStyling? = null,
     @VisibleForTesting(otherwise = PRIVATE)
-    internal var dialog: DownloadDialogFragment = SimpleDownloadDialogFragment.newInstance()
+internal var dialog: DownloadDialogFragment = SimpleDownloadDialogFragment.newInstance(
```
NotWoods: nit: indentation

NotWoods: Also, can this be a `val`?

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
sblatz: Good point. Better to make it clear there's a launch happening on the IO thread :)

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt
sblatz: It cannot be due to the way `override fun start()` handles it a few lines beneath this.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
sblatz: Where should I cancel, then?

samples/browser/src/main/AndroidManifest.xml
sblatz: I've tried changing this and it doesn't seem to merge properly into the sample-browser manifest.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
sblatz: I'm confused what you mean by this. It seems to work properly without these functions being overriden. Can you point me to where you found that? :)

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
sblatz: While I could see the advantage of this, I don't think it's terribly necessary and would require a somewhat heavy refactor of this code. I feel like it's cleaner to have sections like `onReceive` which directly manipulate the reference in the map which I wouldn't be able to do if I had to use a .copy() (I would instead have to constantly reassign `downloadJobs[downloadId]` to a new copy, which gets a lot messier, at least when I tried to do this).

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
NotWoods:
```kt
class Service {
  private val scope = CoroutineScope(IO)

  fun onStartCommand() {
    scope.launch { startDownloadJob() }
  }

  fun onDestroy() {
    scope.cancel()
  }
}

File: components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt (hunk 10 above)
Comment (NotWoods) [reply components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt]: Without `startForeground`, the service is counted as a background service as is more likely to get killed by the system if it needs memory. https://developer.android.com/reference/android/app/Service

Using `Context.startForegroundService` enforces that `startForeground` is called, but I opted not to use it so that the fallback is being a background service rather than crashing.
https://medium.com/pongploydev/why-start-the-service-on-android-o-and-encounter-a-problem-record-context-startforegroundservice-3c8ddd8969ef

Comment (sblatz) [reply components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt]: I am using `startForeground`, though. Just not immediately. Is this really a problem?

Comment (NotWoods) [reply components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt]: Oops I haven't found it, can you link to where it's called? 😅

File: components/browser/state/src/main/java/mozilla/components/browser/state/state/content/DownloadState.kt [hunk 18]
```diff
@@ -16,16 +16,20 @@ import java.util.UUID
...
  * @property destinationDirectory The matching destination directory for this type of download.
+ * @property filePath The file path the file was saved at.
  * @property referrerUrl The site that linked to this download.
+ * @property skipConfirmation Whether or not the confirmation dialog should be shown before the download begins.
+ * @property id The unique identifier of this download.
  */
 data class DownloadState(
     val url: String,
-    val fileName: String? = null,
+    var fileName: String? = null,
```
NotWoods: nit: stuff in `BrowserState` should be immutable

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -18,91 +22,255 @@ import android.os.ParcelFileDescriptor
...
+
+                        currentDownloadJobState.job = CoroutineScope(IO).launch {
+                            startDownloadJob(currentDownloadJobState.state)
+                        }
+
+                        currentDownloadJobState.status = DownloadJobStatus.ACTIVE
+                    }
+
+                    ACTION_CANCEL -> {
+                        currentDownloadJobState.status = DownloadJobStatus.CANCELLED
```
csadilek: nit: not sure if it's a problem, but it looks suspicious, that we some times set the status before we apply the change and some times (other actions) after?

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -18,91 +22,255 @@ import android.os.ParcelFileDescriptor
...
-    private suspend fun performDownload(download: DownloadState) = withContext(IO) {
-        val headers = listOf(
-            CONTENT_TYPE to download.contentType,
-            CONTENT_LENGTH to download.contentLength?.toString(),
-            REFERRER to download.referrerUrl
-        ).mapNotNull { (name, value) ->
-            if (value.isNullOrBlank()) null else Header(name, value)
-        }.toMutableHeaders()
+    internal fun performDownload(download: DownloadState) {
+        val isResumingDownload = downloadJobs[download.id]?.currentBytesCopied ?: 0L > 0L
```
csadilek: Let's turn this expression into a function because we also need it below to set the header e.g. `getCurrentBytesCopied`.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -18,91 +22,255 @@ import android.os.ParcelFileDescriptor
...
 
+        val request = Request(download.url, headers = headers)
         val response = httpClient.fetch(request)
 
+        if (response.status != PARTIAL_CONTENT_STATUS && response.status != OK_STATUS) {
+            // We experienced a problem trying to fetch the file, send a failure notification
+            downloadJobs[download.id]?.currentBytesCopied = 0
+            downloadJobs[download.id]?.status = DownloadJobStatus.FAILED
+            return
+        }
```
csadilek: One thing I am worried about here is that a server responds with 200 OK but doesn't actually respect the byte range. I believe in your testing you mentioned this worked? Just not sure that's always the case. Should we double-check by making sure the response contains the `Content-Range` header?

components/feature/downloads/src/test/java/mozilla/components/feature/downloads/SimpleDownloadDialogFragmentTest.kt
```diff
@@ -45,22 +51,59 @@ class SimpleDownloadDialogFragmentTest {
...
 
-        dialog.onStartDownload = onStartDownload
-        dialog.testingContext = testContext
+        val fragment = Mockito.spy(SimpleDownloadDialogFragment.newInstance())
 
-        performClick(BUTTON_POSITIVE)
+        fragment.onStartDownload = onStartDownload
+        fragment.testingContext = testContext
+
+        Mockito.doReturn(testContext).`when`(fragment).requireContext()
```
csadilek: nit: we usually import statically to avoid `Mockito.doReturn`...

components/browser/state/src/main/java/mozilla/components/browser/state/state/content/DownloadState.kt
```diff
@@ -16,16 +16,20 @@ import java.util.UUID
...
     val contentType: String? = null,
     val contentLength: Long? = null,
     val userAgent: String? = null,
     val destinationDirectory: String = Environment.DIRECTORY_DOWNLOADS,
+    var filePath: String? = null,
     val referrerUrl: String? = null,
     val skipConfirmation: Boolean = false,
-    val id: String = UUID.randomUUID().toString()
+    val id: Long = Random.nextLong()
```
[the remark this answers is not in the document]
NotWoods: The native downloader uses Long ids so we kept Longs to match. I initially had @sblatz switch to strings and found that it required bigger changes to our downloading API.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
NotWoods: +1, I think we should check for `Content-Range` to ensure that range is observed.

components/browser/state/src/main/java/mozilla/components/browser/state/state/content/DownloadState.kt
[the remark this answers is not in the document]
csadilek: Ok, makes sense!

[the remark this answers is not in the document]
sblatz: Yep, basically what @NotWoods said :)

components/browser/state/src/main/java/mozilla/components/browser/state/state/content/DownloadState.kt
```diff
@@ -16,16 +16,20 @@ import java.util.UUID
...
  */
 data class DownloadState(
     val url: String,
-    val fileName: String? = null,
+    var fileName: String? = null,
     val contentType: String? = null,
     val contentLength: Long? = null,
     val userAgent: String? = null,
     val destinationDirectory: String = Environment.DIRECTORY_DOWNLOADS,
+    var filePath: String? = null,
```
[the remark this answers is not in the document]
sblatz: Updated for both of these variables.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
sblatz: I'm fairly confident the order does not matter. That being said, I made them all consistent so there's no confusion surrounding it.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
sblatz: Added a check :)

components/feature/downloads/src/test/java/mozilla/components/feature/downloads/SimpleDownloadDialogFragmentTest.kt
sblatz: Done!

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
csadilek: Yeah, should have marked it as `nit` maybe. It's mostly to simply the expression for the header value: `headers.append(RANGE,"bytes=${downloadJobs[download.id]?.currentBytesCopied}-")`

sblatz:
This is ready for review again. I added tests, addressed all of your comments, and added a "try again" asset per Bram's request.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadDialogFragment.kt
```diff
@@ -53,5 +54,11 @@ abstract class DownloadDialogFragment : DialogFragment() {
         const val KEY_URL = "KEY_URL"
 
         const val FRAGMENT_TAG = "SHOULD_DOWNLOAD_PROMPT_DIALOG"
+
+        const val MEGABYTE = 1024.0 * 1024.0
     }
 }
+
+internal fun Long.toMegabyteString(): String {
```
csadilek: This looks unit testable :).

components/browser/state/src/main/java/mozilla/components/browser/state/state/content/DownloadState.kt
```diff
@@ -16,16 +16,22 @@ import java.util.UUID
...
+@Suppress("Deprecation")
 data class DownloadState(
     val url: String,
     val fileName: String? = null,
     val contentType: String? = null,
     val contentLength: Long? = null,
     val userAgent: String? = null,
     val destinationDirectory: String = Environment.DIRECTORY_DOWNLOADS,
+    val filePath: String =
+        Environment.getExternalStoragePublicDirectory(destinationDirectory).path + "/" + fileName,
```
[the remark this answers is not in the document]
sblatz: I am intending on doing a fast-follow PR with more features to add. I will add this change then so we can start testing this in Fenix at least :)

sblatz:
bors r=@csadilek

PR #4760, opened 2019-10-16
Closes #2923: Remove moment.js from reader view web extension

### LINKED ISSUE #2923: Remove moment.js from reader view web extension
This is currently being used to generate a human-readable and localized string for the reading time e.g. 3 - 5 minutes.

We could use the minified version of this script, but we'd still be better off not needing it. We should likely pass the localized constants down from the feature using messaging, but the String is dependent on the language of the article currently.

csadilek:
Removes moment.js from reader view, and uses [Intl.RelativeTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RelativeTimeFormat) instead to construct a localized reading time string.

Tested this with various sites in Spanish, Arabic, German, and English to verify this works as before. We could improve this more later to detect the language if the article doesn't specify but that's a separate issue (which also doesn't work in Fennec currently).


components/feature/readerview/src/main/assets/extensions/readerview/readerview.js
```diff
@@ -191,15 +191,20 @@ class ReaderView {
...
-
-    if (readingTimeMinsSlow == readingTimeMinsFast) {
-      return moment.duration(readingTimeMinsFast, 'minutes').locale(lang).humanize();
+    // Construct a localized and "humanized" reading time in minutes. If we
+    // have both a fast and slow reading time we'll show both e.g.
+    // "2 - 4 minutes", otherwise we'll just show "4 minutes".
+    var parts = new Intl.RelativeTimeFormat(lang).formatToParts(readingTimeMinsSlow, 'minute');
+    if (parts.length == 3) {
+      // No need to use part[0] which represent "in".
+      var readingTime = `${parts[1].value} ${parts[2].value}`
```
Amejia481: nit: For more clarify, instead of using directly the array's position, we could save the values in variables that give a meaningful names to its content.

csadilek: 👍 I can clean this up some more.

csadilek: Done.

csadilek:
bors r=Amejia481

Issue #4549, opened 2019-09-26
Create abstraction for WebNotificationDelegate for the Engine


jonalmeida:
To fit our engine abstraction model, we need to implement an abstraction for WebNotifications where we can implement the `GeckoWebNotificationDelegate`.

rocketsroger:
Confirmed abstraction is needed for `WebNotificationDelegate`.  Will fix at the same time as #2295

jonalmeida:
> Confirmed abstraction is needed for `WebNotificationDelegate`. Will fix at the same time as #2295

Could you help me understand why we need to expose an abstraction on the Engine? Couldn't we just add `runtime.registerWebNotificationDelegate` in the `init` block of the GeckoEngine?

Having the abstraction on the engine would mean that apps can register their own delegate as well, is that intended?

Issue #4628, opened 2019-10-03
Create permission storage support for WebNotifications and WebPush


jonalmeida:
This would probably be a single storage permission we would need to add for both.

We should have a look at `PromptFeature` to see how it stores permissions in the `SitePermissionsFeature` and (re)-use those APIs as much as possible.

jonalmeida:
Looks like we don't need to do this since the permissions feature already handles this. 🎉

PR #4768, opened 2019-10-17
Closes #2295, #4549: Implement Web Notification Feature

### LINKED ISSUE #2295: Implement GeckoWebNotificationDelegate

rocketsroger:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/webnotifications/GeckoWebNotificationDelegate.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/webnotifications/GeckoWebNotificationDelegate.kt
@@ -1,0 +1,26 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.engine.gecko.webnotifications
+
+import mozilla.components.concept.engine.webnotifications.WebNotification
+import mozilla.components.concept.engine.webnotifications.WebNotificationDelegate
+import org.mozilla.geckoview.WebNotification as GeckoViewWebNotification
+import org.mozilla.geckoview.WebNotificationDelegate as GeckoViewWebNotificationDelegate
+
+internal class GeckoWebNotificationDelegate(
+    private val webNotificationDelegate: WebNotificationDelegate
+) : GeckoViewWebNotificationDelegate {
+    override fun onShowNotification(webNotification: GeckoViewWebNotification) {
+        webNotificationDelegate.onShowNotification(webNotification.toWebNotification())
+    }
+
+    override fun onCloseNotification(webNotification: GeckoViewWebNotification) {
+        webNotificationDelegate.onCloseNotification(webNotification.toWebNotification())
+    }
+
+    private fun GeckoViewWebNotification.toWebNotification(): WebNotification {
+        return WebNotification(title, tag, text, imageUrl, textDirection, lang, requireInteraction)
+    }
+}
--- a/components/concept/engine/src/main/java/mozilla/components/concept/engine/webnotifications/WebNotification.kt
+++ b/components/concept/engine/src/main/java/mozilla/components/concept/engine/webnotifications/WebNotification.kt
@@ -7,31 +7,23 @@ package mozilla.components.concept.engine.webnotifications
 /**
  * A notification sent by the Web Notifications API.
  *
- * @property origin The website that fired this notification.
  * @property title Title of the notification to be displayed in the first row.
- * @property body Body of the notification to be displayed in the second row.
  * @property tag Tag used to identify the notification.
- * @property iconUrl Medium image to display in the notification.
+ * @property body Body of the notification to be displayed in the second row.
+ * @property iconUrl Large icon url to display in the notification.
  * Corresponds to [android.app.Notification.Builder.setLargeIcon].
- * @property vibrate Vibration pattern felt when the notification is displayed.
+ * @property direction Preference for text direction.
+ * @property lang language of the notification.
+ * @property requireInteraction Preference flag that indicates the notification should remain.
  * @property timestamp Time when the notification was created.
- * @property requireInteraction Preference flag that indicates the notification should remain
- * active until the user clicks or dismisses it.
- * @property silent Preference flag that indicates no sounds or vibrations should be made.
- * @property onClick Callback called with the selected action, or null if the main body of the
- * notification was clicked.
- * @property onClose Callback called when the notification is dismissed.
  */
 data class WebNotification(
-    val origin: String,
-    val title: String? = null,
-    val body: String? = null,
-    val tag: String? = null,
-    val iconUrl: String? = null,
-    val vibrate: LongArray = longArrayOf(),
-    val timestamp: Long? = null,
-    val requireInteraction: Boolean = false,
-    val silent: Boolean = false,
-    val onClick: () -> Unit,
-    val onClose: () -> Unit
+    val title: String?,
+    val tag: String,
+    val body: String?,
+    val iconUrl: String?,
+    val direction: String?,
+    val lang: String?,
+    val requireInteraction: Boolean,
+    val timestamp: Long = System.currentTimeMillis()
 )
--- a/components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/NativeNotificationBridge.kt
+++ b/components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/NativeNotificationBridge.kt
@@ -4,12 +4,16 @@
 
 package mozilla.components.feature.webnotifications
 
+import android.app.Activity
 import android.app.Notification
+import android.app.PendingIntent
 import android.content.Context
+import android.content.Intent
 import android.graphics.Bitmap
 import android.net.Uri
 import android.os.Build
 import android.os.Build.VERSION.SDK_INT
+import androidx.annotation.DrawableRes
 import androidx.core.net.toUri
 import mozilla.components.browser.icons.BrowserIcons
 import mozilla.components.browser.icons.Icon.Source
@@ -18,16 +22,23 @@ import mozilla.components.browser.icons.IconRequest.Size
 import mozilla.components.concept.engine.webnotifications.WebNotification
 
 internal class NativeNotificationBridge(
-    private val icons: BrowserIcons
+    private val icons: BrowserIcons,
+    @DrawableRes private val smallIcon: Int
 ) {
+    companion object {
+        private const val EXTRA_ON_CLICK = "mozac.feature.webnotifications.generic.onclick"
+    }
 
     /**
      * Create a system [Notification] from this [WebNotification].
      */
+    @Suppress("LongParameterList")
     suspend fun convertToAndroidNotification(
         notification: WebNotification,
         context: Context,
-        channelId: String
+        channelId: String,
+        activityClass: Class<out Activity>?,
+        requestId: Int
     ): Notification {
         val builder = if (SDK_INT >= Build.VERSION_CODES.O) {
             Notification.Builder(context, channelId)
@@ -37,22 +48,25 @@ internal class NativeNotificationBridge(
         }
 
         with(notification) {
-            loadIcon(iconUrl?.toUri(), origin, Size.DEFAULT)?.let { icon ->
-                builder.setLargeIcon(icon)
+            activityClass?.let {
+                val intent = Intent(context, activityClass).apply {
+                    putExtra(EXTRA_ON_CLICK, tag)
+                }
+
+                PendingIntent.getActivity(context, requestId, intent, 0).apply {
+                    builder.setContentIntent(this)
+                }
             }
 
-            builder.setContentTitle(title).setContentText(body)
+            builder.setSmallIcon(smallIcon)
+                .setContentTitle(title)
+                .setContentText(body)
+                .setShowWhen(true)
+                .setWhen(timestamp)
+                .setAutoCancel(true)
 
-            @Suppress("Deprecation")
-            builder.setVibrate(vibrate)
-
-            timestamp?.let {
-                builder.setShowWhen(true).setWhen(it)
-            }
-
-            if (silent) {
-                @Suppress("Deprecation")
-                builder.setDefaults(0)
+            loadIcon(iconUrl?.toUri(), Size.DEFAULT)?.let { iconBitmap ->
+                builder.setLargeIcon(iconBitmap)
             }
         }
 
@@ -62,10 +76,10 @@ internal class NativeNotificationBridge(
     /**
      * Load an icon for a notification.
      */
-    private suspend fun loadIcon(url: Uri?, origin: String, size: Size): Bitmap? {
+    private suspend fun loadIcon(url: Uri?, size: Size): Bitmap? {
         url ?: return null
         val icon = icons.loadIcon(IconRequest(
-            url = origin,
+            url = url.toString(),
             size = size,
             resources = listOf(IconRequest.Resource(
                 url = url.toString(),
--- a/components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
+++ b/components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
@@ -1,0 +1,102 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.feature.webnotifications
+
+import android.app.Activity
+import android.app.NotificationChannel
+import android.app.NotificationManager
+import android.content.Context
+import android.os.Build
+import androidx.annotation.DrawableRes
+import androidx.core.app.NotificationCompat
+import androidx.core.content.getSystemService
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
+import mozilla.components.browser.icons.BrowserIcons
+import mozilla.components.concept.engine.Engine
+import mozilla.components.concept.engine.webnotifications.WebNotification
+import mozilla.components.concept.engine.webnotifications.WebNotificationDelegate
+import mozilla.components.support.base.log.logger.Logger
+import java.lang.UnsupportedOperationException
+
+private const val NOTIFICATION_CHANNEL_ID = "mozac.feature.webnotifications.generic.channel"
+
+/**
+ * Feature implementation for configuring and displaying web notifications to the user.
+ *
+ * Initialize this feature globally once on app start
+ * ```Kotlin
+ * WebNotificationFeature(
+ *     applicationContext, engine, icons, R.mipmap.ic_launcher, BrowserActivity::class.java
+ * )
+ * ```
+ *
+ * @param context The application Context.
+ * @param engine The browser engine.
+ * @param browserIcons The entry point for loading the large icon for the notification.
+ * @param smallIcon The small icon for the notification.
+ * @param activityClass The Activity that the notification will launch if user taps on it
+ */
+class WebNotificationFeature(
+    private val context: Context,
+    private val engine: Engine,
+    private val browserIcons: BrowserIcons,
+    @DrawableRes private val smallIcon: Int,
+    private val activityClass: Class<out Activity>?
+) : WebNotificationDelegate {
+    private val logger = Logger("WebNotificationFeature")
+    private var pendingRequestId = 0
+    private var notificationId = 0
+    private val notificationIdMap = HashMap<String, Int>()
+    private val notificationManager = context.getSystemService<NotificationManager>()
+    private val nativeNotificationBridge = NativeNotificationBridge(browserIcons, smallIcon)
+
+    init {
+        try {
+            engine.registerWebNotificationDelegate(this)
+        } catch (e: UnsupportedOperationException) {
+            logger.error("failed to register for web notification delegate", e)
+        }
+    }
+
+    override fun onShowNotification(webNotification: WebNotification) {
+        ensureNotificationGroupAndChannelExists()
+
+        notificationIdMap[webNotification.tag]?.let {
+            notificationManager?.cancel(it)
+        }
+
+        pendingRequestId++
+        notificationId++
+        notificationIdMap[webNotification.tag] = notificationId
+
+        GlobalScope.launch(Dispatchers.IO) {
+            val notification = nativeNotificationBridge.convertToAndroidNotification(
+                webNotification, context, NOTIFICATION_CHANNEL_ID, activityClass, pendingRequestId)
+            notificationManager?.notify(notificationId, notification)
+        }
+    }
+
+    override fun onCloseNotification(webNotification: WebNotification) {
+        notificationIdMap[webNotification.tag]?.let {
+            notificationManager?.cancel(it)
+        }
+    }
+
+    private fun ensureNotificationGroupAndChannelExists() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            val channel = NotificationChannel(
+                NOTIFICATION_CHANNEL_ID,
+                context.getString(R.string.mozac_feature_notification_channel_name),
+                NotificationManager.IMPORTANCE_LOW
+            )
+            channel.setShowBadge(true)
+            channel.lockscreenVisibility = NotificationCompat.VISIBILITY_PRIVATE
+
+            notificationManager?.createNotificationChannel(channel)
+        }
+    }
+}
--- a/components/feature/webnotifications/src/main/res/values/strings.xml
+++ b/components/feature/webnotifications/src/main/res/values/strings.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this file,
+   - You can obtain one at http://mozilla.org/MPL/2.0/.  -->
+<resources>
+    <!-- Default Web Notification Channel Name. -->
+    <string name="mozac_feature_notification_channel_name">Site notifications</string>
+</resources>
--- a/samples/browser/src/main/java/org/mozilla/samples/browser/DefaultComponents.kt
+++ b/samples/browser/src/main/java/org/mozilla/samples/browser/DefaultComponents.kt
@@ -44,6 +44,7 @@ import mozilla.components.feature.search.SearchUseCases
 import mozilla.components.feature.session.HistoryDelegate
 import mozilla.components.feature.session.SessionUseCases
 import mozilla.components.feature.tabs.TabsUseCases
+import mozilla.components.feature.webnotifications.WebNotificationFeature
 import mozilla.components.lib.fetch.httpurlconnection.HttpURLConnectionClient
 import org.mozilla.samples.browser.integration.FindInPageIntegration
 import org.mozilla.samples.browser.request.SampleRequestInterceptor
@@ -101,6 +102,9 @@ open class DefaultComponents(private val applicationContext: Context) {
                 .enable()
 
             MediaStateMachine.start(this)
+
+            WebNotificationFeature(applicationContext, engine, icons, R.drawable.ic_notification,
+                BrowserActivity::class.java)
         }
     }
 
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
@@ -10,6 +10,7 @@ import mozilla.components.browser.engine.gecko.integration.LocaleSettingUpdater
 import mozilla.components.browser.engine.gecko.mediaquery.from
 import mozilla.components.browser.engine.gecko.mediaquery.toGeckoValue
 import mozilla.components.browser.engine.gecko.webextension.GeckoWebExtension
+import mozilla.components.browser.engine.gecko.webnotifications.GeckoWebNotificationDelegate
 import mozilla.components.concept.engine.Engine
 import mozilla.components.concept.engine.EngineSession
 import mozilla.components.concept.engine.EngineSession.TrackingProtectionPolicy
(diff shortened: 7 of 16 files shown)


components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
```diff
@@ -0,0 +1,103 @@
...
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
+import mozilla.components.browser.icons.BrowserIcons
+import mozilla.components.concept.engine.Engine
+import mozilla.components.concept.engine.webnotifications.WebNotification
+import mozilla.components.concept.engine.webnotifications.WebNotificationDelegate
+import mozilla.components.support.base.log.logger.Logger
+import java.lang.UnsupportedOperationException
+
+private const val NOTIFICATION_CHANNEL_ID = "mozac.feature.webnotifications.generic"
```
NotWoods: Other browsers create a channel per website and put them under a notification group. Should we follow the same model?

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
```diff
@@ -0,0 +1,103 @@
...
+
+    init {
+        try {
+            engine.registerWebNotificationDelegate(FeatureWebNotificationDelegate())
+        } catch (e: UnsupportedOperationException) {
+            logger.error("failed to register for web notification delegate", e)
+        }
+    }
+
+    private inner class FeatureWebNotificationDelegate : WebNotificationDelegate {
```
NotWoods: Since `WebNotificationFeature` doesn't have any methods, should it implement `WebNotificationDelegate` and contain all these functions instead of having an inner class?

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
```diff
@@ -0,0 +1,103 @@
...
+            val nativeNotificationBridge = NativeNotificationBridge(browserIcons)
+            val channelId = ensureChannelExists(context, notificationManager)
+
+            notificationIdMap[webNotification.tag]?.let {
+                notificationManager.cancel(it)
+            }
+
+            notificationIdMap[webNotification.tag] = notificationId
+            GlobalScope.launch(Dispatchers.IO) {
+                notificationManager.notify(notificationId++, nativeNotificationBridge.convertToAndroidNotification(
```
NotWoods: I'm not a fan of inline `++` because it's hard to remember if the return value is before or after the addition. I think it'd be better to store the desired IDs in a local variable separate from incrementing.

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
```diff
@@ -0,0 +1,103 @@
...
+                        webNotification, context, channelId, launcherIcon, activityClass, pendingRequestId++))
+            }
+        }
+
+        override fun onCloseNotification(webNotification: WebNotification) {
+            val notificationManager = context.getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
```
NotWoods: `notificationManager` could be made a property of the delegate instead of getting it every time.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/webnotifications/GeckoWebNotificationDelegate.kt
```diff
@@ -0,0 +1,29 @@
...
+    override fun onShowNotification(webNotification: GeckoViewWebNotification) {
+        webNotificationDelegate.onShowNotification(webNotification.toWebNotification())
+    }
+
+    override fun onCloseNotification(webNotification: GeckoViewWebNotification) {
+        webNotificationDelegate.onCloseNotification(webNotification.toWebNotification())
+    }
+
+    private fun GeckoViewWebNotification.toWebNotification(): WebNotification {
+        this.apply {
```
NotWoods: `this.apply` is unneeded since its an extension method

components/concept/engine/src/main/java/mozilla/components/concept/engine/webnotifications/WebNotification.kt
```diff
@@ -7,31 +7,23 @@ package mozilla.components.concept.engine.webnotifications
...
-    val vibrate: LongArray = longArrayOf(),
-    val timestamp: Long? = null,
-    val requireInteraction: Boolean = false,
-    val silent: Boolean = false,
-    val onClick: () -> Unit,
-    val onClose: () -> Unit
+    val title: String?,
+    val tag: String,
+    val text: String?,
+    val imageUrl: String?,
```
NotWoods: The [Notification spec](https://developer.mozilla.org/en-US/docs/Web/API/Notification/Notification) has both "image" and "icon". Icon refers to the medium image, but "image" refers to big picture style notifications.

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/NativeNotificationBridge.kt
```diff
@@ -37,22 +43,22 @@ internal class NativeNotificationBridge(
...
-            timestamp?.let {
-                builder.setShowWhen(true).setWhen(it)
+            loadIcon(imageUrl?.toUri(), Size.DEFAULT)?.let { iconBitmap ->
+                builder.setLargeIcon(iconBitmap)
             }
 
-            if (silent) {
-                @Suppress("Deprecation")
-                builder.setDefaults(0)
+            timestamp.let {
```
NotWoods: If `timestamp` is always set there's no need for the `let` block

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/NativeNotificationBridge.kt
```diff
@@ -17,17 +19,21 @@ import mozilla.components.browser.icons.IconRequest
 import mozilla.components.browser.icons.IconRequest.Size
 import mozilla.components.concept.engine.webnotifications.WebNotification
 
+private const val EXTRA_ON_CLICK = "onClick"
```
NotWoods: Extra should be namespaced and maybe on the companion object.

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
rocketsroger: I agree.  We'll need to do this when the push notification feature is implemented.  Currently GV doesn't give us which web site is notifying.  Once we have that information we can create notification groups with channels for each website.  This will be a design decision of push notification.

components/concept/engine/src/main/java/mozilla/components/concept/engine/webnotifications/WebNotification.kt
```diff
@@ -7,31 +7,23 @@ package mozilla.components.concept.engine.webnotifications
...
+    val textDirection: String?,
```
NotWoods: I think we should follow the notification spec for these properties and not GeckoView's specific fields, to be more generic. 
```suggestion
    val dir: String?,
```

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/NativeNotificationBridge.kt
```diff
@@ -37,22 +43,22 @@ internal class NativeNotificationBridge(
...
+                intent.putExtra(EXTRA_ON_CLICK, tag)
+                val pendingIntent = PendingIntent.getActivity(context, requestId, intent, 0)
+                builder.setContentIntent(pendingIntent)
             }
 
-            builder.setContentTitle(title).setContentText(body)
-
-            @Suppress("Deprecation")
-            builder.setVibrate(vibrate)
+            builder.setSmallIcon(launcherIcon).setContentTitle(title).setContentText(text)
```
NotWoods: The small icon should be a dedicated monochrome icon instead of just the launcher icon.

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/NativeNotificationBridge.kt
```diff
@@ -17,17 +19,21 @@ import mozilla.components.browser.icons.IconRequest
...
     /**
      * Create a system [Notification] from this [WebNotification].
      */
+    @Suppress("LongParameterList")
     suspend fun convertToAndroidNotification(
         notification: WebNotification,
         context: Context,
-        channelId: String
+        channelId: String,
+        launcherIcon: Int,
```
NotWoods:
```suggestion
        @DrawableRes launcherIcon: Int,
```

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
rocketsroger: Agreed.  I'll just extend `WebNotificationDelegate` with `WebNotificationFeature`

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
NotWoods: I that case, I think we should probably set up a notification group for now with this channel as the only entry.

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
rocketsroger: Sure, I'll move the increment to the end of the block.

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
rocketsroger: Agreed.  I'll make this a property.

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
```diff
@@ -0,0 +1,103 @@
...
+            context: Context,
+            notificationManager: NotificationManager
+        ): String {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                val channel = NotificationChannel(
+                        NOTIFICATION_CHANNEL_ID,
+                        context.getString(R.string.mozac_browser_errorpages_connection_failure_message),
+                        NotificationManager.IMPORTANCE_LOW
+                )
+                channel.setShowBadge(false)
```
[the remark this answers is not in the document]
rocketsroger: You're right.  Notification badge is probably wanted.  I'll change it to true.

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
```diff
@@ -0,0 +1,103 @@
...
+                channel.lockscreenVisibility = NotificationCompat.VISIBILITY_PUBLIC
```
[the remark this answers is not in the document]
rocketsroger: Agreed.  Shouldn't assume notification is not sensitive.  I'll change it to `PRIVATE`

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/webnotifications/GeckoWebNotificationDelegate.kt
rocketsroger: Good catch, forgot that I left that there.  Thanks,

components/concept/engine/src/main/java/mozilla/components/concept/engine/webnotifications/WebNotification.kt
rocketsroger: Yes, I agree.  But currently GV only gives us a imageURL for the icon.  No other information.
Here's the GV's `WebNotification` spec https://searchfox.org/mozilla-central/source/mobile/android/geckoview/src/main/java/org/mozilla/geckoview/WebNotification.java

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/NativeNotificationBridge.kt
rocketsroger: Oh yeah, I had timestamp as possible `null` but not anymore.  I'll fix.  Thanks,

components/concept/engine/src/main/java/mozilla/components/concept/engine/webnotifications/WebNotification.kt
rocketsroger: I see what you mean.  I agree, I should follow the spec not GV's naming.

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/NativeNotificationBridge.kt
rocketsroger: Yes, but we'll need to get `Notification.Badge` from GV down the road.  However, currently we don't have it.  So I added this to allow user of the feature to pass an icon in.  I'll rename the parameter to `smallIcon` so its more clear.

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/NativeNotificationBridge.kt
rocketsroger: Sure, I'll add this.  Thanks,

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
rocketsroger: Sure, I'll create a channel

components/concept/engine/src/main/java/mozilla/components/concept/engine/webnotifications/WebNotification.kt
NotWoods: I wasn't clear in my comment but I'd prefer to use `iconUrl` as the property name here. I previous discussed the GV version with Alvina and the intent is that it represents the "icon" field.

rocketsroger: I see.  Currently we don't have a icon URL from GV though.  We only get the image URL so that's the image when you expand the notification.  So my implementation can only allow the browser to specify the icon.  In the future when GV pass us the icon URL we can then update this class and use it.

NotWoods: GV's `imageUrl` is actually the `icon` field of a web notification. It represents the medium 48x48 icon rather than the expanded big picture image. GV doesn't give us the `image` field right now.

rocketsroger: I see. I was thinking the image that we put in the large icon is the image.  But looking into this further I see that there's another image (a much bigger image) that we can show in the notification.  Sure, I'll rename it to `iconUrl`  Thanks,

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/NativeNotificationBridge.kt
rocketsroger: Sure, I'll put it in a companion object

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
```diff
@@ -0,0 +1,105 @@
...
+    private val engine: Engine,
+    private val browserIcons: BrowserIcons,
+    @DrawableRes private val smallIcon: Int,
+    private val activityClass: Class<*>?
+) : WebNotificationDelegate {
+    private val logger = Logger("AutoPushFeature")
+    private var pendingRequestId = 0
+    private var notificationId = 0
+    private val notificationIdMap = HashMap<String, Int>()
+    private val notificationManager = context.getSystemService(Context.NOTIFICATION_SERVICE)
```
NotWoods: nit: There's an extension function for `getSystemService` in Android KTX that's shorter:
```kt
context.getSystemService<NotificationManager>()
```

components/feature/webnotifications/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this file,
+   - You can obtain one at http://mozilla.org/MPL/2.0/.  -->
+<resources>
+    <!-- Default Web Notification Group Name. -->
+    <string name="mozac_feature_notification_group_name">Sites</string>
```
rocketsroger: How do I get the official strings?

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
```diff
@@ -0,0 +1,105 @@
...
+        ensureNotificationGroupAndChannelExists()
+
+        notificationIdMap[webNotification.tag]?.let {
+            notificationManager.cancel(it)
+        }
+
+        notificationIdMap[webNotification.tag] = notificationId
+        GlobalScope.launch(Dispatchers.IO) {
+            notificationManager.notify(notificationId, nativeNotificationBridge.convertToAndroidNotification(
+                webNotification, context, NOTIFICATION_CHANNEL_ID, smallIcon, activityClass, pendingRequestId))
```
NotWoods: It might make sense to turn `NativeNotificationBridge` into a class property and move some constant data to its constructor (such as smallIcon and activityClass)

```
private val nativeNotificationBridge = NativeNotificationBridge(
  browserIcons,
  smallIcon,
  activityClass,
  ...
)

...

override fun onShowNotification(webNotification: WebNotification) {
  ...
  nativeNotificationBridge.convertToAndroidNotification(
                webNotification, context, NOTIFICATION_CHANNEL_ID, pendingRequestId))
}
```

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
rocketsroger: Nice, I'll update.

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
rocketsroger: Sure, sounds good.  Thanks,

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
jonalmeida: > Currently GV doesn't give us which web site is notifying.

Maybe this is something we should ask GV if they can provide. We can then group these site permissions together along with web push notifications or at least be able to categories them per website in site settings.

I would argue that a user doesn't know or care about the difference between a WebPush and WebNotification. We should delegate blocking per-site.

We can also punt on this for now and investigate it more with https://github.com/mozilla-mobile/android-components/issues/4628.

components/feature/webnotifications/src/main/res/values/strings.xml
jonalmeida: Let's punt on this for now since we may want to group site permissions together within the SitePermissionsFeature.

jonalmeida: @m.heubusch  would be the right person to ping though.

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
```diff
@@ -0,0 +1,104 @@
...
+
+    override fun onShowNotification(webNotification: WebNotification) {
+        ensureNotificationGroupAndChannelExists()
+
+        notificationIdMap[webNotification.tag]?.let {
+            notificationManager?.cancel(it)
+        }
+
+        notificationIdMap[webNotification.tag] = notificationId
+        GlobalScope.launch(Dispatchers.IO) {
```
jonalmeida: I was confused why the notify call needed to be in a coroutine. It probably reads better to not inline the `convertToAndroidNotification`.

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/NativeNotificationBridge.kt
```diff
@@ -18,16 +21,23 @@ import mozilla.components.browser.icons.IconRequest.Size
...
     /**
      * Create a system [Notification] from this [WebNotification].
      */
+    @Suppress("LongParameterList")
     suspend fun convertToAndroidNotification(
         notification: WebNotification,
         context: Context,
-        channelId: String
+        channelId: String,
+        activityClass: Class<*>?,
```
jonalmeida: This wildcard seems too permissive. Maybe `Class<out Activity>?`?

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
```diff
@@ -0,0 +1,104 @@
...
+
+    override fun onCloseNotification(webNotification: WebNotification) {
+        notificationIdMap[webNotification.tag]?.let {
+            notificationManager?.cancel(it)
+        }
+    }
+
+    private fun ensureNotificationGroupAndChannelExists() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            notificationManager?.createNotificationChannelGroup(NotificationChannelGroup(
```
jonalmeida: Mentioned above as well, we probably want the notification settings to be in site permissions, so we don't need a group here. We can put it at the top level like Send Tab does.

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/NativeNotificationBridge.kt
```diff
@@ -37,22 +47,18 @@ internal class NativeNotificationBridge(
         }
 
         with(notification) {
-            loadIcon(iconUrl?.toUri(), origin, Size.DEFAULT)?.let { icon ->
-                builder.setLargeIcon(icon)
+            activityClass?.let {
+                val intent = Intent(context, activityClass)
```
jonalmeida: Fun Kotlin tip: You can use `.apply { }` to do things on the instance before assignment.

```kotlin
val intent = Intent(context, activityClass).apply {
  putExtra(EXTRA_ON_CLICK, tag)
}
```

What you have here works perfectly well, but I thought I should share. :)

samples/browser/src/main/java/org/mozilla/samples/browser/DefaultComponents.kt
```diff
@@ -100,6 +101,9 @@ open class DefaultComponents(private val applicationContext: Context) {
             MediaFeature(applicationContext)
                 .enable()
 
+            WebNotificationFeature(applicationContext, engine, icons, R.mipmap.ic_launcher,
```
NotWoods: needs a transparent icon, ic_launcher for sample browser will probably be a solid circle.

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
```diff
@@ -0,0 +1,104 @@
...
+ * @param browserIcons The entry point for loading the large icon for the notification.
+ * @param smallIcon The small icon for the notification.
+ * @param activityClass The Activity that the notification will launch if user taps on it
+ */
+class WebNotificationFeature(
+    private val context: Context,
+    private val engine: Engine,
+    private val browserIcons: BrowserIcons,
+    @DrawableRes private val smallIcon: Int,
+    private val activityClass: Class<*>?
```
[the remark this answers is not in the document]
rocketsroger: Agreed.  Thanks,

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
```diff
@@ -0,0 +1,104 @@
...
+) : WebNotificationDelegate {
+    private val logger = Logger("AutoPushFeature")
```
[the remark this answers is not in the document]
rocketsroger: Good catch.  Thanks,

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/NativeNotificationBridge.kt
rocketsroger: Agreed.

components/feature/webnotifications/src/main/res/drawable/ic_test.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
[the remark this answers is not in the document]
rocketsroger: I've removed the icon.  Thanks for the advice.

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
rocketsroger: Sure, I'll do that.

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
rocketsroger: Ah, I'll remove the group.

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/NativeNotificationBridge.kt
rocketsroger: Agreed.

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/NativeNotificationBridge.kt
rocketsroger: I see.  I think I'll have to stick with what it is right now.  The main reason is activityClass could be null and if it is we wouldn't add any pendingIntent to the notification.

components/feature/webnotifications/src/main/res/values/strings.xml
rocketsroger: Created issue.  https://github.com/mozilla-mobile/fenix/issues/6303.

samples/browser/src/main/java/org/mozilla/samples/browser/DefaultComponents.kt
rocketsroger: Agreed.  Issue created https://github.com/mozilla-mobile/android-components/issues/4869

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/NativeNotificationBridge.kt
rocketsroger: Thanks for the offline explanation.  I will update this block.

samples/browser/src/main/java/org/mozilla/samples/browser/DefaultComponents.kt
rocketsroger: Updated with correct icon.

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
```diff
@@ -0,0 +1,102 @@
...
+    private var pendingRequestId = 0
+    private var notificationId = 0
+    private val notificationIdMap = HashMap<String, Int>()
+    private val notificationManager = context.getSystemService<NotificationManager>()
+    private val nativeNotificationBridge = NativeNotificationBridge(browserIcons, smallIcon)
+
+    init {
+        try {
+            engine.registerWebNotificationDelegate(this)
+        } catch (e: UnsupportedOperationException) {
```
jonalmeida: Nice!

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
```diff
@@ -0,0 +1,102 @@
...
+    private val context: Context,
+    private val engine: Engine,
+    private val browserIcons: BrowserIcons,
+    @DrawableRes private val smallIcon: Int,
+    private val activityClass: Class<out Activity>?
+) : WebNotificationDelegate {
+    private val logger = Logger("WebNotificationFeature")
+    private var pendingRequestId = 0
+    private var notificationId = 0
+    private val notificationIdMap = HashMap<String, Int>()
```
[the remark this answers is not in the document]
jonalmeida: Actually, let's fix this before landing since it can cause conflicts and crashes.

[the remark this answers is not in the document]
rocketsroger: That makes sense.  I'll use `NotificationIds`.  Thanks,

samples/browser/src/main/java/org/mozilla/samples/browser/DefaultComponents.kt
```diff
@@ -100,6 +101,9 @@ open class DefaultComponents(private val applicationContext: Context) {
             MediaFeature(applicationContext)
                 .enable()
 
+            WebNotificationFeature(applicationContext, engine, icons, R.drawable.ic_notification,
```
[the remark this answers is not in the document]
rocketsroger: Agreed.  I'll move it.  Thanks,

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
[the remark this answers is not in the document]
rocketsroger: After looking into it.  I think I'll have to update `NotificationIds` before I can use it this use case.  Created https://github.com/mozilla-mobile/android-components/issues/4882

rocketsroger:
bors r=[jonalmeida, NotWoods]

PR #4773, opened 2019-10-17
Closes #4749: Fix for bug 1522829

### LINKED ISSUE #4749: Implement fix for bug 1522829
See: https://bugzilla.mozilla.org/show_bug.cgi?id=1522829

There is a ticket of this in Fenix as well (https://github.com/mozilla-mobile/fenix/issues/5019) but most of the work will need to be done in A-C.

csadilek:
We basically already have the parent session everywhere we need it, we now just forward the corresponding engine session so we can pass it along to GeckoView e.g. when opening a new (regular or private) tab, and when reacting to window requests.

@pocmo one discussion point is whether or not we have to create an engine session for the parent if it's no longer there (e.g. when re-loading a child tab after a restart). I decided **not** to do that as we'd be loading all parent sessions on startup and the child session should already have the right state.

Also, I didn't want to overload the method as named args with default values make this easy enough to read in Kotlin.

### DIFF
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -96,9 +96,9 @@ class GeckoEngineSession(
     /**
      * See [EngineSession.loadUrl]
      */
-    override fun loadUrl(url: String, flags: LoadUrlFlags) {
+    override fun loadUrl(url: String, parent: EngineSession?, flags: LoadUrlFlags) {
         requestFromWebContent = false
-        geckoSession.loadUri(url, flags.value)
+        geckoSession.loadUri(url, (parent as? GeckoEngineSession)?.geckoSession, flags.value)
     }
 
     /**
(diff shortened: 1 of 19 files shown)


pocmo:
> @pocmo one discussion point is whether or not we have to create an engine session for the parent if it's no longer there (e.g. when re-loading a child tab after a restart). I decided not to do that as we'd be loading all parent sessions on startup [..]

I can't follow here. What do you mean with "we'd be loading all parent sessions on startup"? Since we only initialize sessions lazily on startup wouldn't we only need to load the session we render and its parent (and its parent ... up the tree, which should be short walk in reality).

> and the child session should already have the right state.

I'm not sure whether this is still okay for the internals of GeckoView. Do we need to ask them if that is required when restoring?

csadilek:
> Since we only initialize sessions lazily on startup wouldn't we only need to load the session we render and its parent (and its parent ... up the tree, which should be short walk in reality).

Yes, that's what I meant. Didn't mean all parents, just the parents of the sessions we're eventually loading. I have that delta in a branch.

> I'm not sure whether this is still okay for the internals of GeckoView. Do we need to ask them if that is required when restoring?

I will ask.

csadilek:
@pocmo 

> I'm not sure whether this is still okay for the internals of GeckoView.

OK, confirmed that we're good here. We don't need to specify the referring/parent session again after restoring and that information is part of the restored state.

csadilek:
bors r=pocmo

Issue #4261, opened 2019-08-30
Update copy and graphic on QR code scanning screen


ryanfeeley:
Our reticle screen still needs updating using the fancy attached graphic made by @shorlander. We've also dropped the https for legibility (as it redirects to https anyway)
[reticle.zip](https://github.com/mozilla-mobile/android-components/files/3560022/reticle.zip)

<img width="371" alt="image" src="https://user-images.githubusercontent.com/68213/64023616-b34e3480-cb06-11e9-9146-a1d0f473d516.png">

lime124:
@vesta0 no design work needed.

lime124:
strike that - i hadn't realize this was in the A/C backlog

drigio:
Hello @lime124 @ryanfeeley I would like to work on this if available.

NotWoods:
@drigio If you comment on mozilla-mobile/fenix#6198 I can assign that to you too.

sblatz:
In sprint planning we discussed that this is a lot of work due to testing on lots of different aspect ratios. The solution isn't hard, but that's why we're deferring for now.

shorlander:
> In sprint planning we discussed that this is a lot of work due to testing on lots of different aspect ratios. The solution isn't hard, but that's why we're deferring for now.

@sblatz Is there a reason this would be more prone to error than any other piece of UI or even the current implementation of the QR code scanner?

Right now it's centered and if you happen to change the aspect ratio it scales to fit.

The reason I ask is because currently it's not in a good state. You can hardly see the black reticle in many situations and the instructions easily get lost pinned to the top of the view. We are funneling anyone who wants to sign-in to Sync to this slightly broken experience.

liuche:
@NotWoods do you have any context on why this might be difficult? ^

mcarare:
@liuche : After spending some time to investigate this, IMO implementing this will have a degree of risk, as we have to revert changes made in https://github.com/mozilla-mobile/android-components/pull/5985.  Or we could keep them, but we would still have a square preview, instead of the full display one like in the comment 0.

liuche:
Interesting, how about we keep the "square" scan area from #5985, and overlay the black inside the square? I think the important part of this issue is that in the current QR scan implementation, the little window outline that shows where to align the QR code is hard to see.

Can you add the black alpha overlay from the mock to create the "scan window" but keep the square cut?

csadilek:
[...]

PR #4845, opened 2019-10-24
#4261 update graphic qr code

drigio:
- Added an overlay that allows dimming the background and the reticle comes into focus.
- Removed the previous reticle png images.
---

### Pull Request checklist

### After merge


components/feature/qr/src/main/java/mozilla/components/feature/qr/CustomOverlay.kt
```diff
@@ -0,0 +1,105 @@
...
+        const val ALPHA = 170
+    }
+
+    constructor(context: Context) : super(context)
+
+    constructor(context: Context, attrs: AttributeSet) : super(context, attrs) {
+        activity = context as Activity
+    }
+
+    constructor(context: Context, attrs: AttributeSet, defStyleAttr: Int) : super(context, attrs, defStyleAttr)
```
NotWoods: You can use `@JvmOverloads` so that you don't need to write 3 constructors: 

```kt
class CustomOverlay @JvmOverloads constructor(
    context: Context,
    attrs: AttributeSet? = null,
    defStyleAttr: Int = 0
) : LinearLayout(context, attrs, defStyleAttr)
```

components/feature/qr/src/main/java/mozilla/components/feature/qr/CustomOverlay.kt
```diff
@@ -0,0 +1,105 @@
...
+    override fun dispatchDraw(canvas: Canvas) {
+        super.dispatchDraw(canvas)
+
+        if (windowFrame == null) {
+            createWindowFrame()
+        }
+        canvas.drawBitmap(windowFrame!!, 0f, 0f, null)
+    }
+
+    override fun isEnabled(): Boolean { return false }
```
NotWoods: nit: Use `=` shorthand here
```suggestion
    override fun isEnabled() = false
```

components/feature/qr/src/main/java/mozilla/components/feature/qr/CustomOverlay.kt
```diff
@@ -0,0 +1,105 @@
...
+
+    override fun isClickable(): Boolean { return false }
```
NotWoods: ditto `=`

components/feature/qr/src/main/java/mozilla/components/feature/qr/CustomOverlay.kt
```diff
@@ -0,0 +1,105 @@
...
+     * This is called before the canvas is drawn. Initializes the windowFrame only once
+     * @param canvas The canvas to draw upon.
+     */
+    override fun dispatchDraw(canvas: Canvas) {
+        super.dispatchDraw(canvas)
+
+        if (windowFrame == null) {
+            createWindowFrame()
+        }
+        canvas.drawBitmap(windowFrame!!, 0f, 0f, null)
```
NotWoods: nit: You can simplify the null check by having `createWindowFrame` return the newly created windowFrame.

```suggestion
        canvas.drawBitmap(windowFrame ?: createWindowFrame(), 0f, 0f, null)
```

components/feature/qr/src/main/java/mozilla/components/feature/qr/CustomOverlay.kt
```diff
@@ -0,0 +1,105 @@
...
+        val osCanvas = Canvas(windowFrame!!)
+
+        val outerRectangle = RectF(0f, 0f, width.toFloat(), height.toFloat())
+
+        var paint = Paint(Paint.ANTI_ALIAS_FLAG)
+        paint.color = Color.argb(ALPHA, 0, 0, 0)
+        osCanvas.drawRect(outerRectangle, paint)
+
+        paint.xfermode = PorterDuffXfermode(PorterDuff.Mode.SRC_OUT)
+        val innerRectangle = RectF(
```
NotWoods: Maybe use `Rect(...).toRectF()` so you don't need a lot of `toFloat` calls

components/feature/qr/src/main/java/mozilla/components/feature/qr/CustomOverlay.kt
```diff
@@ -0,0 +1,105 @@
...
+                loc[0].toFloat(),
+                (loc[1] - topOffset).toFloat(),
+                (width - (width - right)).toFloat(),
+                (height - (height - bottom)).toFloat()
+        )
+        osCanvas.drawRoundRect(innerRectangle, Companion.RADIUS, Companion.RADIUS, paint)
```
NotWoods: nit: use `RADIUS` directly
```suggestion
        osCanvas.drawRoundRect(innerRectangle, RADIUS, RADIUS, paint)
```

components/feature/qr/src/main/java/mozilla/components/feature/qr/CustomOverlay.kt
```diff
@@ -0,0 +1,105 @@
...
+ * @property windowFrame a reference to a [Bitmap], used to create the window frame.
+ * @property activity a reference to an [Activity], used to refer the child views of
+ * calling parent view.
+ * @property ALPHA defines the opacity of the background.
+ * @property RADIUS defines the radius of the rectangle corners.
+ */
+
+class CustomOverlay : LinearLayout {
+    private var windowFrame: Bitmap? = null
+    private lateinit var activity: Activity
```
NotWoods: I think you can just directly set this:
```suggestion
    private val activity = context as Activity
```

components/feature/qr/src/main/java/mozilla/components/feature/qr/CustomOverlay.kt
drigio: ``drawRoundRect()`` requires ``RectF()``. One of the ``Rectf()`` can be changed to ``Rect()`` which is used for ``outerRectangle``

components/feature/qr/src/main/java/mozilla/components/feature/qr/CustomOverlay.kt
drigio: I think you meant 
```java
windowFrame ?: createWindowFrame()
canvas.drawBitmap(windowFrame!!, 0f, 0f, null)
```
Because ``createWindowFrame()`` is a void fun

drigio:
Wow Thanks @NotWoods ! It was a great experience :)

components/feature/qr/src/main/java/mozilla/components/feature/qr/CustomOverlay.kt
```diff
@@ -0,0 +1,102 @@
...
+ * calling parent view.
+ * @property ALPHA defines the opacity of the background.
+ * @property RADIUS defines the radius of the rectangle corners.
+ */
+
+class CustomOverlay @JvmOverloads constructor (
+    context: Context,
+    attrs: AttributeSet? = null,
+    defStyleAttr: Int = 0
+) : LinearLayout(context, attrs, defStyleAttr) {
```
pocmo: I wonder why this is a `LinearLayout`? It doesn't seem to have any children?

components/feature/qr/src/main/java/mozilla/components/feature/qr/CustomOverlay.kt
```diff
@@ -0,0 +1,102 @@
...
+    companion object {
+        const val RADIUS = 30.0f
+        const val ALPHA = 170
+    }
+
+    /**
+     * This is called before the canvas is drawn. Initializes the windowFrame only once
+     * @param canvas The canvas to draw upon.
+     */
+    override fun dispatchDraw(canvas: Canvas) {
```
pocmo: Is there a reason why you are using `dispatchDraw` instead of `onDraw`?

components/feature/qr/src/main/java/mozilla/components/feature/qr/CustomOverlay.kt
```diff
@@ -0,0 +1,102 @@
...
+    override fun isClickable() = false
+
+    /**
+     * This function creates the windows frame around our R.id.qr_reticle ImageView
+     * dynamically, so this will scale for all display sizes. Sets an alpha of 170%
+     * to give a dimmed look for the background
+     */
+    private fun createWindowFrame() {
+        val dm = DisplayMetrics()
+        val imageView = activity.findViewById(R.id.qr_reticle) as ImageView
```
pocmo: It's a bit fragile that this view looks for other views in its parent activity. I wonder if that implementation could be unified: One ImageView that draws the reticle and the frame (by extending `AppCompatImageView`)?

components/feature/qr/src/main/java/mozilla/components/feature/qr/CustomOverlay.kt
drigio: Previously I had thought of creating a child view that will hold the *reticle*. Hence the ``LinearLayout``.

components/feature/qr/src/main/java/mozilla/components/feature/qr/CustomOverlay.kt
drigio: Well, as there were gonna be child views I wanted to use ``dispatchDraw()`` to get control before the child views are drawn.

components/feature/qr/src/main/java/mozilla/components/feature/qr/CustomOverlay.kt
drigio: Actually I had thought of this, but instead using a ``LinearLayout``. I was going to have child view that would be an ``ImageView`` and then get the reference of its dimensions and draw this overlay. Turns out child views are rendered at a later stage. Hence I had to use the view in its parent, as it would have been drawn already there. Also extending a simple ``View`` instead of ``LinearLayout`` or ``AppCompatImageView`` would have done the work as we have to set the parent view, i.e the overlay to span over the whole screen i.e. ``match_parent``. So no use of ``AppCompatImageView`` there because we are not setting any *drawable* for it.  The child of these views would've been an ImageView which would contain the *reticle*. I hope this clears the reason?

drigio:
Hey @pocmo @NotWoods is there anything that I need to do more / any changes that need to be done? Please do point me in the right direction if possible, as contributing to large organizations is still new to me :)

boek:
@pocmo bump :)

drigio:
Hi @pocmo @NotWoods  , may I know what was remaining so that next time I would take care of those things?

Issue #4711, opened 2019-10-10
Create a per-task tab selection system


NotWoods:
Currently, we represent tabs as a list of objects with IDs, and have a field for the "selected" tab ID.

Unfortunately this system has some limitations. Notably, when dealing with custom tabs we need to track the custom tab session ID separately and pass that around rather than working with the selected tab. However, hardcoding the session ID makes it difficult to deal with pop-up windows and `_blank` links that open a new tab.

If we want to support multiple windows or split-screen (with the same browser app in each window/split) then there will also be multiple "selected" tabs at once.

We need to set up a new system. I suggest a system where the "selected" tab is local to the Fragment, rather than global. This could come in the form of storing it as saved state in the Fragment/Activity and passing it around, or through a map of task -> selected session. Like the current "selected" tab ID field, this ID needs to be mutable so that it can react to new tabs opening. 

Handling the selected tab per-task will also improve the custom tab workflow as custom tabs will be just another task. As a result the session handling will be the same for both custom tabs and normal tabs.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-3960)

pocmo:
Can you elaborate a bit more what you are trying to solve here? (e.g. what you described in the second paragraph).

> If we want to support multiple windows or split-screen (with the same browser app in each window/split) then there will also be multiple "selected" tabs at once.

This is something we did not intent to do and therefore also did not design for it.

> We need to set up a new system. I suggest a system where the "selected" tab is local to the Fragment, rather than global. This could come in the form of storing it as saved state in the Fragment/Activity and passing it around, or through a map of task -> selected session.

I'm not sure how this would work in practice. The selected tab is independent from the view. Even if the view is destroyed and not visible anymore it is still the selected tab and will be selected the next time the app is launched.

jonalmeida:
Taking a shot to explain what Tiger is suggesting: a single Custom Tab can have multiple sessions with new window requests (e.g. a pop-up window to login that closes). 

So we would need to observe the current session and when a new window request shows up, we want to new the same View™ with the new session, giving the impression it was loading in the same custom tab window, and when that child session wants to close itself, we would need to observe the parent session (via the `parentId` I imagine).

jonalmeida:
[...]

PR #4901, opened 2019-10-30
Closes #4553: Migrate WindowFeature to browser-state

### LINKED ISSUE #4553: Migrate WindowFeature to browser-state
Add `WindowRequest` state to browser-state and synchronize it with `browser-session`.

csadilek:
This migrates our `WindowFeature` to `browser-state` and deletes the relevant code from `browser-session`.

Instead of keeping two distinct states for open and close window requests, this uses just one, which simplifies store observers (e.g the feature implementations). This wasn't really useful before anyway, so I decided to simplify here and introduce a type instead.

The `WindowFeature` now also uses our `TabsUseCases` to open and close tabs in response to window requests, instead of interacting with the `SessionManager` directly. This follows the same pattern as our other migrations. The use cases will be the final classes to migrate once everything else is done. For this to work, we need to move `WindowFeature` from `feature-session` to `feature-tabs` because otherwise we'd be introducing a dependency cycle. This is a better fit anyway, as the feature reacts to window request by opening and closing tabs, and a separate feature module seems unnecessary.


csadilek:
PR to fix R-B after this lands: https://github.com/mozilla-mobile/reference-browser/pull/933 (will send one for Fenix once this lands.)

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
```diff
@@ -57,16 +57,29 @@ class CustomTabWindowFeature(
...
     override fun start() {
-        sessionManager.findSessionById(sessionId)?.register(windowObserver)
+        scope = store.flowScoped { flow ->
+            flow.mapNotNull { state -> state.findCustomTab(sessionId) }
+                .ifChanged {
+                    it.content.windowRequest
+                }
+                .collect { state ->
+                    val windowRequest = state.content.windowRequest
+                    if (windowRequest?.type == WindowRequest.Type.OPEN) {
```
grigoryk: What would `windowRequest?.type == WindowRequest.Type.CLOSE` mean in this context? Ideally you'd want to close a custom tab, right?

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
```diff
@@ -57,16 +57,29 @@ class CustomTabWindowFeature(
...
+                        val intent = configToIntent(state.config)
+                        intent.launchUrl(context, windowRequest.url.toUri())
+                        store.dispatch(ContentAction.ConsumeWindowRequestAction(sessionId))
```
grigoryk: If some of the state changes may be dropped due to conflation, this `action` isn't guaranteed to be noticed by the code above... But I think that's probably fine - the reducer, and I'm assuming it will process _every_ action... -  will set the `windowRequest` value to `null`... So some future `UpdateWindowRequestAction` will reset `state.windowRequest` to a brand new instance...

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
```diff
@@ -57,16 +57,29 @@ class CustomTabWindowFeature(
...
     /**
-     * Starts the feature and a observer to listen for window requests.
+     * Starts observing the configured session to listen for window requests.
      */
     override fun start() {
-        sessionManager.findSessionById(sessionId)?.register(windowObserver)
+        scope = store.flowScoped { flow ->
+            flow.mapNotNull { state -> state.findCustomTab(sessionId) }
+                .ifChanged {
+                    it.content.windowRequest
```
grigoryk: `ifChanged` is using referential equality, not structural one (e.g. iiuc it's just checking that this is the same object reference, not that its contents haven't changed).

Is it possible that emitters of `UpdateWindowRequestAction` will reuse instances of `windowRequest`, and just change wrapped `url`? If that's possible, I don't think this check will behave correctly.

components/feature/tabs/src/main/java/mozilla/components/feature/tabs/WindowFeature.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    private var scope: CoroutineScope? = null
+
+    /**
+     * Starts observing the selected session to listen for window requests
+     * and opens / closes tabs as needed.
+     */
+    override fun start() {
+        scope = store.flowScoped { flow ->
+            flow.mapNotNull { state -> state.selectedTab }
```
grigoryk: I think all of my questions above apply here as well, since the same underlying primitives are used.

components/feature/tabs/src/main/java/mozilla/components/feature/tabs/WindowFeature.kt
```diff
@@ -0,0 +1,82 @@
...
+    ) {
+        consume()
+        store.dispatch(ContentAction.ConsumeWindowRequestAction(tabId))
+    }
+
+    /**
+     * Stops observing the selected session for incoming window requests.
+     */
+    override fun stop() {
+        scope?.cancel()
```
grigoryk: `flowScoped` returns `MainScope`, which `has [SupervisorJob] and [Dispatchers.Main] context elements.`.

What effect will cancellation have?

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
```diff
@@ -57,16 +57,29 @@ class CustomTabWindowFeature(
     }
 
     /**
-     * Starts the feature and a observer to listen for window requests.
+     * Starts observing the configured session to listen for window requests.
      */
     override fun start() {
-        sessionManager.findSessionById(sessionId)?.register(windowObserver)
+        scope = store.flowScoped { flow ->
```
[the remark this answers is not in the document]
csadilek: So, the way I see it, the behaviour is slightly different. We won't process new state changes in the reducer before we've notified all observers. So, there's no way for a new `WindowRequest` to come in. This is the main advantage of `browser-state` over `browser-session`. We're able to shield from side-effects that observers themselves might trigger and can guarantee that things happen in a reliable order. This would be a bug in `browser-state` if it didn't work.

You're right though that we'll only process one `WindowRequest` and that was deliberate on my part and the same as we have now. That's fine for users clicking on `target="_blank" links. If a site calls `window.open` in a loop then we won't act on that, we'd only open one window. Do you consider this a problem? If so, let's file a follow-up issue, as that would be quite tricky to fix as part of the migration.

Another reason multiple `window.open` calls won't be acted on is because we immediately switch to the newly opened window/tab and since we're observing the selected session in this flow, we wouldn't be observing `window.open` of the previous tab. So, that's the way I intended for this to work.

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
csadilek: Yeah, exactly. If we ever want to support `window.close` for `CustomTab`s we'd have to add that, but we don't support that right now. In fact, we haven't even wired up engine gecko to do that.

csadilek: We might do this soon for https://github.com/mozilla-mobile/android-components/issues/4711 but for now we should be good.

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
csadilek: That's interesting. All our migrated features right now (prompts, downloads, etc.) work like this. We're not re-using request objects and they should be immutable for that reason! See: https://github.com/mozilla-mobile/android-components/blob/master/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/window/GeckoWindowRequest.kt and https://github.com/mozilla-mobile/android-components/blob/master/components/concept/engine/src/main/java/mozilla/components/concept/engine/prompt/PromptRequest.kt

If we find some that are mutable, we should file as a bug, but I guess immutability should definitely required for any state in our new browser-store. If we don't yet point that out in the docs, we should add some exclamation marks around it :)

csadilek: For `WindowRequest`, we should be fine though and this should behave correctly.

components/feature/tabs/src/main/java/mozilla/components/feature/tabs/WindowFeature.kt
csadilek: Yes, same pattern, see above.

components/feature/tabs/src/main/java/mozilla/components/feature/tabs/WindowFeature.kt
csadilek: Ah yes, this took me a while to get used to as well! I also assumed that `MainScope().cancel` would be a problem, because I misinterpreted what it was doing. This is really just cancelling the scope we create here: https://github.com/mozilla-mobile/android-components/blob/master/components/concept/engine/src/main/java/mozilla/components/concept/engine/prompt/PromptRequest.kt

So, we really just want to stop observing / consuming from the flow, until the feature is started again. It doesn't stop / cancel anything else.

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
grigoryk: :+1:

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
grigoryk: Thanks for checking! I agree that as long as things are immutable, this should be fine.

> If we don't yet point that out in the docs, we should add it with some exclamation marks around it :)

:+1: :scroll: absolutely, yeah.

components/feature/tabs/src/main/java/mozilla/components/feature/tabs/WindowFeature.kt
grigoryk: You linked a wrong url, right? Or i'm just not seeing it.

But OK, I understand it now. This is just cancelling a just-created `SupervisorJob`. Well, this takes some mental gymnastics to read... Maybe `MainScope` could be named differently, like... `JobOnMainDispatcher`? :)

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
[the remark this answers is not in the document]
grigoryk: I'll need to think about this with a fresh brain tomorrow, but you appear confident about producer behaviour, so I'll take that :)

> Do you consider this a problem?

I'm not actually sure tbh, since I don't know how browsers normally behave in this case. But, I don't think handling "just call window.open a ton of times in a loop" is the right thing to do, so I think behaviour as-is is good.

components/feature/tabs/src/main/java/mozilla/components/feature/tabs/WindowFeature.kt
csadilek: Ah yeah wrong URL, meant to link to https://github.com/mozilla-mobile/android-components/blob/master/components/lib/state/src/main/java/mozilla/components/lib/state/ext/StoreExtensions.kt#L182

Yes, the naming isn't ideal, maybe we should create an alias and not use the lib function.

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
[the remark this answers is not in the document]
csadilek: OK, I'll double check and test as well tomorrow before landing!

grigoryk:
I've also not thought very much about the code shuffling you did. It seemed fine from description in https://github.com/mozilla-mobile/android-components/pull/4901#issue-334543493 though.

csadilek:
bors r=grigoryk

PR #4919, opened 2019-10-31
Closes #3486 - Implement Web Share API

### LINKED ISSUE #3486: Add support for Web Share API
GeckoView will be adding a delegate for the Web Share API, which we should handle to show a native sharing prompt.

The delegate API is expected to pass data corresponding to parts of the Web Share API:
- `title`: optional string
- `text`: optional string
- `url`: optional URL string
- `files`: optional list of URIs representing files to share

At least one of those values will be defined.

The delegate API is expected to have the following callbacks:
- `onDismiss`: when the share panel is dismissed or an error occurs, this will tell GeckoView to send an `AbortError` to the web app.
- `onSuccess`: when the share is complete, this will tell GeckoView to resolve the share promise.

## Steps that the implementation should take:

1. If there are no share targets available, call `onDismiss` and abort these steps.
2. Present the user with a choice of one or more share targets, selected at the user agent's discretion. The user MUST be given the option to cancel rather than choosing any of the share targets (done by native share UI). Wait for the user's choice.
3. If the user chose to cancel the share operation, call `onDismiss` and abort these steps.
4. Activate the chosen share target, convert data to a format suitable for ingestion into the target, and transmit the converted data to the target. If an error occurs starting the target or transmitting the data, call `onDismiss` and abort these steps.
5. Once the data has been successfully transmitted to the target, call `onSuccess`.

---

Spec: https://wicg.github.io/web-share/level-2/
GeckoView Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1557039

NotWoods:
ShareData is a separate data class so that Web Share Target can use it down the line.

---

web_share

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/prompt/GeckoPromptDelegate.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/prompt/GeckoPromptDelegate.kt
@@ -15,6 +15,7 @@ import mozilla.components.concept.engine.prompt.PromptRequest
 import mozilla.components.concept.engine.prompt.PromptRequest.MenuChoice
 import mozilla.components.concept.engine.prompt.PromptRequest.MultipleChoice
 import mozilla.components.concept.engine.prompt.PromptRequest.SingleChoice
+import mozilla.components.concept.engine.prompt.ShareData
 import mozilla.components.support.base.log.logger.Logger
 import mozilla.components.support.ktx.kotlin.toDate
 import org.mozilla.geckoview.AllowOrDeny
@@ -41,6 +42,7 @@ typealias GECKO_AUTH_LEVEL = PromptDelegate.AuthPrompt.AuthOptions.Level
 typealias GECKO_PROMPT_FILE_TYPE = PromptDelegate.FilePrompt.Type
 typealias GECKO_PROMPT_CHOICE_TYPE = PromptDelegate.ChoicePrompt.Type
 typealias GECKO_PROMPT_FILE_CAPTURE = PromptDelegate.FilePrompt.Capture
+typealias GECKO_PROMPT_SHARE_RESULT = PromptDelegate.SharePrompt.Result
 typealias AC_AUTH_LEVEL = PromptRequest.Authentication.Level
 typealias AC_AUTH_METHOD = PromptRequest.Authentication.Method
 typealias AC_FILE_FACING_MODE = PromptRequest.File.FacingMode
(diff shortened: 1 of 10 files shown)


components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/prompt/GeckoPromptDelegate.kt
```diff
@@ -300,6 +302,32 @@ internal class GeckoPromptDelegate(private val geckoEngineSession: GeckoEngineSe
...
+        val geckoResult = GeckoResult<PromptResponse>()
+        val onSuccess = { geckoResult.complete(prompt.confirm(GECKO_PROMPT_SHARE_RESULT.SUCCESS)) }
+        val onFailure = { geckoResult.complete(prompt.confirm(GECKO_PROMPT_SHARE_RESULT.FAILURE)) }
+        val onDismiss = { geckoResult.complete(prompt.dismiss()) }
+
+        geckoEngineSession.notifyObservers {
+            onPromptRequest(
+                PromptRequest.Share(
+                    ShareData(
+                        title = prompt.title,
```
jonalmeida: If all values are null, since they're all nullable, what are we sharing then? Should we just dismiss immediately?

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/share/SharePrompt.kt
```diff
@@ -0,0 +1,53 @@
...
+    fun handleShareRequest(promptRequest: Share, session: SessionState) {
+        val sendIntent = promptRequest.data.asSendIntent()
+        try {
+            displayShareSheet(sendIntent, session)
+        } catch (e: ActivityNotFoundException) {
+            feature.onCancel(session.id)
+        }
+    }
+
+    private fun displayShareSheet(sendIntent: Intent, session: SessionState) {
```
jonalmeida: We have `Context.share` (and it seems like you saw it). Does it not work for this case, and/or modify it to work?

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/prompt/GeckoPromptDelegate.kt
NotWoods: Theoretically Gecko shouldn't pass in a request if everything is null, and will [throw a TypeError instead](https://w3c.github.io/web-share/#navigator-interface). We don't have an API to throw that error from here so I don't think we need to double-check.

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt
```diff
@@ -89,18 +93,30 @@ class PromptFeature internal constructor(
         store: BrowserStore,
         customTabId: String? = null,
         fragmentManager: FragmentManager,
+        shareDelegate: ShareDelegate = DefaultShareDelegate(),
```
Amejia481: nit: I think we need to update the docs :)

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/prompt/GeckoPromptDelegate.kt
Amejia481: Did we communicate this issue to the GV team? This way we can enforce this restriction to the type level, they can improve the API to make its values non-null.

jonalmeida: > Did we communicate this issue to the GV team? This way we can enforce this restriction to the type level, they can improve the API to make its values non-null.

I looked at the spec, and all fields are optional, so there's no way for GV to enforce non-null at the GV API surface. 🤷‍♂

jonalmeida:
> The diff coverage is 55.55%.

slack-imgs com

NotWoods:
bors r=Amejia481

PR #4936, opened 2019-11-01
Closes #4824: Throw exceptions on certain Rust errors

### LINKED ISSUE #4824: AutopushFeature: throw exceptions on Rust errors
If we get a Rust error that is NOT `CommunicationError` or `CommunicationServerError`, we should throw on it since those are unrecoverable issues and we need to know about them.

See [RustError.kt][0] for the full list.

[0]: https://github.com/mozilla/application-services/blob/master/components/push/android/src/main/java/mozilla/appservices/push/RustError.kt#L15-L27

jonalmeida:
Moved the internal `CoroutineScope.launchAndTry` extension to be a bit more generic for easier testing.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
+++ b/components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
@@ -15,6 +15,8 @@ import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.asCoroutineDispatcher
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.plus
+import mozilla.appservices.push.CommunicationError
+import mozilla.appservices.push.CommunicationServerError
 import mozilla.appservices.push.SubscriptionResponse
 import mozilla.components.concept.push.Bus
 import mozilla.components.concept.push.EncryptedPushMessage
@@ -120,7 +122,7 @@ class AutoPushFeature(
         service.stop()
 
         DeliveryManager.with(connection) {
-            scope.launch {
+            job = scope.launch {
                 // TODO replace with unsubscribeAll API when available
                 PushType.values().forEach { type ->
                     unsubscribe(type.toChannelId())
@@ -138,7 +140,7 @@ class AutoPushFeature(
      * each push type and notifies the subscribers.
      */
     override fun onNewToken(newToken: String) {
-        scope.launchAndTry {
+        job = scope.launchAndTry {
             logger.info("Received a new registration token from push service.")
 
             connection.updateToken(newToken)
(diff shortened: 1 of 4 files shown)


components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
```diff
@@ -337,6 +335,23 @@ class AutoPushFeature(
...
 
+internal fun CoroutineScope.launchAndTry(
+    block: suspend CoroutineScope.() -> Unit,
+    errorBlock: (Exception) -> Unit
+): Job {
+    return launch {
+        try {
+            block()
+        } catch (e: RustPushError) {
+            if (e !is CommunicationServerError && e !is CommunicationError) {
```
grigoryk: I'd prefer a `when`, but this works. Worst case is that we'll accidentally throw when we shouldn't (e.g. a new "non-terminal" error type is added in the underlying library, and this check degrades), which we should be able to catch in Sentry and resolve.

Perhaps though you can guard against this hypothetical future by forcing an exhaustive `when` treatment? e.g.
```
try {
  block()
} catch (e: RustPushError) {
   if (shouldReThrow(e)) {
     throw e
   }
}

private fun shouldReThrow(e: RustPushError): Boolean {
  return when (e) {
    // Do not re-throw intermittent errors.
    is CommunicationServerError, CommunicationError -> false
    // Re-throw internal errors that are unrecoverable/indicative of bugs. 
    is InternalPanic, CryptoError, AlreadyRegisteredError, StorageError, MissingRegistrationTokenError, StorageSqlError, TranscodingError, UrlParseError -> true
  }
}
```

jonalmeida: You're right, this is a much safer change. I'd rather fail to compile. I was just being a lazy typist.

grigoryk: Well, I did the exact thing as you in my version in FxA... So yeah, we're all lazy :) I'll push a PR to fix up a similar function in FxA.

jonalmeida: Ah, since the Rust errors are just inherited and not a sealed class, we can never truly make this exhaustive; I have to add _every_ exception to make the compiler happy. 

If we take this direction, we would be swallowing up new error types instead of throwing on them i.e. doing the inverse of the worst case you mentioned.

My preference is to throw on a new error type added. If it was added, that means new functionality was added, and that means the new thing is failing already! 😄 

Kotlin when clause still not happy when we've made every error explicit. We should really ask a-s to change the errors, but that's not really worth their time.

grigoryk: Yeah, I just hit that as well.

> My preference is to throw on a new error type added.

Agreed, let's go with that!

jonalmeida:
bors r=grigoryk

PR #4937, opened 2019-11-02
Make sure handleFxaExceptions doesn't swallow newly added fxa exceptions

grigoryk:
@jonalmeida this is a similar change to what we discussed in https://github.com/mozilla-mobile/android-components/pull/4936/files


grigoryk:
Filed an issue for an unrelated intermittent: https://github.com/mozilla-mobile/android-components/issues/4938

Issue #4907, opened 2019-10-31
[Bug]"Search or enter address" strings improperly positioned in RTL


BranescuMihai:
## Prerequisites
Have RTL enabled in Developer options

## Steps to reproduce
1. Tap on the search bar
2. Check the position of the "Search or enter address" from the search bar

### Expected behavior
Should be properly positioned 

### Actual behavior
"Search or enter address" string to close to the left edge of the search box

### Device information

* Android device: 
• Google Pixel 3a XL (Android 9)
• Samsung Galaxy Note 9 (Android 8.1.0)
• Samsung Galaxy S7 (Android 7)
• LG Nexus 4 (Android 5.1.1)

* Fenix version:  Nightly Build #12980627 from 25/10

[Related to #6269 from Fenix](https://github.com/mozilla-mobile/fenix/issues/6269)

![rsz_rsz_screenshot_20191025-143102](https://user-images.githubusercontent.com/51314259/67572721-72a61c80-f73f-11e9-8e2b-b9f4799b40d8.png)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-4156)

boek:
@BranescuMihai it looks like you wrote some code for this. Did it ever make it into a-c?

darkwing:
Pinging again @BranescuMihai :) . Hoping to fix https://github.com/mozilla-mobile/fenix/issues/6269 ASAP!

BranescuMihai:
@boek  @darkwing I'm sorry for not responding faster, somehow I didn't get a notification. There is a PR open on this, I also forgot about it until recently. Feel free to review it, there are 4 possible scenarios mentioned in the description that had to be taken into account.
[https://github.com/mozilla-mobile/android-components/pull/4946](https://github.com/mozilla-mobile/android-components/pull/4946)

PR #4946, opened 2019-11-04
For #4907 - RTL url no margin set

BranescuMihai:
This change takes into account all four possible scenarios:
Legend:
X button = X
ActionContainer = A
UrlView = U

1. (X GONE) (A GONE) <- 8dp -> (U)
2. (X GONE) (A) (U)
3. (X) (A) (U)
4. (X) (A GONE) (U)

So basically, this adds 8dp only when both views in its anchoring chain are GONE.
---

### Pull Request checklist

- [x] **Tests**: This PR has only a visual change, no tests needed

### After merge

<table style="width:100%">
  <tr>
    <th>RTL</th>
    <th>RTL with X button</th>
  </tr>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>

### DIFF
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/edit/EditToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/edit/EditToolbar.kt
@@ -14,6 +14,8 @@ import android.widget.ImageView
 import androidx.annotation.ColorInt
 import androidx.annotation.VisibleForTesting
 import androidx.annotation.VisibleForTesting.PRIVATE
+import androidx.constraintlayout.widget.ConstraintLayout
+import androidx.constraintlayout.widget.ConstraintSet
 import androidx.core.content.ContextCompat
 import androidx.core.view.inputmethod.EditorInfoCompat
 import androidx.core.view.isVisible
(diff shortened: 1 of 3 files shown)


BranescuMihai:
> From the screenshots I wonder if this is a scenario that actually exists in the real world. In this case RTL is forced by the developer setting and the icon is displayed on the right while the text is actually still LTR. Shouldn't the text be on the right side next to the icon too?

@pocmo this was one of my questions as well. Looked into it here: [Material bidirectionality -> mirroring layout](https://material.io/design/usability/bidirectionality.html#), turns out texts that are in English(native LTR) like URLs should be kept LTR and not aligned to the right (which makes sense, otherwise we would need to manually position the cursor after each letter tapped)

pocmo:
@BranescuMihai I think that makes sense. But then this "bug" would never occur for someone using an RTL language I guess? 🤔

BranescuMihai:
@pocmo yes, there are a few reasons why there is a bug there:
1. The main problem of this bug was actually the hint, which is in English, and from what I can see, we do not support RTL languages in Fenix strings (no Hebrew, no Arabic translations)
2. If I am from a RTL country, and I want to write to a friend in English, even though my phone's language is set to Arabic(for example), my keyboard's language can be in English, thus the writing will be LTR

Issue #4915, opened 2019-10-31
Allow toolbar menu buttons to determine their state asynchronously (or provide alternative)


Baron-Severin:
### Use case
As part of https://github.com/mozilla-mobile/fenix/issues/4281, determining whether or not the current site is bookmarked needs to happen within `TwoStateButton#isInPrimaryState`.  Because of how `TwoStateButton` works, this call must be synchronous.  Unfortunately checking if the site is bookmarked is very slow (10-50MS in my tests).

The best fix would be for bookmark checks to be faster (and I've notified the perf team), but I also don't think it is unreasonable to expect that menu items may need to make slow calls to determine their state in the future.  It would be good to have a canonical way of handling this case.

### Fenix workaround
https://github.com/mozilla-mobile/fenix/pull/6310/commits/572e2f6762ebeb5c6d8b65d4d676cb85288a6530

Baron-Severin:
It's also worth noting that, since all the logic is driven by the component, the current toolbar API doesn't play particularly well with LibState.  One possible solution to this could be to have the toolbar handle its presentation details, but to expose `fun update(state: ToolbarState)` and allow client code to drive the logic.  This could also help pave the way for [a future Fenix issue](https://github.com/mozilla-mobile/fenix/issues/6275) that will require updating the toolbar menu while it is open.

NotWoods:
Closing because this is offered in menu2 (#5373)

PR #4951, opened 2019-11-04
Closes #4651: Add checks for WebExtensionBrowserAction references

### LINKED ISSUE #4651: Provide an api to allow partial update of browser actions or page actions in the toolbar
While we are working on web extensions adding browser actions/page actions items to the toolbar. We discovered that we don't have a way to update an existing item without re-render the whole toolbar. 

Browser actions/page actions could have properties like `badgeText` to inform the user about some state on the web extension.
image

They can change pretty frequent, and we need a way to updated a specific single item on the toolbar  (browser actions/page actions)

psymoon:


components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
```diff
@@ -36,6 +36,7 @@ class ToolbarPresenter(
     internal var renderer = URLRenderer(toolbar, urlRenderConfiguration)
 
     private var scope: CoroutineScope? = null
+    private val toolbarActionSet = HashSet<WebExtensionToolbarAction>()
```
Amejia481: As a web extension can only have one browser action or page action, I think we could use a `map` instead of a `HashSet`, where the key will be the `WebExtension.id` and the value will be the `WebExtensionToolbarAction` object. This way we can make sure that `WebExtensionToolbarAction`s are per `WebExtensionToolbarAction`.

Also this way you don't have to override `equals` and `hashcode` in `WebExtensionToolbarAction`.

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/WebExtensionToolbarAction.kt
```diff
@@ -64,4 +61,14 @@ open class WebExtensionToolbarAction(
         textView.setTextColor(badgeTextColor)
         textView.setBackgroundColor(badgeBackgroundColor)
     }
+
+    override fun equals(other: Any?): Boolean {
+        if (other is WebExtensionToolbarAction) {
+            return this.browserAction === other.browserAction
```
Amejia481: I think comparing using reference could cause duplication maybe we could use an approach as the one suggested above where we don't have to override `equals` nor `hashCode`. What do you think?

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
```diff
@@ -97,27 +98,20 @@ class ToolbarPresenter(
...
-                val toolbarAction = convertToToolbarAction(extensionAction)
-                // We need to verify, if we this toolbarAction already exits on the toolbar
-                // if so, we don't need to re-add it, just replace it with the new one.
-                // https://github.com/mozilla-mobile/android-components/issues/4651
-                toolbar.addBrowserAction(toolbarAction)
+                val toolbarAction = WebExtensionToolbarAction(
+                    browserAction = extensionAction,
+                    listener = extensionAction.onClick)
+
+                if (!toolbarActionSet.contains(toolbarAction)) {
```
Amejia481: If we use the map approach, when we notice that the item is already there we need to replace its `WebExtensionToolbarAction.browserAction` references with the new one (`extensionAction`) other wise we are going to added to the toolbar.

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
psymoon: @Amejia481 Hmm, I think `HashSet` has constant lookup time. If we use map as you suggest, it'd be as good as using a list since we'd always iterate every time render() is fired up. Is there any other reason for using a map aside from eliminating duplicates?

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/WebExtensionToolbarAction.kt
psymoon: What other use cases for `WebExtensionToolbarAction` that would deter from overriding object value comparables vs. references?

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
psymoon: Unless I'm misunderstanding here, are we checking against `WebExtension.id`, not `BrowserAction` references?

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/WebExtensionToolbarAction.kt
Amejia481: Normally it is not a good idea to override `equals` and `hashcode` unless is completely necessarily. Because after we do it, we have a maintenance cost, as we have to watch for these functions after every change in the class, for this is better to use [data classes](https://kotlinlang.org/docs/reference/data-classes.html) that save of this maintenance work, as they regenerate `equals()` and `hashCode()` for use.

`WebExtensionToolbarAction` is not **data class** now because, `data classes` doesn't allow inheritance :( and as **browser actions** and **page actions** look exactly the same the only differences is the place where they are located, I was thinking about reusing the behavior on `WebExtensionToolbarAction` for page action because we only really need another type with the same functionality.

**pageAction** Inside of the url
image

**browserAction** Inside of the toolbar
image

csadilek:
bors abort

components/feature/toolbar/src/test/java/mozilla/components/feature/toolbar/ToolbarPresenterTest.kt
```diff
@@ -493,4 +472,56 @@ class ToolbarPresenterTest {
         verify(toolbar).displayProgress(0)
         verify(toolbar).siteSecure = Toolbar.SiteSecurity.INSECURE
     }
+
+    @Test
+    fun `Toolbar render WebExtensionActions`() {
```
csadilek: nit: test name could be a little clearer :)

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
```diff
@@ -97,27 +101,22 @@ class ToolbarPresenter(
...
+                if (existingBrowserAction != null) {
+                    existingBrowserAction.browserAction = extensionAction
+                } else {
+                    val toolbarAction = WebExtensionToolbarAction(
+                        browserAction = extensionAction,
+                        listener = extensionAction.onClick)
+                    toolbar.addBrowserAction(toolbarAction)
+                    webExtensionToolbarMap[extension.id] = toolbarAction
+                }
                 toolbar.invalidateActions()
```
csadilek: Initially, I thought we're gonna look for a way to only call invalidate on the single action that's affected. This will invalidate the menu/browserActions/pageActions/navigationActions every time. For uBlock this will definitely be called a lot, but I am not sure that's actually a performance problem. So, OK to land as is if it's tricky to implement and we can revisit later.

Amejia481: @pocmo I remembered that we talked about this. But I don't recall if you have any concerns related to add functionality for updating  a specific `Toolbar.Action `. Should we worry about calling `toolbar.invalidateActions()` when just one item has changed?

pocmo: I think that is fine. On invalidate the toolbar will:
- call the visible lambda of all actions to check whether they still want to be in the toolbar
- (only if needed it adds or removes views for new/hidden/visible actions)
- it will call bind() on all actions to give them the opportunity to update their view (which gets passed in)

This is cheap: The visible lambda doesn't do much and the bind() call for all default actions doesn't do anything at all because by default actions just don't change.

Amejia481: Thanks!
cc: @psymoon

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
```diff
@@ -37,6 +37,10 @@ class ToolbarPresenter(
 
     private var scope: CoroutineScope? = null
 
+    // {Key: WebExtension id, Value: WebExtensionToolbarAction}
+    @VisibleForTesting
+    internal val webExtensionToolbarMap = HashMap<String, WebExtensionToolbarAction>()
```
pocmo: Do we need this map to be here (which makes the presenter stateful) or could this live in the state directly (e.g. we add this as an optional field to the webextension state)

csadilek: The `WebExtensionToolbarAction` is mutable UI state, so I think at least in its current form we should not be adding it to the webextension state. 

We could make it immutable, but then I think we need toolbar API to support updating individual browser actions.

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
csadilek: Applications may do expensive things in the various lambdas though which are called during `bind` e.g. see also https://github.com/mozilla-mobile/android-components/issues/4915 which is needed because a first version queried the bookmark state from storage in `TwoStateButton#isInPrimaryState`. Of course, we could document that this is a bad idea and hope apps don't do that...

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarPresenter.kt
csadilek: Just to clarify because the naming here overlaps. The web extension `BrowserAction` is already in the state, both globally and per tab (to support tab overrides). The `WebExtensionToolbarAction` is just a wrapper for it for displaying purposes in the toolbar, which is why I wouldn't add it to the state. I would prefer for it to be immutable though, but this would require toolbar API.

PR #4955, opened 2019-11-04
Closes #4682: Add support for browser.tabs.remove

### LINKED ISSUE #4682: Add support for browser.tabs.remove
Currently we only have support for browser.tabs.create with https://github.com/mozilla-mobile/android-components/pull/4680.

csadilek:
Brings in support for `browser.tabs.remove`. All very similar to what we did for `browser.tabs.add`. Refactored / Broke apart the tests so each one only verifies one behaviour.

### DIFF
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
@@ -26,6 +26,7 @@ import mozilla.components.concept.engine.utils.EngineVersion
 import mozilla.components.concept.engine.webextension.WebExtension
 import mozilla.components.concept.engine.webextension.WebExtensionDelegate
 import org.json.JSONObject
+import org.mozilla.geckoview.AllowOrDeny
 import org.mozilla.geckoview.ContentBlocking
 import org.mozilla.geckoview.ContentBlockingController
 import org.mozilla.geckoview.ContentBlockingController.Event
@@ -36,6 +37,7 @@ import org.mozilla.geckoview.GeckoSession
 import org.mozilla.geckoview.GeckoWebExecutor
 import org.mozilla.geckoview.WebExtensionController
 import java.lang.IllegalStateException
+import java.util.WeakHashMap
 
 /**
  * Gecko-based implementation of Engine interface.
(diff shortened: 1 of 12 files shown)


components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
```diff
@@ -158,6 +160,11 @@ class GeckoEngine(
         this.webExtensionDelegate = webExtensionDelegate
 
         val tabsDelegate = object : WebExtensionController.TabDelegate {
+            // We use this map to:
+            // - Ensures a web extension can only close tabs it has opened
```
pocmo: Why is that the case? Looking at the tabs API documentation I can't see that restriction being mentioned?

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
```diff
@@ -158,6 +160,11 @@ class GeckoEngine(
...
+            // - Find the engine session of a given gecko session
```
pocmo: Can't we also do this lookup by going over the engine sessions of all tabs (with a cast)?

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
```diff
@@ -158,6 +160,11 @@ class GeckoEngine(
...
+            private val tabs = WeakHashMap<GeckoEngineSession, String>()
```
pocmo: If we really need this mapping .. would it be an option to put it into our state (e.g. the extension id in the tab) instead of having a separate mini state here?

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
csadilek: OK, as discussed, since we can't currently perform this lookup in an acceptable way, we will keep this map here and live with the limitation that a web extension can only close tabs it has opened. This problem will go away once the browser-state migration is done and the engine has access to the browser store. The issue is tracked here: https://github.com/mozilla-mobile/android-components/issues/4965

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
csadilek: OK, follow-up filed: https://github.com/mozilla-mobile/android-components/issues/4965 (see comment below)

csadilek: OK, follow-up filed: https://github.com/mozilla-mobile/android-components/issues/4965 (see comment below)

csadilek:
bors r=Amejia481

jonalmeida:
bors retry

PR #4958, opened 2019-11-05
Password sync support

grigoryk:
This does a few things in order to enable password syncing.

Our underlying storage for passwords is encrypted at rest. This means that we need to keep the passwords storage layer unlocked for the duration of a sync, which is what this PR achieves.

In order to keep password storage layer unlocked, we need to be able to access the encryption key from the worker. To achieve that, worker is now using the new `KeySharedPreferences` API to obtain a `passwords_key`.

`KeySharedPreferences` was added to make it easier to transparently store key/values pairs such that `value` is encrypted at rest, protected by a key which is stored in AndroidKeyStore. Due to issues with the AndroidKeyStore for API versions prior to 23, on those older API levels `KeySharedPreferences` simply delegates to SharedPreferences, without any encryption.

`KeySharedPreferences` supports an API upgrade scenario when going from pre-API23 to post-API23, but not the other way.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/concept/sync/src/main/java/mozilla/components/concept/sync/Sync.kt
+++ b/components/concept/sync/src/main/java/mozilla/components/concept/sync/Sync.kt
@@ -42,8 +42,21 @@ data class SyncAuthInfo(
     val tokenServerUrl: String
 )
 
+/**
+ * An extension of [SyncableStore] that can be locked/unlocked using an encryption key.
+ */
 interface LockableStore : SyncableStore {
+    /**
+     * Unlocks underlying storage using [encryptionKey]. Storage must be [ensureLocked] after use.
+     *
+     * @param encryptionKey Plaintext encryption key used by the underlying storage implementation (e.g. sqlcipher)
+     * to key the store.
+     */
     suspend fun ensureUnlocked(encryptionKey: String)
+
+    /**
+     * Locks the underlying storage. Call this after you're done with the store following [ensureLocked].
+     */
     suspend fun ensureLocked()
 }
 
--- a/components/concept/sync/src/main/java/mozilla/components/concept/sync/Sync.kt
+++ b/components/concept/sync/src/main/java/mozilla/components/concept/sync/Sync.kt
@@ -47,17 +47,13 @@ data class SyncAuthInfo(
  */
 interface LockableStore : SyncableStore {
     /**
-     * Unlocks underlying storage using [encryptionKey]. Storage must be [ensureLocked] after use.
+     * Executes a [block] while keeping the store in an unlocked state. Store is locked once [block] is finished.
      *
      * @param encryptionKey Plaintext encryption key used by the underlying storage implementation (e.g. sqlcipher)
      * to key the store.
+     * @param block A lambda to execute while the store is unlocked.
      */
-    suspend fun ensureUnlocked(encryptionKey: String)
-
-    /**
-     * Locks the underlying storage. Call this after you're done with the store following [ensureLocked].
-     */
-    suspend fun ensureLocked()
+    suspend fun <T> unlocked(encryptionKey: String, block: (store: LockableStore) -> T): T
 }
 
 /**
--- a/components/concept/sync/src/main/java/mozilla/components/concept/sync/Sync.kt
+++ b/components/concept/sync/src/main/java/mozilla/components/concept/sync/Sync.kt
@@ -42,6 +42,11 @@ data class SyncAuthInfo(
     val tokenServerUrl: String
 )
 
+interface LockableStore : SyncableStore {
+    suspend fun ensureUnlocked(encryptionKey: String)
+    suspend fun ensureLocked()
+}
+
 /**
  * Describes a "sync" entry point for a storage layer.
  */
--- a/components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/KeySharedPreferences.kt
+++ b/components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/KeySharedPreferences.kt
@@ -1,11 +1,10 @@
-/*
- * This Source Code Form is subject to the terms of the Mozilla Public
+/* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/.
- */
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 package mozilla.components.lib.dataprotect
 
+import android.annotation.TargetApi
 import android.content.Context
 import android.content.Context.MODE_PRIVATE
 import android.os.Build
@@ -17,19 +16,47 @@ import java.security.GeneralSecurityException
 
 private const val BASE_64_FLAGS = Base64.URL_SAFE or Base64.NO_PADDING
 
-open class KeySharedPreferences(
+/**
+ * A wrapper around [SharedPreferences] which encrypts contents on supported API versions (23+).
+ * Otherwise, this simply delegates to [SharedPreferences].
+ *
+ * In rare circumstances (such as APK signing key rotation) a master key which protects this storage may be lost,
+ * in which case previously stored values will be lost as well. Applications are encouraged to instrument such events.
+ *
+ * @param context A [Context], used for accessing [SharedPreferences].
+ * @param keystore An optional [Keystore]: required when running on API23+, expected as `null` otherwise.
+ */
+class KeySharedPreferences(
     context: Context,
-    private val keystore: Keystore? = Keystore(BuildConfig.APPLICATION_ID)
+    private val keystore: Keystore? = null
 ) {
 
     companion object {
         const val KEY_PREFERENCES = "key_preferences"
     }
 
+    init {
+        if (Build.VERSION.SDK_INT < M) {
+            require(keystore == null) {
+                "Keystore isn't supported for pre-23 API versions"
+            }
+        } else {
+            require(keystore != null) {
+                "Keystore must be configured for 23+ API versions"
+            }
+        }
+    }
+
     private val logger = Logger("KeySharedPreferences")
     private val prefs = context.getSharedPreferences(KEY_PREFERENCES, MODE_PRIVATE)
 
-    open fun getString(key: String): String? {
+    /**
+     * Retrieves a stored [key]. See [putString] for storing a [key].
+     *
+     * @param key A key name.
+     * @return An optional [String] if [key] is present in the store.
+     */
+    fun getString(key: String): String? {
         return if (Build.VERSION.SDK_INT >= M) {
             getStringM(key)
         } else {
@@ -37,8 +64,33 @@ open class KeySharedPreferences(
         }
     }
 
+    /**
+     * Stores [value] under [key]. Retrieve it using [getString].
+     *
+     * @param key A key name.
+     * @param value A value for [key].
+     */
+    fun putString(key: String, value: String) {
+        if (Build.VERSION.SDK_INT >= M) {
+            putStringM(key, value)
+        } else {
+            prefs.edit().putString(key, value).apply()
+        }
+    }
+
+    /**
+     * Removes key/value pair from storage for the provided [key].
+     */
+    fun remove(key: String) {
+        prefs.edit().remove(key).apply()
+    }
+
+    @TargetApi(M)
     private fun getStringM(key: String): String? {
-        verifyKey()
+        // The fact that we're possibly generating a managed key here implies that this key could be lost after being
+        // for some reason. One possible reason for a key to be lost is rotating signing keys for the APK.
+        // Applications are encouraged to instrument such events.
+        generateManagedKeyIfNecessary()
 
         return if (prefs.contains(key)) {
             val value = prefs.getString(key, "")
@@ -62,16 +114,9 @@ open class KeySharedPreferences(
         }
     }
 
-    open fun putString(key: String, value: String) {
-        if (Build.VERSION.SDK_INT >= M) {
-            putStringM(key, value)
-        } else {
-            prefs.edit().putString(key, value).apply()
-        }
-    }
-
+    @TargetApi(M)
     private fun putStringM(key: String, value: String) {
-        verifyKey()
+        generateManagedKeyIfNecessary()
         val editor = prefs.edit()
 
         val encrypted = keystore!!.encryptBytes(value.toByteArray(StandardCharsets.UTF_8))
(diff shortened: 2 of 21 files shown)


components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/KeySharedPreferences.kt
```diff
@@ -0,0 +1,129 @@
...
+import java.nio.charset.StandardCharsets
+import java.security.GeneralSecurityException
+
+private const val BASE_64_FLAGS = Base64.URL_SAFE or Base64.NO_PADDING
+
+/**
+ * A wrapper around [SharedPreferences] which encrypts contents on supported API versions (23+).
+ * Otherwise, this simply delegates to [SharedPreferences].
+ *
+ * @param context Application context, used for accessing [SharedPreferences].
```
ekager: Can we reword this slightly because getApplicationContext is a thing in Android and this could be confusing

components/lib/dataprotect/src/test/java/mozilla/components/lib/dataprotect/SecureAbove22PreferencesTest.kt
```diff
@@ -0,0 +1,102 @@
+package mozilla.components.lib.dataprotect
```
ekager: nit license

grigoryk:
Still need to test the upgrade scenario in an emulator. Here's what to do:
- run a pre-23 emulator, use the logins storage, put some passwords into it (via sync, as there's no "create" currently)
- then pull the logins.db file from the emulator, along with key_preferences_pre_m.xml;
  - note that logins.db is encrypted with the key stored in key_preferences_pre_m.xml
- run a new emulator, install fenix there, launch it; there should be no passwords in it
- copy logins.db and pref files over to the new emulator, re-start fenix.
- make sure logins created in old emulator are still accessible
- make sure that old pref files is empty on disk

components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/SecureAbove22Preferences.kt
```diff
@@ -0,0 +1,166 @@
...
+
+    /**
+     * Removes key/value pair from storage for the provided [key].
+     */
+    fun remove(key: String) {
+        prefs.edit().remove(key).apply()
+    }
+
+    @TargetApi(M)
+    private fun getStringM(key: String): String? {
```
ekager: Let's rename these to `getStringDecrypted` and `putStringEncrypted`

components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/SecureAbove22Preferences.kt
```diff
@@ -0,0 +1,166 @@
...
+
+    /**
+     * Copies over [String] preferences from [plaintextPrefs].
+     */
+    @VisibleForTesting
+    @Suppress("ApplySharedPref")
+    internal fun migratePrefs(plaintextPrefs: SharedPreferences) {
+        plaintextPrefs.all.forEach {
+            if (it.value is String) {
+                putString(it.key, it.value as String)
```
ekager: Don't we have to use `putStringM` here for it to be migrated :D

ekager:
Currently failing on `> Task :samples-dataprotect:compileDebugKotlin FAILED
[task 2019-11-05T02:28:01.542Z] e: /builds/worker/checkouts/src/samples/dataprotect/src/main/java/org/mozilla/samples/dataprotect/MainActivity.kt: (29, 13): Too many arguments for public constructor SecureAbove22Preferences(context: Context) defined in mozilla.components.lib.dataprotect.SecureAbove22Preferences`

components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/SecureAbove22Preferences.kt
ekager: I reviewed this and then did it but undo it if u dont like it @grigoryk

components/concept/sync/src/main/java/mozilla/components/concept/sync/Sync.kt
```diff
@@ -42,6 +42,24 @@ data class SyncAuthInfo(
     val tokenServerUrl: String
 )
 
+/**
+ * An extension of [SyncableStore] that can be locked/unlocked using an encryption key.
+ */
+interface LockableStore : SyncableStore {
+    /**
+     * Unlocks underlying storage using [encryptionKey]. Storage must be [ensureLocked] after use.
```
pocmo: nit: If we always should call `ensureLocked()` after use then could we design this API to always do that. e.g. something like:

```Kotlin
store.unlocked(encryptionKey) {
  // here is is unlocked and after this block it gets locked automatically
}
```

which internally does something like
```
try {
  store.ensureUnlocked(encryptionKey)
  block(store)
} finally {
  store.ensureLocked()
}
```

Just to prevent misuse? But I do not know enough about how this is used to know whether that makes sense.

components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/KeyUtils.kt
```diff
@@ -0,0 +1,27 @@
...
+
+package mozilla.components.lib.dataprotect
+
+import java.security.SecureRandom
+
+/**
+ * @param keyStrength The strength of the generated key in bits
+ */
+@Suppress("MagicNumber")
+fun generateEncryptionKey(keyStrength: Int): String {
```
pocmo: nit: Does this need to be public? Is this something that app code is going to need to call? Such public functions without a scope are kind of noisy since they will show up in autocomplete everywhere.

components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/SecureAbove22Preferences.kt
```diff
@@ -0,0 +1,168 @@
...
+ *
+ * In rare circumstances (such as APK signing key rotation) a master key which protects this storage may be lost,
+ * in which case previously stored values will be lost as well. Applications are encouraged to instrument such events.
+ *
+ * @param context A [Context], used for accessing [SharedPreferences].
+ */
+class SecureAbove22Preferences(context: Context) {
+    companion object {
+        const val KEY_PREFERENCES_POST_M = "key_preferences_post_m"
+        const val KEY_PREFERENCES_PRE_M = "key_preferences_pre_m"
```
pocmo: nit: Are those public?

components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/SecureAbove22Preferences.kt
```diff
@@ -0,0 +1,168 @@
...
+
+        private const val BASE_64_FLAGS = Base64.URL_SAFE or Base64.NO_PADDING
+    }
+
+    private val keystore: Keystore? = if (Build.VERSION.SDK_INT < M) {
+        null
+    } else {
+        Keystore(context.packageName)
+    }
```
pocmo: nit: There are a lot of `if build version < m else` blocks in this class. Those make it quite hard to follow and mixing that with nullability (`keystore!!`) makes it easy to introduce errors. The support libraries run into this case a lot where they have to support different implementations for different API levels. They solve this nicely by using delegation. In this case they'd delegate internally to something like SecurePreferences21 (for 21+) and SecurePreferences23 (for 23+). Here this would make this code much easier to read and avoid some nullability issues.

An example:
* NotificationCompat: https://chromium.googlesource.com/android_tools/+/refs/heads/master/sdk/sources/android-25/android/support/v4/app/NotificationCompat.java#939
* NotificationCompatApi21: https://chromium.googlesource.com/android_tools/+/refs/heads/master/sdk/sources/android-25/android/support/v4/app/NotificationCompatApi21.java
* ...

components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/SecureAbove22Preferences.kt
```diff
@@ -0,0 +1,168 @@
...
+                logger.error("IllegalArgumentException exception: ", error)
+                null
+            } catch (error: GeneralSecurityException) {
+                logger.error("Decrypt exception: ", error)
+                null
+            }
+
+            plain.let {
+                if (it != null) String(it, StandardCharsets.UTF_8) else null
+            }
```
pocmo: Kotlin nit: For one I think `let` makes code pretty hard to read if you are just doing a null check on an immutable(!!) variable (that's just a fancy "if"). And here it's even more weird since it doesn't do more than translating `plain` to `it`.

There's no difference to this (since `plain` is immutable!):

```Kotlin
 if (plain != null) String(plain, StandardCharsets.UTF_8) else null
```

Since let also returns a value you can use it for the nullability check here though, especially since you want `null` in the else case (no need to move the null check inside):

```Kotlin
plain?.let { String(it, StandardCharsets.UTF_8) }   // this statement will already be null in the "else" case.
```

But I think a good old `if` is just more readable. `let` only seems to be helpful if you are dealing with mutable variables that are not in your control.

components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/SecureAbove22Preferences.kt
```diff
@@ -0,0 +1,168 @@
...
+ */
+class SecureAbove22Preferences(context: Context) {
+    companion object {
+        const val KEY_PREFERENCES_POST_M = "key_preferences_post_m"
+        const val KEY_PREFERENCES_PRE_M = "key_preferences_pre_m"
+
+        private const val BASE_64_FLAGS = Base64.URL_SAFE or Base64.NO_PADDING
+    }
+
+    private val keystore: Keystore? = if (Build.VERSION.SDK_INT < M) {
```
pocmo: nit: If you move those `Build.VERSION.SDK_INT` checks behind a method than you could mock that and test the whole migration flow without calling into `migratePrefs()` manually?

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/sync/WorkManagerSyncManager.kt
```diff
@@ -299,9 +301,24 @@ class WorkManagerSyncWorker(
             // Short-circuit if there are no configured stores.
             // Don't update the "last-synced" timestamp because we haven't actually synced anything.
             return Result.success()
-        }.also { stores ->
+        }
+
+        val passwordStore = syncableStores.entries.find { it.key == SyncEngine.Passwords }?.value
```
pocmo: nit: we could do a null-safe cast here, I guess (`as? LockableStore`)

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/sync/WorkManagerSyncManager.kt
```diff
@@ -299,9 +301,24 @@ class WorkManagerSyncWorker(
...
+        // If password storage is configured to be synced, we need to make sure it's decrypted during a sync.
+        passwordStore?.let {
+            val keySharedPreferences = SecureAbove22Preferences(context)
+
+            (it as LockableStore).ensureUnlocked(
+                keySharedPreferences.getString("passwords_key") ?: throw IllegalStateException(
+                    "'passwords_key' encryption key must be set if password sync is enabled"
+                )
+            )
+        }
```
pocmo: nit: Again my `let` overuse nit :)

I feel like this would be much more readable with just an ` if`. However here I wonder if we could just use something like:

```kotlin
passwordStore?.ensureUnlocked(
    SecureAbove22Preferences(context).getString("passwords_key")
    ?: ....
)
```

docs/changelog.md
```diff
@@ -55,6 +55,15 @@ permalink: /changelog/
...
 
+* **lib-dataprotect**
+  * Added new `SecureAbove22Preferences` helper class, which is an encryption-aware wrapper for `SharedPreferences`. Only actually encrypts stored values when running on API23+.
+
+* **service-firefox-accounts**
+  * Support for keeping `SyncEngine.Passwords` engine unlocked during sync.
+
+* **concept-sync**
+  * Added new `LockableStore` to facilitate syncing of "lockable" stores (such as `SyncableLoginsStore`).
+
```
pocmo: Do I understand this correctly that this is not breaking anything existing (e.g. Lockwise will just continue to work without any migration)? :)

docs/changelog.md
```diff
@@ -55,6 +55,15 @@ permalink: /changelog/
 * **feature-prompts**
   * Adds support for Web Share API using `ShareDelegate`.
 
+* **lib-dataprotect**
+  * Added new `SecureAbove22Preferences` helper class, which is an encryption-aware wrapper for `SharedPreferences`. Only actually encrypts stored values when running on API23+.
+
```
pocmo: What's our plan to QA this? This seems like something that will be rather painful if it doesn't work on some configurations. Will be do that in Fenix once the actual feature is available?

grigoryk:
Another network issue in TC.
```
[task 2019-11-05T05:56:43.328Z] FAILURE: Build failed with an exception.
[task 2019-11-05T05:56:43.328Z] 
[task 2019-11-05T05:56:43.328Z] * What went wrong:
[task 2019-11-05T05:56:43.328Z] Execution failed for task ':concept-push:packageDebugResources'.
[task 2019-11-05T05:56:43.328Z] > Could not resolve all files for configuration ':concept-push:_internal_aapt2_binary'.
[task 2019-11-05T05:56:43.328Z]    > Could not download aapt2-linux.jar (com.android.tools.build:aapt2:3.4.1-5326820)
[task 2019-11-05T05:56:43.328Z]       > Could not get resource 'https://dl.google.com/dl/android/maven2/com/android/tools/build/aapt2/3.4.1-5326820/aapt2-3.4.1-5326820-linux.jar'.
[task 2019-11-05T05:56:43.328Z]          > Connection reset
```

docs/changelog.md
grigoryk: Yeah, not a breaking change.

components/concept/sync/src/main/java/mozilla/components/concept/sync/Sync.kt
grigoryk: That is much better. Fixed.

components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/SecureAbove22Preferences.kt
grigoryk: Re-wrote the class following this pattern, and it's _much_ better now. Thanks!

grigoryk:
@pocmo thanks for feedback, updated version is looking much better!

grigoryk:
bors r=pocmo

PR #4973, opened 2019-11-05
Fix changelog: dataprotect changes in wrong release

grigoryk:
#4958 didn't land in time for 20, but its changelog entries went into 20.


pocmo:
Rebased patch, fixed conflict and pushed to PR.

bors r+

PR #4977, opened 2019-11-05
Closes #4969: Provide API to observe browser state changes on all tabs

### LINKED ISSUE #4969: browser-store: Provide API to observe specific state changes on all tabs
This came up when porting the `WindowFeature` where we need to be able to observe `WindowRequests` on all tabs (not just the selected tab) e.g. a background tab could issue a `window.close` request. We currently have a workaround which I want to get rid of.

We discussed adding a `filterChanged` method to our Flow extensions. This looks promising as it would allow us to collect only tabs for which the provided predicate has changed e.g. only the tabs that have new window requests.

csadilek:
This allows us to map a flow to a list (e.g. to a list of tabs as in `state.tabs`) and `collect` individual elements for which a specific attribute has changed (e.g. only tabs for which `windowRequest` has changed).

@pocmo this is pretty much as discussed, relying on the new `flatMap` flow operators. I did verify that this works and allows me to get rid of the workaround, but curious what you think. I also need to sleep over this :). In general, I think it does make sense that we have a `ifChanged` for flows of states, and `filterChanged` for flows of list of states.

### DIFF
--- a/components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlinx/coroutines/flow/Flow.kt
+++ b/components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlinx/coroutines/flow/Flow.kt
@@ -5,7 +5,9 @@
 package mozilla.components.support.ktx.kotlinx.coroutines.flow
 
 import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.asFlow
 import kotlinx.coroutines.flow.filter
+import kotlinx.coroutines.flow.flatMapConcat
 
 /**
  * Returns a [Flow] containing only values of the original [Flow] that have changed compared to
@@ -29,7 +31,7 @@ fun <T> Flow<T>.ifChanged(): Flow<T> = ifChanged { it }
  * Original Flow: "banana", "bus", "apple", "big", "coconut", "circle", "home"
  * Mapped: b, b, a, b, c, c, h
  * Returned Flow: "banana", "apple", "big", "coconut", "home"
- * ``
+ * ```
  */
 fun <T, R> Flow<T>.ifChanged(transform: (T) -> R): Flow<T> {
     var observedValueOnce = false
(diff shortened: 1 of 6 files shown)


components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlinx/coroutines/flow/Flow.kt
```diff
@@ -47,6 +50,34 @@ fun <T, R> Flow<T>.ifChanged(transform: (T) -> R): Flow<T> {
...
+/**
+ * Returns a [Flow] containing only changed elements of the lists of the original [Flow].
+ *
+ * Example:
+ * ```
+ * Block: x -> x[0]  // Map to first character of input
+ * Original Flow: list("banana"), list("bandanna", "bus"), list("apple", "big", "coconut", "circle"), list("home")
+ * Mapped: ("b"), ("b", "b"), ("a", "b", "c", "c"), ("h")
+ * Returned Flow: "banana", "bus", "apple", "coconut", "circle", "home"
+ * ```
```
pocmo: Even though we came up with this method together: Just reading the example I have no clue what this does. :) It definitely took a while to understand what's going on.

pocmo: I wonder if we can come up with a better example that is at the same time concise.

components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlinx/coroutines/flow/Flow.kt
```diff
@@ -47,6 +50,34 @@ fun <T, R> Flow<T>.ifChanged(transform: (T) -> R): Flow<T> {
...
+ */
+@FlowPreview
+fun <T, R> Flow<List<T>>.filterChanged(transform: (T) -> R): Flow<T> {
```
pocmo: Interesting, you flatten internally and go directly from `Flow<List<T>>` to `Flow<T>`. Originally I was expecting this to return the filtered `Flow<List<T>>` and then the app/component code can decide whether to use the list or flatten. However right now I am not sure if that will be a valid use case. Anyhow, I guess we can always adapt later.

components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlinx/coroutines/flow/Flow.kt
csadilek: @pocmo OK, I'll spend time today to find a better example and adding more comments at least to explain why we should observe a change.

components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlinx/coroutines/flow/Flow.kt
csadilek: Yes, I had started with that, but this made the consuming code easier. Maybe it's too "magical" and we should add another step to the API to flatten. Probably worth revisiting once we have more use cases...

components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlinx/coroutines/flow/Flow.kt
csadilek: @pocmo I've updated this now with more tests and two examples with more details. Can you take another look?

components/feature/tabs/src/test/java/mozilla/components/feature/tabs/WindowFeatureTest.kt
```diff
@@ -30,6 +31,7 @@ import org.mockito.Mockito.spy
 import org.mockito.Mockito.verify
 
 @RunWith(AndroidJUnit4::class)
+@FlowPreview
```
pocmo: Can we control this via a compiler argument (like `ExperimentalCoroutinesApi`) in build.gradle? This will avoid that this annotation ripples up all the way through app code.

components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlinx/coroutines/flow/Flow.kt
```diff
@@ -47,6 +50,62 @@ fun <T, R> Flow<T>.ifChanged(transform: (T) -> R): Flow<T> {
...
+    var lastMappedValues: List<R>? = null
+    return flatMapConcat { values ->
+        val lastMapped = lastMappedValues
+        val changed = if (lastMapped == null) {
+            values
+        } else {
+            values.filterIndexed { index, value ->
+                lastMapped.size <= index || lastMapped[index] !== transform(value)
+            }
```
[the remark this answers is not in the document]
csadilek: @pocmo yes, I think we need to care about this, because otherwise a tab could get added and eventually have a window request `w3` (or any other change we're observing) and we wouldn't notify it because the value wouldn't have changed relative to our last mapped values. So, if there's no value at the index of previous last mapped values, we have to emit as well.

components/feature/tabs/src/test/java/mozilla/components/feature/tabs/WindowFeatureTest.kt
csadilek: Ah yes, good point. I've added this to `build.gradle` now!

components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlinx/coroutines/flow/Flow.kt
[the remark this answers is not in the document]
csadilek: OK, pushed another update now!

csadilek:
bors r=pocmo

Issue #4983, opened 2019-11-06
Download the image icon for the add-on items in the add-on management UI in sample browser


Amejia481:
Right now, we have a fixed icon in the add-on management UI. We have to fetch the image using the approach mentioned on https://github.com/mozilla-mobile/android-components/issues/4867#issuecomment-549486733

![image](https://user-images.githubusercontent.com/773158/68305875-6d699b80-0076-11ea-9804-24301bed0472.png)

PR #4997, opened 2019-11-07
Issue #4983: Add code for migrating Gecko related files.

pocmo:
For now the recommendation is to only remove the prefs.js file since it may contain incompatible prefs.

The tricky part here is that this code **has to** run **before** Gecko gets initialized. Since that happens quite early this will have to happen immediately at app start before any other migration code runs - and *blocking* so that Gecko does not get initialized before that.

In my tests I created a second `FennecMigrator` right in the application object that I ran blocking:

```Kotlin
        val migrator = FennecMigrator.Builder(applicationContext, crashReporter)
            .migrateGecko()
            .build()

        migrator.migrateBlocking()
```

That worked quite well with the current approach and since it is blocking there's no interference / race with the following migration.

@grigoryk What do you think?

---

### Pull Request checklist

### After merge


pocmo:
I am going to write some tests for that now.

components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt
```diff
@@ -384,4 +412,18 @@ class FennecMigrator private constructor(
...
+    private fun migrateGecko(): Result<Unit> {
+        if (profile == null) {
+            crashReporter.submitCaughtException(IllegalStateException("Missing Profile path"))
+            return Result.Failure(IllegalStateException("Missing Profile path"))
+        }
+
+        return try {
+            GeckoMigration.migrate(profile.path)
+        } catch (e: Exception) {
+            Result.Failure(e)
```
grigoryk: `crashReporter.submitCaughtException` here as well?

components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt
```diff
@@ -245,6 +262,16 @@ class FennecMigrator private constructor(
         return runMigrationsAsync(migrationsToRun)
     }
 
+    /**
+     * Performs the configured data migration blocking the current thread until it has been completed.
+     *
+     * @return A [MigrationResults], describing which migrations were performed and if they succeeded.
+     */
+    @CheckResult
+    fun migrateBlocking(): MigrationResults {
```
grigoryk: I think I'd prefer to just do `runBlocking { migrateAsync.await() }` in Fenix instead. Not sure how much value this method really adds.

It'll give us more control over what's going closer to where we actually care about _how_ this runs (befoe gecko init in apps's onCreate).

components/support/migration/src/main/java/mozilla/components/support/migration/MigrationResultsStore.kt
```diff
@@ -56,6 +56,7 @@ internal class MigrationResultsStore(context: Context) : SharedPreferencesCache<
                 Migration.History.javaClass.simpleName -> Migration.History
                 Migration.Bookmarks.javaClass.simpleName -> Migration.Bookmarks
                 Migration.OpenTabs.javaClass.simpleName -> Migration.OpenTabs
+                Migration.Gecko.javaClass.simpleName -> Migration.Gecko // Bzzzp!
```
grigoryk: Great comment ;)

components/support/migration/src/main/java/mozilla/components/support/migration/GeckoMigration.kt
```diff
@@ -0,0 +1,40 @@
...
+internal object GeckoMigration {
+    /**
+     * Tries to migrate Gecko related internal files in [profilePath].
+     */
+    fun migrate(
+        profilePath: String
+    ): Result<Unit> {
+        // The prefs file contains gecko prefs that have been changed by the user (about:config) or
+        // code. Since those may be incompatible with GeckoView and/or app code, we decided to just
+        // reset them. GeckoView will recreated the file with defaults if needed.
```
grigoryk: s/recreated/recreate

components/support/migration/src/main/java/mozilla/components/support/migration/GeckoMigration.kt
```diff
@@ -0,0 +1,40 @@
...
+ * Helper for migrating Gecko related internal files.
+ */
+internal object GeckoMigration {
+    /**
+     * Tries to migrate Gecko related internal files in [profilePath].
+     */
+    fun migrate(
+        profilePath: String
+    ): Result<Unit> {
+        // The prefs file contains gecko prefs that have been changed by the user (about:config) or
```
grigoryk: Maybe also worth noting in a comment here that GeckoView will just happily inherit whatever is there, so our "migration" really is just about removing that prefs file.

components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt
```diff
@@ -384,4 +412,18 @@ class FennecMigrator private constructor(
...
+
+    @SuppressWarnings("TooGenericExceptionCaught")
+    private fun migrateGecko(): Result<Unit> {
+        if (profile == null) {
+            crashReporter.submitCaughtException(IllegalStateException("Missing Profile path"))
+            return Result.Failure(IllegalStateException("Missing Profile path"))
+        }
+
+        return try {
+            GeckoMigration.migrate(profile.path)
```
grigoryk: In every other migration we have debug log printouts before starting, after finishing migraton. Let's add them here as well.

components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt
pocmo: `migrateHistory`, `migrateBookmarks` and `migrateOpenTabs` don't do this for the generic exception case. Should they?

components/support/migration/src/main/java/mozilla/components/support/migration/MigrationResultsStore.kt
pocmo: Heh, yeah, this caused a crash at runtime when missing and the compiler didn't notify me about that since it doesn't know how to make this when statement include all cases (like for enums or sealed classes). Also using `simpleName` is a bit scary here since that can easily change and then breaks at runtime when reading an older state from disk.

components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt
pocmo: The code that runs the migrations logs all failures. Maybe we should report all failure exceptions from that code?

pocmo:
PR updated.

bors r=grigoryk

pocmo:
bors retry

pocmo:
bors retry

PR #5001, opened 2019-11-07
Closes #5000: about links should not trigger external app

### LINKED ISSUE #5000: about: links should not trigger external app
Currently if content opens `about:blank` we end up sending a `VIEW` intent. We should exclude `about:` as we do for `javascript:`.

csadilek:
Added test code for our blacklisting functionality as well, which was missing.


csadilek:
bors retry

Issue #4999, opened 2019-11-07
GeckoInputStream read() method not throwing IOException


BranescuMihai:
This bug is related to [6434](https://github.com/mozilla-mobile/fenix/issues/6434) from Fenix.

I narrowed the problem down to the copyInChunks() method from AbstractFetchDownloadService. Here, I tested with Fenix, Reference Browser and samples-browser from a-c. From my tests I observed that samples-browser, which uses the [okio](https://github.com/square/okio) source as an inputStream, returns the appropriate IOException when WiFi connection is interrupted mid-download, while Fenix and Reference Browser, which both use the GeckoInputStream instead, return -1 (which is considered as EOF, so a success response).

As per [java documentation](https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#read-byte:A-), an input stream implementation should throw an IOException "if the first byte cannot be read for any reason other than the end of the file", so this behaviour should be fixed in GeckoInputStream.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-4316)

sblatz:
Wow, good work for finding this, @BranescuMihai! 

Is this something that we need to change in GV, then, @cpeterso?

cpeterso:
I filed a new GV bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1594905

The GV fix sounds easy. How urgent is the Fenix Wi-Fi download bug?

sblatz:
@cpeterso I would say the fix is pretty urgent. It's currently allowing users to get in a state where they will crash the app (file looks to be downloaded, they press open, it crashes). I'd like to get this in sooner than later especially if it's an easy fix :)

sblatz:
@csadilek is there a reason Fenix & Sample browser are not in sync on which `InputStream` they're using? It seems like this could lead to other inconsistencies in functionality. This is why this issue slipped past me in development 😕

csadilek:
@sblatz yes, sample browser is using `HttpUrlConnectionClient`, not the `GeckoViewFetchClient`. 

We already have different build variants for `engine-system` (webview) and `engine-gecko` and can override individual components. Since this was a footgun, and we both tested with the sample browser when working on this, let's fix it right away. Here's a PR: https://github.com/mozilla-mobile/android-components/pull/5003

cpeterso:
snorp says this is an A-C bug, not a GV bug:

[...]

PR #5003, opened 2019-11-07
Use GeckoViewFetchClient for Gecko variants in sample browser

csadilek:
Let's use the GeckoViewFetch client for our gecko engine build variants. This hurt us when testing downloads: https://github.com/mozilla-mobile/android-components/issues/4999#issuecomment-551288072


samples/browser/src/geckoBeta/java/org/mozilla/samples/browser/Components.kt
```diff
@@ -26,4 +27,6 @@ class Components(applicationContext: Context) : DefaultComponents(applicationCon
             WebCompatFeature.install(it)
         }
     }
+
+    override val client by lazy { GeckoViewFetchClient(applicationContext) }
```
pocmo: ⚡️ The clients should use the same runtime we pass to `GeckoEngine`.

pocmo: Ah, haha, just saw the follow-up PR :)

PR #5005, opened 2019-11-08
sample-browser: Use same Gecko runtime for engine and client

csadilek:
@jonalmeida a quick follow-up to https://github.com/mozilla-mobile/android-components/pull/5003. We'll want to use the same runtime for `engine` and `client`. Noticed that on my way home, but you were too fast :).


jonalmeida:
bors r=jonalmeida

Issue #4956, opened 2019-11-04
Enable round-trip encryption tests


grigoryk:
Currently they're "ignored" (both in dataprotect Keystore and in the new KeySharedPreferences) due to "mac check in GCM failed".

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-4207)

csadilek:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1795959

Change performed by the [Move to Bugzilla add-on](https://addons.mozilla.org/en-US/firefox/addon/move-to-bugzilla/).

PR #5011, opened 2019-11-08
Closes #5004: Improve SecureAbove22Preferences API

### LINKED ISSUE #5004: SecureAbove22Preferences API isn't component-friendly
In #4958 we landed `SecureAbove22Preferences`, which makes it easy to store data securely on API levels which support it.

However, the way that API is structured, it's essentially global memory - any consumer can affect key/value pairs of any other consumer, clear all of the storage, etc.

What we should have landed is an API that's closer to what SharedPreferences has - a "scoped" storage identified by a name. This way different consumers can have their own silo'd storage layers. It's still possible for clashes to occur, but it's much more manageable.

grigoryk:
This changes `SecureAbove22Preferences` to allow isolating its instances by a name. Entries in an instance with name "a" won't be affected by (or visible to) an instance with name "b", and vice versa.

Migration logic for the encrypted version was simplified as well. It's now expressed in terms of basic `SecureAbove22Preferences` operations.

Lack of test coverage for the secure version is starting to become a problem here, so let's address #4956 soon.

Another change that's present here is a new method on `GlobalSyncableStoreProvider`: `configureKeyStorage`. It takes an instance of `SecureAbove22Preferences`, which is then used by the sync worker to obtain an encryption key for passwords storage, if it's configured.

The contract here is that if password syncing is enabled, then a "passwords" key must be available from the configured `SecureAbove22Preferences` instance.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/SecureAbove22Preferences.kt
+++ b/components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/SecureAbove22Preferences.kt
@@ -11,12 +11,18 @@ import android.content.SharedPreferences
 import android.os.Build
 import android.os.Build.VERSION_CODES.M
 import android.util.Base64
-import androidx.annotation.VisibleForTesting
 import mozilla.components.support.base.log.logger.Logger
 import java.nio.charset.StandardCharsets
 import java.security.GeneralSecurityException
 
 private interface KeyValuePreferences {
+    /**
+     * Retrieves all key/value pairs present in the store.
+     *
+     * @return A [Map] containing all key/value pairs present in the store.
+     */
+    fun all(): Map<String, String>
+
     /**
      * Retrieves a stored [key]. See [putString] for storing a [key].
      *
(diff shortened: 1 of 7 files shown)


components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/SecureAbove22Preferences.kt
```diff
@@ -181,6 +186,7 @@ private class SecurePreferencesImpl23(context: Context) : KeyValuePreferences {
      * which stores it in system's secure storage layer exposed via [AndroidKeyStore].
      */
     private fun generateManagedKeyIfNecessary() {
+        // Do we need to check this on every access, or just during instantiation? Is the overhead here worth it?
```
ekager: We could probably just do this during init because it seems unlikely/impossible that you could lose this key mid app run ?

grigoryk: That's what I'm hoping for. Let's address this in a follow-up.

grigoryk:
bors r=ekager

grigoryk:
Filed https://github.com/mozilla-mobile/android-components/issues/5018 to think about moving `keystore.available` check into `init`.

PR #5020, opened 2019-11-08
For #4999: Adds download failure condition for less than content length

sblatz:
---

I'm unsure how to add a test for this. It's adding a small incremental change to the download failed condition. The only way I could think to test is if there's a way to send a mock response that only let's it read a few bytes? Unsure how to do that 😅 

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
+++ b/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
@@ -73,7 +73,7 @@ abstract class AbstractFetchDownloadService : Service() {
         var job: Job? = null,
         var state: DownloadState,
         var currentBytesCopied: Long = 0,
-        var status: DownloadJobStatus,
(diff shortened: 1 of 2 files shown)


components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -264,6 +264,12 @@ abstract class AbstractFetchDownloadService : Service() {
             useFileStream(newDownloadState, isResumingDownload) { outStream ->
                 copyInChunks(downloadJobs[download.id]!!, inStream, outStream)
             }
+
+            downloadJobs[download.id]?.let {
```
csadilek: `copyInChunks` loops as long as `status` is `ACTIVE`. So, we break out of the loop when `pausing` the download. However, with this change, we may set the `status` to failed for paused downloads if the job isn't cancelled fast enough. I think we need to add an additional check here to see if the download was paused. We also need to make sure changes / writes to `status` are visible across threads.

csadilek:
bors retry

PR #5047, opened 2019-11-13
Fixes #5046: Updates guessFileName to use uniqueFIleName

### LINKED ISSUE #5046: [Bug] Downloading files overrides older versions of that file
## Steps to reproduce
1. Navigate to www.thinkbroadband.com/download
2. Click to download any file (let's say 5mb, because it's faster)
3. Wait until it has downloaded
4. Click to download that file again

### Expected behavior
The download popup should display the name followed by the number of same files. Example: 5mb (1)
Also in the download folder, it should display the file followed by the number in parenthesis, and not override the same file with each download

### Actual behavior
The download popup shows the same name, regardless of how many times you download. Also, every time you download the file, it overrides the file with the same name in the Downloads folder

* Fenix version: 
• Develop from 13/11

### Notes
[Video with issue](https://drive.google.com/file/d/10ANoROFo7GooE1OUno5FGLz9jqWJsBgQ/view?usp=sharing)

sblatz:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
+++ b/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
@@ -566,7 +566,8 @@ class SystemEngineView @JvmOverloads constructor(
     internal fun createDownloadListener(): DownloadListener {
         return DownloadListener { url, userAgent, contentDisposition, mimetype, contentLength ->
             session?.internalNotifyObservers {
-                val fileName = DownloadUtils.guessFileName(contentDisposition, url, mimetype)
+
+                val fileName = DownloadUtils.guessFileName(contentDisposition, null, url, mimetype)
                 val cookie = CookieManager.getInstance().getCookie(url)
                 onExternalResource(url, fileName, contentLength, mimetype, cookie, userAgent)
             }
--- a/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt
+++ b/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt
@@ -5,7 +5,6 @@
 package mozilla.components.feature.downloads
 
 import android.content.Context
-import android.os.Environment
 import android.widget.Toast
 import androidx.annotation.ColorRes
 import androidx.annotation.VisibleForTesting
@@ -31,7 +30,6 @@ import mozilla.components.support.base.feature.PermissionsFeature
 import mozilla.components.support.ktx.android.content.appName
 import mozilla.components.support.ktx.android.content.isPermissionGranted
 import mozilla.components.support.ktx.kotlinx.coroutines.flow.ifChanged
-import java.io.File
 
 /**
  * Feature implementation to provide download functionality for the selected
@@ -96,15 +94,7 @@ class DownloadsFeature(
                 .collect { state ->
                     val download = state.content.download
                     if (download != null) {
-                        // Update the file name to ensure it doesn't collide with one already on disk]
-                        val downloadWithUniqueName = download.fileName?.let {
-                            download.copy(fileName = uniqueFileName(
-                                Environment.getExternalStoragePublicDirectory(download.destinationDirectory),
-                                it
-                            ))
-                        } ?: download
-
-                        processDownload(state, downloadWithUniqueName)
+                        processDownload(state, download)
                     }
                 }
         }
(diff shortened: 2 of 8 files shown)


components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -566,7 +566,8 @@ class SystemEngineView @JvmOverloads constructor(
     internal fun createDownloadListener(): DownloadListener {
         return DownloadListener { url, userAgent, contentDisposition, mimetype, contentLength ->
             session?.internalNotifyObservers {
-                val fileName = DownloadUtils.guessFileName(contentDisposition, url, mimetype)
+
+                val fileName = DownloadUtils.guessFileName(contentDisposition, null, url, mimetype)
```
sblatz: Passing `null` in for `destinationDirectory` here worries me a little bit. This means that in WebView implementations, we wouldn't be able to guess the file name? I'm not sure I understand the significance of `SystemEngineView` nor the purpose of this `DownloadListener`. 

Can you help me understand if this is an issue or if there's a work around I can use here without the `destinationDirectory`?

components/support/utils/src/main/java/mozilla/components/support/utils/DownloadUtils.kt
```diff
@@ -40,17 +42,46 @@ object DownloadUtils {
...
+
+        return destinationDirectory?.let {
+            uniqueFileName(Environment.getExternalStoragePublicDirectory(destinationDirectory), fileName)
+        } ?: fileName
+    }
+
+    /**
+     * Checks if the file exists so as not to overwrite one already in the destination directory
+     */
+    fun uniqueFileName(directory: File, fileName: String): String {
```
csadilek: This would behave weirdly for files that don't have an extension if I read this correctly. So, say a file named "test" would result in `test`, `(1).test`, `(2).test` instead of `test`, `test (1)`, `test (2)`, because `substringAfterLast` will return the original value if the delimiter is not found?

csadilek: Since this is a public util method now, we should add some basic tests for it :). Here's an example test for the case I describe above: https://github.com/csadilek/android-components/commit/b1eddbdd5be4c25e841684db6a51a4eea9897fb7

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
csadilek: >  I'm not sure I understand the significance of SystemEngineView nor the purpose of this DownloadListener.

Yes, to support downloads in our system-engine (webview).

You're not changing any behaviour here though as you fallback to the same filename that would've been used before. Can we just use `Environment.DIRECTORY_DOWNLOADS` here as default?

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt
```diff
@@ -98,7 +98,7 @@ class DownloadsFeature(
                     if (download != null) {
                         // Update the file name to ensure it doesn't collide with one already on disk]
                         val downloadWithUniqueName = download.fileName?.let {
-                            download.copy(fileName = uniqueFileName(
+                            download.copy(fileName = DownloadUtils.uniqueFileName(
```
csadilek: @sblatz I missed this in my first review, but @pocmo found it: We shouldn't change the state of the session here. 

The `DownloadState` describes a request to download a file with the name provided by the engine, which is different from the unique file name dependent on the file system on the device i.e. two sessions could download the same file and we can only determine the local file name once we write to disk. So, we should determine the unique file name in the download service when we perform the download.

csadilek: Ah, scratch the updating download state part, as we're only copying here to forward an updated object to the download manager / service. I think it would still be cleaner to do this in the service instead (when we open / start writing to the file)?

sblatz: @csadilek I opted for creating a unique name for the file in `DownloadFeature` because I want the `DownloadPrompt` to display the correct name. That means even **before** a user is writing to that file, we've determined a unique name for the file. This matches the behavior of desktop.

components/support/utils/src/main/java/mozilla/components/support/utils/DownloadUtils.kt
sblatz: Added, thanks for figuring out this test, @csadilek! 😄

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
sblatz: I'm going to opt for keeping this `null`. I don't want to make it `Environment.DIRECTORY_DOWNLOADS` because that will be another place someone has to change if that default directory ever changes. It'd be weird if these were out of sync :)

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt
sblatz: Actually I'm wrong here. This is something that desktop does *not* do. It just seemed to make sense to me that we would show the user the file as its actual name in the prompt.

sblatz: I have moved this logic to the service right before writing the file 😄

sblatz: Hmm. Seems moving the logic to the service causes file corruption when the files conflict. Can we keep it inside of the `DownloadsFeature` and file that as a follow-up?

csadilek: OK, looks good now!

sblatz:
bors retry

sblatz:
bors retry

PR #5065, opened 2019-11-14
Closes #4935: Tracking protection state not updated on browser session

### LINKED ISSUE #4935: Tracking protection is not working without an AppRequestInterceptor in fenix
We have to investigate why tracking protection is not working without [reapplying the tracking protection policy per session.](https://github.com/mozilla-mobile/fenix/blob/a7e139da8eb74722aaab1aae7fde927e0e16e97d/app/src/main/java/org/mozilla/fenix/AppRequestInterceptor.kt#L23)

csadilek:
The cause of this problem is a bit of a gap in our design. When a `GeckoEngineSession` is created we enable (or disable) tracking protection based on settings, and also call `notifyObservers` to make sure the corresponding browser session (`trackerBlockingEnabled` field) is updated. However, when the `GeckoEngineSession` is created no browser session is linked to it yet and therefore there's no observer. The notification is simply lost and the browser session is never updated.

We looked into a few ways to work around this e.g. setting the field in [onLocationChange](https://github.com/mozilla-mobile/android-components/blob/master/components/browser/session/src/main/java/mozilla/components/browser/session/engine/EngineObserver.kt#L41) instead, as we do for `title`, `url`, `webAppManifest`. This would be a bit of a hack though as we'd be doing it way too often, or in an unrelated method / event.

So, how about this idea instead: We introduce a new `notifyAtLeastOneObserver` (or similar) method which queues up the notification if there's no observer and sends it to the first observer that's registered. Wdyt?

@pocmo @Amejia481

### DIFF
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -189,7 +189,7 @@ class GeckoEngineSession(
         if (!enabled) {
             disableTrackingProtectionOnGecko()
         }
-        notifyObservers { onTrackerBlockingEnabledChange(enabled) }
+        notifyAtLeastOneObserver { onTrackerBlockingEnabledChange(enabled) }
     }
 
     /**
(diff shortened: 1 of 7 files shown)


csadilek:
bors r=pocmo, Amejia481

PR #5066, opened 2019-11-14
Add setDynamicToolbarMaxHeight API

hiikezoe:
---
To make dynamic toolbar behavior compatible with other browsers, we did introduce a new API to specify the maximum height of the dynamic toolbar in [bug 1586144](https://bugzilla.mozilla.org/show_bug.cgi?id=1586144).

To be honest, I don't know what we should do for engine-gecko-beta or engine-gecko in this kind of situation (i.e. introducing a new API in gecko-nightly).

### Pull Request checklist

- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) **I am totally unsure whether dynamic toolbar is accessibility friendly or not.**

### After merge

### DIFF
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineView.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineView.kt
@@ -141,6 +141,10 @@ class GeckoEngineView @JvmOverloads constructor(
         currentGeckoView.setVerticalClipping(clippingHeight)
     }
 
+    override fun setDynamicToolbarMaxHeight(height: Int) {
+        // no-op
+    }
+
     override fun captureThumbnail(onFinish: (Bitmap?) -> Unit) {
         val geckoResult = currentGeckoView.capturePixels()
         geckoResult.then({ bitmap ->
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineView.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineView.kt
@@ -141,6 +141,10 @@ class GeckoEngineView @JvmOverloads constructor(
         currentGeckoView.setVerticalClipping(clippingHeight)
     }
 
+    override fun setDynamicToolbarMaxHeight(height: Int) {
+        currentGeckoView.setDynamicToolbarMaxHeight(height)
+    }
+
     override fun captureThumbnail(onFinish: (Bitmap?) -> Unit) {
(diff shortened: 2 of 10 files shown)


hiikezoe:
This is the first step to make the toolbar compatible with other browsers.

We still need to [set the max height](https://github.com/hiikezoe/reference-browser/commit/0499b60b6205509838884610735944ee0d366a9d) and also need to [call setVerticalClipping with minus values](https://github.com/hiikezoe/android-components/commit/1c2533d53302d1bc32d29a74df485265910450a8).  For the place to set the height, I am not sure it's the best place to do that, but I couldn't think of any other places.

r?@pocmo

hiikezoe:
> This looks good to me.
> 
> Are you planning to open another PR for the `EngineViewBottomBehavior` change that you linked to? I assume we want both of those changes to go into the next release.

No, at this moment because the `EngineViewBottomBehavior` change should be merged altogether with calling this new setDynamicToolbarMaxHeight in appropriate place in reference-browser and in fenix, I have [a commit to call it in reference-browser](https://github.com/hiikezoe/reference-browser/commit/0499b60b6205509838884610735944ee0d366a9d), but I am not convinced it's called in the best place.

pocmo:
> No, at this moment because the EngineViewBottomBehavior change should be merged altogether with calling this new setDynamicToolbarMaxHeight in appropriate place in reference-browser and in fenix

Yeah, but the behavior change has to happen in AC. Doesn't it make sense to land both in AC and then update the apps with the next release? Otherwise we may end up with only one of the changes in the next AC release - which is not really usable?

Your RB patch is fine for now. It works if the toolbar has a fixed size in the layout. In situations where the toolbar does not have a size yet (if determined in the "measure" passes) then this would set it to 0 currently. However in both RB and Fenix we know the size of the toolbar (56dp) and can definitely just set it once in the fragment after creating the views.

hiikezoe:
> > No, at this moment because the EngineViewBottomBehavior change should be merged altogether with calling this new setDynamicToolbarMaxHeight in appropriate place in reference-browser and in fenix
> 
> Yeah, but the behavior change has to happen in AC. Doesn't it make sense to land both in AC and then update the apps with the next release? Otherwise we may end up with only one of the changes in the next AC release - which is not really usable?

Right. It's totally useless.

> Your RB patch is fine for now. It works if the toolbar has a fixed size in the layout. In situations where the toolbar does not have a size yet (if determined in the "measure" passes) then this would set it to 0 currently. However in both RB and Fenix we know the size of the toolbar (56dp) and can definitely just set it once in the fragment after creating the views.

Thanks!  I pushed PRs respectively. https://github.com/mozilla-mobile/android-components/pull/5107 and https://github.com/mozilla-mobile/reference-browser/pull/954.  Hope it gets landed.

PR #5068, opened 2019-11-15
For mozilla-mobile/fenix#5540: update copy for 'no internet' error

Baron-Severin:
---

### Pull Request checklist

This only makes changes to copy.

### After merge

### Screenshot
Screen Shot 2019-11-14 at 4 59 05 PM

### DIFF
--- a/components/browser/errorpages/src/main/java/mozilla/components/browser/errorpages/ErrorPages.kt
+++ b/components/browser/errorpages/src/main/java/mozilla/components/browser/errorpages/ErrorPages.kt
@@ -33,7 +33,7 @@ object ErrorPages {
             .use { it.readText() }
             .replace("%pageTitle%", context.getString(R.string.mozac_browser_errorpages_page_title))
             .replace("%backButton%", context.getString(R.string.mozac_browser_errorpages_page_go_back))
-            .replace("%button%", context.getString(R.string.mozac_browser_errorpages_page_refresh))
+            .replace("%button%", context.getString(errorType.buttonRes))
             .replace("%messageShort%", context.getString(errorType.titleRes))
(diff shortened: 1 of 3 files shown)


components/browser/errorpages/src/main/java/mozilla/components/browser/errorpages/ErrorPages.kt
```diff
@@ -45,7 +45,8 @@ object ErrorPages {
  */
 enum class ErrorType(
     @StringRes val titleRes: Int,
-    @StringRes val messageRes: Int
+    @StringRes val messageRes: Int,
+    @StringRes val buttonRes: Int = R.string.mozac_browser_errorpages_page_refresh
```
Amejia481: nit: `refreshButtonRes`

Baron-Severin:
Closing this in favor of making changes in Fenix, because we couldn't find a good solution to determining whether or not the user was online where `ErrorType` was created.  Summary (from Slack) of alternatives we investigated below.

"""
We need context to find out whether or not the internet is available, so we can construct the proper ErrorType.  Unfortunately, it’s currently being created in SystemEngineSession and GeckoEngineSession, the latter of which doesn’t have any reference to context (at least that I can find)

Introducing context to that class sounds immediately unacceptable
we do have a context reference in createErrorPage, where the error type is used.  It would be possible to check there and substitute a new ErrorType as appropriate.  But that is confusing code, couples the error creation to this specific use case, subtly breaks any future use cases, and arguably chips away at type safety (an ErrorType is suddenly not necessary a complete, valid error).  Also unacceptable (edited) 
We could update the functions that return ErrorType  to instead return Context -> ErrorType , but that’s also a very confusing change to make, just to support some updated copy
"""

PR #5082, opened 2019-11-17
Closes #4993: Add tag to notifications for WebNotification

### LINKED ISSUE #4993: Add tag to notifications for WebNotification
When calling `notify` we should add a tag for the WebNotifications to namespace them correctly and avoid ID collisions.

rocketsroger:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
+++ b/components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
@@ -23,6 +23,7 @@ import mozilla.components.support.base.log.logger.Logger
 import java.lang.UnsupportedOperationException
 
 private const val NOTIFICATION_CHANNEL_ID = "mozac.feature.webnotifications.generic.channel"
+private const val NOTIFICATION_TAG_ID = "mozac.feature.webnotifications.generic.tag"
 
 /**
  * Feature implementation for configuring and displaying web notifications to the user.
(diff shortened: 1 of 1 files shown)


components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
```diff
@@ -66,7 +67,7 @@ class WebNotificationFeature(
         ensureNotificationGroupAndChannelExists()
 
         notificationIdMap[webNotification.tag]?.let {
-            notificationManager?.cancel(it)
+            notificationManager?.cancel(NOTIFICATION_TAG_ID, it)
```
pocmo: Should this be a tag that includes the domain so that we can have multiple concurrent notifications?

jonalmeida: @pocmo you should still be able to have multiple notifications from what I understand [in the API](https://developer.android.com/reference/android/app/NotificationManager#notify(int,%20android.app.Notification)). The tag + id pair, is what makes the notification unique to app which we were missing. In Fenix, we've been using a still tag with different IDs for multiple tabs received, which seems to work out fine (minus the `Random()` bit 😅 ).

rocketsroger: As discussed.  Since the tag + id pair will be uniquely identifying each notification.  I'll go ahead and mark this resolved.  Thanks,

jonalmeida:
> bors r+

You can r+ your own PR? 🤔 

That seems like a bug.

rocketsroger:
> That seems like a bug.

I was told r+ is a quick way of doing it.  Maybe I'm not supposed to.

rocketsroger:
> > That seems like a bug.
> 
> I was told r+ is a quick way of doing it. Maybe I'm not supposed to.

Looks like its OK from bors documentation:

bors r=[list] | Same as r+, but   the “reviewer” in the commit log will be recorded as the user(s) given as the argument.

PR #5083, opened 2019-11-17
Closes #5063: better exception reporting during migrations

### LINKED ISSUE #5063: IllegalStateException in FennecMigrator$runMigrationsAsync
Three different stacktraces with multiple reports:

https://crash-stats.mozilla.com/report/index/c2929c40-54fa-4140-8e99-93bad0191112

https://crash-stats.mozilla.com/report/index/6935abd3-27e1-4805-9f7b-dc39c0191112

https://crash-stats.mozilla.com/report/index/881be396-038a-49f2-aa99-3121b0191112

grigoryk:
Patch 1:
>
    We are logging exceptions when no profile is found, which is actually a legit case.
    There will be no profile on a fresh install, for example.
    
    This patch makes these kinds of exceptions more precise/useful: we'll only send a report
    in case there's a profile but a missing dbPath. We'll now also send an exception in case
    of an IOException while parsing profiles.

Patch 2:
>
    Submit unexpected exceptions during migrations

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt
+++ b/components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt
@@ -105,7 +105,7 @@ class FennecMigrator private constructor(
         private var coroutineContext: CoroutineContext = Executors.newSingleThreadExecutor().asCoroutineDispatcher()
 
         private var fxaState = File("${context.filesDir}", "fxa.account.json")
-        private var fennecProfile = FennecProfile.findDefault(context)
+        private var fennecProfile = FennecProfile.findDefault(context, crashReporter)
         private var browserDbName = "browser.db"
 
         /**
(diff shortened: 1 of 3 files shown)


components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt
```diff
@@ -334,6 +371,9 @@ class FennecMigrator private constructor(
             logger.debug("Migrated bookmarks.")
             Result.Success(Unit)
         } catch (e: Exception) {
+            crashReporter.submitCaughtException(
+                FennecMigratorException.MigrateBookmarksException(e::class.qualifiedName)
```
pocmo: Instead of the `name` (String) parameter... doesn't it make more sense to set the `cause` (Exception) parameter so that we get a nice stacktrace including the stacktrace of the exception that happened here? Much more helpful and doesn't require reflection? :)

grigoryk: Done; i've decided to not block on the PII thing right now, and deal with it more generally in a follow-up: https://github.com/mozilla-mobile/android-components/issues/5095

PR #5107, opened 2019-11-20
Use minus offset for vertical clipping.

hiikezoe:
After bug 1586144 the area which is initially covered by the dynamic toolbar
is outside of the ICB so that we need to use minus offsets.

https://bugzilla.mozilla.org/show_bug.cgi?id=1586144

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/engine-gecko-beta/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineViewTest.kt
+++ b/components/browser/engine-gecko-beta/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineViewTest.kt
@@ -80,9 +80,9 @@ class GeckoEngineViewTest {
         val engineView = GeckoEngineView(context)
         engineView.currentGeckoView = mock()
 
-        engineView.setVerticalClipping(42)
+        engineView.setVerticalClipping(-42)
 
-        verify(engineView.currentGeckoView).setVerticalClipping(42)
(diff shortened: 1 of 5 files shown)


hiikezoe:
> +1
> 
> @hiikezoe This only works with GV Nightly currently, right? Do we have any plans to uplift this to beta (what Fenix relases are currently shipping with)?

Right. It only works on nightly, and I am not going to uplift corresponding change in mozilla-central.

> If not, should we maybe move this change into `GeckoEngineView` so that we can add `-` only for Nightly? Not sure if its worth it though since the whole dynamic toolbar is not really working with beta anyways.

Very interesting. You mean we can tweak the given value [here in GeckoEngineView.setVerticalClipping](https://github.com/mozilla-mobile/android-components/blob/d80c05ee84b255ccaada47a0e0134c37eee8053c/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineView.kt#L145), right?  Unfortunately, it's impossible because the change we need is 's/dependency.height - dependency.translationY.toInt()/-dependency.translationY.toInt()/', we don't have `dependency.height` in the GeckoEngineView function. maybe we can because it's the height of the toolbar, but to do that, it would be messy.

pocmo:
Alright, then let's get that in as-is.

bors r+

PR #5116, opened 2019-11-20
Closes #5113: Take User to Market Place If Intent Package is Not Installed

### LINKED ISSUE #5113: Take User to Market Place If Intent Package is Not Installed
When a site tries to launch an app intent with package name specified and the application is not installed, bring the user to the market place so they can install it.

rocketsroger:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/app-links/src/main/java/mozilla/components/feature/app/links/AppLinkRedirect.kt
+++ b/components/feature/app-links/src/main/java/mozilla/components/feature/app/links/AppLinkRedirect.kt
@@ -5,7 +5,6 @@
 package mozilla.components.feature.app.links
 
 import android.content.Intent
-import android.content.pm.ResolveInfo
 
 /**
  * Data class for the external Intent or fallback URL a given URL encodes for.
@@ -13,7 +12,7 @@ import android.content.pm.ResolveInfo
 data class AppLinkRedirect(
     val appIntent: Intent?,
     val fallbackUrl: String?,
-    val info: ResolveInfo? = null
+    val marketplaceIntent: Intent?
 ) {
     /**
      * If there is a third-party app intent.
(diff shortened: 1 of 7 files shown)


rocketsroger:
> I wonder if this could lead to unexpected behavior or even exploitation since that allows every website to open up google play?

This could be something I need to look into.  This implementation will open Google Store (or another market place) if the intent have a specific package name in it.  We then open the Google Store page to the particular page for that package.  There is a possibility if the market place app(Google Play or otherwise) have vulnerability then this is a possible attack.

rocketsroger:
I think this behaviour is safe due to the following reasons:

1. We create the market intent by using the prefix + package name: 
private const val MARKET_INTENT_URI_PACKAGE_PREFIX = "market://details?id=".

2. The user have to click on the link on the intent and the app have to be not installed to open the market app. (Google Play in most cases)

rocketsroger:
> I'm less concerned about the market itself. I wonder if this is the best user experience? There's a difference between an app I installed myself opening or a website randomly pushing me into the store to install some app I do not want. Is that something that was explicitly requested?

This was not something explicitly requested.  I noticed this issue when I was reproducing another issue.  The problem here when an intent is targeting a particular app, if we don't have it, we'll leave it for GeckoView to open which will then show an confusing `Unknown Protocol` error.  I then tried Fennec and it was bringing the user to the market app.  Hence this change to match the behavior on both Fennec and Chrome.

> Furthermore I wonder about the fallback behavior. A website can define a fallback URL that will get opened if the intent cannot be satisfied. If we always jump to the store then we break that behavior. Is that a concern?

In my test case, there is a possible way to ignore the targeted app and open in Fenix.  However, this is not always the case, sometimes an intent is understood by a particular application.

> I think we'd need at least a way to let the app enable/disable this behavior - then I'd feel better landing this? However if it turns out to not be needed then I would suggest to remove it again. :)

I do blame myself not clarify this in the PR comments.  It is controlled by the same switch/toggle for launching third party apps.  If its on, we will redirect to market app.  If its off, we will not handle the intent.  By default the switch/toggle is off.

rocketsroger:
Thanks! I'll be more clear with the why and the details of my future PR so it doesn't require pulling teeth to get the context.  😅

rocketsroger:
bors r=pocmo

Issue #4791, opened 2019-10-18
Allow opening browser/page action default popup


csadilek:
We need to be able to open the configured default popup when tapping on browser / page actions. For now we can open it in a new tab, similar to Fennec.

PR #5121, opened 2019-11-21
Closes #5090: Wire up GeckoView action delegate for BrowserActions

### LINKED ISSUE #5090: Wire up GeckoView action delegate for BrowserActions
This is a follow-up for https://github.com/mozilla-mobile/android-components/issues/4469 now that the GV API has landed.

csadilek:
This follows the same pattern as content/background messaging, with the only difference that we have to be able to register action handlers for each extension on every session. The latter is the reason I moved the session-specific action out of the `EngineObserver` into our web extension module. Ultimately, we want to get rid of`EngineObserver` anyway once we fully migrated to browser-state.

I am also going to add a browser action to our sample-browser extension, but I will do this as part of #4791, as this PR is already pretty big.

We will get new GV API for installing extensions i.e different calls for built-in and third-party extensions. For now I've also introduced a `supportActions` boolean, similar to our `allowContentMessaging` to be able to configure per extension if we need to hook up action delegates. We really don't need that for our current extensions, so it makes sense not to register all these handlers.

### DIFF
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/webextension/GeckoWebExtension.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/webextension/GeckoWebExtension.kt
@@ -5,13 +5,19 @@
 package mozilla.components.browser.engine.gecko.webextension
 
 import mozilla.components.browser.engine.gecko.GeckoEngineSession
+import mozilla.components.browser.engine.gecko.await
 import mozilla.components.concept.engine.EngineSession
+import mozilla.components.concept.engine.webextension.ActionHandler
+import mozilla.components.concept.engine.webextension.BrowserAction
 import mozilla.components.concept.engine.webextension.MessageHandler
 import mozilla.components.concept.engine.webextension.Port
 import mozilla.components.concept.engine.webextension.WebExtension
+import mozilla.components.support.base.log.logger.Logger
 import org.json.JSONObject
 import org.mozilla.geckoview.GeckoResult
+import org.mozilla.geckoview.GeckoSession
 import org.mozilla.geckoview.WebExtension as GeckoNativeWebExtension
+import org.mozilla.geckoview.WebExtension.Action as GeckoNativeWebExtensionAction
 
 /**
  * Gecko-based implementation of [WebExtension], wrapping the native web
@@ -21,13 +27,16 @@ class GeckoWebExtension(
     id: String,
     url: String,
     allowContentMessaging: Boolean = true,
+    supportActions: Boolean = false,
     val nativeExtension: GeckoNativeWebExtension = GeckoNativeWebExtension(
         url,
         id,
         createWebExtensionFlags(allowContentMessaging)
     ),
     private val connectedPorts: MutableMap<PortId, Port> = mutableMapOf()
-) : WebExtension(id, url) {
+) : WebExtension(id, url, supportActions) {
+
+    private val logger = Logger("GeckoWebExtension")
 
     /**
      * Uniquely identifies a port using its name and the session it
@@ -142,6 +151,64 @@ class GeckoWebExtension(
             connectedPorts.remove(portId)
         }
     }
+
+    /**
+     * See [WebExtension.registerActionHandler].
+     */
+    override fun registerActionHandler(actionHandler: ActionHandler) {
+        if (!supportActions) {
+            logger.error("Attempt to register default action handler but browser and page " +
+                "action support is turned off for this extension: $id")
+            return
+        }
+
+        val actionDelegate = object : GeckoNativeWebExtension.ActionDelegate {
+
+            override fun onBrowserAction(
+                ext: GeckoNativeWebExtension,
+                // Session will always be null here for the global default delegate
+                session: GeckoSession?,
+                action: GeckoNativeWebExtensionAction
+            ) {
+                actionHandler.onBrowserAction(this@GeckoWebExtension, null, action.toBrowserAction())
+            }
+        }
+
+        nativeExtension.setActionDelegate(actionDelegate)
+    }
+
+    /**
+     * See [WebExtension.registerActionHandler].
+     */
+    override fun registerActionHandler(session: EngineSession, actionHandler: ActionHandler) {
+        if (!supportActions) {
+            logger.error("Attempt to register action handler on session but browser and page " +
+                "action support is turned off for this extension: $id")
+            return
+        }
+
+        val actionDelegate = object : GeckoNativeWebExtension.ActionDelegate {
+
+            override fun onBrowserAction(
+                ext: GeckoNativeWebExtension,
+                geckoSession: GeckoSession?,
+                action: GeckoNativeWebExtensionAction
+            ) {
+                actionHandler.onBrowserAction(this@GeckoWebExtension, session, action.toBrowserAction())
+            }
+        }
+
+        val geckoSession = (session as GeckoEngineSession).geckoSession
+        geckoSession.setWebExtensionActionDelegate(nativeExtension, actionDelegate)
+    }
+
+    /**
+     * See [WebExtension.hasActionHandler].
+     */
+    override fun hasActionHandler(session: EngineSession): Boolean {
+        val geckoSession = (session as GeckoEngineSession).geckoSession
+        return geckoSession.getWebExtensionActionDelegate(nativeExtension) != null
+    }
 }
 
 /**
@@ -172,3 +239,14 @@ private fun createWebExtensionFlags(allowContentMessaging: Boolean): Long {
         GeckoNativeWebExtension.Flags.NONE
     }
 }
+
+private fun GeckoNativeWebExtensionAction.toBrowserAction() =
+    BrowserAction(
+        title ?: "",
+        enabled ?: true,
+        { size -> icon?.get(size)?.await() },
+        badgeText ?: "",
+        badgeTextColor ?: 0,
+        badgeBackgroundColor ?: 0,
+        { click() }
+    )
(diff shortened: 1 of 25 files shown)


components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/webextension/GeckoWebExtension.kt
```diff
@@ -142,6 +151,64 @@ class GeckoWebExtension(
...
+
+        nativeExtension.setActionDelegate(actionDelegate)
+    }
+
+    /**
+     * See [WebExtension.registerActionHandler].
+     */
+    override fun registerActionHandler(session: EngineSession, actionHandler: ActionHandler) {
+        if (!supportActions) {
+            logger.error("Attempt to register action handler on session but browser and page " +
```
Amejia481: Nice 👍

csadilek: Hehe, yers, I forgot that before :). I just want to turn this off by default for built-in extensions, we can turn it on for third-party extensions, once we get the different methods.

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/WebExtensionToolbarAction.kt
```diff
@@ -38,16 +45,32 @@ open class WebExtensionToolbarAction(
 
         rootView.setBackgroundResource(backgroundResource)
         padding?.let { rootView.setPadding(it) }
+
+        rootView.addOnAttachStateChangeListener(object : View.OnAttachStateChangeListener {
```
Amejia481: Nice ✨ 
Similar to the [pattern used in](https://github.com/mozilla-mobile/android-components/blob/6fef6179e1f4bfe59359b86efa7035116b32237a/components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt#L179), it looks like this is a common need, maybe in the future we can have an utility extension for loading an image  async for everybody to use :)

components/feature/toolbar/src/test/java/mozilla/components/feature/toolbar/WebExtensionToolbarFeatureTest.kt
```diff
@@ -94,22 +94,20 @@ class WebExtensionToolbarFeatureTest {
 
         val browserAction = BrowserAction(
             title = "title",
-            icon = mock(),
+            loadIcon = { mock() },
```
Amejia481: nit: Does worth it adding extra test coverage for the icon loading section and the `iconJob?.cancel()`  ?

components/feature/toolbar/src/test/java/mozilla/components/feature/toolbar/WebExtensionToolbarTest.kt
```diff
@@ -4,53 +4,81 @@
...
 import org.junit.runner.RunWith
 import org.mockito.Mockito.verify
 
 @RunWith(AndroidJUnit4::class)
 class WebExtensionToolbarTest {
+    private val testDispatcher = TestCoroutineDispatcher()
+
+    @Before
+    @ExperimentalCoroutinesApi
+    fun setUp() {
```
Amejia481: nit: I think we could replace this block with [MainCoroutineRule](https://github.com/mozilla-mobile/android-components/blob/9b8e27603efea171c8f3764de9a25cc3f1e84ec9/components/support/test/src/main/java/mozilla/components/support/test/rule/MainCoroutineRule.kt#L28)

components/feature/toolbar/src/test/java/mozilla/components/feature/toolbar/WebExtensionToolbarTest.kt
```diff
@@ -4,53 +4,81 @@
...
 
         val action = WebExtensionToolbarAction(browserAction) {}
-
         action.bind(view)
+        action.iconJob?.joinBlocking()
+        testDispatcher.advanceUntilIdle()
+
+        val iconCaptor = argumentCaptor<BitmapDrawable>()
+        verify(imageView).setImageDrawable(iconCaptor.capture())
+        assertEquals(icon, iconCaptor.value.bitmap)
```
Amejia481: nit: I'm not sure if it worths adding extra test coverage for the `iconJob?.cancel()`, what do you think?

components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
```diff
@@ -63,10 +89,60 @@ object WebExtensionSupport {
...
+                    }
             }
         })
     }
 
+    /**
+     * Observes the provided store to register session-specific [ActionHandler]s
+     * for all installed extensions on newly added sessions.
+     */
+    private fun registerActionHandlersForNewSessions(store: BrowserStore) {
```
Amejia481: Could you help me to clarify `newly added sessions` term mean? Does that mean session that are just linked to its respective `engineSession`, right?

csadilek: Sure, we're already capturing the existing sessions when the extension is installed. So, for every session added after (once it's engine session is created / linked) we have to do this.

Amejia481: Thanks 👍

components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
```diff
@@ -63,10 +89,60 @@ object WebExtensionSupport {
...
+                        action: BrowserAction
+                    ) {
+                        store.dispatch(WebExtensionAction.UpdateBrowserAction(webExtension.id, action))
+                    }
+                })
+
+                // Register action handler for all existing engine sessions on the new extension
+                store.state.tabs
+                    .forEach {
+                        val session = it.engineState.engineSession
```
psymoon: nit:
```						
it.engineState.engineSession?.let { session ->
	registerSessionActionHandler(webExtension, session, SessionActionHandler(store, it.id))
}
```

components/feature/toolbar/src/test/java/mozilla/components/feature/toolbar/WebExtensionToolbarTest.kt
csadilek: Ah, nice! Updated.

components/feature/toolbar/src/test/java/mozilla/components/feature/toolbar/WebExtensionToolbarTest.kt
csadilek: Yeah, might as well. Added a test now to make sure we cancel loading the icon.

components/concept/engine/src/main/java/mozilla/components/concept/engine/webextension/BrowserAction.kt
```diff
@@ -21,8 +20,7 @@ import android.graphics.drawable.Drawable
 data class BrowserAction(
     val title: String,
     val enabled: Boolean,
-    val icon: Drawable,
-    val uri: String,
+    val loadIcon: suspend (Int) -> Bitmap?,
```
psymoon: Is this related to AMO add-on icon?

components/concept/engine/src/main/java/mozilla/components/concept/engine/webextension/WebExtension.kt
```diff
@@ -81,6 +84,53 @@ abstract class WebExtension(
...
      */
     abstract fun disconnectPort(name: String, session: EngineSession? = null)
+
+    /**
+     * Registers an [ActionHandler] for this web extension. The handler will
+     * be invoked whenever browser and page action defaults change. To listen
+     * for session-specific overrides see registerActionHandler(
+     * EngineSession, ActionHandler).
+     */
+    abstract fun registerActionHandler(actionHandler: ActionHandler)
```
psymoon: This appears to be registered for the lifecycle of an application. Is there a reason behind that?

components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
csadilek: 👍 let's save that line :)

csadilek: Updated.

components/concept/engine/src/main/java/mozilla/components/concept/engine/webextension/BrowserAction.kt
csadilek: This icon is specified by extension itself either in the manifest or via `setIcon`:
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/browser_action
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/browserAction/setIcon

components/concept/engine/src/main/java/mozilla/components/concept/engine/webextension/WebExtension.kt
csadilek: So, it's registered globally on the webextension, yes. Later I think we'll remove it when extensions are disabled or uninstalled. We're not doing any UI work in these handlers, just making sure the store is up-to-date.

The UI components observing on the store (e.g. your `WebExtensionsToolbarFeature`) are all tied to a lifecycle (e.g. the BrowserFragment). So, we should be good here.

csadilek:
bors r=psymoon,Amejia481

PR #5123, opened 2019-11-21
Deprecate StorageSync class

eoger:
It is not used by any of our products anymore (fenix, lockwise, fxr).

### DIFF
--- a/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/sync/StorageSync.kt
+++ b/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/sync/StorageSync.kt
@@ -20,6 +20,7 @@ import mozilla.components.support.base.observer.ObserverRegistry
 /**
  * Orchestrates data synchronization of a set of [SyncableStore]-s.
  */
+@Deprecated("Use SyncManager instead.")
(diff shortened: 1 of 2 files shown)


rocketsroger:
Since we're removing a public facing class, we need to follow the steps here:
https://mozac.org/contributing/deprecating

grigoryk:
We may (?) have external consumers of this, so let's definitely follow the documented deprecation process.

This also gives us time to consider other cleanup that we can perform as part of this - e.g. what else do we no longer need now that SyncManager is a thing?

eoger:
Updated with deprecation notice.

eoger:
@rocketsroger can you do it?

rocketsroger:
> @rocketsroger can you do it?

Sure, once the tests passes, I will trigger bors.

grigoryk:
Well, I should have been more clear... There's no `SyncManager` class exposed as a public API in A-C :) I'll update the deprecation notice.

rocketsroger:
> Well, I should have been more clear... There's no `SyncManager` class exposed as a public API in A-C :) I'll update the deprecation notice.

My mistake.  I assumed there is a `SyncManager` public API.

PR #5124, opened 2019-11-21
Closes #4882: Update NotificationIds To Allow Incrementing IDs

### LINKED ISSUE #4882: Update NotificationIds To Allow Incrementing IDs
Currently `NotificationIds` support only allows a tag to retrieve same notification ID every time its called with the same tag.  However, some component might need to get increment IDs but not currently used by another component.  Update implementation to allow retrieval of the next available ID using the same notification tag.

rocketsroger:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
+++ b/components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
@@ -19,11 +19,13 @@ import mozilla.components.browser.icons.BrowserIcons
 import mozilla.components.concept.engine.Engine
 import mozilla.components.concept.engine.webnotifications.WebNotification
 import mozilla.components.concept.engine.webnotifications.WebNotificationDelegate
+import mozilla.components.support.base.ids.NotificationIds
 import mozilla.components.support.base.log.logger.Logger
 import java.lang.UnsupportedOperationException
 
 private const val NOTIFICATION_CHANNEL_ID = "mozac.feature.webnotifications.generic.channel"
-private const val NOTIFICATION_TAG_ID = "mozac.feature.webnotifications.generic.tag"
+private const val NOTIFICATION_TAG_ID = "mozac.feature.webnotifications.generic.notification.tag"
+private const val PENDING_INTENT_TAG_ID = "mozac.feature.webnotifications.generic.pendingintent.tag"
 
 /**
  * Feature implementation for configuring and displaying web notifications to the user.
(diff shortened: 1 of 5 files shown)


components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
```diff
@@ -70,13 +70,13 @@ class WebNotificationFeature(
             notificationManager?.cancel(NOTIFICATION_TAG_ID, it)
         }
 
-        pendingRequestId++
-        notificationId++
+        val notificationId = NotificationIds.getNextIdForTag(context, NOTIFICATION_TAG_ID)
+        val pendingIntentId = NotificationIds.getNextIdForTag(context, PENDING_INTENT_TAG_ID)
```
pocmo: So we always need a new tag and never want to reuse one?

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
```diff
@@ -70,13 +70,13 @@ class WebNotificationFeature(
...
         notificationIdMap[webNotification.tag] = notificationId
 
         GlobalScope.launch(Dispatchers.IO) {
             val notification = nativeNotificationBridge.convertToAndroidNotification(
-                webNotification, context, NOTIFICATION_CHANNEL_ID, activityClass, pendingRequestId)
+                webNotification, context, NOTIFICATION_CHANNEL_ID, activityClass, pendingIntentId)
             notificationManager?.notify(NOTIFICATION_TAG_ID, notificationId, notification)
```
pocmo: Actually, can you help me understand this. Since the notification already comes with a unique tag - do we even need to provide an incrementing id? Isnt (tag, 1) already unique?

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
rocketsroger: Yes, in this case we need a new one.  Not the old one.

components/feature/webnotifications/src/main/java/mozilla/components/feature/webnotifications/WebNotificationFeature.kt
rocketsroger: You're correct.  By using the tag given to us through GeckoView, we can use the same notification ID (ex: 1) and everything still works as expected.  I will cancel this review and update the implementation to use the tag from GeckoView.

rocketsroger:
Won't fix the issue.   A better solution is available for WebNotifications.  This PR is not needed anymore.

PR #5125, opened 2019-11-21
Closes #4398: Add reader state to browser-state

### LINKED ISSUE #4398: Add reader state to browser-state
We currently have two flags in browser session:
readerable - whether or not the current page can be shown in reader mode
readerMode - whether or not reader mode is on

We need to bring these to browser-state for https://github.com/mozilla-mobile/android-components/issues/4397.

Alternatively, we could look for a solution to let features extend browser-state and define their own actions/reducers. We never quite liked that these fields are added to the session directly, but it seems to be the easiest solution.

csadilek:
Closing #4398 to bring in the state required for reader mode so we can port it over soon. Nothing new here, following the existing pattern, and as discussed in https://github.com/mozilla-mobile/android-components/issues/4398#issuecomment-531128780.

### DIFF
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/state/TabSessionState.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/state/TabSessionState.kt
@@ -16,8 +16,9 @@ import java.util.UUID
  * parent. The parent tab is usually the tab that initiated opening this
  * tab (e.g. the user clicked a link with target="_blank" or selected
  * "open in new tab" or a "window.open" was triggered).
- * @property extensionState a map of web extension ids and extensions, that contains the overridden
- * values for this tab.
+ * @property extensionState a map of web extension ids to extensions,
+ * that contains the overridden values for this tab.
+ * @property readerState the [ReaderState] of this tab.
  */
 data class TabSessionState(
     override val id: String = UUID.randomUUID().toString(),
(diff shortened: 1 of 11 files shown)


components/browser/state/src/main/java/mozilla/components/browser/state/state/TabSessionState.kt
```diff
@@ -16,32 +16,37 @@ import java.util.UUID
...
  */
 data class TabSessionState(
     override val id: String = UUID.randomUUID().toString(),
     override val content: ContentState,
     override val trackingProtection: TrackingProtectionState = TrackingProtectionState(),
     override val engineState: EngineState = EngineState(),
     val parentId: String? = null,
-    override val extensionState: Map<String, WebExtensionState> = emptyMap()
+    override val extensionState: Map<String, WebExtensionState> = emptyMap(),
+    val readerState: ReaderState = ReaderState()
```
pocmo: Small nit: I wonder if we should create some kind of `feature` sub-state that contains optional states that our features may add?

csadilek: Yes, was wondering about too, as we talked about this before, but wasn't sure. Let's punt on that and file a follow-up for grouping features together: https://github.com/mozilla-mobile/android-components/issues/5147

I will add a comment here as well.

csadilek:
bors r=pocmo

Issue #5007, opened 2019-11-08
Add Tabs support to SyncEngine


jonalmeida:
Adding an extra element to the `SyncEngine` and similar to the other engine types, to link it to the `WorkManagerSyncManager` and from there the Rust component.

Issue #4557, opened 2019-09-27
Adding new types of SyncEngines


grigoryk:
Follow-up to #4480.
Before we add new types of SyncEngines, think through the implications for the `declined` stuff in the a-s SyncManager. We need to make sure that user choice is respected, and that if engine is already present in the declined/accepted lists, we're behaving in a sensible way.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-3857)

jonalmeida:
Hmm.. I don't see anything actionable to do here since this issue was first reported, so I'm going to close it because it seems more like a "spike" ticket.

PR #5131, opened 2019-11-22
Add Tabs engine to the sync manager

eoger:
Fixes https://github.com/mozilla-mobile/android-components/issues/5007.
This is fairly simple PR as it doesn't even need a direct dependency on the published a-s `tabs` component.

### DIFF
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -27,7 +27,7 @@ object Versions {
     const val disklrucache = "2.0.2"
     const val leakcanary = "1.6.3"
 
-    const val mozilla_appservices = "0.42.2"
+    const val mozilla_appservices = "0.44.0"
 
     const val mozilla_glean = "21.2.0"
 
--- a/components/feature/push/src/test/java/mozilla/components/feature/push/RustPushConnectionTest.kt
+++ b/components/feature/push/src/test/java/mozilla/components/feature/push/RustPushConnectionTest.kt
@@ -14,6 +14,7 @@ import org.junit.Assert.assertNull
 import org.junit.Assert.assertTrue
 import org.junit.Ignore
 import org.junit.Test
+import org.mockito.ArgumentMatchers.nullable
 import org.mockito.Mockito.anyString
 import org.mockito.Mockito.verify
 
@@ -65,7 +66,7 @@ class RustPushConnectionTest {
             connection.subscribe("123", "")
         }
 
-        verify(api).subscribe(anyString(), anyString())
+        verify(api).subscribe(anyString(), anyString(), nullable(String::class.java))
     }
 
     @Test(expected = IllegalStateException::class)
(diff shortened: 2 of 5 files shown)


grigoryk:
Please see https://github.com/mozilla-mobile/android-components/issues/4557 - we need to make sure that this will interact correctly with the `declined` list.

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -27,7 +27,7 @@ object Versions {
     const val disklrucache = "2.0.2"
     const val leakcanary = "1.6.3"
 
-    const val mozilla_appservices = "0.42.2"
+    const val mozilla_appservices = "0.44.0"
```
jonalmeida: TODO me: run a smoke test against this upgrade to make sure native features are still working fine.

eoger:
@grigoryk would the a-s sync manager take care of that?

buildSrc/src/main/java/Dependencies.kt
jonalmeida: Tested with the sample FxA app. Seems fine.

jonalmeida:
@eoger just remember to put a PR up for Fenix to upgrade their a-s version when this lands.

grigoryk:
> @grigoryk would the a-s sync manager take care of that?

I'm not really sure - need to think it through. Maybe? That's why there's an issue filed :-) If you can think through the implications here, it'd be great. I can take a close look at this early next week, as well.

eoger:
> > @grigoryk would the a-s sync manager take care of that?
> 
> @eoger I'm not really sure - need to think it through. Maybe? That's why there's an issue filed :-) If you can think through the implications here, it'd be great. I can take a close look at this early next week, as well.

I did some digging around the code. This is what should happen when we add an engine:

- Before rust sync manager sync: `enabledChanges` [would not contain](https://searchfox.org/mozilla-mobile/rev/b56d507e9f715b69d94b324652e6b346de309812/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/sync/WorkManagerSyncManager.kt#365-379) our new engine, and [neither](https://searchfox.org/mozilla-mobile/rev/b56d507e9f715b69d94b324652e6b346de309812/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/sync/WorkManagerSyncManager.kt#381-384) `enginesToSync` (as it is an `intersection of stores for which we've set a 'handle' and those that are enabled.`)
- After the rust sync manager synced: If the user is connected on another device and disabled tabs, it would be in the resulting `declined` list, otherwise it would be in the `enabled` list [for the next sync](https://searchfox.org/mozilla-mobile/rev/b56d507e9f715b69d94b324652e6b346de309812/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/sync/WorkManagerSyncManager.kt#421-435) as `enabled = allEngines - declined`.

TL;DR: I believe we would respect the user's choice of declining the engine. Let me know if it makes sense!

Amejia481:
bors retry

grigoryk:
bors retry

grigoryk:
Okay, since this bumps the A-S version, let's just wait until Monday; otherwise we'll break Fenix builds for the weekend.

jonalmeida:
The last build seemed to have an intermittent test failure. Filed https://github.com/mozilla-mobile/android-components/issues/5213.

Issue #4879, opened 2019-10-29
Fennec migration: run in a foreground service


grigoryk:
Related to https://github.com/mozilla-mobile/android-components/issues/4872 - our migrations are potentially slow, which creates a large window of time for things to go wrong (OS killing us, user killing us, some bad interaction with Fenix occuring, etc).

To get an additional margin of safety, we could wrap our migration code in a foreground service.

pocmo:
@grigoryk Are you working on this? Otherwise I may pick that up. :)

pocmo:
Just landed in Fenix and should go into the next builds.

ebalazs-sv:
Verified as fixed on Release, Beta, Nightly 12/13 migration builds with Samsung Galaxy S8 (Android 9), Huawei Honor 8 (Android 7.0).
A notification is displayed while migration, and this contains:
warning sign Firefox Preview now
"Migrating" title
"Migrating user data." text
The notification disappears after the migration is done.

Issue #4872, opened 2019-10-28
Fennec migration: "status" observable interface


grigoryk:
Now that I'm trying out migrations against real profiles, it's obvious that they this stuff will take quite a long time to run.

From slack:
> Note the timing. It took 6 seconds to migrate history, then 5 seconds to migrate bookmarks, then finally we got to FxA migration which took 2-3 seconds

I'm pretty sure we'll have some kind of a special "first run after Fenix upgrade" UI, and part of that UI could indicate progress of our migrations as they execute.

We could expose an observable interface, which UI can use to display real-time migration state.

pocmo:
This also means we should be careful with the order of things. For example we should migrate the open tabs first, so that the user can use/see them asap.

pocmo:
@grigoryk I was thinking a bit about that and potential solutions. Right now I'm leaning towards using `lib-state` for that. We could have a migration-specific store - It's a proven pattern and it allows the background service to update the state and the UI to react to updates and render the state. Does this make sense to you?

pocmo:
The initial implementation for that landed. Let's iterate on that if needed.

ebalazs-sv:
This is in preparation for the upcoming UI and there is nothing observable until the UI lands. Due to this I am moving this to Done.

PR #5142, opened 2019-11-25
Issue #4879: Add service for running migration in background.

pocmo:
This is the AC part of the background service. Next step is to open the Fenix PR after this is in the next snapshot.

### DIFF
--- a/components/support/migration/src/main/java/mozilla/components/support/migration/AbstractMigrationService.kt
+++ b/components/support/migration/src/main/java/mozilla/components/support/migration/AbstractMigrationService.kt
@@ -1,0 +1,131 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.support.migration
+
+import android.app.NotificationChannel
+import android.app.NotificationManager
+import android.app.Service
+import android.content.Context
+import android.content.Intent
+import android.os.Build
+import android.os.IBinder
+import androidx.core.app.NotificationCompat
+import kotlinx.coroutines.MainScope
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.launch
+import mozilla.components.support.base.ids.NotificationIds
+import mozilla.components.support.base.log.logger.Logger
+
+private const val NOTIFICATION_TAG = "mozac.support.migration.notification"
+private const val NOTIFICATION_CHANNEL_ID = "mozac.support.migration.generic"
+
+// Non-final strings and icon.
+// https://github.com/mozilla-mobile/android-components/issues/5141
+private const val TEMPORARY_NOTIFICATION_TITLE = "Migrating"
+private const val TEMPORARY_NOTIFICATION_TEXT = "Migrating user data."
+private const val TEMPORARY_NOTIFICATION_CHANNEL_NAME = "Migration"
+private val temporaryNotificationIcon = R.drawable.mozac_lib_crash_notification
+
+/**
+ * Abstract implementation of a background service running a configured [FennecMigrator].
+ *
+ * An application using this implementation needs to extend this class and provide a
+ * [FennecMigrator] instance.
+ */
+abstract class AbstractMigrationService : Service() {
+    protected abstract val migrator: FennecMigrator
+
+    private val logger = Logger("MigrationService")
+    private val scope = MainScope()
+
+    override fun onCreate() {
+        super.onCreate()
+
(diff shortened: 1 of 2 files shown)


components/support/migration/src/main/java/mozilla/components/support/migration/AbstractMigrationService.kt
```diff
@@ -0,0 +1,131 @@
...
+        stopForeground(true)
+
+        // Now after the migration this notification channel is no longer needed and we can just
+        // remove it again.
+        removeChannel()
+
+        stopSelf()
+    }
+
+    private fun ensureChannelExists(): String {
```
grigoryk: I would expect this sort of thing to live elsewhere, but i see that we already have similar `ensureNotificationGroupAndChannelExists` methods elsewhere already... Anyway, that's cleanup for a later date.

components/support/migration/src/main/java/mozilla/components/support/migration/AbstractMigrationService.kt
```diff
@@ -0,0 +1,131 @@
...
+                logger.debug("Stopping migration service")
+                shutdown()
+            }
+        }
+    }
+
+    override fun onBind(intent: Intent?): IBinder? = null
+
+    private suspend fun migrate() {
+        val results = migrator.migrateAsync().await()
```
grigoryk: We'll need to emit a glean ping out of this; is that something we can do in a service? or would we need to communicate w/ the host app? It'd be great to not have to worry about making our results Parcelable...

But I guess we can deal with this later.

components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt
```diff
@@ -236,28 +239,52 @@ class FennecMigrator private constructor(
      * @return A deferred [MigrationResults], describing which migrations were performed and if they succeeded.
      */
     fun migrateAsync(): Deferred<MigrationResults> = synchronized(migrationLock) {
```
grigoryk: this could be `internal` now, right?

components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt
```diff
@@ -236,28 +239,52 @@ class FennecMigrator private constructor(
...
+    fun <T> startMigrationServiceIfNeeded(service: Class<T>) where T : AbstractMigrationService {
+        if (hasMigrationsToRun()) {
+            logger.debug("Has migrations to run. Starting service.")
+            context.startService(Intent(context, service))
+        } else {
+            logger.debug("No migrations to run. Not starting service.")
+        }
+    }
+
+    private fun getMigrationsToRun(): List<VersionedMigration> {
```
grigoryk: I'm a little worried that this is now not protected by the `migrationLock`; we _shouldn't_ have overlapping migrations running, of course...

components/support/migration/src/main/java/mozilla/components/support/migration/AbstractMigrationService.kt
pocmo: That should just work. We are already using Glean from some other services (e.g. media).

components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt
pocmo: The Gecko migration piece is still called from the app code directly.

components/support/migration/src/main/java/mozilla/components/support/migration/AbstractMigrationService.kt
pocmo: Yeah, we could think of a helper maybe. But in this case I hope this code will not stick around forever, so I like it to not spread too much.

components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt
pocmo: Yeah, overlapping migrations could cause a bunch of other issues.

It's private and still protected by the lock when called from `migrateAsync()`. I think I'll add the same protection to `startMigrationServiceIfNeeded()`.

pocmo: Or maybe not! If a migration is currently running then we do not want to block here - because that can take a while. 🤔

pocmo: I think I want to land as-is. Once we have #4872 it will be easier to communicate the current state.

pocmo:
bors r=grigoryk

PR #5183, opened 2019-11-28
Closes #4067: Fennec logins migration

### LINKED ISSUE #4067: Fennec logins & password data import
Sibling of #4066.

Fennec will soon grow an ability to share logins/password data with trusted applications, using the same mechanism we rely on for FxA auto-login. See https://bugzilla.mozilla.org/show_bug.cgi?id=1572550

This work tracks getting logins information Fennec's ContentProvider, and passing it over to a-s for processing.
Depending on how Fennec shapes up, this may be either a path to a database file (preferred), or a bulk set of logins/passwords.

Relevant a-s ticket: https://github.com/mozilla/application-services/pull/1472

grigoryk:
---

### Pull Request checklist

### After merge


grigoryk:
This PR doesn't expose `master password` as part of FennecMigrator's public API - but otherwise, it supports decryption of logins dbs protected with MP.

grigoryk:
Fenix PR that enables logins migration: https://github.com/mozilla-mobile/fenix/pull/6930

components/support/migration/src/main/java/mozilla/components/support/migration/FennecLoginsMigration.kt
```diff
@@ -0,0 +1,541 @@
...
+     * @param key4DbPath Path to key4 db.
+     * @return True if [masterPassword] is correct, false otherwise.
+     */
+    internal fun isMasterPasswordValid(masterPassword: String, key4DbPath: String): Boolean {
+        val db = SQLiteDatabase.openDatabase(key4DbPath, null, OPEN_READONLY)
+        return db.rawQuery("select item1, item2 from metadata where id = 'password'", null).use {
+            it.moveToNext()
+
+            val globalSalt = it.getBlob(it.getColumnIndexOrThrow("item1"))
+            val item2 = it.getBlob(it.getColumnIndexOrThrow("item2"))
```
csadilek: Can we add the details from your comment https://github.com/mozilla-mobile/android-components/issues/4067#issuecomment-558424220 to describe what item2 is?

components/support/migration/src/main/java/mozilla/components/support/migration/FennecLoginsMigration.kt
```diff
@@ -0,0 +1,541 @@
...
+                        timesUsed = timesUsed,
+                        timeCreated = timeCreated,
+                        timeLastUsed = timeLastUsed,
+                        timePasswordChanged = timePasswordChanged,
+                        usernameField = usernameField,
+                        passwordField = passwordField
+                    )
+                    records.add(fullRecord)
+                } catch (e: Exception) {
+                    crashReporter.submitCaughtException(e)
```
csadilek: Here we'll send a caught exception to Sentry/Socorro for every incorrect record. How likely are incorrect records? Should we combine into one exception instead?

components/support/migration/src/main/java/mozilla/components/support/migration/FennecLoginsMigration.kt
```diff
@@ -0,0 +1,541 @@
...
+        val globalSalt = getGlobalSalt(key4DbPath)
+        val db = SQLiteDatabase.openDatabase(key4DbPath, null, OPEN_READONLY)
+        val a11 = db.rawQuery("select a11, a102 from nssPrivate", null).use {
+            while (it.moveToNext()) {
+                val maybeRightKeyId = it.getBlob(it.getColumnIndexOrThrow("a102"))
+                if (maybeRightKeyId.contentEquals(keyId)) {
+                    return@use it.getBlob(it.getColumnIndexOrThrow("a11"))
+                }
+            }
+            null
```
csadilek: nit: `return@use` here as well to make it easier to read?

components/support/migration/src/main/java/mozilla/components/support/migration/FennecLoginsMigration.kt
```diff
@@ -0,0 +1,541 @@
...
+        return db.rawQuery("select item1 from metadata where id = 'password'", null).use {
+            it.moveToNext()
+            it.getBlob(it.getColumnIndexOrThrow("item1"))
+        }
+    }
+
+    /**
+     * Decrypts [ciphertext] using TDES.
+     */
+    private fun decryptTDes(key: ByteArray, iv: ByteArray, ciphertext: ByteArray): ByteArray {
```
csadilek: nit `descryptTDES`?

components/support/migration/src/main/java/mozilla/components/support/migration/FennecLoginsMigration.kt
csadilek: Actually, yes, I think we should move this out of the loop. `decrypt` could throw a `NoSuchAlgorithmException` and then we'd be logging this a lot...

grigoryk:
@csadilek addresses all of your feedback.

I've reached out to @mozkeeler for feedback, as well. Dana thinks this is basically fine (non-ideal as there are hand-rolled ASN.1 decoder bits, but alas). Main concern is that we should be careful about running this migration while Gecko is around. "NSS has historically had issues when anything else touches cert9.db/key4.db".
But, It will _probably_ be fine even if we overlap with NSS, somehow - "since it's read-only, the worst that should happen is some slowness".

grigoryk:
bors r=csadilek

PR #5223, opened 2019-12-03
Update to latest app-services, including metrics from logins component

rfk:
(Top-level bug comment edited for clarity)

This updates a-c to use the latest a-s release, which includes some new metrics from the logins component. There was a lot of back-end-forth about the implications of landing such a thing, which you can read about below. The resolution is that we have data-review+ for landing these metrics in Fenix and Lockwise, and any new consumers of the logins component will need to get data-review for logins-component metrics.

----

Over in https://github.com/mozilla/application-services/pull/2225 we're working on adding some basic performance and quality metrics to the appservices logins component, instrumented via Glean.  I'm happy with how that change has shaped in *in isolation*, but before we land it I want to understand whether it's going to work out as part of our broader component ecosystem.

Here is what I think it would look like to update android-components to this new version of the logins component. There are no new APIs, but there is a new behaviour that consumers need to be aware of, and it's probably important enough to be called out as a breaking change.

I tested this by building the sync-logins sample app (ref https://github.com/mozilla-mobile/android-components/pull/5222, https://github.com/mozilla/application-services/issues/2312) and confirm that it still worked, then adding some additional logging output in glean to confirmed that the expected metrics-gathering was taking place.  Notably, the sync-logins sample app does not actually submit any metrics via glean, so IIUC it would have been gathering the metrics but discarding them without submitting anything to Mozilla. I don't know whether they would have been gathered in memory, or written to disk; I should see if I can find that out.

@grigoryk @pocmo what's your gut reaction to this? Will it be workable as-is, or should we consider some more nuanced API to let consumers opt in or out of getting metrics-gathering behaviour from this component?

### DIFF
--- a/components/service/sync-logins/README.md
+++ b/components/service/sync-logins/README.md
@@ -15,6 +15,16 @@ The **Firefox Sync - Logins Component** provides a way for Android applications 
 
 ## Usage
 
+### Before using this component
+
+The `mozilla.appservices.logins` component collects telemetry using the [Glean SDK](https://mozilla.github.io/glean/).
+Applications that send telemetry via Glean *must ensure* they have received appropriate data-review following
+[the Firefox Data Collection process](https://wiki.mozilla.org/Firefox/Data_Collection) before integrating this component.
+
+Details on the metrics collected by the `mozilla.appservices.logins` component are available
+[here](https://github.com/mozilla/application-services/tree/master/docs/metrics/logins/metrics.md)
+
+
 ### Setting up the dependency
 
 Use Gradle to download the library from [maven.mozilla.org](https://maven.mozilla.org/) ([Setup repository](../../../README.md#maven-repository)):
@@ -78,7 +88,7 @@ When inserting records (e.g. creating records for use with `AsyncLoginsStorage.a
 
 The hostname this record corresponds to. It is an error to attempt to insert or update a record to have a blank hostname, and attempting to do so `InvalidRecordException` being thrown.
 
-#### `username: String? = null`
+#### `username: String = ""`
 
 The username associated with this record, which may be blank if no username is asssociated with this login.
 
@@ -124,11 +134,11 @@ A lower bound on the time that the `password` field was last changed in millisec
 
 This is a metadata field, and as such, is ignored by `AsyncLoginsStorage.add` and `AsyncLoginsStorage.update`.
 
-#### `usernameField: String? = null`
+#### `usernameField: String = ""`
 
 HTML field name of the username, if known.
 
-#### `passwordField: String? = null`
+#### `passwordField: String = ""`
 
 HTML field name of the password, if known.
 
@@ -153,8 +163,6 @@ The errors reported as "raw" `LoginsStorageException` are things like Rust panic
 Yes, sort of. This works, however due to the fact that `lock` closes the database connection, *all data is lost when the database is locked*. This means that doing so will result in a database with different behavior around lock/unlock than one stored on disk.
 
 That said, doing so is simple: Just create a `DatabaseLoginsStorage` with the path `:memory:`, and it will work. You may also use a [SQLite URI filename](https://www.sqlite.org/uri.html) with the parameter `mode=memory`. See https://www.sqlite.org/inmemorydb.html for more options and further information.
-
-Note that we offer a `MemoryLoginsStorage` class which doesn't come with the same limitations (however it cannot sync).
 
 ### How do I set a key for the `DatabaseLoginsStorage`?
 
--- a/components/service/sync-logins/build.gradle
+++ b/components/service/sync-logins/build.gradle
@@ -29,7 +29,11 @@ android {
 
 dependencies {
     // Parts of this dependency are typealiase'd or are otherwise part of this module's public API.
-    api Dependencies.mozilla_sync_logins
+    api(Dependencies.mozilla_sync_logins) {
+      // Use our own version of the Glean dependency,
+      // which might be different from the version declared by A-S.
+      exclude group: 'org.mozilla.components', module: 'service-glean'
+    }
     api Dependencies.mozilla_sync15
 
     // Types defined in concept-sync are part of this module's public API.
@@ -37,6 +41,7 @@ dependencies {
 
     implementation project(':concept-storage')
     implementation project(':support-sync-telemetry')
+    implementation project(':service-glean')
 
     implementation Dependencies.kotlin_stdlib
     implementation Dependencies.kotlin_coroutines
--- a/components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/AsyncLoginsStorage.kt
+++ b/components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/AsyncLoginsStorage.kt
@@ -12,7 +12,6 @@ import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.plus
 import mozilla.appservices.logins.DatabaseLoginsStorage
 import mozilla.appservices.logins.LoginsStorage
-import mozilla.appservices.logins.MemoryLoginsStorage
 import mozilla.components.concept.sync.SyncAuthInfo
 import mozilla.components.concept.sync.SyncStatus
 import mozilla.appservices.sync15.SyncTelemetryPing
@@ -210,8 +209,7 @@ interface AsyncLoginsStorage : AutoCloseable {
      * generated automatically. The format of generated guids
      * are left up to the implementation of LoginsStorage (in
      * practice the [DatabaseLoginsStorage] generates 12-character
-     * base64url (RFC 4648) encoded strings, and [MemoryLoginsStorage]
-     * generates strings using [java.util.UUID.toString])
+     * base64url (RFC 4648) encoded strings.
      *
      * This will return an error result if a GUID is provided but
      * collides with an existing record, or if the provided record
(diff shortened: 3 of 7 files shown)


components/service/sync-logins/build.gradle
```diff
@@ -29,13 +29,18 @@ android {
 
 dependencies {
     // Parts of this dependency are typealiase'd or are otherwise part of this module's public API.
-    api Dependencies.mozilla_sync_logins
+    api(Dependencies.mozilla_sync_logins) {
+      // Use our own version of the Glean dependency,
+      // which might be newer than the version declared by A-S.
+      exclude group: 'org.mozilla.components', module: 'service-glean'
+    }
```
rfk: I copied this from the existing rusthttp integration, in which a-s takes a dependency on a-c to get at concept-fetch.

rfk:
(CI is failing because of the `mavenLocal` stuff, which I'd obviously remove before marking this ready-for-review)

components/service/sync-logins/build.gradle
grigoryk: nit: 'newer' is probably better phrased as 'different'.

rfk:
> I don't know whether they would have been gathered in memory, or written to disk; I should see if I can find that out.

Following up here for completeness: :Dexter says that a small quantity of metrics will be accumulated in memory, but discarded rather than being flushed to disk.

rfk:
Thanks for taking a look! Your feedback is not a surprise, but I wanted to start from the simplest possible option and work from there.

I'm curious whether you'd feel differently if the data-review requirement wasn't present, e.g. if we managed to get a data-review signoff that it's OK for the component to collect these metrics from any Firefox app without having to do any extra reviews at the app level.

> Applications will bump versions without perusing every single notice in the changelog,
> without realizing that they're now collecting more data than before. Even if changelog is studied,
> this doesn't seem like a scaleable approach.

Aye. I think the only way to make this tractable over the long term will be to have good policy and tooling support from Glean, ref [Bug 1598247](https://bugzilla.mozilla.org/show_bug.cgi?id=1598247). Having to set 20 different "enable this metric, enable that metric" flags across various component APIs sounds like a nightmare.

But, we can do it that way in the meantime.

> An explicit flag that's part of component's public API seems like a decent starting point, which will > also enables applications to give users a more granular control over data collection.

Details TBD while I remember the right idioms for Kotlin, but I want to propose some sort of module-level function like `SyncableLoginsStore.enableTelemetry(version)` with an integer version number. The current suite of metrics would only be collected if the app calls this with `version >= 1`. If we add new metrics at a later date, they would only be collected if the app passes `version >= 2`. And so on.

Does that seem workable @grigoryk? (I think it's awful, but I acknowledge that it may be necessary until we figure out something better).

components/service/sync-logins/build.gradle
rfk: Heh, I don't think it will ever be older, but I agree that's probably clearer.

rfk:
> I want to propose some sort of module-level function like SyncableLoginsStore.enableTelemetry(version)

I pushed a strawman of how this might look in practice (but haven't actually implemented the underlying logic on the a-s side).

Dexterp37:
> 
> 
> I'm really not a fan of this, to be honest. As we discussed elsewhere, I don't think a breaking change notice is enough - or a notice in component's README, for that matter.

I think this should be discussed and solved generally, without blocking this specific integration. We have [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1598247) for exploring possibilities and discussing options.

> Applications will bump versions without perusing every single notice in the changelog, without realizing that they're now collecting more data than before. Even if changelog _is_ studied, this doesn't seem like a scaleable approach.

While I agree that README and changelogs changes may skipped over (thus the bug I referenced earlier, even though not reading changelogs is probably **bad** practice :-) ), providing a control mechanism outside of Glean for this would be [problematic for the whole Glean ecosystem](https://github.com/mozilla/application-services/pull/2325#discussion_r353641202): from collection to analysis. And, down the line, of course, for the component itself.

We are discussing a general solution to this problem, which might be either a more explicit review process (Policy) or some other technical expedient.

> Default behaviour of our components must be safe and non-surprising.

I agree with that :-)

> "off by default" telemetry collection seems like an obvious way to avoid this footgun. This implies presence of some lever that needs to be switched for telemetry collection to happen within a component.

Components are meant to collect general telemetry about the feature they implement. Every new metric in every single component goes through data-review. Data-reviewers should be aware of the fact that Android components can be consumed by different Mozilla Android applications.

If a metric is product specific, controversial or not general enough to go in a _component_, then that's a signal that it should live in the product.

Another obvious solution, to be extremely explicit and to be friendly with data analysts down the line (and to the Glean ecosystem), could be to publish two versions of the library. One with telemetry, one without. This would make it extremely easy, at each step (integration, collection, ingestion, ETL, analysis) that app X is not collecting telemetry from the specific component, because that component doesn't come with any telemetry at all.

We have [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1525904) for making it easier, but it can be prioritized accordingly if it's a requirement.

> An explicit flag that's part of component's public API seems like a decent starting point, which will also enables applications to give users a more granular control over data collection.

As mentioned earlier, I'm afraid that this would be [problematic](https://github.com/mozilla/application-services/pull/2325#discussion_r353641202). Moreover, applying the suggested solution exclusively to this component, would make it work differently than other Glean-enabled components in A-C (sync, engine-gecko, ...).

This problem requires a general, end to end solution and I would strongly encourage everyone to take part to the general conversation.

@grigoryk what's the specific requirement/discussion this is coming from? Is there any concrete request from Fenix/Lockwise/FxR to integrate `logins` without collecting the related telemetry?

rfk:
> I pushed a strawman of how this might look in practice (but haven't actually implemented
> the underlying logic on the a-s side).

I've reverted this; based on discussions in this and related issues, it's going to add pipeline complexity that's not justified in practice given our data-review requirements and practices. We can iterate on the concerns raised in this PR over time, but I'm getting a very strong "doesn't need to block metrics in the meantime" vibe from all the surrounding discussions.

rfk:
@grigoryk given all the discussions, and pending an affirmative answer in [Bug 1597895 Comment 15](https://bugzilla.mozilla.org/show_bug.cgi?id=1597895#c15), do you have any objections to landing a change like the one that's currently in this PR to enable basic logins component telemetry?

components/service/sync-logins/README.md
```diff
@@ -10,6 +10,16 @@ The **Firefox Sync - Logins Component** provides a way for Android applications
...
 ## Usage
 
+### Before using this component
+
+The `mozilla.appservices.logins` component collects telemetry using the [Glean SDK](https://mozilla.github.io/glean/).
+Applications that send telemetry via Glean *must ensure* they have received appropriate data-review following
+[the Firefox Data Collection process](https://wiki.mozilla.org/Firefox/Data_Collection) before integrating this component.
+
+Details on the metrics collected by the `mozilla.appservices.logins` component are available
+[here](https://github.com/mozilla/application-services/tree/master/docs/metrics/logins/metrics.md)
```
Dexterp37: This page is currently a 404: is this expected?

rfk: Yes; it'll be live when the upstream logins-component PR lands.

rfk: I just checked to confirm, and this link is now live as expected.

rfk:
Application-services v0.46.0 is now out, so this is ready for review.

rfk:
I'm not sure what's up with the test failure here sorry:

```
[task 2019-12-14T06:53:49.810Z] Execution failed for task ':support-rustlog:jacocoDebug'.
[task 2019-12-14T06:53:49.810Z] > A failure occurred while executing com.android.build.gradle.internal.tasks.Workers$ActionFacade
[task 2019-12-14T06:53:49.810Z]    > java.lang.ExceptionInInitializerError (no error message)
```

docs/changelog.md
```diff
@@ -148,6 +148,12 @@ permalink: /changelog/
     * Upgrade glean_parser to v1.12.0.
     * Implement the deletion request ping in Glean.
 
+* **service-sync-logins**
```
jonalmeida: Drive-by comment: this changelog entry probably needs to go to the latest release.

rfk: Thanks! I guess this had gone more stale than I thought :-)

components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/AsyncLoginsStorage.kt
```diff
@@ -386,13 +384,6 @@ open class AsyncLoginsStorageAdapter<T : LoginsStorage>(private val wrapped: T)
         fun forDatabase(dbPath: String): AsyncLoginsStorageAdapter<DatabaseLoginsStorage> {
             return AsyncLoginsStorageAdapter(DatabaseLoginsStorage(dbPath))
         }
-
-        /**
-         * Creates an [AsyncLoginsStorage] that is backed by a [MemoryLoginsStorage].
-         */
-        fun inMemory(items: List<ServerPassword>): AsyncLoginsStorageAdapter<MemoryLoginsStorage> {
```
rfk: The a-s `MemoryLoginsStorage` class is deprecated. I couldn't find any consumers of `inMemory` so I just removed it, but I could change this to be deprecated as well if that's preferred.

grigoryk: I think it's fine to remove it. We haven't been very strict with our "deprecate major api pieces before outright removal" policy...
Thanks for mentioning an alternative in the changelog.

components/support/migration/build.gradle
```diff
@@ -67,6 +67,7 @@ dependencies {
     jnaForTest Dependencies.thirdparty_jna
     testImplementation files(configurations.jnaForTest.copyRecursive().files)
     testImplementation Dependencies.mozilla_full_megazord_forUnitTests
+    testImplementation Dependencies.mozilla_glean_forUnitTests
```
grigoryk: Why did you need this in `support-migration`?

PR #5228, opened 2019-12-03
Migrate fennec telemetry setting

Mugurell:
Introduces a new "FennecSettingsMigrator" which will be responsible for
migrating all common Fennec - Fenix settings.
Starting with the telemetry opt-in status which in Fenix will be "true" only
if both "Telemetry" and "Firefox Health Report" were enabled on Fennec.

---

### Pull Request checklist

- [x] **Changelog**: This PR includes does not need [a changelog entry](https://github.com/mozilla-mobile/android-components/blob/master/docs/changelog.md).

### After merge

### DIFF
--- a/components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt
+++ b/components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt
@@ -21,6 +21,7 @@ import mozilla.components.lib.crash.CrashReporter
 import mozilla.components.service.fxa.manager.FxaAccountManager
 import mozilla.components.service.sync.logins.AsyncLoginsStorage
 import mozilla.components.support.base.log.logger.Logger
+import mozilla.components.support.migration.FennecMigrator.Builder
 import java.io.File
 import java.lang.Exception
 import java.util.concurrent.Executors
@@ -60,6 +61,11 @@ sealed class Migration(val currentVersion: Int) {
      * Migrates Gecko(View) internal files.
      */
     object Gecko : Migration(currentVersion = 1)
+
+    /**
+     * Migrates all Fennec settings backed by SharedPreferences.
+     */
+    object Settings : Migration(currentVersion = 1)
 }
 
 /**
@@ -106,11 +112,18 @@ sealed class FennecMigratorException(cause: Exception) : Exception(cause) {
      * @param cause Original exception which caused the problem.
      */
     class MigrateOpenTabsException(cause: Exception) : FennecMigratorException(cause)
+
     /**
      * Unexpected exception while migrating gecko profile.
      * @param cause Original exception which caused the problem.
      */
     class MigrateGeckoException(cause: Exception) : FennecMigratorException(cause)
+
+    /**
+     * Unexpected exception while migrating settings.
+     * @param cause Original exception which caused the problem
+     */
+    class MigrateSettingsException(cause: Exception) : FennecMigratorException(cause)
 }
 
 /**
@@ -250,6 +263,14 @@ class FennecMigrator private constructor(
         fun migrateFxa(accountManager: FxaAccountManager, version: Int = Migration.FxA.currentVersion): Builder {
             this.accountManager = accountManager
             migrations.add(VersionedMigration(Migration.FxA, version))
+            return this
+        }
+
+        /**
+         * Enable all Fennec - Fenix common settings migration.
+         */
+        fun migrateSettings(version: Int = Migration.Settings.currentVersion): Builder {
+            migrations.add(VersionedMigration(Migration.Settings, version))
             return this
         }
 
@@ -395,6 +416,7 @@ class FennecMigrator private constructor(
                 Migration.FxA -> migrateFxA()
                 Migration.Gecko -> migrateGecko()
                 Migration.Logins -> migrateLogins()
+                Migration.Settings -> migrateSharedPrefs()
             }
 
             results[versionedMigration.migration] = when (migrationResult) {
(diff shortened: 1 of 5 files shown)


components/support/migration/src/main/java/mozilla/components/support/migration/FennecSettingsMigrator.kt
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+
+private fun migrateTelemetryOptInStatus(
+    fennecPrefs: SharedPreferences,
+    fenixPrefs: SharedPreferences
+): SettingsMigrationResult {
+    val isTelemetryEnabled = fennecPrefs.getBoolean(FENNEC_PREFS_TELEMETRY_KEY, false)
+    val isHealthReportEnabled = fennecPrefs.getBoolean(FENNEC_PREFS_FHR_KEY, false)
+    val isTelemetryEnabledOnFennec = isHealthReportEnabled && isTelemetryEnabled
```
grigoryk: My understanding was that we will simply use the FHR value. That's documented in https://github.com/mozilla-mobile/fenix/issues/5525 - `We are going to map those two values to a single value when migrating - with the FHR setting taking precedence.`

components/support/migration/src/main/java/mozilla/components/support/migration/FennecSettingsMigrator.kt
```diff
@@ -0,0 +1,100 @@
...
+    @VisibleForTesting const val FENNEC_PREFS_FHR_KEY = "android.not_a_preference.healthreport.uploadEnabled"
+    @VisibleForTesting const val FENNEC_PREFS_TELEMETRY_KEY = "toolkit.telemetry.enabled"
+    @VisibleForTesting const val FENIX_PREFS_TELEMETRY_KEY = "pref_key_telemetry"
+
+    /**
+     * Migrate all Fennec - Fenix common SharedPreferences.
+     */
+    fun migrateSharedPrefs(context: Context): Result<SettingsMigrationResult> {
+        fennecAppPrefs = context.getSharedPreferences(FENNEC_APP_SHARED_PREFS_NAME, MODE_PRIVATE)
+        fenixPrefs = context.getSharedPreferences(FENIX_SHARED_PREFS_NAME, MODE_PRIVATE)
```
grigoryk: That's not the right shared pref. From Fenix's Settings.kt:
```
private val appContext = context.applicationContext

val isTelemetryEnabled by booleanPreference(
        appContext.getPreferenceKey(R.string.pref_key_telemetry),
        default = true
    )
```

So the actual pref ends up in something like `shared_prefs/org.mozilla.fenix.preferences.xml` file, or `org.mozilla.fenix.debug_preferences.xml`, or whatever the app id is. There's _also_ `fenix_preferences.xml` but that's not where this pref is read from.

I think if you just use `context.applicationContext` it should be fine though.

components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt
```diff
@@ -494,4 +515,23 @@ class FennecMigrator private constructor(
...
+
+    private fun migrateSharedPrefs(): Result<SettingsMigrationResult> {
+        val result = FennecSettingsMigration.migrateSharedPrefs(context)
+
+        if (result is Result.Failure<SettingsMigrationResult>) {
+            val migrationFailureWrapper = result.throwables.first() as SettingsMigrationException
+            return when (migrationFailureWrapper.failure) {
+                is SettingsMigrationResult.Failure.EmptyPreference -> {
+                    logger.error("Detected a corrupt state: ${migrationFailureWrapper.failure}")
+                    crashReporter.submitCaughtException(
```
grigoryk: I don't think we should report this exception. What if you're just installing this build (e.g. fresh Fenix install)? That's not a problem at all, but as-is, we'll report an exception.

jonalmeida:
Drive-by comment: this PR is referencing an a-c issue that doesn't exist (probably a Fenix one I'm guessing).

components/support/migration/src/main/java/mozilla/components/support/migration/FennecSettingsMigrator.kt
Mugurell: Indeed, thanks!

components/support/migration/src/main/java/mozilla/components/support/migration/FennecSettingsMigrator.kt
Mugurell: Thanks!

components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt
Mugurell: Thanks!.
For migrating Preferences, as of now for only telemetry there is just one possible result: `Success`.
Either because this is a fresh install and there's nothing to migrate, either because we just write a new value to app's shared prefs.
In the future the logic may become more convoluted but for now this is as simple as it gets.

Mugurell:
Thanks @jonalmeida and @grigoryk !
Filed https://github.com/mozilla-mobile/android-components/issues/5267 for the needed a-c work
And will continue on https://github.com/mozilla-mobile/fenix/issues/5525 with the work needed for Fenix to use the new migration.

components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt
grigoryk: Simple is good!

grigoryk:
@Mugurell we use "assigned" on PRs to track who's responsible for reviewing, so that PRs don't go unreviewed for too long.

components/support/migration/src/test/java/mozilla/components/support/migration/FennecMigratorTest.kt
```diff
@@ -419,6 +419,21 @@ class FennecMigratorTest {
...
+        // Fennec SharedPreferences are missing / empty
+        val migrator = FennecMigrator.Builder(testContext, mock())
+            .migrateSettings()
+            .setCoroutineContext(this.coroutineContext)
+            .build()
+
+        with(migrator.migrateAsync().await()) {
+            assertEquals(1, this.size)
+            assertTrue(this.containsKey(Migration.Settings))
+            assertFalse(this.getValue(Migration.Settings).success)
```
grigoryk: This is now wrong. Can you assert here that you succeed with `SettingsMigrationResult.Success.NoFennecPrefs`?

grigoryk:
@Mugurell I've forced push a version which works, with slightly more paranoid error reporting.

I was wrong in my initial review, but I'm not entirely sure why yet:
- `fenix_preferences.xml` _is_ where the telemetry setting should go! This is what you had initially, IIRC.
- but, `org.mozilla.firefox_preferences.xml` _also_ exists, and also has the same setting! And it value even tracks what changes in Fenix's UI. But, updating this value manually doesn't actually change telemetry settings.

grigoryk:
bors retry

Mugurell:
@grigoryk Thanks! 
Also saw that but I was planning to tackle that in Fenix.

grigoryk:
Fenix PR to enable this: https://github.com/mozilla-mobile/fenix/pull/7105

PR #5239, opened 2019-12-05
Closes #4774: Link addons to their installed extensions

### LINKED ISSUE #4774: Query installed extensions on startup and add to browser store
We've already added the required browser state to represent web extensions. This is about adding that state on startup after querying the engine (GV) for installed add-ons.

csadilek:
We don't have the actual `engine.listInstalledExtensions` yet but this brings in functionality to query installed extensions via the engine and allows us to add the installed extension state to the addons. 

This state will ultimately capture more extension metadata (enabled/blocked etc.), but for now it is just the installed version and the `optionsPageUrl`, which I used here for our settings page. I've also moved our `enabled` state to this new installed extension state.

The UI will use the new `AddonManager` to query `Addons` which will make sure they have this new state set. Instead of returning an `emptyList` when something goes wrong, we now wrap all exceptions and throw them. We need to be able to distinguish between "there are no addons" and "there was a problem fetching them".

### DIFF
--- a/components/feature/addons/src/main/java/mozilla/components/feature/addons/Addon.kt
+++ b/components/feature/addons/src/main/java/mozilla/components/feature/addons/Addon.kt
@@ -13,9 +13,6 @@ import kotlinx.android.parcel.Parcelize
  *
  * @property id The unique ID of this add-on.
  * @property authors List holding information about the add-on authors.
- * @property installed Indicates if this [Addon] is installed or not, defaults to false.
- * @property enabled Indicates if this [Addon] is enabled to interact with web content or not,
- * defaults to false.
  * @property categories List of categories the add-on belongs to.
  * @property downloadUrl The (absolute) URL to download the add-on file (eg xpi).
  * @property version The add-on version e.g "1.23.0".
@@ -31,14 +28,14 @@ import kotlinx.android.parcel.Parcelize
  * @property rating The rating information of this add-on.
  * @property createdAt The date the add-on was created.
  * @property updatedAt The date of the last time the add-on was updated by its developer(s).
+ * @property installedState Holds the state of the installed web extension for this add-on. Null, if
+ * the [Addon] is not installed.
  */
 @Parcelize
 data class Addon(
     val id: String,
     val authors: List<Author>,
     val categories: List<String>,
-    val installed: Boolean = false,
-    val enabled: Boolean = false,
     val downloadUrl: String,
     val version: String,
     val permissions: List<String> = emptyList(),
@@ -49,7 +46,8 @@ data class Addon(
     val siteUrl: String = "",
     val rating: Rating? = null,
     val createdAt: String,
-    val updatedAt: String
+    val updatedAt: String,
+    val installedState: InstalledState? = null
 ) : Parcelable {
     /**
      * Represents an add-on author.
@@ -69,6 +67,7 @@ data class Addon(
 
     /**
      * Holds all the rating information of this add-on.
+     *
      * @property average An average score from 1 to 5 of how users scored this add-on.
      * @property reviews The number of users that has scored this add-on.
      */
@@ -80,10 +79,39 @@ data class Addon(
 
     /**
      * Returns a list of id resources per each item on the [Addon.permissions] list.
+     * Holds the state of the installed web extension of this add-on.
+     *
+     * @property id The ID of the installed web extension.
+     * @property version The installed version.
+     * @property enabled Indicates if this [Addon] is enabled to interact with web content or not,
+     * defaults to false.
+     * @property optionsPageUrl the URL of the page displaying the
+     * options page (options_ui in the extension's manifest).
+     */
+    @Parcelize
+    data class InstalledState(
+        val id: String,
+        val version: String,
+        val optionsPageUrl: String,
+        val enabled: Boolean = false
+    ) : Parcelable
+
+    /**
+     * Returns a list of id resources per each item on the [AddOn.permissions] list.
      */
     fun translatePermissions(): List<Int> {
-        return permissions.mapNotNull { it -> permissionToTranslation[it] }
+        return permissions.mapNotNull { permissionToTranslation[it] }
     }
+
+    /**
+     * Returns whether or not this [Addon] is currently installed.
+     */
+    fun isInstalled() = installedState != null
+
+    /**
+     * Returns whether or not this [Addon] is currently enabled.
+     */
+    fun isEnabled() = installedState?.enabled == true
 
     companion object {
         /**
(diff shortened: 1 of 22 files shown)


components/feature/addons/src/main/java/mozilla/components/feature/addons/Addon.kt
```diff
@@ -80,11 +79,40 @@ data class Addon(
...
      */
     fun translatePermissions(): List<Int> {
-        return permissions.mapNotNull { it -> permissionToTranslation[it] }
+        return permissions.mapNotNull { permissionToTranslation[it] }
     }
 
+    /**
+     * Returns whether or not this [Addon] is currently installed.
+     */
+    fun isInstalled() = installedState != null
```
psymoon: nit: 
```    
val isInstalled: Boolean
    get() = installedState != null
```

components/feature/addons/src/main/java/mozilla/components/feature/addons/Addon.kt
```diff
@@ -80,11 +79,40 @@ data class Addon(
...
+
+    /**
+     * Returns whether or not this [Addon] is currently enabled.
+     */
+    fun isEnabled() = installedState?.enabled == true
```
psymoon: nit:
```
val isEnabled: Boolean
    get() = installedState?.enabled == true
```

components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
```diff
@@ -119,21 +133,50 @@ object WebExtensionSupport {
...
+                throwable -> logger.error("Failed to query installed extension", throwable)
+                initializationResult.completeExceptionally(throwable)
+            }
+        )
+    }
+
+    /**
+     * Marks the provided [webExtension] as installed by adding it to the [store].
+     */
+    private fun registerInstalledExtension(store: BrowserStore, webExtension: WebExtension) {
```
psymoon: fun overloading intended?

components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
```diff
@@ -0,0 +1,57 @@
...
+
+package mozilla.components.feature.addons
+
+import mozilla.components.browser.state.action.WebExtensionAction
+import mozilla.components.browser.state.state.WebExtensionState
+import mozilla.components.browser.state.store.BrowserStore
+import mozilla.components.support.webextensions.WebExtensionSupport
+
+/**
+ * Provides access to installed and recommended [Addon]s and manages their state.
```
psymoon: nit: ... manages their state**s**.

components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * Returns the list of all installed and recommended add-ons.
+     *
+     * @return list of all [Addon]s with up-to-date [Addon.installedState].
+     * @throws AddonManagerException in case of a problem reading from
+     * the [addonsProvider] or querying web extension state from the engine / store.
+     */
+    @Throws(AddonManagerException::class)
+    @Suppress("TooGenericExceptionCaught")
+    suspend fun getAddons(): List<Addon> {
```
psymoon: Should we just rename this method to `getInstalledAddons`? `getAddons` sounds too generic for me, I would have to look at the javadoc.

components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
```diff
@@ -119,21 +133,50 @@ object WebExtensionSupport {
...
+        installedExtensions[webExtension.id] = webExtension
+        store.dispatch(WebExtensionAction.InstallWebExtensionAction(webExtension.toState()))
+
+        // Register action handler for all existing engine sessions on the new extension
+        store.state.tabs
+            .forEach {
+                val session = it.engineState.engineSession
```
psymoon: nit:
```
it.engineState.engineSession?.let { session ->
    registerSessionActionHandler(webExtension, session, SessionActionHandler(store, it.id))
}
```

csadilek: thanks, updated!

components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
csadilek: Fixed.

components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
csadilek: The method returns all `Addon`s (incl. the recommended but not installed ones). So, I think we should keep the current name or maybe use `getAllAddons`...

components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
csadilek: @psymoon don't see any overloading here? There's a plural version (`registerInstalledExtensions`) which acts on all installed extensions and one that acts on just one (`registerInstalledExtension`).

components/feature/addons/src/main/java/mozilla/components/feature/addons/Addon.kt
csadilek: hm why do you think it's better to turn this into a property? We don't allow mutating the val anyway.

components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
psymoon: Ah, my mistake 👍

components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
psymoon: Sounds good

components/feature/addons/src/main/java/mozilla/components/feature/addons/Addon.kt
psymoon: Hm now that I think about it, `sInstalled` would never be mutable so I guess func makes more sense

csadilek:
bors r=psymoon,Amejia481

Issue #4500, opened 2019-09-23
Add engine API for Add-ons management


csadilek:
This may need to be broken down more later, but on a high level we need to wrap the Add-on management API (to be) provided by GeckoView to:

- download and install Add-ons
- query installed and available Add-ons
- enable/disable Add-ons
- update add-on

csadilek:
Blocked on: https://bugzilla.mozilla.org/show_bug.cgi?id=1582185

csadilek:
OK, we have everything now that was covered by this ticket. Closing.

pocmo:
Looking forward to getting the Reference Browser update with that. :)

PR #5240, opened 2019-12-05
Issue #4500: Add action / state to enable and disable extension

csadilek:
Included a small refactoring to add re-usable methods for updating both the global and tab-specific web extension state.


components/browser/state/src/main/java/mozilla/components/browser/state/reducer/WebExtensionReducer.kt
```diff
@@ -27,71 +26,64 @@ internal object WebExtensionReducer {
...
+                    }
+                    current.copy(extensionState = updatedExtensions)
                 } else {
                     current
                 }
             }
         )
     }
+
+    private fun BrowserState.updateWebExtensionState(
```
Amejia481: This is exactly the same refractory that I'm doing locally 😄

csadilek: 👍 :)

csadilek:
bors r=Amejia481,psymoon

PR #5241, opened 2019-12-05
Do not inherit LifecycleAwareFeature in RemoteTabsFeature

eoger:
Turns out it doesn't really work out as we need the feature for the awesomebar suggestions.


jonalmeida:
While we're removing the `LifecycleAwareFeature`, we're also allowing a potential rogue coroutine to run without being able to cancel it.

I don't have a code sample to reference (yet), but there might be examples around the code base for features that need to live as long as the app's lifecycle (i.e. ProcessLifecycleOwner).

eoger:
I basically copied what `WebExtensionSupport` is doing: https://github.com/mozilla-mobile/android-components/blob/ac99d412cc3f08554705372b29500c1f410d9ed3/components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt#L183-L199

eoger:
@jonalmeida I have added stop `start`/`stop` back following our in-person discussion, just remove the inheritance which makes no sense.

jonalmeida:
Failed lib-state was from the CI container being killed:
```
[taskcluster:error] Task timeout after 7200 seconds. Force killing container.
```

Issue #5143, opened 2019-11-25
Provide mechanism to trigger Addon updates


csadilek:
GV will provide an `update` call on the web extension controller to check for updates for all installed addons, and automatically install them, if available. We should provide a mechanism to call this method e.g. when the device is idle or periodically in the background?

One complexity here is that the permissions of addons could have changed, requiring a new confirmation from the user. So, we'd have to display our permission confirmation dialog and have users confirm for every addon that has changed permissions. GV will provide a callback for that.

csadilek:
Depends on https://bugzilla.mozilla.org/show_bug.cgi?id=1599581

Amejia481:
Closed by https://github.com/mozilla-mobile/android-components/commit/e2ac3d169f4dfc40400bd099096554b88a112c9d

PR #5242, opened 2019-12-05
For issue #5143 Initial support for triggering Addon updates.

Amejia481:
This draft only includes happy path where an extension is updated without the need of re-granting permissions.
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
@@ -225,6 +225,19 @@ class GeckoEngine(
     }
 
     /**
+     * See [Engine.updateWebExtension].
+     */
+    override fun updateWebExtension(
+        extension: WebExtension,
+        onSuccess: (WebExtension?) -> Unit,
+        onError: (String, Throwable) -> Unit
+    ) {
+        // GeckoView support for updating extensions hasn't been implemented yet
+        // TODO https://bugzilla.mozilla.org/show_bug.cgi?id=1599581
+        onSuccess(null)
+    }
+
+    /**
      * See [Engine.registerWebExtensionDelegate].
      */
     override fun registerWebExtensionDelegate(
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/WebExtensionReducer.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/WebExtensionReducer.kt
@@ -52,6 +52,9 @@ internal object WebExtensionReducer {
                     it.copy(browserAction = action.browserAction)
                 }
             }
+            is WebExtensionAction.UpdateWebExtension -> {
+                state.updateWebExtensionState(action.updatedExtension.id) { action.updatedExtension }
+            }
         }
     }
 
--- a/components/browser/state/src/test/java/mozilla/components/browser/state/action/WebExtensionActionTest.kt
+++ b/components/browser/state/src/test/java/mozilla/components/browser/state/action/WebExtensionActionTest.kt
@@ -188,4 +188,20 @@ class WebExtensionActionTest {
         store.dispatch(WebExtensionAction.UpdateWebExtensionEnabledAction(extension.id, true)).joinBlocking()
         assertTrue(store.state.extensions[extension.id]?.enabled!!)
     }
+
+    @Test
+    fun `UpdateWebExtension -  Update an existing extension`() {
+        val existingExtension = WebExtensionState("id", "url")
+        val updatedExtension = WebExtensionState("id", "url2")
+
+        val store = BrowserStore(
+            initialState = BrowserState(
+                extensions = mapOf("id" to existingExtension)
+            )
+        )
+
+        store.dispatch(WebExtensionAction.UpdateWebExtension(updatedExtension)).joinBlocking()
+        assertEquals(updatedExtension, store.state.extensions.values.first())
+        assertSame(updatedExtension, store.state.extensions.values.first())
+    }
 }
--- a/components/concept/engine/src/main/java/mozilla/components/concept/engine/Engine.kt
+++ b/components/concept/engine/src/main/java/mozilla/components/concept/engine/Engine.kt
@@ -141,6 +141,26 @@ interface Engine {
     ): Unit = onError(id, UnsupportedOperationException("Web extension support is not available in this engine"))
 
     /**
+     * Updates the provided [extension] if a new version is available.
+     *
+     * @param extension the extension to be updated.
+     * @param onSuccess (optional) callback invoked if the extension was updated successfully,
+     * providing access to the [WebExtension] object for bi-directional messaging, if null is provided
+     * that means that the [WebExtension] hasn't been change since the last update.
+     * @param onError (optional) callback invoked if there was an error updating the extension.
+     * This callback is invoked with an [UnsupportedOperationException] in case the engine doesn't
+     * have web extension support.
+     */
+    fun updateWebExtension(
+        extension: WebExtension,
+        onSuccess: ((WebExtension?) -> Unit) = { },
+        onError: ((String, Throwable) -> Unit) = { _, _ -> }
+    ): Unit = onError(
+        extension.id,
+        UnsupportedOperationException("Web extension support is not available in this engine")
+    )
+
+    /**
      * Uninstalls the provided extension from this engine.
      *
      * @param ext the [WebExtension] to uninstall.
--- a/components/concept/engine/src/main/java/mozilla/components/concept/engine/webextension/WebExtensionDelegate.kt
+++ b/components/concept/engine/src/main/java/mozilla/components/concept/engine/webextension/WebExtensionDelegate.kt
@@ -82,4 +82,20 @@ interface WebExtensionDelegate {
      * granted.
      */
     fun onInstallPermissionRequest(webExtension: WebExtension): Boolean = false
+
+    /**
+     * Invoked when a web extension has changed its permissions while trying to update to a
+     * new version. This requires user interaction as the updated extension will not be installed,
+     * until the user grants the new permissions.
+     *
+     * @param current The current [WebExtension].
+     * @param updated The update [WebExtension] that requires extra permissions.
+     * @param onPermissionsGranted A callback to indicate if the new permissions from the [updated] extension
+     * are granted or not.
+     */
+    fun onUpdatePermissionRequest(
+        current: WebExtension,
+        updated: WebExtension,
+        onPermissionsGranted: ((Boolean) -> Unit)
+    ) = Unit
 }
--- a/components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
+++ b/components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
@@ -7,6 +7,11 @@ package mozilla.components.feature.addons
 import mozilla.components.browser.state.action.WebExtensionAction
 import mozilla.components.browser.state.state.WebExtensionState
 import mozilla.components.browser.state.store.BrowserStore
+import mozilla.components.concept.engine.Engine
+import mozilla.components.concept.engine.webextension.WebExtension
+import mozilla.components.feature.addons.update.AddonUpdater
+import mozilla.components.feature.addons.update.AddonUpdater.Status
+import mozilla.components.feature.addons.update.GlobalAddonManagerProvider
 import mozilla.components.support.webextensions.WebExtensionSupport
 
 /**
@@ -17,8 +22,14 @@ import mozilla.components.support.webextensions.WebExtensionSupport
  */
 class AddonManager(
     private val store: BrowserStore,
-    private val addonsProvider: AddonsProvider
+    private val engine: Engine,
+    private val addonsProvider: AddonsProvider,
+    private val addonUpdater: AddonUpdater
 ) {
+
+    init {
+        GlobalAddonManagerProvider.initialize(this)
+    }
 
     /**
      * Returns the list of all installed and recommended add-ons.
@@ -48,6 +59,35 @@ class AddonManager(
         } catch (throwable: Throwable) {
             throw AddonManagerException(throwable)
         }
+    }
+
+    /**
+     * Try updates the addon with the provided [id] by updating it in the the [store] and re-attaching
+     * all the handlers. If there is no update available nothing will happen.
+     */
+    fun updateAddon(id: String, onFinish: ((Status) -> Unit)) {
+        val extension = WebExtensionSupport.installedExtensions[id]
+
+        if (extension == null) {
+            onFinish(Status.NotInstalled)
+            return
+        }
+
+        val onSuccess: ((WebExtension?) -> Unit) = { updatedExtension ->
+            val status = if (updatedExtension == null) {
+                Status.NoUpdateAvailable
+            } else {
+                WebExtensionSupport.markExtensionAsUpdated(store, updatedExtension)
+                Status.SuccessfullyUpdated
+            }
+            onFinish(status)
+        }
+
+        val onError: ((String, Throwable) -> Unit) = { message, exception ->
+            onFinish(Status.Error(message, exception))
+        }
+
+        engine.updateWebExtension(extension, onSuccess, onError)
     }
 }
 
--- a/components/feature/addons/src/main/java/mozilla/components/feature/addons/update/AddonUpdater.kt
+++ b/components/feature/addons/src/main/java/mozilla/components/feature/addons/update/AddonUpdater.kt
@@ -1,0 +1,196 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+package mozilla.components.feature.addons.update
+
+import android.content.Context
+import androidx.annotation.VisibleForTesting
+import androidx.work.Constraints
+import androidx.work.CoroutineWorker
+import androidx.work.Data
+import androidx.work.ExistingPeriodicWorkPolicy
+import androidx.work.NetworkType
+import androidx.work.PeriodicWorkRequest
+import androidx.work.PeriodicWorkRequestBuilder
+import androidx.work.WorkManager
+import androidx.work.WorkerParameters
+import mozilla.components.feature.addons.update.AddonUpdater.Frequency
+import mozilla.components.support.base.log.logger.Logger
+import java.util.concurrent.TimeUnit
+import kotlin.coroutines.resume
+import kotlin.coroutines.suspendCoroutine
+
+/**
+ * Contract to define the behavior for updating addons.
+ */
+interface AddonUpdater {
+    /**
+     * Registers the given [addonId] for periodically check for new updates.
+     * @param addonId The unique id of the addon.
+     */
+    fun registerForFutureUpdates(addonId: String)
+
+    /**
+     * Unregisters the given [addonId] for periodically checking for new updates.
+     * @param addonId The unique id of the addon.
+     */
+    fun unregisterForFutureUpdates(addonId: String)
+
+    /**
+     * Indicates the status of a request for updating an addon.
+     */
+    sealed class Status {
+        /**
+         * The addon is not part of the installed list.
+         */
+        object NotInstalled : Status()
+
+        /**
+         * The addon was successfully updated.
+         */
+        object SuccessfullyUpdated : Status()
+
+        /**
+         * The addon has not been updated since the last update.
+         */
+        object NoUpdateAvailable : Status()
+
+        /**
+         * An error has happened while trying to update.
+         * @property message A string message describing what has happened.
+         * @property exception The exception of the error.
+         */
+        class Error(val message: String, val exception: Throwable) : Status()
+    }
+
+    /**
+     * Indicates how often an extension should be updated.
+     * @property repeatInterval Integer indicating how often the update should happen.
+     * @property repeatIntervalTimeUnit The time unit of the [repeatInterval].
+     */
+    class Frequency(val repeatInterval: Long, val repeatIntervalTimeUnit: TimeUnit)
+}
+
+/**
+ * An implementation of [AddonUpdater] that uses the work manager api for scheduling new updates.
+ * @property applicationContext The application context.
+ * @param frequency (Optional) indicates how often updates should be performed, defaults
+ * to one day.
+ */
+class DefaultAddonUpdater(
+    private val applicationContext: Context,
+    private val frequency: Frequency = Frequency(1, TimeUnit.DAYS)
+) : AddonUpdater {
+    /**
+     * See [AddonUpdater.registerForFutureUpdates]
+     */
+    override fun registerForFutureUpdates(addonId: String) {
+        WorkManager.getInstance(applicationContext).enqueueUniquePeriodicWork(
+            getUniquePeriodicWorkName(addonId),
+            ExistingPeriodicWorkPolicy.KEEP,
+            createPeriodicWorkerRequest(addonId)
+        )
+    }
+
+    /**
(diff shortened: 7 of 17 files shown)


components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
```diff
@@ -163,10 +185,18 @@ object WebExtensionSupport {
...
+    private fun registerInstalledExtension(
+        context: Context,
+        store: BrowserStore,
+        webExtension: WebExtension,
+        updateFrequency: UpdateFrequency
+    ) {
         installedExtensions[webExtension.id] = webExtension
         store.dispatch(WebExtensionAction.InstallWebExtensionAction(webExtension.toState()))
 
+        // Register extension for periodically checking  for new updates
```
Amejia481: This should only happens for external extensions not `isBuiltIn`.

components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
```diff
@@ -163,10 +185,18 @@ object WebExtensionSupport {
...
+        ExtensionUpdater.registerForFutureUpdates(context, updateFrequency, webExtension.id)
```
Amejia481: This should only happens for external extensions not `isBuiltIn`.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
```diff
@@ -192,6 +192,19 @@ class GeckoEngine(
...
     }
 
+    /**
+     * See [Engine.updateWebExtension].
+     */
+    override fun updateWebExtension(
+        extension: WebExtension,
+        onSuccess: (WebExtension?) -> Unit,
+        onError: (String, Throwable) -> Unit
+    ) {
```
Amejia481: Are`onSuccess` or `onError` callbacks  going to be called when GeckoView notifies us that the updated extension have added extra permissions and user interaction is required?

components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
```diff
@@ -133,7 +148,19 @@ object WebExtensionSupport {
...
+                registerInstalledExtension(context, store, webExtension, updateFrequency)
+            }
+
+            override fun onPermissionsChangedWhileUpdating(
+                current: WebExtension,
+                updated: WebExtension,
+                onPermissionsGranted: (Boolean) -> Unit
+            ) {
+                // TODO:
+                //  0 Change the state of the extensions? In which state is the [current] extension?
```
Amejia481: @psymoon @csadilek  I need some help to clarify how should be proceed.

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/WebExtensionReducer.kt
```diff
@@ -46,6 +46,18 @@ internal object WebExtensionReducer {
...
+            is WebExtensionAction.UpdateWebExtension -> {
+                val updatedExtension = action.updatedExtension
+                state.updateWebExtensionState(updatedExtension.id) { existingExtension ->
+                    existingExtension.copy(
+                        url = updatedExtension.url ?: existingExtension.browserActionPopupSession,
+                        browserAction = updatedExtension.browserAction
+                            ?: existingExtension.browserAction,
+                        browserActionPopupSession = updatedExtension.browserActionPopupSession
+                            ?: existingExtension.browserActionPopupSession
+                    )
```
Amejia481: I'm not sure, but I have hunch that could be missing something here

Amejia481:
As my intention is just to show you me code, I use`[ci skip]`  to avoid overhead the CI server and getting charged 💰  without getting any value of it.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
csadilek: @Amejia481 I didn't think so, no. There'll be a `PromptDelegate` which we will set using `setPromptDelegate` on the `WebExtensionController`. So, `onSuccess` will be called once the exension is updated, `onError` in case an actual error occurred, but we can double-check.

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/WebExtensionReducer.kt
csadilek: Yeah, there will be many more fields in the `WebExtension` object we get from GV, and we will get a new instance of the native `WebExtension` as well. So, I think it will be better to remove the old global `WebExtensionState` and add the new one based on the new object.

components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
```diff
@@ -176,6 +227,66 @@ object WebExtensionSupport {
...
+
+    /**
+     * Updates the provided [extensionId] by updating it in the the [store] and re-attaching
+     * all the handlers.
+     */
+    internal fun updateInstalledExtension(extensionId: String, onFinish : ((ExtensionUpdateStatus) -> Unit)) {
+        val extension = installedExtensions[extensionId]
+
+        if (extension == null) {
+            onFinish(ExtensionUpdateStatus.NotInstalled)
```
csadilek: Curious how do you see this happen? The user uninstalled the extension and the background worker to update is still running?

components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
```diff
@@ -231,4 +342,48 @@ object WebExtensionSupport {
     }
 
     private fun WebExtension.toState() = WebExtensionState(id, url)
+
+    internal object ExtensionUpdater {
```
csadilek: Sine we only use this for `Addon`s, `feature-addons` is a better fit.

components/concept/engine/src/main/java/mozilla/components/concept/engine/webextension/WebExtensionDelegate.kt
```diff
@@ -65,4 +65,21 @@ interface WebExtensionDelegate {
...
+     * Invoked when a web extension has changed its permissions while trying to update to a
+     * new version. This requires user interaction as the updated extension will not be installed,
+     * until the user grants the new permissions.
+     *
+     * @param current The current [WebExtension].
+     * @param updated The update [WebExtension] that requires extra permissions.
+     * @param onPermissionsGranted A callback to indicate if the new permissions from the [updated] extension
+     * are granted or not.
+     */
+    fun onPermissionsChangedWhileUpdating(
```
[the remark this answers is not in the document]
Amejia481: In the proposal document we have another callback (`onOptionalPrompt`) that will be called when the extension request extra permissions at run-time, not sure if the names could cause confusion.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
Amejia481: Verified. Only  `onUpdatePrompt` will be called if the permissions changed.

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/WebExtensionReducer.kt
Amejia481: 👍

components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
Amejia481: That maybe the situation, I'm adding api for `unregisterForFutureUpdates` to avoid that.

components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
Amejia481: 👍

components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
```diff
@@ -33,6 +42,12 @@ import mozilla.components.support.ktx.kotlinx.coroutines.flow.filterChanged
 object WebExtensionSupport {
     private val logger = Logger("mozac-webextensions")
 
+    /**
+     * Provides global access to the [BrowserStore] and [Engine] for background workers.
```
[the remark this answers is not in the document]
Amejia481: Moving to a separated class.

components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
Amejia481: 👍

components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
```diff
@@ -49,6 +64,47 @@ class AddonManager(
...
+            onFinish(Status.Error(message, exception))
+        }
+
+        engine.updateWebExtension(extension, onSuccess, onError)
+    }
+
+    @Suppress("EXPERIMENTAL_API_USAGE")
+    private fun registerNewExtensionsForUpdates() {
+        store.flowScoped { flow ->
+            flow.ifChanged { it.extensions }
```
Amejia481: @csadilek @psymoon  would you mien giving me a hand? I'm trying to observe only new extensions when they are installed to register them for checking for updates. But I end up getting the whole collection the ones that hasn't change and the new ones.

csadilek: @Amejia481 yes, that's by design, the way this API behaves is that you always first get the current state (change from no state observed -> current state), and then only changes (last observed state -> changed state).

psymoon: @Amejia481 [`Flow.distinctUntilChanged()`](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/distinct-until-changed.html) may be useful. You may need a mediator observable generally to map/collect the latest data set (speaking from Rx perspective)

Amejia481: Awesome, let me take a look at it, thanks!

csadilek: @Amejia481 @psymoon I don't think this is needed here. We know all existing addons here on startup and can register them for updates. Installing a new addon will also go through the `AddonManager` and we can register for updates then i.e. we're about to add an `installAddon` call to the manager. So, we don't actually need to go through the store here.

csadilek: In other words `addons` can't change without the `AddonManager` triggering it. So, I can just call `registerForFutureUpdates`, as needed.

csadilek: Does that make sense? We can pair up on that too.

Amejia481: Go it 👍 !

Amejia481: I will summit the pr with the `registerForFutureUpdates` function to be integrated in the follow up work.

csadilek: 👍

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
```diff
@@ -224,6 +224,19 @@ class GeckoEngine(
...
+        // GeckoView support for updating extensions hasn't been implemented yet
+        // TODO https://github.com/mozilla-mobile/android-components/issues/4500
```
csadilek: nit: I would link to this instead: https://bugzilla.mozilla.org/show_bug.cgi?id=1599581

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/WebExtensionReducer.kt
```diff
@@ -52,6 +52,10 @@ internal object WebExtensionReducer {
                     it.copy(browserAction = action.browserAction)
                 }
             }
+            is WebExtensionAction.UpdateWebExtension -> {
+                val updatedExtension = action.updatedExtension
```
csadilek: nit: would probably skip the `val`.

components/browser/state/src/test/java/mozilla/components/browser/state/action/WebExtensionActionTest.kt
```diff
@@ -188,4 +188,20 @@ class WebExtensionActionTest {
...
+
+        val store = BrowserStore(
+            initialState = BrowserState(
+                extensions = mapOf("id" to existingExtension)
+            )
+        )
+
+        store.dispatch(WebExtensionAction.UpdateWebExtension(updatedExtension)).joinBlocking()
+        assertEquals(updatedExtension, store.state.extensions.values.first())
+        assertTrue(updatedExtension === store.state.extensions.values.first())
```
csadilek: nit: you can use `assertSame` here.

components/concept/engine/src/main/java/mozilla/components/concept/engine/Engine.kt
```diff
@@ -140,6 +140,26 @@ interface Engine {
         onError: ((String, Throwable) -> Unit) = { _, _ -> }
     ): Unit = onError(id, UnsupportedOperationException("Web extension support is not available in this engine"))
 
+    /**
+     * Update the provided [extension] if a new version is available.
```
csadilek: nit: Updates...

components/concept/engine/src/main/java/mozilla/components/concept/engine/webextension/WebExtensionDelegate.kt
```diff
@@ -82,4 +82,20 @@ interface WebExtensionDelegate {
...
```
csadilek: Can we rename this to `onUpdatePermissionRequest` to be in line with `onInstallPermissionRequest` above?

components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
```diff
@@ -49,6 +60,35 @@ class AddonManager(
             throw AddonManagerException(throwable)
         }
     }
+
+    /**
+     * Try updates the provided [extensionId] by updating it in the the [store] and re-attaching
```
csadilek: nit: Updates the [Addon] with the provided id....

components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
```diff
@@ -49,6 +60,35 @@ class AddonManager(
...
+     * all the handlers.
+     */
+    fun tryUpdateExtension(extensionId: String, onFinish: ((Status) -> Unit)) {
```
csadilek: Since we're in the `AddonManager` here, let's not use `Extension` here. It gets a bit confusing. How about just `updateAddon`? And document that if no update is available, nothing will happen? Or if you prefer `maybeUpdateAddon`?

components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
```diff
@@ -49,6 +60,35 @@ class AddonManager(
...
+        val extension = WebExtensionSupport.installedExtensions[extensionId]
+
+        if (extension == null) {
+            onFinish(Status.NotInstalled)
+            return
+        }
+
+        val onSuccess: ((WebExtension?) -> Unit) = { updatedExtension ->
+            val status = if (updatedExtension == null) {
+                Status.NoNewAvailableUpdate
```
csadilek: nit: Would rename to `NoUpdateAvailable`

components/feature/addons/src/main/java/mozilla/components/feature/addons/update/AddonUpdater.kt
```diff
@@ -0,0 +1,196 @@
...
+import java.util.concurrent.TimeUnit
+import kotlin.coroutines.resume
+import kotlin.coroutines.suspendCoroutine
+
+/**
+ * Contract to define the behavior for updating a web extensions.
+ */
+interface AddonUpdater {
+    /**
+     * Register the given [extensionId] for periodically checking for new updates.
```
csadilek: nit: Registers..to periodically check for new updates.

components/feature/addons/src/main/java/mozilla/components/feature/addons/update/AddonUpdater.kt
```diff
@@ -0,0 +1,196 @@
...
+     * @param extensionId The unique id of the extension.
+     */
+    fun registerForFutureUpdates(extensionId: String)
+
+    /**
+     * Unregister the given [extensionId] for periodically checking for new updates.
```
csadilek: nit: Unregisters..

components/feature/addons/src/main/java/mozilla/components/feature/addons/update/AddonUpdater.kt
```diff
@@ -0,0 +1,196 @@
...
+    /**
+     * Indicates how often an extension should be updated.
+     * @property repeatInterval Integer indicating how often the update should happen.
+     * @property repeatIntervalTimeUnit The time unit of the [repeatInterval].
+     */
+    class Frequency(val repeatInterval: Long, val repeatIntervalTimeUnit: TimeUnit)
+}
+
+/**
+ * An implementation of [AddonUpdater] that usages the work manager api for scheduling new updates.
```
csadilek: that uses...

components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
```diff
@@ -189,6 +187,25 @@ object WebExtensionSupport {
...
     }
 
+    /**
+     * Marks the provided [updatedExtension] as updated in the [store].
+     */
+    fun markExtensionAsUpdated(store: BrowserStore, updatedExtension: WebExtension) {
+        installedExtensions[updatedExtension.id] = updatedExtension
+        store.dispatch(WebExtensionAction.UpdateWebExtension(updatedExtension.toState()))
+
+        // Register action handler for all existing engine sessions on the new extension
```
csadilek: nit: on the updated extension. An issue was filed today to get us an API so we don't have to do this per extension: https://bugzilla.mozilla.org/show_bug.cgi?id=1603559

We could link to it here and in the code above.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
Amejia481: Sure 👍

components/browser/state/src/test/java/mozilla/components/browser/state/action/WebExtensionActionTest.kt
Amejia481: Nice I didn't know about it :)

components/concept/engine/src/main/java/mozilla/components/concept/engine/webextension/WebExtensionDelegate.kt
Amejia481: 👍

components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
Amejia481: 👍

samples/browser/src/main/java/org/mozilla/samples/browser/DefaultComponents.kt
```diff
@@ -132,7 +137,7 @@ open class DefaultComponents(private val applicationContext: Context) {
 
     // Addons
     val addonManager by lazy {
-        AddonManager(store, addonCollectionProvider)
+        AddonManager(store, engine, addonCollectionProvider, addonUpdater)
```
psymoon: Hmm now we have `AddonManager` depends on `DefaultAddonUpdater` and `AddonCollectionProvider`. I wonder if we could have one consumer component declaration instead of needing to declare a multiple components (as it appears as of now most UI integrations are done through `AddonManager`). I'm curious to see if the simplicity of client side integration is worth looking into (since declaring multiple components can be confusing). Thoughts?

components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
```diff
@@ -49,6 +60,35 @@ class AddonManager(
...
+
+    /**
+     * Try updates the provided [extensionId] by updating it in the the [store] and re-attaching
+     * all the handlers.
+     */
+    fun tryUpdateExtension(extensionId: String, onFinish: ((Status) -> Unit)) {
+        val extension = WebExtensionSupport.installedExtensions[extensionId]
+
+        if (extension == null) {
+            onFinish(Status.NotInstalled)
```
psymoon: I find `Status.NotInstalled` a bit out of the scope for an update function vs. `tryUpdateExtension` taking in `extension` as the parameter, and `WebExtensionSupport.installedExtensions[extensionId]` and null check happening outside. 🤔

components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
Amejia481: Great, I'll add it!

components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
Amejia481: I'm using `Status.NotInstalled` to double check, we could just ignore when we don't find in `installedExtensions`  
 
When an extension is installed, we send the `extensionId` to the `workmanager`'s jobs database. It periodically calls `AddonUpdaterWorker` that calls `tryUpdateExtension`.

A situation where the user uninstalls the `extension` while a scheduled work was already in progress causing that the `extension` is not in `installedExtensions`. I'm using  `NotInstalled` to let the callers take action about it. When this happens`AddonUpdaterWorker` is  removing the job from the `workmanager`'s jobs database.

samples/browser/src/main/java/org/mozilla/samples/browser/DefaultComponents.kt
csadilek: Yes, so it's definitely worth looking into at some point, but seeing our current component graph I don't think we need to do this as part of this PR: https://github.com/mozilla-mobile/android-components/blob/master/samples/browser/src/main/java/org/mozilla/samples/browser/DefaultComponents.kt

Like there's lots than can be optimized. In reference-browser, this looks much cleaner as we created component groups by that also has disadvantages (all components in a group are initialized together which is tricky for startup performance).

Amejia481:
To keep moving let's land this and open follow up issues :)

bors r=psymoon,csadilek

samples/browser/src/main/java/org/mozilla/samples/browser/DefaultComponents.kt
psymoon: I agree with optimization being out of this PR

PR #5270, opened 2019-12-09
Closes #5268: Filter out private tabs in RemoteTabsFeature

### LINKED ISSUE #5268: Filter-out private tabs in remote tabs

eoger:
r? @jonalmeida

### DIFF
--- a/components/feature/remotetabs/src/main/java/mozilla/components/feature/remotetabs/RemoteTabsFeature.kt
+++ b/components/feature/remotetabs/src/main/java/mozilla/components/feature/remotetabs/RemoteTabsFeature.kt
@@ -42,7 +42,7 @@ class RemoteTabsFeature(
                 val lastUsed = 0L
                 // TO-DO: https://github.com/mozilla-mobile/android-components/issues/5179
                 val iconUrl = null
-                state.tabs.map { tab ->
+                state.tabs.filter { !it.content.private }.map { tab ->
                     // TO-DO: https://github.com/mozilla-mobile/android-components/issues/1340
                     val history = listOf(TabEntry(tab.content.title, tab.content.url, iconUrl))
                     Tab(history, 0, lastUsed)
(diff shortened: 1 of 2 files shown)


components/feature/remotetabs/src/main/java/mozilla/components/feature/remotetabs/RemoteTabsFeature.kt
```diff
@@ -40,7 +40,10 @@ class RemoteTabsFeature(
                 val lastUsed = 0L
                 // TO-DO: https://github.com/mozilla-mobile/android-components/issues/5179
                 val iconUrl = null
-                state.tabs.map { tab ->
+                state.tabs.mapNotNull { tab ->
```
jonalmeida: Couldn't we just stick a filter before it or did I misunderstand what we wanted to do? 🤔 

```kotlin
state.tabs.filter { !it.content.private }
```

eoger: Yup that works too, I was wondering if the compiler would optimize the two array iterations.

jonalmeida:
SSL error when downloading gradle for the ui-icons build:
```
[task 2019-12-10T16:24:48.150Z] Exception in thread "main" javax.net.ssl.SSLHandshakeException: Remote host closed connection during handshake
```

components/feature/remotetabs/src/main/java/mozilla/components/feature/remotetabs/RemoteTabsFeature.kt
jonalmeida: The Kotlin compiler is quite good at generating optimized bytecode, and the AOT compiler on devices do even more things to reduce runtime costs, so in general, I don't think we need to think about those kinds of optimizations on Android any more.

PR #5273, opened 2019-12-10
Closes #4529: Add method to wait for all dispatched actions to complete

### LINKED ISSUE #4529: Store: Test helper for waiting for all dispatched actions to complete
See discussion:
https://github.com/mozilla-mobile/android-components/pull/4506#discussion_r327751819

When tests call code that dispatches actions on `BrowserStore` then there's no way to wait for those actions to be completed.

In some tests I dispatched a "dummy" action to be able to wait for all previous actions to be completed:

```
store.dispatch(SystemAction.LowMemoryAction).joinBlocking()
```

csadilek:
Putting this up for discussion. It fixes #4529, but doesn't restrict to tests. 

I had a case last week where I needed this (or something like it) outside of tests as well:
- `MethodA` dispatches actions
- `MethodB` awaits `MethodA`s completion and reads state from the store directly

=> There is no guarantee that `MethodB` sees the "latest" state in the store, as we're reducing on  a separate (dedicated) thread. In other words, `MethodB` can await `MethodA`'s completion, but not its desired state. This is definitely an edge case, as usually we'd be observing the store bound to a scope, but there are use cases where a method just wants to read the latest state from the store, and then it would be nice to be able to make sure it is in fact the latest state (includes state changes of already dispatched actions).

More specifically, my use case was getting the list of installed web extensions in an async callback from the engine, then dispatching actions to the store to save the extension's state, and having another component reading that state (just once). I wanted to guarantee that this state is always up-to-date.

@pocmo wdyt?

### DIFF
--- a/.buildconfig.yml
+++ b/.buildconfig.yml
@@ -268,6 +268,10 @@ projects:
     path: components/support/test
     description: 'A collection of helpers for testing components (local unit tests).'
     publish: true
+  support-test-libstate:
+    path: components/support/test-libstate
+    description: 'A collection of helpers for testing functionality that relies on the lib-state component (local unit tests).'
+    publish: true
   support-android-test:
     path: components/support/android-test
     description: 'A collection of helpers for testing components from instrumented (on device) tests.'
--- a/README.md
+++ b/README.md
@@ -216,6 +216,8 @@ _Supporting components with generic helper code._
 
 * 🔵 [**Test Appservices**](components/support/test-appservices/README.md) - A component for synchronizing Application Services' unit testing dependencies used in Android Components.
 
+* 🔵 [**Test LibState**](components/support/test-libstate/README.md) - A collection of helpers for testing functionality that relies on the lib-state component in local unit tests (`src/test`).
+
 * 🔵 [**Utils**](components/support/utils/README.md) - Generic utility classes to be shared between projects.
 
 * 🔵 [**Webextensions**](components/support/webextensions/README.md) - A component containing building blocks for features implemented as web extensions.
--- a/components/feature/contextmenu/build.gradle
+++ b/components/feature/contextmenu/build.gradle
@@ -44,6 +44,7 @@ dependencies {
     testImplementation Dependencies.testing_mockito
 
     testImplementation project(':support-test')
+    testImplementation project(':support-test-libstate')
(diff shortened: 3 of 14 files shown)


csadilek:
> My main concern here is that `processPendingActions()` waits on all children which means you may be waiting for much more than you intent to.

So, this comes from our current workaround not being ideal. We dispatch a random action and call `joinBlocking` on it. According to the docs we should and can't actually rely on coroutine execution order, but our current workaround does, as it assumes that when our pseudo action is processed all other actions are processed too, but that must not necessarily be the case.

> In addition to that you may run into a deadlock if you are blocking on the UI thread...

Yes, that's a problem, if the observers do that and we have a caller of `processPendingActions` on the UI thread. We'd have to document the use cases of this method for testing and for calling on non-ui thread to make sure the state is up-to-date. Definitely not ideal.

>In your example you mention that MethodB has the ability to await MethodAs completion. Doesn't that mean MethodA could return a job (or Deferred) that awaits for all combined dispatched actions by MethodA (since Store.dispatch() returns a Job).

Yes, but I also think that's a bit of  "weakness" in our current design that e.g. Redux doesn't have. So, I tried to make this case somewhat easier to handle. Alternatively, what do you think about bringing in our `syncDispatch` method (which is currently a store extension in Session)? This would also solve this as `MethodA` could just use that.

> as opposed to a "wait for all" which is kind of a moving target.

That's a bug in my implementation. I didn't mean for it to work that way. This would have to be synchronized on `dispatch` and only process pending actions at the time of invocation.

csadilek:
@pocmo OK, I've turned this into a test-only utility now to close #4529. When we get a chance we can discuss and revisit possible store API for this, but let's skip this for now. I didn't make the `scope` public and instead used reflection in the test utility, as we do for GeckoView.

The only wrinkle here is that I had to introduce a separate testing module so we don't introduce a dependency cycle, but this seems OK to me, as we might have more lib-state testing utilities later.

Can you take another look?

csadilek:
@pocmo ok renamed this to `waitUntilIdle`.

csadilek:
bors r=pocmo

pocmo:
> @pocmo ok renamed this to `waitUntilIdle`.

💯

Issue #5280, opened 2019-12-10
Expose LocaleManager clear() method


BranescuMihai:
Related to [Fenix-220](https://github.com/mozilla-mobile/fenix/issues/220), we need to manually clear the locale storage when the user selected the default from the locale list (so the app will follow the system one)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-4605)

csadilek:
This landed back in https://github.com/mozilla-mobile/android-components/pull/5282.

PR #5282, opened 2019-12-10
For #5280 - expose clear method of LocaleManager

BranescuMihai:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/support/locale/src/main/java/mozilla/components/support/locale/LocaleManager.kt
+++ b/components/support/locale/src/main/java/mozilla/components/support/locale/LocaleManager.kt
@@ -7,7 +7,9 @@ package mozilla.components.support.locale
 import android.content.Context
 import android.content.SharedPreferences
 import android.content.res.Configuration
+import android.content.res.Resources
 import androidx.appcompat.app.AppCompatActivity
+import androidx.core.os.ConfigurationCompat
 import mozilla.components.support.base.R
 import mozilla.components.support.base.log.logger.Logger
 import java.util.Locale
@@ -46,6 +48,33 @@ object LocaleManager {
             }
         }
         return currentLocale
+    }
+
+    /**
+     * Change the current locale to the system defined one. As a result, [getCurrentLocale] will
+     * return null.
+     *
+     * After calling this function, to visualize the locale changes you have to make sure all your visible activities
+     * get recreated. If your app is using the single activity approach, this will be trivial just call
+     * [AppCompatActivity.recreate]. On the other hand, if you have multiple activity this could be tricky, one
+     * alternative could be restarting your application process see https://github.com/JakeWharton/ProcessPhoenix
+     *
+     */
+    fun resetToSystemDefault(context: Context) {
+        clear(context)
+        val locale = getSystemDefault()
+        locale?.let {
+            updateSystemLocale(locale)
+            updateConfiguration(context, locale)
+        }
+    }
+
+    /**
+     * Returns the locale set by the system
+     */
+    fun getSystemDefault(): Locale? {
+        val config = Resources.getSystem().configuration
+        return ConfigurationCompat.getLocales(config).get(0)
     }
 
     internal fun updateResources(baseContext: Context): Context {
--- a/components/support/locale/src/test/java/mozilla/components/support/locale/LocaleManagerTest.kt
+++ b/components/support/locale/src/test/java/mozilla/components/support/locale/LocaleManagerTest.kt
@@ -12,6 +12,7 @@ import org.junit.Assert.assertNull
 import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
+import org.robolectric.annotation.Config
 import java.util.Locale
 
 @RunWith(AndroidJUnit4::class)
@@ -20,12 +21,11 @@ class LocaleManagerTest {
     @Before
     fun setup() {
         LocaleManager.clear(testContext)
-        Locale.setDefault("en_US".toLocale())
     }
 
     @Test
+    @Config(qualifiers = "en-rUS")
     fun `changing the language to Spanish must change the system locale to Spanish and change the configurations`() {
-
         var currentLocale = LocaleManager.getCurrentLocale(testContext)
 
         assertNull(currentLocale)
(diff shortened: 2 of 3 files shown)


components/support/locale/src/main/java/mozilla/components/support/locale/LocaleManager.kt
```diff
@@ -48,6 +48,14 @@ object LocaleManager {
         return currentLocale
     }
 
+    /**
+     * Remove the locale code from the storage. Useful when having the option for a default (use
+     * system locale), in which case [getCurrentLocale] will return null
+     */
```
Amejia481: nit: I think it will better to just mention the intended behavior without mentioning the internal details as they could change in the future.

components/support/locale/src/main/java/mozilla/components/support/locale/LocaleManager.kt
```diff
@@ -48,6 +48,14 @@ object LocaleManager {
...
+    fun clear(context: Context) {
```
Amejia481: Maybe instead of making the `clear` function public, we could rename it to something more meaningful, something like`resetToSystemDefault` that will better inform the consumer what is happening. What do you think? :)

Amejia481:
As part of the pr, let's also add an entry to the [change log](https://github.com/mozilla-mobile/android-components/blob/master/docs/changelog.md), mentioning the changes, this way other people will be aware of the changes :)

components/support/locale/src/main/java/mozilla/components/support/locale/LocaleManager.kt
BranescuMihai: Changed the docs, it's now similar to the one left by you above, thanks

components/support/locale/src/main/java/mozilla/components/support/locale/LocaleManager.kt
BranescuMihai: Your proposed name is actually much better, thanks! However, it got me thinking that actually the whole resetting should be done in this class (previously, I did it in an extension file for LocaleManager in Fenix), therefore I moved back here some logic. 

One issue I have with this is that while on Fenix I could easily test this with Mockk, on AC it seems we don't have Mockk or PowerMock, so I don't see a simple way of mocking Resources.getSystem() static method. If you have any ideas regarding this please let me know!:)

BranescuMihai:
> As part of the pr, let's also add an entry to the [change log](https://github.com/mozilla-mobile/android-components/blob/master/docs/changelog.md), mentioning the changes, this way other people will be aware of the changes :)

Added a new entry to the log

components/support/locale/src/main/java/mozilla/components/support/locale/LocaleManager.kt
Amejia481: An alternative could passing the context to `getSystemDefault` and this way you could mock the context functionality, using something like       
`ConfigurationCompat.getLocales(testContext.resources.configuration)`. What do you think?

components/support/locale/src/main/java/mozilla/components/support/locale/LocaleManager.kt
```diff
@@ -48,6 +50,34 @@ object LocaleManager {
...
+        }
+    }
+
+    /**
+     * Returns the locale set by the system
+     */
+    @Suppress("DEPRECATION")
+    fun getSystemDefault(): Locale? {
+        val config = Resources.getSystem().configuration
+        return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) config.locales.get(0) else config.locale
```
Amejia481: Maybe [ConfigurationCompat.getLocales ](https://developer.android.com/reference/androidx/core/os/ConfigurationCompat) could safe us the API checking :)

components/support/locale/src/main/java/mozilla/components/support/locale/LocaleManager.kt
BranescuMihai: It's important to note that the LocaleAwareApplication overrides the app's resources at startup, so in order to get the real system chosen locale, we need to use Resources.getSystem() instead of our resources from an injected context. So I would keep the usage of these system resources inside this manager, instead of relying on the client to pass it, since it's independent of the app itself

components/support/locale/src/main/java/mozilla/components/support/locale/LocaleManager.kt
BranescuMihai: Yep, great find! Changed to this compat class

components/support/locale/src/main/java/mozilla/components/support/locale/LocaleManager.kt
```diff
@@ -48,6 +50,33 @@ object LocaleManager {
...
+     * Change the current locale to the system defined one.
+     * [getCurrentLocale] will return null
+     *
+     * After calling this function, to visualize the locale changes you have to make sure all your visible activities
+     * get recreated. If your app is using the single activity approach, this will be trivial just call
+     * [AppCompatActivity.recreate]. On the other hand, if you have multiple activity this could be tricky, one
+     * alternative could be restarting your application process see https://github.com/JakeWharton/ProcessPhoenix
+     *
+     */
+    fun resetToSystemDefault(context: Context) {
```
Amejia481: Would you mind adding a test for this section? If you need any help just let me know eager to help!

components/support/locale/src/main/java/mozilla/components/support/locale/LocaleManager.kt
```diff
@@ -48,6 +50,33 @@ object LocaleManager {
         return currentLocale
     }
 
+    /**
+     * Change the current locale to the system defined one.
+     * [getCurrentLocale] will return null
```
Amejia481: nit: What do you think about ` "Change the current locale to the system defined one. As a result, [getCurrentLocale] will return null"`?

components/support/locale/src/main/java/mozilla/components/support/locale/LocaleManager.kt
Amejia481: 👍

components/support/locale/src/main/java/mozilla/components/support/locale/LocaleManager.kt
BranescuMihai: Changed, thanks

components/support/locale/src/main/java/mozilla/components/support/locale/LocaleManager.kt
BranescuMihai: Added, turns out there's a [config annotation for robolectric](http://robolectric.org/device-configuration/), which solved my testing problem.

Amejia481: Awesome :)

PR #5285, opened 2019-12-10
Closes #5284: Adds progress bar to download notification

### LINKED ISSUE #5284: Show progress of download on notification
For fenix https://github.com/mozilla-mobile/fenix/issues/4525

sblatz:
---

DEMO: 
progress bar 2019-12-19 14_20_54

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
+++ b/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
@@ -22,6 +22,7 @@ import android.os.IBinder
 import android.os.ParcelFileDescriptor
 import android.provider.MediaStore
 import android.widget.Toast
+import androidx.annotation.GuardedBy
 import androidx.annotation.VisibleForTesting
 import androidx.core.app.NotificationManagerCompat
 import androidx.core.content.FileProvider
@@ -30,6 +31,10 @@ import androidx.localbroadcastmanager.content.LocalBroadcastManager
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers.IO
 import kotlinx.coroutines.Job
+import kotlinx.coroutines.MainScope
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
 import mozilla.components.browser.state.state.content.DownloadState
 import mozilla.components.concept.fetch.Client
@@ -62,6 +67,8 @@ import kotlin.random.Random
 @Suppress("TooManyFunctions", "LargeClass")
 abstract class AbstractFetchDownloadService : Service() {
 
+    private val notificationUpdateScope = MainScope()
+
     protected abstract val httpClient: Client
     @VisibleForTesting
     internal val broadcastManager by lazy { LocalBroadcastManager.getInstance(this) }
@@ -74,7 +81,7 @@ abstract class AbstractFetchDownloadService : Service() {
         var job: Job? = null,
         @Volatile var state: DownloadState,
         var currentBytesCopied: Long = 0,
-        @Volatile var status: DownloadJobStatus,
+        @GuardedBy("context") var status: DownloadJobStatus,
         var foregroundServiceId: Int = 0,
         var downloadDeleted: Boolean = false
     )
@@ -92,6 +99,7 @@ abstract class AbstractFetchDownloadService : Service() {
 
     internal val broadcastReceiver by lazy {
         object : BroadcastReceiver() {
+            @Suppress("LongMethod")
             override fun onReceive(context: Context, intent: Intent?) {
                 val downloadId =
                     intent?.extras?.getLong(DownloadNotification.EXTRA_DOWNLOAD_ID) ?: return
@@ -99,13 +107,17 @@ abstract class AbstractFetchDownloadService : Service() {
 
                 when (intent.action) {
                     ACTION_PAUSE -> {
-                        currentDownloadJobState.status = DownloadJobStatus.PAUSED
+                        synchronized(context) {
+                            currentDownloadJobState.status = DownloadJobStatus.PAUSED
+                        }
                         currentDownloadJobState.job?.cancel()
                         emitNotificationPauseFact()
                     }
 
                     ACTION_RESUME -> {
-                        currentDownloadJobState.status = DownloadJobStatus.ACTIVE
+                        synchronized(context) {
+                            currentDownloadJobState.status = DownloadJobStatus.ACTIVE
+                        }
 
                         currentDownloadJobState.job = CoroutineScope(IO).launch {
                             startDownloadJob(downloadId)
@@ -118,8 +130,9 @@ abstract class AbstractFetchDownloadService : Service() {
                         NotificationManagerCompat.from(context).cancel(
                             currentDownloadJobState.foregroundServiceId
                         )
-                        currentDownloadJobState.status = DownloadJobStatus.CANCELLED
-
+                        synchronized(context) {
+                            currentDownloadJobState.status = DownloadJobStatus.CANCELLED
+                        }
                         currentDownloadJobState.job?.cancel()
 
                         currentDownloadJobState.job = CoroutineScope(IO).launch {
@@ -134,7 +147,10 @@ abstract class AbstractFetchDownloadService : Service() {
                         NotificationManagerCompat.from(context).cancel(
                             currentDownloadJobState.foregroundServiceId
                         )
-                        currentDownloadJobState.status = DownloadJobStatus.ACTIVE
+
+                        synchronized(context) {
+                            currentDownloadJobState.status = DownloadJobStatus.ACTIVE
+                        }
 
                         currentDownloadJobState.job = CoroutineScope(IO).launch {
                             startDownloadJob(downloadId)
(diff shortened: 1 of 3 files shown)


components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -257,10 +262,29 @@ abstract class AbstractFetchDownloadService : Service() {
...
+        val downloadJobState = downloadJobs[downloadID] ?: return
+        val download = downloadJobState.state
+
+        val timerTask = object : TimerTask() {
+            override fun run() {
+                displayOngoingDownloadNotification(download, downloadJobState.currentBytesCopied)
+            }
+        }
+
+        notificationTimer.scheduleAtFixedRate(timerTask, 0, PROGRESS_UPDATE_INTERVAL)
```
pocmo: I am a bit confused about the lifetime of the tasks. We schedule one task for every download here. But in `startDownoadJob()` we cancel all tasks on the timer. So I am wondering:
* Doesn't the `cancel()` call in `startDownoadJob()` cause all other notifications to get stale except the new one?
* We never cancel a specific task after the download is done right? Only if the service stops and the `Timer` object gets garbage collected then they get away (but the docs are not very specific on when that happens).

I wonder if we could use a single Kotlin Coroutine to update all notifications, something like:

```Kotlin
val notificationUpdateScope = MainScope()

override fun onCreate() {
  // [..]
  notificationUpdateScope.launch {
     delay(...)
     updateProgress()
  }
}

override fun onDestroy() {
   // [..]
   notificationUpdateScope.cancel()
}

fun updateProgress() {
   // Update progress of all notifications.
}
```

Does this make sense? I think with a single coroutine it is much easier to understand its lifetime and avoiding multiple zombie threads being left behind. What do you think?

sblatz: All great points. I have updated this PR with the ideas you suggested.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -182,12 +195,33 @@ abstract class AbstractFetchDownloadService : Service() {
...
+     * Android rate limits notifications being sent, so we must send them on a delay so that
+     * notifications are not dropped
+     */
+    @Synchronized
+    private fun updateDownloadNotificationProgress() {
+        for (download in downloadJobs.values) {
+            if (download.status != DownloadJobStatus.ACTIVE) { continue }
+            // We must be synchronized here to avoid the status getting set to PAUSED on this line
+            // and then overwriting that with an ongoing notification anyway
+            displayOngoingDownloadNotification(download.state, download.currentBytesCopied)
```
sblatz: I'm having trouble with this race condition. Do you have any suggestions about how to solve it? I thought putting a synchronized on this and all other state setting would be sufficient but it doesn't seem so 😕 

@pocmo

pocmo: Oh, sorry, this has fallen through the cracks. I can try to help with that this week. What is happening in that race condition? Notifications do not update or with wrong state?

sblatz: What’s happening is a user presses PAUSE, the notification then gets a timed update to display the current progress (ACTIVE state), so the “pause” action gets eaten but the download is still paused and stuck in a state where the user sees it as “frozen”

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -329,9 +373,9 @@ abstract class AbstractFetchDownloadService : Service() {
 
     /**
      * Informs [mozilla.components.feature.downloads.manager.FetchDownloadManager] that a download
-     * has been completed.
+     * is no longer in progress due to being paused, completed, or failed
      */
-    private fun sendDownloadCompleteBroadcast(downloadID: Long, status: DownloadJobStatus) {
+    private fun sendDownloadNotInProgress(downloadID: Long, status: DownloadJobStatus) {
```
sblatz: Perhaps `sendDownloadFinished` is better? I know this name is ugly but I want to be clear about what's going on here.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -99,13 +105,17 @@ abstract class AbstractFetchDownloadService : Service() {
 
                 when (intent.action) {
                     ACTION_PAUSE -> {
-                        currentDownloadJobState.status = DownloadJobStatus.PAUSED
+                        synchronized(context) {
+                            currentDownloadJobState.status = DownloadJobStatus.PAUSED
+                        }
```
pocmo: Now that we synchronize you can remove the `@Volatile` from `DownloadJobState`. Instead you could try adding `@GuardedBy("context")` to the `status` property to denote that this needs to be sychronized on `context` when accessed/modified. That itself doesn't do anything but hopefully the IDE (or lint) will show a warning if you don't do that.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -182,12 +196,71 @@ abstract class AbstractFetchDownloadService : Service() {
...
+     * Android rate limits notifications being sent, so we must send them on a delay so that
+     * notifications are not dropped
+     */
+    private fun updateDownloadNotification() {
+        synchronized(context) {
+            for (download in downloadJobs.values) {
+                // Dispatch the corresponding notification based on the current status
+                val notification = when (download.status) {
+                    DownloadJobStatus.ACTIVE -> {
+                        synchronized(context) {
```
pocmo: It looks like you are already in a synchronized block here (the whole method is?).

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -182,12 +196,71 @@ abstract class AbstractFetchDownloadService : Service() {
...
+                    }
+
+                    DownloadJobStatus.FAILED -> {
+                        DownloadNotification.createDownloadFailedNotification(context, download.state)
+                    }
+
+                    DownloadJobStatus.CANCELLED -> { return }
+                }
+
+                synchronized(context) {
```
pocmo: Here as well. This is already in the outter synchronized block.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
sblatz: Ooo didn't know about that keyword. Thanks!

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
sblatz: Fixed :)

sblatz:
bors retry

sblatz:
bors retry

sblatz:
bors retry

pocmo:
One or two tests are blocking indefinitely.

sblatz:
bors retry

sblatz:
bors retry

sblatz:
bors retry

sblatz:
bors retry

sblatz:
bors cancel

sblatz:
bors retry

sblatz:
bors retry

jonalmeida:
@sblatz could you squash your follow-up commits into one before we merge? :)

sblatz:
@jonalmeida done! 😄

jonalmeida:
bors retry

external-content duckduckgo com

jonalmeida:
@sblatz these failures are out of memory errors from the download tests. Could you take a look at them? I'm guessing there are more intermittent problems with them that Sebastian mentioned.

sblatz:
bors retry

sblatz:
bors r=@pocmo

sblatz:
bors retry

sblatz:
I'm really at a loss here for what the out of memory could be caused by. It looks like it fails on every run of `when the cancel button is clicked onCancelDownload must be called` but *only* on CI. The whole build passes for me locally (and always passed on Github...?). When I debug sample browser I don't see a spike in memory consumption when following the steps of that test :\

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -257,18 +302,6 @@ abstract class AbstractFetchDownloadService : Service() {
...
-
-        NotificationManagerCompat.from(context).notify(
-            downloadJobs[download.id]?.foregroundServiceId ?: 0,
-            ongoingDownloadNotification
-        )
-    }
-
     @Suppress("ComplexCondition")
     internal fun performDownload(download: DownloadState) {
         val isResumingDownload = downloadJobs[download.id]?.currentBytesCopied ?: 0L > 0L
```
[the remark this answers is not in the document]
sblatz: 0L > 0L would always be false, right? 

The logic is: `isResumingDownload` is TRUE if the bytes copied is more than 0. if bytesCopied is null, we assume it's 0, and thus this would evaluate to false.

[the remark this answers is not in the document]
jonalmeida: Why evaluate instead of simplifying this to be `false`?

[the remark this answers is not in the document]
sblatz: I'm confused. I want to compare `currentBytesCopied` to 0L so I can't just do 

`downloadJobs[download.id]?.currentBytesCopied ?: false > 0L`

Is there some syntax I'm missing here? 😅

sblatz:
bors retry

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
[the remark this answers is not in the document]
jonalmeida: Try this so avoid confusion:
```kotlin
(downloadJobs[download.id]?.currentBytesCopied ?: 0L) > 0L
```

Otherwise, it looks like:
```kotlin
downloadJobs[download.id]?.currentBytesCopied ?: (0L > 0L)
```

sblatz:
bors r=@pocmo

sblatz:
TAKE THAT BORS.

PR #5288, opened 2019-12-11
Add peer-to-peer communication

espertus:
Add feature-p2p
---

### Pull Request checklist

### After merge


espertus:
In a pleasant surprise, it turned out not to be hard to inline images, so I've added that.

components/feature/p2p/src/main/java/mozilla/components/feature/p2p/view/P2PBar.kt
```diff
@@ -0,0 +1,180 @@
...
+import mozilla.components.feature.p2p.R
+
+private const val DEFAULT_VALUE = 0
+
+/**
+ * A toolbar for peer-to-peer communication between browsers. Setting [listener] causes the
+ * buttons to become active.
+ */
+@Suppress("TooManyFunctions")
+class P2PBar @JvmOverloads constructor(
```
jonalmeida: A lot of this logic should be removed from the View and put into a presenter (See other examples of interactor/presenter patterns in the codebase). We want our Android Views to do just the one thing: displaying the UI. If there is logic how to decide what to show we should put that in a presenter. TabsTrayPresenter and ToolbarPresenter are other examples.

components/feature/p2p/src/main/java/mozilla/components/feature/p2p/internal/P2PController.kt
```diff
@@ -0,0 +1,232 @@
...
+        nearbyConnection?.unregisterObservers()
+    }
+
+    @Synchronized
+    private fun reportError(msg: String) {
+        Logger.error(msg)
+        view.failure(msg)
+    }
+
+    private inline fun <reified T : ConnectionState> cast() =
```
jonalmeida: Having to do this cast doesn't feel great. 😐  I think we need to re-think adding functionality to a state object.

components/feature/p2p/src/main/java/mozilla/components/feature/p2p/internal/P2PController.kt
```diff
@@ -0,0 +1,232 @@
...
+            sendMessage(URL_INDICATOR, it)
+        }
+    }
+
+    @Suppress("MaxLineLength")
+    override fun onLoadData(context: Context, data: String, newTab: Boolean) {
+        // Store data in file to work around Mozilla bug 1598481, which makes
+        // loading a page from memory extremely inefficient. Instead, we will
+        // write it to a file and then load a URL referencing that file.
+        val file = File.createTempFile("moz", ".html")
```
jonalmeida: Probably need to do this IO in a coroutine instead of on the main thread.

components/feature/p2p/src/main/assets/extensions/p2p/manifest.json
```diff
@@ -0,0 +1,23 @@
...
+  "content_scripts": [
+    {
+      "matches": ["*://*/*"],
+      "js": [ "p2p.js" ],
+      "run_at": "document_idle"
+    }
+  ],
+  "permissions": [
+    "nativeMessaging",
+    "contextMenus",
```
jonalmeida: Where do we use all of these permissions like `contextMenus`, `downloads`, `history`, etc?

components/feature/p2p/src/main/assets/extensions/p2p/p2p.js
```diff
@@ -0,0 +1,136 @@
...
+            css.styleSheet.cssText = styles;
+        } else {
+            css.appendChild(document.createTextNode(styles));
+        }
+        window.document.getElementsByTagName('head')[0].appendChild(css);
+    }
+}
+
+// Adapted from https://stackoverflow.com/a/32596557/631051 [author: sleeplessnerd]
+function captureCss() {
```
jonalmeida: Note for self: review this again.

samples/browser/src/main/java/org/mozilla/samples/browser/BrowserFragment.kt
```diff
@@ -98,7 +120,21 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
...
         readerViewFeature.onBackPressed() || super.onBackPressed()
 
+    @CallSuper
+    override fun onRequestPermissionsResult(
+        requestCode: Int,
+        permissions: Array<String>,
+        grantResults: IntArray
+    ) {
+        if (requestCode == REQUEST_CODE_P2P_PERMISSIONS) {
+            p2pIntegration.get()!!.feature.onPermissionsResult(permissions, grantResults)
```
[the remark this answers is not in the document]
espertus: Yes, I should have known better.

samples/browser/src/main/java/org/mozilla/samples/browser/BaseBrowserFragment.kt
```diff
@@ -222,6 +222,7 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler {
         private const val REQUEST_CODE_DOWNLOAD_PERMISSIONS = 1
         private const val REQUEST_CODE_PROMPT_PERMISSIONS = 2
         private const val REQUEST_CODE_APP_PERMISSIONS = 3
+        internal const val MAX_REQUEST_CODE_PERMISSIONS = 3
```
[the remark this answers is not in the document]
espertus: The issue is that I had to declare `p2p*` in `BrowserFragment`, not `BaseBrowserFragment`. (I currently don't remember why, but I could probably rediscover the reason by trying to move the `p2p` code into `BaseBrowserFragment`.) Thus, I couldn't reference `p2pFeature` in `BaseBrowserFragment`.

components/feature/p2p/src/main/java/mozilla/components/feature/p2p/internal/P2PController.kt
espertus: We could eliminate the `cast` function, but then we'd have to do a lot of `as?` casts. Would I be correct in assuming that's just as bad?

components/feature/p2p/src/main/java/mozilla/components/feature/p2p/internal/P2PController.kt
espertus: Of course. I'm pleased to get a chance to use a coroutine. They're much nicer than `AsyncTask`.

components/feature/p2p/src/main/assets/extensions/p2p/manifest.json
espertus: I should have cleaned that up after copying from another extension.

components/feature/p2p/src/main/java/mozilla/components/feature/p2p/internal/P2PController.kt
jonalmeida: It's fine. I think it's the `cast<ConnectionState.Authenticating>()?.accept()` state calls that require the cast in order to invoke feel awkward to have to use, but I have no good suggestions right now. 🤷‍♂

samples/browser/src/main/java/org/mozilla/samples/browser/BaseBrowserFragment.kt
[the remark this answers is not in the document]
espertus: I've cleaned all of this up.

components/feature/p2p/src/main/java/mozilla/components/feature/p2p/view/P2PBar.kt
espertus: @jonalmeida I'm looking again at this code while my teammates are away. I know you are too, so I won't expect a quick reply.

I want to make sure I understand how Presenter works with MVC. Would the presenter mediate all of the interactions between the view and the controller? `V <--> P <--> C`

espertus:
@jonalmeida Thanks for your help! Please take another look. I particularly welcome your input on the division from P2PController into P2PInteractor/P2PPresenter. As written, they both need access to the same `NearbyConnection`, which is handled kind of clumsily, and they share a hashmap of in-progress messages, which is handled less clumsily.

components/feature/p2p/src/main/java/mozilla/components/feature/p2p/internal/P2PPresenter.kt
```diff
@@ -0,0 +1,132 @@
...
+        view.failure(msg)
+    }
+
+    internal fun stop() {
+        nearbyConnection?.unregisterObservers()
+    }
+
+    companion object {
+        private var savedConnectionState: ConnectionState? = null
+        internal var nearbyConnection: NearbyConnection? = null // TODO: lock up
```
jonalmeida: Not sure what this TODO means.

components/feature/p2p/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,81 @@
+<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
jonalmeida: @espertus nit: Can we rename this file to `mozac_feature_p2p_strings.xml`?

components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionController.kt
```diff
@@ -109,7 +109,8 @@ class WebExtensionController(
             installedExtensions[extensionId]?.let { ext ->
                 val port = ext.getConnectedPort(name, session)
                 port?.postMessage(msg)
-                        ?: logger.error("No port connected for provided session. Message $msg not sent.")
+                        ?: logger.error(
+                            "No port with name $name connected for provided session. Message $msg not sent.")
```
jonalmeida: I think your auto-linter might have touched this file in error.

components/feature/p2p/src/main/assets/extensions/p2p/p2p.js
```diff
@@ -0,0 +1,136 @@
...
+                }
+            } else {
+                for (var j = 0; j < sheet.rules.length; j++) {
+                    insertStyle(sheets[i].rules[j].cssText);
+                }
+            }
+        }
+    }
+    // Remove stylesheet links.
+    window.document.querySelectorAll('link[rel="stylesheet"]')forEach(e => e.parentNode.removeChild(e));
```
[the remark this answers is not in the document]
jonalmeida: @espertus looks like we caught that js error a while back, I didn't even remember myself! 😅

samples/browser/src/main/java/org/mozilla/samples/browser/BaseBrowserFragment.kt
```diff
@@ -185,6 +190,24 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler {
             owner = this,
             view = layout)
 
+        p2pIntegration.set(
+                feature = P2PIntegration(
+                        store = components.store,
+                        engine = components.engine,
+                        view = layout.p2p,
+                        thunk = { -> components.nearbyConnection },
```
jonalmeida:
```suggestion
                        thunk = { components.nearbyConnection },
```

You can omit the arrow if there aren't any arguments for the callback.

samples/browser/src/main/java/org/mozilla/samples/browser/BaseBrowserFragment.kt
```diff
@@ -185,6 +190,24 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler {
             owner = this,
             view = layout)
 
+        p2pIntegration.set(
+                feature = P2PIntegration(
```
jonalmeida: nit: indentation seems off, the linter might complain.

samples/browser/src/main/java/org/mozilla/samples/browser/BaseBrowserFragment.kt
```diff
@@ -212,6 +235,7 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler {
             REQUEST_CODE_DOWNLOAD_PERMISSIONS -> downloadsFeature.get()
             REQUEST_CODE_PROMPT_PERMISSIONS -> promptFeature.get()
             REQUEST_CODE_APP_PERMISSIONS -> sitePermissionsFeature.get()
+            REQUEST_CODE_P2P_PERMISSIONS -> p2pFeature
```
jonalmeida:
```suggestion
            REQUEST_CODE_P2P_PERMISSIONS -> p2pIntegration.get()?.feature
```

It's not great, but you could reference the inner feature directly from the integration in order to remove this second reference to the feature.

```
p2pFeature = p2pIntegration.get()?.feature
```

samples/browser/src/main/java/org/mozilla/samples/browser/BrowserFragment.kt
```diff
@@ -81,7 +81,11 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
         )
 
         thumbnailsFeature.set(
-            feature = ThumbnailsFeature(requireContext(), layout.engineView, components.sessionManager),
+            feature = ThumbnailsFeature(
```
jonalmeida: Same here regarding auto-format.

espertus:
> What's a thunk? 🙂

From the start of the [Wikipedia article](https://en.wikipedia.org/wiki/Thunk):

> In computer programming, a thunk is a subroutine used to inject an additional calculation into another subroutine. Thunks are primarily used to delay a calculation until its result is needed, or to insert operations at the beginning or end of the other subroutine. 

It's a pretty common term in the functional programming community.

Would you like for me to use a more specific term, such as `nearbyConnectionCreator`?

espertus:
> Not sure what this TODO means.

I was unsure whether it was okay to share the `NearbyConnection` between `P2PPresenter` and `P2PInteractor` like this and was hoping for your opinion.

espertus:
> Can we rename this file to mozac_feature_p2p_strings.xml?

Certainly. I believe you had me change it to a non-standard name before to keep it from being picked up by the translation framework.

espertus:
> I think your auto-linter might have touched this file in error.

If you look closely, you'll see that I changed the string, which made the line longer, so I let the auto-linter split it. I've put it back on one line, if that better fits the project's style.

espertus:
I've made the changes you recommended. The only issues found by ktlint or detekt are the two `TODO` comments. I await your opinion on how I split data and functionality between `P2PPresenter` and `P2PInteractor`. Thank you, @jonalmeida .

jonalmeida:
> It's a pretty common term in the functional programming community.
> 
> Would you like for me to use a more specific term, such as nearbyConnectionCreator?

Neat, I learnt something new today. 🙂  This sounds similar to our provider strategy that we follow as well ([example](https://searchfox.org/mozilla-mobile/source/android-components/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt#60)). Never used that term before though. Maybe `connectionProvider`?

> I was unsure whether it was okay to share the NearbyConnection between P2PPresenter and P2PInteractor like this and was hoping for your opinion.

Through the constructor is usually best. When you create your presenter/interactor, you can invoke the `thunk` in the constructor for both (`thunk.invoke()`). If the change turns out to be rather large, we could always clean this up in a follow-up. Since those references are in a companion object which are like thread-safe singletons, so I imagine this is safe.

> Certainly. I believe you had me change it to a non-standard name before to keep it from being picked up by the translation framework.

Oh, I'm sorry, I was still able to see the `strings.xml` file so I thought the comment was missed. Might have been my error.

> If you look closely, you'll see that I changed the string, which made the line longer, so I let the auto-linter split it. I've put it back on one line, if that better fits the project's style.

Sorry, my mistake! Follow the linter in that case.

> The only issues found by ktlint or detekt are the two TODO comments.

And one more lint!

```
[task 2020-03-27T21:18:36.730Z]   /builds/worker/checkouts/src/components/feature/p2p/src/main/res/layout/mozac_feature_p2p_view.xml:81: Error: Using android:src to define resource instead of app:srcCompat [AndroidSrcXmlDetector]
[task 2020-03-27T21:18:36.730Z]           android:src="@drawable/mozac_ic_close"
```

espertus:
> > It's a pretty common term in the functional programming community.
> > Would you like for me to use a more specific term, such as nearbyConnectionCreator?
> 
> Neat, I learnt something new today. 🙂 This sounds similar to our provider strategy that we follow as well ([example](https://searchfox.org/mozilla-mobile/source/android-components/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt#60)). Maybe `connectionProvider`?

Sure. Done.

> > I was unsure whether it was okay to share the NearbyConnection between P2PPresenter and P2PInteractor like this and was hoping for your opinion.
> 
> Through the constructor is usually best. When you create your presenter/interactor, you can invoke the `thunk` in the constructor for both (`thunk.invoke()`).

If each applied the thunk, they'd get a different `NearbyConnection`. Additionally, I found that I needed to do the invoking in `P2PPresenter.start()`, not its constructors.

> If the change turns out to be rather large, we could always clean this up in a follow-up. Since those references are in a companion object which are like thread-safe singletons, so I imagine this is safe.

I just made a minor improvement by making the setter private.

> > The only issues found by ktlint or detekt are the two TODO comments.

Removed.

> And one more lint!
> 
> ```
> [task 2020-03-27T21:18:36.730Z]   /builds/worker/checkouts/src/components/feature/p2p/src/main/res/layout/mozac_feature_p2p_view.xml:81: Error: Using android:src to define resource instead of app:srcCompat [AndroidSrcXmlDetector]
> [task 2020-03-27T21:18:36.730Z]           android:src="@drawable/mozac_ic_close"
> ```

Thank you! I did not know about that linter.

I have rebased, squashed, tested, etc.

jonalmeida:
> If each applied the thunk, they'd get a different `NearbyConnection`. Additionally, I found that I needed to do the invoking in `P2PPresenter.start()`, not its constructors.

Wouldn't the singleton instance passed it avoid that? The `thunk` is invoked, multiple times, but the instance would always be the same.

Let's land this regardless and iterate!

jonalmeida:
bors r-

Ah sorry, the changelog entry is in the wrong release. Let's fix that first.

espertus:
> Wouldn't the singleton instance passed it avoid that? The `thunk` is invoked, multiple times, but the instance would always be the same.

Yes, I forgot that it was backed by a lazy singleton. Fixed.

I've also fixed the changelog.

PR #5291, opened 2019-12-11
Issue #4500: Add API for installing/uninstalling third-party extensions

csadilek:
This brings in the engine/store APIs to install third-party extensions, as well as uninstall (for both built-in and external extensions). I prefer having a single `installWebExtension` method (as opposed to separate ones for built-in and external extensions), because it keeps install/uninstall symmetric and the API surface smaller. We can revisit that later though once the new `installBuiltIn` lands in GV.

I am going to add the `AddonManager` and sample-browser bits in a separate PR.

### DIFF
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
@@ -159,35 +159,67 @@ class GeckoEngine(
         onSuccess: ((WebExtension) -> Unit) = { },
         onError: ((String, Throwable) -> Unit) = { _, _ -> }
     ) {
-        if (ext.supportActions) {
-            // We currently have to install the global action handler before we
-            // install the extension which is why this is done here directly.
-            // This code can be removed from the engine once the new GV addon
-            // management API lands. Then the global handlers will be invoked
-            // with the latest state whenever they are registered:
-            // https://bugzilla.mozilla.org/show_bug.cgi?id=1599897
-            // https://bugzilla.mozilla.org/show_bug.cgi?id=1582185
-            ext.registerActionHandler(object : ActionHandler {
-                override fun onBrowserAction(extension: WebExtension, session: EngineSession?, action: BrowserAction) {
-                    webExtensionDelegate?.onBrowserActionDefined(extension, action)
-                }
+        val actionHandler = object : ActionHandler {
+            override fun onBrowserAction(extension: WebExtension, session: EngineSession?, action: BrowserAction) {
+                webExtensionDelegate?.onBrowserActionDefined(extension, action)
+            }
 
-                override fun onToggleBrowserActionPopup(
-                    extension: WebExtension,
-                    action: BrowserAction
-                ): EngineSession? {
-                    val session = GeckoEngineSession(runtime)
-                    return webExtensionDelegate?.onToggleBrowserActionPopup(extension, session, action)
-                }
-            })
+            override fun onToggleBrowserActionPopup(extension: WebExtension, action: BrowserAction): EngineSession? {
+                return webExtensionDelegate?.onToggleBrowserActionPopup(extension, GeckoEngineSession(runtime), action)
+            }
         }
 
-        runtime.registerWebExtension(ext.nativeExtension).then({
-            webExtensionDelegate?.onInstalled(ext)
-            onSuccess(ext)
+        if (ext.isBuiltIn()) {
+            if (ext.supportActions) {
+                // We currently have to install the global action handler before we
+                // install the extension which is why this is done here directly.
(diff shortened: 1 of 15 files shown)


components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/webextension/GeckoWebExtension.kt
```diff
@@ -218,6 +222,25 @@ class GeckoWebExtension(
...
         return geckoSession.getWebExtensionActionDelegate(nativeExtension) != null
     }
+
+    /**
+     * See [WebExtension.getMetadata].
+     */
+    override fun getMetadata(): Metadata? {
+        return nativeExtension.metaData?.let {
+            Metadata(
+                name = it.name,
```
Amejia481: I guess we can't show this info to the user as I will be not localized, we have to rely on the addon counterpart, right?

csadilek: Yes, exactly. We're going to use what we have in our `Addon` in the UI, but I wanted us to have access to all fields in case we need them.

csadilek: The string are localized based on the GeckoView language setting.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/webextension/GeckoWebExtension.kt
```diff
@@ -218,6 +222,25 @@ class GeckoWebExtension(
...
+                description = it.description,
+                developerName = it.creatorName,
+                developerUrl = it.creatorUrl,
+                homePageUrl = it.homepageUrl,
+                version = it.version,
+                permissions = it.permissions.toList(),
```
Amejia481: We could use `Addon.permissionToTranslation` to localize this one.

https://github.com/mozilla-mobile/android-components/blob/e29cc1990413a93a1b243ae7f131bbe18a66a0e2/components/feature/addons/src/main/java/mozilla/components/feature/addons/Addon.kt#L120

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/webextension/GeckoWebExtension.kt
Amejia481: 👍

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/webextension/GeckoWebExtension.kt
csadilek: Yes, 👍 , but not here in the engine. That's for the follow-up when we integrate with `feature-addons`.

Amejia481: 👍

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
```diff
@@ -159,35 +160,67 @@ class GeckoEngine(
...
-                ): EngineSession? {
-                    val session = GeckoEngineSession(runtime)
-                    return webExtensionDelegate?.onToggleBrowserActionPopup(extension, session, action)
-                }
+            override fun onToggleBrowserActionPopup(extension: WebExtension, action: BrowserAction): EngineSession? {
+                return webExtensionDelegate?.onToggleBrowserActionPopup(extension, GeckoEngineSession(runtime), action)
+            }
+        }
+
+        if (Uri.parse(ext.url).scheme == "resource") {
```
Amejia481: Btw,  adding `isBuiltIn` to the properties of the extension is going to be part of the follow up?

csadilek:
bors abort

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
csadilek: @Amejia481 hm let me add it here, I think that's easy and we'll need it.

Amejia481: Thanks, I will use it on the update part :)

Amejia481:
bors r=Amejia481,psymoon

Amejia481:
bors retry

PR #5292, opened 2019-12-11
Closes #5243: Split AppLinksFeature into feature and interceptor

### LINKED ISSUE #5243: Investigate alternative solution to intercept load requests for AppLinksFeature
Current `AppLinksFeature` is a observer.  However, this is not the best design.  There are issues with the observer architecture that needs to be fixed.  Investigate into how this issue can be solved.

rocketsroger:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -33,6 +33,7 @@ import mozilla.components.support.ktx.android.util.Base64
 import mozilla.components.support.ktx.kotlin.isEmail
 import mozilla.components.support.ktx.kotlin.isGeoLocation
 import mozilla.components.support.ktx.kotlin.isPhone
+import mozilla.components.support.ktx.kotlin.tryGetHostFromUrl
 import org.json.JSONObject
 import org.mozilla.geckoview.AllowOrDeny
 import org.mozilla.geckoview.ContentBlocking
@@ -353,50 +354,37 @@ class GeckoEngineSession(
                 return GeckoResult.fromValue(AllowOrDeny.ALLOW)
             }
 
+            val isSameDomain = this@GeckoEngineSession.currentUrl?.tryGetHostFromUrl() ==
+                request.uri.tryGetHostFromUrl()
             val response = settings.requestInterceptor?.onLoadRequest(
                 this@GeckoEngineSession,
-                request.uri
+                request.uri,
+                request.hasUserGesture,
+                isSameDomain
             )?.apply {
                 when (this) {
                     is InterceptionResponse.Content -> loadData(data, mimeType, encoding)
                     is InterceptionResponse.Url -> loadUrl(url)
+                    is InterceptionResponse.AppIntent -> {
+                        notifyObservers {
+                            onLaunchIntentRequest(url = url, appIntent = appIntent)
+                        }
+                    }
                 }
             }
 
             return if (response != null) {
                 GeckoResult.fromValue(AllowOrDeny.DENY)
-            } else if (!isObserved()) {
-                GeckoResult.fromValue(AllowOrDeny.ALLOW)
             } else {
-                val geckoResult: GeckoResult<AllowOrDeny> = GeckoResult()
-                var completedBy: String? = null
-                val allowOrDeny: (Boolean, String) -> Unit = { shouldAllow, newCaller ->
-
-                    /* Debugging code for Android-components/issues/5127, will remove */
-                    if (completedBy != null) {
-                        throw IllegalStateException("GeckoResult already completed by $completedBy, " +
-                            "new caller is $newCaller")
-                    }
-
-                    val result = if (shouldAllow) AllowOrDeny.ALLOW else AllowOrDeny.DENY
-                    geckoResult.complete(result)
-                    /* Debugging code for Android-components/issues/5127, will remove */
-                    completedBy = newCaller
-                }
-
                 notifyObservers {
-                    // Unlike the name LoadRequest.isRedirect may imply this flag is not about http redirects. The flag
-                    // is "True if and only if the request was triggered by an HTTP redirect."
-                    // See: https://bugzilla.mozilla.org/show_bug.cgi?id=1545170
                     onLoadRequest(
                         url = request.uri,
                         triggeredByRedirect = request.isRedirect,
-                        triggeredByWebContent = request.hasUserGesture,
-                        shouldLoadUri = allowOrDeny
+                        triggeredByWebContent = request.hasUserGesture
                     )
                 }
 
-                geckoResult
+                GeckoResult.fromValue(AllowOrDeny.ALLOW)
             }
         }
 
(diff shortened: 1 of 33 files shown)


components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -243,19 +250,20 @@ class SystemEngineView @JvmOverloads constructor(
...
+                                    Logger.error("failed to start third party app activity", e)
+                                }
+                                super.shouldInterceptRequest(view, request)
+                            }
                         }
                     }
                 }
             }
 
-            if (request.isForMainFrame) {
```
jonalmeida: I'm not sure we should be removing the `isForMainFrame` check here.

components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
```diff
@@ -76,17 +77,16 @@ abstract class EngineSession(
...
          * @param url The string url that was requested.
+         * @param appIntent The Androio Intent that was requested.
          * @param triggeredByRedirect True if and only if the request was triggered by an HTTP redirect.
          * @param triggeredByWebContent True if and only if the request was triggered from within
          * web content (as opposed to via the browser chrome).
-         * @param shouldLoadUri Function to be called when request is consumed.
          */
         fun onLoadRequest(
-            url: String,
+            url: String?,
```
jonalmeida: Why did `url` on the EngineSession need to be nullable?

docs/changelog.md
```diff
@@ -18,6 +18,14 @@ permalink: /changelog/
     * `browser-engine-gecko-beta`: GeckoView 72.0
     * `browser-engine-gecko-nightly`: GeckoView 73.0
 
+* **feature-app-links**
+  * ⚠️ **This is a breaking change**:
+  * Feature now contains two parts.  One part is the AppLinksFeature, the other part is RequestInterceptor.
+  ```Kotlin
```
jonalmeida: nit: lowercase 'k' otherwise markdown doesn't recognize the highlighter language.

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
rocketsroger: I'll look into this more.

components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
rocketsroger: You're right, it doesn't need to be null.  Same with the `appIntent`.  It was old code that should have been cleaned up.  I'll remove.

docs/changelog.md
rocketsroger: Thanks, didn't know that.  Thought the correct format is uppercase.  I'll fix.

jonalmeida: We have some cases of uppercase in the changelog, but highlighting doesn't work on those.

rocketsroger: Should I fix them as well?

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
rocketsroger: Agreed that I shouldn't ignore observers for `SystemEngineView`.  Updated to match the design of `GeckoEngineSession` so its more consistent.  Now `request.isForMainFrame` is checked when handling the response from the interceptor.

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -355,48 +356,30 @@ class GeckoEngineSession(
 
             val response = settings.requestInterceptor?.onLoadRequest(
                 this@GeckoEngineSession,
-                request.uri
+                request.uri,
+                request.hasUserGesture,
+                this@GeckoEngineSession.currentUrl?.tryGetHostFromUrl() == request.uri.tryGetHostFromUrl()
```
csadilek: nit: I would assign this to a `val isSameDomain` to make it easier to read here. 

I also realizes now that your original solution to pass both urls to the interceptor was easier to read, but just a nit. :)

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -355,48 +356,30 @@ class GeckoEngineSession(
...
             )?.apply {
                 when (this) {
                     is InterceptionResponse.Content -> loadData(data, mimeType, encoding)
                     is InterceptionResponse.Url -> loadUrl(url)
+                    is InterceptionResponse.AppIntent -> {
+                        notifyObservers {
```
csadilek: That's a problem / breaking change, as you're now only notifying the observer for app intents. Fenix has an observer too: https://github.com/mozilla-mobile/fenix/blob/fe034226a335a2e48b4fa94e94aa80d7fe278f57/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt#L401

So, it would no longer be called now. Can you just notify in all cases with the app intent being null for regular load requests?

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
rocketsroger: Agreed.  I'll update.

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
rocketsroger: Understood.  I'll fix this.

components/browser/engine-system/src/test/java/mozilla/components/browser/engine/system/SystemEngineSessionTest.kt
```diff
@@ -459,74 +462,17 @@ class SystemEngineSessionTest {
         assertEquals("UTF-8", response.encoding)
     }
 
-    @Test
-    fun `shouldInterceptRequest notifies observers if request was not intercepted`() {
```
csadilek: It's not clear to me why we're removing these tests? Shouldn't they still work now, as we still call `onLoadRequest` as before (just need to remove the lambda?)

rocketsroger: You're correct.  I removed this earlier because these tests became not applicable when I removed the observer call from `SystemEngineView`.  But during this review it was added back in.  I should have remember to add these two tests back.  Thanks for catching this.

rocketsroger:
>     * Can we file a follow-up to "Migrate feature-app-links to browser-state" and put it on the browser-state board?

Issue created: https://github.com/mozilla-mobile/android-components/issues/5341

rocketsroger:
bors r=csadilek

Issue #5296, opened 2019-12-12
Improve a11y for error pages


mcarare:
Original issue in Fenix: https://github.com/mozilla-mobile/fenix/issues/3586.
Additionally, there are other issues like: list items announced twice.

mcarare:
Verified as fixed: https://github.com/mozilla-mobile/fenix/issues/3586#issuecomment-572123658.

PR #5299, opened 2019-12-12
For #5296 Improve a11y for error pages

mcarare:
Use role="presentation" for `<ul>` to remove `<li>` semantics for screen reader.
Remove `<p>` from errorShortDesc because paragraphs cannot contain lists.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/errorpages/src/main/java/mozilla/components/browser/errorpages/ErrorPages.kt
+++ b/components/browser/errorpages/src/main/java/mozilla/components/browser/errorpages/ErrorPages.kt
@@ -36,6 +36,7 @@ object ErrorPages {
             .replace("%button%", context.getString(errorType.refreshButtonRes))
             .replace("%messageShort%", context.getString(errorType.titleRes))
             .replace("%messageLong%", context.getString(errorType.messageRes, uri))
+            .replace("<ul>", "<ul role=\"presentation\">")
             .replace("%css%", css)
     }
 }
--- a/components/browser/errorpages/src/main/res/raw/error_pages.html
+++ b/components/browser/errorpages/src/main/res/raw/error_pages.html
@@ -5,8 +5,8 @@
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
 
 <html xmlns="http://www.w3.org/1999/xhtml">
-    <meta charset="UTF-8">
     <head>
+        <meta charset="UTF-8">
         <meta name="viewport" content="width=device-width; user-scalable=false;" />
         <title>%pageTitle%</title>
         <style>%css%</style>
@@ -26,7 +26,7 @@
 
                 <!-- Short Description -->
                 <div id="errorShortDesc">
-                    <p id="errorShortDescText">%messageLong%</p>
+                   %messageLong%
                 </div>
 
             </div>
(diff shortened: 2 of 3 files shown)


components/browser/errorpages/src/main/res/raw/error_pages.html
```diff
@@ -26,7 +26,7 @@ <h1 id="et_dnsNotFound" class="errorTitleText">%messageShort%</h1>
 
                 <!-- Short Description -->
                 <div id="errorShortDesc">
-                    <p id="errorShortDescText">%messageLong%</p>
```
Amejia481: Do you know why we were using this id `errorShortDescText`? Not sure if removing it could cause any breakage to some JavaScript or CSS that relies on it.

components/browser/errorpages/src/main/java/mozilla/components/browser/errorpages/ErrorPages.kt
```diff
@@ -36,6 +36,7 @@ object ErrorPages {
             .replace("%button%", context.getString(errorType.refreshButtonRes))
             .replace("%messageShort%", context.getString(errorType.titleRes))
             .replace("%messageLong%", context.getString(errorType.messageRes, uri))
+            .replace("<ul>", "<ul role=\"presentation\">")
```
Amejia481: This is a good start but in some cases is not going to match as consumers could use multiple variations of `<ul>`. In addition to this, I believe we have to document the issue and indicate to consumers what is the best approach. What do you think? :)

Amejia481:
Let's add an entry in the changelog :)

components/browser/errorpages/src/main/res/raw/error_pages.html
mcarare: There is no reference in CSS or code to this. The styling is done in the `<div>`. 
Either way, if it was used, it wouldn't work as expected. When parsing the code, after the `<ul>` element a `</p`> would be automatically inserted by the parser.

components/browser/errorpages/src/main/java/mozilla/components/browser/errorpages/ErrorPages.kt
mcarare: IMO in this case <`ul>` was used just for the bullet formatting it provides. The change impacts only the way screen readers handle the content labeling of the items. 
In other cases, like using a `<ol>` the `<li>` semantics is important to be kept ( like announcing a specific ordered). That is why I think we should not touch other use cases unless a specific request will be made.

components/browser/errorpages/src/main/res/raw/error_pages.html
Amejia481: 👍

components/browser/errorpages/src/main/java/mozilla/components/browser/errorpages/ErrorPages.kt
Amejia481: Sorry, I meant that the consumers  could use multiple variations of `<ul>` like `<ul class="name">` or `<ul id="id">` and `replace("<ul>", "<ul role=\"presentation\">")` won't be able to add the `role=\"presentation\"`  property as it won't  match. I just wanted to have a sometime type of documentation to indicate that issue to consumer or given them.

Amejia481: I'm completely agree with you it will fix the actual issue. Feel free to land!

PR #5311, opened 2019-12-13
Issue #4500: Add engine stubs for enabling/disabling extensions

csadilek:


components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
```diff
@@ -292,10 +292,36 @@ class GeckoEngine(
...
+     * See [Engine.enableWebExtension].
+     */
+    override fun enableWebExtension(
+        extension: WebExtension,
+        onSuccess: (WebExtension) -> Unit,
+        onError: (Throwable) -> Unit
+    ) {
+        // TODO https://bugzilla.mozilla.org/show_bug.cgi?id=1599585
+        webExtensionDelegate?.onEnabled(extension)
+        onSuccess(extension)
```
psymoon: I guess it is TODO but I find it odd that we pass `onSuccess` even though `webExtensionDelegate` is nullable.

csadilek: Yeaj, so, I think the `delegate` will always remain nullable, well eventually we will get rid of it, once we have the store available in the engine. The only reason I am calling `onSuccess` now is to be able test integration.

csadilek:
bors r=psymoon,Amejia481

csadilek:
bors r=psymoon,Amejia481

PR #5327, opened 2019-12-16
Issue #4500: Wire up AddonManager -> Engine

csadilek:
Next 🧩. Wiring Up AddonManager and Engine.

Next step after that: Wiring up UI and AddonManager.


csadilek:
bors r=Amejia481,psymoon

PR #5334, opened 2019-12-17
Closes #5329: FennecMigrator: Use startForegroundService().

### LINKED ISSUE #5329: Fennec Nightly Upgrade: IllegalStateException : Not allowed to start service Intent
Crash when upgrading Fennec Nightly with a Fenix build.

```
E  FATAL EXCEPTION: main
E  Process: org.mozilla.fennec_aurora, PID: 29582
E  java.lang.RuntimeException: Unable to create application org.mozilla.fenix.MigratingFenixApplication: java.lang.IllegalStateException
: Not allowed to start service Intent { cmp=org.mozilla.fennec_aurora/org.mozilla.fenix.MigrationService }: app is in background uid
UidRecord{cbc9e67 u0a91 TRNB idle procs:1 seq(0,0,0)}
E      at android.app.ActivityThread.handleBindApplication(ActivityThread.java:5876)
E      at android.app.ActivityThread.access$1100(ActivityThread.java:199)
E      at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1650)
E      at android.os.Handler.dispatchMessage(Handler.java:106)
E      at android.os.Looper.loop(Looper.java:193)
E      at android.app.ActivityThread.main(ActivityThread.java:6669)
E      at java.lang.reflect.Method.invoke(Native Method)
E      at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
E      at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
E  Caused by: java.lang.IllegalStateException: Not allowed to start service Intent { cmp=org.mozilla.fennec_aurora/org.mozilla.fenix.Mig
rationService }: app is in background uid UidRecord{cbc9e67 u0a91 TRNB idle procs:1 seq(0,0,0)}
E      at android.app.ContextImpl.startServiceCommon(ContextImpl.java:1577)
E      at android.app.ContextImpl.startService(ContextImpl.java:1532)
E      at android.content.ContextWrapper.startService(ContextWrapper.java:664)
E      at org.mozilla.fenix.MigratingFenixApplication.setupInMainProcessOnly(MigratingFenixApplication.kt:17)
E      at org.mozilla.fenix.FenixApplication.onCreate(FenixApplication.kt:4)
E      at android.app.Instrumentation.callApplicationOnCreate(Instrumentation.java:1154)
E      at android.app.ActivityThread.handleBindApplication(ActivityThread.java:5871)
E      ... 8 more
```

pocmo:
Whenever the app is in the background while we start a migration - for example when using a widget or some other background action triggers an Application object being created - then we try to launch the migration service and we see a crash since that is not allowed. Since we are using a foreground service anyways we can just use `startForegroundService()` here.

### DIFF
--- a/components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt
+++ b/components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt
@@ -7,6 +7,7 @@ package mozilla.components.support.migration
 import android.content.Context
 import android.content.Intent
 import androidx.annotation.VisibleForTesting
+import androidx.core.content.ContextCompat
 import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Deferred
@@ -375,7 +376,7 @@ class FennecMigrator private constructor(
     fun <T> startMigrationServiceIfNeeded(service: Class<T>) where T : AbstractMigrationService {
         if (hasMigrationsToRun()) {
             logger.debug("Has migrations to run. Starting service.")
-            context.startService(Intent(context, service))
+            ContextCompat.startForegroundService(context, Intent(context, service))
         } else {
             logger.debug("No migrations to run. Not starting service.")
         }
--- a/components/support/migration/src/main/AndroidManifest.xml
+++ b/components/support/migration/src/main/AndroidManifest.xml
@@ -2,4 +2,8 @@
    - License, v. 2.0. If a copy of the MPL was not distributed with this
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="mozilla.components.support.migration" />
+    package="mozilla.components.support.migration">
+
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
+
+</manifest>


components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt
```diff
@@ -375,7 +375,7 @@ class FennecMigrator private constructor(
     fun <T> startMigrationServiceIfNeeded(service: Class<T>) where T : AbstractMigrationService {
         if (hasMigrationsToRun()) {
             logger.debug("Has migrations to run. Starting service.")
-            context.startService(Intent(context, service))
+            context.startForegroundService(Intent(context, service))
```
psymoon: `ContextCompat.startForegroundService()`? 🤔 

I think we also need a permission for Android P:
`<uses-permission android:name=”android.permission.FOREGROUND_SERVICE”/>`

pocmo: Oh, yeah, that is right. We already have that for other reasons, but we should make that explicit in this component too.

pocmo: Updated!

pocmo:
```
/builds/worker/checkouts/src/components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt:378: Error: Call requires API level 26 (current min is 21): android.content.Context#startForegroundService [NewApi]
[task 2019-12-17T17:08:18.164Z]               context.startForegroundService(Intent(context, service))
```

Oh damn. Forgot about that. Impressive how many errors you can have with a one line change! ✨

pocmo:
Updated with `ContextCompat.startForegroundService()`.

PR #5350, opened 2019-12-18
Regenerate metrics.md. (Sorry, not sorry)

pocmo:
The Glean plugin changed how `metrics.md` files are generated and so we should push the new version to avoid it updating all the time locally.


pocmo:
bors r=psymoon

PR #5364, opened 2019-12-19
Add Cliqz search provider.

sammacbeth:
Adds `cliqz.xml` with Cliqz search provider (https://beta.cliqz.com)

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/search/src/main/assets/searchplugins/cliqz.xml
+++ b/components/browser/search/src/main/assets/searchplugins/cliqz.xml
@@ -1,0 +1,18 @@
+<SearchPlugin xmlns="http://www.mozilla.org/2006/browser/search/">
+  <ShortName>Cliqz</ShortName>
+  <InputEncoding>UTF-8</InputEncoding>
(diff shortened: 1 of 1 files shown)


sammacbeth:
Test failure seems to be just to code coverage. Does that apply to XML?

pocmo:
Sorry, it took way too long to get back to this. Almost all time has gone into the Fenix/Fennec migration project and I neglected some PRs because of that ... and this has tumbled to the bottom of the list. :(

> Test failure seems to be just to code coverage. Does that apply to XML?

No, this seems to be a false report from codecov - which unfortunately happens quite often.


Something I haven't thought about here is "packaging". Currently `list.json` controls what search engines a user will get. However right now *ALL* search plugins will end up in apps using `browser-search` and that is something we should solve before adding more plugins, I think.

Looking at the issue it seems like including the plugin is also a bit more complicated from the policy point of view. Let's get that sorted out. In the meantime I would like to close the PR to reduce the list of open PRs (since this one is not actionable) at the moment.

Note that in an app using `browser-search` you can always package additional plugins. We are doing this in several of our own apps. So not landing this at least does not block anyone from using this or other search engines in their apps. :)

Issue #4928, opened 2019-11-01
Intermittent test failure: :feature-sitepermissions:jacocoDebug


grigoryk:
An example of this intermittent making for a very sad PR: https://github.com/mozilla-mobile/android-components/pull/4912

A TC failure:
https://taskcluster-artifacts.net/UmPE2Aw9TKuH6FVg631Ojw/0/public/logs/live_backing.log
```
[task 2019-11-01T16:34:53.702Z] > Task :feature-sitepermissions:jacocoDebug FAILED
[task 2019-11-01T16:34:53.770Z] 
[task 2019-11-01T16:34:53.770Z] FAILURE: Build failed with an exception.
[task 2019-11-01T16:34:53.770Z] 
[task 2019-11-01T16:34:53.770Z] * What went wrong:
[task 2019-11-01T16:34:53.770Z] Execution failed for task ':feature-sitepermissions:jacocoDebug'.
[task 2019-11-01T16:34:53.770Z] > There was a failure while executing work items
[task 2019-11-01T16:34:53.770Z]    > A failure occurred while executing com.android.build.gradle.internal.tasks.JacocoTaskDelegate$JacocoWorkerAction
[task 2019-11-01T16:34:53.770Z]       > java.lang.ExceptionInInitializerError (no error message)
```

jonalmeida:
Seeing more of this now. TC: https://firefox-ci-tc.services.mozilla.com/tasks/RvLbeTl-QxqFX9ZG3bpmlA/runs/0/logs/https%3A%2F%2Ffirefox-ci-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FRvLbeTl-QxqFX9ZG3bpmlA%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log

```
[task 2019-12-19T19:47:06.438Z] * What went wrong:
[task 2019-12-19T19:47:06.438Z] Execution failed for task ':feature-session:jacocoDebug'.
[task 2019-12-19T19:47:06.438Z] > A failure occurred while executing com.android.build.gradle.internal.tasks.Workers$ActionFacade
[task 2019-12-19T19:47:06.438Z]    > java.lang.ExceptionInInitializerError (no error message)
```

jonalmeida:
I'm going to close this since I don't think we see this failure any more. Please re-open if I'm wrong.

PR #5381, opened 2019-12-20
Issue #4928: Gradle: Force jacoco out of process..

pocmo:
Trying to find a solution to our intermittent jacoco problems (#4928) I came across this flag that with AGP 3.5+ should move jacoco to a separate JVM. There's no guarantee that this actually helps but let's try how it behaves on taskcluster..


jonalmeida:
Not merging this in case we want to run this a couple more times on CI to test it out.

jonalmeida:
I'm going to merge this since many PRs are seeing this jacoco exception and it's worth trying it out (and reverting if not).

jonalmeida:
bors r=psymoon

jonalmeida:
bors ping

Issue #5376, opened 2019-12-20
Simplify implementation of URLStringUtils.isURLLike


hawkinsw:
Hello!

First, let me say thank you for taking the time and effort to build such a comprehensive function for detecting whether a string is a valid URL. The Java SDK versions are so restrictive that they are unusable. What you provide is really, really important to have.

In Fenix, the `URLStringUtils.isURLLike` function is used to determine whether or not what the user's input to the Awesome Bar is a valid URL. That classification determines whether to load a page with that input as a URL or start a query with that input as the search terms. In other words, the application blocks while waiting for that decision to be made.

Unfortunately, due to the robustness of the regular expression underlying the `URLStringUtils.isURLLike` function, the first classification takes a very, very long time. According to our profiles, it takes almost 500ms (on a modern smartphone) to compile the regular expression. For context, that is 2-3x longer than it takes to instantiate and initialize Gecko on the same hardware. Of course, each comparison using the compiled regular expression is blindingly fast.

It would be ideal to maintain the functionality provided by your robust classification but the Fenix UI  cannot be stalled for that long. It is very perceptible to the user.

We experimented with precompiling the regular expression on a background thread prior to the first call to `URLStringUtils.isURLLike`. This worked and users were happy! However, there was concern among developers that using such a trick made the code less maintainable. 

So it is that I come to you asking for your help to improve the performance of `URLStringUtils.isURLLike`. 

I certainly don't mean to assume that I can do your job, but I thought through a few options for doing this:

1. Looking through the history, I noticed that earlier this summer AC migrated away from a more-straightforward regular expression for classifying input as valid URLs to one that matches Fennec's implementation. AC could reincorporate that regular expression and expose its capabilities in a new function named, say, `URLStringUtils.isURLSimple`. AC users could choose which one to use based on their needs -- those wanting an exhausting classification could you use `URLStringUtils.isURLLike` (aware of the performance penalty for the first classification) and and those wanting a fast classification could use `URLStringUtils.isURLSimple` (aware of the possibility of mismatching certain inputs)

2. AC could abandon regular expression-based classification and use an algorithmic approach like the one taken in https://github.com/jsdom/whatwg-url. 

[...]

PR #5384, opened 2019-12-20
Issue #5376: Simpler method for determining whether a string is a search query

hawkinsw:
Update URLStringUtils.isSearchTerm to use a simpler method for
determining whether a string is a search query or a URL.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/support/utils/src/main/java/mozilla/components/support/utils/URLStringUtils.kt
+++ b/components/support/utils/src/main/java/mozilla/components/support/utils/URLStringUtils.kt
@@ -6,16 +6,39 @@ package mozilla.components.support.utils
 
 import android.net.Uri
 import android.text.TextUtils
+import androidx.annotation.VisibleForTesting
+import java.util.regex.Pattern
 
 object URLStringUtils {
-    fun isURLLike(string: String, safe: Boolean = false) =
+    /**
+     * Determine whether a string is a URL.
+     *
+     * This method performs a strict check to determine whether a string is a URL. It takes longer
+     * to execute than isURLLike() but checks whether, e.g., the TLD is ICANN-recognized. Consider
+     * using isURLLike() unless these guarantees are required.
+     */
+    fun isURLLikeStrict(string: String, safe: Boolean = false) =
         if (safe) {
             string.matches(WebURLFinder.fuzzyUrlRegex)
         } else {
             string.matches(WebURLFinder.fuzzyUrlNonWebRegex)
         }
 
-    fun isSearchTerm(string: String) = !isURLLike(string, false)
+    /**
+     * Determine whether a string is a URL.
+     *
+     * This method performs a lenient check to determine whether a string is a URL. Anything that
+     * contains a :, ://, or . and has no internal spaces is potentially a URL. If you need a
+     * stricter check, consider using isURLLikeStrict().
+     */
+    fun isURLLike(string: String) = isURLLenient.matcher(string).matches()
+
+    /**
+     * Determine whether a string is a search term.
+     *
+     * This method recognizes a string as a search term as anything other than a URL.
+     */
+    fun isSearchTerm(string: String) = !isURLLike(string)
 
     /**
      * Normalizes a URL String.
@@ -28,4 +51,39 @@ object URLStringUtils {
         }
         return uri.toString()
     }
+
+    private val isURLLenient by lazy {
+        // Be lenient about what is classified as potentially a URL. Anything that contains a :,
+        // ://, or . and has no internal spaces is potentially a URL.
+        //
+        // Use java.util.regex because it is always unicode aware on Android.
+        // https://developer.android.com/reference/java/util/regex/Pattern.html
+
+        // Use both the \w+ and \S* after the punctuation because they seem to match slightly
+        // different things. The \S matches any non-whitespace character (e.g., '~') and \w
+        // matches only word characters. In other words, the regex is requiring that there be a
+        // non-symbol character somewhere after the ., : or :// and before any other character
+        // or the end of the string. For example, match
+        // mozilla.com/~userdir
+        // and not
+        // mozilla./~ or mozilla:/
+        // Without the [/]* after the :// in the alternation of the characters required to be a
+        // valid URL,
+        // file:///home/user/myfile.html
+        // is considered a search term; it is clearly a URL.
+        Pattern.compile("^\\s*\\w+(://[/]*|:|\\.)\\w+\\S*\\s*$", flags)
+    }
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal const val UNICODE_CHARACTER_CLASS: Int = 0x100
+
+    // To run tests on a non-Android device (like a computer), Pattern.compile
+    // requires a flag to enable unicode support. Set a value like flags here with a local
+    // copy of UNICODE_CHARACTER_CLASS. Use a local copy because that constant is not
+    // available on Android platforms < 24 (Fenix targets 21). At runtime this is not an issue
+    // because, again, Android REs are always unicode compliant.
+    // NB: The value has to go through an intermediate variable; otherwise, the linter will
+    // complain that this value is not one of the predefined enums that are allowed.
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal var flags = 0
 }
--- a/components/support/utils/src/test/java/mozilla/components/support/utils/URLStringUtilsTest.kt
+++ b/components/support/utils/src/test/java/mozilla/components/support/utils/URLStringUtilsTest.kt
@@ -7,16 +7,23 @@ package mozilla.components.support.utils
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import mozilla.components.support.utils.URLStringUtils.isSearchTerm
 import mozilla.components.support.utils.URLStringUtils.isURLLike
+import mozilla.components.support.utils.URLStringUtils.isURLLikeStrict
 import mozilla.components.support.utils.URLStringUtils.toNormalizedURL
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertFalse
 import org.junit.Assert.assertNotEquals
 import org.junit.Assert.assertTrue
+import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
 
 @RunWith(AndroidJUnit4::class)
 class URLStringUtilsTest {
+
+    @Before
+    fun configurePatternFlags() {
+        URLStringUtils.flags = URLStringUtils.UNICODE_CHARACTER_CLASS
+    }
 
     @Test
     fun toNormalizedURL() {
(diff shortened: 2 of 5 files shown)


components/support/utils/src/main/java/mozilla/components/support/utils/URLStringUtils.kt
```diff
@@ -28,4 +28,10 @@ object URLStringUtils {
         }
         return uri.toString()
     }
+
+    private val isURLLenient by lazy {
```
mcomella: I'll leave this up to the lib owners but I think we should remove `isURLLike` or at least have it share the `isSearchTerm` impl because:
- `isURLLike` seems redundant to `isSearchTerm`
- What is `isURLLike` supposed to mean? Without further details, how are consumers supposed to use it confidently? Note that it doesn't look like our code uses it ([query](https://searchfox.org/mozilla-mobile/search?path=&q=isurllike))

components/support/utils/src/main/java/mozilla/components/support/utils/URLStringUtils.kt
```diff
@@ -28,4 +28,10 @@ object URLStringUtils {
...
+        // Be lenient about what is classified as potentially a URL. Anything that contains a :
+        // or . and has no internal spaces is potentially a URL.
+        "^\\s*\\w+(://|:|\\.)\\w+\\S*\\s*$".toRegex()
```
mcomella: I don't think `\w` matches unicode characters so http://faß.de would not be considered valid. It [sounds like](https://www.regular-expressions.info/wordboundaries.html) Java might support unicode for `\b`.

Can we make sure there are unicode tests in the test file (if there aren't already)?

mcomella: Why is `\\w+\\S*` necessary? It seems redundant.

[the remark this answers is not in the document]
hawkinsw: This was the original AC regular expression.

hawkinsw: Good point!

hawkinsw: We need both the \w+ and \S* because they seem to match slightly different things. The \S matches any non-whitespace character (e.g., '~') and \w matches only word characters. In other words, the regex is requiring that there be a non-symbol character somewhere after the ., : or :// and before any other character or the end of the string. 

For example, we want to match something like mozilla.com/~userdir but not something like mozilla./~ or mozilla:/

mcomella: I see: I think we should we include a comment to explain that.

From my perspective, I'm not sure being so strict is necessary but I haven't done the research. :)

components/support/utils/src/test/java/mozilla/components/support/utils/URLStringUtilsTest.kt
```diff
@@ -67,9 +67,20 @@ class URLStringUtilsTest {
...
+        assertFalse(isSearchTerm("about:config:8000"))
         assertFalse(isSearchTerm("http://192.168.255.255"))
+        // Per https://bugs.chromium.org/p/chromium/issues/detail?id=31405, ICANN will accept
+        // purely numeric gTLDs.
+        assertFalse(isSearchTerm("3.14.2019"))
+        assertFalse(isSearchTerm(" cnn.com "))
+        assertFalse(isSearchTerm(" cnn.com"))
+        assertFalse(isSearchTerm("cnn.com "))
+        assertFalse(isSearchTerm("mozilla.com/~userdir"))
+        assertFalse(isSearchTerm("http://faß.de//"))
```
mcomella: Could we include a URL with a unicode character in the TLD?

components/support/utils/src/main/java/mozilla/components/support/utils/URLStringUtils.kt
```diff
@@ -28,4 +29,12 @@ object URLStringUtils {
         }
         return uri.toString()
     }
+
+    private val isURLLenient by lazy {
+        // Be lenient about what is classified as potentially a URL. Anything that contains a :,
+        // ://, or . and has no internal spaces is potentially a URL. Use java.util.regex because
```
mcomella: "Use java.util.regex over Kotlin Regex" (iiuc)

To be fair, Kotlin Regex probably uses java.util.Regex under the hood and our unit tests would catch if that changes.

components/support/utils/src/main/java/mozilla/components/support/utils/URLStringUtils.kt
```diff
@@ -28,4 +29,12 @@ object URLStringUtils {
...
+        // of its explicit unicode support
+        // (https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html).
+        Pattern.compile("^\\s*\\w+(://|:|\\.)\\w+\\S*\\s*$")
```
mcomella: Why did we decide to do our own expression over using `java.net.URL` or similar classes? Perhaps we should explain in code comments. It's also possible this is a "back out the regressing change" which is cool by me too. :P

I know `java.net.URL` throws an Exception when a result is not a URL so, due to the performance penalty, it's not appropriate for rapid (negative) matching but I don't know if there are other ways to use it or other possibilities.

components/support/utils/src/main/java/mozilla/components/support/utils/URLStringUtils.kt
```diff
@@ -15,7 +16,7 @@ object URLStringUtils {
             string.matches(WebURLFinder.fuzzyUrlNonWebRegex)
         }
 
-    fun isSearchTerm(string: String) = !isURLLike(string, false)
+    fun isSearchTerm(string: String) = !isURLLenient.matcher(string).matches()
```
mcomella: This probably needs a changelog entry.

components/support/utils/src/main/java/mozilla/components/support/utils/URLStringUtils.kt
hawkinsw: `java.net.URL` does not recognized as URLs strings that do not begin with "http://" or "https://".

components/support/utils/src/test/java/mozilla/components/support/utils/URLStringUtilsTest.kt
hawkinsw: good idea!

components/support/utils/src/main/java/mozilla/components/support/utils/URLStringUtils.kt
```diff
@@ -28,4 +30,20 @@ object URLStringUtils {
...
         return uri.toString()
     }
+
+    private val isURLLenient by lazy {
+        // Be lenient about what is classified as potentially a URL. Anything that contains a :,
+        // ://, or . and has no internal spaces is potentially a URL.
+        //
+        // Use java.util.regex because
+        // of its explicit unicode support when using the UNICODE_CHARACTER_CLASS flag
+        // (https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html).
```
mcomella: nit: should we link to Android docs instead?

components/support/utils/src/main/java/mozilla/components/support/utils/URLStringUtils.kt
```diff
@@ -28,4 +30,20 @@ object URLStringUtils {
...
+        //
+        // We need both the \w+ and \S* after the punctuation because they seem to match slightly
+        // different things. The \S matches any non-whitespace character (e.g., '~') and \w
+        // matches only word characters. In other words, the regex is requiring that there be a
+        // non-symbol character somewhere after the ., : or :// and before any other character
+        // or the end of the string.
```
mcomella: nit: include an example? I think they were helpful

components/support/utils/src/test/java/mozilla/components/support/utils/URLStringUtilsTest.kt
```diff
@@ -67,9 +67,21 @@ class URLStringUtilsTest {
...
+        assertTrue(isSearchTerm("sf: help"))
+        assertTrue(isSearchTerm("mozilla./~"))
+        assertTrue(isSearchTerm("cnn.com politics"))
 
         assertFalse(isSearchTerm("about:config"))
+        assertFalse(isSearchTerm("about:config:8000"))
         assertFalse(isSearchTerm("http://192.168.255.255"))
+        // Per https://bugs.chromium.org/p/chromium/issues/detail?id=31405, ICANN will accept
+        // purely numeric gTLDs.
+        assertFalse(isSearchTerm("3.14.2019"))
```
mcomella: Should we include `file://` URLs too (even if they're disabled in fenix)?

components/support/utils/src/test/java/mozilla/components/support/utils/URLStringUtilsTest.kt
```diff
@@ -67,9 +67,21 @@ class URLStringUtilsTest {
...
+        assertFalse(isSearchTerm(" cnn.com "))
+        assertFalse(isSearchTerm(" cnn.com"))
+        assertFalse(isSearchTerm("cnn.com "))
+        assertFalse(isSearchTerm("mozilla.com/~userdir"))
+        assertFalse(isSearchTerm("http://faß.de//"))
```
mcomella: I noticed our algorithm looks for one of `://`, `:`, and `.`: that means `http://faß.de//` is a significantly different test case than, `faß.de`. Should we include both?

Maybe this is why the tests didn't catch the non-unicode error before (if I understood you correctly).

hawkinsw: Yes. I think.

I believe that what happened was the unicode B matched \s* in that case because it fell after the ://. When it was before the . (the first required punctuation element) (as in the latter example), it would not match in the `\w` until I passed the proper option to the `compile` method. I see no reason to change our approach, but I suppose I haven't thought about it too deeply and reserve my ability to change my mind :-)

mcomella: I do think it a bit odd that the protocol and TLD are handled in the same branch but, given the guided principle of,

> Be lenient about what is classified as potentially a URL. Anything that contains ://, or . and has no internal spaces is potentially a URL.

Maybe it's good enough? In any case, I think landing this as is – as a glorified back-out to improve perf quickly – and handing it off to the a-c if they feel the implementation needs to be more robust is a fine approach for us.

hawkinsw: This is definitely a very lenient approach to deciding what is a URL. This is exactly the method that Fennec uses, so I think it is reasonable.

components/support/utils/src/main/java/mozilla/components/support/utils/URLStringUtils.kt
hawkinsw: I think that we should link to the Java documentation since we are specifically referring to the capability of the SDK. But, I see your point: keeping up to date with the versions of the SDK may be hard. I honestly don't care which one we do. Do you feel strongly one way or another?

mcomella: The Android framework contains a copy of the SDK and, for all we know, may not be representative of the SDK (for example, the behavior around unicode). Therefore, I strongly prefer to reference the Android documentation.

components/support/utils/src/main/java/mozilla/components/support/utils/URLStringUtils.kt
```diff
@@ -6,16 +6,17 @@ package mozilla.components.support.utils
...
-            string.matches(WebURLFinder.fuzzyUrlNonWebRegex)
-        }
+            if (safe) {
+                string.matches(WebURLFinder.fuzzyUrlRegex)
+            } else {
+                string.matches(WebURLFinder.fuzzyUrlNonWebRegex)
+            }
 
-    fun isSearchTerm(string: String) = !isURLLike(string, false)
+    fun isSearchTerm(string: String) = !isURLLenient.matcher(string).matches()
```
csadilek: @hawkinsw Thanks for taking this on! 

I am curious what the plan is in Fenix. Is it to switch these `isUrl` calls to `!isSearchTerm` (since that's the faster method now but currently has no caller): https://github.com/mozilla-mobile/fenix/search?utf8=✓&q=isUrl()

We then should also update the `ToolbarInteractor` in A-C to get the benefits: https://github.com/mozilla-mobile/android-components/blob/master/components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarInteractor.kt#L29

We could also have a `String.isUrlStrict` and `String.isUrl` and only use the strict method where needed.

hawkinsw: > @hawkinsw Thanks for taking this on!
> 
> I am curious what the plan is in Fenix. Is it to switch these `isUrl` calls to `!isSearchTerm` (since that's the faster method now but currently has no caller)? [https://github.com/mozilla-mobile/fenix/search?utf8=✓&q=isUrl()](https://github.com/mozilla-mobile/fenix/search?utf8=%E2%9C%93&q=isUrl())

That is exactly the path that I thought we would take. However, see below ...
> 
> We then should also update the `ToolbarInteractor` in A-C to get the benefits: https://github.com/mozilla-mobile/android-components/blob/master/components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarInteractor.kt#L29
> 

Thanks for pointing this out. I will push an updated PR to address this! Thanks again!
> We could also have a `String.isUrlStrict` and `String.isUrl` and only use the strict method where needed.

I actually like this better. I am sending up an update to this PR that addresses this and the previous point. Thanks for the review. I will rerequest a review as soon as it is up!

hawkinsw:
Do not review this PR until further notice. There is a discrepancy in compiling regular expressions that work on the host (the developer's computer, that is) and that work on the device.

mcomella:
> Do not review this PR until further notice. There is a discrepancy in compiling regular expressions that work on the host (the developer's computer, that is) and that work on the device.

Could we compile regex that work on device and, for now, not worry about those on the host? We could run our tests on device and not run the tests as unit tests. This assumes a-c has tests on device (which they should! 😄)

mcomella:
Removing review request: I believe a-c will take it from here.

PR #5441, opened 2019-12-30
Issue #4500: Integrate GV API to list installed web extensions

csadilek:
Just wiring up the "real" method call now that we have it. With this the installed addon keeps working after a restart now.


csadilek:
bors r=psymoon

PR #5459, opened 2020-01-02
Closes #5148: Add notify and cancel extension functions to NotificationManagerCompat

### LINKED ISSUE #5148: Get rid of NotificationIds
I implemented `NotificationIds` and the extension method helpers, so that we could get a unique id for a `tag`. It turns out `NotificationManager` already has methods to additionally accept a tag: https://developer.android.com/reference/android/app/NotificationManager#notify(java.lang.String,%20int,%20android.app.Notification)

This means we could get rid of NotificationIds and the extension methods and use that directly.

rocketsroger:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/support/base/src/main/java/mozilla/components/support/base/ids/SharedIdsHelper.kt
+++ b/components/support/base/src/main/java/mozilla/components/support/base/ids/SharedIdsHelper.kt
@@ -4,8 +4,11 @@
 
 package mozilla.components.support.base.ids
 
+import android.app.Notification
+import android.app.NotificationManager
 import android.content.Context
 import androidx.annotation.VisibleForTesting
+import androidx.core.app.NotificationManagerCompat
 
 // If the tag is not used again in one week then clear the id
 private const val ID_LIFETIME: Long = 1000 * 60 * 60 * 24 * 7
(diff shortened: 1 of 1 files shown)


components/support/base/src/main/java/mozilla/components/support/base/ids/SharedIdsHelper.kt
```diff
@@ -39,3 +42,39 @@ object SharedIdsHelper {
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
     internal fun clear(context: Context) = ids.clear(context)
 }
+
+/**
+ * Post a notification to be shown in the status bar.
+ *
+ * Uses a unique [String] tag instead of an [Int] id like [NotificationManager.notify].
+ */
+fun NotificationManager.notify(context: Context, tag: String, notification: Notification) {
```
jonalmeida: Could we also mention in all the docs that this will use the existing ID for the tag if it exists?

My minor worry is that someone will use this on two different notifications with the same ID, and without realizing that it'll overwrite the previous notification.

Maybe also reference the SharedIdsHelper? e.g. "See also: [SharedIdsHelper]."

components/support/base/src/main/java/mozilla/components/support/base/ids/SharedIdsHelper.kt
```diff
@@ -4,8 +4,11 @@
 
 package mozilla.components.support.base.ids
```
jonalmeida: Should these go under `support-ktx` instead? It already has `support-base` as a dependency so I don't think we'd be adding overhead there.

jonalmeida:
Can we write some simple tests to assert SharedIdsHelper is being called? Not sure if it's possible to do that when you touch Android APIs directly.

components/support/base/src/main/java/mozilla/components/support/base/ids/SharedIdsHelper.kt
rocketsroger: Sure, I'll move it.

components/support/base/src/main/java/mozilla/components/support/base/ids/SharedIdsHelper.kt
rocketsroger: OK, I'll document this.  Thanks,

components/support/base/src/main/java/mozilla/components/support/base/ids/SharedIdsHelper.kt
rocketsroger: There is one issue when I move it, it will still break Fenix.

jonalmeida: Oh, they're using these extensions from a-c already? 🤔

jonalmeida: That is to say, wouldn't they have to import the extension from this package to use it?

rocketsroger: Yes, in `PoweredByNotification.kt`.  Another choice is I don't add these extensions but remove the call in Fenix?

rocketsroger: Yes, they are importing `mozilla.components.support.base.ids.cancel` and `mozilla.components.support.base.ids.notify`

jonalmeida: Spoke offline: these extensions existed before, and we're reintroducing them since it was breaking Fenix. We can refactor them to the right place later then. 👍

rocketsroger:
bors r=jonalmeida

PR #5464, opened 2020-01-03
Do not publish glean gradle plugin to fix snapshot builds

csadilek:
This currently breaks our snapshot builds as the renameSnapshotArtifacts task is not defined for tooling/glean-gradle-plugin.

/cc @mdboom


csadilek:
Note, the plugin is not yet being used, but we will need to fix this so that https://github.com/mozilla-mobile/fenix/pull/6476 can land.

PR #5483, opened 2020-01-07
Migration telemetry + fennec telemetry identifiers migration

grigoryk:
This is a rolling WIP for migration telemetry; will mark it 'ready' when it's ready for review.

---

### Pull Request checklist

### After merge


grigoryk:

Operational details about Fennec->Fenix migration, such as success/error rates, raw counts, general types of errors.

2) Why does Mozilla need to answer these questions?  Are there benefits for users? Do we need this information to address product or business requirements? Some example responses:

This information is essential to ensure Fennec->Fenix migration proceeds smoothly for our users, avoiding unnecessary data loss and quality issues.

Crash and failed assertion reporting via Socorro and Sentry is in place, but that isn't detailed enough for us to have a good picture of quality of the migrations in the wild.

No.

5) List all proposed measurements and indicate the category of data collection for each measurement, using the [Firefox data collection categories](https://wiki.mozilla.org/Firefox/Data_Collection) found on the Mozilla wiki.

<table>
  <tr>
    <td>Tracking Bug #</td>
  </tr>
  <tr>
    <td>History migration</td>
    <td>Category 2, Interaction data</td>
    <td>https://github.com/mozilla-mobile/fenix/issues/5888</td>
  </tr>

  <tr>
    <td>Bookmarks migration</td>
    <td>Category 2, Interaction data</td>
    <td>https://github.com/mozilla-mobile/fenix/issues/5888</td>
  </tr>

  <tr>
    <td>Open tabs migration</td>
    <td>Category 2, Interaction data</td>
    <td>https://github.com/mozilla-mobile/fenix/issues/5888</td>
  </tr>

  <tr>
    <td>Fxa migration</td>
    <td>Category 2, Interaction data</td>
    <td>https://github.com/mozilla-mobile/fenix/issues/5888</td>
  </tr>

  <tr>
    <td>Logins migration</td>
    <td>Category 2, Interaction data</td>
    <td>https://github.com/mozilla-mobile/fenix/issues/5888</td>
  </tr>

  <tr>
    <td>Gecko migration</td>
    <td>Category 1, techical data</td>
    <td>https://github.com/mozilla-mobile/fenix/issues/5888</td>
  </tr>

  <tr>
    <td>Settings migration</td>
    <td>Category 2, Interaction data</td>
    <td>https://github.com/mozilla-mobile/fenix/issues/5888</td>
  </tr>
<tr>
    <td>Addons migration</td>
    <td>Category 2, Interaction data</td>
    <td>https://github.com/mozilla-mobile/fenix/issues/5888</td>
  </tr>
<tr>
    <td>Telemetry identifiers migration</td>
    <td>Category 1, technical data</td>
    <td>https://github.com/mozilla-mobile/fenix/issues/5888</td>
  </tr>
</table>


* I want this data to be collected for 6 months initially (potentially renewable).

* Which release channels?
all

* Which countries?
all

* Which locales?
all

no

No opt-out mechanism.

A telemetry dashboard for Fennec->Fenix migration will be built and closely monitored during the migration.
 
Internally within the core Fenix/A-C/A-S teams.

11) Is there a third-party tool (i.e. not Telemetry) that you are proposing to use for this data collection? If so:

n/a

grigoryk:
I haven't included the very detailed `logins` metrics as they're emitted from the rust library; we could do this in a follow-up. For now there are high level counts indicating total detected, # failed to parse, # failed to import. These basic metrics should be a good start for nightly.

components/support/migration/build.gradle
```diff
@@ -73,3 +78,6 @@ dependencies {
 
 apply from: '../../../publish.gradle'
 ext.configurePublish(config.componentsGroupId, archivesBaseName, project.ext.description)
+
+ext.gleanGenerateMarkdownDocs = true
+apply from: '../../service/glean/scripts/sdk_generator.gradle'
```
Dexterp37: Please note that the SDK generator is [deprecated since A-C 27](https://github.com/mozilla-mobile/android-components/blob/master/docs/changelog.md#2700). You need to use [this](https://github.com/mozilla-mobile/android-components/blob/cabd445aae1a995c7154c343301ce600788be85d/components/support/sync-telemetry/build.gradle#L11-L13) and [this](https://github.com/mozilla-mobile/android-components/blob/cabd445aae1a995c7154c343301ce600788be85d/components/support/sync-telemetry/build.gradle#L62-L63)

components/support/migration/metrics.yaml
```diff
@@ -0,0 +1,508 @@
...
+      - telemetry_identifiers
+    send_in_pings:
+      - migration
+    bugs:
+      - https://github.com/mozilla-mobile/fenix/issues/5888
+    data_reviews:
+      - https://github.com/mozilla-mobile/android-components/pull/5483
+    notification_emails:
+      - fenix-core@mozilla.com
+    expires: never
```
Dexterp37: Are you sure these should be `never` expiring? When will the migration process end?

components/support/migration/metrics.yaml
```diff
@@ -0,0 +1,508 @@
...
+      - fxa
+      - gecko
+      - settings
+      - telemetry_identifiers
+    send_in_pings:
+      - migration
+    bugs:
+      - https://github.com/mozilla-mobile/fenix/issues/5888
+    data_reviews:
+      - https://github.com/mozilla-mobile/android-components/pull/5483
```
Dexterp37: You should add a link to the data review response here. Do not forget to update this and all the other links in this file.

components/support/migration/metrics.yaml
```diff
@@ -0,0 +1,508 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+$schema: moz://mozilla.org/schemas/glean/metrics/1-0-0
+
+migration:
+  migration_versions:
```
Dexterp37: nit: `migration` is redundant in this case, since the category is already called `migration`. The final metric name in the ping would be `migration.migration_version`, which seems.. awkward.

components/support/migration/metrics.yaml
```diff
@@ -0,0 +1,508 @@
...
+    data_reviews:
+      - https://github.com/mozilla-mobile/android-components/pull/5483
+    notification_emails:
+      - fenix-core@mozilla.com
+    expires: never
+    lifetime: ping
+
+migration.telemetry_identifiers:
+  fennec_client_id:
+    type: string
```
Dexterp37: Isn't this a UUID?

components/support/migration/metrics.yaml
```diff
@@ -0,0 +1,508 @@
...
+      - https://github.com/mozilla-mobile/fenix/issues/5888
+    data_reviews:
+      - https://github.com/mozilla-mobile/android-components/pull/5483
+    notification_emails:
+      - fenix-core@mozilla.com
+    expires: never
+    lifetime: ping
+  fennec_profile_creation_date:
+    type: datetime
+    time_unit: millisecond
```
Dexterp37: Is this the resolution of the profile creation date in Fennec? I would be very surprised if that would be something other than "daily" precision!

components/support/migration/metrics.yaml
```diff
@@ -0,0 +1,508 @@
...
+    notification_emails:
+      - fenix-core@mozilla.com
+    expires: never
+    lifetime: ping
+
+migration.history:
+  migrated:
+    type: labeled_counter
+    description: >
+      Counters describing how history migration went; should add-up to 'detected'.
```
Dexterp37:
```suggestion
      Counters describing how history migration went; should add-up to 'migration.history.detected'.
```

components/support/migration/metrics.yaml
```diff
@@ -0,0 +1,508 @@
...
+    notification_emails:
+      - fenix-core@mozilla.com
+    expires: never
+    lifetime: ping
+
+migration.bookmarks:
+  migrated:
+    type: labeled_counter
+    description: >
+      Counters describing how bookmark migration went; should add-up to 'detected'.
```
Dexterp37:
```suggestion
      Counters describing how bookmark migration went; should add-up to 'migration.bookmarks.detected'.
```

components/support/migration/metrics.yaml
```diff
@@ -0,0 +1,508 @@
...
+      - https://github.com/mozilla-mobile/fenix/issues/5888
+    data_reviews:
+      - https://github.com/mozilla-mobile/android-components/pull/5483
+    notification_emails:
+      - fenix-core@mozilla.com
+    expires: never
+    lifetime: ping
+
+migration.open_tabs:
+  detected:
```
Dexterp37: Oh, so you don't get the number of failures for migrated tabs?

components/support/migration/metrics.yaml
```diff
@@ -0,0 +1,508 @@
...
+    bugs:
+      - https://github.com/mozilla-mobile/fenix/issues/5888
+    data_reviews:
+      - https://github.com/mozilla-mobile/android-components/pull/5483
+    notification_emails:
+      - fenix-core@mozilla.com
+    expires: never
+    lifetime: ping
+  failure_reason:
+    type: string
```
Dexterp37: Is there a pre-defined list of reasons why this could fail? Would it make sense to send failure codes instead?

components/support/migration/metrics.yaml
```diff
@@ -0,0 +1,508 @@
...
+      - migration
+    bugs:
+      - https://github.com/mozilla-mobile/fenix/issues/5888
+    data_reviews:
+      - https://github.com/mozilla-mobile/android-components/pull/5483
+    notification_emails:
+      - fenix-core@mozilla.com
+    expires: never
+    lifetime: ping
+  success_reason:
```
Dexterp37: Is there a pre-defined list of reasons why this could succeed? Would it make sense to send success codes instead?

components/support/migration/metrics.yaml
```diff
@@ -0,0 +1,508 @@
...
+    bugs:
+      - https://github.com/mozilla-mobile/fenix/issues/5888
+    data_reviews:
+      - https://github.com/mozilla-mobile/android-components/pull/5483
+    notification_emails:
+      - fenix-core@mozilla.com
+    expires: never
+    lifetime: ping
+  failure_reason:
+    type: string
```
Dexterp37: Same as the other considerations for strings.

components/support/migration/metrics.yaml
```diff
@@ -0,0 +1,508 @@
...
+      - migration
+    bugs:
+      - https://github.com/mozilla-mobile/fenix/issues/5888
+    data_reviews:
+      - https://github.com/mozilla-mobile/android-components/pull/5483
+    notification_emails:
+      - fenix-core@mozilla.com
+    expires: never
+    lifetime: ping
+  success_reason:
```
Dexterp37: Same as the other considerations for strings.

components/support/migration/metrics.yaml
```diff
@@ -0,0 +1,508 @@
...
+    bugs:
+      - https://github.com/mozilla-mobile/fenix/issues/5888
+    data_reviews:
+      - https://github.com/mozilla-mobile/android-components/pull/5483
+    notification_emails:
+      - fenix-core@mozilla.com
+    expires: never
+    lifetime: ping
+  failure_reason:
+    type: string
```
Dexterp37: Same as the other considerations for strings.

components/support/migration/pings.yaml
```diff
@@ -0,0 +1,16 @@
...
+migration:
+  description: >
+    A ping sent after a Fennec->Fenix migration was completed.
+  include_client_id: true
+  bugs:
+  - https://github.com/mozilla-mobile/fenix/issues/5888
+  notification_emails:
+  - fenix-core@mozilla.com
+  data_reviews:
+  - https://github.com/mozilla-mobile/android-components/pull/5483
```
Dexterp37: Please do not forget to update this to the response of the data review.

Dexterp37:
>     1. No opt-out mechanism.

Please note that the opt-out mechanism is provided by the Glean SDK itself.

components/support/migration/metrics.yaml
grigoryk: Ah, forgot to change the expiration labels; will change these to "in 6 months".

grigoryk:
> Please note that the opt-out mechanism is provided by the Glean SDK itself.
How does that work in the case of a migration ping from the user's point of view, though?

components/support/migration/metrics.yaml
grigoryk: I did this to make sure "versions of migrations" aren't understood as "migration of versions". Both are a little awkward in their own ways :)

components/support/migration/pings.yaml
grigoryk: Okay, will do.

We haven't followed this pattern in at least some other metrics that were added (e.g. a lot just link to the PR with a data review), and this isn't mentioned in the glean docs, either.

components/support/migration/metrics.yaml
grigoryk: It could be implied during analysis from two other metrics: `failed = detected - migrated`

components/support/migration/metrics.yaml
grigoryk: That's a good call, will save a lot of bandwidth/storage.

Dexterp37:
> 
> 
> > Please note that the opt-out mechanism is provided by the Glean SDK itself.
> 
> How does that work in the case of a migration ping from the user's point of view, though? E.g. there's no access to Settings while a migration is in process.

Any previously set Telemetry setting is honoured: if user has telemetry off before the migration starts, no ping will be sent.

components/support/migration/pings.yaml
Dexterp37: > We haven't followed this pattern in at least some other metrics that were added (e.g. a lot just link to the PR with a data review), and this isn't mentioned in the glean docs, either.

Unfortunately this is something we cannot enforce at build-time and we have to resort to documentation and education.

I created a PR to [clarify the language in our docs](https://github.com/mozilla/glean/pull/641).

components/support/migration/metrics.yaml
Dexterp37: I see, ok!

components/support/migration/build.gradle
```diff
@@ -73,3 +90,6 @@ dependencies {
 
 apply from: '../../../publish.gradle'
 ext.configurePublish(config.componentsGroupId, archivesBaseName, project.ext.description)
+
+ext.gleanGenerateMarkdownDocs = true
+apply plugin: "org.mozilla.telemetry.glean-gradle-plugin"
```
Dexterp37: nit: missing newline at the end of the file :)

components/support/migration/metrics.yaml
Dexterp37: Is the same true for the other metrics for which you're also explicitly counting the number of failures? Note that this is not a blocker for this PR, just a recommendation for analysis clarity: I'd strive for consistency within this file, if possible. If you're counting the failures for some classes of problems, then I'd count them for all classes of problems. It might be confusing to count failures explicitly for a subset of problems and then infer it at analysis time for another class of problems.

components/support/migration/metrics.yaml
Dexterp37: Moreover, it will save you some troubles :) String metric types have a limit at [100 characters](https://mozilla.github.io/glean/book/user/metrics/string.html#limits), and it's very unlikely that the limit will ever grow.

grigoryk:
> > > Please note that the opt-out mechanism is provided by the Glean SDK itself.
> > 
> > 
> > How does that work in the case of a migration ping from the user's point of view, though? E.g. there's no access to Settings while a migration is in process.
> 
> Any previously set Telemetry setting is honoured: if user has telemetry off before the migration starts, no ping will be sent.

In the telemetry migration, we're looking at the FHR setting in Fennec. It's worth noting that while it's probably true that (i will need to double check that there aren't some timing/initialization quirks at play) a user may disable this migration ping by setting FHR to 'off' in Fennec _before_ the upgrade, it's really not a great opt-out mechanism for most users.

Dexterp37:
> In the telemetry migration, we're looking at the FHR setting in Fennec. It's worth noting that while it's probably true that (i will need to double check that there aren't some timing/initialization quirks at play) a user may disable this migration ping by setting FHR to 'off' in Fennec _before_ the upgrade,

The Glean SDK *requires* somebody to tell the SDK if it's ok or not to send data. That "ok or not" part comes from the product itself (Fenix, in this case) which is responsible for propagating the user intent to the library. I suspect, but please correct me if I'm wrong, that the workflow is the following:

- Fennec is told that needs to updated;
- Fenix is downloaded;
- Fenix starts and tries to migrate things over;
- The "FHR" preference gets migrated over;
- Fenix inits telemetry and the Glean SDK is initialized with the new preference.

In that case, the FHR setting is honored.

Both legacy telemetry and the Glean SDK don't offer a mechanism to opt out from _single_ data points. They do offer a mechanism to shut down data upload completely, which is the opt-out mechanism used by Fenix+Glean SDK.

mdboom:
> a user may disable this migration ping by setting FHR to 'off' in Fennec _before_ the upgrade, it's really not a great opt-out mechanism for most users.

@grigoryk : Can you elaborate on what kind of mechanism you'd like to see?  It sounds like the opt-on preference in Fennec will be honoured, both for migration to Fenix and then within Fenix itself.

grigoryk:
> @grigoryk : Can you elaborate on what kind of mechanism you'd like to see? It sounds like the opt-on preference in Fennec will be honoured, both for migration to Fenix and then within Fenix itself.

To be clear, I think the current behaviour is perfectly fine in this case. We already made the decisions around how we migrate telemetry settings, and they apply here as well.

My only issue with this is pedantic - I don't want to claim we have a viable opt-out mechanism, while in reality it requires an detailed understanding of the migration process from the user which can't be expected. E.g. this situation somehow feels different from just saying that users may opt-out by going to Settings->Telemetry.

grigoryk:
> 
    Fenix starts and tries to migrate things over;
    The "FHR" preference gets migrated over;
    Fenix inits telemetry and the Glean SDK is initialized with the new preference.
    In that case, the FHR setting is honored.

We need to confirm timing here, actually... Now that you mention it, I'm not sure that we wait to initialize Glean SDK before FHR setting is migrated. I'll double check and comment on this issue.

boek:
@mdboom @Dexterp37 This conversation has me curious.

Does Glean collect metrics even if it hasn't been enabled yet? So if like @grigoryk says we wait until after the migration to actually decide to enable telemetry will the migration ping still collect and send?

boek:
R+

# Data Review Form (to be filled by Data Stewards)

1) Is there or will there be **documentation** that describes the schema for the ultimate data set in a public, complete, and accurate way?
Yes, metrics.yaml, pings.yaml and metrics.md

2) Is there a control mechanism that allows the user to turn the data collection on and off?
Yes, glean provides APIs to allow the user to turn off data collection.

3) If the request is for permanent data collection, is there someone who will monitor the data over time?
Has expiry

4) Using the **[category system of data types](https://wiki.mozilla.org/Firefox/Data_Collection)** on the Mozilla wiki, what collection type of data do the requested measurements fall under?
Type1, Type 2

5) Is the data collection request for default-on or default-off?
The default is handled by the client. It is default-on for Fenix with plans to adhere to previously applied Fennec setting as applicable

6) Does the instrumentation include the addition of **any *new* identifiers** (whether anonymous or otherwise; e.g., username, random IDs, etc.  See the appendix for more details)?
Yes, `fennec_client_id` is the existing client id of this Firefox for Android installation, and will be carried forward during migration for telemetry continuity

7) Is the data collection covered by the existing Firefox privacy notice?
Yes

8) Does there need to be a check-in in the future to determine whether to renew the data?
`fenix-core` Will check-in at expiry.

9) Does the data collection use a third-party collection tool?
No

grigoryk:
@Dexterp37 addressed all of your feedback. I've left failure counts as-is, since they reflect the way migration code is structured - e.g. when a migration produced an explicit failure count, it's reported (and can be cross-checked against detected/success counts); otherwise, it can be implied during analysis. We _can_ synthesize the implied values, but that doesn't seem to be valuable beyond the somewhat more consistent in shape.

I also wouldn't want to imply that the synthesized failure counts are also subject to cross-checks, since they're not in these specific cases. In case when these counts are explicitly reported, we can verify them to ensure that error reporting code is bug free.

Thanks for your time reviewing this stuff, much appreciated!

components/support/migration/metrics.yaml
grigoryk: Checked locally, and it is just a regular unix timestamp.

components/support/migration/metrics.yaml
```diff
@@ -0,0 +1,550 @@
...
+      - https://github.com/mozilla-mobile/android-components/pull/5483#issuecomment-573971458
+    notification_emails:
+      - fenix-core@mozilla.com
+    expires: 2020-10-01
+    lifetime: ping
+  duration:
+    type: timespan
+    time_unit: millisecond
+    description: >
+      Duration of the bookmarks migration.
```
rfk: nit: s/bookmarks/history/

grigoryk:
One caveat, is that since we currently run gecko migration separately, this PR will emit _two_ migration pings - one with results of the gecko migration, and one with everything else. I'm not sure if this is problematic or not. If it is, fixing it is - I think - trivial - we just need to move the `submit` call outside of the migrator, and into Fenix where it can be executed once for both migrations (and not once for each separately).

grigoryk:
@Dexterp37 also, I'm seeing the following in the logs, when settings various counts:
`01-13 20:11:30.712 10796 10915 W libglean_ffi: glean_core::error_recording: migration.logins.failure_counts/process: Added negative or zero value 0`

`migration.logins.failure_counts` is type `counter`, which is why it doesn't seem to like being incremented by nothing. However, `0` is a valid result here and something we'd like to track.

It seems like we'd need to treat omission of this count as a 0 value during analysis?

components/support/migration/metrics.yaml
```diff
@@ -0,0 +1,550 @@
...
+      - migration
+    bugs:
+      - https://github.com/mozilla-mobile/fenix/issues/5888
+    data_reviews:
+      - https://github.com/mozilla-mobile/android-components/pull/5483#issuecomment-573971458
+    notification_emails:
+      - fenix-core@mozilla.com
+    expires: 2020-10-01
+    lifetime: ping
+  success_reason:
```
rfk: Nit: I had many questions about why there could be multiple different reasons for success, but they were answered when I found the corresponding values in the code. I wonder if it would be worth fleshing out the descriptions here with a little more detail, to help future analysts avoid similar spelunking.

Dexterp37:
> E.g. this situation somehow feels different from just saying that users may opt-out by going to Settings->Telemetry.

For what is worth, I respectfully disagree: if user has expressed the intent of disabling Telemetry before migration, that intent is carried over (**must** be carried over). Happy to dig more into this topic, if you feel like, in a separate meeting.

Dexterp37:
> 
> 
> @mdboom @Dexterp37 This conversation has me curious.
> 
> Does Glean collect metrics even if it hasn't been enabled yet? So if like @grigoryk says we wait until after the migration to actually decide to enable telemetry will the migration ping still collect and send?

No, it doesn't collect. It batches up a certain amount of recordings before init ([see our docs](https://mozilla.github.io/glean/book/user/general-api.html#behavior-when-uninitialized)). When Glean inits, it attempts to replay back, honouring whatever value of `setUploadEnabled` set before init. If `false`, everything is dropped on the floor.

grigoryk:
bors r=pocmo

grigoryk:
bors r=pocmo

PR #5492, opened 2020-01-07
Closes #5174: Add telemetry crash service in lib-crash.

### LINKED ISSUE #5174: Crash Reporter: Allow collecting crash metrics without confirmation in the crash prompt
Our other products send crash statistics based on the telemetry setting. However in our case `GleanCrashReporterService` pretends to be a crash reporter and therefore only gets triggered if the user decides to send a crash report (and has telemetry enabled).

We should make it possible to trigger a service without showing the prompt first (for that particular service).

rocketsroger:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/lib/crash/src/main/java/mozilla/components/lib/crash/CrashReporter.kt
+++ b/components/lib/crash/src/main/java/mozilla/components/lib/crash/CrashReporter.kt
@@ -9,9 +9,9 @@ import android.content.Context
 import android.content.Intent
 import androidx.annotation.StyleRes
 import androidx.annotation.VisibleForTesting
+import androidx.core.content.ContextCompat
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
@@ -19,6 +19,8 @@ import mozilla.components.lib.crash.handler.ExceptionHandler
 import mozilla.components.lib.crash.notification.CrashNotification
 import mozilla.components.lib.crash.prompt.CrashPrompt
 import mozilla.components.lib.crash.service.CrashReporterService
+import mozilla.components.lib.crash.service.SendCrashReportService
+import mozilla.components.lib.crash.service.SendCrashTelemetryService
 import mozilla.components.support.base.log.logger.Logger
 
 /**
@@ -40,6 +42,7 @@ import mozilla.components.support.base.log.logger.Logger
  * @property enabled Enable/Disable crash reporting.
  *
  * @param services List of crash reporting services that should receive crash reports.
+ * @param telemetryServices List of telemetry crash reporting services that should receive crash reports.
  * @param shouldPrompt Whether or not the user should be prompted to confirm sending crash reports.
  * @param enabled Enable/Disable crash reporting.
  * @param promptConfiguration Configuration for customizing the crash reporter prompt.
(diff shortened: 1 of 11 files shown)


components/lib/crash/src/main/java/mozilla/components/lib/crash/CrashReporter.kt
```diff
@@ -96,13 +97,33 @@ class CrashReporter(
         }
     }
 
+    /**
+     * Submit a crash report to all registered services.
```
pocmo: nit: "to all registered services" -> "to all registered telemetry services"?

I wonder if that needs to be public at all. Are we expecting anyone to call this?

components/lib/crash/src/main/java/mozilla/components/lib/crash/CrashReporter.kt
```diff
@@ -96,13 +97,33 @@ class CrashReporter(
...
+                when (crash) {
+                    is Crash.NativeCodeCrash -> telemetryService.report(crash)
+                    is Crash.UncaughtExceptionCrash -> telemetryService.report(crash)
+                }
+            }
+
+            logger.info("Crash report submitted to ${services.size} telemetry services")
+            withContext(Dispatchers.Main) {
+                then()
+            }
```
pocmo: Is this callback mechanism with `then` even needed here?

components/lib/crash/src/main/java/mozilla/components/lib/crash/CrashReporter.kt
```diff
@@ -129,17 +150,22 @@ class CrashReporter(
 
         logger.info("Received crash: $crash")
 
-        if (CrashPrompt.shouldPromptForCrash(shouldPrompt, crash)) {
-            if (shouldSendIntent(crash)) {
-                // This crash was not fatal and the app has registered a pending intent: Send Intent to app and let the
-                // app handle showing a confirmation UI.
-                launchCrashIntent(context, crash)
+        /* Telemetry services does not require user prompts */
+        submitTelemetryCrashReport(crash)
```
pocmo: I am wondering here whether we may be executing this on the crashing process here (e.g. uncaught exception case) and whether that is problematic (after the crash it is not guaranteed that anything is working). But since we are also submitting here for the non-prompt case ... maybe that is fine? But then again we never run a non-prompt configuration in an app we ship right now.
If we step back from what we have right now... then I assume we should always launch our service here and the service will take care of uploading (non prompt case) and/or invoking telemetry from a new process? 🤔

rocketsroger: Agreed.  I'll simplify this so even non-prompt cases go through the service.  This way there is less that can go wrong.

components/lib/crash/src/main/java/mozilla/components/lib/crash/CrashReporter.kt
rocketsroger: I left it in so its available if needed.  But you're right that if we can always add it in later.    I'll remove it.

components/lib/crash/src/main/java/mozilla/components/lib/crash/CrashReporter.kt
rocketsroger: I don't think this needs to be public.  Are you thinking that someone would want to submit a caught exception to telemetry services?

rocketsroger: Actually I take that back.  If I'm moving this to the service then I'll have to make this public.

components/lib/crash/src/main/java/mozilla/components/lib/crash/CrashReporter.kt
```diff
@@ -80,7 +85,7 @@ class CrashReporter(
     /**
      * Submit a crash report to all registered services.
      */
-    fun submitReport(crash: Crash, then: () -> Unit = {}): Job {
+    fun submitCrashReport(crash: Crash, then: () -> Unit = {}): Job {
```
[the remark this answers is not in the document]
rocketsroger: Maybe at this stage I shouldn't make this API change.  This will require a fix in Fenix if I rename it.

components/lib/crash/src/main/java/mozilla/components/lib/crash/CrashReporter.kt
rocketsroger: FYI this is now needed since its now called from a service.

rocketsroger:
bors r=pocmo

rocketsroger:
bors r=pocmo

rocketsroger:
bors retry

PR #5438, opened 2019-12-30
Closes #5436: ReaderViewControlsBar should not unconditionally inflate

colintheshots:
---

### Pull Request checklist

- [x] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [x] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/android-components/blob/master/docs/changelog.md) or does not need one
- [x] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features

### After merge
- [ ] **Milestone**: Make sure issues closed by this pull request are added to the [milestone](https://github.com/mozilla-mobile/android-components/milestones) of the version currently in development.
- [ ] **Breaking Changes**: If this is a breaking change, please push a draft PR on [Reference Browser](https://github.com/mozilla-mobile/reference-browser) to address the breaking issues.

PR #5494, opened 2020-01-07
Closes #5436: Lazy inflate ReaderViewControlsBar

### LINKED ISSUE #5436: ReaderViewControlsBar should not unconditionally inflate
This is the A-C companion bug to fenix#4309.

The ReaderViewControlsBar unconditionally inflates, taking an extra 50ms on slower devices even on warm starts of Fenix from an external link. There is no reason this UI element needs to be visible immediately. It's infrequently used on any given session as it only appears when the reader view appearance button is pressed. Therefore, it makes sense to inflate this lazily only when needed. It certainly shouldn't delay our warm startup time from a link.

jonalmeida:
---

Building off https://github.com/mozilla-mobile/android-components/pull/5438, we're adding conditional inflation to the ReaderViewControlsBar with some ideas that Colin and I worked on offline.

I'm currently running into unrelated build issues with testing this in Fenix to assert the gains expected, but the basis of the change is largely the same so the code.

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/readerview/src/main/java/mozilla/components/feature/readerview/view/ReaderViewControlsBar.kt
+++ b/components/feature/readerview/src/main/java/mozilla/components/feature/readerview/view/ReaderViewControlsBar.kt
@@ -10,7 +10,6 @@ import android.util.AttributeSet
 import android.view.View
 import android.widget.RadioGroup
 import androidx.annotation.IdRes
-import androidx.annotation.VisibleForTesting
 import androidx.appcompat.widget.AppCompatButton
 import androidx.constraintlayout.widget.ConstraintLayout
 import mozilla.components.feature.readerview.R
@@ -29,14 +28,14 @@ class ReaderViewControlsBar @JvmOverloads constructor(
     defStyleAttr: Int = 0
 ) : ConstraintLayout(context, attrs, defStyleAttr), ReaderViewControlsView {
 
-    private var inflated: Boolean = false
-
     override var listener: ReaderViewControlsView.Listener? = null
 
     private lateinit var fontIncrementButton: AppCompatButton
     private lateinit var fontDecrementButton: AppCompatButton
     private lateinit var fontGroup: RadioGroup
     private lateinit var colorSchemeGroup: RadioGroup
+
+    private var view: View? = null
 
     init {
         isFocusableInTouchMode = true
@@ -49,8 +48,6 @@ class ReaderViewControlsBar @JvmOverloads constructor(
      * @param font The applicable font types available.
      */
     override fun setFont(font: FontType) {
-        inflateIfNeeded()
-
         val selected = when (font) {
             FontType.SERIF -> R.id.mozac_feature_readerview_font_serif
             FontType.SANSSERIF -> R.id.mozac_feature_readerview_font_sans_serif
@@ -67,8 +64,6 @@ class ReaderViewControlsBar @JvmOverloads constructor(
      * @param size An integer value in the range [MIN_TEXT_SIZE] to [MAX_TEXT_SIZE].
      */
     override fun setFontSize(size: Int) {
-        inflateIfNeeded()
-
         val (incrementState, decrementState) = when {
             size <= MIN_TEXT_SIZE -> {
                 Pair(first = true, second = false)
@@ -90,8 +85,6 @@ class ReaderViewControlsBar @JvmOverloads constructor(
      * @param scheme The applicable colour schemes available.
      */
     override fun setColorScheme(scheme: ColorScheme) {
-        inflateIfNeeded()
-
         val selected = when (scheme) {
             ColorScheme.DARK -> R.id.mozac_feature_readerview_color_dark
             ColorScheme.SEPIA -> R.id.mozac_feature_readerview_color_sepia
@@ -105,8 +98,6 @@ class ReaderViewControlsBar @JvmOverloads constructor(
      * Updates visibility to [View.VISIBLE] and requests focus for the UI controls.
      */
     override fun showControls() {
-        inflateIfNeeded()
-
         visibility = View.VISIBLE
         requestFocus()
     }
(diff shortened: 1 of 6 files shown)


components/feature/readerview/src/main/java/mozilla/components/feature/readerview/view/ReaderViewControlsBar.kt
```diff
@@ -111,6 +109,16 @@ class ReaderViewControlsBar @JvmOverloads constructor(
         visibility = View.GONE
     }
 
+    override fun tryInflate(): Boolean {
```
mcomella: nit: can we comment this is done for Fenix performance reasons: I.e. it's not created until it's needed? Not sure if this is the best place for it.

I wonder if "not inflated until necessary" would be useful in the class kdoc (maybe not, as an implementation detail).

components/feature/readerview/src/main/java/mozilla/components/feature/readerview/view/ReaderViewControlsView.kt
```diff
@@ -45,6 +45,15 @@ interface ReaderViewControlsView {
...
     fun asView(): View = (this as View)
 
+    /**
+     * Tries to inflate the view if needed.
+     *
+     * See: https://github.com/mozilla-mobile/android-components/issues/5491
+     *
+     * @return true if the inflation was completed, false if the view was already inflated.
+     */
+    fun tryInflate(): Boolean
```
mcomella: nit: should this be internal? I don't like that this is exposed to consumers.

components/feature/readerview/src/test/java/mozilla/components/feature/readerview/view/ReaderViewControlsBarTest.kt
```diff
@@ -185,4 +185,12 @@ class ReaderViewControlsBarTest {
 
         verify(listener).onColorSchemeChanged(ReaderViewFeature.ColorScheme.SEPIA)
     }
+
+    @Test
+    fun `tryInflate is only successfully once`() {
+        val bar = ReaderViewControlsBar(appCompatContext)
+
+        assertTrue(bar.tryInflate())
```
mcomella: nit: I don't love that the return value is not used in production. As such, I'd prefer to remove it and instead write a test that mocks `context.getService<LayoutInflater>` (or whatever) to verify the inflation occurs.

That being said, we're pressed for time and this is a trivial detail so I actually don't think it's that important to fix this.

mcomella:
> I'm currently running into unrelated build issues with testing this in Fenix to assert the gains expected, but the basis of the change is largely the same so the code.

@colintheshots Maybe you can take a look?

components/feature/readerview/src/main/java/mozilla/components/feature/readerview/view/ReaderViewControlsBar.kt
jonalmeida: I had docs on the API, but I copied them over here as well.

components/feature/readerview/src/main/java/mozilla/components/feature/readerview/view/ReaderViewControlsView.kt
jonalmeida: Can't make it internal. 🤷‍♂ 

This is where we have to give way for the implementation sake.

components/feature/readerview/src/test/java/mozilla/components/feature/readerview/view/ReaderViewControlsBarTest.kt
jonalmeida: > That being said, we're pressed for time and this is a trivial detail so I actually don't think it's that important to fix this.

I agree, but maybe it'll be more useful if Fenix wants to "warm up" later?

(Spitballing ideas to make it feel worth it :)

jonalmeida:
> > I'm currently running into unrelated build issues with testing this in Fenix to assert the gains expected, but the basis of the change is largely the same so the code.
> 
> @colintheshots Maybe you can take a look?

Fenix builds are busted from missing dependencies from what I can tell.

> Failed to resolve: org.mozilla.components:feature-search:28.0.0-SNAPSHOT

colintheshots:
> @colintheshots Do you think this is an effective solution?

@mcomella I wrote much of this solution, so I expect it's a reasonable approach for this specific case of the problem.

@jonalmeida It looks like the build is fixed. If this is ready for review, we could make this a non-draft PR.

jonalmeida:
> @jonalmeida It looks like the build is fixed. If this is ready for review, we could make this a non-draft PR.

I just verified that the patch functions as your original one.

I'll leave this for @csadilek to do a final pass on it.

jonalmeida:
The lib-crash failure is fixed in a PR ready to land, the other two failures are connectiong reset network errors which releng are fixing soon.

components/feature/readerview/src/main/java/mozilla/components/feature/readerview/view/ReaderViewControlsView.kt
csadilek: nit: We're not doing (and aren't planning to do) anything with the returned boolean. So this also just be `maybeInflate` or `inflateIfNeeded` without a return value?

jonalmeida: I was using the return value only for writing out a test; Mike commented about that too. I wasn't sure what better way to get a test out of this without something more complicated by finding child views maybe?

csadilek: Yes, that's fair, it does make testing easier. I think it's fine to land as is.

jonalmeida:
bors r=csadilek

PR #5495, opened 2020-01-07
Closes #5460: Fennec migration: Disable currently unsupported add-ons

### LINKED ISSUE #5460: Fennec migration: Disable (currently) unsupported add-ons
This is for https://github.com/mozilla-mobile/fenix/issues/4983 to disable all (currently) unsupported add-ons.

Depends on: 
https://bugzilla.mozilla.org/show_bug.cgi?id=1599585
https://bugzilla.mozilla.org/show_bug.cgi?id=1604222

csadilek:
This is for the first iteration and currently disabling *all* installed addons. Later we want to match against supported addons from our AMO collection and/or an offline whitelist.

Made sure to capture all outcomes in `AddonMigrationResult`s so we can add telemetry/logs, as needed. Let me know if I missed anything :).

### DIFF
--- a/components/support/migration/src/main/java/mozilla/components/support/migration/AddonMigration.kt
+++ b/components/support/migration/src/main/java/mozilla/components/support/migration/AddonMigration.kt
@@ -1,0 +1,140 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.support.migration
+
+import kotlinx.coroutines.CompletableDeferred
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import mozilla.components.concept.engine.Engine
+import mozilla.components.concept.engine.webextension.EnableSource
+import mozilla.components.concept.engine.webextension.WebExtension
+
+/**
+ * Wraps [AddonMigrationResult] in an exception so that it can be returned via [Result.Failure].
+ *
+ * @property failure Wrapped [AddonMigrationResult] indicating exact failure reason.
+ */
+class AddonMigrationException(val failure: AddonMigrationResult.Failure) : Exception(failure.toString())
+
+/**
+ * Result of an add-on migration.
+ */
+sealed class AddonMigrationResult {
+    /**
+     * Success variants of an add-on migration.
+     */
+    sealed class Success : AddonMigrationResult() {
+        /**
+         * No add-ons installed to migrate.
+         */
+        object NoAddons : Success()
+
+        /**
+         * Successfully migrated add-ons.
+         */
+        data class AddonsMigrated(
+            val migratedAddons: List<WebExtension>
+        ) : Success() {
+            override fun toString(): String {
+                return "Successfully migrated ${migratedAddons.size} add-ons."
+            }
+        }
+    }
+
+    /**
+     * Failure variants of an add-on migration.
+     */
+    sealed class Failure : AddonMigrationResult() {
+        /**
+         * Failed to query installed add-ons via the [Engine].
+         */
+        internal data class FailedToQueryInstalledAddons(val throwable: Throwable) : Failure() {
+            override fun toString(): String {
+                return "Failed to query installed add-ons: ${throwable::class}"
+            }
+        }
+
+        /**
+         * Failed to migrate some or all of the installed add-ons.
+         */
+        internal data class FailedToMigrateAddons(
+            val migratedAddons: List<WebExtension>,
+            val failedAddons: Map<WebExtension, Exception>
+        ) : Failure() {
+            override fun toString(): String {
+                return "Failed to migrate ${failedAddons.size} of ${failedAddons.size + migratedAddons.size} add-ons."
+            }
+        }
+    }
+}
+
+/**
+ * Helper for migrating add-ons.
+ */
+internal object AddonMigration {
+
+    /**
+     * Performs a migration of all installed add-ons. The only migration step involved
+     * is to make sure we disable all currently unsupported add-ons.
+     */
+    @SuppressWarnings("TooGenericExceptionCaught")
+    suspend fun migrate(engine: Engine): Result<AddonMigrationResult> {
+        val installedAddons =
+        try {
+            getInstalledAddons(engine)
+        } catch (e: Exception) {
+            return Result.Failure(AddonMigrationException(AddonMigrationResult.Failure.FailedToQueryInstalledAddons(e)))
+        }
+
+        if (installedAddons.isEmpty()) {
+            return Result.Success(AddonMigrationResult.Success.NoAddons)
+        }
+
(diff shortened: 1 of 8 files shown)


components/support/migration/src/main/java/mozilla/components/support/migration/AddonMigration.kt
```diff
@@ -0,0 +1,135 @@
...
+            withContext(Dispatchers.Main) {
+                engine.listInstalledWebExtensions(
+                    onSuccess = { result.complete(it) },
+                    onError = { result.completeExceptionally(it) }
+                )
+            }
+        }.await()
+    }
+
+    private suspend fun migrateAddon(engine: Engine, addon: WebExtension): WebExtension {
```
grigoryk: Is the `addon` that's returned here actually different from `addon` that's passed-in? Or are you just round-tripping it (in which case, that seems unnecessary)?

components/support/migration/src/main/java/mozilla/components/support/migration/AddonMigration.kt
```diff
@@ -0,0 +1,135 @@
...
+            }
+
+            return if (failures.isEmpty()) {
+                Result.Success(AddonMigrationResult.Success.AddonsMigrated(migratedAddons))
+            } else {
+                val failure = AddonMigrationResult.Failure.FailedToMigrateAddons(migratedAddons, failures)
+                Result.Failure(AddonMigrationException(failure))
+            }
+        } catch (e: Exception) {
+            return Result.Failure(AddonMigrationException(AddonMigrationResult.Failure.FailedToQueryInstalledAddons(e)))
```
grigoryk: A bit of a nit, but `FailedToQueryInstalledAddons` won't be strictly the right failure if something other than `getInstalledAddons` fails.

components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt
```diff
@@ -709,4 +737,35 @@ class FennecMigrator private constructor(
...
+                val migrationFailureWrapper = result.throwables.first() as AddonMigrationException
+                return when (val failure = migrationFailureWrapper.failure) {
+                    is AddonMigrationResult.Failure.FailedToQueryInstalledAddons -> {
+                        logger.error("Failed to query installed add-ons: $failure")
+                        crashReporter.submitCaughtException(migrationFailureWrapper)
+                        result
+                    }
+                    is AddonMigrationResult.Failure.FailedToMigrateAddons -> {
+                        logger.error("Failed to migrate add-ons: $failure")
+                        crashReporter.submitCaughtException(migrationFailureWrapper)
```
grigoryk: I don't think we'll actually see the collected underlying exceptions in Sentry, right? In a similar situation in the logins exception, I collapse collected exceptions into a set and submit them individually - e.g. see `collectExceptionsIntoASet`

components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt
```diff
@@ -70,6 +71,11 @@ sealed class Migration(val currentVersion: Int) {
      * Migrates all Fennec settings backed by SharedPreferences.
      */
     object Settings : Migration(currentVersion = 1)
+
+    /**
+     * Migrates / Disables all currently unsupported Add-ons.
+     */
+    object Addons : Migration(currentVersion = 1)
```
grigoryk: You'll also need to add this migration to `MigrationResultsStore#fromJSON`, otherwise we'll crash when reading the persisted migrations.

pocmo: This has come up every time we add a migration. We should have modeled this in a way that the type system enforces it. But, well, I also do not want to refactor anything that late anymore.

components/support/migration/src/main/java/mozilla/components/support/migration/AddonMigration.kt
csadilek: Yes, the `addons` are different. The returned once have their enabled state updated. We want this so we can check the successfully migrated addons and we'd know if they were enabled/disabled (supported/unsupported). So, I think this makes sense to keep.

components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt
csadilek: Ah, thanks, good catch! Updated.

components/support/migration/src/main/java/mozilla/components/support/migration/AddonMigration.kt
csadilek: Yeah, that's fair. There's no need for a catch-all here as I am doing that in `FennecMigrator#migrateAddons` anyway. Refactored.

components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt
csadilek: Ah, yes, I remember this. Case here is slightly different as I am mapping the exceptions to their addon, but I made part of your logic re-usable and applied it here as well so we now
- properly report the underlying exception
- and only do it once (and not spam Sentry)

csadilek:
@grigoryk @pocmo I've addressed all comments now, so this is ready for another pair of 👀 .

components/support/migration/src/main/java/mozilla/components/support/migration/AddonMigration.kt
grigoryk: 👍 I thought it may be about the state flip, thank you for confirming.

components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt
grigoryk: Sounds great!

components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt
grigoryk: Totally agreed @pocmo, both that this is bad now, and that it's not really worth refactoring at this point.

components/support/migration/src/main/java/mozilla/components/support/migration/AddonMigration.kt
```diff
@@ -0,0 +1,137 @@
...
+                engine.listInstalledWebExtensions(
+                    onSuccess = { result.complete(it) },
+                    onError = { result.completeExceptionally(it) }
+                )
+            }
+        }.await()
+    }
+
+    private suspend fun migrateAddon(engine: Engine, addon: WebExtension): WebExtension {
+        // TODO (For final migration): check if addon is supported (against AMO via
```
grigoryk: Can you file an issue for this and link it here? Don't want to loose track of this.

csadilek: Done, linked to the issue now.

csadilek:
bors r=pocmo,grigoryk

csadilek:
bors retry

PR #5496, opened 2020-01-07
Closes #5373 - Add menu2 component

### LINKED ISSUE #5373: Redesign menu items to be pure data classes
The browser menu was written before we started using lib-state everywhere, and is now very difficult to integrate. Data is mixed with views, functionality has to be duplicated for different combinations of menu item features, and Fenix has needed to avoid lib-state to work with the menu.

We should rewrite the menu items to be more similar to `ContextMenuCandidate`. The menu should be represented as pure data classes, and functions like `visible` and `isHighlighted` should be changed to just be properties. The display code would no longer be part of the menu item classes.

This will also help with `MenuButton`, which sometimes displays a highlight depending on the state of the menu. Currently we need to know when to invalidate the menu button, so code must be duplicated to trigger it.

NotWoods:
browser-menu v2: A new browser menu designed to work with lib-state.

This introduces a new API using immutable data classes to represent options. Stateful functionality like the two-state button is replaced with swapping out data classes and submitting a new list of options.

Internally this uses a recycler view and the view holders intentionally do not map 1:1 to data classes. This way we can use different views based on properties instead of introducing all new classes for every combination of properties.

#### New features:
- No more mutable state.
- Simpler collection of items (no more separate HighlightableItem, HighlightableSwitch, HighlightableCheckbox...).
- Image buttons that allow for secondary actions (ie: last app that was shared to).
- Text icons that allow for extra text at the end of an option (ie: keyboard shortcut descriptions).
- Better keyboard support (focus starts at the beginning of the list and no items receive double-focus).
- Width of the menu can be customized.
- Multiple observers can watch for menu dismiss.

#### Removed features:
- No more support for custom menu items.

---

### Pull Request checklist

### After merge


components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenu.kt
```diff
@@ -135,7 +135,6 @@ class BrowserMenu internal constructor(
     }
```
Baron-Severin: (unrelated to this LOC)

I've been playing around with this API and I'm really loving it.  This solves just about every problem I can remember running into with the toolbar, and builds a strong framework for adding new functionality we might need in the future.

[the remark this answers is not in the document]
NotWoods: I'd be willing to do that, but this module also is 100% backwards compatible. If you don't use the new API and functions then you can continue to use the old `BrowserMenu` with no trouble.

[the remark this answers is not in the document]
NotWoods: We spoke about this during the Android meeting. This is fine to keep as its not removing the old version.

NotWoods:
To make this easier to review, I created #5649 with *just* the data classes and `asCandidate` function.

NotWoods:
What's currently blocking this?

NotWoods:
I could potentially split this PR again and separate the RecyclerView code from the rest of it

docs/changelog.md
```diff
@@ -12,6 +12,17 @@ permalink: /changelog/
...
+* **browser-menu2**
+  * Added new component to replace **browser-menu**. menu2 uses immutable data types and better integrates with `BrowserStore`.
+
+* **concept-menu**
+  * Added concept component to contain data classes for menu items and interfaces for menu controllers.
+
+* **browser-toolbar**
+  * Added support for the new `MenuController` interface for menu2.
+    When a menu controller is added to a toolbar, it will be used in place of the `BrowserMenuBuilder`.
+    The builder will supply items to the `MenuController` in `invalidateMenu` if it is kept.
```
psymoon, line 24: Unless I'm missing something, this doesn't apply to custom menu items, right? [WebExtensionBrowserMenuItem](https://github.com/mozilla-mobile/android-components/blob/master/components/browser/menu/src/main/java/mozilla/components/browser/menu/item/WebExtensionBrowserMenuItem.kt) for instance.

components/browser/menu2/src/main/java/mozilla/components/browser/menu2/view/MenuButton2.kt
```diff
@@ -0,0 +1,116 @@
...
+    defStyleAttr: Int = 0
+) : FrameLayout(context, attrs, defStyleAttr),
+    MenuButton,
+    View.OnClickListener,
+    Observable<MenuButton.Observer> by ObserverRegistry() {
+
+    /**
+     * Sets a [MenuController] that will be used to create a menu when this button is clicked.
+     */
+    override var menuController: MenuController? = null
```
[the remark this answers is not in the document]
NotWoods: The API is primarily like this so the `MenuButton` can be used in XML. I think the nullability doesn't hurt, although we could use a dummy controller by default if you prefer.

docs/changelog.md
NotWoods: Anything that implements `asCandidate` will work automatically. However, I don't plan to rely on this compat behaviour and I want to just migratate addons all at once.

components/browser/menu/src/main/java/mozilla/components/browser/menu/view/MenuButton.kt
```diff
@@ -28,24 +37,63 @@ class MenuButton @JvmOverloads constructor(
...
      */
     var getOrientation: () -> Orientation = {
         BrowserMenu.determineMenuOrientation(parent as? View?)
     }
 
+    /**
+     * Sets a [MenuController] that will be used to create a menu when this button is clicked.
+     * If present, [menuBuilder] will be ignored.
+     */
+    override var menuController: MenuController? = null
```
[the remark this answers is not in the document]
NotWoods: This property needs to be nullable in order to check if we're using the new menu2 controller or still in legacy "menu 1" mode.

components/browser/menu2/src/main/java/mozilla/components/browser/menu2/BrowserMenuController.kt
```diff
@@ -0,0 +1,92 @@
...
+    private var menuCandidates: List<MenuCandidate> = emptyList()
+    private var currentPopup: PopupWindow? = null
+
+    /**
+     * @param anchor The view on which to pin the popup window.
+     * @param orientation The preferred orientation to show the popup window.
+     * @param width The width of the popup menu. The height is always set to wrap content.
+     */
+    fun show(
+        anchor: View,
```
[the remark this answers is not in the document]
NotWoods: I've added a new candidate type and code to handle nested lists in the controller.

NotWoods:
Broke out some base functionality to #7466 and #7467

components/concept/menu/build.gradle
```diff
@@ -26,6 +26,7 @@ dependencies {
 
     implementation Dependencies.androidx_annotation
     implementation Dependencies.androidx_appcompat
+    implementation project(':support-base')
```
pocmo, line 29: I recently ran into a similar problem and wanted to avoid having a "concept" depend on actual code here. I opened the following RFC for creating a `concept-base` component just for interfaces. That may be relevant. :) https://github.com/mozilla-mobile/android-components/pull/7776

(This doesn't need to stop landing this though)

(You primarily need the `Observable` interface here?)

docs/changelog.md
```diff
@@ -120,6 +120,17 @@ permalink: /changelog/
 * **feature-addons**
   * Add `Addon.createdAtDate` and `Addon.updatedAtDate` extensions to get `Addon.createdAt` and `Addon.updatedAt` as a `Date`.
 
+* **browser-menu2**
```
pocmo, line 123: This needs to be moved to the latest version now. :)

components/concept/menu/build.gradle
NotWoods: Yup, this is for the `Observable` interface.

NotWoods:
bors r=pocmo

PR #5517, opened 2020-01-09
Closes #5345: Migration service: Use lowest priority and set category.

### LINKED ISSUE #5345: Give migration notification low priority
It's way too "in your face" right now. This may cause the user to try to kill the app to get rid of it. Let's prevent that by moving the priority down.

pocmo:
Closes #5345.


grigoryk:
From the [Android docs](https://developer.android.com/guide/components/services#Foreground):
```
If an app tries to hide its services by using a low-priority notification, this can impair the
performance of the app the user is actively interacting with. For this reason, if an app tries to run a
service with a minimum-priority notification, the system calls out the app's behavior in the
notification drawer's bottom section. 
```

Are we going to be hit by this?

pocmo:
I didn't notice any issue while testing. But let's play it safe. I'll get back to LOW and only add the code that sets the category (which also seems to impact how a notification shows up).

pocmo:
Okay, PR updated. Now only setting the category.

Issue #5260, opened 2019-12-08
Sample browser crashing due to FATAL EXCEPTION: IconThread


psymoon:
```
2019-12-08 16:30:55.911 11827-11924/org.mozilla.samples.browser E/AndroidRuntime: FATAL EXCEPTION: IconThread
    Process: org.mozilla.samples.browser, PID: 11827
    java.lang.IllegalArgumentException: Could not process image.
        at org.mozilla.gecko.mozglue.GeckoLoader.nativeRun(Native Method)
        at org.mozilla.gecko.GeckoThread.run(GeckoThread.java:476)
```

Seems to be happening on `geckoNightlyDebug`

psymoon:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1602565

psymoon:
@csadilek and I both verify the new gecko nightly update fixed this bug.

PR #5538, opened 2020-01-10
Issue #5260: Fall back to default icon if action icon fails to load

csadilek:
GV patch is pending, but this is good to keep anyway in case we ever fail to load the icon.


csadilek:
bors r=psymoon

PR #5546, opened 2020-01-12
Do not identify IPv6 addresses as search terms.

YoRyan:
Currently it is impossible to navigate to IPv6 literals (`[::1]:8080`) on the Android browsers because the addresses are misidentified as search terms. This PR incorporates a new regex adopted from [this notebook](https://nbviewer.jupyter.org/github/rasbt/python_reference/blob/master/tutorials/useful_regex.ipynb#Ipv6) and includes some test cases.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/support/utils/src/main/java/mozilla/components/support/utils/URLStringUtils.kt
+++ b/components/support/utils/src/main/java/mozilla/components/support/utils/URLStringUtils.kt
@@ -71,7 +71,7 @@ object URLStringUtils {
         // valid URL,
         // file:///home/user/myfile.html
         // is considered a search term; it is clearly a URL.
-        Pattern.compile("^\\s*\\w+(://[/]*|:|\\.)\\w+\\S*\\s*$", flags)
+        Pattern.compile("^\\s*\\w+(://[/]*|:|\\.)\\[?[\\w:]+\\S*\\s*$", flags)
     }
 
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
--- a/components/support/utils/src/main/java/mozilla/components/support/utils/WebURLFinder.kt
+++ b/components/support/utils/src/main/java/mozilla/components/support/utils/WebURLFinder.kt
@@ -204,6 +204,24 @@ class WebURLFinder {
         private const val PORT_NUMBER = "\\:\\d{1,5}"
 
         /**
+         * Regular expression to match IPv6 addresses, with optional zone identifier.
+         *
+         * Taken from:
+         * https://nbviewer.jupyter.org/github/rasbt/python_reference/blob/master/tutorials/useful_regex.ipynb#Ipv6
+         */
+        private const val IPV6_ADDRESS = (
+                "\\[" +
+                        "((([0-9A-Fa-f]{1,4}:){7}([0-9A-Fa-f]{1,4}|:))" +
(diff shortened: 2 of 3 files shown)


csadilek:
@YoRyan Thanks for your contribution and sorry we didn't get back to you sooner! We faced numerous performance issues with regular expressions in `WebURLFinder`:

https://github.com/mozilla-mobile/android-components/issues/5594
https://github.com/mozilla-mobile/android-components/issues/5880

As this fix would introduce a performance regression for us, would you mind filing an issue for us instead to support IPV6 addresses? Then we can prioritize it, discuss the trade-offs and investigate possibly faster solutions? Thanks again.

YoRyan:
[Done!](https://bugzilla.mozilla.org/show_bug.cgi?id=1631962)

YoRyan:
An issue has already [been filed](https://github.com/mozilla-mobile/fenix/issues/4343) on the Fenix side, but I believe the bug lies within this repository. Would you like another issue opened here?

lordgurke:
I have very little experience with Java, so this is just a quick thought on how I would combat the performance problems with those regular expressions:  
Would it be a feasible solution to do a very simple RegEx match first, i.E. `([0-9a-fA-F\:]+)|([0-9]+)` on the value entered into the address bar and then try to convert it with `inet_pton` into a binary representation (or the respective equivalent in Java).  
If the address entered is a valid IPv4 address you'll get a binary string with 4 bytes length, if it's an IPv6 address you'll get 16 bytes length, and if it's not a valid address at all, you'll get nothing (or an exception).  
If the latter happens, it's not a valid IP address and we can safely interpret the input as a hostname or a search term.  
With that way, we only have a very simple and fast RegEx matching on IP addresses and the underlying system has to decide if the input entered is a valid IP address representation - and that should also be fast enough.  
Also, this would safely include all possible variants on IP address representations.  
  
For information on `inet_pton` see for example https://linux.die.net/man/3/inet_pton  
  
  
So, like in Javaish pseudocode:
```
public static function isIpAddress(string possibleIpAddress) {
   try {
      // try to convert input string (i.e. "2001:db8:a:b:c:d::1") into byte array representing an IP address
      byte[] testAddr = inet_pton(possibleIpAddress);
      // if input is an IP address, we'll get either a 4 or 16 bytes long byte array
      if(testAddr.length == 4 || testAddr.length == 16) {
          return true;
      }
      else {
          return false;
      }
   }
   catch (Exception e) {
       // if input is not closely an IP address, we might get an exception
       return false;
   }
}
```

redturtlepower:
I'd rather have a browser that works. Did not bother looking at the performance related bugs, because seriously, you set a wrong priority. Just made an update, got new shiny UI makeover. Yay! I am on the edge of switching back to Chrome. Wake up, it is 2021.

NiKiZe:
> @YoRyan Thanks for your contribution and sorry we didn't get back to you sooner! We faced numerous performance issues with regular expressions in `WebURLFinder`:
> 
> #5594 #5880
> 
> As this fix would introduce a performance regression for us, would you mind filing an issue for us instead to support IPV6 addresses? Then we can prioritize it, discuss the trade-offs and investigate possibly faster solutions? Thanks again.

So you prefer a broken browser over some performance issue? (Just trying to make sure I understand your priorities)

Issue #5509, opened 2020-01-08
Fennec migration: Blocking migration UI


jonalmeida:
Feature: https://github.com/mozilla-mobile/fenix/issues/7092
Designs: https://miro.com/app/board/o9J_kw1CGow=/

While a migration is taking place, we want to block all entry points to the browser until the migration is complete as it would be in a volatile state.

#### Browser Entry Points
- Sharing
- Voice Search
- Custom Tabs
- Send Tab

jonalmeida:
We probably need a combination of things below:

- Disable the intentprocessors or make one that would consume and drop intents if we're migrating and takes you to the migration screen.
  - Covers sharing, but maybe not voice search?
- Custom Tabs
  - Should we disable the service?
  - Drop the intent and/or redirect to the migration UI?
- Push
  - Should be disable from working via FxaAccountManager during migration?
- Fullscreen app activity.
  - Can we lock it so you can't come out?

jonalmeida:
> Disable the intentprocessors or make one that would consume and drop intents if we're migrating and takes you to the migration screen. 

This is being done in https://github.com/mozilla-mobile/android-components/issues/5537.

> Fullscreen app activity. 

This will be an `AbstractMigrationProgressActivity` to make it simple for apps to extend and implement the UI updates when the relevant callbacks are invoked for in-progress and completed states.

> Search Widget
>   - These go directly to the HomeActivity and not through the IntentReceiverActivity.

This seems to be accurate - in Fenix's [`SearchWidgetProvider`][0], it creates an intent to go straight to the `HomeActivity` and therefore by-passing our Intent Processor. A better change would be to refactor that code so that it does go through the intent processor and make a `SearchWidgetIntentProcessor` to handle this intent mutation.

We need to make this change in:
 1. [`createTextSearchIntent`][1]
 2. [`createVoiceSearchIntent`][2]

Maybe we can live without this for nightly if we run out of time.

[0]: https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/gecko/search/SearchWidgetProvider.kt
[1]: https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/gecko/search/SearchWidgetProvider.kt#L106
[2]: https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/gecko/search/SearchWidgetProvider.kt#L116

liuche:
[...]

PR #5548, opened 2020-01-13
Closes #5537: Add IntentProcessor to handle intents when migrating

### LINKED ISSUE #5537: Create a MigrationIntentProcessor that handles intents when migration is active
We want to handle incoming intents gracefully while the migration is still taking place and the browser is in a volatile state.

Any incoming intents we should either drop and/or redirect you to the blocking Migration UI.

jonalmeida:
---

Since an app would need to handle what to do when migrating, with our case being to show a blocking UI, we need to delegate that to the app to do since the activity would live in the app.

In https://github.com/mozilla-mobile/android-components/issues/5509, we'll create an `AbstractMigrationProgressActivity` so that we can notify frontend code when the show progress and finished states without the app needing to observe the store directly.

### Pull Request checklist

### After merge

### DIFF
--- a/components/support/migration/src/main/java/mozilla/components/support/migration/MigrationIntentProcessor.kt
+++ b/components/support/migration/src/main/java/mozilla/components/support/migration/MigrationIntentProcessor.kt
@@ -1,0 +1,35 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.support.migration
+
+import android.content.Intent
+import mozilla.components.feature.intent.processing.IntentProcessor
+import mozilla.components.support.migration.state.MigrationProgress
+import mozilla.components.support.migration.state.MigrationStore
+
+/**
+ * An [IntentProcessor] that checks if we're in a migration state.
+ *
+ * ⚠️ When using this processor, ensure this is the first processor to be invoked if there are multiple.
+ */
+class MigrationIntentProcessor(private val store: MigrationStore) : IntentProcessor {
+
+    /**
+     * Matches itself with all intents in order to ensure processing all of incoming intents.
+     */
+    override fun matches(intent: Intent): Boolean = store.state.progress == MigrationProgress.MIGRATING
+
+    /**
+     * Processes all incoming intents if a migration is in progress.
+     *
+     * If this is true, we should show an instance of AbstractMigrationProgressActivity.
+     */
+    override suspend fun process(intent: Intent): Boolean {
+        return when (store.state.progress) {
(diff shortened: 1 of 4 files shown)


jonalmeida:
@pocmo while trying to add this to Fenix for testing I've reached an interesting problem with this intent processor solution: the `IntentReceiverActivity` lives in the `main` package and the intent processor needs a MigrationStore which only lives in the `migration` package.

I'm not sure if I can alias the IntentReceiverActivity in `migration` somehow.

OR

We move the migration store to the main package which isn't really ideal from the way it's currently setup.

jonalmeida:
> We move the migration store to the main package which isn't really ideal from the way it's currently setup.

I tried to do this locally in Fenix, and it didn't turn out to be as bad of a change needed. Just looking for comment if we're okay with doing this and there isn't something I've missed).

pocmo:
I was expecting this to come up sooner or later, when I added the MigrationStore (app code requiring access to the migration package).

Initially we tried to only add this package to the `fennec*` build variants since then it is nicely isolated. However I think it is fine to loosen this restriction: The Fennec variants more or less are going to be our **main app** (with fenix production and fenix beta going away) - so at that point the migration code will be in "all builds" anyways. We should still make sure to only ever try to migrate something for fennec builds only - but that should be guarded by `MigratingFenixApplication` already.

components/support/migration/src/main/java/mozilla/components/support/migration/MigrationIntentProcessor.kt
```diff
@@ -0,0 +1,48 @@
...
+ * An [IntentProcessor] that checks if we're in a migration state.
+ *
+ * ⚠️ When using this processor, ensure this is the first processor to be invoked if there are multiple.
+ */
+class MigrationIntentProcessor(private val store: MigrationStore) : IntentProcessor {
+
+    /**
+     * Matches itself with [ACTION_VIEW] intents in order to ensure processing all of incoming intents.
+     */
+    override fun matches(intent: Intent): Boolean = intent.toSafeIntent().action == ACTION_VIEW
```
pocmo: Wouldn't we want to handle other intents like `ACTION_SEND` here too?

jonalmeida: I realized that we can just return true instead while migrating so I'm going to change this to that instead. :)

jonalmeida: For reference, this is the change, so we end up catching all intents. I didn't need to do any fancy in the end:
```diff
-override fun matches(intent: Intent): Boolean = intent.toSafeIntent().action == ACTION_VIEW
+override fun matches(intent: Intent): Boolean = store.state.progress == MigrationProgress.MIGRATING
```

jonalmeida:
bors r=pocmo

Issue #5141, opened 2019-11-25
Finalize migration notification UX


pocmo:
Issue #4879 will land with temporary title, text and icon in the "We are migrating" notification. This issue is about finalizing the UX once we know how that notification should look like.

ebalazs-sv:
Verified as fixed on Nightly migration build 01/15 with Samsung Galaxy S8 (Android 9), Huawei Honor 8 (Android 7.0), Xiaomi Mi4i (Android 5.0.2) and the notification is as in [this](https://app.abstract.com/share/b1b9f719-fbb5-4c94-b45d-556232a3cd52?collectionLayerId=2c6d75d5-2f40-49be-b249-ef758aca0d6e&present=true&preview=false&sha=a9396e0ef08450c14d9e0554a86f8b70d860da05) abstract.
I will mark this as Done.

PR #5554, opened 2020-01-13
Issue #5141: Update strings of migration notification.

pocmo:
Replacing our temporary strings with the strings from the mock.


pocmo:
bors r=grigoryk

PR #5567, opened 2020-01-14
Closes #5432: Add API to query unsupported addons

### LINKED ISSUE #5432: AddonManager: Add API to query unsupported addons
During migration we will disable installed addons that are not yet supported. Since we want to display theses addons separately we need a way to query them (see "Not yet suported" section in https://miro.com/app/board/o9J_kw8Lt8g=/?moveToWidget=3074457347099638860).

For this we need to be able to differentiate if an addon was disabled by the user or if it was disabled during migration. See https://bugzilla.mozilla.org/show_bug.cgi?id=1604222.

csadilek:
This just brings in the flag/state so we know why an extension was disabled when we're querying it from GV. This is the only part that wasn't available in your original PR @psymoon. With this we can later add UI to indicate that the extension is installed but disabled because it wasn't supported, or we could write code to automatically enable these addons.

Let's talk more about it later...

### DIFF
--- a/components/feature/addons/src/main/java/mozilla/components/feature/addons/Addon.kt
+++ b/components/feature/addons/src/main/java/mozilla/components/feature/addons/Addon.kt
@@ -96,7 +96,8 @@ data class Addon(
         val version: String,
         val optionsPageUrl: String,
         val enabled: Boolean = false,
-        val supported: Boolean = true
+        val supported: Boolean = true,
+        val disabledAsUnsupported: Boolean = false
     ) : Parcelable
 
     /**
@@ -120,6 +121,14 @@ data class Addon(
      * Returns whether or not this [Addon] is currently supported by the browser.
      */
     fun isSupported() = installedState?.supported == true
+
+    /**
+     * Returns whether or not this [Addon] is currently disabled because it is not
+     * supported. This is based on the installed extension state in the engine. An
+     * addon can be disabled as unsupported and later become supported, so
+     * both [isSupported] and [isDisabledAsUnsupported] can be true.
+     */
+    fun isDisabledAsUnsupported() = installedState?.disabledAsUnsupported == true
 
     companion object {
         /**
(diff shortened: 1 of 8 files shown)


components/feature/addons/src/main/java/mozilla/components/feature/addons/Addon.kt
```diff
@@ -121,6 +122,11 @@ data class Addon(
      */
     fun isSupported() = installedState?.supported == true
 
+    /**
+     * Returns whether or not this [Addon] is currently disabled because it is not supported.
+     */
+    fun isDisabledAsUnsupported() = installedState?.disabledAsUnsupported == true
```
Amejia481: If `isDisabledAsUnsupported` is true should `supported` be `false`?

components/feature/addons/src/main/java/mozilla/components/feature/addons/Addon.kt
```diff
@@ -121,6 +122,11 @@ data class Addon(
      */
     fun isSupported() = installedState?.supported == true
```
psymoon: nit: comment mentioning this is based on AMO collection

components/feature/addons/src/main/java/mozilla/components/feature/addons/Addon.kt
```diff
@@ -121,6 +122,11 @@ data class Addon(
...
 
+    /**
+     * Returns whether or not this [Addon] is currently disabled because it is not supported.
```
psymoon: nit: comment mentioning this is based on [Engine]

components/feature/addons/src/main/java/mozilla/components/feature/addons/Addon.kt
csadilek: @Amejia481 I've added a comment. Both flags can be `true` at the same time. We might disable an extension during migration. It's state is then disabled as unsupported. Later this extension might be come supported as indicated by our AMO response. The both flags are true. The extension is now supported, but it is still disabled as unsupported. This is important because we can then trigger a flow to automatically enable those extensions. (We still need to discuss how this will look like to the user, see mocks.)

components/feature/addons/src/main/java/mozilla/components/feature/addons/Addon.kt
csadilek: Updated!

components/feature/addons/src/main/java/mozilla/components/feature/addons/Addon.kt
Amejia481: 👍

csadilek:
bors r=psymoon,Amejia481

PR #5575, opened 2020-01-14
Closes #5097: Add web domain for autofill in authentication dialog

### LINKED ISSUE #5097: Add Site Domain for Autofill to AuthenticationDialogFragment in feature-prompts
Currently the Fenix chrome in `AuthenticationDialogFragment` doesn't know anything about the website underneath, and thus shows only app ID related password autofill suggestions (ie shows only suggestions related to mozilla.org or firefox for every dialog). 

## Steps to reproduce
Using device with 3rd party password manager installed and logged in (Lastpass, Lockwise, etc).
Open `https://jigsaw.w3.org/HTTP/Basic/` in Samples Browser/Fenix/RB

### Expected behavior
Any saved credentials for the `w3.org` website should be shown

### Actual behavior
Saved credentials for `mozilla.org` are shown

See: https://github.com/mozilla-mobile/fenix/issues/4205

In theory we should follow @jhugman 's suggestion there to add the domain to the dialog view so the system password manager's know what to fill.

> This is a missing feature of Fenix's BasicAuth dialog. Android triggers an autofill request when an `EditText` is entered, and guesses at the autofill `ViewStructure` that password managers will use.
> 
> The password managers use the package name to find the account to offer to autofill.
> 
> For browsers, it needs to add the [`webDomain` property](https://developer.android.com/reference/android/view/ViewStructure#setWebDomain(java.lang.String)) for page specific text boxes. GeckoView is already doing this, so this is why it works in web content already, but not in Fenix's chrome.
> 
> You need to implement [`onProvideAutofillStructure`](https://developer.android.com/reference/android/view/View#onProvideAutofillStructure(android.view.ViewStructure,%20int)) on the dialog, and the set the [`webDomain` property](https://developer.android.com/reference/android/view/ViewStructure#setWebDomain(java.lang.String)).
> 
> This will fix the problem in all password managers, including Lockwise. Until then, it's impossible for them to know the correct hostname.
> 
> The Android doc [Optimize your app for autofill](https://developer.android.com/guide/topics/text/autofill-optimize) seems to be pretty comprehensive.

One caveat - I timeboxed this to a few hours and tried to make custom edit texts for the dialog that overrode `onProvideAutofillStructure` and set the domain passed into the dialog, and I couldn't get it to work.

rocketsroger:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/dialog/AuthenticationDialogFragment.kt
+++ b/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/dialog/AuthenticationDialogFragment.kt
@@ -13,7 +13,6 @@ import android.text.TextWatcher
 import android.view.LayoutInflater
 import android.view.View
 import android.view.View.GONE
-import android.widget.EditText
 import androidx.annotation.StringRes
 import androidx.annotation.VisibleForTesting
 import androidx.annotation.VisibleForTesting.PRIVATE
@@ -23,6 +22,7 @@ import mozilla.components.feature.prompts.R
 private const val KEY_USERNAME_EDIT_TEXT = "KEY_USERNAME_EDIT_TEXT"
 private const val KEY_PASSWORD_EDIT_TEXT = "KEY_PASSWORD_EDIT_TEXT"
 private const val KEY_ONLY_SHOW_PASSWORD = "KEY_ONLY_SHOW_PASSWORD"
+private const val KEY_SESSION_URL = "KEY_SESSION_URL"
 
 /**
  * [android.support.v4.app.DialogFragment] implementation to display a
@@ -32,6 +32,12 @@ private const val KEY_ONLY_SHOW_PASSWORD = "KEY_ONLY_SHOW_PASSWORD"
 internal class AuthenticationDialogFragment : PromptDialogFragment() {
 
     internal val onlyShowPassword: Boolean by lazy { safeArguments.getBoolean(KEY_ONLY_SHOW_PASSWORD) }
+
+    private var sessionUrl: String
+        get() = safeArguments.getString(KEY_SESSION_URL, "")
+        set(value) {
+            safeArguments.putString(KEY_SESSION_URL, value)
+        }
 
     internal var username: String
         get() = safeArguments.getString(KEY_USERNAME_EDIT_TEXT, "")
@@ -80,7 +86,10 @@ internal class AuthenticationDialogFragment : PromptDialogFragment() {
     }
 
     private fun bindUsername(view: View) {
-        val usernameEditText = view.findViewById<EditText>(R.id.username)
+        // Username field uses the AutofillEditText so if the user focus is here, the autofill
+        // application can get the web domain info without searching through the view tree.
+        val usernameEditText = view.findViewById<AutofillEditText>(R.id.username)
+        usernameEditText.sessionUrl = sessionUrl
 
         if (onlyShowPassword) {
             usernameEditText.visibility = GONE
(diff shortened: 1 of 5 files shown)


components/feature/prompts/src/main/java/mozilla/components/feature/prompts/dialog/AutofillEditText.kt
```diff
@@ -0,0 +1,27 @@
...
+    internal var sessionUrl: String = ""
+
+    constructor (context: Context) : super(context)
+
+    constructor (context: Context, attrs: AttributeSet?) : super(context, attrs)
+
+    constructor (context: Context, attrs: AttributeSet?, defStyleAttr: Int) : super(context, attrs, defStyleAttr)
+
+    override fun onProvideAutofillStructure(structure: ViewStructure?, flags: Int) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
grigoryk: Do you know how this supposed to work prior to 26?

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/dialog/AutofillEditText.kt
```diff
@@ -0,0 +1,27 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+package mozilla.components.feature.prompts.dialog
+
+import android.content.Context
+import android.os.Build
+import android.util.AttributeSet
+import android.view.ViewStructure
+import androidx.appcompat.widget.AppCompatEditText
+
+internal class AutofillEditText : AppCompatEditText {
```
grigoryk: A class comment here would be useful, explaining that this exists to make sure we call `setWebDomain` on the view structure.

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/dialog/AuthenticationDialogFragment.kt
```diff
@@ -80,7 +86,8 @@ internal class AuthenticationDialogFragment : PromptDialogFragment() {
     }
 
     private fun bindUsername(view: View) {
-        val usernameEditText = view.findViewById<EditText>(R.id.username)
+        val usernameEditText = view.findViewById<AutofillEditText>(R.id.username)
+        usernameEditText.sessionUrl = sessionUrl
```
grigoryk: You need this for both `username` and `password` fields, because a "login" might have an empty username and still be legit, right? If so, can you add a comment describing this?

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/dialog/AutofillEditText.kt
rocketsroger: No, Autofill framework was only available starting API 26
https://developer.android.com/guide/topics/text/autofill

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/dialog/AutofillEditText.kt
rocketsroger: Agreed.  I'll add the comment.  Thanks,

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/dialog/AuthenticationDialogFragment.kt
rocketsroger: Sure, I'll add comments to clarify why both needs to use `AutofillEditText`

rocketsroger:
bors r=grigoryk

PR #5581, opened 2020-01-15
FennecMigrator: Notify Glean about whether telemetry should be collected after migrating Fennec setting.

pocmo:


pocmo:
bors r=grigoryk

PR #5585, opened 2020-01-15
Closes #5355: Add web extension browser action browser-menu support

### LINKED ISSUE #5355: Web Extensions Fenix integration - BrowserAction in Menu
Similarly to `WebExtensionToolbarFeature`, Fenix mockup requires browser action exposure in context menu

psymoon:
Added `WebExtensionBrowserMenu`, `WebExtensionBrowserMenuBuilder` and `WebExtensionBrowserMenuItem` with live browser action update.


components/browser/menu/build.gradle
```diff
@@ -40,7 +43,7 @@ dependencies {
     testImplementation Dependencies.androidx_test_junit
     testImplementation Dependencies.testing_robolectric
     testImplementation Dependencies.testing_mockito
-    testImplementation Dependencies.androidx_test_core
```
psymoon: To reviewers: This was a dupe dependency

components/browser/menu/src/main/java/mozilla/components/browser/menu/WebExtensionBrowserMenu.kt
```diff
@@ -0,0 +1,107 @@
...
+    private var scope: CoroutineScope? = null
+
+    @kotlinx.coroutines.ExperimentalCoroutinesApi
+    override fun show(
+        anchor: View,
+        orientation: Orientation,
+        endOfMenuAlwaysVisible: Boolean,
+        onDismiss: () -> Unit
+    ): PopupWindow {
+        scope = store.flowScoped { flow ->
```
Amejia481: @psymoon Could we find a more generic alternative without having to create a custom `BrowserMenu`? I think there are similar needs to this one, other developers may want to add dynamically items to the menu as well, but this solution only works for `WebExtensions`. What do you think?

psymoon: `menuItems` is an immutable list by design so I am trying to follow the same pattern. We don't dynamically add new items to the menu - just update tab specific browser action overrides that are already in the menu (for add-on live update while the menu is opened like in fennac). I think the use case here is niche enough for a custom `BrowserMenu`.

AFAIK, @NotWoods is working on a long term solution for menu [here](https://github.com/mozilla-mobile/android-components/pull/5496)

Amejia481: What I'm a bit concern is the case when we want to add live updates to other items, which approuch should we take? Are we going to create a new type of   `BrowserMenu` that includes web exts and the new items? For moment, this solution is going work but it will be hard to scale it. If we don't other short term alternatives, I'm OK with keeping this one and refactor it later with a solution like  @NotWoods is proposing.

psymoon: @Amejia481 I'm just updating the object references and invalidating the respective views so you could technically accomplish that with current `BrowserMenu` implementation as well.

Amejia481: A lighter approach could be adding two callback to`BrowserMenu` on  `show` and another to the `onDismissListener`. This way we could have a  `WebExtensionBrowserMenuFeature`, that gets the `menu` and starts listing for changes on `onShow` and stops listing when `onDismissListener`. Exactly as you are doing here. What do you think?

psymoon: @Amejia481 I think it is a viable concern and we should address it in the near future, especially when @NotWoods is ready to land his refactor. For now, the priority is to change as little as possible and easy for us to integrate.

Amejia481: 👍

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/WebExtensionToolbarFeature.kt
```diff
@@ -87,20 +87,8 @@ class WebExtensionToolbarFeature(
...
-                    toolbarAction.browserAction = BrowserAction(
-                        title = it.title ?: browserAction.title,
-                        enabled = it.enabled ?: browserAction.enabled,
-                        badgeText = it.badgeText ?: browserAction.badgeText,
-                        badgeBackgroundColor = it.badgeBackgroundColor ?: browserAction.badgeBackgroundColor,
-                        badgeTextColor = it.badgeTextColor ?: browserAction.badgeTextColor,
-                        loadIcon = it.loadIcon ?: browserAction.loadIcon,
-                        onClick = it.onClick
-                    )
+                    toolbarAction.browserAction = toolbarAction.browserAction.copyWithOverride(it)
```
csadilek: This should be: `toolbarAction.browserAction = browserAction.copyWithOverride(it)`

components/browser/menu/src/main/java/mozilla/components/browser/menu/WebExtensionBrowserMenu.kt
```diff
@@ -0,0 +1,95 @@
...
+                    val browserMenuItem = webExtensionBrowserActions.getOrPut(extension.id) {
+                        val browserMenuItem = WebExtensionBrowserMenuItem(
+                            browserAction = browserAction,
+                            listener = browserAction.onClick
+                        )
+                        browserMenuItem
+                    }
+
+                    tab?.extensionState?.get(extension.id)?.browserAction?.let {
+                        browserMenuItem.browserAction = browserMenuItem.browserAction.copyWithOverride(it)
```
csadilek: This needs to be `browserMenuItem.browserAction = browserAction.copyWithOverride(it)` -> We want to start from the global one and applying the overrides of the tab one.

components/browser/menu/src/main/java/mozilla/components/browser/menu/WebExtensionBrowserMenu.kt
```diff
@@ -0,0 +1,95 @@
...
+                currentPopup = null
+                scope?.cancel()
+                webExtensionBrowserActions.clear()
+                onDismiss()
+            }
+        }
+    }
+
+    companion object {
+        internal val webExtensionBrowserActions = ConcurrentHashMap<String, WebExtensionBrowserMenuItem>()
```
csadilek: This doesn't need to be a concurrent map. We're not accessing from multiple threads.

components/browser/menu/src/main/java/mozilla/components/browser/menu/WebExtensionBrowserMenuBuilder.kt
```diff
@@ -0,0 +1,43 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.menu
+
+import android.content.Context
+import mozilla.components.browser.state.selector.selectedTab
+import mozilla.components.browser.state.store.BrowserStore
+
+/**
+ * Browser menu builder with web extension support.
```
csadilek: nit: Let's elaborate a bit here. How is it different than the regular [BrowserMenuBuilder]? i.e. it can add webextension actions...

components/browser/menu/src/main/java/mozilla/components/browser/menu/WebExtensionBrowserMenuBuilder.kt
```diff
@@ -0,0 +1,43 @@
...
+ * @param store [BrowserStore] required to render web extension browser actions
+ * @param appendExtensionAtStart true if web extensions appear at the top (start) of the menu,
+ * false if web extensions appear at the bottom of the menu. Default to false (bottom).
+ */
+class WebExtensionBrowserMenuBuilder(
+    items: List<BrowserMenuItem>,
+    extras: Map<String, Any> = emptyMap(),
+    endOfMenuAlwaysVisible: Boolean = false,
+    private val store: BrowserStore,
+    private val appendExtensionAtStart: Boolean = false
```
csadilek: `appendExtensionActionsAtStart` or similar to indicate what we're appending. It's not the extensions...

components/browser/menu/src/test/java/mozilla/components/browser/menu/WebExtensionBrowserMenuBuilderTest.kt
```diff
@@ -0,0 +1,110 @@
...
+import org.junit.Assert.assertNotNull
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.Mockito
+
+@RunWith(AndroidJUnit4::class)
+class WebExtensionBrowserMenuBuilderTest {
+
+    @Test
+    fun `web extension is added at the start when appendExtensionAtStart is true`() {
```
csadilek: nit `appendExtensionsAtStart` (see above)

components/browser/menu/src/test/java/mozilla/components/browser/menu/WebExtensionBrowserMenuTest.kt
```diff
@@ -0,0 +1,226 @@
...
+            WebExtensionBrowserAction("default_title", false, mock(), "", 0, 0) {}
+        val defaultExtensions: Map<String, WebExtensionState> = mapOf(
+            "id" to WebExtensionState("id", "url", true, defaultBrowserAction)
+        )
+
+        createTab(
+            "https://www.example.org", id = "tab1",
+            extensions = defaultExtensions
+        )
+
```
csadilek: nit: test is a bit hard to read with all these empty lines :)

components/browser/menu/src/test/java/mozilla/components/browser/menu/WebExtensionBrowserMenuTest.kt
```diff
@@ -0,0 +1,226 @@
...
+    @Test
+    fun `webExtensionBrowserActions is only updated when the menu is shown`() {
+        webExtensionBrowserActions.clear()
+        val browserAction = WebExtensionBrowserAction("browser_action", false, mock(), "", 0, 0) {}
+
+        val extensions = mapOf(
+            "browser_action" to WebExtensionState("browser_action", "url", true, browserAction)
+        )
+
+        val store = Mockito.spy(
```
csadilek: Why is the store a `spy`? We're not verifying anything on it, or are we?

components/browser/menu/src/test/java/mozilla/components/browser/menu/WebExtensionBrowserMenuTest.kt
```diff
@@ -0,0 +1,226 @@
...
+
+        val browserMenuItems = getOrUpdateWebExtensionMenuItems(store.state, store.state.selectedTab)
+        assertEquals(1, browserMenuItems.size)
+
+        val extension = browserMenuItems[0]
+        assertEquals("overridden_title", (extension as WebExtensionBrowserMenuItem).browserAction.title)
+    }
+
+    @Test
+    fun `getOrUpdateWebExtensionMenuItems does not include actions from disabled extensions `() {
```
csadilek: nit: ends with` `

components/browser/menu/src/test/java/mozilla/components/browser/menu/WebExtensionBrowserMenuTest.kt
```diff
@@ -0,0 +1,226 @@
...
+        webExtensionBrowserActions.clear()
+        val enabledAction = WebExtensionBrowserAction("enabled", false, mock(), "", 0, 0) {}
+        val disabledAction = WebExtensionBrowserAction("disabled", false, mock(), "", 0, 0) {}
+
+        val extensions = mapOf(
+            "enabled" to WebExtensionState("enabled", "url", true, enabledAction),
+            "disabled" to WebExtensionState("disabled", "url", false, disabledAction)
+        )
+
+        val store = Mockito.spy(
```
csadilek: Why a spy? See above? :)

components/concept/engine/src/main/java/mozilla/components/concept/engine/webextension/BrowserAction.kt
```diff
@@ -25,4 +25,21 @@ data class BrowserAction(
     val badgeTextColor: Int?,
     val badgeBackgroundColor: Int?,
     val onClick: () -> Unit
-)
+) {
+    /**
+     * Return a copy of [BrowserAction] with tab-specific [override] if it exists.
```
csadilek: nit: `Returns a copy of this [BrowserAction] with the provided overrides applied e.g. for tab-specific overrides`.

components/concept/engine/src/main/java/mozilla/components/concept/engine/webextension/BrowserAction.kt
```diff
@@ -25,4 +25,21 @@ data class BrowserAction(
...
+     *
+     * @param override tab-specific action only contain the values that should be overridden.
```
csadilek: nit: `@param override the action to use for overriding properties. Note that only the provided (non-null) properties of the override will be applied, all other properties will remain unchanged.`

samples/browser/src/main/java/org/mozilla/samples/browser/DefaultComponents.kt
```diff
@@ -215,7 +215,8 @@ open class DefaultComponents(private val applicationContext: Context) {
             },
             SimpleBrowserMenuItem("Add-ons") {
                 val intent = Intent(applicationContext, AddonsActivity::class.java)
-                intent.flags = Intent.FLAG_ACTIVITY_NEW_TASK
+                intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+                intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TASK)
```
csadilek: @psymoon oops, looks like this made it in by mistake.

samples/browser/src/main/java/org/mozilla/samples/browser/addons/AddonsActivity.kt
```diff
@@ -23,4 +25,12 @@ class AddonsActivity : AppCompatActivity() {
             }
         }
     }
+
+    override fun onBackPressed() {
```
csadilek: Looks like this code made it in by mistake.

components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/WebExtensionToolbarFeature.kt
```diff
@@ -87,20 +87,8 @@ class WebExtensionToolbarFeature(
                     toolbarAction
                 }
 
-                // Check if we have a tab-specific override and apply it.
-                // Note that tab-specific actions only contain the values that
-                // should be overridden. All other values will be null and should
-                // therefore *not* be applied / overridden.
```
csadilek: nit: Let's add a comment here `// Apply tab-specific override of browser action`

components/browser/menu/src/main/java/mozilla/components/browser/menu/WebExtensionBrowserMenu.kt
```diff
@@ -0,0 +1,94 @@
...
+                extension.browserAction?.let { browserAction ->
+                    // Add the global browser action if it doesn't exist
+                    val browserMenuItem = webExtensionBrowserActions.getOrPut(extension.id) {
+                        val browserMenuItem = WebExtensionBrowserMenuItem(
+                            browserAction = browserAction,
+                            listener = browserAction.onClick
+                        )
+                        browserMenuItem
+                    }
+
```
csadilek: nit: Let's add a comment here `// Apply tab-specific override of browser action`

psymoon:
bors r= csadilek

PR #5587, opened 2020-01-15
Refresh constellation state if not present

eoger:
Technically the caller of `SyncedTabsFeature.getSyncedTabs` could refresh devices beforehand but it's just easier to do it here.


jonalmeida:
@grigoryk this looks fine to me, do you have any concerns?

eoger:
This seems agreeable, let's do that!

PR #5614, opened 2020-01-16
For Fenix#7517: Update menu container style to use 16dp padding

psymoon:
Also allow `mozac_browser_menu_web_extension` to use `@style/Mozac.Browser.Menu.Item.Container`.

### DIFF
--- a/components/browser/menu/src/main/res/values/dimens.xml
+++ b/components/browser/menu/src/main/res/values/dimens.xml
@@ -11,8 +11,8 @@
     <!--Menu Item -->
     <dimen name="mozac_browser_menu_item_text_size">16sp</dimen>
     <dimen name="mozac_browser_menu_item_container_layout_height">48dp</dimen>
-    <dimen name="mozac_browser_menu_item_container_padding_start">24dp</dimen>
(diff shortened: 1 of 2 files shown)


components/browser/menu/src/main/res/values/dimens.xml
```diff
@@ -11,8 +11,8 @@
     <!--Menu Item -->
     <dimen name="mozac_browser_menu_item_text_size">16sp</dimen>
     <dimen name="mozac_browser_menu_item_container_layout_height">48dp</dimen>
-    <dimen name="mozac_browser_menu_item_container_padding_start">24dp</dimen>
```
Amejia481: Is it there any specific reason why do we want to change padding for all the menu items?

psymoon: @Amejia481 https://github.com/mozilla-mobile/fenix/issues/7157

Amejia481: Thanks, I just saw the change in isolation web ext!

psymoon:
bors r=Amejia481

PR #5620, opened 2020-01-17
For mozilla-mobile/fenix#7506 - Expose MenuButton

NotWoods:
New MenuButton to use in Fenix for the What's New option. This button is also used internally by the browser toolbar.

---

### Pull Request checklist

### After merge


components/browser/menu/src/main/java/mozilla/components/browser/menu/view/MenuButton.kt
```diff
@@ -0,0 +1,141 @@
...
+import mozilla.components.browser.menu.BrowserMenu
+import mozilla.components.browser.menu.BrowserMenu.Orientation
+import mozilla.components.browser.menu.BrowserMenuBuilder
+import mozilla.components.browser.menu.BrowserMenuHighlight
+import mozilla.components.browser.menu.R
+
+/**
+ * A `three-dot` button used for expanding menus.
+ *
+ * If you are using a browser toolbar, do not use this class directly.
```
Baron-Severin: Nit: up to you, but I would consider renaming this to `MenuButtonImpl`, just to further reinforce the point that this really isn't intended for public consumption.  Might also add something like this to the kdoc:

-----
WARNING: Prefer using a browser toolbar, which internally delegates to this class.

If you need this class (i.e., browser toolbar does not support one of your requirements), you will have to trigger most of its behavior manually, including invalidating menu items, setting highlighted state, etc.  See `browser-toolbar` `MenuButton` for example usage.

NotWoods: Part of the reason I rewrote this component was to make it more suited to public consumption :) I personally prefer to leave the current name, as I don't think `Impl` is a good name for a view.

The invalidation work is intentionally pulled out because it's not needed for the button itself, and I want to make it more generic so that its easier to integrate with the new menu component I've written later.

NotWoods:
bors r=Baron-Severin

PR #5649, opened 2020-01-20
For #5373 - Add menu candidate class

NotWoods:
Spun out of #5496

This is part of a new browser-menu v2 API, designed to work with lib-state.

This introduces a new API using immutable data classes to represent options. Stateful functionality like the two-state button is replaced with swapping out data classes and submitting a new list of options.

This PR *only* includes the data classes, plus some helpers to migrate from the old style to the new style. It does not affect the existing menus, which continue to work just fine.

See #5496 for an example of rendering code.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/menu/src/main/java/mozilla/components/browser/menu/item/WebExtensionBrowserMenuItem.kt
+++ b/components/browser/menu/src/main/java/mozilla/components/browser/menu/item/WebExtensionBrowserMenuItem.kt
@@ -4,15 +4,24 @@
 
 package mozilla.components.browser.menu.item
 
-import android.graphics.drawable.BitmapDrawable
+import android.content.Context
+import android.graphics.drawable.Drawable
 import android.view.View
 import android.widget.ImageView
 import android.widget.TextView
+import androidx.appcompat.content.res.AppCompatResources.getDrawable
+import androidx.core.graphics.drawable.toDrawable
 import kotlinx.coroutines.MainScope
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
 import mozilla.components.browser.menu.BrowserMenu
 import mozilla.components.browser.menu.BrowserMenuItem
 import mozilla.components.browser.menu.R
+import mozilla.components.browser.menu2.candidate.ContainerStyle
+import mozilla.components.browser.menu2.candidate.DrawableMenuIcon
+import mozilla.components.browser.menu2.candidate.TextMenuCandidate
+import mozilla.components.browser.menu2.candidate.TextMenuIcon
+import mozilla.components.browser.menu2.candidate.TextStyle
 import mozilla.components.concept.engine.webextension.Action
 import mozilla.components.support.base.log.Log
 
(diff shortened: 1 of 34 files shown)


components/browser/menu/src/main/java/mozilla/components/browser/menu/item/WebExtensionBrowserMenuItem.kt
```diff
@@ -83,4 +91,38 @@ class WebExtensionBrowserMenuItem(
         labelView.invalidate()
         badgeView.invalidate()
     }
+
+    @Suppress("TooGenericExceptionCaught")
```
Baron-Severin: Nit: why is `Throwable` an appropriate exception here?  It doesn't look like you follow this policy in AC, but on FFTV any suppression needed an explanatory comment.  If it didn't need one, then the rule should probably be globally relaxed.

components/browser/menu2/README.md
```diff
@@ -0,0 +1,44 @@
...
+Use Gradle to download the library from [maven.mozilla.org](https://maven.mozilla.org/) ([Setup repository](../../../README.md#maven-repository)):
+
+```Groovy
+implementation "org.mozilla.components:browser-menu2:{latest-version}"
+```
+
+### MenuController
+Sample code can be found in [Sample Toolbar app](https://github.com/mozilla-mobile/android-components/tree/master/samples/toolbar).
+
+There are multiple properties that you customize of the menu browser by just adding them into your dimens.xml file.
```
Baron-Severin: Will this still be true?  Not that I would expect these to change all that often, but it might be more internally consistent to include these at the top level of passed state (with sensible defaults).

.buildconfig.yml
```diff
@@ -176,6 +176,10 @@ projects:
     path: components/browser/menu
     description: 'A customizable menu for browsers.'
     publish: true
+  browser-menu2:
```
Baron-Severin: I'm split on this name.  Part of me thinks something like `immutable-browser-menu` would be better, part of me prefers that this one clearly communicates which module should be preferred.  I think it's probably best left as is, but I'm on the fence.

components/browser/menu/src/main/java/mozilla/components/browser/menu/item/WebExtensionBrowserMenuItem.kt
NotWoods: In this case I just copied the code that was already used. @psymoon Can you comment here?

components/browser/menu2/README.md
NotWoods: This will still be true, although the dimens mentioned have the old names (new one have menu2, not menu)

.buildconfig.yml
NotWoods: I'm going to stick with this name, since its already been merged in and agreed upon. I'm not a big number fan either which is why I originally built this in the same component.

NotWoods:
bors r=Baron-Severin

NotWoods:
bors retry

components/browser/menu/src/main/java/mozilla/components/browser/menu/item/WebExtensionBrowserMenuItem.kt
psymoon: @Baron-Severin @NotWoods generic throwable is fine here since this is for a fallback option like an else branch.

NotWoods:
bors retry

NotWoods:
bors retry

NotWoods:
bors retry

Issue #5263, opened 2019-12-09
Remove service stop from AutoPushFeature


jonalmeida:
With multiple push services, we don't want to obstruct those from running if we need to shutdown our push support.

We should remove `pushService.stop` from the `shutdown` call for this reason.

PR #5658, opened 2020-01-21
Issue #4925: Add support for scopes in AutoPushFeature and others

### LINKED ISSUE #3418: [WebPush] Support multiple push subscriptions
From https://github.com/mozilla-mobile/android-components/issues/1919#issuecomment-486677816:

> With MessageRouter and Delivery Manager (mozilla/application-services#545), we should have this covered.

There should be some work involved migrating from the a-c baked-in version of DeliveryManager to the a-s one.

This should support WebPush/PWA subscriptions.

GeckoView issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1343678

### LINKED ISSUE #4925: Add support for scopes in AutoPushFeature
For FxA support, we ignored supporting this since we didn't require it. For WebPush we need to pass the scope down to the native component.

This means adding support to:
 - `subscribeForType`
 - `unsubscribeForType`
 - Potentially removing `subscribeAll`

### LINKED ISSUE #5037: Add support for dispatchInfoForChid to RustPushConnection
In order to better support scopes in PushSubscriptions (#3946), we need to make use of the native query to reconstruct a subscription when we receive:
 - Messages
 - Subscriptions

jonalmeida:
---

We're updating the push API to be based on push scopes instead of channel IDs since that's the identifier we'd like to use going forward.

Optional: if we're unsure about moving to this new API, we can land everything up to part 3.

Else, this closes #4925, closes #5037, and closes #3418.

### Pull Request checklist

### After merge


jonalmeida:
Tested this with Send Tab and it still works.

Did a rough integration fit with WebPush, and it works too (with some encoding nits). :)

jonalmeida:
Screen Shot 2020-01-21 at 7 41 03 PM

My pride and joy. ❤️

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
```diff
@@ -193,72 +179,30 @@ class AutoPushFeature(
...
+    fun unsubscribe(scope: String, onUnsubscribe: ((Boolean) -> Unit) = {}) {
         DeliveryManager.with(connection) {
-            scope.launchAndTry {
-                PushType.values().forEach { type ->
-                    val sub = subscribe(type.toChannelId()).toPushSubscription()
-                    subscriptionObservers.notifyObservers { onSubscriptionAvailable(sub) }
+            coroutineScope.launchAndTry {
+                val result = unsubscribe(scope)
+                if (result) {
+                    onUnsubscribe(result)
```
grigoryk: If there's a possibility of failure for this operation, shouldn't you also have a `onError` callback as part of this API then? For consumers, currently it's a little strange - the block they passed in for `onUnsubscribe` may never get called - and there's a good chance this is what will "pump" consumer's logic forward.

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
```diff
@@ -193,72 +179,30 @@ class AutoPushFeature(
...
-     */
-    fun subscribeForType(type: PushType) {
+    fun subscribe(scope: String, appServerKey: String? = null, onSubscribe: ((AutoPushSubscription) -> Unit) = {}) {
         DeliveryManager.with(connection) {
-            scope.launchAndTry {
-                val sub = subscribe(type.toChannelId()).toPushSubscription()
-                subscriptionObservers.notifyObservers { onSubscriptionAvailable(sub) }
+            coroutineScope.launchAndTry {
+                val sub = subscribe(scope, appServerKey)
+                onSubscribe(sub)
```
grigoryk: same comment as for `unsubscribe` - don't we also want a `onError` callback here?

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
```diff
@@ -338,6 +286,22 @@ class AutoPushFeature(
...
 
+    /**
+     * Observers that want to receive updates for new subscriptions and messages.
+     */
+    interface Observer {
+
+        /**
+         * A subscription for the scope is available.
+         */
+        fun onSubscriptionChanged(scope: String) = Unit
```
grigoryk: In the README and inline docs above, you have `override fun onSubscriptionChanged(subscription: AutoPushSubscription) { }`, which doesn't match this signature.

components/feature/push/src/main/java/mozilla/components/feature/push/Connection.kt
```diff
@@ -22,18 +22,47 @@ import java.util.UUID
...
+
+    /**
+     * Updates the registration token to the native Push API if it changes.
+     */
     suspend fun updateToken(token: String): Boolean
+
+    /**
+     * Checks validity of current push subscriptions.
+     *
+     * @return true if a push subscription was updated and a subscriber should be notified of the change.
```
grigoryk: This will change to a `Map<PushScope, Boolean>`, right? Surely we don't want to notify _everyone_ if just one subscription is updated? If so, can you add a link to a github follow-up here?

components/feature/push/src/main/java/mozilla/components/feature/push/Connection.kt
```diff
@@ -49,23 +78,34 @@ internal class RustPushConnection(
...
-    override suspend fun subscribe(channelId: String, scope: String): SubscriptionResponse = synchronized(this) {
+    override suspend fun subscribe(
+        scope: PushScope,
+        appServerKey: AppServerKey?
+    ): AutoPushSubscription = synchronized(this) {
         val pushApi = api
         check(pushApi != null) { "Rust API is not initiated; updateToken hasn't been called yet." }
-        return pushApi.subscribe(channelId, scope)
+
+        val channelId = scope.toChannelId()
```
grigoryk: Somewhere here would be an excellent place to describe in a comment that Autopush really operates over these `channelID` things, while the WebPush world is built around a similar (but different) concept of `scope`. Here is the point of transition.

components/feature/push/src/main/java/mozilla/components/feature/push/Connection.kt
```diff
@@ -106,26 +146,32 @@ internal class RustPushConnection(
...
         encoding: String,
         salt: String,
         cryptoKey: String
-    ): ByteArray = synchronized(this) {
+    ): Pair<PushScope, ByteArray> = synchronized(this) {
         val pushApi = api
         check(pushApi != null) { "Rust API is not initiated; updateToken hasn't been called yet." }
-        return pushApi.decrypt(channelID = channelId, body = body, encoding = encoding, salt = salt, dh = cryptoKey)
+
+        val scope = pushApi.dispatchInfoForChid(channelId)?.scope.orEmpty()
```
grigoryk: So `scope` here could be empty? And the pair below would essentially by `("", 'some data')`?

Is this something you expect to see? seems like if this happens, we can't meaningfully process the decrypted message - how can consumers determine that the message is directed at them?

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
```diff
@@ -338,6 +286,22 @@ class AutoPushFeature(
...
+
+        /**
+         * A messaged has been received for the [scope].
+         */
+        fun onMessageReceived(scope: String, message: ByteArray?) = Unit
```
grigoryk: Looking at your FxA code, this API feels a bit like a foot-gun. I _think_ you already have a problem in one consumer, where it's not filtering messages by `scope` and just tries to process everything. How can we then expect other consumers not to repeat this error themselves?

Feels to me that you'd ideally have something like `observeMessages(scope: String, handler:...)`, and `observeAllMessages(handler:...)`. Currently you only have the latter (`register(...)`).

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
grigoryk: Almost everywhere else `scope` is typed as `PushScope`, but here it's a String. Can you stick to one type throughout?

components/feature/push/src/main/java/mozilla/components/feature/push/Connection.kt
```diff
@@ -143,3 +189,24 @@ internal fun ServiceType.toBridgeType() = when (this) {
 
 @VisibleForTesting
 internal fun Protocol.asString() = name.toLowerCase(Locale.ROOT)
+
+internal typealias PushScope = String
+internal typealias AppServerKey = String
```
grigoryk: Are these `typealiase` definitions valuable?

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
```diff
@@ -128,8 +128,6 @@ class AutoPushFeature(
      * This should only be done on an account logout or app data deletion.
      */
     override fun shutdown() {
-        service.stop()
```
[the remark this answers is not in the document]
jonalmeida: This is for https://github.com/mozilla-mobile/android-components/issues/5263.

With multiple features that rely on push, we shouldn't shut down the service when only the feature needs it. In Fenix, for example, Leanplum push service should still be allowed to function.

components/feature/accounts-push/src/main/java/mozilla/components/feature/accounts/push/FxaPushSupportFeature.kt
```diff
@@ -44,29 +42,45 @@ internal const val LAST_VERIFIED = "last_verified_push_subscription"
...
     owner: LifecycleOwner = ProcessLifecycleOwner.get(),
     autoPause: Boolean = false
 ) {
+
+    /**
+     * A unique scope for the FxA push subscription.
+     */
+    private val pushScope: String by lazy {
+        // Generate a unique scope if one doesn't exist.
+        val scope = UUID.randomUUID().toString().replace("-", "")
```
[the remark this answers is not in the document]
jonalmeida: > Even though this is in a lazy, you should still look it up first in the prefs and avoid generating a random UUID and doing string manipulations on it when you don't need to. 

For `SharedPreferences.getString` I need to provide a default string even if I do the pref check first (and for push that can't be null), so generating a UUID is pretty cheap since it's only done once per-startup.

components/feature/accounts-push/src/main/java/mozilla/components/feature/accounts/push/FxaPushSupportFeature.kt
```diff
@@ -18,18 +17,17 @@ import mozilla.components.concept.sync.DeviceConstellationObserver
...
 import mozilla.components.support.base.log.logger.Logger
 import mozilla.components.support.base.utils.SharedPreferencesCache
 import org.json.JSONObject
+import java.util.UUID
 
-internal const val PREFERENCE_NAME = "mozac_feature_accounts"
-internal const val LAST_VERIFIED = "last_verified_push_subscription"
+internal const val PREFERENCE_NAME = "mozac_feature_accounts_push"
+internal const val PREF_LAST_VERIFIED = "last_verified_push_subscription"
+internal const val PREF_FXA_SCOPE = "mozac_fxa_push_scope"
```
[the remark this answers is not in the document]
jonalmeida: > I guess we don't want to use the same scope/channelId for different FxA accounts, right? Is this the reason for generating a random one? It'll be great to have a comment pointing out the reason.

Having a deterministic scope means the UUID is deterministic too, and anything we put for the scope would mean the same UUID is used for all Fenix devices. Collision between accounts isn't an issue here, but it has a bit of code smell if every account has the same push UUID? Generating a random string to generate a random UUID seemed the same as just using a random UUID string to begin with.

components/feature/accounts-push/src/main/java/mozilla/components/feature/accounts/push/FxaPushSupportFeature.kt
```diff
@@ -119,7 +142,7 @@ internal class ConstellationObserver(
     private val verifier: VerificationDelegate = VerificationDelegate(context)
 ) : DeviceConstellationObserver {
 
-    private val logger = Logger("ConstellationObserver")
+    private val logger = Logger(ConstellationObserver::class.java.simpleName)
```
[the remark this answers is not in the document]
jonalmeida: Yeah, that's fair. I noticed that when other classes were renamed the loggers was frequently forgotten, so I figured this was an easy way to ensure they always have the same name.

components/feature/accounts-push/src/main/java/mozilla/components/feature/accounts/push/FxaPushSupportFeature.kt
```diff
@@ -141,33 +164,16 @@ internal class ConstellationObserver(
...
-            }
-        }
-    }
+internal class AutoPushObserver(private val accountManager: FxaAccountManager) : AutoPushFeature.Observer {
+    private val logger = Logger(AutoPushObserver::class.java.simpleName)
 
-    override fun onEvent(type: PushType, message: String) {
-        logger.debug("Received new push message for $type")
+    override fun onMessageReceived(scope: String, message: ByteArray?) {
+        logger.info("Received new push message for $scope")
```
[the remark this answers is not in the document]
jonalmeida: We don't need to from our previous conversations since fxa will just drop messages it can't parse, but I agree, a simple check here since we have the context of our own scope short-circuits that out a lot quicker.

components/feature/push/README.md
```diff
@@ -37,21 +37,21 @@ feature.initialize()
 feature.shutdown()
 
 // To receive the subscription info for all the subscription changes.
-feature.registerForSubscriptions(object : PushSubscriptionObserver {
-  override fun onSubscriptionAvailable(subscription: AutoPushSubscription) {
-    // handle subscription info here.
+feature.register(object : AutoPushFeature.Observer {
```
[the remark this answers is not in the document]
jonalmeida: I can't change the API for [`Observable`](https://github.com/mozilla-mobile/android-components/blob/master/components/support/base/src/main/java/mozilla/components/support/base/observer/Observable.kt), but maybe a better naming for `subscribe`/`unsubscribe`?

`createSubscription`/`removeSubscription`?

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
```diff
@@ -163,25 +157,17 @@ class AutoPushFeature(
...
-                val decrypted = decrypt(
+                val (scope, decryptedMessage) = decryptMessage(
                     channelId = message.channelId,
                     body = message.body,
                     encoding = message.encoding,
                     salt = message.salt,
                     cryptoKey = message.cryptoKey
                 )
-                messageObserverBus.notifyObservers(type, String(decrypted))
+                notifyObservers { onMessageReceived(scope, decryptedMessage) }
```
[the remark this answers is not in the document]
jonalmeida: > I also wonder if this could be a source of problems (perf, at least?) if websites decide to programmatically register for a very large number of unique scopes.

We had the MessageBus here for that reason. 🙂 

Do you think we should bring it back?

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
jonalmeida: Oops, you're right, we shouldn't filter that out.

> shouldn't you also have a onError callback as part of this API then?
We shouldn't need to be explicit calling `onError` since `launchAndTry` will invoke that if we receive an error.

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
jonalmeida: Thanks, I recently simplified that API and forgot about it. :)

components/feature/push/src/main/java/mozilla/components/feature/push/Connection.kt
jonalmeida: > Surely we don't want to notify everyone if just one subscription is updated?

In the case that we have a single subscription updated, we will be notifying everyone. Changing this also involves re-introducing the MessageBus.

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
jonalmeida: Should be the same as the comment above.

components/feature/push/src/main/java/mozilla/components/feature/push/Connection.kt
jonalmeida: > Is this something you expect to see? 

This can happen if there there is some corrupt subscription in the internal database from what I understand.

> seems like if this happens, we can't meaningfully process the decrypted message - how can consumers determine that the message is directed at them?

Messages can still be decrypted since the scope is only required for the final delivery and not for the decryption. A consumer would simply miss that message.

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
jonalmeida: I'm a bit confused by this since we wanted to remove the MessageBus and have consumers ensure they weren't invoking their code unnecessarily. Do we want to go back on this?

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
jonalmeida: The `typealias` was internal to just the `Connection` to avoid confusion between the `scope` and `appServerKey` since they're all strings.

I can expose that to consumers if it seems clearer.

components/feature/push/src/main/java/mozilla/components/feature/push/Connection.kt
jonalmeida: They're internal to avoid confusion when switching between the different String values since they don't have types and were getting confusing during implementation.

components/feature/accounts-push/src/main/java/mozilla/components/feature/accounts/push/FxaPushSupportFeature.kt
```diff
@@ -76,36 +90,45 @@ class FxaPushSupportFeature(
...
 
-        push.unsubscribeForType(PushType.Services)
+        push.unsubscribe(pushScope)
 
-        // Delete cached value of last verified timestamp when we log out.
-        preference(context).edit().remove(LAST_VERIFIED).apply()
+        // Delete cached value of last verified timestamp and scope when we log out.
+        preference(context).edit()
+            .remove(PREF_LAST_VERIFIED)
+            .remove(PREF_FXA_SCOPE)
```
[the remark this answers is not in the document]
jonalmeida: Same comment above, there's no strong reasoning to, but it seems odd if every FxA user had the exact same subscription ID for FxA.

> why is it important to erase the scope value?

If we stick with that reasoning, removal is so that a new user doesn't have the same subscription ID as the previous user.

components/feature/accounts-push/src/main/java/mozilla/components/feature/accounts/push/FxaPushSupportFeature.kt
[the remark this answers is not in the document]
grigoryk: That's a good point, one less thing to worry about!

components/feature/push/src/main/java/mozilla/components/feature/push/Connection.kt
grigoryk: 👍

components/feature/push/src/main/java/mozilla/components/feature/push/Connection.kt
grigoryk: "Missing a message" is partially what I meant by "we can't meaningfully process the decrypted message". What's the point of sending this message onto observers if we know they can't process it anyway?

components/feature/accounts-push/src/main/java/mozilla/components/feature/accounts/push/FxaPushSupportFeature.kt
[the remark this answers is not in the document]
grigoryk: I would frame thinking about this in the context of the larger system. Thinking through how autopush/FxA/etc all work together, what does it mean for every account to have the same channelId?

How is this different from a WebPush scenario, where every user of a webapp gets the same scope generated? The FxA case doesn't seem meaningfully different to me.

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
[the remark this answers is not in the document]
grigoryk: Okay, makes sense. I would document this right where you `start` the service, to explain why there isn't a matching `stop`. The contract here is that we `start` it if it's needed, but don't `stop` it when we don't need it, since other components may be around that rely on the service.

This also means that we never `stop` this service at all, right?

components/feature/accounts-push/src/main/java/mozilla/components/feature/accounts/push/FxaPushSupportFeature.kt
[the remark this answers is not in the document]
grigoryk: Why is that weird? It seems fine to me. As mentioned above, this also matches semantics of webpush. Each user has a unique UAID, and with that UAID you may associate "fxa channel id", "cnn.com channel id", etc. These will all end up as unique endpoints exposed by autopush.

components/feature/accounts-push/src/main/java/mozilla/components/feature/accounts/push/FxaPushSupportFeature.kt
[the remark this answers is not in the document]
grigoryk: The less possibility for things to go wrong, the better :)

components/feature/push/README.md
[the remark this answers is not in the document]
grigoryk: I don't think that helps... Leave it as-is, but maybe try to document better in the docs what these do?

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
grigoryk: Thinking about GV API is useful here - I'm assuming it's shaped something like "onMessageReceived(scope, message)", right? Since it needs to handle messages for arbitrary scopes, I'm not sure what else it can be. If that's the case, you're totally right about the shape of this!

Something to consider for the future is a perf improvement - if we _know_ that a message is received for the deterministic FxA scope (see my other comments about this), we know that we don't need to send it to GV, and it could go straight to our FxA consumer. And vice versa, no need to send GV-bound messages to FxA. This may imply an API level change (you need to somehow know which scopes observers care/don't care about), or you could do this by inspecting observers themselves (e.g. which class is the observer?). The latter does feel a bit like magic to me and seemingly violates the public API... but it's an easy way to achieve this improvement, if we need it.

components/feature/push/src/main/java/mozilla/components/feature/push/Connection.kt
jonalmeida: I've added implementation notes for the `RustPushConnection` to document the relation between those two concepts.

components/feature/push/src/main/java/mozilla/components/feature/push/Connection.kt
```diff
@@ -66,6 +82,30 @@ interface PushConnection : Closeable {
...
 }
 
+/**
+ * An implementation of [PushConnection] for the native component using the [PushAPI].
+ *
+ * Implementation notes: There are few important concepts to note here - for WebPush, the identifier that is required
+ * for notifying clients is called a "scope". This can be a site's host URL or an other uniquely identifying string to
+ * create a push subscription for that given scope.
+ *
+ * In the native Rust component the concept is similar, in that we have a unique ID for each subscription, however
```
grigoryk: The unique thing here is the pair `(uaid, channelId)`, right? Which is hidden from consumers of this API. E.g. this code running on different devices, when encountering the same `scope`, won't collapse into the same notification endpoint on the autopush side.

components/feature/push/src/main/java/mozilla/components/feature/push/Connection.kt
```diff
@@ -66,6 +82,30 @@ interface PushConnection : Closeable {
...
+ * the ID is called a "channel ID" (or chid), which is a UUID. In the implementation, the scope can also be provided
+ * to the native layer, however it is mostly just stored in the database but not used in any significant way.
```
grigoryk: Maybe worth calling out that it _may_ be used for stuff in the future, e.g. rate limiting, but without immediate plans to do so (e.g. no tracking issues in github!)

components/feature/push/src/main/java/mozilla/components/feature/push/Connection.kt
```diff
@@ -66,6 +82,30 @@ interface PushConnection : Closeable {
...
+ * With this in mind, we decided to write our implementation to be a 1-1 mapping of the public API 'scope' to the
+ * internal API 'channel ID' by generating a UUID from the scope value. This means that we have a reproducible way to
+ * always retrieve the chid when the caller passes the scope to us.
+ *
+ * Some nuances are that we also need to provide the native API the same scope value along with the
+ * scope-based chid value to satisfy the internal API requirements, at the cost of some noticeable duplication of
+ * information, so that we can retrieve those values later; see [RustPushConnection.subscribe] and
+ * [RustPushConnection.unsubscribe] implementations for details.
+ *
+ * Another nuance, in order to satisfy the API shape we need for push subscriptions, is to query the database for
```
grigoryk: Maybe I'm not reading this right, but english seems off here. Can you reword this so it's a bit clearer? I'm not 100% what you're trying to say here.

components/feature/push/src/main/java/mozilla/components/feature/push/Connection.kt
jonalmeida: The `uaid` isn't something that's exposed to us in anyway so it's hidden to a-c as well, but I see the point you're making, that we should call out the uniqueness is per device as well.

components/feature/push/src/main/java/mozilla/components/feature/push/Connection.kt
jonalmeida: I swallowed some words there. Fixed! 😄

jonalmeida:
We spoke about the FxA scope offline and decided the current implementation will work for now since we're generating a random scope for FxA and this would avoid collisions with other apps that may use the same scope.

components/feature/push/src/main/java/mozilla/components/feature/push/Connection.kt
jonalmeida: Ah fair, I guess we can use the lack of chid/scope to avoid decrypting the message and notifying clients since that probably means there are bigger problems if we get a valid message but the chid/scope doesn't exist.

jonalmeida:
bors r=grigoryk

I verified that push notifications work on Fenix and R-B. There are some site permission bugs with notifications that are unrelated to push.

PR #5669, opened 2020-01-21
F5695 add search to context menu

Baron-Severin:
For mozilla-mobile/fenix#5695
 
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineView.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineView.kt
@@ -10,9 +10,11 @@ import android.util.AttributeSet
 import android.widget.FrameLayout
 import androidx.annotation.VisibleForTesting
 import androidx.core.view.ViewCompat
+import mozilla.components.browser.engine.gecko.selection.GeckoSelectionActionDelegate
 import mozilla.components.concept.engine.EngineSession
 import mozilla.components.concept.engine.EngineView
 import mozilla.components.concept.engine.permission.PermissionRequest
+import mozilla.components.concept.engine.selection.SelectionActionDelegate
 import org.mozilla.geckoview.BasicSelectionActionDelegate
 import org.mozilla.geckoview.GeckoResult
 import java.lang.IllegalStateException
@@ -75,6 +77,8 @@ class GeckoEngineView @JvmOverloads constructor(
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
     internal var currentSelection: BasicSelectionActionDelegate? = null
 
+    override var selectionActionDelegate: SelectionActionDelegate? = null
+
     init {
         // Currently this is just a FrameLayout with a single GeckoView instance. Eventually this
         // implementation should handle at least two GeckoView so that we can switch between
@@ -104,8 +108,9 @@ class GeckoEngineView @JvmOverloads constructor(
 
             try {
                 currentGeckoView.setSession(internalSession.geckoSession)
-                currentSelection =
-                    internalSession.geckoSession.selectionActionDelegate as? BasicSelectionActionDelegate
+                currentSelection = GeckoSelectionActionDelegate.maybeCreate(context, selectionActionDelegate)
+                    ?: internalSession.geckoSession.selectionActionDelegate as? BasicSelectionActionDelegate
+                internalSession.geckoSession.selectionActionDelegate = currentSelection
             } catch (e: IllegalStateException) {
                 // This is to debug "display already acquired" crashes
                 val otherActivityClassName =
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/selection/GeckoSelectionActionDelegate.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/selection/GeckoSelectionActionDelegate.kt
@@ -1,0 +1,61 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.engine.gecko.selection
+
+import android.app.Activity
+import android.content.Context
+import android.view.MenuItem
+import mozilla.components.concept.engine.selection.SelectionActionDelegate
+import org.mozilla.geckoview.BasicSelectionActionDelegate
+
+/**
+ * An adapter between the GV [BasicSelectionActionDelegate] and a generic [SelectionActionDelegate].
+ *
+ * @property customDelegate handles as much of this logic as possible.
+ */
+open class GeckoSelectionActionDelegate(
+    activity: Activity,
+    private val customDelegate: SelectionActionDelegate
+) : BasicSelectionActionDelegate(activity) {
+
+    companion object {
+        /**
+         * @returns a [GeckoSelectionActionDelegate] if [customDelegate] is non-null and [context]
+         * is an instance of [Activity]. Otherwise, returns null.
+         */
+        fun maybeCreate(context: Context, customDelegate: SelectionActionDelegate?): GeckoSelectionActionDelegate? {
+            return if (context !is Activity || customDelegate == null) {
+                null
+            } else {
+                GeckoSelectionActionDelegate(context, customDelegate)
+            }
+        }
+    }
+
+    override fun getAllActions(): Array<String> {
+        return super.getAllActions() + customDelegate.getAllActions()
+    }
+
+    override fun isActionAvailable(id: String): Boolean {
+        val customActionIsAvailable = customDelegate.isActionAvailable(id) &&
+            !mSelection?.text.isNullOrEmpty()
+
+        return customActionIsAvailable ||
+            super.isActionAvailable(id)
+    }
+
+    override fun prepareAction(id: String, item: MenuItem) {
+        val title = customDelegate.getActionTitle(id)
+            ?: return super.prepareAction(id, item)
+
+        item.title = title
+    }
+
+    override fun performAction(id: String, item: MenuItem): Boolean {
+        val selectedText = mSelection?.text ?: return super.performAction(id, item)
+
+        return customDelegate.performAction(id, selectedText) || super.performAction(id, item)
+    }
+}
--- a/components/browser/engine-gecko-beta/src/test/java/mozilla/components/browser/engine/gecko/selection/GeckoSelectionActionDelegateTest.kt
+++ b/components/browser/engine-gecko-beta/src/test/java/mozilla/components/browser/engine/gecko/selection/GeckoSelectionActionDelegateTest.kt
@@ -1,0 +1,65 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.engine.gecko.selection
+
+import android.app.Activity
+import android.app.Application
+import android.app.Service
+import mozilla.components.concept.engine.selection.SelectionActionDelegate
+import mozilla.components.support.test.mock
+import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertNull
+import org.junit.Assert.assertTrue
+import org.junit.Test
+
+class GeckoSelectionActionDelegateTest {
+
+    @Test
+    fun `maybe create with non-activity context should return null`() {
+        val customDelegate = mock<SelectionActionDelegate>()
+
+        assertNull(GeckoSelectionActionDelegate.maybeCreate(mock<Application>(), customDelegate))
+        assertNull(GeckoSelectionActionDelegate.maybeCreate(mock<Service>(), customDelegate))
+    }
+
+    @Test
+    fun `maybe create with null delegate context should return null`() {
+        assertNull(GeckoSelectionActionDelegate.maybeCreate(mock<Activity>(), null))
+    }
+
+    @Test
+    fun `maybe create with expected inputs should return non-null`() {
+        assertNotNull(GeckoSelectionActionDelegate.maybeCreate(mock<Activity>(), mock()))
+    }
+
+    @Test
+    fun `getAllActions should contain all actions from the custom delegate`() {
+        val customActions = arrayOf("1", "2", "3")
+        val customDelegate = object : SelectionActionDelegate {
+            override fun getAllActions(): Array<String> = customActions
+            override fun isActionAvailable(id: String): Boolean = false
+            override fun getActionTitle(id: String): CharSequence? = ""
+            override fun performAction(id: String, selectedText: String): Boolean = false
+        }
+
+        val geckoDelegate = TestGeckoSelectionActionDelegate(mock(), customDelegate)
+
+        val actualActions = geckoDelegate.allActions
+
+        customActions.forEach {
+            assertTrue(actualActions.contains(it))
+        }
+    }
+}
+
+/**
+ * Test object that overrides visibility for [getAllActions]
+ */
+class TestGeckoSelectionActionDelegate(
+    activity: Activity,
+    customDelegate: SelectionActionDelegate
+) : GeckoSelectionActionDelegate(activity, customDelegate) {
+    public override fun getAllActions() = super.getAllActions()
+}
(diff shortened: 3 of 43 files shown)


Baron-Severin:
Just noticed there are some detekt issues on these changes.  I'll fix them up tomorrow.

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineView.kt
```diff
@@ -24,7 +26,8 @@ import java.lang.IllegalStateException
 class GeckoEngineView @JvmOverloads constructor(
     context: Context,
     attrs: AttributeSet? = null,
-    defStyleAttr: Int = 0
+    defStyleAttr: Int = 0,
+    private val selectionActionDelegate: SelectionActionDelegate? = null
```
NotWoods: Is there another way for this to be passed to GeckoView? Stringing it through the view constructors feels brittle, especially if we start adding more properties like this in the constructor down the line. Would a setter suffice?

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/selection/GeckoSelectionActionDelegate.kt
```diff
@@ -0,0 +1,59 @@
...
+    private val customDelegate: SelectionActionDelegate
+) : BasicSelectionActionDelegate(activity) {
+
+    companion object {
+        /**
+         * @returns a [GeckoSelectionActionDelegate] if [customDelegate] is non-null and [context]
+         * is an instance of [Activity]. Otherwise, returns null.
+         */
+        fun maybeCreate(context: Context, customDelegate: SelectionActionDelegate?): GeckoSelectionActionDelegate? {
+            return if (context !is Activity || customDelegate == null) {
```
NotWoods: Personally I think this is easier to read with the truthy statement first
```suggestion
            return if (context is Activity && customDelegate != null) {
```

components/concept/engine/src/main/java/mozilla/components/concept/engine/search/SearchRequest.kt
```diff
@@ -0,0 +1,15 @@
...
+package mozilla.components.concept.engine.search
+
+/**
+ * Value type that represents a request for showing a search to the user.
+ */
+data class SearchRequest(val type: Type, val query: String) {
+    /**
+     * Whether the search should be made in a private or normal session.
+     */
+    enum class Type { NORMAL, PRIVATE }
```
NotWoods: We've largely just used a boolean to represent private mode in Android Components. Could that be used instead?

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineView.kt
Baron-Severin: Hm, I'm not sure I follow.  I'm definitely unhappy about adding a dependency to `GeckoEngineView`, but imo adding a setter carries the same drawbacks as a constructor param in addition to adding an unnecessary lifecycle (pre vs post setter call).  I feel like it obfuscates the problem, rather than fixing it.  I'm open to alternatives here, but strongly against moving this to a setter.

components/concept/engine/src/main/java/mozilla/components/concept/engine/search/SearchRequest.kt
Baron-Severin: I prefer enums here, but I'm happy to follow the local convention.  Changing now.

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineView.kt
Baron-Severin: Ekager showed me a way we might be able to get this to the `GeckoSession` directly, so I'm going to try to get that working.

Baron-Severin:
@boek I wrote down what I was trying to change [here](https://github.com/mozilla-mobile/android-components/pull/5669/files#diff-6f5d2f74131ca43dcd5bd4ee1709ebccR111).  Please ping me if anything is unclear, I'll respond to direct Slack messages while on PTO.

Baron-Severin:
@boek and I spoke with @csadilek about this one earlier today and he asked to be tagged for review.

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/selection/GeckoSelectionActionDelegate.kt
```diff
@@ -0,0 +1,61 @@
...
+import android.app.Activity
+import android.content.Context
+import android.view.MenuItem
+import mozilla.components.concept.engine.selection.SelectionActionDelegate
+import org.mozilla.geckoview.BasicSelectionActionDelegate
+
+/**
+ * An adapter between the GV [BasicSelectionActionDelegate] and a generic [SelectionActionDelegate].
+ *
+ * @param customDelegate handles as much of this logic as possible.
```
csadilek: nit `@property`, the `@param` is the activity which needs KDocs.

components/browser/engine-gecko-beta/src/test/java/mozilla/components/browser/engine/gecko/selection/GeckoSelectionActionDelegateTest.kt
```diff
@@ -0,0 +1,31 @@
...
+import android.app.Activity
+import android.app.Application
+import android.app.Service
+import mozilla.components.concept.engine.selection.SelectionActionDelegate
+import mozilla.components.support.test.mock
+import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertNull
+import org.junit.Test
+
+class GeckoSelectionActionDelegateTest {
```
csadilek: We should add some tests here e.g. to make sure `getAllActions` contains the actions of the custom delegate.

components/browser/engine-gecko-nightly/src/test/java/mozilla/components/browser/engine/gecko/selection/GeckoSelectionActionDelegateTest.kt
```diff
@@ -0,0 +1,31 @@
...
+import android.app.Activity
+import android.app.Application
+import android.app.Service
+import mozilla.components.concept.engine.selection.SelectionActionDelegate
+import mozilla.components.support.test.mock
+import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertNull
+import org.junit.Test
+
+class GeckoSelectionActionDelegateTest {
```
csadilek: nit: improve test coverage, see above.

components/browser/state/src/main/java/mozilla/components/browser/state/store/BrowserStore.kt
```diff
@@ -4,24 +4,29 @@
...
 
 /**
  * The [BrowserStore] holds the [BrowserState] (state tree).
  *
  * The only way to change the [BrowserState] inside [BrowserStore] is to dispatch an [Action] on it.
  */
 class BrowserStore(
-    initialState: BrowserState = BrowserState()
+    initialState: BrowserState = BrowserState(),
+    dispatcher: CoroutineDispatcher = Executors.newSingleThreadExecutor().asCoroutineDispatcher()
```
csadilek: We shouldn't need to expose the `dispatcher` here. Is this for testing purposes?

components/concept/engine/src/main/java/mozilla/components/concept/engine/search/SearchAdapter.kt
```diff
@@ -0,0 +1,21 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.concept.engine.search
+
+/**
+ * May be implemented by client code in order to allow a component to start searches.
+ */
+interface SearchAdapter {
```
csadilek: I am not convinced that this should be in `concept-engine`. It's not really related to the browser engine. Can this just live in feature-search?

components/concept/engine/src/main/java/mozilla/components/concept/engine/search/SearchAdapter.kt
```diff
@@ -0,0 +1,21 @@
...
+
+    /**
+     * Called by the component to indicate that the user should be shown a search.
+     */
+    fun sendSearch(isPrivate: Boolean, text: String)
+
+    /**
+     * Called by the component to check whether or not the currently selected session is private.
+     */
+    fun isPrivate(): Boolean
```
csadilek: nit: maybe `isPrivateSession()`...

components/concept/engine/src/main/java/mozilla/components/concept/engine/selection/SelectionActionDelegate.kt
```diff
@@ -0,0 +1,30 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.concept.engine.selection
+
+/**
+ * Generic delegate for handling the context menu that is shown when text is selected.
+ */
+interface SelectionActionDelegate {
+    /**
```
csadilek: nit: Would prefer a first line explaining what the method does, and then describe what it returns.

components/concept/engine/src/main/java/mozilla/components/concept/engine/selection/SelectionActionDelegate.kt
```diff
@@ -0,0 +1,30 @@
...
+     * @returns String IDs for each action that could possibly be shown in the context menu. This
+     * array must include all actions, available or not, and must not change over the class lifetime.
+     */
+    fun getAllActions(): Array<String>
```
csadilek: nit: newline.

components/feature/search/src/main/java/mozilla/components/feature/search/SearchFeature.kt
```diff
@@ -0,0 +1,57 @@
...
+ * NOTE: that this only consumes SearchRequests, and will not hook up the [store] to a source of
+ * SearchRequests. See [SelectionActionDelegate] for use in generating SearchRequests.
+ */
+class SearchFeature(
+    private val store: BrowserStore,
+    private val performSearch: (SearchRequest) -> Unit
+) : LifecycleAwareFeature {
+    private var scope: CoroutineScope? = null
+
+    @FlowPreview
```
csadilek: We usually do this instead: https://github.com/mozilla-mobile/android-components/blob/master/components/support/webextensions/build.gradle#L29

components/lib/state/src/main/java/mozilla/components/lib/state/Store.kt
```diff
@@ -39,9 +40,9 @@ typealias Observer<S> = (S) -> Unit
  */
 open class Store<S : State, A : Action>(
     initialState: S,
-    private val reducer: Reducer<S, A>
+    private val reducer: Reducer<S, A>,
+    dispatcher: CoroutineDispatcher = Executors.newSingleThreadExecutor().asCoroutineDispatcher()
```
csadilek: We shouldn't need expose our dispatcher. I am a bit worried what client-code could do with this :). It also shouldn't be necessary for tests.

components/feature/search/src/test/java/mozilla/components/feature/search/SearchFeatureTest.kt
```diff
@@ -0,0 +1,114 @@
...
+    private lateinit var performSearch: (SearchRequest) -> Unit
+    private lateinit var store: BrowserStore
+    private lateinit var searchFeature: SearchFeature
+
+    @Before
+    fun before() {
+        Dispatchers.setMain(testDispatcher)
+        store = BrowserStore(
+            mockBrowserState(),
+            dispatcher = TestCoroutineDispatcher()
```
csadilek: You shouldn't need this here? (see comments above) :)

You're calling `joinBlocking` on the dispatching job anyway. Here's an example of another test using our `MainCoroutineRule`: https://github.com/mozilla-mobile/android-components/blob/master/components/support/webextensions/src/test/java/mozilla/components/support/webextensions/WebExtensionPopupFeatureTest.kt

components/browser/state/src/main/java/mozilla/components/browser/state/store/BrowserStore.kt
Baron-Severin: It is, there might be a better solution I'm missing though.  `SearchFeature` responds to `UpdateSearchRequestAction` then dispatches a `ConsumeSearchRequestAction` in its collect block.  AFAICT, using `joinBlocking` [here](https://github.com/mozilla-mobile/android-components/pull/5669/files#diff-3bde79026c2bd8586e7ab86d45f42a39R102-R113) doesn't cause the consume action to block, leading to a flaky test.  It's possible I'm misunderstanding something though?

components/feature/search/src/main/java/mozilla/components/feature/search/SearchFeature.kt
Baron-Severin: Oh neat!

components/browser/state/src/main/java/mozilla/components/browser/state/store/BrowserStore.kt
Baron-Severin: I'll try your suggestion in https://github.com/mozilla-mobile/android-components/pull/5669#discussion_r378431112

components/concept/engine/src/main/java/mozilla/components/concept/engine/search/SearchAdapter.kt
Baron-Severin: It's used by the `DefaultSelectionActionDelegate` in feature-contextmenu and `BrowserStoreSearchAdapter` in feature-search.  It's possible that there's a more recent ancestor of the two though, I'm not super familiar with the component structure.

components/browser/state/src/main/java/mozilla/components/browser/state/store/BrowserStore.kt
csadilek: @Baron-Severin ah we have an extension function you can use in tests for this (to make sure actions have been processed): https://github.com/mozilla-mobile/android-components/blob/f7d2a4648162bebc98f852e2f2f9d6bd8b6f0636/components/feature/prompts/src/test/java/mozilla/components/feature/prompts/PromptFeatureTest.kt#L218

components/concept/engine/src/main/java/mozilla/components/concept/engine/search/SearchAdapter.kt
csadilek: @Baron-Severin yeah there's no common ancestor, but `feature-contextmenu` can depend on `feature-search`. It already depends on `feature-tabs` and `feature-app-links`

Baron-Severin: That makes a lot of sense.  Moving it.

Baron-Severin:
> Overall this looks good to me. I share the same concern raised earlier about having `selectionActionDelegate` in the constructor:
> 
>     * It makes it awkward/impossible to use when `GeckoEngineView` is dropped into an XML layout (you'll always need your own inflate implementation to use it).
> 
>     * It has a ripple effect on `Engine.createView()` which now also needs this parameter. This is fine if you call this yourself, but other component code will just not be able to call it with a delegate because they do not know which delegate you want to use.
> 
>     * Together with `@JvmOverloads` it will just create a lot of more permutations of constructors that are not really needed.

I still don't fully agree.  It's an optional parameter, and if it's missing the class will function as it did before this PR.  That said, I'm happy to change as the consensus is against me here.  Updating the code now.

Baron-Severin:
bors r=csadilek

csadilek:
@NotWoods just removed you to try and make bors happy. We'll wait until tomorrow in case you want to review as well.

And @pocmo if you want to take another look before landing. :)

NotWoods:
I don't have any ideas for alternatives to the constructor parameter, so feel free to remove my review here.

csadilek:
> I don't have any ideas for alternatives to the constructor parameter, so feel free to remove my review here.

@NotWoods 👍 , this was already refactored to use a setter instead :).

csadilek:
@Baron-Severin There's a lint error, an unused import in `TabCollectionStorageTest`.

@NotWoods if you're around and could approve, it would be great, then I don't have to force merge. bors insists on your review :).

PR #5679, opened 2020-01-22
Closes #5677: Catch all known non-fatal push errors

### LINKED ISSUE #5677: CryptoError: Error(
https://sentry.prod.mozaws.net/operations/fenix-fennec-nightly/issues/7131337/

```
CryptoError: Error(

Crypto error: Error(

Crypto error: Decryption error))
    at mozilla.appservices.push.RustError.intoException(RustError.kt:14)
    at mozilla.appservices.push.PushManager.decrypt(PushManager.kt:16)
    at mozilla.components.feature.push.RustPushConnection.decrypt(Connection.kt:3)
    at mozilla.components.feature.push.AutoPushFeature$onMessageReceived$1.invokeSuspend(AutoPushFeature.kt:15)
    at mozilla.components.feature.push.AutoPushFeature$onMessageReceived$1.invoke
...
(6 additional frame(s) were not displayed)

Error(

Crypto error: Error(

Crypto error: Decryption error))
```

jonalmeida:
Previously, we wanted to throw on all unknown push errors so that we
were notified on them. Since this seems to be more common than
originally expected, we should just catch them and in a future version,
we should log them without crashing.

All of these push errors can be considered recoverable except
for InternalPanic.

---

Spoke offline with @jrconlin, and it's safe to not crash on these errors.

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
+++ b/components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
@@ -15,10 +15,17 @@ import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.asCoroutineDispatcher
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.plus
+import mozilla.appservices.push.AlreadyRegisteredError
 import mozilla.appservices.push.CommunicationError
 import mozilla.appservices.push.CommunicationServerError
+import mozilla.appservices.push.CryptoError
+import mozilla.appservices.push.MissingRegistrationTokenError
 import mozilla.appservices.push.RecordNotFoundError
+import mozilla.appservices.push.StorageError
+import mozilla.appservices.push.StorageSqlError
 import mozilla.appservices.push.SubscriptionResponse
+import mozilla.appservices.push.TranscodingError
+import mozilla.appservices.push.UrlParseError
 import mozilla.components.concept.push.Bus
 import mozilla.components.concept.push.EncryptedPushMessage
 import mozilla.components.concept.push.PushError
@@ -182,17 +189,9 @@ class AutoPushFeature(
     }
 
     override fun onError(error: PushError) {
-        logger.error("${error.javaClass.simpleName} error: ${error.desc}")
+        logger.error("${error.javaClass.simpleName} error: ${error.message}")
 
-        // Submit via our configured CrashReporter, as well.
-        when (error) {
-            is PushError.Rust -> {
-                crashReporter?.submitCaughtException(error.cause)
-            }
-            else -> {
-                crashReporter?.submitCaughtException(GenericPushError(error.desc))
-            }
-        }
+        crashReporter?.submitCaughtException(error)
     }
 
     /**
@@ -319,7 +318,7 @@ class AutoPushFeature(
 
     private fun CoroutineScope.launchAndTry(block: suspend CoroutineScope.() -> Unit): Job {
         return launchAndTry(block, { e ->
-            onError(PushError.Rust(e))
+            onError(PushError.Rust(e, e.message.orEmpty()))
         })
     }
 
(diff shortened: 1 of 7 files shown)


components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
```diff
@@ -359,7 +361,21 @@ internal fun CoroutineScope.launchAndTry(
...
+                is CommunicationError,
+                is CommunicationServerError,
+                is AlreadyRegisteredError,
+                is StorageError,
+                is MissingRegistrationTokenError,
+                is StorageSqlError,
+                is TranscodingError,
+                is RecordNotFoundError,
+                is UrlParseError -> false
+                else -> true
```
grigoryk: So, new exception types will throw? Not sure that's worth creating us extra work in the future (to keep this list up-to-date). Either way, these exceptions will end-up in Sentry - either as `info`s, or as `fatal`.

There's an idea in building robust software systems - be lenient with what you accept, and strict with what you send out. I think it applies here. We're handling an external message, subject to all sorts of unexpected issues outside of our control. If we crash here in response to a bad message, we're not gaining much! So, maybe our push stack got into a pickle and doesn't function correctly (either due to a server issue, or a client issue, or some network corruption, or..) - what do we gain by crashing here? The rest of the app is, most likely, perfectly functional. You'll just be getting in a user's way.

As long as we're aware of these problems, we can hopefully fix them - and that's what `submitCaughtException` already buys you. Submit these would-be crashes as `info`s, and let the app live :)

jonalmeida: From [the original discussion](https://github.com/mozilla-mobile/android-components/pull/4936#discussion_r341782953), we wanted to crash on newly added exceptions but we had to list them all out, so we opted for a whitelist instead. Now that we know that we don't need to crash on fatal errors, we're just catching and submitting them and we're at least aware of changes to the system this way.

Are you suggesting we don't crash at all? The exceptions that would go through here would be non-push specific (besides InternalPanic) since we're only logging them with `submitCaughtException`.

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
```diff
@@ -184,15 +191,7 @@ class AutoPushFeature(
...
-        // Submit via our configured CrashReporter, as well.
-        when (error) {
-            is PushError.Rust -> {
-                crashReporter?.submitCaughtException(error.cause)
-            }
-            else -> {
-                crashReporter?.submitCaughtException(GenericPushError(error.desc))
-            }
-        }
+        crashReporter?.submitCaughtException(error)
```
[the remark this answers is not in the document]
jonalmeida: These questions were answered offline. I've overridden the `message` for extra certainty and for the RustError we override the `throwable` as well.

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
grigoryk: Let's differentiate between error handling for `onMessageReceived` and for other types of exceptions (one way is to add `onMessageReceivedError` to AutoPushFeature in addition to `onError` - but see last paragraph).

And not crash at all (but rather, log and submit via crash reporter) in case of those types of errors. The other errors we can be more crashy with, I think.

Underlying idea is that it doesn't feel right to me to crash in case of a failure of a fairly isolated system due to an external event.

Also, you'll probably notice that there's currently a bit of a circle dance around error handling - e.g. AbstractFirebasePushService.onMessageReceived -> AutopushFeature.onMessageReceived -> throws -> AbstractFirebasePushService catches -> AutopushFeature.onError.

It seems like this could be simplified. e.g. AutopushFeature.onMessageReceived can just deal with exceptions directly, and if it throws, consumers should re-throw.

jonalmeida:
> LGTM, our conversation about never throwing in `onMessageReceived` notwithstanding :) I'm fine with either direction, but would prefer to never see `fatal` crashes coming out of that method.

I've filed https://github.com/mozilla-mobile/android-components/issues/5691 as follow-up.

jonalmeida:
bors r=grigoryk

jonalmeida:
bors status

PR #5681, opened 2020-01-22
Closes #5661: Caching AppLinksUseCases results

### LINKED ISSUE #5661: Caching AppLinksUseCases results to avoid redundant work
Currently whenever `BrowserMenuItem`is getting highlight effect it narrows down the list by checking if the item is visible.  This causes some expensive operations for `AppLinksUseCases` trying to determining if it should be visible.  Avoid doing this to speed up performance

rocketsroger:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/app-links/src/main/java/mozilla/components/feature/app/links/AppLinksUseCases.kt
+++ b/components/feature/app-links/src/main/java/mozilla/components/feature/app/links/AppLinksUseCases.kt
@@ -17,6 +17,9 @@ import java.util.UUID
 private const val EXTRA_BROWSER_FALLBACK_URL = "browser_fallback_url"
 private const val MARKET_INTENT_URI_PACKAGE_PREFIX = "market://details?id="
 
+@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+internal const val APP_LINKS_CACHE_INTERVAL = 30 * 1000L // 30 seconds
+
 /**
  * These use cases allow for the detection of, and opening of links that other apps have registered
  * an [IntentFilter]s to open.
@@ -87,6 +90,15 @@ class AppLinksUseCases(
         private val includeInstallAppFallback: Boolean = false
     ) {
         operator fun invoke(url: String): AppLinkRedirect {
+            val urlHash = (url + includeHttpAppLinks + ignoreDefaultBrowser + includeHttpAppLinks).hashCode()
+            val currentTimeStamp = System.currentTimeMillis()
+            // since redirectCache is mutable, get the latest
+            val cache = redirectCache
+            if (cache != null && urlHash == cache.cachedUrlHash &&
+                    currentTimeStamp <= cache.cacheTimeStamp + APP_LINKS_CACHE_INTERVAL) {
+                return cache.cachedAppLinkRedirect
+            }
+
             val redirectData = createBrowsableIntents(url)
             val isAppIntentHttpOrHttps = redirectData.appIntent?.data?.isHttpOrHttps ?: false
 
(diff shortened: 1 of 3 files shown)


components/feature/app-links/src/main/java/mozilla/components/feature/app/links/AppLinksUseCases.kt
```diff
@@ -17,6 +17,9 @@ import java.util.UUID
 private const val EXTRA_BROWSER_FALLBACK_URL = "browser_fallback_url"
 private const val MARKET_INTENT_URI_PACKAGE_PREFIX = "market://details?id="
 
+@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+internal const val APP_LINKS_CACHE_INTERVAL = 30 * 1000L // 30 seconds
```
rocketsroger: No data backing up this choice for 30 seconds.  This could be shorter (5 seconds could work as well) since this is to avoid the performance hits that occurs when `BrowserMenu` checks for visibility of the menu item.  Any thoughts?

components/feature/app-links/src/main/java/mozilla/components/feature/app/links/AppLinksUseCases.kt
```diff
@@ -218,4 +230,16 @@ class AppLinksUseCases(
...
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal data class AppLinkRedirectCache(
+        var cacheTimeStamp: Long = 0,
+        var cachedUrlHash: Int = 0,
+        var cachedAppLinkRedirect: AppLinkRedirect = AppLinkRedirect(null, null, null)
+    )
+
+    companion object {
+        @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+        internal var redirectCache = AppLinkRedirectCache()
```
NotWoods: Instead of having default parameters for the cache type, just make this field nullable.

rocketsroger: > Instead of having default parameters for the cache type, just make this field nullable.

I was on the fence about this.  Do you mean making `redirectCache` nullable but then when returning, do `return redirectCache?.cachedAppLinkRedirect ?: AppLinkRedirectCache()`

I like how clean it is right now though since we don't have to check what's in the cache when returning.

NotWoods: Since you don't have to return the cache, I think you can just make this change:
```kt
if (urlHash == redirectCache?.cachedUrlHash && ...) {
```

rocketsroger: Oh doh, that makes perfect sense.  I'll update.  Thanks 😅

rocketsroger: Oh I remembered why now.  Since `redirectCache` is mutable.  I can't be sure that after the check it can be set to null again.  So guarding the return inside doesn't work.  I can't change it to a Nullable.

rocketsroger: Updated.  Thanks for the suggestion.

components/feature/app-links/src/main/java/mozilla/components/feature/app/links/AppLinksUseCases.kt
```diff
@@ -218,4 +232,16 @@ class AppLinksUseCases(
         val marketplaceIntent: Intent? = null,
         val resolveInfo: ResolveInfo? = null
     )
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal data class AppLinkRedirectCache(
+        var cacheTimeStamp: Long = 0,
```
NotWoods: nit: Remove the default arguments. I don't think it makes sense to have defaults for a cache now that we've made the property nullable.
```suggestion
        var cacheTimeStamp: Long,
        var cachedUrlHash: Int,
        var cachedAppLinkRedirect: AppLinkRedirect
```

rocketsroger: Agreed.  I'll remove.  Thanks,

rocketsroger:
bors r=NotWoods

rocketsroger:
bors retry

PR #5705, opened 2020-01-23
Clean up web extension action menu popup 

psymoon:
- Remove onActionPopupToggleOverride and add openPopupInTab flag.
- Add store flow to observe for `extension.popUpSession` change when it's not null in sample browser.
- Fix web extension menu item label view being clickable; only the container should be.

This cleans up and makes it easier for Fenix integration as well as easier to deprecate `WebExtensionSupport` class later down the road when browser store refactor is complete.

### DIFF
--- a/components/browser/menu/src/test/java/mozilla/components/browser/menu/WebExtensionBrowserMenuTest.kt
+++ b/components/browser/menu/src/test/java/mozilla/components/browser/menu/WebExtensionBrowserMenuTest.kt
@@ -59,6 +59,7 @@ class WebExtensionBrowserMenuTest {
             "browser_action" to WebExtensionState(
                 "browser_action",
                 "url",
+                "name",
                 true,
                 browserAction = browserAction,
                 pageAction = pageAction
@@ -94,6 +95,7 @@ class WebExtensionBrowserMenuTest {
             "id" to WebExtensionState(
                 "id",
                 "url",
+                "name",
                 true,
                 browserAction = defaultBrowserAction,
                 pageAction = defaultPageAction
@@ -116,6 +118,7 @@ class WebExtensionBrowserMenuTest {
             "id2" to WebExtensionState(
                 "id2",
                 "url",
+                "name",
                 true,
                 browserAction = anotherBrowserAction,
                 pageAction = anotherPageAction
@@ -146,6 +149,7 @@ class WebExtensionBrowserMenuTest {
             "id" to WebExtensionState(
                 "id",
                 "url",
+                "name",
                 true,
                 browserAction = defaultBrowserAction,
                 pageAction = defaultPageAction
@@ -155,6 +159,7 @@ class WebExtensionBrowserMenuTest {
             "id" to WebExtensionState(
                 "id",
                 "url",
+                "name",
                 true,
                 browserAction = overriddenBrowserAction,
                 pageAction = overriddenPageAction
@@ -198,6 +203,7 @@ class WebExtensionBrowserMenuTest {
             "enabled" to WebExtensionState(
                 "enabled",
                 "url",
+                "name",
                 true,
                 browserAction = enabledBrowserAction,
                 pageAction = enabledPageAction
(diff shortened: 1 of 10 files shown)


samples/browser/src/main/java/org/mozilla/samples/browser/BrowserActivity.kt
```diff
@@ -43,6 +53,32 @@ open class BrowserActivity : AppCompatActivity(), ComponentCallbacks2 {
         }
     }
 
+    @kotlinx.coroutines.ExperimentalCoroutinesApi
+    override fun onStart() {
+        super.onStart()
+        scope = components.store.flowScoped { flow ->
+            flow.ifChanged { it.extensions }
+                .map { it.extensions.filterValues { extension -> extension.popupSession != null } }
+                .collect { pair ->
```
psymoon: TODO rename; this is not a pair...

components/browser/menu/src/test/java/mozilla/components/browser/menu/WebExtensionBrowserMenuTest.kt
```diff
@@ -59,6 +59,7 @@ class WebExtensionBrowserMenuTest {
             "browser_action" to WebExtensionState(
                 "browser_action",
                 "url",
+                "",
```
csadilek: nit: Might as well put a "title" in there :)

samples/browser/src/main/java/org/mozilla/samples/browser/BrowserActivity.kt
csadilek: Yeah, maybe just extensionState?

csadilek: nit: I would (also for Fenix) move this to a private function to make it clear what this does

```
webExtScope = observeWebExtensionPopups()
```

samples/browser/src/main/java/org/mozilla/samples/browser/BrowserActivity.kt
```diff
@@ -43,6 +53,32 @@ open class BrowserActivity : AppCompatActivity(), ComponentCallbacks2 {
...
+                    if (pair.values.isNotEmpty()) {
+                        // There could only be one active popupSession instance at a time
```
csadilek: nit: // We currently limit to one active popup session at a time ?

samples/browser/src/main/java/org/mozilla/samples/browser/BrowserActivity.kt
psymoon: I like it

samples/browser/src/main/java/org/mozilla/samples/browser/BrowserActivity.kt
```diff
@@ -43,6 +53,16 @@ open class BrowserActivity : AppCompatActivity(), ComponentCallbacks2 {
         }
     }
 
+    override fun onStart() {
+        super.onStart()
+        scope = observeWebExtensionPopups()
```
csadilek: nit: would give the scope a name "webExtScope" or similar :)

samples/browser/src/main/java/org/mozilla/samples/browser/BrowserActivity.kt
```diff
@@ -70,4 +90,23 @@ open class BrowserActivity : AppCompatActivity(), ComponentCallbacks2 {
         components.sessionManager.onLowMemory()
         components.icons.onLowMemory()
     }
+
+    private fun observeWebExtensionPopups(): CoroutineScope {
+        return components.store.flowScoped { flow ->
+            flow.ifChanged { it.extensions }
+                .map { it.extensions.filterValues { extension -> extension.popupSession != null } }
```
csadilek: I think here, after the map, we need another empty `.ifChanged()` to make sure that an unrelated change to the global extension state doesn't get collected, otherwise we'd try to display the same popup again.

In other words, we only want to collect if we found a new non-null popupSession.

psymoon:
bors r=csadilek

psymoon:
bors retry

PR #5714, opened 2020-01-24
Closes #5708: Browser/Page actions don't have stable ordering

### LINKED ISSUE #5708: Browser/Page actions don't have stable ordering in toolbar and menu
STR
- Start sample app
- Install multiple web extensions
- Click on first web extension browser action in the toolbar (from the left)
- Transition to another screen (enabling action popup or background/foreground app or transition to addons list fragment)
- Go back to browser screen

Expected
- Browser actions stay the same in the toolbar

Actual
- Browser actions order changed in the toolbar

csadilek:


components/browser/menu/src/main/java/mozilla/components/browser/menu/WebExtensionBrowserMenu.kt
```diff
@@ -70,7 +70,7 @@ class WebExtensionBrowserMenu internal constructor(
         ): List<BrowserMenuItem> {
             val menuItems = ArrayList<BrowserMenuItem>()
             val extensions = state.extensions.values.toList()
-            extensions.filter { it.enabled }
+            extensions.filter { it.enabled }.sortedBy { it.name }
```
psymoon: Let's make sure name being nullable doesn't cause a crash.

csadilek: 👍 done.

csadilek:
bors r=psymoon

Issue #5667, opened 2020-01-21
Know which PWAs are installed


NotWoods:
We want to be able to check if a PWA is installed for a couple different use cases:
- Open notification in corresponding installed PWA
- Display PWA's Web Share Target
- Open corresponding PWA when on a website in the browser

On Android 7.1 and above, we can get a list of pinned shortcuts and use that to figure out which PWAs are installed. On older devices, we could come up with some creative solution or just fallback to different behaviour. (For example, opening notifications in the browser instead.)

TitanNano:
> On Android 7.1 and above, we can get a list of pinned shortcuts and use that to figure out which PWAs are installed. On older devices, we could come up with some creative solution or just fallback to different behaviour. (For example, opening notifications in the browser instead.)

@NotWoods I spent some time pondering about this. As there is no possibility to know the installed shortcuts, there would be only the option to manually track the installation, but I think this has a very high chance to be wrong.

So I came up with the following solution. We introduce a tri-state, `IS_NOT_INSTALLED`, `IS_INSTALLED` and `IS_POSSIBLY_INSTALLED`. The first two work as expected, the third though would actually be more of a suggestion. Once the user adds the shortcut on android <= 7.1 this information is stored somewhere and from now on the state of that PWA scope is `IS_POSSIBLY_INSTALLED`. 
This would enable us to cover the following situations:

- opening a notification:
  - if `IS_NOT_INSTALLED` just open the page in a browser tab
  - if `IS_INSTALLED` open the page in the PWA
  - if `IS_POSSIBLY_INSTALLED` tapping the notification still opens a browser tab, but the notification also has an action "open in PWA"
- display web share target:
  - if `IS_NOT_INSTALLED` no target is shown
  - if `IS_INSTALLED` show target
  - if `IS_POSSIBLY_INSTALLED` we could show a target, but it would never go away anymore
- open PWA when entering it's scope while navigating in the browser
  - if `IS_NOT_INSTALLED` nothing
  - if `IS_INSTALLED` launch PWA in standalone mode (like the Fenix setting "Open links in apps")
  - if `IS_POSSIBLY_INSTALLED` do nothing, but provide a button or menu entry to launch PWA

[...]

PR #5724, opened 2020-01-26
Know which PWAs are installed #5667

TitanNano:
For #5667

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/pwa/src/main/java/mozilla/components/feature/pwa/ManifestStorage.kt
+++ b/components/feature/pwa/src/main/java/mozilla/components/feature/pwa/ManifestStorage.kt
@@ -15,8 +15,12 @@ import mozilla.components.feature.pwa.db.ManifestEntity
 
 /**
  * Disk storage for [WebAppManifest]. Other components use this class to reload a saved manifest.
+ *
+ * @param context the application context this storage is associated with
+ * @param usedAtTimeout a timeout in milliseconds after which the storage will consider a manifest
+ *                      as unused. By default this is 30 days.
  */
-class ManifestStorage(context: Context) {
+class ManifestStorage(context: Context, private val usedAtTimeout: Long = DEFAULT_TIMEOUT) {
 
     @VisibleForTesting
     internal var manifestDao = lazy { ManifestDatabase.get(context).manifestDao() }
@@ -29,6 +33,26 @@ class ManifestStorage(context: Context) {
      */
     suspend fun loadManifest(startUrl: String): WebAppManifest? = withContext(Dispatchers.IO) {
         manifestDao.value.getManifest(startUrl)?.manifest
+    }
+
+    /**
+     * Load all Web App Manifests with a matching scope for the given URL from disk.
+     * If no manifest is found, an empty list is returned.
+     *
+     * @param url URL of site. Should correspond to an url covered by the scope of a stored manifest.
+     */
+    suspend fun loadManifestsByScope(url: String): List<WebAppManifest> = withContext(Dispatchers.IO) {
+        manifestDao.value.getManifestsByScope(url).map { it -> it.manifest }
+    }
+
+    /**
+     * Checks whether there is a currently used manifest with a scope that matches the url.
+     *
+     * @param url the url to match with manifest scopes.
+     * @param currentTime the current time in milliseconds.
+     */
+    suspend fun hasRecentManifest(url: String, currentTime: Long): Boolean = withContext(Dispatchers.IO) {
+        manifestDao.value.hasRecentManifest(url, deadline = currentTime - usedAtTimeout) > 0
     }
 
     /**
@@ -54,9 +78,26 @@ class ManifestStorage(context: Context) {
     }
 
     /**
+     * Update the last time a web app was used.
+     *
+     * @param manifest the manifest to update
+     */
+    suspend fun updateManifestUsedAt(manifest: WebAppManifest) = withContext(IO) {
+        manifestDao.value.getManifest(manifest.startUrl)?.let { existing ->
+            val update = existing.copy(usedAt = System.currentTimeMillis())
+            manifestDao.value.updateManifest(update)
+        }
+    }
+
+    /**
      * Delete all manifests associated with the list of URLs.
      */
     suspend fun removeManifests(startUrls: List<String>) = withContext(IO) {
         manifestDao.value.deleteManifests(startUrls)
+    }
+
+    companion object {
+        const val MILLISECONDS_IN_A_DAY: Long = 86400000
+        const val DEFAULT_TIMEOUT = MILLISECONDS_IN_A_DAY * 30
     }
 }
--- a/components/feature/pwa/src/main/java/mozilla/components/feature/pwa/db/ManifestDao.kt
+++ b/components/feature/pwa/src/main/java/mozilla/components/feature/pwa/db/ManifestDao.kt
@@ -21,6 +21,14 @@ internal interface ManifestDao {
     fun getManifest(startUrl: String): ManifestEntity?
 
     @WorkerThread
+    @Query("SELECT * from manifests WHERE :url LIKE scope||'%' ORDER BY LENGTH(scope) DESC")
+    fun getManifestsByScope(url: String): List<ManifestEntity>
+
+    @WorkerThread
+    @Query("SELECT count(start_url) from manifests WHERE :url LIKE scope||'%' AND used_at < :deadline")
+    fun hasRecentManifest(url: String, deadline: Long): Int
+
+    @WorkerThread
     @Insert(onConflict = OnConflictStrategy.REPLACE)
     fun insertManifest(manifest: ManifestEntity): Long
 
--- a/components/feature/pwa/src/main/java/mozilla/components/feature/pwa/feature/ManifestUpdateFeature.kt
+++ b/components/feature/pwa/src/main/java/mozilla/components/feature/pwa/feature/ManifestUpdateFeature.kt
@@ -37,6 +37,7 @@ class ManifestUpdateFeature(
 
     private var scope: CoroutineScope? = null
     private var updateJob: Job? = null
+    private var updateUsageJob: Job? = null
 
     /**
      * When the manifest is changed, compare it to the existing manifest.
@@ -46,6 +47,8 @@ class ManifestUpdateFeature(
     override fun onWebAppManifestChanged(session: Session, manifest: WebAppManifest?) {
         if (manifest?.startUrl == initialManifest.startUrl && manifest != initialManifest) {
             updateJob?.cancel()
+            updateUsageJob?.cancel()
+
             updateJob = scope?.launch { updateStoredManifest(manifest) }
         }
     }
@@ -63,6 +66,11 @@ class ManifestUpdateFeature(
     override fun start() {
         scope = MainScope()
         sessionManager.findSessionById(sessionId)?.register(this)
+        updateUsageJob?.cancel()
+
+        updateUsageJob = scope?.launch {
+            storage.updateManifestUsedAt(initialManifest)
+        }
     }
 
     override fun stop() {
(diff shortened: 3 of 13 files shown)


components/feature/pwa/src/main/java/mozilla/components/feature/pwa/ManifestStorage.kt
```diff
@@ -16,7 +16,7 @@ import mozilla.components.feature.pwa.db.ManifestEntity
 /**
  * Disk storage for [WebAppManifest]. Other components use this class to reload a saved manifest.
  */
-class ManifestStorage(context: Context) {
+class ManifestStorage(context: Context, private val usedAtTimeout: Long = 2592000000) {
```
NotWoods: Let's add some constant like `MILLISECONDS_IN_A_DAY` then set this to be `MILLISECONDS_IN_A_DAY * 30`, so its clear what this number represents.

It would also be good to have some doc comments that indicate this is in milliseconds and what its used for.
```
@param usedAtTimeout ...
```

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/ManifestStorage.kt
```diff
@@ -25,10 +25,24 @@ class ManifestStorage(context: Context) {
...
      * If no manifest is found, null is returned.
      *
-     * @param startUrl URL of site. Should correspond to manifest's start_url.
+     * @param url URL of site. Should correspond to manifest's start_url.
      */
-    suspend fun loadManifest(startUrl: String): WebAppManifest? = withContext(Dispatchers.IO) {
-        manifestDao.value.getManifest(startUrl)?.manifest
+    suspend fun loadManifest(
+        url: String,
+        allowTimedout: Boolean = false
```
NotWoods: nit: `includeTimedOut`

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/ManifestStorage.kt
```diff
@@ -25,10 +25,24 @@ class ManifestStorage(context: Context) {
      * Load a Web App Manifest for the given URL from disk.
      * If no manifest is found, null is returned.
      *
-     * @param startUrl URL of site. Should correspond to manifest's start_url.
+     * @param url URL of site. Should correspond to manifest's start_url.
```
NotWoods: nit: Add a `@param` comment for the boolean

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/WebAppLauncherActivity.kt
```diff
@@ -92,7 +92,7 @@ class WebAppLauncherActivity : AppCompatActivity() {
 
     @VisibleForTesting(otherwise = PRIVATE)
     internal suspend fun loadManifest(startUrl: String): WebAppManifest? {
-        return storage.loadManifest(startUrl)
+        return storage.loadManifest(startUrl, true)
```
NotWoods: nit: Used named arguments for the boolean

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/ManifestStorage.kt
```diff
@@ -25,10 +25,24 @@ class ManifestStorage(context: Context) {
...
      */
-    suspend fun loadManifest(startUrl: String): WebAppManifest? = withContext(Dispatchers.IO) {
-        manifestDao.value.getManifest(startUrl)?.manifest
+    suspend fun loadManifest(
+        url: String,
+        allowTimedout: Boolean = false
+    ): WebAppManifest? = withContext(Dispatchers.IO) {
+        val entry = manifestDao.value.getManifest(url)
+
+        if (entry == null) {
```
NotWoods: You can simplify this with the elvis operator
```kt
val entry = manifestDao.value.getManifest(url) ?: return@withContext null
```
Does return without the `@withContext` label work?

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/ManifestStorage.kt
```diff
@@ -25,10 +25,24 @@ class ManifestStorage(context: Context) {
...
+            return@withContext null
+        }
+
+        // ignore entry if it hasn't been used for more than 30 days
+        if (!allowTimedout && (System.currentTimeMillis() - entry.usedAt) > usedAtTimeout) {
```
NotWoods: Dealing with current time will be a bit of a pain in tests. Maybe we can add the current time as a function parameter.

```kt
suspend fun loadManifest(url: String, includeTimedOut: Boolean = false, currentTime: Long = System.currentTimeMillis())
```

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/db/ManifestDao.kt
```diff
@@ -17,8 +17,8 @@ import androidx.room.Update
 @Dao
 internal interface ManifestDao {
     @WorkerThread
-    @Query("SELECT * from manifests WHERE start_url = :startUrl")
-    fun getManifest(startUrl: String): ManifestEntity?
+    @Query("SELECT * from manifests WHERE :url LIKE scope||'%' ORDER BY LENGTH(scope) DESC LIMIT 1")
+    fun getManifest(url: String): ManifestEntity?
```
NotWoods: Let's make a separate query for getting the manifest using an arbitrary URL. That way we can continue to use the primary key when its known, such as in `updateManifest`. We should also make `scope` an index so its faster to query.

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/feature/ManifestUpdateFeature.kt
```diff
@@ -63,6 +66,24 @@ class ManifestUpdateFeature(
...
+    /**
+     * Updates manifest parameters like last time of usage.
+     */
+    @OnLifecycleEvent(Lifecycle.Event.ON_RESUME)
+    fun resume() {
+        scope = MainScope()
+        updadeUsageJob?.cancel()
+
+        updadeUsageJob = scope?.launch {
+            storage.updateManifestUsedAt(initialManifest.startUrl)
```
NotWoods: Since its a one-month timescale, updating only when the PWA is started should be fine. We don't need to update on every pause and resume.

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/feature/ManifestUpdateFeature.kt
```diff
@@ -63,6 +66,24 @@ class ManifestUpdateFeature(
     override fun start() {
         scope = MainScope()
         sessionManager.findSessionById(sessionId)?.register(this)
+        updadeUsageJob?.cancel()
```
NotWoods: This job should also be cancelled when the manifest changes.

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/db/ManifestDao.kt
TitanNano: @NotWoods scope should already be an index https://github.com/mozilla-mobile/android-components/pull/5724/commits/e673751375fc375269a9d8d43b19715753b53b7e#diff-a37ee58885a902172361ba486bd51783R23-R24

NotWoods: Oops, missed that. Sorry!

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/ManifestStorage.kt
TitanNano: @NotWoods I always get `'return' is not allowed here` without the label.

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/ManifestStorage.kt
```diff
@@ -53,10 +71,27 @@ class ManifestStorage(context: Context) {
         }
     }
 
+    /**
+     * Update the last time a web app was used.
+     *
+     * @param url an url that is part of an installed web app
+     */
+    suspend fun updateManifestUsedAt(url: String) = withContext(IO) {
```
NotWoods: I think this can just the `startUrl` instead of the scope, since you'll only update the manifest in a running PWA. It should just take the manifest as a parameter to avoid confusion when calling the function.

TitanNano:
@NotWoods I just noticed that we might get problems if a PWA changes its `start_url`. Is this already handled somewhere? Because from the databases point of view there could be two manifests with the same scope but different `start_url`s.

NotWoods:
The start URL is tied to the corresponding shortcut on the homescreen. If a site changes the start URL, it gets a new shortcut. 

I think having conflicting scopes will be unusual enough that we can just pick one manifest out of the matching set. Conflicting scopes should only happen on sites controlled by the same developer. We can always revisit it if we get reports that it doesn't work properly.

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/ManifestStorage.kt
```diff
@@ -25,10 +29,24 @@ class ManifestStorage(context: Context) {
...
      *
-     * @param startUrl URL of site. Should correspond to manifest's start_url.
+     * @param url URL of site. Should correspond to an url covered by the scope of a stored manifest.
+     * @param includeTimeout Whether the manifest should only be loaded if the last usage
+     *                       has not passed the defined timeout.
+     * @param currentTime timestamp in milliseconds that is used to validate timeout.
      */
-    suspend fun loadManifest(startUrl: String): WebAppManifest? = withContext(Dispatchers.IO) {
-        manifestDao.value.getManifest(startUrl)?.manifest
+    suspend fun loadManifest(
```
NotWoods: Let's leave `loadManifest` for loading based on the `startUrl`. We use this for launching the shortcut and we want to maintain lookup by the primary key.

I think we should have two new functions: `loadManifestByScope`, which includes all manifests under a given scope, and `hasRecentManifest`, which just returns a boolean.

- `loadManifest` corresponds to `getManifest`
- `loadManifestByScope` corresponds to `getManifestByScope`
- `hasRecentManifest` corresponds to a new SQL method. Perhaps we can add a method that just uses `COUNT` instead of returning the manifest, then return if count is at least 1.

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/ManifestStorage.kt
```diff
@@ -31,6 +35,47 @@ class ManifestStorage(context: Context) {
...
+     * If no manifest is found, an empty list is returned.
+     *
+     * @param url URL of site. Should correspond to an url covered by the scope of a stored manifest.
+     * @param includeTimeout Whether the manifest should only be loaded if the last usage
+     *                       has not passed the defined timeout.
+     * @param currentTime timestamp in milliseconds that is used to validate timeout.
+     */
+    suspend fun loadManifestsByScope(
+        url: String,
+        includeTimeout: Boolean = false,
```
NotWoods: I think we can take out the timeout code since its really only needed for install state, which is now done by checking `hasRecentManifest`. We can simplify this function and only take `url` as a parameter.

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/ManifestStorage.kt
```diff
@@ -31,6 +35,47 @@ class ManifestStorage(context: Context) {
...
+        currentTime: Long = System.currentTimeMillis()
+    ): List<WebAppManifest> = withContext(Dispatchers.IO) {
+        val entityList = manifestDao.value.getManifestsByScope(url)
+        val first = entityList.firstOrNull() ?: return@withContext emptyList<WebAppManifest>()
+
+        entityList.filter { it ->
+            if (it.scope != first.scope) {
```
NotWoods: I think we should either return the entire list even if some manifests have a narrower scope, or always return only 1 item instead of the list. The scope check here seems arbitrary to me.

TitanNano: I agree, let's just return the entire list. Since it's always sorted by narrowest scope first, I think it also will be simple to get the best match by just taking the first entry.

NotWoods:
bors r=NotWoods

PR #5734, opened 2020-01-27
Fixes #5725 - Set manifest scope to be not-null

### LINKED ISSUE #5725: Add comments about what Gecko manifest processing is performed
Usually when working with the manifest parser, we assume the JSONObject is identical to whatever the website provides. However, Gecko and GeckoView do some processing first before giving the object to Android Components. Let's write down the adjustments somewhere.

- Changes URLs to be absolute
- Changes some space-separated strings into arrays
- Changes colors to follow Android format (`#AARRGGBB`)
- Removes invalid enum values (ie: `display: halfscreen`)
- Ensures `display`, `dir`, `start_url`, and `scope` always have a value
- Trims most strings (`name`, `short_name`, ...)

NotWoods:
Gecko always provides us with a scope, so it doesn't need to be nullable. I've set it to default to start URL for tests, which is close to what the spec specifies.

Additionally added comments about what changes Gecko performs.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/concept/engine/src/main/java/mozilla/components/concept/engine/manifest/WebAppManifest.kt
+++ b/components/concept/engine/src/main/java/mozilla/components/concept/engine/manifest/WebAppManifest.kt
@@ -49,6 +49,7 @@ import mozilla.components.concept.engine.manifest.WebAppManifest.ExternalApplica
 data class WebAppManifest(
     val name: String,
     val startUrl: String,
+    val scope: String,
     val shortName: String? = null,
     val display: DisplayMode = DisplayMode.BROWSER,
     @ColorInt val backgroundColor: Int? = null,
@@ -57,7 +58,6 @@ data class WebAppManifest(
     val dir: TextDirection = TextDirection.AUTO,
     val lang: String? = null,
     val orientation: Orientation = Orientation.ANY,
-    val scope: String? = null,
     @ColorInt val themeColor: Int? = null,
     val relatedApplications: List<ExternalApplicationResource> = emptyList(),
     val preferRelatedApplications: Boolean = false,
--- a/components/concept/engine/src/main/java/mozilla/components/concept/engine/manifest/WebAppManifestParser.kt
+++ b/components/concept/engine/src/main/java/mozilla/components/concept/engine/manifest/WebAppManifestParser.kt
@@ -44,6 +44,16 @@ class WebAppManifestParser {
     /**
      * Parses the provided JSON and returns a [WebAppManifest] (wrapped in [Result.Success] if parsing was successful.
      * Otherwise [Result.Failure].
+     *
+     * Gecko performs some initial parsing on the Web App Manifest, so the [JSONObject] we work with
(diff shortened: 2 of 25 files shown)


components/concept/engine/src/main/java/mozilla/components/concept/engine/manifest/WebAppManifest.kt
```diff
@@ -57,7 +57,7 @@ data class WebAppManifest(
     val dir: TextDirection = TextDirection.AUTO,
     val lang: String? = null,
     val orientation: Orientation = Orientation.ANY,
-    val scope: String? = null,
+    val scope: String = startUrl,
```
pocmo: If we always get one then should we just not have any default at all?

components/concept/engine/src/main/java/mozilla/components/concept/engine/manifest/WebAppManifestParser.kt
```diff
@@ -44,22 +44,34 @@ class WebAppManifestParser {
...
                 name = name,
                 shortName = shortName,
-                startUrl = json.getString("start_url"),
+                startUrl = startUrl,
                 display = parseDisplayMode(json),
                 backgroundColor = parseColor(json.tryGetString("background_color")),
                 description = json.tryGetString("description"),
                 icons = parseIcons(json),
-                scope = json.tryGetString("scope"),
+                scope = json.tryGetString("scope") ?: startUrl,
```
pocmo: How does this affect manifest that we already have saved to disk? Could we have saved a `null` at any point in the past?

Using the start url as default would mean that we always switch to a custom tab once the user leaves the start url? Is that the expected default? I wonder if we need to be more protective against changes in Gecko(View)? Since the preprocessing is not part of the standard we can't always rely on it?

components/concept/engine/src/main/java/mozilla/components/concept/engine/manifest/WebAppManifest.kt
NotWoods: I left a default to make writing tests a bit easier and avoid breaking changes. We could break and make this a required parameter instead.

components/concept/engine/src/main/java/mozilla/components/concept/engine/manifest/WebAppManifestParser.kt
NotWoods: Yes, the default isn't perfect. I put in a poor default because we should always get a value from storage and Gecko...so I should probably change this to throw if the scope isn't defined :)

It might be worthwhile to schedule a meeting with the Gecko side and decide what to do long term. I think its OK to rely on the preprocessing, especially since it differs the standard in some areas like `sizes`. The standard is being updated to include a concept of a "processed manifest", which is closer to what we're working with here.

pocmo:
There are some test failures here.

pocmo:
@NotWoods Closing this to get our PR count down. Let's revive this once you are back and have time to fix the failures. :)

PR #5752, opened 2020-01-29
Add menu scrollbar display options

psymoon:


psymoon:
Will verify solution with UX

PR #5784, opened 2020-02-03
Closes #5783: Deadlock when dispatching actions during lifecycle change

### LINKED ISSUE #5783: lib-state: Deadlock when dispatching actions during lifecycle change
We've run into an ANR when switching tabs with add-ons installed. The cause of this is a deadlock that can occur when the installed add-on sends a series of browser actions (translated to actions on our browser store) during or overlapping with lifecycle changes. Actions are dispatched (processed) on a dedicated thread and guarded by a lock on the store. The lifecyle changes cause synchronized method invocations on the subscription (on the main thread). Since Subscription.dispatch is guarded by the same subscription lock as these "lifecycle" methods, a deadlock can occur when any of these "lifecycle" method also requires the store lock (e.g. https://github.com/mozilla-mobile/android-components/blob/master/components/lib/state/src/main/java/mozilla/components/lib/state/Store.kt#L142). 

If a lifecycle event occurs and waits on the store lock we have to make sure we can still dispatch messages. Otherwise, the store lock would be never become available.

csadilek:
The fix here is to make sure we can `resume` without a lock on the store. In general, reading the state on the store shouldn't need to be synchronized with actions being processed/dispatched. Reading the last written state is perfectly fine for other components in the system i.e. reading should not block on pending actions being processed.

@pocmo Can you take a look, fix is as discussed :).

### DIFF
--- a/components/lib/state/src/main/java/mozilla/components/lib/state/Store.kt
+++ b/components/lib/state/src/main/java/mozilla/components/lib/state/Store.kt
@@ -56,13 +56,12 @@ open class Store<S : State, A : Action>(
         scope.cancel()
     }
     private val dispatcherWithExceptionHandler = dispatcher + exceptionHandler
-    private var currentState = initialState
+    @Volatile private var currentState = initialState
 
     /**
      * The current [State].
      */
     val state: S
-        @Synchronized
         get() = currentState
 
     /**
@@ -151,6 +150,11 @@ open class Store<S : State, A : Action>(
             active = false
         }
 
+        /**
+         * Notifies this subscription's observer of a state change.
+         *
+         * @param state the updated state.
+         */
         @Synchronized
         internal fun dispatch(state: S) {
             if (active) {


components/lib/state/src/main/java/mozilla/components/lib/state/Store.kt
```diff
@@ -174,6 +167,27 @@ open class Store<S : State, A : Action>(
...
+        // and need the store lock as well (e.g. to read the current state on
+        // resume). Therefore, these subscription "lifecycle" methods
+        // can not be guarded by the same lock as dispatch. If a lifecycle
+        // event occurs and waits on the store lock we have to make sure we
+        // can still dispatch messages. Otherwise, the store lock would be
+        // never become available.
+        internal fun dispatch(state: S) {
+            if (active) {
+                observer.invoke(state)
+            }
```
pocmo: I think we introduced a small possibility of a race condition here. Since we are on two threads, it is in theory possible that we check for active here (`true`) and briefly after that the other thread calls `pause()` before `observer.invoke(state)` gets executed. In that situation we may invoke observer with a paused lifecycle - which before was not possible due to the synchronization? I wonder if that could lead to lifecycle issues in the observers?

Still approving the PR for now. I'll look at this again today - I wonder if there are other options.

pocmo: Looked a bit more into this and was trying to find a solution that does not need a bigger risky refactoring.

* Idea 1: Get rid of the lock on `Store.state`: This definitely only exists for data visibility reasons. We could use `volatile`, synchronize on anything else or just leave it to the caller (but that's a risky change). That looked pretty good until I saw that `Subscription.unsubscribe` goes through the `Store` as well. It is not synchronized on the store instance though. However it synchronizes on `subscriptions` which is also used in the dispatch loop. So I wonder if this will just deadlock too? But then this would be a separate one that is not handled here in any case? 🤔 

* Idea 2: Just use the same lock in store and subscription. The only reason why its two locks is convenience: The subscription does not have a strong reference to the store and therefore it wasn't straight-forward to lock on it. We could pass in a lock and use that lock in the store and subscriptions. That was easy to prototype and passes the unit tests. One potential downside of this is that lifecycle methods may have to wait for an internal reduce/dispatch to complete. Due to the double lock this was already kind of the case whenever we did not deadlock. Maybe that is negligible?

csadilek: OK, after more discussion, we decided to remove the the unneeded lock when reading the store's state. It causes the deadlock on resume, and also (unnecessarily we think) blocks reading the state when actions are being processed. So, this solves the deadlock and improves performance. Let's file a follow-up issue for the potential problem in unsubscribe which we haven't seen yet, but should fix this sprint as well.

csadilek:
bors r=pocmo

PR #5797, opened 2020-02-04
Closes #5744: Replace 'in memory' session store parser with stream parser.

### LINKED ISSUE #5744: OutOfMemoryError: Failed to allocate a 67644144 byte allocation with 25165824 free bytes and 61MB until OOM, max al...
https://sentry.prod.mozaws.net/operations/fenix-fennec-nightly/issues/7170540/

```
OutOfMemoryError: Failed to allocate a 67644144 byte allocation with 25165824 free bytes and 61MB until OOM, max allowed footprint 229046520, growth limit 268435456
    at java.lang.StringFactory.newStringFromChars(StringFactory.java:267)
    at java.lang.StringBuffer.toString(StringBuffer.java:671)
    at java.io.StringWriter.toString(StringWriter.java:210)
    at kotlin.io.FilesKt__FileReadWriteKt.readText(FileReadWrite.kt:5)
    at kotlin.io.FilesKt__FileReadWriteKt.readText$default(FileReadWrite.kt:2)
...
(8 additional frame(s) were not displayed)

Failed to allocate a 67644144 byte allocation with 25165824 free bytes and 61MB until OOM, max allowed footprint 229046520, growth limit 268435456
```

pocmo:
This replaces our in-memory session store parser (using `org.json`) with a stream-based version (using `JsonReader`) avoiding reading the whole store into memory and then parsing it completely. This hopefully will avoid memory issues like we saw in #5744 (although for only one user so far).

At first I wasn't sure whether I should trust the new parser and I was thinking about maybe running both in parallel first and reporting back if they yield different results. However I think we should just go ahead and use the new parser, because:

* The new parser passes all existing tests
* There's a new test checking that both parsers yield the same result for our existing test sessions stores.
* This patch adds more reporting of parsing failures to Sentry.
* The new build will get QA coverage again.
* We are still in Nightly rollout and can take a bit more risk


components/support/migration/src/main/java/mozilla/components/support/migration/FennecSessionMigration.kt
```diff
@@ -36,13 +36,21 @@ internal object FennecSessionMigration {
 
         for (file in sessionFiles) {
             try {
-                return parseJSON(file.readText())
+                return StreamingSessionStoreParser.parse(file)
+            } catch (e: FileNotFoundException) {
+                logger.error("FileNotFoundException while trying to parse sessopm stpre", e)
```
grigoryk: You're not submitting this exception because we expect to see it, or..? Maybe change the logger level to `warn` to indicate that?

components/support/migration/src/main/java/mozilla/components/support/migration/session/StreamingSessionStoreParser.kt
```diff
@@ -0,0 +1,227 @@
...
+
+import android.util.JsonReader
+import mozilla.components.browser.session.Session
+import mozilla.components.browser.session.SessionManager
+import mozilla.components.support.migration.Result
+import java.io.File
+import java.io.IOException
+
+/**
+ * Session store parser reading the session store as a stream reducing memory impact.
```
grigoryk: uber-nit: missing comma after "stream" :p

components/support/migration/src/main/java/mozilla/components/support/migration/session/StreamingSessionStoreParser.kt
```diff
@@ -0,0 +1,227 @@
...
+                    index = reader.nextInt()
+                }
+                else -> {
+                    reader.skipValue()
+                }
+            }
+        }
+
+        return if (index < 1 || entries.size < index) {
+            // This tab has no entries at all. Let's just stkip it.
```
grigoryk: nit: "skip"

components/support/migration/src/main/java/mozilla/components/support/migration/session/StreamingSessionStoreParser.kt
```diff
@@ -0,0 +1,227 @@
...
+                "index" -> {
+                    index = reader.nextInt()
+                }
+                else -> {
+                    reader.skipValue()
+                }
+            }
+        }
+
+        return if (index < 1 || entries.size < index) {
```
grigoryk: Isn't it strange, a tab with no entries? Well, not much you can do here, other than maybe report it...

components/support/migration/src/main/java/mozilla/components/support/migration/FennecSessionMigration.kt
pocmo: Yeah, I think it is expected that the store or its backup do not exist. Agree, warning makes sense here.

components/support/migration/src/main/java/mozilla/components/support/migration/session/StreamingSessionStoreParser.kt
pocmo: Yeah, since I didn't write the format and since it is not really documented, I wanted to be over-protective here.

pocmo:
bors r=grigoryk

PR #5809, opened 2020-02-04
Closes #5805: Addon name missing after uninstall / reinstall

### LINKED ISSUE #5805: Addon name not present until restart after uninstall / reinstall
This is a minor bug in our reducer. We are (and need to be) tolerant to browser actions arriving before we get notified that an extension was successfully installed. If that happens we find existing state when processing the install action and simply return that. However, this causes us to lose state only present in the install action (e.g. the name of the extension).

csadilek:


csadilek:
bors r=gabrielluong,Amejia481

Issue #5818, opened 2020-02-05
Migrate default search engine


pocmo:
Currently we do not migrate the default search engine when a user updates from Fennec to Fenix. This means that a user will get the default search engine for that configuration after the upgrade.

We want to try a "best effort" migration:
- Get the name of the default search engine from Fennec
- Try to find a match in the list of search engines for this Fenix installation
- If a match is found set the default to the matching

This will only work in cases where the previous search engine name is still in the list of search engines Fenix has determined for this installation. This can fail for custom search engines or when the user is in a completely different region that does not offer the previous choice. However the approach looks very promising and should migrate a decent chunk of defaults.

pocmo:
This is in the latest build.

sflorean:
Verified as fixed for the default search engine added in Fennec and Fenix. Areas tested:
- search widget
- settings
- URL bar 
- performed different searches 
Environment: 
- Migration build 7 - 02/06
- Google Pixel XL (Android Q)
- Samsung Galaxy Note 8 (Android 9)
- Motorola Moto G6 (Android 8.0)
- Sony Xperia Z5 Lite (Android 7)
- LG G4  Android 6.0.1)
- Xiaomi Mi Pad 2 (Android 5.1) - x86 architecture

PR #5819, opened 2020-02-05
Issue #5818: Try to migrate default search engine.

pocmo:
Implementation for migrating the default search engine from Fennec to Fenix.

This is a "best effort" migration. We will recover the default search engine name from Fennec and then try to find a matching search engine in the search engine manager for the current Fenix installation. If we find a match we will update the default search engine to that; otherwise the regular default search engine for this installation will be used.


components/support/migration/src/main/java/mozilla/components/support/migration/TelemetryHelpers.kt
```diff
@@ -16,6 +16,7 @@ internal fun Migration.telemetryIdentifier(): String {
         Migration.Settings -> "settings"
         Migration.Addons -> "addons"
         Migration.TelemetryIdentifiers -> "telemetry_identifiers"
+        Migration.SearchEngine -> "search_engine"
```
grigoryk: in metrics.yaml `migration_versions` you called it `search` - these labels need to be the same (also, see comment on top of this function).

components/support/migration/src/main/java/mozilla/components/support/migration/SearchEngineMigration.kt
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    private suspend fun determineNewDefault(
+        context: Context,
+        searchEngineManager: SearchEngineManager,
+        name: String
+    ): Result<SearchEngineMigrationResult> {
+        val searchEngines = searchEngineManager.getSearchEnginesAsync(context)
+
+        searchEngines.forEach { engine ->
```
grigoryk: maybe a comment here would be useful - if passed-in fennec default doesn't match anything from the search manager, we do nothing - which is equivalent of keeping fenix's default search engine as a "fallback".

pocmo: That is mentioned in the class comment. But I agree, here in context, we could mention that too.

pocmo:

This is an extension of the already existing migration telemetry ([data review](https://github.com/mozilla-mobile/android-components/pull/5483#issuecomment-573971458)). This patch set extends the telemetry probes to cover the new "default search engine" migration, with the goal of measuring the success/error rates of this new migration type.

2) Why does Mozilla need to answer these questions?  Are there benefits for users? Do we need this information to address product or business requirements?

This information is essential to ensure the default search engine migration as part of the Fennec->Fenix migration proceeds without unexpected errors. The information is used to determine the overall success rate and to decide whether additional work is needed to achieve a higher success rate.

Crash and failed assertion reporting via Socorro and Sentry is in place to debug individual errors, but that isn't detailed enough for us to have a good picture of quality of the migrations in the wild.

No.

5) List all proposed measurements and indicate the category of data collection for each measurement, using the [Firefox data collection categories](https://wiki.mozilla.org/Firefox/Data_Collection) found on the Mozilla wiki.

<table>
  <tr>
    <td>Tracking Bug #</td>
  </tr>
  <tr>
    <td>Search engine migration</td>
    <td>Category 2, Interaction data</td>
    <td>https://github.com/mozilla-mobile/fenix/issues/5818</td>
  </tr>
</table>

In detail:

* We add an additional label to the `migration_versions` probe to collect the version of the search engine migration we ran (`search` / `labeled_string`)
* We add `migration.search.any_failures` (`boolean`) to indicate whether failures occured while migrating the search engine.
* For failures we add `migration.search.failure_reason` to indicate the reason why the migration failed (one out of three different constants)
* For successful migrations we add `migration.success_reason` to indicate why/how the migration succeeded (currently only one constant)

For the same time as the other migration probes (see [data review](https://github.com/mozilla-mobile/android-components/pull/5483#issuecomment-573971458)).

* Which release channels?
all

* Which countries?
all

* Which locales?
all

no

No opt-out mechanism.

Via the migration dashboard on sql.telemetry.mozilla.org.
 
Internally within the core Fenix/A-C/A-S teams.

11) Is there a third-party tool (i.e. not Telemetry) that you are proposing to use for this data collection? If so:

n/a

components/support/migration/src/main/java/mozilla/components/support/migration/TelemetryHelpers.kt
pocmo: Argh, wrote that before any other telemetry code. :)

pocmo:
Added some tests. I think this is now ready to land after the data review.

PR #5822, opened 2020-02-05
Closes #5807 - Add a callback to notify when experiments are updated

### LINKED ISSUE #5807: Add a callback mechanism to inform application when experiments are updated.
Originally from Bugzilla [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1569647)

There currently isn't a way for an app that consumes service-experiments to be notified when enrollment in a new experiment happens.  Normally it would be detected on the next launch of the application, but it may be desirable to the consuming application to know about enrollment into new experiments on the current execution.  

This was brought up by @csadilek [here](https://github.com/mozilla-mobile/fenix/pull/4098/files#r306329116)

travis79:


samples/glean/src/main/java/org/mozilla/samples/glean/ExperimentUpdateReceiver.kt
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    override fun onReceive(p0: Context?, p1: Intent?) {
+        listener.onExperimentsUpdated()
+    }
+
+    interface ExperimentUpdateListener {
+        fun onExperimentsUpdated()
+    }
+}
```
Dexterp37: nit: missing newline

samples/glean/src/main/java/org/mozilla/samples/glean/MainActivity.kt
```diff
@@ -13,12 +14,20 @@ import org.mozilla.samples.glean.GleanMetrics.Test
...
-open class MainActivity : AppCompatActivity() {
+open class MainActivity : AppCompatActivity(), ExperimentUpdateReceiver.ExperimentUpdateListener {
+
+    private var experimentUpdateReceiver: ExperimentUpdateReceiver? = null
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_main)
 
+        // Set up the ExperimentUpdateReceiver to receive experiment updated Intents
```
Dexterp37: Can you add a comment stating that this is not relevant to the Glean SDK but to the experiments library?

samples/glean/src/main/java/org/mozilla/samples/glean/MainActivity.kt
```diff
@@ -53,16 +62,27 @@ open class MainActivity : AppCompatActivity() {
...
-            textViewExperimentStatus.text = getString(R.string.experiment_not_active)
+            onExperimentsUpdated()
+        }
+    }
+
+    /**
+     * This function will be called by the ExperimentUpdateListener interface when the experiments
+     * are updated.
+     */
+    override fun onExperimentsUpdated() {
```
Dexterp37: Can you add a comment stating that this is not relevant to the Glean SDK but to the experiments library?

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
```diff
@@ -56,7 +58,8 @@ open class ExperimentsInternalAPI internal constructor() {
      */
     fun initialize(
         applicationContext: Context,
-        configuration: Configuration = Configuration()
+        configuration: Configuration = Configuration(),
+        onExperimentsUpdated: (() -> Unit)? = null
```
Dexterp37: **Note*: the owners of the experiments library has the final word on this!

While this solves the problem you're seeing (which I guess is the fact that `withExperiment` can be called while the library is not ready, yet!), it also pushes the boilerplate to the user of the API, which now has to deal with the library state (which should be an internal detail).

I think that `withExperiment` should not have a dependency on when `initialize` is called and its dependency should be handled internally, as we do with Glean.

With that said, I think the owner of the library can make the call :)

travis79: Understood about waiting for approval from the owners of the library.

Experiments in this PR will still work the exact same way it does today, if the callback is not used. The experiment will be received and enrollment determined, only the app won't be able to tell this until the next time it is launched (or if it polls `withExperiment`).  

This adds to the experiments API the ability for an application to receive a notification when the experiments have been updated through the callback, that way it can act on the enrollment without having to wait for launch or polling `withExperiment`.

mythmon: I agree that `withExperiment` should not have an ordering dependency on calls to `initialize`. It would be great if `withExperiment` acted as more of an event listener and less of an if block.

That being said, I think that this change provides a useful improvement to the use cases we are seeing today, and is worth keeping. I think that ultimately we should prefer the improvements to `withExperiment`, or some other reactive experiment checking. Until then, lets take the easy wins that get us unblocked.

PR #5835, opened 2020-02-06
Closes #4779 - Fix `ExperimentsDebugActivity` command to change server

### LINKED ISSUE #4779: Experimenter testing command doesn't fetch experiments from overridden server endpoint (like staging server)
In setting up the first Mobile Experimenter experiment with [ETP on Fenix](https://github.com/mozilla-mobile/fenix/issues/5651), we can't seem to use the records from the staging server.

### Acceptance Criteria
- [ ] Overriding endpoint in adb testing instructions fetches experiments from the correct endpoint
- [ ] Includes a test
- [ ] Timeline: in the next few weeks (**done before 11/8**), before any upcoming Fenix experiments
- [ ] The experiment testing (e.g. changing an endpoint) doesn't make assumptions about where experiments are init-ed (specifically, whether experiments can be init-ed in MainApplication).
- [ ] ^If not possible, or too difficult, update documentation to be clear about any constraints testing creates.  

Even though we're overriding the endpoint to use staging, it's still fetching from prod.
I don't see [a test testing pulling experiments from the updated endpoint](https://github.com/mozilla-mobile/android-components/blob/master/components/service/experiments/src/test/java/mozilla/components/service/experiments/debug/ExperimentsDebugActivityTest.kt#L101).

staging: https://settings-writer.stage.mozaws.net/v1
prod: https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/mobile-experiments/records?_since=1566233826256
command: `adb shell am start -n org.mozilla.fenix/mozilla.components.service.experiments.debug.ExperimentsDebugActivity  --ez updateExperiments true  --es setKintoInstance "staging" --es overrideExperiment "fenix-etp-5651" --es branch “treatment_standard"`

```
2019-10-17 15:15:34.740 6476-6542/org.mozilla.fenix.debug I/experiments: Fetching experiments from https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/mobile-experiments/records?_since=1566233826256
2019-10-17 15:15:34.741 6476-6476/org.mozilla.fenix.debug I/ExperimentsDebugActivity: Changing Kinto instance to https://settings-writer.stage.mozaws.net/v1
2019-10-17 15:15:34.794 6476-6542/org.mozilla.fenix.debug D/experiments: Fetched experiments from https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/mobile-experiments/records?_since=1566233826256:
    {"data":[]}
2019-10-17 15:15:34.795 6476-6542/org.mozilla.fenix.debug I/experiments: Experiments update from server: ExperimentsSnapshot(experiments=[Experiment(id=fenix-test-2019-08-05, description=End-to-end test of the Mobiles Experiments platform for Firefox Preview., lastModified=1566233826256, schemaModified=1566228455761, buckets=Buckets(start=0, count=667), branches=[Branch(name=a, ratio=1), Branch(name=b, ratio=1), Branch(name=c, ratio=1)], match=Matcher(appId=^org.mozilla.fenix$, appDisplayVersion=null, appMinVersion=null, appMaxVersion=null, localeLanguage=null, localeCountry=null, deviceManufacturer=null, deviceModel=null, regions=[], debugTags=[]))], lastModified=1566233826256)
2019-10-17 15:15:34.799 6476-6542/org.mozilla.fenix.debug I/experiments: onExperimentsUpdated - no experiment currently active, looking for match
2019-10-17 15:15:59.725 6476-6476/org.mozilla.fenix.debug I/ExperimentsDebugActivity: Override to the following experiment/branch:fenix-etp-5651/treatment_standard
```

A few things to investigate:
- Is this dependent on where the experiments are initialized in the `MainApplication`? The `ExperimentsDebugActivity` needs the Application to have been initialized before it can run, so does this preclude having the experiment initialization in `MainApplication`? (The testing framework shouldn't require code structure changes.)
- Should the experiments be updated [after checking the endpoint](https://github.com/mozilla-mobile/android-components/blob/master/components/service/experiments/src/main/java/mozilla/components/service/experiments/debug/ExperimentsDebugActivity.kt#L79)?
- Is the updated endpoint in the configuration somehow not the same one [used to fetch the new experiments](https://github.com/mozilla-mobile/android-components/blob/master/components/service/experiments/src/main/java/mozilla/components/service/experiments/debug/ExperimentsDebugActivity.kt#L102)?

travis79:
This fixes the issue of not being able to change the Kinto endpoint using the `adb` command.  

The issue was due to the fact that the source was being updated, but then subsequently not being used for the update because the `ExperimentUpdater` was building a new `ExperimentSource` using the old config rather than using the updated one.

### DIFF
--- a/components/service/experiments/src/main/java/mozilla/components/service/experiments/ExperimentsUpdater.kt
+++ b/components/service/experiments/src/main/java/mozilla/components/service/experiments/ExperimentsUpdater.kt
@@ -42,7 +42,7 @@ internal class ExperimentsUpdater(
     private val experiments: ExperimentsInternalAPI
 ) {
     private val logger: Logger = Logger(LOG_TAG)
-    private lateinit var config: Configuration
+    internal lateinit var config: Configuration
 
     internal lateinit var source: KintoExperimentSource
 
@@ -119,7 +119,7 @@ internal class ExperimentsUpdater(
     @Synchronized
     internal fun updateExperiments(): Boolean {
         return try {
-            val serverExperiments = getExperimentSource(config).getExperiments(experiments.experimentsResult)
+            val serverExperiments = source.getExperiments(experiments.experimentsResult)
             logger.info("Experiments update from server: $serverExperiments")
             experiments.onExperimentsUpdated(serverExperiments)
             true
--- a/components/service/experiments/src/main/java/mozilla/components/service/experiments/debug/ExperimentsDebugActivity.kt
+++ b/components/service/experiments/src/main/java/mozilla/components/service/experiments/debug/ExperimentsDebugActivity.kt
@@ -98,6 +98,8 @@ class ExperimentsDebugActivity : Activity() {
                 logger.info("Changing Kinto instance to ${config.kintoEndpoint}")
                 // Update the default config
                 Experiments.configuration = config
+                // Also update the updater's copy of the config
+                Experiments.updater.config = config
                 // Reset the source to use the new endpoint
                 Experiments.updater.source = Experiments.updater.getExperimentSource(config)
             }


travis79:
@mythmon This is ready for your review.

travis79:
> This change looks like it fixes a decently clear bug to me. Are there any tests for this feature that need updated? Should there be?

I tried to come up with a good way to test this better and, barring some sort of integration test that actually pulls from staging or dev (and all the overhead that comes along with that), I'm not sure we can easily do more than what we have here.

The current test coverage for this was correctly looking at the `ExperimentSource` associated with the updater to verify the endpoint was changed, but because the updater wasn't using the existing source and was creating a new one with a stale copy of the config, it never used the updated endpoint.  

That being said, if you think that there is a particular case or need for extra test coverage here, I can try and add an integration test that validates this functionality to the sample app as a follow up to this.

mythmon:
(Wow there a lot of test statuses reported here)

Given that this is an important part of our ability to successfully launch a recipe that has already bitrotted once, I think it would be good to have an automated test somewhere that triggers this code path and validates that it works. That being said, I don't think we need to block this change on that.

PR #5837, opened 2020-02-06
Closes #5836: Remove unneeded locales from add-ons

### LINKED ISSUE #5836: TransactionTooLargeException when serializing add-on
In Fenix: https://sentry.prod.mozaws.net/operations/fenix-fennec-nightly/issues/7221928/

The add-ons returned from the `AddonManager` contain all translations of description/summary/name etc. and really don't need to. 

We can keep only the translations we need: the current default lang and the fallback. We have all translations returned from the provider anyway, so if the user changes to a different language, the next time they open the add-ons page they will get the updated language if it exists.

csadilek:
I've also verified that switching languages still works with this. Our provider still has all languages, with this we just prevent storing them all in the add-on instance which isn't needed and causes problems when serializing (sending the add-on instance to a different fragment).


csadilek:
bors r=Amejia481

Issue #5278, opened 2019-12-10
Fennec migration: about:home is restored in Open Tabs after reopening the app


ebalazs-sv:
## Preconditions:
Have Firefox Beta/Nightly installed and logged in.
## Steps to reproduce
1. Replace Fennec Beta/Nightly with Fenix using a Beta/Nightly build from https://github.com/mozilla-mobile/fenix/wiki/Fennec-Migration;
2. Launch Fenix and remove about:home from open tabs;
3. Close the app from Recent app and reopen it;
4. Observe the Open tabs section,
### Expected behavior
The Open tabs section is empty.
### Actual behavior
The about:home page is displayed in Open tabs.
### Notes:
[Video](https://drive.google.com/file/d/1DpHLjhuTQ1cyTXZA7FFuggOMDvmrfInh/view?usp=sharing)

### Device information

* Android device: Samsung Galaxy S8 (Android 9) - ARM64; Motorola Nexus 6 (Android 7.1.1); Xiaomi Mi Pad 2 (Android 5.1).
* Fenix version: Beta 12/10 GV 71; Nightly 191210 GV 72

codrut-topliceanu:
@ebalazs-sv 
I'm unable to reproduce the issue. (Pixel 4 Android 10 & LG K4 Android 6)
Is the issue still present for you?

pocmo:
I can still reproduce that (I think the key is that the tab needs to have more history than only `about:home` which will otherwise not kept by Fennec and therefore can't be restored) and will provide a patch.

sflorean:
Verified as fixed on Fennec Nightly Migration build 8. about:home tabs are not migrated or displayed after migration. 
Devices: 
- Google Pixel 3 (Android Q)
- Nexus 6P (Android 8.1.0)
- Xiaomi Mi Pad 2 (Android 5.1) - x86 architecture

PR #5853, opened 2020-02-07
Closes #4466, #5278: Migration: Filter about:home tabs and rewrite about:reader URLs.

### LINKED ISSUE #4466: Fennec migration: Decide how to deal with about:reader URLs.
We need to decide how we want to deal with `about:reader` URLs.

* What do we want to do if a tab with an `about:reader` URL is open? Just point it to the original page? Or try to transform it into a "new" reader mode Session?

* What do we want to do with `about:reader` URLs in bookmarks? Those showed up as "offline available" content in Fennec.

pocmo:
For #4466 an #5278.

We are now:
- Skipping tabs that are pointing to `about:home` - Fenix can't handle that URL.
- Rewriting `about:reader` URLs to the original URL. Fenix can't handle `about:reader` URLs but can just load the original URL.


components/support/migration/src/main/java/mozilla/components/support/migration/FennecSessionMigration.kt
```diff
@@ -36,7 +37,7 @@ internal object FennecSessionMigration {
 
         for (file in sessionFiles) {
             try {
-                return StreamingSessionStoreParser.parse(file)
+                return StreamingSessionStoreParser.parse(file).filter(logger)
             } catch (e: FileNotFoundException) {
                 logger.warn("FileNotFoundException while trying to parse sessopm stpre", e)
```
grigoryk: nit: "sessopm stpre" is probably "session store" :)

components/support/migration/src/main/java/mozilla/components/support/migration/FennecSessionMigration.kt
```diff
@@ -73,3 +74,48 @@ internal object FennecSessionMigration {
...
+                if (index < selectedIndex) {
+                    selectedIndex--
+                }
+                null
+            }
+
+            // We rewrite about:reader URLs to their actual URLs. Fenix does not handle about:reader
+            // URLs, but it can load the original URL.
+            item.session.url.startsWith(ABOUT_READER) -> {
+                val url = Uri.decode(item.session.url.substring(ABOUT_READER.length))
```
grigoryk: Any chance Uri.decode can fail? If so, we'll fail the whole session store parse, vs just ignoring an individual (bad) url.

pocmo: Just read through the code and saw that it could theoretically at least throw `IllegalStateException`. I think I'll just catch `Exception` here and report this to Sentry.

pocmo:
bors r=grigoryk

PR #5877, opened 2020-02-10
Closes #5875: onTitleChange blocking on history storage write

### LINKED ISSUE #5875: GeckoEngineSession.onTitleChange blocking on history storage write
We're currently blocking on `historyDelegate.onTitleChanged` and therefore `historyStorage.recordObservation` which seems like an oversight. This can be fairly costly and may be the cause of:  https://github.com/mozilla-mobile/fenix/issues/8238

Let's move this to an IO thread.

csadilek:
Looking through other call-sites, this seems save to move to an IO thread esp., as it writes to storage. We also already have a custom scope in the engine session which I am using here.

### DIFF
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -10,7 +10,6 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.MainScope
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.runBlocking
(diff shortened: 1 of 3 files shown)


csadilek:
> in some cases, we may fail to persist title changes (if the job is cancelled before it starts)

We discussed this more and found that to be acceptable. Strictly speaking, this was possible before too. A tab could've been opened, a page partially loaded, and the tab closed again before we got the callback on the delegate. This window of time is now bigger, of course, and the case more likely.

> we may notify observers of the new title before it's persisted in the storage

This is a change in behaviour, as we always notify the observer even if we haven't persisted yet. I've added comments explicitly calling this out. Observers of `onTitleChange` don't (and now can't) depend on the visit being persisted to disk.

//cc @pocmo do you want to take a look as well?

csadilek:
Did a round of manual testing to confirm we're no longer on the main thread and there's are no obvious regressions.

pocmo:
bors r=grigoryk,pocmo

csadilek:
bors retry

PR #5888, opened 2020-02-11
Closes #5815: Update unsupported addons adapter when uninstall occurs

### LINKED ISSUE #5815: List of unsupported add-ons is not refreshed when add-on is uninstalled
Noticed this too during migration. Uninstall worked, but the list wasn't refreshed. We could also go back a step if there's no unsupported add-on left.

psymoon:
One concern I have with this approach is if adapter can ever be null (i.e. garbage collected) at the wrong time (when `onUninstallSuccess` is invoked but activity/fragment is being destroyed). I tried testing uninstalling and pressing back quickly a few times and I haven't seen any crashes (perhaps the nullable adapter check is sufficient)

### DIFF
--- a/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/UnsupportedAddonsAdapter.kt
+++ b/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/UnsupportedAddonsAdapter.kt
@@ -10,6 +10,7 @@ import android.view.ViewGroup
 import android.widget.ImageButton
 import android.widget.ImageView
 import android.widget.TextView
+import androidx.annotation.VisibleForTesting
 import androidx.recyclerview.widget.RecyclerView
 import mozilla.components.feature.addons.Addon
 import mozilla.components.feature.addons.AddonManager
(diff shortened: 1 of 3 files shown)


components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/UnsupportedAddonsAdapter.kt
```diff
@@ -28,9 +29,11 @@ class UnsupportedAddonsAdapter(
     private val unsupportedAddonsAdapterDelegate: UnsupportedAddonsAdapterDelegate,
     private val unsupportedAddons: List<Addon>
```
Amejia481: nit: To avoid using `_` we could remove the `private` modifier from  the`unsupportedAddons`and use another name for the list.

```kotlin
class UnsupportedAddonsAdapter(
    private val addonManager: AddonManager,
    private val unsupportedAddonsAdapterDelegate: UnsupportedAddonsAdapterDelegate,
    val addons: List<Addon>
) : RecyclerView.Adapter<RecyclerView.ViewHolder>() {
    private val unsupportedAddons : MutableList<Addon> = addons.toMutableList()
```

components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/UnsupportedAddonsAdapter.kt
```diff
@@ -40,18 +43,35 @@ class UnsupportedAddonsAdapter(
...
 
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    @Suppress("TooGenericExceptionCaught")
+    internal fun removeUninstalledAddonAtPosition(position: Int) {
+        try {
+            val uninstalledAddon = _unsupportedAddons[position]
+            _unsupportedAddons.remove(uninstalledAddon)
+            notifyItemChanged(position)
+            unsupportedAddonsAdapterDelegate.onUninstallSuccess()
+        } catch (throwable: IndexOutOfBoundsException) {
```
Amejia481: I'm a bit confuse of how we can get an `IndexOutOfBoundsException` as we are always deleting an item that is in bounds.

psymoon: Spamming the remove button would cause this. I set the visibility of the remove button to View.GONE to defend against this, but you just never know 🤷‍♂

psymoon:
bors r=Amejia481

PR #5893, opened 2020-02-11
Move add-ons translations on the right path

Amejia481:
Related issue: https://github.com/mozilla-mobile/fenix/issues/8188
---

### Pull Request checklist

### After merge


l10n.toml
```diff
@@ -163,7 +163,7 @@ branches = [
 
 [[paths]]
   reference = "components/feature/addons/src/main/res/values/strings.xml"
-  l10n = "components/feature/addons/src/main/res/values/values-{android_locale}/strings.xml"
+  l10n = "components/feature/addons/src/main/res/values-{android_locale}/strings.xml"
```
Amejia481: @Pike I noticed that I put the wrong path on the `l10n.toml` file. I'm moving the files to the right place on this pr, I believe you have to do the same on your end sorry for that.

Pike: So this is gonna be an interesting experience, 'cause you're the first one trying.

CC @mathjazz to make sure he's aware of this coming in. Also @flodolo.

We might need to remove files by hand.

On a positive note, the values-foo files are only written, not read, so we won't loose any data, AFAICT.

flodolo: Without [work on Pontoon's side](https://mozilla-l10n.github.io/documentation/tools/pontoon/renaming_a_file.html), are we going to lose a bunch of things? History, attributions, pending suggestions.

Pike: The Resource path doesn't change, just where Pontoon syncs l10n files to, and reads from.

This would be a problem if we had changes coming in to Pontoon through VCS, but for android-l10n we don't, so no problem.

I think.

Famous last words.

Related, that downloads on projects with project config are broken, https://bugzilla.mozilla.org/show_bug.cgi?id=1530988.

components/feature/addons/src/main/res/values-pt-rBR/strings.xml
```diff
@@ -53,7 +53,7 @@
     <!-- The authors of an add-on. -->
     <string name="mozac_feature_addons_authors">Autores</string>
     <!-- The last date that the add-on was updated. -->
-    <string name="mozac_feature_addons_last_updated">Ultima atualização</string>
+    <string name="mozac_feature_addons_last_updated">Última atualização</string>
```
Amejia481: @Pike Android lint was complaining about this on [Task Cluster](https://firefox-ci-tc.services.mozilla.com/tasks/X7fakBMrTsqfogf6xqzd7g/runs/0/logs/https%3A%2F%2Ffirefox-ci-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FX7fakBMrTsqfogf6xqzd7g%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log) Should I keep the change here or just ignore it? 

```
/builds/worker/checkouts/src/components/feature/addons/src/main/res/values-pt-rBR/strings.xml:56: Error: "Ultima" is a common misspelling; did you mean "Última" ? [Typos]
<string name="mozac_feature_addons_last_updated">Ultima atualização</string>
                                                                                         
```

Pike: I've made the corresponding change in Pontoon now, too. Matches what the translation memory has, that's why.

l10n.toml
Amejia481: Is OK to land as it is?

Pike: Yes, thanks.

I started feeling bad for you for all that jazz flying by, none of which is actionable for you.

mathjazz: I also believe this should JustWork™. We might want to test on stage first, though. TM

Amejia481: No worries!

PR #5896, opened 2020-02-11
Closes #5894: AddonManager API to suspend until all actions are processed

### LINKED ISSUE #5894: AddonManager: Provide API to suspend until all actions are processed
We have numerous bugs filed in Fenix were switching fragments quickly between async add-on operations causes inconsistencies when re-opening the add-on manager.

These could all be solved if we kept showing the loading state in add-on manager until these async actions complete (in success or error). Currently we are displaying the most up-to-date state e.g. while installing on a slow network, and switching to a different fragment and back to the add-on manager we could still show "recommended" which would allow the user to click + again. The same can happen with enabled/disabled/uninstall.

If `getAddons` only returned once the actions are processed we could avoid these problems.

csadilek:


csadilek:
bors r=psymoon,Amejia481

PR #5902, opened 2020-02-12
For Fenix #8324 - Adds flags for optionally not accessing Keystore

ekager:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/SecureAbove22Preferences.kt
+++ b/components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/SecureAbove22Preferences.kt
@@ -59,9 +59,13 @@ private interface KeyValuePreferences {
  *
  * @param context A [Context], used for accessing [SharedPreferences].
  * @param name A name for this storage, used for isolating different instances of [SecureAbove22Preferences].
+ * @param forceInsecure A flag indicating whether to force plaintext storage. If set to `true`,
+ * [InsecurePreferencesImpl21] will be used as a storage layer, otherwise a storage implementation
+ * will be decided based on Android API version, with a preference given to secure storage
  */
-class SecureAbove22Preferences(context: Context, name: String) : KeyValuePreferences {
-    private val impl = if (Build.VERSION.SDK_INT >= M) {
+class SecureAbove22Preferences(context: Context, name: String, forceInsecure: Boolean = false) :
+    KeyValuePreferences {
+    private val impl = if (Build.VERSION.SDK_INT >= M && !forceInsecure) {
         SecurePreferencesImpl23(context, name)
     } else {
         InsecurePreferencesImpl21(context, name)
@@ -82,12 +86,39 @@ class SecureAbove22Preferences(context: Context, name: String) : KeyValuePrefere
  * A simple [KeyValuePreferences] implementation which entirely delegates to [SharedPreferences] and doesn't perform any
  * encryption/decryption.
  */
-private class InsecurePreferencesImpl21(context: Context, name: String) : KeyValuePreferences {
+@SuppressWarnings("TooGenericExceptionCaught")
+private class InsecurePreferencesImpl21(
+    context: Context,
+    name: String,
+    migrateFromSecureStorage: Boolean = true
+) : KeyValuePreferences {
     companion object {
         private const val SUFFIX = "_kp_pre_m"
     }
 
+    internal val logger = Logger("mozac/InsecurePreferencesImpl21")
+
     private val prefs = context.getSharedPreferences("$name$SUFFIX", MODE_PRIVATE)
+
+    init {
+        // Check if we have any encrypted values stored on disk.
+        if (migrateFromSecureStorage && Build.VERSION.SDK_INT >= M && prefs.all.isEmpty()) {
+            val secureStorage = SecurePreferencesImpl23(context, name, false)
+            // Copy over any old values.
+            try {
+                secureStorage.all().forEach {
+                    putString(it.key, it.value)
+                }
+            } catch (e: Exception) {
+                // Certain devices crash on various Keystore exceptions. While trying to migrate
+                // to use the plaintext storage we don't want to crash if we can't access secure
+                // storage, and just catch the errors.
+                logger.error("Migrating from secure storage failed", e)
+            }
+            // Erase old storage.
+            secureStorage.clear()
+        }
+    }
 
     override fun all(): Map<String, String> {
         return prefs.all.mapNotNull {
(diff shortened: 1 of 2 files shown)


components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/SecureAbove22Preferences.kt
```diff
@@ -59,9 +59,16 @@ private interface KeyValuePreferences {
...
+ * @param useSecureIfAvailable A flag indicating whether or not to use encrypted storage when available.
+ * If set to `true`, will only by encrypted on Android API 23+. Otherwise, even if this flag is set to `true`,
+ * account state will be stored in plaintext.
  */
-class SecureAbove22Preferences(context: Context, name: String) : KeyValuePreferences {
-    private val impl = if (Build.VERSION.SDK_INT >= M) {
+class SecureAbove22Preferences(
+    context: Context,
+    name: String,
+    useSecureIfAvailable: Boolean = true
```
grigoryk: I think something like `forceInsecure` is closer to your intent here, right?

An alternative would have been to just expose the `insecureimpl` class as public API, and let consumers choose it directly instead of going through the "compat" version.
I think I'd actually prefer that, it seems cleaner?

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
```diff
@@ -1008,7 +1008,12 @@ open class FxaAccountManager(
         return if (deviceConfig.secureStateAtRest) {
             SecureAbove22AccountStorage(context, crashReporter)
         } else {
-            SharedPrefAccountStorage(context, crashReporter)
+            // Don't migrate here so we don't hit the KeyStore path if the secureStateAtRest flag is false
```
grigoryk: This change is interesting - it removes our ability to flip back and forth between the two storage modes, changing the public API in a subtle way.

Before, application could switch `secureStateAtRest` in between versions without any problems, in theory - state would be migrated between storage layers.

After this change, if the application has secureStateAtRest=true, but then switches it to `false`, FxA state will be lost. So flipping this flag and releasing, every single user will get logged out. That's a pretty big foot gun!

I'm not against doing this temporarily while we learn more about scale of they keystore problems, but we should really put up a big, fat warning sign both in Fenix where this flag is set, and in `DeviceConfig` as well.

grigoryk:
Let's also consider how this can play out over time. Will we need to have an ability to switch back and forth between the two modes? With this API change, we don't really have that since we will log everyone out when disabling secure storage.

On the other hand, this still leaves us with an ability to _turn on_ secure storage -  just not _off_.

I _think_ that's enough for our needs? What do you think?

components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/SecureAbove22Preferences.kt
```diff
@@ -59,9 +59,12 @@ private interface KeyValuePreferences {
  *
  * @param context A [Context], used for accessing [SharedPreferences].
  * @param name A name for this storage, used for isolating different instances of [SecureAbove22Preferences].
+ * @param forceInsecure A flag indicating whether to force plaintext storage. If set to `true`,
+ * [InsecurePreferencesImpl21] will be used as a storage layer.
```
grigoryk: expand the comment? "[...], otherwise a storage implementation will be decided based on Android API version, with a preference given to secure storage."

components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/SecureAbove22Preferences.kt
```diff
@@ -82,13 +85,36 @@ class SecureAbove22Preferences(context: Context, name: String) : KeyValuePrefere
...
+        if (migrateFromSecureStorage && Build.VERSION.SDK_INT >= M) {
+            if (prefs.all.isEmpty()) {
+                val secureStorage = SecurePreferencesImpl23(context, name, false)
+                // Copy over any old values.
+                try {
+                    secureStorage.all().forEach {
+                        putString(it.key, it.value)
+                    }
+                } catch (e: Error) {
+                    // Migrating from encrypted storage didn't work, but we tried
```
grigoryk: in a follow-up, can we record this into sentry as an `info`?

components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/SecureAbove22Preferences.kt
```diff
@@ -82,13 +85,36 @@ class SecureAbove22Preferences(context: Context, name: String) : KeyValuePrefere
...
         private const val SUFFIX = "_kp_pre_m"
     }
 
     private val prefs = context.getSharedPreferences("$name$SUFFIX", MODE_PRIVATE)
 
+    init {
+        // Check if we have any encrypted values stored on disk.
+        if (migrateFromSecureStorage && Build.VERSION.SDK_INT >= M) {
+            if (prefs.all.isEmpty()) {
+                val secureStorage = SecurePreferencesImpl23(context, name, false)
```
grigoryk: Can this throw at all?

components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/SecureAbove22Preferences.kt
grigoryk: a log statement would be nice, too :)

components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/SecureAbove22Preferences.kt
```diff
@@ -82,13 +85,36 @@ class SecureAbove22Preferences(context: Context, name: String) : KeyValuePrefere
...
+                // Copy over any old values.
+                try {
+                    secureStorage.all().forEach {
+                        putString(it.key, it.value)
+                    }
+                } catch (e: Error) {
```
grigoryk: You don't do the try/catch in the other storage (since we don't believe it can fail). Can you add a comment here, describing why this is necessary?

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/AccountStorage.kt
```diff
@@ -36,9 +36,13 @@ internal class SharedPrefAccountStorage(
...
-            secureStorage.read()?.let { secureAccount ->
-                this.write(secureAccount.toJSONString())
-                secureStorage.clear()
+            try {
+                secureStorage.read()?.let { secureAccount ->
+                    this.write(secureAccount.toJSONString())
+                    secureStorage.clear()
+                }
+            } catch (e: Error) {
+                // Migrating the storage didn't work, but we tried
```
grigoryk: log this, and (follow-up is fine), consider sending it to sentry so that we can keep track of these events in the wild.

Also, can you add an explanatory comment, similar to one in InsecurePreferencesImpl21 migration logic?

components/lib/dataprotect/src/main/java/mozilla/components/lib/dataprotect/SecureAbove22Preferences.kt
ekager: I don't think so now that I made the Keystore lazy nothing should be throwable until we try to get `all()`

grigoryk: 👍

ekager:
Opened https://github.com/mozilla-mobile/android-components/issues/5950 for adding Sentry here

ekager:
bors r=grigoryk

Issue #5557, opened 2020-01-13
engine-gecko: Integrate certificate information


csadilek:
We currently only use the full (unparsed) issuer name in our `SecurityInfo`. GeckoView now provides more certificate information we should integrate.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-4735)

csadilek:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1795970

Change performed by the [Move to Bugzilla add-on](https://addons.mozilla.org/en-US/firefox/addon/move-to-bugzilla/).

PR #5916, opened 2020-02-12
For #5557 Get Cert Issuer Name

kglazko:
--- 

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -453,7 +453,7 @@ class GeckoEngineSession(
 
             notifyObservers {
                 // TODO provide full certificate info: https://github.com/mozilla-mobile/android-components/issues/5557
-                onSecurityChange(securityInfo.isSecure, securityInfo.host, securityInfo.certificate?.issuerDN?.name)
+                onSecurityChange(securityInfo.isSecure, securityInfo.host, securityInfo.getIssuerName())
             }
         }
 
@@ -740,6 +740,10 @@ class GeckoEngineSession(
         }
 
         return cookiesPolicies
+    }
+
+    internal fun GeckoSession.ProgressDelegate.SecurityInformation.getIssuerName(): String? {
+        return certificate?.issuerDN?.name?.substringAfterLast("O=")?.substringBeforeLast(",C=")
     }
 
     private operator fun Int.contains(mask: Int): Boolean {
--- a/components/browser/engine-gecko-nightly/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineSessionTest.kt
+++ b/components/browser/engine-gecko-nightly/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineSessionTest.kt
@@ -73,6 +73,8 @@ import org.mozilla.geckoview.WebRequestError
 import org.mozilla.geckoview.WebRequestError.ERROR_CATEGORY_UNKNOWN
 import org.mozilla.geckoview.WebRequestError.ERROR_MALFORMED_URI
 import org.mozilla.geckoview.WebRequestError.ERROR_UNKNOWN
+import java.security.Principal
+import java.security.cert.X509Certificate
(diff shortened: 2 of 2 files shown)


kglazko:
This still needs to be tested with a local build of Fenix, do not merge.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -742,6 +742,10 @@ class GeckoEngineSession(
         return cookiesPolicies
     }
 
+    internal fun GeckoSession.ProgressDelegate.SecurityInformation.getIssuerName(): String? {
+            return certificate?.issuerDN?.name?.substringAfterLast("O=")?.substringBeforeLast(",C=")
```
pocmo: That is an interesting idea to parse this!

I saw that Android also has a `DName` class. I wonder if we can throw the `String` into that class and get the results back:
https://developer.android.com/reference/kotlin/android/net/http/SslCertificate.DName.html?hl=en

kglazko: Oh my gosh that is amazing, I had no idea that existed. Let me try it!

kglazko:
@csadilek @pocmo I added a comment, I couldn't get Sebastian's implementation to work (first AS was complaining and then when I modified it to work, the unit tests failed). Please let me know how I should proceed! I added 3 tests now, please let me know how I should proceed :)

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -742,6 +744,12 @@ class GeckoEngineSession(
         return cookiesPolicies
     }
 
+    internal fun GeckoSession.ProgressDelegate.SecurityInformation.getIssuerName(): String? {
+        return certificate?.issuerDN?.name?.substringAfterLast("O=")?.substringBeforeLast(",C=")
+        // I can't get the below implementation to work correctly
+        // certificate?.issuerDN?.name?.let { SslCertificate.DName(it).getDName() }
```
csadilek: This seems to work using:
```
return SslCertificate(certificate).issuedBy.oName
```

You will need to create an instance of `SslCertificate` first to create an instance of `DName`. Tested and this works in the app, but we need to figure out how to mock this correctly for unit testing.

csadilek: At least with the androidx test runner I can't get a unit test running that depends on `android.net.http.SslCertificate.DName`. I am getting `java.lang.NoClassDefFoundError: com/android/org/bouncycastle/asn1/x509/X509Name`... tried for a while to resolve this without much luck.

So, I am OK landing this simple parser for now which we can cover with tests at least. @pocmo wdyt?

pocmo: With the simple parser I am not sure whether we will be guaranteed to always get the string ordered in the same way with all elements always present. If not that could be causing issues. If we want to land this then maybe we should get another review from the GV team to make sure this is okay.

The other option here is to move this specific test to `androidTest` to run it on an actual device/emulator where those classes exist. We already run other `androidTest` tests for `browser-engine-gekco-nightly` on Firebase (and others hopefully soon too). I think I would prefer that. What do you think?

csadilek: @pocmo and I talked about this: Let's land the PR as is with the simple solution for now. 

Otherwise we have to parse the full SSL certificate every time, which is costly. We can address this in https://github.com/mozilla-mobile/android-components/issues/5557 and also talk to GV team again to see if they can provide a better API for us.

@kglazko There are some lint issues to address and would you mind squashing your commits then we can land this :).

kglazko: Thank you! Sounds great.

csadilek:
bors retry

(network failures)

PR #5917, opened 2020-02-12
Closes #5915: Built-in extensions are disabled by default

### LINKED ISSUE #5915: Built-in extensions are disabled by default
Built-in extensions are currently installed using "registerWebExtension" and don't have metadata when returned from GV. This causes us to mark them as disabled which results in browser/page actions to not be displayed. We should default to true if we don't know the enabled state.

csadilek:
Also, make sure we remove actions from disabled extensions from the toolbar. There's no bug here for add-ons or existing built-in extensions (as the current ones don't have browser actions), but let's fix it.


csadilek:
bors r=gabrielluong

Issue #5930, opened 2020-02-13
Hash-only URL Changes Clear Out Title


darkwing:
While investigating https://github.com/mozilla-mobile/fenix/issues/6826, I've found that the reason the session titles are cleared is due to a URL comparison:

https://github.com/mozilla-mobile/android-components/blob/master/components/browser/session/src/main/java/mozilla/components/browser/session/engine/EngineObserver.kt#L42

The only difference in URL is the second URL has a hash, and thus EngineObserver is clearing the title every time a hash changes, which is a problem.  You can also observe the problem in this demo:

https://lunar-glazer-eoiwdi9ooo.glitch.me/

We may want to make the `if (session.url != url) {` logic more specific, or remove hashes before comparing titles.

PR #5948, opened 2020-02-14
For #5930 - Don't clear title when hash changes

darkwing:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/engine/EngineObserver.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/engine/EngineObserver.kt
@@ -6,6 +6,7 @@ package mozilla.components.browser.session.engine
 
 import android.content.Intent
 import android.graphics.Bitmap
+import android.net.Uri
 import android.os.Environment
 import androidx.core.net.toUri
 import mozilla.components.browser.session.Session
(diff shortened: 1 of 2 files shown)


components/browser/session/src/main/java/mozilla/components/browser/session/engine/EngineObserver.kt
```diff
@@ -64,6 +65,13 @@ internal class EngineObserver(
         }
     }
 
+    private fun isUrlSame(originalUrl: String, newUrl: String): Boolean {
+        val originalUri = URI(originalUrl)
+        val uri = URI(newUrl)
```
pocmo: The constructor of [URI](https://developer.android.com/reference/java/net/URI.html#URI(java.lang.String)) can throw `URISyntaxException`. In other places we use [Uri.parse()](https://developer.android.com/reference/android/net/Uri.html?hl=en#parse(java.lang.String)) which is more forgiving.

components/browser/session/src/main/java/mozilla/components/browser/session/engine/EngineObserver.kt
```diff
@@ -64,6 +65,16 @@ internal class EngineObserver(
...
 
+    private fun isUrlSame(originalUrl: String, newUrl: String): Boolean {
+        val originalUri = Uri.parse(originalUrl)
+        val uri = Uri.parse(newUrl)
+
+        return uri.port == originalUri.port &&
+            uri.host == originalUri.host &&
+            uri.path?.trimStart('/') == originalUri.path?.trimStart('/') &&
+            uri.query == originalUri.query
+    }
```
pocmo: Eventually we could consider landing this as an extension method on `Uri` in `support-ktx`. But no need to do that now. :)

PR #5949, opened 2020-02-14
Closes #5944: Allow addons to specify how to open options_ui

### LINKED ISSUE #5944: Addons should be able to specify if settings page should open in tab or popup
We're not using this flag provided by GeckoView yet. Let's bring it in and use it in sample browser, and also in Fenix after checking back with UX. Addon settings aren't guaranteed to work in popups e.g. uBlock currently also wants a regular tab, but we're ignoring it and put it in a popup based on latest UX.

csadilek:


samples/browser/src/main/java/org/mozilla/samples/browser/addons/InstalledAddonDetailsActivity.kt
```diff
@@ -93,13 +94,21 @@ class InstalledAddonDetailsActivity : AppCompatActivity() {
...
-            intent.putExtra("add_on", addOn)
-            this.startActivity(intent)
+            if (addon.installedState?.openOptionsPageInTab == true) {
+                components.tabsUseCases.addTab(optionsPageUrl as String)
+                val intent = Intent(this, BrowserActivity::class.java)
+                intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP)
+                this.startActivity(intent)
+            } else {
+                val intent = Intent(this, AddonSettingsActivity::class.java)
+                intent.putExtra("add_on", addon)
```
psymoon: I think we should send `optionsPageUrl` and `title` here instead of sending add-on parcel (I think there was a discussion over slack about add-on parcel size being too large). Don't start this activity unless `addon.installedState?.openOptionsPageInTab` exists (maybe log it or show indication if null?). 

Otherwise, we may show a blank page in `AddonSettingsActivity` and this may get flagged as a bug :). What do you think?

csadilek: - We've fixed that add-on parcel size already. That was just because we had all translations in there as well.

> Don't start this activity unless addon.installedState?.openOptionsPageUrl exists (maybe log it or show indication if null?).
That can't happen as we only enable the button if it exists. That's why I made it nullable, or did I miss anything?

csadilek:
bors r=Amejia481,psymoon

PR #5953, opened 2020-02-14
Closes #6027 Add support for video autoplay permissions

### LINKED ISSUE #6027: Add support for video autoplay permissions
GeckoView exposed AUTOPLAY_AUDIBLE and AUTOPLAY_INAUDIBLE.  Update feature-sitepermissions to handle them.

Baron-Severin:
For Fenix#8411: fix crash in SitePermissionsFeature#updateSitePermissionsStatus

---

@Amejia481 I've spent a few hours trying to reproduce this crash without any luck.  Hopefully this fixes the crash (and enables https://github.com/mozilla-mobile/fenix/issues/5636), but at the very least it looks unlikely to break anything.

### Pull Request checklist

I don't think this one needs a changelog entry?  Happy to add one if I'm wrong.

### After merge

### DIFF
--- a/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/permission/GeckoPermissionRequest.kt
+++ b/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/permission/GeckoPermissionRequest.kt
@@ -17,6 +17,8 @@ import org.mozilla.geckoview.GeckoSession.PermissionDelegate.MediaSource.SOURCE_
 import org.mozilla.geckoview.GeckoSession.PermissionDelegate.MediaSource.SOURCE_MICROPHONE
 import org.mozilla.geckoview.GeckoSession.PermissionDelegate.MediaSource.SOURCE_OTHER
 import org.mozilla.geckoview.GeckoSession.PermissionDelegate.MediaSource.SOURCE_SCREEN
+import org.mozilla.geckoview.GeckoSession.PermissionDelegate.PERMISSION_AUTOPLAY_AUDIBLE
+import org.mozilla.geckoview.GeckoSession.PermissionDelegate.PERMISSION_AUTOPLAY_INAUDIBLE
 import org.mozilla.geckoview.GeckoSession.PermissionDelegate.PERMISSION_DESKTOP_NOTIFICATION
 import org.mozilla.geckoview.GeckoSession.PermissionDelegate.PERMISSION_GEOLOCATION
 
@@ -50,7 +52,9 @@ sealed class GeckoPermissionRequest constructor(
         companion object {
             val permissionsMap = mapOf(
                 PERMISSION_DESKTOP_NOTIFICATION to Permission.ContentNotification(),
-                PERMISSION_GEOLOCATION to Permission.ContentGeoLocation()
+                PERMISSION_GEOLOCATION to Permission.ContentGeoLocation(),
+                PERMISSION_AUTOPLAY_AUDIBLE to Permission.ContentAutoPlayAudible(),
+                PERMISSION_AUTOPLAY_INAUDIBLE to Permission.ContentAutoPlayInaudible()
             )
         }
     }
--- a/components/browser/engine-gecko/src/test/java/mozilla/components/browser/engine/gecko/permission/GeckoPermissionRequestTest.kt
+++ b/components/browser/engine-gecko/src/test/java/mozilla/components/browser/engine/gecko/permission/GeckoPermissionRequestTest.kt
@@ -16,6 +16,8 @@ import org.junit.runner.RunWith
 import org.mockito.Mockito.verify
 import org.mozilla.geckoview.GeckoSession
 import org.mozilla.geckoview.GeckoSession.PermissionDelegate.MediaSource
+import org.mozilla.geckoview.GeckoSession.PermissionDelegate.PERMISSION_AUTOPLAY_AUDIBLE
+import org.mozilla.geckoview.GeckoSession.PermissionDelegate.PERMISSION_AUTOPLAY_INAUDIBLE
 import org.mozilla.geckoview.GeckoSession.PermissionDelegate.PERMISSION_DESKTOP_NOTIFICATION
 import org.mozilla.geckoview.GeckoSession.PermissionDelegate.PERMISSION_GEOLOCATION
 
@@ -34,6 +36,14 @@ class GeckoPermissionRequestTest {
         request = GeckoPermissionRequest.Content(uri, PERMISSION_GEOLOCATION, callback)
         assertEquals(uri, request.uri)
         assertEquals(listOf(Permission.ContentGeoLocation()), request.permissions)
+
+        request = GeckoPermissionRequest.Content(uri, PERMISSION_AUTOPLAY_AUDIBLE, callback)
+        assertEquals(uri, request.uri)
+        assertEquals(listOf(Permission.ContentAutoPlayAudible()), request.permissions)
+
+        request = GeckoPermissionRequest.Content(uri, PERMISSION_AUTOPLAY_INAUDIBLE, callback)
+        assertEquals(uri, request.uri)
+        assertEquals(listOf(Permission.ContentAutoPlayInaudible()), request.permissions)
 
         request = GeckoPermissionRequest.Content(uri, 1234, callback)
         assertEquals(uri, request.uri)
--- a/components/feature/session/src/test/java/mozilla/components/feature/session/SettingsUseCasesTest.kt
+++ b/components/feature/session/src/test/java/mozilla/components/feature/session/SettingsUseCasesTest.kt
@@ -41,16 +41,16 @@ class SettingsUseCasesTest {
 
     @Test
     fun `UpdateSettingUseCase will update all sessions`() {
-        val autoplaySetting = object : UpdateSettingUseCase<Boolean>(settings, sessionManager) {
+        val allowFileAccessSetting = object : UpdateSettingUseCase<Boolean>(settings, sessionManager) {
             override fun update(settings: Settings, value: Boolean) {
-                settings.allowAutoplayMedia = value
+                settings.allowFileAccess = value
             }
         }
 
-        autoplaySetting(true)
-        verify(settings).allowAutoplayMedia = true
-        verify(engineSessionA.settings).allowAutoplayMedia = true
-        verify(engineSessionB.settings).allowAutoplayMedia = true
+        allowFileAccessSetting(true)
+        verify(settings).allowFileAccess = true
+        verify(engineSessionA.settings).allowFileAccess = true
+        verify(engineSessionB.settings).allowFileAccess = true
     }
 
     @Test
--- a/components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
+++ b/components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
@@ -26,6 +26,8 @@ import mozilla.components.browser.session.runWithSessionIdOrSelected
 import mozilla.components.concept.engine.permission.Permission
 import mozilla.components.concept.engine.permission.Permission.ContentAudioCapture
 import mozilla.components.concept.engine.permission.Permission.ContentAudioMicrophone
+import mozilla.components.concept.engine.permission.Permission.ContentAutoPlayAudible
+import mozilla.components.concept.engine.permission.Permission.ContentAutoPlayInaudible
 import mozilla.components.concept.engine.permission.Permission.ContentGeoLocation
 import mozilla.components.concept.engine.permission.Permission.ContentNotification
 import mozilla.components.concept.engine.permission.Permission.ContentVideoCamera
@@ -229,7 +231,8 @@ class SitePermissionsFeature(
             storage.findSitePermissionsBy(request.host)
         }
 
-        val prompt = if (shouldApplyRules(permissionFromStorage)) {
+        val prompt = if (shouldApplyRules(permissionFromStorage) ||
+            request.isForAutoplay()) {
             handleRuledFlow(request, session)
         } else {
             handleNoRuledFlow(permissionFromStorage, request, session)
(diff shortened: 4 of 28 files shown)


components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/db/SitePermissionsEntity.kt
```diff
@@ -37,6 +37,12 @@ internal data class SitePermissionsEntity(
     @ColumnInfo(name = "local_storage")
     var localStorage: SitePermissions.Status,
 
+    @ColumnInfo(name = "autoplay_audible")
+    var autoplayAudible: SitePermissions.Status,
+
+    @ColumnInfo(name = "autoplay_inaudible")
+    var autoplayInaudible: SitePermissions.Status,
```
Baron-Severin: IIUC, these should default to 0 (`NO_DECISION`), so we _shouldn't_ need to migrate anything?

Amejia481: Agree!

Amejia481:
I almost forget to mention, to automatically allow or deny auto play we need to add it to the [SitePermissionsRules](https://github.com/mozilla-mobile/android-components/blob/1ae746a6e85e55c3324c1d2c2f22999e46440999/components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsRules.kt#L19) class

Amejia481:
Related  https://github.com/mozilla-mobile/fenix/issues/7507#issuecomment-587432919

components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/permission/GeckoPermissionRequest.kt
```diff
@@ -50,7 +52,9 @@ sealed class GeckoPermissionRequest constructor(
         companion object {
             val permissionsMap = mapOf(
                 PERMISSION_DESKTOP_NOTIFICATION to Permission.ContentNotification(),
-                PERMISSION_GEOLOCATION to Permission.ContentGeoLocation()
+                PERMISSION_GEOLOCATION to Permission.ContentGeoLocation(),
```
Amejia481: We need this change also in nightly and beta :)

components/browser/engine-gecko/src/test/java/mozilla/components/browser/engine/gecko/permission/GeckoPermissionRequestTest.kt
```diff
@@ -35,6 +37,14 @@ class GeckoPermissionRequestTest {
         assertEquals(uri, request.uri)
         assertEquals(listOf(Permission.ContentGeoLocation()), request.permissions)
 
+        request = GeckoPermissionRequest.Content(uri, PERMISSION_AUTOPLAY_AUDIBLE, callback)
```
Amejia481: Let's add this test for the remaining variants (nightly and beta) :)

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsRules.kt
```diff
@@ -12,21 +12,53 @@ import mozilla.components.feature.sitepermissions.SitePermissionsRules.Action.BL
...
+            ALLOWED -> SitePermissions.Status.ALLOWED
             BLOCKED -> SitePermissions.Status.BLOCKED
             ASK_TO_ALLOW -> SitePermissions.Status.NO_DECISION
         }
     }
 
+    /**
+     * Autoplay requests will never prompt the user
+     */
+    enum class AutoplayAction {
```
Amejia481:
```
  /**
     * Autoplay requests will never prompt the user
     */
```
I think can keep autoplay as an `Action`as in a future follow up we want that autoplay shows an UI in the toolbar instead of a prompt

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
```diff
@@ -229,10 +231,9 @@ class SitePermissionsFeature(
             storage.findSitePermissionsBy(request.host)
         }
 
-        val prompt = if (shouldApplyRules(permissionFromStorage)) {
```
Amejia481: I was think about simpler solution wdyt? `shouldApplyRules` can stay as it is now and just verify that the request is for autoplay as for now we always want to handle autoplay with global rules.

```kotlin
        val prompt = if (request.isForAutoplay() || shouldApplyRules(permissionFromStorage)) {
            handleRuledFlow(request, session)
        } else {
            handleNoRuledFlow(permissionFromStorage, request, session)
        }
```

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
```diff
@@ -265,10 +266,15 @@ class SitePermissionsFeature(
 
     private fun handleRuledFlow(
         permissionRequest: PermissionRequest,
```
Amejia481: The function remain as it was, just adding an extra validation for when we have an autoplay request and we don't have `sitePermissionsRules` .
 
```kotlin
    private fun handleRuledFlow(
            permissionRequest: PermissionRequest,
            session: Session
    ): SitePermissionsDialogFragment? {
        //  For now we only support autoplay via sitePermissionsRules until we add support  for autoplay for specific sites see issue  ###
        return if (permissionRequest.isForAutoplay() && sitePermissionsRules == null) {
            permissionRequest.reject()
            session.contentPermissionRequest.consume { true }
            null
        } else {
            val action = requireNotNull(sitePermissionsRules).getActionFrom(permissionRequest)
            when (action) {
                SitePermissionsRules.Action.ALLOWED -> {
                    permissionRequest.grant()
                    session.contentPermissionRequest.consume { true }
                    null
                }
                SitePermissionsRules.Action.BLOCKED -> {
                    permissionRequest.reject()
                    session.contentPermissionRequest.consume { true }
                    null
                }
                SitePermissionsRules.Action.ASK_TO_ALLOW -> {
                    createPrompt(permissionRequest, session)
                }
            }
        }
    }
```

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
```diff
@@ -265,10 +266,15 @@ class SitePermissionsFeature(
...
-        session: Session
+        session: Session,
+        sitePermissionsRules: SitePermissionsRules
     ): SitePermissionsDialogFragment? {
-        val action = requireNotNull(sitePermissionsRules).getActionFrom(permissionRequest)
+        val action = sitePermissionsRules.getActionFrom(permissionRequest)
         return when (action) {
+            SitePermissionsRules.Action.ALLOWED -> {
+                permissionRequest.grant()
```
Amejia481: We need to consume the request after granting   `session.contentPermissionRequest.consume { true }`

components/feature/sitepermissions/src/test/java/mozilla/components/feature/sitepermissions/SitePermissionsFeatureTest.kt
```diff
@@ -285,6 +289,75 @@ class SitePermissionsFeatureTest {
...
+            location = SitePermissionsRules.Action.ALLOWED,
+            camera = SitePermissionsRules.Action.ALLOWED,
+            notification = SitePermissionsRules.Action.ALLOWED,
+            microphone = SitePermissionsRules.Action.ALLOWED,
+            autoplayAudible = SitePermissionsRules.Action.ALLOWED,
+            autoplayInaudible = SitePermissionsRules.Action.ALLOWED
+        )
+
+        doReturn(siteRules).`when`(mockStorage).findSitePermissionsBy(anyString())
+        sitePermissionFeature.sitePermissionsRules = defaultRules
```
Amejia481: 👍  Additionally we could also add a case where `sitePermissionsRules` are not provided but `SitePermissions` are

components/lib/crash/docs/metrics.md
```diff
@@ -11,11 +11,8 @@ This means you might have to go searching through the dependency tree to get a f
 
 
 ## metrics
-
```
Amejia481: I think we can ignore the `*.md` for the change set

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
Amejia481: Also it will be good to add a reference to ticket where we are going to add support for specific sites  autoplay request

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsRules.kt
Baron-Severin: I don't think we can here, because even once we're showing autoplay in the UI, it won't be a prompt.  If it's anything like desktop, it'll be an icon that the user can press instead of a dialog.

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
Baron-Severin: I feel like the control flow is becoming pretty confusing here, and it'll stay that way even after we make the eventual autoplay changes.  But we're short of time and we've probably already spent too much on this PR, so I'll take your advice and keep it close to what's in master.

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsRules.kt
Baron-Severin: Although this could probably be renamed to be more general.  It's an action that will never prompt the user, what would be a good name for that?

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
Amejia481: After we add support for adding autoplay settings per site, we can delete the part below:

```kotlin
        return if (permissionRequest.isForAutoplay() && sitePermissionsRules == null) {
            permissionRequest.reject()
            session.contentPermissionRequest.consume { true }
            null
        } else {
```
As autoplay is going to behave the same way as other permissions, with the only difference in `ASK_TO_ALLOW`

```kotlin
                SitePermissionsRules.Action.ASK_TO_ALLOW -> {
                   if(!permissionRequest.isForAutoplay()){
                       createPrompt(permissionRequest, session)
                    }else{
                      notifyAutoplaydelagate()  // TBD  
                      null
                  }   
                }
```

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsRules.kt
Amejia481: What I meant was that `SitePermissionsRules.ASK_TO_ALLOW`  is lower level API is doesn't imply how we are going to ask to allow(`ASK_TO_ALLOW`), it just says that we need to ask, this will be the job of `SitePermissionsFeature` as it is a higher level (closer to the UI) it will determine which permissions should prompt and which not.

Amejia481: I'm OK to leave as it is I just want to minimize the change and reuse the concepts that we already have.

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
```diff
@@ -267,15 +270,28 @@ class SitePermissionsFeature(
...
-            }
-            SitePermissionsRules.Action.ASK_TO_ALLOW -> {
-                createPrompt(permissionRequest, session)
+        // For now we only support autoplay via sitePermissionsRules until we add support for
+        // autoplay for specific sites see Fenix issue  #8603
+        return if (permissionRequest.isForAutoplay() && sitePermissionsRules == null) {
+            permissionRequest.reject()
+            session.contentPermissionRequest.consume { true }
+            null
+        } else {
```
Amejia481: After we add support for adding autoplay settings per site, we can delete this part :)

components/feature/session/src/test/java/mozilla/components/feature/session/SettingsUseCasesTest.kt
```diff
@@ -41,16 +41,16 @@ class SettingsUseCasesTest {
...
-                settings.allowAutoplayMedia = value
+                settings.allowFileAccess = value
             }
         }
 
-        autoplaySetting(true)
-        verify(settings).allowAutoplayMedia = true
-        verify(engineSessionA.settings).allowAutoplayMedia = true
-        verify(engineSessionB.settings).allowAutoplayMedia = true
+        allowFileAccessSetting(true)
```
Amejia481: I'm not quite aware of the context of this change, could you help me to clarify?

Baron-Severin: It's possible that I'm misunderstanding, but based on the name of the test (`UpdateSettingUseCase will update all sessions`) it appeared that this test was setting an arbitrary property and ensuring that it was updated on all sessions.  Instead of removing the test, it seemed more appropriate to switch it over to a new property.

Amejia481: Got it, what confused me was that most of the test was related to autoplay 👍

liuche:
@Baron-Severin @Amejia481 so is this something that will land today? I'd like to get it landed at least some time early this week to pick up into a later Beta build, for migration, but let me know how that looks.

Baron-Severin:
> @Baron-Severin @Amejia481 so is this something that will land today? I'd like to get it landed at least some time early this week to pick up into a later Beta build, for migration, but let me know how that looks.

We plan to land it this morning.

Baron-Severin:
bors r=Amejia481

PR #5954, opened 2020-02-15
Fenix 5236 - Fix search alignment

darkwing:
Per https://github.com/mozilla-mobile/fenix/issues/5236, we need to move the alignment of the search suggestion title -1dp if there's no description.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/awesomebar/src/main/java/mozilla/components/browser/awesomebar/layout/DefaultSuggestionViewHolder.kt
+++ b/components/browser/awesomebar/src/main/java/mozilla/components/browser/awesomebar/layout/DefaultSuggestionViewHolder.kt
@@ -7,6 +7,7 @@ package mozilla.components.browser.awesomebar.layout
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import android.view.ViewGroup.MarginLayoutParams
 import android.widget.ImageView
 import android.widget.TextView
 import mozilla.components.browser.awesomebar.BrowserAwesomeBar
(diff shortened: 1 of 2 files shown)


components/browser/awesomebar/src/main/java/mozilla/components/browser/awesomebar/layout/DefaultSuggestionViewHolder.kt
```diff
@@ -42,6 +43,10 @@ internal sealed class DefaultSuggestionViewHolder {
             } else {
                 descriptionView.visibility = View.VISIBLE
                 descriptionView.text = suggestion.description
+
+                val titleParams = titleView.layoutParams as MarginLayoutParams
+                titleParams.setMargins(0, 0, 0, 0)
+                titleView.requestLayout()
```
pocmo: I'm not really sure whys this part is needed. Doesn't that remove all margin? 🤔

darkwing: Apologies, I come from the web world -- is there a better way to only modify the top margin?

hakkikaancaliskan: You mean something like this? https://stackoverflow.com/a/50765990

pocmo: @darkwing Only modifying top margin is not directly possible but you can set the value of the other items to the currently assigned value (e.g. `setMargins(leftMargin, [..])`). Or we set it explicitly to 8dp like in the XML?

What's important here is that the views get recycled (or reused) if they are no longer needed, which means we  need to make sure that the if/else cases are symmetric and we potentially reset all values. So we should make sure to set the margin in both cases!

PR #5976, opened 2020-02-18
Improve add-ons update flow.

Amejia481:
Both the simple flow (with NO new permissions) and the complex update flow with new permissions work as expected, I did some minor changes as I noticed that we didn't  remove from a previously allowed add-on from the status store after It was processed, also Geckoview requires us to process the updates in the main dispatcher.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/addons/src/main/java/mozilla/components/feature/addons/update/AddonUpdater.kt
+++ b/components/feature/addons/src/main/java/mozilla/components/feature/addons/update/AddonUpdater.kt
@@ -24,6 +24,7 @@ import androidx.work.PeriodicWorkRequest
 import androidx.work.PeriodicWorkRequestBuilder
 import androidx.work.WorkManager
 import androidx.work.WorkerParameters
+import kotlinx.coroutines.Dispatchers
 import mozilla.components.concept.engine.webextension.WebExtension
 import mozilla.components.feature.addons.Addon
 import mozilla.components.feature.addons.R
@@ -122,6 +123,7 @@ class DefaultAddonUpdater(
     private val applicationContext: Context,
     private val frequency: Frequency = Frequency(1, TimeUnit.DAYS)
 ) : AddonUpdater {
+    private val logger = Logger("DefaultAddonUpdater")
 
     @VisibleForTesting
     internal val updateStatusStorage = UpdateStatusStorage()
@@ -135,6 +137,7 @@ class DefaultAddonUpdater(
             ExistingPeriodicWorkPolicy.KEEP,
             createPeriodicWorkerRequest(addonId)
         )
+        logger.info("registerForFutureUpdates $addonId")
     }
 
     /**
@@ -143,6 +146,7 @@ class DefaultAddonUpdater(
     override fun unregisterForFutureUpdates(addonId: String) {
         WorkManager.getInstance(applicationContext)
             .cancelUniqueWork(getUniquePeriodicWorkName(addonId))
+        logger.info("unregisterForFutureUpdates $addonId")
     }
 
     /**
@@ -154,6 +158,7 @@ class DefaultAddonUpdater(
             ExistingWorkPolicy.KEEP,
             createImmediateWorkerRequest(addonId)
         ).enqueue()
+        logger.info("update $addonId")
     }
 
     /**
@@ -165,6 +170,8 @@ class DefaultAddonUpdater(
         newPermissions: List<String>,
         onPermissionsGranted: (Boolean) -> Unit
     ) {
+        logger.info("onUpdatePermissionRequest $current")
+
         val wasPreviouslyAllowed =
             updateStatusStorage.isPreviouslyAllowed(applicationContext, updated.id)
 
@@ -172,6 +179,8 @@ class DefaultAddonUpdater(
 
         if (!wasPreviouslyAllowed) {
             createNotification(updated, newPermissions)
+        } else {
+            updateStatusStorage.markAsUnallowed(applicationContext, updated.id)
         }
     }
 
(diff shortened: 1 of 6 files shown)


components/feature/addons/src/main/java/mozilla/components/feature/addons/update/AddonUpdater.kt
```diff
@@ -413,6 +422,16 @@ class DefaultAddonUpdater(
                 .apply()
         }
 
+        @Synchronized
+        fun remove(context: Context, addonId: String) {
```
psymoon: nit: Let's make it consistent with method name here - `unmarkAsAllowed()`?

samples/browser/src/geckoNightly/java/org/mozilla/samples/browser/Components.kt
```diff
@@ -19,7 +19,7 @@ import org.mozilla.geckoview.GeckoRuntimeSettings
 class Components(private val applicationContext: Context) : DefaultComponents(applicationContext) {
     private val runtime by lazy {
         // Allow for exfiltrating Gecko metrics through the Glean SDK.
-        val builder = GeckoRuntimeSettings.Builder()
+        val builder = GeckoRuntimeSettings.Builder().aboutConfigEnabled(true)
```
psymoon: Maybe a stupid question - why didn't we want `about:config` allowed before?

components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
```diff
@@ -70,7 +70,8 @@ class AddonManager(
                 .map { addon -> addon.filterTranslations(locales) }
                 .map { addon ->
                     installedExtensions[addon.id]?.let {
-                        addon.copy(installedState = it.toInstalledState())
+                        addon.copy(installedState = it.toInstalledState(),
```
psymoon: nit: assign a val to `it.toInstalledState()`

components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
```diff
@@ -70,7 +70,8 @@ class AddonManager(
...
+                                version = it.toInstalledState().version.ifEmpty { addon.version })
```
psymoon: So is this not where we update version after updating?

samples/browser/src/geckoNightly/java/org/mozilla/samples/browser/Components.kt
Amejia481: I'm not sure 😄,  maybe no body need it before.

components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
Amejia481: As we are not re-querying the AMO endpoints after update, we are using the info that Gecko has about the installed add-on, this way we are always going to have the latest version without without hitting the network.

components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
Amejia481: 👍

components/feature/addons/src/main/java/mozilla/components/feature/addons/update/AddonUpdater.kt
Amejia481: 👍  let's use `unmarkAsAllowed`

components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
psymoon: Okay cool, makes sense to me 👍

components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
```diff
@@ -70,7 +70,9 @@ class AddonManager(
                 .map { addon -> addon.filterTranslations(locales) }
                 .map { addon ->
                     installedExtensions[addon.id]?.let {
-                        addon.copy(installedState = it.toInstalledState())
+                        val installedState = it.toInstalledState()
+                        addon.copy(installedState = installedState,
+                                version = installedState.version.ifEmpty { addon.version })
```
csadilek: As discussed, I think it's better to keep these separate, the latest add-on version as provided by or AMO collection and the installed version. This will allow us to debug problems, or have telemetry, and also makes it easier to support manual updates later.

components/feature/addons/src/main/java/mozilla/components/feature/addons/update/AddonUpdater.kt
```diff
@@ -407,10 +416,14 @@ class DefaultAddonUpdater(
...
             allowSet.add(addonId)
-            getSettings(context)
-                .edit()
-                .putStringSet(KEY_ALLOWED_SET, allowSet)
-                .apply()
+            setData(context, allowSet)
+        }
+
+        @Synchronized
+        fun unmarkAsAllowed(context: Context, addonId: String) {
```
csadilek: nit: markAsUnallowed?

components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
csadilek: So, in other words, we don't need to do this and just use the `installedState.version` in the UI if it exists.

Amejia481:
bors r=csadilek,psymoon

PR #5984, opened 2020-02-18
Closes #3729: allow tokenserver url override

### LINKED ISSUE #3729: Allow specifying tokenserverUrl in ServerConfig
This is a follow-up to https://github.com/mozilla-mobile/android-components/pull/3579.

Currently, the only way to use a self-hosted sync server is by specifying a custom FxA endpoint via ServerConfig (which will then be used during tokenserver url configuration internally). However, self-hosting FxA server currently is not well documented, and generally quite experimental.

A much more common path - and this is what both Firefox Desktop and Firefox for Android currently support - is to specify a tokenserver URL. Token server in our current architecture is what's responsible for pointing the client to a specific storage node. In case of self-hosting, both a sync storage server and a token server are easily deployed via available docker containers. 

By allowing specifying tokenserver url as part of the ServerConfig passed into the account manager, we're enabling easy use of self-hosted sync infrastructure.

Actual work here should be pretty trivial - making sure we override tokenserver url stored within the `SyncAuthInfoCache`.

tublitzed:
The first step in supporting self hosted sync for Fenix is allowing the tokenserver url to be passed into the FxaAccountManager. This allows users to override the default tokenserver url, via Sync Config.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/Config.kt
+++ b/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/Config.kt
@@ -47,10 +47,12 @@ data class DeviceConfig(
  * @property supportedEngines A set of supported sync engines, exposed via [GlobalSyncableStoreProvider].
  * @property syncPeriodInMinutes Optional, how frequently periodic sync should happen. If this is `null`,
  * periodic syncing will be disabled.
+ * @property tokenServerUrl Optional, override default tokenServerUrl to support self-hosted sync storage.
  */
 data class SyncConfig(
     val supportedEngines: Set<SyncEngine>,
-    val syncPeriodInMinutes: Long? = null
+    val syncPeriodInMinutes: Long? = null,
+    val tokenServerUrl: String? = null
 )
 
 /**
--- a/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
+++ b/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
@@ -908,10 +908,16 @@ open class FxaAccountManager(
             return
         }
 
+        var tokenServerUrl = syncConfig?.tokenServerUrl
+        if (tokenServerUrl == null) {
+            tokenServerUrl = account.getTokenServerEndpointURL()
+        }
+
+        logger.info("caching tokenServerUrl: $tokenServerUrl")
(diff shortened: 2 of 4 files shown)


components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/Config.kt
```diff
@@ -47,10 +47,12 @@ data class DeviceConfig(
  * @property supportedEngines A set of supported sync engines, exposed via [GlobalSyncableStoreProvider].
  * @property syncPeriodInMinutes Optional, how frequently periodic sync should happen. If this is `null`,
  * periodic syncing will be disabled.
+ * @property tokenServerUrl Optional, override default tokenServerUrl to support self-hosted sync storage.
  */
 data class SyncConfig(
```
tublitzed: @grigoryk - So it seems that the ServerConfig (which was the suggestion on where to overwrite this value in the original [issue](https://github.com/mozilla-mobile/android-components/issues/3729)) is pulled from [here](https://github.com/mozilla/application-services/blob/master/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/Config.kt) in application services and doesn't currently support setting a custom token server url.

Since #3729 was filed specifically for this repo, I wanted to first confirm with you that you still think it makes sense for the ServerConfig to set this value. If so, I can close this PR and first open a PR over in application services to support setting a token server url. Then, once that merges, I can re-open this one and adjust to use that value instead.

Since that's a much larger scope of changes, I figured it'd make sense to confirm this approach with you first.

tublitzed:
It's also worth noting here that I think we may also need to make additional changes to support [#3762](https://github.com/mozilla-mobile/fenix/issues/3762). I'm able to verify here that I _can_ successfully overwrite the tokenserver url used for sync....but cannot get a successful sync after doing so. When setting to local sync server(http://localhost:5000/token/1.0/sync/1.5), I get a network error (is this a known issue with the android studio emulator and localhost?) and when setting to staging sync server (https://token.stage.mozaws.net/1.0/sync/1.5) I get a 401. I'm not sure if there's something else that's happening in desktop when you overwrite the token server url via `identity.sync.tokenserver.uri` in `about:config`...still need to dig into that. I think we might be missing an extra step here though. Flagging this here in case anyone has ideas as to what that might be. Otherwise, I'll keep digging.

tublitzed:
Hey, @grigoryk can I get a r+ here?

tublitzed:
Thanks for the tips, @grigoryk! App services PR coming next.

PR #5985, opened 2020-02-18
Closes #5983: QR scanner slow and fails to scan on some devices

### LINKED ISSUE #5983: QR scanner slow and fails to scan on some devices
This is caused by distorted images as we're setting the aspect ratio of the texture view. Easiest solution right now is to make the scan area a square.

csadilek:
We're making the scan area a square which we found to work much better on various devices (esp. one plus 6T) when testing. Also limiting the maximum resolution to get faster scan results.


csadilek:
bors r=Amejia481

PR #5987, opened 2020-02-18
Closes #5911: Update test coverage for feature-addons/ui

### LINKED ISSUE #5911: Increase unit test coverage in feature-addons/ui
The UI package in feature-addons doesn't have sufficient coverage. 

In particular, these have zero coverage:
` AddonsManagerAdapterDelegate`
`CustomViewHolder`
` Extensions`
` AddonPermissionsAdapter`

psymoon:
Add more coverage for:
- `AddonsManagerAdapterDelegate`
- `CustomViewHolder`
- `Extensions`

I don't think we need test coverage for `AddonPermissionsAdapter` since it's mostly android `RecyclerView` logic.

### DIFF
--- a/components/feature/addons/src/test/java/mozilla/components/feature/addons/ui/AddonsManagerAdapterTest.kt
+++ b/components/feature/addons/src/test/java/mozilla/components/feature/addons/ui/AddonsManagerAdapterTest.kt
@@ -4,7 +4,9 @@
 
 package mozilla.components.feature.addons.amo.mozilla.components.feature.addons.ui
 
+import android.view.View
 import android.widget.ImageView
+import android.widget.TextView
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.runBlocking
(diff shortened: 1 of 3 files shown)


components/feature/addons/src/test/java/mozilla/components/feature/addons/ui/AddonsManagerAdapterTest.kt
```diff
@@ -105,4 +111,92 @@ class AddonsManagerAdapterTest {
             }
         }
     }
+
+    @Test
+    fun bindAddon() {
```
Amejia481: nit: to keep consistency, I think we should use the backtick(\`bind add-on`\) function names

components/feature/addons/src/test/java/mozilla/components/feature/addons/ui/ExtensionsTest.kt
```diff
@@ -30,4 +31,18 @@ class ExtensionsTest {
 
         assertEquals("Hello", map.translate())
     }
+
```
csadilek: Let's add a test for Addon.translatedName as well as it's easy it shouldn't remain untested. :)

components/feature/addons/src/test/java/mozilla/components/feature/addons/ui/AddonsManagerAdapterTest.kt
```diff
@@ -105,4 +111,92 @@ class AddonsManagerAdapterTest {
...
+            id = "id",
+            authors = emptyList(),
+            categories = emptyList(),
+            downloadUrl = "downloadUrl",
+            version = "version",
+            permissions = emptyList(),
+            rating = Addon.Rating(4.5f, 1000),
+            createdAt = "",
+            updatedAt = "",
+            translatableName = mapOf("en-US" to "name", "de" to "Name", "es" to "nombre"),
```
Amejia481: nit: Should we also test when we don't get the translatable fields like `translatableName` and `translatableSummary` as they have a different paths.

psymoon:
bors r=csadilek, Amejia481

PR #5997, opened 2020-02-19
Closes #5792: Potential deadlock between unsubscribe and dispatch

### LINKED ISSUE #5792: Subscription.unsubscribe() <-> Store.dispatchInternal() can potentially deadlock
We think that in theory it is possible that `Subscription.unsubscribe()` and `Store.dispatchInternal()` can deadlock.

Locks:
* `unsubscribe()` is synchronized on the `Subscription` object and calling into `removeSubscription()` which needs the `subscriptions` list as a lock.
* `dispatchInternal` is synchronized on the `subscriptions` list and calls into `Subscription.dispatch()` which requires a lock on the `Subscription` object.

Threads:
* `unsubscribe()` is likely to get called from the main thread after a lifecycle change
* `dispatchInternal` is called on the internal thread of the `Store`

With the right timing it seems to be possible that every thread has only one lock and requires the other lock - essentially deadlocking both threads.

csadilek:
As with our previous deadlock fix, we do **not** want to change the lock on the `Subscription`, as it guarantees we never dispatch to paused or stopped observers. However, the inner lock on `subscriptions` is only there for visibility purposes. We can achieve the same, namely making sure retrievals reflect the results of the most recently completed update, by using a concurrent HashSet: A set projection of https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html

This removes the potential of a deadlock and I don't see any change in semantics: The `removeSubscription` case is guarded by the `Subscription` lock and can't overlap with dispatching to the same subscription. The `addSubscription` case is guarded by the lock on the `Store` which is also required for `dipsatching`, so add and dispatch can't overlap either.

@pocmo wdyt?

### DIFF
--- a/components/lib/state/src/main/java/mozilla/components/lib/state/Store.kt
+++ b/components/lib/state/src/main/java/mozilla/components/lib/state/Store.kt
@@ -13,6 +13,8 @@ import kotlinx.coroutines.asCoroutineDispatcher
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.launch
 import java.lang.ref.WeakReference
+import java.util.Collections
+import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.Executors
 
 /**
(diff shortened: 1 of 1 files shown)


csadilek:
> In theory the `Store` could still keep a reference to a `Subscription` after `removeSubscription()` if we are looping over them (for dispatch) and we keeping a reference to a "stable" iterator. But I think this should only be very briefly, so negligible.

Yes, it also shouldn't matter, as the subscription is no longer active and dispatch wouldn't do anything.

>Iterators being "stable" in ConcurrentHashMap hopefully works nicely with Kotlin's forEach() ... if they loop manually then this may not work, right? But looking at the code this seems to be just a regular for (x in y) loop which should use the iterator.

Yes, that's the way I understand it too.

csadilek:
bors r=pocmo

PR #6034, opened 2020-02-21
Closes #6011: Integrate DebuggerDelegate to support temporary extensions

### LINKED ISSUE #6011: Integrate WebExtensionController.DebuggerDelegate to support temporary extensions
We need to know when a temporary extension was installed so we can query the list of installed extensions again and make sure we have action handlers hooked up.

GV: https://bugzilla.mozilla.org/show_bug.cgi?id=1614295

csadilek:
GeckoView calls us back now when a debug / temporary extension was installed so we can refresh the list of installed extensions to make sure action handlers are hooked up. 

We're gonna have to uninstall all extensions and call `listInstalledExtensions` again to re-install everything that is there now incl. the new temporary extension. This is fine as it's for debugging purposes only.

### DIFF
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
@@ -332,8 +332,15 @@ class GeckoEngine(
             }
         }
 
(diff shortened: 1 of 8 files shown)


csadilek:
Marking as "blocked" because we still need to verify running web-ext before landing.

csadilek:
OK, tested and verified web-ext run works with this. Ready to review/land.

csadilek:
bors r=Amejia481

Issue #6030, opened 2020-02-21
Improve add-ons icon transition in AddonsManagerAdapter


Amejia481:
When we change the icons too quickly is causes flickering, for more info see related issue https://github.com/mozilla-mobile/fenix/issues/8111

csadilek:
Closed by: https://github.com/mozilla-mobile/android-components/pull/6031

PR #6050, opened 2020-02-24
Closes issue #6030: Set the default add-on icon when failing to fetch it

Amejia481:
Fixes https://github.com/mozilla-mobile/fenix/issues/8111#issuecomment-590220306 and https://github.com/mozilla-mobile/fenix/issues/8698

### DIFF
--- a/components/feature/addons/src/main/res/layout/mozac_feature_addons_item.xml
+++ b/components/feature/addons/src/main/res/layout/mozac_feature_addons_item.xml
@@ -28,8 +28,8 @@
                 android:layout_width="32dp"
                 android:layout_height="32dp"
                 android:layout_gravity="center"
-                android:src="@drawable/mozac_ic_extensions"
                 android:importantForAccessibility="no"
+                android:src="@android:color/transparent"
                 android:scaleType="fitCenter" />
 
     </androidx.cardview.widget.CardView>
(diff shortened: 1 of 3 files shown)


Amejia481:
bors r=csadilek

components/feature/addons/src/main/res/layout/mozac_feature_addons_item.xml
```diff
@@ -28,8 +28,8 @@
                 android:layout_width="32dp"
                 android:layout_height="32dp"
                 android:layout_gravity="center"
-                android:src="@drawable/mozac_ic_extensions"
                 android:importantForAccessibility="no"
+                android:src="@android:color/transparent"
```
psymoon: `@null` works better here I think, not sure if transparent is the best practice for this fwiw :)

Amejia481: Great observation 👍  `@null` is better for performance, but in this case we want to have a `drawable` to transitioning from it to the new icon and if we use `@null` `image.getDrawable() `will [return null](https://stackoverflow.com/a/9311334/5533820) and cause the app to crash [here](https://github.com/mozilla-mobile/android-components/blob/270857e1e0ccf1dcec6bd29611f2ce0763b8aa27/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonsManagerAdapter.kt#L292)

PR #6052, opened 2020-02-24
Closes #6015: Deny installing add-ons if built-in permissions required

### LINKED ISSUE #6015: Installing built-in extensions as external add-ons should fail
We should fail on an attempt to install an external add-on that requires permissions of built-in extensions (e.g. the `geckoViewAddons` permission).

csadilek:
Just a fallback in case add-ons do ever land in our AMO collection with permissions of built-in extensions.


csadilek:
bors r=Amejia481

PR #6057, opened 2020-02-25
Closes #6022: Let GeckoView prompt user when native code crash is non-fatal

### LINKED ISSUE #6022: Non-fatal crash should not crash the app
When testing with `about:crashcontent` it should not crash the app.  Currently both Reference Browser and Fenix crashes when non-fatal crash occurs in GeckoView.

rocketsroger:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/lib/crash/src/main/java/mozilla/components/lib/crash/CrashReporter.kt
+++ b/components/lib/crash/src/main/java/mozilla/components/lib/crash/CrashReporter.kt
@@ -153,13 +153,15 @@ class CrashReporter(
 
         logger.info("Received crash: $crash")
 
-        if (shouldSendIntent(crash)) {
-            // This crash was not fatal and the app has registered a pending intent
-            launchCrashIntent(context, crash)
-        }
-
         if (telemetryServices.isNotEmpty()) {
             sendCrashTelemetry(context, crash)
+        }
+
+        // If crash is native code and non fatal then the view will handle the user prompt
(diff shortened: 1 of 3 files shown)


components/lib/crash/src/main/java/mozilla/components/lib/crash/CrashReporter.kt
```diff
@@ -153,15 +153,17 @@ class CrashReporter(
...
         if (telemetryServices.isNotEmpty()) {
             sendCrashTelemetry(context, crash)
         }
 
+        // If crash is native code and non fatal then the view will handle the user prompt
+        if (crash is Crash.NativeCodeCrash && !crash.isFatal) {
+            // App has registered a pending intent
+            sendNonFatalCrashIntent(context, crash)
+            return
+        }
```
pocmo: Why do we switch to always doing this? Previously we did this only when we had an intent and that kind of made sense? Currently we explicit allow apps to not provide an Intent if they do not intent to handle crashes in their UI. 🤔

rocketsroger: > Why do we switch to always doing this? Previously we did this only when we had an intent and that kind of made sense? Currently we explicit allow apps to not provide an Intent if they do not intent to handle crashes in their UI. 🤔

You're right.  I'll update so if there's no intent it should continue on to report the crash.

Issue #5531, opened 2020-01-09
Measure perf impact of GeckoLoginStorageDelegate#onLoginFetch


severinrudie:
GV `LoginStorage.Delegate#onLoginFetch` requires a synchronous response, but retrieving that information requires unlocking and querying storage backed by a DB.  This performance hit could be mitigated by changing the GV API to expect a Future, or potentially by doing some complex caching in `LoginsStorage`.  Both of these options would require substantial work.  First, we should measure the impact of the current solution (wrap the call in runBlocking) in order to determine 

#### Acceptance Criteria
- [ ] Add a probe to determine time spent in `GeckoLoginStorageDelegate#onLoginFetch`
- [ ] File a follow-up to analyze resulting telemetry and determine whether or not further work is required

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-4708)

ekager:
We removed the `runBlocking` so we are not as concerned about the perf now. Maybe would still be good to add probes, but not urgent

grigoryk:
While this is useful, the impact of these calls is greatly reduced after https://github.com/mozilla-mobile/android-components/pull/6128 - we no longer have an overhead of unlocking the store, or needing to key material.

cpeterso:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1796417

Change performed by the [Move to Bugzilla add-on](https://addons.mozilla.org/en-US/firefox/addon/move-to-bugzilla/).

PR #6059, opened 2020-02-25
#6084 #5531 - Implement onLoginFetch in non-blocking way

ekager:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/autofill/GeckoLoginDelegateWrapper.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/autofill/GeckoLoginDelegateWrapper.kt
@@ -4,6 +4,9 @@
 
 package mozilla.components.browser.engine.gecko.autofill
 
+import kotlinx.coroutines.Dispatchers.IO
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
 import mozilla.components.concept.storage.Login
 import mozilla.components.concept.storage.LoginStorageDelegate
 import org.mozilla.geckoview.GeckoResult
@@ -14,18 +17,27 @@ import org.mozilla.geckoview.LoginStorage
  * them to [storageDelegate]. This allows us to avoid duplicating [LoginStorageDelegate] code
  * between different versions of GeckoView, by duplicating this wrapper instead.
  */
-class GeckoLoginDelegateWrapper(private val storageDelegate: LoginStorageDelegate) : LoginStorage.Delegate {
+class GeckoLoginDelegateWrapper(private val storageDelegate: LoginStorageDelegate) :
+    LoginStorage.Delegate {
+
     override fun onLoginSave(login: LoginStorage.LoginEntry) {
         storageDelegate.onLoginSave(login.toLogin())
     }
 
     override fun onLoginFetch(domain: String): GeckoResult<Array<LoginStorage.LoginEntry>>? {
-        val storedLogins = storageDelegate.onLoginFetch(domain)
+        val result = GeckoResult<Array<LoginStorage.LoginEntry>>()
 
-        return storedLogins
-            .map { it.toLoginEntry() }
-            .toTypedArray()
-            .let { GeckoResult.fromValue(it) }
+        GlobalScope.launch(IO) {
+            val storedLogins = storageDelegate.onLoginFetch(domain)
+
+            val logins = storedLogins.await()
+                .map { it.toLoginEntry() }
+                .toTypedArray()
+
+            result.complete(logins)
+        }
+
+        return result
     }
 
     /**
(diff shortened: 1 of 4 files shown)


components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/autofill/GeckoLoginDelegateWrapper.kt
```diff
@@ -14,18 +16,26 @@ import org.mozilla.geckoview.LoginStorage
...
     override fun onLoginFetch(domain: String): GeckoResult<Array<LoginStorage.LoginEntry>>? {
-        val storedLogins = storageDelegate.onLoginFetch(domain)
+        val result = GeckoResult<Array<LoginStorage.LoginEntry>>()
 
-        return storedLogins
-            .map { it.toLoginEntry() }
-            .toTypedArray()
-            .let { GeckoResult.fromValue(it) }
+        GlobalScope.launch {
+            val storedLogins = storageDelegate.onLoginFetch(domain)
```
Baron-Severin: Nit: we should cancel this job if it is restarted (e.g., someone refreshes in the middle).

```kotlin
var job: Job? = null

override fun onLoginFetch(domain: String): GeckoResult<Array<LoginStorage.LoginEntry>>? {
    ...
    job?.cancel()
    job = GlobalScope.launch {
    ...
}
```

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/autofill/GeckoLoginDelegateWrapper.kt
```diff
@@ -14,18 +16,26 @@ import org.mozilla.geckoview.LoginStorage
...
+
```
Baron-Severin: I remember we tested some edge cases for this, but I can't remember exactly what.  The following are the ones that come to mind.

- Scope is launched -> user backs to home fragment (this one I know we checked
- Scope is launched -> user navigates to a new page
- Scope is launched -> user backgrounds the app
- Scope is launched -> user refreshes the page

If all of these are OK, then I'm fine merging this.

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/autofill/GeckoLoginDelegateWrapper.kt
ekager: This makes sense, yeah!

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/autofill/GeckoLoginDelegateWrapper.kt
ekager: I think all of these are fine, but we should probably also explicitly define a dispatcher?

pocmo: Yeah, I think with us using the storage here it makes sense to explicitly use the IO dispatcher.

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/autofill/GeckoLoginDelegateWrapper.kt
```diff
@@ -14,18 +17,29 @@ import org.mozilla.geckoview.LoginStorage
...
     override fun onLoginSave(login: LoginStorage.LoginEntry) {
         storageDelegate.onLoginSave(login.toLogin())
     }
 
     override fun onLoginFetch(domain: String): GeckoResult<Array<LoginStorage.LoginEntry>>? {
-        val storedLogins = storageDelegate.onLoginFetch(domain)
+        val result = GeckoResult<Array<LoginStorage.LoginEntry>>()
+
+        job?.cancel()
```
pocmo: None of the calls in the `launch` block suspend, so we only cancel if this coroutine has not actually been launched yet. Is that a problem?

In theory if we cancel then the `GeckoResult` we return here will never complete.. Is this a problem? Should we catch a cancellation inside the launch block and make the `GeckoResult` complete? I am not sure if GeckoView expects us to eventually complete it?

components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/GeckoLoginStorageDelegate.kt
```diff
@@ -74,11 +75,9 @@ class GeckoLoginStorageDelegate(
         }
     }
 
-    override fun onLoginFetch(domain: String): List<Login> {
-        if (!isAutofillEnabled.invoke()) return listOf()
-        return runBlocking {
-            // GV expects a synchronous response. Blocking here hasn't caused problems during
-            // testing, but we should add telemetry to verify. See #5531
+    override fun onLoginFetch(domain: String): Deferred<List<Login>> {
+        if (!isAutofillEnabled.invoke()) return CompletableDeferred(listOf())
```
pocmo: nit: This was already the case in the previous code... but I think `isAutofillEnabled()` will be easier to read instead of using `invoke()` - which we can omit here. :)

pocmo:
@ekager I filed the following issue for that in AC: #6084 - Can you add this to your commit message before landing? We are trying to file an issue in AC for every change so that it will be easier to track back why we did certain changes.  :)

pocmo:
Failure in `feature-readerview` is unrelated. Since bors is down I am going to merge this manually.

PR #6060, opened 2020-02-25
Update to application-services 0.52.0. Fixes #5999

### LINKED ISSUE #5999: Update to application-services v0.52.0
We should get the previous update in https://github.com/mozilla-mobile/android-components/pull/5882 landed first, but filing this for tracking since there are quite a few important changes in this release.  I think it would be reasonable for someone from the a-s team to pick this up.

Here's my android-specific summary of the changelog:

Sync:
* Significantly reduced network activity during the normal course of syncing, thanks to:
    * Better caching of the tokenserver token and info/configuration response.
    * Less network requests will be made in the case nothing has changed on the server.

Places:
* Added a new field reasons which is a List of SearchResultReasons in SearchResult.
* Some places import related issues fixed
* **Breaking change**: The PlacesWriterConnection.resetHistorySyncMetadata and PlacesWriterConnection.resetBookmarkSyncMetadata methods have been moved to the PlacesApi class.

FxA:
* **Breaking change**: A few changes were made in order to decouple device commands from "account events":
        * The AccountEvent enum has been refactored: .TabReceived has been replaced by .IncomingDeviceCommand(IncomingDeviceCommand), IncomingDeviceCommand itself is another enum that contains TabReceived.
        * FirefoxAccount.pollDeviceCommands now returns an array of IncomingDeviceCommand.
* FirefoxAccount.handlePushMessage now handles all possible FxA push payloads and will return new AccountEvents:
        * .ProfileUpdated which should be handled by fetching the newest profile.
        * .AccountAuthStateChanged should be handled by checking if the authentication state is still valid.
        * .AccountDestroyed should be handled by removing the account information (no need to call FirefoxAccount.disconnect) from the device.
        * .DeviceConnected can be handled by showing a " is connected to this account" notification.
        * .DeviceDisconnected should be handled by showing a "re-auth" state to the user if isLocalDevice is true. There is no need to call FirefoxAccount.disconnect as it will fail.

* Added a deviceId property to the AccountEvent.deviceDisconnected enum case.
* Added context=oauth_webchannel_v1 in getManageDevicesURL methods for WebChannel redirect URLs.

Push:
* Android: Exposed GeneralError to the Kotlin layer.

mhammond:
---

cc @grigoryk and @eoger - I'm sure I've done something wrong here and have missed a few steps, and I'm possibly even on the wrong branch! Note that the PR which caused most of these changes is [here](https://github.com/mozilla/application-services/pull/2522)

### Pull Request checklist

### After merge


components/concept/sync/src/main/java/mozilla/components/concept/sync/Devices.kt
```diff
@@ -87,12 +87,12 @@ interface DeviceConstellation : Observable<DeviceEventsObserver> {
...
     /**
-     * Polls for any pending [DeviceEvent] events.
-     * In case of new events, registered [DeviceEventsObserver] observers will be notified.
+     * Polls for any pending [DeviceCommandIncoming] commands.
+     * In case of new commands, registered [AccountEventsObserver] observers will be notified.
      *
      * @return A [Deferred] that will be resolved with a success flag once operation is complete.
      */
-    fun pollForEventsAsync(): Deferred<Boolean>
+    fun pollForCommandsAsync(): Deferred<Boolean>
```
eoger: Nice, we also called this `pollForCommands` in the Swift account manager.

components/concept/sync/src/main/java/mozilla/components/concept/sync/OAuthAccount.kt
```diff
@@ -293,6 +293,8 @@ sealed class AuthType {
 
 /**
  * Observer interface which lets its users monitor account state changes and major events.
+ * (XXX - there's some tension between this and the
+ * mozilla.components.concept.sync.AccountEvent we should resolve!)
```
grigoryk: Can you file & link an issue for this? I don't think we need to block this PR on updating this stuff.

Issue #5996, opened 2020-02-19
[A-C] Update complete notification is not routing to onboarding screen


Sparky93:
1. Update from Fennec to Fenix
2. Place the app in the background
3. Tap on "Update complete" notification

Expected behavior:
Onboarding screen is being shown.

Actual behavior:
Migration progress screen is being shown.

jonalmeida:
@Sparky93 if you're working on this, I'll close the other issue as a duplicate. 🙂

Sparky93:
@jonalmeida Agree, but also I talked with @pocmo and this ticket is being to be closed by his work on [#8488](https://github.com/mozilla-mobile/fenix/issues/8488)

Sparky93:
Sorry @jonalmeida , I misunderstood, both of the tickets are required.

ebalazs-sv:
Verified as fixed on Fennec Beta Migration Build 2, 02/27 with LG G7 fit (Android 8.1), Samsung Galaxy S5 (Android 6.0.1) and tapping the "Update complete" notification displays the onboarding screen.
I will move this to Done. Thanks.

PR #6065, opened 2020-02-25
Fix for #5996 - Update complete notification is not routing to onboarding screen

Sparky93:
### Pull Request checklist

- [x] **Tests**: This code is already covered by tests
- [x] **Changelog**: This PR does not need a changelog entry

### After merge

### DIFF
--- a/components/support/migration/src/main/java/mozilla/components/support/migration/AbstractMigrationService.kt
+++ b/components/support/migration/src/main/java/mozilla/components/support/migration/AbstractMigrationService.kt
@@ -4,10 +4,11 @@
 
 package mozilla.components.support.migration
 
-import android.app.NotificationChannel
+import android.app.Activity
 import android.app.NotificationManager
 import android.app.PendingIntent
 import android.app.Service
+import android.app.NotificationChannel
 import android.content.Context
 import android.content.Intent
 import android.os.Build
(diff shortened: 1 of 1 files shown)


components/support/migration/src/main/java/mozilla/components/support/migration/AbstractMigrationService.kt
```diff
@@ -4,10 +4,7 @@
 
 package mozilla.components.support.migration
 
-import android.app.NotificationChannel
-import android.app.NotificationManager
-import android.app.PendingIntent
-import android.app.Service
+import android.app.*
```
pocmo: This generic import will fail our style checks.

components/support/migration/src/main/java/mozilla/components/support/migration/AbstractMigrationService.kt
```diff
@@ -38,7 +35,7 @@ private const val TEMPORARY_NOTIFICATION_CHANNEL_NAME = "Migration"
 abstract class AbstractMigrationService : Service() {
     protected abstract val migrator: FennecMigrator
     protected abstract val store: MigrationStore
-    protected abstract val migrationActivity: Class<out AbstractMigrationProgressActivity>
+    protected abstract val migrationActivity: Class<out Activity>
```
pocmo: Don't we need two activity references here so that we can route to either the migration activity or the browser/onboarding?

pocmo: Ah, I see you decided to route to the decision activity instead. That should work too.

pocmo: Maybe we should rename this reference though? :)

pocmo:
I'll trigger a snapshot release as soon as bors has merged this. Then we can get the Fenix PR in. :)

pocmo:
Snapshots are triggered!

PR #6111, opened 2020-02-28
Closes #1912: Call speculativeConnect from toolbar and awesomebar

### LINKED ISSUE #1912: Use `speculativeConnect` for URLs we're likely to load after autocompletion
In Fennec we do this as a load time optimization.

csadilek:
This introduces calls to `Engine.speculativeConnect` from our toolbar and awesomebar.

In the toolbar, we now call `speculativeConnect` on successful autocompletion from our domain list or history (re-using the existing logic to discard stale results.)

In the awesomebar, we now call `speculativeConnect` for the highest ranked search suggestion. This worked quite well when I tested this. When typing a search term instead of a URL, `speculativeConnect` will be called with the corresponding search URL of the configured search engine. So, when hitting enter, the URL will have already been "pre-fetched."

This is "inspired by": 
- https://searchfox.org/mozilla-esr68/source/mobile/android/base/java/org/mozilla/gecko/home/BrowserSearch.java#522-525
- https://searchfox.org/mozilla-esr68/source/mobile/android/base/java/org/mozilla/gecko/home/BrowserSearch.java#623-627

@pocmo I've decided to break this apart from the speculative gecko session mechanism, which I'll work on as a follow-up.

### DIFF
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
@@ -465,7 +465,7 @@ private class AsyncAutocompleteDelegate(
     override val coroutineContext: CoroutineContext,
     private val logger: Logger = Logger("AsyncAutocompleteDelegate")
 ) : AutocompleteDelegate, CoroutineScope {
-    override fun applyAutocompleteResult(result: AutocompleteResult) {
+    override fun applyAutocompleteResult(result: AutocompleteResult, onApplied: () -> Unit) {
         // Bail out if we were cancelled already.
         if (!parentScope.isActive) {
             logger.debug("Autocomplete request cancelled. Discarding results.")
@@ -483,6 +483,7 @@ private class AsyncAutocompleteDelegate(
                         totalItems = result.totalItems
                     )
                 )
+                onApplied()
             } else {
                 logger.debug("Discarding stale autocomplete result.")
             }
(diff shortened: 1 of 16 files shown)


csadilek:
I've also tested that the UI performance isn't affected by this. The individual calls to `speculativeConnect` are very fast (can't measure in ms).

csadilek:
> Desktop does not do a speculative connect for the search engine if search suggestions are disabled. I assume we indirectly do that too since we wouldn't use SearchSuggestionProvider in that case?

Yes, exactly, that would work the same way. Apps / Fenix removes the provider if it's disabled.

> Desktop never does a speculative connect in private contexts. At the moment that's something Fenix would need to control. Do we need additional API for that in AC or would we expect them to just not pass in an Engine?

For now let's not add more API. We can handle that in Fenix easily when creating the feature instance, and as you wrote, not pass in the engine for private contexts. Just double-checked that this will work.

> think Fennec seems to do a speculative connect for the first history/bookmark item in the list too? 🤔

Hmm interesting. I can add it to our history provider as well. Do we want to do this now or wait and see how this all pans out? Wdyt?

> Furthermore they at least do this when the URL bar initially gets focused - which seems a good place to do this once? 🤔 Our onInputStarted() callback may be a good place to do that.

Interesting. Let's punt on this one for now until we understand how effective the connects are while typing and address in a follow-up.

csadilek:
@pocmo this is ready for another pass. I've applied the change to the history, bookmarks, and clipboard providers as well.

csadilek:
Fenix PR: https://github.com/mozilla-mobile/fenix/pull/8909/

PR #6122, opened 2020-02-28
Fixes #6094: Check if we are able to open url 

### LINKED ISSUE #6094: Crash on handling schemes other than http/https
Fenix issue/crash [here](https://github.com/mozilla-mobile/fenix/issues/8412).

The issue is that we are not handling other schemes than http and https. Basically if a user wants to access `blob:https://mega.nz/...` (like seen in a recent sentry report) the app would not know how to handle this and crash.

ValentinTimisica:
There are cases when a user opens url with scheme that we are not handling in
the app so we need to check if we can resolve them before opening.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
+++ b/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
@@ -19,7 +19,6 @@ import mozilla.components.browser.state.selector.findCustomTab
 import mozilla.components.browser.state.state.CustomTabConfig
 import mozilla.components.browser.state.store.BrowserStore
 import mozilla.components.concept.engine.window.WindowRequest
-import mozilla.components.lib.crash.CrashReporter
 import mozilla.components.lib.state.ext.flowScoped
 import mozilla.components.support.base.feature.LifecycleAwareFeature
 import mozilla.components.support.ktx.kotlinx.coroutines.flow.ifChanged
@@ -33,7 +32,7 @@ class CustomTabWindowFeature(
     private val activity: Activity,
     private val store: BrowserStore,
     private val sessionId: String,
-    private val crashReporter: CrashReporter? = null
+    @VisibleForTesting internal val handleError: (Exception) -> Unit
 ) : LifecycleAwareFeature {
 
     private var scope: CoroutineScope? = null
@@ -76,10 +75,15 @@ class CustomTabWindowFeature(
                     val windowRequest = state.content.windowRequest
                     if (windowRequest?.type == WindowRequest.Type.OPEN) {
                         val intent = configToIntent(state.config)
+                        // This could only fail if the above intent is for our application
+                        // and we are not registered to handle its schemes.
+                        // Let's log this to better asses how often this happens in real world and
+                        // if we need to add new schemes to properly support this workflow.
+                        // See Fenix #8412
                         try {
                             intent.launchUrl(activity, windowRequest.url.toUri())
                         } catch (e: ActivityNotFoundException) {
-                            crashReporter?.submitCaughtException(e)
+                            handleError(e)
                         }
                         store.dispatch(ContentAction.ConsumeWindowRequestAction(sessionId))
                     }
--- a/samples/browser/src/main/java/org/mozilla/samples/browser/ExternalAppBrowserFragment.kt
+++ b/samples/browser/src/main/java/org/mozilla/samples/browser/ExternalAppBrowserFragment.kt
@@ -68,7 +68,9 @@ class ExternalAppBrowserFragment : BaseBrowserFragment(), UserInteractionHandler
             requireActivity(),
             components.store,
             sessionId!!
-        )
+        ) {
+            // No-op. Client may override this
+        }
         lifecycle.addObserver(windowFeature)
 
         if (manifest != null) {
--- a/components/feature/customtabs/build.gradle
+++ b/components/feature/customtabs/build.gradle
@@ -39,7 +39,6 @@ dependencies {
     implementation project(':support-ktx')
     implementation project(':support-utils')
     implementation project(':ui-icons')
-    implementation project(':lib-crash')
 
     implementation Dependencies.androidx_core_ktx
     implementation Dependencies.kotlin_stdlib
--- a/components/feature/customtabs/src/test/java/mozilla/components/feature/customtabs/CustomTabWindowFeatureTest.kt
+++ b/components/feature/customtabs/src/test/java/mozilla/components/feature/customtabs/CustomTabWindowFeatureTest.kt
@@ -17,7 +17,6 @@ import mozilla.components.browser.state.state.CustomTabMenuItem
 import mozilla.components.browser.state.state.createCustomTab
 import mozilla.components.browser.state.store.BrowserStore
 import mozilla.components.concept.engine.window.WindowRequest
-import mozilla.components.lib.crash.CrashReporter
 import mozilla.components.support.test.any
 import mozilla.components.support.test.ext.joinBlocking
 import mozilla.components.support.test.mock
@@ -31,6 +30,8 @@ import org.junit.runner.RunWith
 import org.mockito.Mockito.never
 import org.mockito.Mockito.spy
 import org.mockito.Mockito.verify
+import org.mockito.Mockito.verifyZeroInteractions
+import java.lang.Exception
 
 @RunWith(AndroidJUnit4::class)
 class CustomTabWindowFeatureTest {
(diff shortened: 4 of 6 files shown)


components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
```diff
@@ -72,13 +73,23 @@ class CustomTabWindowFeature(
...
+                            handleError()
+                        }
                     }
                 }
         }
     }
 
+    @VisibleForTesting
+    internal fun canHandleIntent(customTabIntent: CustomTabsIntent): Boolean =
+        customTabIntent.intent.resolveActivity(activity.packageManager) != null
```
pocmo: I _think_ `resolveActivity()` will also return activities that may have `exported=false` set and cannot actually be launched? We may need to call `resolveActivity()` on the `PackageManager` and check whether the returned `ActivityInfo` has the exported flag set.

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
```diff
@@ -72,13 +73,23 @@ class CustomTabWindowFeature(
                     val windowRequest = state.content.windowRequest
                     if (windowRequest?.type == WindowRequest.Type.OPEN) {
                         val intent = configToIntent(state.config)
-                        intent.launchUrl(activity, windowRequest.url.toUri())
-                        store.dispatch(ContentAction.ConsumeWindowRequestAction(sessionId))
+                        // Setting url for intent in order to check if we are able to open that url
+                        intent.intent.data = windowRequest.url.toUri()
+                        if (canHandleIntent(intent)) {
+                            intent.launchUrl(activity, windowRequest.url.toUri())
+                            store.dispatch(ContentAction.ConsumeWindowRequestAction(sessionId))
```
pocmo: I assume we want to consume the window request in any case? 🤔

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
```diff
@@ -72,13 +73,23 @@ class CustomTabWindowFeature(
                     val windowRequest = state.content.windowRequest
                     if (windowRequest?.type == WindowRequest.Type.OPEN) {
                         val intent = configToIntent(state.config)
-                        intent.launchUrl(activity, windowRequest.url.toUri())
-                        store.dispatch(ContentAction.ConsumeWindowRequestAction(sessionId))
+                        // Setting url for intent in order to check if we are able to open that url
+                        intent.intent.data = windowRequest.url.toUri()
+                        if (canHandleIntent(intent)) {
+                            intent.launchUrl(activity, windowRequest.url.toUri())
```
pocmo: nit: We are calling `windowRequest.url.toUri()` twice.

samples/browser/src/main/java/org/mozilla/samples/browser/ExternalAppBrowserFragment.kt
```diff
@@ -68,7 +68,9 @@ class ExternalAppBrowserFragment : BaseBrowserFragment(), UserInteractionHandler
             requireActivity(),
             components.store,
             sessionId!!
-        )
+        ) {
+            // No-op. Client may override this
+        }
```
pocmo: Isn't that already the default behavior? Or do you just want to make this explicit here so that people reading the sample code will know about this?

ValentinTimisica: This piece of code is for a lambda parameter that I added and will be a breaking change for our clients who will now have to provide a specific behaviour for handling the error.

ValentinTimisica:
Check comments from #6374.

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
```diff
@@ -76,17 +76,36 @@ class CustomTabWindowFeature(
...
+                        if (canHandleIntent(intent)) {
+                            intent.launchUrl(activity, url)
+                        } else {
+                            // We could only get here if the above intent is for our application
+                            // and we are not registered to handle its schemes.
+                            // Let's log this to better asses how often this happens in real world and
+                            // if we need to add new schemes to properly support this workflow.
+                            // See Fenix #8412
+                            try {
+                                intent.launchUrl(activity, url)
```
rocketsroger: Looks like both `canHandleIntent(activity, url)` true and false case both calls `intent.launchUrl()`.  Why do the extra check?

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
```diff
@@ -33,7 +33,7 @@ class CustomTabWindowFeature(
     private val activity: Activity,
     private val store: BrowserStore,
     private val sessionId: String,
-    private val crashReporter: CrashReporter? = null
+    @VisibleForTesting internal val handleError: (Exception) -> Unit
```
rocketsroger: I don't know if this is what we want as a fix.  This failure to handle the link/intent is an A-C issue.

ValentinTimisica: I have found the issue and I have steps to reproduce in the Fenix issue - https://github.com/mozilla-mobile/fenix/issues/8412#issuecomment-592020541.

Seems like this is a Fenix issue and for fixing it we have to add new intent filter schemes.
Any touch on a link gets to AC which will try to open a client activity (in this case Fenix) and in our case Fenix doesn't know how to handle specific link schemes.

Because at the moment we can't easily add new intent filter schemes without a product decision the best we can do is:
1) Show a message to the user that his touch event on a specific link cannot be handled by the app (UX suggestion, you can check it here https://github.com/mozilla-mobile/fenix/pull/8903#issuecomment-595875216).
2) Log the error to our crashstats to better assess how many users are affected by this. (solution inspired by your patch).

Because the issue actually is in the AC client (it doesn't handle all schemes) AC will simply pass this error to the client and it will be its responsibility to handle this case. We on Fenix will use both of the above two solutions:
https://github.com/mozilla-mobile/fenix/blob/fdecc03b19b3f6b46874bafe14cb85fd278ee45a/app/src/main/java/org/mozilla/fenix/customtabs/ExternalAppBrowserFragment.kt#L87

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
ValentinTimisica: When `canHandleIntent` returns:
- true: 
We know that the client can open the link and we simply delegate it to do so.
- false:
We know that the client can't handle the link but we try to give to it a stacktrace with even more information about why the call to open the link failed.
In this case the `intent.launchUrl(activity, url)` it's clear that it will not actually open the url, but just throw an exception which we'll forward to the client.
To give a bit more context about this I wrote the comment above the `try-catch`.

rocketsroger: I see.  But couldn't we (like the change I landed) just call `launchUrl` in the try catch and handle the error?

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
rocketsroger: This is interesting.  I think this makes sense.

components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
ValentinTimisica: I just found it to be more clear and I already had the patch and tests for that scenario. I think the difference is minor.

rocketsroger: Agreed.  I would preferred we go straight to try catch without the extra check.  Thanks,

ValentinTimisica: Hello @rocketsroger, I agree with you that we should only keep the `try catch` without the extra check and I updated the PR accordingly. Thanks!

rocketsroger:
> Looks good. Can you please squash the changes together?
> https://github.com/wprig/wprig/wiki/How-to-squash-commits

Looks clean enough.  Probably don't need to squash.  I'll approve.

rocketsroger:
bors r=rocketsroger

PR #6151, opened 2020-03-02
Closes #6018: Do not load url during link if session is restored

### LINKED ISSUE #6018: Do not always load Session URL when EngineSession gets linked
When linking a `Session` with an `EngineSession` we perform a load of the `Session.url`:
https://github.com/mozilla-mobile/android-components/blob/34227ba6c29f6af3a2bb356f7a5d1ccf788254ef/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt#L46-L52

This makes sense in situations where we created a `Session` with a URL and once we lazily create the `EngineSession`. Now we actually need to load the URL.

However this behavior is wrong in situations where we restore the state:
https://github.com/mozilla-mobile/android-components/blob/34227ba6c29f6af3a2bb356f7a5d1ccf788254ef/components/browser/session/src/main/java/mozilla/components/browser/session/LegacySessionManager.kt#L275-L281

In this situation we already restored the last state and then in `link()` will load the last URL again. Not only does this do an unnecessary fresh load, it also pollutes the back history of the session.

psymoon:
The potential pitfall of this patch is we are assuming `geckoSession.restoreState()` would always be successful. Perhaps we could have a follow-up for GV to provide us a callback for when `restoreState()` is completed with a status? 🤔

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/LegacySessionManager.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/LegacySessionManager.kt
@@ -273,20 +273,21 @@ class LegacySessionManager(
         getEngineSession(session)?.let { return it }
 
         return engine.createSession(session.private).apply {
+            var restored = false
             session.engineSessionHolder.engineSessionState?.let { state ->
-                restoreState(state)
+                restored = restoreState(state)
                 session.engineSessionHolder.engineSessionState = null
             }
 
-            link(session, this)
+            link(session, this, restored)
         }
     }
 
-    fun link(session: Session, engineSession: EngineSession) {
+    private fun link(session: Session, engineSession: EngineSession, restored: Boolean = false) {
         val parent = values.find { it.id == session.parentId }?.let {
             this.getEngineSession(it)
         }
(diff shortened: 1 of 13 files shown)


components/browser/session/src/main/java/mozilla/components/browser/session/LegacySessionManager.kt
```diff
@@ -273,20 +273,21 @@ class LegacySessionManager(
...
         return engine.createSession(session.private).apply {
+            var restored = false
             session.engineSessionHolder.engineSessionState?.let { state ->
-                restoreState(state)
+                restored = restoreState(state)
                 session.engineSessionHolder.engineSessionState = null
             }
 
-            link(session, this)
+            link(session, this, restored)
```
pocmo: Ah, interesting, you return whether the restore was successful or not. Initially I thought the engine session you keep a flag internally after restore() was called (meaning "I am a session that was restored") which we then could use to decide whether we need to call `load()`. But I think your implementation makes more sense!

psymoon:
bors r=pocmo, csadilek

PR #6153, opened 2020-03-02
Closes #6145: Do not invoke close callback on start

### LINKED ISSUE #6145: Tabs tray closes when tab switcher is pressed and no tabs open
Related: https://github.com/mozilla-mobile/reference-browser/issues/1033

## STR:
1. Open Sample or Reference Browser.
2. Close all tabs.
3. Toggle between the browser view and the tabs tray view.

## Expected
 - Toggling works seamlessly without any non-user interacted changes.

## Actual
 - Clicking on the menu, flashes the tabs tray and switches back to the browser view.

jonalmeida:
In the tabs tray Flow, `ifChanged` is invoked because the initial state starts with `null` and that in turn invokes the close tabs callback.

We also need to pass the initial filter of tabs from the feature to satisfy changes on the app's side that we need to use here.

---

Closes #6145.

Related app changes: https://github.com/mozilla-mobile/reference-browser/pull/1039

### Pull Request checklist

### After merge
- [x] **Breaking Changes**: If this is a breaking change, please push a draft PR on Reference Browser to address the breaking issues.
  - **See:** https://github.com/mozilla-mobile/reference-browser/pull/1039

### DIFF
--- a/components/feature/tabs/src/main/java/mozilla/components/feature/tabs/tabstray/TabsFeature.kt
+++ b/components/feature/tabs/src/main/java/mozilla/components/feature/tabs/tabstray/TabsFeature.kt
@@ -14,17 +14,22 @@ import mozilla.components.support.base.feature.LifecycleAwareFeature
 
 /**
  * Feature implementation for connecting a tabs tray implementation with the session module.
+ *
+ * @param defaultTabsFilter A tab filter that is used for the initial presenting of tabs that will be used by
+ * [TabsFeature.filterTabs] by default as well.
  */
 class TabsFeature(
     tabsTray: TabsTray,
     private val store: BrowserStore,
     tabsUseCases: TabsUseCases,
+    private val defaultTabsFilter: (TabSessionState) -> Boolean = { true },
     closeTabsTray: () -> Unit
 ) : LifecycleAwareFeature {
     @VisibleForTesting
     internal var presenter = TabsTrayPresenter(
         tabsTray,
         store,
+        defaultTabsFilter,
         closeTabsTray
     )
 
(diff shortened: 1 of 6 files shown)


components/feature/tabs/src/main/java/mozilla/components/feature/tabs/tabstray/TabsFeature.kt
```diff
@@ -14,18 +14,23 @@ import mozilla.components.support.base.feature.LifecycleAwareFeature
...
+    private val defaultTabsFilter: (TabSessionState) -> Boolean = { true },
     closeTabsTray: () -> Unit
 ) : LifecycleAwareFeature {
     @VisibleForTesting
     internal var presenter = TabsTrayPresenter(
         tabsTray,
         store,
-        closeTabsTray
+        closeTabsTray,
+        defaultTabsFilter
```
pocmo: What is a bit weird about this API is that I can provide a filter and it will not be used until I also call `filterTabs()`. Instead of having a default parameter in `filterTabs()` can we setup the presenter to use it?

jonalmeida: We're passing in the filter to the presenter because we need it for the initial state of the tab filter which we weren't doing before. So in the tabs tray on the app, we would present all the tabs, and then the app would invoke `filterTabs` to get only the non-private ones as we do [in r-b](https://github.com/mozilla-mobile/reference-browser/pull/1039/files#diff-145bf304b2e445da6fb25020ce03c650R37).

The idea behind the `filterTabs` is that it can use the same filter passed in from init or pass a new filter in.

I do agree this API is weird though. I'll try to look for something better than this.

jonalmeida: Looking into this more, I think we're doing the right thing here. What makes this diff look weird is that we had the same default param on the presenter and we're adding the same one on the feature, but we really should have **removed it** from the presenter when we put it on the feature - I'm making those changes now.

jonalmeida:
bors r=rocketsroger

PR #6165, opened 2020-03-03
Closes #6162: Provide API to speculatively create engine session

### LINKED ISSUE #6162: Speculatively create Gecko session to start content process early
Since we now call `speculativeConnect` from the toolbar and awesombar we can use it as a hook to create a gecko session as well.

See discussion in https://github.com/mozilla-mobile/fenix/issues/8291

csadilek:
See https://github.com/mozilla-mobile/android-components/issues/6162#issuecomment-594093161.

Another advantage of decoupling this from `speculativeConnect` is that we can also handle private sessions. Applied to all 3 engine variants.

### DIFF
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
@@ -6,6 +6,7 @@ package mozilla.components.browser.engine.gecko
 
 import android.content.Context
 import android.util.AttributeSet
+import androidx.annotation.VisibleForTesting
 import mozilla.components.browser.engine.gecko.integration.LocaleSettingUpdater
 import mozilla.components.browser.engine.gecko.mediaquery.from
 import mozilla.components.browser.engine.gecko.mediaquery.toGeckoValue
@@ -35,6 +36,7 @@ import mozilla.components.concept.engine.webextension.WebExtensionRuntime
 import mozilla.components.concept.engine.webnotifications.WebNotificationDelegate
 import mozilla.components.concept.engine.webpush.WebPushDelegate
 import mozilla.components.concept.engine.webpush.WebPushHandler
+import mozilla.components.support.utils.ThreadUtils
 import org.json.JSONObject
 import org.mozilla.geckoview.AllowOrDeny
 import org.mozilla.geckoview.ContentBlocking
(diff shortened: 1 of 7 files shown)


csadilek:
Fenix PR: https://github.com/mozilla-mobile/fenix/pull/8955

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
```diff
@@ -146,6 +155,16 @@ class GeckoEngine(
         return GeckoEngineSessionState.fromJSON(json)
     }
 
+    /**
+     * See [Engine.speculativeCreateSession].
+     */
+    override fun speculativeCreateSession(private: Boolean) {
+        val speculativeSession = this.speculativeSession
+        if (speculativeSession == null || speculativeSession.geckoSession.settings.usePrivateMode != private) {
+            this.speculativeSession = GeckoEngineSession(runtime, private, defaultSettings)
```
pocmo: In the case where `usePrivateMode != private` we drop the previous `speculativeSession`. Internally `GeckoEngineSession` calls `open()` on the `GeckoSession`. Wouldn't we need to call `close()` here to avoid leaking that session since we never told Gecko to close it?

pocmo: Related to that I wonder if, since both methods are not syncrhonized, we could end up in a situation where we create two sessions in a race and one gets leaked and never closed? Same concern for `createSession()`: Could we end up in a situation where we, due to a race, end up using the same `GeckoEngineSession` twice? Of course all that depends on the calling thread - if it is always `main` then I guess it doesn't matter.

csadilek: > Wouldn't we need to call close() here to avoid leaking that session since we never told Gecko to close it?

Ah, yes, good catch! I'll update in a bit.

> ... if it is always main then I guess it doesn't matter.

Yes, we really want to be on the main thread for these methods. Let's add an assert on main thread for `speculativeCreateSession` to prevent weirdness. This way we can also get rid of `volatile` which wouldn't be sufficient.

csadilek: OK, updated.

csadilek:
bors r=pocmo

pocmo:
bors retry

PR #6170, opened 2020-03-04
On memory pressure close engine sessions and keep state for restoring later.

### LINKED ISSUE #5933: SessionManager never closes opened sessions, leading to OOMs
See [Fenix #4151](https://github.com/mozilla-mobile/fenix/issues/4151)
See attempted fix PR https://github.com/mozilla-mobile/android-components/pull/5919 (with [several problems](https://github.com/mozilla-mobile/android-components/pull/5919#pullrequestreview-358056532))

Any `EngineSession`s that are actively opened (I _**believe**_ the only way to 'open' an `EngineSession` is to select and load its tab during the current user session) will never be closed.  If too many are opened (on my device ~20) this will cause an OOM crash.  I haven't measured, but presumably there is a perf impact before that point is reached.

It's on GVs roadmap to handle this memory pressure intelligently, but it hasn't been prioritized.  For now, they recommend that we open/close `EngineSession`s ourselves.
[Riot conversation](https://matrix.to/#/!kgzqJPKsBZFakDVGxV:mozilla.org/$FRkgsypkk5TWgICW28a25EIAacze0uBUyDI_e_zdmrw?via=mozilla.org&via=matrix.org)
[BZ1](https://bugzilla.mozilla.org/show_bug.cgi?id=1587853)
[BZ2](https://bugzilla.mozilla.org/show_bug.cgi?id=1454752)

A very basic solution here could be:
- `SessionManager#onLowMemory` closes all `EngineSession`s except for the currently selected one.
- When a new tab is selected, more than MAX_SESSIONS are open, close the LRU `EngineSession`

pocmo:
Unfortunately the implementation for this needs to live in SessionManager as long as it keeps
references to EngineSession instances and thumbnails. Therefore we determine what to trim in
SessionManager and notify BrowserStore to perform the same changes.

Hopefully in a not to distant future we can move that to BrowserStore.

Closes #5933.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
@@ -4,6 +4,7 @@
 
 package mozilla.components.browser.session
 
+import android.content.ComponentCallbacks2
 import mozilla.components.browser.session.engine.EngineObserver
 import mozilla.components.browser.session.ext.syncDispatch
 import mozilla.components.browser.session.ext.toCustomTabSessionState
@@ -18,6 +19,7 @@ import mozilla.components.browser.state.store.BrowserStore
 import mozilla.components.concept.engine.Engine
 import mozilla.components.concept.engine.EngineSession
 import mozilla.components.concept.engine.EngineSessionState
+import mozilla.components.support.base.memory.MemoryConsumer
 import mozilla.components.support.base.observer.Observable
 import java.lang.IllegalArgumentException
 
@@ -29,7 +31,7 @@ class SessionManager(
     val engine: Engine,
     private val store: BrowserStore? = null,
     private val delegate: LegacySessionManager = LegacySessionManager(engine, EngineSessionLinker(store))
-) : Observable<SessionManager.Observer> by delegate {
+) : Observable<SessionManager.Observer> by delegate, MemoryConsumer {
 
     /**
      * This class only exists for migrating from browser-session
@@ -318,9 +320,16 @@ class SessionManager(
      * Informs this [SessionManager] that the OS is in low memory condition so it
      * can reduce its allocated objects.
      */
+    @Deprecated("Use onTrimMemory instead.", replaceWith = ReplaceWith("onTrimMemory"))
     fun onLowMemory() {
         delegate.onLowMemory()
         store?.syncDispatch(SystemAction.LowMemoryAction)
+    }
+
+    override fun onTrimMemory(level: Int) {
+        if (level >= ComponentCallbacks2.TRIM_MEMORY_RUNNING_LOW) {
+            delegate.onLowMemory()
+        }
     }
 
     companion object {
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
@@ -19,6 +19,7 @@ import mozilla.components.browser.state.store.BrowserStore
 import mozilla.components.concept.engine.Engine
 import mozilla.components.concept.engine.EngineSession
 import mozilla.components.concept.engine.EngineSessionState
+import mozilla.components.support.base.log.logger.Logger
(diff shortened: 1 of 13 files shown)


components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
```diff
@@ -62,17 +67,22 @@ class SessionManager(
...
+                engineSession.unregister(observer)
             }
 
+            session.engineSessionHolder.engineSession = null
+            session.engineSessionHolder.engineObserver = null
+            session.engineSessionHolder.engineSessionState = null
+
             store?.syncDispatch(UnlinkEngineSessionAction(session.id))
+
+            // Closing the engine session after updating the store so that neither session manager
```
csadilek: nit: I misread this comment a few times: Should this be "Closing the engine session after updating *now* that neither the session manager nor the store keep a reference anymore."?

components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
```diff
@@ -318,9 +328,55 @@ class SessionManager(
...
+                        it.engineSessionHolder.engineSessionState = state
+                        states[it.id] = state
+                    }
+
+                    unlinkedEngineSessions++
+                }
+            }
+        }
+
+        logger.debug("Cleared $clearThumbnails thumbnail(s) and unlinked $unlinkedEngineSessions engine sessions()")
```
csadilek: This should be $clearedThumbnails (not the $clearThumbnails boolean).

csadilek: Also nit: don't need to end sentence in `()` :)

components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
```diff
@@ -318,9 +328,55 @@ class SessionManager(
...
+                }
+
+                if (closeEngineSessions) {
+                    val state = it.engineSessionHolder.engineSession?.saveState()
+                    linker.unlink(it)
+
+                    if (state != null) {
+                        it.engineSessionHolder.engineSessionState = state
+                        states[it.id] = state
+                    }
```
csadilek: We're doing this on `TRIM_MEMORY_RUNNING_CRITICAL,` but presumably `saveState` is also requiring some memory. Just wondering if we might fail to do this because there's not enough memory left in which case `TRIM_MEMORY_RUNNING_LOW` would be better for clearing engine sessions as well? 🤔 Of course, that would happen more often then. Just a thought. +1 for landing as is.

components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
pocmo: @jonalmeida doesn't like me to end my sentences with `.` - now you are telling me `()` is also not an option?? :)

pocmo: (That should have been `session(s)` - plurals, yay!)

jonalmeida: Only commit messages! 😅

components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
pocmo: You are right in theory, but: The implementation of `saveState` has changed in the past. Nowadays `GeckoEngineSession` always has a reference to the last state (since GV pushes it to us) and `saveState()` only returns that reference :)

components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
csadilek: 😆

pocmo:
Updated!

bors r=csadilek

Issue #4965, opened 2019-11-05
browser.tabs.remove support is limited to tabs opened by web extension


csadilek:
We currently have this limitation (a web extension can only close tabs it has opened), because we have no way of accessing all available engine sessions from within the engine. This will change once we finish our `browser-state` migration and once the engine has access to the browser store.

csadilek:
This can be addressed with changes coming to GV next week. We will have to attach the tab delegate per session / per extension, and therefore also know which session/tab should be closed without looking it up again.

csadilek:
Since we need the `TabDelegate API` to work in both Nightly and Beta, I am pushing this to next week (post merge day). If we refactor now, it wouldn't work in Beta and we'd have to somehow keep two different solutions in `support-webextensions` that will work with both Beta and Nightly. That's not worth it for a couple of days.

PR #6172, opened 2020-03-04
Manual GV Nightly upgrade to 75.0.20200304084140

csadilek:
Fixes breaking changes for web ext. APIs. I will address the deprecation of the tab delegate later today as part of https://github.com/mozilla-mobile/android-components/issues/4965 which we can fix with this.

### DIFF
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
@@ -256,6 +256,7 @@ class GeckoEngine(
     /**
      * See [Engine.registerWebExtensionDelegate].
      */
+    @Suppress("Deprecation")
     override fun registerWebExtensionDelegate(
         webExtensionDelegate: WebExtensionDelegate
     ) {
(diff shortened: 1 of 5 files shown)


components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
```diff
@@ -256,6 +256,7 @@ class GeckoEngine(
     /**
      * See [Engine.registerWebExtensionDelegate].
      */
+    @Suppress("Deprecation")
```
Amejia481: nit: Should we create an issue to keep track of when this is going to be addressed? e.g         `runtime.webExtensionController.tabDelegate = tabsDelegate`

csadilek: @Amejia481 nope, see my comment above and the linked issue which will address this in this Sprint :)

csadilek: (maybe even today) :)

Amejia481: Awesome 👍

csadilek:
bors r=Amejia481

csadilek:
bors retry

csadilek:
bors retry

PR #6181, opened 2020-03-05
Closes #6180: Update browser-engine-gecko to GeckoView 74 RC2.

### LINKED ISSUE #6180: Update browser-engine-gecko to 74 RC
There's a GeckoView 74 RC build available now and with that we can already update `browser-engine-gecko` before the actual merge day / release happens next week. In this case we want to ship Focus with GV 74, so there's an incentive to update earlier, before merge day.

pocmo:
The actual merge day is next week. But we do already have a stable RC build for GeckoView 74 and therefore would like to update `browser-engine-gecko` already to ship it with Focus. `browser-engine-beta` and `browser-engine-nightly` will be updated next week after merge day.


pocmo:
Thank you.

bors r=jonalmeida

PR #6211, opened 2020-03-09
Closes #5716: Do less during account recovery

### LINKED ISSUE #5716: Do less while recovering from authentication problems
When we recover from authentication problems, we'll currently execute `initDeviceAsync`, `refreshDevicesAsync` and `pollForEventsAsync` - as we're just going through the regular motions of being logged-in.

It's possible that due to currently unknown server-side issues, clients sometimes could get into a loop of failed to sync (401), tried to recover, recovered successfully (got a new access token), repeat. This loop will generate a lot of extra traffic to our FxA servers that we can easily avoid without any real downsides (I think!).

grigoryk:
It is not strictly necessary to re-initialize the device, and refresh device constellation when
we perform an account recovery. This patch removes these operations from the recovery flow,
which helps ease the server load.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
+++ b/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
@@ -812,11 +812,8 @@ open class FxaAccountManager(
                         logger.info("Registering device constellation observer")
                         account.deviceConstellation().register(accountEventsIntegration)
 
-                        logger.info("Initializing device")
-                        // NB: underlying API is expected to 'ensureCapabilities' as part of device initialization.
-                        account.deviceConstellation().initDeviceAsync(
-                                deviceConfig.name, deviceConfig.type, deviceConfig.capabilities
-                        ).await()
+                        // NB: we're running neither `initDevice` nor `ensureCapabilities` here, since
+                        // this is a recovery flow, and these calls already ran at some point before.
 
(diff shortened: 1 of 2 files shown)


components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
```diff
@@ -968,8 +965,8 @@ open class FxaAccountManager(
             type = deviceConfig.type.intoSyncType()
         ))
 
-        // If device supports SEND_TAB...
-        if (deviceConfig.capabilities.contains(DeviceCapability.SEND_TAB)) {
+        // If device supports SEND_TAB, and we're not recovering from an auth problem...
+        if (deviceConfig.capabilities.contains(DeviceCapability.SEND_TAB) && authType != AuthType.Recovered) {
```
jonalmeida: This seems sensible from reading the ticket. Would this cause the device constellation we have to be invalid on genuine recovered states vs. 401 error states?

grigoryk: I don't think so - other devices are unlikely to have changed just because your current one just recovered from a 401. Also - we will refresh the constellation next time user tries to "sync now", and during a startup, so even if it is out of date a bit, that's not for long.

grigoryk:
bors r=jonalmeida

Issue #6215, opened 2020-03-10
Expose media width/height metadata 


ekager:
For https://github.com/mozilla-mobile/fenix/issues/7602

PR #6216, opened 2020-03-10
For #6215 - Expose media height and width

ekager:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/concept/engine/src/main/java/mozilla/components/concept/engine/media/Media.kt
+++ b/components/concept/engine/src/main/java/mozilla/components/concept/engine/media/Media.kt
@@ -197,7 +197,9 @@ abstract class Media(
      * @property duration Indicates the duration of the media in seconds.
      */
     data class Metadata(
-        val duration: Double = -1.0
+        val duration: Double = -1.0,
+        val height: Long = 0L,
+        val width: Long = 0L
     )
 
     /**
--- a/components/concept/engine/src/main/java/mozilla/components/concept/engine/media/Media.kt
+++ b/components/concept/engine/src/main/java/mozilla/components/concept/engine/media/Media.kt
@@ -43,6 +43,11 @@ abstract class Media(
     abstract val metadata: Metadata
 
     /**
+     * The fullscreen info for this media element.
+     */
+    abstract val fullscreen: Boolean
+
+    /**
      * The [Volume] for this media element.
      */
     abstract val volume: Volume
@@ -59,6 +64,8 @@ abstract class Media(
         fun onMetadataChanged(media: Media, metadata: Metadata) = Unit
         /** Notify the observer that media volume changed. */
         fun onVolumeChanged(media: Media, volume: Volume) = Unit
+        /** Notify the observer that media fullscreen changed. */
+        fun onFullscreenChanged(media: Media, fullscreen: Boolean) = Unit
     }
 
     /**
--- a/components/feature/media/src/main/java/mozilla/components/feature/media/ext/MediaState.kt
+++ b/components/feature/media/src/main/java/mozilla/components/feature/media/ext/MediaState.kt
@@ -29,6 +29,26 @@ fun MediaState.getPlayingMediaIdsForTab(tabId: String?): List<String> {
 }
 
 /**
+ * Finds any fullscreen video status and orientation of fullscreen media.
+ */
+internal fun MediaState.findFullscreenMediaOrientation(): MediaState.FullscreenOrientation? {
+    elements.forEach { (_, media) ->
+        val mediaInFullScreen = media.find { it.fullscreen } ?: return null
+        // Don't predict the orientation if we're not in full screen or we don't have media height/width
+        if (mediaInFullScreen.metadata.height != 0L &&
+            mediaInFullScreen.metadata.width != 0L
+        ) {
+            return if (mediaInFullScreen.metadata.height > mediaInFullScreen.metadata.width) {
+                MediaState.FullscreenOrientation.PORTRAIT
+            } else {
+                MediaState.FullscreenOrientation.LANDSCAPE
+            }
+        }
+    }
+    return null
+}
+
+/**
  * Find a [SessionState] with playing [Media] and return this [Pair] or `null` if no such
  * [SessionState] could be found.
  */
@@ -97,8 +117,9 @@ internal fun MediaState.toPlaybackState() =
     PlaybackStateCompat.Builder()
         .setActions(
             PlaybackStateCompat.ACTION_PLAY_PAUSE or
-                PlaybackStateCompat.ACTION_PLAY or
-                PlaybackStateCompat.ACTION_PAUSE)
+                    PlaybackStateCompat.ACTION_PLAY or
+                    PlaybackStateCompat.ACTION_PAUSE
+        )
         .setState(
             when (aggregate.state) {
                 MediaState.State.PLAYING -> PlaybackStateCompat.STATE_PLAYING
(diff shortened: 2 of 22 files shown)


samples/browser/src/main/java/org/mozilla/samples/browser/BrowserFragment.kt
```diff
@@ -80,6 +87,33 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
...
+                    // I tried adding a metadata "fullscreen" to onFullscreenChange
+                    // in Media but it's called after this callback
+                    val media =
+                        components.sessionManager.selectedSessionOrThrow.media.firstOrNull {
+                            it.state == Media.State.PLAYING
+                        }
+                    val height = media?.metadata?.height
+                    val width = media?.metadata?.width
+                    Logger("Height $height Width $width")
+                    activity?.enterToImmersiveMode()
```
pocmo: Following the breadcrumbs back to the Fenix issue... the goal here is to decide whether we should be in portrait mode or landscape mode when playing a video (depending on its aspect ratio)?

I wonder if another workaround (until we can actually know which `Media` we should look at) would be to not lock orientation in fullscreen at all. That is how RB does it and during fullscreen video playback you can still rotate the phone. This is still not perfect, but gives the user the option to choose (and if needed lock orientation changes globally).

components/concept/engine/src/main/java/mozilla/components/concept/engine/media/Media.kt
```diff
@@ -178,7 +178,9 @@ abstract class Media(
      * @property duration Indicates the duration of the media in seconds.
      */
     data class Metadata(
-        val duration: Double = -1.0
+        val duration: Double = -1.0,
+        val height: Long = 0L,
+        val width: Long = 0L
     )
```
pocmo: In #4287 I am currently refactoring media / feature-media to not use `browser-session` anymore (and `browser-state` instead). Let's try to not break each other's code. :)

Maybe after the refactoring we can also make it easier to choose the right orientation - since with `browser-state` we do not depend on callback order that much. But that's something I'd have to check after I have it working. :)

ekager:
Sorry for the very long delay and lack of context D: 

We no longer _lock_ orientation in full screen (this was a stupid idea that I had without realizing how common portrait videos are these days and have since removed), but it's still not ideal if you are trying to watch landscape videos on YouTube and get launched in portrait every time and have to rotate, or vice versa (whichever orientation we would set as the default). We want to be better about knowing whether to launch in landscape or portrait.

GV could add a more info to the fullscreen API so we could get the aspect ratio in the fullscreen callback (see https://bugzilla.mozilla.org/show_bug.cgi?id=1600962#c4). Agi is waiting on us to tell him whether we can use this height/width metadata or not before adding new APIs.

Maybe we can chat about this next week because it looks like your media refactor has landed and maybe that has changed things?

pocmo:
> Sorry for the very long delay and lack of context D:

Sorry, now it took me a while to find back to this PR... :)

The `feature-media` refactoring has landed and now we can land patches that add width/height if needed. As you mention adding a `fullscreen` property would be great but if it gets called later than this does not help. From what I can see the order of those events should not have changed and is completely in the hands of GV. Let's ask them if they can change (and guarantee) the order or if they have a different proposal.

pocmo:
I commented on Bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1600962

While writing the comment I noticed that we may have another option:
* (1) We continue to switch to fullscreen from the feature - without locking
* (2) We add another feature that does locking later, based on playing fullscreen media if it sees any.

For (2) we could add a little data object (with fullscreen, ratio or orientation) to `MediaState` or `MediaState.Aggregate`. `MediaMiddleware` could look at fullscreen changes of individual media elements and update this "combined" state if needed.

ekager:
Okay again sorry for the crazy delay 😓 
Attempted to work off your idea from https://github.com/mozilla-mobile/android-components/pull/6216#issuecomment-607808297 so now looking for feedback if I'm weaving through all the media layers correctly 😅 
This seems to work decently well on popular sites ie Vimeo, Youtube
Portrait video to test with: https://vimeo.com/265111898

pocmo:
> Okay again sorry for the crazy delay 😓

Heh, we will get this landed sooner or later! :)

> Attempted to work off your idea from #6216 (comment) so now looking for feedback if I'm weaving through all the media layers correctly 😅

Yeah, I think that looks good. I may have some nits, but I don't want to annoy you as long as this is only a draft. Let's add some tests and then feel free to turn this into a PR and flag me for review. :)

components/feature/media/src/main/java/mozilla/components/feature/media/ext/MediaState.kt
```diff
@@ -28,6 +28,28 @@ fun MediaState.getPlayingMediaIdsForTab(tabId: String?): List<String> {
         ?: emptyList()
 }
 
+/**
+ * Finds any fullscreen video status and orientation of fullscreen media.
+ */
+fun MediaState.findFullscreenMediaOrientation(): MediaState.FullscreenOrientation? {
```
pocmo, line 34: nit: This can be `internal` and is not expected to be called by users of the library, I think?

components/feature/media/src/main/java/mozilla/components/feature/media/middleware/MediaMiddleware.kt
```diff
@@ -90,7 +90,8 @@ class MediaMiddleware(
             is MediaAction.RemoveMediaAction,
             is MediaAction.RemoveTabMediaAction,
             is MediaAction.UpdateMediaVolumeAction,
-            is MediaAction.UpdateMediaStateAction -> {
+            is MediaAction.UpdateMediaStateAction,
+            is MediaAction.UpdateMediaFullscreenAction -> {
```
pocmo, line 94: I wonder if it is a bit too aggressive to run the whole aggregator whenever the fullscreen mode changes (since that should not affect the media state) - and also vise versa looking up the fullscreen mode whenever we process a playing mode change.
But then again a large part of that code may change soon with the new GeckoView API anyways - and also the fullscreen state doesn't change all the time either. So probably fine. Also, I guess, not easy to isolate since the aggregator always creates a new aggregate.

components/feature/media/src/main/java/mozilla/components/feature/media/fullscreen/MediaFullscreenOrientationFeature.kt
```diff
@@ -0,0 +1,52 @@
...
+import mozilla.components.browser.state.state.MediaState.FullscreenOrientation
+import mozilla.components.browser.state.store.BrowserStore
+import mozilla.components.lib.state.ext.flowScoped
+import mozilla.components.support.base.feature.LifecycleAwareFeature
+import mozilla.components.support.ktx.kotlinx.coroutines.flow.ifChanged
+
+/**
+ * Feature that will auto-rotate the device to the correct orientation for the media aspect ratio.
+ */
+class MediaFullscreenOrientationFeature(
```
pocmo, line 23: This looks nice! I think eventually we could merge this with `FullScreenFeature`. But since that is not migrated to `BrowserStore` yet ... not in this PR. :)

components/concept/engine/src/main/java/mozilla/components/concept/engine/media/Media.kt
```diff
@@ -197,7 +204,18 @@ abstract class Media(
...
+    )
+
+    /**
+     * Fullscreen status of [Media].
+     *
+     * @property inFullscreen Indicates if the media is in fullscreen.
+     */
+    data class Fullscreen(
+        val inFullscreen: Boolean = false
     )
```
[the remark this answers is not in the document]
ekager: You're right we don't need it 🤷‍♀️

pocmo:
Rebased the PR to fix the changelog conflict. Let's land this! :)

bors r+

PR #6231, opened 2020-03-10
Closes #6227: Mark extensions as allowed for private browsing by default

### LINKED ISSUE #6227: Webextension intermittently fails to detect active tab
This is a regression from https://bugzilla.mozilla.org/show_bug.cgi?id=1599139, described in https://bugzilla.mozilla.org/show_bug.cgi?id=1621385. 

We don't integrate with this API yet. We'll need it once working on https://github.com/mozilla-mobile/fenix/issues/6123.

For now, let's make sure we call `setAllowedinPrivateBrowsing` with true, if needed. The default in GV is false.

csadilek:


csadilek:
bors r=Amejia481

Issue #6235, opened 2020-03-11
lib-state: Add support for middleware


pocmo:
With the refactoring of `feature-media` (#4287) we now have a use case that could benefit from a middleware.

https://redux.js.org/advanced/middleware/

csadilek:
Closed by: https://github.com/mozilla-mobile/android-components/commit/f4d962d1d4e2743eada9016d85ed20841fa26752

PR #6241, opened 2020-03-11
Issue #6235: lib-state/browser-state: Add support for middleware.

pocmo:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/lib/state/src/main/java/mozilla/components/lib/state/Middleware.kt
+++ b/components/lib/state/src/main/java/mozilla/components/lib/state/Middleware.kt
@@ -1,0 +1,33 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.lib.state
+
+/**
+ * A [Middleware] sits between the store and the reducer. It provides an extension point between
+ * dispatching an action, and the moment it reaches the reducer.
+ *
+ * A [Middleware] can rewrite an [Action], it can intercept an [Action], dispatch additional
+ * [Action]s or perform side-effects when an [Action] gets dispatched.
+ *
+ * The [Store] will create a chain of [Middleware] instances and invoke them in order. Every
+ * [Middleware] can decide to continue the chain (by calling `next`), intercept the chain (by not
+ * invoking `next`). A [Middleware] has no knowledge of what comes before or after it in the chain.
+ */
+typealias Middleware<S, A> = (store: MiddlewareStore<S, A>, next: (A) -> Unit, action: A) -> Unit
+
+/**
+ * A simplified [Store] interface for the purpose of passing it to a [Middleware].
+ */
+interface MiddlewareStore<S, A> {
+    /**
+     * Returns the current state of the [Store].
+     */
+    val state: S
+
+    /**
+     * Dispatches an [Action] synchronously on the [Store].
+     */
+    fun dispatch(action: A)
+}
--- a/components/lib/state/src/main/java/mozilla/components/lib/state/Store.kt
+++ b/components/lib/state/src/main/java/mozilla/components/lib/state/Store.kt
@@ -12,23 +12,11 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.asCoroutineDispatcher
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.launch
+import mozilla.components.lib.state.internal.ReducerChainBuilder
 import java.lang.ref.WeakReference
 import java.util.Collections
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.Executors
-
-/**
- * Reducers specify how the application's [State] changes in response to [Action]s sent to the [Store].
- *
- * Remember that actions only describe what happened, but don't describe how the application's state changes.
- * Reducers will commonly consist of a `when` statement returning different copies of the [State].
- */
-typealias Reducer<S, A> = (S, A) -> S
-
-/**
- * Listener called when the state changes in the [Store].
- */
-typealias Observer<S> = (S) -> Unit
 
 /**
  * A generic store holding an immutable [State].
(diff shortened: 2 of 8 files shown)


components/lib/state/src/main/java/mozilla/components/lib/state/Store.kt
```diff
@@ -38,11 +25,23 @@ typealias Observer<S> = (S) -> Unit
...
  * @param reducer A function that gets the current [State] and [Action] passed in and will return a new [State].
+ * @param middleware Optional list of [Middleware] sitting between the [Store] and the [Reducer].
  */
 open class Store<S : State, A : Action>(
     initialState: S,
-    private val reducer: Reducer<S, A>
+    reducer: Reducer<S, A>,
+    middleware: List<Middleware<S, A>> = emptyList()
 ) {
+    private val chain: (S, A) -> S by lazy {
```
csadilek: nit: what do you think about calling this function `reduce`. Chain introduces an additional name/concept and we're really only reducing the state here using the configured reducer and middleware.

components/lib/state/src/main/java/mozilla/components/lib/state/Middleware.kt
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * A [Middleware] can rewrite an [Action] or a [State], it can intercept an [Action], dispatch
+ * additional [Action]s or perform side-effects when an [Action] gets dispatched.
+ *
+ * The [Store] will create a chain of [Middleware] instances and invoke them in order. Every
+ * [Middleware] can decide to continue the chain or intercept/modify the [State] or [Action] before
+ * finally returning a [State]. A [Middleware] has no knowledge of what comes before or after it in
+ * the chain.
+ */
+typealias Middleware<S, A> = (state: S, action: A, next: (S, A) -> S) -> S
```
csadilek: I am wondering if we should pass the Store instead of the State which is what Redux does. Since we want the middleware to dispatch actions, I think this would make sense? Of course we could pass the store when creating the middleware but that could technically be a different instance than the one it is being applied on.

This is only if we want middlewares to be able to dispatch (not just rewrite) additional actions, which is mentioned in the docs above...

components/lib/state/src/main/java/mozilla/components/lib/state/Store.kt
pocmo: Yeah, that makes sense. Wasn't super happy about that name. :)

components/lib/state/src/main/java/mozilla/components/lib/state/Middleware.kt
pocmo: Ah, yes, that is a good idea. I may end up refactoring the middleware for `feature-media` into something that will need to dispatch too - and then I can try that. I'll keep this PR open until then.

pocmo: Also passing in the `Store` to the `Middleware` doesn't really seem to work since the `Middleware` will also need to be passed to the `Store` constructor which turns into a recursive problem. :)

pocmo: Not passing in the `State` means the `Store` needs to have the new `State` applied already - which currently also is not the case (since the "chain" only goes down to the reducer and not invokes all of dispatch yet). I'll try to refactor this and test this with my `feature-media` middleware and then report back. :)

pocmo:
@csadilek Alright. Updated the PR.

This is now more closely following what Redux does. A Middleware no longer returns a `State` and instead only invokes `next` (or not), but has access to a (limited) `Store` to get the `State` or dispatch additional `Action`s.

I tested this with my media middleware and it simplified it a lot. It also solved a problem where I wanted to do someting *after* the state in the store has changed and this is now possible after `next` was called. Previously this only invoked the reducer and not update the state in the store yet (and not notify observers). I am pretty happy with that result.

pocmo:
Okay, I am pretty much done with the `feature-media` refactoring. Still need to do some cleanup and (re)write a bunch of tests. But this part looks good and should be ready to land.

bors r=csadilek

pocmo:
Seeing some test failures in other components. Didn't expect that since the store tests all pass and the public API should not have changed. Investigating!

pocmo:
Seeing for example `AddonManagerTest.updateAddon - when a extension is updated successfully` fail.

csadilek:
@pocmo I found the cause of the problem:

This happens for tests that use a `spy(BrowserStore(..))`. The spy is a wrapper of the actual store and since we now capture `this` inside the reducer lambda we're only transitioning the state of the inner store (the store being wrapped, not the spy). Before, this was working, as the reducer returned the state and we set it as part of dispatch on the same instance we invoked dispatch on.

In Redux, a middleware (same as reducers) return the state, which we could use to fix our problem. I've pushed this here (tests passing): https://github.com/csadilek/android-components/commit/91759ea6265bddf2454f4d9f391e433b454f4802

pocmo:
Aaah, spy!! This is not the first time that a spy has caused trouble.

* In your patch we are now calling `transitionTo()` twice. Once at the end of the chain after the reducer was invoked (as before) and once at the end of `dispatch()` with the state passed back through the chain. That does not happen for `MiddlewareStore.dispatch()` though. This is somewhat complicated to grasp and something I think we should avoid.

* As I understand the Redux docs their middleware does not return the state, but can return _something_. It could be the state or something else, but it doesn't seem to affect the store. And I think I would like to avoid returning the state through the chain since that allows modifying the state outside of the reducers - which I tried at the beginning of the patch and it is complicated and hard to understand. It also means that after calling `next()` a Middleware may not see the new state since it can still change on the way down.

I will experiment a bit more today. Right now my gut feeling is that I would prefer to change/fix the tests instead of changing the behavior here. We seem to use a spy to verify that some action was dispatched on the store. Another option may be that we verify only that the state is what we expect and ignore the implementation detail of how it changed - which may not be the worst idea in general and closer to a black-box-y unit test.

Okay, I'm going back into that rabbit hole now. :)

csadilek:
@pocmo yes, have the same concern about the final additional `transitionTo`. I basically does nothing as the state is already the same, but it adds complexity just to making spies work.

pocmo:
bors r=csadilek

PR #6243, opened 2020-03-11
Downgrade application services dependencies to 0.52.0.

pocmo:
Due to a regression in AS 0.53.0 we are downgrading to 0.52.0 for the upcoming Fenix 4.1 release. See:
* https://github.com/mozilla-mobile/fenix/issues/9027#issuecomment-597328582
* https://github.com/mozilla/application-services/issues/2791


pocmo:
Closing: We downgraded to AS 0.52.0 in AC 36.0.1. We will not perform the downgrade on `master` and instead work on updating to AS 0.54.0 (and 0.55.0 with the regression fix once available).

PR #6251, opened 2020-03-12
Closes #6012: Run the FxA extension on non-prod servers

### LINKED ISSUE #6012: Add support for FxA staging servers to fxawebchannel extension
The extension's content script currently doesn't run for the pages returned from the staging server.

eoger:
Thanks to `GeckoPort.senderUrl`  exposing the content URL, we can now assert in `FxaWebChannelFeature` that a WebChannel message comes from the right content page.
This allows us to inject the extension content script in more FxA server instances while ensuring that a non-prod instance can mess up a prod-instance account data.
 
---

Also fixes https://github.com/mozilla/application-services/issues/2795.

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/accounts/src/main/assets/extensions/fxawebchannel/manifest.json
+++ b/components/feature/accounts/src/main/assets/extensions/fxawebchannel/manifest.json
@@ -4,7 +4,13 @@
   "version": "1.0",
   "content_scripts": [
     {
-      "matches": ["https://accounts.firefox.com/*"],
+      "matches": [
+        "https://accounts.firefox.com/*",
+        "https://stable.dev.lcip.org/*",
+        "https://accounts.stage.mozaws.net/*",
+        "https://latest.dev.lcip.org/*",
+        "https://accounts.firefox.com.cn/*"
+      ],
       "js": ["fxawebchannel.js"],
       "run_at": "document_start"
     }
--- a/components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FxaWebChannelFeature.kt
+++ b/components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FxaWebChannelFeature.kt
@@ -15,17 +15,20 @@ import mozilla.components.concept.engine.webextension.MessageHandler
 import mozilla.components.concept.engine.webextension.Port
 import mozilla.components.concept.sync.AuthType
 import mozilla.components.service.fxa.FxaAuthData
+import mozilla.components.service.fxa.ServerConfig
 import mozilla.components.service.fxa.SyncEngine
 import mozilla.components.service.fxa.manager.FxaAccountManager
 import mozilla.components.service.fxa.sync.toSyncEngines
 import mozilla.components.service.fxa.toAuthType
 import mozilla.components.support.base.feature.LifecycleAwareFeature
 import mozilla.components.support.base.log.logger.Logger
+import mozilla.components.support.ktx.kotlin.isSameOriginAs
 import mozilla.components.support.webextensions.WebExtensionController
 import org.json.JSONArray
 import org.json.JSONException
 import org.json.JSONObject
 import java.lang.ClassCastException
+import java.net.URL
 
 /**
  * Configurable FxA capabilities.
@@ -53,6 +56,7 @@ class FxaWebChannelFeature(
     private val engine: Engine,
     private val sessionManager: SessionManager,
     private val accountManager: FxaAccountManager,
+    private val serverConfig: ServerConfig,
     private val fxaCapabilities: Set<FxaCapability> = emptySet()
 ) : SelectionAwareSessionObserver(sessionManager), LifecycleAwareFeature {
 
@@ -98,9 +102,16 @@ class FxaWebChannelFeature(
      */
     private class WebChannelViewContentMessageHandler(
         private val accountManager: FxaAccountManager,
+        private val serverConfig: ServerConfig,
         private val fxaCapabilities: Set<FxaCapability>
     ) : MessageHandler {
+        @SuppressWarnings("ComplexMethod")
         override fun onPortMessage(message: Any, port: Port) {
+            if (!isCommunicationAllowed(serverConfig, port)) {
+                logger.error("Communication disallowed, ignoring WebChannel message.")
+                return
+            }
+
             val json = try {
                 message as JSONObject
             } catch (e: ClassCastException) {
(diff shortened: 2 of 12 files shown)


eoger:
Not sure who's the right person to review on a-c side, a-s side @rfk since you were having security concerns in https://github.com/mozilla-mobile/android-components/issues/6012#issuecomment-589341798.

components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FxaWebChannelFeature.kt
```diff
@@ -98,9 +101,19 @@ class FxaWebChannelFeature(
      */
     private class WebChannelViewContentMessageHandler(
         private val accountManager: FxaAccountManager,
+        private val serverConfig: ServerConfig,
         private val fxaCapabilities: Set<FxaCapability>
     ) : MessageHandler {
+        @SuppressWarnings("ComplexMethod")
         override fun onPortMessage(message: Any, port: Port) {
+            val senderOrigin = URL(port.senderUrl()).host
+            val expectedOrigin = URL(serverConfig.contentUrl).host
```
rfk: I'm not super familiar with `java.net.URL` but a quick read of the docs suggests that `host` may not be sufficient here, because it may not include e.g. the port number or URL scheme. I've seen enough origin-matching bugs in my time to be very wary of getting this right.  I'd like to see this check factored out into some sort of `isSameOrigin()` function and have tests for at least the following cases:

* `isSameOrigin("https://example.com", https://example.com") -> true` (simple match)
* `isSameOrigin("https://example.com/foo", https://example.com/bar") -> true` (matches ignoring path)
* `isSameOrigin("https://example.com", https://example.com:443") -> true` (default port for scheme)
* `isSameOrigin("https://example.com:8000", https://example.com:8000") -> true` (non-default port match)
* `isSameOrigin("https://example.com", https://example.net") -> false` (host mismatch)
* `isSameOrigin("https://example.com", http://example.com") -> false` (scheme mismatch)
* `isSameOrigin("https://example.com", https://example.com:444") -> false` (non-default port mismatch)
* `isSameOrigin("https://example.com:444", https://example.com:445") -> false` (explicit port mismatch)
* `isSameOrigin("https://example.com", http://example.com:443") -> false` (scheme mismatch despite explicit port)
* `isSameOrigin("https://example.com:8000", http://example.com:8000") -> true` (scheme mismatch despite non-default port)

I had a bit of a look around but wasn't able to find an easy way to do this ([this package](https://kotlinlang.org/api/latest/jvm/stdlib/org.w3c.dom.url/) provides a `URL` class with an `origin` property, but looks like it's only when targetting javascript?). Absent something better, I suggest something like (pardon my pseudocode):

```
def canonicalizeOrigin(v):
  u = URL(v)
  c = URL(u.protocol, u.host, u.port == -1 ? u.defaultPort : u.port, "")
  return c.toString()

def isSameOrigin(v1, v2):
    return canonicalizeOrigin(v1) == canonicalizeOrigin(v2)
```

It's not very efficient, but it's easy to understand.

I also think it would be valuable to add some belt-and-braces security checks for things we would never expect to se in the wild:

* Reject any messages received from a non-https non-localhost URL
* Reject any messages received with from a URL with non-empty userinfo component

components/feature/accounts/src/test/java/mozilla/components/feature/accounts/FxaWebChannelFeatureTest.kt
```diff
@@ -144,17 +150,61 @@ class FxaWebChannelFeatureTest {
...
+        val messageHandler = argumentCaptor<MessageHandler>()
+        val port = mock<Port>()
+
+        WebExtensionController.installedExtensions[FxaWebChannelFeature.WEB_CHANNEL_EXTENSION_ID] = ext
+
+        whenever(accountManager.supportedSyncEngines()).thenReturn(null)
+        whenever(sessionManager.getOrCreateEngineSession(session)).thenReturn(engineSession)
+        whenever(port.engineSession).thenReturn(engineSession)
+        whenever(port.senderUrl()).thenReturn("https://bar.foo/email")
+        whenever(serverConfig.contentUrl).thenReturn("https://foo.bar")
```
rfk: As mentioned above, I think we should test a lot more mis-matched combinations here, but perhaps we can do that in a way that only tests the matching logic rather than having to spin up all this infra and actually send a request.

components/feature/accounts/src/main/assets/extensions/fxawebchannel/manifest.json
```diff
@@ -4,7 +4,12 @@
   "version": "1.0",
   "content_scripts": [
     {
-      "matches": ["https://accounts.firefox.com/*"],
+      "matches": [
```
eoger: I should also add `latest` here.

eoger:
I have addressed @rfk's good comments, r? @csadilek

docs/changelog.md
```diff
@@ -12,6 +12,14 @@ permalink: /changelog/
 * [Gecko](https://github.com/mozilla-mobile/android-components/blob/master/buildSrc/src/main/java/Gecko.kt)
 * [Configuration](https://github.com/mozilla-mobile/android-components/blob/master/buildSrc/src/main/java/Config.kt)
 
+* **browser-engine-gecko**, **browser-engine-gecko-beta**, **browser-engine-gecko-nightly**
+  * Added `GeckoPort.senderUrl` which returns the associated content URL.
+
+* **feature-accounts**
+  * It should now be possible to log-in on stable, stage and china FxA servers using a WebChannel flow.
+  * `FxaWebChannelFeature` takes a `ServerConfig` as 4th parameter to ensure incoming WebChannel messages
```
csadilek: Let's mark this as a breaking change (there are examples in this files). We need to put up PRs to Fenix and RB in anticipation of this landing too, if you don't mind. Otherwise we'll break master.

components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FxaWebChannelFeature.kt
```diff
@@ -299,5 +309,48 @@ class FxaWebChannelFeature(
...
+                return false
+            }
+            return true
+        }
+
+        /**
+         * Compares 2 URLs and returns true if they have the same origin,
+         * which means: same protocol, same host, same port.
+         */
+        private fun isSameOrigin(a: String, b: String): Boolean {
```
csadilek: Nice! Let's move this to a `String` extension function similar to https://github.com/mozilla-mobile/android-components/blob/master/components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlin/String.kt#L105

Then we can also move the extensive test case over to StringUtilsTest. You can then also invoke this as `senderOrigin.isSameOriginAs(expectedOrigin)`.

eoger:
- Opened https://github.com/mozilla-mobile/fenix/pull/9151 for Fenix, r-b doesn't have the webchannel feature.
- Addressed your comments

Thanks @csadilek! If you're happy with it this is ready to land.

eoger:
Rebased and ready to land, let me know how you want to proceed.

csadilek:
@rfk can you approve this PR to make bors happy :)

PR #6264, opened 2020-03-13
Closes #6263 Sanitize the exceptions sent by the SupportedAddonsChecker

### LINKED ISSUE #6263: Sanitize the exceptions sent by DefaultSupportedAddonsChecker
After putting in production the `DefaultSupportedAddonsChecker`, we have been getting a lot of [noise](https://sentry.prod.mozaws.net/operations/fenix-nightly/issues/7510929/?query=is:unresolved) as we are reporting all the exceptions that happen while trying to check for new  supported add-ons. Let's just report only exceptions that are actionable for us.

Amejia481:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/addons/src/main/java/mozilla/components/feature/addons/migration/SupportedAddonsChecker.kt
+++ b/components/feature/addons/src/main/java/mozilla/components/feature/addons/migration/SupportedAddonsChecker.kt
@@ -30,6 +30,7 @@ import mozilla.components.support.base.log.logger.Logger
 import mozilla.components.support.ktx.android.content.appName
 import mozilla.components.support.ktx.android.notification.ChannelData
 import mozilla.components.support.ktx.android.notification.ensureNotificationChannelExists
+import java.io.IOException
 import java.util.concurrent.TimeUnit
 
 /**
@@ -80,7 +81,7 @@ class DefaultSupportedAddonsChecker(
     override fun registerForChecks() {
         WorkManager.getInstance(applicationContext).enqueueUniquePeriodicWork(
                 CHECKER_UNIQUE_PERIODIC_WORK_NAME,
-                ExistingPeriodicWorkPolicy.KEEP,
+                ExistingPeriodicWorkPolicy.REPLACE,
                 createPeriodicWorkerRequest()
         )
         logger.info("Register check for new supported add-ons")
@@ -160,7 +161,9 @@ internal class SupportedAddonsWorker(
             }
         } catch (exception: Exception) {
             logger.error("An exception happened trying to check for new supported add-ons, re-schedule ${exception.message}", exception)
-            GlobalAddonDependencyProvider.onCrash?.invoke(exception)
+            if (!exception.shouldReport()) {
+                GlobalAddonDependencyProvider.onCrash?.invoke(exception)
+            }
         }
         return Result.success()
     }
@@ -243,4 +246,6 @@ internal class SupportedAddonsWorker(
 
         internal var onNotificationClickIntent: Intent = Intent()
     }
+
+    private fun Exception.shouldReport(): Boolean = cause is IOException
 }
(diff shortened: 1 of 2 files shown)


components/feature/addons/src/main/java/mozilla/components/feature/addons/migration/SupportedAddonsChecker.kt
```diff
@@ -80,7 +82,7 @@ class DefaultSupportedAddonsChecker(
     override fun registerForChecks() {
         WorkManager.getInstance(applicationContext).enqueueUniquePeriodicWork(
                 CHECKER_UNIQUE_PERIODIC_WORK_NAME,
-                ExistingPeriodicWorkPolicy.KEEP,
+                ExistingPeriodicWorkPolicy.REPLACE,
```
Amejia481: We need this to change how often the worker will run [see](https://developer.android.com/reference/androidx/work/ExistingPeriodicWorkPolicy). If we don't add it, the old constraints will be maintained when consumers (Fenix) change frequency of the worker.

components/feature/addons/src/main/java/mozilla/components/feature/addons/migration/SupportedAddonsChecker.kt
```diff
@@ -160,7 +162,9 @@ internal class SupportedAddonsWorker(
             }
         } catch (exception: Exception) {
             logger.error("An exception happened trying to check for new supported add-ons, re-schedule ${exception.message}", exception)
-            GlobalAddonDependencyProvider.onCrash?.invoke(exception)
+            if (!exception.isNoise()) {
```
csadilek: Let's just add another catch block to handle `IOException`s first. Let's have the type system do this for you. `SocketTimeOutException`s is an IO exception too. So, we don't need this extensions method.

components/feature/addons/src/main/java/mozilla/components/feature/addons/migration/SupportedAddonsChecker.kt
csadilek: 👍

components/feature/addons/src/main/java/mozilla/components/feature/addons/migration/SupportedAddonsChecker.kt
Amejia481: 👍

Amejia481: It will be very convenient to just add an additional catch block, but unfortunately we are wrapping all exceptions that come from [getAddons as AddonManagerException](https://github.com/mozilla-mobile/android-components/blob/e8e4806526aeb895a69e003c7917d3c0ceb986a2/components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt#L100), as result we need to unwrap them using `exception.cause`

csadilek: Ah, I missed that! OK, maybe just call it `exception.shouldReport()`?

components/feature/addons/src/test/java/mozilla/components/feature/addons/migration/SupportedAddonsWorkerTest.kt
```diff
@@ -92,6 +96,33 @@ class SupportedAddonsWorkerTest {
         }
     }
 
+    @Test
+    fun `doWork - will NOT pass any IOException or SocketTimeoutException  exceptions to the crashReporter`() {
```
csadilek: nit: extra space.

components/feature/addons/src/main/java/mozilla/components/feature/addons/migration/SupportedAddonsChecker.kt
Amejia481: Sounds good! :)

Amejia481:
bors r=csadilek

Issue #6269, opened 2020-03-16
Tracking Protection icon not updated properly


ValentinTimisica:
When all exceptions are removed the icon still shows a wrong state. We should notify the sessions when exceptions are removed.

See Fenix issue [here](https://github.com/mozilla-mobile/fenix/issues/8597) for more details.

ValentinTimisica:
Verified as fixed on Fenix [issue](https://github.com/mozilla-mobile/fenix/issues/8597#issuecomment-601150082).

PR #6270, opened 2020-03-16
For #6269: Notify active sessions when all exceptions are removed.

ValentinTimisica:
I also updated the tests accordingly.
Made this change because the tracking protection icon is not updated after the
exceptions are removed.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/concept/engine/src/main/java/mozilla/components/concept/engine/content/blocking/TrackingProtectionExceptionStorage.kt
+++ b/components/concept/engine/src/main/java/mozilla/components/concept/engine/content/blocking/TrackingProtectionExceptionStorage.kt
@@ -47,8 +47,10 @@ interface TrackingProtectionExceptionStorage {
 
     /**
      * Removes all domains from the exception list.
+     * @param activeSessions A list of all active sessions (including CustomTab
+     * sessions) to be notified.
      */
-    fun removeAll()
+    fun removeAll(activeSessions: List<EngineSession?>? = null)
 
     /**
      * Restore all domains stored in the storage.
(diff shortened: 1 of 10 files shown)


components/concept/engine/src/main/java/mozilla/components/concept/engine/content/blocking/TrackingProtectionExceptionStorage.kt
```diff
@@ -48,7 +48,7 @@ interface TrackingProtectionExceptionStorage {
     /**
      * Removes all domains from the exception list.
      */
-    fun removeAll()
+    fun removeAll(activeSessions: List<EngineSession?>)
```
Amejia481: Adding a new parameter will cause a breaking change on the API. I believe we are the only consumers but to be sure, you can add a default a value  `fun removeAll(activeSessions: List<EngineSession?> = null)`. also we need to add some docs about this new parameter :)

components/feature/session/src/main/java/mozilla/components/feature/session/TrackingProtectionUseCases.kt
```diff
@@ -85,7 +85,9 @@ class TrackingProtectionUseCases(
          * Removes all domains from the exception list.
          */
         operator fun invoke() {
-            engine.trackingProtectionExceptionStore.removeAll()
+            engine.trackingProtectionExceptionStore.removeAll(sessionManager.all.map {
```
Amejia481: As we are applying this on all the sessions it could cause a performance hit, right now we don't have a straightforward way to avoid it, as it will required that we introduce the browser store, we are going to address this on https://github.com/mozilla-mobile/android-components/issues/6283. Would you mind adding a comment with [this link](https://github.com/mozilla-mobile/android-components/issues/6283) please ? :)

PR #6286, opened 2020-03-16
Closes #6281: Do not automatically dismiss confirm dialogs

### LINKED ISSUE #6281: feature-prompts: DialogFragments (confirm prompts) sometimes not shown
I can reproduce this reliably using uBlock's settings page, which triggers a file prompt, then immediately follows with a `window.confirm` prompt. We get all the prompt requests, but somehow fail to display the confirmation window, despite calling `dialog.show` for it.

csadilek:
The feature to automatically dismiss can maybe be made more robust to not be affected by lifecycle changes (https://github.com/mozilla-mobile/android-components/issues/6281#issuecomment-599739212), but in either case we don't want to automatically dismiss confirmation dialogs, as these require user interaction before the next page load anyway.

Similar fix to: https://github.com/mozilla-mobile/android-components/pull/5704/

### DIFF
--- a/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt
+++ b/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt
@@ -533,6 +533,7 @@ class PromptFeature private constructor(
                         title,
                         message,
                         promptAbuserDetector.areDialogsBeingAbused(),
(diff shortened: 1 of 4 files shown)


csadilek:
bors r=Amejia481

Amejia481:
bors retry

csadilek:
bors retry

PR #6289, opened 2020-03-17
Closes #6290: Do not remove FxA push scope on logout

### LINKED ISSUE #6290: Push scope for Sync is out of sync between the in-memory and persisted value
The STR for one of the outcomes of this bug are here: https://github.com/mozilla-mobile/fenix/issues/9120

Another issue that I believe is related is https://github.com/mozilla-mobile/fenix/issues/9033.

### LINKED ISSUE #9033: Fallback to httpClient when the download.response throw an IOException
In cases where users with slow networks start a download [but quickly press pause and then resume](https://github.com/mozilla-mobile/fenix/issues/9354#issuecomment-730256235), there will be not enough time for bytes to be copied, but the stream in [[download.response] ](https://github.com/mozilla-mobile/android-components/blob/d19d167a48658d10ccd95a53990e368eb16595f1/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt#L622) will be [closed](https://github.com/mozilla-mobile/fenix/issues/9354#issuecomment-731267368), and when the user will retry we will try to use [[download.response] ](https://github.com/mozilla-mobile/android-components/blob/d19d167a48658d10ccd95a53990e368eb16595f1/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt#L622) over and over with a closed stream as a result the download will failed every time, we have to fallback to httpClient

jonalmeida:
We're caching the `fxaPushScope` in memory and also saving it to our
prefs. When we log out, we delete the pref, but the in-memory isn't
removed.

We could always fetch from the prefs instead of using an in-memory
caching, however this could be a bit expensive to always read from prefs
since we need to do this check when receiving any push message.

An easier solution would be to never remove the scope in prefs once
generated. It is safe to persist the push scope across different
accounts since the combination of scope + UAID is unique per install.

---

Closes #6290

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/accounts-push/src/main/java/mozilla/components/feature/accounts/push/FxaPushSupportFeature.kt
+++ b/components/feature/accounts-push/src/main/java/mozilla/components/feature/accounts/push/FxaPushSupportFeature.kt
@@ -141,7 +141,6 @@ internal class AccountObserver(
         // Delete cached value of last verified timestamp and scope when we log out.
         preference(context).edit()
             .remove(PREF_LAST_VERIFIED)
-            .remove(PREF_FXA_SCOPE)
             .apply()
     }
 }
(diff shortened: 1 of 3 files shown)


jonalmeida:
bors r=rocketsroger

PR #6298, opened 2020-03-17
For #6303 - Change logic for updating vs creating logins

ekager:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/concept/storage/src/main/java/mozilla/components/concept/storage/LoginsStorage.kt
+++ b/components/concept/storage/src/main/java/mozilla/components/concept/storage/LoginsStorage.kt
@@ -112,6 +112,14 @@ interface LoginsStorage : AutoCloseable {
      * @return A list of [Login] objects, representing matching logins.
      */
     suspend fun getByBaseDomain(origin: String): List<Login>
+
+    /**
+     * Fetch the list of potential duplicate logins from the underlying storage layer, ignoring username.
+     *
+     * @param login The [Login] to compare against for finding dupes.
+     * @return A list of [Login] objects, representing matching potential dupe logins.
+     */
+    suspend fun getPotentialDupesIgnoringUsername(login: Login): List<Login>
 }
 
 /**
@@ -196,12 +204,17 @@ interface LoginValidationDelegate {
          * Indicates that a matching [Login] was found in storage, and the [Login] can be used
          * to update its information.
          */
-        object CanBeUpdated : Result()
+        data class CanBeUpdated(val foundLogin: Login) : Result()
 
         /**
          * The [Login] cannot be saved.
          */
         sealed class Error : Result() {
+            /**
+             * Indicates that a duplicate [Login] was found in storage, we should not save it again.
+             */
+            object Duplicate : Result()
+
             /**
              * The passed [Login] had an empty password field, and so cannot be saved.
              */
@@ -216,14 +229,24 @@ interface LoginValidationDelegate {
     }
 
     /**
-     * Checks whether or not [login] can be persisted.
+     *
+     * Calls underlying storage methods to fetch list of [Login]s that could be potential dupes of [newLogin]
      *
      * Note that this method is not thread safe.
      *
-     * @returns a [LoginValidationDelegate.Result], detailing whether [login] can be saved as a new
-     * value, used to update an existing one, or an error occured.
+     * @returns a list of potential duplicate [Login]s
      */
-    fun validateCanPersist(login: Login): Deferred<Result>
+    fun getPotentialDupesIgnoringUsernameAsync(newLogin: Login): Deferred<List<Login>>
+
+    /**
+     *
+     * Checks whether a [login] should be saved or updated.
+     *
+     * Note that this method is not thread safe.
+     *
+     * @returns a [Result], detailing whether a [login] should be saved or updated.
+     */
+    fun shouldUpdateOrCreateAsync(newLogin: Login, potentialDupes: List<Login>? = null): Deferred<Result>
 }
 
 /**
--- a/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/dialog/LoginDialogFragment.kt
+++ b/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/dialog/LoginDialogFragment.kt
@@ -30,12 +30,12 @@ import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withContext
 import mozilla.components.concept.storage.Login
-import mozilla.components.concept.storage.LoginValidationDelegate
 import mozilla.components.concept.storage.LoginValidationDelegate.Result
 import mozilla.components.feature.prompts.R
 import mozilla.components.feature.prompts.ext.onDone
 import mozilla.components.support.ktx.android.content.appName
 import mozilla.components.support.ktx.android.view.toScope
+import java.util.concurrent.CopyOnWriteArrayList
 import kotlin.reflect.KProperty
 import com.google.android.material.R as MaterialR
 
@@ -71,7 +71,11 @@ internal class LoginDialogFragment : PromptDialogFragment() {
     private var username by SafeArgString(KEY_LOGIN_USERNAME)
     private var password by SafeArgString(KEY_LOGIN_PASSWORD)
 
-    private var stateUpdate: Job? = null
+    private var validateStateUpdate: Job? = null
+
+    // List of potential dupes ignoring username. We could potentially both read and write this list
+    // from different threads, so we are using a copy-on-write list.
+    private var potentialDupesList: CopyOnWriteArrayList<Login>? = null
 
     override fun shouldDismissOnLoad(): Boolean = false
 
@@ -180,7 +184,6 @@ internal class LoginDialogFragment : PromptDialogFragment() {
             override fun afterTextChanged(editable: Editable) {
                 // Note that password is accessed by `fun update`
                 password = editable.toString()
-
                 if (password.isEmpty()) {
                     setViewState(
                         confirmButtonEnabled = false,
@@ -224,42 +227,44 @@ internal class LoginDialogFragment : PromptDialogFragment() {
         )
         // The below code is not thread safe, so we ensure that we block here until we are sure that
         // the previous call has been canceled
-        runBlocking { stateUpdate?.cancelAndJoin() }
+        runBlocking { validateStateUpdate?.cancelAndJoin() }
         var validateDeferred: Deferred<Result>? = null
-        stateUpdate = scope.launch(IO) {
-            validateDeferred = feature?.loginValidationDelegate?.validateCanPersist(login)
+        validateStateUpdate = scope.launch(IO) {
+            val validationDelegate = feature?.loginValidationDelegate ?: return@launch
+            // We only want to fetch this list once. Since we are ignoring username the results will
+            // not change when user changes username or password fields
+            if (potentialDupesList == null) {
+                this@LoginDialogFragment.potentialDupesList = CopyOnWriteArrayList(
+                    validationDelegate.getPotentialDupesIgnoringUsernameAsync(
+                        login
+                    ).await()
+                )
+            }
+            // Passing a copy of this potential dupes list, not a reference, so we know for certain
+            // that what shouldUpdateOrCreateAsync is using can't change.
+            validateDeferred = validationDelegate.shouldUpdateOrCreateAsync(
+                login,
+                potentialDupesList?.toList()
+            )
             val result = validateDeferred?.await()
-
             withContext(Main) {
                 when (result) {
-                    is LoginValidationDelegate.Result.CanBeCreated ->
+                    Result.CanBeCreated -> {
                         setViewState(
                             confirmText =
                             context?.getString(R.string.mozac_feature_prompt_save_confirmation)
                         )
-                    is LoginValidationDelegate.Result.CanBeUpdated ->
+                    }
+                    is Result.CanBeUpdated -> {
                         setViewState(
                             confirmText =
                             context?.getString(R.string.mozac_feature_prompt_update_confirmation)
-                        )
-                    is LoginValidationDelegate.Result.Error.EmptyPassword ->
-                        setViewState(
-                            confirmButtonEnabled = false,
-                            passwordErrorText =
-                            context?.getString(R.string.mozac_feature_prompt_error_empty_password)
-                        )
-                    is LoginValidationDelegate.Result.Error.GeckoError -> {
-                        // TODO handle these errors more robustly. See:
-                        // https://github.com/mozilla-mobile/fenix/issues/7545
-                        setViewState(
-                            passwordErrorText =
-                            context?.getString(R.string.mozac_feature_prompt_error_unknown_cause)
                         )
                     }
                 }
             }
         }
-        stateUpdate?.invokeOnCompletion {
+        validateStateUpdate?.invokeOnCompletion {
             if (it is CancellationException) {
                 validateDeferred?.cancel()
             }
--- a/components/service/sync-logins/build.gradle
+++ b/components/service/sync-logins/build.gradle
@@ -27,6 +27,23 @@ android {
     }
 }
 
+configurations {
+    // There's an interaction between Gradle's resolution of dependencies with different types
+    // (@jar, @aar) for `implementation` and `testImplementation` and with Android Studio's built-in
+    // JUnit test runner.  The runtime classpath in the built-in JUnit test runner gets the
+    // dependency from the `implementation`, which is type @aar, and therefore the JNA dependency
+    // doesn't provide the JNI dispatch libraries in the correct Java resource directories.  I think
+    // what's happening is that @aar type in `implementation` resolves to the @jar type in
+    // `testImplementation`, and that it wins the dependency resolution battle.
+    //
+    // A workaround is to add a new configuration which depends on the @jar type and to reference
+    // the underlying JAR file directly in `testImplementation`.  This JAR file doesn't resolve to
+    // the @aar type in `implementation`.  This works when invoked via `gradle`, but also sets the
+    // correct runtime classpath when invoked with Android Studio's built-in JUnit test runner.
+    // Success!
+    jnaForTest
+}
+
 dependencies {
     // Parts of this dependency are typealiase'd or are otherwise part of this module's public API.
     api(Dependencies.mozilla_sync_logins) {
@@ -57,6 +74,11 @@ dependencies {
     testImplementation Dependencies.testing_mockito
     testImplementation Dependencies.testing_coroutines
     testImplementation project(':support-test')
+
+    jnaForTest Dependencies.thirdparty_jna
+    testImplementation files(configurations.jnaForTest.copyRecursive().files)
+    testImplementation Dependencies.mozilla_glean_forUnitTests
+    testImplementation Dependencies.mozilla_full_megazord_forUnitTests
 }
 
 apply from: '../../../publish.gradle'
--- a/components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/DefaultLoginValidationDelegate.kt
+++ b/components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/DefaultLoginValidationDelegate.kt
@@ -24,27 +24,103 @@ class DefaultLoginValidationDelegate(
     private val scope: CoroutineScope = CoroutineScope(IO)
 ) : LoginValidationDelegate {
 
-    @Suppress("ComplexMethod") // This method is not actually complex
-    override fun validateCanPersist(login: Login): Deferred<Result> {
+    /**
+     * Compares a [Login] to a passed in list of potential dupes [Login] or queries underlying
+     * storage for potential dupes list of [Login] to determine if it should be updated or created.
+     */
+    override fun shouldUpdateOrCreateAsync(
+        newLogin: Login,
+        potentialDupes: List<Login>?
+    ): Deferred<Result> {
+        return scope.async {
+            val potentialDupesList =
+                potentialDupes ?: storage.value.getPotentialDupesIgnoringUsername(newLogin)
+            val foundLogin = if (potentialDupesList.isEmpty()) {
+                // If list is empty, no records match -> create
+                null
+            } else {
+                // Matching guid, non-blank matching username -> update
+                potentialDupesList.find { it.guid == newLogin.guid && it.username == newLogin.username }
+                // Matching guid, blank username -> update
+                    ?: potentialDupesList.find { it.guid == newLogin.guid && it.username.isEmpty() }
+                    // Matching username -> update
+                    ?: potentialDupesList.find { it.username == newLogin.username }
+                    // Non matching guid, blank username -> update
+                    ?: potentialDupesList.find { it.username.isEmpty() }
+                    // else create
+            }
+            if (foundLogin == null) Result.CanBeCreated else Result.CanBeUpdated(foundLogin)
+        }
+    }
+
+    override fun getPotentialDupesIgnoringUsernameAsync(newLogin: Login): Deferred<List<Login>> {
+        return scope.async {
+            storage.value.getPotentialDupesIgnoringUsername(newLogin)
+        }
+    }
+
+    /**
+     * Queries underlying storage for a new [Login] to determine if should be updated or created,
+     * and can be merged and saved, or any error states.
+     *
+     * @return a [Result] detailing whether we should create or update (and the merged login to update),
+     * else any error states
+     */
+    @Suppress("ComplexMethod")
+    fun ensureValidAsync(login: Login): Deferred<Result> {
         return scope.async {
             try {
-                // We're setting guid=null here in order to trigger de-duping logic properly.
-                // Internally, the library ensures to not dupe records against themselves
-                // (via a `guid <> :guid` check), which means we need to "pretend" this is a new record,
-                // in order for `ensureValid` to actually throw `DUPLICATE_LOGIN`.
-                storage.value.ensureValid(login.copy(guid = null))
-                Result.CanBeCreated
+                var result = shouldUpdateOrCreateAsync(login).await()
+                val loginToCheck = when (result) {
+                    // If we are updating, let's validate the merged login
+                    is Result.CanBeUpdated -> result.foundLogin.mergeWithLogin(login)
+                    is Result.CanBeCreated -> login
+                    else -> login
+                }
+                storage.value.ensureValid(loginToCheck)
+                // Return the merged login to update to avoid doing this merge twice
+                if (result is Result.CanBeUpdated) {
+                    result = Result.CanBeUpdated(loginToCheck)
+                }
+                result
             } catch (e: InvalidRecordException) {
                 when (e.reason) {
-                    InvalidLoginReason.DUPLICATE_LOGIN -> Result.CanBeUpdated
+                    InvalidLoginReason.DUPLICATE_LOGIN -> Result.Error.Duplicate
                     InvalidLoginReason.EMPTY_PASSWORD -> Result.Error.EmptyPassword
                     InvalidLoginReason.EMPTY_ORIGIN -> Result.Error.GeckoError(e)
                     InvalidLoginReason.BOTH_TARGETS -> Result.Error.GeckoError(e)
                     InvalidLoginReason.NO_TARGET -> Result.Error.GeckoError(e)
-                    // TODO in what ways can the login fields be illegal? represent these in the UI
                     InvalidLoginReason.ILLEGAL_FIELD_VALUE -> Result.Error.GeckoError(e)
                 }
             }
+        }
+    }
+
+    companion object {
+        /**
+         * Will use values from [login] if they are 1) non-null and 2) non-empty.  Otherwise, will fall
+         * back to values from [this].
+         */
+        fun Login.mergeWithLogin(login: Login): Login {
+            infix fun String?.orUseExisting(other: String?) =
+                if (this?.isNotEmpty() == true) this else other
+
+            infix fun String?.orUseExisting(other: String) =
+                if (this?.isNotEmpty() == true) this else other
+
+            val origin = login.origin orUseExisting origin
+            val username = login.username orUseExisting username
+            val password = login.password orUseExisting password
+            val httpRealm = login.httpRealm orUseExisting httpRealm
+            val formActionOrigin = login.formActionOrigin orUseExisting formActionOrigin
+
+            return copy(
+                origin = origin,
+                username = username,
+                password = password,
+                httpRealm = httpRealm,
+                formActionOrigin = formActionOrigin
+            )
         }
     }
 }
--- a/components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/GeckoLoginStorageDelegate.kt
+++ b/components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/GeckoLoginStorageDelegate.kt
@@ -4,7 +4,6 @@
 
 package mozilla.components.service.sync.logins
 
-import androidx.annotation.VisibleForTesting
 import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Deferred
@@ -13,14 +12,8 @@ import kotlinx.coroutines.async
 import kotlinx.coroutines.launch
 import mozilla.components.concept.storage.Login
 import mozilla.components.concept.storage.LoginStorageDelegate
+import mozilla.components.concept.storage.LoginValidationDelegate
 import mozilla.components.concept.storage.LoginsStorage
-import mozilla.components.support.base.log.logger.Logger
-
-/**
- * A type of persistence operation, either 'create' or 'update'.
- */
-@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
-enum class Operation { CREATE, UPDATE }
 
 /**
  * [LoginStorageDelegate] implementation.
@@ -58,8 +51,6 @@ class GeckoLoginStorageDelegate(
     private val scope: CoroutineScope = CoroutineScope(Dispatchers.IO)
 ) : LoginStorageDelegate {
 
-    private val logger = Logger("GeckoLoginStorageDelegate")
-
     override fun onLoginUsed(login: Login) {
         val guid = login.guid
         // If the guid is null, we have no way of associating the login with any record in the DB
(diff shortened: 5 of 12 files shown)


components/service/sync-logins/src/test/java/mozilla/components/service/sync/logins/DefaultLoginValidationDelegateTest.kt
```diff
@@ -0,0 +1,202 @@
...
+            val operation = validationDelegate.getPersistenceOperation(newLogin).await()
+            assertEquals(Operation.CREATE, operation)
+        }
+
+    @Test
+    @Config(sdk = [21])
+    fun `WHEN existing login matches by important elements, should update`() = runBlockingTest {
+        val newLogin = createLogin(guid = "1")
+        val savedLogin = createLogin("2")
+        whenever(loginsStorage.getStoredLoginByElements(any())).thenReturn(savedLogin)
```
ekager: This fails locally (and probably on CI too) with an NPE. :/ Not sure what I'm doing wrong. Clearly since the other new tests work it's something with the `any()`

ekager: > java.lang.NullPointerException at mozilla.components.service.sync.logins.DefaultLoginValidationDelegate$Companion.getStoredLoginByElements(DefaultLoginValidationDelegate.kt:105)
> 	at mozilla.components.service.sync.logins.DefaultLoginValidationDelegateTest$WHEN existing login matches by important elements, should update$1.invokeSuspend(DefaultLoginValidationDelegateTest.kt:100)
> 	at mozilla.components.service.sync.logins.DefaultLoginValidationDelegateTest$WHEN existing login matches by important elements, should update$1.invoke(DefaultLoginValidationDelegateTest.kt)
> 	at kotlinx.coroutines.test.TestBuildersKt$runBlockingTest$deferred$1.invokeSuspend(TestBuilders.kt:50)
> 	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
> 	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:56)
> 	at kotlinx.coroutines.test.TestCoroutineDispatcher.dispatch(TestCoroutineDispatcher.kt:50)
> 	at kotlinx.coroutines.DispatchedContinuationKt.resumeCancellableWith(DispatchedContinuation.kt:271)
> 	at kotlinx.coroutines.intrinsics.CancellableKt.startCoroutineCancellable(Cancellable.kt:26)
> 	at kotlinx.coroutines.CoroutineStart.invoke(CoroutineStart.kt:109)
> 	at kotlinx.coroutines.AbstractCoroutine.start(AbstractCoroutine.kt:158)
> 	at kotlinx.coroutines.BuildersKt__Builders_commonKt.async(Builders.common.kt:89)
> 	at kotlinx.coroutines.BuildersKt.async(Unknown Source)
> 	at kotlinx.coroutines.BuildersKt__Builders_commonKt.async$default(Builders.common.kt:82)
> 	at kotlinx.coroutines.BuildersKt.async$default(Unknown Source)
> 	at kotlinx.coroutines.test.TestBuildersKt.runBlockingTest(TestBuilders.kt:49)
> 	at kotlinx.coroutines.test.TestBuildersKt.runBlockingTest$default(TestBuilders.kt:45)
> 	at mozilla.components.service.sync.logins.DefaultLoginValidationDelegateTest.WHEN existing login matches by important elements, should update(DefaultLoginValidationDelegateTest.kt:97)
> 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
> 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
> 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
> 	at java.lang.reflect.Method.invoke(Method.java:498)
> 	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
> 	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
> 	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
> 	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
> 	at org.robolectric.RobolectricTestRunner$HelperTestRunner$1.evaluate(RobolectricTestRunner.java:601)
> 	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
> 	at org.robolectric.internal.SandboxTestRunner$2.evaluate(SandboxTestRunner.java:260)
> 	at org.robolectric.internal.SandboxTestRunner.runChild(SandboxTestRunner.java:130)
> 	at org.robolectric.internal.SandboxTestRunner.runChild(SandboxTestRunner.java:42)
> 	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
> 	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
> 	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
> 	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
> 	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
> 	at org.robolectric.internal.SandboxTestRunner$1.evaluate(SandboxTestRunner.java:84)
> 	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
> 	at androidx.test.ext.junit.runners.AndroidJUnit4.run(AndroidJUnit4.java:104)
> 	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
> 	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:68)
> 	at com.intellij.rt.execution.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:47)
> 	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:242)
> 	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:70)

components/concept/storage/src/main/java/mozilla/components/concept/storage/LoginsStorage.kt
```diff
@@ -220,10 +218,22 @@ interface LoginValidationDelegate {
...
+     * A type of persistence operation, either 'create' or 'update'.
+     */
+    enum class Operation { CREATE, UPDATE }
+
+    /**
+     * Checks whether a [login] should be saved or updated.
+     *
+     * @returns a [LoginValidationDelegate.Operation], detailing whether [login] should be saved or updated
+     */
+    fun getPersistenceOperation(newLogin: Login): Deferred<Operation>
```
Baron-Severin: I feel like you explained this to me during our conversation and I'm just being dense, but this seems to cover a really similar conceptual space to `validateCanPersist`.  Why can't they be consolidated into one function?

components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/DefaultLoginValidationDelegate.kt
```diff
@@ -24,27 +27,117 @@ class DefaultLoginValidationDelegate(
...
             }
         }
     }
+
+    /**
+     * Searches storage to find a matching login to all A-S defined important elements
+     * (username, hostname, and formSubmitUrl or httpRealm)
+     * in case a user enters the same information as another login but doesn't match GUID
+     */
+    suspend fun getStoredLoginByElements(newLogin: Login): Login? {
```
Baron-Severin: Nit: what about `getStoredLoginByImportantFields`?  Or `getStoredLoginByStructualEquality`?  Except it's only partial structural equality....   This one's hard to name.

components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/DefaultLoginValidationDelegate.kt
```diff
@@ -24,27 +27,117 @@ class DefaultLoginValidationDelegate(
...
+            try {
+                URL(newLogin.origin).host
+            } catch (e: MalformedURLException) {
+                return null
+            }
+        )
+        return storageList.find {
+            it.username == newLogin.username &&
+                    (it.formActionOrigin == newLogin.formActionOrigin ||
+                            it.httpRealm == newLogin.httpRealm)
```
Baron-Severin: You mentioned during our call that this is based on how AS considers login equality, right?  Could you add a comment to that effect?

components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/DefaultLoginValidationDelegate.kt
```diff
@@ -24,27 +27,117 @@ class DefaultLoginValidationDelegate(
...
+        }
+    }
+
+    companion object {
+        /**
+         * Will use values from [this] if they are 1) non-null and 2) non-empty.  Otherwise, will fall
+         * back to values from [this].
```
Baron-Severin: Looks like I messed up this comment.  It should begin with `Will use values from [login] if`

components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/DefaultLoginValidationDelegate.kt
Baron-Severin: I almost feel like this should probably be a method on `LoginStorage` instead of here, what do you think?

ekager: Oops that's how I had it and I moved it out when trying to get testing working and didn't move it back. Done!

components/concept/storage/src/main/java/mozilla/components/concept/storage/LoginsStorage.kt
ekager: So the way I was thinking about it was from the UI perspective I want to separate the logic of "should we save or update?" (`getPersistenceOperation`) and "CAN we save or update this login?" (`validateCanPersist`). In `LoginDialogFragment` these are used for the button text and the error states/enabled states respectively. Does that make sense?

components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/DefaultLoginValidationDelegate.kt
ekager: Yeah it's in the fun level javadoc, do you want it in the fun too?

components/concept/storage/src/main/java/mozilla/components/concept/storage/LoginsStorage.kt
Baron-Severin: I understand the distinction from the UI perspective, but I'm not sure why they can't rely on the same underlying check.  If it's possible to consolidate them, I think the code would be easier to follow.

components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/DefaultLoginValidationDelegate.kt
Baron-Severin: No I guess you're right, probably fine as is.

components/concept/storage/src/main/java/mozilla/components/concept/storage/LoginsStorage.kt
ekager: Maybe - I think this would turn into a 70+ line behemoth function though and I see them as fundamentally different logic. I can look again tomorrow to see what this could look like.

ekager: Okay attempted! Let me know if you like this better. It's fairly large but maybe more understandable?

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/dialog/LoginDialogFragment.kt
```diff
@@ -210,31 +200,40 @@ internal class LoginDialogFragment : PromptDialogFragment() {
...
                         setViewState(
                             confirmButtonEnabled = false,
                             passwordErrorText =
                             context?.getString(R.string.mozac_feature_prompt_error_empty_password)
                         )
-                    is LoginValidationDelegate.Result.Error.GeckoError -> {
+                    is Result.Error.GeckoError -> {
                         // TODO handle these errors more robustly. See:
                         // https://github.com/mozilla-mobile/fenix/issues/7545
                         setViewState(
```
Baron-Severin: I think I probably should have had this disable the button as well.  Would you mind adding that in?

components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/DefaultLoginValidationDelegate.kt
```diff
@@ -24,27 +26,105 @@ class DefaultLoginValidationDelegate(
...
+        }
+    }
+
+    companion object {
+        /**
+         * Searches storage to find a matching login to all A-S defined important elements
+         * (username, hostname, and formSubmitUrl or httpRealm)
+         * in case a user enters the same information as another login but doesn't match GUID
+         */
+        suspend fun LoginsStorage.getStoredLoginByImportantFields(newLogin: Login): Login? {
```
Baron-Severin: Optional nit: I meant that I think this should probably be part of the `LoginsStorage` class itself, rather than an extension function.  I won't block on it though.

components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/GeckoLoginStorageDelegate.kt
```diff
@@ -78,68 +71,26 @@ class GeckoLoginStorageDelegate(
...
         scope.launch {
             val existingLogin = login.guid?.let { loginStorage.get(it) }
-
-            when (getPersistenceOperation(login, existingLogin)) {
-                Operation.UPDATE -> {
-                    existingLogin?.let { loginStorage.update(it.mergeWithLogin(login)) }
+                ?: loginStorage.getStoredLoginByImportantFields(login)
+            when (validationDelegate.getNewLoginPersistenceAsync(login).await()) {
+                LoginValidationDelegate.Result.CanBeUpdated -> {
+                    val mergedLogin = requireNotNull(existingLogin).mergeWithLogin(login)
```
Baron-Severin: It doesn't seem like `existingLogin` is guaranteed to be nonnull here.  Could we noop in that case instead of crashing?  It's what we're doing here with other errors anyway.

components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/DefaultLoginValidationDelegate.kt
```diff
@@ -24,27 +26,105 @@ class DefaultLoginValidationDelegate(
...
+                (savedLoginByGuid ?: savedLoginByElements)?.mergeWithLogin(newLogin)
+                    ?.copy(password = newLogin.password) ?: newLogin
+            val result = try {
+                storage.value.ensureValid(loginToCheck)
+                when {
+                    // GUIDs don't match, but everything else important matches so we can merge and update
+                    savedLoginByElements != null -> {
+                        Result.CanBeUpdated
+                    }
+                    newLogin.guid.isNullOrEmpty() || savedLoginByGuid == null -> {
```
grigoryk: If `newLogin.guid == null`, `savedLoginByGuid` will always be `null` as well. So this `||` should be unnecessary..?

components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/DefaultLoginValidationDelegate.kt
```diff
@@ -24,27 +26,105 @@ class DefaultLoginValidationDelegate(
...
+                        Result.CanBeCreated
+                    }
+                    newLogin.guid != savedLoginByGuid.guid -> {
```
grigoryk: If `storage.get` is functioning correctly, this is impossible. And I really hope `get` is functioning correctly :)

components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/GeckoLoginStorageDelegate.kt
```diff
@@ -78,68 +71,27 @@ class GeckoLoginStorageDelegate(
...
     @Synchronized
     override fun onLoginSave(login: Login) {
+        val validationDelegate = DefaultLoginValidationDelegate(storage = loginStorage)
         scope.launch {
             val existingLogin = login.guid?.let { loginStorage.value.get(it) }
-
-            when (getPersistenceOperation(login, existingLogin)) {
-                Operation.UPDATE -> {
-                    existingLogin?.let { loginStorage.value.update(it.mergeWithLogin(login)) }
+                ?: loginStorage.value.getStoredLoginByImportantFields(login)
```
grigoryk: This is quite inefficient - inside of `getNewLoginPersistenceAsync` you'll be running both of these "getters"!
It's also very complex to actually reason about what's going on here.

components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/DefaultLoginValidationDelegate.kt
```diff
@@ -24,27 +26,105 @@ class DefaultLoginValidationDelegate(
...
-                // We're setting guid=null here in order to trigger de-duping logic properly.
-                // Internally, the library ensures to not dupe records against themselves
-                // (via a `guid <> :guid` check), which means we need to "pretend" this is a new record,
-                // in order for `ensureValid` to actually throw `DUPLICATE_LOGIN`.
-                storage.value.ensureValid(login.copy(guid = null))
-                Result.CanBeCreated
+            val savedLoginByGuid = newLogin.guid?.let { storage.value.get(it) }
+            val savedLoginByElements = storage.value.getStoredLoginByImportantFields(newLogin)
+            // Check the merged login with the new password to display correct UI error states
+            val loginToCheck =
```
grigoryk: I'm trying to understand what's going on here, and for such a small function`getNewLoginPersistenceAsync` is _very_ hard to reason about. As-is, I don't think we can land it - just imagine trying to understand this a few months from now! I've been reading and re-reading it, and I can't quite tell if it's correct.

You basically have three cases to deal with:

1) trivial stuff - malformed record, or a complete dupe of an existing record
2) entirely new record - guid, username+origin, etc, nothing matches any existing records
3) existing record being updated - either matched by guid, or by "the essence" of the login. Note that `username` is immutable for a record, so a username change is same as creating an entirely new record and covered by case (2).
   - Also: We may or may not know `newLogin.guid` - what does that depend on? When will we have a `guid` set on `newLogin` and when will we not?

Can you update this code in a way that makes it very clear which cases you're dealing with, and how?

Also: all of this logic should live inside of the `logins` rust library. It'll be easier to understand, and much faster since I bet you can just write it out in SQL. This code runs during user input, so we should try to make it quick!

Having all of this logic live in one place will ensure that we can actually easily understand what's going on, and aren't duplicating ourselves. For an example of conceptual duplication, take a look at `getStoredLoginByImportantFields` and the dupe check within `ensureValid` - https://github.com/mozilla/application-services/blob/master/components/logins/src/db.rs#L819.

I don't mind if we have a concise, clear and correct implementation of this logic here first, but we _really_ should move it down a layer.

components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/GeckoLoginStorageDelegate.kt
ekager: Okay this was a very valid criticism :)
I changed this a bit so that the `CanBeUpdated` `Result` needs to pass which login we found back to the caller. Should help with being more explicit about what is going on here too.

components/service/sync-logins/src/test/java/mozilla/components/service/sync/logins/DefaultLoginValidationDelegateTest.kt
```diff
@@ -0,0 +1,203 @@
...
+class DefaultLoginValidationDelegateTest {
+
+    private lateinit var loginsStorage: LoginsStorage
+    private lateinit var validationDelegate: DefaultLoginValidationDelegate
+    private lateinit var scope: TestCoroutineScope
+
+    @Before
+    @Config(sdk = [21])
+    fun before() {
+        loginsStorage = mockLoginsStorage()
```
grigoryk: Let's use real storage implementation here (see `browser-storage-sync` for an example of how to hook that up.).
Your logic depends on behaviour of `ensureValid`, which you're not really testing below. Once you're using real storage implementation for these tests, the tests will actually be nicely end-to-end - you can insert real records to simulate these scenarios, and exercise our whole stack.

Once/if this code largely moves into Rust, test exercising our whole stack will remain valuable as integration tests. As they are now, they won't be testing much at that point.

What I'd like to see is a top-down approach here, working from the basic scenarios. Group the tests together so it's clear that we're testing all of these scenarios in an exhaustive way:
- various duplicate, invalid record scenarios
- various 'new record' scenarios. e.g. changed username, new record entirely, new username for an origin with an already saved login, etc.
- updated records - changed password, what else? which fields are considered immutable?

components/service/sync-logins/src/test/java/mozilla/components/service/sync/logins/DefaultLoginValidationDelegateTest.kt
```diff
@@ -0,0 +1,203 @@
...
+        scope = TestCoroutineScope()
+        validationDelegate =
+            DefaultLoginValidationDelegate(storage = lazy { loginsStorage }, scope = scope)
+    }
+
+    @Test
+    @Config(sdk = [21])
+    fun `WHEN guid is null or empty THEN should create a new record`() = runBlockingTest {
```
grigoryk: What if the record has a `null` guid, but matches an existing record exactly except for a password change?

grigoryk:
Filed an A-S issue for this API: https://github.com/mozilla/application-services/issues/2878

grigoryk:
Actually, new plan emerges! There's this: https://github.com/mozilla/application-services/pull/2542 - which is a better fit here.

For a given login record, it will return a list of matching records from the store using the deduping logic, but ignoring the username. Username is really the only change a user can make to a record to make it a dupe.

So, instead of all of the logic we have currently, we can instead:
- query this API, get a list back; if list is empty, we know we're creating
- as user is changing the username, check against the list we got back and see if there's a matching record (then, we can update if pw changed)
- if there isn't a matching record, we can create a new one (new username)

There's also a case of changing a username from empty to non-empty, while the password stays the same - on desktop, that triggers a "do you want to add a username to this login?" UI.

components/concept/storage/src/main/java/mozilla/components/concept/storage/LoginsStorage.kt
```diff
@@ -216,14 +229,14 @@ interface LoginValidationDelegate {
...
+     * Checks whether a [login] should be saved or updated, and whether or not [login] can be persisted.
      *
      * Note that this method is not thread safe.
      *
-     * @returns a [LoginValidationDelegate.Result], detailing whether [login] can be saved as a new
-     * value, used to update an existing one, or an error occured.
+     * @returns a [Result], detailing whether a [login] can and should be saved or updated or any error states
      */
-    fun validateCanPersist(login: Login): Deferred<Result>
+    fun getNewLoginPersistenceAsync(newLogin: Login): Deferred<Result>
```
grigoryk, line 239: nit: what about something like `shouldUpdateOrCreateAsync`? Seems a little less confusing

components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/DefaultLoginValidationDelegate.kt
```diff
@@ -24,27 +24,73 @@ class DefaultLoginValidationDelegate(
...
+                    newLogin
+                }
+                // If no usernames match in list, this is a new username, we know we're creating
+                !potentialDupes.any { it.username.isBlank() || it.username == newLogin.username } -> {
+                    shouldCreate = true
+                    newLogin
+                }
+                // Otherwise there's a matching record, let's merge with new password to update
+                else -> {
+                    potentialDupes.first().mergeWithLogin(newLogin)
```
grigoryk, line 48: This isn't quite right.

At this point, you know that:
- there _could_ be a record with an exact username match; if it exists, you should pick that record for updating (there could be more than one of these, too, in theory - but you can't depend on that being true or not, since then you'd be assuming behaviour details/restrictions internal to the underlying storage layer)
- otherwise, there _could_ be a record with a blank username and the same `guid` as `newLogin`. if it exists, you should pick that record for updating. Ideally, we'd have a separate `Result` for this case - desktop displays a "Add username to this login?" prompt

If none of those are true, you know that list of dupes isn't empty, there aren't any exact username matches, and there are no guid-matching records with a blank username.
If there are any non-guid-matching blank-username records, you could pick one of them for updating, to reduce number of total records present. `guid` could be mismatching for all sorts of sync-related reasons.

Otherwise, it's a "create a new record" case.

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/dialog/LoginDialogFragment.kt
```diff
@@ -179,19 +179,8 @@ internal class LoginDialogFragment : PromptDialogFragment() {
...
-                        context?.getString(R.string.mozac_feature_prompt_error_empty_password)
-                    )
-                } else {
-                    setViewState(
-                        confirmButtonEnabled = true,
-                        passwordErrorText = ""
-                    )
-                }
+                // Update accesses member state, so it must be called after password is set
+                update()
```
grigoryk, line 183: A password change can't cause a record to become a dupe (or stop being one). I don't think you need to query the store in this case at all.

ekager:
> Currently, you're calling `update` in LoginDialogFragment every time a username is changed. Which will trigger a `getNewLoginPersistenceAsync`.
> 
I ended up separating out the functions again to be able to have the `LoginDialogFragment` keep track of its own potential duplicates list. The logic in `shouldUpdateOrCreateAsync` now only checks for update vs create (and not error states). I figure we were eating the errors anyway (and the only one we would display is empty password which we can do without validating), so this should be fine for the UI layer. For the storage layer `GeckoLoginStorageDelegate`, we should actually be validating so we don't crash on some unexpected error/we can eat the duplicate logins, so the `ensureValid` is now split out for only the storage layer to call. Wasn't sure if this was the right direction, but it seemed like splitting it up into new functions was the only way to make sure we aren't calling `getPotentialDupesIgnoringUsername` multiple times, but can still check the update/create logic correctly 🤔

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/dialog/LoginDialogFragment.kt
```diff
@@ -71,7 +70,10 @@ internal class LoginDialogFragment : PromptDialogFragment() {
     private var username by SafeArgString(KEY_LOGIN_USERNAME)
     private var password by SafeArgString(KEY_LOGIN_PASSWORD)
 
-    private var stateUpdate: Job? = null
+    private var validateStateUpdate: Job? = null
+
+    // List of potential dupes ignoring username
+    private var potentialDupesList: List<Login>? = null
```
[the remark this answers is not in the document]
ekager: Do you think this ^ will work? Synchronizing the getter and setter? I also just think the syntax is cute 💁🏼‍♀️

components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/DefaultLoginValidationDelegate.kt
```diff
@@ -24,27 +24,94 @@ class DefaultLoginValidationDelegate(
...
+        return scope.async {
+            val potentialDupesList =
+                potentialDupes ?: storage.value.getPotentialDupesIgnoringUsername(newLogin)
+            val foundLogin = if (potentialDupesList.isEmpty()) {
+                // If list is empty, no records match -> create
+                null
+            } else {
+                // Matching username -> update
+                potentialDupesList.find { it.username == newLogin.username }
+                // Matching guid, blank username -> update
```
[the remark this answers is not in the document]
ekager: filed https://github.com/mozilla-mobile/fenix/issues/9642

ekager:
These updated tests pass locally for me... But on CI they're all failing with
```
[task 2020-04-02T22:29:01.793Z] 	at mozilla.components.service.sync.logins.SyncableLoginsStorage$wipeLocal$2.invokeSuspend(SyncableLoginsStorage.kt:149)
[task 2020-04-02T22:29:01.793Z] 	(Coroutine boundary)
[task 2020-04-02T22:29:01.793Z] 	at mozilla.components.service.sync.logins.DefaultLoginValidationDelegateTest$before$1.invokeSuspend(DefaultLoginValidationDelegateTest.kt:41)
[task 2020-04-02T22:29:01.793Z] Caused by: java.lang.NoClassDefFoundError: Could not initialize class org.mozilla.appservices.logins.GleanMetrics.LoginsStore
[task 2020-04-02T22:29:01.793Z] 	at mozilla.appservices.logins.DatabaseLoginsStorage$unlockCounters$2.invoke(DatabaseLoginsStorage.kt:380)
[task 2020-04-02T22:29:01.793Z] 	at mozilla.appservices.logins.DatabaseLoginsStorage$unlockCounters$2.invoke(DatabaseLoginsStorage.kt:32)
[task 2020-04-02T22:29:01.793Z] 	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
[task 2020-04-02T22:29:01.793Z] 	at mozilla.appservices.logins.DatabaseLoginsStorage.getUnlockCounters(DatabaseLoginsStorage.kt)
[task 2020-04-02T22:29:01.793Z] 	at mozilla.appservices.logins.DatabaseLoginsStorage.unlock(DatabaseLoginsStorage.kt:73)
[task 2020-04-02T22:29:01.793Z] 	at mozilla.components.service.sync.logins.LoginStorageConnection.init$service_sync_logins_debug(SyncableLoginsStorage.kt:283)
[task 2020-04-02T22:29:01.793Z] 	at mozilla.components.service.sync.logins.SyncableLoginsStorage$conn$2.invoke(SyncableLoginsStorage.kt:122)
[task 2020-04-02T22:29:01.793Z] 	at mozilla.components.service.sync.logins.SyncableLoginsStorage$conn$2.invoke(SyncableLoginsStorage.kt:114)
[task 2020-04-02T22:29:01.793Z] 	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
[task 2020-04-02T22:29:01.793Z] 	at mozilla.components.service.sync.logins.SyncableLoginsStorage.getConn(SyncableLoginsStorage.kt)
[task 2020-04-02T22:29:01.793Z] 	at mozilla.components.service.sync.logins.SyncableLoginsStorage.access$getConn$p(SyncableLoginsStorage.kt:114)
[task 2020-04-02T22:29:01.793Z] 	at mozilla.components.service.sync.logins.SyncableLoginsStorage$wipeLocal$2.invokeSuspend(SyncableLoginsStorage.kt:149)
[task 2020-04-02T22:29:01.793Z] 	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
[task 2020-04-02T22:29:01.794Z] 	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:56)
[task 2020-04-02T22:29:01.794Z] 	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
[task 2020-04-02T22:29:01.794Z] 	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:738)
[task 2020-04-02T22:29:01.794Z] 	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678)
[task 2020-04-02T22:29:01.794Z] 	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)
```

grigoryk:
@ekager try adding `testImplementation Dependencies.mozilla_glean_forUnitTests` to the `build.gradle` file, it'll probably fix that failure.

components/service/sync-logins/src/test/java/mozilla/components/service/sync/logins/DefaultLoginValidationDelegateTest.kt
```diff
@@ -0,0 +1,233 @@
...
+    @Before
+    fun before() = runBlocking {
+        loginsStorage.wipeLocal()
+        scope = TestCoroutineScope()
+        validationDelegate =
+            DefaultLoginValidationDelegate(storage = lazy { loginsStorage }, scope = scope)
+    }
+
+    @Test
+    fun `WHEN potential dupes list is empty, should create`() {
```
grigoryk, line 48: shouldn't these tests be `runBlocking`? to wait for completion of the `launch`?

components/service/sync-logins/src/test/java/mozilla/components/service/sync/logins/DefaultLoginValidationDelegateTest.kt
```diff
@@ -0,0 +1,233 @@
...
+        scope.launch {
+            val newLogin = createLogin(guid = "1")
+            val existingLogin = createLogin(
+                guid = "1",
+                password = "password1"
+            )
+            loginsStorage.add(existingLogin)
+            val operation = validationDelegate.shouldUpdateOrCreateAsync(newLogin).await()
+            assertEquals(Result.CanBeUpdated(existingLogin), operation)
+            loginsStorage.delete("1")
```
grigoryk, line 67: `@before` function does a `wipeLocal` of the storage, is that not enough?

components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/SyncableLoginsStorage.kt
```diff
@@ -194,7 +194,6 @@ class SyncableLoginsStorage(
      */
     @Throws(IdCollisionException::class, InvalidRecordException::class, LoginsStorageException::class)
     override suspend fun add(login: Login): String = withContext(coroutineContext) {
-        check(login.guid == null) { "'guid' for a new login must be `null`" }
```
ekager, line 197: @grigoryk this was breaking the tests because we want to be able to test for matching guids. Think we can get rid of this check or should I make a testing-specific add?

ekager: Updated

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/dialog/LoginDialogFragment.kt
```diff
@@ -71,7 +71,17 @@ internal class LoginDialogFragment : PromptDialogFragment() {
     private var username by SafeArgString(KEY_LOGIN_USERNAME)
     private var password by SafeArgString(KEY_LOGIN_PASSWORD)
 
-    private var stateUpdate: Job? = null
+    private var validateStateUpdate: Job? = null
+
+    /**
+     * These are essential to the function of this fragment.
+     */
+    private val loginValidationDelegate by lazy {
```
grigoryk, line 79: nit: do you need this to be `lazy`? that has overhead we may not need here. And below, replace `requireNotNull` with just `!!`?

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/dialog/LoginDialogFragment.kt
```diff
@@ -224,42 +233,42 @@ internal class LoginDialogFragment : PromptDialogFragment() {
...
+                this@LoginDialogFragment.potentialDupesList = CopyOnWriteArrayList(
+                    loginValidationDelegate.getPotentialDupesIgnoringUsernameAsync(
+                        login
+                    ).await()
+                )
+                potentialDupesList?.toList()
+            } else {
+                potentialDupesList?.toList()
+            }
+            validateDeferred =
```
grigoryk, line 251: nit: move the `.toList()` on the invocation line, instead, makes the intent of passing in a copy of the list clearer. it's also really worth your time to leave a comment here explaining why we're passing in a copy, and not the original.

this will also let you simplify the above statement, removing the `else` clause.

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/dialog/LoginDialogFragment.kt
```diff
@@ -71,7 +71,17 @@ internal class LoginDialogFragment : PromptDialogFragment() {
...
+
+    /**
+     * These are essential to the function of this fragment.
+     */
+    private val loginValidationDelegate by lazy {
+        requireNotNull(requireNotNull(feature).loginValidationDelegate)
+    }
+
+    // List of potential dupes ignoring username
+    private var potentialDupesList: CopyOnWriteArrayList<Login>? = null
```
grigoryk, line 84: whenever you use concurrency primitives, it's a really good idea to describe _why_, and what concerns you're trying to address. Otherwise, the only place this is currently documented is in the github discussion.

components/service/sync-logins/build.gradle
```diff
@@ -27,12 +27,29 @@ android {
...
+    jnaForTest
+}
+
 dependencies {
     // Parts of this dependency are typealiase'd or are otherwise part of this module's public API.
     api(Dependencies.mozilla_sync_logins) {
-      // Use our own version of the Glean dependency,
-      // which might be different from the version declared by A-S.
-      exclude group: 'org.mozilla.components', module: 'service-glean'
+        // Use our own version of the Glean dependency,
```
grigoryk, line 50: nit: unnecessary whitespace churn

ekager:
bors r=grigoryk

PR #6321, opened 2020-03-19
Closes #6315: Include engine session when migrating custom tab

### LINKED ISSUE #6315: No browser action badges when moving tabs from custom tab
STR

Install uBlock on Fenix
Open page in Fenix custom tab
Three dots -> Open in Fenix Nightly
Three dots -> notice no badge for uBlock
Refresh page -> still no badge

Screenshot_20200318-155009_Firefox Nightly

csadilek:
When migrating a custom tab to a regular tab we forgot to migrate the engine session. This is the easy solution.

Alternatively, we could migrate the engine state as part of `toTabSessionState`: https://github.com/mozilla-mobile/android-components/blob/master/components/browser/session/src/main/java/mozilla/components/browser/session/ext/SessionExtensions.kt#L20

This affects a lot of parts though and is probably not needed for other callers e.g. `add` and `restore`? In those cases we either don't have an engine session or we will link one anyway.


csadilek:
bors r=pocmo

PR #6340, opened 2020-03-23
Closes #4632. Add contains function for suggestion providers

### LINKED ISSUE #4632: Change suggestion providers to be a mutableSet, not a mutableList
Since we're currently just crashing if we try to add a `SuggestionProvider` that has the same ID as one already in use, why not just make this into a set?

Doing so would allow us in Fenix to [easily update which providers are available](https://github.com/mozilla-mobile/fenix/pull/5778/files#diff-de2161f65b2eebcd91d886bc5d00bbf3R228-R238) without having to keep track of this internally in Fenix.

It's a lot of bookkeeping and it's likely very useful for all consumers of AwesomeBar to be able to find out if their provider is already in the recyclerview. Either a `contains` function that checks the ID or using a set would work here 😄

levimatheri:
---

MutableSet would have been better but hashCode() and equals() would have to be generated which doesn't work with interfaces. Thoughts?

### Pull Request checklist

### After merge


components/browser/awesomebar/src/main/java/mozilla/components/browser/awesomebar/BrowserAwesomeBar.kt
```diff
@@ -95,12 +95,12 @@ class BrowserAwesomeBar @JvmOverloads constructor(
     @Synchronized
     override fun addProviders(vararg providers: AwesomeBar.SuggestionProvider) {
         providers.forEach { provider ->
-            val existingProvider = this.providers.find { it.id == provider.id }
-            existingProvider?.let {
+            if (containsProvider(provider)) {
                 throw IllegalStateException("Failed to add provider " +
                         "${provider.id} of type ${provider::class.java.name}. " +
-                        "Provider with the same ID already exists: ${it::class.java.name}")
+                        "Provider with the same ID already exists.")
```
csadilek: I think it's valuable to have the other providers type in the exception message. Theoretically, there could be two providers with the same ID but different types (or packages/versions) on the classpath. And that would be good to know as part of the exception.

My vote is to keep the original implementation of this method, and applications can use the new `containsProvider`.

components/browser/awesomebar/src/test/java/mozilla/components/browser/awesomebar/BrowserAwesomeBarTest.kt
```diff
@@ -385,6 +385,29 @@ class BrowserAwesomeBarTest {
...
     }
 
+    @Test
+    fun `provider exists for the awesome bar`() {
+        val awesomeBar = BrowserAwesomeBar(testContext)
+        val provider1 = mockProvider()
+        `when`(provider1.id).thenReturn("1")
+
+        awesomeBar.addProviders(provider1)
+        assertEquals(true, awesomeBar.containsProvider(provider1))
```
csadilek: nit: `assertTrue(awesomeBar.containsProvider(provider1))`

components/browser/awesomebar/src/test/java/mozilla/components/browser/awesomebar/BrowserAwesomeBarTest.kt
```diff
@@ -385,6 +385,29 @@ class BrowserAwesomeBarTest {
...
+    fun `provider does not exist for the awesome bar`() {
+        val awesomeBar = BrowserAwesomeBar(testContext)
+        val provider1 = mockProvider()
+        `when`(provider1.id).thenReturn("1")
+
+        val provider2 = mockProvider()
+        `when`(provider2.id).thenReturn("2")
+
+        awesomeBar.addProviders(provider1)
+        assertEquals(false, awesomeBar.containsProvider(provider2))
```
csadilek: nit: `assertTrue(awesomeBar.containsProvider(provider2))`

components/browser/awesomebar/src/test/java/mozilla/components/browser/awesomebar/BrowserAwesomeBarTest.kt
```diff
@@ -385,6 +385,29 @@ class BrowserAwesomeBarTest {
...
+        val awesomeBar = BrowserAwesomeBar(testContext)
+        val provider1 = mockProvider()
+        `when`(provider1.id).thenReturn("1")
+
+        awesomeBar.addProviders(provider1)
+        assertEquals(true, awesomeBar.containsProvider(provider1))
+    }
+
+    @Test
+    fun `provider does not exist for the awesome bar`() {
```
csadilek: Since the two tests are essentially the same, let's combine them into one, e.g:

```Kotlin
@Test
fun `containsProvider - checks if provider exists`() {
val awesomeBar = BrowserAwesomeBar(testContext)
    val provider1 = mockProvider()
    `when`(provider1.id).thenReturn("1")

    val provider2 = mockProvider()
    `when`(provider2.id).thenReturn("2")

    awesomeBar.addProviders(provider1)
    assertTrue(awesomeBar.containsProvider(provider1))
    assertFalse(awesomeBar.containsProvider(provider2))
}
```

components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
```diff
@@ -33,6 +33,11 @@ interface AwesomeBar {
      */
     fun removeAllProviders()
 
+    /**
+     * Returns true if this awesome bar contains the following [SuggestionProvider]
```
csadilek: nit: Returns whether or not this....

components/browser/awesomebar/src/main/java/mozilla/components/browser/awesomebar/BrowserAwesomeBar.kt
levimatheri: I agree. Will make the changes you suggested

levimatheri:
@csadilek Looks good?

PR #6346, opened 2020-03-23
Closes #6344: Use activityName instead of packageName as primary key.

### LINKED ISSUE #6344: RecentApp should use activity name as a primary key
Using `packageName` as a primary key is not viable, as multiple options in share options could have the same package name. For example, "Save to drive" and "Copy to clipboard" both have `com.google.android.apps.docs`.
This leads to cases where options could be added to recent apps list in Fenix if another option with same package name has been used.

mcarare:
---

### Pull Request checklist

### After merge

https://github.com/mozilla-mobile/fenix/pull/9317

### DIFF
--- a/components/feature/share/src/main/java/mozilla/components/feature/share/db/RecentAppsDatabase.kt
+++ b/components/feature/share/src/main/java/mozilla/components/feature/share/db/RecentAppsDatabase.kt
@@ -8,11 +8,14 @@ import android.content.Context
 import androidx.room.Database
 import androidx.room.Room
 import androidx.room.RoomDatabase
+import androidx.room.migration.Migration
+import androidx.sqlite.db.SupportSQLiteDatabase
+import mozilla.components.feature.share.db.RecentAppsDatabase.Companion.RECENT_APPS_TABLE
 
 /**
  * Internal database for storing apps and their scores that determine if they are most recently used.
  */
-@Database(entities = [RecentAppEntity::class], version = 1)
+@Database(entities = [RecentAppEntity::class], version = 2)
 internal abstract class RecentAppsDatabase : RoomDatabase() {
     abstract fun recentAppsDao(): RecentAppsDao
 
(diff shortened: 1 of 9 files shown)


mcarare:
> As we are changing the structure of the database, [we need to increase the database version number](https://github.com/mozilla-mobile/android-components/blob/260d9b43ebe00da2b0bf929fbacc2fbc84f67e83/components/feature/share/src/main/java/mozilla/components/feature/share/db/RecentAppsDatabase.kt#L15) and add a migration for deleting the `packageName` column.

I'not sure if we can migrate de current entries, because we might not know the activityName that corresponds to the existing packageNames.

Amejia481:
> > As we are changing the structure of the database, [we need to increase the database version number](https://github.com/mozilla-mobile/android-components/blob/260d9b43ebe00da2b0bf929fbacc2fbc84f67e83/components/feature/share/src/main/java/mozilla/components/feature/share/db/RecentAppsDatabase.kt#L15) and add a migration for deleting the `packageName` column.
> 
> I'not sure if we can migrate de current entries, because we might not know the activityName that corresponds to the existing packageNames.

As we are creating a new field on the table(`activityName`) the old one (`packageName`) will be remained there, what I'm suggesting is in the migration code is to drop the `packageName` field as it contains information that we are not going to use anymore.

.buildconfig.yml
```diff
@@ -1,4 +1,4 @@
-componentsVersion: 37.0.0
+componentsVersion: 37.0.20200322190046
```
Amejia481: not sure if we want to include this change

mcarare: I did not :)

components/feature/share/src/main/java/mozilla/components/feature/share/db/RecentAppsDatabase.kt
```diff
@@ -31,9 +35,28 @@ internal abstract class RecentAppsDatabase : RoomDatabase() {
...
             }
         }
     }
 }
+
+internal object Migrations {
+    val migration_1_2 = object : Migration(1, 2) {
+        override fun migrate(database: SupportSQLiteDatabase) {
+
+            database.execSQL("DROP TABLE RECENT_APPS_TABLE")
```
Amejia481: 👍

Amejia481:
> > > As we are changing the structure of the database, [we need to increase the database version number](https://github.com/mozilla-mobile/android-components/blob/260d9b43ebe00da2b0bf929fbacc2fbc84f67e83/components/feature/share/src/main/java/mozilla/components/feature/share/db/RecentAppsDatabase.kt#L15) and add a migration for deleting the `packageName` column.
> > 
> > 
> > I'not sure if we can migrate de current entries, because we might not know the activityName that corresponds to the existing packageNames.
> 
> As we are creating a new field on the table(`activityName`) the old one (`packageName`) will be remained there, what I'm suggesting is in the migration code is to drop the `packageName` field as it contains information that we are not going to use anymore.

Go it, the whole entries becomes invalid, I thought we have info that we could reuse 👍

PR #6360, opened 2020-03-24
Closes issue #6388: Add remove functionality to the RecentAppsStorage

DrCesar:
---

### DIFF
--- a/components/feature/share/src/main/java/mozilla/components/feature/share/RecentAppsStorage.kt
+++ b/components/feature/share/src/main/java/mozilla/components/feature/share/RecentAppsStorage.kt
@@ -38,6 +38,14 @@ class RecentAppsStorage(context: Context) {
     }
 
     /**
+     * Deletes an app form the recent apps list
+     * @param activityName - name of the activity of the app
+     */
+    fun deleteRecentApp(activityName: String) {
+        database.value.recentAppsDao().deleteRecentApp(activityName)
+    }
+
+    /**
      * Get a descending ordered list of the most recent apps
      * @param limit - size of list
      */
--- a/components/feature/share/src/main/java/mozilla/components/feature/share/db/RecentAppsDao.kt
+++ b/components/feature/share/src/main/java/mozilla/components/feature/share/db/RecentAppsDao.kt
@@ -23,6 +23,11 @@ internal interface RecentAppsDao {
     @Update
     fun updateRecentApp(recentApp: RecentAppEntity)
 
+    @Query("""
+        DELETE FROM recent_apps_table
+        WHERE activityName = :activityName""")
+    fun deleteRecentApp(activityName: String)
+
     @Update
     fun updateAllRecentApp(recentApps: List<RecentAppEntity>)
 
--- a/components/feature/share/src/test/java/mozilla/components/feature/share/RecentAppStorageTest.kt
+++ b/components/feature/share/src/test/java/mozilla/components/feature/share/RecentAppStorageTest.kt
@@ -102,4 +102,14 @@ class RecentAppStorageTest {
 
         verify(recentAppsDao).insertRecentApp(result)
     }
+
+    @Test
+    fun `delete an app from our database`() {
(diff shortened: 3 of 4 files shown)


Amejia481:
Hi @DrCesar,
Thanks for putting this pr.
Is there any Fenix related issue where I can find more info about this new feature?

DrCesar:
Hi, yes the issue in feniz is this https://github.com/mozilla-mobile/fenix/issues/8759.
I added this so an RecentShareApp can be deleted from the list.

components/feature/share/src/main/java/mozilla/components/feature/share/db/RecentAppsDao.kt
```diff
@@ -23,6 +24,9 @@ internal interface RecentAppsDao {
     @Update
     fun updateRecentApp(recentApp: RecentAppEntity)
 
+    @Delete
+    fun deleteRecentApp(packageName: RecentAppEntity)
```
Amejia481: The parameter `packageName` should be renamed to `activityName`as now it is the[ primary key](https://github.com/mozilla-mobile/android-components/blob/311864c26b1b3c779936701cbcb3ca4a99dff014/components/feature/share/src/main/java/mozilla/components/feature/share/db/RecentAppEntity.kt#L19).

Also we can take advantage that `activityName` is primary key to improve the delete statement.
Instead of passing the object, we could pass the primary key (`activityName`), this way we can avoid having to query the table before and then deleting. You can find an example [here](https://github.com/mozilla-mobile/android-components/blob/311864c26b1b3c779936701cbcb3ca4a99dff014/components/feature/addons/src/main/java/mozilla/components/feature/addons/update/db/UpdateAttemptDao.kt#L23)

components/feature/share/src/main/java/mozilla/components/feature/share/RecentAppsStorage.kt
```diff
@@ -37,6 +37,17 @@ class RecentAppsStorage(context: Context) {
         }
     }
 
+    /**
+     * Deletes an app form the recent apps list
+     * @param activityName - name of the activity of the app
+     */
+    fun deleteRecentApp(activityName: String) {
```
Amejia481: It will be nice to have a test for this, you could find some similar tests [here](https://github.com/mozilla-mobile/android-components/blob/master/components/feature/share/src/test/java/mozilla/components/feature/share/RecentAppStorageTest.kt) :)

components/feature/share/src/main/java/mozilla/components/feature/share/RecentAppsStorage.kt
```diff
@@ -37,6 +37,17 @@ class RecentAppsStorage(context: Context) {
...
+        val recentAppEntity = database.value.recentAppsDao().getRecentApp(activityName)
+        recentAppEntity?.let {
+            database.value.recentAppsDao().deleteRecentApp(recentAppEntity);
```
Amejia481: See https://github.com/mozilla-mobile/android-components/pull/6360#discussion_r398069021

DrCesar:
Ok thanks for the help. Im going to fix the pr. And sorry for not linking the Issue of fenix, it would have helped.

Amejia481:
No worries, I'm glad to help :)

components/feature/share/src/main/java/mozilla/components/feature/share/RecentAppsStorage.kt
DrCesar: Added the test for RecentAppsStorage.deleteRecentApp

DrCesar:
> 
> 
> @DrCesar thanks for this patch!
> Before approving the patch, I squashed the commits and then rebased :)

Thanks @Amejia481. Yeah I should have cleaned those commits sorry for the mess. And thanks for all the help.

Amejia481:
No worries, you did a great job 👍 
I'm always glad to help!

PR #6371, opened 2020-03-24
Closes #6366: Improve service-experiments (Mako) startup performance

### LINKED ISSUE #6366: Improve service-experiments (Mako) `initialize` performance.
Fenix identified Mako initialization as a factor in startup performance.  We can address this by pushing the init off of the main thread and using lazy instantiation, where appropriate.

travis79:
- Adds a private coroutine scope and uses it to initialize Mako off of the calling thread so as to not block it.
- Updates to tests that rely on `initialize` making them await the async operation so as to not break tests.

### DIFF
--- a/components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
+++ b/components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
@@ -9,6 +9,11 @@ import android.content.Context
 import mozilla.components.support.base.log.logger.Logger
 import mozilla.components.service.glean.Glean
 import androidx.annotation.VisibleForTesting
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.ObsoleteCoroutinesApi
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.newSingleThreadContext
 import java.io.File
 
 /**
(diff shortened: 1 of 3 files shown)


components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
```diff
@@ -53,49 +64,53 @@ open class ExperimentsInternalAPI internal constructor() {
...
+     *
+     * @return Returns a [Job] so that off-thread initialization can be monitored and interacted
+     * with.
      */
     fun initialize(
         applicationContext: Context,
         configuration: Configuration = Configuration(),
         onExperimentsUpdated: (() -> Unit)? = null
-    ) {
+    ): Job? {
```
Dexterp37: Do we *need* to provide external consumers the ability to monitor the experiments initialization? If yes, why?

If this is only needed in tests, we can createa an `internal` variable from `coroutineScope.launch` and use that for waiting.

travis79: Good idea! I just pushed this change.

components/service/experiments/src/main/java/mozilla/components/service/experiments/Experiments.kt
```diff
@@ -44,6 +49,16 @@ open class ExperimentsInternalAPI internal constructor() {
 
     private var experimentsUpdatedCallback: (() -> Unit)? = null
 
+    /**
+     * A [CoroutineScope] used to dispatch initialization as to not block the calling thread.
+     */
+    @ObsoleteCoroutinesApi
+    private val coroutineScope by lazy { CoroutineScope(newSingleThreadContext("Mako")) }
```
Dexterp37:
```suggestion
    private val coroutineScope by lazy { CoroutineScope(newSingleThreadContext("MakoInitializerThread")) }
```

travis79: Done!

travis79:
bors r=@Dexterp37

Issue #5866, opened 2020-02-10
ActivityNotFoundException: No Activity found to handle Intent { act=android.intent.action.VIEW cat=[mozilla.components.pwa.c...


pocmo:
https://sentry.prod.mozaws.net/operations/fenix-fennec-nightly/issues/7183008/

```
ActivityNotFoundException: No Activity found to handle Intent { act=android.intent.action.VIEW cat=[mozilla.components.pwa.category.SHORTCUT] dat= pkg=org.mozilla.fennec_aurora (has extras) }
    at android.app.Instrumentation.checkStartActivityResult(Instrumentation.java:2014)
    at android.app.Instrumentation.execStartActivity(Instrumentation.java:1675)
    at android.app.Activity.startActivityForResult(Activity.java:4586)
    at androidx.fragment.app.FragmentActivity.startActivityForResult(FragmentActivity.java:6)
    at android.app.Activity.startActivityForResult(Activity.java:4544)
...
(17 additional frame(s) were not displayed)

No Activity found to handle Intent { act=android.intent.action.VIEW cat=[mozilla.components.pwa.category.SHORTCUT] dat= pkg=org.mozilla.fennec_aurora (has extras) }
```

pocmo:
Sentry issue: [FENIX-FENNEC-BETA-1S](https://sentry.prod.mozaws.net/operations/fenix-fennec-beta/issues/7658852/)

PR #6374, opened 2020-03-24
For #5866: Don't crash when failed to launch new window request in custom tab

rocketsroger:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/customtabs/build.gradle
+++ b/components/feature/customtabs/build.gradle
@@ -39,6 +39,7 @@ dependencies {
     implementation project(':support-ktx')
     implementation project(':support-utils')
     implementation project(':ui-icons')
+    implementation project(':lib-crash')
 
     implementation Dependencies.androidx_core_ktx
     implementation Dependencies.kotlin_stdlib
--- a/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
+++ b/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
@@ -5,7 +5,9 @@
 package mozilla.components.feature.customtabs
 
 import android.app.Activity
+import android.content.ActivityNotFoundException
 import androidx.annotation.VisibleForTesting
+import androidx.annotation.VisibleForTesting.PRIVATE
 import androidx.browser.customtabs.CustomTabsIntent
 import androidx.core.net.toUri
 import kotlinx.coroutines.CoroutineScope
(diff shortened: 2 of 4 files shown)


jonalmeida:
I'm not sure what the context for the fix is so this may be out of place:

Shouldn't we figure out why we weren't able to handle the intent since it looks like it was meant for ourselves (`mozilla.components.pwa.category.SHORTCUT`) instead of try-catching it?

```
No Activity found to handle Intent { act=android.intent.action.VIEW cat=[mozilla.components.pwa.category.SHORTCUT] dat= pkg=org.mozilla.fennec_aurora (has extras) }
```

rocketsroger:
> I'm not sure what the context for the fix is so this may be out of place:
> 
> Shouldn't we figure out why we weren't able to handle the intent since it looks like it was meant for ourselves (`mozilla.components.pwa.category.SHORTCUT`) instead of try-catching it?
> 
> ```
> No Activity found to handle Intent { act=android.intent.action.VIEW cat=[mozilla.components.pwa.category.SHORTCUT] dat= pkg=org.mozilla.fennec_aurora (has extras) }
> ```

Agreed, this is not a fix.  This is more to avoid these crashes by catching them and sending them as a caught exception.  There's a bigger underlying issue here.  I'm not closing this issue yet and will be continue to investigate.  However, I want to avoid crashing first.

rocketsroger:
> LGTM but leaving it to you to merge since you might have wanted a review from Sebastian as well from the looks of it.

Your approval is just as good :)

rocketsroger:
bors r=jonalmeida

ValentinTimisica:
Hello @rocketsroger, @jonalmeida, I already had for a long time a PR created for this crash which actually fixes the issue.

You can check it [here](https://github.com/mozilla-mobile/android-components/pull/6122).
#6094 is the AC issue and [this](https://github.com/mozilla-mobile/fenix/issues/8412) is the Fenix issue (which contain the result of my investigation with steps to reproduce) for which I also have a [PR](https://github.com/mozilla-mobile/fenix/pull/8903) (with directions from UX to show a message to the user).

I would appreciate if you guys could review my patch.

rocketsroger:
> Hello @rocketsroger, @jonalmeida, I already had for a long time a PR created for this crash which actually fixes the issue.

Sorry I didn't notice your PR before merging mine.  However, mine PR is not meant to fix this issue.  There is a underlying issue that we need to investigate here.  My change is just to avoid crashing but still report the crash so it is being tracked.

ValentinTimisica:
> > Hello @rocketsroger, @jonalmeida, I already had for a long time a PR created for this crash which actually fixes the issue.
> 
> Sorry I didn't notice your PR before merging mine. However, mine PR is not meant to fix this issue. There is a underlying issue that we need to investigate here. My change is just to avoid crashing but still report the crash so it is being tracked.

I investigated and found the underlying issue: any url touches in web pages gets in AC which will then try to start a client activity to handle that specific url.

In our case (Fenix) and in other clients's cases also there might not be an activity with an intent filter to match that url's scheme and this will cause an ActivityNotFoundException in AC.
We can't fix this in AC because is a client issue that can only be fixed there by registering new intent filters for the new schemes.
The best what we can do in this case is pass that information to the client - that it cannot handle that specific url.

rocketsroger:
> We can't fix this in AC because is a client issue that can only be fixed there by registering new intent filters for the new schemes.

Yes with the current way that we have no way of fixing this.  But I was thinking there might be another way if A-C launch new window request differently in custom tab or understand that this is not a supported URL in GeckoView and launch in a third party app. (similar to applinks feature)

Issue #5529, opened 2020-01-09
History of Single Session Saved in Collection Does Not Restore


ekager:
Originally filed in Fenix: https://github.com/mozilla-mobile/fenix/issues/5208

We are still _trying_ to save with [`addTabsToCollection`](https://github.com/mozilla-mobile/android-components/blob/master/components/feature/tab-collections/src/main/java/mozilla/components/feature/tab/collections/TabCollectionStorage.kt#L51) and [restore](https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt#L174) and it looks like page and loading state is restored correctly, but history no longer is. 

Did something change in the save/restore of individual sessions that would remove history?

ekager:
Just checked this today and looks like loading state is no longer restored either. Logging showed we _are_ restoring. 

I also wonder if this has to do with: https://github.com/mozilla-mobile/fenix/issues/9039

pocmo:
Ah, yeah, it looks like we restore the session and the engine state and then only return the pure session while dropping the engine state on the floor. I'll prepare a patch.

Issue #6380, opened 2020-03-25
Fix on-device tests in feature-tab-collections


pocmo:
I noticed that some of the on-device tests in `feature-tab-collections` are failing. Since we are not running those tests in automation yet, we may just have missed those failures. Let's fix them!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-5352)

pocmo:
Once they are fixed let's also add tests for the `SessionManager.restore()` extension methods.

csadilek:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1795373

Change performed by the [Move to Bugzilla add-on](https://addons.mozilla.org/en-US/firefox/addon/move-to-bugzilla/).

PR #6381, opened 2020-03-25
Issue #5529: Restore engine state of tabs in tab collections.

pocmo:
Changes here:

- `TabCollection` and `Tab` return a `Snapshot.Item` instead of only a `Session`
- `SessionManager.add()` allows to pass in an `EngineSessionState`
- `SessionManager.restore()` helper functions to restore a `Tab`/`TabCollection`

### DIFF
--- a/components/browser/session/src/test/java/mozilla/components/browser/session/SessionManagerMigrationTest.kt
+++ b/components/browser/session/src/test/java/mozilla/components/browser/session/SessionManagerMigrationTest.kt
@@ -1281,6 +1281,58 @@ class SessionManagerMigrationTest {
         verify(engineSession3, never()).close()
         verify(engineSession4, never()).close()
     }
+
+    @Test
+    fun `Adding session with engine session state`() {
+        val store = BrowserStore()
+        val manager = SessionManager(engine = mock(), store = store)
+
+        val state: EngineSessionState = mock()
+        val session = Session("https://www.mozilla.org")
+
+        manager.add(session, engineSessionState = state)
+
+        assertEquals(state, session.engineSessionHolder.engineSessionState)
+        assertNull(session.engineSessionHolder.engineSession)
+
+        assertEquals(state, store.state.tabs[0].engineState.engineSessionState)
+        assertNull(store.state.tabs[0].engineState.engineSession)
+    }
+
+    @Test
+    fun `Adding session with engine session and engine session state`() {
+        val store = BrowserStore()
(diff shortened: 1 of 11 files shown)


pocmo:
Seeing test failures in `feature-search` and `feature-customtabs`. Investigating..

pocmo:
Fixed! Needed to update some matchers to use the new `add()` API.

components/browser/session/src/test/java/mozilla/components/browser/session/SessionManagerMigrationTest.kt
```diff
@@ -1281,6 +1281,58 @@ class SessionManagerMigrationTest {
...
+
+        assertNull(session.engineSessionHolder.engineSessionState)
+        assertEquals(engineSession, session.engineSessionHolder.engineSession)
+
+        assertNull(store.state.tabs[0].engineState.engineSessionState)
+        assertEquals(engineSession, session.engineSessionHolder.engineSession)
+    }
+
+    @Test
+    fun `Adding session with engine session session`() {
```
csadilek: nit: session session :)

components/feature/tab-collections/src/main/java/mozilla/components/feature/tab/collections/ext/SessionManager.kt
```diff
@@ -0,0 +1,71 @@
...
+import mozilla.components.browser.session.SessionManager
+import mozilla.components.concept.engine.Engine
+import mozilla.components.feature.tab.collections.Tab
+import mozilla.components.feature.tab.collections.TabCollection
+
+/**
+ * Restores the given [Tab] from a [TabCollection]. Will invoke [onTabRestored] on successful restore
+ * and [onFailure] otherwise.
+ */
+fun SessionManager.restore(
```
csadilek: These methods are not used yet, right? Should we add tests for it? We can also file a follow-up to cover them.

pocmo: They are getting used in this PR: https://github.com/mozilla-mobile/fenix/pull/9399

We only have on-device tests for this component (because of Room) and I noticed that they are failing and we didn't notice since they do not run in automation. I filed an issue to fix them. As part of that we could add tests for those methods too:
https://github.com/mozilla-mobile/android-components/issues/6380

pocmo:
bors r=csadilek

PR #6393, opened 2020-03-26
Closes #6317: Prevent inserting duplicate record into content resolver

### LINKED ISSUE #6317: Downloading 50mb+ files from thinkbroadband.com makes the browser to crash
### Steps to reproduce
1. Go to thinkbroadband.com/download;
2. Tap on the 50mb file;
3. Tap on the "Download" button from the pop-up.

### Expected behavior
The file starts to download without any problems.

### Actual behavior
The app crash when downloading 50+mb files.

### Device information

* Android device: Google Pixel 3 (Android Q)
* Reference Browser version: 1.0.2012(build #20761207)
* Not reproducible using: Google Pixel 4 (Android Q), OnePlus 5T (Android 9) and LG g7 fit (Android 8.1).
[crash download.txt](https://github.com/mozilla-mobile/reference-browser/files/4353210/crash.download.txt)

csadilek:
OK, this hole was deep :). The cause of this crash is that we were unconditionally inserting into the content resolver which may already have a row / record of the download URI: https://github.com/mozilla-mobile/android-components/issues/6317#issuecomment-601119812

This can happen if a download fails or gets cancelled before we write the file. We will have a unique file name generated based on existing files, but also need to check if we have a record of the file in the resolver. If so, use it, otherwise create a new record.

I've tried for a few hours to write a meaningful test for this, but there are simply too many static methods involved here and the resulting refactoring was terrible: `ContentUris.withAppendedId`, `MediaStore.setIncludePending`, `MediaStore.Downloads.getContentUri` etc. That's properly the reason we don't have an existing test for this method :(

This also makes sure we now won't crash if for some reason there's another problem inserting into the content resolver, but fail the download instead.

### DIFF
--- a/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
+++ b/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
@@ -10,11 +10,13 @@ import android.app.DownloadManager.EXTRA_DOWNLOAD_ID
 import android.app.Service
 import android.content.ActivityNotFoundException
 import android.content.BroadcastReceiver
+import android.content.ContentUris
 import android.content.ContentValues
 import android.content.Context
 import android.content.Intent
 import android.content.Intent.ACTION_VIEW
 import android.content.IntentFilter
+import android.net.Uri
 import android.os.Build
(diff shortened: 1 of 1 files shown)


csadilek:
bors r=Amejia481

PR #6397, opened 2020-03-26
Closes #6171: Simplify push error handling; reduce non-fatal reporting

### LINKED ISSUE #6171: Add better error vs info logging for non-fatal push exceptions
Related to https://github.com/mozilla-mobile/android-components/issues/5691, we don't need to log exceptions that are device network connectivity issues to our CrashReporter.

We need a better way to filter the valuable ones from the non-fatal.

jonalmeida:
Based on the our crash reporting there are only a hand few of exceptions
that are non-fatal. For the rest, we can simply notify the crash
reporter.

 - Simplified state requirements for native invocation by replacing the
 `DeliveryManager.with` to an extension `PushConnection.ifInitialized`.
 - The new `CoroutineExceptionHandler` now notifies the crash reporter
 of fatal or unknown exceptions.
 - Moved extension functions to their equivalent files in an `.ext`
 package.
 - Added more tests to increase coverage. 🎉

---

Closes #6171

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
+++ b/components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
@@ -7,35 +7,29 @@ package mozilla.components.feature.push
 import android.content.Context
 import android.content.SharedPreferences
 import androidx.annotation.VisibleForTesting
+import kotlinx.coroutines.CoroutineExceptionHandler
 import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Job
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.asCoroutineDispatcher
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.plus
-import mozilla.appservices.push.AlreadyRegisteredError
 import mozilla.appservices.push.CommunicationError
 import mozilla.appservices.push.CommunicationServerError
 import mozilla.appservices.push.CryptoError
 import mozilla.appservices.push.GeneralError
-import mozilla.appservices.push.MissingRegistrationTokenError
-import mozilla.appservices.push.RecordNotFoundError
-import mozilla.appservices.push.StorageError
-import mozilla.appservices.push.StorageSqlError
-import mozilla.appservices.push.TranscodingError
-import mozilla.appservices.push.UrlParseError
 import mozilla.components.concept.push.EncryptedPushMessage
 import mozilla.components.concept.push.PushError
 import mozilla.components.concept.push.PushProcessor
 import mozilla.components.concept.push.PushService
 import mozilla.components.support.base.crash.CrashReporting
+import mozilla.components.feature.push.ext.launchAndTry
+import mozilla.components.feature.push.ext.ifInitialized
 import mozilla.components.support.base.log.logger.Logger
 import mozilla.components.support.base.observer.Observable
 import mozilla.components.support.base.observer.ObserverRegistry
 import java.io.File
 import java.util.concurrent.Executors
 import kotlin.coroutines.CoroutineContext
-import mozilla.appservices.push.PushError as RustPushError
 
 /**
  * A implementation of a [PushProcessor] that should live as a singleton by being installed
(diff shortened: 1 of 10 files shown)


jonalmeida:
cc: @grigoryk

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
```diff
@@ -340,53 +323,18 @@ class AutoPushFeature(
...
-internal object DeliveryManager {
-    /**
-     * Executes the block if the Push Manager is initialized.
-     */
-    fun with(connection: PushConnection, block: PushConnection.() -> Unit) {
-        if (connection.isInitialized()) {
-            block(connection)
-        }
+    if (isFatal) {
+        onError(PushError.Rust(e, e.message.orEmpty()))
```
grigoryk, line 337: maybe at least log the non-fatal ones?

components/feature/push/src/main/java/mozilla/components/feature/push/ext/Connection.kt
```diff
@@ -0,0 +1,19 @@
...
+import mozilla.components.support.base.log.logger.Logger
+
+/**
+ * Executes the block if the Push Manager is initialized.
+ */
+internal inline fun PushConnection.ifInitialized(block: PushConnection.() -> Unit) {
+    if (isInitialized()) {
+        block()
+    } else {
+        Logger.error("Native push layer is not yet initialized.")
```
grigoryk, line 17: do we never need to hard-assert this condition and crash out?

components/feature/push/src/main/java/mozilla/components/feature/push/ext/CoroutineScope.kt
```diff
@@ -0,0 +1,29 @@
...
+
+package mozilla.components.feature.push.ext
+
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.launch
+import mozilla.appservices.push.PushError
+
+/**
+ * Catches all known non-fatal push errors to notify the callback before re-throwing.
```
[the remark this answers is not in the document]
jonalmeida: Yeah, that's bad documentation. It's a known _fatal_ push error.

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
jonalmeida: This is the extension function, at the calling site, we use the logger there.

components/feature/push/src/main/java/mozilla/components/feature/push/ext/Connection.kt
jonalmeida: We spent some time thinking about this: push fatal errors are limited to those are not in this list: https://github.com/mozilla-mobile/android-components/pull/6397/files#diff-c2cb12518c4508eaf1fb172996e64fd5R328

The fatal errors we can receive are malformed messages or unrecoverable database errors. Fatal for push doesn't mean we should crash out for the user because we have no real recovery process aside using `verifyConnection` on the AutoPush side and `renewRegistration` for resetting our push tokens.

So for now, the best we can do is report those fatal errors to Sentry and fix them in future releases. App services, could always handle or expose a way to reset the internal db, but that's out of the scope right now until we know what are the situations we want to handle for that solution.

jonalmeida: This was our way of avoiding any timing issues where a message or token could be received from the push service _before_ our rust code had initialized. This shouldn't happen in practice since we don't allow receiving messages until everything is initialized first - it ends up being a way to satisfy the compiler checks that what could happen in practice.

If we lose:
 - A message, this would a lost push message that we cannot recover. In theory we could create a buffer to process these messages layer but in the worse case this would be only the first message loss (see above why this wouldn't happen).
- A push token, we still store the token in our prefs and re-init our native layer with it on the next initialization.

jonalmeida:
bors r=grigoryk

jonalmeida:
bors retry

Issue #6399, opened 2020-03-26
Expose a method to find existing session by its URL


mcarare:
Fenix needs a method to find an existing session by its URL. This is needed for Fenix issue: https://github.com/mozilla-mobile/fenix/issues/5201 & https://github.com/mozilla-mobile/fenix/issues/9421

PR #6401, opened 2020-03-26
For #6399: Add method to find session by url

mcarare:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/selector/Selectors.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/selector/Selectors.kt
@@ -79,6 +79,16 @@ fun BrowserState.findTabOrCustomTabOrSelectedTab(tabId: String? = null): Session
 }
 
 /**
+ * Finds and returns the tab with the given url. Returns null if no matching tab could be found.
+ *
+ * @param url A mandatory url of the searched tab.
+ * @return The tab with the provided url
+ */
+fun BrowserState.findTabByUrl(url: String): TabSessionState? {
+    return tabs.firstOrNull { tab -> tab.content.url == url }
+}
+
+/**
  * List of private tabs.
  */
 val BrowserState.privateTabs: List<TabSessionState>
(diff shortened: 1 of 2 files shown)


components/browser/session/src/main/java/mozilla/components/browser/session/LegacySessionManager.kt
```diff
@@ -516,6 +516,14 @@ class LegacySessionManager(
         values.find { session -> session.id == id }
     }
 
+    /**
+     * Finds and returns the first session with the given url. Returns null if no matching session could be
+     * found.
+     */
+    fun findSessionByUrl(url: String): Session? = synchronized(values) {
+        values.firstOrNull { session -> session.url == url }
+    }
```
pocmo: Would it be possible to implement this as an extension function in `browser-state` on `BrowserState` instead? If possible I would like to avoid adding anything new to `SessionManager` anymore since we are trying to get rid of it.

Even here I think you would want to maybe limit it to `sessions` instead of `values` to not search in custom tabs and web apps?

mcarare: One question: Presumably I get an  `existingTab`:`TabSessionState`. How can I use that to navigate to the opened tab? 
Without using something like :
`sessionManager.findSessionById(existingTab.id)?.let {
                sessionManager.select(it)
}`

pocmo: Yeah, we'd still need this workaround "translation" code. That will be something we can get rid of once we start migrating Fenix code from `SessionManager` to `BrowserStore`. But at least we do not add new functionality that will end up getting called in various places makign things harder to migrate. :)

mcarare: I agree, I just wanted to see if there is something already migrated so I could avoid using `sessionManager` in yet another place in Fenix. Thank you for the details!

PR #6411, opened 2020-03-26
For #3182 - canScrollVerticallyUp considers websites handling motion events.

Mugurell:
Beside the check for the website being scrolled to top we also need to check if
it wants the consume the swipe down motion event, case in which
`canScrollVerticallyUp` will always return false.

The result of this method can be used for the `pull down to refresh`
functionality.

---

### Pull Request checklist

- [x] **Tests**: This PR adds very little code. No tests.

Video showcasing the result - https://drive.google.com/file/d/1USVgp8IirznWWJoYPi6XPETdROYYxzTQ/view

### After merge

### DIFF
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineView.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineView.kt
@@ -7,6 +7,7 @@ package mozilla.components.browser.engine.gecko
 import android.content.Context
 import android.graphics.Bitmap
 import android.util.AttributeSet
+import android.view.MotionEvent
 import android.widget.FrameLayout
 import androidx.annotation.VisibleForTesting
 import androidx.core.view.ViewCompat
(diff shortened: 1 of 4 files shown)


jonalmeida:
I was keen to see this work so I tried it out in r-b and it looks great. :)

There might be more to handling just `INPUT_RESULT_HANDLED_CONTENT` though. I've seen some gestures like long presses cause the screen to not respond to touch events until the page is refreshed. Twitter is where I can easily show it happening, but I was able to reproduce this on Google Maps as well inconsistently with some squiggle gestures (maybe those are interpreted as long press events as well?)

device-2020-03-26-234247 2020-03-27 00_08_13

Mugurell:
@jonalmeida I don't think the patch and feature as a whole could cause this issue.
We are not modifying any behavior but just ask GV if the website is scrolled to top and doesn't consume touch events.

I cannot reproduce the issue you described on twitter but I see it easily reproduced on Google Maps.
Since this issue (on google.com/maps) can be reproduced on Firefox (Fenix) Nightly but not on Firefox Preview or on latest geckoview-example I think this is a recent regression already existing in our Nightly.
I'll look into what might cause this.

Mugurell:
After more tests I think I've managed to reproduce the issue with Google Maps - https://drive.google.com/file/d/1QhNqHpZNTWDdyzppjSrwlkmcju_DRBap/view 
Based on this video and jonalmeida's gif I think the real issue lies in GV
- in my video it's seen seen that when the pin is dropped and the maps view is updated the whole page stops responding to touches
- in jonalmeida's gif it's seen that both the first and second selected trending remain with their background darkened. I think it should normally get whitened again when we leave their bounds.

This tells me that GV gets stucked. Maybe a race condition, maybe a weird infinite loop...?

Mugurell:
Dropping this in favor of https://github.com/mozilla-mobile/android-components/pull/6457
Sorry for the confusion.

PR #6412, opened 2020-03-26
Closes #6367: Do not intercept app initiated load requests by default

### LINKED ISSUE #6367: engine-gecko: Differentiate between initiated and other load requests
This is a performance optimization based on the discussion in https://github.com/mozilla-mobile/fenix/issues/8782.

With this check (either as described here https://github.com/mozilla-mobile/fenix/issues/8782#issuecomment-603411945 or using a new parameter provided by GV) we can decide if we need to run through our interceptors or just notify observers. That latter is fast, the former more expensive (even after recent optimizations). The goal is to let GV continue loading the URL as soon as possible.

csadilek:
Our interceptors in Fenix and R-B do not need to be invoked for load requests initiated by the application (e.g. the user loading a URL via the toolbar). With this, the interceptors can control whether or not they want to be invoked in that case.

This is a performance optimization, as we want to minimize the time we block GeckoView from initiating the page load (it has to wait for the result of `onLoadRequest`.)

Full discussion in: https://github.com/mozilla-mobile/fenix/issues/8782#issuecomment-603411945

### DIFF
--- a/buildSrc/src/main/java/Gecko.kt
+++ b/buildSrc/src/main/java/Gecko.kt
@@ -6,7 +6,7 @@ internal object GeckoVersions {
     /**
      * GeckoView Nightly Version.
      */
-    const val nightly_version = "76.0.20200327094805"
+    const val nightly_version = "76.0.20200330094747"
 
     /**
      * GeckoView Beta Version.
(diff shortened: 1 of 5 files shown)


rocketsroger:
bors r=rocketsroger

csadilek:
Marking as "do not land" until we have confirmation from GV that this is the right approach.

csadilek:
OK, looks like we're getting a new parameter / flag so we don't have to use `triggerUri` which isn't a reliable way to check.

rocketsroger:
bors r=rocketsroger

rocketsroger:
bors r=rocketsroger

PR #6432, opened 2020-03-27
Do not pause or display media notification when all media is muted.

everett1992:
Firefox preview will display a notification and request audio focus even
when the playing media is a muted gif, webm or other 'lite' media.

My goal is to stop requesting audio focus when the media is not playing
audio.
 - I only considered muted media. I ignored media with volume == 0.0,
 and media with audioTrackCount == 0. I'm not _sure_ that
 audioTrackCount 0 means the media is not playing sound. A similar
 commit could track and check those values.
 - I added the check for muted audio in MediaAggregateUpdaterTest
 because I am nnoyed by the notifications firefox creates if I leave a
 page with a webm in an open tab. I think aggregating player state to
 NONE will disable the notification. People won't be able to play/pause
 muted media from a notification, but I don't expect that's a common
 usecase. The check can be moved to MediaStateDelegate to avoid chaning
 the notification behavior.

Fixes [mozilla-mobile/fenix # 6146](https://github.com/mozilla-mobile/fenix/issues/6146)

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
@@ -459,6 +459,16 @@ sealed class MediaAction : BrowserAction() {
     ) : MediaAction()
 
     /**
+     * Updates the [Media.Volume] for the [MediaState.Element] with id [mediaId] owned by the tab
+     * with id [tabId].
+     */
+    data class UpdateMediaVolumeAction(
+        val tabId: String,
+        val mediaId: String,
+        val volume: Media.Volume
+    ) : MediaAction()
+
+    /**
      * Updates [MediaState.Aggregate] in the [MediaState].
      */
     data class UpdateMediaAggregateAction(
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/media/GeckoMedia.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/media/GeckoMedia.kt
@@ -17,6 +17,7 @@ import org.mozilla.geckoview.MediaElement.MEDIA_STATE_SEEKING
 import org.mozilla.geckoview.MediaElement.MEDIA_STATE_STALLED
 import org.mozilla.geckoview.MediaElement.MEDIA_STATE_SUSPEND
 import org.mozilla.geckoview.MediaElement.MEDIA_STATE_WAITING
+import kotlin.properties.Delegates
 
 /**
  * [Media] (`concept-engine`) implementation for GeckoView.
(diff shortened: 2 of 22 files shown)


everett1992:
Do you want the checks to be all green?
lint-detekt fails on undocumented public method in an interface where all methods are callbacks and none had public documentation before my change. It fails for 'too many methods', I'm don't want to jump into a refactor to reduce the methods.

complete-pr and complete-pr-2 are failing for what look like instrumentation issues in the python harness.

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -448,6 +448,16 @@ sealed class MediaAction : BrowserAction() {
...
+    /**
+     * Updates the [Media.Volume] for the [MediaState.Element] with id [mediaId] owned by the tab
+     * with id [tabId].
+     */
+    data class UpdateMediaVolumeAction(
+        val tabId: String,
+        val mediaId: String,
+        val volume: Media.Volume
+    ) : MediaAction()
+
```
pocmo, line 460: Please add a test for this to `MediaActionTest`.

pocmo:
> Do you want the checks to be all green?

Yes. Of course it is always a trade-off and if it is absolutely not reasonable to change/refactor something at this point then we use `@Suppress` to explicitly allow this case.

everett1992:
> > Do you want the checks to be all green?
> 
> Yes. Of course it is always a trade-off and if it is absolutely not reasonable to change/refactor something at this point then we use `@Suppress` to explicitly allow this case.

What about the complete-pr* tasks? I don't understand that condition that's failing
```
[task 2020-03-29T16:42:36.556Z] + /builds/worker/checkouts/src/taskcluster/scripts/are_dependencies_completed.py J6HfwGEySAqrq72n8aoaHg TpEPK7iKQPuF2kFHuT2Yfg
[task 2020-03-29T16:42:36.765Z] Fetching status of 2 dependencies...
[task 2020-03-29T16:42:36.765Z] Statuses fetched.
[task 2020-03-29T16:42:36.765Z] Traceback (most recent call last):
[task 2020-03-29T16:42:36.765Z]   File "/builds/worker/checkouts/src/taskcluster/scripts/are_dependencies_completed.py", line 52, in <module>
[task 2020-03-29T16:42:36.765Z]     main()
[task 2020-03-29T16:42:36.765Z]   File "/builds/worker/checkouts/src/taskcluster/scripts/are_dependencies_completed.py", line 46, in main
[task 2020-03-29T16:42:36.765Z]     check_all_dependencies_are_completed(result.dependencies_task_ids)
[task 2020-03-29T16:42:36.765Z]   File "/builds/worker/checkouts/src/taskcluster/scripts/are_dependencies_completed.py", line 32, in check_all_dependencies_are_completed
[task 2020-03-29T16:42:36.765Z]     raise ValueError('Some tasks are not completed: {}'.format(non_completed_tasks))
```

everett1992:
I'm going to take another swing at this, I'll add a `muted` value to MediaState.Aggregate. IMO MediaStateAggregator and aggregateNewState should be refactored, each property in MediaState.Aggregate should be reduced on it's own. Any advice?

pocmo:
> What about the complete-pr* tasks? I don't understand that condition that's failing

They are kind of "gatekeepers". They fail if a previous other checks failed. So you can ignore them. If the other tasks are green then they will be green too. :)

pocmo:
I think your approach is a good compromise for now. Not showing the notification and therefore not requesting audio focus for muted content makes sense to me. It also seems like that is what Chrome does. And yeah, I agree that this is not the right tool for making the user aware of background data usage - we need something else for that. So I am in favor of landing this implementation and adapt further in the future if needed.

I tested your patch today and was surprised to see it not work. But then I noticed that a change of `GeckoMedia.volume` does not trigger `onVolumeChange()`. The same is true for the already existing `GeckoMedia.metadata`. Seems like this is something we missed. After adding this everything works nicely. However so far I only tested with Twitter. Will test on some more sites and see if there are any downsides.

everett1992:
Thanks for testing, I don't have a Android emulator set up. I based my
change on the metadata workflow so if it was broken, makes sense volume was
broken too.

PR #6475, opened 2020-03-31
Closes #6314: Support fetching / downloading data URIs

### LINKED ISSUE #6314: Unable to download data: urls
We currently can't consume downloads from data urls e.g.

```html
<a href="data:text/plain;base64,SGVsbG8sIFdvcmxkIQ==" type="text/plain" download="test.txt">Test</a>
```

This is also the cause of the uBlock "Back up to file" functionality not working:
https://github.com/gorhill/uBlock/blob/08d370d32e648c2823e1bc31fcd9f7e1f96d155a/platform/chromium/vapi-common.js#L172

This is similar to: https://github.com/mozilla-mobile/android-components/issues/1159

```
 java.lang.IllegalArgumentException: URI scheme must be http or https
        at org.mozilla.geckoview.GeckoWebExecutor.fetch(GeckoWebExecutor.java:128)
        at mozilla.components.browser.engine.gecko.fetch.GeckoViewFetchClient.fetch(GeckoViewFetchClient.kt:57)
        at mozilla.components.feature.downloads.AbstractFetchDownloadService.performDownload$feature_downloads_debug(AbstractFetchDownloadService.kt:326)
        at mozilla.components.feature.downloads.AbstractFetchDownloadService.startDownloadJob$feature_downloads_debug(AbstractFetchDownloadService.kt:292)
        at mozilla.components.feature.downloads.AbstractFetchDownloadService$onStartCommand$1.invokeSuspend(AbstractFetchDownloadService.kt:205)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:56)
        at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:561)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:727)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:667)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:655)
```

csadilek:
With this we can fetch from data URIs and downloads "just work" (we don't require any special handling). So, e.g. this works now (which is also used by uBlock to save backups):

```html
<a href="data:text/plain;base64,SGVsbG8sIFdvcmxkIQ==" download="backup.txt">Backup</a>
```
https://github.com/gorhill/uBlock/blob/08d370d32e648c2823e1bc31fcd9f7e1f96d155a/platform/chromium/vapi-common.js#L172

GV didn't want to add API for it so I took a stab at adding this to our fetch libs: https://github.com/mozilla-mobile/android-components/issues/6314#issuecomment-606278886

@pocmo can you take a look since we looked at download stuff today :). I've time-boxed this and limited to base64 encoded URIs. The device and unit tests pass and the download works / can be tested here:

https://jsfiddle.net/nue9rg0L/2/

### DIFF
--- a/components/concept/fetch/src/main/java/mozilla/components/concept/fetch/Client.kt
+++ b/components/concept/fetch/src/main/java/mozilla/components/concept/fetch/Client.kt
@@ -4,6 +4,10 @@
 
 package mozilla.components.concept.fetch
 
+import android.util.Base64
+import mozilla.components.concept.fetch.Response.Companion.CONTENT_LENGTH_HEADER
+import mozilla.components.concept.fetch.Response.Companion.CONTENT_TYPE_HEADER
+import java.io.ByteArrayInputStream
 import java.io.IOException
 
 /**
@@ -41,6 +45,44 @@ abstract class Client {
     abstract fun fetch(request: Request): Response
 
     /**
+     * Generates a [Response] by decoding a base64 encoded data URI.
+     *
+     * @param request The [Request] for the data URI.
+     * @return The generated [Response] including the decoded bytes as body.
+     */
+    @Suppress("ThrowsCount", "TooGenericExceptionCaught")
+    protected fun fetchDataUri(request: Request): Response {
+        if (!request.isDataUri()) {
+            throw IOException("Not a data URI")
+        }
+
+        val dataUri = request.url
+        if (!dataUri.contains(DATA_URI_BASE64_EXT)) {
+            throw IOException("Data URI must be base64 encoded")
+        }
+
+        return try {
+            val contentType = dataUri.substringAfter(DATA_URI_SCHEME).substringBefore(DATA_URI_BASE64_EXT)
+            val bytes = Base64.decode(dataUri.substring(dataUri.lastIndexOf(',') + 1), Base64.DEFAULT)
+            val headers = MutableHeaders().apply {
+                set(CONTENT_LENGTH_HEADER, bytes.size.toString())
+                if (contentType.isNotEmpty()) {
+                    set(CONTENT_TYPE_HEADER, contentType)
+                }
+            }
+
+            Response(
+                dataUri,
+                Response.SUCCESS,
+                headers,
+                Response.Body(ByteArrayInputStream(bytes), contentType)
+            )
+        } catch (e: Exception) {
+            throw IOException("Failed to decode data URI")
+        }
+    }
+
+    /**
      * List of default headers that should be added to every request unless overridden by the headers in the request.
      */
     protected val defaultHeaders: Headers = MutableHeaders(
@@ -61,4 +103,9 @@ abstract class Client {
         // We expect all clients to support and use keep-alive by default.
         "Connection" to "keep-alive"
     )
+
+    companion object {
+        const val DATA_URI_BASE64_EXT = ";base64"
+        const val DATA_URI_SCHEME = "data:"
+    }
 }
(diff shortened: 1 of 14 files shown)


components/concept/fetch/src/main/java/mozilla/components/concept/fetch/Client.kt
```diff
@@ -62,3 +66,31 @@ abstract class Client {
...
 }
+
+/**
+ * "Fetches" a base64 encoded data URI.
+ *
+ * @param request The [Request] for the data URI.
+ * @return The generated [Response] including the decoded bytes as body.
+ */
+@Suppress("TooGenericExceptionCaught")
+fun fetchDataUri(request: Request): Response {
```
pocmo, line 77: Should we put this into `Client` and make it `protected` (like `defaultHeaders`) so that it is not used outside of a `Client` implementation? This could now end up getting called by app code.

components/lib/fetch-httpurlconnection/src/test/java/mozilla/components/lib/fetch/httpurlconnection/HttpUrlConnectionFetchTestCases.kt
```diff
@@ -34,4 +38,18 @@ class HttpUrlConnectionFetchTestCases : mozilla.components.tooling.fetch.tests.F
         connection.setupWith((Request("https://mozilla.org", useCaches = false)))
         assertFalse(connection.useCaches)
     }
+
+    @Test
+    override fun getDataUri() {
+        // We can't run this test case because Base64 encoding is not available but we can verify
+        // that we attempt to fetch the data URI.
```
pocmo, line 45: If we run this with `AndroidJUnit4` then we should be able to use the Base64 classes, I think?

components/concept/fetch/src/main/java/mozilla/components/concept/fetch/Client.kt
```diff
@@ -62,3 +66,31 @@ abstract class Client {
...
+    return try {
+        val dataUri = request.url
+        val contentType = dataUri.substringAfter("data:").substringBefore(";base64")
+
+        val bytes = Base64.decode(dataUri.substring(dataUri.lastIndexOf(',') + 1), Base64.DEFAULT)
+        val headers = MutableHeaders(
+            CONTENT_TYPE_HEADER to contentType,
+            CONTENT_LENGTH_HEADER to bytes.size.toString()
+        )
```
pocmo, line 86: Looking at [RFC 2397](https://tools.ietf.org/html/rfc2397) and [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs) I wonder if we should be a bit more protective regarding optional elements:
* If the data uri doesn't specify base64 encoding then bail out early?
* If the content type is not specified then do not add header for it?

components/concept/fetch/src/main/java/mozilla/components/concept/fetch/Client.kt
csadilek: OK, makes sense. Updated!

components/lib/fetch-httpurlconnection/src/test/java/mozilla/components/lib/fetch/httpurlconnection/HttpUrlConnectionFetchTestCases.kt
csadilek: Ah yes, good catch! I've upgraded the lib and removed the test override as the base test is running fine now!

components/concept/fetch/src/main/java/mozilla/components/concept/fetch/Client.kt
csadilek: OK, added another test case and made this more robust too.

csadilek:
bors r=Amejia481,pocmo

csadilek:
bors retry

Issue #6499, opened 2020-04-02
Support dynamic width for the BrowserMenu


Mugurell:
As seen in ticket https://github.com/mozilla-mobile/fenix/issues/7157 there might be some menu options wider than we currently support.
To allow a little more width for them our BrowserMenu should allow for a dynamic width instead of always using the fixed `250dp` we currently use.

Mugurell:
Closing this since it already landed.

PR #6531, opened 2020-04-06
For #6499 - Add min/max width support for BrowserMenu

Mugurell:
BrowserMenu can now have a dynamic width allowing for a better fit for all menu items.:
- Using the same value for `mozac_browser_menu_width_min` and `mozac_browser_menu_width_max` means BrowserMenu will have a fixed width - `mozac_browser_menu_width`.
_This is the default behavior_.
- Different values for `mozac_browser_menu_width_min` and `mozac_browser_menu_width_max` means BrowserMenu will have a dynamic width depending on the widest BrowserMenuItem and between the aforementioned dimensions also taking into account display width.

[Video showing this being used by Fenix with a  `mozac_browser_menu_width_max` of 600dp.](https://drive.google.com/file/d/1MO8s3W0MN8v9Y2M0h3WyFFhgKEQal8gg/view?usp=sharing)

---

### Pull Request checklist

- [x] **Tests**: No tests. Small layout change.

Video showing how Fenix will 

### After merge

### DIFF
--- a/components/browser/menu/README.md
+++ b/components/browser/menu/README.md
@@ -29,11 +29,20 @@ There are multiple properties that you customize of the menu browser by just add
     <!--Change the width of the menu-->
     <dimen name="mozac_browser_menu_width" tools:ignore="UnusedResources">250dp</dimen>
 
+    <!--Change the dynamic width of the menu-->
+    <dimen name="mozac_browser_menu_width_min" tools:ignore="UnusedResources">200dp</dimen>
+    <dimen name="mozac_browser_menu_width_max" tools:ignore="UnusedResources">300dp</dimen>
+
     <!--Change the top and bottom padding of the menu-->
     <dimen name="mozac_browser_menu_padding_vertical" tools:ignore="UnusedResources">8dp</dimen>
 
 </resources>
 ```
+BrowserMenu can have a dynamic width:
+- Using the same value for `mozac_browser_menu_width_min` and `mozac_browser_menu_width_max` means BrowserMenu will have a fixed width - `mozac_browser_menu_width`.
+_This is the default behavior_.
+- Different values for `mozac_browser_menu_width_min` and `mozac_browser_menu_width_max` means BrowserMenu will have a dynamic width depending on the widest BrowserMenuItem and between the aforementioned dimensions also taking into account display width.
+
 
 ### BrowserMenuDivider
 ```kotlin
--- a/components/browser/menu/src/main/java/mozilla/components/browser/menu/view/DynamicWidthRecyclerView.kt
+++ b/components/browser/menu/src/main/java/mozilla/components/browser/menu/view/DynamicWidthRecyclerView.kt
@@ -1,0 +1,76 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.menu.view
+
+import android.annotation.SuppressLint
+import android.content.Context
+import android.util.AttributeSet
+import androidx.annotation.VisibleForTesting
+import androidx.recyclerview.widget.RecyclerView
+import mozilla.components.browser.menu.R
+
+/**
+ * [RecylerView] with automatically set width between widthMin / widthMax xml attributes.
+ */
+class DynamicWidthRecyclerView @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet? = null
+) : RecyclerView(context, attrs) {
+    @VisibleForTesting internal var minWidth: Int = -1
+    @VisibleForTesting internal var maxWidth: Int = -1
+
+    init {
+        context.obtainStyledAttributes(attrs, R.styleable.DynamicWidthRecyclerView).apply {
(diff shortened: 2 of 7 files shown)


components/browser/menu/README.md
```diff
@@ -26,8 +26,9 @@ There are multiple properties that you customize of the menu browser by just add
     <!--Change how much shadow the menu should have-->
     <dimen name="mozac_browser_menu_elevation" tools:ignore="UnusedResources">4dp</dimen>
 
-    <!--Change the width of the menu-->
-    <dimen name="mozac_browser_menu_width" tools:ignore="UnusedResources">250dp</dimen>
```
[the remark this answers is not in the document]
Mugurell: I was thinking that all clients would find this as a cool feature and all would like it :-))
And they still could set the same value for both properties so they could get the old effect.
But I agree, would be better to not break the old functionality.

Mugurell:
> @Mugurell This is in reference of #3703. (Can you rename the commit for that issue?)

@psymoon Thank you for referencing #3703.
For this patch I only went by the acceptance criteria from https://github.com/mozilla-mobile/fenix/issues/7157 :
"Expand the menu width up to a certain max_width we determine (314dp) & truncate items that don't fit."

I don't think my patch is the solution for #3703
As I understand that ticket it needed the menu to be as small as the widest child, with no value imposed for a minimum width, so it could be even 100 px for maybe a simple a/b/c menu.
Or allow for setting a minimum width for each individual menu where desired so.

My patch will hardcode a min/max width for all menus and allow them to vary their width inside that constraints based on the widest child.
But because the min/max width values would apply to all menus we needed some .. reasonable defaults so extremes like 100px which the patch for #3703 needs would not be possible.
And it would definitely not be possible to have different min or min/max widths imposed at runtime for different menus.

I propose finishing this patch based on your review as it was intended for #6499 and https://github.com/mozilla-mobile/fenix/issues/7157 and then work on #3703 and https://github.com/mozilla-mobile/fenix/issues/1871 to bring support for runtime passed width values.

psymoon:
@Mugurell that sounds good. Do you think you can add tests and address the breaking change bit?

Mugurell:
Closed/reopened PR to restart failing CI jobs.

components/browser/menu/src/main/java/mozilla/components/browser/menu/view/DynamicWidthRecyclerView.kt
```diff
@@ -0,0 +1,54 @@
...
+
+import android.content.Context
+import android.util.AttributeSet
+import androidx.recyclerview.widget.RecyclerView
+import mozilla.components.browser.menu.R
+
+/**
+ * [RecylerView] with automatically set width between widthMin / widthMax xml attributes.
+ */
+class DynamicWidthRecyclerView @JvmOverloads constructor(
```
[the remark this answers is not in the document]
Mugurell: Added. Thanks!

Mugurell:
Rebased the PR to include this patch in the changelog of v. 40 and also apply the recommendation from here - https://proandroiddev.com/misconception-about-kotlin-jvmoverloads-for-android-view-creation-cb88f432e1fe

components/browser/menu/src/test/java/mozilla/components/browser/menu/view/DynamicWidthRecyclerViewTest.kt
```diff
@@ -0,0 +1,121 @@
...
+import org.junit.runner.RunWith
+import org.mockito.ArgumentMatchers.anyInt
+import org.mockito.Mockito.never
+import org.mockito.Mockito.spy
+import org.mockito.Mockito.verify
+import org.robolectric.Robolectric
+import org.robolectric.annotation.Config
+
+@RunWith(AndroidJUnit4::class)
+class DynamicWidthRecyclerViewTest {
```
psymoon, line 22: Can we add a test here that this would work with `mozac_browser_menu_width` as well?

Mugurell: Sorry but I'm not really sure what to test in this regards.
In the tests from line 33 and 45 I am testing the cases in which we will not try to calculate and set a "dynamic width". In which cases I verify that just the default behavior will be executed, onMeasure will use it's parameters - layout_width and layout_height. I added a small comment in this regard in DynamicWidthRecyclerView.
If you have other scenarios in mind I could use some pointers.

psymoon: @Mugurell I meant more specifically a case where you don't define `minWidth` and `maxWidth`? :)

psymoon: @Mugurell I meant more specifically a case where you don't define minWidth and maxWidth? :)

Mugurell: Thanks!
I added 3 new tests:
- If minWidth and maxWidth are not provided view should use layout_width
- If only minWidth is provided view should use layout_width
- If only maxWidth is provided view should use layout_width

psymoon:
bors r=psymoon

PR #6541, opened 2020-04-06
 GeckoView update (nightly) (20200406-140518)

MickeyMoz:


psymoon:
I don't see any breaking changes in GV log and it appears, `Error transferring file: Server returned HTTP response code: 504 for URL: https://oss.sonatype.org/content/groups/public/org/robolectric/android-all/8.1.0-robolectric-4611349/android-all-8.1.0-robolectric-4611349.pom` due to intermittently fails to download dependencies, is the reason for CI failures.

PR #6544, opened 2020-04-06
Close #5795: Support viewport-fit descriptor via onMetaViewportFitChange.

### LINKED ISSUE #5795: Support viewport-fit descriptor via `onMetaViewportFitChange`
Spec: https://drafts.csswg.org/css-round-display/#viewport-fit-descriptor

I have added `<meta name="viewport" content="viewport-fit=cover">` support to GV by https://bugzilla.mozilla.org/show_bug.cgi?id=1574307. ac should override `onMetaViewportFitChange` of `ContentDelegate`, then ac should set `layoutInDisplayCutoutMode` by `viewport-fit value`.

GVE implementation for example is https://bugzilla.mozilla.org/show_bug.cgi?id=1612785

rocketsroger:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -5,6 +5,8 @@
 package mozilla.components.browser.engine.gecko
 
 import android.annotation.SuppressLint
+import android.os.Build
+import android.view.WindowManager
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
@@ -679,6 +681,18 @@ class GeckoEngineSession(
             val parsed = WebAppManifestParser().parse(manifest)
             if (parsed is WebAppManifestParser.Result.Success) {
                 notifyObservers { onWebAppManifestLoaded(parsed.manifest) }
+            }
+        }
+
+        override fun onMetaViewportFitChange(session: GeckoSession, viewportFit: String) {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+                val layoutInDisplayCutoutMode = when (viewportFit) {
+                    "cover" -> WindowManager.LayoutParams.LAYOUT_IN_DISPLAY_CUTOUT_MODE_SHORT_EDGES
+                    "contain" -> WindowManager.LayoutParams.LAYOUT_IN_DISPLAY_CUTOUT_MODE_NEVER
+                    else -> WindowManager.LayoutParams.LAYOUT_IN_DISPLAY_CUTOUT_MODE_DEFAULT
+                }
+
+                notifyObservers { onMetaViewportFitChanged(layoutInDisplayCutoutMode) }
             }
         }
     }
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
@@ -17,6 +17,7 @@ import mozilla.components.browser.session.ext.toTabSessionState
 import mozilla.components.browser.state.action.ContentAction.AddFindResultAction
 import mozilla.components.browser.state.action.ContentAction.ClearFindResultsAction
 import mozilla.components.browser.state.action.ContentAction.ConsumeHitResultAction
+import mozilla.components.browser.state.action.ContentAction.FullScreenChangedAction
 import mozilla.components.browser.state.action.ContentAction.RemoveThumbnailAction
 import mozilla.components.browser.state.action.ContentAction.UpdateHitResultAction
 import mozilla.components.browser.state.action.ContentAction.UpdateLoadingStateAction
@@ -26,6 +27,7 @@ import mozilla.components.browser.state.action.ContentAction.UpdateSecurityInfoA
 import mozilla.components.browser.state.action.ContentAction.UpdateThumbnailAction
 import mozilla.components.browser.state.action.ContentAction.UpdateTitleAction
 import mozilla.components.browser.state.action.ContentAction.UpdateUrlAction
+import mozilla.components.browser.state.action.ContentAction.ViewportFitChangedAction
 import mozilla.components.browser.state.action.CustomTabListAction.RemoveCustomTabAction
 import mozilla.components.browser.state.action.EngineAction
 import mozilla.components.browser.state.action.TabListAction.AddTabAction
@@ -98,6 +100,10 @@ class Session(
         fun onFindResult(session: Session, result: FindResult) = Unit
         fun onDesktopModeChanged(session: Session, enabled: Boolean) = Unit
         fun onFullScreenChanged(session: Session, enabled: Boolean) = Unit
+        /**
+         * @param layoutInDisplayCutoutMode value of defined in https://developer.android.com/reference/android/view/WindowManager.LayoutParams#layoutInDisplayCutoutMode
+         */
+        fun onMetaViewportFitChanged(session: Session, layoutInDisplayCutoutMode: Int) = Unit
         fun onThumbnailChanged(session: Session, bitmap: Bitmap?) = Unit
         fun onContentPermissionRequested(session: Session, permissionRequest: PermissionRequest): Boolean = false
         fun onAppPermissionRequested(session: Session, permissionRequest: PermissionRequest): Boolean = false
@@ -406,7 +412,17 @@ class Session(
      * Exits fullscreen mode if it's in that state.
      */
     var fullScreenMode: Boolean by Delegates.observable(false) { _, old, new ->
-        notifyObservers(old, new) { onFullScreenChanged(this@Session, new) }
+        if (notifyObservers(old, new) { onFullScreenChanged(this@Session, new) }) {
+            store?.syncDispatch(FullScreenChangedAction(id, new))
+        }
+    }
+
+    /**
+     * Display cutout mode state.
+     */
+    var layoutInDisplayCutoutMode: Int by Delegates.observable(0) { _, _, new ->
+        notifyObservers { onMetaViewportFitChanged(this@Session, new) }
+        store?.syncDispatch(ViewportFitChangedAction(id, new))
     }
 
     /**
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
@@ -246,6 +246,19 @@ sealed class ContentAction : BrowserAction() {
      * Removes the [SearchRequest] of the [ContentState] with the given [sessionId].
      */
     data class ConsumeSearchRequestAction(val sessionId: String) : ContentAction()
+
+    /**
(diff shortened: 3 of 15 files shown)


components/feature/session/src/main/java/mozilla/components/feature/session/FullScreenFeature.kt
```diff
@@ -17,6 +17,7 @@ open class FullScreenFeature(
     sessionManager: SessionManager,
     private val sessionUseCases: SessionUseCases,
     private val sessionId: String? = null,
+    private val viewportFitChange: (String) -> Unit = {},
     private val fullScreenChanged: (Boolean) -> Unit
```
jonalmeida, line 21:
```suggestion
    private val fullScreenViewPortChanged: (Boolean, String) -> Unit
```

Should we have a single callback from `viewportFitChange` events along with `fullScreenChanged`? I'm not sure if consumers will find it easy to handle those separately since they seem coupled together (e.g. when fullscreen changes, you would want to know the viewportFit, and vice versa).

The internal implementation would call the session value for the other (see below).

components/feature/session/src/main/java/mozilla/components/feature/session/FullScreenFeature.kt
```diff
@@ -28,6 +29,7 @@ open class FullScreenFeature(
     }
 
     override fun onFullScreenChanged(session: Session, enabled: Boolean) = fullScreenChanged(enabled)
+    override fun onMetaViewportFitChanged(session: Session, viewportFit: String) = viewportFitChange(viewportFit)
```
jonalmeida, line 32:
```suggestion
    override fun onFullScreenChanged(session: Session, enabled: Boolean) = fullScreenViewPortChanged(enabled, session.layoutInDisplayCutoutMode)
    override fun onMetaViewportFitChanged(session: Session, viewportFit: String) = (session.fullScreenMode, viewportFit)
```

See above for related change.

components/feature/session/src/main/java/mozilla/components/feature/session/FullScreenFeature.kt
```diff
@@ -17,6 +17,7 @@ open class FullScreenFeature(
     sessionManager: SessionManager,
     private val sessionUseCases: SessionUseCases,
     private val sessionId: String? = null,
+    private val viewportFitChange: (String) -> Unit = {},
```
jonalmeida, line 20: nit: `viewportFitChange` -> `viewportFitChanged`

components/feature/session/src/main/java/mozilla/components/feature/session/FullScreenFeature.kt
rocketsroger: Oh I just changed to from `onMetaViewportFitChange` to `onMetaViewportFitChanged` in the last patch to match the naming above.  Should I have not?

components/feature/session/src/main/java/mozilla/components/feature/session/FullScreenFeature.kt
rocketsroger: The GeckoView API is separate though, if we coupled them then the consumers will get fullScreenChanged called twice?

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -681,6 +681,10 @@ class GeckoEngineSession(
                 notifyObservers { onWebAppManifestLoaded(parsed.manifest) }
             }
         }
+
+        override fun onMetaViewportFitChange(session: GeckoSession, viewportFit: String) {
```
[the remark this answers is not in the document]
rocketsroger: Agreed. I'll make a enum from GeckoView values.

rocketsroger:
> We should also add support for the `viewportFit` to BrowserState since that's the new API we're moving towards.
> 
> I recall @Amejia481 mentioning something similar in sprint planning, so he might have more comment.

Agreed.  But it looks like I will have to move the `FullScreenFeature` to use BrowserState since they are related.  Should I move the feature first, then apply this change on top after?

components/feature/session/src/main/java/mozilla/components/feature/session/FullScreenFeature.kt
rocketsroger: Yeah if they are coupled together, then each call will be called once.  The consumer will then get both the full screen state and viewportFit value twice.

components/feature/session/src/main/java/mozilla/components/feature/session/FullScreenFeature.kt
jonalmeida: That's a good point.

When I think about this code from a consumer's side, it seems like we're giving them multiple state changes to be handled separately, but they would really need them all together either way to make correct decisions.

```kotlin
FullscreenFeature(// ...
  ::handleViewPortChanged,
  ::handleFullscreenChanged)

fun handleViewPortChanged(viewPort: String) {
  if (sessionManager.findSession().fullscreen) {
    if (viewPort == "contain") {
      // do fullscreen with viewport change
    }
  }
}

fun handleFullscreenChanged(fullscreen: Boolean) {
  if (fullscreen) {
    if (sessionManager.findSession().viewPort == "contain") {
      // do fullscreen with viewport change
    }
  } else {
    if (sessionManager.findSession().viewPort == "none") {
      // do non-fullscreen with viewport
    }
  }
}
```

The unknown in my knowledge is how apps want to handle the viewport value with and without fullscreen, so maybe the above code is the outcome regardless?

components/feature/session/src/main/java/mozilla/components/feature/session/FullScreenFeature.kt
jonalmeida: The nit was on feature's constructor. :)

rocketsroger: Ah.  Thanks!

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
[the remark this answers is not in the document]
rocketsroger: Actually I will translate to the `layoutInDisplayCutoutMode` right here so the consumers doesn't have to understand GeckoView string values.

components/feature/session/src/main/java/mozilla/components/feature/session/FullScreenFeature.kt
rocketsroger: I see.  This is my mistake in my explanation.  The consumers doesn't have to need them together to make the correct decision.  These two state changes can be handled separately by the consumers without knowing the other state.

components/feature/session/src/main/java/mozilla/components/feature/session/FullScreenFeature.kt
```diff
@@ -28,6 +30,11 @@ open class FullScreenFeature(
     }
 
     override fun onFullScreenChanged(session: Session, enabled: Boolean) = fullScreenChanged(enabled)
+    override fun onMetaViewportFitChanged(session: Session, layoutInDisplayCutoutMode: Int) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
```
jonalmeida, line 34: Do we need this check here if we already have it in the engine?

components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
```diff
@@ -98,6 +100,7 @@ class Session(
         fun onFindResult(session: Session, result: FindResult) = Unit
         fun onDesktopModeChanged(session: Session, enabled: Boolean) = Unit
         fun onFullScreenChanged(session: Session, enabled: Boolean) = Unit
+        fun onMetaViewportFitChanged(session: Session, layoutInDisplayCutoutMode: Int) = Unit
```
jonalmeida, line 103: Since we're not going the enum route, we should probably document what this int value references and the supported values.

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -246,6 +246,16 @@ sealed class ContentAction : BrowserAction() {
...
+
+    /**
+     * Updates the [fullScreenEnabled] with the given [sessionId].
+     */
+    data class FullScreenChangedAction(val sessionId: String, val fullScreenEnabled: Boolean) : ContentAction()
+
+    /**
+     * Updates the [layoutInDisplayCutoutMode] with the given [sessionId].
+     */
+    data class ViewportFitChangedAction(val sessionId: String, val layoutInDisplayCutoutMode: Int) : ContentAction()
```
jonalmeida, line 258: Same here. Docs for the int value or point to where one place in the codebase to find them.

components/feature/session/src/main/java/mozilla/components/feature/session/FullScreenFeature.kt
rocketsroger: you're right.  Forgot to clean it up.  Thanks,

components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
rocketsroger: Agreed.  I'll like to Android docs

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
rocketsroger: Agreed

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -246,6 +246,19 @@ sealed class ContentAction : BrowserAction() {
...
+
+    /**
+     * Updates the [fullScreenEnabled] with the given [sessionId].
+     */
+    data class FullScreenChangedAction(val sessionId: String, val fullScreenEnabled: Boolean) : ContentAction()
+
+    /**
+     * Updates the [layoutInDisplayCutoutMode] with the given [sessionId].
+     *
+     * layoutInDisplayCutoutMode value is defined in
```
jonalmeida, line 258: nit: `@param` so that the KDocs render correctly. :)

components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
```diff
@@ -98,6 +100,11 @@ class Session(
         fun onFindResult(session: Session, result: FindResult) = Unit
         fun onDesktopModeChanged(session: Session, enabled: Boolean) = Unit
         fun onFullScreenChanged(session: Session, enabled: Boolean) = Unit
+        /**
+         * layoutInDisplayCutoutMode value is defined in
```
jonalmeida, line 104: Same here for `@param`.

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
rocketsroger: Agreed.  Thanks,

rocketsroger:
bors r=jonalmeida

PR #6547, opened 2020-04-06
Closes #4397: Migrate feature-readerview to browser-state

### LINKED ISSUE #4397: Migrate feature-readerview to use browser-state

csadilek:
@pocmo Super happy with how the middleware worked out. It basically maps state to other (action/state), as relevant to reader view e.g. the url has changed -> a new reader check is required.

The other cases handled by the middleware are:
- User opens a tab via the context menu (may or may not switch to it directly)
- A new tab is selected
- A tab is removed (see comment in middleware)

All other functionality remains the same. Diff is big mostly because of test refactorings :)

I found another way that allows us to remove all reader state from the session right away! I am storing the state in the snapshot directly (same as `engineState`) and let the `SessionManager` deal with adding/deleting it based on the state in the store. This way we don't need to pass the store along to the serializer / session storage.

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
@@ -13,8 +13,11 @@ import mozilla.components.browser.state.action.CustomTabListAction
 import mozilla.components.browser.state.action.EngineAction.LinkEngineSessionAction
 import mozilla.components.browser.state.action.EngineAction.UpdateEngineSessionStateAction
 import mozilla.components.browser.state.action.EngineAction.UnlinkEngineSessionAction
+import mozilla.components.browser.state.action.ReaderAction
 import mozilla.components.browser.state.action.SystemAction
 import mozilla.components.browser.state.action.TabListAction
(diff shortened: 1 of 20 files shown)


csadilek:
Fenix PR: https://github.com/mozilla-mobile/fenix/pull/9752

csadilek:
R-B PR: https://github.com/mozilla-mobile/reference-browser/pull/1110

components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
```diff
@@ -98,12 +101,36 @@ class SessionManager(
...
+        // migration is complete and the session storage uses the store instead
+        // of the session manager.
+        return snapshot.copy(
+            sessions = snapshot.sessions.map { item ->
+                store?.state?.findTab(item.session.id)?.let {
+                    item.copy(readerState = it.readerState)
+                } ?: item
+            }
+        )
+    }
```
pocmo, line 118: Ah, nice. Yeah, that is a good interim solution. :)

csadilek:
bors r=pocmo

Issue #6558, opened 2020-04-07
Allow simple highlightable menu items


ValentinTimisica:
We need to create a simple menu item which can be highlighted. Currently we have `BrowserMenuHighlightableItem` but we need a simple item (no images/icons) like `SimpleBrowserMenuItem`.

See [Fenix issue](https://github.com/mozilla-mobile/fenix/issues/9505) and [UX mocks](https://share.goabstract.com/33afd8de-f90e-4504-90e3-ee2a3bcb1925?mode=design&sha=0239115bee7129b68b46631245ea89540bda7e15).

ValentinTimisica:
Verified as fixed on [Fenix issue](https://github.com/mozilla-mobile/fenix/issues/9505#issuecomment-626576943).

PR #6559, opened 2020-04-07
For #6558: Adds a menu with simple highlightable items

ValentinTimisica:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/menu/src/main/java/mozilla/components/browser/menu/item/SimpleBrowserMenuHighlightableItem.kt
+++ b/components/browser/menu/src/main/java/mozilla/components/browser/menu/item/SimpleBrowserMenuHighlightableItem.kt
@@ -1,0 +1,107 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.menu.item
+
+import android.content.Context
+import android.view.View
+import android.widget.TextView
+import androidx.annotation.ColorInt
+import androidx.annotation.ColorRes
+import androidx.core.content.ContextCompat
+import mozilla.components.browser.menu.BrowserMenu
+import mozilla.components.browser.menu.BrowserMenuItem
+import mozilla.components.browser.menu.R
+import mozilla.components.browser.menu.ext.addRippleEffect
+import mozilla.components.browser.menu2.candidate.ContainerStyle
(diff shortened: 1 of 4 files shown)


eliserichards:
Hey @pocmo @jonalmeida @Amejia481! Any update on when this will be merged? 🤗

ValentinTimisica:
@jonalmeida I don't really understand what do you mean by "just use the mozac_browser_menu_item_simple.xml layout". Do you want me to modify the existing `SimpleBrowserMenuItem.kt` to support highlight? (this will break the purpose of 'Simple', that is why I created a new item which support highlight and can be reused everywhere).

jonalmeida:
For the PR, we're discussing the changes offline to only add the things that we need for this feature.

components/browser/menu/src/main/java/mozilla/components/browser/menu/item/SimpleBrowserMenuHighlightableItem.kt
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        wasHighlighted = isHighlighted()
+        updateHighlight(view, wasHighlighted)
+    }
+
+    private fun bindText(view: View) {
+        val textView = view as TextView
+        textView.text = label
+        textView.setColorResource(textColorResource)
```
jonalmeida, line 65: Probably need the same `NO_ID` check here?

components/browser/menu/src/main/java/mozilla/components/browser/menu/item/SimpleBrowserMenuHighlightableItem.kt
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+
+    private fun updateHighlight(view: View, isHighlighted: Boolean) {
+        val textView = view as TextView
+
+        if (isHighlighted) {
+            textView.setBackgroundColor(backgroundTint)
+        } else {
+            textView.addRippleEffect()
```
jonalmeida, line 86: nit: Not sure if this is what you want, but if you wanted to keep the ripple for the TextView, you could add it as a `foreground` so that you don't lose that animation. Up to you!

ValentinTimisica: Seems like to set the `foreground` requires API level 23, and our current min is 21, so I suggest we keep the current implementation. What do you think?

jonalmeida: Ah never mind then. :)

PR #6561, opened 2020-04-07
Issue #6314: Support fetching non base64 encoded data URIs

csadilek:
Simple change to support data URI that are not base64 encoded. With this backup/restore is now finally working in uBlock.


csadilek:
bors r=Amejia481

PR #6566, opened 2020-04-07
For #6564 - Only display URL hostname for tabstray

### LINKED ISSUE #6564: Tabstray items should only display URL hostname, not entire URL
At the moment the tabstray receives an entire URL (i.e. `https://davidwalsh.name`); we should ensure it only displays `davidwalsh.name`.

darkwing:
Fixes #6564. We don't want the entire URL, just the hostname, per @topotropic's direction.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/tabstray/src/main/java/mozilla/components/browser/tabstray/TabViewHolder.kt
+++ b/components/browser/tabstray/src/main/java/mozilla/components/browser/tabstray/TabViewHolder.kt
@@ -15,6 +15,7 @@ import mozilla.components.browser.tabstray.thumbnail.TabThumbnailView
 import mozilla.components.concept.tabstray.Tab
 import mozilla.components.concept.tabstray.TabsTray
 import mozilla.components.support.base.observer.Observable
+import mozilla.components.support.ktx.kotlin.tryGetHostFromUrl
 
 /**
  * A RecyclerView ViewHolder implementation for "tab" items.
@@ -47,7 +48,7 @@ class TabViewHolder(
         }
 
         titleView.text = title
-        urlView?.text = tab.url
+        urlView?.text = tab.url.tryGetHostFromUrl()
 
         itemView.setOnClickListener {
             observable.notifyObservers { onTabSelected(tab) }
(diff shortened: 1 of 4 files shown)


components/browser/tabstray/src/main/java/mozilla/components/browser/tabstray/TabViewHolder.kt
```diff
@@ -47,7 +49,13 @@ class TabViewHolder(
...
 
         titleView.text = title
-        urlView?.text = tab.url
+
+        var url = tab.url
+        try {
+            url = URL(tab.url).host
+        }
+        catch (e: MalformedURLException) { /* ignore error */ }
+        urlView?.text = url
```
jonalmeida, line 58: Let's use the catch clause properly to set the fallback URL if the parsing fails.

```suggestion
        try {
            val tabHost = URL(tab.url).host
            urlView?.text = tabHost
        } catch (e: MalformedURLException) {
            urlView?.text = tab.url
        }
```

It's generally not good practice to not handle the exceptions. If this is not the place you want to handle it, you can annotate the as a method that throws exceptions and the caller will then have to put the try-catch clause appropriately

jonalmeida:
We also need to add a changelog entry. :)

components/browser/tabstray/src/main/java/mozilla/components/browser/tabstray/TabViewHolder.kt
csadilek: nit: we should lift the assignment out of the try catch:

```kotlin
urlView?.text = try {
    URL(tab.url).host          
} catch (e: MalformedURLException) {
    tab.url
}
```

csadilek: We also have an extension function for this we could re-use here: https://github.com/mozilla-mobile/android-components/blob/f362c94e32c715f334b3880f7bb696a94cdbca6a/components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlin/String.kt#L107

So, you can just call:

```kotlin
tab.url.tryGetHostFromUrl()
```

jonalmeida:
bors retry

PR #6599, opened 2020-04-09
Closes #6584: TabUseCases may call loadUrl twice

### LINKED ISSUE #6584: AddNewTabUseCase calls loadUrl twice
This happens for initial page loads if an engine session needs to still be linked. Then we load the URL once while linking and once right after via the call `loadUrl`:

https://github.com/mozilla-mobile/android-components/blob/master/components/feature/tabs/src/main/java/mozilla/components/feature/tabs/TabsUseCases.kt#L129

csadilek:


csadilek:
bors retry

PR #6634, opened 2020-04-13
Closes #6633: Reduce AMO read timeout for migration

### LINKED ISSUE #6633: Reduce AMO read timeout to 5s for migration
Our current timeout is 20s. Let's reduce to speed up migration runs.

csadilek:
Makes the timeout configurable and uses 5s for migration.

### DIFF
--- a/components/feature/addons/src/main/java/mozilla/components/feature/addons/amo/AddonCollectionProvider.kt
+++ b/components/feature/addons/src/main/java/mozilla/components/feature/addons/amo/AddonCollectionProvider.kt
@@ -31,7 +31,7 @@ internal const val DEFAULT_SERVER_URL = "https://addons.mozilla.org"
 internal const val DEFAULT_COLLECTION_NAME = "7e8d6dc651b54ab385fb8791bf9dac"
 internal const val COLLECTION_FILE_NAME = "mozilla_components_addon_collection_%s.json"
 internal const val MINUTE_IN_MS = 60 * 1000
-internal const val READ_TIMEOUT_IN_SECONDS = 20L
+internal const val DEFAULT_READ_TIMEOUT_IN_SECONDS = 20L
 
 /**
  * Provide access to the collections AMO API.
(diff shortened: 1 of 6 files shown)


components/feature/addons/src/main/java/mozilla/components/feature/addons/amo/AddonCollectionProvider.kt
```diff
@@ -64,11 +64,14 @@ class AddonCollectionProvider(
...
      * from a previously cached response, defaults to true.
+     * @param readTimeoutInSeconds optional timeout in seconds to use when fetching
+     * available add-ons from a remote endpoint. If not specified [DEFAULT_READ_TIMEOUT_IN_SECONDS]
+     * will be used.
      * @throws IOException if the request failed, or could not be executed due to cancellation,
      * a connectivity problem or a timeout.
      */
     @Throws(IOException::class)
-    override suspend fun getAvailableAddons(allowCache: Boolean): List<Addon> {
+    override suspend fun getAvailableAddons(allowCache: Boolean, readTimeoutInSeconds: Long?): List<Addon> {
```
grigoryk, line 74: nit:
I'd prefer `readTimeoutInSeconds: Long? = DEFAULT_READ_TIMEOUT_IN_SECONDS` since it makes it obvious at first glance at the function signature what the default is; as-is, you need to read through the method to find out what happens when `readTimeoutInSeconds` is null.

... but I to do this, I think you'd need to declare the default at the interface level, and change it from optional to a required value.

grigoryk: ... at which point, it's no longer super obvious what's happening when you actually read the implementation (since the default value is one step removed). So, yeah... Nevermind, I guess :)

csadilek: :) Yeah, I came down that same path. This seems to be the best compromise.

csadilek:
bors r=grigoryk

Issue #6644, opened 2020-04-14
Integrate UI to cancel add-on installation


csadilek:
Breaking this off https://github.com/mozilla-mobile/android-components/issues/5748. This is to add the "Cancel" button to the UI and wire it up in Sample Browser/R-B and Fenix.

Blocked on: https://bugzilla.mozilla.org/show_bug.cgi?id=1629952

csadilek:
New GV issues filed.

PR #6645, opened 2020-04-14
Closes #5748: Provide API to cancel add-on installation

### LINKED ISSUE #5748: Provide API to cancel add-on installation
We currently have no way of cancelling after installation of an add-on has started. 

This requires:
https://bugzilla.mozilla.org/show_bug.cgi?id=1612097

csadilek:
This is working but ran into a crash that happens when cancelling too late (after the extension was downloaded but before it was installed): https://bugzilla.mozilla.org/show_bug.cgi?id=1629952

So, let's not wire up the UI just yet: https://github.com/mozilla-mobile/android-components/issues/6644

Notes:
- We may eventually turn the returned operation into a full `CompletableFuture`-like type that supports `cancel` as well, and remove our callbacks, but this is a pretty big refactoring that I am not fully convinced of yet :)
- Cancel returns another `GeckoResult` to indicate if cancellation worked, which I am wrapping in a `Deferred` so we can `await` on it and prevent another round of callbacks. This works out quite well in the UI: https://github.com/mozilla-mobile/android-components/commit/1dbbd7eba02cdac80c43f9ed8eaf12bb0a3c754b

### DIFF
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
@@ -14,6 +14,7 @@ import mozilla.components.browser.engine.gecko.webextension.GeckoWebExtension
 import mozilla.components.browser.engine.gecko.webnotifications.GeckoWebNotificationDelegate
 import mozilla.components.browser.engine.gecko.webpush.GeckoWebPushDelegate
 import mozilla.components.browser.engine.gecko.webpush.GeckoWebPushHandler
+import mozilla.components.concept.engine.CancellableOperation
 import mozilla.components.concept.engine.Engine
 import mozilla.components.concept.engine.EngineSession
 import mozilla.components.concept.engine.EngineSession.SafeBrowsingPolicy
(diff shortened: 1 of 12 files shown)


components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
```diff
@@ -196,17 +197,17 @@ class GeckoEngine(
...
     }
 
     internal fun installWebExtension(
         ext: GeckoWebExtension,
         onSuccess: ((WebExtension) -> Unit) = { },
         onError: ((String, Throwable) -> Unit) = { _, _ -> }
-    ) {
-        if (ext.isBuiltIn()) {
+    ): CancellableOperation {
+        val geckoResult = if (ext.isBuiltIn()) {
```
psymoon, line 210: nit: `return if`?  Although arguably this is more readable :).

csadilek: Yeah, let's leave it, it gets harder to read otherwise ;).

csadilek:
bors r=Amejia481,psymoon

PR #6652, opened 2020-04-15
Closes #6651 - Don't restore session source in SnapshotSerializer

### LINKED ISSUE #6651: Restoring source should be optional in SnapshotSerializer
When we save to a collection and then restore the tab, it's strange behavior to keep the source, but there's currently no way to strip the source when the session is already created AFAIK

ekager:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
@@ -174,7 +174,12 @@ class Session(
         /**
          * User entered a URL or search term
          */
-        USER_ENTERED
+        USER_ENTERED,
+
+        /**
+         * This session was restored
+         */
+        RESTORED
     }
 
     /**
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/storage/SnapshotSerializer.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/storage/SnapshotSerializer.kt
@@ -83,10 +83,12 @@ class SnapshotSerializer(
     fun itemFromJSON(engine: Engine, json: JSONObject): SessionManager.Snapshot.Item {
         val sessionJson = json.getJSONObject(Keys.SESSION_KEY)
         val session = deserializeSession(sessionJson, restoreSessionIds, restoreParentIds)
-        val readerState = ReaderState(active = sessionJson.optBoolean(Keys.SESSION_READER_MODE_KEY, false))
+        val readerState =
+            ReaderState(active = sessionJson.optBoolean(Keys.SESSION_READER_MODE_KEY, false))
         val engineState = engine.createSessionState(sessionJson)
 
-        return SessionManager.Snapshot.Item(session,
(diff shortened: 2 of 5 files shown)


components/browser/session/src/main/java/mozilla/components/browser/session/ext/AtomicFile.kt
```diff
@@ -48,7 +48,12 @@ fun AtomicFile.readSnapshotItem(
     engine: Engine,
     restoreSessionId: Boolean = true,
     restoreParentId: Boolean = true,
-    serializer: SnapshotSerializer = SnapshotSerializer(restoreSessionId, restoreParentId)
+    restoreSource: Boolean = true,
```
pocmo, line 51: Just commented on the issue. Let's see why we decided to restore it in the first place. Maybe we shouldn't do that. 🤔

pocmo: Found it: b640faedb64f13d2a3e0f438233061c90c574ea1 - Issue #407

pocmo: Looking at the issue and changes I think we primarily wanted to add the source to session and there was no strong reason to save and restore it.
The only downside of not saving it to disk is that when we switch apps and the app gets destroyed and recreated then we would no longer have the original source assigned. 🤔

ekager: For a normal tab I guess this could make sense to save/restore? Since we would still maybe want to return to the app that created the ACTION_VIEW session on back press if the app got recreated? But that seems like an edge case where the user would have left the tab/app to begin with and maybe we don't care about restoring the source anymore there. 

Happy to follow your lead here! I don't have strong opinions except that I don't want it restored from a collection :)

pocmo: > maybe want to return to the app that created the ACTION_VIEW session on back press if the app got recreated?

I think so too for fast switches between apps. But imagine launching the apps three days later, we restore, you press back on a tab and leave the browser. That seems confusing too since you have no clue why this happens.

> Happy to follow your lead here! I don't have strong opinions except that I don't want it restored from a collection :)

Both options are not perfect, I think. What I like about the "restore" source is that we will not need to pass down the boolean, resulting in less code and less configurations to test and maintain... and hopefully less bugs.

Maybe the "back" behavior is actually more consistent and understandable that way: If you open a link from an app and press back then you get back to the app and we close the tab. If you continue to do something else then the tab will start to stick around and behave like a normal tab eventually.

wdyt?

ekager: Yeah I agree, let's not restore the source. The edge case of the app dying before the user takes any action seems not that important compared to how confusing it is to have inconsistent behavior.

ekager:
bors r=pocmo

Issue #2930, opened 2019-05-05
Pinch to Zoom Triggers Toolbar Behavior and Doesn't Zoom


ekager:
Similar to #2695 
See Fenix issue https://github.com/mozilla-mobile/fenix/issues/2278
"Pinch to zoom trigger the address bar, even if the page doesn't scroll."

Mugurell:
I think this issue stems from the fact that we don't have a clear way of knowing if the panning gesture was used for scrolling or for zooming. We treat both the same.

This explains why [panning even with two fingers](https://drive.google.com/file/d/1cNNq3renCvr1JUs5ZkEj-QR6C3uTs0GO) will animate the navbar.
But even [one finger zoom](https://drive.google.com/file/d/1BMdSf5jk-s26NG5HEpoFOWCrjb_flyD3) will have the same effect.

We need to have a proper way to differentiate between this 2 gestures.

PR #6666, opened 2020-04-16
For #2930 - Different BrowserToolbar behaviors depending on gestures

Mugurell:
We'll now have a more in-depth understanding of user's gestures based on which
we can trigger different BrowserToolbar behaviors depending on if the user is
scrolling or zooming.

The ConstraintLayout parent will still inform us about when the scroll begins
and when the scroll ends but then actually incrementally updating the toolbar's
translation will be done in response to GestureDetector callbacks.
This also allows us to easily differentiate between horizontal and vertical
scrolls and only react for the latter.

If the user's gestures as inferred as zoom gestures (multi-finger scale or
quick-scale) the toolbar should not be animated.
This scenario is though affected by the fact that normally scroll gestures have
a much higher sensitivity (a 1 pixel movement is enough for a gesture to be
considered a scroll gesture) and so it is possible that even in the case of a
desired zoom gesture the scroll callbacks will be fired first.
When the gesture is identified as a zoom gesture we will immediately snap the
toolbar up/down but there might be a minor visual glitch because of this.
In an effort to reduce such races between scroll and zoom gestures I've lowered
the sensitivity of the ScaleDetector, this having the other effect that
multi-finger scrolls must now be more precise. Pinching while zooming will now
more often mean a zoom gesture and not animate the toolbar.

---

### Pull Request checklist

- [x] **Tests**: This PR includes tests and and explanation of why not even more thorough tests.

### After merge

### DIFF
--- a/components/browser/toolbar/src/test/java/mozilla/components/browser/toolbar/behavior/TestUtils.kt
+++ b/components/browser/toolbar/src/test/java/mozilla/components/browser/toolbar/behavior/TestUtils.kt
@@ -1,0 +1,62 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.toolbar.behavior
+
+import android.view.MotionEvent
+
+object TestUtils {
+    fun getMotionEvent(
+        action: Int,
+        x: Float = 0f,
+        y: Float = 0f,
+        previousEvent: MotionEvent? = null
+    ): MotionEvent {
+        val currentTime = System.currentTimeMillis()
+        val downTime = previousEvent?.downTime ?: System.currentTimeMillis()
+
+        var pointerCount = previousEvent?.pointerCount ?: 0
+        if (action == MotionEvent.ACTION_POINTER_DOWN) {
+            pointerCount++
+        } else if (action == MotionEvent.ACTION_DOWN) {
+            pointerCount = 1
+        }
+
+        val properties = Array(pointerCount, ::getPointerProperties)
+        val pointerCoords = getPointerCoords(x, y, pointerCount)
+
+        return MotionEvent.obtain(
+            downTime, currentTime,
+            action, pointerCount, properties,
+            pointerCoords, 0, 0, 1f, 1f, 0, 0, 0, 0
+        )
+    }
+
+    private fun getPointerCoords(
+        x: Float,
+        y: Float,
+        pointerCount: Int,
+        previousEvent: MotionEvent? = null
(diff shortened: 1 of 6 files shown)


Mugurell:
This patch might the previous behavior in order to better support scroll / zoom differentiation.
Scrolls have an innate higher sensitivity than zoom gestures. As such even in the event of a desired zoom gesture, one finger might generate an action before the other finger and such generating first a scroll movement or even all fingers to be 
This cannot be overcome other than having the ScaleDetector be the sole responder to two finger gestures.
Lowering the sensitivity of the GestureDetector (used for scrolls) is not a solution as the user expects to see an update on the screen even for 1px scrolls (very small, slows pan gestures), a much lower value than what we need for detecting a zoom.
I went with a compromise: increase with 50% the sensitivity of the ScaleDetector.
Pinching while zooming will now more often mean a zoom gesture and not animate the toolbar and in so minimizing an innate "flicker" for when scrolls first animate the y translation of the toolbar.
This issue exists in other apps also and with the proposed 50% sensitivity increase we obtain a similar experience as on Chrome for example.
Without this - https://drive.google.com/open?id=1U_PWaCZF5B1zTpl76HJxp0LJNoj3XRmy
After this - https://drive.google.com/open?id=1MpXSIPHsIRHme29gIwFtpWKs_Zfhq8oq

components/browser/toolbar/src/test/java/mozilla/components/browser/toolbar/behavior/TestUtils.kt
```diff
@@ -0,0 +1,58 @@
+package mozilla.components.browser.toolbar.behavior
```
pocmo, line 1: nit: license header.

docs/changelog.md
```diff
@@ -35,6 +35,9 @@ permalink: /changelog/
 
 * **browser-tabstray**
   * The iconView is no longer required in the template.
+  
+* **browser-toolbar**
+  * It will only be animated for vertical scrolls inside the EngineView. Not for horizontal scrolls. Not for zoom gestures.
```
pocmo, line 40: nit: Can you move this to the latest release? (Sorry, that the review took so long)

components/browser/toolbar/src/test/java/mozilla/components/browser/toolbar/behavior/TestUtils.kt
Mugurell: Thanks!

docs/changelog.md
Mugurell: Sure, np!

Mugurell:
"An error occurred while preparing SDK package Android SDK Build-Tools 28.0.3: Connection reset.:"
Restarting CI jobs.

Mugurell:
bors r=pocmo

PR #6694, opened 2020-04-16
Adds UI Widgets component with Button styling

sblatz:
---

### Pull Request checklist

### After merge

### DIFF
--- a/.buildconfig.yml
+++ b/.buildconfig.yml
@@ -224,6 +224,10 @@ projects:
     path: components/ui/colors
     description: 'The standard set of Photon colors.'
     publish: true
+  ui-widgets:
+    path: components/ui/widgets
+    description: 'The standard set of Mozilla widgets.'
+    publish: true
   ui-fonts:
     path: components/ui/fonts
     description: 'Convenience accessor for fonts used by Mozilla.'
--- a/components/ui/widgets/src/main/AndroidManifest.xml
+++ b/components/ui/widgets/src/main/AndroidManifest.xml
@@ -1,0 +1,4 @@
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<manifest package="mozilla.components.ui.widgets" />
--- a/components/ui/widgets/src/main/res/values/colors.xml
+++ b/components/ui/widgets/src/main/res/values/colors.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+
+<resources>
+    <!-- Button Colors -->
+    <color name="mozac_grey_button_color">#E0E0E6</color>
+    <color name="mozac_destructive_button_text_color">#C50042</color>
+    <color name="mozac_grey_button_text_color">#312A65</color>
+</resources>
--- a/components/ui/widgets/src/main/res/values/styles.xml
+++ b/components/ui/widgets/src/main/res/values/styles.xml
@@ -1,0 +1,35 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+
+<resources xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <!-- Button styling -->
+    <style name="Mozac.Widgets.NeutralButton" parent="Widget.MaterialComponents.Button.TextButton">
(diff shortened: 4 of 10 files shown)


.buildconfig.yml
```diff
@@ -224,6 +224,10 @@ projects:
     path: components/ui/colors
     description: 'The standard set of Photon colors.'
     publish: true
+  ui-components:
+    path: components/ui/components
```
pocmo, line 228: nit: Let's not use `component(s)` in the name of a component. If we say "ui components" then we usually mean everything that is `ui-*`. If there's also a component that has that exact name then things start to get confusing. :)

Some other options may be: `theme` or `styles` (in case you want to only drop styles here) or `widgets` if this is for ui widgets (The name `widget` may be misleading to. But it is consistent with the [Android package naming](https://developer.android.com/reference/android/widget/package-summary?hl=en)).

components/ui/components/README.md
```diff
@@ -0,0 +1,19 @@
+# [Android Components](../../../README.md) > UI > Colors
+
+The standard set of [Photon](https://design.firefox.com/photon/) colors.
```
pocmo, line 3: nit: This seems to be an unchanged copy of `ui-colors` and needs some edits. :)

components/ui/components/src/main/res/values/styles.xml
```diff
@@ -0,0 +1,28 @@
...
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+
+<resources
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools">
+
+<!-- UI button styling -->
+<style name="NeutralButton" parent="Widget.MaterialComponents.Button.TextButton">
```
pocmo, line 12: Nice! Great so see that a style seems to be sufficient. Does it work nicely in Fenix?

components/ui/widgets/src/main/res/values/styles.xml
```diff
@@ -0,0 +1,35 @@
...
+        <item name="fontFamily">@font/metropolis_semibold</item>
+    </style>
+
+    <style name="Mozac.Widgets.DestructiveButton" parent="Mozac.Widgets.NeutralButton">
+        <item name="iconTint">@color/destructive_button_text_color</item>
+        <item name="android:textColor">@color/destructive_button_text_color</item>
+    </style>
+
+    <style name="Mozac.Widgets.PositiveButton" parent="Mozac.Widgets.NeutralButton">
+        <item name="backgroundTint">?accent</item>
```
sblatz, line 31: @pocmo how do recommend approaching attributes we have defined in Fenix? Would it make sense to create an `accent` attr here and then we just "override" what that is in Fenix? I don't really want to duplicate a bunch color values here 😕

components/ui/widgets/src/main/res/values/styles.xml
```diff
@@ -0,0 +1,35 @@
...
+        <item name="android:textAlignment">center</item>
+        <item name="android:background">@drawable/rounded_button_background</item>
+        <item name="android:layout_width">match_parent</item>
+        <item name="android:layout_height">48dp</item>
+        <item name="android:textStyle">bold</item>
+        <item name="android:textAllCaps">false</item>
+        <item name="backgroundTint">@color/grey_button_color</item>
+        <item name="android:textColor">@color/button_text_color</item>
+        <item name="android:letterSpacing">0</item>
+        <item name="fontFamily">@font/metropolis_semibold</item>
```
sblatz, line 22: @pocmo Is it okay that I added fonts as part of this component? My thought is: if we include fonts in this component, our UI will look more consistent across product lines. If I don't include it here, we could just create a `NeutralButton` in fenix that overrides `fontFamily` but if I'm *already* forking the ui-widget, is there a point in it even being inside of AC? 😓

Amejia481: For non-Mozilla consumer, they will have to pay the space price without using the font :(

components/ui/widgets/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,4 @@
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<manifest package="mozilla.components.ui.components" />
```
pocmo, line 4: `components` -> `widgets`

sblatz: fixed, thanks!

components/ui/widgets/src/main/res/values/colors.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+
+<resources>
+    <!-- Button Colors -->
+    <color name="grey_button_color">#E0E0E6</color>
```
Amejia481, line 8: Should this be prefixed with `mozac` too?

Amejia481:
bors r=pocmo,Amejia481

Issue #6548, opened 2020-04-06
Force expand toolbar if content cannot handle touches


sblatz:
This is to avoid Fenix bug: https://github.com/mozilla-mobile/fenix/issues/9517

PR #6710, opened 2020-04-20
For 10040 & 10045 (fenix) - Toolbar: Consistently hide/show bottom to…

JLRoberts:
…olbar

Relying on input to be handled  by Engine View provided inconsistent showing / hiding of the toolbar.

If the site is able to scroll and scrolling, show and hide the toolbar. This will consistently show the toolbar when scrolling up and hide it scrolling down.

Also if multitouch event, dont translate the toolbar to support pinch/zoom gestures without affecting the toolbar.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/behavior/BrowserToolbarBottomBehavior.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/behavior/BrowserToolbarBottomBehavior.kt
@@ -53,20 +53,6 @@ class BrowserToolbarBottomBehavior(
      */
     private var engineView: EngineView? = null
 
-    /**
-     * Depending on how user's touch was consumed by EngineView / current website,
-     *
-     * we will animate the dynamic navigation bar if:
-     * - touches were used for zooming / panning operations in the website.
-     *
-     * We will do nothing if:
-     * - the website is not scrollable
-     * - the website handles the touch events itself through it's own touch event listeners.
-     */
-    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
-    internal val shouldScroll: Boolean
-        get() = engineView?.getInputResult() == EngineView.InputResult.INPUT_RESULT_HANDLED
-
     override fun onStartNestedScroll(
         coordinatorLayout: CoordinatorLayout,
         child: BrowserToolbar,
@@ -75,19 +61,18 @@ class BrowserToolbarBottomBehavior(
         axes: Int,
         type: Int
     ): Boolean {
-        return if (shouldScroll && axes == ViewCompat.SCROLL_AXIS_VERTICAL) {
-            shouldSnapAfterScroll = type == ViewCompat.TYPE_TOUCH
+        val started = axes == ViewCompat.SCROLL_AXIS_VERTICAL
+        shouldSnapAfterScroll = type == ViewCompat.TYPE_TOUCH
+
+        if (snapAnimator.isRunning && started) {
             snapAnimator.cancel()
-            true
-        } else if (engineView?.getInputResult() == EngineView.InputResult.INPUT_RESULT_UNHANDLED) {
-            // Force expand the toolbar if event is unhandled, otherwise user could get stuck in a
-            // state where they cannot show the toolbar
-            snapAnimator.cancel()
-            forceExpand(child)
-            false
-        } else {
-            false
         }
+
+        if (engineView?.getInputResult() == EngineView.InputResult.INPUT_RESULT_UNHANDLED) {
+            forceExpand(child)
+        }
+
+        return started
     }
 
     override fun onStopNestedScroll(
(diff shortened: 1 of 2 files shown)


JLRoberts:
Need to add testing..

Posting for feedback on changes.

JLRoberts:
Need to fix testing, updated to fix compile, but tests fail. Update pending. 

Still need feedback on the change of behavior.

Mugurell:
Hello and thanks for quickly coming with a solution for the aforementioned issues!

Using EV's `canScrollVerticallyUp` and `canScrollVerticallyDown` is an interesting idea, though at the moment `canScrollVerticallyDown` [will always return true](https://github.com/mozilla-mobile/android-components/blob/efe29011666b238661253702974003423d5a469e/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineView.kt#L171) so we will always expand the toolbar for a vertical up scroll.
I think this would be ok..ish, maybe with a comment in code, but I'm not sure it would be ok with all edgecases the toolbar currently has to account for.
Also I was thinking that the issues described in https://github.com/mozilla-mobile/fenix/issues/10045 and https://github.com/mozilla-mobile/fenix/issues/10046 could be the result of a recent regression from https://github.com/mozilla-mobile/android-components/issues/6548 which in the meantime has already been resolved.
Asking for @sblatz's opinion for the above.

&nbsp;
Regarding the second part, ignoring multi touch events (>= 2 finger scrolls), I don't think we should go by this route because:
- other browsers are supporting this
- quick scale - one finger zoom (double tap followed by a drag)  would still animate the toolbar while two finger zoom not.

For differentiating between scroll and zoom gestures I have an open PR - https://github.com/mozilla-mobile/android-components/pull/6666

&nbsp;
Regarding failed tests I see 3 of them failing in `BrowserToolbarBottomBehaviorTest` followin the refactorization. Should be fairly easy to fix. If you need our support please tell.

&nbsp;
Regarding PRs ceremony, since android-components and fenix are different projects we would need an issue opened in android-components referring the issues in fenix and describing the problem, then this PR commited against that new a-c issue. Again, we can help here if needed.
Also because of the behavior change for all clients I think a changelog entry would be helpful.

JLRoberts:
Thank you for the feedback. First I will update to remove the multitouch scroll part and just focus on the basic toolbar behavior. I did see the commits for force expand. I should have put a video up as well to describe. But losing the toolbar is not fixed by that commit. Scrolling on the main GitHub page (logged out) you'll find several areas that you will end up with no toolbar. 

I can post a video of this if needed in the issue that is created on this side. And update the commit message to reference those commits shortly. 

I didn't think the canscrollverticallyup / down would work as is since I don't think you would be able to test it like it is either. 

As far as down always returning true, Ill relook at engineview. It was working for me to prevent the toolbar from hiding on short sites and boxes inside sites. But I'll look again.

JLRoberts:
Oh actually on second thought. I guess that side of the check can probably be removed them anyways. Since all that check was made to do was prevent hiding/showing toolbar on short sites. But if you can't scroll up. You don't need to check for scrolling down since the toolbar is never going anywhere. Ok I'll update that too. Thanks for the quick feedback btw.

Mugurell:
> Oh actually on second thought. I guess that side of the check can probably be removed them anyways. Since all that check was made to do was prevent hiding/showing toolbar on short sites. But if you can't scroll up. You don't need to check for scrolling down since the toolbar is never going anywhere. Ok I'll update that too. Thanks for the quick feedback btw.

Indeed. That's why I said this would be `ok..ish`. Don't really know how you could get in such a scenario. Maybe a comment in code would help to show future others that we thought about this though.

JLRoberts:
Updated changes to what we discussed
- removed multitouch check
- added comment to section in question, although I might should update that even further to explain why we don't check the other direction
- opened issue here in android-components to refer to

If this is what we want, I'll find help for the tests as I'm unsure of the changes that need to happen there.

JLRoberts:
@sblatz - Updated commit with an idea. If we don't hide the toolbar on scrolling up when unhandled by engineview, then we shouldn't need to force expand and lock, same as if the site was unscrollable.

I still haven't been able to find a site to really test this outside of bing.com homepage which has a really strange scrolling behavior for the top bar too currently. Adding this bit though results in a static toolbar instead of showing the whitespace like it did previously.

sblatz:
@JLRoberts hmm this new solution leads to a situation where the toolbar can't be expanded if I go from scrollable content -> hide the toolbar -> unscrollable content

scroll fails 2020-04-21 13_21_48

This happens on Instagram for example, when scrolling a user's profile without an account and then you're stuck on this page without a toolbar:

image

JLRoberts:
Yes I was thinking about this while I was digging a hole earlier. I hadn't run into it but it's possible. I do think it's getting closer. I don't get as much time to work on stuff the rest of the week but I'll keep thinking about it. If there's any ideas that we don't want to pollute this commit further on I'm also in the matrix chats and what not. Thanks for checking it out though.

JLRoberts:
@sblatz @Mugurell thanks for the feedback on this. I'm going to close this for now. As it seems at the moment getting a more reliable showing of the toolbar breaks it in other ways without maybe more information from engineview. And that probably needs more choices than me poking around here. 

The last push fixes instances like the Instagram. I think really the main thing gained is to avoid returning false in onstart and handling what needs to be done on the actual scroll since returning false means you get no other scroll events for the duration of that scroll.

But it does break it in instances like a small scroll box where the web page view doesn't actually move. Take for instance like stack overflow code snippets. The page isn't moving but the toolbar would hide. So the input is being handled but I don't think the user would expect that the toolbar hides when the actual page isn't moving. But I don't think there's a way currently to know the difference. This also results in sometimes clipped views where the toolbar would be. 

The top toolbar behaves also much differently since it does seem to react more to the entire view being moved and not just scroll events inside. But it's also placed differently being on top of the view and not overlaid above.

PR #6716, opened 2020-04-20
Bug 1625126 - Cache external dependencies in a single task and let gr…

JohanLorenzo:
…adle tasks use it

Backports https://github.com/mozilla-mobile/reference-browser/pull/1117 with some changes to grasp the complexity of android-components.

---

### Pull Request checklist

### After merge


pocmo:
> Task timeout after 1800 seconds

Heh, yeah, running all this in a single task takes quite some time.. :)

pocmo:
Hm, they all seem to have problems with jacoco? 🤔 

> Could not find org.jacoco:org.jacoco.agent:0.7.9.

pocmo:
Also all tasks seem to download gradle themselves:

> [task 2020-04-21T09:38:15.931Z] + automation/taskcluster/helper/get-secret.py -s project/mobile/android-components/public-tokens -k codecov -f .cc_token
[task 2020-04-21T09:38:16.069Z] + ./gradlew -PgoogleRepo=file:///builds/worker/fetches/android-gradle-dependencies/google -PjcenterRepo=file:///builds/worker/fetches/android-gradle-dependencies/jcenter listRepositories -Pcoverage :ui-tabcounter:assemble :ui-tabcounter:assembleAndroidTest :ui-tabcounter:test :ui-tabcounter:lintRelease
> [task 2020-04-21T09:38:16.151Z] Downloading https://services.gradle.org/distributions/gradle-5.6.4-bin.zip
> [task 2020-04-21T09:38:18.042Z] .........................................................................................

JohanLorenzo:
> Hm, they all seem to have problems with jacoco? 🤔

Yeah, I just looked into it and jacoco is present in the archive, but not version 0.7.9. The archive only contains 0.8.4 and 0.8.5. I guess the dependencies task isn't building everything yet.

JohanLorenzo:
> Also all tasks seem to download gradle themselves:

You're right. @tomprince called it out in mozilla-mobile/reference-browser#1093 (comment). I haven't had time to look into it. I'm still unsure whether caching it works the same way as regular deps.

pocmo:
@JohanLorenzo Ah, if you haven't figured it out yet: I think we need to add `--Pcoverage` because only that will enable jacoco and test coverage. :)

JohanLorenzo:
Oh, thanks for pointing this out! I've made a solution that rebuilds every single component. I'm unsure that's the right way to go. I'll wait for the test results. If it all goes well, I'll try to optimize it with your approach.

taskcluster/ci/toolchain/android.yml
```diff
@@ -30,3 +27,34 @@ linux64-android-sdk-linux-repack:
...
+            # Aliases aren't allowed for toolchains depending on toolchains.
+            - linux64-android-sdk-linux-repack
+    run:
+        script: android-gradle-dependencies.sh
+        sparse-profile: null
+        resources:
+            - taskcluster/scripts/toolchain/android-gradle-dependencies.sh
+            - taskcluster/scripts/toolchain/android-gradle-dependencies/**
+            - buildSrc/src/main/java/Dependencies.kt
+            - '*.gradle'
```
JohanLorenzo, line 48: I expanded the list to include all components. 

⚠️ This task is spun if a single component gets its dependencies updated. This introduces a 2-hour-delay before getting results. I think it's okay if dependencies aren't updated that often, but it becomes a problem if many PRs do it. Does this sound okay-ish to you @pocmo?

taskcluster/ci/toolchain/android.yml
```diff
@@ -30,3 +27,34 @@ linux64-android-sdk-linux-repack:
...
+        toolchain-artifact: public/build/android-gradle-dependencies.tar.xz
+        toolchain-alias: android-gradle-dependencies
+    treeherder:
+        symbol: TL(gradle-dependencies)
+    worker:
+        docker-image: {in-tree: android-build}
+        env:
+            # TODO do no hardcode
+            ANDROID_SDK_ROOT: /builds/worker/fetches/android-sdk-linux
+        max-run-time: 14400
```
JohanLorenzo, line 60: The task takes 2 hours at the moment. I expanded the deadline to 4 hours, in case it gets slower from times to times.

taskcluster/scripts/toolchain/android-gradle-dependencies.sh
```diff
@@ -0,0 +1,44 @@
...
+CURRENT_DIR="$(get_abs_path $0)"
+PROJECT_DIR="$(get_abs_path $CURRENT_DIR/../../../..)"
+
+pushd $PROJECT_DIR
+
+. taskcluster/scripts/toolchain/android-gradle-dependencies/before.sh
+
+GRADLE_FLAGS='--no-daemon -PgoogleRepo=http://localhost:8081/nexus/content/repositories/google/ -PjcenterRepo=http://localhost:8081/nexus/content/repositories/jcenter/'
+
+./gradlew $GRADLE_FLAGS ktlint detekt
```
JohanLorenzo, line 23: The rest of the script is different from r-b. We basically read `.buildconfig.yml` to get all components and build each one of them. It currently takes 2 hours. I wonder if using `--parallel` is a good thing here. I know it caused some issues in the past. What do you think @pocmo?

pocmo: I wonder why call gradle for every component instead of once to build everything like previously?

Yeah, `--parallel` has been causing intermittent issues in tests. I think that would be particularly annoying when this long running task fails.

taskcluster/ci/toolchain/android.yml
pocmo: 2 hours is harsh. At what point would we have to wait that long, when opening the PR? Or when merging it? Or in both cases?

What was the reason for including `build.gradle`?

JohanLorenzo: Great, after a lot of trial and error, I managed to reduce the task time down to 20-30 minutes, which looks more acceptable to me.

This task is rebuilt when a PR changes once of the listed files then when we merge it. The reason is: the cache built in a PR is considered untrusted. It becomes trusted once it's merged and once it runs on trusted workers. 

As far as I see, `build.gradle` lists the dependencies each component has (the main `build.gradle` also lists some) so we need to track these files to make sure we cache the latest dependencies.

taskcluster/scripts/toolchain/android-gradle-dependencies.sh
JohanLorenzo: I ended using `--parallel` which shows the best results in terms of performance and reliability (I tried other ways and it didn't work out). I made the tests not fail the cache anymore, so I think it won't be too annoying.

JohanLorenzo:
I think we're good for a round of review. The current solution functionally works and is not as long as it was yesterday! r? @pocmo @tomprince

taskcluster/ac_taskgraph/job.py
```diff
@@ -57,13 +56,20 @@ def configure_gradlew(config, job, taskdesc):
...
 
-    gradle_command = ["./gradlew"] + run.pop("gradlew")
+    maven_dependencies_dir = path.join(fetches_dir, "android-gradle-dependencies")
+    gradle_repos_args = [
+        "-P{repo_name}Repo=file://{dir}/{repo_name}".format(
+            dir=maven_dependencies_dir, repo_name=repo_name
+        )
+        for repo_name in ("google", "jcenter")
+    ]
+    gradle_command = ["./gradlew"] + gradle_repos_args + ["listRepositories"] + run.pop("gradlew")
```
tomprince, line 72: I think there is a limit to the amount of complexity there should be constructing complex commands in taskgraph, vs having that complexity in a script[1]. I think this code is at least nearing that point (if not already past it).

[1] which would incidentally allow having richer error reporting and logging

taskcluster/scripts/toolchain/android-gradle-dependencies.sh
```diff
@@ -0,0 +1,50 @@
...
+}
+
+CURRENT_DIR="$(get_abs_path $0)"
+PROJECT_DIR="$(get_abs_path $CURRENT_DIR/../../../..)"
+
+pushd $PROJECT_DIR
+
+. taskcluster/scripts/toolchain/android-gradle-dependencies/before.sh
+
+COMPONENT_REGEX='^  ([-a-z]+):$'
```
tomprince, line 21: The code here has definitely outstripped what is reasonable to do in shell, and this should be rewritten in python (parsing yaml is definitely a point at which python should be used).

JohanLorenzo: I definitely agree. I changed this logic many times in the PR, and I've thought several times it would just be better in Python.

I'd like to tackle this in a separate bug so I can make all repos use the same python script.

taskcluster/ac_taskgraph/job.py
JohanLorenzo: That's a good idea! This script could then be used by both `job.py` and `android-gradle-dependencies.sh`. I'll have a look at it when I convert `android-gradle-dependencies.sh` into a Python script.

JohanLorenzo:
bors r=pocmo,tomprince

JohanLorenzo:
> Can you send a short mail to the public list or android-components-team so that no one is surprised if they end up with a long running task in their PR? :)

Done!

PR #6730, opened 2020-04-21
Close #6601: Update targetSdkVersion to 29

### LINKED ISSUE #6601: Update compile and target SDK to 29
With the Android 11 (SDK 30) soon to be released, we would be two versions below that.

We should update our `compileSdkVersion` and `targetSdkVersion` in `Config.kt` to 29 sooner than later.

jonalmeida:
---

This didn't cause any breaking API changes as I expected, but let's see what bors thinks.

I recall we had to make sure our clients had updated either their compile or target SDK as well the last time we did this, but I'll need to verify.

### Pull Request checklist

### After merge


jonalmeida:
I found this [Fire TV issue](https://github.com/mozilla-mobile/firefox-tv/issues/991) when we needed to update to 28 immediately. There doesn't seem to be any information there about consumers needing to update as well.

Robolectric hasn't updated to support API 29 (I tried version 4.3), so instead I offered for forcing it  to target 28 first.

I wonder if there are any other considerations to make that I'm missing.. cc @pocmo maybe?

pocmo:
> I wonder if there are any other considerations to make that I'm missing.. cc @pocmo maybe?

I'd just try building Fenix and RB with that locally, if you haven't. Just to make sure there are no surprises once this hits nightly. Maybe even do a local publish and/or kdoc run to make sure that still works too.

CC @JohanLorenzo This is maybe an example of a change (SDK version) that may need new dependencies and the change is not in one of the defined files? But then again I am not even sure if we cache the SDK yet since that is not a "gradle dependency" and downloaded by the android plugin as needed.

JohanLorenzo:
Great to know. Thanks for highlighting this, @pocmo. If I understand this PR correctly, then we can track `buildSrc/src/main/java/Config.kt`. If this file changes, the SDK will be downloaded and cached once more. 

I think that's a good-enough solution. What do you think?

jonalmeida:
> I'd just try building Fenix and RB with that locally, if you haven't. Just to make sure there are no surprises once this hits nightly. Maybe even do a local publish and/or kdoc run to make sure that still works too.

I verified that kdocs generate as well - I hadn't done that. Thanks!

> If I understand this PR correctly, then we can track buildSrc/src/main/java/Config.kt. If this file changes, the SDK will be downloaded and cached once more.

That sounds like a great idea to me.

JohanLorenzo:
Great!

Could you change https://github.com/mozilla-mobile/android-components/blob/9dfa58613e995577e7e9a6acb3187f51c2a76390/taskcluster/ci/toolchain/android.yml#L27 to be 

```yaml
      resources:
          - buildSrc/src/main/java/Config.kt
```
?

This way, this file is going to be tracked :)

jonalmeida:
> Great!
> 
> Could you change
> 
> https://github.com/mozilla-mobile/android-components/blob/9dfa58613e995577e7e9a6acb3187f51c2a76390/taskcluster/ci/toolchain/android.yml#L27
> to be
> 
> ```yaml
>       resources:
>           - buildSrc/src/main/java/Config.kt
> ```
> 
> ?
> 
> This way, this file is going to be tracked :)

Done, thanks!

jonalmeida:
bors r=pocmo

PR #6732, opened 2020-04-21
Closes #6233: Zero-cost autopublish workflow

### LINKED ISSUE #6233: Auto-publication: improve up-to-date checks
Currently in order to check if a-c needs to be re-published, a compilation is done. Specifically, Fenix will run `compileReleaseKotlin`, and look at gradle output. The logic is:
```
def compileTasksExecuted = compileOutput.toString().split('\n').last().split(':')[1].split(' ')[1]
if (compileTasksExecuted.equals("1")) {
    log("android-components are up-to-date, skipping publication.")
} else {
    log("android-components changed, publishing locally...")
}
```

This is problematic for two reasons:
- if a-c was compiled "out of band" of the autoPublish flow - e.g. manually, to run tests, and then a Fenix auto-publication build runs - it will execute a compile task, determine that nothing changed, and skipped re-publication. Thus, the out-of-band changes won't make it into the build.
- when nothing changed in a-c, we're incurring a compilation overhead - builds are incremental, so this is rather small; on my machine (top spec 16" macbook) it's 5-6 seconds of overhead, and could be much worse on slower machines

One alternative I'd like to explore is to rely on git hashes in order to see if any files changed. Not all tracked files changing will equate to "we need to re-publish a-c", but let's consider that an optimization, and deal with it in a follow-up.

grigoryk:
This is pretty much a copy-paste of what @rfk did for a-s, with minor changes in how this is triggered by the consuming projects.

Adds a python script which is responsible for checking if there are any local changes,
and publishing to mavenLocal if there are.

Before, consuming projects were responsible for this logic; now, they can simply call this command.

This also behaves differently from what fenix did prior.
Before, to determine if there are local changes, we'd run the build and see if any tasks were
actually performed (meaning, there were code changes that triggered a rebuild).
This way had its pros - it wouldn't consider changes to .gitignore, for example, as something that would
affect the build.

However, for the most common case, when there are no changes, this approach would still run through the build.
Doing so comes with a significant overhead of running through all of gradle's build phases, even if there isn't
actually anything to re-compile.

New approach optimizes for the common case. When there are no changes, we can now determine that almost instantly
by looking at an aggregate of git hash values for our project. And, by allowing consuming projects to call the python
script directly, we're skipping the gradle overhead as well.

Old approach was also potentially quite brittle, since it relied on parsing gradle's output.

The end result of this is a zero-cost auto-publication workflow.

---

### Pull Request checklist

### After merge


automation/publish_to_maven_local_if_modified.py
```diff
@@ -0,0 +1,94 @@
...
+    while ln:
+        with open(ln, "rb") as f:
+            contents_hash.update(f.read())
+        contents_hash.update(b"\x00")
+        ln = next(changes_lines)
+except StopIteration:
+    pass
+contents_hash.update(b"\x00")
+
+contents_hash = contents_hash.hexdigest()
```
rfk, line 75: Something that occurred to me after doing the a-s version: are the `.gitignore`d files whose contents we should also take into account here? I suspect `local.properties` is one, but there may be others.

grigoryk: 👍 took your changes from the a-s PR.

grigoryk:
`build-feature-intent` failure is unrelated to the PR (connection reset).

grigoryk:
bors r=csadilek

Issue #1705, opened 2019-01-11
Debug UI for seeing past crashes (like about:crashes)


pocmo:
On Fennec we have `about:crashes`. This works partially for GeckoView for native crashes - but that's more like a lucky coincidence than something we expect.

The only piece that actually knows about all crashes (native crashes vs. uncaught exceptions) is `lib-crash`. We could provide some UI piece (maybe a fragment) in the library that apps can decide to show (up to them what triggers it).

We would like to see a list of crashes. For native crashes we would like to link to the crash on Socorro. I think for Sentry we can't do that - but we could at least expose parts of the stacktrace so that we can find it on Sentry. Or maybe append a random UUID to the crash report sent to Sentry - we can then expose the same UUID in the UI.

EDIT: Sentry internally already assigns a generated UUID to the event and we can get that via the event builder. Pasting this UUID into the search query field of the web UI directly jumps to the crash report and generates a URL like : `[SENTRY SERVER]?query=[UUID]&limit=25&sort=date&statsPeriod=24h&shortIdLookup=1`

tarekziade:
I would also love to have the ability to grab them from the host that runs the test. (maybe via a directory?) 

The use case would to automate the collecting of crashes when we're doing tests on the CI in raptor and other places (we can't visit about: pages easily from webdriver)

tarekziade:
Also, if we could grab it from a chrome API that would be fantastic, we would just do a webdriver call and would avoid having to read/write on the android file system.

Dexterp37:
> The use case would to automate the collecting of crashes when we're doing tests on the CI in raptor and other places (we can't visit about: pages easily from webdriver).

Fly-by: if the use-case is to capture the number of times something crashes, we could also work for making CI send `crash` pings to our ingestion (or grab them locally on the CI). Perf has a similar use-case for perf metrics.

tarekziade:
> Fly-by: if the use-case is to capture the number of times something crashes, we could also work for making CI send `crash` pings to our ingestion (or grab them locally on the CI). Perf has a similar use-case for perf metrics.

For my use case it's to try to bring Fenix on par with other platforms where we push crash dumps as downloadable artifacts on jobs in Tascluster, so developers can investigate when something goes wrong, right after the test ended.

Dexterp37:
> 
> 
[...]

PR #6738, opened 2020-04-22
Issue #1705: Allow CrashReporterService to return a unique identifier for reported crashes.

pocmo:
This is one small piece of #1705. I am going to land some pieces independently to avoid ending up with a huge patch.

Here we are allowing a `CrashReporterService` to return an ID that we can later use to lookup that crash. For Socorro we return the crash ID that we get when uploading. We can later use that to create a URL pointing to this crash. For Sentry we return the UUID of the event. We can at least ask Sentry with this UUID for the actual crash URL.


pocmo:
bors r=rocketsroger

PR #6739, opened 2020-04-22
Issue #1705: Assign UUID to every crash object.

pocmo:
Assigning every crash a unique ID so that we can later reference this crash when saving some metadata about it to disk.

### DIFF
--- a/components/lib/crash/src/main/java/mozilla/components/lib/crash/Crash.kt
+++ b/components/lib/crash/src/main/java/mozilla/components/lib/crash/Crash.kt
@@ -7,6 +7,7 @@ package mozilla.components.lib.crash
 import android.content.Intent
 import android.os.Bundle
 import java.io.Serializable
+import java.util.UUID
 
 // Intent extra used to store crash data under when passing crashes in Intent objects
 private const val INTENT_CRASH = "mozilla.components.lib.crash.CRASH"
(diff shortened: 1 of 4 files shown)


components/lib/crash/src/main/java/mozilla/components/lib/crash/Crash.kt
```diff
@@ -81,11 +93,12 @@ sealed class Crash {
 
         companion object {
             internal fun fromBundle(bundle: Bundle) = NativeCodeCrash(
-                bundle.getString(INTENT_MINIDUMP_PATH, null),
-                bundle.getBoolean(INTENT_MINIDUMP_SUCCESS, false),
-                bundle.getString(INTENT_EXTRAS_PATH, null),
-                bundle.getBoolean(INTENT_FATAL, false),
-                bundle.getParcelableArrayList(INTENT_BREADCRUMBS) ?: arrayListOf()
+                uuid = bundle.getString(INTENT_UUID) as String,
```
pocmo, line 96: Here I'll need to add `?: UUID.randomUUID().toString()`. We are also processing crashes from GV via `CrashHandlerService` and they do not contain any UUID, so in those cases we need to generate one.

rocketsroger: You'll be adding this in a separate change?

pocmo: Just added it to this PR. :)

pocmo:
bors r=rocketsroger

PR #6746, opened 2020-04-22
Close #6421: Download try again should not crash when download does not exist

### LINKED ISSUE #6421: IllegalStateException: download must not be null
https://sentry.prod.mozaws.net/operations/fenix-fennec-beta/issues/7663623/

```
IllegalStateException: download must not be null
    at mozilla.components.feature.downloads.manager.FetchDownloadManager.tryAgain(FetchDownloadManager.kt:3)
    at mozilla.components.feature.downloads.DownloadsFeature.tryAgain(DownloadsFeature.kt:1)
    at org.mozilla.fenix.browser.BaseBrowserFragment$initializeUI$$inlined$also$lambda$7$1.invoke(BaseBrowserFragment.kt:3)
    at -$$LambdaGroup$js$vSvBtMcliXwjaE59xzbJRgV9aME.onClick(java-style lambda group:148)
    at android.view.View.performClick(View.java:5721)
...
(9 additional frame(s) were not displayed)

download must not be null
```

rocketsroger:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/manager/FetchDownloadManager.kt
+++ b/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/manager/FetchDownloadManager.kt
@@ -76,7 +76,7 @@ class FetchDownloadManager<T : AbstractFetchDownloadService>(
     }
 
     override fun tryAgain(downloadId: Long) {
-        val download = queuedDownloads[downloadId]
+        val download = queuedDownloads[downloadId] ?: return
 
         val intent = Intent(applicationContext, service.java)
         intent.putDownloadExtra(download)
--- a/components/feature/downloads/src/test/java/mozilla/components/feature/downloads/manager/FetchDownloadManagerTest.kt
+++ b/components/feature/downloads/src/test/java/mozilla/components/feature/downloads/manager/FetchDownloadManagerTest.kt
@@ -30,6 +30,7 @@ import org.junit.Test
 import org.junit.runner.RunWith
 import org.mockito.Mockito.verify
 import mozilla.components.feature.downloads.AbstractFetchDownloadService.DownloadJobStatus
+import org.junit.Assert.assertFalse
 import org.mockito.Mockito.times
 
 @RunWith(AndroidJUnit4::class)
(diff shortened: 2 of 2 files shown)


components/feature/downloads/src/main/java/mozilla/components/feature/downloads/manager/FetchDownloadManager.kt
```diff
@@ -76,7 +76,7 @@ class FetchDownloadManager<T : AbstractFetchDownloadService>(
     }
 
     override fun tryAgain(downloadId: Long) {
-        val download = queuedDownloads[downloadId]
+        val download = queuedDownloads[downloadId] ?: return
```
Amejia481, line 79: Are there any steps to reproduce this issue? I want to find the root cause, If we do that, are we just preventing the crash, but the notification is going to persist, without any behaviour? I noticed in [unregisterListeners](https://github.com/mozilla-mobile/android-components/blob/ba60c8df899b4e6fd32341fc6a2fc011b1009376/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/manager/FetchDownloadManager.kt#L91), we are clearing the `queuedDownloads`, should we also remove all the pending notifications too, as the `downloads` in the downloads notification are not going to be valid anymore?

rocketsroger: I didn't find any good way to reproduce it.  I can only hard code the failure locally in my testing.  You bring up a good point.  Since `unregisterListeners` clears the download queue we should also clear all download notifications.  I'll look more into this.  Thanks,

Amejia481: 👍 
We have to be careful if we clear notifications as they could be in a downloading state

rocketsroger: Sure, I'll confirm if clearing notification is doable.

rocketsroger: I think this requires a different fix.  I'm only fixing the symptoms here but not the cause.  The more I look the more issues I see with this download queue.  I'll find a different solution to fix this.  Thanks,

Amejia481: Don't worry, proceed with this patch. I'm working on a related issue https://github.com/mozilla-mobile/fenix/issues/9884 and I believe I found the source of the issue.  👍

rocketsroger:
bors r=Amejia481

PR #6755, opened 2020-04-23
Closes #6754: Intent.asForegroundServicePendingIntent: Set PendingIntent.FLAG_UPDATE_CURRENT by default.

### LINKED ISSUE #6754: Crash notification: Sometimes wrong crash is getting reported
Sometimes we seem to report a wrong (older) crash. At least that is what I am seeing when using the crash sample app.

This seems to be caused by `Intent.asForegroundServicePendingIntent` not setting `PendingIntent.FLAG_UPDATE_CURRENT` and therefore we may dispatch an older version of the pending intent.

pocmo:
While testing with the crash sample app I noticed that sometimes we'd report an older crash via the notification. We need to set `PendingIntent.FLAG_UPDATE_CURRENT` so that we will always dispatch the latest intent.


pocmo:
bors r=psymoon

PR #6769, opened 2020-04-24
Closes #6538: Add navigation state to browser state

### LINKED ISSUE #6538: BrowserState: Add canGoBack, canGoForward to ContentState
Let's add those to `ContentState` and make sure they are in sync.

csadilek:
Need this for my reader mode refactor :).

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
@@ -18,6 +18,8 @@ import mozilla.components.browser.state.action.ContentAction.AddFindResultAction
 import mozilla.components.browser.state.action.ContentAction.ClearFindResultsAction
 import mozilla.components.browser.state.action.ContentAction.ConsumeHitResultAction
 import mozilla.components.browser.state.action.ContentAction.FullScreenChangedAction
+import mozilla.components.browser.state.action.ContentAction.UpdateBackNavigationStateAction
+import mozilla.components.browser.state.action.ContentAction.UpdateForwardNavigationStateAction
 import mozilla.components.browser.state.action.ContentAction.RemoveThumbnailAction
 import mozilla.components.browser.state.action.ContentAction.UpdateHitResultAction
 import mozilla.components.browser.state.action.ContentAction.UpdateLoadingStateAction
@@ -227,6 +229,7 @@ class Session(
      */
     var canGoBack: Boolean by Delegates.observable(false) { _, old, new ->
         notifyObservers(old, new) { onNavigationStateChanged(this@Session, new, canGoForward) }
+        store?.syncDispatch(UpdateBackNavigationStateAction(id, canGoBack))
     }
 
     /**
@@ -234,6 +237,7 @@ class Session(
      */
     var canGoForward: Boolean by Delegates.observable(false) { _, old, new ->
         notifyObservers(old, new) { onNavigationStateChanged(this@Session, canGoBack, new) }
+        store?.syncDispatch(UpdateForwardNavigationStateAction(id, canGoForward))
     }
 
     /**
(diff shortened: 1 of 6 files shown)


components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
```diff
@@ -227,13 +229,15 @@ class Session(
...
         notifyObservers(old, new) { onNavigationStateChanged(this@Session, new, canGoForward) }
+        store?.syncDispatch(UpdateNavigationStateAction(id, NavigationState(canGoBack, canGoForward)))
     }
 
     /**
      * Navigation state, true if there's an history item to go forward to, otherwise false.
      */
     var canGoForward: Boolean by Delegates.observable(false) { _, old, new ->
         notifyObservers(old, new) { onNavigationStateChanged(this@Session, canGoBack, new) }
+        store?.syncDispatch(UpdateNavigationStateAction(id, NavigationState(canGoBack, canGoForward)))
```
pocmo, line 240: I wonder if always dispatching a combined `NavigationState` is the best solution when in the future we will remove it from `Session` and only have it on `SessionState`. I did the same but then looked at `EngineObserver` which gets a nullable `canGoBack` and `canGoForward` (one level lower Gecko also only invokes `onCanGoForward` or `onCanGoBack`).
Once we get rid of `Session` we will not have the combined state and only the `back` or `forward` change to dispatch. In my local version I then changed the action to have nullable fields and only dispatch what actually changed and only update that field in the reducer. But at this point we could also just have two actions. What do you think?

components/browser/state/src/main/java/mozilla/components/browser/state/state/NavigationState.kt
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * A value type holding navigation state information for a Session.
+ *
+ * @property canGoBack whether or not there's an history item to go back to.
+ * @property canGoForward whether or not there's an history item to go forward to.
+ */
+data class NavigationState(
+    val canGoBack: Boolean = false,
+    val canGoForward: Boolean = false
+)
```
pocmo, line 16: If we consider only dispatching the actual change (forward/back, see above)  then we could also consider flattening the state and adding those fields to `ContentState` directly?

components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
csadilek: Heh, I actually started with that and then changed it to using a single `NavigationState`. This seemed cleaner to me, as it's a bit easier to read than adding two more boolean states to the `ContentState` directly.

> Once we get rid of Session we will not have the combined state and only the back or forward change to dispatch. 

Yeah, that's a good point. We're using independent / separate actions everywhere. So, I will change this to separate actions as well.

@pocmo btw, I noticed that restored sessions lose their history i.e. navigation doesn't work once a session was restored. I can't seem to remember if this worked before? So, open a tab, follow a few links, save, close and open the app. We should be able to go/navigate back but can't. Do we have an issue on file (GV or A-C)? Kinda confused how we missed this, probably just me :).

pocmo: > @pocmo btw, I noticed that restored sessions lose their history i.e. navigation doesn't work once a session was restored. I can't seem to remember if this worked before? 

I am pretty sure this worked and that I have been doing this a bunch of times to test something not that long ago. If I remember correctly then the only way this could have worked is when GV invokes the callbacks on restore, I think? In theory this sounds like a simple test we could add to the on-device tests of the `browser-engine-gecko-nightly` component - and then maybe use that to bisect if that worked earlier?

csadilek: @pocmo OK, I filed an issue: https://github.com/mozilla-mobile/android-components/issues/6783 - It also doesn't work in Fenix release. So, the regression isn't too recent.

csadilek:
@pocmo ok, updated.

PR #6777, opened 2020-04-27
Issue #1705: Create separate interface for crash telemetry services.

pocmo:
In #1705 I am going to add more methods/properties to the `CrashReporterService` interface. Those will not make any sense for a telemetry service. So here I am going to create a separate interface for telemetry service and make glean and the crash reporter class use that instead.


pocmo:
Looks like I forgot to change the tests using the interfaces. Updating PR!

pocmo:
bors r=rocketsroger

PR #6779, opened 2020-04-27
MozillaSocorroService: Use internal logger instance.

pocmo:
I couldn't find the log statements until I noticed that we didn't use the internal `logger` instance and instead called static methods on `Logger` directly. :)


pocmo:
bors r=rocketsroger

Issue #5680, opened 2020-01-22
Deprecate/Remove URLStringUtils.isURLLikeStrict


csadilek:
This function is no longer being used so we should deprecate and remove it so it doesn't get re-introduced by accident.

It uses a regex that takes > 500ms to compile. More details:
https://github.com/mozilla-mobile/android-components/issues/5376
https://github.com/mozilla-mobile/android-components/issues/5594

mcomella:
Deprecation completed. We'll need to follow up with deprecation in a https://github.com/mozilla-mobile/android-components/issues/6815

PR #6784, opened 2020-04-27
Issue #5680: Deprecate URLStringUtils.isURLLikeStrict.

mcomella:
I want to remove this method and the expensive regex it relies on
because:

1. unused code should generally be removed because it's more likely to
break
2. if the APIs exist and fulfill someone's needs, they'll use them.
However, if they had to write their own, they may be able to come up
with a simpler, alternative implementation that would better fit their
needs and, in this case, be much more performant.
3. On a mobile device with limited battery, we shouldn't be
unnecessarily compiling a regex for a second or more if it's not
necessary, especially if a simpler implementation is viable
4. This hard-coded list of TLDs (from 2015) is already out-of-date and
liable to cause problems: more TLDs are getting all the time so, if
keeping a hard-coded list of TLDs is actually a need, we should find an
alternative implementation that lends itself to getting frequent updates
like our public suffix list

---

### Pull Request checklist

### After merge

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -15,6 +15,12 @@ permalink: /changelog/
 * **feature-top-sites**
   * Added `isDefault` to the top site entity, which allows application to specify a default top site that is added by the application. This is called through `TopSiteStorage.addTopSite`.
 
+* **support-utils**
+  * `URLStringUtils.isURLLikeStrict` is now deprecated due to performance issues. Consider using the less strict `isURLLike` instead or creating a new method using `:lib-publicsuffixlist`.
+
+* **support-ktx**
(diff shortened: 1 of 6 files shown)


docs/changelog.md
```diff
@@ -12,6 +12,12 @@ permalink: /changelog/
 * [Gecko](https://github.com/mozilla-mobile/android-components/blob/master/buildSrc/src/main/java/Gecko.kt)
 * [Configuration](https://github.com/mozilla-mobile/android-components/blob/master/buildSrc/src/main/java/Config.kt)
 
+* **support-utils**
+  * `URLStringUtils.isURLLikeStrict` is now deprecated due to performance issues. Consider using the less strict `isURLLike` instead or creating a new method using `:lib-publicsuffixlist`.
+  
+* **support-test**
```
csadilek, line 18: This should be `support-ktx`.

mcomella:
@csadilek How can I land this? The required status checks seem to be hanging.

PR #6785, opened 2020-04-27
Fix component name in changelog

jonalmeida:
---

Fixing a small nit that we didn't catch during the code review. 🙂


jonalmeida:
bors retry

PR #6787, opened 2020-04-27
Closes #6783: Engine session state is not restored

### LINKED ISSUE #6783: History navigation does not work for restored sessions
STR:

- Open tab.
- Follow a few links.
- Send app to background and foreground or wait to make sure session is persisted.
- Restart the app.

Expected:
- Back navigation is possible in tab.

Actual:
- Back navigation doesn't work / closes the app.

csadilek:
Regression is fairly recent and my mistake. 

Introduced by this change: 
https://github.com/mozilla-mobile/android-components/commit/91bc92fc5ef889e53c77981b81a54e6edb6c22b5#diff-d386a820f3dab49f970fb183c2a57201

We now use the wrong key to look up the engine session JSON object. We use `Keys.SESSION_KEY` instead of `Keys.ENGINE_SESSION_KEY`. 🤦 

I've also add a unit test here so this doesn't happen again.


pocmo:
bors retry

PR #6801, opened 2020-04-29
Closes #6553: Use SystemClock.elapsedRealtimeNanos for logElapsedTime

### LINKED ISSUE #6553: logElapsedTime should use SystemClock.elapsedRealtimeNanos instead of System.nanoTime()
`logElapsedTime` is using `System.nanoTime`: https://github.com/mozilla-mobile/android-components/blob/da1b07d69d3eee83fc0f35e26ffb56acd96d6e3c/components/support/utils/src/main/java/mozilla/components/support/utils/Performance.kt#L21

However, System.nanoTime may not count during deep sleep:
- see https://stackoverflow.com/a/11092021 and comments
- it doesn't comment about deep sleep in the documentation
- it's a java API so is unlikely to take into account Android specific implementation details

`SystemClock` is an Android API that's specifically documented to take into account deep slepe so we should use it instead: https://developer.android.com/reference/android/os/SystemClock

Note: when consulting the Glean team in Fenix, we decided to use Glean's built-in timing facilities to do any type of timing capture and logging specifically to avoid maintaining two implementations of these types of functions. I'm not sure that this will apply to all use cases nor that it applies to a-c but I thought I'd mention it as our ideal going forward.

CC @grigoryk

grigoryk:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/support/utils/src/main/java/mozilla/components/support/utils/Performance.kt
+++ b/components/support/utils/src/main/java/mozilla/components/support/utils/Performance.kt
@@ -4,6 +4,7 @@
 
 package mozilla.components.support.utils
 
+import android.os.SystemClock
 import mozilla.components.support.base.log.logger.Logger
 
 /**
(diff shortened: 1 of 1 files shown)


grigoryk:
bors r=csadilek

PR #6802, opened 2020-04-29
Closes #4992: Emit provider duration facts from BrowserAwesomeBar

### LINKED ISSUE #4992: Instrument awesomebar suggestion providers
In a conversation with @rfk we've discussed importance of instrumenting some of the internals behind high-level product experiences.
There are many levels to this conversation - at a lower level, app-services is starting to work on defining and shipping quality metrics for its libraries.
At a higher level, we can take awesomebar as a good example - we have product features which depend on performance of different parts of the product, and keeping track of how these sub-parts are performing is valuable for making sure we deliver a good & consistent experience.

As a starting point, we can start by instrumenting performance of querying our configured awesomebar providers, and ideally how that correlates to the amount of data providers need to churn through (e.g. number of recorded history items in moz_places, number of open tabs, etc).

For historical context - in Fennec we had histograms tracking performance distribution of various hot-path queries.

grigoryk:
Some things to consider and trade-offs:

When recording durations of various providers into a glean metric, we have a bit of a hurdle.
One approach would be to encapsulate all of the awesomebar-related perf telemetry entirely within the
awesomebar a-c component. But, set of providers isn't entirely known to us at the a-c level.
We know what providers we have defined, but we don't know what providers applications will provide themselves.
Also, glean doesn't have a metric of type (string->timespan), which would allow us to work-around this.
So, we can't define ping/metrics in the a-c component. This means that they need to be defined elsewhere, while
the measurement happens within the component. An established pattern for that in the codebase is emitting "facts",
which is what this patch does.

We delegate to the consuming application to then actually do something with these facts - e.g. map providers
to concrete glean metric definitions.

Another consideration is if we should try to group timings related to a single query. That's hard to do reliably,
and will introduce additional complexity into what's otherwise a super simple setup. Source of that complexity
is that we actively try to cancel query jobs as user is typing; our providers could take an arbitrary time to resolve,
and so grouping becomes difficult. However, we care about improving how long each individual provider takes,
since that's a good proxy for "how responsive is this UI?". Simply recording each individual timing we see
should be enough for that.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/awesomebar/src/main/java/mozilla/components/browser/awesomebar/BrowserAwesomeBar.kt
+++ b/components/browser/awesomebar/src/main/java/mozilla/components/browser/awesomebar/BrowserAwesomeBar.kt
@@ -5,6 +5,7 @@
 package mozilla.components.browser.awesomebar
 
 import android.content.Context
+import android.os.SystemClock
 import android.util.AttributeSet
 import android.util.LruCache
 import androidx.annotation.VisibleForTesting
@@ -17,6 +18,7 @@ import kotlinx.coroutines.Job
 import kotlinx.coroutines.asCoroutineDispatcher
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
+import mozilla.components.browser.awesomebar.facts.emitProviderQueryTimingFact
 import mozilla.components.browser.awesomebar.layout.SuggestionLayout
 import mozilla.components.browser.awesomebar.transform.SuggestionTransformer
 import mozilla.components.concept.awesomebar.AwesomeBar
@@ -30,7 +32,7 @@ internal const val INITIAL_NUMBER_OF_PROVIDERS = 5
 /**
  * A customizable [AwesomeBar] implementation.
  */
-@Suppress("TooManyFunctions")
+@Suppress("TooManyFunctions", "LargeClass")
 class BrowserAwesomeBar @JvmOverloads constructor(
     context: Context,
     attrs: AttributeSet? = null,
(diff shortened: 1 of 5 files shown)


components/browser/awesomebar/src/main/java/mozilla/components/browser/awesomebar/BrowserAwesomeBar.kt
```diff
@@ -154,7 +156,24 @@ class BrowserAwesomeBar @JvmOverloads constructor(
 
             providers.forEach { provider ->
                 launch {
+                    // At this point, we have a timing value for a provider.
+                    // We have a choice here - we can try grouping different timings together for a
+                    // single user input value, or we can just treat them as entirely independent values.
+                    // These timings are correlated with each other in a sense that that act on the same
```
mcomella, line 162: nit: each other in a sense that they

components/browser/awesomebar/src/main/java/mozilla/components/browser/awesomebar/BrowserAwesomeBar.kt
```diff
@@ -154,7 +156,24 @@ class BrowserAwesomeBar @JvmOverloads constructor(
...
                 launch {
+                    // At this point, we have a timing value for a provider.
+                    // We have a choice here - we can try grouping different timings together for a
+                    // single user input value, or we can just treat them as entirely independent values.
+                    // These timings are correlated with each other in a sense that they act on the same
+                    // inputs, and are executed at the same time. However, our goal here is to track performance
+                    // of the providers. Each provider acts independently from another; recording their performance
+                    // at an individual level will allow us to track that performance over time.
+                    // Tracked value will be reflected both in perceived user experience (how quickly results from
+                    // a provider show up),
```
csadilek, line 167: nit: looks like part of the sentence got lost: both in perceived user experience.., and...

grigoryk:
bors r=csadilek

PR #6819, opened 2020-04-29
Bump app services to v0.58.2

eoger:
r? @grigoryk


grigoryk:
I've tested this in local builds, and it seems to resolve the problem we've been seeing with 5/6. I've tried this on emulators running API 21/23/29, and on a device running API 21.

bors r+

PR #6828, opened 2020-04-30
Issue #1705 - lib-crash: Add database storage for crashes.

pocmo:
This patch adds the database classes to store information about crashes on disk. In this state it is not being used yet. Landing this separately to continue working on the UI and the integration. I'm not 100% happy with that part yet and therefore want to continue iterating. However the database piece seems to be stable enough and probably won't change.

### DIFF
--- a/components/lib/crash/src/main/java/mozilla/components/lib/crash/db/ReportEntity.kt
+++ b/components/lib/crash/src/main/java/mozilla/components/lib/crash/db/ReportEntity.kt
@@ -1,0 +1,43 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.lib.crash.db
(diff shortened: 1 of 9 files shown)


components/lib/crash/src/main/java/mozilla/components/lib/crash/db/ReportEntity.kt
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * Database internal primary key of the entry.
+     */
+    @PrimaryKey(autoGenerate = true)
+    @ColumnInfo(name = "id")
+    var id: Long? = null,
+
+    /**
+     * UUID of the crash that was reported.
+     */
```
rocketsroger, line 28: nit: should we name this column as well to follow the format of the other columns?

pocmo: Yeah, makes sense. I'll update the PR.

pocmo:
bors r=rocketsroger

pocmo:
bors retry

PR #6843, opened 2020-05-01
Closes #2985: Launch reader view in extension page

### LINKED ISSUE #2985: Open Reader View in an extension page
When activating reader view, we are currently transforming the existing document in our web extension content script. 

This leaves two problems:

- We don't get a history entry for reader view (although we could use pushState and other approaches see https://github.com/mozilla-mobile/android-components/issues/2879 for details)

- We effectively can't isolate reader view from JavaScript that was loaded with the original page e.g. a setTimeout call that shows a popup / ad etc.

A solution to this would be to open reader view in an [extension page](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/user_interface/Extension_pages). So, our content script would send a message to a background script, which would call `browser.tabs.update()`. We can then transform the DOM using a script loaded by the extension page. We can pass the original document along in the message, and follow Fennec's approach and download the document again (using XHR) as a fallback (e.g. when the extension page is navigated to directly, see also https://github.com/mozilla-mobile/android-components/issues/2980).

Another big advantage of this approach is that we would no longer need to inject all scripts e.g.`Readability.js` into every single page. We only need to inject the functionality that checks whether or not the page is readerable. The remaining scripts only need to be loaded by the extension page once Reader View is active.

In order to make this work, we need support for `browser.tabs.update()` for pages packaged in the web extension. This currently doesn't seem to work in GV: https://bugzilla.mozilla.org/show_bug.cgi?id=1550877

**Status - Remaining GV issues we're blocked on / still need fixed and uplifted to beta:**

* [x] ~~https://bugzilla.mozilla.org/show_bug.cgi?id=1554302 (no history entry for extension pages)~~
* [x] ~~https://bugzilla.mozilla.org/show_bug.cgi?id=1554302 (`browser.tabs.update` only works once per session)~~
* [x] ~~https://bugzilla.mozilla.org/show_bug.cgi?id=1554260 (WebExtension Page messages are sent to the Runtime instead of GeckoSessions)~~
* [x] ~~https://bugzilla.mozilla.org/show_bug.cgi?id=1553371 (WebExtension pages don't have access to privileged APIs)~~
* [x] ~~https://bugzilla.mozilla.org/show_bug.cgi?id=1551377 (WebExtension `tabId` should return a unique ID for each tab)~~

Other related GV issues, but not blockers:
* [x] https://bugzilla.mozilla.org/show_bug.cgi?id=1634423 (capturePixels returns empty bitmap for extension pages)
* [x] https://bugzilla.mozilla.org/show_bug.cgi?id=1634428 (Intermediate loads of about:blank when navigating to/from extension page)
* [ ] https://bugzilla.mozilla.org/show_bug.cgi?id=1550144 (Provide API to add custom history entry (e.g. about:reader?url=))
* [x] ~~https://bugzilla.mozilla.org/show_bug.cgi?id=1560048 (Expose internal web extension ID)~~
* [x] ~~https://bugzilla.mozilla.org/show_bug.cgi?id=1551278 (Content messaging delegate receives messages from all extensions)~~

csadilek:
Design is as described in #2985. Instead of transforming the current DOM, we now open reader view in an extension page. We inject only a small script (instead of the whole logic) into every page to check whether or not the page is readerable. If the reader view is activated, we send a message to a background script to open the reader page which contains the rest of the logic and its dependencies.

This still requires a few workarounds, which I documented with links to the corresponding issues. However, this is now finally in a functioning and landable state. It solves many problems users reported in Fenix e.g. reader stuck in wrong zoom level, reader refreshing or reloading the original page, overlays/popups appearing etc. Also, it is now possible to navigate back and forth into reader view as we have a history entry (the extension page).

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
@@ -25,6 +25,7 @@ import mozilla.components.concept.engine.EngineSessionState
 import mozilla.components.support.base.log.logger.Logger
 import mozilla.components.support.base.memory.MemoryConsumer
 import mozilla.components.support.base.observer.Observable
+import mozilla.components.support.ktx.kotlin.isExtensionUrl
 import java.lang.IllegalArgumentException
 
 /**
@@ -62,7 +63,14 @@ class SessionManager(
                 this.engineObserver = EngineObserver(session, store).also { observer ->
                     engineSession.register(observer)
                     if (!sessionRestored && !skipLoading) {
-                        engineSession.loadUrl(session.url, parentEngineSession)
+                        if (session.url.isExtensionUrl()) {
+                            // The parent tab/session is used as a referrer which is not accurate
+                            // for extension pages. The extension page is not loaded by the parent
+                            // tab, but opened by an extension e.g. via browser.tabs.update.
+                            engineSession.loadUrl(session.url)
+                        } else {
+                            engineSession.loadUrl(session.url, parentEngineSession)
+                        }
                     }
                 }
             }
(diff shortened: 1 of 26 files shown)


csadilek:
Tests added.

csadilek:
Did a round of testing Fenix as well. Looking good. No changes required there.

csadilek:
@jonalmeida pushed an update now and addressed your comments.

csadilek:
bors r=jonalmeida

Issue #6757, opened 2020-04-23
Expose GeckoSession CompositorController clearColor


ekager:
This will allow Fenix to set the default clear color on a session. https://mozilla.github.io/geckoview/javadoc/mozilla-central/org/mozilla/geckoview/CompositorController.html
https://github.com/mozilla-mobile/fenix/issues/6313

ekager:
I filed this in AC because this is where the onFirstContentfulPaint callbacks live and I figure at the minimum we would need to expose those to Fenix in order to show/hide some dummy background view.

pocmo:
> at the minimum we would need to expose those to Fenix in order to show/hide some dummy background view.

Would it also be an option to hide the geckoview until we get `onFirstContentfulPaint()` and just show the window background in the meantime?

ekager:
That sounds way less complicated if that works :slightly_smiling_face:

pocmo:
> That sounds way less complicated if that works 🙂

I saw your PR and wonder if you tried this and if that works? If so then we could consider maybe adding this to `GeckoEngineView`. It already registers an `EngineSession.Observer` and could use that to hide/show the internal `GeckoView`. With that we wouldn't need to add anything to `Session(State)` and the app wouldn't need to do anything manually. :)

PR #6844, opened 2020-05-01
For #6757 - Adds firstContentfulPaint callback to SessionState

ekager:
---

Testing out @pocmo ['s idea](https://github.com/mozilla-mobile/android-components/issues/6757#issuecomment-619942875) from #6757 via a session callback and var.
See accompanying Fenix PR for how I've integrated this: https://github.com/mozilla-mobile/fenix/pull/10358

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineView.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineView.kt
@@ -7,6 +7,7 @@ package mozilla.components.browser.engine.gecko
 import android.content.Context
 import android.graphics.Bitmap
 import android.util.AttributeSet
+import android.view.View
 import android.widget.FrameLayout
 import androidx.annotation.VisibleForTesting
 import androidx.core.view.ViewCompat
@@ -65,6 +66,10 @@ class GeckoEngineView @JvmOverloads constructor(
 
         override fun onProcessKilled() {
             rebind()
+        }
+
+        override fun onFirstContentfulPaint() {
+            visibility = View.VISIBLE
         }
 
         override fun onAppPermissionRequest(permissionRequest: PermissionRequest) = Unit
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -600,6 +600,7 @@ class GeckoEngineSession(
 
         override fun onFirstContentfulPaint(session: GeckoSession) {
             firstContentfulPaint = true
+            notifyObservers { onFirstContentfulPaint() }
         }
 
         override fun onContextMenu(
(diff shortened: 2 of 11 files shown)


components/browser/session/src/main/java/mozilla/components/browser/session/Session.kt
```diff
@@ -91,6 +91,10 @@ class Session(
             triggeredByRedirect: Boolean,
             triggeredByWebContent: Boolean
         ) = Unit
+        /**
+         * Indicates this session has had it's first engine contentful paint of page content.
+         */
+        fun onFirstContentfulPaint(session: Session) = Unit
```
pocmo, line 97: Is there any chance we can move this to `SessionState` and not add it to `Session`? Since we are moving away from `Session` I would love to not add anything new here if possible. :)

ekager: Yep! We aren't using `TabSessionState` very much yet in Fenix I think? But something like this looks like it would work the same? 
https://github.com/mozilla-mobile/fenix/pull/10358/files#diff-bf0b57d467da0e3d5ef96c23fc828376R177

pocmo: Yeah, we could either subscribe to `BrowserStore` or lookup the `TabSessionState` if it is a one time thing.

pocmo: I just realized that I left my other comment in the issue yesterday and not here on the PR: https://github.com/mozilla-mobile/android-components/issues/6757#issuecomment-623379531 - If that is possible then maybe we can simplify it even more? 🤔

ekager: I'm pretty sure Fenix still needs to be able to access whether the first contentful paint has happened for the case of navigate to browser -> click URL bar to enter SearchFragment -> Press Back. We are setting the EngineView to GONE while leaving the BrowserFragment for visual and performance issues, and need to set it to VISIBLE again (and we won't get the onFirstContentfulPaint callback for an existing session).

pocmo: Ah, yeah, that makes sense - we need to keep that state around. I still think we could perform the showing/hiding inside `GeckoEngineView` and show/hide the internal `GeckoView` based on whether the paint has happened or when it happens. Do you agree? (Of course this does not need to happen here in the PR).

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineView.kt
```diff
@@ -67,6 +68,10 @@ class GeckoEngineView @JvmOverloads constructor(
             rebind()
         }
 
+        override fun onFirstContentfulPaint() {
+            visibility = View.VISIBLE
+        }
```
pocmo, line 73: When I saw that I was wondering whether we could also set it to `VISIBLE` or `GONE` in `render()` based on whether the `Session` has already had its first contentful paint or not. But then I realized that we are passing `EngineSession` and not `Session`. 🤔

ekager:
bors r=pocmo

Issue #6825, opened 2020-04-30
Implement a custom tooltip for browser TwoStateButton specific for SDK 26


mcarare:
Original Fenix issue: https://github.com/mozilla-mobile/fenix/issues/9387.

The crash is specific to devices with  SDK version == 26.
In `TooltopCompat.java`:   
    
```
        if (Build.VERSION.SDK_INT >= 26) {
            view.setTooltipText(tooltipText);
        } else {
            TooltipCompatHandler.setTooltipText(view, tooltipText);
        }
```

As `TooltipCompatHandler` is package-private in `androidx.appcompat.widget`, we should implement a custom tooltip for SDK 26.

Note: Tested different implementations, and this also happens in sample-browser, and different mock apps (completely independent of AC) when tooltip is added for a view in a `PopupWindow`

mcarare:
Closing as per comment in https://github.com/mozilla-mobile/fenix/issues/9387#issuecomment-625812563.

PR #6863, opened 2020-05-04
For #6825: Implement a custom tooltip for SDK 26.

mcarare:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/menu/src/main/java/mozilla/components/browser/menu/item/CustomTooltip.kt
+++ b/components/browser/menu/src/main/java/mozilla/components/browser/menu/item/CustomTooltip.kt
@@ -1,0 +1,153 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.menu.item
+
+import android.annotation.SuppressLint
+import android.text.TextUtils
+import android.view.Gravity
+import android.view.LayoutInflater
+import android.view.View
+import android.view.View.OnLongClickListener
+import android.view.WindowManager
+import android.widget.LinearLayout
+import android.widget.PopupWindow
+import android.widget.TextView
+import androidx.core.view.ViewCompat
+import androidx.core.widget.PopupWindowCompat
+import mozilla.components.browser.menu.R
+
+/**
+ * A tooltip shown on long click on an anchor view.
+ * There can be only one tooltip shown at a given moment.
+ */
+internal class CustomTooltip private constructor(
+    private val anchor: View,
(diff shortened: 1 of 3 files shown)


components/browser/menu/src/main/java/mozilla/components/browser/menu/item/CustomTooltip.kt
```diff
@@ -0,0 +1,152 @@
...
+    override fun onViewDetachedFromWindow(v: View) {
+        hide()
+        anchor.removeCallbacks(hideTooltipRunnable)
+    }
+
+    companion object {
+        private const val LONG_CLICK_HIDE_TIMEOUT_MS: Long = 2500
+        private const val TOOLTIP_EXTRA_VERTICAL_OFFSET_DP = 8
+        // The tooltip currently being shown
+        private var activeTooltip: CustomTooltip? = null
```
Amejia481, line 125: Let's address this check
```
[task 2020-05-04T16:33:07.856Z]   Errors found:
[task 2020-05-04T16:33:07.856Z]   
[task 2020-05-04T16:33:07.856Z]   /builds/worker/checkouts/src/components/browser/menu/src/main/java/mozilla/components/browser/menu/item/CustomTooltip.kt:124: Error: Do not place Android context classes in static fields (static reference to CustomTooltip which has field anchor pointing to View); this is a memory leak [StaticFieldLeak]
[task 2020-05-04T16:33:07.856Z]           // The tooltip currently being shown
[task 2020-05-04T16:33:07.856Z]           ^
[task 2020-05-04T16:33:07.856Z]   
[task 2020-05-04T16:33:07.856Z] 
```

mcarare: @Amejia481 I cannot replicate the exact behavior of a tooltip without using using a static reference to the _active_ (currently visible) tooltip. 

The options are either to suppress this warning or slightly change the behavior to allow for the tooltip to disappear when you click outside it (a normal tooltip does not disappear until the timeout has passed, another tooltip is triggered to be shown or the anchor view is detached from window). It's not a big difference IMO. What do you think would be the best option?

Amejia481: Got it, we can suppress it, but let's make sure we are cleaning `activeTooltip` after is not needed any more to avoid memory leaks 👍

mcarare: Thanks for the review! I added the suppress annotation and explanation in comment. I have investigated with both Leak Canary and A-S profiler and they showed no memory leaks.

Amejia481: Awesome thank you for the great work!

PR #6865, opened 2020-05-04
Close #6535: Add breadcrumbs to CrashReporting interface

### LINKED ISSUE #6535: Add option for adding bread crumbs to CrashReporting interface
For debugging some crashes it would be very helpful if components could add breadcrumbs too.

rocketsroger:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/lib/crash/src/main/java/mozilla/components/lib/crash/service/SentryService.kt
+++ b/components/lib/crash/src/main/java/mozilla/components/lib/crash/service/SentryService.kt
@@ -17,8 +17,8 @@ import io.sentry.event.Event
 import io.sentry.event.EventBuilder
 import io.sentry.event.interfaces.ExceptionInterface
 import mozilla.components.Build
-import mozilla.components.lib.crash.Breadcrumb as crashBreadCrumbs
 import mozilla.components.lib.crash.Crash
+import mozilla.components.support.base.crash.Breadcrumb as CrashBreadcrumb
 
 /**
  * A [CrashReporterService] implementation that uploads crash reports to a Sentry server.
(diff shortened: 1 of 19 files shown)


components/lib/crash/src/main/java/mozilla/components/lib/crash/service/SentryService.kt
```diff
@@ -17,8 +17,8 @@ import io.sentry.event.Event
 import io.sentry.event.EventBuilder
 import io.sentry.event.interfaces.ExceptionInterface
 import mozilla.components.Build
-import mozilla.components.lib.crash.Breadcrumb as crashBreadCrumbs
 import mozilla.components.lib.crash.Crash
+import mozilla.components.support.base.crash.Breadcrumb as crashBreadCrumb
```
pocmo, line 21: nit: This has been the case before, but I just noticed it here... A class name that starts with a lower letter is confusing to read in code imo. Can we just name it `CrashBreadCrumb`? :) Or since the original class name doesn't use that camel-casing maybe `CrashBreadcrumb`?

docs/changelog.md
```diff
@@ -24,7 +24,13 @@ permalink: /changelog/
 
 * **feature-push**
   * Simplified error handling and reduced non-fatal exception reporting.
+ 
+* **lib-crash**
+  * ⚠️ **This is a breaking change**: breadcrumbs now have `support-base` dependency.
```
pocmo, line 29: The wording on this changelog entry is not super clear, I think. Just from reading this I wouldn't be sure what that means. 🤔 
I also wonder if we should make `support-base` an `api` dependency in `build.gradle` so that a consumer of `lib-crash` does not have to import `support-base` manually to get access to breadcrumbs. `support-base` is something that helps us internally but an end-user should not have to bother with it.

components/lib/crash/src/main/java/mozilla/components/lib/crash/service/SentryService.kt
rocketsroger: Agreed, I'll fix it.  Thanks,

docs/changelog.md
rocketsroger: Ok, I'll remove this entry.

rocketsroger:
bors r=pocmo

PR #6875, opened 2020-05-05
Remove service-glean's hard-dependency on httpurlconnection

### LINKED ISSUE #6660: glean/experiments: Hard dependency on lib-httpurlconnection
I was investigating why we still see (I think unused) fragments of `lib-httpurlconnection` in Fenix APKs. I noticed that `service-glean` as well as `service-experiments` have a hard (non-test) dependency on `lib-httpurlconnection`.

* https://github.com/mozilla-mobile/android-components/blob/7b25cd702dcfc5e22be436d6c5c39fb8249e5a9c/components/service/glean/build.gradle#L56

* https://github.com/mozilla-mobile/android-components/blob/7b25cd702dcfc5e22be436d6c5c39fb8249e5a9c/components/service/experiments/build.gradle#L33

My assumption is that we can get rid of that dependency and instead always enforce passing in a `lib-fetch` implementation.

badboy:
Remove service-glean's hard-dependency on httpurlconnection

We still test with it, so we keep it around.

*BREAKING CHANGE*:

Users will need to supply a configuration with a ping uploader to Glean
now.
A default lib-fetch-httpurlconnection implementation can be used like
this:

```kotlin
import mozilla.components.lib.fetch.httpurlconnection.HttpURLConnectionClient;
import mozilla.components.service.glean.config.Configuration;
import mozilla.components.service.glean.net.ConceptFetchHttpUploader;

val config = Configuration(httpClient = ConceptFetchHttpUploader(lazy { HttpURLConnectionClient() }))
Glean.initialize(context, true, config)
```

For Java this becomes:

```java
import mozilla.components.lib.fetch.httpurlconnection.HttpURLConnectionClient;
import mozilla.components.service.glean.config.Configuration;
import mozilla.components.service.glean.net.ConceptFetchHttpUploader;

ConceptFetchHttpUploader httpClient = ConceptFetchHttpUploader.fromClient(new HttpURLConnectionClient());
Configuration config = new Configuration(httpClient);
Glean.INSTANCE.initialize(context, true, config);
```

Fixes #6660 (at least partly)

I'm not familiar enough with service-experiments, so @travis79 would need to take a look.

cc @pocmo for visibility.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
@@ -4,26 +4,27 @@
 
 package mozilla.components.service.glean.config
 
-import mozilla.components.lib.fetch.httpurlconnection.HttpURLConnectionClient
-import mozilla.components.service.glean.net.ConceptFetchHttpUploader
 import mozilla.telemetry.glean.net.PingUploader
 import mozilla.telemetry.glean.config.Configuration as GleanCoreConfiguration
 
 /**
  * The Configuration class describes how to configure the Glean.
  *
- * @property serverEndpoint the server pings are sent to. Please note that this is
- *           is only meant to be changed for tests.
- * @property channel the release channel the application is on, if known. This will be
- *           sent along with all the pings, in the `client_info` section.
- * @property maxEvents the number of events to store before the events ping is sent
  * @property httpClient The HTTP client implementation to use for uploading pings.
+ *                      If you don't provide your own networking stack with an HTTP client to use,
+ *                      you can fall back to a simple implementation on top of `java.net` using
+ *                      `ConceptFetchHttpUploader(lazy { HttpURLConnectionClient() as Client })`
+ * @property serverEndpoint (optional) the server pings are sent to. Please note that this is
+ *           is only meant to be changed for tests.
+ * @property channel (optional )the release channel the application is on, if known. This will be
+ *           sent along with all the pings, in the `client_info` section.
+ * @property maxEvents (optional) the number of events to store before the events ping is sent
  */
 data class Configuration @JvmOverloads constructor (
+    val httpClient: PingUploader,
     val serverEndpoint: String = DEFAULT_TELEMETRY_ENDPOINT,
     val channel: String? = null,
-    val maxEvents: Int? = null,
-    val httpClient: PingUploader = ConceptFetchHttpUploader(lazy { HttpURLConnectionClient() })
+    val maxEvents: Int? = null
 ) {
     // The following is required to support calling our API from Java.
     companion object {
--- a/components/service/glean/src/main/java/mozilla/components/service/glean/net/ConceptFetchHttpUploader.kt
+++ b/components/service/glean/src/main/java/mozilla/components/service/glean/net/ConceptFetchHttpUploader.kt
@@ -14,9 +14,11 @@ import mozilla.components.concept.fetch.isClientError
 import mozilla.components.concept.fetch.isSuccess
 import mozilla.components.support.base.log.logger.Logger
 import mozilla.telemetry.glean.net.HeadersList
-import mozilla.telemetry.glean.net.PingUploader
+import mozilla.telemetry.glean.net.PingUploader as CorePingUploader
 import java.io.IOException
 import java.util.concurrent.TimeUnit
+
+typealias PingUploader = CorePingUploader
 
 /**
  * A simple ping Uploader, which implements a "send once" policy, never
@@ -33,6 +35,16 @@ class ConceptFetchHttpUploader(
         const val DEFAULT_CONNECTION_TIMEOUT = 10000L
         // The timeout, in milliseconds, to use when reading from the server.
         const val DEFAULT_READ_TIMEOUT = 30000L
+
+        /**
+         * Export a constructor that is usable from Java.
+         *
+         * This looses the lazyness of creating the `client`.
+         */
+        @JvmStatic
+        fun fromClient(client: Client): ConceptFetchHttpUploader {
+            return ConceptFetchHttpUploader(lazy { client })
+        }
     }
 
     /**
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -12,6 +12,13 @@ permalink: /changelog/
 * [Gecko](https://github.com/mozilla-mobile/android-components/blob/master/buildSrc/src/main/java/Gecko.kt)
 * [Configuration](https://github.com/mozilla-mobile/android-components/blob/master/buildSrc/src/main/java/Config.kt)
 
+* **service-glean**
+  * ⚠️ **This is a breaking change**: Glean's configuration now requires explicitly setting an http client. Users need to pass one at construction.
+    A default `lib-fetch-httpurlconnection` implementation is available.
+    Add it to the configuration object like this:
+    `val config = Configuration(httpClient = ConceptFetchHttpUploader(lazy { HttpURLConnectionClient() as Client }))`.
+    See [PR #6875](https://github.com/mozilla-mobile/android-components/pull/6875) for details and full code examples.
+
 # 41.0.0
 
 * [Commits](https://github.com/mozilla-mobile/android-components/compare/v40.0.0...v41.0.0)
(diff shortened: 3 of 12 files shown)


components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
```diff
@@ -37,6 +36,12 @@ data class Configuration @JvmOverloads constructor (
      * @return a [mozilla.telemetry.glean.config.Configuration] instance.
      */
     fun toWrappedConfiguration(): GleanCoreConfiguration {
-        return GleanCoreConfiguration(serverEndpoint, channel, maxEvents, httpClient)
+        // Only pass on the HTTP client if one was configured,
+        // otherwise fall back to the default in Glean core.
+        if (httpClient != null) {
+            return GleanCoreConfiguration(serverEndpoint, channel, maxEvents, httpClient)
+        } else {
+            return GleanCoreConfiguration(serverEndpoint, channel, maxEvents)
```
Dexterp37, line 44: Is our default behaving the same as `lib-fetch-httpurlconnection`? Please note that this will have impact on Lockwise and Firefox Reality. I'd rather making this explicit..

badboy: Make it explicit how?
We don't want to depend on `lib-fetch-httpurlconnection`.
We implement a ping uploader here: https://github.com/mozilla/glean/blob/4aab82a58c0e9db1577ebd10019447155bb846e7/glean-core/android/src/main/java/mozilla/telemetry/glean/net/HttpURLConnectionUploader.kt#L20
It's defering upload to `java.net.HttpURLConnection`.

IMO it makes more sense to _not_ redefine anything like that in a-c.
That only means we would now introduce a second place we need to change if we ever touch the Glean core uploader.

Dexterp37: Sorry I wasn't clear :-) I mean let's just drop any default and force users to explicitly state what they want.

badboy: If we only drop it here than that configuration API differs between a-c and direct use, which seems ... not good

Dexterp37: The API already differs between a-c and direct use: a-c depends on `ConceptFetch` while direct use doesn't know about it.

If the intent here is to make sure Mozilla users are intentional with the upload mechanism, I'm afraid this is the only option.

badboy: That would mean we're also breaking all the current users, as they are now forced to pass this explicitly.
Are we okay with doing that?

Dexterp37: > Are we okay with doing that?

This is what @pocmo [was asking](https://github.com/mozilla-mobile/android-components/issues/6660#issuecomment-624045948): I'm not a fan of this dance, to be honest, but we're living in A-C and we should respect the A-C's guidelines :-)

pocmo: > but we're living in A-C and we should respect the A-C's guidelines :-)

I'm open to discussing this though. :)

I think it is a fairly cheap breakage: Apps can fix this with one line of code. Fenix shouldn't break since Fenix should never use anything but Gecko for network communication. If it breaks Fenix then this would be good. :)

Lockwise: Yeah, they would probably need to add `HttpUrlConnectionClient` manually once. I think that is fine.

Firefox Reality: Haven't looked at it yet. But if it breaks then this may actually be good since I assume there is a strong reason to also use Gecko for all network communication in Firefox Reality. And if they have been using `HttpUrlConnection` instead then this may exactly be one of the cases where the default was not obvious (or having the choice) and where the default is the opposite of what we wanted.

badboy:
I updated the PR now. I still need to test that it still works in the mentioned apps and probably prepare PRs for them.

components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
pocmo: Firefox Reality seems to use the default and I filed an issue for changing that: https://github.com/MozillaReality/FirefoxReality/issues/3339

components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt
```diff
@@ -4,26 +4,28 @@
...
+ *                      you can fall back to a simple implementation on top of `java.net` using
+ *                      `ConceptFetchHttpUploader(lazy { HttpURLConnectionClient() })`
+ * @property serverEndpoint (optional) the server pings are sent to. Please note that this is
  *           is only meant to be changed for tests.
- * @property channel the release channel the application is on, if known. This will be
+ * @property channel (optional )the release channel the application is on, if known. This will be
  *           sent along with all the pings, in the `client_info` section.
- * @property maxEvents the number of events to store before the events ping is sent
- * @property httpClient The HTTP client implementation to use for uploading pings.
+ * @property maxEvents (optional )the number of events to store before the events ping is sent
```
Dexterp37, line 22:
```suggestion
 * @property maxEvents (optional) the number of events to store before the events ping is sent
```

docs/changelog.md
```diff
@@ -12,6 +12,13 @@ permalink: /changelog/
 * [Gecko](https://github.com/mozilla-mobile/android-components/blob/master/buildSrc/src/main/java/Gecko.kt)
 * [Configuration](https://github.com/mozilla-mobile/android-components/blob/master/buildSrc/src/main/java/Config.kt)
 
+* **service-glean**
+  * ⚠️ **This is a breaking change**: Glean's configuration now requires an http client. Users need to pass one at construction.
```
Dexterp37, line 16:
```suggestion
  * ⚠️ **This is a breaking change**: Glean's configuration now requires explicitly setting an http client. Users need to pass one at construction.
```

components/service/glean/src/main/java/mozilla/components/service/glean/net/ConceptFetchHttpUploader.kt
```diff
@@ -33,6 +35,16 @@ class ConceptFetchHttpUploader(
...
         // The timeout, in milliseconds, to use when reading from the server.
         const val DEFAULT_READ_TIMEOUT = 30000L
+
+        /**
+         * Export a constructor that is usable from Java.
+         *
+         * This looses the lazyness of creating the `client`.
+         */
+        @JvmStatic
+        fun fromClient(client: Client): ConceptFetchHttpUploader {
```
Dexterp37, line 45: Firefox Reality would need to use a different uploader. Do we need to do this for other uploaders as well? @pocmo

docs/changelog.md
Dexterp37, line 16: We should also update the glean sdk docs to mention this ...

components/service/glean/src/main/java/mozilla/components/service/glean/net/ConceptFetchHttpUploader.kt
pocmo: I'm not sure I fully understand. The 1:1 migration for Firefox Reality would be to continue using `HttpUrlConnectionClient` (https://github.com/MozillaReality/FirefoxReality/issues/3339)?

badboy: `Client` is coming from concept-fetch.
So as long as whatever they want to use adheres to that they're fine.

components/service/glean/src/main/java/mozilla/components/service/glean/net/ConceptFetchHttpUploader.kt
```diff
@@ -33,6 +35,16 @@ class ConceptFetchHttpUploader(
         const val DEFAULT_CONNECTION_TIMEOUT = 10000L
         // The timeout, in milliseconds, to use when reading from the server.
         const val DEFAULT_READ_TIMEOUT = 30000L
+
+        /**
+         * Export a constructor that is usable from Java.
+         *
+         * This looses the lazyness of creating the `client`.
```
travis79, line 42:
```suggestion
         * This loses the laziness of creating the `client`.
```

badboy: english. hard.

docs/changelog.md
badboy: Will do https://bugzilla.mozilla.org/show_bug.cgi?id=1636951

Dexterp37:
Please go ahead and merge this ;)

badboy:
bors r=Dexterp37

Issue #6884, opened 2020-05-06
SessionManager.onTrimMemory(): Do not remove engine session state from sessions


pocmo:
When trimming memory we want to save the state of an engine session, attach it to the holder and close the engine session. Right now we try to do this for all sessions and then "unlink" them. This causes the engine session state to be cleared from sessions that either did not have an engine session (e.g. restored) or that have been trimmed before. We should only perform this action for sessions that actually have an engine session. All other sessions can be left as-is.

Related Fenix issues:
- https://github.com/mozilla-mobile/fenix/issues/9039
- https://github.com/mozilla-mobile/fenix/issues/6726

PR #6885, opened 2020-05-06
Issue #6884: SessionManager.onTrimMemory(): Do not remove engine session state from sessions.

pocmo:
When trimming memory we want to save the state of an engine session, attach it to the holder and close the engine session. Right now we try to do this for all sessions and then "unlink" them. This causes the engine session state to be cleared from sessions that either did not have an engine session (e.g. restored sessions) or that have been trimmed before. We should only perform this action for sessions that actually have an engine session. All other sessions can be left as-is.

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
@@ -80,6 +80,13 @@ class SessionManager(
 
             session.engineSessionHolder.engineSession = null
             session.engineSessionHolder.engineObserver = null
+
+            // Note: We could consider not clearing the engine session state here. Instead we could
+            // try to actively set it here by calling save() on the engine session (if we have one).
+            // That way adding the same Session again could restore the previous state automatically
+            // (although we would need to make sure we do not override it with null in link()).
+            // Clearing the engine session state would be left to the garbage collector whenever the
+            // session itself gets collected.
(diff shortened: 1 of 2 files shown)


components/browser/session/src/main/java/mozilla/components/browser/session/SessionManager.kt
```diff
@@ -80,6 +80,13 @@ class SessionManager(
 
             session.engineSessionHolder.engineSession = null
             session.engineSessionHolder.engineObserver = null
+
+            // Note: We could consider not clearing the engine session state here. Instead we could
+            // try to actively set it here by calling save() on the engine session (if we have one).
+            // That way adding the same Session again could restore the previous state automatically
+            // (although we would need to make sure we do not override it with null in link()).
+            // Clearing the engine session state would be left to the garbage collector whenever the
+            // session itself gets collected.
```
csadilek, line 89: Interesting, yes. I do also prefer the current fix, as it's a bit easier to understand i.e. we only save the engine session state and unlink if we have an engine session, and eagerly remove the old state in unlink, but both seem fine.

pocmo: Yeah, always keeping (or actively saving) the state may make it easier for the current "undo" workaround of re-adding a `Session`. But I also would like to not make this change for now if not really needed. :)

pocmo:
bors r=csadilek

PR #6889, opened 2020-05-06
Closes #6756: Addon data lost during migration

### LINKED ISSUE #6756: Addon data lost upon migration
STR:

install ublock on fennec
add more custom lists in "Filter lists" under settings
install fenix
notice custom lists selected above are gone now

looks like this is due to us deleting `prefs.js` during migration, which the extension code uses to determine the internal UUID for the extension. This UUID is used to key all data in the `indexdb` database, so without the `prefs.js` file we lose all data.

csadilek:
Grisha and I tested this and it's looking good so far:

- Addon data (uBlock filters and settings) are still there post migration
- Addon ID is stable now
- In case migration leaves a broken prefs.js nothing bad happens and the file is just overridden with defaults

Open questions:
- ~Do we need to report more fine-grained success/failure results~ DONE
- ~Do we need to be more tolerant parsing the file~ DONE


csadilek:
OK, updated and ready for review.

components/support/migration/src/main/java/mozilla/components/support/migration/GeckoMigration.kt
```diff
@@ -4,10 +4,87 @@
...
+         * Failed to write backup file for prefs.js.
+         */
+        internal data class FailedToWriteBackup(val throwable: Throwable) : Failure() {
+            override fun toString(): String {
+                return "Failed to write backup of prefs.js file: ${throwable::class}"
+            }
+        }
+
+        /**
+         * Failed to write backup file for prefs.js.
```
grigoryk, line 70: wrong comment

components/support/migration/src/main/java/mozilla/components/support/migration/GeckoMigration.kt
```diff
@@ -4,10 +4,87 @@
...
+import androidx.annotation.VisibleForTesting
+import mozilla.components.support.ktx.util.writeString
 import java.io.File
 import java.io.IOException
 
 private const val PREFS_FILE = "prefs.js"
+private const val PREFS_BACKUP_FILE = "prefs.js.backup.v"
+
+@VisibleForTesting
+internal var userPrefsToKeep = setOf("extensions.webextensions.uuids")
```
grigoryk, line 17: const?

components/support/migration/src/main/java/mozilla/components/support/migration/GeckoMigration.kt
```diff
@@ -17,27 +94,73 @@ internal object GeckoMigration {
...
+        // Now let's transform the original to contain only prefs we want to keep.
+        val transformed = prefsjs.useLines {
+            it.filter { line ->
+                userPrefsToKeep.any { pref -> line.startsWith("$USER_PREF(\"$pref\"") }
+            }.toList()
+        }
+
+        return if (transformed.isEmpty()) {
+            removePrefsFile(prefsjs)
+        } else {
```
grigoryk, line 123: style nit: you don't _need_ to wrap this block in an `else` if you `return` inside of the `if`. that will help flatten this code a bit. the cost is repeating `return` twice, though.

components/support/migration/src/main/java/mozilla/components/support/migration/GeckoMigration.kt
```diff
@@ -17,27 +94,73 @@ internal object GeckoMigration {
...
+        }
     }
 
-    private fun removePrefsFile(profilePath: File): Result<Unit> {
-        val prefsFile = File(profilePath, PREFS_FILE)
-        if (prefsFile.exists()) {
-            return if (prefsFile.delete()) {
-                Result.Success(Unit)
+    private fun isLineValid(pref: String): Boolean {
+        return pref.lastIndexOf("$USER_PREF") == 0 && pref.endsWith(");")
```
grigoryk, line 135: maybe add a comment saying that we're checking if `user_pref` is present exactly once _and_ as the very first thing? it's not super obvious.

components/support/migration/src/main/java/mozilla/components/support/migration/GeckoMigration.kt
```diff
@@ -17,27 +94,73 @@ internal object GeckoMigration {
...
+    }
+
+    @Suppress("TooGenericExceptionCaught")
+    private fun removePrefsFile(prefsJs: File, invalid: Boolean = false): Result<GeckoMigrationResult> {
+        return try {
+            if (prefsJs.delete()) {
```
grigoryk, line 141: if you care about logging the exact exception: ` Note that the Files class defines the delete method to throw an IOException when a file cannot be deleted. This is useful for error reporting and to diagnose why a file cannot be deleted.`

components/support/migration/src/main/java/mozilla/components/support/migration/GeckoMigration.kt
```diff
@@ -17,27 +94,73 @@ internal object GeckoMigration {
...
+                return if (invalid) {
+                    Result.Success(GeckoMigrationResult.Success.PrefsFileRemovedInvalidPrefs)
+                } else {
+                    Result.Success(GeckoMigrationResult.Success.PrefsFileRemovedNoPrefs)
+                }
             } else {
-                Result.Failure(IOException("$PREFS_FILE exists but could not be deleted"))
+                Result.Failure(GeckoMigrationException(GeckoMigrationResult.Failure.FailedToDeleteFile()))
             }
+        } catch (e: Exception) {
```
grigoryk, line 150: this will be a `SecurityException`, according to java docs.

components/support/migration/src/main/java/mozilla/components/support/migration/GeckoMigration.kt
csadilek: Can't be a const because it's a set. Only primitives can be const :(.

components/support/migration/src/main/java/mozilla/components/support/migration/GeckoMigration.kt
csadilek: True, just tried, but to me at least, the single return is easier to read, but that's definitely subjective 🤷‍♂️ .

components/support/migration/src/main/java/mozilla/components/support/migration/GeckoMigration.kt
csadilek: Yeah, good call. Added a comment.

components/support/migration/src/main/java/mozilla/components/support/migration/GeckoMigration.kt
csadilek: Yeah, we're already catching the exception, but in case it returns false without throwing we also report that as an error. Or did I misunderstand what you meant?

csadilek: Oh, I see you suggest using `Files.delete` instead.

components/support/migration/src/main/java/mozilla/components/support/migration/GeckoMigration.kt
csadilek: Updated.

csadilek:
> This definitely looks more complicated than it is :)

Haha. All the error handling / telemetry made this blow up in size, but glad we did it.

grigoryk:
Re-tested with the telemetry in place, it's looking good!

grigoryk:
bors retry

csadilek:
bors retry

grigoryk:
bors retry

csadilek:
bors retry

Issue #6890, opened 2020-05-04
[Bug][a11y] The app crash when you open a PiP screen using Talkback


sv-sdeiac:
### Prerequisites
The talkback option is enabled from the Android settings.

## Steps to reproduce
1. Open a video in youtube.com;
2. Go to full-screen;
3. Display the Picture-in-picture video mode.

### Expected behavior
The Picture-in-Picture mode is displayed and the user can interact with the floating screen.

### Actual behavior
The app crash when the PiP mode is displayed.

### Device information

Android device: Samsung Galaxy Note S8 (Android 9).
Fenix version: Firefox Preview Nightly 4/30 #21210606.
Not reproducible using LG G7 Fit (Android 8.1).
[crashTalkback-PiP.txt](https://github.com/mozilla-mobile/fenix/files/4572922/crashTalkback-PiP.txt)

PR #6891, opened 2020-05-07
For #6890 - Catch IllegalStateException for PiP crash on Samsung devices

ekager:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/session/src/main/java/mozilla/components/feature/session/PictureInPictureFeature.kt
+++ b/components/feature/session/src/main/java/mozilla/components/feature/session/PictureInPictureFeature.kt
@@ -11,6 +11,8 @@ import android.os.Build
 import androidx.annotation.RequiresApi
 import mozilla.components.browser.session.SessionManager
 import mozilla.components.browser.session.runWithSessionIdOrSelected
+import mozilla.components.support.base.crash.CrashReporting
+import mozilla.components.support.base.log.logger.Logger
 
 /**
  * A simple implementation of Picture-in-picture mode if on a supported platform.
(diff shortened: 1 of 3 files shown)


components/feature/session/src/main/java/mozilla/components/feature/session/PictureInPictureFeature.kt
```diff
@@ -39,7 +39,14 @@ class PictureInPictureFeature(
...
                 session.fullScreenMode
             }
-        return fullScreenMode && enterPipModeCompat()
+        return fullScreenMode && try {
+            enterPipModeCompat()
+        } catch (e: IllegalStateException) {
+            // On certain Samsung devices, if accessibility mode is enabled, this will throw an
+            // IllegalStateException even if we check for the system feature beforehand. So let's just
+            // catch it, and not enter PiP. See https://stackoverflow.com/q/55288858
+            false
```
jonalmeida, line 48: Nice find! Might be worth logging the error though with the crash reporter so we know how often this happens. `IllegalStateException` is quite broad of an exception so we may catch other unknown exceptions here too.

components/feature/session/src/main/java/mozilla/components/feature/session/PictureInPictureFeature.kt
```diff
@@ -39,7 +42,15 @@ class PictureInPictureFeature(
...
                 session.fullScreenMode
             }
-        return fullScreenMode && enterPipModeCompat()
+        return fullScreenMode && try {
+            enterPipModeCompat()
+        } catch (e: IllegalStateException) {
+            // On certain Samsung devices, if accessibility mode is enabled, this will throw an
+            // IllegalStateException even if we check for the system feature beforehand. So let's
+            // catch it, log it, and not enter PiP. See https://stackoverflow.com/q/55288858
+            logger.error("Entering PipMode failed", e)
```
jonalmeida, line 51: The logger would add the error to the on-device logs only. We should pass in `CrashReporting` instance here so that we can log it on one of our crash reporters. Here is an [example usage](https://searchfox.org/mozilla-mobile/source/android-components/components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt#653).

components/feature/session/src/main/java/mozilla/components/feature/session/PictureInPictureFeature.kt
```diff
@@ -11,22 +11,28 @@ import android.os.Build
...
  * @param activity the activity with the EngineView for calling PIP mode when required; the AndroidX Fragment
  * doesn't support this.
+ * @param crashReporting Instance of `CrashReporting` to record unexpected caught exceptions
  * @param customTabSessionId ID of custom tab session.
  * @param pipChanged a change listener that allows the calling app to perform changes based on PIP mode.
  */
 class PictureInPictureFeature(
     private val sessionManager: SessionManager,
     private val activity: Activity,
+    private val crashReporting: CrashReporting,
```
jonalmeida, line 30: This is a breaking API that would need to be fixed on the next snapshots update.

```kotlin
private val crashReporting: CrashReporting? = null
```

If you want to that update to the feature your own pace, a default nullable works too.

ekager:
bors r=jonalmeida

Issue #6897, opened 2020-05-07
Add 'dismiss' property to DownloadState and handle it in DownloadFeature


codrut-topliceanu:
AC work for https://github.com/mozilla-mobile/fenix/issues/9044

In order to comply with the [changes requested here](https://github.com/mozilla-mobile/fenix/issues/9044#issuecomment-609522470), which describe a non-interfering  onDownloadStopped notification 'dialog', there needs to be a way to keep track if the user has dismissed one of these notification dialogs or not and to tie them to their individual tabs.

Thus changes needed would be a 'dismissed' property for `DownloadState`, that state to survive the `ConsumeDownloadAction` in `ContentStateReducer` and for `DownloadFeature` to be aware when trying to process downloads to ignore dismissed downloadStates.

codrut-topliceanu:
Following [the discussion here ](https://github.com/mozilla-mobile/android-components/pull/6899) I've decided to close this issue and try to complete any necessary work in Fenix.

PR #6899, opened 2020-05-07
For #6897 - Add dismiss property to DownloadState

codrut-topliceanu:
DownloadState.dismissed will be used to keep track of dismissed download notification dialogs in Fenix.
---

For https://github.com/mozilla-mobile/android-components/issues/6897

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt
+++ b/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt
@@ -94,7 +94,9 @@ class DownloadsFeature(
                 .ifChanged { it.content.download }
                 .collect { state ->
                     val download = state.content.download
-                    if (download != null) {
+                    // [dismissed] Indicates that the user has dismissed the onDownloadStopped
+                    // notification tab, so there is no need to process anything.
+                    if (download != null && !download.dismissed) {
                         processDownload(state, download)
                     }
                 }
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/ContentStateReducer.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/ContentStateReducer.kt
@@ -63,7 +63,9 @@ internal object ContentStateReducer {
                 it.copy(download = action.download)
             }
             is ContentAction.ConsumeDownloadAction -> updateContentState(state, action.sessionId) {
-                if (it.download != null && it.download.id == action.downloadId) {
+                if (it.download != null && it.download.id == action.downloadId &&
+                    it.download.dismissed
+                ) {
                     it.copy(download = null)
                 } else {
                     it
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/state/content/DownloadState.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/state/content/DownloadState.kt
@@ -22,6 +22,7 @@ import kotlin.random.Random
  * @property referrerUrl The site that linked to this download.
  * @property skipConfirmation Whether or not the confirmation dialog should be shown before the download begins.
  * @property id The unique identifier of this download.
+ * @property dismissed Indicates that the user has dismissed the onDownloadStopped notification tab.
  */
 @Suppress("Deprecation")
 @Parcelize
@@ -34,7 +35,8 @@ data class DownloadState(
     val destinationDirectory: String = Environment.DIRECTORY_DOWNLOADS,
     val referrerUrl: String? = null,
     val skipConfirmation: Boolean = false,
-    val id: Long = Random.nextLong()
+    val id: Long = Random.nextLong(),
+    val dismissed: Boolean = false
 ) : Parcelable {
     val filePath: String get() =
         Environment.getExternalStoragePublicDirectory(destinationDirectory).path + "/" + fileName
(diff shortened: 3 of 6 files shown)


codrut-topliceanu:
**Note:** Needs landing **before** https://github.com/mozilla-mobile/fenix/pull/10491

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt
```diff
@@ -94,7 +94,9 @@ class DownloadsFeature(
                 .ifChanged { it.content.download }
                 .collect { state ->
                     val download = state.content.download
-                    if (download != null) {
+                    // [dismissed] Indicates that the user has dismissed the onDownloadStopped
+                    // notification tab, so there is no need to process anything.
+                    if (download != null && !download.dismissed) {
```
Amejia481, line 99: Hi @codrut-topliceanu thanks for the patch!
I'm bit confused about why do we need a `dismissed` property in the `DownloadState`, would you mind help me to understand? :)

I see you are using the `dismissed` property to avoid the download to be processed, but as far as I'm aware the download starts after the user clicks the "Download" button in the download dialog.

codrut-topliceanu: Hey @Amejia481 
Sure thing. I'll go over my train of thought.
I needed a way to remember the state of the download in case the user would have the new download-stopped-view (un-inspiringly named `DownloadNotification`) on screen and decided to switch tabs or do some other work, and I noticed that `DownloadState` was attached to `ContentState` which was in turn attached to `TabSessionState`, but the `DownloadState` was getting consumed in `ContentStateReducer`. So I added the `dismissed` property to try to keep the `DownloadState` alive until the user is done with said view. However when the `dismissed` property is changed it triggers the flow from `DownloadsFeature.kt`, which in turn would make it try to process the download state again: 

  ```
store.flowScoped { flow ->
            flow.mapNotNull { state -> state.findTabOrCustomTabOrSelectedTab(tabId) }
                .ifChanged { it.content.download }....
```
...so I added that 'if' there.

pocmo: I agree that I do not fully understand the usage of a `dismissed` flag here. The `DownloadState` in the `ContentState` signals a "download request". It gets handled by the download feature which will display the download request. The user then decides to download the file or not. After that the state should get consumed and no longer exist on the ContentState. 🤔

codrut-topliceanu: Hey @pocmo , thank you for taking a look at this.
I really need `DownloadState` to survive so I can keep the download notification alive even if the user changes tabs, and this is the easiest way I found to do that. This PR doesn't make sense on its own. (https://github.com/mozilla-mobile/fenix/pull/10491)

I suppose I could leave `DownloadState` alone and add the needed properties to something else, like `ContentState`, but I'm not sure that would be an improvement over this.

Amejia481: @codrut-topliceanu  Ok, I think now, we are in the same page, what got us both confused was the name that you are using for referring to the new dialog ([DownloadNotification.kt ](https://github.com/mozilla-mobile/fenix/blob/5caf9a39230ab588bb03349513f6061f801b4a8c/app/src/main/java/org/mozilla/fenix/downloads/DownloadNotification.kt)), we thought you were referring to the notification in the status bar, I think it would be better to rename the file to `DownloadDialog` or something similar that include "Dialog" as it could cause a lot of confusion.

As for the `dismissed` property, I think this is something that could be better handled directly in the app as it's UI specific for Fenix, not necessary the reality for all consumers of Android Components as a library. We let consumer know the download was successful via `downloadFeature.onDownloadStopped`  then apps(consumers) decide which is the best UI to let their users know about it. Let me know what you think!

Amejia481: Additionally, In my opinion  dismissing the dialog as the user interacts with any other UI content/switch tabs is good idea and avoids issues like the one that I mentioned here https://github.com/mozilla-mobile/fenix/pull/10491#issuecomment-626790698

codrut-topliceanu: @Amejia481 thanks for your continued patience in this matter.
- I have just changed the name of the `DownloadNotification`, which was a very poorly chosen one, and will commit the changes in Fenix.

- I think you are right, other users of the Android Components library would have no need of the `dismissed` property, and would, no doubt, be puzzled by its presence, so I will remove it. 

- The `dismissed` property would be required in the case that we need to keep the dialog (and `DownloadState`) alive and attached to each individual tab. I have send a Slack message to @brampitoyo to inquire as to whether this should be the case or if the dialog should just be dismissed as soon as the user interacts with any other UI content or switches tabs. Either way I will make the changes in Fenix.

I will close this PR and the related issue. Again, thank you for all your help!

Amejia481: No problem I'm glad to help, thanks for all  great we work we really appreciate it ❤️

brampitoyo: I’ve responded to @codrut-topliceanu on Slack, but want to repost it here.

I think that the download complete notification shouldn’t be dismissed when users interact with other elements in the UI (for example: open the main menu, open Settings, switch tabs).

If it causes any layout problem as @Amejia481 posted on https://github.com/mozilla-mobile/fenix/pull/10491#issuecomment-626790698, I’d be happy to redesign the dialog so that it’s narrow and only takes up one line of text.

Let me know!

PR #6900, opened 2020-05-07
AddonMigration: Update fallback list with M2 add-ons

csadilek:
Updates the fallback list (used in case of network problems) to include the newly supported add-ons (beta and release).

List comes from:
https://addons.mozilla.org/api/v4/accounts/account/mozilla/collections/16f6e5d9a40448b8955db57ced6d75/addons/


csadilek:
bors r=grigoryk,Amejia481

PR #6906, opened 2020-05-07
Issue #2985: Support loading reader view ext. page in new session

csadilek:
Basically, a leftover from the refactoring. We used to turn on reader mode when the port connected and reader mode was active for this session. Now we have a separate page which could be loaded in a new session. When the reader extension page connect, we can simply "show" reader mode (and mark it as active for the session, but that part already worked.)


csadilek:
bors r=jonalmeida

PR #6909, opened 2020-05-07
Close #6842: Migrate to browser-state in BrowserThumbnails

jonalmeida:
---

Closes #6842 

### Pull Request checklist

- [x] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [x] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/android-components/blob/master/docs/changelog.md) or does not need one
- [x] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features

### After merge
- **Milestone**: Make sure issues closed by this pull request are added to the [milestone](https://github.com/mozilla-mobile/android-components/milestones) of the version currently in development.
- **Breaking Changes**: If this is a breaking change, please push a draft PR on [Reference Browser](https://github.com/mozilla-mobile/reference-browser) to address the breaking issues.

Issue #6905, opened 2020-05-07
Update heurestic for more requesting thumbnail screenshots


jonalmeida:
In our current implementation, we request a screenshot when the page is done loading. This isn't often enough unfortunately since other events can make the thumbnail screenshot out of sync with what is on the screen:
 - Javascript rendering.
 - User interactions before the page is done.

jonalmeida:
The easiest way would be to take a screenshot when leaving the BrowserFragment which would handle the case of getting the latest screenshot from the last time the user saw it.

We may need to also consider de-duping requests for screenshots from other features that want one too. In Fenix, I've seen the `BrowserAnimator` make requests for it during possible entry/exit as well.

jonalmeida:
In #6842, we're backing out the `onTitleChanged` observer that invokes a capture in place of is.

jonalmeida:
Updated the title to accurately reflect what this ticket needs to be done.

PR #6918, opened 2020-05-08
Issue #2754: Adds ThumbnailDiskCache for storing and restoring thumbnails

### LINKED ISSUE #2754: Persist and restore thumbnails
Thumbnails are not persisted. This means the next time a session is restored there are not thumbnails. We could consider persisting them to make that a more pleasant experience.

There are two options:
* Have the session storage take care of persisting and restoring thumbnails
* Have an independent cache of thumbnails (that may be useful if we want to display them eventually in other places).

gabrielluong:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/utils/ThumbnailDiskCache.kt
+++ b/components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/utils/ThumbnailDiskCache.kt
@@ -1,0 +1,86 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.thumbnails.utils
+
+import android.content.Context
+import android.graphics.Bitmap
+import com.jakewharton.disklrucache.DiskLruCache
+import mozilla.components.support.base.log.logger.Logger
+import java.io.File
+import java.io.IOException
+
+private const val MAXIMUM_CACHE_THUMBNAIL_DATA_BYTES: Long = 1024 * 1024 * 100 // 100 MB
+private const val THUMBNAIL_DISK_CACHE_VERSION = 1
+private const val WEBP_QUALITY = 90
+
+/**
+ * Caching thumbnail bitmaps on disk.
+ */
+class ThumbnailDiskCache {
+    private val logger = Logger("ThumbnailDiskCache")
+    private var thumbnailCache: DiskLruCache? = null
+    private val thumbnailCacheWriteLock = Any()
+
+    /**
+     * Retrieves the thumbnail data from the disk cache for the given session ID or URL.
+     *
+     * @param context the application [Context].
+     * @param sessionIdOrUrl the session ID or URL of the thumbnail to retrieve.
+     * @return the [ByteArray] of the thumbnail or null if the snapshot of the entry does not exist.
+     */
+    internal fun getThumbnailData(context: Context, sessionIdOrUrl: String): ByteArray? {
+        val snapshot = getThumbnailCache(context).get(sessionIdOrUrl) ?: return null
+
+        return try {
+            snapshot.getInputStream(0).use {
+                it.buffered().readBytes()
+            }
+        } catch (e: IOException) {
+            logger.info("Failed to read thumbnail bitmap from disk", e)
+            null
+        }
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


gabrielluong:
Fixes #2754. 

This focuses on adding a `ThumbnailDiskCache` to read and store thumbnail screenshots in WEBP format with their associated `sessionId`. In addition, we added the `ThumbnailDecoder` to decode the data we read from the DiskLruCache to bitmap.  

We will add a middleware to integrate this feature when this lands along with the work to migrate BrowserThumbnails to use browser-state in https://github.com/mozilla-mobile/android-components/pull/6909.

components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/decoder/ThumbnailDecoder.kt
```diff
@@ -0,0 +1,66 @@
...
+import androidx.annotation.VisibleForTesting
+import mozilla.components.support.base.log.logger.Logger
+import kotlin.math.floor
+import kotlin.math.max
+import kotlin.math.min
+
+/**
+ * A thumbnail decoder that can decode a [ByteArray] into a [Bitmap].
+ */
+class ThumbnailDecoder {
```
pocmo, line 19: Since this is very close to the icon decoder implementation: Do you think there's enough overlap so that we could extract it and share it via a `support-*` component?

components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/utils/ThumbnailDiskCache.kt
```diff
@@ -0,0 +1,72 @@
...
+import java.io.IOException
+
+private const val MAXIMUM_CACHE_THUMBNAIL_DATA_BYTES: Long = 1024 * 1024 * 100 // 100 MB
+private const val THUMBNAIL_DISK_CACHE_VERSION = 1
+private const val WEBP_QUALITY = 90
+
+/**
+ * Caching bitmaps and resource URLs on disk.
+ */
+class ThumbnailDiskCache {
```
pocmo, line 21: I am surprised that we are using the session id as a key here. This means this is not something we could use for top sites etc. too. Fennec had a cache using the URL as a key, I think.

If this is purely for the current list of sessions (e.g. tabs tray) then I wonder if we should integrate this into `SessionStorage` instead of having this as a standalone cache 🤔

jonalmeida: I suggested using the session ID for the tabs tray use case where we would want a thumbnail per session. I wasn't aware of top sites or other features that would want thumbnails too, but we could look into making sure we support both.

> I wonder if we should integrate this into SessionStorage instead of having this as a standalone cache 🤔

SessionStorage sounds like a good idea. We would need to make more `save` calls whenever a new site loads in order to keep the thumbnail fresh, but maybe that could be abstracted away from the in-memory/disk cache? We haven't reached https://github.com/mozilla-mobile/android-components/issues/6925 and https://github.com/mozilla-mobile/android-components/issues/6905 yet.

components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/decoder/ThumbnailDecoder.kt
gabrielluong: Fixed in https://github.com/mozilla-mobile/android-components/pull/7003

components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/utils/ThumbnailDiskCache.kt
gabrielluong: I am wondering if I should switch to using sessionIdOrUrl as the key value to the disk cache to resolve the different use case.

jonalmeida: > I am wondering if I should switch to using sessionIdOrUrl as the key value to the disk cache to resolve the different use case.

That sounds like a good idea. Thinking about it more after our conversations, it's probably best to not have the Thumbnails in session storage when we want to re-use this (eventually) for top sites. It would be more difficult to control an in-memory cache along with the session storage as this gets more complicated.

gabrielluong:
bors cancel

gabrielluong:
bors r=jonalmeida

Issue #6356, opened 2020-03-23
Install built-in extensions using GV extension controller


csadilek:
For built-in extensions we're still using the original (soon-to-be-deprecated) `registerWebExtension` API. Let's switch over to the new `runtime.webExtensionController.install` call which we already use for add-ons.

csadilek:
One issue we have to solve here is how change will be applied. Will installing the same version again, simply override the old one or do we have to call update? For new we can simply install again on startup (same as we do now with register), assuming that works.

csadilek:
Another change needed here:

- The new ID format (e.g. `icons@mozac.org`) is not a valid string for `runtime.connectNative` so we need to separate extension ID from app ID.

csadilek:
Decided not to land this right now:

- We're seeing a performance impact on startup. We'd have to do more testing to quantify precisely, but given that there's no user value landing this we don't need to rush it.

- Installing the built-in extensions on every startup introduces another problem. It causes all open extension tabs to be closed. This is a problem for reader view, which can't cleanly be worked around.

Conclusion here is that we shouldn't install on every startup, but only once, or when there's a newer version of the extension. Then there's also value in switching from the transient `registerWebExtension` to the persistent `installBuiltInExtension`, as it would address both concerns above (and also make startup faster).

GV bug for API to install only if needed:
https://bugzilla.mozilla.org/show_bug.cgi?id=1635564

Another major concern I have is that developers would forget to increment the version in the manifest, then we'd ship changes that never get applied correctly: https://github.com/mozilla-mobile/android-components/issues/7278

PR #6919, opened 2020-05-08
Update GeckoView Nightly version to 78.0.20200508093543

psymoon:


components/browser/engine-gecko-nightly/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineTest.kt
```diff
@@ -68,6 +68,7 @@ import java.lang.Exception
 import org.mozilla.geckoview.WebExtension as GeckoWebExtension
 
 @RunWith(AndroidJUnit4::class)
+@Suppress("Deprecation")
```
csadilek, line 71: I would prefer if we only put that on the methods that use `registerWebExtension`. It's easier to see what's deprecated and we can address it soon.

components/browser/engine-gecko-nightly/src/test/java/mozilla/components/browser/engine/gecko/webextension/GeckoWebExtensionTest.kt
```diff
@@ -40,6 +40,7 @@ import org.mozilla.geckoview.WebExtension
 import org.mozilla.geckoview.WebExtensionController
 
 @RunWith(AndroidJUnit4::class)
+@Suppress("Deprecation")
```
csadilek, line 43: I would prefer if we only put that on the methods that use `registerWebExtension`. It's easier to see what's deprecated and we can address it soon.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/webextension/GeckoWebExtension.kt
```diff
@@ -29,7 +29,7 @@ import org.mozilla.geckoview.WebExtension.Action as GeckoNativeWebExtensionActio
  * Gecko-based implementation of [WebExtension], wrapping the native web
  * extension object provided by GeckoView.
  */
-@Suppress("TooManyFunctions")
+@Suppress("TooManyFunctions", "Deprecation")
```
csadilek, line 32: Let's put the `@Suppress("Deprecation")` on the actual inner constructor call, instead of the whole class, so we know exactly why were doing it. Let's also link to: https://github.com/mozilla-mobile/android-components/issues/6356

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
```diff
@@ -203,6 +203,7 @@ class GeckoEngine(
         return installWebExtension(ext, onSuccess, onError)
     }
 
+    @Suppress("Deprecation")
```
csadilek, line 206: See below. Let's put this on the `register` call with a link to https://github.com/mozilla-mobile/android-components/issues/6356

components/browser/engine-gecko-nightly/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineTest.kt
psymoon: It's gonna have to be the methods that initializes gecko `WebExtension` constructors as well.

psymoon:
bors r=Amejia481, csadilek

PR #6924, opened 2020-05-08
For 6922 - Scroll to selected tab when entering tab tray

darkwing:
---

With my manual testing, I:

1.  Open enough tabs to make the tab tray scroll
2.  Select a tab toward the bottom of the screen
3.  Hit the tab tray button
4.  Scroll to the top of the tab tray
5.  Press / swipe back
6.  Selected tab is still scrolled down to

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/tabstray/src/main/java/mozilla/components/browser/tabstray/TabsAdapter.kt
+++ b/components/browser/tabstray/src/main/java/mozilla/components/browser/tabstray/TabsAdapter.kt
@@ -43,7 +43,11 @@ class TabsAdapter(
     private var tabs: Tabs? = null
 
     override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): TabViewHolder {
-        return viewHolderProvider.invoke(parent, tabsTray)
+        val result = viewHolderProvider.invoke(parent, tabsTray)
+        tabs?.let {
+            this.tabsTray.scrollToPosition(it.selectedIndex)
+        }
+        return result
     }
 
     override fun getItemCount() = tabs?.list?.size ?: 0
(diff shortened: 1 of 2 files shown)


darkwing:
The problem with this current implementation is by putting the `scrollToPosition` in `updateTabs`, the tab gets scrolled down to when closing any other tab, so `updateTabs` is too aggressive

components/browser/tabstray/src/main/java/mozilla/components/browser/tabstray/TabsAdapter.kt
```diff
@@ -43,7 +43,11 @@ class TabsAdapter(
     private var tabs: Tabs? = null
 
     override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): TabViewHolder {
-        return viewHolderProvider.invoke(parent, tabsTray)
+        val result = viewHolderProvider.invoke(parent, tabsTray)
+        tabs?.let {
```
jonalmeida, line 47: `onCreateViewHolder` is a called every time a new tabs needs to be created, so this would be invoked n-tab times.

What we probably want is for an observer to know when the RecyclerView is done laying out the views so we can scroll to right one. I did some googling around on how others use this API:

What we can do is add the [`RecyclerView.AdapterDataObserver`](https://developer.android.com/reference/androidx/recyclerview/widget/RecyclerView.AdapterDataObserver) to know when the data has changed and we need to scroll to the right position.

This could be done in `BrowserTabsTray`.

Please also add tests, we have a guide on getting full code coverage to aid with testing too: https://mozac.org/contributing/code-coverage

jonalmeida: > What we probably want is for an observer to know when the RecyclerView is done laying out the views so we can scroll to right one. I did some googling around on how others use this API:
> 
> What we can do is add the RecyclerView.AdapterDataObserver to know when the data has changed and we need to scroll to the right position.

It looks like this is already setup for you in the tabs tray, does it not work as expected? https://github.com/mozilla-mobile/android-components/blob/master/components/feature/tabs/src/main/java/mozilla/components/feature/tabs/tabstray/TabsTrayPresenter.kt#L87

PR #6936, opened 2020-05-11
Upgrade Glean to v29.1.0

badboy:
---

### Pull Request checklist

### After merge

### DIFF
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -29,7 +29,7 @@ object Versions {
 
     const val mozilla_appservices = "0.58.2"
 
-    const val mozilla_glean = "28.0.0"
+    const val mozilla_glean = "29.1.0"
 
     const val material = "1.1.0"
     const val nearby = "17.0.0"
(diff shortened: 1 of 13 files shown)


mdboom:
It will change the range of acceptable values on GV metrics at the upper end of the range.  This will result in fewer errors when they overflow, and more correct results, and I think this is what we want.

At the lower end, incoming milliseconds could never be less than one millisecond anyway, so there is no change.

Dexterp37:
> 
> 
> It will change the range of acceptable values on GV metrics at the upper end of the range. This will result in fewer errors when they overflow, and more correct results, and I think this is what we want.
> 
> At the lower end, incoming milliseconds could never be less than one millisecond anyway, so there is no change.

Cool, thanks @mdboom - so we don't expect any disruption in things that are already being collected, right?

mdboom:
> Cool, thanks @mdboom - so we don't expect any disruption in things that are already being collected, right?

Nope.  Should be no disruption.  It didn't touch the multiplication of values or anything, just increased the maximum.

Dexterp37:
> 
> 
> > Cool, thanks @mdboom - so we don't expect any disruption in things that are already being collected, right?
> 
> Nope. Should be no disruption. It didn't touch the multiplication of values or anything, just increased the maximum.

Lovely! @badboy let's get this party started then! 🎸

badboy:
bors r=Dexterp37

PR #6939, opened 2020-05-11
For #4261: Add custom viewfinder with message to QR scan fragment.

mcarare:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFeature.kt
+++ b/components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFeature.kt
@@ -30,6 +30,8 @@ typealias OnScanResult = (result: String) -> Unit
  * need to be requested before a QR scan can be performed. Once the request
  * is completed, [onPermissionsResult] needs to be invoked. This feature
  * will request [android.Manifest.permission.CAMERA].
+ * @property scanMessage (Optional) String resource for an optional message
+ * to be laid out below the QR scan viewfinder
  */
 class QrFeature(
     private val context: Context,
--- a/components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFeature.kt
+++ b/components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFeature.kt
@@ -7,6 +7,7 @@ package mozilla.components.feature.qr
 import android.Manifest.permission.CAMERA
 import android.content.Context
 import androidx.annotation.MainThread
+import androidx.annotation.StringRes
 import androidx.annotation.VisibleForTesting
 import androidx.fragment.app.FragmentManager
 import mozilla.components.support.base.feature.UserInteractionHandler
@@ -34,7 +35,9 @@ class QrFeature(
     private val context: Context,
     private val fragmentManager: FragmentManager,
     private val onScanResult: OnScanResult = { },
-    override val onNeedToRequestPermissions: OnNeedToRequestPermissions = { }
+    override val onNeedToRequestPermissions: OnNeedToRequestPermissions = { },
+    @StringRes
+    private var scanMessage: Int? = null
 ) : LifecycleAwareFeature, UserInteractionHandler, PermissionsFeature {
     private var containerViewId: Int = 0
 
@@ -75,7 +78,7 @@ class QrFeature(
 
         return if (context.isPermissionGranted(CAMERA)) {
             fragmentManager.beginTransaction()
-                .add(containerViewId, QrFragment.newInstance(scanCompleteListener), QR_FRAGMENT_TAG)
+                .add(containerViewId, QrFragment.newInstance(scanCompleteListener, scanMessage), QR_FRAGMENT_TAG)
                 .commit()
             true
         } else {
--- a/components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt
+++ b/components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt
@@ -41,6 +41,8 @@ import android.view.Surface
 import android.view.TextureView
 import android.view.View
 import android.view.ViewGroup
+import androidx.annotation.StringRes
+import androidx.core.content.ContextCompat.getColor
 import androidx.fragment.app.Fragment
 import com.google.zxing.BinaryBitmap
 import com.google.zxing.MultiFormatReader
@@ -48,6 +50,7 @@ import com.google.zxing.NotFoundException
 import com.google.zxing.PlanarYUVLuminanceSource
 import com.google.zxing.common.HybridBinarizer
 import mozilla.components.feature.qr.views.AutoFitTextureView
+import mozilla.components.feature.qr.views.CustomViewFinder
 import mozilla.components.support.base.log.logger.Logger
 import java.io.Serializable
 import java.util.ArrayList
@@ -93,6 +96,9 @@ class QrFragment : Fragment() {
     }
 
     internal lateinit var textureView: AutoFitTextureView
+    internal lateinit var customViewFinder: CustomViewFinder
+    @StringRes
+    private var scanMessage: Int? = null
     internal var cameraId: String? = null
     private var captureSession: CameraCaptureSession? = null
     internal var cameraDevice: CameraDevice? = null
@@ -114,6 +120,11 @@ class QrFragment : Fragment() {
                 override fun onScanComplete(result: String) {
                     Handler(Looper.getMainLooper()).apply {
                         post {
+                            context?.let {
+                                customViewFinder.setViewFinderColor(
+                                    getColor(it, R.color.mozac_feature_qr_scan_success_color)
+                                )
+                            }
                             value?.onScanComplete(result)
                         }
                     }
@@ -197,6 +208,10 @@ class QrFragment : Fragment() {
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         textureView = view.findViewById<View>(R.id.texture) as AutoFitTextureView
+        customViewFinder = view.findViewById<View>(R.id.view_finder) as CustomViewFinder
+        scanMessage?.let {
+            CustomViewFinder.setMessage(it)
+        }
         qrState = STATE_FIND_QRCODE
     }
 
@@ -471,9 +486,15 @@ class QrFragment : Fragment() {
 
         private const val CAMERA_CLOSE_LOCK_TIMEOUT_MS = 2500L
 
-        fun newInstance(listener: OnScanCompleteListener): QrFragment {
+        /**
+         * Returns a new instance of QR Fragment
+         * @param listener Listener invoked when the QR scan completed successfully.
+         * @param scanMessage (Optional) Scan message to be displayed.
+         */
+        fun newInstance(listener: OnScanCompleteListener, scanMessage: Int? = null): QrFragment {
             return QrFragment().apply {
                 scanCompleteListener = listener
+                this.scanMessage = scanMessage
             }
         }
 
@@ -492,7 +513,7 @@ class QrFragment : Fragment() {
          * @param aspectRatio The aspect ratio
          * @return The optimal `Size`, or an arbitrary one if none were big enough.
          */
-        @Suppress("LongParameterList")
+        @Suppress("LongParameterList", "ComplexMethod")
         internal fun chooseOptimalSize(
             choices: Array<Size>,
             textureViewWidth: Int,
--- a/components/feature/qr/src/main/java/mozilla/components/feature/qr/views/CustomViewFinder.kt
+++ b/components/feature/qr/src/main/java/mozilla/components/feature/qr/views/CustomViewFinder.kt
@@ -1,0 +1,284 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.feature.qr.views
+
+import android.content.Context
+import android.content.res.Configuration
+import android.graphics.Canvas
+import android.graphics.Color
+import android.graphics.Paint
+import android.graphics.Path
+import android.graphics.Rect
+import android.os.Build
+import android.text.Layout
+import android.text.StaticLayout
+import android.text.TextPaint
+import android.util.AttributeSet
+import android.view.View
+import android.view.ViewGroup
+import androidx.annotation.ColorInt
+import androidx.annotation.Px
+import androidx.appcompat.widget.AppCompatImageView
+import androidx.annotation.StringRes
+import androidx.annotation.VisibleForTesting
+import androidx.core.content.ContextCompat
+import androidx.core.text.HtmlCompat
+import mozilla.components.support.ktx.android.util.dpToPx
+import mozilla.components.support.ktx.android.util.spToPx
+import kotlin.math.min
+import kotlin.math.roundToInt
+
+/**
+ * A [View] that shows a ViewFinder positioned in center of the camera view and draws an Overlay
+ */
+@Suppress("LargeClass", "TooManyFunctions")
+class CustomViewFinder @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet? = null
+) : AppCompatImageView(context, attrs) {
+    private var messageResource: Int? = null
+    private val overlayPaint: Paint = Paint(Paint.ANTI_ALIAS_FLAG)
+    internal val viewFinderPaint: Paint = Paint(Paint.ANTI_ALIAS_FLAG)
+    private var viewFinderPath: Path = Path()
+    private var viewFinderPathSaved: Boolean = false
+    private var overlayPath: Path = Path()
+    private var overlayPathSaved: Boolean = false
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal lateinit var viewFinderRectangle: Rect
+    private var viewFinderCornersSize: Float = 0f
+    private var viewFinderCornersRadius: Float = 0f
+    private var viewFinderTop: Float = 0f
+    private var viewFinderLeft: Float = 0f
+    private var viewFinderRight: Float = 0f
+    private var viewFinderBottom: Float = 0f
+    private var normalizedRadius: Float = 0f
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal var scanMessageLayout: StaticLayout? = null
+    private lateinit var messageTextPaint: TextPaint
+
+    init {
+        isSaveEnabled = true
+        overlayPaint.style = Paint.Style.FILL
+        viewFinderPaint.style = Paint.Style.STROKE
+        overlayPath.fillType = Path.FillType.EVEN_ODD
+        viewFinderPath.fillType = Path.FillType.EVEN_ODD
+
+        this.layoutParams = ViewGroup.LayoutParams(
+            ViewGroup.LayoutParams.MATCH_PARENT,
+            ViewGroup.LayoutParams.MATCH_PARENT)
+        this.setOverlayColor(DEFAULT_OVERLAY_COLOR)
+        this.setViewFinderColor(DEFAULT_VIEWFINDER_COLOR)
+        this.setViewFinderStroke(DEFAULT_VIEWFINDER_THICKNESS_DP.dpToPx(resources.displayMetrics))
+        this.setViewFinderCornerRadius(DEFAULT_VIEWFINDER_CORNERS_RADIUS_DP.dpToPx(resources.displayMetrics))
+    }
+
+    /** Calculates viewfinder rectangle and triggers calculating message layout that depends on it */
+    internal fun computeViewFinderRect(width: Int, height: Int) {
+        if (width > 0 && height > 0) {
+            val minimumDimension = min(width.toFloat(), height.toFloat())
+
+            val viewFinderSide = (minimumDimension * DEFAULT_VIEWFINDER_WIDTH_RATIO).roundToInt()
+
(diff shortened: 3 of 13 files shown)


components/feature/qr/src/main/java/mozilla/components/feature/qr/views/CustomViewFinder.kt
```diff
@@ -0,0 +1,284 @@
...
+        this.layoutParams = ViewGroup.LayoutParams(
+            ViewGroup.LayoutParams.MATCH_PARENT,
+            ViewGroup.LayoutParams.MATCH_PARENT)
+        this.setOverlayColor(DEFAULT_OVERLAY_COLOR)
+        this.setViewFinderColor(DEFAULT_VIEWFINDER_COLOR)
+        this.setViewFinderStroke(DEFAULT_VIEWFINDER_THICKNESS_DP.dpToPx(resources.displayMetrics))
+        this.setViewFinderCornerRadius(DEFAULT_VIEWFINDER_CORNERS_RADIUS_DP.dpToPx(resources.displayMetrics))
+    }
+
+    private fun computeViewFinderRect(width: Int, height: Int) {
```
csadilek, line 76: Tricky to unit test view code, but it seems reasonable to add a test for this method and maybe also `drawOverlay/drawViewFinder` if `Path` can be spied on?

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt
```diff
@@ -471,9 +491,16 @@ class QrFragment : Fragment() {
...
 
-        fun newInstance(listener: OnScanCompleteListener): QrFragment {
+        /**
+         * Returns a new instance of QR Fragment
+         * @param listener Listener invoked when the QR scan completed successfully.
+         * @param scanMessage (Optional) Scan message to be displayed.
+         */
+        fun newInstance(listener: OnScanCompleteListener, scanMessage: Int? = null): QrFragment {
             return QrFragment().apply {
+                retainInstance = true
```
csadilek, line 501: Was this set to keep the new `scanMessage` when the fragment gets re-created? If so, we already solve that for the `scanCompleteListener` i.e. the feature could just set the `scanMessage` onStart. Then we don't have to retain the instance: https://github.com/mozilla-mobile/android-components/blob/master/components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFeature.kt#L52

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt
```diff
@@ -197,6 +210,13 @@ class QrFragment : Fragment() {
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         textureView = view.findViewById<View>(R.id.texture) as AutoFitTextureView
+        customViewFinder = view.findViewById<View>(R.id.view_finder) as CustomViewFinder
+        scanMessage?.let {
+            MainScope().launch {
```
csadilek, line 215: Can you add a comment here why we're launching? Presumably because the `customViewFinder` isn't drawn yet?

samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/MainActivity.kt
```diff
@@ -54,6 +54,7 @@ open class MainActivity : AppCompatActivity(), LoginFragment.OnLoginCompleteList
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+        RustLog.disable()
```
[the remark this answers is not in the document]
mcarare: I was testing the behavior in the sample app, and enable() is called at activity recreation and it crashes because it's already enabled. It was the quickest fix rather than check if it was enabled.

components/feature/qr/src/main/java/mozilla/components/feature/qr/views/CustomViewFinder.kt
```diff
@@ -0,0 +1,284 @@
...
+    /** Converts a value in density independent pixels (dp) to a px value. */
+    private fun Float.dpToPx(displayMetrics: DisplayMetrics) = TypedValue.applyDimension(
+        TypedValue.COMPLEX_UNIT_DIP, this, displayMetrics)
+
+    /** Converts a value in scale independent pixels (sp) to a px value. */
+    private fun Float.spToPx(displayMetrics: DisplayMetrics) = TypedValue.applyDimension(
+        TypedValue.COMPLEX_UNIT_SP, this, displayMetrics)
+
+    companion object {
+        private const val DEFAULT_VIEWFINDER_WIDTH_RATIO = 0.5f
```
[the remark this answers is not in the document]
mcarare: I added multiple functions that could add customization, and tried to make this as flexible as possible, so, if needed, in the future, constants could be added as parameters, or changed with a function (like changing the viewfinder color on scan result)

components/feature/qr/src/main/java/mozilla/components/feature/qr/views/CustomViewFinder.kt
```diff
@@ -0,0 +1,284 @@
...
+    /** Sets the corner radius for the ViewFinder. */
+    private fun setViewFinderCornerRadius(@Px radius: Float) {
+        viewFinderCornersRadius = radius
+        if (isLaidOut) {
+            invalidate()
+        }
+    }
+
+    /** Converts a value in density independent pixels (dp) to a px value. */
+    private fun Float.dpToPx(displayMetrics: DisplayMetrics) = TypedValue.applyDimension(
```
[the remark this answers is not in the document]
mcarare: At first I worked on this when it was included in support.base, where I did not have access to those (had to add as dependency, resulting in a circular dependency). So, should I move these to support.ktx ? (I would like to keep the float parameters, instead using Int)

components/feature/qr/src/main/java/mozilla/components/feature/qr/views/CustomViewFinder.kt
mcarare: I will give it a try. :)

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt
mcarare: I had some problems with testing in Fenix. It would either not show the message in the first draw, or it would not change it on rotation or not changing the StaticLayout width on rotation. Maybe it has to do with Fenix declaring the handling of config changes?

components/feature/qr/src/main/java/mozilla/components/feature/qr/views/CustomViewFinder.kt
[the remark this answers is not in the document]
mcarare: The. in problem I had with passing customization (like scan message) is that you have to pass it through QR Feature and then QR Fragment, and then to CustomViewFinder...

mcarare:
@csadilek As for the aspect ratio, I was planning on taking a look  at `PreviewView` from CameraX, cause it claims to have improved compatibility for devices, screen sizes, camera hardware. I am also concerned though with the testing that implementation will need.

samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/MainActivity.kt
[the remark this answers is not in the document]
csadilek: Ah interesting! Figured something must be wrong :). Fine to leave it then. Let's file an issue that it crashes if invoked twice and we should either prevent that or provide a method to check.

components/feature/qr/src/main/java/mozilla/components/feature/qr/views/CustomViewFinder.kt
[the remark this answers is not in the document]
csadilek: Yeah, let's, `dpToPx` already exists and the other one we can add.

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt
csadilek: I am not sure, but passing along `scanMessage` seems preferable over keeping the instance of the fragment around. Let's try / debug the UI issue you're seeing if still present?

samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/MainActivity.kt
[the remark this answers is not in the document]
mcarare: Filed https://github.com/mozilla-mobile/android-components/issues/6965.

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt
mcarare: I remembered the other way around. Removing `retainInstance = true` has no implications in Fenix, but in sample app the message disappears on rotation. I think Fenix retains instance by default, but other clients might not.

components/feature/qr/src/main/java/mozilla/components/feature/qr/views/CustomViewFinder.kt
mcarare: I added some unit tests. I did not find a way to spy on the path or canvas. Also, I avoided using Robolectric. There are some framework (https://facebook.github.io/screenshot-tests-for-android/) but I don't think it's worth adopting it just for these tests.

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFeature.kt
```diff
@@ -34,7 +35,9 @@ class QrFeature(
     private val context: Context,
     private val fragmentManager: FragmentManager,
     private val onScanResult: OnScanResult = { },
-    override val onNeedToRequestPermissions: OnNeedToRequestPermissions = { }
+    override val onNeedToRequestPermissions: OnNeedToRequestPermissions = { },
+    @StringRes
+    private var scanMessage: Int? = null
```
Amejia481, line 40: nit: Let's add some docs for this new property :)

mcarare: @Amejia481 Comments in code or entry in changelog?

Amejia481: both 😄

Amejia481: Initially I was thinking about the code but now that you mentioned the changelog let's add it too XD

components/support/ktx/src/main/java/mozilla/components/support/ktx/android/util/DisplayMetrics.kt
```diff
@@ -17,3 +17,11 @@ fun Int.dpToFloat(displayMetrics: DisplayMetrics) = TypedValue.applyDimension(
...
  */
 fun Int.dpToPx(displayMetrics: DisplayMetrics) = dpToFloat(displayMetrics).toInt()
+
+/** Converts a value in density independent pixels (dp) to a px value. */
+fun Float.dpToPx(displayMetrics: DisplayMetrics) = TypedValue.applyDimension(
+    TypedValue.COMPLEX_UNIT_DIP, this, displayMetrics)
+
+/** Converts a value in scale independent pixels (sp) to a px value. */
+fun Float.spToPx(displayMetrics: DisplayMetrics) = TypedValue.applyDimension(
+    TypedValue.COMPLEX_UNIT_SP, this, displayMetrics)
```
Amejia481, line 27: Should we add some tests for this new utilities? it will be nice to mentioning them in the changelog too :)

samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/MainActivity.kt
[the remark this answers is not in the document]
Amejia481: Should we document it in the code as it is not straightforward why it should be there? We could also add a link to the issue in the code.

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt
csadilek: OK, but that's an even stronger reason to remove it ;). If it's fine in Fenix, we can debug the UI hiccup in sample browser later. So, I would treat `scanMessage` the same as `scanCompleteListener` and make sure we set it when the features is started. I don't think we should keep the fragment instance longer than we need to.

samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/MainActivity.kt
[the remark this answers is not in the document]
csadilek: 👍

components/feature/qr/src/main/java/mozilla/components/feature/qr/views/CustomViewFinder.kt
csadilek: ok, thanks!

components/feature/qr/src/main/res/values/colors.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <color name="scan_success_color">#9059FF</color>
```
Amejia481, line 3: nit: We could prefix it with `mozac_feature_qr`

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFeature.kt
mcarare: Done! Let me know if you have any feedback on content.

Amejia481: Looks good!

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt
mcarare: I wasn't thinking about sample, but other clients that would use the QR scan feature.

components/support/ktx/src/main/java/mozilla/components/support/ktx/android/util/DisplayMetrics.kt
mcarare: Not really sure how we could test this. 🤔

Amejia481: You could assert that value you are getting from `spToPx` will be the same as executing 
```TypedValue.applyDimension(TypedValue.COMPLEX_UNIT_SP, this, displayMetrics)```,  as `spToPx` is a shortcut

File: components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt (hunk 2 above)
Comment (mcarare) [reply components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt]: I've removed it. We can pick it up at a latter time if the message disappearing issue arises.

File: components/support/ktx/src/main/java/mozilla/components/support/ktx/android/util/DisplayMetrics.kt (hunk 8 above)
Comment (mcarare) [reply components/support/ktx/src/main/java/mozilla/components/support/ktx/android/util/DisplayMetrics.kt:27]: I thought of that, but it would look like we are just testing how extensions work. Or maybe I am seeing it wrong.

File: components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt (hunk 2 above)
Comment (csadilek) [reply components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt]: @mcarare but we still need to pass the `scanMessage` in `onStart` similar to: https://github.com/mozilla-mobile/android-components/blob/master/components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFeature.kt#L52

Otherwise, you will lose the string if the fragment gets re-created which might be exactly the problem you run into in sample browser?

Comment (mcarare) [reply components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt]: Yes, but in Fenix the fragment does not get recreated.

File: components/support/ktx/src/main/java/mozilla/components/support/ktx/android/util/DisplayMetrics.kt (hunk 8 above)
Comment (Amejia481) [reply components/support/ktx/src/main/java/mozilla/components/support/ktx/android/util/DisplayMetrics.kt:27]: The idea is we want to assert the algorithm of converting a `Float` from `sp` to `px` will return the same value as using `spToPx`, if there are other ways to convert from an `sp` to `px` we can use it too.

Comment (mcarare) [reply components/support/ktx/src/main/java/mozilla/components/support/ktx/android/util/DisplayMetrics.kt:27]: Good point! Thanks!

File: components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt (hunk 2 above)
Comment (mcarare) [reply components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt]: The message still disappears even with `   it.scanMessage = scanMessage` added in `start()`. Only calling setupMessage again could fix that, but it would have to be synchronized somehow with the viewfinder redrawing.

Comment (csadilek) [reply components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt]: Ok, can you add the change here? I think we want that either way. I can take a quick look later today and worst case we can investigate as a follow-up if we can't get it to work, but let's definitely make sure we have the message in the fragment when it re-starts.

File: components/support/ktx/src/main/java/mozilla/components/support/ktx/android/util/DisplayMetrics.kt (hunk 8 above)
Comment (mcarare) [reply components/support/ktx/src/main/java/mozilla/components/support/ktx/android/util/DisplayMetrics.kt:27]: @Amejia481 Added the tests, let me know what you think.

File: components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt (hunk 2 above)
Comment (mcarare) [reply components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt]: Sorry, do you mean add `retainInstance = true` here? `this.scanMessage = scanMessage` is already called in `newInstance()`.

Comment (csadilek) [reply components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt]: @mcarare ah no, I mean passing along the `scanMessage` in onStart: https://github.com/mozilla-mobile/android-components/blob/master/components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFeature.kt#L52

**Comment by Amejia481** [comment]
@mcarare there is a crash after cancelling and trying to pair again:

**STR:**
1.  Tap "FXA SIGN N: PAIR"
2. Scan the QR code.
3. Click the "Cancel" button.
4. Press back
5. Tap "FXA SIGN N: PAIR"

```
kotlin.UninitializedPropertyAccessException: lateinit property viewFinderRectangle has not been initialized
        at mozilla.components.feature.qr.views.CustomViewFinder.setupMessage(CustomViewFinder.kt:204)
        at mozilla.components.feature.qr.QrFragment$onViewCreated$$inlined$let$lambda$1.invokeSuspend(QrFragment.kt:217)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:56)
        at android.os.Handler.handleCallback(Handler.java:883)
        at android.os.Handler.dispatchMessage(Handler.java:100)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7356)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)

```

mcarare:
> @mcarare there is a crash after cancelling and trying to pair again:
> 
> **STR:**
> 
> 1. Tap "FXA SIGN N: PAIR"
> 2. Scan the QR code.
> 3. Click the "Cancel" button.
> 4. Press back
> 5. Tap "FXA SIGN N: PAIR"
> 
 Haven't been able to reproduce, but, nevertheless, I am changing the implementation slightly, to ensure setup message is called **only** after `viewFinderRectangle` is drawn, as the message position depends entirely on `viewFinderRectangle` width and position.

Amejia481:
> > @mcarare there is a crash after cancelling and trying to pair again:
> > **STR:**
> > 
> > 1. Tap "FXA SIGN N: PAIR"
> > 2. Scan the QR code.
> > 3. Click the "Cancel" button.
> > 4. Press back
> > 5. Tap "FXA SIGN N: PAIR"
> 
> Haven't been able to reproduce, but, nevertheless, I am changing the implementation slightly, to ensure setup message is called **only** after `viewFinderRectangle` is drawn, as the message position depends entirely on `viewFinderRectangle` width and position.

Awesome, now it's working!

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt
csadilek: OK, yeah, making the message static as in your latest change works too! Tested with "don't keep activities" and the custom finder and message are still fine after the activity gets recreated.

mcarare:
@Amejia481 I am going to move the color update into the `Handler(Looper.getMainLooper()).post` although I would have preferred the `runOnUiThread` approach. Any thoughts?

Amejia481:
> @Amejia481 I am going to move the color update into the `Handler(Looper.getMainLooper()).post` although I would have preferred the `runOnUiThread` approach. Any thoughts?

That sounds right!
It's exactly what is the OS is complaining about:    
` Caused by: android.view.ViewRootImpl$CalledFromWrongThreadException: Only the original thread that created a view hierarchy can touch its views.`

mcarare:
bors r=csadilek, Amejia481

Issue #4910, opened 2019-10-31
Use notification grouping for downloads


sblatz:
Per @NotWoods' suggestion, we should consider using notification grouping for our download manager. This would also address us not using a `startForeground` which could lead to us getting killed under memory pressure.

 https://developer.android.com/training/notify-user/group

Amejia481:
We covered this work on https://github.com/mozilla-mobile/android-components/pull/6954

PR #6954, opened 2020-05-12
Closes #6893 Fulfill the foreground service contract on AbstractFetchDownloadService

### LINKED ISSUE #6893: Fulfill the foreground service contract on AbstractFetchDownloadService
For `AbstractFetchDownloadService` to be considered as a [foreground service](https://developer.android.com/guide/components/services#Foreground) it needs to call `startForeground()` and provides an ongoing notification this signal Android that this task  isn't a candidate for the system to kill when low on memory, avoiding issues like https://github.com/mozilla-mobile/fenix/issues/9884

Amejia481:
For fulfilling the Android foreground service api, we need to provide a notification, this notification will be marked as ongoing by the OS, and we won't be able to remove it until we are not longer a foreground a service by calling `stopForeground`. For this reason, I used two approaches:

**Devices that support notification group:**  [Demo 🎮 ](https://drive.google.com/file/d/1xaGh4qx_3913t4s1GenvIJphRFWNQvWP/view?usp=sharing)

A separate notification (a summary of all notification) which will be the foreground notification, it will be always present until we don't have more active downloads.

**Devices that DO NOT support notification group:**  [Demo 🎮  (Tested on Android 5.1.1) ](https://drive.google.com/file/d/10f26Ggh5uXVn5gCbrCBp3o5s2eNoAbzU/view?usp=sharing)

Set the latest active notification as the foreground notification and  keep changing it to the latest active download.

**Note:**
To reduce battery consumption when all the downloads notifications have been dismissed we stop the service.

Related issues https://github.com/mozilla-mobile/android-components/issues/4910

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
+++ b/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
@@ -7,6 +7,7 @@ package mozilla.components.feature.downloads
 import android.annotation.TargetApi
 import android.app.DownloadManager.ACTION_DOWNLOAD_COMPLETE
 import android.app.DownloadManager.EXTRA_DOWNLOAD_ID
+import android.app.Notification
 import android.app.Service
 import android.content.ActivityNotFoundException
 import android.content.BroadcastReceiver
@@ -45,6 +46,12 @@ import mozilla.components.concept.fetch.Headers.Names.CONTENT_RANGE
 import mozilla.components.concept.fetch.Headers.Names.RANGE
 import mozilla.components.concept.fetch.MutableHeaders
 import mozilla.components.concept.fetch.Request
+import mozilla.components.feature.downloads.AbstractFetchDownloadService.DownloadJobStatus.CANCELLED
+import mozilla.components.feature.downloads.AbstractFetchDownloadService.DownloadJobStatus.COMPLETED
+import mozilla.components.feature.downloads.AbstractFetchDownloadService.DownloadJobStatus.ACTIVE
+import mozilla.components.feature.downloads.AbstractFetchDownloadService.DownloadJobStatus.FAILED
+import mozilla.components.feature.downloads.AbstractFetchDownloadService.DownloadJobStatus.PAUSED
+import mozilla.components.feature.downloads.DownloadNotification.NOTIFICATION_DOWNLOAD_GROUP_ID
 import mozilla.components.feature.downloads.ext.addCompletedDownload
 import mozilla.components.feature.downloads.ext.getDownloadExtra
 import mozilla.components.feature.downloads.ext.withResponse
@@ -78,6 +85,8 @@ abstract class AbstractFetchDownloadService : Service() {
     internal val broadcastManager by lazy { LocalBroadcastManager.getInstance(this) }
     @VisibleForTesting
     internal val context: Context get() = this
+    @VisibleForTesting
+    internal var compatForegroundNotificationId: Int = COMPAT_DEFAULT_FOREGROUND_ID
 
     internal var downloadJobs = mutableMapOf<Long, DownloadJobState>()
 
@@ -146,27 +155,25 @@ abstract class AbstractFetchDownloadService : Service() {
 
                 when (intent.action) {
                     ACTION_PAUSE -> {
-                        setDownloadJobStatus(currentDownloadJobState, DownloadJobStatus.PAUSED)
+                        setDownloadJobStatus(currentDownloadJobState, PAUSED)
                         currentDownloadJobState.job?.cancel()
                         emitNotificationPauseFact()
                     }
 
                     ACTION_RESUME -> {
-                        setDownloadJobStatus(currentDownloadJobState, DownloadJobStatus.ACTIVE)
+                        setDownloadJobStatus(currentDownloadJobState, ACTIVE)
 
                         currentDownloadJobState.job = CoroutineScope(IO).launch {
-                            startDownloadJob(downloadId)
+                            startDownloadJob(currentDownloadJobState)
                         }
 
                         emitNotificationResumeFact()
                     }
 
                     ACTION_CANCEL -> {
-                        NotificationManagerCompat.from(context).cancel(
-                            currentDownloadJobState.foregroundServiceId
-                        )
+                        removeNotification(context, currentDownloadJobState)
                         currentDownloadJobState.lastNotificationUpdate = System.currentTimeMillis()
-                        setDownloadJobStatus(currentDownloadJobState, DownloadJobStatus.CANCELLED)
+                        setDownloadJobStatus(currentDownloadJobState, CANCELLED)
                         currentDownloadJobState.job?.cancel()
 
                         currentDownloadJobState.job = CoroutineScope(IO).launch {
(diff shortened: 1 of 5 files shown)


components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -223,14 +235,25 @@ abstract class AbstractFetchDownloadService : Service() {
         val foregroundServiceId = downloadJobs[download.id]?.foregroundServiceId ?: Random.nextInt()
 
         // Create a new job and add it, with its downloadState to the map
-        downloadJobs[download.id] = DownloadJobState(
+        val downloadJobState = DownloadJobState(
```
NotWoods, line 238: Since this class is getting pretty big, we should probably move `DownloadJobState` to its own file

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -372,9 +384,134 @@ abstract class AbstractFetchDownloadService : Service() {
         context.unregisterReceiver(broadcastReceiver)
     }
 
+    @VisibleForTesting
```
NotWoods, line 387: If this function is `open` should we make it `PROTECTED` instead of the default `PRIVATE`?

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -223,14 +235,25 @@ abstract class AbstractFetchDownloadService : Service() {
...
+
+        setForegroundNotification(downloadJobState)
+
+        downloadJobs[download.id] = downloadJobState
+
+        // We are canceling the notificationUpdateScope, because we don't want to keep it
+        // running when we don't have more downloads to process. As a result, we need to
+        // create a new one when there are more downloads to process and the scope is canceled
+        if (!notificationUpdateScope.isActive) {
+            notificationUpdateScope = MainScope()
```
pocmo, line 256: I haven't read all the code yet, but here I do not understand why we need to create new scopes instead of having one scope that is tied to the lifetime of the service - because like the notification/service it should only be around if there's a download? 🤔

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -372,9 +384,134 @@ abstract class AbstractFetchDownloadService : Service() {
...
     }
 
+    @VisibleForTesting
+    internal open fun removeDownloadJob(downloadJobState: DownloadJobState) {
+        downloadJobs.remove(downloadJobState.state.id)
+        if (downloadJobs.isEmpty()) {
+            stopForeground(false)
+            NotificationManagerCompat.from(context).cancel(getForegroundId())
+            clearAllDownloadsNotificationsAndJobs()
+            stopSelf()
```
pocmo, line 394: Won`t stopping the service already run through `clearAllDownloadsNotificationsAndJobs()` when the service gets to the destroy step? 🤔 I wonder if calling this here an additional times has made this code more complicated since now `clearAllDownloadsNotificationsAndJobs()` has two responsibilities as previously it was only handling clean up when the service ends.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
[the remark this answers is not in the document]
Amejia481: I thought not active meant that it was already cancel. Is that correct?

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadNotification.kt
```diff
@@ -126,9 +161,19 @@ internal object DownloadNotification {
...
+            .setDeleteIntent(createDismissPendingIntent(context, downloadState.id))
+            .setCompatGroup(NOTIFICATION_GROUP_KEY)
             .build()
     }
 
+    @VisibleForTesting
+    internal fun getDownloadSummary(context: Context, notifications: List<DownloadJobState>): String {
+        val summaryList = notifications.take(2)
+        return summaryList.joinToString("\n") { downloadState ->
+            "${downloadState.state.fileName} ${downloadState.getStatusDescription(context)}"
```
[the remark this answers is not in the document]
Amejia481: 👍

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
Amejia481: That make sense, I will try to tied it to the service's lifecycle.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
Amejia481: Thanks, for spotting this. As I changed the structure of the code,  I do not need neither `open`, `PROTECTED` nor `PRIVATE` for testing.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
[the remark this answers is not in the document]
Amejia481: As I changed the structure of the code, we don't need to recreate the scope anymore.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
Amejia481: I think we can do this in follow up, as it will cause the diff more difficult to be reviewed. What do you think?

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
Amejia481: You were right, after updating the code and doing the cleanup only when the service get destroyed made the code more easier to understand/maintain 👍

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
Amejia481: Now `notificationUpdateScope` is tied to the same scope as the service 👍

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
NotWoods: Sounds good to me!

Amejia481:
@pocmo  this is ready for review :)

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -260,74 +272,60 @@ abstract class AbstractFetchDownloadService : Service() {
...
             }
+        }
+    }
 
-            notification?.let {
-                NotificationManagerCompat.from(context).notify(
-                    download.foregroundServiceId,
-                    it
-                )
+    /***
```
pocmo, line 286: nit: Tripple star 🙃

Amejia481: Fixing it 😄 
image

Amejia481:
> Looks good to me.
> 
> My only recommendation would be to eventually (not in this PR, so ✅) consider moving the notification code to its own class that we delegate to - so that we do not blow up the service class that much. Maybe we could even have two classes (depending on API version) with a shared interface. If we then then create an instance based on the API level then we wouldn't need all those API level checks in all the methods.

Agree 👍 !

Amejia481:
bors r=notWoods,pocmo

Amejia481:
bors retry

PR #6956, opened 2020-05-12
Close #6953: Report native fatal crash to Sentry as level FATAL

### LINKED ISSUE #6953: Native fatal crashes should be reported to Sentry as level fatal
Currently we're reporting all native code crashes as level info to Sentry.  This is incorrect.  The level should be fatal when the crash is fatal, error when the crash is non fatal.

rocketsroger:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/lib/crash/src/main/java/mozilla/components/lib/crash/service/SentryService.kt
+++ b/components/lib/crash/src/main/java/mozilla/components/lib/crash/service/SentryService.kt
@@ -90,13 +90,18 @@ class SentryService(
 
     override fun report(crash: Crash.NativeCodeCrash): String? {
         if (sendEventForNativeCrashes) {
+            val level = when (crash.isFatal) {
+                true -> Event.Level.FATAL
+                else -> Event.Level.ERROR
+            }
+
             crash.breadcrumbs.forEach {
                 client.context.recordBreadcrumb(it.toSentryBreadcrumb())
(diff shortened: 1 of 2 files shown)


components/lib/crash/src/main/java/mozilla/components/lib/crash/service/SentryService.kt
```diff
@@ -90,13 +90,18 @@ class SentryService(
...
+            }
+
             crash.breadcrumbs.forEach {
                 client.context.recordBreadcrumb(it.toSentryBreadcrumb())
             }
 
             val eventBuilder = EventBuilder()
                 .withMessage(createMessage(crash))
-                .withLevel(Event.Level.INFO)
+                .withLevel(level)
```
pocmo, line 104: The initial idea here was that this is just a "log message" as opposed to an actual crash report that we can act on. But I guess we can also send it with "error" level if you think that is more helpful.

rocketsroger: I agree.  But there are some issues where other teams thought native crashes are not actual crashes.  Because caught exceptions are also sent as INFO.  This will make it easier to spot real crashes in Sentry. 👍

rocketsroger:
bors r=pocmo

PR #6961, opened 2020-05-12
Closes #6862: Migrate feature-accounts to browser-state

### LINKED ISSUE #6862: Migrate feature-accounts to browser-state
This is really just the `FxaWebChannelFeature` which we can migrate similar to what we did when migrating `ReaderViewFeature`.

FxR wants to re-use this feature and it would be a lot easier without the `SessionManager` dependency.

csadilek:
For the FxA web channel, we don't need to connect to the content script if tabs are opened in the background (e.g. via long press->open in new tab) so I removed that code. The port is disconnected automatically when the window is "unloaded".

The actual change here is minimal, it's mostly the tests, which I refactored to reduce duplication.

Tested manually in Fenix as well.

### DIFF
--- a/components/feature/accounts/build.gradle
+++ b/components/feature/accounts/build.gradle
@@ -21,13 +21,19 @@ android {
     }
 }
 
+tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
+    kotlinOptions.freeCompilerArgs += [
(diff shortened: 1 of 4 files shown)


csadilek:
Fenix PR: https://github.com/mozilla-mobile/fenix/pull/10613

csadilek:
bors r=grigoryk

PR #6395, opened 2020-03-26
Closes #6354 - Take Lazy references to storage in FennecMigrator

grigoryk:
This is a follow-up to https://github.com/mozilla-mobile/android-components/issues/6304, in which I forget that FennecMigrator exists.

---

### Pull Request checklist

- [ ] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/android-components/blob/master/docs/changelog.md) or does not need one
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features

### After merge
- [ ] **Milestone**: Make sure issues closed by this pull request are added to the [milestone](https://github.com/mozilla-mobile/android-components/milestones) of the version currently in development.
- [ ] **Breaking Changes**: If this is a breaking change, please push a draft PR on [Reference Browser](https://github.com/mozilla-mobile/reference-browser) to address the breaking issues.

PR #6964, opened 2020-05-13
Initialize accountManager on the main thread during migration

grigoryk:
Fenix bug for this - https://github.com/mozilla-mobile/fenix/issues/10432

This fixes fallout from #6395 in which
FennecMigrator changed to accept a Lazy-wrapped reference to the account manager.

This was done so that during regular (non-migrating) startup, a simple act of building a no-op
migrator will not cause account manager initialization.

However, during a migrating startup this creates a race to initialize account manager.
Two participants in this race are:
1) FenixApplication running expensive operations (such as storage init, account manager init)
after visual completeness. However, currently that just entails a 5 second delay from `onStart`.
2) FxA migration, which runs roughly at the mid-point of the migration.

Account manager currently can't be initialized on a background thread (or it will crash).
FxA migration runs on a background thread, so if (2) wins (that is, migrations before FxA migration take
less than 5 seconds), then migration will crash.

This patch adds an eager initialization right before we kick-off migrations but are still on the main thread,
iff an FxA migration will be executed.

---

### Pull Request checklist

### After merge


components/lib/crash/docs/metrics.md
```diff
@@ -20,7 +20,7 @@ The following metrics are added to the ping:
...
+| crash_metrics.crash_count |[labeled_counter](https://mozilla.github.io/glean/book/user/metrics/labeled_counters.html) |Counts the number of crashes that occur in the application. This measures only the counts of each crash in association with the labeled type of the crash. The labels correspond to the types of crashes handled by lib-crash. Deprecated: `native_code_crash` replaced by `fatal_native_code_crash` and `nonfatal_native_code_crash`  |[1](https://bugzilla.mozilla.org/show_bug.cgi?id=1553935#c3), [2](https://github.com/mozilla-mobile/android-components/pull/5700#pullrequestreview-347721248)|<ul><li>uncaught_exception</li><li>caught_exception</li><li>fatal_native_code_crash</li><li>nonfatal_native_code_crash</li></ul>|never |
```
pocmo, line 23: Not related, but I'll just merge this here too. :)

PR #6975, opened 2020-05-13
Update lifecycle version

NotWoods:
The new version [splits the dependency up](https://developer.android.com/jetpack/androidx/releases/lifecycle) into separate libraries to use.

---

### Pull Request checklist

### After merge


components/feature/tab-collections/build.gradle
```diff
@@ -49,8 +49,7 @@ dependencies {
     implementation Dependencies.kotlin_coroutines
 
     implementation Dependencies.androidx_paging
-    implementation Dependencies.androidx_lifecycle_extensions
-    kapt Dependencies.androidx_lifecycle_compiler
+    implementation Dependencies.androidx_lifecycle_livedata
```
psymoon, line 52: 🤔 What do we need `livedata` dependency for?

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -90,6 +90,10 @@ object Dependencies {
     const val androidx_core_ktx = "androidx.core:core-ktx:${Versions.AndroidX.core}"
     const val androidx_fragment = "androidx.fragment:fragment:${Versions.AndroidX.fragment}"
     const val androidx_lifecycle_extensions = "androidx.lifecycle:lifecycle-extensions:${Versions.AndroidX.lifecycle}"
```
psymoon, line 92: Looks like this is deprecated, does this still need to be here?

components/support/webextensions/build.gradle
```diff
@@ -37,7 +37,6 @@ dependencies {
     implementation project(':browser-state')
     implementation project(':support-base')
     implementation project(':support-ktx')
-    implementation Dependencies.androidx_lifecycle_extensions
```
psymoon, line 40: I think this was added due to `WebExtensionPopupFeature`. Can we verify that works? (Sample browser -> install ublock -> click on ublock browser action)

components/feature/tab-collections/build.gradle
NotWoods: `androidx_lifecycle_extensions` had both LiveData and ViewModel and was replaced with two separate dependencies. `TabCollectionDao` currently uses `LiveData` to return some results so I left a dependency on `androidx_lifecycle_livedata`.

buildSrc/src/main/java/Dependencies.kt
NotWoods: No, I'll remove it

components/support/webextensions/build.gradle
NotWoods: Yes, it still works.

psymoon:
@NotWoods Looks like there are some dependency failures in the CI with `feature-share`:
https://firefoxci.taskcluster-artifacts.net/dP-x-rgMQBKTXo14owasPw/0/public/logs/live_backing.log

Do you also see this locally?

NotWoods:
I don't see the failure locally, but I'll try adding a livedata dependency and see if it helps.

NotWoods:
Actually, it seems like there are a lot of unneeded dependencies. I'll try pulling out some.

NotWoods:
bors r=psymoon

PR #6984, opened 2020-05-14
Bug 1635766 - Remove hard dependency on lib-httpurlconnection from service-experiments

travis79:
This removes the dependency on lib-httpurlconnection from service-experiments aka Mako

### DIFF
--- a/components/service/experiments/src/main/java/mozilla/components/service/experiments/Configuration.kt
+++ b/components/service/experiments/src/main/java/mozilla/components/service/experiments/Configuration.kt
@@ -5,7 +5,6 @@
 package mozilla.components.service.experiments
 
 import mozilla.components.concept.fetch.Client
-import mozilla.components.lib.fetch.httpurlconnection.HttpURLConnectionClient
 
 /**
  * The Configuration class describes how to configure Experiments.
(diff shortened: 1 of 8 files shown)


components/service/experiments/src/main/java/mozilla/components/service/experiments/Configuration.kt
```diff
@@ -16,6 +15,6 @@ import mozilla.components.lib.fetch.httpurlconnection.HttpURLConnectionClient
  * [ExperimentsUpdater.KINTO_ENDPOINT_PROD]
  */
 data class Configuration(
-    val httpClient: Lazy<Client> = lazy { HttpURLConnectionClient() },
```
jonalmeida, line 19: drive-by comment 🚗 : was this a lazy value before for perf reasons? We've had patterns like this before where we pass lazily initialized objects around so we don't hit them immediately and cause other initializations until we actually need to.

travis79: It was, as far as I can tell/remember.  In this case, we use the `Client` almost immediately during init so I'm not sure that lazy buys us a lot in this case.  Most of the initialization of experiments was moved off the main thread which is where the real performance bottleneck was.  If this comes up in perf testing, please let's flag it then and we can address it.

travis79: I guess if this were Lazy, it would help more if the configuration object was created some time before calling init, but if the config object is created right before or when calling initialize, then the Client is going to be used immediately to make a request for the latest experiments.

PR #6990, opened 2020-05-15
Closes #6985: Make sure suggestion with long title/desc doesn't freeze UI

### LINKED ISSUE #6985: [Bug] 8-second hang in nComputeLineBreaks when presenting a certain set of suggested URLs
I can reproduce this reliably on my device with my account history synced.

The pathology here looks similar to mozilla-mobile/fenix#1824, except I'm not sure whether there's an actual long URL involved, and if so, what it is.

Performance profile: https://perfht.ml/2WHW9ef

STR (for me):
* Open up Fenix
* type "pen" in the URL box, which offers some search suggestions, and history-based suggestions.
* Tap any of the suggestions (in the case of the profile, penny-arcade.com/comic). Alternatively, just scrolling around the suggestions also hangs.

Results:
* 8 second hang.
* Bottom half (or so) of the viewport is black.

### Device information
* Android device: Android 10
* Fenix version: Fenix Nightly 20200505094621

csadilek:
We're already truncating in the view itself, but if the string is extremely long e.g. a 30K bookmarklet the UI hangs (see #6985). So, with this, we truncate any excessive title/description to prevent this case.

### DIFF
--- a/components/browser/awesomebar/src/main/java/mozilla/components/browser/awesomebar/layout/DefaultSuggestionViewHolder.kt
+++ b/components/browser/awesomebar/src/main/java/mozilla/components/browser/awesomebar/layout/DefaultSuggestionViewHolder.kt
@@ -9,10 +9,20 @@ import android.view.View
 import android.view.ViewGroup
 import android.widget.ImageView
 import android.widget.TextView
+import androidx.annotation.VisibleForTesting
 import mozilla.components.browser.awesomebar.BrowserAwesomeBar
 import mozilla.components.browser.awesomebar.R
 import mozilla.components.browser.awesomebar.widget.FlowLayout
 import mozilla.components.concept.awesomebar.AwesomeBar
+
+// This is used for truncating title and description to prevent extreme cases
+// from slowing down UI rendering e.g. in case of a bookmarklet or a data URI.
+// So this is just to make the actual truncation in the view faster which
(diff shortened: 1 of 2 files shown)


components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlin/String.kt
```diff
@@ -133,3 +133,17 @@ fun String.isSameOriginAs(other: String): Boolean {
...
     return canonicalizeOrigin(this) == canonicalizeOrigin(other)
 }
+
+/**
+ * Truncates this [String] to the provided size.
+ *
+ * @param size to truncate to.
+ * @throws StringIndexOutOfBoundsException in case size is < 0.
+ */
+fun String.truncate(size: Int): String {
```
pocmo, line 143: Was looking around for an existing Kotlin function. They have so many helpers that I was wondering if they had one for that too. `take()` looks like it could be it?
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/take.html

csadilek: Oh, nice. Let me try and remove our own then :).

csadilek: Updated to switch to `take`.

csadilek:
bors r=pocmo

Issue #6991, opened 2020-05-15
Build correct crash report URL in SentryService


pocmo:
I accidentally landed the patch with hard-coded `sample-crash` in the URL:
https://github.com/mozilla-mobile/android-components/blob/master/components/lib/crash/src/main/java/mozilla/components/lib/crash/service/SentryService.kt#L48

PR #6992, opened 2020-05-15
Issue #6991: SentryService: Allow consumer to define sentry project URL.

pocmo:
I accidentally landed the patch with hard-coded sample-crash in the URL. :)

Instead we need the app to provide the URL to the right project (In Fenix it currently depends on the build type).


pocmo:
bors r=csadilek

Issue #6687, opened 2020-04-16
prompt.Choice[] cannot be cast to prompt.Choice


st3fan:
> https://sentry.prod.mozaws.net/operations/fenix-fennec-beta/issues/7843411

```
java.lang.ClassCastException: mozilla.components.concept.engine.prompt.Choice[] cannot be cast to mozilla.components.concept.engine.prompt.Choice
    at mozilla.components.feature.prompts.PromptFeature.onConfirm(PromptFeature.kt:25)
    at -$$LambdaGroup$js$CAs0Y7NWl2_rHUSA2p5MNFfDxg0.onClick(java-style lambda group:3)
    at androidx.appcompat.app.AlertController$ButtonHandler.handleMessage(AlertController.java:3)
    at android.os.Handler.dispatchMessage(Handler.java:102)
    at android.os.Looper.loop(Looper.java:154)
    at android.app.ActivityThread.main(ActivityThread.java:6316)
    at java.lang.reflect.Method.invoke(Method.java)
    at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:872)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:762)
```

rocketsroger:
Looks like the frequency of this is very low.  However, there are also other similar cases with very low frequency as well.  I'll catch the exception and send a crash reporting with breadcrumb.  Once we have the breadcrumb we can figure out which case is causing the crash.

rocketsroger:
I'll continue to monitor this for a week to see if the patch works.  If it works, I'll use the information to find the root cause.

kbrosnan:
Looks like this is fixed. Not seeing any crashes on 78 or 79.

PR #6995, opened 2020-05-15
For #6687: Add crash reporting information for ClassCastException in PromptFeature

rocketsroger:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt
+++ b/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt
@@ -312,47 +312,52 @@ class PromptFeature private constructor(
     @Suppress("UNCHECKED_CAST", "ComplexMethod")
     override fun onConfirm(sessionId: String, value: Any?) {
         store.consumePromptFrom(sessionId, activePrompt) {
-            when (it) {
-                is TimeSelection -> it.onConfirm(value as Date)
-                is Color -> it.onConfirm(value as String)
-                is Alert -> {
-                    val shouldNotShowMoreDialogs = value as Boolean
-                    promptAbuserDetector.userWantsMoreDialogs(!shouldNotShowMoreDialogs)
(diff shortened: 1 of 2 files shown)


components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt
```diff
@@ -312,47 +322,58 @@ class PromptFeature private constructor(
...
                     }
                 }
+            } catch (e: ClassCastException) {
+                crashReporting?.recordCrashBreadcrumb(
+                    Breadcrumb("PromptFeature onConsume cast failed",
+                        hashMapOf("class name" to " ${it.javaClass}"), "crash",
+                        Breadcrumb.Level.DEBUG, Breadcrumb.Type.NAVIGATION)
+                )
+
+                crashReporting?.submitCaughtException(e)
```
pocmo, line 376: I don't really like the idea of catching `ClassCastException` here every time since it will just hide errors while developing. Maybe we could just consider adding regular breadcrumbs to the feature (showing prompt, confirming, cancelling) that are helpful for understanding all kinds of errors?

rocketsroger: Sure, I'll add breadcrumb and throw the exception after.

pocmo: Oh, I was thinking about just generic breadcrumbs independent of the exception here.

But talking about the exception. Maybe you do not want a breadcrumb and instead just rethrow a new exception with a better message (and the original one as cause)?

rocketsroger: I am worried this is called too many times to add a generic breadcrumb without an exception.  Sure, I could modify the message.  But wouldn't that be similar to adding a breadcrumb since we will be try catching the exception and rethrowing?

rocketsroger:
bors r=pocmo

PR #6999, opened 2020-05-16
Use helper to change state of tab in BrowserStore

NotWoods:
This reduces some of the repeated code in `BrowserStore`, and additionally optimizes by stopping iteration once the correct tab list is found.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/BrowserStateReducer.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/BrowserStateReducer.kt
@@ -15,6 +15,9 @@ import mozilla.components.browser.state.action.TabListAction
 import mozilla.components.browser.state.action.TrackingProtectionAction
 import mozilla.components.browser.state.action.WebExtensionAction
 import mozilla.components.browser.state.state.BrowserState
+import mozilla.components.browser.state.state.CustomTabSessionState
+import mozilla.components.browser.state.state.SessionState
+import mozilla.components.browser.state.state.TabSessionState
 import mozilla.components.browser.state.store.BrowserStore
 import mozilla.components.lib.state.Action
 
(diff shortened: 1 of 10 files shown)


components/browser/state/src/main/java/mozilla/components/browser/state/reducer/BrowserStateReducer.kt
```diff
@@ -39,3 +42,38 @@ internal object BrowserStateReducer {
...
+ * @param update Returns a new version of the tab state. Must be the same class,
+ * preferably using [SessionState.copySession].
+ */
+@Suppress("Unchecked_Cast")
+internal fun BrowserState.updateTabState(
+    tabId: String,
+    update: (SessionState) -> SessionState
+): BrowserState {
+    val newTabs = tabs.updateTabs(tabId, update) as List<TabSessionState>?
+    if (newTabs != null) return copy(tabs = newTabs)
```
pocmo, line 58: I always prefer explicit blocks with `{}`. But since our tooling does not complain I will accept it. 🙈

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/BrowserStateReducer.kt
```diff
@@ -39,3 +42,38 @@ internal object BrowserStateReducer {
...
+
+    val newCustomTabs = customTabs.updateTabs(tabId, update) as List<CustomTabSessionState>?
+    if (newCustomTabs != null) return copy(customTabs = newCustomTabs)
+
+    return this
```
pocmo, line 63: I like that this no longer maps over two lists if not needed.

components/browser/state/src/main/java/mozilla/components/browser/state/state/SessionState.kt
```diff
@@ -24,4 +24,17 @@ interface SessionState {
     val engineState: EngineState
     val extensionState: Map<String, WebExtensionState>
     val contextId: String?
+
+    /**
+     * Copy the class and override some parameters.
+     */
+    @Suppress("LongParameterList")
+    fun copySession(
```
pocmo, line 32: I wish we could make this `internal` since we only need this inside the reducers. But I guess this only possible with an extension method that does instance checks and this is awkward as well.

nit: I feel like something like `createCopy()` would be a better name. Calling `copySession()` on `SessionState/TabSessionState/CustomTabSessionState` would make me wonder what Session I am copying where.

NotWoods:
bors r=pocmo

Issue #5217, opened 2019-12-02
Crash when downloading a large data URI


MarcLeclair:
## Steps to reproduce
1. Launch sample-browser.
2. Open a website that redirects using a data URI. The one I used was: https://mars.nasa.gov/participate/send-your-name/mars2020/certificate/599173796266
3. Click download

### Expected behavior
If the Data URI is an image, a dialogue should pop up to ask the user if they want to download the image. 
### Actual behavior
Application crashes with a TransactionTooLargeException
### Device information

* Android device: **Google Pixel 3(Android 10).**

### Notes
The application redirects because ( from what I can gather), we check for the "new" url hostname versus our session, which in this case will be different. From there, we redirect with an intent. In the example above, the picture on the NASA website is ~1.2MB which will lead to the application trying to send an intent that's too big (data parcel size 3473380 bytes in my case).

jonalmeida:
cc: @rocketsroger for reference.

csadilek:
Removing the app-link label. This is no longer redirecting, but crashing because we're storing the data URI in the download object which passed as intent extras:

```
 java.lang.RuntimeException: android.os.TransactionTooLargeException: data parcel size 1486032 bytes
        at android.app.ContextImpl.startServiceCommon(ContextImpl.java:1621)
        at android.app.ContextImpl.startService(ContextImpl.java:1571)
        at android.content.ContextWrapper.startService(ContextWrapper.java:669)
        at mozilla.components.feature.downloads.manager.FetchDownloadManager.download(FetchDownloadManager.kt:8)
        at mozilla.components.feature.downloads.DownloadsFeature.access$processDownload(DownloadsFeature.kt:16)
        at mozilla.components.feature.downloads.DownloadsFeature$start$2$invokeSuspend$$inlined$collect$1.emit(Collect.kt:4)
        at mozilla.components.support.ktx.kotlinx.coroutines.flow.FlowKt$ifChanged$$inlined$filter$1$2.emit(Collect.kt:7)
        at mozilla.components.feature.downloads.DownloadsFeature$start$2$invokeSuspend$$inlined$mapNotNull$1$2.emit(Collect.kt:6)
        at kotlinx.coroutines.CoroutineScopeKt.emitAllImpl$FlowKt__ChannelsKt(CoroutineScope.kt:14)
        at kotlinx.coroutines.flow.FlowKt__ChannelsKt$emitAllImpl$1.invokeSuspend(Unknown Source:11)
[...]

PR #7028, opened 2020-05-19
Issue #5217: Skip check to open about/data/js URLs in external apps

csadilek:
This prevents looking up external apps for `data:`, `javascript:` and `about:` URLs. We don't need to look for external apps in these cases and the former actually causes crashes when the URL is too large `TransactionTooLargeException` as the intent contains the entire URL.


csadilek:
bors r=rocketsroger

PR #7031, opened 2020-05-19
Closes #5249: Long URI in the address bar will hang the UI

### LINKED ISSUE #5249: Pasting/getting redirected to extremely-long data URI in the address bar will hang the UI
## Steps to reproduce
1. Launch sample-browser.
2. Open a website that redirects using a Data URI* (for an example, please refer to this test page I made when I was testing: https://github.com/MarcLeclair/DataURIStaticTestPage if you want to test using a data:application/json;base64 and get redirected) 
3. Click on the element that redirects to Data URI

OR

1. Launch sample-browser
2. paste long data URI* in the address bar ( keep in mind that copying something that's bigger than the allowed clipboard size might crash the application on an emulator). 

*: If you want to see noticeable lag, using a URL > 25k chars

### Expected behavior
The load time shouldn't be different 
### Actual behavior
The webpage loads in ~156 seconds when the data URI is extremely long (The example using my test webpage above  was 65k characters)
### Device information

* Android device: **Google Pixel 1(Android 10).** and **Moto G5 plus(Android 7)**

### Notes

It was loading all fine in Fennec because we actually limited the URL length in Browser.js  as seen here: 

 `let message = {
      [...]
      uri: truncate(fixedURI.displaySpec, MAX_URI_LENGTH),
     [...]
    };`

where the MAX_URI_LENGTH was 25000

This image below shows the stack call when loading the 65k character URL.
image

I'd like to highlight a few  possible solutions:

1. We could cache the full URL and use `ellipsize()` from android to set the text in the TextView. However, to allow users to copy the  full URL (if the devices' clipboard allows it), the cache version can be copied to the clipboard by setting a `OnPrimaryClipChangedListener` on the edit state of the toolbar. 

2. Since the issue is the StaticLayout builder, we could do all the lifting on our own and try to work through the `breakstrategy` set by the layout itself. However, that's a lot of work for such a small edge case

I've also attached the trace as a zip if anyway is interested in seeing it. To open it, just open it with android studio after unzipping it.

[java-trace-sample-uri-motog5-dec4.trace.zip](https://github.com/mozilla-mobile/android-components/files/3934043/java-trace-sample-uri-motog5-dec4.trace.zip)

csadilek:
We've had two cases now where this was a problem. A bookmarklet (see https://github.com/mozilla-mobile/android-components/issues/6985) and a data URI (https://github.com/mozilla-mobile/android-components/issues/5249).

This fix is similar to the one we landed last week for our Awesomebar (https://github.com/csadilek/android-components/commit/bba846d12eb12e7bb91e112aba3901c034057fde), but tapping on long URLs or copying them from the clipboard etc. would still cause the UI to hang. With data URIs I can reproduce seconds/minutes long hangs that are fixed with this.

Took this from Fennec (https://github.com/pocmo/fennec-graveyard/blob/master/mobile/android/chrome/content/browser.js#L4308), and applied to both display and edit toolbar (otherwise the UI would hang once the toolbar is in edit mode).

### DIFF
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
@@ -32,6 +32,11 @@ import mozilla.components.ui.autocomplete.InlineAutocompleteEditText
 import mozilla.components.ui.autocomplete.OnFilterListener
 import kotlin.coroutines.CoroutineContext
 
+// This is used for truncating URLs to prevent extreme cases from
+// slowing down UI rendering e.g. in case of a bookmarklet or a data URI.
+// https://github.com/mozilla-mobile/android-components/issues/5249
+const val MAX_URI_LENGTH = 25000
+
 /**
  * A customizable toolbar for browsers.
  *
@@ -99,7 +104,7 @@ class BrowserToolbar @JvmOverloads constructor(
             // We update the display toolbar immediately. We do not do that for the edit toolbar to not
             // mess with what the user is entering. Instead we will remember the value and update the
             // edit toolbar whenever we switch to it.
-            display.url = value
+            display.url = value.take(MAX_URI_LENGTH)
         }
 
     override var siteSecure: Toolbar.SiteSecurity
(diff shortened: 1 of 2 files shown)


csadilek:
I have a version of this where we only truncate `data:` and `javascript:` URLs explicitly, but I don't think that's worth it. The app will become unresponsive for the large URLs like this, and I don't quite see how that's safer either.

csadilek:
bors r=pocmo

PR #7033, opened 2020-05-20
Bug 1639405 - Remove build.gradle files from triggering toolchain task

jonalmeida:
---

We're already watching the `Dependencies.kt` file so the only change to do here is remove the build.gradle entries.

### After merge

### DIFF
--- a/taskcluster/ci/toolchain/android.yml
+++ b/taskcluster/ci/toolchain/android.yml
@@ -45,9 +45,7 @@ linux64-android-gradle-dependencies:
             - taskcluster/scripts/toolchain/android-gradle-dependencies.sh
             - taskcluster/scripts/toolchain/android-gradle-dependencies/**
             - buildSrc/src/main/java/Dependencies.kt
-            - '*.gradle'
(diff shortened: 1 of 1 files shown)


taskcluster/ci/toolchain/android.yml
```diff
@@ -45,9 +45,6 @@ linux64-android-gradle-dependencies:
             - taskcluster/scripts/toolchain/android-gradle-dependencies.sh
             - taskcluster/scripts/toolchain/android-gradle-dependencies/**
             - buildSrc/src/main/java/Dependencies.kt
-            - '*.gradle'
-            - components/**/build.gradle
-            - samples/**/build.gradle
```
pocmo, line 50: We should keep the root `build.gradle` file since that contains some plugin versions!

jonalmeida: Good point! Instead of the `*.gradle` then we can use the file directly as `build.gradle` so we don't watch other files unnecessarily.

jonalmeida:
bors r=JohanLorenzo

PR #7039, opened 2020-05-20
Issue #7037: CrashReporter: Log ids and URLs of crash reports when submitting crash reports.

### LINKED ISSUE #7037: Crash Reporter: Log crash ids and/or crash report URLs
Sometimes all you have is a device log. Especially for native code crashes it would be really helpful if we could get to the crash report from the log. So let's log the crash ids from the services we reported the crash too (socorro, sentry) and also the URL to the report if we have it.

pocmo:
Closes #7037.


pocmo:
Failures look network related. Let's try to land.

bors r=psymoon

PR #7044, opened 2020-05-20
Issue #2985: Add fallback in case Promise.any not available

csadilek:
This is handling `promise.any` not being available (current beta). The only part we miss is the optimization to let the faster operation of fetch|cache win, but that's fine for now.


csadilek:
bors r=jonalmeida

csadilek:
bors retry

PR #7045, opened 2020-05-20
Issue #7012: Add share option to crash list.

### LINKED ISSUE #7012: Add share option to crash list
Often you may not want to read the crash report on a phone on crash-stats/sentry, but instead share/copy all information to post it somewhere or open an issue. This is especially annoying with Sentry because it will always redirect you to a login page (if you are not logged in already) and that makes it impossible to copy the link for anyone that does not have Sentry access. So let's add a share option! :)

pocmo:
Closes #7012. This will make it easier to share crash reports.


pocmo:
>  Do you think we should add some tests? Or are we gong to do it in a follow up pr? :)

Yeah, I think we could consider adding some tests for the UI. But I also think it doesn't need to block this share functionality. I'll file a new issue! -> #7095 

bors r=Amejia481

PR #7048, opened 2020-05-20
For #7064: New error message to indicate we don't support opening Zip files

kglazko:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
+++ b/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
@@ -24,6 +24,7 @@ import android.os.IBinder
 import android.os.ParcelFileDescriptor
 import android.provider.MediaStore
 import android.widget.Toast
+import android.webkit.MimeTypeMap
 import androidx.annotation.GuardedBy
 import androidx.annotation.VisibleForTesting
 import androidx.core.app.NotificationManagerCompat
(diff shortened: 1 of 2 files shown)


components/feature/downloads/src/main/res/values/strings.xml
```diff
@@ -25,6 +25,9 @@
     <!-- Message that appears when the downloaded file could not be opened-->
     <string name="mozac_feature_downloads_could_not_open_file">Could not open file</string>
 
+    <!-- Message that appears when the downloaded file is a Zip file and Android doesn't support opening-->
+    <string name="mozac_feature_downloads_could_not_open_zip">Opening ZIP files not supported</string>
```
Amejia481, line 29: What do you think about just having one message and replacing **ZIP** for the extension of the not supported file?

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -196,10 +196,18 @@ abstract class AbstractFetchDownloadService : Service() {
                                 filePath = currentDownloadJobState.state.filePath,
                                 contentType = currentDownloadJobState.state.contentType
                             )) {
-                            Toast.makeText(applicationContext,
-                                applicationContext.getString(R.string.mozac_feature_downloads_could_not_open_file),
-                                Toast.LENGTH_SHORT
-                            ).show()
+                            val zipExt = ".zip"
+                            if (currentDownloadJobState.state.filePath.endsWith(zipExt)) {
```
Amejia481, line 200: I thought we were looking to have a generic message where we replace the file extension for all non supported files. Instead of having one specific message for zips. What do you think?

`<string name="mozac_feature_downloads_open_not_supported">Opening %1$s files not supported</string>`


 ```kotlin
val errorMessage = applicationContext.getString(R.string.mozac_feature_downloads_open_not_supported, fileExtension)

Toast.makeText(applicationContext,errorMessage,Toast.LENGTH_SHORT).show()

```

kglazko: I see, I misunderstood! My concern here is that some cases like .zip are not supported.. and potentially others like .rar as well. But, what about the case where we download an image file but for some reason it is corrupted or can't be opened, then we will display a message saying 'Opening PNG not supported'- when we clearly do. I wanted to isolate the cases that are likely caused by Android's own issues with their intent such as .Zip into their own message so that the user can differentiate between some kind of download failure and an inability to support opening files.

PR #7065, opened 2020-05-21
Update room version

NotWoods:
Also switch to the KTX variant of Room, so we can use suspend and Flow!


NotWoods:
bors r=psymoon

PR #7072, opened 2020-05-22
Fix EditToolbar keyEvent handling

### LINKED ISSUE #3600: Fix EditToolbar keyEvent handling
The [EditToolbar](https://github.com/mozilla-mobile/android-components/blob/40653c6f1df71f0c4990c04a4c334bf9b75fd28f/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/edit/EditToolbar.kt#L44) class applies the following `setOnDispatchKeyEventPreImeListener` function on the `urlView` text view. 
```Kotlin
setOnDispatchKeyEventPreImeListener { event ->
            if (event?.keyCode == KeyEvent.KEYCODE_BACK) {
                toolbar.onEditCancelled()
            }
            false
}
```
The method checks if the pressed key is the BACK key and if it is, it calls `toolbar.onEditCancelled()`. However, it doesn't check the event action resulting in `toolbar.onEditCancelled()` getting called twice (once for ACTION_DOWN and once for ACTION_UP). I might be mistaken but as far as I can see this makes no sense, it should only trigger `onEditCancelled` once. I would thus suggest another check that makes sure it only reacts to UP (or DOWN).

After the KEYCODE_BACK check, the function always returns `false`. Again, I don't know all the mechanisms in this library and it might be intentional, but I am very sure that this causes [this Fenix issue](https://github.com/mozilla-mobile/fenix/issues/3767). Because the function returns `false`, the Android runtime passes the action down the chain and in Fenix it ultimately triggers the `onBackPressed` function in [HomeActivity](https://github.com/mozilla-mobile/fenix/blob/55f5c5d049fc4a93c3dbd65fb8d87a7f0832304a/app/src/main/java/org/mozilla/fenix/HomeActivity.kt#L164). This causes the app to close. I think it makes sense to let it return `true` because the event is already processed through the `onEditCancelled` method.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-2944)

miDeb:
Closes #3600.

Return true in `setOnDispatchKeyEventPreImeListener` when the back button was pressed. This prevents handling the back button press twice in the case that the keyboard has already been closed.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/edit/EditToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/edit/EditToolbar.kt
@@ -49,7 +49,7 @@ private const val AUTOCOMPLETE_QUERY_THREADS = 3
  * - actions: Optional action icons injected by other components (e.g. barcode scanner)
  * - exit: Button that switches back to display mode or invoke an app-defined callback.
  */
-@Suppress("TooManyFunctions")
+@Suppress("TooManyFunctions", "LargeClass")
 class EditToolbar internal constructor(
     context: Context,
     private val toolbar: BrowserToolbar,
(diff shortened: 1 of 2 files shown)


Scripterr:
Hi, can someone please review this PR, this bug is really annoying and results in data loss if I accidentally press back in the url bar while editing or viewing url, it load previous page and the form or loaded data is lost and user have to refill the information and users valuable time and internet data is wasted.

miDeb:
I guess https://github.com/mozilla-mobile/fenix/pull/11871 will fix this.

Issue #7096, opened 2020-05-25
URLs containing double dash "--" should be seen as valid


codrut-topliceanu:
Needed to fix https://github.com/mozilla-mobile/fenix/issues/10747

Currently `isURLLenient` from `URLStringUtils.kt` does not allow for URLs to contain double dash ("--"). This results in an URL, like the one mentioned in https://github.com/mozilla-mobile/fenix/issues/10747 , as being a search term in Fenix.

PR #7097, opened 2020-05-25
For #7096 - Allow for URL to contain double dash

codrut-topliceanu:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/support/utils/src/main/java/mozilla/components/support/utils/URLStringUtils.kt
+++ b/components/support/utils/src/main/java/mozilla/components/support/utils/URLStringUtils.kt
@@ -59,9 +59,9 @@ object URLStringUtils {
 
     private val isURLLenient by lazy {
         // Be lenient about what is classified as potentially a URL.
-        // (\w+-)*\w+(://[/]*|:|\.)(\w+-)*\w+([\S&&[^\w-]]\S*)?
+        // (\w+-+)*\w+(://[/]*|:|\.)(\w+-+)*\w+([\S&&[^\w-]]\S*)?
         // -------                 -------
-        // 0 or more pairs of consecutive word letters followed by a dash
+        // 0 or more pairs of consecutive word letters or dashes
         //        ---                     ---
         // followed by at least a single word letter.
         // -----------             ----------
@@ -91,7 +91,7 @@ object URLStringUtils {
         // http://www-.com
         // www.c-c-
         // 3-3
-        Pattern.compile("^\\s*(\\w+-)*\\w+(://[/]*|:|\\.)(\\w+-)*\\w+([\\S&&[^\\w-]]\\S*)?\\s*$", flags)
+        Pattern.compile("^\\s*(\\w+-+)*\\w+(://[/]*|:|\\.)(\\w+-+)*\\w+([\\S&&[^\\w-]]\\S*)?\\s*$", flags)
     }
 
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
--- a/components/support/utils/src/test/java/mozilla/components/support/utils/URLStringUtilsTest.kt
+++ b/components/support/utils/src/test/java/mozilla/components/support/utils/URLStringUtilsTest.kt
@@ -114,6 +114,9 @@ class URLStringUtilsTest {
         assertFalse(isURLLike("http://www-.com"))
         assertFalse(isURLLike("www.c-c-  "))
         assertFalse(isURLLike("3-3 "))
+
+        // Examples from issues
+        assertTrue(isURLLike("https://abc--cba.com/")) // #7096
     }
 
     @Test
@@ -160,6 +163,9 @@ class URLStringUtilsTest {
         assertTrue(isSearchTerm("http://www-.com"))
         assertTrue(isSearchTerm("www.c-c-  "))
         assertTrue(isSearchTerm("3-3 "))
+
+        // Examples from issues
+        assertFalse(isSearchTerm("https://abc--cba.com/")) // #7096
     }
 
     @Test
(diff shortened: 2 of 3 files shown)


components/support/utils/src/main/java/mozilla/components/support/utils/URLStringUtils.kt
```diff
@@ -90,8 +90,9 @@ object URLStringUtils {
         // -x.com
         // http://www-.com
         // www.c-c-
+        // https://disposable-wide4--solidsoft-blog-disposable.netlify.app/ (fenix/issues/10747)
```
Amejia481, line 93: Let's not include the Fenix reference, as we already have a reference to an Android component issue in the commit message.

This `url` is added below the "There are some examples of non-URLs according to this pattern:",
Instead of adding the whole `url`, I think it would be better if we describe, what this new quantifier is doing.

Like the above comment:
```

 0 or more pairs of consecutive word letters followed by a dash
        //        ---                     ---
        // followed by at least a single word letter.
        // -----------             ----------
        // Combined, that means "w", "w-w", "w-w-w", etc match, but "w-", "w-w-", "w-w-w-" do not.
        //          --------------
        // That surrounds :, :// or .
        //                                                    -
        // At the end, there may be an optional
        //                                    ------------
        // non-word, non-- but still non-space character (e.g., ':', '/', '.', '?' but not 'a', '-', '\t')
```

components/support/utils/src/main/java/mozilla/components/support/utils/URLStringUtils.kt
```diff
@@ -90,8 +90,9 @@ object URLStringUtils {
...
         // 3-3
-        Pattern.compile("^\\s*(\\w+-)*\\w+(://[/]*|:|\\.)(\\w+-)*\\w+([\\S&&[^\\w-]]\\S*)?\\s*$", flags)
+        Pattern.compile("^\\s*(\\w+-)*\\w+(://[/]*|:|\\.)(\\w+-+)*\\w+([\\S&&[^\\w-]]\\S*)?\\s*$", flags)
```
csadilek, line 95: We should also update the first `(\\w+-)` group here to `(\\w+-+)`, same as the second one, because now this is a valid URL:

`http://abc-abc--abc-abc.com`

but this isn't

`abc-abc--abc-abc.com`

Seem my comment above :)

So, the regex becomes:
```suggestion
        Pattern.compile("^\\s*(\\w+-+)*\\w+(://[/]*|:|\\.)(\\w+-+)*\\w+([\\S&&[^\\w-]]\\S*)?\\s*$", flags)
```

components/support/utils/src/main/java/mozilla/components/support/utils/URLStringUtils.kt
csadilek: Yes, the regex in the comment above needs an update to reflect the change:

`(\w+-+)*\w+(://[/]*|:|\.)(\w+-+)*\w+([\S&&[^\w-]]\S*)?`  (both groups updated -> see comment below)

`// 0 or more pairs of consecutive word letters followed by a dash` -> This should be come `// 0 or more pairs of consecutive word letters or dashes`

components/support/utils/src/test/java/mozilla/components/support/utils/URLStringUtilsTest.kt
```diff
@@ -114,6 +114,10 @@ class URLStringUtilsTest {
         assertFalse(isURLLike("http://www-.com"))
         assertFalse(isURLLike("www.c-c-  "))
         assertFalse(isURLLike("3-3 "))
+
+        // Examples from issues
+        assertTrue(isURLLike(
+            "https://disposable-wide4--solidsoft-blog-disposable.netlify.app/")) // fenix/10747
```
csadilek, line 120: Let's remove the fenix ticket references. I'd use a generic example and link to the A-C issue.

components/support/utils/src/test/java/mozilla/components/support/utils/URLStringUtilsTest.kt
```diff
@@ -160,6 +164,10 @@ class URLStringUtilsTest {
         assertTrue(isSearchTerm("http://www-.com"))
         assertTrue(isSearchTerm("www.c-c-  "))
         assertTrue(isSearchTerm("3-3 "))
+
+        // Examples from issues
+        assertFalse(isSearchTerm(
+            "https://disposable-wide4--solidsoft-blog-disposable.netlify.app/")) // fenix/10747
```
csadilek, line 170: Let's remove the fenix ticket references. I'd use a generic example and link to the A-C issue.

codrut-topliceanu:
Hey, thanks for taking a look at this :D
I've completed the requested changes.

components/support/utils/src/main/java/mozilla/components/support/utils/URLStringUtils.kt
```diff
@@ -61,7 +61,7 @@ object URLStringUtils {
         // Be lenient about what is classified as potentially a URL.
         // (\w+-)*\w+(://[/]*|:|\.)(\w+-)*\w+([\S&&[^\w-]]\S*)?
```
csadilek, line 62: @codrut-topliceanu The regex in this comment still needs to be updated to `(\w+-+)` for the two groups otherwise it's confusing to read the new explanation :).

codrut-topliceanu: Darn, you're right. Good eye! Sorry I missed it, fixing it now.

csadilek: 👍

Issue #7011, opened 2020-05-18
Memory trimming to aggresive?


pocmo:
Currently we drop engine sessions (and keep their state) when the trimming level is at least `TRIM_MEMORY_RUNNING_CRITICAL` (15). That sounded good, but `TRIM_MEMORY_UI_HIDDEN` (20) is a higher level and happens whenever the app goes to the background. And that seems very aggressive when switching between apps. Either we should ignore "ui hidden" or pick a higher level? 🤔

pocmo:
The Android documentation confirms:

```
The values provided by onTrimMemory(int) do not represent a single linear progression of memory limits, but provide you different types of clues about memory availability: [..]
```

PR #7100, opened 2020-05-25
Issue #7011: Correctly evaluate memory trimming levels.

pocmo:
Previously we assumed the value passed to onTrimMemory() represents a single linear progression of memory limits. But this
is not the case. We need to look at the values individually.

See:
https://developer.android.com/reference/android/content/ComponentCallbacks2


pocmo:
> This makes sense to me. This is an useful API - I wonder if memory trimming evaluation could be refactored into a centralized haven 🤔.

Do you have an idea how this could look like? Right now it looks like every component may have its own opinion on what levels are interesting to it.

bors r=psymoon

Issue #6879, opened 2020-05-05
Using a syncable storage in the correct state can lead to multiple syncs


jonalmeida:
### Steps to reproduce
1. Do not have any references to the account manager that can sync first.
1. Add a SyncStatusObserver to the account manager to request remote tabs:
    ```kotlin
    accountManager.register(object: SyncStatusObserver {
      override fun onIdle() {
        remoteStorage.getTabs
      }
    }
    ```
2. Run:
    ```kotlin
      accountManager.syncNowAsync()
    ```
## Expected
 - According to the docs for `onIdle`:

    > Gets called at the end of a sync, after every configured syncable has been synchronized.

 - So therefore, `tabsStorage.getSyncedTabs()` should be safe to call here.

## Actual
 - It isn't safe to make that call here since the device constellation is not synchronized yet, requiring an `AccountObserver` and a `DeviceConstellationObserver` that leads to two tab syncs. One from `onDevicesUpdate` and a second from the following `onIdle`:

    ```kotlin
      class AccountObserver(
          private val observer: DeviceConstellationObserver,
          private val lifecycleOwner: LifecycleOwner
      ) : AccountObserver {
          override fun onAuthenticated(account: OAuthAccount, authType: AuthType) {
              account.deviceConstellation().registerDeviceObserver(observer, lifecycleOwner, false)
          }
      }

      class ConstellationObserver(
          private val tabsStorage: SyncedTabsStorage
      ) : DeviceConstellationObserver {
          override fun onDevicesUpdate(constellation: ConstellationState) {
              tabsStorage.getSyncedTabs()
          }
      }
    ```

**Note:** We can't remove the `SyncStatusObserver.onIdle` because otherwise we have to always check if we need to observe for an `OAuthAccount` directly on the `FxaAccountManager` or wait for the `AccountObserver.onAuthenticated` to observe. 

Daisy chaining these observers is quite tedious as well, so it might be worth considering a new API.

cc: @grigoryk

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-5860)

Issue #6878, opened 2020-05-05
Syncing with debounce enabled on the AccountManager does not always invoke onIdle


jonalmeida:
### Steps to reproduce

1. Add an observer on the Account Manager:
    ```kotlin
    accountManager.register(object: SyncStatusObserver {
      override fun onIdle() {
        Logger.d("Idle should be executed eventually")
      }
    }
    ```
2. Execute this a couple of times rapidly:
    ```kotlin
      accountManager.syncNowAsync(SyncReason.User, true)
    ```

### Expected
 - We eventually get `onIdle` executed.

### Actual
 - This doesn't always happen.

cc: @grigoryk

grigoryk:
Hmm, this is strange. Reading through the code, I don't immediately see why this would be the case. Worker performs a `isDebounced` check if this flag is set (checking if sync already happened recently), and short-circuits if so. There's nothing special about that short-circuit, it's treated just like any other sync worker result.

So what this makes me think is that there's something else is going on, e.g. not related to this flag, but perhaps a more general bug with handling of worker results.

jonalmeida:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1794210

Change performed by the [Move to Bugzilla add-on](https://addons.mozilla.org/en-US/firefox/addon/move-to-bugzilla/).

PR #7114, opened 2020-05-26
Close #6963: Sync account when SyncedTabsFeature is started

### LINKED ISSUE #6963: Use account sync to trigger a Sync Tabs refresh in feature start
Use `FxaAccountManager.syncNowAsync` to trigger a sync tabs refresh in the presenter instead of a direct sync tabs poll. Currently, we're using `controller.syncTabs()` which can fail for edge cases.

We should put this in the controller since we invoke this in the `onRefresh` call as well.

jonalmeida:
Instead of syncing tabs on start, we refresh the devices and sync the account. The allows our mechanism to sync tabs once that is complete to be triggered where we know we're in the most current state we can be in.

---

Closes #6963 

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/syncedtabs/src/main/java/mozilla/components/feature/syncedtabs/controller/DefaultController.kt
+++ b/components/feature/syncedtabs/src/main/java/mozilla/components/feature/syncedtabs/controller/DefaultController.kt
@@ -12,19 +12,23 @@ import mozilla.components.feature.syncedtabs.view.SyncedTabsView
 import mozilla.components.feature.syncedtabs.view.SyncedTabsView.ErrorType
 import mozilla.components.service.fxa.manager.FxaAccountManager
 import mozilla.components.service.fxa.manager.ext.withConstellation
+import mozilla.components.service.fxa.sync.SyncReason
 import kotlin.coroutines.CoroutineContext
 
 internal class DefaultController(
     override val provider: SyncedTabsProvider,
     override val accountManager: FxaAccountManager,
     override val view: SyncedTabsView,
-    private val coroutineContext: CoroutineContext
+    coroutineContext: CoroutineContext
 ) : SyncedTabsController {
 
-    override fun syncTabs() {
-        view.startLoading()
+    private val scope = CoroutineScope(coroutineContext)
 
-        val scope = CoroutineScope(coroutineContext)
+    /**
+     * See [SyncedTabsController.refreshSyncedTabs]
+     */
+    override fun refreshSyncedTabs() {
+        view.startLoading()
 
         scope.launch {
             accountManager.withConstellation {
@@ -43,6 +47,18 @@ internal class DefaultController(
             scope.launch(Dispatchers.Main) {
                 view.stopLoading()
             }
+        }
+    }
+
+    /**
+     * See [SyncedTabsController.syncAccount]
+     */
+    override fun syncAccount() {
+        scope.launch {
+            accountManager.withConstellation {
+                refreshDevicesAsync().await()
+            }
+            accountManager.syncNowAsync(SyncReason.User)
         }
     }
 }
--- a/components/feature/syncedtabs/src/main/java/mozilla/components/feature/syncedtabs/interactor/DefaultInteractor.kt
+++ b/components/feature/syncedtabs/src/main/java/mozilla/components/feature/syncedtabs/interactor/DefaultInteractor.kt
@@ -4,19 +4,13 @@
 
 package mozilla.components.feature.syncedtabs.interactor
 
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.launch
 import mozilla.components.browser.storage.sync.Tab
+import mozilla.components.feature.syncedtabs.controller.SyncedTabsController
 import mozilla.components.feature.syncedtabs.view.SyncedTabsView
-import mozilla.components.service.fxa.manager.FxaAccountManager
-import mozilla.components.service.fxa.manager.ext.withConstellation
-import mozilla.components.service.fxa.sync.SyncReason
-import kotlin.coroutines.CoroutineContext
 
 internal class DefaultInteractor(
-    override val accountManager: FxaAccountManager,
+    override val controller: SyncedTabsController,
     override val view: SyncedTabsView,
-    private val coroutineContext: CoroutineContext,
     override val tabClicked: (Tab) -> Unit
 ) : SyncedTabsInteractor {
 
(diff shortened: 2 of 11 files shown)


jonalmeida:
> Merging [#7114](https://codecov.io/gh/mozilla-mobile/android-components/pull/7114?src=pr&el=desc) into [master](https://codecov.io/gh/mozilla-mobile/android-components/commit/0f6e197c573757d904597960ffe248758862b9dd&el=desc) will **increase** coverage by `0.35%`.
> The diff coverage is `90.00%`.

This is some quality feel-good content, right here.

grigoryk:
It may be worth expanding the `syncNow` API to allow specifying not just a reason, but which engine you want to be synced. `appservices.SyncManager` can be configured to request syncing specific engines via `SyncParams`, currently we just tell it to always sync everything.

I'm assuming that the synced tabs feature will be started when you open up some synced-tabs UI? Depending on what else is happening w/ sync on other clients, you could greatly speed-up how quickly tabs are actually synced when you ask for a sync. E.g. if there's lots of changes to history/bookmarks, or just a bad connection, a full sync could take a while.

This is what Fennec did when you'd pull-to-refresh the synced tabs UI - it'll just sync the tabs collection.

jonalmeida:
> I'm assuming that the synced tabs feature will be started when you open up some synced-tabs UI?

We actually have both. A user-initiated sync (in Fenix this is a pull-to-refresh) and the other is the `onStart` of the feature.

I've filed https://github.com/mozilla-mobile/android-components/issues/7119 for adding more `SyncReason`s.

grigoryk:
> and the other is the onStart of the feature.

When does that happen in Fenix?

jonalmeida:
> When does that happen in Fenix? We'll run a sync while first initializing account manager, which will be pretty closely after app startup.

We would start the feature (and therefore sync account) when the view is visible. In R-B, this is a separate UI view, which can happen before or long after an account sync happens. We can be a stale state if we don't sync. 

Being able to sync individual engines sounds like a neat idea! In the Sync Tabs case, would that solve the problem of new devices that are added/removed if we would sync the engine? I assume we would still need to refresh the device constellation first?

components/feature/syncedtabs/src/main/java/mozilla/components/feature/syncedtabs/controller/DefaultController.kt
```diff
@@ -45,4 +49,16 @@ internal class DefaultController(
             }
         }
     }
+
+    /**
+     * See [SyncedTabsController.syncAccount]
+     */
+    override fun syncAccount() {
+        scope.launch {
+            accountManager.withConstellation {
```
grigoryk, line 58: Not for this PR, but just thinking out-loud:

This makes me think if it's worth consolidating these two things, perhaps as a flag once can pass into `syncNowAsync`.

e.g. this will be a nice entry point:
`syncNowAsync(reason = SyncReason.User, engines = [Tabs], updateConstellation = true)`

grigoryk, line 58: `accountManager.withConstellation` is potentially misleading. If account isn't logged-in, this is a no-op, but just glancing at this code doesn't make it obvious.

I much prefer the long-form `authenticatedAccount()?.deviceConstellation()?.refreshDevicesAsync()?.await()` for that reason, it makes things obvious at a glance even if it's a little clumsy; perhaps we can add `constellation(): DeviceConstellation?` to the account manager? Then this would become `accountManager.constellation()?.refreshDevicesAsync()?.await()`.

components/feature/syncedtabs/src/main/java/mozilla/components/feature/syncedtabs/interactor/DefaultInteractor.kt
```diff
@@ -33,11 +27,6 @@ internal class DefaultInteractor(
     }
 
     override fun onRefresh() {
-        CoroutineScope(coroutineContext).launch {
-            accountManager.withConstellation {
-                refreshDevicesAsync()
-            }
-            accountManager.syncNowAsync(SyncReason.User, true)
-        }
+        controller.syncAccount()
```
grigoryk, line 30: So if i'm reading this correctly, this is just a refactor? You were already requesting a syncNow before this PR.

components/feature/syncedtabs/src/main/java/mozilla/components/feature/syncedtabs/controller/DefaultController.kt
```diff
@@ -45,4 +49,16 @@ internal class DefaultController(
...
+                refreshDevicesAsync().await()
+            }
+            accountManager.syncNowAsync(SyncReason.User)
```
grigoryk, line 61: Before this used `debounce=true`, now it's `debounce=false`. Any reason for changing this?

jonalmeida: Yes, it's due to this bug: https://github.com/mozilla-mobile/android-components/issues/6878

components/feature/syncedtabs/src/main/java/mozilla/components/feature/syncedtabs/interactor/DefaultInteractor.kt
jonalmeida: Before we were calling `syncTabs` which pulls the tabs from the last sync. Now, we're calling `syncAccount` which is what we were doing when a user activated sync from the UI. When the account syncs and the account manager is idle, we do the `syncTabs` call.

components/feature/syncedtabs/src/main/java/mozilla/components/feature/syncedtabs/controller/DefaultController.kt
jonalmeida: > I much prefer the long-form `authenticatedAccount()?.deviceConstellation()?.refreshDevicesAsync()?.await()` for that reason, it makes things obvious at a glance even if it's a little clumsy; perhaps we can add `constellation(): DeviceConstellation?` to the account manager? Then this would become `accountManager.constellation()?.refreshDevicesAsync()?.await()`.

I've found handling the device constellation a bit clumsy at times to get into the right state, I had filed https://github.com/mozilla-mobile/android-components/issues/6879 to address that.

> `accountManager.withConstellation` is potentially misleading. If account isn't logged-in, this is a no-op, but just glancing at this code doesn't make it obvious.

Would an intermediate fix be to change the naming of the extension function to suggest it's a "maybe" action.

`withConstellationIfExists` for example?

jonalmeida: I filed https://github.com/mozilla-mobile/android-components/issues/7139 to address this in a follow-up.

jonalmeida:
bors r=grigoryk

Since we have follow-ups issues already filed, I think we can merge this.

PR #7118, opened 2020-05-26
Closes #7117: Enable local Gradle Build Cache

### LINKED ISSUE #7117: Gradle build cache for local development
A gradle build cache helps speed up local builds by caching outputs of tasks and avoiding running them even after a clean build if outputs are present in the cache from previous runs.

In https://github.com/mozilla-mobile/android-components/issues/638 @pocmo started investigating this at a larger scale, in context of remote caches and Gradle Enterprise.

This issue is about enabling a local gradle build cache as a starting point, to improve local development flow. I did a few build scans as quick tests, and results are quite promising.

grigoryk:
WIP, let's see what CI does with this.

---

### Pull Request checklist

### After merge

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -25,6 +25,9 @@ permalink: /changelog/
   * Added `getTopFrecentSites` to `PlacesHistoryStorage`, which returns a list of the top frecent site infos
     sorted by most to least frecent.
 
+* **local development**
+  * Enable local Gradle Build Cache to speed-up local builds. Build cache is located in `.build-cache/`, clear it if you run into strange problems and please file an issue.
+
 # 44.0.0
 
 * [Commits](https://github.com/mozilla-mobile/android-components/compare/v43.0.0...v44.0.0)
--- a/gradle.properties
+++ b/gradle.properties
@@ -17,6 +17,9 @@ org.gradle.jvmargs=-Xmx4096m
 # XXX Deactivated because of broken releases?
 org.gradle.parallel=false
 
+# Build cache - further configured in settings.gradle
+org.gradle.caching=true
+
 libRepositoryName=Mozilla-Mobile
 libUrl=https://github.com/mozilla-mobile/android-components
 libVcsUrl=https://github.com/mozilla-mobile/android-components.git
--- a/settings.gradle
+++ b/settings.gradle
@@ -14,6 +14,13 @@ buildscript {
     }
 }
 
+buildCache {
+    local {
+        directory = new File(rootDir, '.build-cache')
+        removeUnusedEntriesAfterDays = 30
+    }
+}
+
 def setupProject(name, path, description) {
     settings.include(":$name")
 
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ out/
 .gradle/
 build/
 !taskcluster/ci/build/
+.build-cache/
 
 # Local configuration file (sdk path, etc)
 local.properties
--- a/buildSrc/settings.gradle
+++ b/buildSrc/settings.gradle
@@ -1,0 +1,10 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+buildCache {
+    local {
+        directory = new File(rootDir, '../.build-cache')
+        removeUnusedEntriesAfterDays = 30
+    }
+}


grigoryk:
From a random task in CI - `561 actionable tasks: 386 executed, 175 from cache`. So, it's picking up the `local` cache and we're doing quite a bit less work.

Now, what needs investigation is what effect this change will have across different PRs, what the lifetime of that cache is going to be, etc...

grigoryk:
My best uneducated guess is that this should have no effect across PRs, since the build cache is local to the worker in question and the state produced isn't maintained. If that's the case, this should be harmless to enable (famous last words..).

grigoryk:
Looking at the logs from the force-push CI run, early tasks to complete didn't see many (if at all) cache hits, and tasks running later would hit more and more of the cached results.

grigoryk:
Trying this out with `./gradlew clean :sample-browser:assemble`.

With a cold local cache, this takes quite a bit longer with caching enabling, since cache needs to be populated during a build. We're spending extra two minutes packing task outputs for the cache:
Screen Shot 2020-05-26 at 4 32 19 PM

However, subsequent clean builds don't need to pack as much. This is just a straight clean rebuild with no changes:
Screen Shot 2020-05-26 at 4 33 52 PM

Depending on what's changing, there could be things to re-cache. This is with a few code changes in sample-browser and a dependencies change (adding a new component as a dependency):
Screen Shot 2020-05-26 at 7 15 31 PM

... and this is just a code change:
Screen Shot 2020-05-26 at 7 18 36 PM

So, packing of certain task outputs is quite expensive, but on average it's not done very frequently due to the nature of what triggers them, so most of local builds do not incur that cost.

It's possible to annotate certain tasks with `«task».outputs.cacheIf("executing task is faster", { false })` to try and optimize this further.

grigoryk:
So with CI performance when it comes to a local cache (so, not shared across PRs), this change is both a positive (using the cached outputs) and a negative (caching outputs for tasks that won't be hit again).

I think we should still land this, it seems on average it's a net positive in CI and a net positive locally. But, before we do I'd like to first land some CI telemetry to actually measure impact of these changes.

docs/changelog.md
```diff
@@ -48,6 +48,9 @@ permalink: /changelog/
     menu also contains an access entry for Add-ons Manager, for which `onAddonsManagerTapped` needs to be passed in the
     constructor.
 
+* **local development**
+  * Enable local Gradle Build Cache to speed-up local builds.
```
pocmo, line 52: Let's make sure everyone knows about this in case they run into errors and may need to nuke their cache.

settings.gradle
```diff
@@ -14,6 +14,13 @@ buildscript {
     }
 }
 
+buildCache {
+    local {
+        directory = new File(rootDir, '.build-cache')
+        removeUnusedEntriesAfterDays = 30
+    }
+}
```
pocmo, line 22: And you did see improvements in CI here too? Otherwise this seems not do much on a clean CI run. But I guess it also doesn't have downsides to have it on. 🤷

gradle.properties
```diff
@@ -17,6 +17,9 @@ org.gradle.jvmargs=-Xmx4096m
 # XXX Deactivated because of broken releases?
 org.gradle.parallel=false
 
+# Build cache - further configured in settings.gradle
+org.gradle.caching=true
```
pocmo, line 21: The only fear I have here is that we may have non-standard tasks that are not cacheable. Or that we cache tasks that actually should run everytime. Like things that depend on timestamps or constants in BuildConfig. Can you do some sanity checks to make sure they look like what you'd expect? 🤔 
Does this have any impact on our custom substitution build setup?

docs/changelog.md
grigoryk: I'll send out an email after this lands.

settings.gradle
grigoryk: I think so? It's hard to say due to noise in durations of the CI runs. Caches are certainly being used in CI, so it should help as long as we're not taking too long packing outputs for the cache.

gradle.properties
grigoryk: This shouldn't affect autopublication stuff, but I'll verify that.

Build scans list a few hundred of our tasks as non-cacheable already. I'll do a sanity pass to see if something that's marked as cacheable stands out to me as "shouldn't be".

Otherwise, we should benefit from going through the non-cacheable ones and figuring out if they can be/should be cached. But that's a follow-up.

pocmo:
@grigoryk Should we go ahead and land this?

grigoryk:
@pocmo yes, let's do it; now that I know I can retroactively track our impact on CI, we can land this now and then look for any regressions in the CI dashboard for a-c that doesn't yet exist, but will :)

grigoryk:
bors r=pocmo

PR #7127, opened 2020-05-27
Issue #7021: Integrate the ThumbnailDiskCache with BrowserThumbnails

### LINKED ISSUE #7021: Integrate the ThumbnailDiskCache with BrowserThumbnails

gabrielluong:
---

Fixes #7021 

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/ThumbnailsMiddleware.kt
+++ b/components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/ThumbnailsMiddleware.kt
@@ -1,0 +1,36 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.thumbnails
+
+import mozilla.components.browser.state.action.BrowserAction
+import mozilla.components.browser.state.action.ContentAction
+import mozilla.components.browser.state.state.BrowserState
+import mozilla.components.browser.thumbnails.storage.ThumbnailStorage
+import mozilla.components.lib.state.Middleware
+import mozilla.components.lib.state.MiddlewareStore
+
+/**
+ * [Middleware] implementation for handling [ContentAction.UpdateThumbnailAction] and storing
+ * the thumbnail to the disk cache.
+ */
+class ThumbnailsMiddleware(
+    private val thumbnailStorage: ThumbnailStorage
+) : Middleware<BrowserState, BrowserAction> {
+    override fun invoke(
+        store: MiddlewareStore<BrowserState, BrowserAction>,
+        next: (BrowserAction) -> Unit,
+        action: BrowserAction
+    ) {
+        when (action) {
+            is ContentAction.UpdateThumbnailAction -> {
+                // Store the captured tab screenshot from the EngineView when the session's
+                // thumbnail is updated.
+                thumbnailStorage.saveThumbnail(action.sessionId, action.thumbnail)
+            }
+        }
+
+        next(action)
+    }
+}
--- a/components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/ThumbnailsUseCases.kt
+++ b/components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/ThumbnailsUseCases.kt
@@ -1,0 +1,53 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.thumbnails
+
+import android.graphics.Bitmap
+import mozilla.components.browser.state.selector.findTab
+import mozilla.components.browser.state.store.BrowserStore
+import mozilla.components.browser.thumbnails.storage.ThumbnailStorage
+import mozilla.components.support.base.log.logger.Logger
+
+/**
+ * Contains use cases related to the thumbnails feature.
+ */
+class ThumbnailsUseCases(
+    store: BrowserStore,
+    thumbnailStorage: ThumbnailStorage
+) {
+    /**
+     * Load thumbnail use case.
+     */
+    class LoadThumbnailUseCase internal constructor(
+        private val store: BrowserStore,
+        private val thumbnailStorage: ThumbnailStorage
+    ) {
+        private val logger = Logger("ThumbnailsUseCases")
+
+        /**
+         * Loads the thumbnail of a tab from its in-memory [ContentState] or from the disk cache
+         * of [ThumbnailStorage].
+         */
+        suspend operator fun invoke(sessionIdOrUrl: String): Bitmap? {
+            val tab = store.state.findTab(sessionIdOrUrl)
+            tab?.content?.thumbnail?.let {
+                logger.debug(
+                    "Loaded thumbnail from memory (sessionIdOrUrl = $sessionIdOrUrl, " +
+                        "generationId = ${it.generationId})"
+                )
+                return@invoke it
+            }
+
+            return thumbnailStorage.loadThumbnail(sessionIdOrUrl).await()
+        }
+    }
+
+    val loadThumbnail: LoadThumbnailUseCase by lazy {
+        LoadThumbnailUseCase(
+            store,
+            thumbnailStorage
+        )
+    }
+}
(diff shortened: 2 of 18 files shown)


components/feature/tabs/src/main/java/mozilla/components/feature/tabs/tabstray/TabsTrayPresenter.kt
```diff
@@ -43,6 +45,12 @@ class TabsTrayPresenter(
 
     private suspend fun collect(flow: Flow<BrowserState>) {
         flow.map { it.toTabs(tabsFilter) }
+            .map { tabs ->
+                // Load the tab thumbnail from the memory or disk caches.
+                tabs.copy(list = tabs.list.map { tab ->
+                    tab.copy(thumbnail = thumbnailsUseCases.loadThumbnail(tab.id))
```
jonalmeida, line 51: This has a downside that I see after testing it in Sample Browser: since the in-memory version is null, we always load it from disk. This means the bitmap from the previous list is different from the one we just fetched now.

When the RecyclerView's DiffUtils checks if the tabs have changed to process a UI update, it sees that the references have changed so updates all the tab thumbnails even if the image itself hasn't changed.

components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/storage/ThumbnailStorage.kt
```diff
@@ -0,0 +1,79 @@
...
+            }
+        }
+    }
+
+    @WorkerThread
+    private fun loadThumbnailInternal(sessionIdOrUrl: String): Bitmap? {
+        val desiredSize = DesiredSize(
+            targetSize = context.resources.getDimensionPixelSize(R.dimen.mozac_browser_thumbnails_size_default),
+            maxSize = maximumSize,
+            maxScaleFactor = MAXIMUM_SCALE_FACTOR
```
jonalmeida, line 59: I noticed the icons are noticeably blurry when comparing from the cache to the ones fresh in-memory. Maybe we need to change the scaling value?

Screen Shot 2020-05-27 at 1 30 10 PM

components/browser/thumbnails/src/main/res/values/dimens.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<resources xmlns:tools="http://schemas.android.com/tools">
+    <dimen name="mozac_browser_thumbnails_size_default">102dp</dimen>
+
+    <dimen name="mozac_browser_thumbnails_maximum_size" tools:ignore="PxUsage">2153px</dimen>
```
jonalmeida, line 8: Where did this value come from? I wonder if there is a better DP value we can choose here.

Future: if we could ask the app to provide the app with this attribute based on the thumbnail visible and we can store an image size that is optimized for what is visible.

components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/ThumbnailsUseCases.kt
```diff
@@ -0,0 +1,45 @@
...
+import android.graphics.Bitmap
+import mozilla.components.browser.state.selector.findTab
+import mozilla.components.browser.state.store.BrowserStore
+import mozilla.components.browser.thumbnails.storage.ThumbnailStorage
+import mozilla.components.support.base.log.logger.Logger
+
+/**
+ * Contains use cases related to the thumbnails feature.
+ */
+class ThumbnailsUseCases(
```
jonalmeida, line 16: Nice use case!

components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/ThumbnailsUseCases.kt
```diff
@@ -0,0 +1,45 @@
...
+        private val thumbnailStorage: ThumbnailStorage
+    ) {
+        private val logger = Logger("ThumbnailsUseCases")
+
+        /**
+         * Loads the thumbnail of a tab from its in-memory [ContentState] or from the disk cache
+         * of [ThumbnailStorage].
+         */
+        suspend operator fun invoke(sessionIdOrUrl: String): Bitmap? {
+            val tab = store.state.findTab(sessionIdOrUrl)
```
jonalmeida, line 34: Thinking about this more, I wonder if we can find a way to request a screenshot as part of this method OR a new use case that would do that.

I can see it getting complicated though because we would need to pass a reference to the view in here..

components/browser/thumbnails/src/main/res/values/dimens.xml
gabrielluong: I took the max of the height/width bounds I was seeing from the Bitmap coming from engineView.captureThumbnails. I expect we will need to tweak all the DesiredSizes numbers.

jonalmeida: With pixel values, those are absolute numbers. With browser icons I think the limitation is that favicons have low image quality so we keep a hard pixel value on the image. With thumbnails though, I imagine the limitation is what the displayed view size is which is relative to your device.

> I took the max of the height/width bounds I was seeing from the Bitmap coming from engineView.captureThumbnails.

This value will change if you test it on a device with a different display.

components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/storage/ThumbnailStorage.kt
gabrielluong: Gonna address this as a followup.

components/browser/thumbnails/src/main/res/values/dimens.xml
gabrielluong: Will explore this in a followup.

components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/ThumbnailsUseCases.kt
gabrielluong: Follow up in https://github.com/mozilla-mobile/android-components/issues/7144

components/feature/tabs/src/main/java/mozilla/components/feature/tabs/tabstray/TabsTrayPresenter.kt
gabrielluong: Override the `equals()` in the `Tab` data class to check for the `thumbnail.generationId` 
The generation ID changes whenever the bitmap is modified. This can be used as an efficient way to check if a bitmap has changed.(https://developer.android.com/reference/kotlin/android/graphics/Bitmap#getgenerationid)

gabrielluong:
bors r=jonalmeida

components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/ThumbnailsMiddleware.kt
```diff
@@ -0,0 +1,36 @@
...
+    override fun invoke(
+        store: MiddlewareStore<BrowserState, BrowserAction>,
+        next: (BrowserAction) -> Unit,
+        action: BrowserAction
+    ) {
+        when (action) {
+            is ContentAction.UpdateThumbnailAction -> {
+                // Store the captured tab screenshot from the EngineView when the session's
+                // thumbnail is updated.
+                thumbnailStorage.saveThumbnail(action.sessionId, action.thumbnail)
```
pocmo, line 30: I still think the session id is not a good key for thumbnails since it not a 1:1 mapping. Loading a thumbnail for an ID gives you no guarantee it is actually for the page that the session is pointing to. Using the URL is not an option?

components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/storage/ThumbnailStorage.kt
```diff
@@ -0,0 +1,87 @@
...
+    private val decoders = AndroidImageDecoder()
+    private val logger = Logger("ThumbnailStorage")
+    private val maximumSize =
+        context.resources.getDimensionPixelSize(R.dimen.mozac_browser_thumbnails_maximum_size)
+    private val scope = CoroutineScope(jobDispatcher)
+
+    /**
+     * Asynchronously loads a thumbnail [Bitmap] for the given session ID or url.
+     */
+    fun loadThumbnail(sessionIdOrUrl: String): Deferred<Bitmap?> = scope.async {
```
pocmo, line 46: Having `sessionIdOrUrl` here is somewhat weird. 🤔

Issue #7132, opened 2020-05-27
Fix Gecko Login prompt API breaking changes


Amejia481:
GeckoView nightly introduced some [new api changes](https://mozilla.github.io/geckoview/javadoc/mozilla-central/org/mozilla/geckoview/doc-files/CHANGELOG#v78), we need to adapt our code for it.

> ⚠️ Refactored LoginStorage to the Autocomplete API to support login form autocomplete delegation. Refactored ‘LoginStorage.Delegate’ to ‘Autocomplete.LoginStorageDelegate’. Refactored GeckoSession.PromptDelegate.onLoginStoragePrompt to GeckoSession.PromptDelegate.onLoginSave. Added GeckoSession.PromptDelegate.onLoginSelect. (bug 1618058)

jonalmeida:
Closing as this was fixed with #7133.

PR #7133, opened 2020-05-27
For #7132 - Update to new Autocomplete API

ekager:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/prompt/GeckoPromptDelegate.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/prompt/GeckoPromptDelegate.kt
@@ -29,7 +29,7 @@ import org.mozilla.geckoview.GeckoSession.PromptDelegate.DateTimePrompt.Type.MON
 import org.mozilla.geckoview.GeckoSession.PromptDelegate.DateTimePrompt.Type.TIME
 import org.mozilla.geckoview.GeckoSession.PromptDelegate.DateTimePrompt.Type.WEEK
 import org.mozilla.geckoview.GeckoSession.PromptDelegate.PromptResponse
-import org.mozilla.geckoview.LoginStorage
+import org.mozilla.geckoview.Autocomplete
 import java.io.FileOutputStream
 import java.io.IOException
 import java.security.InvalidParameterException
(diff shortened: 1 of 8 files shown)


components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/prompt/GeckoPromptDelegate.kt
```diff
@@ -80,17 +80,60 @@ internal class GeckoPromptDelegate(private val geckoEngineSession: GeckoEngineSe
...
+            )
+        }
+        return geckoResult
+    }
+
+    override fun onLoginSelect(
+        session: GeckoSession,
+        prompt: PromptDelegate.AutocompleteRequest<Autocomplete.LoginSelectOption>
+    ): GeckoResult<PromptResponse>? {
+        fun Autocomplete.LoginEntry.toLogin() = Login(
```
ekager, line 106: This isn't actually doing anything, now we ignore this prompt type for now. Filed https://github.com/mozilla-mobile/android-components/issues/7134 about following up for the feature

Amejia481: We can document it in the code and link to the issue :)

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/prompt/GeckoPromptDelegate.kt
```diff
@@ -56,41 +56,65 @@ typealias AC_FILE_FACING_MODE = PromptRequest.File.FacingMode
...
-            .origin(origin)
-            .formActionOrigin(formActionOrigin)
-            .httpRealm(httpRealm)
-            .username(username)
-            .password(password)
-            .build()
+        geckoEngineSession.notifyObservers {
+            onPromptRequest(
+                PromptRequest.SaveLoginPrompt(
+                    hint = prompt.options[0].hint,
```
Amejia481, line 92: The pr looks great the only. The only thing that I'm a bit confused is about if we are sure that always we are going to have at least one option

ekager: Yeah probably not ideal and I can try to make this a bit more safe when I'm adding #7134 but we are already doing this for the logins too, just later https://github.com/mozilla-mobile/android-components/blob/master/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt#L429

Since they are broken up now into save and select now, the SaveLoginPrompt should only ever deal with exactly 1 option

Amejia481: 👍

PR #7135, opened 2020-05-27
Closes #7130: Stop sending native crashes to Sentry

### LINKED ISSUE #7130: Stop sending native crashes to Sentry
Currently native crashes are being monitored and triaged in Socorro.  This is because of the integration with Bugzilla.  Stop sending native crashes to Sentry.

rocketsroger:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/lib/crash/src/main/java/mozilla/components/lib/crash/service/SentryService.kt
+++ b/components/lib/crash/src/main/java/mozilla/components/lib/crash/service/SentryService.kt
@@ -93,25 +93,7 @@ class SentryService(
     }
 
     override fun report(crash: Crash.NativeCodeCrash): String? {
-        if (sendEventForNativeCrashes) {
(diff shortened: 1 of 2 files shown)


pocmo:
> This is already possible via the sendEventForNativeCrashes flag that we can pass in via the constructor. Maybe let's use that in the app and then consider removing the code at a later time if never used?

Or alternatively get rid of the flag altogether if it doesn't do what it says it does :)

rocketsroger:
> > This is already possible via the sendEventForNativeCrashes flag that we can pass in via the constructor. Maybe let's use that in the app and then consider removing the code at a later time if never used?
> 
> Or alternatively get rid of the flag altogether if it doesn't do what it says it does :)

You're right, I didn't think about that.  I think using the flag is a good idea.  That way we keep the functionality if we every decides to turn it back on later.  I'll update.

Issue #7050, opened 2020-05-20
Move all download state to browser store


csadilek:
We currently keep additional application state for downloads in the `DownloadService` (see `downloadJobs`) and `DownloadManager` (see `queuedDownloads`).

Having all this state in the store will make it easier to implement additional download UI in the future. It also solves the problem described in https://github.com/mozilla-mobile/android-components/issues/5217 where we have to pass the download object (and the full URL) to the download intent which may be too large in case of data URIs. If the service has access to the store it can simply read the state from there.

csadilek:
This is a bigger refactoring and we may need to be broken down more once work starts.

csadilek:
First part here is done in: https://github.com/mozilla-mobile/android-components/issues/7103

codrut-topliceanu:
Changes made for this have affected the fix for https://github.com/mozilla-mobile/fenix/issues/9044 . I'll take a look on the Fenix side and see how I can resolve those conflicts.

Amejia481:
> Changes made for this have affected the fix for [mozilla-mobile/fenix#9044](https://github.com/mozilla-mobile/fenix/issues/9044) . I'll take a look on the Fenix side and see how I can resolve those conflicts.
> 
> Edit: Or should I wait for this to be finished?

I think you could continue on Fenix, Is there something in particular that it's blocking you?

codrut-topliceanu:
@Amejia481 Yeah, you're right, if the `DownloadState` move has been completed, I should be able to make the needed changes in https://github.com/mozilla-mobile/fenix/issues/11371 without any conflicts from the proposed changes in https://github.com/mozilla-mobile/android-components/issues/7227 .

marekchen:
I want to add a Download Manager Center Activity in my browser powered by android-components.
I want show all download tasks' progress in my activity.But now DownloadState in BrowserState's queuedDownloads don't include download progress.
So I can wait this change to be done.Then I can get download progress from BrowserState?
English is not my mother  tongue, if you have any doubt about what I say,please comment me. Thank you very much.

Amejia481:
> I want to add a Download Manager Center Activity in my browser powered by android-components.
> I want show all download tasks' progress in my activity.But now DownloadState in BrowserState's queuedDownloads don't include download progress.
> So I can wait this change to be done.Then I can get download progress from BrowserState?
> English is not my mother tongue, if you have any doubt about what I say,please comment me. Thank you very much.

[...]

PR #7136, opened 2020-05-27
Closes #7103 #5217: DownloadService: Read initial download state from store

### LINKED ISSUE #5217: Crash when downloading a large data URI

### LINKED ISSUE #7103: DownloadService: Read initial download state from store
This is a first step towards https://github.com/mozilla-mobile/android-components/issues/7050.

We can add the store to the download service and read the initial download state from there, based on the download ID provided in the intent, instead of serializing the whole download state as intent extra. This will allow us to fix: https://github.com/mozilla-mobile/android-components/issues/5217

csadilek:
This moves the queued download state from the manager to the store, which cleans things up and also fixes #5217.

Keeping the changeset as small as possible so we can continue in #7050 and move the remaining state as well.

Tested various cases will do more now and also open PRs for R-B and Fenix.


docs/changelog.md
```diff
@@ -13,6 +13,40 @@ permalink: /changelog/
...
+          ),
+          tabId = sessionId,
+          onNeedToRequestPermissions = { permissions ->
+              requestPermissions(permissions, REQUEST_CODE_DOWNLOAD_PERMISSIONS)
+          }),
+      owner = this,
+      view = layout
+  )
+  ```
+
```
Amejia481, line 49: nit: Extra line

csadilek: 👍

NotWoods:
Since we'll be adding more properties later, should `queuedDownloads` be a map of longs to a new data class? That way we could just save `DownloadState` as a property, and move properties from `DownloadJobState` later on.

```kt
val queuedDownloads: Map<Long, ActiveDownloadState>

data class ActiveDownloadState(
  val download: DownloadState
  // add more properties in the future
)
```

csadilek:
> Since we'll be adding more properties later, should `queuedDownloads` be a map of longs to a new data class? 

@NotWoods Yes, I thought about that too! I think it's not more work to do as a follow-up though. So I would do that when we tackle the next step. I agree that this will definitely end up being some nested structure...

csadilek:
bors r=Amejia481

csadilek:
bors r=Amejia481

PR #7152, opened 2020-05-28
Closes #7150: Media notification: Set flag to replace existing pending intent.

### LINKED ISSUE #7150: MediaService: Set flag to replace pending intent
Currently we are not setting the flag to replace the pending intent for the media notification. This may cause an old pending intent getting invoked when the notification is clicked. So far this didn't matter since we just switched to the browser.

But now that we want to switch to the playing tab (which needs to read data from the intent) we need to get the actual latest data to know which tab to switch to.

Fenix:
https://github.com/mozilla-mobile/fenix/issues/5148

pocmo:


pocmo:
Argh. This seems to run into a Robolectric bug.

```
Exception in thread "pool-1-thread-1 @coroutine#1" java.lang.NullPointerException
	at org.robolectric.shadows.ShadowPendingIntent.create(ShadowPendingIntent.java:356)
	at org.robolectric.shadows.ShadowPendingIntent.getActivity(ShadowPendingIntent.java:65)
	at android.app.PendingIntent.getActivity(PendingIntent.java)
	at mozilla.components.feature.media.notification.MediaNotificationKt.toNotificationData(MediaNotification.kt:117)
[..]
```

It looks like Robolectric falsly assumes that with `FLAG_UPDATE_CURRENT` there's always an existing Intent and then it fails when accessing the non existing intent. 🤔

pocmo:
Filed https://github.com/mozilla-mobile/android-components/issues/7155 to consider updating Robolectric (not sure if this is going to fix this though). Only alternative here is to disable this test for now.

pocmo:
Updating robolectric is a rabbit hole. Maybe ignoring this test for now is a better option.

pocmo:
Yeah, I'll disable the test. Updating Robolectric is not straightforward and will take a while.

pocmo:
bors r=rocketsroger

PR #7154, opened 2020-05-28
Add blog posting about engine dependent feature flags.

pocmo:


pocmo:
> Should this be linked somewhere in the Contributing page?

I don't think it needs. It's not directly related to contributing to the project and more meant for consuming projects.

PR #7172, opened 2020-05-29
Close #7164: Add getSubscription to AutoPushFeature if one exists

### LINKED ISSUE #7164: Add support for checking if a push subscription exists
For web consumers, the API for `WebPushDelegate.onGetSubscription` is expected to function differently from how we implemented it.

Querying that API, should check if a subscription exists and return it, otherwise return null.

Instead, we were eagerly creating a subscription when that query was made.

This resulted in sites being given a push subscription without user permission.

---

We don't have a straight-forward mechanism to query if a subscription already exists, but we can check if the scope and endpoint exist. That should be enough to satisfy this API without creating a new subscription during the query and _only then_ invoking the subscribe API to get-or-create.

This might also be the cause of https://github.com/mozilla-mobile/android-components/issues/7161 but let's fix this first and verify that after.

jonalmeida:
---

There are two parts to this fix:

First, we added the ability to check if a subscription exists first to match the
API requirement for WebPush support in GeckoView. We do this because the
call to `getSubscription` on the engine side expects it to work as a
check before creating a subscription, where-as our implementation used a
"get or create" pattern instead.

The fallout from this is that visiting sites that check for
subscriptions were immediately given one without user permission.

Second, with the previous patch, we fixed a bug where we need to check if a
subscription exists. This new (correct) behaviour broke our original
implementation of how to handle the null case in the engine.

Integration PR for Reference Browser: https://github.com/mozilla-mobile/reference-browser/pull/1208

Fenix PR will be a fast-follow.

Closes #7164 

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/webpush/GeckoWebPushDelegate.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/webpush/GeckoWebPushDelegate.kt
@@ -22,11 +22,7 @@ internal class GeckoWebPushDelegate(private val delegate: WebPushDelegate) : Gec
         val result: GeckoResult<GeckoWebPushSubscription> = GeckoResult()
 
         delegate.onGetSubscription(scope) { subscription ->
-            if (subscription != null) {
-                result.complete(subscription.toGeckoSubscription())
-            } else {
-                result.completeExceptionally(WebPushException("Retrieving subscription failed."))
-            }
+            result.complete(subscription?.toGeckoSubscription())
         }
 
         return result
@@ -39,11 +35,7 @@ internal class GeckoWebPushDelegate(private val delegate: WebPushDelegate) : Gec
         val result: GeckoResult<GeckoWebPushSubscription> = GeckoResult()
 
         delegate.onSubscribe(scope, appServerKey) { subscription ->
-            if (subscription != null) {
-                result.complete(subscription.toGeckoSubscription())
-            } else {
-                result.completeExceptionally(WebPushException("Creating subscription failed."))
-            }
+            result.complete(subscription?.toGeckoSubscription())
         }
 
         return result
--- a/components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
+++ b/components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
@@ -231,6 +231,32 @@ class AutoPushFeature(
     }
 
     /**
+     * Checks if a subscription for the [scope] already exists.
+     *
+     * @param scope The subscription identifier which usually represents the website's URI.
+     * @param appServerKey An optional key provided by the application server.
+     * @param block The callback invoked when a subscription for the [scope] is found, otherwise null. Note: this will
+     * not execute on the calls thread.
+     */
+    fun getSubscription(
+        scope: String,
+        appServerKey: String? = null,
+        block: (AutoPushSubscription?) -> Unit
+    ) {
+        connection.ifInitialized {
+            coroutineScope.launchAndTry {
+                if (containsSubscription(scope)) {
+                    // If we have a subscription, calling subscribe will give us the existing subscription.
+                    // We do this because we do not have API symmetry across the different layers in this stack.
+                    subscribe(scope, appServerKey, {}, block)
+                } else {
+                    block(null)
+                }
+            }
+        }
+    }
+
+    /**
      * Deletes the registration token locally so that it forces the service to get a new one the
      * next time hits it's messaging server.
      */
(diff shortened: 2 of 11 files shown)


components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
```diff
@@ -230,6 +230,31 @@ class AutoPushFeature(
...
+     * @param scope The subscription identifier which usually represents the website's URI.
+     * @param appServerKey An optional key provided by the application server.
+     * @param block The callback invoked when a subscription for the [scope] is found, otherwise null.
+     */
+    fun getSubscription(
+        scope: String,
+        appServerKey: String? = null,
+        block: (AutoPushSubscription?) -> Unit
+    ) {
+        connection.ifInitialized {
```
grigoryk, line 245: So the `block` may never even get called?

components/feature/push/src/main/java/mozilla/components/feature/push/Connection.kt
```diff
@@ -154,6 +159,14 @@ internal class RustPushConnection(
         return pushApi.unsubscribeAll()
     }
 
+    @GuardedBy("this")
+    override suspend fun containsSubscription(scope: PushScope): Boolean = synchronized(this) {
+        val pushApi = api
```
grigoryk, line 164: Why do you need a local variable? you already `synchronzied` the call to this, so as long as every access to `api` is also synchronzied it seems unnecessary?

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/webpush/GeckoWebPushDelegate.kt
```diff
@@ -22,11 +22,7 @@ internal class GeckoWebPushDelegate(private val delegate: WebPushDelegate) : Gec
         val result: GeckoResult<GeckoWebPushSubscription> = GeckoResult()
 
         delegate.onGetSubscription(scope) { subscription ->
-            if (subscription != null) {
-                result.complete(subscription.toGeckoSubscription())
-            } else {
-                result.completeExceptionally(WebPushException("Retrieving subscription failed."))
-            }
+            result.complete(subscription?.toGeckoSubscription())
```
grigoryk, line 25: Due to `connection.ifInitialized` this may never get called, I think. That seems like a problem, if we never complete the result.

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
```diff
@@ -230,6 +230,31 @@ class AutoPushFeature(
...
+     * Checks if a subscription for the [scope] already exists.
+     *
+     * @param scope The subscription identifier which usually represents the website's URI.
+     * @param appServerKey An optional key provided by the application server.
+     * @param block The callback invoked when a subscription for the [scope] is found, otherwise null.
+     */
+    fun getSubscription(
+        scope: String,
+        appServerKey: String? = null,
+        block: (AutoPushSubscription?) -> Unit
```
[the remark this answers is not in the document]
jonalmeida: That's a fair point that we should invoke the callback on the caller's thread. I'll have to spend some time to figure out how this would affect our integrations.

When we last tried exposing the async functions, it was a bit more complicated to complete the GeckoResult async call with the kotlin async call. We opted to follow the WebExtension learnings since it had a similar one-to-one mapping of APIs.

Now that we have these nested calls where we need to make async checks before another async call, we should re-visit that.

For that reason, I stuck with the current system and hid the `containsSubscription` call internally.

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
jonalmeida: Yes, it could if the push native layer is not initialized, but we would have much bigger problems here if we can get this far in the call stack having not crashed yet.

This has come up before so maybe we call simplify this call by putting it within the `Connection`. It was meant to satisfy the push state along and remove the null-ability when needing to use the native API.

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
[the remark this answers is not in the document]
jonalmeida: We haven't (thankfully) run into issues with being on the wrong thread since consumers always need to be on a background thread to wait for messages, but that doesn't fix the problem with this design that you rightfully spotted.

components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
jonalmeida: Filed https://github.com/mozilla-mobile/android-components/issues/7184 to track this work.

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/webpush/GeckoWebPushDelegate.kt
jonalmeida: Answered below.

jonalmeida:
bors r=grigoryk

PR #7195, opened 2020-05-31
Issue #7194: Create a Browser Containers component

### LINKED ISSUE #7194: Create a Browser Containers component
In GV [Bug 1638878](https://bugzilla.mozilla.org/show_bug.cgi?id=1638878), we're looking to forward all the `contextualidentities` API calls to the embedder (AC). In AC, we can start adding a storage of these contextual identities in parallel to this work.

From MDN, [`ContextualIdentity`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/contextualIdentities/ContextualIdentity) can have the following properties:

- cookieStoreId: String
- color: String
- colorCode: String
- icon: String
- iconUrl: String
- name: String

However, `colorCode` and `iconUrl` are not actually exposed in [`contextualIdentities.create()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/contextualIdentities/create) so we should be able to ignore these 2 properties for the time being.

This issue is to introduce a storage component to store contextual identities in a Room database and provide the necessary APIs to get, add and remove contextual identities.

gabrielluong:
---

Fixes #7194. For ContextualIdentityStorage.addContextualIdentity, I will note that I followed the API of [contextualIdentities.create()](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/contextualIdentities/create) since this is what most web extensions care about.

### Pull Request checklist

### After merge

### DIFF
--- a/.buildconfig.yml
+++ b/.buildconfig.yml
@@ -52,6 +52,10 @@ projects:
     path: components/feature/app-links
     description: 'Component opening URLs in other non-browser apps.'
     publish: true
+  feature-containers:
+    path: components/feature/containers
+    description: 'Feature component for working with contextual identities also known as containers.'
+    publish: true
   feature-contextmenu:
     path: components/feature/contextmenu
     description: 'Component for displaying context menus for web content.'
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Before you attempt to make a contribution please read the [Community Participati
 
 * [List of good first issues](https://github.com/mozilla-mobile/android-components/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) (**New contributors start here!**) and [List of "help wanted" issues](https://github.com/mozilla-mobile/android-components/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22).
 
-* Matrix: [android-components:mozilla.org chat room](https://chat.mozilla.org/#/room/#android-components:mozilla.org) ([How to connect](https://wiki.mozilla.org/Matrix#Connect_to_Matrix)). 
+* Matrix: [android-components:mozilla.org chat room](https://chat.mozilla.org/#/room/#android-components:mozilla.org) ([How to connect](https://wiki.mozilla.org/Matrix#Connect_to_Matrix)).
 
 * Subscribe to our mailing list [android-components@](https://lists.mozilla.org/listinfo/android-components) to keep up to date ([Archives](https://lists.mozilla.org/pipermail/android-components/)).
 
@@ -125,6 +125,8 @@ _Combined components to implement feature-specific use cases._
 * 🔵 [**Accounts Push**](components/feature/accounts-push/README.md) - Feature of use cases for FxA Account that work with push support.
 
 * 🔵 [**Awesomebar**](components/feature/awesomebar/README.md) - A component that connects a [concept-awesomebar](components/concept/awesomebar/README.md) implementation to a [concept-toolbar](components/concept/toolbar/README.md) implementation and provides implementations of various suggestion providers.
+
+* 🔴 [**Containers**](components/feaature/containers/README.md) - A component for working with contextual identities also known as containers.
 
 * 🔵 [**Context Menu**](components/feature/contextmenu/README.md) - A component for displaying context menus when *long-pressing* web content.
 
--- a/components/feature/containers/README.md
+++ b/components/feature/containers/README.md
@@ -1,0 +1,19 @@
+# [Android Components](../../../README.md) > Feaature > Containers
+
+Feature component for working with contextual identities also known as containers.
+
+## Usage
+
+### Setting up the dependency
+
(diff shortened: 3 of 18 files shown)


gabrielluong:
Ah seems like CI doesn't like the component naming:

`RuntimeError: Treeherder group and symbol names must not be longer than 25 characters: feature-contextualidentities(BN) (see https://github.com/taskcluster/taskcluster-treeherder/blob/master/schemas/task-treeherder-config.yml)`

gabrielluong:
I am going to move this over to a browser component after talking with @jonalmeida since this is a component of a browser over a feature on its own.

gabrielluong:
> Ah seems like CI doesn't like the component naming:
> 
> `RuntimeError: Treeherder group and symbol names must not be longer than 25 characters: feature-contextualidentities(BN) (see https://github.com/taskcluster/taskcluster-treeherder/blob/master/schemas/task-treeherder-config.yml)`

Also, considering renaming this to browser-containers to get over the RuntimeError.

gabrielluong:
We might want to consider moving to coroutines simialr to https://github.com/mozilla-mobile/android-components/pull/7071

components/browser/containers/src/androidTest/java/mozilla/components/browser/containers/ContainerStorageTest.kt
```diff
@@ -0,0 +1,121 @@
...
+import org.junit.Assert.assertNotNull
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.Executors
+
+@ExperimentalCoroutinesApi
+@Suppress("LargeClass")
+class ContainerStorageTest {
```
pocmo, line 27: Let's make sure we get those running in automation too, after #7196 lands.

components/browser/containers/src/main/java/mozilla/components/browser/containers/Container.kt
```diff
@@ -0,0 +1,30 @@
...
+package mozilla.components.browser.containers
+
+/**
+ * A container also known as a contextual identity.
+ */
+interface Container {
+    /**
+     * The cookie store ID for the container.
+     */
+    val cookieStoreId: String
```
pocmo, line 14: Is that the id for the context? If yes, should we name this `contextId` like we do in `Session` etc.?

.buildconfig.yml
```diff
@@ -160,6 +160,10 @@ projects:
     path: components/browser/awesomebar
     description: 'A customizable awesomebar component for browsers.'
     publish: true
+  browser-containers:
```
pocmo, line 163: My gut reaction here is that I think this should be a `feature` component? 🤔 Similar to `feature-top-sites` and `feature-tab-collections`.
And I could imagine that we may add more "container feature code" here?

components/browser/containers/src/main/java/mozilla/components/browser/containers/Container.kt
gabrielluong: Yes `cookieStoreId` is equivalent to `contextId`. There are just a lot of different naming conventions when you look at the GV naming and compare it to the public documentations of Contextual Identities.

I don't have any strong preferences since it does make sense to align with what we have already named it internally.

components/browser/containers/src/androidTest/java/mozilla/components/browser/containers/ContainerStorageTest.kt
gabrielluong: Will do

.buildconfig.yml
jonalmeida: @pocmo Heh, it was originally a feature component but I suggested to gl that it should be a browser component. I figured it didn't have any feature aspect to it but could be used as part of our features instead as we decide how to build those. For example, PWA containers or the desktop version of multi-account containers. We do use feature components within other feature components though.

Either way could work really.

gabrielluong: @pocmo @jonalmeida I would like a stronger preference if anyone has one. I already changed it from `feature-contextual-identities` to `browser-container` once. Also, finalizing a naming would also be great. `feature-contextual-identities` ran into the issue with CI error: `RuntimeError: Treeherder group and symbol names must not be longer than 25 characters: feature-contextualidentities`. Alternatively, we could call this `feature-containers`.

pocmo: Just as an FYI, there's a workaround to make longer names work with treeherder (keep the long name and add a shorter override for treeherder):
https://github.com/mozilla-mobile/android-components/blob/11e259c2c347b7d608a3c9cc616d1549b9a7427c/taskcluster/ci/build/kind.yml#L96

> We do use feature components within other feature components though.

Yeah, dependencies on the same level are fine.

>  Alternatively, we could call this feature-containers

That sounds good to me! 👍

gabrielluong:
bors r=pocmo

PR #7212, opened 2020-06-02
Remove browser-session dependency from feature-privatemode

csadilek:
This was on the browser-state board as a TODO to migrate, but the feature actually doesn't need browser-session, so ✔️ .


csadilek:
bors r=NotWoods,Amejia481

PR #7231, opened 2020-06-03
Closes #7230: Add downloads to the system download database for devices older than Q

### LINKED ISSUE #7230: Add completed downloads to the system database for devices older than Q
Related Fenix issue https://github.com/mozilla-mobile/fenix/issues/8754 

Right now we trying to update the download system database when [we start the download](https://github.com/mozilla-mobile/android-components/blob/ba7a2383b093a775a7bd06dc962f78af76c13543/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt#L682) but we should be doing it when the download is completed for this reason we are experiencing https://github.com/mozilla-mobile/fenix/issues/8754  on devices older than Q

Amejia481:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
+++ b/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
@@ -299,12 +299,16 @@ abstract class AbstractFetchDownloadService : Service() {
      * Be aware that you need to pass [latestUIStatus] as [DownloadJobState.status] can be modified
      * from another thread, causing inconsistencies in the ui.
      */
-    private fun updateDownloadNotification(latestUIStatus: DownloadJobStatus, download: DownloadJobState) {
+    @VisibleForTesting
+    internal fun updateDownloadNotification(latestUIStatus: DownloadJobStatus, download: DownloadJobState) {
         val notification = when (latestUIStatus) {
             ACTIVE -> DownloadNotification.createOngoingDownloadNotification(context, download)
             PAUSED -> DownloadNotification.createPausedDownloadNotification(context, download)
-            COMPLETED -> DownloadNotification.createDownloadCompletedNotification(context, download)
             FAILED -> DownloadNotification.createDownloadFailedNotification(context, download)
+            COMPLETED -> {
+                addToDownloadSystemDatabaseCompat(download.state)
+                DownloadNotification.createDownloadCompletedNotification(context, download)
+            }
             CANCELLED -> {
                 removeNotification(context, download)
                 download.lastNotificationUpdate = System.currentTimeMillis()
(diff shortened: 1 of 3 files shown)


components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -360,6 +364,33 @@ abstract class AbstractFetchDownloadService : Service() {
...
     }
 
+    /**
+     * Adds a file to the downloads database system, so it could appear in Downloads App
+     * (and thus become eligible for management by the Downloads App) only for compatible devices
+     * otherwise nothing will happen.
+     */
+    @VisibleForTesting
+    internal fun addToDownloadSystemDatabaseCompact(download: DownloadState) {
+        if (SDK_INT <= Build.VERSION_CODES.P) {
```
NotWoods, line 374: I think it reads better as `< Q`, since Q introduces the new download system that causes issues.

Amejia481: 👍

Amejia481: Fixed :)

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -299,12 +299,16 @@ abstract class AbstractFetchDownloadService : Service() {
...
-    private fun updateDownloadNotification(latestUIStatus: DownloadJobStatus, download: DownloadJobState) {
+    @VisibleForTesting
+    internal fun updateDownloadNotification(latestUIStatus: DownloadJobStatus, download: DownloadJobState) {
         val notification = when (latestUIStatus) {
             ACTIVE -> DownloadNotification.createOngoingDownloadNotification(context, download)
             PAUSED -> DownloadNotification.createPausedDownloadNotification(context, download)
-            COMPLETED -> DownloadNotification.createDownloadCompletedNotification(context, download)
             FAILED -> DownloadNotification.createDownloadFailedNotification(context, download)
+            COMPLETED -> {
+                addToDownloadSystemDatabaseCompact(download.state)
```
kglazko, line 309: So this will be always called regardless of whether or not the OS version is less than Android Q? It looks addToDownloadSystem has only an if{} but not an else, and is only actionable for versions less than Android Q; is it possible to check this outside the function to reduce unnecessary function calls? (I'm not sure what the best practices are here but just wanted to point this out)

components/feature/downloads/src/test/java/mozilla/components/feature/downloads/AbstractFetchDownloadServiceTest.kt
```diff
@@ -1058,6 +1058,27 @@ class AbstractFetchDownloadServiceTest {
         verify(service).unregisterNotificationActionsReceiver()
     }
 
+    @Test
+    @Config(sdk = [Build.VERSION_CODES.P])
+    fun `WHEN a download is completed the file MUST be added to the download system database`() {
```
kglazko, line 1063: I'm not sure if you want to rename this test since it's being configured for and checking a specific SDK version, and since there is different MediaStore approach for build versions > P.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
Amejia481: Good catch, normally on Android when something does actions that are backward support compatibility the functions or clases has the  **compat** word on it to indicate that support older versions for example [ContextCompat](https://developer.android.com/reference/androidx/core/content/ContextCompat) and [ActivityCompat](https://developer.android.com/reference/androidx/core/app/ActivityCompat). I put it on the function as I didn't want to enlarge the `updateDownloadNotification` function, also I added the note on the description of the function :)

kglazko: Oh then ultra-nit but should it be renamed to Compat instead of Compact? I think that had me a little confused, I didn't catch the reference/

components/feature/downloads/src/test/java/mozilla/components/feature/downloads/AbstractFetchDownloadServiceTest.kt
Amejia481: 👍 
I updated the test to include that only apply on devices older that Q:
"WHEN a download is completed on devices older than Q the file MUST be added manually to the download system database"

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
Amejia481: No problem, I already updated it :)

kglazko: Awesome, you changed it! Thank you 👍

PR #7239, opened 2020-06-04
Close #7236: Fix last verification check in AutoPushFeature

### LINKED ISSUE #7236: Push subscription verification is not always executed on first install
To avoid hitting our push servers too often, we check our last successful verification first and verify again after the [`PERIODIC_INTERVAL_MILLISECONDS`][1] interval.

A problem arises when we first access the `prefLastVerified`: if the value does not exist, we use `System.currentTimeMillis()` which should be fine in most cases, except on it's first usage:

```kotlin
private fun shouldVerifyNow(): Boolean {
    return (System.currentTimeMillis() - prefLastVerified) >= PERIODIC_INTERVAL_MILLISECONDS
}
```

Here, we compare the current time with the pref value (which doesn't exist), so we query the current time again. The difference of that is therefore a negative value which is always fail the verify check on that first attempt.

[0]: https://searchfox.org/mozilla-mobile/rev/f1321a39b6154079614c894c01df30cba1fac16f/android-components/components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt#88
[1]: https://searchfox.org/mozilla-mobile/rev/f1321a39b6154079614c894c01df30cba1fac16f/android-components/components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt#348

jonalmeida:
When we first access the `lastVerified` pref we do not have an value
stored so we get the value from the current time.

The first usage is where we compare the delta of the current time with
the `lastVerified` so this lead us to have a negative value which fails.

---

Closes #7236 

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
+++ b/components/feature/push/src/main/java/mozilla/components/feature/push/AutoPushFeature.kt
@@ -85,7 +85,7 @@ class AutoPushFeature(
     private val prefToken: String?
         get() = preferences(context).getString(PREF_TOKEN, null)
     private var prefLastVerified: Long
-        get() = preferences(context).getLong(LAST_VERIFIED, System.currentTimeMillis())
+        get() = preferences(context).getLong(LAST_VERIFIED, 0)
         set(value) = preferences(context).edit().putLong(LAST_VERIFIED, value).apply()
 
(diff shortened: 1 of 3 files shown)


jonalmeida:
bors r=Amejia481

Issue #468, opened 2018-07-20
Define design axioms


pocmo:
Example:
http://www.gwtproject.org/makinggwtbetter.html#designaxioms

pocmo:
https://mozac.org/contributing/design-axioms

PR #7240, opened 2020-06-04
Issue #468: Define initial set of design axioms.

pocmo:
Hey folks. Issue #468 has been around for a while and I took the time this sprint to try to define an initial set of design axioms for AC.

See those axioms as a set of generic "base rules" that define how we want to build our components and what is important to us. I do not intend to dictate those. See this PR as a proposal and that's why I will flag you all for review. Mostly I tried to capture what we have been following already... but we never wrote it down somewhere. Let me know what you think and then we can tweak as needed.

Also do not see them as "set in stone". We can adapt and complement them along the way. For example I feel like we should have something about "performance", "security" or "privacy" in here too eventually.

Here some other examples of axioms that I looked at:
- http://www.gwtproject.org/makinggwtbetter.html#designaxioms
- https://www.designprinciplesftw.com/collections/design-axioms
- http://www.mit.edu/~juhan/design_axioms.html

### DIFF
--- a/docs/contribute/design_axioms.md
+++ b/docs/contribute/design_axioms.md
@@ -1,0 +1,39 @@
+---
+layout: page
+title: Design Axioms
+permalink: /contributing/design-axioms
+---
+
+## 💚 100% Kotlin
+
+Kotlin is the language of choice for components we develop, and it is the recommended language for application projects integrating components. We make use of Kotlin idioms and official Kotlin extension libraries (e.g. Coroutines).
+
+Java interoperability is not a goal of this project. We may add annotations for Java interoperability (e.g. `@JvmStatic`) at the request of a consumer to facilitate integration, but we are not considering replacing Kotlin idioms for the sake of operability.
+
+## 🔄 Third-party dependencies
+
+As providers of a framework, we strive to be independent of other third-party libraries unless strictly needed (e.g. Android Jetpack libraries).
+
+It is our goal to let our consumers decide what third-party libraries fit their needs, without our components making this choice or introducing an additional, duplicated stack of third-party frameworks.
+
+Additionally, by limiting external dependencies, we intent to keep the byte size of our components and the consumer apps as small as possible.
+
+## 🧰 Simple and applicable
+
+We are building components that directly satisfy the needs of our consumers. We tend to not design components in a vacuum without a defined use case. Components are preferred to be simple and "to the point". Additional functionality is added as needed, never in advance without knowing if it will ever be used (or how).
+
+## 🎨 Customizability
+
+Different consumers have different requirements. If possible, we try to provide customization options (e.g. styling) as needed by the consumers. We aim to balance customizability and complexity. When too complex of customization options are required, we prefer alternative component implementations.
+
+## 🧩 Pluggability
+
+Even with customization options, not all components will be able to satisfy all needs of all consumers. As such, we strive to depend on interfaces (allowing different implementations) instead of concrete classes. At a component level we prefer the same abstraction and try to depend on "concept components" (e.g. `concept-toolbar`) and their interfaces instead of components providing a specific implementation (e.g. `browser-toolbar`).
+
+## ✅ Testing
+
+We strive to have a high code coverage with a high quality suite of tests. Tests are meant to prevent regressions, exercise various parts of the code base, prove correctness and assert an always shippable state.
+
+## 📓 API Documentation
+
+While we strive to keep the public API surface as simple and self-explanatory as possible, we provide KDocs for all public API methods to describe the methods purpose, parameters, return types, and possible exceptions, if applicable.
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -21,6 +21,7 @@ Before contributing, please review our [Community Participation Guidelines](http
 
 ### Development
 
+* [Design Axioms]({{ site.baseurl }}/contributing/design-axioms)
 * [Architecture and Overview]({{ site.baseurl }}/contributing/architecture)
 * [Code coverage]({{ site.baseurl }}/contributing/code-coverage)
 * [Working on unreleased component code in an app]({{ site.baseurl }}{% link contribute/components_inside_app.md %})


docs/contribute/design_axioms.md
```diff
@@ -0,0 +1,35 @@
...
+
+## 💚 100% Kotlin
+
+Kotlin is the language of choice for components we develop and is the recommended language for application projects integrating components. We make use of Kotlin idioms and official Kotlin extension libraries (e.g. Coroutines).
+
+Java interoperability is not a goal of this project. We may add annotations for Java interoperability (e.g. `@JvmStatic`) at the request of a consumer to facilitate integration, but we are not considering replacing Kotlin idioms for the sake of operability.
+
+## 🔄 Third-party dependencies
+
+As providers of a framework we strive to not depend on other third-party libraries as long as not strictly needed (e.g. Android Jetpack libraries).
```
csadilek, line 15: nit: We strive to be independent of other third-party libraries unless strictly needed...

docs/contribute/design_axioms.md
```diff
@@ -0,0 +1,35 @@
...
+
+We are building components that directly satisfy the needs of our consumers. We tend to not design components in a vacuum without a defined use case. Components are preferred to be simple and "to the point". Additional functionality is added as needed, never in advance without knowing if it will ever be used (or how).
+
+## 🎨 Customizability
+
+Different consumers have different requirements. If possible we try to provide customization options (e.g. styling) as and if needed by consumers. We try to balance customizability and complexity. Sometimes we prefer alternative component implementations over too complex customization options.
+
+## 🧩 Pluggability
+
+Even with customization options not all components will be able to satisfy all needs of all consumers. As such, we strive to depend on interfaces (allowing different implementations) instead of classes in our code where applicable and useful. At a component level we prefer the same abstraction and try to depend on "concept components" (e.g. `concept-toolbar`) and their interfaces instead of components providing a specific implementation (e.g. `browser-toolbar`).
```
csadilek, line 31: nit: instead of concrete classes.

docs/contribute/design_axioms.md
```diff
@@ -0,0 +1,35 @@
...
+
+## ✅ Testing
```
csadilek, line 33: ❤️

docs/contribute/design_axioms.md
```diff
@@ -0,0 +1,35 @@
...
+
+We strive to have a high code coverage with a high quality suite of tests. Tests are meant to prevent regressions, exercise various parts of the code base, prove correctness and assert an always shippable state.
```
csadilek, line 35: Suggestion:

📓 API Documentation

While we strive to keep the public API surface as simple and self-explanatory as possible, we provide KDocs for all public API methods to describe the methods purpose, parameters, return types, and possible exceptions, if applicable.

pocmo:
> I wonder if it's worth mentioning that breaking APIs are inevitable and we don't go out of our way to avoid it (but we keep them documented).

Yes, I think that is a great idea and I totally think we should add something about API stability and guarantees we make.

docs/contribute/design_axioms.md
pocmo: That is a great addition! 👍

docs/contribute/design_axioms.md
psymoon: nit: As providers of a framework, we strive [...]

docs/contribute/design_axioms.md
```diff
@@ -0,0 +1,35 @@
...
+
+Java interoperability is not a goal of this project. We may add annotations for Java interoperability (e.g. `@JvmStatic`) at the request of a consumer to facilitate integration, but we are not considering replacing Kotlin idioms for the sake of operability.
+
+## 🔄 Third-party dependencies
+
+As providers of a framework we strive to not depend on other third-party libraries as long as not strictly needed (e.g. Android Jetpack libraries).
+
+It is our goal to let our consumers decide what third-party libraries fit their needs, without our components making this choice or introducing an additional, duplicated stack of third-party frameworks.
+
+Additionally by avoiding excessive frameworks we intent to keep the byte size of our components, and essentially the consuming apps, as small as possible.
```
psymoon, line 19: nit: Additionally, by limiting external dependencies, we intent to keep the byte size of our components and the consumer apps as small as possible.

docs/contribute/design_axioms.md
```diff
@@ -0,0 +1,35 @@
...
+
+Kotlin is the language of choice for components we develop and is the recommended language for application projects integrating components. We make use of Kotlin idioms and official Kotlin extension libraries (e.g. Coroutines).
+
+Java interoperability is not a goal of this project. We may add annotations for Java interoperability (e.g. `@JvmStatic`) at the request of a consumer to facilitate integration, but we are not considering replacing Kotlin idioms for the sake of operability.
+
+## 🔄 Third-party dependencies
+
+As providers of a framework we strive to not depend on other third-party libraries as long as not strictly needed (e.g. Android Jetpack libraries).
+
+It is our goal to let our consumers decide what third-party libraries fit their needs, without our components making this choice or introducing an additional, duplicated stack of third-party frameworks.
```
psymoon, line 17: nit: I think this is a run-on sentence.

"It is our goal to let our consumers decide what third-party libraries fit their needs; our components will not be making this choice or introducing additional and duplicated stack of third-party frameworks."

docs/contribute/design_axioms.md
```diff
@@ -0,0 +1,35 @@
+---
+layout: page
+title: Design Axioms
+permalink: /contributing/design-axioms
+---
+
+## 💚 100% Kotlin
+
+Kotlin is the language of choice for components we develop and is the recommended language for application projects integrating components. We make use of Kotlin idioms and official Kotlin extension libraries (e.g. Coroutines).
```
psymoon, line 9: nit: Kotlin is the language of choice for the components we develop, and it is the recommended [...]

docs/contribute/design_axioms.md
```diff
@@ -0,0 +1,35 @@
...
+
+Additionally by avoiding excessive frameworks we intent to keep the byte size of our components, and essentially the consuming apps, as small as possible.
+
+## 🧰 Simple and applicable
+
+We are building components that directly satisfy the needs of our consumers. We tend to not design components in a vacuum without a defined use case. Components are preferred to be simple and "to the point". Additional functionality is added as needed, never in advance without knowing if it will ever be used (or how).
+
+## 🎨 Customizability
+
+Different consumers have different requirements. If possible we try to provide customization options (e.g. styling) as and if needed by consumers. We try to balance customizability and complexity. Sometimes we prefer alternative component implementations over too complex customization options.
```
psymoon, line 27: nit: [...] "If possible, we try to provide customization options (e.g. styling) as needed by the consumers. We aim to balance customizability and complexity. When too complex of customization options are required, we prefer alternative component implementations."

docs/contribute/design_axioms.md
psymoon: nit: Even with customization options, not all components will be [...]

pocmo:
Updated the PR with your suggestions. :)

pocmo:
bors r=csadilek,Amejia481, jonalmeida,NotWoods,psymoon

PR #7245, opened 2020-06-04
Closes #6356: Install built-in extensions using GV extension controller

### LINKED ISSUE #6356: Install built-in extensions using GV extension controller

csadilek:
**DO NOT LAND** - We're first making sure this doesn't regress performance.

This is replacing the deprecated `runtime.registerWebExtension` with the new `webExtensionController.installBuiltIn` for installing built-in extensions. 

I've described the required steps here: https://github.com/mozilla-mobile/android-components/issues/6356#issuecomment-638387500

This is best reviewed by collapsing the tests. The actual changeset is pretty small and cleaned up the engine logic nicely (net negative lines of code). 🎉

### DIFF
--- a/components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt
+++ b/components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt
@@ -137,7 +137,7 @@ class BrowserIcons(
      */
     fun install(engine: Engine, store: BrowserStore) {
         engine.installWebExtension(
-            id = "mozacBrowserIcons",
+            id = "icons@mozac.org",
             url = "resource://android/assets/extensions/browser-icons/",
             allowContentMessaging = true,
             onSuccess = { extension ->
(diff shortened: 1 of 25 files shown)


csadilek:
Another note: Once this moves through the beta|release variants, we can simplify our API a bit by removing `allowContentMessaging` (which is a permission now), and `supportActions` (which is no longer needed). This will popup automatically though as the params will eventually be unused.

components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt
```diff
@@ -137,7 +137,7 @@ class BrowserIcons(
      */
     fun install(engine: Engine, store: BrowserStore) {
         engine.installWebExtension(
-            id = "mozacBrowserIcons",
+            id = "icons@mozac.org",
```
Amejia481, line 140: Is there any specifics about changing the ID? As the ext are persisted, doing that could duplicate the extension on devices that already have this ext?

Amejia481: No problem, I saw the comments !

csadilek: 👍 Yeah, new ID format is a requirement. Previous API was transient so extension isn't persisted yet. Going forward, we can't change IDs easily, not without landing some migration code that first removes the old one...

csadilek:
Decided not to land this short-term. Details and next steps are here: https://github.com/mozilla-mobile/android-components/issues/6356#issuecomment-639822748

PR #7251, opened 2020-06-04
Move PIP to BrowserState, use for toolbar visibility

NotWoods:
Moves everything that affects toolbar visibility into the stores.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/pwa/src/main/java/mozilla/components/feature/pwa/feature/WebAppHideToolbarFeature.kt
+++ b/components/feature/pwa/src/main/java/mozilla/components/feature/pwa/feature/WebAppHideToolbarFeature.kt
@@ -4,94 +4,120 @@
 
 package mozilla.components.feature.pwa.feature
 
-import android.net.Uri
 import android.view.View
 import androidx.core.net.toUri
 import androidx.core.view.isVisible
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.MainScope
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.flow.combine
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.launch
 import mozilla.components.browser.session.Session
-import mozilla.components.browser.session.SessionManager
+import mozilla.components.browser.state.selector.findTabOrCustomTabOrSelectedTab
+import mozilla.components.browser.state.state.CustomTabSessionState
+import mozilla.components.browser.state.state.SessionState
+import mozilla.components.browser.state.store.BrowserStore
 import mozilla.components.concept.engine.manifest.WebAppManifest
+import mozilla.components.feature.customtabs.store.CustomTabState
+import mozilla.components.feature.customtabs.store.CustomTabsServiceState
+import mozilla.components.feature.customtabs.store.CustomTabsServiceStore
+import mozilla.components.feature.pwa.ext.getTrustedScope
+import mozilla.components.feature.pwa.ext.trustedOrigins
+import mozilla.components.lib.state.ext.flow
 import mozilla.components.support.base.feature.LifecycleAwareFeature
+import mozilla.components.support.ktx.android.net.isInScope
+import mozilla.components.support.ktx.kotlinx.coroutines.flow.ifChanged
 
 /**
  * Hides a custom tab toolbar for Progressive Web Apps and Trusted Web Activities.
  *
  * When the [Session] is inside a trusted scope, the toolbar will be hidden.
  * Once the [Session] navigates to another scope, the toolbar will be revealed.
+ * The toolbar is also hidden in fullscreen mode or picture in picture mode.
+ *
+ * In standard custom tabs, no scopes are trusted.
+ * As a result the URL has no impact on toolbar visibility.
  *
  * @param toolbar Toolbar to show or hide.
- * @param sessionId ID of the custom tab session.
- * @param trustedScopes Scopes to hide the toolbar at.
- * Scopes correspond to [WebAppManifest.scope]. They can be a path (PWA) or just an origin (TWA).
- * @param onToolbarVisibilityChange Called when the toolbar is changed to be visible or hidden.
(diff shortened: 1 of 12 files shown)


components/feature/pwa/src/main/java/mozilla/components/feature/pwa/feature/WebAppHideToolbarFeature.kt
```diff
@@ -20,78 +39,78 @@ import mozilla.components.support.base.feature.LifecycleAwareFeature
  * Once the [Session] navigates to another scope, the toolbar will be revealed.
```
grigoryk, line 39: update this doc to mention PIP state

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/feature/WebAppHideToolbarFeature.kt
```diff
@@ -4,14 +4,33 @@
...
+import mozilla.components.feature.customtabs.store.CustomTabsServiceStore
+import mozilla.components.feature.pwa.ext.getTrustedScope
+import mozilla.components.feature.pwa.ext.trustedOrigins
+import mozilla.components.lib.state.ext.flow
 import mozilla.components.support.base.feature.LifecycleAwareFeature
+import mozilla.components.support.ktx.android.net.isInScope
+import mozilla.components.support.ktx.kotlinx.coroutines.flow.ifChanged
 
 /**
  * Hides a custom tab toolbar for Progressive Web Apps and Trusted Web Activities.
```
grigoryk, line 36: can you explain in the doc what happens for custom tabs?

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/feature/WebAppHideToolbarFeature.kt
```diff
@@ -20,78 +39,78 @@ import mozilla.components.support.base.feature.LifecycleAwareFeature
...
  */
 class WebAppHideToolbarFeature(
-    private val sessionManager: SessionManager,
     private val toolbar: View,
-    private val sessionId: String,
-    private var trustedScopes: List<Uri>,
-    private val onToolbarVisibilityChange: (visible: Boolean) -> Unit = {}
-) : Session.Observer, LifecycleAwareFeature {
+    private val store: BrowserStore,
+    private val customTabsStore: CustomTabsServiceStore? = null,
```
grigoryk, line 50: `manifest` will be `null` if you're not using this feature for a PWA? Would it be better to split PWA-related functionality into a feature which extends this class?

Why are `customTabStore` and `tabId` nullable? At the very least I'd expect `tabId` to be not-null.

NotWoods: `customTabStore` is null for backwards-compatibility. `tabId` is null to support working with the selected session, which is fairly common in Fenix.

components/support/ktx/src/main/java/mozilla/components/support/ktx/android/net/Uri.kt
```diff
@@ -25,3 +25,27 @@ val Uri.hostWithoutCommonPrefixes: String?
...
+    val path = path.orEmpty()
+    return trustedScopes.any { scope ->
+        sameOriginAs(scope) && path.startsWith(scope.path.orEmpty())
+    }
+}
+
+/**
+ * Checks that Uri has the same host as [other].
+ */
+fun Uri.sameHostAs(other: Uri) = scheme == other.scheme && host == other.host
```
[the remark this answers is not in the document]
NotWoods: I made these methods public because there are similar checks all over Fenix and AC. I kept the refactoring limited to PWA-related code, but other checks exist with the same names. I believe same host code in the engine used to be called in multiple locations.

NotWoods:
Yup, I can break up this code into separate commits if you prefer.

1. Expose all the logic needed to derive if toolbar should be hidden, so that an app can make their own version of the hide fullscreen component if desired.
   a. Make origin/scope logic a public helper
   b. Move PIP logic in BrowserState
2. Switch WebAppHideToolbarFeature so it uses our stores
   a. Switch to BrowserState for everything stored there
   b. Instead of computing "trustedOrigins" outside the feature and passing in the computed list, handling the computation automatically using the custom tab store. (See sample browser for moved logic.)
3. Add fullscreen/pip to WebAppHideToolbarFeature

Does that sound about right?

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/feature/WebAppHideToolbarFeature.kt
NotWoods: We can make a non-PWA/customtab version of this feature, but there isn't much logic to share. The store subscription code would be different if we're only listening to BrowserStore. I can add a separate version of this feature if we want.

grigoryk: Okay, thanks. Can you expand docstrings to explain these choices?

NotWoods: 👍

components/feature/session/src/main/java/mozilla/components/feature/session/PictureInPictureFeature.kt
```diff
@@ -57,25 +56,33 @@ class PictureInPictureFeature(
...
         activity.enterPictureInPictureMode()
         true
     }
 
-    fun onPictureInPictureModeChanged(enabled: Boolean) = pipChanged?.invoke(enabled)
+    /**
+     * Should be called when then system informs you of changes to and from picture-in-picture mode.
+     * @param enabled True if the activity is in picture-in-picture mode.
+     */
+    fun onPictureInPictureModeChanged(enabled: Boolean): Job {
```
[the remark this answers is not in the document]
NotWoods: I've already updated this function to return unit.

NotWoods:
bors r=grigoryk

PR #7258, opened 2020-06-04
Issue #7261: Add an isPinned parameter to Top Sites

### LINKED ISSUE #7261: Add an isPinned parameter to Top Sites
Adds an isPinned parameter to Top Sites so that we can differentiate whether or not the top site is pinned by the user or not (versus autogenerated by frequently visited sites).

gabrielluong:
Fixes #7261. Adds an `isPinned` parameter to Top Sites so that we can differentiate whether or not the top site is pinned by the user or not (versus autogenerated by frequently visited sites).

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/top-sites/schemas/mozilla.components.feature.top.sites.db.TopSiteDatabase/4.json
+++ b/components/feature/top-sites/schemas/mozilla.components.feature.top.sites.db.TopSiteDatabase/4.json
@@ -1,0 +1,64 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 4,
+    "identityHash": "67b4635c36f8c6c11dc45b8733686660",
+    "entities": [
+      {
+        "tableName": "top_sites",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`id` INTEGER PRIMARY KEY AUTOINCREMENT, `title` TEXT NOT NULL, `url` TEXT NOT NULL, `is_default` INTEGER NOT NULL, `is_pinned` INTEGER NOT NULL, `created_at` INTEGER NOT NULL)",
+        "fields": [
+          {
+            "fieldPath": "id",
+            "columnName": "id",
+            "affinity": "INTEGER",
+            "notNull": false
+          },
+          {
+            "fieldPath": "title",
+            "columnName": "title",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "url",
+            "columnName": "url",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "isDefault",
+            "columnName": "is_default",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
+          {
+            "fieldPath": "isPinned",
+            "columnName": "is_pinned",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
+          {
+            "fieldPath": "createdAt",
+            "columnName": "created_at",
+            "affinity": "INTEGER",
+            "notNull": true
+          }
+        ],
+        "primaryKey": {
+          "columnNames": [
+            "id"
+          ],
+          "autoGenerate": true
+        },
+        "indices": [],
+        "foreignKeys": []
+      }
+    ],
+    "views": [],
+    "setupQueries": [
+      "CREATE TABLE IF NOT EXISTS room_master_table (id INTEGER PRIMARY KEY,identity_hash TEXT)",
+      "INSERT OR REPLACE INTO room_master_table (id,identity_hash) VALUES(42, '67b4635c36f8c6c11dc45b8733686660')"
+    ]
+  }
+}
--- a/components/feature/top-sites/src/androidTest/java/mozilla/components/feature/top/sites/TopSiteStorageTest.kt
+++ b/components/feature/top-sites/src/androidTest/java/mozilla/components/feature/top/sites/TopSiteStorageTest.kt
@@ -247,6 +247,67 @@ class TopSiteStorageTest {
         }
     }
 
+    @Test
+    fun migrate3to4() {
+        val dbVersion3 = helper.createDatabase(MIGRATION_TEST_DB, 3).apply {
+            execSQL(
+                "INSERT INTO " +
+                    "top_sites " +
+                    "(title, url, is_default, created_at) " +
+                    "VALUES " +
+                    "('Mozilla','mozilla.org',0,1)," +
+                    "('Top Articles','https://getpocket.com/fenix-top-articles',0,2)," +
+                    "('Wikipedia','https://www.wikipedia.org/',0,3)," +
+                    "('YouTube','https://www.youtube.com/',0,4)"
+            )
+        }
+
+        dbVersion3.query("SELECT * FROM top_sites").use { cursor ->
+            assertEquals(5, cursor.columnCount)
+        }
+
+        val dbVersion4 = helper.runMigrationsAndValidate(
+            MIGRATION_TEST_DB, 4, true, Migrations.migration_3_4
+        ).apply {
+            execSQL(
+                "INSERT INTO  " +
+                    "top_sites " +
+                    "(title, url, is_default, is_pinned, created_at) " +
+                    "VALUES " +
+                    "('Firefox','firefox.com',1,1,5)," +
+                    "('Monitor','https://monitor.firefox.com/',0,0,5)"
+            )
+        }
+
+        dbVersion4.query("SELECT * FROM top_sites").use { cursor ->
+            assertEquals(6, cursor.columnCount)
+
+            // Check is_pinned for Mozilla
+            cursor.moveToFirst()
+            assertEquals(1, cursor.getInt(cursor.getColumnIndexOrThrow("is_pinned")))
+
+            // Check is_pinned for Top Articles
+            cursor.moveToNext()
+            assertEquals(1, cursor.getInt(cursor.getColumnIndexOrThrow("is_pinned")))
+
+            // Check is_pinned for Wikipedia
+            cursor.moveToNext()
+            assertEquals(1, cursor.getInt(cursor.getColumnIndexOrThrow("is_pinned")))
+
+            // Check is_pinned for YouTube
+            cursor.moveToNext()
+            assertEquals(1, cursor.getInt(cursor.getColumnIndexOrThrow("is_pinned")))
+
+            // Check is_pinned for Firefox
+            cursor.moveToNext()
+            assertEquals(1, cursor.getInt(cursor.getColumnIndexOrThrow("is_pinned")))
+
+            // Check is_pinned for Monitor
+            cursor.moveToNext()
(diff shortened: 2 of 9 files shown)


liuche:
Is `isPinned` already a parameter in the Bookmarks places storage? I don't know if we use it in sync or other clients, but I'm curious about the two sources of truth? cc @grigoryk

Amejia481:
bors try

Amejia481:
> Fixes #7261. Adds an `isPinned` parameter to Top Sites so that we can differentiate whether or not the top site is pinned by the user or not (versus autogenerated by frequently visited sites).

Could we use the `isDefault` property for the same purpose ?

gabrielluong:
> > Fixes #7261. Adds an `isPinned` parameter to Top Sites so that we can differentiate whether or not the top site is pinned by the user or not (versus autogenerated by frequently visited sites).
> 
> Could we use the `isDefault` property for the same purpose ?

No, `isDefault` are added by Fenix (as a default). In this sense, it is already pinned and `isDefault` is also used for telemetry purposes to tell which top sites was added by Fenix as a default.

`isPinned` helps user differentiate from `isDefault` and autogenerated top sites since pinned will tell us that this is pinned by the user. 

This illustration might help:
image

grigoryk:
> Additionally, we have to guard to not allow consumer for entering invalid combinations

There isn't an invalid combination, though. Any default site (e.g. a support forum url, or an FxA sign-in page) can also be pinned by the user as well as one of the frecent sites.

When we're displaying top sites, the key bit of information is _why_ a top site got returned by the API - this allows consumer to treat it in a certain way while displaying the result. E.g. a pinned site will be in the front, default site may have different styling, etc.
If you model this as three boolean flags, this is pretty trivial, and consumers get to decide how they want to handle displaying individual result while ignoring states they don't care about. There's no loss of information, and the API is simple to use.

If you're trying to model this from an "origin" point of view - that is, you'll have a `reason` or `origin` field of sorts instead of boolean flags, you'll have to pick sides at the API level - which origin is more important? e.g. in cases when there are several reasons why a top site exists, is it more important that the site was default, or frecent? This creates unnecessary complexity.
And your final option is to return a single explicit state, while also not loosing any information (letting consumers decide). This is essentially the same as "three boolean flags", except that you'll be making the API arguably more confusing. You'll have 2^3=8 states:

default | frecent | pinned
-|-|-|
1 | 1 | 1
1 | 1 | 0
1 | 0 | 1
1 | 0 | 0
0 | 1 | 1
0 | 1 | 0
0 | 0 | 1
0 | 0 | 0

You could do a bitmask for this, but I'd say just returning three flags is pretty simple to understand and use.

grigoryk:
Ugh, my apologies - I somehow read this as "api for returning top sites" vs "api for creating top sites", as it actually is.

In that case:
- either an explicit api to add default sites
- or a `creation reason` field

... both are better options.

gabrielluong:
> The change itself is fine, but I wonder if it's considering the slightly expanded version of this feature.
> 
> A few questions to help make sure this is going in the right direction:
> 
> higher level:
> 
> * how do we manage a set of default top sites? Does the application define them on startup?

Default top sites are set by the application on startup. 

> * what happens if you want to change a set of default sites during an upgrade? either add a site or remove some. how would you do that currently? how will that interact with the pinned sites that user may have created?

Thus far, we only had to change the default top sites for new installs and not worry about migration of existing default top sites. I would be curious about what Firefox Desktop does and if it also only resets the default top sites of new profiles.

> 
> schema:
> 
> * can a site be both "default" and "pinned"?

Let's say the site had both "default" and "pinned" set to true. For Fenix's case, we would render the ordering of "default" sites follow by "pinned" sites. However, an application might get confused here by which category/ordering we should place a site that is both "default" and "pinned". For Fenix, we would set "default" to false when a site sets "pinned" to true. We would similarly flip these bits into 3 categories - default, frecent, pinned in Fenix. 

So, there are cases in the table:

default | frecent | pinned | is it useful?
-|-|-|-
1 | 1 | 1 | No
1 | 1 | 0 | No
1 | 0 | 1 | No
1 | 0 | 0 | Yes
0 | 1 | 1 | No
0 | 1 | 0 |  Yes
0 | 0 | 1 | Yes
0 | 0 | 0 | No

From Fenix's perspective, adding an enum with the 3 possible states is most useful for it. However, I can imagine have all 3 boolean flags to keep track of the states, and allowing the application to determine what is useful to it, but it leaves the application to really figure out its allowable states. My recommendation is to go with an enum solution with the possible states. Let me know if that sounds good.

> * what about ordering? surely at _some point_ we'd want to expose that in the UI, right?

Yes, we want Default, Pinned and followed by Frecent as the ordering in Fenix. This is largely handled by reading the flags and the application determines how it wants to render their top sites.

> 
> regarding the storage api:
> 
> * with above questions in mind, what state transitions do you have here? e.g. "pinning a new site", "unpinning a site", "defining/chaning a set of defaults sites during startup", "user removing a default site", what else?

- Pinning a new site
- Unpinning a site
- Defining/changing a set of defaults sites during startup
- User removing a default site
- Adding a frecent site dynamically to populate up to X top sites
- Changing a frecent site to a pinned site
- Reordering a pinned top site
- Never showing a particular frecent site

> * how do you express these at the api surface?

Any changes to boolean/state flag happens through the `TopSiteStorage` API. We partly do this by adding/removing a top site, but should add an update API for toggling/changing flags/states.

For handling frecent sites, here is a current rough wip in https://github.com/mozilla-mobile/fenix/pull/11247/files. You will see we call the history storage API to get a list of frecent sites and choose to replace or add the sites depending on how the ordering has changed and how many frecent top sites we need to add to satisfy having X number of top sites.

> * what sorts of internal details are different for these operations? E.g. "defining a set of default top sites" sounds to me like a transaction wrapping a few insert statements, which is different from pinning a site (which could be either an insert, or an update, or possibly both!).

We add the default top sites when the `TopSiteStorage` is initialized on Fenix startup - https://github.com/gabrielluong/fenix/blob/36fdc25f02f0eb7b178827ae590f8d7e4c12365a/app/src/main/java/org/mozilla/fenix/components/TopSiteStorage.kt#L58. This does a few insert statements on a new profile for Fenix.

Adding a new frecent/default/pinned site is done with an insert statement.
Pinning/unpinning a site should be done with an update statement to flip the state/flag.

> 
> regarding fennec->fenix migrations:
> 
> * take a look at the current migration logic: https://github.com/mozilla-mobile/android-components/blob/master/components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt#L1259 - we read pinned sites from fennec and insert the as "top sites" via the old api.
> * now that you're adding a concept of "pinned sites", you'd need to modify the migration logic as well so that pinned sites in fennec are migrated as "pinned" in the new system.
> * fennec pinned sites have an explicit order - you'd need to retain this order; currently that's done via the `reversed` bit in that iterator linked above. Make sure that doesn't regress somehow
> * now, consider users we've already migrated. What will happen after your changes? Current behaviour is fine I think - e.g. "everything is marked as pinned", but let's be explicit about it and spell it out somewhere in the commit comment, at least, and have tests to cover that scenario.
> 
> Happy to chat about this in more detail; it'd be good to think through the feature a bit, and avoid db migrations and unnecessary churn of this API.

I asked @liuche a week ago if we wanted to do the fennec pinned site migration (specifically preserving the isPinned flag) and that was a no since it was too late and not as important to product. I assume we will still want to add the migration logic regardless with this PR?

My final proposal here is to add a 3 state enum (PINNED, DEFAULT, FRECENT) and position for PINNED sites so that we can be explicit about reordering the top sites in the future.

Amejia481:
I like the enum idea 👍 but I'm also opened for the other alternative

grigoryk:
> For Fenix's case, we would render the ordering of "default" sites follow by "pinned" sites. However, an application might get confused here by which category/ordering we should place a site that is both "default" and "pinned". For Fenix, we would set "default" to false when a site sets "pinned" to true. We would similarly flip these bits into 3 categories - default, frecent, pinned in Fenix.

If we want mix-up default, pinned and frecent sites into a single carousel, then if user pins a site, I think it should take precedence over whatever "default" site we may have. Otherwise, user's only recourse against a long row of "default" sites is to remove them manually, which users may not be able to easily discover. This adds friction, since the UX isn't just automatically adapting to how the browser is being used.

In Fennec, we displayed pinned first, then top sites (frecent), then suggest (default) is there was any room left. See https://searchfox.org/mozilla-esr68/source/mobile/android/base/java/org/mozilla/gecko/db/BrowserProvider.java#943

In general, I think Fennec's approach will work out better, and is simpler than trying to flip states back and forth.
Assume a maximum number of slots, say 10 (should be a flexible number). First, display pinned sites, up to 10. If any room is left, then display frecent sites, filling out any remaining slots with default sites.

An alternative approach here is to split up "default", "pinned" and "frecent" into separate rows, or just generally separate parts of the UI. That will take up much more room though.

> My final proposal here is to add a 3 state enum (PINNED, DEFAULT, FRECENT)

I don't think you need to track "frecent" state explicitly. Two booleans (pinned, default) and an order should be fine.

You get a list of frecent sites from a completely separate system, the History API. Trying to keep two separate storages in-sync isn't necessary, and will add a bunch of complexity. As far as TopSiteStorage is concerned, there is no such thing as a "frecent" site. If that storage layer isn't storing/managing them, it shouldn't know about them either. Otherwise you're coupling two very different systems.

So this TopSiteStorage will provide APIs for:
- querying and managing pinned sites - get all ordered, pin, unpin, reorder
- querying and managing default sites - get all, optionally supporting migrations, handling app updates, etc

You can also easily add a higher level "wrapper" API which operates over two systems (HistoryStorage, TopSiteStorage) and provides a higher level querying API which merges everything together into a single, ordered list. You can then use your proposed `enum` in that API to annotate sites in the final merged list.

Alternatively, application code can directly interact with History API and TopSiteStorage API, and perform the "merge" as it wants.

grigoryk:
> I asked @liuche a week ago if we wanted to do the fennec pinned site migration (specifically preserving the isPinned flag) and that was a no since it was too late and not as important to product. I assume we will still want to add the migration logic regardless with this PR?

We already have a pinned sites migration! As part of this PR you will need to modify it to work with the new API, specifically make sure that both "pinned" flag and "order" are set correctly. See `migratePinnedSites` in `FennecMigrator`.

grigoryk:
> Thus far, we only had to change the default top sites for new installs and not worry about migration of existing default top sites. I would be curious about what Firefox Desktop does and if it also only resets the default top sites of new profiles.

Ignoring the migration use case seems fine to me, as long as we're very explicit about it. In practice there's probably very little value from migrating "default" sites during upgrades, since active users will not see them once they visited a bunch of pages and hopefully pinned a few.

grigoryk:
Thinking further about this, you keep _really_ simplify this:
- keep a set of default sites as a static, hard-coded list (either in Fenix passed-in via constructor to a `TopSitesFeature`, or in a-c directly)
- Instead of `TopSitesStorage` you'll just have a single purpose `PinnedSites` storage
- `TopSitesFeature` provides a higher-level querying API which merges everything together (hard-coded default list, pinned sites, frecent sites)
- app upgrades for default sites are free (just change the hard-coded list!)

gabrielluong:
Closing PR since we don't need to add a isPinned parameter anymore.

Issue #7264, opened 2020-06-05
Use base URL when caching icons


codrut-topliceanu:
This issue is needed to fix  https://github.com/mozilla-mobile/fenix/issues/5039

At the moment icons are cached on disk (`IconDiskCache.kt`) using a key made from their respective sites' full URL, but when cached in memory they use the URL that points to the icon resource. This leads to the following situation:

1. Tab1 is loaded (www.abc.com/news), icon is downloaded and saved in disk cache using a key created from the full URL (`IconDiskCache.createKey()`), and also stored in memory as a resource. E.g.: www.abc.com/icon_type.png

2. Tab2 is loaded (www.abc.com/international) and its icon, identical to the one from tab1, is found to already be stored in the memory cache and is _no longer_ saved to the disk cache.

3. App is restarted.

4. Tab1's icon is found in disk cache saved with "www.abc.com/news" key and is properly displayed.

5. Tab2's icon does not exist as there is no key with "www.abc.com/international" key stored in the disk cache, so it defaults to the URL's auto generated icon.

PR #7266, opened 2020-06-05
For #7264 - Icons now use base URL instead of full URL when being cached

codrut-topliceanu:
---

For https://github.com/mozilla-mobile/android-components/issues/7264

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/icons/src/main/java/mozilla/components/browser/icons/processor/DiskIconProcessor.kt
+++ b/components/browser/icons/src/main/java/mozilla/components/browser/icons/processor/DiskIconProcessor.kt
@@ -17,7 +17,15 @@ class DiskIconProcessor(
     private val cache: ProcessorDiskCache
 ) : IconProcessor {
     interface ProcessorDiskCache {
-        fun put(context: Context, request: IconRequest, resource: IconRequest.Resource, icon: Icon)
+        /**
+         * Saves icon resources to cache.
+         * */
+        fun putResources(context: Context, request: IconRequest)
+
+        /**
+         * Saves icon bitmap to cache.
+         * */
+        fun putIcon(context: Context, resource: IconRequest.Resource, icon: Icon)
     }
 
     override fun process(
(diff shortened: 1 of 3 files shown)


codrut-topliceanu:
@pocmo Hey, thanks for taking a look at this. I'll see about coming up with a different solution :D

components/browser/icons/src/main/java/mozilla/components/browser/icons/processor/DiskIconProcessor.kt
```diff
@@ -27,10 +35,12 @@ class DiskIconProcessor(
         icon: Icon,
         desiredSize: DesiredSize
     ): Icon {
-        if (resource != null && icon.shouldCacheOnDisk && !request.isPrivate) {
-            cache.put(context, request, resource, icon)
+        if (resource != null && !request.isPrivate) {
+            cache.putResources(context, request)
```
pocmo, line 39: Overall this looks good. I wonder if we can find a way to optimize this `putResources` away in situations where we already have them on disk? Otherwise we will always perform a disk write here which seems wasteful?

However I am not sure if there is a good way to do that. This would require that we know here (or inside the cache) whether there are new resources or if they all have been loaded from memory (like the icon). Do you think it would work if we'd [change the order of the processors here](https://github.com/mozilla-mobile/android-components/blob/master/components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt#L96-L99) to first execute the disk processor and then the memory processor? Then we could maybe check here if the mapping is already in the memory cache? And then we know we also wrote it to disk already?

Or maybe you have another idea?

Anyhow, if that is not possible or too complicated then I guess we could also land as-is.

codrut-topliceanu: But the resources are just 
  ```
data class Resource(
        val url: String,
        val type: Type,
        val sizes: List<HtmlSize> = emptyList(),
        val mimeType: String? = null,
        val maskable: Boolean = false
    )
```

serialized to JSON, not the actual bitmaps, and they're saved to a `DiskLruCache` with maxsize of `MAXIMUM_CACHE_RESOURCES_BYTES: Long = 1024 * 1024 * 10 // 10 MB`

I'm not entirely sure if changing the processors order would help. But I'd need to investigate some more to be sure.

codrut-topliceanu:
bors r=pocmo

pocmo:
bors retry

PR #7272, opened 2020-06-05
Closes #7269: Only register extension tab handler if needed

### LINKED ISSUE #7269: Web extension tab handlers registered too often
We have to register web extension action and tab handlers (delegates) per extension and per session. For action handlers we check if we already have one and only register if not. For tab handlers we forgot to do that. As a consequence, every time we see an engine session change we register them all again, which is unnecessary.

Haven't seen a big impact of this, but it makes debugging very confusing. Let's fix this.

csadilek:
See https://github.com/mozilla-mobile/android-components/issues/7269#issue-631645512.

Introduces `hasTabHandler`, similar to `hasActionHandler` so we don't re-register tab handlers every time session state changes.


csadilek:
bors r=Amejia481

Issue #5178, opened 2019-11-27
Add a lastUsed/lastAccessed property to tabs


eoger:
On Firefox Desktop, we sort synced tabs by their `lastUsed` property.
It can be defined as roughly the time when a tab was last made active.

jonalmeida:
This issue is for correcting the tabs to be in a "last used" order. There are two parts to this issue, we can break them up into separate issues to make it easier:
1.  Every time a tab is selected, we want to give it the timestamp at the time it was selected. You can have a look at the `BrowserAction` to see how the reducer modifies the state when the [`SelectTabAction`][0] is passed.
2. Once we have that, for synced tabs, we just pass along that property to the [`SyncedTabsStorage`][1].

[0]: https://searchfox.org/mozilla-mobile/source/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt#78
[1]: https://searchfox.org/mozilla-mobile/source/android-components/components/feature/syncedtabs/src/main/java/mozilla/components/feature/syncedtabs/storage/SyncedTabsStorage.kt#41

PR #7277, opened 2020-06-05
#5178 last used tabs property

Vishwa-Mozilla:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TabListReducer.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TabListReducer.kt
@@ -62,7 +62,16 @@ internal object TabListReducer {
             }
 
             is TabListAction.SelectTabAction -> {
-                state.copy(selectedTabId = action.tabId)
+
+                val updatedTabsList = state.tabs.map {
+                    if (it.id == action.tabId) {
+                        it.copy(lastAccess = action.timeSelected)
+                    } else {
+                        it
+                    }
+                }
+
+                state.copy(tabs = updatedTabsList, selectedTabId = action.tabId)
             }
 
             is TabListAction.RemoveTabAction -> {
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
@@ -75,7 +75,7 @@ sealed class TabListAction : BrowserAction() {
      *
      * @property tabId the ID of the tab to select.
      */
-    data class SelectTabAction(val tabId: String) : TabListAction()
+    data class SelectTabAction(val tabId: String, val timeSelected: Long = System.currentTimeMillis()) : TabListAction()
 
     /**
      * Removes the [TabSessionState] with the given [tabId] from the list of sessions.
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/state/TabSessionState.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/state/TabSessionState.kt
@@ -30,7 +30,8 @@ data class TabSessionState(
     val parentId: String? = null,
     override val extensionState: Map<String, WebExtensionState> = emptyMap(),
     val readerState: ReaderState = ReaderState(),
-    override val contextId: String? = null
+    override val contextId: String? = null,
+    val lastAccess: Long = 0L
 ) : SessionState {
 
     override fun createCopy(
@@ -40,7 +41,7 @@ data class TabSessionState(
         engineState: EngineState,
         extensionState: Map<String, WebExtensionState>,
         contextId: String?
-    ) = copy(
+    ): SessionState = copy(
         id = id,
         content = content,
         trackingProtection = trackingProtection,
@@ -63,7 +64,8 @@ fun createTab(
     readerState: ReaderState = ReaderState(),
     title: String = "",
     thumbnail: Bitmap? = null,
-    contextId: String? = null
+    contextId: String? = null,
+    lastAccess: Long = 0L
 ): TabSessionState {
     return TabSessionState(
         id = id,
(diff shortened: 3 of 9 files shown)


components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TabListReducer.kt
```diff
@@ -62,7 +63,16 @@ internal object TabListReducer {
             }
 
             is TabListAction.SelectTabAction -> {
-                state.copy(selectedTabId = action.tabId)
+
+                val updatedTabsList = state.tabs.map {
```
Amejia481, line 67: nit: I think we could use instead ` state.findTab(action.tabId)` :)

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TabListReducer.kt
```diff
@@ -62,7 +63,16 @@ internal object TabListReducer {
...
+                    if (it.id == action.tabId) {
+                        it.lastAccess = System.currentTimeMillis()
+                        Log.log(Log.Priority.DEBUG, "")
```
Amejia481, line 70: nit: a leftover of debugging :)

components/browser/state/src/main/java/mozilla/components/browser/state/state/CustomTabSessionState.kt
```diff
@@ -24,7 +25,8 @@ data class CustomTabSessionState(
     val config: CustomTabConfig,
     override val engineState: EngineState = EngineState(),
     override val extensionState: Map<String, WebExtensionState> = emptyMap(),
-    override val contextId: String? = null
+    override val contextId: String? = null,
+    override var lastAccess: Long = System.currentTimeMillis()
```
Amejia481, line 29: 🤔  Not sure if we should have a default value? I don't know if creating that tab would count as accessing it.

components/browser/state/src/main/java/mozilla/components/browser/state/state/SessionState.kt
```diff
@@ -24,6 +25,7 @@ interface SessionState {
     val engineState: EngineState
     val extensionState: Map<String, WebExtensionState>
     val contextId: String?
+    var lastAccess: Long
```
Amejia481, line 28: It would be better if we you `val` as every time we change this object we are mutating it.

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TabListReducer.kt
Amejia481: Instead of changing the value `lastAccess` of directly it will be better to use `copy` to change value

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TabListReducer.kt
jonalmeida: I also thought this would work, but since we're updating the `tabs` for the `BrowserState`, we need to map the timestamp to the selected tab in the tab list, and then `copy` that over.

components/browser/state/src/main/java/mozilla/components/browser/state/state/SessionState.kt
jonalmeida, line 28: I don't think we need this on the `SessionState`. I'm not sure what Custom tabs would get out of this if we have a `lastAccess` attribute there. 🤔 

It might better leaving this on only the `TabSessionState` for now and we can always add it to Custom Tabs later if we need it.

jonalmeida: +1

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TabListReducer.kt
jonalmeida: +1

components/browser/state/src/main/java/mozilla/components/browser/state/state/CustomTabSessionState.kt
Vishwa-Mozilla: yes, I will remove it

jonalmeida: @Vishwa-Mozilla I checked with @eoger who recommended keeping this to `System.currentTimeMillis()` instead of defaulting to `0L`.

We could sync the tabs before accessing them and that might cause issues for the other synced devices.

eoger: Actually, sorry for going back on my word and for the resulting confusion, let's go with `0L`. I'm concerned this might interact badly with session restore, and it agrees with [the current implementation](https://searchfox.org/mozilla-central/rev/0e09b9191c02097034e46b193930f91c45b7885d/services/sync/modules/engines/tabs.js#209).

ekager:
Sorry for the driveby - I'm not as familiar with the new tab/browser state so not sure if we're doing this anymore, but should we also add a way to save/restore this value of the session with the `SnapshotSerializer`? (If that still makes sense)

jonalmeida:
> Sorry for the driveby - I'm not as familiar with the new tab/browser state so not sure if we're doing this anymore, but should we also add a way to save/restore this value of the session with the `SnapshotSerializer`? (If that still makes sense)

Nice catch! Yes, we probably should serialize that. I'll try to amend this PR or make another one based off it. Thanks @ekager!

jonalmeida:
I'll do this in a follow-up actually, that should make it easier to land this then.

Filed https://github.com/mozilla-mobile/android-components/issues/7476.

jonalmeida:
bors r-

Forgot to run some more tests against readerview.

ekager:
@jonalmeida does this need more changes or should we land it?

jonalmeida:
@ekager Yeah, we can land it. It was cutting it too close to the release so we had decided to wait.

PR #7287, opened 2020-06-07
Closes #7286 - Hide the keyboard when the find in page bar is unfocused

### LINKED ISSUE #7286: The keyboard doesn't always close when unfocusing the find in page bar
STR:
- Open a webpage
- Open the find in page bar
- Tap somewhere on the webpage

This seems to happen only the first time the find in page bar is opened.

Screen-Recording-20200607-202826

miDeb:
---

### Pull Request checklist

I'd like to add a test for this, but I haven't found a way to check if the keyboard is opened.

### After merge

### DIFF
--- a/components/feature/findinpage/src/main/java/mozilla/components/feature/findinpage/view/FindInPageBar.kt
+++ b/components/feature/findinpage/src/main/java/mozilla/components/feature/findinpage/view/FindInPageBar.kt
@@ -92,8 +92,6 @@ class FindInPageBar @JvmOverloads constructor(
     }
 
     override fun clear() {
-        queryEditText.hideKeyboard()
-
         queryEditText.text = null
         queryEditText.clearFocus()
         resultsCountTextView.text = null
@@ -177,7 +175,8 @@ class FindInPageBar @JvmOverloads constructor(
         resultsCountTextView.setTextColorIfNotDefaultValue(styling.resultCountTextColor)
     }
 
-    private fun bindQueryEditText() {
+    @VisibleForTesting
+    internal fun bindQueryEditText() {
         with(queryEditText) {
             setTextSizeIfNotDefaultValue(styling.queryTextSize)
             setTextColorIfNotDefaultValue(styling.queryTextColor)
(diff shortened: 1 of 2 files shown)


components/feature/findinpage/src/main/java/mozilla/components/feature/findinpage/view/FindInPageBar.kt
```diff
@@ -92,8 +92,6 @@ class FindInPageBar @JvmOverloads constructor(
     }
 
     override fun clear() {
-        queryEditText.hideKeyboard()
```
Amejia481, line 95: Is there any strong reason why we shouldn't hide the keyboard here?

miDeb: `queryEditText.clearFocus()` will trigger the hiding of the keyboard anyways, so wouldn't it just be duplicated?

Amejia481: Ok, I see it's because we are clearing focus after  `queryEditText.clearFocus()`.
Let's add add some test to verify we are `hideKeyboard` when after `clear` is called and when the view doesn't have the focus 👍

miDeb: I will do that!

miDeb: Sorry if this is a stupid question, but how can I verify `hideKeyboard` since it is an extension method?

Amejia481: It's not a problem,  it's a really good question:
As `hideKeyboard` is static is bit tricky to test, but you could have an internal function on `FindInPageBar` and that calls the extension function and in a test you can spy on `FindInPageBar` and verify if it was called.

`FindInPageBar`
 ```kotlin
   @VisibleForTesting
    internal fun hideKeyboard(view: View){
        view.hideKeyboard()
    }
```

In the test:

```kotlin
    fun `test`() {
        val findInPageBar = spy(FindInPageBar(testContext))

        findInPageBar.clear()
        verify(findInPageBar).hideKeyboard(findInPageBar)
    }
```

miDeb: Sure, thank you

PR #7296, opened 2020-06-08
Closes #7223: Clear speculative session on engine settings change

### LINKED ISSUE #7223: Speculative GeckoSession cannot be configured
At first I thought I saw the following issue in Focus for Android after integrating the engine, but this is not the case since Focus doesn't use any code that creates a speculative `GeckoSession`. However it still looks like this could be a problem. I am not sure whether this is only theoretically possible or whether this can actually happen in Fenix right now.

Our Engine allows it to create a speculative `EngineSession` that will be used later when`createSession()` is called:
https://github.com/mozilla-mobile/android-components/blob/master/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt#L177

The `SettingsUseCases` class is a helper to apply a settings change to the engine and all its engine sessions. For example `UpdateTrackingProtectionUseCase` updates the tracking protection policy for all engine sessions. It does so by looping over all available engine sessions and applying the new policy:
* https://github.com/mozilla-mobile/android-components/blob/master/components/feature/session/src/main/java/mozilla/components/feature/session/SettingsUseCases.kt#L75-L77
* https://github.com/mozilla-mobile/android-components/blob/master/components/feature/session/src/main/java/mozilla/components/feature/session/SettingsUseCases.kt#L42-L46

`UpdateSettingUseCase.forEachSession` does not update the speculative `EngineSession`. So if the settings change after the speculative `EngineSession` then the change will not apply to that session since `createSession()` will not create an engine session with the new settings but instead return the speculative engine session with the old settings.

Not sure if other code is affected that tries to do things with "all engine sessions".

What are our options to fix that? Should we allow to access the speculative engine session so that the use case can change it? Should we add a way to "drop" or "recreate" the speculative session?

csadilek:
Since the speculative session only exists for performance reasons I think it's fine to simply clear it and take a slight performance hit in case the user changes engine settings. This only affects the first session after a settings change anyway, and then we don't have to recreate it with the previous parameters.

Similar to our `AwesomeBarFeature`, the `SettingsUseCase` now takes an optional engine reference to support clearing the speculative session.

### DIFF
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
@@ -179,6 +179,14 @@ class GeckoEngine(
     }
 
     /**
+     * See [Engine.clearSpeculativeSession].
+     */
+    override fun clearSpeculativeSession() {
+        this.speculativeSession?.geckoSession?.close()
+        this.speculativeSession = null
+    }
+
+    /**
      * Opens a speculative connection to the host of [url].
      *
      * This is useful if an app thinks it may be making a request to that host in the near future. If no request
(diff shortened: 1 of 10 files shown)


csadilek:
@pocmo PR updated so we only pass the engine, instead of both engine and engine.settings.

Fenix: https://github.com/mozilla-mobile/fenix/pull/11339
R-B: https://github.com/mozilla-mobile/reference-browser/pull/1225

csadilek:
bors r=pocmo

PR #7298, opened 2020-06-08
Issue #6644: Integrate UI to cancel add-on installation

csadilek:
This brings in the missing piece of our add-on installation UI: https://miro.com/app/board/o9J_kw8Lt8g=/?moveToWidget=3074457347043892302&cot=3

There is still one crash left (cancellation after the add-on was downloaded), but it has been fixed in GV and will be in tomorrow's Nightly. Either way, we don't need to wait with this any longer as all problems have been addressed.

Once we have this here, let's leave the ticket open and land it in Fenix as well.


csadilek:
bors r=Amejia481

Issue #7295, opened 2020-06-08
Certain actions with afterChange() followed by syncDispatch() take from 200ms - 800ms


MarcLeclair:
I've noticed that some actions block the Java Main Thread when navigating to a new webpage. I've noticed it in the following build:

- locally with `geckoBetaFenixBeta` builds as shown in this profile: https://share.firefox.dev/2Y47G8D and https://share.firefox.dev/2zZ3KOu  . In this one,  `UpdateBackNavigationStateAction` takes up to 500ms at times.

- From the playstore, I noticed it with the `Nightly for Developers` build  as shown in these profiles: 
   G5 plus: https://share.firefox.dev/373tsgR
   Pixel 4: https://share.firefox.dev/3dE1Np8
  However, in this scenario, it's actually the `UpdateTitleAction` that seems to take a long time to finish.

It is worth to note that I tried it on `Firefox for Android` channel too on the play store and didn't see this behavior happening.  (profile for reference) (https://share.firefox.dev/2MAwxLY

I'm not sure if there's anything to do about this, but it seems as if a simple update action shouldn't take this long on our Main Thread.

csadilek:
> locally with `geckoBetaFenixBeta` builds as

Just clarifying that this happens on master only, also with`geckoBetaFenixBeta` build from master, but not on the latest published beta build. That means it's a pretty recent regression. @pocmo pointed to the new `ThumbnailsMiddleware` which seems to run IO operations on the main thread?

pocmo:
Locally I enabled `StrictMode` on the store thread and that was also the first thing it caught:

```
D  StrictMode policy violation: android.os.strictmode.DiskReadViolation
D      at android.os.StrictMode$AndroidBlockGuardPolicy.onReadFromDisk(StrictMode.java:1589)
D      at libcore.io.BlockGuardOs.access(BlockGuardOs.java:71)
D      at libcore.io.ForwardingOs.access(ForwardingOs.java:72)
D      at android.app.ActivityThread$AndroidOs.access(ActivityThread.java:7351)
D      at java.io.UnixFileSystem.checkAccess(UnixFileSystem.java:281)
D      at java.io.File.exists(File.java:815)
D      at android.app.ContextImpl.ensurePrivateDirExists(ContextImpl.java:677)
D      at android.app.ContextImpl.ensurePrivateCacheDirExists(ContextImpl.java:673)
D      at android.app.ContextImpl.getCacheDir(ContextImpl.java:784)
[...]

PR #7299, opened 2020-06-08
Issue #7295: Don't block store thread when saving thumbnail

csadilek:
We could use separate dispatchers (separate thread pools) for saving and loading, but we discussed and didn't think it was necessary right now.

Could also think of just using an IO dispatcher instead of a custom thread pool, but that's independent of the issue here.

### DIFF
--- a/components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/storage/ThumbnailStorage.kt
+++ b/components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/storage/ThumbnailStorage.kt
@@ -10,8 +10,10 @@ import androidx.annotation.WorkerThread
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Deferred
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.asCoroutineDispatcher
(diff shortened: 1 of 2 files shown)


components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/storage/ThumbnailStorage.kt
```diff
@@ -77,11 +79,13 @@ class ThumbnailStorage(
...
      * as its key.
      */
-    fun saveThumbnail(sessionIdOrUrl: String, bitmap: Bitmap) {
-        logger.debug(
-            "Saved thumbnail to disk (sessionIdOrUrl = $sessionIdOrUrl, " +
-                "generationId = ${bitmap.generationId})"
-        )
-        sharedDiskCache.putThumbnailBitmap(context, sessionIdOrUrl, bitmap)
-    }
+    @WorkerThread
```
pocmo, line 82: With this now doing the work asynchronously this doesn't seem to need a `WorkerThread` annotation? 🤔

csadilek: Ah, yeah, that doesn't help here. Removed.

csadilek:
bors r=pocmo,jonalmeida,gabrielluong

PR #7311, opened 2020-06-09
Closes #7310: Support custom use cases in TabsFeature

### LINKED ISSUE #7310: Make TabsFeature take tabsUseCase interfaces, not concrete classes.
We want to be able to add some behavior to the TabsFeature in Fenix. Let's turn these into interfaces 🙌

csadilek:
Because `RemoveTabsUseCase` can't be customized, making undo possible in Fenix requires a workaround and an additional observer. Seems simple enough to get rid of...


csadilek:
bors r=jonalmeida

Issue #6910, opened 2020-05-08
Output markdown for reports to be displayed in task runs


pocmo:
The taskcluster github integration can now add markdown to the task run on GitHub. We are doing this here for ui tests:
https://github.com/mozilla-mobile/fenix/runs/653344259

This is done here:
https://github.com/mozilla-mobile/fenix/blob/master/automation/taskcluster/androidTest/parse-ui-test.py

And was implemented here:
https://github.com/mozilla-mobile/fenix/pull/10089

We should do the same for:
- unit tests
- detekt
- ktlint
- lint

.. then it will be much easier to see why something failed without digging through logs and/or having to rerun everything locally. :)

NotWoods:
For detekt & ktlint, we can have results outputted in a checkstyle XML file. This would give us an easily parsed data file to use. I'm hoping to see Taskcluster add support for Github annotations so that we can show those errors inline.

PR #7324, opened 2020-06-10
For #6910 - Expose reports as artifacts

NotWoods:
This exposes the HTML reports from tests, lint, & jacoco as Taskcluster build artifacts. After this change they can be accessed from the "Artifacts" menu in Taskcluster, and linked to!


NotWoods:
bors r=grigoryk

PR #7345, opened 2020-06-11
Change WebAppHideToolbarFeature to use callback

NotWoods:
One more tweak for the PWA toolbar system. This makes the Fenix integration easier.


NotWoods:
bors r=grigoryk

PR #7360, opened 2020-06-12
Closes #7359: Register WebCompat add-on with ID matching manifest

### LINKED ISSUE #7359: WebCompat add-on is registered with wrong ID
The add-on is using `webcompat@mozilla.org` in its manifest, but `webcompat@mozilla.com` when registered.

Let's change that now while we're still using the transient `registerWebExtension`. Later it will be tricky to change.

csadilek:
The ID should match the one in the manifest. Noticed this when testing migration and found two entries with separate UUIDs in prefs.js.

Safe to change to now as we're still using the transient `registerWebExtension`, tricky later once we switch to `installBuiltIn`.

r? @denschub


csadilek:
bors r=Amejia481

Issue #7364, opened 2020-06-12
Add sessionId to DownloadState to keep track of the tab that started the download


codrut-topliceanu:
Needed to fix https://github.com/mozilla-mobile/fenix/issues/11371 (and https://github.com/mozilla-mobile/fenix/issues/9044)

In order to track the session from which a download was started we need to add a sessionId to the `DownloadState` object. This property should be populated when processing the download (`DownloadFeature`).

PR #7365, opened 2020-06-12
For #7364 - Adds sessionId to DownloadState and populates it in DowloadFeature

codrut-topliceanu:
---

For https://github.com/mozilla-mobile/android-components/issues/7364

### Pull Request checklist

- [x] **Tests**: This PR doesn't include tests as it is only a minor change.
- [x] **Changelog**: This PR does not need [a changelog entry](https://github.com/mozilla-mobile/android-components/blob/master/docs/changelog.md).

### After merge

### DIFF
--- a/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt
+++ b/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt
@@ -93,9 +93,8 @@ class DownloadsFeature(
             flow.mapNotNull { state -> state.findTabOrCustomTabOrSelectedTab(tabId) }
                 .ifChanged { it.content.download }
                 .collect { state ->
-                    val download = state.content.download
-                    if (download != null) {
-                        processDownload(state, download)
+                    state.content.download?.let { downloadState ->
+                        processDownload(state, downloadState)
                     }
                 }
         }
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/ContentStateReducer.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/ContentStateReducer.kt
@@ -61,7 +61,7 @@ internal object ContentStateReducer {
                 it.copy(thumbnail = action.thumbnail)
             }
             is ContentAction.UpdateDownloadAction -> updateContentState(state, action.sessionId) {
-                it.copy(download = action.download)
+                it.copy(download = action.download.copy(sessionId = action.sessionId))
             }
             is ContentAction.ConsumeDownloadAction -> updateContentState(state, action.sessionId) {
                 if (it.download != null && it.download.id == action.downloadId) {
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/state/content/DownloadState.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/state/content/DownloadState.kt
@@ -22,6 +22,8 @@ import kotlin.random.Random
  * @property referrerUrl The site that linked to this download.
  * @property skipConfirmation Whether or not the confirmation dialog should be shown before the download begins.
  * @property id The unique identifier of this download.
+ * @property sessionId Identifier of the session that spawned the download.
+ * @
  */
 @Suppress("Deprecation")
 @Parcelize
@@ -34,7 +36,8 @@ data class DownloadState(
     val destinationDirectory: String = Environment.DIRECTORY_DOWNLOADS,
     val referrerUrl: String? = null,
     val skipConfirmation: Boolean = false,
-    val id: Long = Random.nextLong()
+    val id: Long = Random.nextLong(),
+    val sessionId: String? = null
 ) : Parcelable {
     val filePath: String get() =
         Environment.getExternalStoragePublicDirectory(destinationDirectory).path + "/" + fileName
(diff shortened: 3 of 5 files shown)


codrut-topliceanu:
Some unit tests are failing. :( Looking into them now.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt
```diff
@@ -93,9 +93,8 @@ class DownloadsFeature(
             flow.mapNotNull { state -> state.findTabOrCustomTabOrSelectedTab(tabId) }
                 .ifChanged { it.content.download }
                 .collect { state ->
-                    val download = state.content.download
-                    if (download != null) {
-                        processDownload(state, download)
+                    state.content.download?.let { downloadState ->
```
Amejia481, line 96: I think will be better if we populate this property when the download request is issued https://github.com/mozilla-mobile/android-components/blob/9e12caf058910ed0a2b1d5fa78c8ffdb01b6ff1c/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/ContentStateReducer.kt#L65

This way we are going to have a consistent state in `BrowserStore`. what do you think?

codrut-topliceanu: @Amejia481 Thanks for taking a look at this. That's a really good solution that hadn't occurred to me! I just tried it now and it works great. Thank you.

csadilek: @Amejia481 @codrut-topliceanu Thanks and sorry for the late response. 

When creating a `DownloadState` we always know the `sessionId` (it should be called `tabId` in browser-state, btw). So, we can simply set the `tabId` here when creating the `download`: https://github.com/mozilla-mobile/android-components/blob/master/components/browser/session/src/main/java/mozilla/components/browser/session/engine/EngineObserver.kt#L185

It doesn't need to be a side-effect / unexpected change of reducing.

codrut-topliceanu: I already tried that and it didn't work, I'm afraid. It only worked for download that had skip confirmation == true.

csadilek: @codrut-topliceanu Hm can you share some more details? It effectively shouldn't matter when the sessionId is added, so why not add it directly when the download is created? It seems like a potential source of error to do this "later".

codrut-topliceanu: I'll look some more into it now, but I remember that it just wouldn't populate the sessionId in `EngineObserver` when `onExternalResource` was called (except when it was a larger download and it needed a confirmation from the user).

codrut-topliceanu: @csadilek Correction: `EngineObserver.onExternalResource` doesn't get called at all if the download is small enough that it doesn't need to show the confirmation dialog. I'm not entirely sure why,  or how to proceed.

Amejia481: I think the source event is on [here](https://github.com/mozilla-mobile/android-components/blob/28c9be258747ff964d0a92808a57ae31f791cb31/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt#L667), if you debug that line you should able to see why is not getting called.

codrut-topliceanu: Yep that's the source. Can't go further up the tree from that as it leads into Gecko.

Amejia481: Do you mean that when you try to download a file in some cases this [line](https://github.com/mozilla-mobile/android-components/blob/28c9be258747ff964d0a92808a57ae31f791cb31/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt#L667) does get executed?  If that it's the case then this is a GeckoView issue and it's a separate issue from this one.

codrut-topliceanu: Yep that's the case.

codrut-topliceanu:
bors r=csadilek

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt
Amejia481: Would you mind opening a [ bug in Bugzilla](https://bugzilla.mozilla.org/enter_bug.cgi?product=GeckoView) describing the the issue and the step to reproduce it?

codrut-topliceanu: Ok, I'll give it a try. I've never opened one in Bugzilla before.

Amejia481: If you need any help just let me know, happy to help!

Issue #7366, opened 2020-06-12
Provide request interceptor to automatically navigate into PWAs


TitanNano:
For mozilla-mobile/fenix#5772 a request interceptor is needed that will launch a PWA intent when the user navigates to a URL which falls into the scope of a PWA.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-6356)

csadilek:
Landed back in https://github.com/mozilla-mobile/android-components/pull/7367.

PR #7367, opened 2020-06-12
Provide request interceptor to automatically navigate into PWAs

TitanNano:
For #7366 and mozilla-mobile/fenix#5772

- a WebAppInterceptor is needed to redirect to a separate WebApp
  activity
- the manifest dao currently reports a web app as installed when it was
  last used before the deadline / expiration time. This is the oposite
  of what it's supposed to do.
- The intent extension should also hold an override URL so a WebApp
  intent can be launched with a deeplink.
- The WebAppIntentProcessor currently launches new sessions for each
  intent. This will break user expectations when they start to be
  switched to a already running WebApp activity but loose their session.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/pwa/src/main/java/mozilla/components/feature/pwa/WebAppInterceptor.kt
+++ b/components/feature/pwa/src/main/java/mozilla/components/feature/pwa/WebAppInterceptor.kt
@@ -1,0 +1,60 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.feature.pwa
+
+import android.content.Context
+import android.content.Intent
+import android.net.Uri
+import mozilla.components.concept.engine.EngineSession
+import mozilla.components.concept.engine.request.RequestInterceptor
+import mozilla.components.feature.pwa.ext.putUrlOverride
+import mozilla.components.feature.pwa.intent.WebAppIntentProcessor
+
+/**
+ * This feature will intercept requests and reopen them in the corresponding installed PWA, if any.
+ *
+ * @param shortcutManager current shortcut manager instance to lookup web app install states
+ */
+class WebAppInterceptor(
+    private val context: Context,
+    private val manifestStorage: ManifestStorage,
+    private val launchFromInterceptor: Boolean = true
+) : RequestInterceptor {
+
+    @Suppress("ReturnCount")
+    override fun onLoadRequest(
+        engineSession: EngineSession,
+        uri: String,
+        hasUserGesture: Boolean,
+        isSameDomain: Boolean,
+        isRedirect: Boolean,
+        isDirectNavigation: Boolean
+    ): RequestInterceptor.InterceptionResponse? {
+        val scope = manifestStorage.getInstalledScope(uri) ?: return null
+        val startUrl = manifestStorage.getStartUrlForInstalledScope(scope) ?: return null
+        val intent = createIntentFromUri(startUrl, uri)
+
+        if (!launchFromInterceptor) {
+            return RequestInterceptor.InterceptionResponse.AppIntent(intent, uri)
+        }
+
+        intent.flags = intent.flags or Intent.FLAG_ACTIVITY_NEW_TASK
+        context.startActivity(intent)
+
+        return RequestInterceptor.InterceptionResponse.Deny
+    }
+
+    /**
+     * Creates a new VIEW_PWA intent for a URL.
+     *
+     * @param uri target URL for the new intent
+     */
+    private fun createIntentFromUri(startUrl: String, urlOverride: String = startUrl): Intent {
+        return Intent(WebAppIntentProcessor.ACTION_VIEW_PWA, Uri.parse(startUrl)).apply {
+            this.addCategory(Intent.CATEGORY_DEFAULT)
+            this.putUrlOverride(urlOverride)
+        }
+    }
+}
--- a/components/feature/pwa/src/main/java/mozilla/components/feature/pwa/intent/WebAppIntentProcessor.kt
+++ b/components/feature/pwa/src/main/java/mozilla/components/feature/pwa/intent/WebAppIntentProcessor.kt
@@ -10,7 +10,10 @@ import kotlinx.coroutines.runBlocking
 import mozilla.components.browser.session.Session
 import mozilla.components.browser.session.Session.Source
 import mozilla.components.browser.session.SessionManager
+import mozilla.components.browser.state.state.ExternalAppType
 import mozilla.components.concept.engine.EngineSession
+import mozilla.components.concept.engine.manifest.WebAppManifest
+import mozilla.components.feature.pwa.ext.getUrlOverride
 import mozilla.components.feature.intent.ext.putSessionId
 import mozilla.components.feature.intent.processing.IntentProcessor
 import mozilla.components.feature.pwa.ManifestStorage
@@ -45,13 +48,14 @@ class WebAppIntentProcessor(
 
         return if (!url.isNullOrEmpty() && matches(intent)) {
             val webAppManifest = runBlocking { storage.loadManifest(url) } ?: return false
+            val targetUrl = intent.getUrlOverride() ?: url
 
-            val session = Session(url, private = false, source = Source.HOME_SCREEN)
-            session.webAppManifest = webAppManifest
-            session.customTabConfig = webAppManifest.toCustomTabConfig()
+            val session = findExistingSession(webAppManifest) ?: createSession(webAppManifest, url)
 
-            sessionManager.add(session)
-            loadUrlUseCase(url, session, EngineSession.LoadUrlFlags.external())
+            if (targetUrl !== url) {
+                loadUrlUseCase(targetUrl, session, EngineSession.LoadUrlFlags.external())
+            }
+
             intent.flags = FLAG_ACTIVITY_NEW_DOCUMENT
             intent.putSessionId(session.id)
             intent.putWebAppManifest(webAppManifest)
@@ -60,6 +64,30 @@ class WebAppIntentProcessor(
         } else {
             false
         }
+    }
+
+    /**
+     * Returns an existing web app session that matches the manifest.
+     */
+    private fun findExistingSession(webAppManifest: WebAppManifest): Session? {
+        return sessionManager.all.find {
+            it.customTabConfig?.externalAppType == ExternalAppType.PROGRESSIVE_WEB_APP &&
+                    it.webAppManifest?.startUrl == webAppManifest.startUrl
+        }
+    }
+
+    /**
+     * Returns a new web app session.
+     */
+    private fun createSession(webAppManifest: WebAppManifest, url: String): Session {
+        return Session(url, private = false, source = Source.HOME_SCREEN)
+                .apply {
+                    this.webAppManifest = webAppManifest
+                    this.customTabConfig = webAppManifest.toCustomTabConfig()
+                }.also {
+                    sessionManager.add(it)
+                    loadUrlUseCase(url, it, EngineSession.LoadUrlFlags.external())
+                }
     }
 
     companion object {
--- a/samples/browser/src/main/java/org/mozilla/samples/browser/request/SampleRequestInterceptor.kt
+++ b/samples/browser/src/main/java/org/mozilla/samples/browser/request/SampleRequestInterceptor.kt
@@ -27,8 +27,17 @@ class SampleRequestInterceptor(val context: Context) : RequestInterceptor {
     ): InterceptionResponse? {
         return when (uri) {
             "sample:about" -> InterceptionResponse.Content("<h1>I am the sample browser</h1>")
-            else -> context.components.appLinksInterceptor.onLoadRequest(
-                engineSession, uri, hasUserGesture, isSameDomain, isRedirect, isDirectNavigation)
+            else -> {
+                var response = context.components.appLinksInterceptor.onLoadRequest(
+                        engineSession, uri, hasUserGesture, isSameDomain, isRedirect, isDirectNavigation)
+
+                if (response == null && !isDirectNavigation) {
+                    response = context.components.webAppInterceptor.onLoadRequest(
+                        engineSession, uri, hasUserGesture, isSameDomain, isRedirect, isDirectNavigation)
+                }
+
+                response
+            }
         }
     }
 
--- a/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
+++ b/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
@@ -269,6 +269,8 @@ class SystemEngineView @JvmOverloads constructor(
 
                                 super.shouldInterceptRequest(view, request)
                             }
+
+                            is InterceptionResponse.Deny -> super.shouldInterceptRequest(view, request)
                         }
                     }
                 }
--- a/components/concept/engine/src/main/java/mozilla/components/concept/engine/request/RequestInterceptor.kt
+++ b/components/concept/engine/src/main/java/mozilla/components/concept/engine/request/RequestInterceptor.kt
@@ -26,6 +26,11 @@ interface RequestInterceptor {
         data class Url(val url: String) : InterceptionResponse()
 
         data class AppIntent(val appIntent: Intent, val url: String) : InterceptionResponse()
+
+        /**
+         * Deny request without further action.
+         */
+        object Deny : InterceptionResponse()
     }
 
     /**
--- a/components/feature/pwa/src/main/java/mozilla/components/feature/pwa/ManifestStorage.kt
+++ b/components/feature/pwa/src/main/java/mozilla/components/feature/pwa/ManifestStorage.kt
@@ -19,10 +19,12 @@ import mozilla.components.feature.pwa.db.ManifestEntity
  * @param activeThresholdMs a timeout in milliseconds after which the storage will consider a manifest
  *                      as unused. By default this is [ACTIVE_THRESHOLD_MS].
  */
+@Suppress("TooManyFunctions")
 class ManifestStorage(context: Context, private val activeThresholdMs: Long = ACTIVE_THRESHOLD_MS) {
 
     @VisibleForTesting
     internal var manifestDao = lazy { ManifestDatabase.get(context).manifestDao() }
+    internal var installedScopes: MutableMap<String, String>? = null
 
     /**
      * Load a Web App Manifest for the given URL from disk.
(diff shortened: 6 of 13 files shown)


components/feature/pwa/src/main/java/mozilla/components/feature/pwa/intent/WebAppIntentProcessor.kt
```diff
@@ -61,6 +65,34 @@ class WebAppIntentProcessor(
...
+                .apply {
+                    this.webAppManifest = webAppManifest
+                    this.customTabConfig = webAppManifest.toCustomTabConfig()
+                }.also {
+                    sessionManager.add(it)
+                    loadUrl(url, it)
+                }
+    }
+
+    private fun loadUrl(url: String, session: Session) {
```
NotWoods, line 92: I don't feel like `loadUrl` is needed, we can just keep `loadUrlUseCase` inlined in `createSession`

components/feature/intent/src/main/java/mozilla/components/feature/intent/ext/IntentExtensions.kt
```diff
@@ -25,6 +26,14 @@ fun Intent.getSessionId(): String? = getStringExtra(EXTRA_SESSION_ID)
  */
 fun SafeIntent.getSessionId(): String? = getStringExtra(EXTRA_SESSION_ID)
 
+/**
+ * Retrieves [String] Url override from the intent.
+ *
+ * @return The URL override previously added with [putUrlOverride],
+ * or null if no URL was found.
+ */
+fun Intent.getUrlOverride(): String? = getStringExtra(EXTRA_URL_OVERRIDE)
```
NotWoods, line 35: Since this is just used for PWAs, let's move it to feature-pwa.

NotWoods, line 35: Personally I think it would be better to have either the manifest or the start url included as an extra, rather than the real URL. We already have some code for embedding a web app manifest into an intent.

components/feature/intent/src/main/java/mozilla/components/feature/intent/ext/IntentExtensions.kt
```diff
@@ -8,6 +8,7 @@ import android.content.Intent
 import mozilla.components.support.utils.SafeIntent
 
 const val EXTRA_SESSION_ID = "activeSessionId"
+const val EXTRA_URL_OVERRIDE = "urlOverride"
```
NotWoods, line 11: I'd like to move toward namespaced extras, so I'd prefer to use something like `mozilla.components.support.utils.extra.URL_OVERRIDE`. This matches the behaviour of the built-in Android actions & extras.

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/WebAppInterceptor.kt
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * This feature implements use cases for detecting and handling redirects to standalone web app activities.
+ *
+ * In the case of Android Intents that are not installed, and with no fallback, the user is prompted
+ * to search the installed market place.
+ *
+ * @param shortcutManager current shortcut manager instance to lookup web app install states
+ */
+@Suppress("LongParameterList")
```
NotWoods, line 24: Move the suppress call to the `onLoadRequest` function instead of the entire class

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/WebAppInterceptor.kt
```diff
@@ -0,0 +1,55 @@
...
+import kotlinx.coroutines.runBlocking
+import mozilla.components.concept.engine.EngineSession
+import mozilla.components.concept.engine.request.RequestInterceptor
+import mozilla.components.feature.intent.ext.putUrlOverride
+import mozilla.components.feature.pwa.intent.WebAppIntentProcessor
+
+/**
+ * This feature implements use cases for detecting and handling redirects to standalone web app activities.
+ *
+ * In the case of Android Intents that are not installed, and with no fallback, the user is prompted
```
NotWoods, line 19: I'm confused, can you explain a bit more about what this interceptor is doing? I don't see any marketplace code here. (although perhaps that's what happens if this class doesn't do any intercepting.)

Could we just move the current Fenix session on install, rather than creating a new interceptor?

Would the interceptor automatically move any requests into the corresponding PWA? For example, if I have the Twitter PWA installed and I happen to open a Twitter link in Fenix, does it automatically open the PWA? Is there a user prompt?

NotWoods: Well I wrote the above a little bit late in the day, and I was misreading my own feature request ticket. I can see that this is for intercepting requests and then reopening them in the corresponding installed PWA.

I think we should adjust the KDocs on this class, as the marketplace is mentioned but that's not handled by this interceptor. Theoretically an app could choose to only use this request interceptor. Perhaps something like:
```
This feature will intercept requests and reopen them in the corresponding installed PWA, if any.
```

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/WebAppInterceptor.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    override fun onLoadRequest(
+        engineSession: EngineSession,
+        uri: String,
+        hasUserGesture: Boolean,
+        isSameDomain: Boolean
+    ): RequestInterceptor.InterceptionResponse? {
+        return runBlocking {
+            when (manifestStorage.hasRecentManifest(uri, System.currentTimeMillis())) {
+                true -> {
```
NotWoods, line 37: I'd prefer an `if` statement instead of a `when` statement here :)

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/WebAppInterceptor.kt
```diff
@@ -0,0 +1,55 @@
...
+    private val manifestStorage: ManifestStorage
+) : RequestInterceptor {
+
+    override fun onLoadRequest(
+        engineSession: EngineSession,
+        uri: String,
+        hasUserGesture: Boolean,
+        isSameDomain: Boolean
+    ): RequestInterceptor.InterceptionResponse? {
+        return runBlocking {
```
NotWoods, line 35: I'm wary about having blocking IO on every request, but `AppLinksInterceptor` is similar. @rocketsroger @mcomella can you take a look?

components/feature/intent/src/main/java/mozilla/components/feature/intent/ext/IntentExtensions.kt
TitanNano: But we need the real URL to deep link into the PWA. If I for example click on a link for a tweet I want to end up on that tweet inside the PWA and not the start URL.

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/WebAppInterceptor.kt
TitanNano: I will fix the KDoc, I just coppied it form the AppLink feature and didn't adjust it properly. 

About your earlier comment:
> Would the interceptor automatically move any requests into the corresponding PWA? For example, if I have the Twitter PWA installed and I happen to open a Twitter link in Fenix, does it automatically open the PWA? Is there a user prompt?
Yes currently it directly will switch into the Twitter PWA. I was also thinking about a possible setting like for app links, or a prompt, but I don't know what's best here UX wise.

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/WebAppInterceptor.kt
TitanNano: I also had doubts about this when I wrote it. It doesn't feel any different, but I can not say what the real performance impact is.

components/feature/intent/src/main/java/mozilla/components/feature/intent/ext/IntentExtensions.kt
TitanNano: We also need this because we always have to use the same intent URI or otherwise android will create new Activities for each URI.

components/feature/pwa/src/main/java/mozilla/components/feature/pwa/WebAppInterceptor.kt
mcomella: Thanks for looping me in - sorry for the delay, I'm behind on my email.

I haven't measured it but we shouldn't block page load requests on IO. In general, we should never be doing blocking IO on any critical code paths (such as startup and page load) or on the main thread (I'm guessing this isn't the main thread but I wanted to include it for completeness): `runBlocking` is a red flag that we should be very careful about adding.

Some possible solutions:
- if possible, cache the results so we don't need to read from disk (seems based on the 
- kick off an asynchronous query for this value. When it returns, if it returns true, cancel the load and launch the intent (if I'm reading this correctly)
  - I'm slightly concerned that reading from this disk during page load could affect page load speed... but I'm not sure what to do about that @acreskeyMoz Do you have thoughts on this?

TitanNano: @NotWoods caching all scopes would be the easiest solution. We could just load all installed scopes into a cache. @mcomella when should we populate the cache though? Can we do this during start up, or would it negatively affect the app launch? We also would have to regularly refresh the cache.

`onLoadRequest` acutally expects an async value so doing this async would also be possible, but since android-components implemented the `RequestInterceptor`s synchronous, this would mean we have to now make interceptors optionally asynchronous. That seems a bit complicated.

rocketsroger: Can you please add an example on how you will be using this in the sample browser?

mcomella: > @mcomella when should we populate the cache though? Can we do this during start up, or would it negatively affect the app launch? We also would have to regularly refresh the cache.

We have a visual completeness queue that can run code just after startup: https://searchfox.org/mozilla-mobile/rev/d81a3c3fce58d3a6c3ebd3257ad1566c9bd462ef/fenix/app/src/main/java/org/mozilla/fenix/FenixApplication.kt#192 Be sure to launch it on a background thread though, assuming it doesn't need to be on the main thread.

> We also would have to regularly refresh the cache.

I don't know how Fenix does periodic jobs like this – @NotWoods do you know?

NotWoods: Rather than being periodic, we could just update the cache whenever a new PWA is installed. Old PWAs just "expire" after a certain amount of days so I think just re-populating the cache on startup is fine.

TitanNano: > Can you please add an example on how you will be using this in the sample browser? I am a bit worried about adding more `RequestInterceptor` since every `onLoadRequest` goes though them.

@rocketsroger I added this to example browser now. While doing so I noticed that it currently depends on the `AppLinkFeature` to actually launch the `Intent` because `AppLinkFeature` does exactly the right thing here. Is this acceptable or do we need a separate implementation of `onLaunchIntentRequest`?

acreskeyMoz: >       * I'm slightly concerned that reading from this disk during page load could affect page load speed... but I'm not sure what to do about that @acreskeyMoz Do you have thoughts on this?

Yes, any delay (even 10-20ms) in the initial pageload request will end up hurting our visual metrics. (Especially for simple and fast pages like www.google.com)
I can help to evaluate this change when it's ready.

TitanNano: @NotWoods @mcomella @acreskeyMoz this is no longer blocking and uses a cache which is populated during start up.

samples/browser/src/main/java/org/mozilla/samples/browser/BaseBrowserFragment.kt
```diff
@@ -207,6 +209,18 @@ abstract class BaseBrowserFragment : Fragment(), UserInteractionHandler {
             view = layout
         )
 
+        appLinksFeature.set(
```
rocketsroger, line 212: Is this needed? I don't see any usage of this.

samples/browser/src/main/java/org/mozilla/samples/browser/request/SampleRequestInterceptor.kt
```diff
@@ -21,12 +21,22 @@ class SampleRequestInterceptor(val context: Context) : RequestInterceptor {
...
         return when (uri) {
             "sample:about" -> InterceptionResponse.Content("<h1>I am the sample browser</h1>")
-            else -> context.components.appLinksInterceptor.onLoadRequest(
-                engineSession, uri, hasUserGesture, isSameDomain)
+            else -> {
+                var response = context.components.appLinksInterceptor.onLoadRequest(
+                        engineSession, uri, hasUserGesture, isSameDomain, isAppInitiated)
+
+                if (response == null && !isAppInitiated) {
+                    response = context.components.webAppInterceptor.onLoadRequest(
```
rocketsroger, line 34: Looks like we're opening PWA (if the session exists) regardless what the user preference is.  Is this what we want?

samples/browser/src/main/java/org/mozilla/samples/browser/BaseBrowserFragment.kt
TitanNano: @rocketsroger See my earlier comment: https://github.com/mozilla-mobile/android-components/pull/7367#discussion_r442032905

samples/browser/src/main/java/org/mozilla/samples/browser/request/SampleRequestInterceptor.kt
TitanNano: @rocketsroger currently this is not tied to any perference, yes. I personally think it would be good to let users enable / disable this behavior, but a preference was not mentioned in the original fenix issue mozilla-mobile/fenix#5772.

NotWoods:
Because there's still some UI needed to make this a good user experience, here's what I propose moving forward:

Let's merge in this PR for Android Components, but close the Fenix work for now. We can have this working in sample browser (and maybe Reference Browser) to test with.

We'll need to try and figure out a good UI for this feature. One option is to make it a user toggle, another is to display a list of choices when the link is opened. We'll also need to make this work alongside app links. (ie, what happens if both "Open in app" and "Open in PWA" are turned on?)

samples/browser/src/main/java/org/mozilla/samples/browser/BaseBrowserFragment.kt
rocketsroger: Sorry, I didn't see that comment.  My mistake.  Please take a look at:
https://github.com/mozilla-mobile/android-components/blob/552b63f50640604bef348e56ebe7cb2eb6fd015c/components/feature/app-links/src/main/java/mozilla/components/feature/app/links/AppLinksInterceptor.kt#L86-L87

This might fit what you want.  I'm not sure if you want your feature to be tied to app links.

TitanNano: @rocketsroger thanks for the input. I refactored this to take a `launchFromInterceptor` like the `AppLinksInterceptor`.

https://github.com/mozilla-mobile/android-components/pull/7367/files#diff-ff99e8e29f17903954c251d6b5aa67e3R33-R46

This way this feature does not depend on the `AppLinksFeature`, but if desired the intent could be launched via the app links feature. I think this is important, because the app links feature already has some behavior which might also be desired for Web Apps:
https://github.com/mozilla-mobile/android-components/blob/29cc847c254a78685f5a3bb80b4c21e47c52deb9/components/feature/app-links/src/main/java/mozilla/components/feature/app/links/AppLinksFeature.kt#L46-L75

TitanNano:
> (ie, what happens if both "Open in app" and "Open in PWA" are turned on?)

@NotWoods Right now app links take precedence over web apps.

samples/browser/src/main/java/org/mozilla/samples/browser/BaseBrowserFragment.kt
rocketsroger: Looks good to me.  Thanks

rocketsroger:
@TitanNano, make sure tests are fixed since you changed the `onLoadRequest` interface.  Also, please add tests for your component.  Take a look at `AppLinksInterceptorTest`.  Thanks,

mcomella:
@MarcLeclair Can you give this a look over from a perf perspective while I'm out of the office? I don't think we need formal measurement but ensure nothing looks suspicious. See https://github.com/mozilla-mobile/android-components/pull/7367#discussion_r439712570 for a previous concern.

MarcLeclair:
@mcomella (I'm late to answer, forgot to do so) but nothing jumps out of the ordinary in terms of performance. Unless we need those things right away to populate our UI on MAIN start up (or even VIEW), then it could delay but it doesn't seem like this is the case so I think we're good.

TitanNano:
@rocketsroger I added new tests and everything else also passes now.

rocketsroger:
> @rocketsroger I added new tests and everything else also passes now.

@TitanNano I have couple of concerns:
1. This should be disabled until UX is ready.
2. Have you tested with Fenix? Currently Fenix does not intercept any app initiated request.  https://github.com/mozilla-mobile/android-components/blob/f70216b224852172f9f781b905f94f8e2c0b4d9f/components/concept/engine/src/main/java/mozilla/components/concept/engine/request/RequestInterceptor.kt#L96.

TitanNano:
> > @rocketsroger I added new tests and everything else also passes now.
> 
> @TitanNano I have couple of concerns:
> 
>     1. This should be disabled until UX is ready.
> 
>     2. Have you tested with Fenix? Currently Fenix does not intercept any app initiated request.  https://github.com/mozilla-mobile/android-components/blob/f70216b224852172f9f781b905f94f8e2c0b4d9f/components/concept/engine/src/main/java/mozilla/components/concept/engine/request/RequestInterceptor.kt#L96
>        .

@rocketsroger 

1. Where would you like to disable this? If I understood @NotWoods correctly https://github.com/mozilla-mobile/android-components/pull/7367#issuecomment-648267039, he'd like to have it in sample browser. For Fenix this has to be integrated first before it will do anything.
2. a. There is a patch for Fenix https://github.com/mozilla-mobile/fenix/issues/5772 which worked with a previous iteration of this patch, but I will have to make some slight adjustments to it, to account for the most recent modifications. @NotWoods wanted to put the Fenix patch on hold until UX is ready though, so I haven't put more effort into it for now. 
b. Regarding the app initiated requests: The `WebAppInterceptor` doesn't want to intercept app initiated requests, so it's perfectly fine that Fenix doesn't intercept them. That's also the reason I filtered them out in the `SampleRequestInterceptor` https://github.com/mozilla-mobile/android-components/pull/7367/files#diff-511c28810156a10021024868446d5f52R34.

mcomella:
Removing self as reviewer as I defer to MarcLeclair from https://github.com/mozilla-mobile/android-components/pull/7367#issuecomment-653930390

rocketsroger:
> > > @rocketsroger I added new tests and everything else also passes now.
> > 
> > 
> > @TitanNano I have couple of concerns:
> > ```
> > 1. This should be disabled until UX is ready.
> > 
> > 2. Have you tested with Fenix? Currently Fenix does not intercept any app initiated request.  https://github.com/mozilla-mobile/android-components/blob/f70216b224852172f9f781b905f94f8e2c0b4d9f/components/concept/engine/src/main/java/mozilla/components/concept/engine/request/RequestInterceptor.kt#L96
> >    .
> > ```
> 
> @rocketsroger
> 
>     1. Where would you like to disable this? If I understood @NotWoods correctly [#7367 (comment)](https://github.com/mozilla-mobile/android-components/pull/7367#issuecomment-648267039), he'd like to have it in sample browser. For Fenix this has to be integrated first before it will do anything.
> 
>     2. a. There is a patch for Fenix [mozilla-mobile/fenix#5772](https://github.com/mozilla-mobile/fenix/issues/5772) which worked with a previous iteration of this patch, but I will have to make some slight adjustments to it, to account for the most recent modifications. @NotWoods wanted to put the Fenix patch on hold until UX is ready though, so I haven't put more effort into it for now.
>        b. Regarding the app initiated requests: The `WebAppInterceptor` doesn't want to intercept app initiated requests, so it's perfectly fine that Fenix doesn't intercept them. That's also the reason I filtered them out in the `SampleRequestInterceptor` https://github.com/mozilla-mobile/android-components/pull/7367/files#diff-511c28810156a10021024868446d5f52R34.

Ah I see.  I'm good with the change then.  Since this is only available in the sample browser.

TitanNano:
@rocketsroger @NotWoods who will give the final approval? 😄

rocketsroger:
@NotWoods I'm good with the change from my end.  But I'll defer to your approval since you have a outstanding change request on the review. ✅

Issue #7380, opened 2020-06-14
Let EngineSessions request release from their EngineView


TitanNano:
In mozilla-mobile/fenix#11543 we encounter a situation where one PWA activity would supersede another one, and they both should use the same session, but neither `onDestroy` nor `onStop` will be called until the new activity has started. The best solution here would be to let a `EngineSession` use it's `observer` to notify observing `EngineView`s that this session has to be released. We then can safely use the session in a new `EngineView`.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-6370)

cpeterso:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1796436

Change performed by the [Move to Bugzilla add-on](https://addons.mozilla.org/en-US/firefox/addon/move-to-bugzilla/).

PR #7381, opened 2020-06-14
Let EngineSessions request release from their EngineView

TitanNano:
For #7380 and mozilla-mobile/fenix#5772

`EngineSession`s should be able to notifiy observers that they want to be
releasd from their current `EngineView`.

---

### Pull Request checklist

### After merge


TitanNano:
> I haven't fully traced this change back to the problem in Fenix. But I think this is not safe solution since releasing the engine session like this is not guaranteed to stick. If an `EngineView` is rendering the selected session and you release here then it will still re-render the session the next time the feature is bound (e.g. after an app resume).

The problem in Fenix is that we are trying to use a Web App session in a new activity (which is replacing an existing activity) for the same Web App, but the system doesn't destroy the old activity before the new one is created.

TitanNano:
@pocmo do you have a suggestion for how I could proceed here?

PR #7427, opened 2020-06-17
Closes #7397: Integrate GV API for runtime.openOptionsPage

### LINKED ISSUE #7397: Integrate GV API for runtime.openOptionsPage
Currently `runtime.openOptionsPage` works for add-ons configured with `open_in_tab = true` as we get a `onNewTab` called on the tab delegate. However, for `open_in_tab = false`, there's a new mehod we haven't implemented yet:
https://bugzilla.mozilla.org/show_bug.cgi?id=1619766

csadilek:
Decided to go with the simplest solution that works, similar to Fennec/Desktop, and open a new tab when the options page is triggered via JavaScript.

We may want to make this customizable in the future and support deep linking into our AddosManager/Addon/Settings UI, but this doesn't seem needed right now. It's also tricky for us as we don't use `about:addons`, but have a native UI instead. The user experience definitely feels better this way too.

Other changes:
- Use M3 add-ons so we can test decentraleyes etc.
- Make sure our action popup (in Sample Browser) responds to `window.close`


csadilek:
bors r=Amejia481

PR #7429, opened 2020-06-17
Delay reading from shared preferences in reader view

NotWoods:
This is causing strict mode violations because the reading is done when the browser fragment is built.

---

### Pull Request checklist

### After merge


components/feature/readerview/src/main/java/mozilla/components/feature/readerview/internal/ReaderViewConfig.kt
```diff
@@ -0,0 +1,102 @@
...
+            .put(ACTION_VALUE_SHOW_FONT_SIZE, fontSize)
+            .put(ACTION_VALUE_SHOW_FONT_TYPE, fontType.name.toLowerCase(Locale.ROOT))
+            .put(ACTION_VALUE_SHOW_COLOR_SCHEME, colorScheme.name.toLowerCase(Locale.ROOT))
+
+        return JSONObject()
+            .put(ACTION_MESSAGE_KEY, ACTION_SHOW)
+            .put(ACTION_VALUE, configJson)
+    }
+
+    private inline fun <reified T : Enum<T>> getEnum(key: String, default: T): T {
```
csadilek, line 87: nit: getEnumFromPrefs?

components/feature/readerview/src/main/java/mozilla/components/feature/readerview/internal/ReaderViewConfig.kt
```diff
@@ -0,0 +1,102 @@
...
+        val enumName = prefs.getString(key, default.name) ?: default.name
+        return enumValueOf(enumName)
+    }
+
+    private fun <T : Enum<T>> setEnumAndSend(key: String, action: String, value: T) {
```
csadilek, line 92: nit: I think it's much easier to read if we just inline those two calls again and remove `setEnumAndSend`. We only have two of those and `setEnumAndSend` doesn't make this easier to read...

components/feature/readerview/src/main/java/mozilla/components/feature/readerview/internal/ReaderViewConfig.kt
```diff
@@ -0,0 +1,102 @@
...
+            return fontSizeCache!!
+        }
+        set(value) {
+            val diff = value - fontSize
+            fontSizeCache = value
+            prefs.edit().putInt(FONT_SIZE_KEY, value).apply()
+            sendMessage(ACTION_CHANGE_FONT_SIZE) { put(ACTION_VALUE, diff) }
+        }
+
+    fun createShowReaderMessage(): JSONObject {
```
csadilek, line 76: See above, let's move this back out. I think it makes more sense for it to live in the feature, together with `createHideReaderMessage`.

components/feature/readerview/src/main/java/mozilla/components/feature/readerview/ReaderViewFeature.kt
```diff
@@ -65,46 +61,19 @@ class ReaderViewFeature(
     internal var extensionController = WebExtensionController(READER_VIEW_EXTENSION_ID, READER_VIEW_EXTENSION_URL)
 
     @VisibleForTesting
-    internal val config = Config(context.getSharedPreferences(SHARED_PREF_NAME, Context.MODE_PRIVATE))
+    internal val config = ReaderViewConfig(
+        prefs = context.getSharedPreferences(SHARED_PREF_NAME, Context.MODE_PRIVATE)
```
csadilek, line 65: I assume the call to `getSharedPreferences` is fine and doesn't cause a violation? It's only the `getString`/`getInt` calls?

NotWoods: I'm not sure, cc @mcomella

sraturi: It should give you  `onDiskRead` violation. There is a `StrictMode.allowThreadPolicy.resetAfter` function you can use if moving this to a IO coroutine is not possible.

csadilek: @NotWoods Your goal here is to *not* read from shared prefs when the fragment (and feature) is created, right? Instead wait until its needed and the appearance controls are used. I think that's good enough, because appearance controls won't be perf. sensitive (and are rarely used, if ever).

So, probably not worth the complexity of moving to an IO thread in this case.

NotWoods: Yeah, my intent here is just to move the logic away from startup. I refactored the code to make `getSharedPreferences` lazy as well.

csadilek: I was about to suggest making this a `lazy` ref., but you already did that in your latest change. :).

Let me do a round of testing, but otherwise, I think this can land.

NotWoods:
bors r=csadilek

PR #7431, opened 2020-06-18
Issue #7185: Delete thumbnails when a session is removed

### LINKED ISSUE #7185: Delete thumbnails when a session is removed
If we continue to follow the direction of using the session ID for identifying tabs, we should also make sure we remove the thumbnails for the tab when the tab is removed.

We can do this using the same `ThumbnailMiddleware` we added for getting new thumbnails.

gabrielluong:
---

Fixes #7185

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/storage/ThumbnailStorage.kt
+++ b/components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/storage/ThumbnailStorage.kt
@@ -43,6 +43,25 @@ class ThumbnailStorage(
     private val scope = CoroutineScope(jobDispatcher)
 
     /**
+     * Clears all the stored thumbnails in the disk cache.
+     */
+    fun clearThumbnails(): Job =
+        scope.launch {
+            logger.debug("Cleared all thumbnails from disk")
+            sharedDiskCache.clear(context)
+        }
+
+    /**
+     * Deletes the given thumbnail [Bitmap] from the disk cache with the provided session ID or url
+     * as its key.
+     */
+    fun deleteThumbnail(sessionIdOrUrl: String): Job =
+        scope.launch {
+            logger.debug("Removed thumbnail from disk (sessionIdOrUrl = $sessionIdOrUrl)")
+            sharedDiskCache.removeThumbnailData(context, sessionIdOrUrl)
+        }
+
+    /**
      * Asynchronously loads a thumbnail [Bitmap] for the given session ID or url.
      */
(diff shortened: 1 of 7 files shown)


gabrielluong:
I think we need to move ahead with simply removing the associated thumbnail whenever the session is removed, and not handle the undo case as a first iteration. We can be filling up the disk cache and provide no way to clear it, which isn't great otherwise.

components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/storage/ThumbnailStorage.kt
```diff
@@ -42,6 +42,24 @@ class ThumbnailStorage(
         context.resources.getDimensionPixelSize(R.dimen.mozac_browser_thumbnails_maximum_size)
     private val scope = CoroutineScope(jobDispatcher)
 
+    /**
+     * Clears all the stored thumbnails in the disk cache.
+     */
+    fun clearThumbnails() {
+        logger.debug("Cleared all thumbnails from disk")
```
jonalmeida, line 49: This should be a in coroutine scope too.

components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/utils/ThumbnailDiskCache.kt
```diff
@@ -24,7 +23,6 @@ class ThumbnailDiskCache {
     private var thumbnailCache: DiskLruCache? = null
     private val thumbnailCacheWriteLock = Any()
 
-    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
     internal fun clear(context: Context) {
         getThumbnailCache(context).delete()
```
jonalmeida, line 27: Must have missed this during the previous review, shouldn't this be in a synchronized block as well?

components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/storage/ThumbnailStorage.kt
gabrielluong: Fixed.

jonalmeida:
Whoops, approved the wrong PR. We should still fix the other review comment.

components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/utils/ThumbnailDiskCache.kt
gabrielluong: Fixed.

gabrielluong:
bors r=jonalmeida

PR #7447, opened 2020-06-19
For #7475: Fix tabcounter flicker

hakkikaancaliskan:
**Warning**: I'm amateur. Double check everything.

I've fixed tab counter flicker issue in fenix and @gabrielluong wanted to upstream that changes to ac.

### Pull Request checklist

### After merge

### DIFF
--- a/components/ui/tabcounter/src/main/java/mozilla/components/ui/tabcounter/TabCounter.kt
+++ b/components/ui/tabcounter/src/main/java/mozilla/components/ui/tabcounter/TabCounter.kt
@@ -19,7 +19,6 @@ import mozilla.components.support.ktx.android.view.onNextGlobalLayout
 import mozilla.components.support.utils.DrawableUtils
 import java.text.NumberFormat
 
-@Suppress("LargeClass")
 open class TabCounter @JvmOverloads constructor(
     context: Context,
     attrs: AttributeSet? = null,
@@ -35,7 +34,6 @@ open class TabCounter @JvmOverloads constructor(
     internal var currentTextRatio: Float = 0.toFloat()
 
     init {
-
         // Default TabCounter tint, could be override by the caller
         @ColorInt val defaultTabCounterTint = ContextCompat.getColor(context, R.color.mozac_ui_tabcounter_default_tint)
         val typedArray = context.obtainStyledAttributes(attrs, R.styleable.TabCounter, defStyle, 0)
@@ -48,10 +46,8 @@ open class TabCounter @JvmOverloads constructor(
         box = findViewById(R.id.counter_box)
         bar = findViewById(R.id.counter_bar)
         text = findViewById(R.id.counter_text)
-        text.text = DEFAULT_TABS_COUNTER_TEXT
-        val shiftOneDpForDefaultText = TypedValue.applyDimension(
-            TypedValue.COMPLEX_UNIT_DIP, 1f, context.resources.displayMetrics).toInt()
-        text.setPadding(0, 0, 0, shiftOneDpForDefaultText)
+
+        setCount(count)
 
         if (tabCounterTint != defaultTabCounterTint) {
             tintDrawables(tabCounterTint)
@@ -60,6 +56,9 @@ open class TabCounter @JvmOverloads constructor(
         animationSet = createAnimatorSet()
     }
 
+    /**
+     * Returns current tab count from tab counter view.
+     */
     fun getText(): CharSequence {
         return text.text
     }
@@ -182,7 +181,7 @@ open class TabCounter @JvmOverloads constructor(
         // Fadeout in 66ms, after firstAnimator with delay 32ms (54~58, 4 frames).
         val fadeOut = ObjectAnimator.ofFloat(bar, "alpha",
             ANIM_BAR_FADEOUT_FROM, ANIM_BAR_FADEOUT_TO).setDuration(ANIM_BAR_FADEOUT_DURATION)
-        fadeOut.startDelay = (ANIM_BAR_FADEOUT_DELAY).toLong() // delay 3 frames after firstAnimator
+        fadeOut.startDelay = (ANIM_BAR_FADEOUT_DELAY) // delay 3 frames after firstAnimator
 
         // Move down on y-axis, from -7.0 to 0 in 16ms, after fadeOut (58~59 1 frame).
         val moveDown2 = ObjectAnimator.ofFloat(bar, "translationY",
@@ -216,7 +215,7 @@ open class TabCounter @JvmOverloads constructor(
         // FadeIn in 66 ms, after fadeOut with delay 96ms (57~61, 4 frames).
         val fadeIn = ObjectAnimator.ofFloat(text, "alpha",
             ANIM_TEXT_FADEIN_FROM, ANIM_TEXT_FADEIN_TO).setDuration(ANIM_TEXT_FADEIN_DURATION)
-        fadeIn.startDelay = (ANIM_TEXT_FADEIN_DELAY).toLong() // delay 6 frames after fadeOut
+        fadeIn.startDelay = (ANIM_TEXT_FADEIN_DELAY) // delay 6 frames after fadeOut
 
         // Move down on y-axis, from 0 to 4.4 in 66ms, with fadeIn (57~61, 4 frames).
         val moveDown = ObjectAnimator.ofFloat(text, "translationY",
@@ -259,11 +258,9 @@ open class TabCounter @JvmOverloads constructor(
     }
 
     companion object {
-
         internal const val MAX_VISIBLE_TABS = 99
 
         internal const val SO_MANY_TABS_OPEN = "∞"
-        internal const val DEFAULT_TABS_COUNTER_TEXT = ":)"
 
         internal const val ONE_DIGIT_SIZE_RATIO = 0.6f
         internal const val TWO_DIGITS_SIZE_RATIO = 0.5f
--- a/components/ui/tabcounter/src/test/java/mozilla/components/ui/tabcounter/TabCounterTest.kt
+++ b/components/ui/tabcounter/src/test/java/mozilla/components/ui/tabcounter/TabCounterTest.kt
@@ -6,7 +6,6 @@ package mozilla.components.ui.tabcounter
 
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import mozilla.components.support.test.robolectric.testContext
-import mozilla.components.ui.tabcounter.TabCounter.Companion.DEFAULT_TABS_COUNTER_TEXT
 import mozilla.components.ui.tabcounter.TabCounter.Companion.ONE_DIGIT_SIZE_RATIO
 import mozilla.components.ui.tabcounter.TabCounter.Companion.SO_MANY_TABS_OPEN
 import mozilla.components.ui.tabcounter.TabCounter.Companion.TWO_DIGITS_SIZE_RATIO
@@ -18,12 +17,12 @@ import org.junit.runner.RunWith
 class TabCounterTest {
 
     @Test
-    fun `Default tab count is a smiley face`() {
+    fun `Default tab count is 0`() {
         val tabCounter = TabCounter(testContext)
 
-        assertEquals(DEFAULT_TABS_COUNTER_TEXT, tabCounter.getText())
+        assertEquals("0", tabCounter.getText())
 
-        assertEquals(0.toFloat(), tabCounter.currentTextRatio)
+        assertEquals(ONE_DIGIT_SIZE_RATIO, tabCounter.currentTextRatio)
     }
 
     @Test


components/ui/tabcounter/src/main/java/mozilla/components/ui/tabcounter/TabCounter.kt
```diff
@@ -19,8 +19,7 @@ import mozilla.components.support.ktx.android.view.onNextGlobalLayout
 import mozilla.components.support.utils.DrawableUtils
 import java.text.NumberFormat
 
-@Suppress("LargeClass")
-open class TabCounter @JvmOverloads constructor(
```
rocketsroger, line 23: Is there a reason that you're removing `open`?

hakkikaancaliskan: Because it was pointed by Android studio and I'm unable to find any reason to be let it be as is.
Of course I'm amateur, i can add it back if you want.

components/ui/tabcounter/src/main/java/mozilla/components/ui/tabcounter/TabCounter.kt
```diff
@@ -259,11 +258,11 @@ open class TabCounter @JvmOverloads constructor(
     }
 
     companion object {
+        internal var INTERNAL_COUNT = 0
```
rocketsroger, line 261: Please don't use all cap for variable that is not constant.

components/ui/tabcounter/src/main/java/mozilla/components/ui/tabcounter/TabCounter.kt
rocketsroger: yes, I think it would be better to keep `open` if we're not sure the original intention.  Thanks

components/ui/tabcounter/src/main/AndroidManifest.xml
```diff
@@ -2,4 +2,6 @@
    - License, v. 2.0. If a copy of the MPL was not distributed with this
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="mozilla.components.ui.tabcounter" />
+    package="mozilla.components.ui.tabcounter" >
+    <application android:supportsRtl="true" />
```
rocketsroger, line 6: I think this will require another issue.  Adding RTL support makes sense but this requires a lot of testing to confirm that we do support it correctly in `tabcounter`

components/ui/tabcounter/src/main/java/mozilla/components/ui/tabcounter/TabCounter.kt
hakkikaancaliskan: This was necessary in Fenix because count got reset while navigating. I did it also here too.
https://github.com/mozilla-mobile/fenix/pull/11738

components/ui/tabcounter/src/main/AndroidManifest.xml
hakkikaancaliskan: Hmm, i just added it for detekt.

hakkikaancaliskan:
@rocketsroger Line 54 in this pr

rocketsroger:
> @rocketsroger Line 54 in this pr

Ah, I see.  That's a great find.  I think you're on the right track.  
I would suggest two things:
1. Let's create a Android component issue so the commit message can link to it properly.
2. I think you're absolutely right to remove line 51 - 54.  But simplify everything else and just set the count to 0 in the constructor.

Thanks for finding this!

components/ui/tabcounter/src/main/AndroidManifest.xml
hakkikaancaliskan: how should it be because it was wanted by detekt. Need to convert it to false?

hakkikaancaliskan:
@rocketsroger
Annotation 2020-06-24 003540
What do you think about this? Should we make TabCounter final?

hakkikaancaliskan:
>     1. Let's create an Android component issue so the commit message can link to it properly.
#7475

components/ui/tabcounter/src/main/AndroidManifest.xml
hakkikaancaliskan: Completely removed.

components/ui/tabcounter/src/main/java/mozilla/components/ui/tabcounter/TabCounter.kt
```diff
@@ -60,32 +57,35 @@ open class TabCounter @JvmOverloads constructor(
         animationSet = createAnimatorSet()
     }
 
-    fun getText(): CharSequence {
-        return text.text
+    /**
```
rocketsroger, line 60: Don't think we need to change this

components/ui/tabcounter/src/main/java/mozilla/components/ui/tabcounter/TabCounter.kt
```diff
@@ -19,19 +19,18 @@ import mozilla.components.support.ktx.android.view.onNextGlobalLayout
 import mozilla.components.support.utils.DrawableUtils
 import java.text.NumberFormat
 
-@Suppress("LargeClass")
 open class TabCounter @JvmOverloads constructor(
     context: Context,
     attrs: AttributeSet? = null,
-    defStyle: Int = 0
+    defStyle: Int = 0,
+    private var tabCount: Int = 0
```
rocketsroger, line 26: Is there a reason we need this to be a property?

components/ui/tabcounter/src/main/AndroidManifest.xml
```diff
@@ -1,5 +1,4 @@
 <!-- This Source Code Form is subject to the terms of the Mozilla Public
    - License, v. 2.0. If a copy of the MPL was not distributed with this
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
```
rocketsroger, line 4: This sense but I would suggest not touching this unless the change needs it

components/ui/tabcounter/src/test/java/mozilla/components/ui/tabcounter/TabCounterTest.kt
```diff
@@ -6,24 +6,28 @@ package mozilla.components.ui.tabcounter
...
 class TabCounterTest {
 
     @Test
-    fun `Default tab count is a smiley face`() {
+    // This "A" is intended because we want to run this function first
+    fun `A Default tab count is 0`() {
         val tabCounter = TabCounter(testContext)
 
-        assertEquals(DEFAULT_TABS_COUNTER_TEXT, tabCounter.getText())
+            assertEquals("0", tabCounter.getText())
```
rocketsroger, line 28: Extra tab space.  Thanks,

components/ui/tabcounter/src/main/java/mozilla/components/ui/tabcounter/TabCounter.kt
hakkikaancaliskan: >But simplify everything else and just set the count to 0 in the constructor.
You said set in constructor and i did, i get you wrong?

components/ui/tabcounter/src/main/java/mozilla/components/ui/tabcounter/TabCounter.kt
hakkikaancaliskan: But we compare strings in test. They will fail, right?

components/ui/tabcounter/src/main/java/mozilla/components/ui/tabcounter/TabCounter.kt
rocketsroger: Sorry, I think I made it more confusing.  I meant making a call inside the constructor to set it to 0.  Didn't mean to sound like I wanted a new property.  My bad.

components/ui/tabcounter/src/main/java/mozilla/components/ui/tabcounter/TabCounter.kt
rocketsroger: Ah I see. You can just call `toString()` in the test.  Changing this changes the public API of the class which might have unforeseen consequences.

components/ui/tabcounter/src/main/java/mozilla/components/ui/tabcounter/TabCounter.kt
hakkikaancaliskan: I'm completely lost :)
The `count` used in `setCountWithAnimation` is different than current `tabCount`. We need `tabCount` to get initial state.
There is two seperate tab count currently and before this pr. `tabCount` is used for `don't animate on initial state` etc. but `count` is used for getting tab count from `homefragment` etc.

If you mean something another, please send the code how it should be :(

components/ui/tabcounter/src/test/java/mozilla/components/ui/tabcounter/TabCounterTest.kt
```diff
@@ -6,24 +6,28 @@ package mozilla.components.ui.tabcounter
...
 
+// We need order because we want to run "Default tab count is 0" function first.
 @RunWith(AndroidJUnit4::class)
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
 class TabCounterTest {
 
     @Test
-    fun `Default tab count is a smiley face`() {
+    // This "A" is intended because we want to run this function first
+    fun `A Default tab count is 0`() {
```
rocketsroger, line 25: Each test should be creating its own brand new `tabCounter`.  Does order matter here?  Can you please help investigate this so we don't need the tests to run in a specific order?  Thanks,

components/ui/tabcounter/src/main/java/mozilla/components/ui/tabcounter/TabCounter.kt
hakkikaancaliskan: There was too much but okey, gonna change them.
Should I change TabCounterToolbarButtonTest.kt because i didn't touch them.

resim

components/ui/tabcounter/src/test/java/mozilla/components/ui/tabcounter/TabCounterTest.kt
hakkikaancaliskan: This didn't worked while i was trying but now they're working, confused :)
Gonna roll it back.

components/ui/tabcounter/src/main/java/mozilla/components/ui/tabcounter/TabCounter.kt
rocketsroger: That's weird.  Why didn't we have this issue before this?  Is there something we're missing?

hakkikaancaliskan: Wait, its working fine without `.toString()`.
Does kotlin compare CharSequence and string?

hakkikaancaliskan: I've pushed commit, wait and see :)

components/ui/tabcounter/src/main/java/mozilla/components/ui/tabcounter/TabCounter.kt
rocketsroger: It could be me.  I don't know this code as well as you do.  But I was thinking keeping the `private var count: Int = 0`.  Then on line 51 just call setCount(0).  Do you see an issue with this?

hakkikaancaliskan: I was moved this to companion object because everytime we run TabCounter.kt, count got reset thus stays in initial state and never animating tab counter. Line 69.
I've also opened issue about this https://github.com/mozilla-mobile/fenix/issues/11822. Tab tray covers tab counter and animations never seen by user.
What do you think?

rocketsroger: I see.  Moving to companion object is to fix the other issue?  If we don't move to a companion object we won't fix the flickering issue?  

I am asking this because I would prefer one fix for one issue.  If we can fix the flickering without moving to companion object, I would rather merge the change without moving to companion object.  After this you can always create a new A-C issue and put up a new PR for the animations problem

hakkikaancaliskan: We can fix flicker without this change. Gonna roll it back.

components/ui/tabcounter/src/main/java/mozilla/components/ui/tabcounter/TabCounter.kt
hakkikaancaliskan: It's working fine.

hakkikaancaliskan:
Ok, looks like pr ready to land. I suggest you to review pr from scratch :)

rocketsroger:
bors r=rocketsroger

gabrielluong:
I would also recommend squashing all your commits.

hakkikaancaliskan:
> 
> 
> I would also recommend squashing all your commits.

I could try but I'm not good with git :) thus I will fuck this pr up.
Gonna try it.

rocketsroger:
> I would also recommend squashing all your commits.

Agreed.  @hakkikaancaliskan 
Here's a good link on how to do it:
https://github.com/wprig/wprig/wiki/How-to-squash-commits

hakkikaancaliskan:
> 
> 
> > I would also recommend squashing all your commits.
> 
> Agreed. @hakkikaancaliskan
> Here's a good link on how to do it:
> https://github.com/wprig/wprig/wiki/How-to-squash-commits

Done! Yay!
I should save this commands somewhere lol

rocketsroger:
bors r=rocketsroger

gabrielluong:
@hakkikaancaliskan Thank you for patches! 🌟

Issue #3648, opened 2019-07-04
Add share option to image context menu


NotWoods:
[Originally filed](https://github.com/mozilla-mobile/fenix/issues/3621) by @BlackLotus:
> It's not possible to send/share media through the media menu.
> Examples 1:
> Graphics share menu only has save and send url, you can't share images out of the menu anymore. So right now you have to save, open and then share the image.
> ...
> 
> It would be nice to have at least feature parity with the mobile firefox.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-2995)

pocmo:
This dropped of the list because that requires a "download" step before sharing and we didn't have anything for that in place. :)

NotWoods:
Blocked by #3551

davidwill91:
Is there any chance this will be added in the future?

csadilek:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1793138

Change performed by the [Move to Bugzilla add-on](https://addons.mozilla.org/en-US/firefox/addon/move-to-bugzilla/).

PR #7462, opened 2020-06-23
Close #7461: Add share image context menu

### LINKED ISSUE #7461: Context Menu: Add share image option
Needed for https://github.com/mozilla-mobile/focus-android/issues/4527.

We don't have GV support for retrieving the Bitmap for this, so we share the source URI instead.

GV bug filed: https://bugzilla.mozilla.org/show_bug.cgi?id=1647568

jonalmeida:
---

Fixes #7461 for https://github.com/mozilla-mobile/focus-android/issues/4527.

Filed bug for getting a better behaviour from GV: https://bugzilla.mozilla.org/show_bug.cgi?id=1647568
### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/contextmenu/src/main/java/mozilla/components/feature/contextmenu/ContextMenuCandidate.kt
+++ b/components/feature/contextmenu/src/main/java/mozilla/components/feature/contextmenu/ContextMenuCandidate.kt
@@ -15,6 +15,7 @@ import mozilla.components.browser.state.state.content.DownloadState
 import mozilla.components.concept.engine.HitResult
 import mozilla.components.feature.tabs.TabsUseCases
 import mozilla.components.feature.app.links.AppLinksUseCases
+import mozilla.components.support.ktx.android.content.share
 
 /**
  * A candidate for an item to be displayed in the context menu.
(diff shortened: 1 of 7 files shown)


jonalmeida:
> Should we use the download feature for this? Looks good otherwise

I left that note in the GV bug as well. It seemed a bit odd to me to have to do re-download an image we already have in web content so I opted for matching the current Focus behaviour for now. Depending on the result of the GV bug, we can amend this.

jonalmeida:
bors r=NotWoods

pocmo:
The actual sharing is #3648, I think. Focus distinguishes between the two too:
https://github.com/mozilla-mobile/focus-android/blob/master/app/src/main/java/org/mozilla/focus/menu/context/WebContextMenu.kt#L239-L258

I wonder if sharing the image URL would find more targets if we had a mime type for the intent. But looks like Focus didn't have that either and GV doesn't give us one. We could only guess it from the file extension.

jonalmeida:
> This dropped of the list because that requires a "download" step before sharing and we didn't have anything for that in place. :)

@pocmo would that mean downloading the image to share is reasonable solution? My preference is to get something from GV but thinking about it more, that probably is harder for them to reason about with other mime types.

PR #7473, opened 2020-06-23
Bug 1647821 - Make `onExperimentUpdated` execute when using debug tools.

travis79:


components/service/experiments/proguard-rules.pro
```diff
@@ -19,3 +19,6 @@
 # If you keep the line number information, uncomment this to
 # hide the original source file name.
 #-renamesourcefileattribute SourceFile
+
+# Experiments specific protections
+-keep class mozilla.components.service.experiments.** { *; }
```
Dexterp37, line 24: What's the intent of this? If you want users of the library to retain this, you need to add this rule to a `proguard-rules-consumer.pro` (see [this example](https://searchfox.org/glean/source/glean-core/android/proguard-rules-consumer.pro)) and then advertise it in your [gradle file](https://searchfox.org/glean/rev/c6e1349fdb05baec59aaa5f465ba65e6cdd3b7d3/glean-core/android/build.gradle#53,59).

travis79: Ahh, thanks for this!  I wasn't involved with messing with proguard/R8 rules when we were having problems with it so I guess I didn't realize that they needed to be implemented this way for consuming apps.  Let me change it :)

travis79:
The context on calling the callback here is that it will trigger the callback that the app is expecting when a developer or QA analyst uses the debug tools to force override the experiment enrollment.  Since the callback is intended for the app to know when it is appropriate to check experiment enrollment, it is important to trigger this when the debug tools change the active experiment so that the app can take action.

sblatz:
Looks like the builds are all really sad

PR #7481, opened 2020-06-24
Issue #7480: Migrate feature-session to use BrowserStore.

pocmo:
Starting to migrate some classes in `feature-session`.


components/feature/session/build.gradle
```diff
@@ -21,7 +21,14 @@ android {
     }
 }
 
+tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
```
NotWoods, line 24: Let's use lazy tasks here
```suggestion
tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).configureEach() {
```

components/feature/session/src/main/java/mozilla/components/feature/session/FullScreenFeature.kt
```diff
@@ -39,12 +63,22 @@ open class FullScreenFeature(
...
+            if (it.inFullScreen) {
+                sessionUseCases.exitFullscreen(it.tabId)
                 return true
             }
         }
         return false
     }
 }
+
+private data class Observation(
```
NotWoods, line 76: Some simple KDoc would be nice to have here that mentions Observation is used to just hold the fields that this feature cares about

components/feature/session/src/test/java/mozilla/components/feature/session/FullScreenFeatureTest.kt
```diff
@@ -4,97 +4,386 @@
...
 class FullScreenFeatureTest {
+    private val testDispatcher = TestCoroutineDispatcher()
 
-    private val sessionManager: SessionManager = mock()
-    private val selectedSession: Session = mock()
-    private val useCases = spy(SessionUseCases(sessionManager))
+    @Before
+    @ExperimentalCoroutinesApi
+    fun setUp() {
+        Dispatchers.setMain(testDispatcher)
```
NotWoods, line 41: Can `MainCoroutineRule` be used instead here?

pocmo: We should consider refactoring `MainCoroutineRule` to use a `TestCoroutineDispatcher` by default. So that it is easier to get access to methods like `advanceUntilIdle()`.

pocmo:
I'll keep that PR open until the next AC release went out. And I may add some more refactorings to it. Will re-request a review after that. :)

pocmo:
Okay, migrated some more classes in `feature-session`. This is ready for another review. :)

components/feature/session/src/test/java/mozilla/components/feature/session/FullScreenFeatureTest.kt
NotWoods: I'll file another issue for that. We can still use the rule here as you can pass the test dispatcher into the rule's constructor.

components/feature/session/src/main/java/mozilla/components/feature/session/engine/EngineViewPresenter.kt
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * Stop presenter from updating view.
+     */
+    fun stop() {
+        scope?.cancel()
+    }
+
+    private fun onEngineSession(engineSession: EngineSession?) {
+        if (engineSession == null) {
+            engineView.release()
```
csadilek, line 51: Because we implemented `ifChanged` so it always notifies the initial value (if not observed before) we would now call `release` one time, initially, right? This doesn't seem to a problem though.

pocmo: Yeah, that is correct. In cases where there is no `EngineSession` yet we'd call `release()` once. I also think that shouldn't be a problem.

pocmo:
bors r=NotWoods,csadilek

PR #7492, opened 2020-06-24
Closes #7483: Fix telemetry foreground service notification string

### LINKED ISSUE #7483: SendCrashTelemetryService should use a different notification text
Currently `SendCrashTelemetryService` uses the same notification has the service that sends crash reports:
https://github.com/mozilla-mobile/android-components/blob/master/components/lib/crash/src/main/java/mozilla/components/lib/crash/service/SendCrashTelemetryService.kt#L30

This is slightly confusing since we will show the crash reporter prompt (where we ask the user whether they want to send a crash report) and at the same time the notification says "Sending crash report to Mozilla". Maybe we could use a string like "Collecting crash data" or something like that?

rocketsroger:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/lib/crash/src/main/res/values/strings.xml
+++ b/components/lib/crash/src/main/res/values/strings.xml
@@ -24,6 +24,9 @@
     <!-- Label of notification showing that the crash report service is running. %1$s will be replaced with the name of the organization (e.g. Mozilla). -->
     <string name="mozac_lib_send_crash_report_in_progress">Sending crash report to %1$s</string>
 
+    <!-- Label of notification showing that the telemetry report service is running. %1$s will be replaced with the name of the organization (e.g. Mozilla). -->
+    <string name="mozac_lib_collecting_telemetry_data_in_progress">Collecting telemetry data for %1$s</string>
+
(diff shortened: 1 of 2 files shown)


components/lib/crash/src/main/res/values/strings.xml
```diff
@@ -24,6 +24,9 @@
     <!-- Label of notification showing that the crash report service is running. %1$s will be replaced with the name of the organization (e.g. Mozilla). -->
     <string name="mozac_lib_send_crash_report_in_progress">Sending crash report to %1$s</string>
 
+    <!-- Label of notification showing that the telemetry report service is running. %1$s will be replaced with the name of the organization (e.g. Mozilla). -->
+    <string name="mozac_lib_send_telemetry_report_in_progress">Sending telemetry report to %1$s</string>
```
pocmo, line 28: I think "sending" could still be misleading. If telemetry is disabled then I think we'd still start this service, but Glean itself would not send any data. Maybe let's say something about us collecting the data?

rocketsroger: That's true.  We are just collecting the data, not technically sending it.  I'll update.

Issue #7337, opened 2020-06-11
[Bug] Text in error pages is being read twice by screen readers.


mcarare:
Original Fenix issue: https://github.com/mozilla-mobile/fenix/issues/8635.

This was fixed for error pages in https://github.com/mozilla-mobile/android-components/pull/5299, but the new error pages use `createUrlEncodedErrorPage` that does not contain that fix. Same fix will be implemented here, with some changes, depending on the response to https://github.com/mozilla-mobile/fenix/issues/8635#issuecomment-642583934.

mcarare:
Closing this, as the Fenix issue is marked as fixed:https://github.com/mozilla-mobile/fenix/issues/8635#issuecomment-663414977.

PR #7503, opened 2020-06-25
For #7337: Prevent text in <ul> tags being read twice by screen readers.

mcarare:
See also : https://github.com/mozilla-mobile/android-components/pull/5299.
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/errorpages/src/main/java/mozilla/components/browser/errorpages/ErrorPages.kt
+++ b/components/browser/errorpages/src/main/java/mozilla/components/browser/errorpages/ErrorPages.kt
@@ -93,7 +93,7 @@ object ErrorPages {
          * Warning: When updating these params you WILL cause breaking changes that are undetected
          * by consumers. Update the README accordingly.
          */
-        return "resource://android/assets/$htmlResource?" +
+        var urlEncodedErrorPage = "resource://android/assets/$htmlResource?" +
             "&title=$title" +
             "&button=$button" +
             "&description=$description" +
@@ -103,6 +103,10 @@ object ErrorPages {
             "&badCertTechInfo=$badCertTechInfo" +
             "&badCertGoBack=$badCertGoBack" +
             "&badCertAcceptTemporary=$badCertAcceptTemporary"
+
+        urlEncodedErrorPage = urlEncodedErrorPage
+            .replace("<ul>", "<ul role=\"presentation\">")
+        return urlEncodedErrorPage
     }
 }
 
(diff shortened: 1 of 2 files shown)


components/browser/errorpages/src/main/java/mozilla/components/browser/errorpages/ErrorPages.kt
```diff
@@ -103,6 +103,10 @@ object ErrorPages {
             "&badCertTechInfo=$badCertTechInfo" +
             "&badCertGoBack=$badCertGoBack" +
             "&badCertAcceptTemporary=$badCertAcceptTemporary"
+
+        urlEncodedErrorPage = urlEncodedErrorPage
+            .replace("<ul>", "<ul role=\"presentation\">")
+        return urlEncodedErrorPage
```
pocmo, line 109: Hmmmm, those `<ul>` tags are comming from the translations?

mcarare: Yes. They are in all string.xml translation, they are used to show the items as a "list" with bullets.

pocmo: I'm not fully sure if we should land: Of course this works, but it also feels hacky! I wonder what @Delphine or @flodolo think here? Maybe it is finally time to break this HTML snippet up into multiple strings?

flodolo: Leaving the question (and product) back in Delphine's capable hands, since she's back ;-)

Delphine: So, I'm not entirely sure I understand the whole question here :)
What does this correspond to in the actual strings.xml file? Are we talking about strings that have already been localized?

Also, please note that Betty let me know that we should be string frozen for both a-c and Fenix since June 12th, so I believe anything new needs to wait post Fenix launch.

pocmo: Oh, sorry, here is the context:

This is about the strings on the error pages:
https://github.com/mozilla-mobile/android-components/blob/master/components/browser/errorpages/src/main/res/values/strings.xml#L69

Some of them contain HTML and we would like to change that (`<ul>` to `<ul role="presentation">`). This patch tries to do this in code by modifying the translated strings at runtime. That works, but is a bit hacky.

Creating new strings is probably not right either and that problem may come up again if we change the HTML again. So I wonder if there are other options, for example breaking those large strings with HTML up into multiple strings without HTML that we can then layout as we like...? 🤔

Delphine: As long as we're not losing all the translations on our side, I don't really have much of a preference on how to do this tbh 😶

PR #7510, opened 2020-06-25
Add firefox profiler marker API to the GeckoEngine

canova:
Hi all, I'm on the Firefox Profiler team and I was working on the geckoview API for the Firefox and just landed the GeckoView part in [Bug 1624993](https://bugzilla.mozilla.org/show_bug.cgi?id=1624993). I also want to add this API to android-components so we can use that API in the consumers like Fenix. This is still not exactly complete but I wanted to get some early feedback from you. It would be great if you can take a look at it.

- I added a `ProfilerRuntime` interface and added that interface to `GeckoEngine`. These ProfilerRuntime methods calls the geckoview APIs. I'm not exactly sure about the place. I added it to GeckoEngine class but I can move it to somewhere else too if you think it's not great.
- Also currently I didn't add any tests and I would like to get your feedback on how to write this as well. This feels like a bit tricky, because when the profiler is not running, these methods are no-op. And currently there is no way to enable profiler from geckoview. It's only possible with[ remote profiling via `about:debugging`](https://profiler.firefox.com/docs/#/./guide-remote-profiling). So I'm not sure what to test here. Maybe if we have some devtools tests in the codebase, I can try to do some things like these.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
@@ -10,6 +10,7 @@ import androidx.annotation.VisibleForTesting
 import mozilla.components.browser.engine.gecko.integration.LocaleSettingUpdater
 import mozilla.components.browser.engine.gecko.mediaquery.from
 import mozilla.components.browser.engine.gecko.mediaquery.toGeckoValue
+import mozilla.components.browser.engine.gecko.profiler.Profiler
 import mozilla.components.browser.engine.gecko.webextension.GeckoWebExtension
 import mozilla.components.browser.engine.gecko.webnotifications.GeckoWebNotificationDelegate
 import mozilla.components.browser.engine.gecko.webpush.GeckoWebPushDelegate
@@ -495,6 +496,11 @@ class GeckoEngine(
             GeckoResult<Void>()
         })
     }
+
+    /**
+     * See [Engine.profiler].
+     */
+    override val profiler: Profiler? = Profiler(runtime)
 
     override fun name(): String = "Gecko"
 
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
@@ -27,6 +27,7 @@ import mozilla.components.concept.engine.content.blocking.TrackerLog
 import mozilla.components.concept.engine.content.blocking.TrackingProtectionExceptionStorage
 import mozilla.components.concept.engine.history.HistoryTrackingDelegate
 import mozilla.components.concept.engine.mediaquery.PreferredColorScheme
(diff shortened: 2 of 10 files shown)


canova:
The new profiler marker API is landed to geckoview nightly as well. Closed and opened the PR to re-run the tasks. This PR is ready for a review now. I asked the test question in the android-components Matrix channel but couldn't find a solution there either. I've tried this API manually with Fenix and it works, but still unsure about the possible unit tests. Maybe we can land this without one for now?

components/concept/engine/src/main/java/mozilla/components/concept/engine/profiler/ProfilerRuntime.kt
```diff
@@ -0,0 +1,151 @@
...
+    /**
+     * Returns true if profiler is active and it's allowed the add markers.
+     * It's useful when it's computationally heavy to get startTime or the
+     * additional text for the marker. That code can be wrapped with
+     * isProfilerActive if check to reduce the overhead of it.
+     *
+     * @return true if profiler is active and safe to add a new marker.
+     */
+    fun isProfilerActive(): Boolean {
+        throw UnsupportedOperationException("Firefox Profiler support is not available in this engine")
```
canova, line 67: Currently I'm throwing some exceptions here. But maybe leaving these functions as no-op is a better option. Since these will be no-op if profiler is not running anyway. This way we can start adding profiler markers to the codebase without worrying about exceptions? What do you think?

pocmo: Do we need a default implementation at all? I'd expect `Engine` implementations that do not support `ProfilerRuntime` to just not declare it?

Is your intention here to make all Gecko variants of `Engine` implement them so that you can start using them in Fenix regardless what Gecko version is used? In that case I think I'd still recommend not having a default implementation here and no-op in the specific `GeckoEngine` that doesn't support it.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
```diff
@@ -496,6 +497,55 @@ class GeckoEngine(
...
+        runtime.profilerController.addMarker(markerName, text)
+    }
+
+    /**
+     * See [Engine.addProfilerMarker].
+     */
+    override fun addProfilerMarker(markerName: String) {
+        runtime.profilerController.addMarker(markerName)
+    }
+
```
pocmo, line 548: Nit: I wonder if we should layer this too, like the `GeckoRuntime`, so that the interface of `Engine` does not get too crowded (and it also avoids adding "profiler" to all methods).

Maybe something like a `profiler` property on `Engine`? If it is nullable then you already can use it safely and you do not need to care about no-op/throwing:

```Kotlin
interface Engine {
   val profiler: Profiler?
}


engine.profiler?.addMarker(..)
```

canova: Thanks for the recommendation @pocmo, that definitely makes the code better! Pushed the latest code with this change applied. Could you take a look at it again please?

pocmo:
👍 

bors r+

Issue #5842, opened 2020-02-07
Search engine migration: IndexOutOfBoundsException: Index: 0, Size: 0


pocmo:
https://sentry.prod.mozaws.net/operations/fenix-fennec-nightly/issues/7227032/

```
IndexOutOfBoundsException: Index: 0, Size: 0
    at java.util.ArrayList.get(ArrayList.java:437)
    at mozilla.components.browser.search.SearchEngineManager.getProvidedDefaultSearchEngineAsync(SearchEngineManager.kt:9)
    at mozilla.components.browser.search.SearchEngineManager.getDefaultSearchEngineAsync(SearchEngineManager.kt:7)
    at mozilla.components.browser.search.SearchEngineManager$getDefaultSearchEngineAsync$1.invokeSuspend
    at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:2)
...
(3 additional frame(s) were not displayed)

Index: 0, Size: 0
```

pocmo:
This is not migration related and happens in Fenix Nightly too. Still not often though. Maybe there is a very specific configuration that ends up with no search engines?

https://sentry.prod.mozaws.net/operations/fenix-nightly/issues/6722820

pocmo:
Sentry issue: [FENIX-FENNEC-NIGHTLY-AG](https://sentry.prod.mozaws.net/operations/fenix-fennec-nightly/issues/7288989/)

st3fan:
This is probably a dupe, but this is now a high frequency startup crash on 79b4 and 79b4 - so I am marking it as a blocker.

kbrosnan:
This looks like it was fixed by https://github.com/mozilla-mobile/fenix/issues/11875 as long as we have that on the branch we should be good.

codrut-topliceanu:
https://sentry.prod.mozaws.net/operations/fenix-fennec-beta/issues/8615027
@kbrosnan Seems issue is still present in org.mozilla.firefox_beta-79.0.0-beta.4 should I open another issue for it?

pocmo:
This is a somewhat older build too. 81 is currently on the beta track.

PR #7531, opened 2020-06-26
For #5842 Index Out of Bounds Exception

kglazko:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/search/src/main/java/mozilla/components/browser/search/SearchEngineManager.kt
+++ b/components/browser/search/src/main/java/mozilla/components/browser/search/SearchEngineManager.kt
@@ -130,8 +130,8 @@ class SearchEngineManager(
         val providedDefault = getProvidedDefaultSearchEngineAsync(context)
 
         return when (name) {
-            EMPTY -> defaultSearchEngine ?: providedDefault
-            else -> searchEngineList.list.find { it.name == name } ?: providedDefault
+            EMPTY -> defaultSearchEngine ?: providedDefault!!
+            else -> searchEngineList.list.find { it.name == name } ?: providedDefault!!
         }
     }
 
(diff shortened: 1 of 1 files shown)


kglazko:
I'm not sure if this is the right solution or if there is a better way to prevent this issue. I'm not sure if the right solution here would be to return null instead of crashing? Or what it would even mean to have null for search providers? Would there just be no search provider then?

components/browser/search/src/main/java/mozilla/components/browser/search/SearchEngineManager.kt
```diff
@@ -149,9 +149,9 @@ class SearchEngineManager(
      * Returns the provided default search engine or the first search engine if the default
      * is not set.
      */
-    suspend fun getProvidedDefaultSearchEngineAsync(context: Context): SearchEngine {
+    suspend fun getProvidedDefaultSearchEngineAsync(context: Context): SearchEngine? {
         val searchEngineList = getSearchEngineListAsync(context)
-        return searchEngineList.default ?: searchEngineList.list[0]
+        return searchEngineList.default ?: searchEngineList.list.getOrNull(0)
```
pocmo, line 154: Instead of returning `null` here, I think we should try to figure out what caused this since we should always have search engines, there's a default config in `list.json` so that even with no locale/region we should have a list of search engines. If you follow the stacktrace can you find any reason for us not returning a SearchEngine? 🤔

st3fan:
@kglazko This may be related to the bug that @Mugurell is looking at https://github.com/mozilla-mobile/fenix/issues/11906

I am also not in favor of doing a workaround until we understand the root cause of this IndexOutOfBounds - which may be related to the above issue.

kglazko:
Yeah @st3fan I should've updated the bug after talking to Arturo, I learned that it is best to not do it this way and instead I can change the PR to throw an exception in the case that there are no search engines. An interesting question for the future though is whether or not we will allow a user who wants no search engines (seems like a really niche thing), if we decide so in the future, then the way of handling this may need to be looked at again but for now just throwing an exception seems like the way to go.

st3fan:
@Mugurell has been working on patches. I think we can close this one.

kglazko:
Perfect, closing it now!

PR #7534, opened 2020-06-26
Closes #7533 - Allow loading a thumbnail at maximum size.

### LINKED ISSUE #7533: Allow requesting a thumbnail of maximum size.
#7448 added the ability to specify the thumbnail size to load, but the only available value is 102dp. My usecase is for mozilla-mobile/fenix#3481, I would like to load the full size thumbnail to use as a sort of tab preview.

person808:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/thumbnails/src/main/res/values/dimens.xml
+++ b/components/browser/thumbnails/src/main/res/values/dimens.xml
@@ -2,6 +2,7 @@
 <!-- This Source Code Form is subject to the terms of the Mozilla Public
    - License, v. 2.0. If a copy of the MPL was not distributed with this
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
-<resources xmlns:tools="http://schemas.android.com/tools">
-    <dimen name="mozac_browser_thumbnails_maximum_size" tools:ignore="PxUsage">2153px</dimen>
+<resources>
+    <!-- Maximum size to save thumbnails at. We want full size thumbnails, so we use a large value -->
+    <dimen name="mozac_browser_thumbnails_maximum_size">99999dp</dimen>
 </resources>
--- a/components/support/images/src/main/res/values/dimens.xml
+++ b/components/support/images/src/main/res/values/dimens.xml
@@ -1,7 +1,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!-- This Source Code Form is subject to the terms of the Mozilla Public
-   - License, v. 2.0. If a copy of the MPL was not distributed with this
-   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
-<resources>
-    <dimen name="mozac_support_images_size_default">102dp</dimen>
-</resources>
--- a/components/browser/tabstray/src/main/java/mozilla/components/browser/tabstray/TabViewHolder.kt
+++ b/components/browser/tabstray/src/main/java/mozilla/components/browser/tabstray/TabViewHolder.kt
@@ -8,14 +8,17 @@ import android.content.res.ColorStateList
 import android.view.View
 import android.widget.ImageView
 import android.widget.TextView
+import androidx.annotation.Dimension
+import androidx.annotation.Dimension.DP
 import androidx.appcompat.widget.AppCompatImageButton
 import androidx.recyclerview.widget.RecyclerView
 import mozilla.components.browser.tabstray.thumbnail.TabThumbnailView
 import mozilla.components.concept.tabstray.Tab
 import mozilla.components.concept.tabstray.TabsTray
 import mozilla.components.support.base.observer.Observable
-import mozilla.components.support.images.ImageRequest
+import mozilla.components.support.images.ImageLoadRequest
 import mozilla.components.support.images.loader.ImageLoader
+import mozilla.components.support.ktx.android.util.dpToPx
 import mozilla.components.support.ktx.kotlin.tryGetHostFromUrl
 
 /**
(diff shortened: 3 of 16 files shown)


components/support/images/src/main/res/values/dimens.xml
```diff
@@ -2,6 +2,7 @@
 <!-- This Source Code Form is subject to the terms of the Mozilla Public
    - License, v. 2.0. If a copy of the MPL was not distributed with this
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
-<resources>
+<resources xmlns:tools="http://schemas.android.com/tools">
     <dimen name="mozac_support_images_size_default">102dp</dimen>
+    <dimen name="mozac_support_images_size_max" tools:ignore="PxUsage">2153px</dimen>
```
Amejia481, line 7: Is there any specific reason why the unit are not the same as `mozac_support_images_size_default`? :)

person808: I copied it from [here](https://github.com/mozilla-mobile/android-components/blob/master/components/browser/thumbnails/src/main/res/values/dimens.xml). Seems to be the hard coded maximum px size for thumbnails.

gabrielluong:
So, the current 2153px was just an arbitrary max that I added while testing capturePixels to see if the width/height returns would vary and what was the range of values.

I spoke with Agi about what is the maximum size (width x height) limitation for the bitmap returned by `capturePixels` and it generally allows you to request any size and catches an OOM exception otherwise. 

This is what the test does https://searchfox.org/mozilla-central/source/mobile/android/geckoview/src/androidTest/java/org/mozilla/geckoview/test/ScreenshotTest.kt#332-343.

I think we probably don't need this `DEFAULT` vs `MAX` sizing. Most of the existing `Sizes` are simply because of the `maxSize` parameter needed for `DesiredSize` in the `AndroidImageDecoder`. The current intent with the `2153px` value as the `DEFAULT` is essentially to get the full size of the bitmap when decoding. 

My recommendation is simply to set the `DEFAULT` to `99999` to align with GV. cc @jonalmeida to see if there's other paths forward.

person808:
@gabrielluong I bumped the max size to `99999px`. I left the default size because loading thumbnails at max size in the Fenix tab tray would make them disappear after opening around 6-8 tabs on my Pixel 2 XL. Haven't been able to track down why.

person808:
Ah, it seems like `isGoodSize` is failing because the image is always going to be smaller than a `targetSize` of `99999`. Perhaps its better to make the `size` property of `ImageRequest` an `Int` and make the default value `102dp` (converted to px) so anyone can attempt to load images at an arbitrary size.

gabrielluong:
Also, pinging @jonalmeida for review since he mentioned he had opinions about this as well.

components/support/images/src/main/res/values/dimens.xml
```diff
@@ -2,6 +2,7 @@
 <!-- This Source Code Form is subject to the terms of the Mozilla Public
    - License, v. 2.0. If a copy of the MPL was not distributed with this
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
-<resources>
+<resources xmlns:tools="http://schemas.android.com/tools">
     <dimen name="mozac_support_images_size_default">102dp</dimen>
+    <dimen name="mozac_support_images_size_max" tools:ignore="PxUsage">99999px</dimen>
```
jonalmeida, line 7: Interchanging px values with dp seems wrong. Following along the code, these values are used in the [`AndroidImageDecoder`][0] for the [`inSampleSize`][1] Option used with the `BitmapFactory.decodeByteArray` call, so that seems to expect a px value but for the default cause we can't provide that since it depends on the device.

`getDimensionPixelSize` gives us the px value from dp but looking at the docs, it's unclear whether it can handle being provided a px value accurately. Why not just provide a dp value that equates to a max px like 99999dp?

[0]: https://github.com/mozilla-mobile/android-components/blob/cdef87e847f4272fcb43e441794013e37feb97b9/components/support/images/src/main/java/mozilla/components/support/images/decoder/AndroidImageDecoder.kt#L85
[1]: https://developer.android.com/reference/android/graphics/BitmapFactory.Options#inSampleSize

components/browser/thumbnails/src/main/res/values/dimens.xml
```diff
@@ -3,5 +3,5 @@
    - License, v. 2.0. If a copy of the MPL was not distributed with this
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
 <resources xmlns:tools="http://schemas.android.com/tools">
-    <dimen name="mozac_browser_thumbnails_maximum_size" tools:ignore="PxUsage">2153px</dimen>
+    <dimen name="mozac_browser_thumbnails_maximum_size" tools:ignore="PxUsage">99999px</dimen>
```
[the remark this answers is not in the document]
person808: Don't we need to provide some value [here](https://github.com/mozilla-mobile/android-components/blob/74c1e918dc1dce575a4aae95ef7b264c7b4f7d3f/components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/storage/ThumbnailStorage.kt#L85) though? Looks like the only place where this values ends up being used is in [`isGoodSize`](https://github.com/person808/android-components/blob/eab2bd86b3729e88a962d3d8f35a5770dce7e1c4/components/support/images/src/main/java/mozilla/components/support/images/decoder/AndroidImageDecoder.kt#L53)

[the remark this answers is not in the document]
jonalmeida: Isn't this a duplicate of the values below? 🤔

person808:
@jonalmeida updated with the changes we discussed on slack

person808:
bors r=jonalmeida

PR #7541, opened 2020-06-27
Closes #7278: Add task to generate versions for built-in extensions

### LINKED ISSUE #7278: Automatically generate version in manifest for builtIn extensions
This will be needed to safely land https://github.com/mozilla-mobile/android-components/issues/6356. 

Right now we're using `registerWebExtension` for built-in extensions, which is transient. So our built-in extensions are registered on every startup.

The new `installBuiltInExtension` persists the extension and we don't want to re-install on every startup, only when it changed. See discussion in https://github.com/mozilla-mobile/android-components/issues/6356.

However, if we change a built-in extension and forget to increment the version, the changes would never be applied for users who already have the extension installed. Ideally, we could detect changes and write a version in the manifest.json.

csadilek:
First working version and proposal how to handle versioning once we switch to GV's`ensureBuiltIn` for installing our extension modules. `ensureBuiltIn` only installs/updates the extensions if the manifest has a newer version, so we don't pay the cost of installing on every startup.

This makes sure that
- during development extensions are always installed and we don't have to update the manifest manually on every change
- releases always get a distinct version so we don't ship extension changes that never get applied (because we forgot to update the version)

It's the simplest solution using Gradle's expand / templating functionality from the copy task. We can probably make this more generic with a re-usable task? Not sure if worth it. This is only done for browser-icons here to show the plan:
- Turn manifest.json files into templates 
- Generate the real manifest.json before we build
- gitignore it so we don't have a diff or check it in by mistake

@pocmo @grigoryk wdyt?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -107,6 +107,23 @@ subprojects {
         apply from: "${rootProject.projectDir}/${appServicesSrcDir}/build-scripts/substitute-local-appservices.gradle"
     }
 
+    // Define a reusable task for updating the version in manifest.json for modules that package
+    // a web extension. We automate this to make sure we never forget to update the version, either
+    // in local development or for releases. In both cases, we want to make sure the latest version
+    // of all extensions (including their latest changes) are installed on first start-up.
+    ext.updateExtensionVersion = { task, extDir ->
+        configure(task) {
+            from extDir
+            include 'manifest.template.json'
+            rename { 'manifest.json' }
+            into extDir
+
+            def values = ['version': rootProject.ext.config.componentsVersion + "." + new Date().format('MMddHHmmss')]
+            inputs.properties(values)
+            expand(values)
+        }
+    }
+
     afterEvaluate {
         if (it.hasProperty('android')) {
             jacoco {
@@ -179,6 +196,10 @@ subprojects {
 
                 packagingOptions {
                     exclude 'META-INF/atomicfu.kotlin_module'
+                }
+
+                aaptOptions {
+                    ignoreAssetsPattern "manifest.template.json"
                 }
 
                 compileOptions {
--- a/components/browser/icons/.gitignore
+++ b/components/browser/icons/.gitignore
@@ -1,2 +1,1 @@
-/build
 manifest.json
--- a/components/browser/icons/build.gradle
+++ b/components/browser/icons/build.gradle
@@ -39,6 +39,10 @@ tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
     ]
 }
 
+tasks.register("updateBuiltInExtensionVersion", Copy) { task ->
+    updateExtensionVersion(task, 'src/main/assets/extensions/browser-icons')
+}
+
 dependencies {
     implementation project(':concept-engine')
     implementation project(':concept-fetch')
@@ -71,3 +75,5 @@ dependencies {
 
 apply from: '../../../publish.gradle'
 ext.configurePublish(config.componentsGroupId, archivesBaseName, project.ext.description)
+
+preBuild.dependsOn updateBuiltInExtensionVersion
--- a/components/browser/icons/src/main/assets/extensions/browser-icons/manifest.template.json
+++ b/components/browser/icons/src/main/assets/extensions/browser-icons/manifest.template.json
@@ -6,7 +6,7 @@
     }
   },
   "name": "Mozilla Android Components - Browser Icons",
-  "version": "1.0",
+  "version": "${version}",
   "content_scripts": [
     {
       "matches": ["*://*/*"],
--- a/components/feature/accounts/.gitignore
+++ b/components/feature/accounts/.gitignore
@@ -1,2 +1,1 @@
-/build
 manifest.json
--- a/components/feature/accounts/build.gradle
+++ b/components/feature/accounts/build.gradle
@@ -27,6 +27,10 @@ tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
     ]
 }
 
+tasks.register("updateBuiltInExtensionVersion", Copy) { task ->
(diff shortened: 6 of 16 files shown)


components/browser/icons/.gitignore
```diff
@@ -0,0 +1 @@
+manifest.json
```
pocmo, line 1: You mentioned there may be add-ons like `webcompat` that may not need the templating, do we want to make them use it inside AC?

components/browser/icons/build.gradle
```diff
@@ -25,6 +25,10 @@ android {
         exclude 'META-INF/proguard/androidx-annotations.pro'
     }
 
+    aaptOptions {
+        ignoreAssetsPattern "manifest.json.template"
```
pocmo, line 29: Does it make sense to use something like `manifest.template.json` so that the editor can guess from the file extension that it is JSON?

components/browser/icons/build.gradle
```diff
@@ -39,6 +43,16 @@ tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
...
 }
 
+tasks.register("updateExtensionVersion", Copy) {
+    from 'src/main/assets/extensions/browser-icons'
+    include 'manifest.json.template'
+    rename { 'manifest.json' }
+    into 'src/main/assets/extensions/browser-icons'
+    expand([
+        version: rootProject.ext.config.componentsVersion + "." + new Date().format('MMddHHmmss')
+    ])
```
pocmo, line 53: Nice! That is pretty cool!

components/browser/icons/build.gradle
```diff
@@ -39,6 +43,16 @@ tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
     ]
 }
 
+tasks.register("updateExtensionVersion", Copy) {
+    from 'src/main/assets/extensions/browser-icons'
+    include 'manifest.json.template'
+    rename { 'manifest.json' }
+    into 'src/main/assets/extensions/browser-icons'
```
pocmo, line 50: I wonder if there's a target where we can put it and gradle picks it up - without it being in the src directory? Where do other code generators put there things? But not super important. :)

components/browser/icons/build.gradle
grigoryk: We didn't want this file to be accidentally used for anything by the extension code, since we currently won't be stripping this file from the apk; I'm not sure if that's an actual problem or just a healthy dose of paranoia, but it does also make sense to make the IDE experience a bit nicer.

pocmo: I thought this aapt config is for stripping it from the apk?

csadilek: > Does it make sense to use something like manifest.template.json

Yeah, good point, since it is valid JSON even with the template/placeholder, let's do this. Updated.

> I thought this aapt config is for stripping it from the apk?

Yes, stripping it from the APK, to be paranoid ... making sure it doesn't get used by mistake somehow and we have an extension running with version "${version}". :).

components/browser/icons/.gitignore
csadilek: @pocmo Yes, so what I thought will do is apply this change to every project we want to. Here it's just for browser-icons, but we'd copy this to all projects, as needed. webcompat could decide to keep everything as it is.

I will talk to them to see if they want to keep versioning themselves. I will also see if I can make the gradle task reusable in simple way so we don't have to duplicate it (it's small, but still.)

components/browser/icons/build.gradle
csadilek: I tried. And failed :). Couldn't figure out how to do this, but then this approach is also less magical. At least it's clear looking at the source that it's a template that will be replaced at build-time. Happy to revisit though.

components/browser/icons/.gitignore
pocmo: Makes sense. I also just noticed that this is the module `.gitignore` and not the root one. 👍

components/browser/icons/build.gradle
pocmo: Alright, then let's go ahead as-is :)

components/browser/icons/.gitignore
csadilek: @pocmo OK, I made the task re-usable now and applied it to all projects. The webcompat folks would like to continue using their own versioning as they already have scripts to export, plus they'd like to keep their versions in sync, which makes sense.

I have also added it to sample browser and the two test extensions to demo how this will look like if apps package their own extensions (e.g. Fenix and the two search telemetry extensions).

csadilek:
This is ready for review but not to land (doesn't need to) until the `installBuiltIn` PR lands.

build.gradle
```diff
@@ -107,6 +107,19 @@ subprojects {
         apply from: "${rootProject.projectDir}/${appServicesSrcDir}/build-scripts/substitute-local-appservices.gradle"
     }
 
+    // Define a re-usable task for updating the version in manifest.json (for modules that package a web extension)
+    ext.updateExtensionVersion = { task, extDir ->
```
grigoryk, line 111: Ah, nice.

build.gradle
```diff
@@ -107,6 +107,19 @@ subprojects {
         apply from: "${rootProject.projectDir}/${appServicesSrcDir}/build-scripts/substitute-local-appservices.gradle"
     }
 
+    // Define a re-usable task for updating the version in manifest.json (for modules that package a web extension)
```
grigoryk, line 110: It would be nice to describe why we actually need to do this. You don't mention the reason for this when registering the task, either.

csadilek: Good point. Added docs why we're doing this.

csadilek:
bors r=pocmo,grigoryk

csadilek:
bors retry

PR #7559, opened 2020-06-29
Get addon createdAt/updatedAt as Date

NotWoods:
Right now Fenix has its own date parser for this, but we should provide a standard parser in A-C. Uses the ROOT locale because the date format should always be the same for addons.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/Extensions.kt
+++ b/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/Extensions.kt
@@ -12,9 +12,19 @@ import mozilla.components.feature.addons.update.AddonUpdater
 import mozilla.components.feature.addons.update.AddonUpdater.Status.Error
 import mozilla.components.feature.addons.update.AddonUpdater.Status.NoUpdateAvailable
 import mozilla.components.feature.addons.update.AddonUpdater.Status.SuccessfullyUpdated
-import java.text.NumberFormat
 import java.text.DateFormat
+import java.text.NumberFormat
+import java.text.SimpleDateFormat
+import java.util.Date
 import java.util.Locale
+import java.util.TimeZone
+
+/**
+ * Used to parse [Addon.createdAt] and [Addon.updatedAt].
+ */
+private val dateParser = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'", Locale.ROOT).apply {
+    timeZone = TimeZone.getTimeZone("GMT")
+}
 
 /**
  * A shortcut to get the localized name of an add-on.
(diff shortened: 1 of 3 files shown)


components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/Extensions.kt
```diff
@@ -31,6 +41,17 @@ val Addon.translatedSummary: String get() = translatableSummary.translate(this)
  */
 val Addon.translatedDescription: String get() = translatableDescription.translate(this)
 
+/**
+ * The date the add-on was created, as a JVM date object.
+ */
+val Addon.createdAtDate: Date get() = dateParser.parse(createdAt)!!
```
psymoon, line 47: Can we throw when null instead of `!!` or return a nullable type?

NotWoods: The `parse` function doesn't actually return null. It always throws if the date is invalid.

psymoon: Hmm correct me if I'm missing something here but the API doc suggests nullability:
```
Returns:
    A Date parsed from the string. In case of error, returns null.
Throws:
    NullPointerException - if text or pos is null.
```
https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html#parse(java.lang.String,%20java.text.ParsePosition)

If this is the case, we should not crash with NullPointerException - we should log it with appropriate parsing error. What do you think?

psymoon: 🤔 correct me if I'm missing something here but the API doc suggests nullability:
```
Returns:
    A Date parsed from the string. In case of error, returns null.
Throws:
    NullPointerException - if text or pos is null.
```

https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html#parse(java.lang.String,%20java.text.ParsePosition)

If this is the case, we should not crash with NullPointerException - we should log it with an appropriate parsing error. What do you think?

NotWoods: The version with just one argument seems to handle throwing the parse exception:
```java
    public Date parse(String source) throws ParseException
    {
        ParsePosition pos = new ParsePosition(0);
        Date result = parse(source, pos);
        if (pos.index == 0)
            throw new ParseException("Unparseable date: \"" + source + "\"" ,
                pos.errorIndex);
        return result;
    }
```

I can add some additional code to throw a parse exception, but it won't be testable as my attempts to make this function return null always get this error checking code called instead.

psymoon: `Date result = parse(source, pos);` is a nullable instance:
```
     * If parsing succeeds, then the index of <code>pos</code> is updated
     * to the index after the last character used (parsing does not necessarily
     * use all characters up to the end of the string), and the parsed
     * date is returned. The updated <code>pos</code> can be used to
     * indicate the starting point for the next call to this method.
     * If an error occurs, then the index of <code>pos</code> is not
     * changed, the error index of <code>pos</code> is set to the index of
     * the character where the error occurred, and null is returned.

     * @return A <code>Date</code> parsed from the string. In case of
     *         error, returns null.
```

psymoon: It does seem like a strange design on the Java end. Can you maybe add a comment saying this will never return null?

NotWoods:
bors r=psymoon

PR #7560, opened 2020-06-29
Update to App Services 61.0.7

vladikoff:
Fixes https://github.com/mozilla-mobile/fenix/issues/11294

# v61.0.7 (_2020-06-29_)

[Full Changelog](https://github.com/mozilla/application-services/compare/v61.0.6...v61.0.7)

## Synced Tabs

- JSON is used instead of protocol buffers for sending data over the FFI in `remote_tabs_update_local`. ([#3285](https://github.com/mozilla/application-services/pull/3285))

## General

- The default branch has been renamed from `master` to `main`.

### DIFF
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -27,7 +27,7 @@ object Versions {
     const val disklrucache = "2.0.2"
     const val leakcanary = "1.6.3"
 
-    const val mozilla_appservices = "61.0.6"
+    const val mozilla_appservices = "61.0.7"
 
     const val mozilla_glean = "31.2.3"
 


liuche:
AC 48 is getting cut tomorrow morning (european time) and that's what RC for Fenix Release is aiming to ship with. If this is the AS version that we want to ship with, let's make sure that this gets into AC 48 EOD today cc @pocmo

pocmo:
I'm cutting the AC 48 release now. This will not be in it for now.

CC @grigoryk

st3fan:
I understand this release fixes the https://sentry.prod.mozaws.net/operations/fenix-fennec-nightly/issues/8617877 crash

st3fan:
I put the 79ReleaseBlocker label on this issue but I am getting mixed signals whether it really is. I'd like to keep it in the blockers list until we know for sure.

grigoryk:
> Whether that crash is important enough to delay the release is not clear.

I think it can wait, Sentry isn't looking that bad. 

> It looks like this update changes the underlying protocol from JSON to ProtoBuffers .. https://github.com/mozilla/application-services/pull/3285/files .. that sounds like a rather big architectural change. I don't feel super comfortable making a change like that rushed. I think we need to see this work on a channe first before we ship it?

We already use JSON for a few things (like sync), and it was used widely before switching most of the API surface to protobuffs. So in that sense, the risk is mostly in the unexpected side-effects from this change, not the change itself.

This PR also bumped version of one of the dependencies, but that's a [trivial change](https://github.com/marshallpierce/rust-base64/compare/v0.12.2...v0.12.3).

So - this is fixing a crash with a decent volume, and I'd really like to see this get some bake-time on nightly first. I'm not sure what our process looks like for getting patches merged into the release builds mid-release; can we let this bake first, but still get it into a release sometime before we're at 100%? @st3fan

st3fan:
Bumping this questions to @liuche 

> So - this is fixing a crash with a decent volume, and I'd really like to see this get some bake-time on nightly first. I'm not sure what our process looks like for getting patches merged into the release builds mid-release; can we let this bake first, but still get it into a release sometime before we're at 100%?

What do you think? Bake this on Beta?

st3fan:
@grigoryk can you include this so that we can bake it on beta?

grigoryk:
@st3fan okay, will land this today.

pocmo:
In a mail Chenxia indicated that she would like to have this patch for Beta, so I am going to test it a bit now and land + uplift to 48.

pocmo:
Looks good!

* ✅ Building Fenix against it
* ✅ Running build on device
* ✅ Storage works (bookmarks, history) - Adding and deleting
* ✅ FxA Sign in works
* ✅ Renaming device works (The first couple of times it did not right after sign-in, later getting back to the screen it worked, that didn't seem related to this AS build afaik - Do we show any errors in renaming failed? Because nothing happened and the UI just switched back to the original name)
* ✅ Syncing works
* ✅ Receiving tabs works (manual sync since I do not have any push keys for local builds)
* ✅ Sending tab works.

grigoryk:
> ✅ Renaming device works (The first couple of times it did not right after sign-in, later getting back to the screen it worked, that didn't seem related to this AS build afaik - Do we show any visual error messages if renaming fails? Because nothing happened and the UI just switched back to the original name)

I think that's https://github.com/mozilla-mobile/fenix/issues/10792

PR #7562, opened 2020-06-29
Closes #6910 - Add custom check run text

### LINKED ISSUE #6910: Output markdown for reports to be displayed in task runs

NotWoods:
Closes #6910


build.gradle
```diff
@@ -369,4 +369,12 @@ tasks.register("listRepositories") {
...
         project.repositories.each { println "Name: " + it.name + "; url: " + it.url }
    }
-}
\ No newline at end of file
+}
+
+tasks.register("writeGithubBuildDetails", GithubDetailsTask) {
+    text = """### Test result files
+
+- [Unit Tests (Debug)](/reports/tests/testDebugUnitTest/index.html)
```
grigoryk, line 377: Won't it be less confusing and simpler to set this to `https://firefoxci.taskcluster-artifacts.net/$taskId/0/public` and just replace $taskId with a real one in your task? The regex you have is clever, but that's almost never a good attribute of code :)

NotWoods: I'm trying to avoid the long URL here. Maybe something like `- [Unit Tests (Debug)]({reportsUrl}/reports/tests/testDebugUnitTest/index.html)`? Would allow for shorter URL with simpler replace semantics.

grigoryk: Yeah, that works too!

buildSrc/src/main/java/GithubDetailsTask.kt
```diff
@@ -0,0 +1,43 @@
...
+    @Input
+    var text: String = ""
+
+    private val detailsFile = File("/builds/worker/github/customCheckRunText.md")
+    private val suffix = "\n\n_(404 if compilation failed)_"
+
+    /**
+     * Finds instances of the string {reportsUrl}.
+     */
+    private val replacementRegex = """(\{reportsUrl})""".toRegex()
```
grigoryk, line 29: don't need this

buildSrc/src/main/java/GithubDetailsTask.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Text to write to the file.
+     * Links are automatically rewritten to point to the correct Taskcluster URL.
+     */
+    @Input
+    var text: String = ""
+
+    private val detailsFile = File("/builds/worker/github/customCheckRunText.md")
+    private val suffix = "\n\n_(404 if compilation failed)_"
```
grigoryk, line 24: or if paths of anything changed, right? i'd remove this.

build.gradle
```diff
@@ -369,4 +369,11 @@ tasks.register("listRepositories") {
...
    }
-}
\ No newline at end of file
+}
+
+tasks.register("githubBuildDetails", GithubDetailsTask) {
+    text = """### Test result files
+
+- [Unit Tests]({reportsUrl}/tests/testReleaseUnitTest/index.html)
+- [Android Lint]({reportsUrl}/lint-results-release.html)"""
```
grigoryk, line 378: No detekt?

buildSrc/src/main/java/GithubDetailsTask.kt
NotWoods: It's unlikely for paths to change without adjusting gradle, but compilation failure can happen on any PR. I'd prefer to keep this for now and [switch to redirecting to taskcluster](https://github.com/mozilla-mobile/fenix/issues/12785) in the future.

build.gradle
```diff
@@ -369,4 +369,15 @@ tasks.register("listRepositories") {
...
+}
+
+tasks.register("githubLintDetektDetails", GithubDetailsTask) {
+    text = "### [Detekt Results](/reports/detekt.html)"
```
grigoryk, line 382: You forgot `{reportsUrl}`. But even then, I don't think `https://firefoxci.taskcluster-artifacts.net/{reportsUrl}/0/public/reports/detekt.html` is a correct url (or at least, it doesn't work for me).

Also, this url renders as a sub-header in markdown, probably not what you want.

NotWoods:
bors r=grigoryk

NotWoods:
bors retry

Issue #7563, opened 2020-06-29
Allow sorting of context GeckoViewActions


sblatz:
For https://github.com/mozilla-mobile/fenix/issues/10466

PR #7564, opened 2020-06-29
For #7563: Allow GeckoActions to be sorted

sblatz:
---

Putting this up as a draft for @pocmo to check and make sure we're on the same page about this implementation.

I know I'll need to put this in all three Gecko's, just doing one implementation for now to make sure this solution is sufficient 😄 

Here's how Fenix would consume this change: https://github.com/mozilla-mobile/fenix/pull/12112

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
@@ -60,7 +60,8 @@ class GeckoEngine(
     private val runtime: GeckoRuntime = GeckoRuntime.getDefault(context),
     executorProvider: () -> GeckoWebExecutor = { GeckoWebExecutor(runtime) },
     override val trackingProtectionExceptionStore: TrackingProtectionExceptionStorage =
-        TrackingProtectionExceptionFileStorage(context, runtime)
+        TrackingProtectionExceptionFileStorage(context, runtime),
+    private val actionSorter: ((Array<String>) -> Array<String>)? = null
 ) : Engine, WebExtensionRuntime {
     private val executor by lazy { executorProvider.invoke() }
     private val localeUpdater = LocaleSettingUpdater(context, runtime)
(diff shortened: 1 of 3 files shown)


components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/selection/GeckoSelectionActionDelegate.kt
```diff
@@ -19,25 +19,30 @@ import org.mozilla.geckoview.BasicSelectionActionDelegate
...
         }
     }
 
     override fun getAllActions(): Array<String> {
-        return super.getAllActions() + customDelegate.getAllActions()
+        return if (actionSorter != null) {
+            actionSorter.invoke(super.getAllActions() + customDelegate.getAllActions())
+        } else {
+            super.getAllActions() + customDelegate.getAllActions()
+        }
```
pocmo, line 45: nit: I think we could ask for a `(String) -> R?` mapping, so that we can pass it directly to `sortedBy` here?

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
```diff
@@ -137,7 +138,7 @@ class GeckoEngine(
      * Creates a new Gecko-based EngineView.
      */
     override fun createView(context: Context, attrs: AttributeSet?): EngineView {
-        return GeckoEngineView(context, attrs)
+        return GeckoEngineView(context, attrs, actionSorter = actionSorter)
```
pocmo, line 141: The approach overall looks good to me. However it is a bit annoying that we will have to pass it to the `Engine` and here down to the `View` as well. Especially since that breaks the view constructor contract and so that doesn't work if you add the view to your XML.
Are there other options how we can get this passed down? Ideally we can set our sorter function on our custom delegate (or return it from there), is that something that could work?

sblatz:
Closing in favor of: https://github.com/mozilla-mobile/android-components/pull/7658/files

PR #7575, opened 2020-06-30
Just run release tests on components

NotWoods:
Currently we run tests on both debug and release builds. Theoretically we can half our CI times by turning off one set of tests :)


NotWoods:
Looks like just testing release doesn't have any issues: https://firefoxci.taskcluster-artifacts.net/A-kdpCPyR4Oy4ElBHr5zpw/0/public/reports/tests/testReleaseUnitTest/classes/mozilla.components.feature.pwa.ext.WebAppManifestKtTest.html#should%20return%20the%20scope%20as%20a%20uri

PR #7578, opened 2020-06-30
Closes #6460: Upstream RunWhenReadyQueue

### LINKED ISSUE #6460: Move RunWhenReadyQueue into a-c
https://github.com/mozilla-mobile/fenix/pull/9239 adds `RunWhenReadyQueue` which could be useful in a few places; let's move it into a-c, and add some tests for it.

NotWoods:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/support/utils/build.gradle
+++ b/components/support/utils/build.gradle
@@ -27,7 +27,7 @@ dependencies {
     implementation project(':support-base')
 
     implementation Dependencies.kotlin_stdlib
-
+    implementation Dependencies.kotlin_coroutines
     implementation Dependencies.androidx_annotation
     implementation Dependencies.androidx_core
 
@@ -38,6 +38,7 @@ dependencies {
 
     testImplementation Dependencies.androidx_test_core
     testImplementation Dependencies.androidx_test_junit
+    testImplementation Dependencies.testing_coroutines
     testImplementation Dependencies.testing_mockito
     testImplementation Dependencies.testing_robolectric
 }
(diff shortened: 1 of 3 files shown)


components/support/base/src/main/java/mozilla/components/support/base/concurrent/RunWhenReadyQueue.kt
```diff
@@ -0,0 +1,62 @@
...
+     * Runs the [task] if this queue is marked as ready, or queues it for later execution.
+     *
+     * @param task: The task to run now if queue is ready or queue for later execution.
+     */
+    fun runIfReadyOrQueue(task: () -> Unit) {
+        if (isReady.get()) {
+            scope.launch { task.invoke() }
+        } else {
+            tasks.add(task)
+        }
```
pocmo, line 42: This part doesn't seem thread-safe is `runIfReadyOrQueue()` and `ready()` are called on different threads? I think we would need to synchronize here? Or alternatively assert that we are on the main thread if that is what we expect.

NotWoods: Would `AtomicBoolean` handle the threads? cc @grigoryk who worked on the class

grigoryk: I think what _could_ happen here is that `tasks` is modified by `runIfReadyOrQueue` while we're iterating through it in `ready`. This won't result in an exception thanks to `CopyOnWriteArrayList`, but we will drop that task on the floor in the `tasks.clear` call right after the iteration, I think.

A `synchronized` block wrapping the iteration/clearing of `tasks`, and wrapping `tasks.add` should be fine, though. With that in place, you can switch to a regular array, as well.

PR #7580, opened 2020-07-01
Closes #7570: Add an auth-recovery circuit breaker

### LINKED ISSUE #7570: Add circuit-breaker to prevent FxA auth recovery loop
In [this reddit thread](https://www.reddit.com/r/firefox/comments/hi79tg/firefox_beta_sync_issues/) the user reports a very strange behaviour from Fenix, saying that it "tries to sync, signs out, re-connects automatically and signs out automatically once every second". There is a similar report [here](https://www.reddit.com/r/firefox/comments/hhp7a4/sync_acting_buggy_in_android_andor_windows/) saying that a "purple notification kept floating up from the bottom of the app (repeatedly/non-stop) saying that sync was on".

From discussion with @grigoryk, we suspect these users are somehow stuck in an FxA "auth problem recovery" retry loop, going roughly like:

1. A request using FxA state in the browser fails with an auth error, pushing the FxaAccountManager into the `AuthenticationProblem` state.
2. The FxaAccountManager tries to get a fresh auth token from the FxA server, succeeds, and transitions back to its happy `AuthenticatedNoProfile` state.
    * As part of this, it fires a `AuthType.Recovered` notification to observers, causing Fenix to show a little notification about the account being connected
3. The thing that failed at (1) tries again, but still fails.
    * We go to (1), and repeat indefinitely.

We have this theory that such a loop *should* never happen, and that whatever triggered the failure at step (1) *should* always recover if we are able to get a fresh OAuth token from FxA. But it appears this is not happening for at least two users in the wild, and I'm worried it could happen for more when we roll this out to the Fenix release population.

Could we please consider adding some sort of circuit-breaker here to prevent an infinite loop of retries? Even better if we're able to log the cause of the infinite loop to sentry, but if that's not possible, just being able to avoid DoSing the FxA servers would be excellent :-)

I'm not entirely sure what the implementation of this would look like. Perhaps it would suffice to keep a last-successful-retry timestamp in the in-memory state of FxaAccountManager, and refuse to loop again if we already retried recently, instead driving to the persistent "need to reconnect" state.

grigoryk:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
+++ b/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
@@ -19,8 +19,6 @@ import kotlinx.coroutines.cancel
 import kotlinx.coroutines.withContext
 import mozilla.appservices.syncmanager.DeviceSettings
 import mozilla.components.concept.sync.AccountObserver
-import mozilla.components.concept.sync.AuthException
-import mozilla.components.concept.sync.AuthExceptionType
 import mozilla.components.concept.sync.DeviceCapability
 import mozilla.components.concept.sync.AuthFlowUrl
 import mozilla.components.concept.sync.AuthType
@@ -30,6 +28,7 @@ import mozilla.components.concept.sync.InFlightMigrationState
 import mozilla.components.concept.sync.OAuthAccount
 import mozilla.components.concept.sync.Profile
 import mozilla.components.concept.sync.StatePersistenceCallback
+import mozilla.components.service.fxa.AccountManagerException
 import mozilla.components.service.fxa.AccountStorage
 import mozilla.components.service.fxa.DeviceConfig
 import mozilla.components.service.fxa.FxaDeviceSettingsCache
@@ -65,27 +64,6 @@ import java.util.concurrent.Executors
 import kotlin.coroutines.CoroutineContext
 
 /**
- * Exposes an instance of [FxaAccountManager] for internal consumption; populated during initialization of
- * [FxaAccountManager]. This exists to allow various internal parts without a direct reference to an instance of
- * [FxaAccountManager] to notify it of encountered auth errors.
- */
-internal object GlobalAccountManager {
-    private var instance: WeakReference<FxaAccountManager>? = null
-
-    internal fun setInstance(am: FxaAccountManager) {
-        instance = WeakReference(am)
-    }
-
-    internal fun close() {
-        instance = null
-    }
-
-    internal suspend fun authError(cause: Exception? = null) {
-        instance?.get()?.encounteredAuthError(cause)
-    }
-}
-
-/**
  * Observer interface which lets its consumers respond to authentication requests.
  */
 private interface OAuthObserver {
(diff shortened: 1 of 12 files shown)


components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
```diff
@@ -80,8 +81,16 @@ internal object GlobalAccountManager {
         instance = null
     }
 
-    internal suspend fun authError(cause: Exception? = null) {
-        instance?.get()?.encounteredAuthError(cause)
+    internal suspend fun authError(operation: String) {
```
rfk, line 84: Is there any chance of two concurrent threads of activity both discovering an auth error at the same time, resulting in two closely-spaced calls to this method without the opportunity to try a recovery?

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
```diff
@@ -865,10 +881,30 @@ open class FxaAccountManager(
...
 
+                        // Ensure we clear any auth-relevant internal state, such as access tokens.
+                        account.authErrorDetected()
+
                         // Clear our access token cache; it'll be re-populated as part of the
                         // regular state machine flow if we manage to recover.
                         SyncAuthInfoCache(context).clear()
 
+                        // Circuit-breaker logic to protect ourselves from getting into endless authorization
+                        // check loops. If we determine that application is trying to check auth status too
```
rfk, line 892: "that application" -> "that the application"

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
```diff
@@ -865,10 +881,30 @@ open class FxaAccountManager(
...
+                        // frequently, drive the state machine into an unauthorized state.
```
rfk, line 893: Driving to "unauthorized" seems reasonable here; it won't necessarily help the user to access their account data, but at least the rest of the browser will be useable.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
```diff
@@ -107,6 +116,8 @@ const val SCOPE_PROFILE = "profile"
 const val SCOPE_SYNC = "https://identity.mozilla.com/apps/oldsync"
 // Necessary to obtain a sessionToken, which gives full access to the account.
 const val SCOPE_SESSION = "https://identity.mozilla.com/tokens/session"
+// Not allowed to check authorization status more frequently than 10 seconds.
+const val AUTH_CHECK_CIRCUIT_BREAKER_WINDOW_MS = 1000L * 10
```
rfk, line 120: As you mentioned in slack, it's hard to know what the right value is here. I worry that 10 seconds may be too short, depending on how far removed the failing action is from the recovery process here. How would you feel about 30 seconds or even 1 minute? Our OAuth tokens should all live *much* longer than that in practice!

The tradeoff I guess, is that we increase the risk of spuriously driving the user to the "unauthenticated" state if they happen to trigger several legitimate auth errors within this time period.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
```diff
@@ -71,6 +71,7 @@ import kotlin.coroutines.CoroutineContext
  */
 internal object GlobalAccountManager {
     private var instance: WeakReference<FxaAccountManager>? = null
+    private var lastAuthErrorCheckPoint: Long = 0L
```
rfk, line 74: *If* we wanted to avoid potential issues with concurrently triggering multiple auth errors, we could make this an array of timestamps for the last N auth errors, and then only allow at most N within the last WINDOW seconds. That's substantial extra complexity though, which may not be worth it.

(I can think of a more thorough approach, but it would require cooperation from the underlying Rust library, and would require callers to pass back the actual token from which they received an auth error rather than just calling a generic callback)

rfk:
> I'm going to see if I can get some numbers to back up an appropriate window size.

I inspected the traffic from some users who I believe to be stuck in this state, and they're making any oauth token requests only a second or two apart. So I think even a small window will make a big difference here. 30 seconds?

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
grigoryk: I've switched to a simple sliding window, that should help with the multiple errors case.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
grigoryk: To close the loop - yes, and we've worked around this by allowing a few auth errors (but not too many) through.

pocmo:
bors r=rfk,pocmo

PR #7591, opened 2020-07-02
For #7592 - Adds Login Exception Storage Feature 

ekager:
---

### Pull Request checklist

### After merge

### DIFF
--- a/.buildconfig.yml
+++ b/.buildconfig.yml
@@ -72,6 +72,10 @@ projects:
     path: components/feature/media
     description: 'Feature component for website media related features.'
     publish: true
+  feature-logins:
+    path: components/feature/logins
+    description: 'Feature component for logins related features.'
+    publish: true
   feature-p2p:
     path: components/feature/p2p
     description: 'Feature used for peer-to-peer communication.'
--- a/components/feature/logins/README.md
+++ b/components/feature/logins/README.md
@@ -1,0 +1,19 @@
+# [Android Components](../../../README.md) > Feature > Logins
+
+Feature component with features related to logins.
+
+## Usage
+
+### Setting up the dependency
+
+Use Gradle to download the library from [maven.mozilla.org](https://maven.mozilla.org/) ([Setup repository](../../../README.md#maven-repository)):
+
+```Groovy
+implementation "org.mozilla.components:feature-logins:{latest-version}"
+```
+
+## License
+
+    This Source Code Form is subject to the terms of the Mozilla Public
+    License, v. 2.0. If a copy of the MPL was not distributed with this
+    file, You can obtain one at http://mozilla.org/MPL/2.0/
--- a/components/feature/logins/build.gradle
+++ b/components/feature/logins/build.gradle
@@ -1,0 +1,72 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+apply plugin: 'com.android.library'
+apply plugin: 'kotlin-android'
+apply plugin: 'kotlin-kapt'
+
+android {
+    compileSdkVersion config.compileSdkVersion
+
+    defaultConfig {
+        minSdkVersion config.minSdkVersion
+        targetSdkVersion config.targetSdkVersion
+        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+
+        kapt {
+            arguments {
+                arg("room.schemaLocation", "$projectDir/schemas".toString())
+            }
+        }
+    }
+
+    buildTypes {
+        release {
+            minifyEnabled false
+            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+        }
+    }
+
+    packagingOptions {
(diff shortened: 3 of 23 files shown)


components/feature/prompts/build.gradle
```diff
@@ -34,6 +34,7 @@ dependencies {
     implementation project(':support-ktx')
     implementation project(':support-utils')
     implementation project(':ui-icons')
+    implementation project(':feature-loginsexceptions')
```
ekager, line 37: Q - is there a way to not make the downstream consumers of prompt-feature also _have_ to add this feature since it's optional?

components/feature/loginsexceptions/schemas/mozilla.components.feature.loginsexceptions.db.LoginExceptionDatabase/1.json
```diff
@@ -0,0 +1,40 @@
+{
```
ekager, line 1: q - am i supposed to include this schema file?

.buildconfig.yml
```diff
@@ -72,6 +72,10 @@ projects:
     path: components/feature/media
     description: 'Feature component for website media related features.'
     publish: true
+  feature-loginsexceptions:
```
pocmo, line 75: Does it make sense to turn this into `feature-logins` so that we could land other things related to logins here too in the future?

components/feature/loginsexceptions/schemas/mozilla.components.feature.loginsexceptions.db.LoginExceptionDatabase/1.json
pocmo: Yes, this is great, those can later be used to create older versions of the database to test migrations 👍

components/feature/loginsexceptions/src/androidTest/java/mozilla/components/feature/loginsexceptions/LoginExceptionStorageTest.kt
```diff
@@ -0,0 +1,140 @@
...
+import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertNull
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.Executors
+
+@Suppress("LargeClass")
+class LoginExceptionStorageTest {
```
pocmo, line 29: Lets make sure this component is listed here so that on-device tests for it run on Firebase:
https://github.com/mozilla-mobile/android-components/blob/master/taskcluster/ci/test/kind.yml#L51

I do not remember whether this is the only thing needed, @Amejia481 recently did this for an other component and may know. :)

components/feature/prompts/build.gradle
pocmo: I can see two options:
* Use `compileOnly`. That works if no class of that is actually somehow needed at runtime if you do not use that feature. But it's rather awkward since if you do it will randomly fail and no tooling will warn you about it.
* For everything that `feature-prompts` needs access to, create an interface in `feature-prompts`. Then in your new component implement those interfaces. That way `feature-prompts` does not need to know about the implementation and you are essentially reversing the dependencies. This will need some setup/glue code so that you can tell `feature-prompts` about which implementation to use though.

components/feature/loginsexceptions/src/androidTest/java/mozilla/components/feature/loginsexceptions/LoginExceptionStorageTest.kt
Amejia481: Yes, that's correct. You only need to add another entry.

Amejia481: @ekager to verify that android tests were added correctly just use `bors try` and you will see a task with the symbol name that you chosen with the prefix `test-` e.g `test-YOUR_SYMBOL_NAME`

ekager:
bors try

components/feature/loginsexceptions/src/androidTest/java/mozilla/components/feature/loginsexceptions/LoginExceptionStorageTest.kt
ekager: 💃 
Screen Shot 2020-07-06 at 3 54 53 PM

components/feature/logins/schemas/mozilla.components.feature.loginsexceptions.db.LoginExceptionDatabase/1.json
```diff
@@ -0,0 +1,40 @@
+{
+  "formatVersion": 1,
```
Amejia481, line 2: It looks like we have to `LoginExceptionDatabase/1.json ` files 🤔

ekager: Oops I moved/renamed the feature so I need to remove this one 😅

Amejia481: 👍

ekager:
bors r=pocmo

PR #7599, opened 2020-07-02
Fix the signature of one of the engine.profiler.addMarker overloads

canova:
We added profiler marker API in #7510. This `addMarker` overload was accepting `Double` only but it should be `Double?` instead to be able to allow nullable values as well. `getProfilerTime` returns nullable double time as well and they should be compatible. Other overloads are correct but I miss this one.

It also moves the changelog item to 49 instead of 48 (which we missed).

---

### Pull Request checklist

### After merge


canova:
@pocmo hey, I missed this signature, and it's good to change it to a nullable like the rest of them. This is not a breaking change though, since nullable also includes the non-null double.
also moved the profiler changelog entry to version 49.

PR #7605, opened 2020-07-02
Revert "Closes #7483: Fix telemetry foreground service notification string"

### LINKED ISSUE #7483: SendCrashTelemetryService should use a different notification text

rocketsroger:
Reverts mozilla-mobile/android-components#7492


rocketsroger:
String freeze.  This change needs to be backed out

rocketsroger:
bors r=Amejia481

PR #7609, opened 2020-07-03
Allow local publication workflows of app-services to work on Windows.

mhammond:
The end result is that Android Studio and native Windows can be used
to build the android-components and its consumers using a local
application-services directory.

The equivalent for Fenix landed in https://github.com/mozilla-mobile/fenix/pull/11508

cc @grigoryk

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -95,7 +95,13 @@ if (localProperties != null) {
 
     if (appServicesLocalPath != null) {
         logger.lifecycle("Enabling automatic publication of application-services from: $appServicesLocalPath")
-        def publishAppServicesCmd = ["./automation/publish_to_maven_local_if_modified.py"]
+        // Windows can't execute .py files directly, so we assume a "manually installed" python,
+        // which comes with a "py" launcher and respects the shebang line to specify the version.
+        def publishAppServicesCmd = [];
+        if (System.properties['os.name'].toLowerCase().contains('windows')) {
+            publishAppServicesCmd << "py";
+        }
+        publishAppServicesCmd << "./automation/publish_to_maven_local_if_modified.py";
         runCmd(publishAppServicesCmd, appServicesLocalPath, "Published application-services for local development.")
     } else {
         logger.lifecycle("Disabled auto-publication of application-services. Enable it by settings '$settingAppServicesPath' in local.properties")


settings.gradle
```diff
@@ -96,7 +96,20 @@ if (localProperties != null) {
...
+        // enable these workflows on Windows, even if it isn't quite as automatic as elsewhere.
+        // So rather than fail we make noise...
+        if (System.properties['os.name'].toLowerCase().contains('windows')) {
+            logger.lifecycle('NOTE: The autoPublish workflows do not work on native windows.');
+            logger.lifecycle('You must follow the instructions at:');
+            logger.lifecycle('> https://github.com/mozilla/application-services/blob/master/docs/howtos/setup-android-build-environment.md#using-windows')
+            logger.lifecycle('then, manually ensure that the following command has completed successfully in WSL:');
+            logger.lifecycle("> $publishAppServicesCmd");
+            logger.lifecycle("(from the '$appServicesLocalPath' directory)");
+            logger.lifecycle('Then restart the build');
```
grigoryk, line 109: So IIUC, if the WSL build env is configured correctly (w/ a correctly configured local maven repo), autopublication _will_ work? Or would we still fail to run `publish_to_maven_local_if_modified`? (well, we wouldn't even be able to run that script, I think, due to `./` bit?)

And if you do run it all manually and publish a-s to the locally configured maven in WSL, it will get picked up correctly later on.

If that's the case, I don't quite understand why we wouldn't both update `publishAppServicesCmd` to work on windows (and run that script) and print a message pointing to WSL build instructions.

grigoryk:
@mhammond happy to land this as-is, but could you please answer my question above? Thanks :)

settings.gradle
mhammond: I'm not sure I understand the question :) Are you suggesting that we:

* Give this code the "if on windows, use py.exe" for Windows [like we did for Fenix](https://github.com/mozilla-mobile/fenix/blob/9ed85510ae75508871cc5094c5ff73c546e53a4c/settings.gradle#L76), but still execute $publishAppServicesCmd

* Have that Python script detect windows and print the same message we are printing here?

Then yeah, that would work fine. Happy to make that change if you think that's better.

mhammond: (I guess the key reason it's better is that the limitation is caused by app-services, so it makes sense for this message and the conditions under which it's true to also live in app-services!)

grigoryk: Yeah, I should have phrased that better - your two bullets sounds better than the current patch. Make the execution of `publishAppServicesCmd` work in a-c on windows, and have that script print out specific instructions under certain conditions (e.g. running in WSL).

mhammond:
Thanks Grisha - the app-services side is in https://github.com/mozilla/application-services/pull/3381/files. I probably should do the same for Fenix now too.

Issue #7598, opened 2020-07-02
Don't build unchanged components in CI


mozbhearsum:
At the moment, we always build all components as part of CI jobs. I did some analysis on the most recent ~100 commits to this repository and found that roughly ~85% of the builds being done were not explicitly changed, or depended on by a component that changed in a commit. For example, https://github.com/mozilla-mobile/android-components/commit/c51cbcb0c88e6af502249966004b4248186ad69c only modifies `feature-webnotifications`. That component directly or indirectly depends on: browser-icons, browser-session, browser-state, concept-engine, concept-fetch feature-sitepermissions, support-base, support-images, support-ktx, support-utils, ui-icons

Getting rid of builds for components unaffected by a commit would both speed up CI and save money. I've spoken with both @JohanLorenzo and @pocmo about this, and the rough idea would is:
* Make it possible to inspect the full dependency chain for the local components. I've got some hacky local code that I used for doing analysis, but there's probably better ways to do it.
* Use the list of changed files in a commit to see what components have changed
* Send the information above to `taskgraph` so it can decide which components need to be built. Eg: build all of the components that have changed files, plus everything they depend on

One thing I'm not sure about is what should be built when commits touch things that aren't explicitly part of a component. For example, https://github.com/mozilla-mobile/android-components/commit/c7494a5045808dcafeedb3a4f905fcff00a381d1 is a version bump. Should we build everything for that? Nothing? Something else?

PR #7617, opened 2020-07-03
Prune unnecessary tasks from ci tasks

mozbhearsum:
This is a proof of concept for #7598. There's a number of things that need be improved & addressed (see below), but with commands like:
`taskgraph decision --pushlog-id=0 --pushdate=0 --project=android-components --message= --owner=skaspari+mozlando@mozilla.com --level=3 --base-repository=https://github.com/mozilla-mobile/android-components --head-repository=https://github.com/mozilla-mobile/android-components --head-ref=refs/heads/master --head-rev=6131efe915b92ad10eec23620a71dc54510bd3e9 --head-tag= --repository-type=git --tasks-for=github-push`

...it produces a significantly smaller set of tasks, like:
```
[u'build-concept-storage',
 u'build-samples-browser-beta',
 u'build-samples-browser-nightly',
 u'build-samples-browser-release',
 u'build-samples-browser-system',
 u'build-service-glean',
 u'build-support-base',
 u'build-support-ktx',
 u'build-support-sync-telemetry',
 u'build-support-utils',
 u'complete-push',
 u'complete-push-1',
 u'complete-push-2',
 u'lint-compare-locales',
 u'lint-detektAndKtlint',
 u'test-android-feature-containers',
 u'test-android-feature-pwa',
 u'test-android-feature-share',
 u'test-android-feature-sitepermissions',
 u'test-android-feature-top-sites',
 u'test-ui-browser',
 u'test-ui-glean',
 u'test-unit-browser-engine-gecko-nightly',
 u'toolchain-linux64-android-gradle-dependencies',
 u'toolchain-linux64-android-sdk-linux-repack']
```

Known TODOs include:
* ~~Use a Github API token to avoid getting rate limited~~ - This probably isn't necessary in the end, because we can do 60 requests/hour/IP. Each decision task will make 2 requests, and most decision tasks won't run on the same workers, so the IPs will differ.
* ~~Figure out how to find all files changed for PRs (something like https://github.com/mozilla-mobile/fenix/blob/master/taskcluster/fenix_taskgraph/parameters.py probably)~~
* ~~Ensure we're using the right API endpoints for commits/pushes~~
* Figure out what actual entries we need in `FILE_PATTERNS_TO_AFFECTED_COMPONENTS`
* See if we can get dependencies dumped from gradle in JSON or another more parseable format
* Determine whether or not the `build-samples-browser` tasks should be considered here, too.

### DIFF
--- a/taskcluster/ac_taskgraph/loader/build_config.py
+++ b/taskcluster/ac_taskgraph/loader/build_config.py
@@ -4,16 +4,105 @@
 
 from __future__ import print_function, unicode_literals
 
+import logging
 import os
+import re
+import subprocess
 
 from copy import deepcopy
 from taskgraph.loader.transform import loader as base_loader
+from taskgraph.util.memoize import memoize
+from taskgraph.util.taskcluster import get_session
 from taskgraph.util.templates import merge
 
+from ..files_changes import get_files_changed_pr, get_files_changed_push
 from ..build_config import get_components
 
 
+logger = logging.getLogger(__name__)
+
+ALL_COMPONENTS = object()
+
+def get_components_changed(files_changed):
+    """Translate a list of files changed into a list of components. Eg:
+        [
+            "components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/GeckoLoginStorageDelegate.kt",
+            "components/service/sync-logins/src/test/java/mozilla/components/service/sync/logins/GeckoLoginStorageDelegateTest.kt",
+        ]
+        ->
+        {service-sync-logins}
+    """
+    return set(["-".join(f.split("/")[1:3]) for f in files_changed if f.startswith("components")])
+
+
+cached_deps = {}
+
+# Gradle is expensive to run. Cache the results for each component.
+@memoize
+def get_deps_for_component(component):
+    """Return the full list of local upstream dependencies of a component."""
+    deps = set()
+    # Parsing output like this is not ideal but bhearsum couldn't find a way
+    # to get the dependencies printed in a better format. If we could convince
+    # gradle to spit out JSON that would be much better.
+    # This is filed as https://github.com/mozilla-mobile/android-components/issues/7814
+    for line in subprocess.check_output(['./gradlew', '%s:dependencies' % component, '--configuration', 'implementation']).splitlines():
+        if line.startswith("+--- project"):
+            deps.add(line.split(" ")[2])
+
+    cached_deps[component] = deps
+    # So far, we only have the direct local upstream dependencies. We need to look at
+    # each of those to find _their_ upstreams as well.
+    for d in deps.copy():
+        deps.update(get_deps_for_component(d))
+
+    return deps
+
+def get_affected_components(files_changed, files_affecting_components):
+    affected_components = set()
+
+    # First, find the list of changed components
+    for c in get_components_changed(files_changed):
+        affected_components.add(c)
+
+    # Then, look for any other affected components based on the files changed.
+    for pattern, components in files_affecting_components.items():
+        if any([re.match(pattern, f) for f in files_changed]):
+            # Some file changes may necessitate rebuilding _all_ components.
+            # In this we can just return immediately.
+            if components == "all-components":
+                return ALL_COMPONENTS
+
+            affected_components.update(components)
+
+    # Finally, go through all of the affected components and recursively
+    # find their dependencies.
+    for c in affected_components.copy():
+        affected_components.update(get_deps_for_component(c))
+
+    return affected_components
+
+
 def loader(kind, path, config, params, loaded_tasks):
+    if params["tasks_for"] == "github-pull-request":
+        logger.info("Processing pull request %s" % params["pull_request_number"])
+        files_changed = get_files_changed_pr(params["base_repository"], params["pull_request_number"])
+        affected_components = get_affected_components(files_changed, config.get("files-affecting-components"))
+    elif params["tasks_for"] == "github-push":
+        logger.info("Processing push for commit range %s -> %s" % (params["base_rev"], params["head_rev"]))
+        files_changed = get_files_changed_push(params["base_repository"], params["base_rev"], params["head_rev"])
+        affected_components = get_affected_components(files_changed, config.get("files-affecting-components"))
+    # Unknown tasks_for, can't safely prune tasks
+    else:
+        files_changed = []
+        affected_components = ALL_COMPONENTS
+
+    logger.info("Files changed: %s" % " ".join(files_changed))
+    if affected_components is ALL_COMPONENTS:
+        logger.info("Affected components: ALL")
+    else:
+        logger.info("Affected components: %s" % " ".join(affected_components))
+
     not_for_components = config.get("not-for-components", [])
     jobs = {
         '{}{}'.format(
@@ -28,10 +117,14 @@ def loader(kind, path, config, params, loaded_tasks):
         for component in get_components()
         for build_type in ('regular', 'nightly', 'release')
         if (
-            component['name'] not in not_for_components
+            (affected_components is ALL_COMPONENTS or component['name'] in affected_components)
+            and component['name'] not in not_for_components
             and (component['shouldPublish'] or build_type == 'regular')
         )
     }
+    # Filter away overridden jobs that we wouldn't build anyways to avoid ending up with
+    # partial job entries.
+    overridden_jobs = {k: v for k, v in config.pop('overriden-jobs', {}).items() if affected_components is ALL_COMPONENTS or k in affected_components}
     jobs = merge(jobs, config.pop('overriden-jobs', {}))
 
     config['jobs'] = jobs
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -68,6 +68,13 @@ tasks:
                           else:
                               $if: 'tasks_for in ["cron", "action"]'
                               then: '${push.branch}'
+              base_sha:
+                  $if: 'tasks_for == "github-push"'
+                  then: '${event.before}'
+                  else:
+                      $if: 'tasks_for == "github-pull-request"'
+                      then: '${event.pull_request.base.sha}'
+                      else: 'UNDEFINED'
               head_sha:
                   $if: 'tasks_for == "github-push"'
                   then: '${event.after}'
@@ -213,6 +220,7 @@ tasks:
                                     # to `mach taskgraph decision` are all on the command line.
                                     $merge:
                                         - MOBILE_BASE_REPOSITORY: '${baseRepoUrl}'
+                                          MOBILE_BASE_REV: '${base_sha}'
                                           MOBILE_HEAD_REPOSITORY: '${repoUrl}'
                                           MOBILE_HEAD_REF: '${head_branch}'
                                           MOBILE_HEAD_REV: '${head_sha}'
--- a/taskcluster/ac_taskgraph/__init__.py
+++ b/taskcluster/ac_taskgraph/__init__.py
@@ -8,10 +8,15 @@ import os
 
 from importlib import import_module
 from taskgraph.parameters import extend_parameters_schema
-from voluptuous import Required
+from voluptuous import All, Any, Range, Required
 
 from .build_config import get_components, get_version
 
+
+extend_parameters_schema({
+    Required("pull_request_number"): Any(All(int, Range(min=1)), None),
+    Required("base_rev"): Any(basestring, None),
+})
 
 def register(graph_config):
     """
@@ -36,6 +41,10 @@ def _fill_treeherder_groups(graph_config):
 
 
 def get_decision_parameters(graph_config, parameters):
+    pr_number = os.environ.get("MOBILE_PULL_REQUEST_NUMBER", None)
+    parameters["pull_request_number"] = None if pr_number is None else int(pr_number)
+    parameters["base_rev"] = os.environ.get("MOBILE_BASE_REV")
+
     if parameters["tasks_for"] == "github-release":
         head_tag = parameters["head_tag"].decode("utf-8")
         if not head_tag:
--- a/taskcluster/ac_taskgraph/files_changes.py
+++ b/taskcluster/ac_taskgraph/files_changes.py
@@ -1,0 +1,22 @@
+from taskgraph.util.taskcluster import get_session
+
+
+def get_files_changed_pr(base_repository, pull_request_number):
+    url = base_repository.replace("github.com", "api.github.com/repos")
+    url += "/pulls/%s/files" % pull_request_number
+
+    r = get_session().get(url, timeout=60)
+    r.raise_for_status()
+    files = [f["filename"] for f in r.json()]
+    return files
+
+
+def get_files_changed_push(base_repository, base_rev, head_rev):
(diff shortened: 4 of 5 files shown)


mozbhearsum:
I think this is ready for a round of feedback/review. I'll note that I haven't gone to the trouble of setting up an entire other repo to push `github-push` behaviour in context, but I've done lots of local testing by hand with `taskgraph` for that and `github-pull-request`. This PR demonstrates the ability to filter away all of the build tasks, and my local tests have testing against other PRs with only some changed components.

I originally implemented this as a new `target tasks` filter for CI, but that didn't work out because the `complete` kind depends on all of the builds, and would re-add them to the final task graph even if they were filtered away. Adjusting those dependencies is really difficult, because they're set at task generation time. The easiest way to get rid of them was to not generate them at all. @tomprince suggested the alternative of something like https://hg.mozilla.org/ci/taskgraph/file/tip/src/taskgraph/morph.py#l187, but I _think_ for this case it's best to keep the solution in repo. I could be convinced otherwise, though. The main downside to not generating the tasks at all is that they're totally unrunnable, but I don't see any reason we'd want to run them after the fact, so I think that's an acceptable trade off.

As noted above, there's still a few open questions. @pocmo - do you have any thoughts or insight to the following?
* Figure out what actual entries we need in FILE_PATTERNS_TO_AFFECTED_COMPONENTS. That is, what files that are _not_ part of a component's source tree imply that we need to do builds of one or more components?
* See if we can get dependencies dumped from gradle in JSON or another more parseable format (not a blocker, but it would be nice to have).
* Determine whether or not the build-samples-browser tasks should be subject to conditional building.

mozbhearsum:
I'm also not sure why bors is not reporting in -- I'm wondering if it's being held up on the fact that not all of the tasks have run? (I don't know where to get more information on it.)

taskcluster/ac_taskgraph/loader/build_config.py
```diff
@@ -13,7 +17,113 @@
...
+
+def get_deps_for_component(component):
+    """Return the full list of local upstream dependencies of a component."""
+    # Gradle is expensive to run. Cache the results for each component.
+    # TODO: replace me with a functools.lru_cache decorator when this runs on py3
+    if component in cached_deps:
+        return cached_deps[component]
+
+    deps = set()
+    # Parsing output like this is not ideal but bhearsum couldn't find a way
```
JohanLorenzo, line 73: `gradle` allows us to define custom tasks. We made `printVariants` in Fenix, for instance: https://github.com/mozilla-mobile/fenix/blob/667261f511b7fdd7f45e289502fa7a481512b108/app/build.gradle#L660. 

I spent an hour and half on trying to make such a task but I couldn't. @pocmo would you have an idea?

taskcluster/ac_taskgraph/loader/build_config.py
```diff
@@ -13,7 +17,113 @@
...
+    # to get the dependencies printed in a better format. If we could convince
+    # gradle to spit out JSON that would be much better.
+    for line in subprocess.check_output(['./gradlew', '%s:dependencies' % component, '--configuration', 'implementation']).splitlines():
```
JohanLorenzo, line 76: TIL this command!

taskcluster/ac_taskgraph/loader/build_config.py
```diff
@@ -13,7 +17,113 @@
...
+def get_affected_components(files_changed):
+    affected_components = set()
+
+    # First, find the list of changed components as well as all of their
+    # local upstreams.
+    for c in get_components_changed(files_changed):
+        affected_components.add(c)
+        affected_components.update(get_deps_for_component(c))
+
+    # Then, look for any other affected components based on the files changed.
```
JohanLorenzo, line 97: Good thinking! On tope of my head, I don't know if we should track any other files. Any idea @pocmo?

taskcluster/ac_taskgraph/loader/build_config.py
```diff
@@ -28,12 +138,19 @@ def loader(kind, path, config, params, loaded_tasks):
...
         )
     }
+    # Filter away overridden jobs that we wouldn't build anyways to avoid ending up with
+    # partial job entries.
+    overridden_jobs = {k: v for k, v in config.pop('overriden-jobs', {}).items() if affected_components is ALL_COMPONENTS or k in affected_components}
     jobs = merge(jobs, config.pop('overriden-jobs', {}))
 
     config['jobs'] = jobs
 
+    import pprint
```
JohanLorenzo, line 153: Nit: Remove this `print` before landing.

taskcluster/ac_taskgraph/loader/build_config.py
```diff
@@ -13,7 +17,113 @@
...
+    # Just a placeholder example
+    #"^.*changelog.md.*$": ["build-browser-awesomebar"],
+    "^build.gradle$": ALL_COMPONENTS,
+}
+
+def get_files_changed_pr(base_repository, pull_request_number):
+    url = base_repository.replace("github.com", "api.github.com/repos")
+    url += "/pulls/%s/files" % pull_request_number
+
+    r = requests.get(url, timeout=60)
```
JohanLorenzo, line 36: Followup: Let's retry this call

taskcluster/ac_taskgraph/loader/build_config.py
```diff
@@ -13,7 +17,113 @@
...
+    r.raise_for_status()
+    files = [f["filename"] for f in r.json()]
+    return files
+
+def get_files_changed_push(base_repository, base_rev, head_rev):
+    url = base_repository.replace("github.com", "api.github.com/repos")
+    url += "/compare/"
+    url += "%s...%s" % (base_rev, head_rev)
+
+    r = requests.get(url, timeout=60)
```
JohanLorenzo, line 46: Followup: Let's retry this call.

taskcluster/ac_taskgraph/transforms/complete.py
```diff
@@ -17,6 +17,11 @@
 @transforms.add
 def fill_dependencies(config, tasks):
     for task in tasks:
+        # TODO: This currently includes some tasks that have been removed from
+        # target tasks. How do we avoid including these here?
+        # Best ideas are:
+        # 1) Move this entire task to a morph (like https://hg.mozilla.org/ci/taskgraph/file/tip/src/taskgraph/morph.py#l187)
```
JohanLorenzo, line 23: I kinda like this idea, actually 🙂

The complete task as it is now, has this known bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1642977. When I first implemented it, I felt like I was fighting against the Taskcluster paradigm. Moving it to morph would actually solve bug 1642977, if I understand correctly. Moreover, it will be easier to maintain the dependency list of the complete task since we want to wait on all tasks. As of now, we have to manually update `kind-dependencies`.

JohanLorenzo:
> I'm also not sure why bors is not reporting in -- I'm wondering if it's being held up on the fact that not all of the tasks have run? (I don't know where to get more information on it.)

Oh, that's expected while somewhat confusing, I concede. `bors` only runs when someone requests it. You get a yellow badge because it's a required Github check. This means: even though bors did nothing, Github waits for it. If I remember correctly, we made this check required in order to prevent manual landings. Bors takes care of everything.

taskcluster/ac_taskgraph/loader/build_config.py
```diff
@@ -13,7 +17,113 @@
 from ..build_config import get_components
 
 
+logger = logging.getLogger(__name__)
+
+ALL_COMPONENTS = object()
+# We always build components that have changed files or dependencies, but we also need
+# to rebuild some or all components when other files change. This structure
+# maintains the latter mapping.
+FILE_PATTERNS_TO_AFFECTED_COMPONENTS = {
```
tomprince, line 26: This would be better as a top-level key in the `kind.yml`.

taskcluster/ac_taskgraph/loader/build_config.py
tomprince: You can use `taskgraph.util.taskcluster.get_session` to get this.

taskcluster/ac_taskgraph/loader/build_config.py
```diff
@@ -13,7 +17,113 @@
...
+    # Just a placeholder example
+    #"^.*changelog.md.*$": ["build-browser-awesomebar"],
+    "^build.gradle$": ALL_COMPONENTS,
+}
+
+def get_files_changed_pr(base_repository, pull_request_number):
```
tomprince, line 32: Maybe put this and the next utility functions in `ac_taskgraph.files_changes` maybe (I suspect they'll eventually want to live in the [corresponding upstream file](https://hg.mozilla.org/ci/taskgraph/file/tip/src/taskgraph/files_changed.py)).

taskcluster/ac_taskgraph/loader/build_config.py
```diff
@@ -13,7 +17,113 @@
...
+def get_components_changed(files_changed):
+    """Translate a list of files changed into a list of components. Eg:
+        [
+            "components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/GeckoLoginStorageDelegate.kt",
+            "components/service/sync-logins/src/test/java/mozilla/components/service/sync/logins/GeckoLoginStorageDelegateTest.kt",
+        ]
+        ->
+        {service-sync-logins}
+    """
+    return set(["-".join(f.split("/")[1:3]) for f in files_changed if f.startswith("components")])
```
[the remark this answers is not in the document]
mozbhearsum: That's probably another thing that @pocmo should comment on :). That's a big assumption on my part right now.

taskcluster/ac_taskgraph/loader/build_config.py
```diff
@@ -13,7 +17,113 @@
...
+
+
+cached_deps = {}
+
+def get_deps_for_component(component):
```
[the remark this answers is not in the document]
mozbhearsum: TIL! That's great!

taskcluster/ac_taskgraph/loader/build_config.py
mozbhearsum: I'm hoping this list will be short (and I'd be happy for it to be empty), but I suspect at least some top level gradle files will need be included...

taskcluster/ac_taskgraph/loader/build_config.py
mozbhearsum: Good idea.

taskcluster/ac_taskgraph/loader/build_config.py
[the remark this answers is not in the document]
pocmo: This is something that has naturally evolved and wasn't really designed, I think. The first folder is the categories we think of (browser, feature, concept, ..) and then it's the name of the component in that category. But I think we can make that mandatory if it helps us setting up automation. :)

taskcluster/ac_taskgraph/loader/build_config.py
pocmo: I don't have any good ideas here. Like Johan I'd probably spend quite some time trying various things.. :) Maybe we can file this separately as a follow-up and then we can see if we can find a better solution using gradle? :)

taskcluster/ac_taskgraph/loader/build_config.py
pocmo: Of the top of my head I'd say a full rebuild should be needed for:
* Changes in `buildSrc`
* Changes in root `build.gradle` and `settings.gradle`
* Changes in gradle.properties
* Changes in gradle/wrapper/gradle-wrapper.properties

taskcluster/ac_taskgraph/loader/build_config.py
mozbhearsum: That sounds good to me. I'll open a new Issue for it.

taskcluster/ac_taskgraph/loader/build_config.py
mozbhearsum: > Just to clarify and make sure I understand it correctly...
> 
> This patch will:
> 
>     * Schedule only the tasks for components that should be affected by a push

Correct.
> 
> This patch will NOT:
> 
>     * Share build artifacts between tasks (when building component C depending on A and B then the task for C will also build A and B).

Correct - the internals of the build tasks and their task dependencies are totally unchanged. This patch _only_ prunes out tasks from the existing list of tasks.

>     * Rebuild the dependency tree in taskcluster, e.g. if A depends on B and B depends on C then we won't build C first, then B then A, but instead build A, B, C in parallel (which makes sense without sharing build artifacts)

Right. We're extracting the dependency tree from gradle in the decision task. I have no desire to maintain a parallel dependency tree :).

I'd love to see the latter two things happen at some point, but I think it's much more work for much less of a win - so I have no intention to look at them in the near future.

mozbhearsum:
> Just to clarify and make sure I understand it correctly...
> 
> This patch will:
> 
>     * Schedule only the tasks for components that should be affected by a push

Correct.
> 
> This patch will NOT:
> 
>     * Share build artifacts between tasks (when building component C depending on A and B then the task for C will also build A and B).

Correct - the internals of the build tasks and their task dependencies are totally unchanged. This patch _only_ prunes out tasks from the existing list of tasks.

>     * Rebuild the dependency tree in taskcluster, e.g. if A depends on B and B depends on C then we won't build C first, then B then A, but instead build A, B, C in parallel (which makes sense without sharing build artifacts)

Right. We're extracting the dependency tree from gradle in the decision task. I have no desire to maintain a parallel dependency tree :).

I'd love to see the latter two things happen at some point, but I think it's much more work for much less of a win - so I have no intention to look at them in the near future.

taskcluster/ac_taskgraph/loader/build_config.py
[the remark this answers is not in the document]
mozbhearsum: OK, sounds good. Even if we end up with a deeper directory structure we can likely deal with it. This is probably something to worry about if/when it happens.

taskcluster/ac_taskgraph/loader/build_config.py
mozbhearsum: https://github.com/mozilla-mobile/android-components/issues/7814

taskcluster/ac_taskgraph/transforms/complete.py
mozbhearsum: This sounds like good follow-up fodder :). Do you want a new bug or issue for it?

mozbhearsum:
I think this is ready to go now. I fixed all of the review comments, and also a bug I found that caused forced rebuilds of specific components (based on file patterns) to not consider dependencies. We don't have any of those right now (all of the forced ones are `ALL_COMPONENTS`), but at least it won't bite us in the future.

taskcluster/ac_taskgraph/target_tasks.py
```diff
@@ -4,7 +4,7 @@
 
 from __future__ import absolute_import, print_function, unicode_literals
 
-from taskgraph.target_tasks import _target_task, filter_for_tasks_for
+from taskgraph.target_tasks import _target_task, filter_for_tasks_for, standard_filter
```
JohanLorenzo, line 7: Nit: unused `standard_filter`

JohanLorenzo:
bors r=pocmo,JohanLorenzo

PR #7618, opened 2020-07-04
Record send-tab telemetry for Fenix.

mhammond:
application-services needs to be closely involved in creating the
data to be recorded in telemetry - however, we are trying to avoid
exposing this telemetry directly to the android-components "concepts"
or "features" - for example, the part of android-components that sends
a tab ideally wouldn't have to deal with any interface changes just because
telemetry is being sent or changed - otherwise new/changed telemetry is always
a "breaking change"

In the future, we expect the rust components to interact directly with Glean,
and consumers like android-components wouldn't need to get involved at all.
However, until then...

The idea with this PR is that there's one new public account function,
`processPendingTelemetry()`, which should be called whenever something is
done which might record telemetry. However, it doesn't need to know
what is actually recorded - that knowledge is only in the telemetry
code - the code that is tightly bound to glean.

This hasn't been tested yet - but it builds :) I'm soliciting feedback on
the general shape of this before I invest too much more in tests etc.

I also considered trying to do with Observers(), but that didn't seem
any easier. I'm obviously open to alternative approaches!

See also https://github.com/mozilla/application-services/pull/3308

---

CC @grigoryk @linacambridge @rfk 

### Pull Request checklist

### After merge


components/concept/sync/src/main/java/mozilla/components/concept/sync/OAuthAccount.kt
```diff
@@ -247,6 +247,24 @@ interface OAuthAccount : AutoCloseable {
...
+     * should be able to record telemetry directly, at which time this can be
+     * removed.
+     *
+     * The nature of telemetry is that it's somewhat ephemeral, so exactly when
+     * this should be called and what is recorded isn't documented here.
+     *
+     * While this is an async function, it is "fire and forget" - there's no
+     * need to await.
+     */
+    fun processPendingTelemetry()
```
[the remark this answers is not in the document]
mhammond: That's much simpler - I failed to see the wood for the trees :) I didn't rename `SyncTelemetry` but let me know if you would like me to.

settings.gradle
```diff
@@ -95,7 +95,13 @@ if (localProperties != null) {
 
     if (appServicesLocalPath != null) {
         logger.lifecycle("Enabling automatic publication of application-services from: $appServicesLocalPath")
-        def publishAppServicesCmd = ["./automation/publish_to_maven_local_if_modified.py"]
+        // Windows can't execute .py files directly, so we assume a "manually installed" python,
```
grigoryk, line 98: I guess just need to rebase this PR? this stuff landed.

components/support/sync-telemetry/src/main/java/mozilla/components/support/sync/telemetry/SyncTelemetry.kt
```diff
@@ -284,4 +287,42 @@ object SyncTelemetry {
         val message = reason.message ?: "Unexpected error: ${reason.code}"
         metric.set(message.take(MAX_FAILURE_REASON_LENGTH))
     }
+
+    fun processFxaTelemetry(jsonStr: String) {
+        val json = try {
+            JSONObject(jsonStr)
+        } catch (e: JSONException) {
+            logger.error("Invalid JSON in FxA telemetry", e)
```
grigoryk, line 295: (throughout this function)

Do you care to know if this actually happens in the wild?

Three options:
- crash the app here by throwing
- submit a caught exception to Sentry (via CrashReporter, which you'd need to pass-into this function)
- set some "faulty fxa telemetry detected" glean boolean metric

I think we should do at least one of these (and probably not crash), unless there's some other way to verify that telemetry is broken.

components/support/sync-telemetry/src/main/java/mozilla/components/support/sync/telemetry/SyncTelemetry.kt
```diff
@@ -284,4 +287,42 @@ object SyncTelemetry {
         val message = reason.message ?: "Unexpected error: ${reason.code}"
         metric.set(message.take(MAX_FAILURE_REASON_LENGTH))
     }
+
+    fun processFxaTelemetry(jsonStr: String) {
```
grigoryk, line 291: Maybe you could link to wherever this schema is defined in the app-services repo? It'll be nice to have that handy here.

components/support/sync-telemetry/src/main/java/mozilla/components/support/sync/telemetry/SyncTelemetry.kt
```diff
@@ -284,4 +287,42 @@ object SyncTelemetry {
...
+                val one = recd.getJSONObject(i)
+                val extras = mapOf(
+                    FxaTab.receivedKeys.flowId to one.getString("flow_id"),
+                    FxaTab.receivedKeys.streamId to one.getString("stream_id"),
+                    FxaTab.receivedKeys.reason to one.getString("reason")
+                )
+                FxaTab.received.record(extras)
+            }
+            logger.info("Reported telemetry for ${recd.length()} received commands")
+        } catch (e: JSONException) {
```
grigoryk, line 324: maybe one big try/catch, with centralized error processing? i guess you don't get granular bits of telemetry in that case though..

components/support/sync-telemetry/metrics.yaml
```diff
@@ -443,3 +443,50 @@ bookmarks_sync:
...
+    type: event
+    description: >
+      Recorded when a tab is sent.
+    send_in_pings:
+      - sync
+    extra_keys:
+      flow_id:
+        description: >
+          A guid, unique for the URL being sent but common for all target
+          devices.
```
liuche, line 457: Looking at other documentation and the Desktop data review, this should be fine, but could you clarify here?

Does "unique guid for the url being sent" imply that this is some kind of one-way hash from the url? Is its uniqueness is scoped to the Firefox account, and this `flow_id` would not be the same for other clients?

Mainly I want to verify something like "the ID is a randomly generated GUID so can not identify a **user or tab**" from
https://bug1639781.bmoattachments.org/attachment.cgi?id=9157561
(but that's from stream id)

Since this is also documented in the main firefox documentation, would recommend linking to this doc here as well:
https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/data/sync-ping.html

components/support/sync-telemetry/src/main/java/mozilla/components/support/sync/telemetry/SyncTelemetry.kt
```diff
@@ -284,4 +288,45 @@ object SyncTelemetry {
         val message = reason.message ?: "Unexpected error: ${reason.code}"
         metric.set(message.take(MAX_FAILURE_REASON_LENGTH))
     }
+
+    fun processFxaTelemetry(jsonStr: String, crashReporter: CrashReporting? = null) {
+        val json = try {
+            JSONObject(jsonStr)
+        } catch (e: JSONException) {
+            crashReporter?.submitCaughtException(e)
```
grigoryk, line 296: Nice! I found it helpful to create a separate exception class for "caught" exceptions, makes it much easier to actually filter for them in Sentry: https://github.com/mozilla-mobile/android-components/blob/master/components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt#L137

components/support/sync-telemetry/metrics.yaml
mhammond: The GUID is randomly generated and unrelated to the tab or user - I've updated the descriptions to reflect that and also add the link back to desktop.

components/support/sync-telemetry/src/main/java/mozilla/components/support/sync/telemetry/SyncTelemetry.kt
mhammond: OK, I did what I think you were suggesting :)

components/support/sync-telemetry/src/main/java/mozilla/components/support/sync/telemetry/SyncTelemetry.kt
```diff
@@ -4,21 +4,33 @@
...
+import mozilla.components.support.sync.telemetry.GleanMetrics.FxaTab
 import mozilla.components.support.sync.telemetry.GleanMetrics.HistorySync
 import mozilla.components.support.sync.telemetry.GleanMetrics.LoginsSync
 import mozilla.components.support.sync.telemetry.GleanMetrics.Sync
 import mozilla.components.support.sync.telemetry.GleanMetrics.Pings
 
 const val MAX_FAILURE_REASON_LENGTH = 100
 
+// The exceptions we report to the crash reporter, but otherwise don't escape this module.
+sealed class InvalidTelemetryException(cause: Exception) : Exception(cause) {
```
grigoryk, line 27: nit: could be `internal` so that it doesn't into the public API (and linter won't force you to write kdocs).

grigoryk:
Overall this looks good to me! I guess you should mark it for a review at some point :-)

mhammond:
> Overall this looks good to me! I guess you should mark it as ready for a review at some point :-)

Thanks - the app-services part is yet to land :/ Once that does I'll see about making a new release etc and marking this ready.

grigoryk:
@mhammond I think this can go ahead now that mozilla/application-services#3308 landed?

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
```diff
@@ -1015,7 +1015,7 @@ open class FxaAccountManager(
 
     @VisibleForTesting
     open fun createAccount(config: ServerConfig): OAuthAccount {
-        return FirefoxAccount(config)
+        return FirefoxAccount(config, null, crashReporter)
```
mhammond, line 1018: I'm a bit confused by why the `PersistCallback` is being passed as null here - is there maybe some other creator of a `FirefoxAccount` which is passed a non-null `PersistCallback` but will now be getting a null `crashReporter`?

mhammond:
That taskcluster failure looks unrelated to this change

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FxaDeviceConstellation.kt
```diff
@@ -111,6 +114,7 @@ class FxaDeviceConstellation(
                 when (outgoingCommand) {
                     is DeviceCommandOutgoing.SendTab -> {
                         account.sendSingleTab(targetDeviceId, outgoingCommand.title, outgoingCommand.url)
+                        SyncTelemetry.processFxaTelemetry(account.gatherTelemetry())
```
grigoryk, line 117: you're not passing in `crashReporter` to `processFxaTelemetry`

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FirefoxAccount.kt
```diff
@@ -27,7 +28,8 @@ typealias PersistCallback = mozilla.appservices.fxaclient.FirefoxAccount.Persist
  */
 @Suppress("TooManyFunctions")
 class FirefoxAccount internal constructor(
-    private val inner: InternalFxAcct
+    private val inner: InternalFxAcct,
+    private val crashReporter: CrashReporting? = null
```
grigoryk, line 32: did you mean to use this for the FxaDeviceConstellation constructor?

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
grigoryk: as mentioned on slack, this should be fine, just a bad API for FirefoxAccount. It will register a persist callback internally regardless of what you pass in here.

grigoryk:
bors retry

PR #7623, opened 2020-07-06
Issue #7529: Add ContainerState to BrowserState

### LINKED ISSUE #7529: Add ContainerState to BrowserState

gabrielluong:
- Adds initial container state, actions and reducers to BrowserState
- Refactor `feature/containers/Container.kt` to `browser/state/state/ContainerState.kt`

---

Fixes #7529

### Pull Request checklist

### After merge


components/browser/state/src/main/java/mozilla/components/browser/state/state/ContainerState.kt
```diff
@@ -2,7 +2,22 @@
...
+ *
+ * @property contextId The session context ID also known as cookie store ID for the container.
+ * @property name Name of the container.
+ * @property color The color for the container. This can be shown in tabs belonging to this container.
+ * @property icon The name of an icon for the container.
+ */
+data class ContainerState(
+    override val contextId: String,
+    override val name: String,
+    override val color: String,
```
pocmo, line 18: I wonder why `Color` is a `String`? Wouldn't it make more sense to have an `Int` here that we can use as a color in UI code?

components/browser/state/src/main/java/mozilla/components/browser/state/state/ContainerState.kt
```diff
@@ -2,7 +2,22 @@
...
+    override val icon: String
```
pocmo, line 19: I'm also wondering about the type here, what does "name of an icon" actually mean? How can I display it? Could this be a Bitmap? Or a Drawable? Or a file path? How do I currently turn this name into something I can display?

components/browser/state/src/main/java/mozilla/components/browser/state/state/ContainerState.kt
```diff
@@ -2,7 +2,22 @@
...
+) : Container
```
pocmo, line 20: Where is this interface coming from? 🤔 Is it needed?

components/browser/state/src/main/java/mozilla/components/browser/state/state/BrowserState.kt
```diff
@@ -13,7 +13,8 @@ import mozilla.components.lib.state.State
  * @property tabs the list of open tabs, defaults to an empty list.
  * @property selectedTabId the ID of the currently selected (active) tab.
  * @property customTabs the list of custom tabs, defaults to an empty list.
- * @property extensions A map of extension ids and web extensions of all installed web extensions.
+ * @property containers A map of session context IDs and their respective container [ContainerState].
```
NotWoods, line 16: The use of session here makes me think of session IDs. Maybe just say "A map of context IDs" or "A map of [SessionState.contextId]"?

gabrielluong: Thanks, this is a good change. Will have this resolved in my next revision.

components/browser/state/src/main/java/mozilla/components/browser/state/state/ContainerState.kt
gabrielluong: Changed to `@ColorInt val color: Int,` in my next revision.

components/browser/state/src/main/java/mozilla/components/browser/state/state/ContainerState.kt
gabrielluong: Changed to Drawable

components/browser/state/src/main/java/mozilla/components/browser/state/state/ContainerState.kt
```diff
@@ -0,0 +1,56 @@
...
+ * @property name Name of the container.
+ * @property color The color for the container. This can be shown in tabs belonging to this container.
+ * @property icon The icon for the container.
+ */
+data class ContainerState(
+    val contextId: String,
+    val name: String,
+    val color: Color,
+    val icon: Icon
+) {
```
gabrielluong, line 20: Curious if this should be a Parcelable from comparing it to `SitePermissions`?

pocmo: No, I do not think so. I can't see any need for passing instances around inside a bundle (e.g. from activity to activity). Everything can talk to the store directly. If we need to pass anything around then it's probably the id and then the receiver can lookup the right state from the store with that id.

gabrielluong: Thank you that is very insightful about Parcelable.

gabrielluong:
> ContainerStorage

Yes, I believe we can make this `internal` since only the middleware will needs this from my prototyping.

gabrielluong:
bors r=pocmo

PR #7625, opened 2020-07-06
Update LeakCanary to 2.4 and add to Sample Browser

csadilek:
Sets up LeakCanary in Sample Browser. With the upgrade to 2.4 we also don't need custom code for instrumented tests anymore: https://square.github.io/leakcanary/upgrading-to-leakcanary-2.0/


taskcluster/scripts/toolchain/android-gradle-dependencies.sh
```diff
@@ -19,15 +19,15 @@ pushd $PROJECT_DIR
...
-COMPONENTS_TO_EXCLUDE_IN_FIRST_PASS='(samples-browser|tooling-detekt|tooling-lint)'
+# Components to ignore on first pass. They don't have any extra dependencies so we can
+# filter them out. Other components are built in second pass.
+COMPONENTS_TO_EXCLUDE_IN_FIRST_PASS='(tooling-detekt|tooling-lint)'
 FIRST_PASS_COMPONENTS=$(grep -E "$COMPONENT_REGEX" "$PROJECT_DIR/.buildconfig.yml" | sed -E "s/$COMPONENT_REGEX/:\1/g" | grep -E -v "$COMPONENTS_TO_EXCLUDE_IN_FIRST_PASS")
 
 ASSEMBLE_COMMANDS=$(echo "$FIRST_PASS_COMPONENTS" | sed "s/$/:assemble/g")
 ASSEMBLE_TEST_COMMANDS=$(echo "$FIRST_PASS_COMPONENTS" | sed "s/$/:assembleAndroidTest/g")
 TEST_COMMANDS=$(echo "$FIRST_PASS_COMPONENTS" | sed "s/$/:test/g")
-LINT_COMMANDS=$(echo "$FIRST_PASS_COMPONENTS" | sed "s/$/:lintRelease/g")
```
csadilek, line 30: There's no `lintRelease` in Sample Browser, so switching to `lint`.

csadilek:
@JohanLorenzo Can you take a look at this too, please? New `LeakCanary` no longer requires custom code to set up for tests and can therefore be added to Sample Browser directly. However, Sample Browser was excluded when building up dependencies so we included it here...otherwise dependency resolution fails.

taskcluster/scripts/toolchain/android-gradle-dependencies.sh
JohanLorenzo: I don't remember why we prefer using `lintRelease` over `lint`. That said, if we change this value here, we need to update https://github.com/mozilla-mobile/android-components/blob/f074469a2fe08cf0ca52e57d86d45c2900d9ff88/taskcluster/ci/build/kind.yml#L86. 

In my opinion, we should keep filtering out `samples-browser` in the first pass, and add it alongside `tooling-lint`https://github.com/mozilla-mobile/android-components/blob/f074469a2fe08cf0ca52e57d86d45c2900d9ff88/taskcluster/scripts/toolchain/android-gradle-dependencies.sh#L43-L46

Both have no `lintRelease` per https://github.com/mozilla-mobile/android-components/blob/f074469a2fe08cf0ca52e57d86d45c2900d9ff88/taskcluster/ci/build/kind.yml#L70

What do you think @csadilek?

csadilek:
> Thanks for looping me in, @csadilek!
> 
> I looked at the changes under `taskcluster/`. I'm glad `samples-browser` can now be built! Although, I'm not sure we should change all components to use `lint` instead of `lintRelease`. What do you think?

Thanks, @JohanLorenzo! I've updated the PR, can you take another look?

components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
```diff
@@ -35,7 +36,7 @@ import java.util.Locale
 @Suppress("LargeClass")
 class AddonManager(
     private val store: BrowserStore,
-    private val engine: Engine,
+    private val engine: WebExtensionRuntime,
```
csadilek, line 39: Already removed this change, but GitHub is slow to update today...

csadilek:
bors r=pocmo,JohanLorenzo,jonalmeida

csadilek:
bors retry

PR #7631, opened 2020-07-06
Turn on Gradle configure on demand

NotWoods:
Gradle optimizations for CI. When this flag is turned on, only modules that are dependencies will be configured in a CI task. For example, build-support-base only needs to compile support-base and support-test. Sample Browser still depends on everything, so this just helps with individual tasks.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -13,9 +13,14 @@ org.gradle.jvmargs=-Xmx4096m
 
 # When configured, Gradle will run in incubating parallel mode.
 # This option should only be used with decoupled projects. More details, visit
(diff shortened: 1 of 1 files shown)


grigoryk:
Looking at two build scans, one with this flag [turned off](https://scans.gradle.com/s/mcqeavjani3kw/performance/configuration) and another with it [turned on](https://scans.gradle.com/s/n4cirab3zlxta/performance/configuration), I see no absolutely difference in how tasks are configured.

This makes sense, since we now do a decent job of using the task avoidance APIs, and it's not very clear to me what if any additional improvements this flag gets us.

We can still land this - I don't immediately see any regressions. But yeah, seems like a no-op to me.

Issue #7647, opened 2020-07-08
Releasing session to switch from custom tab to browser does not work anymore


pocmo:
Fenix issue: https://github.com/mozilla-mobile/fenix/issues/12311

This looks like it is a regression from the refactoring to use `BrowserStore` in the `SessionFeature`.

pocmo:
After adding a similar switch action to sample browser I can reproduce it there too:

```
         AndroidRuntime  E  FATAL EXCEPTION: main
                         E  Process: org.mozilla.samples.browser, PID: 515
                         E  java.lang.IllegalStateException: SET SESSION: Current activity: BrowserActivity Other activity: ExternalAppBrowserActivity
                         E      at mozilla.components.browser.engine.gecko.GeckoEngineView.render(GeckoEngineView.kt:124)
                         E      at mozilla.components.feature.session.engine.EngineViewPresenter.onEngineSession(EngineViewPresenter.kt:53)
                         E      at mozilla.components.feature.session.engine.EngineViewPresenter.access$onEngineSession(EngineViewPresenter.kt:22)
                         E      at mozilla.components.feature.session.engine.EngineViewPresenter$start$1$invokeSuspend$$inlined$collect$1.emit(Collect.kt:137)
                         E      at mozilla.components.support.ktx.kotlinx.coroutines.flow.FlowKt$ifChanged$$inlined$filter$1$2.emit(Collect.kt:147)
                         E      at mozilla.components.feature.session.engine.EngineViewPresenter$start$1$invokeSuspend$$inlined$map$2$2.emit(Collect.kt:139)
                         E      at mozilla.components.feature.session.engine.EngineViewPresenter$start$1$invokeSuspend$$inlined$map$1$2.emit(Collect.kt:139)
                         E      at kotlinx.coroutines.flow.FlowKt__ChannelsKt.emitAllImpl$FlowKt__ChannelsKt(Channels.kt:59)
                         E      at kotlinx.coroutines.flow.FlowKt__ChannelsKt$emitAllImpl$1.invokeSuspend(Unknown Source:11)
                         E      at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
                         E      at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:56)
                         E      at android.os.Handler.handleCallback(Handler.java:907)
                         E      at android.os.Handler.dispatchMessage(Handler.java:99)
[...]

PR #7650, opened 2020-07-08
Issue #7647: Add SessionFeature.release() to allow app to stop rendering engine sessions and release the current one.

pocmo:
Previously Fenix just called `engineView.release()`. However since `EnginePresenter` is now listening to `BrowserStore` it may re-render the session that we just released. This new methods stops the presenter from updating and releases the current session.

### DIFF
--- a/components/feature/session/src/main/java/mozilla/components/feature/session/SessionFeature.kt
+++ b/components/feature/session/src/main/java/mozilla/components/feature/session/SessionFeature.kt
@@ -56,4 +56,15 @@ class SessionFeature(
     override fun stop() {
         presenter.stop()
     }
+
+    /**
+     * Stops the feature from rendering sessions on the [EngineView] (until explicitly started again)
+     * and releases an already rendering session from the [EngineView].
+     */
+    fun release() {
+        // Once we fully migrated to BrowserStore we may be able to get rid of the need for cleanup().
+        // See https://github.com/mozilla-mobile/android-components/issues/7657
+        presenter.stop()
+        engineView.release()
+    }
 }
--- a/components/feature/session/src/test/java/mozilla/components/feature/session/SessionFeatureTest.kt
+++ b/components/feature/session/src/test/java/mozilla/components/feature/session/SessionFeatureTest.kt
@@ -279,6 +279,56 @@ class SessionFeatureTest {
     }
 
     @Test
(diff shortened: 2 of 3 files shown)


components/feature/session/src/main/java/mozilla/components/feature/session/SessionFeature.kt
```diff
@@ -56,4 +56,13 @@ class SessionFeature(
     override fun stop() {
         presenter.stop()
     }
+
+    /**
+     * Stops the feature from rendering sessions on the [EngineView] (until explicitly started again)
+     * and releases an already rendering session from the [EngineView].
+     */
+    fun release() {
```
Amejia481, line 64: We need a new patch in Fenix for calling `sessionFeature.release()` instead of `engineView.release()`, right?

pocmo: Yeah, that will happens as part of https://github.com/mozilla-mobile/fenix/issues/12311

Amejia481: nit: We could also add `release()` to the changelog :)

pocmo: Yeah, good point, I keep forgetting about the changelog.

Amejia481: > Yeah, that will happens as part of [mozilla-mobile/fenix#12311](https://github.com/mozilla-mobile/fenix/issues/12311)

Awesome :)

csadilek: Filed a follow-up, as discussed: https://github.com/mozilla-mobile/android-components/issues/7657

With a new ID, we should be able to remove the need for the app to call `release`, as the custom tab presenter wouldn't ever render the regular tab.

pocmo: Great, thank you!

pocmo:
bors r=Amejia481,csadilek

st3fan:
I removed the 79ReleaseBlocker label - do not uplift to the beta branch, we do not need it there.

PR #7651, opened 2020-07-08
[RFC] Introducing an RFC process.

pocmo:
Hey folks. A long time ago we briefly talked about having something like an RFC process to discuss larger changes before writing any code (I kind of tried that [for brower-state](https://github.com/pocmo/android-components/pull/3) back then and it was quite helpful). I would like to revive that effort. And in true tautological fashion this is an RFC for having an RFC process.

A bunch of other projects have RFC processes too. I think what we need is a very lightweight version of that. We probably do not need a super elaborate and complicated process. The main goal here is to foster and document discussions around substantial(!) changes before writing code and opening a PR - and having this discussion open for everyone.

What do you think? I have been drafting a more detailed document describing the process for our documentation site. I'll attach it to that PR later.


docs/rfcs/0001-rfc-process.md
```diff
@@ -0,0 +1,64 @@
...
+title: RFC process
+permalink: /rfc/0001-rfc-process
+---
+
+* Start date: 2020-07-07
+* RFC PR: [#7651](https://github.com/mozilla-mobile/android-components/pull/7651)
+
+## Summary
+
+Introducing a lightweight RFC ("request for comments") process for proposing and discussing "substantial" changes and for building consensus.
```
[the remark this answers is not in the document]
pocmo: I think you are right, we need some examples. I tried to formulate some in the documentation (that is not attached to the PR yet), but I only have "rewriting a component" - where we will soon have an actual example of :)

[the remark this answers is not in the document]
csadilek: I'd also suggest elaborating a bit on when we should do this and define what we mean by "substantial." 

Here's a first suggestion:

A change is substantial if it
- affects multiple components;
- affects how components interact through either their public or internal APIs;
- fundamentally changes how a component is implemented, how it manages state or reacts to changes, in a way that isn't self-explanatory or a direct result of a bug fix.

docs/rfcs/0001-rfc-process.md
```diff
@@ -0,0 +1,64 @@
...
+* Building consensus among the team before potentially writing a lot of code.
+
+## Guide-level explanation
+
+The high-level process of creating an RFC is:
+
+* Create an RFC document (like this one) using the template.
+* Open a pull request for the RFC document.
+* Ask for feedback on the pull request, via the [mailing list]() or in [chat](https://chat.mozilla.org/#/room/#android-components:mozilla.org).
+
```
csadilek, line 32: Personally, I would also like to see some sample code. Do you think that's reasonable requirement? It doesn't have to be working, it could be a sketch as part of the RFC PR or a link to a draft PR. It will make it much easier to understand the bigger picture. Of course, it has to be clear that this is not about the specific code, but the design / architecture.

docs/rfcs/0001-rfc-process.md
```diff
@@ -0,0 +1,64 @@
...
+After potentially some rounds of changes and after the discussion phase has concluded:
+
+* If consensus has been reached then the RFC is considered "accepted" and gets merged into the repository for documentation purposes.
+* If consensus has not been reached then the RFC is considered "rejected" and the pull request gets closed. An RFC proposal may get revived in the future if requirements have changed.
+
+Once an RFC becomes accepted then authors may implement it and submit the feature as a pull request.
+
+## Drawbacks
+
+* Writing an RFC is additional overhead and may feel slower or cumbersome. The assumption is that the advantages still outnumber this drawback. Additionally this lightweight RFC process is simplified compared to other projects, in the hope of better fitting the needs of the team at this time.
```
csadilek, line 47: nit: s/outnumber/outweigh

csadilek: nit Additionally, this....

docs/rfcs/0001-rfc-process.md
pocmo: In addition to this RFC (once landed) I will write a documentation (I can share my draft) that includes a template that can be copied and pasted to start an RFC. Do you think that would be sufficient?

pocmo:
Thank you everyone for the feedback so far. I just updated the RFC and added everything from your comments. :)

I also want to point you to the following PR where we start testing the process with an actual change we want to make (the RFC is still in draft though):
https://github.com/mozilla-mobile/android-components/pull/7655

docs/rfcs/0001-rfc-process.md
csadilek: Yes, sounds good. If there's a snippet part of the template I think that's sufficient.

pocmo:
I'm going to merge this now: After 12 days it seems like everyone who commented was in favor of having an RFC process. I'm going to put up another PR adding documentation about the RFC process. In that PR we can fine-tune the details. :)

pocmo:
bors r=csadilek,Amejia481,jonalmeida,psymoon,rocketsroger

PR #7652, opened 2020-07-08
Added additional tests for functionality FxaAccountManager and FxaDev…

chartjes:
…iceConstellation used by sendtab

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
+++ b/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
@@ -150,7 +150,7 @@ open class FxaAccountManager(
     private val logger = Logger("FirefoxAccountStateMachine")
 
     @Volatile
-    private var latestAuthState: String? = null
+    var latestAuthState: String? = null
 
     // This is used during 'beginAuthenticationAsync' call, which returns a Deferred<String>.
     // 'deferredAuthUrl' is set on this observer and returned, and resolved once state machine goes
--- a/components/service/firefox-accounts/src/test/java/mozilla/components/service/fxa/FxaAccountManagerTest.kt
+++ b/components/service/firefox-accounts/src/test/java/mozilla/components/service/fxa/FxaAccountManagerTest.kt
@@ -8,6 +8,7 @@ import android.content.Context
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LifecycleOwner
 import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.work.testing.WorkManagerTestInitHelper
 import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Deferred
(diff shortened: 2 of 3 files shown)


chartjes:
@eoger Thanks for the review, I will ask Grisha about next steps in getting in merged

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
```diff
@@ -150,7 +150,7 @@ open class FxaAccountManager(
     private val logger = Logger("FirefoxAccountStateMachine")
 
     @Volatile
-    private var latestAuthState: String? = null
+    var latestAuthState: String? = null
```
grigoryk, line 153: Consider adding `@VisibleForTesting` annotation.

components/service/firefox-accounts/src/test/java/mozilla/components/service/fxa/FxaAccountManagerTest.kt
```diff
@@ -1886,6 +1887,180 @@ class FxaAccountManagerTest {
         verify(syncManager, times(2)).stop()
         verifyNoMoreInteractions(syncManager)
     }
+    @Test
+    fun `beginAuthenticationAuthSync handles any pairingUrl correctly`() {
+        // Initialize things so the test can run
+        WorkManagerTestInitHelper.initializeTestWorkManager(testContext)
+
+        // Build our dependencies for creating an account manager
+        val tServerConfig = ServerConfig(
```
grigoryk, line 1896: nit: var name prefixes aren't really a convention in Kotlin, maybe just `serverConfig`? or `testServerConfig` if you want to spell it out.

components/service/firefox-accounts/src/test/java/mozilla/components/service/fxa/FxaAccountManagerTest.kt
```diff
@@ -1886,6 +1887,180 @@ class FxaAccountManagerTest {
...
+        val tSyncConfig = SyncConfig(setOf(SyncEngine.Bookmarks))
+        val manager = FxaAccountManager(
+                testContext,
+                serverConfig = tServerConfig,
+                deviceConfig = tDeviceConfig,
+                syncConfig = tSyncConfig
+        )
+        val pairingUrl = "https://localhost/pairingUrl".toString()
+        val beginResponse = manager.beginAuthenticationAsync(pairingUrl)
+        assertTrue(beginResponse is CompletableDeferred<String?>)
```
grigoryk, line 1915: Unless you `await` for this to resolve, you're not really testing anything. This `CompletableDeferred<String?>` is just a promise to return a value of `String?` at some point in the future, e.g. once this API request goes through the internal state machine.

What you want to do is wrap this whole method in `runBlocking`, set the corresponding coroutine context on your test account manager (`FxaAccountManager(... coroutineContext = this@runBlocking.coroutineContext ...)` and do something like `assertEquals("auth://url", manager.beginAuthenticationAsync().await())`. This will ensure that the internal account manager state machine will actually go through its motions in response to  a beginAuthenticationAsync call, using the same dispatcher as is used for your test code.

However, lots of other tests in this class already exercise `beginAuthenticationAsync`, so as-is this test isn't adding any value.

Name of the test says "any pairing url", however you're just passing in one. Did you intend to test `beginAuthenticationAsync` against a few different inputs?

components/service/firefox-accounts/src/test/java/mozilla/components/service/fxa/FxaAccountManagerTest.kt
```diff
@@ -1886,6 +1887,180 @@ class FxaAccountManagerTest {
...
+                setOf()
+        )
+        val tSyncConfig = SyncConfig(setOf(SyncEngine.Bookmarks))
+        val manager = FxaAccountManager(
+                testContext,
+                serverConfig = tServerConfig,
+                deviceConfig = tDeviceConfig,
+                syncConfig = tSyncConfig
+        )
+        val beginResponse = manager.beginAuthenticationAsync(null)
```
grigoryk, line 1941: Same comment as above, currently this is (almost) not testing anything. You want to at least `await` here, and assert the actual returned value.

components/service/firefox-accounts/src/test/java/mozilla/components/service/fxa/FxaAccountManagerTest.kt
```diff
@@ -1886,6 +1887,180 @@ class FxaAccountManagerTest {
...
+        assertTrue(beginResponse is CompletableDeferred<String?>)
+    }
+
+    @Test
+    fun `integration test with latest auth state matching current auth state`() {
```
grigoryk, line 1946: We already have tests which assert that state validation is performed correctly. Please see `auth state verification while finishing authentication` and neighbouring tests.

Tests below are incomplete (same comments about `await`, etc) and appear to be mostly redundant to me?

If existing tests aren't covering cases you're thinking of, please modify them instead of adding entirely new tests.

components/service/firefox-accounts/src/test/java/mozilla/components/service/fxa/FxaAccountManagerTest.kt
```diff
@@ -1886,6 +1887,180 @@ class FxaAccountManagerTest {
...
+        val expectedAuthData = FxaAuthData(
+                authType = AuthType.Signin,
+                code = "testCode1",
+                state = "testState",
+                declinedEngines = emptySet()
+        )
+        // Override lastAuthState
+        manager.latestAuthState = null
+
+        // Make sure the logger is triggered
```
grigoryk, line 2017: You're not asserting that a log call is performed, though?

components/service/firefox-accounts/src/test/java/mozilla/components/service/fxa/FxaDeviceConstellationTest.kt
```diff
@@ -232,6 +239,68 @@ class FxaDeviceConstellationTest {
         assertEquals(ConstellationState(currentDeviceExpired.into(), listOf(testDevice2.into())), constellation.state())
     }
 
+    @Test
+    @ExperimentalCoroutinesApi
+    fun `polling for commands with real account`() = runBlockingTest {
```
grigoryk, line 244: Any specific reason you're using `runBlockingTest` instead of `runBlocking`? It's mainly useful if you need to skip over any delays in your test code, which I don't think is the case here.

grigoryk: See all the other tests in this file for an established pattern for dealing with coroutines.

components/service/firefox-accounts/src/test/java/mozilla/components/service/fxa/FxaDeviceConstellationTest.kt
```diff
@@ -232,6 +239,68 @@ class FxaDeviceConstellationTest {
...
+            val targetDeviceId = "testDevice"
+            val title = "Mozilla"
+            val url = "https://mozilla.org"
+            val response = testConstellation.sendCommandToDeviceAsync(
+                    targetDeviceId = targetDeviceId,
+                    outgoingCommand = DeviceCommandOutgoing.SendTab(title, url)
+            )
+
+            assertNotNull(response)
+            verify(testAccount.sendSingleTab(targetDeviceId, title, url))
```
grigoryk, line 300: You're note awaiting for a response to return, so I'm not entirely sure how this is supposed to work...

components/service/firefox-accounts/src/test/java/mozilla/components/service/fxa/FxaDeviceConstellationTest.kt
```diff
@@ -232,6 +239,68 @@ class FxaDeviceConstellationTest {
...
+    }
+
+    @Test
+    fun `test sending one tab`() = runBlockingTest {
+        // Create a scope to run this test inside
+        val testScope = GlobalScope
+
+        testScope.launch {
+            // Initialize things so the megazord is happy
+            WorkManagerTestInitHelper.initializeTestWorkManager(testContext)
```
grigoryk, line 277: Why do you need the work manager here?

components/service/firefox-accounts/src/test/java/mozilla/components/service/fxa/FxaDeviceConstellationTest.kt
```diff
@@ -232,6 +239,68 @@ class FxaDeviceConstellationTest {
...
+            Megazord.init()
```
grigoryk, line 278: Consider moving this to this test class' `init` block, otherwise you're calling `init` for each test.

components/service/firefox-accounts/src/test/java/mozilla/components/service/fxa/FxaDeviceConstellationTest.kt
```diff
@@ -232,6 +239,68 @@ class FxaDeviceConstellationTest {
...
+
+            // Create a test account
+            val testAccount = spy(NativeFirefoxAccount(
```
grigoryk, line 281: Not sure you actually need to `spy` here at all. You just need to verify that `sendCommandToDeviceAsync` resolves to `true`, meaning that there were not exceptions/etc encountered during that API call.

I suspect that it actually won't resolve to `true`, but fail with some `FxaException`.

components/service/firefox-accounts/src/test/java/mozilla/components/service/fxa/FxaDeviceConstellationTest.kt
```diff
@@ -232,6 +239,68 @@ class FxaDeviceConstellationTest {
...
+    fun `polling for commands with real account`() = runBlockingTest {
+        val testScope = GlobalScope
+
+        testScope.launch {
+            // Initialize things so the megazord is happy
+            WorkManagerTestInitHelper.initializeTestWorkManager(testContext)
+            Megazord.init()
+
+            // Create a test account
+            val testAccount = spy(NativeFirefoxAccount(
```
grigoryk, line 253: I think it will be more valuable to attempt to switch existing tests in this class to use `NativeFirefoxAccount` (instead of the mock that they're currently using).

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
grigoryk: But actually, after reading the test, I don't think you need to make this public at all. Existing tests which ensure we validate auth state don't need  this to be public.

chartjes: I did figure out how to set that value without making that variable public via some spelunking through the code. That has been fixed.

components/service/firefox-accounts/src/test/java/mozilla/components/service/fxa/FxaAccountManagerTest.kt
chartjes: I have refactored all those variables to use 'test' instead of 't' as the prefix.

PR #7655, opened 2020-07-08
[RFC] Refactoring browser-search component.

pocmo:
This RFC is not really ready for full review yet. However I still want to open the PR already because:

* It acts as an actual example for an RFC (https://github.com/mozilla-mobile/android-components/pull/7651)
* I want to work on the proposal with @boek and @psymoon

### DIFF
--- a/docs/rfcs/0002-browser-search-rewrite.md
+++ b/docs/rfcs/0002-browser-search-rewrite.md
@@ -1,0 +1,128 @@
+---
+layout: page
+title: Replacing `browser-search` component with state in `browser-store`
+permalink: /rfc/0002-search-state-in-browser-store
+---
+
+* Start date: 2020-07-07
+* RFC PR: [#7655](https://github.com/mozilla-mobile/android-components/pull/7655)
+
+## Summary
+
+Moving all search related state to `BrowserState` and replacing `SearchEngineManager` with a `BrowserStore` middleware that deals with saving and loading state; better satisfying the requirements of [Fenix](https://github.com/mozilla-mobile/fenix) and [Focus](https://github.com/mozilla-mobile/focus-android).
+
+## Motivation
+
+The integration of `browser-search` into [Fenix](https://github.com/mozilla-mobile/fenix) and [Firefox for Fire TV](https://github.com/mozilla-mobile/firefox-tv/) showed that the current API is not enough to satisfy all use cases. Some examples include:
+
+* Fenix supports custom search engines. This can only be achieved with `browser-search` by passing a custom `SearchEngineProvider` to `SearchEngineManager`. Everything else is left to the app.
+* `SearchEngineManager` initially only provided synchronous methods for accessing search engines. Later asynchronous methods were added. The result is a hard to understand mix that can cause blocking I/O to happen on the wrong threads.
+* The `browser-search` component does not follow our abstraction model of implementing a `concept-search` component. This makes it impossible to switch or customize implementations.
+* The wrapper code in Fenix makes it hard to argue about the state that is split over two implementations. In addition to that it makes it hard to add functionality to `SearchEngineManager` since it is not used directly.
+* Nowadays Android Components bundles all state in `BrowserState` observable through `BrowserStore` (provided by `browser-state`). On the application side we are using an UI-scoped `Store` (provided by `lib-store`). `SearchEngineManager` creates a secondary source of truth just for search and circumvents this architecture pattern.
+* In "Firefox for Fire TV" and "Firefox for Echo Show" we needed to override the default search configuration, adding and replacing loaded search engines. This turned out to be quite complicated and required wrapping multiple classes that deal with loading the search configuration and engines.
+
+## Guide-level explanation
+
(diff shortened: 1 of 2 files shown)


pocmo:
Jeff, Simon and I met yesterday to talk about this and the conclusion of the meeting was:

image

(`SearchEngine`s should live in `BrowserState` and the persistence layer should be handled transparently by a `Middleware` on `BrowserStore`)

I'm going to update the RFC to reflect that.

pocmo:
@psymoon @boek I think this is ready for an initial review. Do you think that makes sense? Do you see problems or missing pieces?

pocmo:
@Mugurell You had the pleasure to work with the search code in Fenix too. We are considering rewriting the search component to make it nicer and satisfy the use cases of Fenix without the need for wrapper code in Fenix. This is the RFC for that change. :)

docs/rfcs/0002-browser-search-rewrite.md
```diff
@@ -0,0 +1,127 @@
...
+internal class SearchEngineStorage {
+    suspend fun getSearchEngines() = withContext(Dispatchers.IO) {
+      // ...
+    }
+}
+```
+
+### Middleware
+
+A `SearchMiddleware` that will be installed on `BrowserStore` will be responsible for:
```
psymoon, line 87: What about the name `SearchManager`? I think we have manager classes in a-c with similar responsibilities as below.

docs/rfcs/0002-browser-search-rewrite.md
```diff
@@ -0,0 +1,127 @@
...
+
+Using the `BrowserStore` for state will make the app use asynchronous patterns to observe the state and with that avoid blocking the main thread, as well as force the app to deal with the initial state of having no (default) `SearchEngine` yet.
+
+Since the state will move to `browser-state`, additional functionality (querying suggestions, storage, middleware) can move to `feature-search`. This will make `browser-search` obsolete. With that introducing a `concept-search` is not required. An app can use a custom implementation by replacing the middleware, storage or handling search state completely differently.
+
+## Reference-level explanation
+
+The new implementation will use `browser-state` to model all search state.
+
+Functionality on top of the state (querying suggestions, storage, middleware) will live in `feature-search`. With that `browser-search` will no longer be needed and can stay around as-is until no longer used by any app.
```
Mugurell, line 38: How can `until no longer used by any app` be measured?

docs/rfcs/0002-browser-search-rewrite.md
```diff
@@ -0,0 +1,127 @@
...
+ */
+data class SearchState(
+  val searchEngines: List<SearchEngine>,
+  val defaultSearchEngineId: String
+)
+```
+
+`SearchState` will contain all `SearchEngine`s (bundled and custom). Additional extension methods will make it easier to select specific subsets or the default `SearchEngine`.
+
+`SearchEngine` will be turned into a data class and moved to `browser-store`. Other methods like `buildSearchUrl()` will be implemented as extension methods.
```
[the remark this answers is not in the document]
pocmo: Yeah, I was also thinking about something like a `type` property backed by an enum. I will add this to the RFC.

docs/rfcs/0002-browser-search-rewrite.md
pocmo: I think using `*Middleware` is more in line with other `*Middleware` that we have implemented now and it describes its purpose (to be added to the store as a middleware).
Using `*Manager` too vague? And it doesn't really "manage" the state anymore.

Some thing using "manager" is an anti pattern:
* http://www.bright-green.com/blog/2003_02_25/naming_java_classes_without_a.html
* https://blog.codinghorror.com/i-shall-call-it-somethingmanager/

docs/rfcs/0002-browser-search-rewrite.md
pocmo: This sentence was not really precise. I think what we want to do here is:
* Migrate our own applications over (primarily Fenix, but then also Focus etc.)
* Follow our deprecation process to deprecate and eventually remove the component: https://mozac.org/contributing/deprecating

docs/rfcs/0002-browser-search-rewrite.md
```diff
@@ -0,0 +1,127 @@
...
+### Middleware
+
+A `SearchMiddleware` that will be installed on `BrowserStore` will be responsible for:
+
+* Loading the initial set of search engines (from the two storages above) and adding them to `BrowserState` by dispatching actions on `BrowserStore`.
+* Persisting state changes it observes in the specific storage:
+  * Adding search engines
+  * Removing search engines
+  * Changing the default search engine
+* Updating search engines at runtime (e.g. region or locale change)
```
Mugurell, line 94: Region change from MLS?
How would this work?
Polling at regular intervals seems rather wasteful.
Maybe this could be tried once per app start?

Could updating search engines change even the default currently used by the user?

pocmo: This is pretty much undefined at this point. The primary point here is that the middleware will take care of something like that and we will not pass that burden onto the app.

* Issue #1720 is about looking at more data sources than only MLS (or locale). In that same bug Mike Connor (who from our point of view owns the search engine plugins and region mapping) confirmed that we are in general okay with switching defaults. Of course that would need further clarification of how and when that would happen.
* As far as I know the desktop version will soon roll out a change that will make it re-request and in certain situation change its region.

pocmo:
bors r=csadilek,psymoon,Mugurell

PR #7668, opened 2020-07-09
Closes #7664: Add EngineSession.goToHistoryIndex to jump to a specifi…

### LINKED ISSUE #7664: Expose GeckoSession.goToHistoryIndex
For mozilla-mobile/fenix#1048, we need a way to jump to an arbitrary index in a session's history. It looks like GeckoView already provides the API, it just isn't surfaced in A-C.

person808:
…c index in a sessions history.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -153,6 +153,13 @@ class GeckoEngineSession(
     }
 
     /**
+     * See [EngineSession.goToHistoryIndex]
+     */
+    override fun goToHistoryIndex(index: Int) {
+        geckoSession.gotoHistoryIndex(index)
+    }
+
+    /**
      * See [EngineSession.saveState]
      */
     override fun saveState(): EngineSessionState {
--- a/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
+++ b/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
@@ -38,7 +38,7 @@ internal val xRequestHeader = mapOf(
 /**
  * WebView-based EngineSession implementation.
  */
-@Suppress("TooManyFunctions")
+@Suppress("LargeClass", "TooManyFunctions")
 class SystemEngineSession(
     context: Context,
     private val defaultSettings: Settings? = null
@@ -123,6 +123,14 @@ class SystemEngineSession(
      */
     override fun goForward() {
         webView.goForward()
+    }
+
+    /**
+     * See [EngineSession.goToHistoryIndex]
+     */
+    override fun goToHistoryIndex(index: Int) {
+        val historyList = webView.copyBackForwardList()
+        webView.goBackOrForward(index - historyList.currentIndex)
     }
 
     /**
--- a/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
+++ b/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
@@ -466,6 +466,15 @@ abstract class EngineSession(
     abstract fun goForward()
 
     /**
+     * Navigates to the specified index in the [HistoryState] of this session. The current index of
+     * this session's [HistoryState] will be updated but the items within it will be unchanged.
+     * Invalid index values are ignored.
+     *
+     * @param index the index of the session's [HistoryState] to navigate to
+     */
+    abstract fun goToHistoryIndex(index: Int)
+
+    /**
      * Saves and returns the engine state. Engine implementations are not required
      * to persist the state anywhere else than in the returned map. Engines that
      * already provide a serialized state can use a single entry in this map to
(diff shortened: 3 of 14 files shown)


components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
```diff
@@ -465,6 +465,11 @@ abstract class EngineSession(
      */
     abstract fun goForward()
 
+    /**
+     * Navigates to a specific index in the history of this sesson
```
jonalmeida, line 469: Could you add more docs on how this API is supposed to be used to move forward/backward?

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -152,6 +152,13 @@ class GeckoEngineSession(
         geckoSession.goForward()
     }
 
+    /**
+     * See [EngineSession.goToHistoryIndex]
+     */
+    override fun goToHistoryIndex(index: Int) {
+        geckoSession.gotoHistoryIndex(index)
```
jonalmeida, line 159: What happens if you provide an index outside the range of the index available? We should check to see what the different engine requirements and see if we can follow the same pattern for them all.

Maybe something like, [`coerceIn`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/coerce-in.html) either at the engine-level or at the use case level?

config/detekt-baseline.xml
```diff
@@ -210,6 +210,7 @@
     <ID>UndocumentedPublicClass:SessionUseCases.kt$SessionUseCases$ExitFullScreenUseCase</ID>
     <ID>UndocumentedPublicClass:SessionUseCases.kt$SessionUseCases$GoBackUseCase</ID>
     <ID>UndocumentedPublicClass:SessionUseCases.kt$SessionUseCases$GoForwardUseCase</ID>
+    <ID>UndocumentedPublicClass:SessionUseCases.kt$SessionUseCases$GoToHistoryIndexUseCase</ID>
```
jonalmeida, line 213: You could always add the docs too. :)

These detekt warnings existed before we modified our lint configuration to catch it. They were ignored as part of a baseline so that you don't have to add docs for all the code that didn't already exist, but just for the new methods.we add.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
person808: `GeckoSession` seems to ignore invalid values. I'm not sure how to test `SystemEngineSession` though

jonalmeida: @person808 in android components you can switch the build variant to use the System engine with the Sample Browser.

Screen Shot 2020-07-09 at 6 38 43 PM

components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
jonalmeida: Also any side-effects if invalid values used.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
person808: `SystemEngineSession` also ignores invalid values. Maybe it would be best to document that implementations should ignore invalid values? Coercing to a valid value would mean accessing `BrowserStore` to get the history for the session since AFAIK thats the only way to get the session history for `GeckoSession`. What do you think?

jonalmeida: Documenting sounds good. This is a particularly awkward API that uses an index value, no need to make it more complex.

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
```diff
@@ -125,6 +125,14 @@ class SystemEngineSession(
         webView.goForward()
     }
 
+    /**
+     * See [EngineSession.goToHistoryIndex]
+     */
+    override fun goToHistoryIndex(index: Int) {
+        val historyList = webView.copyBackForwardList()
+        webView.goBackOrForward(index - historyList.currentIndex)
```
jonalmeida, line 133:
```suggestion
        webView.goBackOrForward(index)
```
If I read [the docs](https://developer.android.com/reference/android/webkit/WebView#goBackOrForward(int)) correctly, you shouldn't need to use the currentIndex to calculate the direction, right?

person808: The `index` parameter corresponds to the indices in `HistoryState` which are `0..n`. So if we want to go to say index `0` when our `currentIndex = 3` we need to pass `-3`. If we don't do the math then we would call `webView.goBackOrForward(0)` which does nothing.

jonalmeida: I read it again and it makes sense now. I prefer the GV API more. 😄

jonalmeida:
@person808 we can try to merge your change again if you can fix the tests. :)

person808:
bors r=jonalmeida

PR #7681, opened 2020-07-10
Closes #7680: Allow new window onLoadRequest to be intercepted

### LINKED ISSUE #7680: Intercept new window requests to external apps
Currently any new window request in `onLoadRequest` is ignored and not intercepted.  In some cases when user taps on a link, it request a new window but the URL should be opened in a third party application.  Support this use case.

rocketsroger:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -364,10 +364,6 @@ class GeckoEngineSession(
             session: GeckoSession,
             request: NavigationDelegate.LoadRequest
         ): GeckoResult<AllowOrDeny> {
-            if (request.target == NavigationDelegate.TARGET_WINDOW_NEW) {
-                return GeckoResult.fromValue(AllowOrDeny.ALLOW)
-            }
-
             // The process switch involved when loading extension pages will
             // trigger an initial load of about:blank which we want to
             // avoid:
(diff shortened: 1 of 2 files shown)


components/browser/engine-gecko-nightly/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineSessionTest.kt
```diff
@@ -1902,15 +1902,58 @@ class GeckoEngineSessionTest {
     }
 
     @Test
-    fun `Handle new window load requests`() {
```
csadilek, line 1905: The original test was testing load requests, now the new test is "just" covering the app link interceptor. Let's also cover the case where we don't intercept?

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -364,10 +364,6 @@ class GeckoEngineSession(
             session: GeckoSession,
             request: NavigationDelegate.LoadRequest
         ): GeckoResult<AllowOrDeny> {
-            if (request.target == NavigationDelegate.TARGET_WINDOW_NEW) {
-                return GeckoResult.fromValue(AllowOrDeny.ALLOW)
-            }
```
csadilek, line 369: One difference here is that, before when handling a new window request, we didn't notify observers of `onLoadRequest`, now we will notify. Did you check the consequences of that? There's two observers in Fenix which seem fine? One is for telemetry and one for "expanding" the toolbar.

rocketsroger: Agreed, I'll double check Fenix side usage.

components/browser/engine-gecko-nightly/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineSessionTest.kt
rocketsroger: Sure, I'll add test cases where we don't intercept.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
rocketsroger: Looking more into this.  You're right, I don't think we should notify the observers when `TARGET_WINDOW_NEW` is set as the target.  I'll add an extra check before notifying the observer.  Thanks,

components/browser/engine-gecko-nightly/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineSessionTest.kt
```diff
@@ -1902,15 +1902,105 @@ class GeckoEngineSessionTest {
...
+            mock(), mockLoadRequest("sample:about", null,
+            GeckoSession.NavigationDelegate.TARGET_WINDOW_NEW, triggeredByRedirect = false)
+        )
+
+        assertNotNull(observedIntent)
+        assertEquals("result", observedUrl)
+    }
+
+    @Test
+    fun `onLoadRequest will not intercept new window blank page load requests`() {
```
csadilek, line 1960: Hm this isn't quite right. What you're testing here is that we can open new windows e.g. (target="_blank" pages). These aren't blank pages, just pages we want to open in a new tab.

The test is otherwise right. I would use a different example than `about:blank` and a different name e.g.

```onLoadRequest allows new window requests if not intercepted``` (or similar) :)

rocketsroger: Sure, I'll add another URL in the test as part of the update as well.  Thanks,

rocketsroger:
bors r=csadilek

Issue #3322, opened 2019-06-07
Check for authentication erorrs coming out of sync via typed exception


grigoryk:
As of #3283 we are checking for auth errors that can happen during sync by looking at the message embedded in the `PlacesException`. Once places lib grows a `PlacesAuthException` type, we should switch to that.

PR #7694, opened 2020-07-13
Bumps Application Services to v62.0.0

tarikeshaq:
Connects to https://github.com/mozilla/application-services/issues/3337
Bumps Application Services to 62.0.0

# v62.0.0 (_2020-07-13_)

[Full Changelog](https://github.com/mozilla/application-services/compare/v61.0.7...v62.0.0)

### ⚠️ Breaking changes ⚠️

- Adds support for `entrypoint` in oauth flow APIs: consumers of `beginOAuthFlow` and `beginPairingFlow` (`beginAuthentication` and `beginPairingAuthentication` in ios) are now ***required*** to pass an `entrypoint` argument that would be used for metrics. This puts the `beginOAuthFlow` and `beginPairingFlow` APIs inline with other existing APIs, like `getManageAccountUrl`.  ([#3265](https://github.com/mozilla/application-services/pull/3265))
- Changes the `authorizeOAuthCode` API to now accept an `AuthorizationParams` object instead of the individual parameters. The `AuthorizationParams` also includes optional `AuthorizationPKCEParams` that contain the `codeChallenge`, `codeChallengeMethod`. `AuthorizationParams` also includes an optional `keysJwk` for requesting keys ([#3264](https://github.com/mozilla/application-services/pull/3264))

### What's new ###
- Consumers can now optionally include parameters for metrics in `beginOAuthFlow` and `beginPairingFlow` (`beginAuthentication` and `beginPairingAuthentication` in ios). Those parameters can be passed in using a `MetricsParams` struct/class. `MetricsParams` is defined in both the Kotlin and Swift bindings. The parameters are the following ([#3328](https://github.com/mozilla/application-services/pull/3328)):
  - flow_id
  - flow_begin_time
  - device_id
  - utm_source
  - utm_content
  - utm_medium
  - utm_term
  - utm_campaign
  - entrypoint_experiment
  - entrypoint_variation

### What's fixed ###

- Fixed a bug where attempting to edit a login with an empty `form_submit_url` would incorrectly
  reject the entry as invalid ([#3331](https://github.com/mozilla/application-services/pull/3331)).

### What's new ###

- Tab records now have an explicit TTL set when storing on the server, to match the behaviour
  of Firefox Desktop clients ([#3322](https://github.com/mozilla/application-services/pull/3322)).

### DIFF
--- a/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FirefoxAccount.kt
+++ b/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FirefoxAccount.kt
@@ -11,6 +11,7 @@ import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.plus
+import mozilla.appservices.fxaclient.AuthorizationParams
 import mozilla.appservices.fxaclient.FirefoxAccount as InternalFxAcct
 import mozilla.components.concept.sync.AccessType
 import mozilla.components.concept.sync.AuthFlowUrl
@@ -101,7 +102,7 @@ class FirefoxAccount internal constructor(
 
     override fun beginOAuthFlowAsync(scopes: Set<String>) = scope.async {
         handleFxaExceptions(logger, "begin oauth flow", { null }) {
-            val url = inner.beginOAuthFlow(scopes.toTypedArray())
+            val url = inner.beginOAuthFlow(scopes.toTypedArray(), "none")
             val state = Uri.parse(url).getQueryParameter("state")!!
             AuthFlowUrl(state, url)
         }
@@ -109,7 +110,7 @@ class FirefoxAccount internal constructor(
 
     override fun beginPairingFlowAsync(pairingUrl: String, scopes: Set<String>) = scope.async {
         handleFxaExceptions(logger, "begin oauth pairing flow", { null }) {
-            val url = inner.beginPairingFlow(pairingUrl, scopes.toTypedArray())
+            val url = inner.beginPairingFlow(pairingUrl, scopes.toTypedArray(), "none")
             val state = Uri.parse(url).getQueryParameter("state")!!
             AuthFlowUrl(state, url)
         }
@@ -140,7 +141,8 @@ class FirefoxAccount internal constructor(
         accessType: AccessType
     ) = scope.async {
         handleFxaExceptions(logger, "authorizeOAuthCode", { null }) {
-            inner.authorizeOAuthCode(clientId, scopes, state, accessType.msg)
+            val authParams = AuthorizationParams(clientId, scopes, state, accessType.msg)
+            inner.authorizeOAuthCode(authParams)
         }
     }
 
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -27,7 +27,7 @@ object Versions {
     const val disklrucache = "2.0.2"
     const val leakcanary = "1.6.3"
 
-    const val mozilla_appservices = "61.0.7"
+    const val mozilla_appservices = "62.0.0"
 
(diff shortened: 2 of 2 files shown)


components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FirefoxAccount.kt
```diff
@@ -101,15 +102,15 @@ class FirefoxAccount internal constructor(
...
+            val url = inner.beginOAuthFlow(scopes.toTypedArray(), "notset")
             val state = Uri.parse(url).getQueryParameter("state")!!
             AuthFlowUrl(state, url)
         }
     }
 
     override fun beginPairingFlowAsync(pairingUrl: String, scopes: Set<String>) = scope.async {
         handleFxaExceptions(logger, "begin oauth pairing flow", { null }) {
-            val url = inner.beginPairingFlow(pairingUrl, scopes.toTypedArray())
+            val url = inner.beginPairingFlow(pairingUrl, scopes.toTypedArray(), "notset")
```
tarikeshaq, line 113: Will create an issue for this, probably going to require upstream consumers to set the `entrypoint`.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FirefoxAccount.kt
```diff
@@ -140,7 +141,8 @@ class FirefoxAccount internal constructor(
         accessType: AccessType
     ) = scope.async {
         handleFxaExceptions(logger, "authorizeOAuthCode", { null }) {
-            inner.authorizeOAuthCode(clientId, scopes, state, accessType.msg)
+            val authParams = AuthorizationParams(clientId, scopes, state, accessType.msg)
+            inner.authorizeOAuthCode(authParams)
```
rfk, line 145: Not necessarily for this PR, but flagging for future consideration - we should consider pushing the `AuthorizationParams` abstraction up through the API layers here, so that callers of the higher-level API can pass all the new properties in a sensible way.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FirefoxAccount.kt
rfk: Right. Adding a bit of context: FxA basically considers any attempt to load the login page without an `entrypoint` parameter to be a metrics bug, as we have no way of attributing the signin metrics for that flow back to a particular UI touchpoint. Making this a required argument surfaces that requirement to the application, so we probably don't want to paper over it at this level. (Although @grigoryk might prefer to paper over it here for a short period while considering the API implications).

jonalmeida:
bors try

^ To run UI automation.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FirefoxAccount.kt
rfk: If we do decide to temporarily hard-code a default entrypoint here, I suggest the string "none"; this is what FxA uses internally when you don't provide an explicit entrypoint value.

eliserichards:
Ping! What's the timeline on getting this merged? 🙏 😉

grigoryk:
@eliserichards should merge today, hopefully.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FirefoxAccount.kt
grigoryk: Agreed, we can do that in a follow-up.

grigoryk:
New plan: we'll land 61.0.9 first (with a logins crash fix and the metrics extension), and hopefully try to get it onto beta/nightly first.

That gives us time to actually do the API bits for this PR properly.

grigoryk:
Closing this PR, I'll do a pass on this instead which will make the necessary API changes (#7729 , #7730 ).

Issue #7712, opened 2020-07-14
Allow splitting search suggestion provider for awesomebar into two providers


pocmo:
For: https://github.com/mozilla-mobile/fenix/issues/7877

On a slow network connection the search suggestion that matches the entered text is appearing slow as well. We can work around that by filtering exact matches from `SearchSuggestionProvider` and providing a distinct provider that only returns the exact match.

This also seems to feel faster on devices with good network connections.

PR #7713, opened 2020-07-14
Issue #7712: Add flag for filtering exact matches from SearchSuggestionProvider and add SearchActionProvider.

pocmo:
This allows a consuming app to filter suggestions that match the entered text from `SearchSuggestionProvider`. Then they can add `SearchActionProvider` that provides a suggestion for the entered text. The benefit is that `SearchActionProvider` does not depend on querying a search engine first and therefore the suggestion will appear faster - especially on slow networks.


pocmo:
bors r=csadilek

PR #7733, opened 2020-07-15
Closes #7722: Record uncaught exception in Android log

### LINKED ISSUE #7722: Logging fatal exceptions
Since we stopped passing caught exceptions to the default exception handler of the system I no longer see the full stacktrace in the log.

Only when disabling the crash reporter I see them again:
Screenshot 2020-07-15 at 12 24 06

We should consider logging the stacktrace ourselves if we do not pass the exception to the system anymore.

rocketsroger:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/lib/crash/src/main/java/mozilla/components/lib/crash/handler/ExceptionHandler.kt
+++ b/components/lib/crash/src/main/java/mozilla/components/lib/crash/handler/ExceptionHandler.kt
@@ -4,11 +4,14 @@
 
 package mozilla.components.lib.crash.handler
 
+import android.annotation.SuppressLint
 import android.content.Context
 import android.os.Process
+import android.util.Log
 import mozilla.components.lib.crash.Crash
 import mozilla.components.lib.crash.CrashReporter
 
+private const val TAG = "ExceptionHandler"
 /**
  * [Thread.UncaughtExceptionHandler] implementation that forwards crashes to the [CrashReporter] instance.
  */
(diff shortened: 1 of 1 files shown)


pocmo:
> [task 2020-07-15T22:12:59.086Z]   /builds/worker/checkouts/src/components/lib/crash/src/main/java/mozilla/components/lib/crash/handler/ExceptionHandler.kt:39: Error: Using Android Log instead of base component [LogUsage]
> [task 2020-07-15T22:12:59.086Z]               Log.e(TAG, "Uncaught exception handled: ", throwable)
> [task 2020-07-15T22:12:59.086Z]               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Our check is complaining that you are not using our logger. But in that case this may be fine. Let's just suppress it?

Also I wonder if we should move the log statement all the way to the top of `uncaughtException`:
* This will guarantee that it gets logged even if the following code does something unexpected
* In the even of a crash loop we exit early (`crashing` property). In the case this happens it would be great to log that exception too :)

rocketsroger:
> > [task 2020-07-15T22:12:59.086Z]   /builds/worker/checkouts/src/components/lib/crash/src/main/java/mozilla/components/lib/crash/handler/ExceptionHandler.kt:39: Error: Using Android Log instead of base component [LogUsage]
> > [task 2020-07-15T22:12:59.086Z]               Log.e(TAG, "Uncaught exception handled: ", throwable)
> > [task 2020-07-15T22:12:59.086Z]               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> 
> Our check is complaining that you are not using our logger. But in that case this may be fine. Let's just suppress it?
> 
> Also I wonder if we should move the log statement all the way to the top of `uncaughtException`:
> 
>     * This will guarantee that it gets logged even if the following code does something unexpected
> 
>     * In the even of a crash loop we exit early (`crashing` property). In the case this happens it would be great to log that exception too :)

Agreed.  I've updated the change.  Thanks

rocketsroger:
bors r=pocmo

PR #7736, opened 2020-07-16
Improve file picker selection

Amejia481:
---

### Pull Request checklist

### After merge


components/support/ktx/src/main/java/mozilla/components/support/ktx/android/net/Uri.kt
```diff
@@ -53,3 +57,22 @@ fun Uri.sameSchemeAndHostAs(other: Uri) = scheme == other.scheme && host == othe
...
+
+/**
+ * Indicate if the [this] uri is under the application private directory.
+ */
+fun Uri.isUnderPrivateAppDirectory(context: Context): Boolean {
+    return when (this.scheme) {
+        ContentResolver.SCHEME_FILE -> {
+            try {
+                val uriPath = path ?: return true
+                val uriCanonicalPath = File(uriPath).canonicalPath
```
Mugurell, line 69: Not sure why..
but on my Android 11 emulator canonicalPath would not resolve the `data/user/0` symlink to `data/data`.
Don't have a physical device with A11 to test..

components/feature/prompts/src/androidTest/java/mozilla/components/feature/prompts/file/OnDeviceFilePickerTest.kt
```diff
@@ -0,0 +1,119 @@
...
+    private val badUri2 = "file:///data/data/${context.packageName}/any_directory/file.text".toUri()
+    private val goodUri = "file:///data/directory/${context.packageName}/any_directory/file.text".toUri()
+
+    @Test
+    fun filterUrisUnderPrivateAppDirectory() {
+        val uris = arrayOf(badUri1, badUri2, goodUri)
+
+        val filterUris = uris.filterUrisUnderPrivateAppDirectory(context)
+
+        Assert.assertEquals(1, filterUris.size)
```
Mugurell, line 33: This fails for me on an Android 11 emulator.
Not sure why..
but the `data/user/0` symlink is not resolved to `data/data`.
Don't have a physical device with A11 to test..

components/feature/prompts/src/androidTest/java/mozilla/components/feature/prompts/file/OnDeviceFilePickerTest.kt
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    @Test
+    fun unsafeUrisWillNotBeSelected() {
+        val promptContainer = PromptContainer.TestPromptContainer(context)
+
+        val filePicker = FilePicker(promptContainer, BrowserStore()) { }
+        var onDismissWasExecuted = false
+
+        val filePickerRequest = PromptRequest.File(
```
Mugurell, line 44: Maybe it would be good have dismiss/selected asserts in all tests? - have all pickers constructed the same and then do all the same asserts, with only one difference depending on the Intent.
Just to make sure some later refactorization doesn't silently break something?

taskcluster/ci/test/kind.yml
```diff
@@ -111,4 +111,18 @@ jobs:
...
+                - ['automation/taskcluster/androidTest/ui-test.sh', 'feature-prompts', 'arm', '1']
+        treeherder:
+            symbol: 'unit-feature-prompts'
+    android-support-ktx:
+        description: 'Run unit tests on device for support ktx'
+        run:
+            post-gradlew:
+                - ['automation/taskcluster/androidTest/ui-test.sh', 'support-ktx', 'arm', '1']
+        treeherder:
+            symbol: 'unit-support-ktx'
```
Mugurell, line 128: nit: no newline

components/feature/prompts/src/androidTest/java/mozilla/components/feature/prompts/file/OnDeviceFilePickerTest.kt
```diff
@@ -0,0 +1,119 @@
...
+                onMultipleFilesSelected = { _, _ -> }
+        )
+
+        val intent = Intent()
+        intent.data = goodUri
+        filePicker.handleFilePickerIntentResult(intent, filePickerRequest)
+
+        assertTrue(onSingleFileSelectedWasExecuted)
+    }
+}
```
Mugurell, line 119: nit: no newline

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptContainer.kt
```diff
@@ -51,4 +52,10 @@ internal sealed class PromptContainer {
 
         override fun getString(res: Int) = fragment.getString(res)
     }
+
+    @VisibleForTesting
+    internal class TestPromptContainer(override val context: Context) : PromptContainer() {
```
csadilek, line 57: We should be able to move this somewhere under `src/test`.

components/support/ktx/src/main/java/mozilla/components/support/ktx/android/net/Uri.kt
```diff
@@ -53,3 +57,22 @@ fun Uri.sameSchemeAndHostAs(other: Uri) = scheme == other.scheme && host == othe
  * https://html.spec.whatwg.org/multipage/origin.html#same-origin
  */
 fun Uri.sameOriginAs(other: Uri) = sameSchemeAndHostAs(other) && port == other.port
+
+/**
+ * Indicate if the [this] uri is under the application private directory.
+ */
+fun Uri.isUnderPrivateAppDirectory(context: Context): Boolean {
```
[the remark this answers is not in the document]
Amejia481: Yes, that's correct. I would like to keep it. In case we introduce support in the future, and miss this check.  I comment could clarify why it's there. What do you think?

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/file/FilePicker.kt
```diff
@@ -159,13 +161,22 @@ internal class FilePicker(
         if (intent?.clipData != null && request.isMultipleFilesSelection) {
             intent.clipData?.run {
                 val uris = Array<Uri>(itemCount) { index -> getItemAt(index).uri }
-                request.onMultipleFilesSelected(container.context, uris)
+                val sanitizedUris = uris.filterUrisUnderPrivateAppDirectory(container.context)
```
csadilek, line 164: nit++ (just naming): `filter` retains elements for which the predicate is true. so, `filterUrisUnderPrivateAppDirectory` would suggest that we retain the URIs under the private app directory. Of course, we do the opposite here, we remove (or filter out) the URIs under private app dirs. Maybe `removeUrisUnderPrivateAppDir` or `filterOutUris...`. I think it would make it easier to read.

components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlin/String.kt
```diff
@@ -136,3 +137,11 @@ fun String.isSameOriginAs(other: String): Boolean {
 fun String.sanitizeURL(): String {
     return this.trim()
 }
+
+/**
+ * Remove any unwanted character from string containing file name.
+ * For example for with an input of "/../../../../../../directory/file.txt" you will get "file.txt"
```
csadilek, line 143: nit: unnecessary `with`.

components/support/ktx/src/main/java/mozilla/components/support/ktx/android/net/Uri.kt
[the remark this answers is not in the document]
kanytu: Then I see no problem in adding it. The logic seems okay 👍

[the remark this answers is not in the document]
Amejia481: 👍

components/feature/prompts/src/androidTest/java/mozilla/components/feature/prompts/file/OnDeviceFilePickerTest.kt
Amejia481: @kanytu  Do you have any thoughts on this? As the variations between Android 10 and 11 can affect the reliability of `isUnderPrivateAppDirectory` as it is.

Amejia481: @csadilek any thoughts ? :)

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptContainer.kt
csadilek: Or is it because you don't want to open up `PromptContainer`?

Amejia481: Yes

csadilek: OK, let's keep it then.

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/file/FilePicker.kt
Amejia481: 👍

components/feature/prompts/src/androidTest/java/mozilla/components/feature/prompts/file/OnDeviceFilePickerTest.kt
kanytu: I can confirm that in Android R, symlinks are not getting resolved with canonical path. I'm unsure if thats intended (as part of Scoped Storage) or if that's a bug. 

Either way, using `dataDir` instead of an hardcoded `/data/data/` will always be sure to match against private files, so I'm pretty sure this is still a reliable approach.

kanytu: Actually, I was wrong. If that doesn't resolve `/data/data/`, `/data/data/` files will still be considered non-private files. I would say that this should be checked once a stable version of R lands.

csadilek: How about we add a check for `data/user` specifically (in addition to the canonical path comparison) so the tests pass and we cover this case for current Android 11, where the symlink isn't resolved. We can file a follow-up ticket to clean this up / check again once Android 11 is out. This seems acceptable as we don't support `file://` URIs anyway right now?

Amejia481: it sounds reasonable to me!

components/support/ktx/src/main/java/mozilla/components/support/ktx/android/net/Uri.kt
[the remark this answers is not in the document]
kbrosnan: I second @Amejia481's comment about including this. Apps other than Fenix can consume the AC project. GV supports down to API 19 so there are still usecases where `file://` could be passed to AC.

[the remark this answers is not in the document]
kanytu: That is in fact a valid use case. However, as I said, AC still can't process those Uris. Unless they change how `getFileName` calculates a file name. Because, otherwise, AC will try to load a cursor from a `file://` which will just fail.

Amejia481:
bors try

Amejia481:
bors r=csadilek,Mugurell

st3fan:
@Amejia481 Since this landed on Nightly (right?) can we ask QA to test in this area? Maybe do similar functional testing as we did on Fennec? cc @sflorean @kbrosnan

Amejia481:
> @Amejia481 Since this landed on Nightly (right?) can we ask QA to test in this area? Maybe do similar functional testing as we did on Fennec? cc @sflorean @kbrosnan

Yes, we just have to verify that ac version that Fenix is using contains this patch. I will verify and confirm on this issue

Amejia481:
Fenix nightly contains the change, we can start testing!

sflorean:
@Amejia481 @st3fan, we are working on it, @softvision-miralobontiu is currently testing and she will leave a comment with the results.

sflorean:
Test results Fenix 79.0.0: https://bugzilla.mozilla.org/show_bug.cgi?id=1652364#c43.

sflorean:
Extra testing results: https://bugzilla.mozilla.org/show_bug.cgi?id=1652364#c49

PR #7740, opened 2020-07-16
Remove more properties from Session.

pocmo:
Removing a bunch of properties where our code is already reading them exclusively from `BrowserState`. The only change needed in Fenix is to no longer pass `SessionManager` to `ContextMenuUseCases`.


pocmo:
bors r=csadilek

PR #7746, opened 2020-07-16
Issue #7712: SearchActionProvider: Pass in suspending function instead of a Deferred<SearchEngine>

pocmo:
While integrating the provider into Fenix I came to the conclusion that it makes much more sense
to pass in a suspending function. With that the caller does not need to deal with scopes and
dispatchers themselves.

(Follow-up from #7713)

### DIFF
--- a/components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/AwesomeBarFeature.kt
+++ b/components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/AwesomeBarFeature.kt
@@ -8,7 +8,6 @@ import android.content.Context
 import android.content.res.Resources
 import android.graphics.Bitmap
 import android.view.View
-import kotlinx.coroutines.Deferred
 import mozilla.components.browser.icons.BrowserIcons
 import mozilla.components.browser.search.SearchEngine
 import mozilla.components.browser.search.SearchEngineManager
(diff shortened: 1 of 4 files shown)


components/feature/awesomebar/src/main/java/mozilla/components/feature/awesomebar/AwesomeBarFeature.kt
```diff
@@ -149,13 +148,13 @@ class AwesomeBarFeature(
      * @param showDescription whether or not to add the search engine name as description.
      */
     fun addSearchActionProvider(
-        searchEngine: Deferred<SearchEngine>,
+        searchEngineGeetter: suspend () -> SearchEngine,
```
hakkikaancaliskan, line 151: Getter*

samples/browser/src/main/java/org/mozilla/samples/browser/BrowserFragment.kt
```diff
@@ -70,7 +68,7 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
                 components.tabsUseCases.selectTab
             )
             .addSearchActionProvider(
-                searchEngine = GlobalScope.async {
+                searchEngineGeetter = suspend {
```
csadilek, line 71: nit: typo in parameter name

pocmo:
bors r=csadilek

pocmo:
bors r=csadilek

PR #7751, opened 2020-07-16
Closes #7741: Move Session.Source to browser state

### LINKED ISSUE #7741: Move Session.Source to browser state
From the store/state we currently can't tell the "source" of a tab. Let's try to move the type over and keep in sync with browser-session.

csadilek:
Luckily `Session.Source` isn't mutable so this is mostly mechanical. Moving the type to browser-state and making sure the source is set when syncing from `SessionManager`, plus setting defaults.

Fenix PR: https://github.com/mozilla-mobile/fenix/pull/12647


csadilek:
bors r=pocmo

Issue #7753, opened 2020-07-17
White page rendered after switching tabs quickly


pocmo:
Fenix issue: https://github.com/mozilla-mobile/fenix/issues/12593

Emily investigated and it looks like this is a regression from migrating `SessionFeature` to use `BrowserStore`.

pocmo:
In the debugger I can see that before that happens we call `release()` on the GeckoView once. I do not know why yet and it shouldn't be a problem. But that is definitely a new behavior that we introduced (releasing if the GeckoSession is not assigned to a tab anymore).

```
SKDBG: Render 208541118
SKDBG: Render 234939922
SKDBG: Render 208541118
SKDBG: Release
SKDBG: Render 234939922
```

pocmo:
Open questions:

* Where does the release come from?
* Why is it a problem (for GeckoView?) if we release and then render? That seems like a normal use case except that it happens quite fast here.

pocmo:
Adding dispatched store actions to the store it looks like this:

```
[..]
SKDBG: dispatch SelectTabAction(tabId=93d0a2d9-ff06-4430-b772-76cc60513f36, timeSelected=1594990146272)
SKDBG: Render 89248200
[..]
SKDBG: dispatch UnlinkEngineSessionAction(sessionId=93d0a2d9-ff06-4430-b772-76cc60513f36)
SKDBG: dispatch RemoveTabAction(tabId=93d0a2d9-ff06-4430-b772-76cc60513f36, selectParentIfExists=true)
SKDBG: Release
SKDBG: Render 35157153
[..]
```

We switch to the new selected tab, then we remove it. The store first sees an unlink action. When the presenter sees that state then the selected tab has no engine session anymore, so we release. Then the store sees the remove action and will determine a new selected tab and render that.

This is different from the previous `SessionManager` state because previously unlink+remove would happen inside the SessionManager as one action and the presenter would only see the new selected session.

However the new behavior is not really wrong. Once we are no longer using `SessionManager` it may go away since we probably can unlink when removing the tab (like SessionManager does).

I wonder what a good intermediate solution here is though.

Additionally I want to try if that (release+render fast after each other) trips up GeckoView in general, then this may be a bug worth filing too.

pocmo:
Not releasing the session from the view definitely does prevent this. But I still would like to understand the underlying cause and whether we need another fix somewhere.

I tried to replicate this in sample browser by releasing before rendering but that does not seem to cause it.

[...]

PR #7759, opened 2020-07-17
Issue #7753: EngineViewPresenter: Temporarily disable releasing of engine session.

pocmo:
That was a new behavior that was introduced with the migration to `BrowserStore/State` and, surprise!, it can cause a white page of death - even though the order of events look correct and make sense. Anyhow, more in #7753. I think we should temporarily disable this and continue to investigate the actual cause of the white page issue.


pocmo:
Oh, yeah, the tests need to change too of course.

pocmo:
bors r=csadilek

PR #7776, opened 2020-07-20
RFC: Adding a concept-base component for basic interfaces needed by multiple components.

pocmo:
More RFCs for the RFC gods. :)


pocmo:
Yeah, agree, "base" should only contain foundational pieces that are needed almost everywhere. For `Observable` and `ObserverRegistry` ... once we have migrated to `BrowserStore` completely, I would like to see if we can get rid of them eventually too.

pocmo:
bors r=csadilek,NotWoods

PR #7778, opened 2020-07-20
Update Glean to v31.4.1

travis79:


travis79:
> r+ - please only merge after tests are green.

All Green!

PR #7801, opened 2020-07-21
Closes #7782: Fix crashes in addon installation dialog

### LINKED ISSUE #7782: Crash in [@ org.mozilla.geckoview.GeckoResult$UncaughtException: at org.mozilla.geckoview.GeckoResult.dispatchLocked(GeckoResult.java)]
https://crash-stats.mozilla.org/signature/?product=Fenix&signature=org.mozilla.geckoview.GeckoResult%24UncaughtException%3A%20at%20org.mozilla.geckoview.GeckoResult.dispatchLocked%28GeckoResult.java%29

bz-1644999
```
org.mozilla.geckoview.GeckoResult$UncaughtException: java.lang.IllegalStateException: Can not perform this action after onSaveInstanceState
	at org.mozilla.geckoview.GeckoResult.dispatchLocked(GeckoResult.java:4)
	at org.mozilla.geckoview.GeckoResult.completeExceptionally(GeckoResult.java:4)
	at org.mozilla.geckoview.GeckoResult.lambda$thenInternal$2$GeckoResult(GeckoResult.java:10)
	at org.mozilla.geckoview.-$$Lambda$GeckoResult$gwCgOUK_EYQn2g6GolfZvo6A_WE.run(lambda)
	at org.mozilla.geckoview.GeckoResult.lambda$dispatchLocked$3(GeckoResult.java:2)
	at org.mozilla.geckoview.-$$Lambda$GeckoResult$PehjG2jgyDa_p37vZrKrSi2I94s.run(lambda)
	at android.os.Handler.handleCallback(Handler.java:815)
	at android.os.Handler.dispatchMessage(Handler.java:104)
	at android.os.Looper.loop(Looper.java:207)
	at android.app.ActivityThread.main(ActivityThread.java:5740)
	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:905)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:766)
Caused by: java.lang.IllegalStateException: Can not perform this action after onSaveInstanceState
	at androidx.fragment.app.FragmentManager.enqueueAction(FragmentManager.java:6)
	at androidx.fragment.app.BackStackRecord.commitInternal(BackStackRecord.java:14)
	at androidx.fragment.app.BackStackRecord.commit(BackStackRecord.java:1)
	at androidx.fragment.app.DialogFragment.show(DialogFragment.java:5)
	at -$$LambdaGroup$ks$UtaWBn0vzFfZ64mP0sa7JCFLt3Q.invoke(com.android.tools.r8.jetbrains.kotlin-style lambda group:83)
	at -$$LambdaGroup$ks$_HIPVZOvbmOxwps1M61uurZH_lk.invoke(com.android.tools.r8.jetbrains.kotlin-style lambda group:15)
	at mozilla.components.browser.engine.gecko.GeckoEngine$installWebExtension$$inlined$apply$lambda$3.onValue(GeckoEngine.kt:14)
	at org.mozilla.geckoview.GeckoResult.lambda$thenInternal$2$GeckoResult(GeckoResult.java:2)
```

csadilek:
This fixes the following problems with `AddonInstallationDialogFragment`:

- It didn't have a default constructor. So any attempt by Android to restore it caused a crash e.g. when rotating the device.
- Since we're displaying the fragment as a result of an async operation it can happen that the "parent" fragment is still attached, our check of `runIfFragmentAttached` succeeds, but the activity is in the process of being destroyed. Then we run into the crash described in #7782. I've added the workaround here (seems pretty common based on my research) with a description of why it seems safe to me, but thoughts more than welcome. :)
- The icon fetch job was returned but never cancelled. It's also not need to fetch the icon again when rotating.

### DIFF
--- a/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonInstallationDialogFragment.kt
+++ b/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonInstallationDialogFragment.kt
@@ -6,6 +6,7 @@ package mozilla.components.feature.addons.ui
 
 import android.annotation.SuppressLint
 import android.app.Dialog
+import android.graphics.Bitmap
 import android.graphics.Color
 import android.graphics.drawable.BitmapDrawable
 import android.graphics.drawable.ColorDrawable
@@ -25,6 +26,7 @@ import androidx.annotation.VisibleForTesting
 import androidx.appcompat.app.AppCompatDialogFragment
 import androidx.appcompat.widget.AppCompatCheckBox
 import androidx.core.content.ContextCompat
+import androidx.fragment.app.FragmentManager
 import kotlinx.android.synthetic.main.mozac_feature_addons_fragment_dialog_addon_installed.view.*
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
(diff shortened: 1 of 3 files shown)


components/feature/addons/src/test/java/mozilla/components/feature/addons/ui/AddonInstallationDialogFragmentTest.kt
```diff
@@ -165,6 +172,32 @@ class AddonInstallationDialogFragmentTest {
...
+        val addon = mock<Addon>()
+        val mockedCollectionProvider = mock<AddonCollectionProvider>()
+        val fragment = createAddonInstallationDialogFragment(addon, mockedCollectionProvider)
+
+        val fragmentManager = mock<FragmentManager>()
+        val fragmentTransaction = mock<FragmentTransaction>()
+        `when`(fragmentManager.beginTransaction()).thenReturn(fragmentTransaction)
+
+        fragment.show(fragmentManager, "test")
+        verify(fragmentTransaction).commitAllowingStateLoss()
```
Amejia481, line 186: Not sure if this is the right place, but should we verify that `addonCollectionProvider` is assigned and not null?

components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonInstallationDialogFragment.kt
```diff
@@ -206,6 +225,19 @@ class AddonInstallationDialogFragment(
         }
     }
 
+    override fun show(manager: FragmentManager, tag: String?) {
+        // This dialog is shown as a result of an async operation (installing
+        // an add-on). Once installation succeeds, the activity may already be
+        // in the process of being destroyed. Since the dialog doesn't have any
+        // state we need to keep, and since it's also fine to not display the dialog
+        // at all in case  the user navigates away, we can simply use
```
Amejia481, line 233: Nit: double space `// at all in case  ..the user`

components/feature/addons/src/test/java/mozilla/components/feature/addons/ui/AddonInstallationDialogFragmentTest.kt
csadilek: I don't think this test is good for covering that, but I added some assertions to the `build dialog` test to cover it now.

components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonInstallationDialogFragment.kt
csadilek: 👍 fixed.

csadilek:
bors r=Amejia481

pocmo:
This is already in 52.0. Removing labels.

Issue #7788, opened 2020-07-21
Cannot upload immediately after granting storage permission


Mugurell:
Reported on Fenix - https://github.com/mozilla-mobile/fenix/issues/11150

## Steps to reproduce
- fresh install so that no permission is granted for the app
- go to a site that allows file uploads (ex: https://viljamis.com/filetest/)
- press the picker button (usually titled "Browse")
- the permission dialog appears, press "Allow" to grant storage permission
- the file picker opens, pick any file

When returning to the website
### Expected behavior
The picked file appears as selected and ready to be upload

### Actual behavior
No file is ready to be uploaded, Browser must be pressed again.

### Device information
Seems like an old issue happening on all devices.

sv-ohorvath:
This was verified in https://github.com/mozilla-mobile/fenix/issues/11150 on Nightly and now on RC 79.0.5 & Beta 80.0.1. I'll close the issue.

PR #7804, opened 2020-07-21
For #7788 - Cache request to be reused after storage permission is granted

Mugurell:
In the previous flow following storage permission being granted the prompt that
initiated the file picking would be consumed.
Normally, after actually picking a file we would make sure the prompt that
initiated this would be dismissed and following that we would actually do the
upload.
Because this prompt is already consumed following permission being granted the
upload wouldn't succeed and the user would have to redo that upload action.

As such, we need to persist the original prompt even after permission being
granted.
---

### Pull Request checklist

- [x] **Changelog**: This PR does not need one a changelog entry.
- [x] **Accessibility**: The code in this PR does not include any user accessibility changes.

### After merge

### DIFF
--- a/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/file/FilePicker.kt
+++ b/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/file/FilePicker.kt
@@ -45,6 +45,12 @@ internal class FilePicker(
      */
     private var captureUri: Uri? = null
 
+    /**
+     * Cache of the current request to be used after permission is granted.
+     */
+    @VisibleForTesting
+    internal var currentRequest: PromptRequest? = null
+
     @Suppress("ComplexMethod")
     fun handleFileRequest(promptRequest: File, requestPermissions: Boolean = true) {
         // Track which permissions are needed.
@@ -89,7 +95,7 @@ internal class FilePicker(
 
             container.startActivityForResult(chooser, FILE_PICKER_ACTIVITY_REQUEST_CODE)
         } else {
-            onNeedToRequestPermissions(neededPermissions.toTypedArray())
+            askAndroidPermissionsForRequest(neededPermissions, promptRequest)
         }
     }
 
@@ -128,6 +134,7 @@ internal class FilePicker(
         } else {
             onPermissionsDenied()
         }
+        currentRequest = null
     }
 
     /**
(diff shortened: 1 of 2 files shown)


components/feature/prompts/src/main/java/mozilla/components/feature/prompts/file/FilePicker.kt
```diff
@@ -140,9 +146,7 @@ internal class FilePicker(
      */
     @VisibleForTesting(otherwise = PRIVATE)
     internal fun onPermissionsGranted() {
-        store.consumePromptFrom(sessionId) { promptRequest ->
-            handleFileRequest(promptRequest as File, requestPermissions = false)
-        }
+        handleFileRequest(currentRequest as File, requestPermissions = false)
```
csadilek, line 149: Ah yes, that makes perfect sense. We were consuming the prompt request here and then it was no longer in the store once we hit `onActivityResult`. Would you mind adding a comment here to explain why we're consuming in `onPermissionDenied`, but not in `onPermissionGranted`? (Because then we will consume in `onActivityResult`).

Amejia481: Should `currentRequest` be null able? As it will be there after we handle the request (`handleFileRequest`) should it be null after that?

Mugurell: > Should `currentRequest` be nullable? At some point we should clean it up as it will persist there until a new request arrive

I used `currentRequest` as nullable and didn't check it for null just to be able to catch early any situations in which we would not cache the prompt as expected.. though there should be any.
To clean it up, just to reduce the memory footprint, I'll do it in `handleFileRequest`.

csadilek: nit: If we're setting it to `null` now, should we also set it to null if permissions were denied? We could do that in line 137 after either `onPermissionGranted` or `onPermissionDenied` have been called.

Mugurell: Thanks!

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/file/FilePicker.kt
```diff
@@ -45,6 +45,12 @@ internal class FilePicker(
      */
     private var captureUri: Uri? = null
 
+    /**
+     * Cache of the current request to be used after permission is granted.
+     */
+    @VisibleForTesting
+    var currentRequest: PromptRequest? = null
```
Amejia481, line 52: nit: Let's make this one `internal` :)

Mugurell:
bors r=Amejia481,csadilek

PR #7807, opened 2020-07-21
Closes #7797: Strip common subdomain before checking if the new URL is in the same domain

### LINKED ISSUE #7797: Strip common subdomain before checking if the new URL is in the same domain
Currently the following domains are considered different:
example.com, www.example.com, m.example.com.

However, for app links purposes, these should be considered the same domain.  This will prevent unwanted redirect to 3rd party apps when redirecting between same domains.

rocketsroger:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -70,6 +70,7 @@ class GeckoEngineSession(
 
     internal lateinit var geckoSession: GeckoSession
     internal var currentUrl: String? = null
+    internal var lastLoadRequestUri: String? = null
     internal var scrollY: Int = 0
 
     internal var job: Job = Job()
@@ -454,7 +455,7 @@ class GeckoEngineSession(
             isSubframeRequest: Boolean
         ): InterceptionResponse? {
             val interceptor = settings.requestInterceptor
-            return if (
+            val interceptionResponse = if (
                 interceptor != null && (!request.isDirectNavigation || interceptor.interceptsAppInitiatedRequests())
             ) {
                 val engineSession = this@GeckoEngineSession
@@ -462,6 +463,7 @@ class GeckoEngineSession(
                 interceptor.onLoadRequest(
                     engineSession,
                     request.uri,
+                    lastLoadRequestUri,
                     request.hasUserGesture,
                     isSameDomain,
                     request.isRedirect,
@@ -481,6 +483,9 @@ class GeckoEngineSession(
             } else {
                 null
             }
+
+            lastLoadRequestUri = request.uri
+            return interceptionResponse
         }
     }
 
--- a/components/browser/engine-gecko-beta/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineSessionTest.kt
+++ b/components/browser/engine-gecko-beta/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineSessionTest.kt
@@ -1267,6 +1267,7 @@ class GeckoEngineSessionTest {
             override fun onLoadRequest(
                 engineSession: EngineSession,
                 uri: String,
+                lastUri: String?,
                 hasUserGesture: Boolean,
                 isSameDomain: Boolean,
                 isRedirect: Boolean,
@@ -1298,6 +1299,7 @@ class GeckoEngineSessionTest {
             override fun onLoadRequest(
                 engineSession: EngineSession,
                 uri: String,
+                lastUri: String?,
                 hasUserGesture: Boolean,
                 isSameDomain: Boolean,
                 isRedirect: Boolean,
@@ -1334,6 +1336,7 @@ class GeckoEngineSessionTest {
             override fun onLoadRequest(
                 engineSession: EngineSession,
                 uri: String,
+                lastUri: String?,
                 hasUserGesture: Boolean,
                 isSameDomain: Boolean,
                 isRedirect: Boolean,
@@ -1380,6 +1383,7 @@ class GeckoEngineSessionTest {
             override fun onLoadRequest(
                 engineSession: EngineSession,
                 uri: String,
+                lastUri: String?,
                 hasUserGesture: Boolean,
                 isSameDomain: Boolean,
                 isRedirect: Boolean,
@@ -2048,6 +2052,7 @@ class GeckoEngineSessionTest {
             override fun onLoadRequest(
                 engineSession: EngineSession,
                 uri: String,
+                lastUri: String?,
                 hasUserGesture: Boolean,
                 isSameDomain: Boolean,
                 isRedirect: Boolean,
@@ -2085,6 +2090,43 @@ class GeckoEngineSessionTest {
     }
 
     @Test
+    fun `onLoadRequest keep track of the last onLoadRequest uri correctly`() {
+        val engineSession = GeckoEngineSession(mock(),
+            geckoSessionProvider = geckoSessionProvider)
+
+        captureDelegates()
+
+        var observedUrl: String? = null
+
+        engineSession.settings.requestInterceptor = object : RequestInterceptor {
+            override fun interceptsAppInitiatedRequests() = true
+
+            override fun onLoadRequest(
+                engineSession: EngineSession,
+                uri: String,
+                lastUri: String?,
+                hasUserGesture: Boolean,
+                isSameDomain: Boolean,
+                isRedirect: Boolean,
+                isDirectNavigation: Boolean,
+                isSubframeRequest: Boolean
+            ): RequestInterceptor.InterceptionResponse? {
+                observedUrl = lastUri
+                return null
+            }
+        }
+
+        navigationDelegate.value.onLoadRequest(mock(), mockLoadRequest("test1"))
+        assertEquals(null, observedUrl)
+
+        navigationDelegate.value.onLoadRequest(mock(), mockLoadRequest("test2"))
+        assertEquals("test1", observedUrl)
+
+        navigationDelegate.value.onLoadRequest(mock(), mockLoadRequest("test3"))
+        assertEquals("test2", observedUrl)
+    }
+
+    @Test
     fun `onSubframeLoadRequest will notify onLaunchIntent observers if request was intercepted with app intent`() {
         val engineSession = GeckoEngineSession(mock(),
             geckoSessionProvider = geckoSessionProvider)
@@ -2101,6 +2143,7 @@ class GeckoEngineSessionTest {
             override fun onLoadRequest(
                 engineSession: EngineSession,
                 uri: String,
+                lastUri: String?,
                 hasUserGesture: Boolean,
                 isSameDomain: Boolean,
                 isRedirect: Boolean,
@@ -2159,6 +2202,7 @@ class GeckoEngineSessionTest {
             override fun onLoadRequest(
                 engineSession: EngineSession,
                 uri: String,
+                lastUri: String?,
                 hasUserGesture: Boolean,
                 isSameDomain: Boolean,
                 isRedirect: Boolean,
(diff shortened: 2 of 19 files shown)


components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -458,7 +460,9 @@ class GeckoEngineSession(
                 interceptor != null && (!request.isDirectNavigation || interceptor.interceptsAppInitiatedRequests())
             ) {
                 val engineSession = this@GeckoEngineSession
-                val isSameDomain = engineSession.currentUrl?.tryGetHostFromUrl() == request.uri.tryGetHostFromUrl()
+                val isSameDomain = lastOnLoadRequestUrl?.tryGetHostFromUrl()?.toDisplayUrl() ==
+                    request.uri.tryGetHostFromUrl().toDisplayUrl()
+                lastOnLoadRequestUrl = request.uri
```
csadilek, line 465: Your setting this in both branches of the if statement.

csadilek, line 465: Why do we need to introduce new state here? I assume `currentUrl` is not set yet because `onLocationChange` wasn't called yet? Any chance we can prevent adding this?

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -458,7 +460,9 @@ class GeckoEngineSession(
                 interceptor != null && (!request.isDirectNavigation || interceptor.interceptsAppInitiatedRequests())
             ) {
                 val engineSession = this@GeckoEngineSession
-                val isSameDomain = engineSession.currentUrl?.tryGetHostFromUrl() == request.uri.tryGetHostFromUrl()
+                val isSameDomain = lastOnLoadRequestUrl?.tryGetHostFromUrl()?.toDisplayUrl() ==
```
csadilek, line 463: Hm, I don't think we should be using `toDisplayUrl` in the engine. This is an extension function to format the URL for the toolbar, it doesn't seem right to re-use it this way. Let's refactor and add a utility/extension function to strip common subdomains and then compare the domains?

components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlin/String.kt
```diff
@@ -49,6 +49,8 @@ fun String.isEmail() = re.emailish.matches(this)
 
 fun String.isGeoLocation() = re.geoish.matches(this)
 
+fun String.toDisplayUrl() = URLStringUtils.toDisplayUrl(this)
```
csadilek, line 52: See comment above. This doesn't seem right. Wdyt?

config/detekt-baseline.xml
```diff
@@ -478,6 +478,7 @@
     <ID>UndocumentedPublicFunction:Store.kt$Store.Subscription.Binding$fun unbind()</ID>
     <ID>UndocumentedPublicFunction:String.kt$fun String.isEmail()</ID>
     <ID>UndocumentedPublicFunction:String.kt$fun String.isGeoLocation()</ID>
+    <ID>UndocumentedPublicFunction:String.kt$fun String.toDisplayUrl()</ID>
```
csadilek, line 481: Let's add Kdocs to new public functions :).

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
csadilek: My concern with this is that the last `loadRequest` could have failed and then our check wouldn't be accurate anymore. Right now we compare the `currentUrl` with the requested on which is accurate.

csadilek: We talked it through and this is in fact the problem.  Let's double check with GV that there's now way for us to get to the `triggerUri` in this case or maybe file a ticket?

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
csadilek: Let's keep this check as it was and handle the special cases in the `AppLinksInteceptor`. So, we'd pass the previous URL down as well. This way the `isSameDomain` flag isn't affected by the special cases needed for app links.

rocketsroger: Understood, I'll move it.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
rocketsroger: I'll simplify this. Thanks,

components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlin/String.kt
rocketsroger: Agreed.  I'll move this.

components/browser/engine-gecko-beta/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -70,6 +70,7 @@ class GeckoEngineSession(
 
     internal lateinit var geckoSession: GeckoSession
     internal var currentUrl: String? = null
+    internal var lastOnLoadRequestUri: String? = null
```
csadilek, line 73: nit: maybe just `lastLoadRequestUri`.

components/browser/engine-gecko-beta/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineSessionTest.kt
```diff
@@ -1267,6 +1267,7 @@ class GeckoEngineSessionTest {
             override fun onLoadRequest(
                 engineSession: EngineSession,
                 uri: String,
+                lastOnLoadRequestUri: String?,
```
csadilek, line 1270: We should test that we pass the right value (previous lost request uri).

components/browser/engine-gecko-nightly/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineSessionTest.kt
```diff
@@ -1267,6 +1267,7 @@ class GeckoEngineSessionTest {
             override fun onLoadRequest(
                 engineSession: EngineSession,
                 uri: String,
+                lastOnLoadRequestUri: String?,
```
csadilek, line 1270: Same as above, let's actually assert in the test :)

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -248,6 +249,7 @@ class SystemEngineView @JvmOverloads constructor(
                     interceptor.onLoadRequest(
                         session,
                         request.url.toString(),
+                        lastOnLoadRequestUri,
```
csadilek, line 252: For engine-system, we can just use `session.currentUrl`, right? We don't have the same problem as in engine-gecko.

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
```diff
@@ -77,6 +77,7 @@ class SystemEngineView @JvmOverloads constructor(
 ) : FrameLayout(context, attrs, defStyleAttr), EngineView, View.OnLongClickListener {
     @VisibleForTesting(otherwise = PRIVATE)
     internal var session: SystemEngineSession? = null
+    internal var lastOnLoadRequestUri: String? = null
```
csadilek, line 80: see below, might not be needed in this engine?

components/concept/engine/src/main/java/mozilla/components/concept/engine/request/RequestInterceptor.kt
```diff
@@ -72,6 +72,7 @@ interface RequestInterceptor {
     fun onLoadRequest(
         engineSession: EngineSession,
         uri: String,
+        lastOnLoadRequestUri: String?,
```
csadilek, line 75: Let's give this a better name here and API docs! Maybe just `lastUri` or `previousUri`?

components/feature/app-links/src/main/java/mozilla/components/feature/app/links/AppLinksInterceptor.kt
```diff
@@ -131,4 +135,19 @@ class AppLinksInterceptor(
...
         return null
     }
+
+    // for app links interceptor any domains such as example.com, www.example.com and m.example.com
+    // are considered as the same domain
+    private fun isSameDomain(url1: String?, url2: String?): Boolean {
+        return stripUrlSubDomain(url1?.tryGetHostFromUrl()) == stripUrlSubDomain(url2?.tryGetHostFromUrl())
+    }
+
+    private fun stripUrlSubDomain(url: String?): String? {
```
csadilek, line 145: nit: maybe `stripCommonSubDomains`?

components/concept/engine/src/main/java/mozilla/components/concept/engine/request/RequestInterceptor.kt
csadilek: While we're at it the API docs could still need some cleanup. :).

nit: `uri The the` (double the)
nit: ` hasUserGesture` If the request *is* triggered by...
nit: `isRedirect` If the request is due to *a* redirect then true, else false....

components/browser/engine-gecko-beta/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineSessionTest.kt
rocketsroger: Agreed.  I'll add the test.

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
rocketsroger: I see.  I'll double check.   Thanks,

components/concept/engine/src/main/java/mozilla/components/concept/engine/request/RequestInterceptor.kt
rocketsroger: Sure, I'll update with `lastUri`

components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineView.kt
rocketsroger: Agreed, I'll fix.  Thanks,

rocketsroger:
> Looks good to me now! 🚢
> 
> Do you already have a Fenix PR up for the breaking API change in the interceptor? (We should also point that out in the changelog.)

Yes, I should add this to the changelog as well.  I'll update

rocketsroger:
Related https://github.com/mozilla-mobile/fenix/pull/13006

rocketsroger:
bors r=csadilek

PR #7812, opened 2020-07-21
Closes #7811: Handle unsuccessful load when reloading

### LINKED ISSUE #7811: GeckoEngine: Unsuccessfully loaded pages cannot be reloaded
If a page load is interrupted or stopped by the user before `onLocationChange`, it currently cannot be reloaded. An attempt to reload will actually reload `about:blank` yielding a white page, see https://github.com/mozilla-mobile/fenix/issues/11501.

Suggestion from GV is that we handle this in A-C, by calling `loadUri` again if a reload is triggered for a page that wasn't fully loaded: https://github.com/mozilla-mobile/fenix/issues/11501#issuecomment-659004864

csadilek:
Based on snorp's suggestion: https://github.com/mozilla-mobile/fenix/issues/11501#issuecomment-659004864

Easy way to reproduce is to configure an emulator to use a slow network and open a page via an intent (e.g. in Sample Browser). Stop the page load quickly before `onLocationChange` is called. Reload will not work as we keep reloading `about:blank` even though the correct URL is being displayed in the toolbar.

This does fix the problem and makes interrupted pages "reloadable". It's not pretty though.

@pocmo wdyt? Let's discuss before adding tests and cleaning this up :).

### DIFF
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -7,6 +7,7 @@ package mozilla.components.browser.engine.gecko
 import android.annotation.SuppressLint
 import android.os.Build
 import android.view.WindowManager
+import androidx.annotation.VisibleForTesting
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
@@ -101,6 +102,24 @@ class GeckoEngineSession(
     }
 
     /**
+     * Represents a request to load a [url].
+     *
+     * @param url the url to load.
+     * @param parent the parent (referring) [EngineSession] i.e. the session that
+     * triggered creating this one.
+     * @param flags the [LoadUrlFlags] to use when loading the provided url.
+     * @param additionalHeaders the extra headers to use when loading the provided url.
+     **/
+    data class LoadRequest(
+        val url: String,
+        val parent: EngineSession?,
+        val flags: LoadUrlFlags,
+        val additionalHeaders: Map<String, String>?
+    )
+    @VisibleForTesting
+    internal var initialLoadRequest: LoadRequest? = null
+
+    /**
      * See [EngineSession.loadUrl]
      */
     override fun loadUrl(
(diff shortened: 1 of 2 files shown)


components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -352,6 +376,8 @@ class GeckoEngineSession(
 
             currentUrl = url
             initialLoad = false
+            pendingLoadRequest = null
```
csadilek, line 379: `onLocationChange` seems to work best here. `onPageStop` (with success==true) would be another option but that didn't work reliably for me when testing.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -133,7 +152,12 @@ class GeckoEngineSession(
      * See [EngineSession.reload]
      */
     override fun reload(flags: LoadUrlFlags) {
-        geckoSession.reload(flags.value)
+        val loadRequest = pendingLoadRequest
+        if (loadRequest != null) {
+            loadUrl(loadRequest.url, loadRequest.parent, loadRequest.flags, loadRequest.additionalHeaders)
```
csadilek, line 157: Another question here is what would happen with a session that's persisted and restored in that state? Currently would be restored as a white tab and also can't be reloaded. Maybe handle that separately? Or should we persist a flag somehow and load again? 🤔

pocmo, line 157: I wonder if we should do this only for the first load? If I am on site A and load site B, stop the load, remain on site A and press reload then this would load site B now? Maybe this is only a problem for the `about:blank` load that no one wanted?

csadilek: Thinking through options here:
- We could not persist sessions that never had a completed load (never completed in either error of success)
- We could try and persist/restore this additional state somehow
- We could punt on that for now :)

csadilek: @pocmo ah yes, definitely! I pushed an update that handles this now and adds comments and tests.

csadilek: Looks maybe option 1 is best? File a follow-up to not persists tabs that never had a successful load?

csadilek:
OK, this is ready for review :).

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
csadilek: https://github.com/mozilla-mobile/android-components/issues/7859

csadilek:
bors r=pocmo

PR #7830, opened 2020-07-23
Closes #7829: Startup crashes trigger crash loop

### LINKED ISSUE #7829: Startup crashes trigger crash loop
Currently any startup crash causes a crash loop.

This seems to be a regression of: https://github.com/mozilla-mobile/android-components/pull/7206/

csadilek:


csadilek:
bors r=pocmo

Issue #7839, opened 2020-07-24
Overwriting files with same name when downloading to some folders


BranescuMihai:
This is related to a Fenix ticket: https://github.com/mozilla-mobile/fenix/issues/12159

We override files from subfolders when we download them to the main folder. The problem appears on Android 10, because the MediaStore queried subfolders as well when searching for occurrences, even though it should look only in the destination directory.

Mugurell:
Based on @Amejia481's PR comment about the sample-browser being unaffected I investigated a bit what the differences between Fenix and sample-browser are.

Found the real issue to be related to the use of `MediaStore.Downloads` [when running on >=AndroidQ](https://github.com/mozilla-mobile/android-components/blob/59c509410e9ceef960860776db6375a05919ff7d/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt#L666).
As part of the "scoped storage" changes this would only give access to the files downloaded by the current app.
This is why the issue is not reproducing on sample-browser.

But Fenix uses `requestLegacyExternalStorage="true"` (and I think other clients of AC would also) which means the app would get access to all files, including the ones not owned by it.

An easy fix would be to also check for the legacy mode,
so the
>     if (SDK_INT >= Build.VERSION_CODES.Q) {
>        useFileStreamScopedStorage()
>     } else {
>        ...
>     }
would become
>     if (SDK_INT >= Build.VERSION_CODES.Q && !Environment.isExternalStorageLegacy()) {
>         ...
>     } else {
>         useFileStreamLegacy()
>     }

Amejia481:
Awesome, thanks @Mugurell for the investigation, that make sense! We can use this alternative and do some testing :)

PR #7842, opened 2020-07-24
For #7839 - Stop overriding files with same name from subfolders when downloading

BranescuMihai:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
+++ b/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
@@ -4,6 +4,7 @@
 
 package mozilla.components.feature.downloads
 
+import android.annotation.SuppressLint
 import android.annotation.TargetApi
 import android.app.DownloadManager.ACTION_DOWNLOAD_COMPLETE
 import android.app.DownloadManager.EXTRA_DOWNLOAD_ID
@@ -24,8 +25,9 @@ import android.os.Environment
 import android.os.IBinder
 import android.os.ParcelFileDescriptor
 import android.provider.MediaStore
-import android.widget.Toast
 import android.webkit.MimeTypeMap
+import android.widget.Toast
+import androidx.annotation.RequiresApi
 import androidx.annotation.VisibleForTesting
 import androidx.core.app.NotificationManagerCompat
 import androidx.core.content.FileProvider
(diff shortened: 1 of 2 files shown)


Amejia481:
Thanks for the patch @BranescuMihai  :)

I was testing locally but on Sample browser I wasn't able to reproduce https://github.com/mozilla-mobile/fenix/issues/12159 but in Fenix I can, that means there's something that we are doing differently on Fenix that is causing this behaviour before going on with this patch let's try to understand why this is happening on Fenix only, but both are using the same stack for downloading.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -670,7 +670,7 @@ abstract class AbstractFetchDownloadService : Service() {
         val downloadWithUniqueFileName = makeUniqueFileNameIfNecessary(download, append)
         updateDownloadState(downloadWithUniqueFileName)
 
-        if (SDK_INT >= Build.VERSION_CODES.Q) {
+        if (SDK_INT >= Build.VERSION_CODES.Q && !Environment.isExternalStorageLegacy()) {
```
Amejia481, line 673: Let's add tests to verify this behaviour :)

Amejia481:
@BranescuMihai I noticed some users commented on the change of behaviour https://github.com/mozilla-mobile/fenix/issues/12159#issuecomment-667684656 after this change in Fenix https://github.com/mozilla-mobile/fenix/pull/12823 does this patch fix the comments ?

BranescuMihai:
> @BranescuMihai I noticed some users commented on the change of behaviour [mozilla-mobile/fenix#12159 (comment)](https://github.com/mozilla-mobile/fenix/issues/12159#issuecomment-667684656) after this change in Fenix [mozilla-mobile/fenix#12823](https://github.com/mozilla-mobile/fenix/pull/12823) does this patch address the comments ?

Yes, the comment in question is mostly a variant of the situation described in the bug, in the sense that the file is being overridden even if it doesn't have the same content, but since we're only checking via name, that is within the bug's usual steps.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
BranescuMihai: @Amejia481 I would, but there are 2 issues:
- We don't use PowerMock, so besides giving the whole `Environment.isExternalStorageLegacy()` as a boolean param, which I don't feel like we should do, I don't know exactly how to mock that
- We need to update Roboelectric to version 4.3.1 to support Android Q which comes with some issues, and from what I see here: https://github.com/robolectric/robolectric/issues/5245 there's a workaround available, but excluding a library in all our AC modules seems not worth it

So I think we'll have to skip testing it for now

Amejia481: That's correct. There is an alternative that we could use, it's to wrap `isExternalStorageLegacy` in a internal function on `AbstractFetchDownloadService` and then `spy` on `AbstractFetchDownloadService` and change the return of the function, for example:

```kotlin
abstract class AbstractFetchDownloadService : Service() {
   @VisibleForTesting
    internal fun isExternalStorageLegacy(): Boolen{
        return Environment.isExternalStorageLegacy()
   }
}


```


```kotlin

    @Test
    fun `example`() = runBlocking {

        val mockStore = mock<BrowserStore>()
        val service = spy(object : AbstractFetchDownloadService() {
            override val httpClient = client
            override val store = mockStore
        })

        doReturn(false).`when`(service).isExternalStorageLegacy()
       
         assertFalse(service.isExternalStorageLegacy())
       
        doReturn(true).`when`(service).isExternalStorageLegacy()

        assertTrue(service.isExternalStorageLegacy())
    }
```

BranescuMihai: @Amejia481 you're right, I added tests, thanks!

I'm usually not a fan of adding wrapper methods just for testing, but here it makes sense, since it balances the abstraction inside the method.

components/feature/downloads/src/test/java/mozilla/components/feature/downloads/AbstractFetchDownloadServiceTest.kt
```diff
@@ -1135,6 +1136,62 @@ class AbstractFetchDownloadServiceTest {
...
     }
 
+    @Test
+    fun `WHEN we download on devices with version higher than Q which use legacy external storage THEN we use legacy file stream`() {
+        val service = spy(object : AbstractFetchDownloadService() {
+            override val httpClient = client
+            override val store = browserStore
+        })
+        val uniqueFile: DownloadState = mock()
+        val qSdkVersion = 29
```
Amejia481, line 1146: Nit: We can change the SDK via [@Config(sdk )](http://robolectric.org/configuring/) :)

PR #7856, opened 2020-07-27
FxA Account Manager refactor

### LINKED ISSUE #7335: Connecting to sync immediately syncs twice
I'm still diagnosing this, but thought I'd get it out there for comments.

STR:
* Start Fenix without an fxa user
* Sign in to FxA
* Observe via logcat that 2 syncs immediately (but sequentially) happen

I see `syncNow:161, WorkManagerSyncDispatcher` called twice in quick succession - the first time, `reason` says FirstSync (which seems reasonable). However, the second time has `reason` as Startup - even though the app has been started long enough to complete the sign in flow.

It smells as though that second sync should not be happening in this case, but I'm yet to work out exactly why it does.

(And unsurprisingly given that "startup" sync, it doesn't reproduce when disconnecting then reconnecting - once it has reproduced once, you need to uninstall and start fresh to reproduce again)

cc @grigoryk

### LINKED ISSUE #7536: [Bug] Crash @kotlin.KotlinNullPointerException: at mozilla.components.service.fxa.manager.FxaAccountManager.postAuthenticated(FxaAccountManager.kt:36)
Filed from https://github.com/mozilla-mobile/fenix/issues/12000

https://crash-stats.mozilla.org/signature/?product=Fenix&signature=kotlin.KotlinNullPointerException%3A%20at%20mozilla.components.service.fxa.manager.FxaAccountManager.postAuthenticated%28FxaAccountManager.kt%3A36%29

```
kotlin.KotlinNullPointerException
	at mozilla.components.service.fxa.manager.FxaAccountManager.postAuthenticated(FxaAccountManager.kt:36)
	at mozilla.components.service.fxa.manager.FxaAccountManager$postAuthenticated$1.invokeSuspend(FxaAccountManager.kt)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:2)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:19)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1115)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:590)
	at java.lang.Thread.run(Thread.java:818)
```

grigoryk:
### State machine changes:
- "profile" states have been removed, `profile` is now handled outside of the state machine
- states and events have been split into two subtypes: external and internal. External states are the `AccountState`s (authenticated, not authenticated, auth problem, etc) and internal states are `InternalState` (beginning auth, completing auth, recovering from auth problems, etc)
- this split should hopefully make navigating the state machine more intuitive
- retry logic and error handling was added for all internal transitions which require networking (Closes #7536)

### account manager changes:
- removed ability to "set sync config" from the public API. we weren't using it (and have no plans to use this), so the added complexity wasn't benefiting anything.

### async/suspend API changes:
account manager, account and device constellation APIs are now `suspend`, which makes both internal code and consuming of these APIs nicer and more flexible. Actual semantics of what's going on didn't change.

### sync manager changes:
introduced a delay to the periodic sync scheduling, so that we can properly manage what goes on during startup (Closes #7335)  

Tests are half-way being updating, so this is still a WIP.

cc @csadilek @rfk @eoger @jonalmeida 

---

### Pull Request checklist

### After merge


components/concept/sync/src/main/java/mozilla/components/concept/sync/OAuthAccount.kt
```diff
@@ -53,16 +37,25 @@ enum class InFlightMigrationState {
...
-     * "Reuse" in-flight migration present. Can retry migration via [OAuthAccount.retryMigrateFromSessionTokenAsync].
+     * "Reuse" in-flight migration present. Can retry migration via [OAuthAccount.retryMigrateFromSessionToken].
      */
     REUSE_SESSION_TOKEN
 }
 
+/**
+ * Data structure describing FxA and Sync credentials necessary to sign-in into an FxA account.
+ */
+data class AuthInfo(
```
rfk, line 53: naming nit: this is a very general name for data that's specific to one very narrow purpose, one could be forgiven for thinking they need an "`AuthInfo`" for all sorts of things. It may be worth giving it a more narrowly-scoped name like `MigratingAccountInfo` or similar.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FirefoxAccount.kt
```diff
@@ -133,21 +133,21 @@ class FirefoxAccount internal constructor(
...
         state: String,
         accessType: AccessType
-    ) = scope.async {
+    ) = withContext(scope.coroutineContext) {
         handleFxaExceptions(logger, "authorizeOAuthCode", { null }) {
             inner.authorizeOAuthCode(clientId, scopes, state, accessType.msg)
         }
     }
 
     override fun getSessionToken(): String? {
```
rfk, line 147: Do any of your consumers actually use `getSessionToken`, or is it exposed mostly for completeness? It doesn't feel like anyone should be calling this method in practice.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FxaDeviceConstellation.kt
```diff
@@ -43,31 +42,43 @@ class FxaDeviceConstellation(
...
-    override fun ensureCapabilitiesAsync(capabilities: Set<DeviceCapability>): Deferred<Boolean> {
-        return scope.async {
-            handleFxaExceptions(logger, "ensuring capabilities") {
-                account.ensureCapabilities(capabilities.map { it.into() }.toSet())
+    override suspend fun finalizeDevice(authType: AuthType, config: DeviceConfig) = withContext(scope.coroutineContext) {
+        val finalizeAction = when (authType) {
+            AuthType.Signin,
+            AuthType.Signup,
+            AuthType.Pairing,
+            AuthType.MigratedCopy -> DeviceFinalizeAction.Initialize
```
rfk, line 56: Hrm, the fact that there are so many ways to spell "did a fresh signin" seems like a tiny bit of a code-smell to me. What things do we need to do that differ between `Signin`, `Signup` and `Pairing`? (I seem to recall this being added initially for metrics purposes?)

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
```diff
@@ -357,38 +307,38 @@ open class FxaAccountManager(
...
-            // Access to syncManager is guarded by `this`.
-            synchronized(this) {
-                if (syncManager == null) {
-                    throw IllegalStateException(
-                        "Sync is not configured. Construct this class with a 'syncConfig' or use 'setSyncConfig'"
-                    )
+    ) = withContext(coroutineContext) {
+        when (val s = state) {
+            // Can't sync while we're still doing stuff.
+            is State.Active -> Unit
```
rfk, line 316: IIUC this means the `syncNow` request is just dropped on the floor. Is it possible that this might cause a bad user experience where they request a "sync now" but we don't honour it because we happened to be doing something else at the time?

grigoryk:
Thanks for taking a look, @rfk! I'm still making some tweaks, but the core of it is there.

> FWIW the PR would be easier to digest with the async->suspend stuff landed separately 😁

Yup, I got a little carried away there... At this point, it'll be quite a task to split the comments into a sensible "story", happy to chat on zoom about any of this.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
grigoryk: That's right, we're dropping it here. FWIW, I think this can't really happen in practice - if we're in any of the "active" states, there isn't really a way for the user to request a sync. We're either starting/completing auth, logging out, migrating, etc - Fenix's UI, at least, doesn't expose "sync" button in any of these scenarios.

I'm not quite sure if having this check is necessary (it shouldn't happen), but left it here for completeness. This may as well produce a caught exception that we submit into Sentry, since it is an illegal event essentially (but, it exists outside of the state machine).

Another alternative is to make "sync" an external event, then it won't be dropped on the floor, and will only be processed when it makes sense.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FxaDeviceConstellation.kt
grigoryk: Another place where we use this is to kick-off sync after an `onAuthenticated` event (`first sync` vs `startup sync`), with a different mapping from device's init/ensure.

Otherwise, this is plumbed around in such detail mostly for telemetry, yeah.

grigoryk: I'll think about perhaps a different internal/external representation of this.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FirefoxAccount.kt
grigoryk: In practice, this is just used internally by the account manager (to get sync going).

In theory, if consumers aren't using the account manager this lets them do something useful with the account object beyond just fetching a profile, but... yeah, not really something we want to encourage.

Next on my plate is to work through the "we have a firefox-shaped oauth account abstraction with just a single implementation" bit, and simplify things on that front... Perhaps we can clean this up as part of that work.

components/concept/sync/src/main/java/mozilla/components/concept/sync/OAuthAccount.kt
grigoryk: Yup, renamed!

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/Config.kt
```diff
@@ -4,24 +4,30 @@
...
  * @property supportedEngines A set of supported sync engines, exposed via [GlobalSyncableStoreProvider].
- * @property syncPeriodInMinutes Optional, how frequently periodic sync should happen. If this is `null`,
- * periodic syncing will be disabled.
+ * @property periodicSyncConfig Optional configuration for running sync periodically.
+ * Periodic sync is disabled if this is `null`.
  */
 data class SyncConfig(
     val supportedEngines: Set<SyncEngine>,
-    val syncPeriodInMinutes: Long? = null
+    val periodicSyncConfig: PeriodicSyncConfig?
```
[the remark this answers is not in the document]
grigoryk: I'm not so sure this is bad, the upside is that it gives consumers a single point of interaction with all of these things.

If you mean something closer to "it's too bad sync manager is managed in one repo, but sync scheduling in another", then yes, I agree that it is unfortunate and something we should look into fixing 👍

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
```diff
@@ -558,357 +490,277 @@ open class FxaAccountManager(
...
+            }
+            is Event.Internal.UnrecoverableErrorMigrating -> {
+                notifyObservers { onFlowError(AuthFlowError.FailedToMigrate) }
+            }
+            else -> Unit
+        }
+        AccountState.IncompleteMigration -> {
+            via as Event.Internal.IncompleteMigration
+            Unit
+            // TODO notifyObservers?
```
csadilek, line 512: Probably worth deciding now? :). Or filing a ticket?

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
```diff
@@ -924,53 +776,71 @@ open class FxaAccountManager(
...
+    }
 
-        // Before any sync workers have a chance to access it, make sure our SyncAuthInfo cache is hot.
-        maybeUpdateSyncAuthInfoCache()
+    sealed class RetryResult<out T> {
+        data class Success<out T>(val value: T) : RetryResult<T>()
+        object Failure : RetryResult<Nothing>()
+    }
+
+    private suspend fun <T> withRetries(retryCount: Int, block: suspend () -> T): RetryResult<T> {
```
csadilek, line 813: Nice! Let's add a unit test for this one though. It could potentially also live outside this module, if we want it later.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/State.kt
```diff
@@ -4,168 +4,133 @@
...
-    override fun toString(): String {
-        // For a better logcat experience.
-        return this.javaClass.simpleName
-    }
-
-    internal object Init : Event()
-
-    object AccountNotFound : Event()
-    object AccountRestored : Event()
+internal enum class InternalState {
```
csadilek, line 18: This class looks really ✨ now. `InternalState` could also be `ProgressState` or similar to indicate that all these states point to auth being in progress, and the internal events would be progress events. Just a thought.

components/service/firefox-accounts/src/test/java/mozilla/components/service/fxa/manager/FxaStateMatrixTest.kt
```diff
@@ -4,163 +4,235 @@
...
+                val ss = State.Idle(state)
+                assertNextStateForStateEventPair(
+                    ss,
+                    event,
+                    ss.next(event)
+                )
+            }
+        }
+
+        // GARBAGE:
```
csadilek, line 105: nit: 🗑️  :)

components/feature/accounts-push/src/main/java/mozilla/components/feature/accounts/push/FxaPushSupportFeature.kt
```diff
@@ -190,9 +195,9 @@ internal class AutoPushObserver(
         // Ignore push messages that do not have data.
         val rawEvent = message ?: return
 
-        accountManager.withConstellation {
-            processRawEventAsync(String(rawEvent))
-        }
+        accountManager.withConstellation { CoroutineScope(Dispatchers.Main).launch {
+            processRawEvent(String(rawEvent))
```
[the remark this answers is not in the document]
csadilek: Looks like we maybe want to make these observer methods `suspend` functions as well, so we don't need to context switch?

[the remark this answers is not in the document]
grigoryk: Yeah, this will run as before in practice (underneath we have a dedicated dispatcher), and marking the interface funs as `suspend` will avoid switching and certainly make the intention here clearer.

components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FirefoxAccountsAuthFeature.kt
```diff
@@ -80,11 +80,13 @@ class FirefoxAccountsAuthFeature(
                     val state = parsedUri.getQueryParameter("state") as String
 
                     // Notify the state machine about our success.
-                    accountManager.finishAuthenticationAsync(FxaAuthData(
-                        authType = authType,
-                        code = code,
-                        state = state
-                    ))
+                    CoroutineScope(Dispatchers.Main).launch {
```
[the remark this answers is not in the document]
grigoryk: Looking at the `MainScope`, and it's the same thing except w/ a `SupervisorJob`. I _think_ that's okay, but other slightly more compact code i don't think we're winning much here.

components/feature/accounts-push/src/main/java/mozilla/components/feature/accounts/push/FxaPushSupportFeature.kt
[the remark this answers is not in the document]
grigoryk: Filed https://github.com/mozilla-mobile/android-components/issues/7899

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/State.kt
grigoryk: Renamed to `ProgressState`, `Event.Progress.*`, `Event.Account.*`, definitely more descriptive names!

grigoryk:
One remaining aspect of this PR that I need to change before this can land is improving how we handle the "offline startup" case. Current version will likely fail to finalize the account after multiple attempts, and will drive the user to the NotAuthenticated state, which isn't really something we'd ship. Existing state machine would just kind of give up at that point, and some things may not work correctly, but at least the user remains signed-in.

Ideally, we shouldn't be concerned with this stuff at the startup at all. Current fxaclient APIs don't make any guarantees around what happens during ensure/init calls, so we assume they talk to the network during a regular startup flow.
It would be nice if this wasn't the case - regular startup shouldn't _require_ any network interactions.

We can "paper over" this at the state machine level, but I'm inclined to first investigate changing the a-s APIs first. cc @rfk

components/feature/accounts-push/src/main/java/mozilla/components/feature/accounts/push/FxaPushSupportFeature.kt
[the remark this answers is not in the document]
jonalmeida: @grigoryk I don't follow why we need to switch to the main dispatcher here. Previously the FxaPushSupportFeature would execute the call on any dispatcher since account manager had it's own dispatcher. Shouldn't that still be the case?

grigoryk:
This should be good to land now. However, since I'm away until mid-September, and due to how various releases are aligning right now, let's land it once I'm back on the 16th.

grigoryk:
@csadilek once you're back from PTO, let's get this landed. Should be good to go.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/Utils.kt
```diff
@@ -70,14 +73,53 @@ suspend fun handleFxaExceptions(logger: Logger, operation: String, block: () ->
...
-        else -> true
+/**
+ * Simplified version of Kotlin's inline class version that can be used as a return value.
+ */
+internal sealed class Result<out T> {
+    data class Success<out T>(val value: T) : Result<T>()
+    object Failure : Result<Nothing>()
+}
+
+internal suspend fun <T> withRetries(logger: Logger, retryCount: Int, block: suspend () -> T): Result<T> {
```
csadilek, line 84: nit: Let's add some KDocs here :).

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/Utils.kt
```diff
@@ -70,14 +73,53 @@ suspend fun handleFxaExceptions(logger: Logger, operation: String, block: () ->
...
+    var attempt = 0
+    var res: T? = null
+    while (attempt < retryCount && (res == null || res == false)) {
+        attempt += 1
+        logger.info("attempt $attempt/$retryCount")
```
csadilek, line 89: nit: Now that this is in Utils, should we prefix the log statement so it's clear that it's from `withRetries`?

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/Utils.kt
```diff
@@ -70,14 +73,53 @@ suspend fun handleFxaExceptions(logger: Logger, operation: String, block: () ->
...
+        res = block()
+    }
+    return if (res == null || res == false) {
+        logger.warn("all attempts failed")
```
csadilek, line 93: nit: see above.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/Utils.kt
```diff
@@ -70,14 +73,53 @@ suspend fun handleFxaExceptions(logger: Logger, operation: String, block: () ->
...
+        Result.Failure
+    } else {
+        Result.Success(res)
+    }
+}
+
+internal suspend fun withServiceRetries(
```
csadilek, line 100: nit: KDocs :)

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/Utils.kt
```diff
@@ -70,14 +73,53 @@ suspend fun handleFxaExceptions(logger: Logger, operation: String, block: () ->
...
+    logger: Logger,
+    retryCount: Int,
+    block: suspend () -> ServiceResult
+): ServiceResult {
+    var attempt = 0
+    do {
+        attempt += 1
+        logger.info("attempt $attempt/$retryCount")
```
csadilek, line 108: nit: see above :).

csadilek (approved):
@grigoryk OK did another pass and also looked at latest commits. Just some nits, but good to land from my perspective!

grigoryk:
@csadilek address your feedback, thanks! Let's land this 👍 

bors r=csadilek

PR #7892, opened 2020-07-28
RFC: Introduce library for reading & writing JSON

NotWoods:
We've had some internal discussion about adding a JSON library, but I'd like to revive it as an RFC as @ekager is currently looking at pulling Gson into Fenix. I'd like to align our JSON implementation between AC and Fenix and simplify our JSON logic by relying on a library.

### DIFF
--- a/docs/rfcs/0005-json-library.md
+++ b/docs/rfcs/0005-json-library.md
@@ -1,0 +1,45 @@
+---
+layout: page
+title: RFC process
+permalink: /rfc/0005-json-library
+---
+
+* Start date: 2020-07-28
+* RFC PR: [#7892](https://github.com/mozilla-mobile/android-components/pull/7892)
+
+## Summary
+
+Introduce a library for reading and writing JSON into Android Components and Fenix.
+
+## Motivation
+
+We currently interact with JSON in many areas of Android Components and Fenix. Currently serialization and deserialization is all done manually with `JSONObject`, which creates more work for developers and potential bugs. Fenix also is interested in using a library and we'd like to avoid shipping multiple JSON libraries to users. Using a JSON library allows us to reduce code that needs to be written, potentially gain some performance wins from a well maintained lib, and avoid bugs in our custom JSON implementations. It would also be easier to build streaming JSON parsers.
+
+Using a standard library would also create some consistency in our JSON API surface. Parsing JSON looks totally different for every API that interacts with it, with no guarantees about error handling or null safety.
+
+Some examples of JSON usage today are:
+- Reading a `WebAppManifest`
+- Parsing API calls in various service components
+- Sending messages to and from extensions
+
+## Guide-level explanation
+
+We'll start with selecting a JSON library to integrate into Android Components and Fenix. Some options we've looked at include:
+- [kotlinx-serialization](https://github.com/Kotlin/kotlinx.serialization): Multi-platform, multi-format, built by JetBrains. Also much newer than other solutions so documentation is relatively lacking.
+- [Gson](https://github.com/google/gson): Long standing library built by Google. Slower than other options.
+- [Moshi](https://github.com/square/moshi): Simpler API, built by Square. Fewer features than Gson but also faster.
+
+kotlinx-serialization feels the most in-line with our design axioms as it extends the Kotlin standard library. We can use it when building new components that use JSON and slowly migrate existing code.
+
+## Drawbacks
+
+* Using a new tool means we have to invest some time in learning how to use it. We're at least familiar with JSONObject as are most people. This "having another thing to learn" thing is especially problematic if Fenix and other apps adopt different technologies to fill the same gap.
+* We want to be conservative adding third-party libraries.
+
+## Rationale and alternatives
+
+We currently use `JSONObject` for our JSON decoding and encoding. This forces duplication of logic (need to functions for both String -> JSON and JSON -> String) and creates potential crashes if we don't catch JSON exceptions properly. A well maintained library allows us to write less code and potentially gain performance wins.
+
+## Prior art
+
+Many other open-source projects are using an JSON library.


ekager:
Nice! I don't have super strong opinions but seems like kotlinx-serialization might make the most sense versus other external libraries because Kotlin devs are already familiar with `@Serializable` and helps us avoid adding "more Google dependencies". ;)

person808:
I was under the impression that Moshi is essentially Gson 3.0 (see [this](https://www.reddit.com/r/androiddev/comments/684flw/why_use_moshi_over_gson/dgx3gpm/) comment from one of the maintainers from 3 years ago). IIRC, Moshi also has good Kotlin support and having used it I can say it was a very pleasant experience.

docs/rfcs/0005-json-library.md
```diff
@@ -0,0 +1,48 @@
...
+- [kotlinx-serialization](https://github.com/Kotlin/kotlinx.serialization): Multi-platform, multi-format, built by JetBrains. Also much newer than other solutions so documentation is relatively lacking.
+- [Gson](https://github.com/google/gson): Long standing library built by Google. Slower than other options.
+- [Moshi](https://github.com/square/moshi): Simpler API, built by Square. Fewer features than Gson but also faster.
+
+Once a library is chosen we can use it when building new components that use JSON and slowly migrate existing code.
+
+## Drawbacks
+
+* Using a new tool means we have to invest some time in learning how to use it. We're at least familiar with JSONObject as are most people. This "having another thing to learn" thing is especially problematic if Fenix and other apps adopt different technologies to fill the same gap.
+* We want to be conservative adding third-party libraries.
```
pocmo, line 35: Our design axioms list why we want to be conservative, maybe we can repeat some of those arguments here as drawbacks: https://mozac.org/contributing/design-axioms, we could list here.

Specifically APK size is something that we should consider and explicitly measure for the options here before making a decision.

docs/rfcs/0005-json-library.md
```diff
@@ -0,0 +1,48 @@
...
+- Reading a `WebAppManifest`
+- Parsing API calls in various service components
+- Sending messages to and from extensions
+
+## Guide-level explanation
+
+We'll start with selecting a JSON library to integrate into Android Components and Fenix. Some options we've looked at include:
+- [kotlinx-serialization](https://github.com/Kotlin/kotlinx.serialization): Multi-platform, multi-format, built by JetBrains. Also much newer than other solutions so documentation is relatively lacking.
+- [Gson](https://github.com/google/gson): Long standing library built by Google. Slower than other options.
+- [Moshi](https://github.com/square/moshi): Simpler API, built by Square. Fewer features than Gson but also faster.
```
pocmo, line 28: I think using Gson or Moshi (as much as I like them) is kind of a violation of our design axioms: We'd essentially start packaging those libraries into every app that uses AC.
`kotlinx.serialization` feels more in line, as an extension of the Kotlin standard library, that we explicitly make use of. The last time I looked at it, it was full of "experimental" annotations and didn't seem ready for production use yet though. Of course the APK size impact is still important for this library.

docs/rfcs/0005-json-library.md
```diff
@@ -0,0 +1,48 @@
...
+Introduce a library for reading and writing JSON into Android Components and Fenix.
+
+## Motivation
+
+We currently interact with JSON in many areas of Android Components and Fenix. Currently serialization and deserialization is all done manually with `JSONObject`, which creates more work for developers and potential bugs. Fenix also is interested in using a library and we'd like to avoid shipping multiple JSON libraries to users. Using a JSON library allows us to reduce code that needs to be written, potentially gain some performance wins from a well maintained lib, and avoid bugs in our custom JSON implementations.
+
+Some examples of JSON usage today are:
+- Reading a `WebAppManifest`
+- Parsing API calls in various service components
+- Sending messages to and from extensions
```
pocmo, line 21: One additional motivation for me is that it would be much easier to build streaming JSON parsers. Using `JSONObject` always means loading everything into a String first and then parsing it completely into a `JSONObject` tree. In some areas we have seen memory problems with that approach. There's `JsonReader` in Android, but no one will voluntarily use that. Using a JSON library it is much easier to parse directly from an `InputStream`.

docs/rfcs/0005-json-library.md
```diff
@@ -0,0 +1,48 @@
...
+We currently use `JSONObject` for our JSON decoding and encoding. This forces duplication of logic (need to functions for both String -> JSON and JSON -> String) and creates potential crashes if we don't catch JSON exceptions properly. A well maintained library allows us to write less code and potentially gain performance wins.
+
+## Prior art
+
+Many other open-source projects are using an JSON library. Some examples are:
+
+* [Rust RFCs - Active RFC List](https://rust-lang.github.io/rfcs/)
+* [Bors: About the Draft RFCs category](https://forum.bors.tech/t/about-the-draft-rfcs-category/291)
+* [seL4: The RFC Process](https://docs.sel4.systems/processes/rfc-process.html)
+* [The TensorFlow RFC process](https://www.tensorflow.org/community/contribute/rfc_process?hl=en)
```
pocmo, line 48: This looks like a copy from the "RFC process" RFC. :)

docs/rfcs/0005-json-library.md
```diff
@@ -0,0 +1,48 @@
...
+* Start date: 2020-07-28
+* RFC PR: TBD
+
+## Summary
+
+Introduce a library for reading and writing JSON into Android Components and Fenix.
+
+## Motivation
+
+We currently interact with JSON in many areas of Android Components and Fenix. Currently serialization and deserialization is all done manually with `JSONObject`, which creates more work for developers and potential bugs. Fenix also is interested in using a library and we'd like to avoid shipping multiple JSON libraries to users. Using a JSON library allows us to reduce code that needs to be written, potentially gain some performance wins from a well maintained lib, and avoid bugs in our custom JSON implementations.
```
pocmo, line 16: > Fenix also is interested in using a library and we'd like to avoid shipping multiple JSON libraries to users.

Note that this is something we would want to allow for all consuming apps. Using no library in AC also allows Fenix to make a choice and not ship multiple libraries. :)

docs/rfcs/0005-json-library.md
```diff
@@ -0,0 +1,48 @@
...
+Once a library is chosen we can use it when building new components that use JSON and slowly migrate existing code.
+
+## Drawbacks
+
+* Using a new tool means we have to invest some time in learning how to use it. We're at least familiar with JSONObject as are most people. This "having another thing to learn" thing is especially problematic if Fenix and other apps adopt different technologies to fill the same gap.
+* We want to be conservative adding third-party libraries.
+
+## Rationale and alternatives
+
+We currently use `JSONObject` for our JSON decoding and encoding. This forces duplication of logic (need to functions for both String -> JSON and JSON -> String) and creates potential crashes if we don't catch JSON exceptions properly. A well maintained library allows us to write less code and potentially gain performance wins.
```
pocmo, line 39: I agree that using a library means much less code and comfort for AC developers. From my experience with GSON it is also possible to have the same kind of crashes if you end up with incompatible Kotlin models and JSON. On that front I wonder how well those libraries can deal with versioned/flexible formats? They all look great in isolated examples - how do they deal with changing formats (forward and backward compatibility)?

NotWoods: kotlinx-serialization is able to handle optional/required fields, and different schemas could be represented by different classes if needed.

docs/rfcs/0005-json-library.md
NotWoods: It seems that the code is stable and production ready, but the API surface may change between versions.

NotWoods:
I've written a lot of JSON heavy code for Web App Manifest and REST APIs, and I've found that using JSONObject can be error-prone. As Sebastian said, we don't support streaming data which would be nice to have when working with REST APIs. API surfaces are inconsistent, and its possible for serialization code to be out of sync with deserialization code.

pocmo:
* kotlinx-serialization released 1.0: https://blog.jetbrains.com/kotlin/2020/10/kotlinx-serialization-1-0-released
* I recently migrated some code to Android's [JsonWriter](https://developer.android.com/reference/android/util/JsonWriter) / [JsonReader](https://developer.android.com/reference/kotlin/android/util/JsonReader) - which wasn't super nice to write, but it meant supporting streams without adding a library.

pocmo:
In the planning meeting we talked about closing the RFC for now and getting back to it once it comes up again.

PR #7897, opened 2020-07-28
Closes #7896: Change failure messages to not rely on Kotlin reflection

### LINKED ISSUE #7896: Some migration exception messages rely on Kotlin reflection
Just saw that we have a few error messages where we attempt to append the class name of the underlying exception using "text {e::class}".

This will be reported as "text (Kotlin reflection is not available)".

csadilek:
Another option would be to use the entire `toString` of the exception, instead of just the class name, but sticking to what we had so we don't expose anything by mistake.

The `Exception.uniqueId` extension function should also be affected by this so changed it here as well.

@grigoryk @pocmo Relatively minor, but maybe we want land this / fix it and cherry-pick on the branch for next round?

### DIFF
--- a/components/support/migration/src/main/java/mozilla/components/support/migration/AddonMigration.kt
+++ b/components/support/migration/src/main/java/mozilla/components/support/migration/AddonMigration.kt
@@ -72,7 +72,7 @@ sealed class AddonMigrationResult {
          */
         internal data class FailedToQueryInstalledAddons(val throwable: Throwable) : Failure() {
             override fun toString(): String {
-                return "Failed to query installed add-ons: ${throwable::class}"
+                return "Failed to query installed add-ons: $throwable"
             }
         }
 
--- a/components/support/migration/src/main/java/mozilla/components/support/migration/FennecFxaMigration.kt
+++ b/components/support/migration/src/main/java/mozilla/components/support/migration/FennecFxaMigration.kt
@@ -95,7 +95,7 @@ sealed class FxaMigrationResult {
          */
         data class CorruptAccountState(val e: JSONException) : Failure() {
             override fun toString(): String {
-                return "Corrupt account state, exception type: ${e::class}"
+                return "Corrupt account state, exception: $e"
             }
         }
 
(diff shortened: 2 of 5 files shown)


components/support/migration/src/main/java/mozilla/components/support/migration/Exception.kt
```diff
@@ -10,5 +10,5 @@ package mozilla.components.support.migration
 fun Exception.uniqueId(): String {
     // In extreme cases, VM is allowed to return an empty stacktrace.
     val topStackFrameString = this.stackTrace.getOrNull(0)?.toString() ?: ""
-    return "${this::class.java.canonicalName} $topStackFrameString"
+    return "${this.javaClass.canonicalName} $topStackFrameString"
```
grigoryk, line 13: i guess this continues to work as designed, right?

csadilek: @grigoryk yeah, it does, but there's also no need to change that here as the original works fine. I will remove.

grigoryk: Does the original not rely on the reflection lib? it _looks_ like it does

csadilek: Yeah, it's deceiving but the original actually works, I think `this::class.java` is equivalent to `this.javaClass`.

csadilek:
Updated this now to use the exception's `toString`, rather than just the class name. This should give us some more information too.

grigoryk:
🚢 🇮🇹 

bors r+

csadilek:
Fixed the tests, but let's take another look tomorrow. Need rest :).

csadilek:
bors r=pocmo,grigoryk

csadilek:
@pocmo Should I cherry-pick this on the release branches or wait / label somehow?

pocmo:
Yeah, that seems not risky to uplift :)

pocmo:
Uplifted to 52.0.

PR #7908, opened 2020-07-29
Closes #7860: Introduce SearchState to BrowserState

### LINKED ISSUE #7860: Create SearchState in browser-state
RFC: https://github.com/mozilla-mobile/android-components/pull/7655

To start moving the search related state to `browser-state` we first need to define how the state should look like.

From the RFC:

`BrowserState` will get a new `search` property with `SearchState` type that lives in `browser-state`:

```Kotlin
data class BrowserState(
   // ...
   val search: SearchState
)
/**
 * Value type that represents the state of available search engines.
 *
 * @property searchEngines List of loaded search engines.
 * @property defaultSearchEngineId ID of the default search engine.
 */
data class SearchState(
  val searchEngines: List<SearchEngine>,
  val defaultSearchEngineId: String
)
```

`SearchState` will contain all `SearchEngine`s (bundled and custom). Additional extension methods will make it easier to select specific subsets or the default `SearchEngine`.

`SearchEngine` will be turned into a data class and moved to `browser-store`. A `type` property will make custom and provided default search engines distinguishable. Other methods like `buildSearchUrl()` will be implemented as extension methods.

psymoon:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/search/SearchEngine.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/search/SearchEngine.kt
@@ -1,0 +1,34 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.state.search
+
+import android.graphics.Bitmap
+
+/**
+ * A data class representing a search engine.
+ *
+ * @property id the ID of this search engine.
+ * @property name the name of this search engine.
+ * @property icon the icon of this search engine.
+ * @property type the type of this search engine.
+ * @property resultUrls the list of the queried suggestions result urls.
+ * @property suggestUrl the search suggestion url.
+ */
+data class SearchEngine(
+    val id: String,
+    val name: String,
+    val icon: Bitmap,
+    val type: Type,
+    val resultUrls: List<String> = emptyList(),
+    val suggestUrl: String? = null
+) {
+    /**
+     * A enum class representing a search engine type.
+     */
+    enum class Type {
+        BUNDLED,
+        CUSTOM
+    }
+}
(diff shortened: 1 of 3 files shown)


components/browser/state/src/main/java/mozilla/components/browser/state/search/SearchEngine.kt
```diff
@@ -0,0 +1,34 @@
...
+ * @property name the name of this search engine.
+ * @property icon the icon of this search engine.
+ * @property type the type of this search engine.
+ * @property resultUrls the list of the queried suggestions result urls.
+ * @property suggestUrl the search suggestion url.
+ */
+data class SearchEngine(
+    val id: String,
+    val name: String,
+    val icon: Bitmap? = null,
```
pocmo, line 22: Seeing that `Bitmap` in `SearchEngine` in `browser-search` is not nullable. Maybe we can get away with not making it nullable here too? 🤔 But not super important right now, I think.

components/browser/state/src/main/java/mozilla/components/browser/state/search/SearchEngine.kt
```diff
@@ -0,0 +1,34 @@
...
+    val type: Type,
+    val resultUrls: List<String>? = null,
```
pocmo, line 24: Let's use `emptyList()` here as default. :)

components/browser/state/src/main/java/mozilla/components/browser/state/state/SearchState.kt
```diff
@@ -0,0 +1,18 @@
...
+import mozilla.components.browser.state.search.SearchEngine
+
+/**
+ * Value type that represents the state of search.
+ *
+ * @property searchEngines The map of search engines (bundled and custom) to their respective IDs.
+ * @property defaultSearchEngineId The ID of default [SearchEngine]
+ */
+data class SearchState(
+    val searchEngines: Map<String, SearchEngine>,
```
pocmo, line 16: nit: I think we should move the `emptyMap()` here instead of passing it in. :)

components/browser/state/src/main/java/mozilla/components/browser/state/search/SearchEngine.kt
psymoon: I'll make it not nullable.

psymoon:
bors r=pocmo

Issue #7876, opened 2020-07-28
Remove info logging for SearchEngineMigrationResult


pocmo:
We send telemetry for those and that is the only important metric to look at (overall success rate). The individual log messages in Sentry are not helping us.

* https://sentry.prod.mozaws.net/share/issue/e02aae082d0f404d817fc49cdb20dbbd/
* https://sentry.prod.mozaws.net/share/issue/f854493cd8624c1b805d65d554eb0552/

ebalazs-sv:
Confirmed with pocmo, that this issue is not to be verified and will be visible after the rollout, since it makes the error messages better in Sentry.

PR #7909, opened 2020-07-29
Issue #7876: Remove info logging for SearchEngineMigrationResult.

pocmo:
The log messages themselves are not actionable and just add nosie to Sentry. We collect the same
information via telemetry and that gives us a much better view of the success/failure distribution.


pocmo:
bors r=psymoon

psymoon:
bors retry

PR #7912, opened 2020-07-29
Closes #7893: Prevent crash in QrFragment on devices without camera

### LINKED ISSUE #7893: IllegalStateException: No camera found on device
In theory, apps should check if the device has a camera using our extension function `Context.hasCamera`. Looks like in Fenix we forgot to do that for the `PairFragment` and we're seeing quite a few crashes: https://github.com/mozilla-mobile/fenix/issues/13037

We could handle the exception and show an empty screen / log instead.

csadilek:
Let's add this as a fallback in case applications forget to check if the device has a camera e.g. https://github.com/mozilla-mobile/fenix/issues/13037

We really shouldn't be crashing in this case. So, this just displays an error message instead, screenshot below. Working on a unit test too :).

Screenshot_20200729-132835 Screenshot_20200729-131514

### DIFF
--- a/components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt
+++ b/components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt
@@ -41,6 +41,7 @@ import android.view.Surface
 import android.view.TextureView
 import android.view.View
 import android.view.ViewGroup
+import android.widget.TextView
 import androidx.annotation.StringRes
 import androidx.core.content.ContextCompat.getColor
 import androidx.fragment.app.Fragment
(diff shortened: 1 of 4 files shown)


csadilek:
@pocmo for uplift / next release :).

csadilek:
bors r=pocmo

liuche:
@pocmo could we get an uplift of this to AC 52.0.1? That will be the release version right after rollout completes.

If there are other items to uplift, we could also put them into 52.0.1 (with Fenix cut on 7/29) so we don't regress anything in the follow-up release.

pocmo:
I've put the AC 52 label on this PR to remember to uplift this. Right now I am still busy with the 48 dot release. The 52 dot release is not urgent, right? It's a bit overwhelming having two releases to take care of in parallel.

liuche:
Sorry about that, @pocmo, that is a lot of releases in-flight. I'll send an email with what we need for the 52.0.1 dot release.

pocmo:
Uplifted to 52.

PR #7929, opened 2020-07-30
Closes #7861: Add SearchActions and SearchReducer

### LINKED ISSUE #7861: Create actions and reducer to modify SearchState
RFC: #7655

In #7860 we are defining the state. Additionally we need a reducer and a set of actions to modify the `SearchState`.

We may need at least actions for:
* Adding a list of search engines (after loading initial list from disk)
* Adding a custom search engine
* Removing a custom search engine
* Maybe (later?) updating the list of search engines (e.g. locale or region switch)

psymoon:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
@@ -5,6 +5,7 @@
 package mozilla.components.browser.state.action
 
 import android.graphics.Bitmap
+import mozilla.components.browser.state.search.SearchEngine
 import mozilla.components.browser.state.state.BrowserState
 import mozilla.components.browser.state.state.ContainerState
 import mozilla.components.browser.state.state.ContentState
@@ -19,6 +20,7 @@ import mozilla.components.browser.state.state.WebExtensionState
 import mozilla.components.browser.state.state.content.DownloadState
 import mozilla.components.browser.state.state.content.FindResultState
 import mozilla.components.browser.state.state.MediaState
+import mozilla.components.browser.state.state.SearchState
 import mozilla.components.concept.engine.EngineSession
 import mozilla.components.concept.engine.EngineSessionState
 import mozilla.components.concept.engine.HitResult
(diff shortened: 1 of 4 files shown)


components/browser/state/src/main/java/mozilla/components/browser/state/reducer/SearchReducer.kt
```diff
@@ -0,0 +1,74 @@
...
+): BrowserState {
+    return copy(search = search.copy(
+        searchEngines = searchEngineList.map { it.id to it }.toMap()
+    ))
+}
+
+private fun BrowserState.updateSearchEngine(
+    searchEngine: SearchEngine
+): BrowserState {
+    val mutableSearchEngineMap = search.searchEngines.toMutableMap()
```
csadilek, line 45: fly-by nit: :) I'd prefer if we didn't use mutable state in the reducers, even if it's in local scopes. The method would also become shorter:

```
private fun BrowserState.updateSearchEngine(
    searchEngine: SearchEngine
): BrowserState {
    return copy(search = search.copy(
        searchEngines = search.searchEngines + (searchEngine.id to searchEngine)
    ))
}
```

Same for `removeSearchEngine` below.

csadilek: Example where we use an immutable map: https://github.com/mozilla-mobile/android-components/blob/master/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/MediaReducer.kt#L69

psymoon: Awesome, thanks :).

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -614,3 +616,28 @@ sealed class ContainerAction : BrowserAction() {
...
+sealed class SearchAction : BrowserAction() {
+    /**
+     * Updates [BrowserState.search] to add/modify [SearchState.searchEngines].
+     */
+    data class UpdateSearchEngineListAction(val searchEngineList: List<SearchEngine>) : SearchAction()
+
+    /**
+     * Updates [BrowserState.search] to add/modify a custom [SearchEngine].
+     */
+    data class UpdateCustomSearchEngineAction(val searchEngine: SearchEngine) : SearchAction()
```
pocmo, line 632: Looking at this I wonder if we should rethink the `type` we introduced on `SearchEngine`. Maybe we should use actually different types (classes). 🤔  This will make it more obvious and safe: Here we say `UpdateCustomSearchEngine` but at the end of the day we accept all types of `SearchEngine` as parameter. And that can get confusing for the consuming app which may start to use the wrong actions? Not necessarily something for this PR, but maybe a follow-up?

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -614,3 +616,28 @@ sealed class ContainerAction : BrowserAction() {
...
 }
+
+/**
+ * [BrowserAction] implementations related to updating search engines in [SearchState].
+ */
+sealed class SearchAction : BrowserAction() {
+    /**
+     * Updates [BrowserState.search] to add/modify [SearchState.searchEngines].
+     */
+    data class UpdateSearchEngineListAction(val searchEngineList: List<SearchEngine>) : SearchAction()
```
pocmo, line 627: Naming nit regarding `Update` (also for action below): Maybe we should make it explicit and name it `Add` or `Set` and only do that. Actions that potentially do two things can cause issues later.

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
psymoon: Filed https://github.com/mozilla-mobile/android-components/issues/7946 :).

psymoon:
bors r=pocmo

psymoon:
bors retry

Issue #7459, opened 2020-06-22
Provide API call to fetch top sites


MarcLeclair:
We added a `prefetch` for topsites that uses `observerForever` from the `FenixApplication` class. The hiccup here is that it doesn't seem very clean and using `LiveData` doesn't return our data quick enough (as in, the coroutine doesn't dispatch on the main thread since its busy doing other things). The goal of prefetching was to draw everything on the first pass instead of having to wait for a callback.  Here's the issue associated with the prefetch : https://github.com/mozilla-mobile/fenix/issues/11660

The goal of this ticket would be to ideally have a query that doesn't return liveData  in order to control the flow of our fragment creation. 

I'm not sure how feasible this is, this was just an idea to avoid using `observeForever` in Fenix.

Edit: Just an aside, this is important from the product perspective since it allows us to get faster to our "final" state of our homescreen, and therefore speeds our start up time.

data-sync-user:
➤ Patricia Lawless commented:

The GH ticket was moved to "Done" 2 weeks ago.

mcomella:
@csadilek Can you confirm this was addressed and put it back into the Top 10 issues column if not? The GH history implies this was moved via JIRA but may not actually have been fixed.

mcomella:
Actually, I did that steps – please close if it's actually done.

csadilek:
@mcomella nope, this issue was somehow eaten by Jira / closed by mistake.

csadilek:
Talked to @gabrielluong about this. He is very likely going to address this as part of https://github.com/mozilla-mobile/fenix/issues/10342.

I think we can just turn `getTopSites` into a `suspend` function and not have it return livedata, which would fix this issue here as well.

csadilek:
Closing as we have a suspend function for this now: `TopSitesStorage.getTopSites`

PR #7931, opened 2020-07-30
RFC: Introduce a Top Sites Feature

jonalmeida:
@gabrielluong and I wrote up an RFC on the changes that should land for Top Sites.

### DIFF
--- a/docs/rfcs/0006-top-sites-feature.md
+++ b/docs/rfcs/0006-top-sites-feature.md
@@ -21,7 +21,7 @@ To begin, we should rename the `TopSitesStorage`  to `PinnedSitesStorage` and cr
 
 We can then follow the architecture of existing features by using the presenter/interactor/view model, and introduce a `TopSitesStorage` which abstract out the complexity of the multiple data sources from `PinnedSitesStorage` and `PlacesHistoryStorage`.
 
-To allow the ability for adding a top site from different parts of the app (e.g. context menus), we introduce the `TopSitesUseCases` that acts on the storage directly. If the `TopSitesFeature` is started, it will then be notified by the `Observer` to update the UI.
+To allow the ability for adding a top site from different parts of the app (e.g. context menus), we introduce the `PinnedSitesUseCases` that acts on the storage directly. If the `TopSitesFeature` is started, it will then be notified by the `Observer` to update the UI.
 
 ```kotlin
 
@@ -56,7 +56,7 @@ interface TopSitesStorage {
    * Return a unified list of top sites based on the given number of sites desired.
    * If `includeFrecent` is true, fill in any missing top sites with frecent top site results.
    */
-  fun getTopSites(totalNumberOfSites: Int, includeFrecent: Boolean): List<TopSite>
+  suspend fun getTopSites(totalNumberOfSites: Int, includeFrecent: Boolean): List<TopSite>
 
   interface Observer {
     /**
@@ -73,22 +73,23 @@ interface PlacesHistoryStorage {
 
 class DefaultTopSitesStorage(
   val pinnedSitesStorage: PinnedSitesStorage,
-  val historyStorage: HistoryStorage? = null
+  val historyStorage: PlacesHistoryStorage
 ) : TopSitesStorage {
 
   /**
    * Merge data sources here, return a single list of top sites.
    */
-  override fun getTopSites(numFrecent): List<TopSite>
+  override suspend fun getTopSites(totalNumberOfSites: Int, includeFrecent: Boolean): List<TopSite>
 }
 
 /**
- * Use cases can be used for adding a top from different places like a context menu.
+ * Use cases can be used for adding a pinned site from different places like a context menu.
  */
-class TopSitesUseCases(topSitesStorage: TopSitesStorage) {
-  val addTopSites: AddTopSiteUseCase
-  val removeTopSites: RemoveTopSiteUseCase
+class PinnedSitesUseCases(pinnedSitesStorage: PinnedSitesStorage) {
+  val addPinnedSites: AddPinnedSiteUseCase
+  val removePinnedSites: RemovePinnedSiteUseCase
 }
+
 /**
  * View-bound feature that updates the UI when changes are made.
  */
@@ -97,9 +98,9 @@ class TopSitesFeature(
   val presenter: TopSitesPresenter,
   val view: TopSitesView,
   val defaultSites: () -> List<TopSites>
-) {
-  fun start()
-  fun stop()
+) : LifecycleAwareFeature {
+  override fun start()
+  override fun stop()
 }
 ```
 
--- a/docs/rfcs/0006-top-sites-feature.md
+++ b/docs/rfcs/0006-top-sites-feature.md
@@ -1,0 +1,125 @@
+---
+layout: page
+title: Introduce a Top Sites Feature
+permalink: /rfc/0006-top-sites-feature
+---
+
+* Start date: 2020-07-30
+* RFC PR: [#7931](https://github.com/mozilla-mobile/android-components/pull/7931)
+
+## Summary
+
+Create a TopSitesFeature that abstracts out the co-ordination between the multiple data storages.
+
+## Motivation
+
+A large part of the top sites implementation is spread across A-C and Fenix. There is a lot of co-ordination between the History API and the current Top Sites API to manage the final state of what sites should be displayed that can exist as a single implementation.
+
+## Guide-level explanation
+
+To begin, we should rename the `TopSitesStorage`  to `PinnedSitesStorage` and create a `TopSitesFeature` that would orchestrate the different storages and present them to the app.
+
+We can then follow the architecture of existing features by using the presenter/interactor/view model, and introduce a `TopSitesStorage` which abstract out the complexity of the multiple data sources from `PinnedSitesStorage` and `PlacesHistoryStorage`.
+
+To allow the ability for adding a top site from different parts of the app (e.g. context menus), we introduce the `TopSitesUseCases` that acts on the storage directly. If the `TopSitesFeature` is started, it will then be notified by the `Observer` to update the UI.
+
+```kotlin
+
+/**
+ * Implemented by the application for displaying onto the UI.
+ */
+interface TopSitesView {
+  /**
+   * Update the UI.
+   */
+  fun displaySites(sites: List<TopSite>)
+
+  interface Listener {
+    /**
+     * Invoked by the UI for us to add to our storage.
+     */
+    fun onAddTopSite(topSite: TopSite)
+
+    /**
+     * Invoked by the UI for us to remove from our storage.
+     */
+    fun onRemoveTopSite(topSite: TopSite)
+  }
+}
+
+/**
+ * Abstraction layer above the multiple storages.
+ */
+interface TopSitesStorage {
+
+  /**
+   * Return a unified list of top sites based on the given number of sites desired.
+   * If `includeFrecent` is true, fill in any missing top sites with frecent top site results.
+   */
+  fun getTopSites(totalNumberOfSites: Int, includeFrecent: Boolean): List<TopSite>
+
+  interface Observer {
+    /**
+     * Invoked when changes are made to the storage.
+     */
+    fun onStorageUpdated()
+  }
+}
+
+// Already exists in browser-storage-sync.
+interface PlacesHistoryStorage {
+  fun getTopFrecentSites(num: Int)
+}
+
+class DefaultTopSitesStorage(
+  val pinnedSitesStorage: PinnedSitesStorage,
+  val historyStorage: HistoryStorage? = null
+) : TopSitesStorage {
+
+  /**
+   * Merge data sources here, return a single list of top sites.
+   */
+  override fun getTopSites(numFrecent): List<TopSite>
+}
+
+/**
+ * Use cases can be used for adding a top from different places like a context menu.
+ */
+class TopSitesUseCases(topSitesStorage: TopSitesStorage) {
+  val addTopSites: AddTopSiteUseCase
+  val removeTopSites: RemoveTopSiteUseCase
+}
+/**
+ * View-bound feature that updates the UI when changes are made.
+ */
+class TopSitesFeature(
+  val storage: TopSitesStorage,
+  val presenter: TopSitesPresenter,
+  val view: TopSitesView,
+  val defaultSites: () -> List<TopSites>
+) {
+  fun start()
+  fun stop()
+}
+```
+
+## Drawbacks
+
+* `TopSitesStorage` always pulls sites from a persistent storage (Room). This could cause an excessive amount of reads from disk - this is drawback that exists in today's implementation as well. Places has an in-memory cache that should not be affected by this.
+
+## Rationale and alternatives
+
+- This will remove our dependency on `LiveData` in Fenix by updating the `TopSitesView` when we have the top sites available to display; see [#7459](https://github.com/mozilla-mobile/android-components/issues/7459).
+- We can reduce the multiple implementations of the TopSitesFeature by introducing it into Android Components.
+
+## Prior art
+
+* Fenix abstraction of Top Sites: [TopSiteStorage.kt](https://github.com/mozilla-mobile/fenix/blob/3d3153039ce794df09a243968b888ae7cb856d9b/app/src/main/java/org/mozilla/fenix/components/TopSiteStorage.kt)
+* The RFC structure can be see in other components such as, [FindInPageFeature](https://github.com/mozilla-mobile/android-components/blob/master/components/feature/findinpage/src/main/java/mozilla/components/feature/findinpage/FindInPageFeature.kt)  & [SyncedTabsFeature](https://github.com/mozilla-mobile/android-components/blob/master/components/feature/syncedtabs/src/main/java/mozilla/components/feature/syncedtabs/SyncedTabsFeature.kt).
+
+## Unresolved questions
+
+* When the user removes a top site from the UI that comes from `TopSitesStorage`, we remove it from the storage. When a top site comes from frecent, what do we do? Should we be removing the frecent result?
+
+
+


gabrielluong:
cc @csadilek to look over the LiveData

docs/rfcs/0006-top-sites-feature.md
```diff
@@ -0,0 +1,125 @@
...
+
+A large part of the top sites implementation is spread across A-C and Fenix. There is a lot of co-ordination between the History API and the current Top Sites API to manage the final state of what sites should be displayed that can exist as a single implementation.
+
+## Guide-level explanation
+
+To begin, we should rename the `TopSitesStorage`  to `PinnedSitesStorage` and create a `TopSitesFeature` that would orchestrate the different storages and present them to the app.
+
+We can then follow the architecture of existing features by using the presenter/interactor/view model, and introduce a `TopSitesStorage` which abstract out the complexity of the multiple data sources from `PinnedSitesStorage` and `PlacesHistoryStorage`.
+
+To allow the ability for adding a top site from different parts of the app (e.g. context menus), we introduce the `TopSitesUseCases` that acts on the storage directly. If the `TopSitesFeature` is started, it will then be notified by the `Observer` to update the UI.
```
grigoryk, line 24: `PinnedSitesUseCases`? You're not adding a top site (which could be either pinned, frecent or default), you'd be specifically adding a pinned site.

docs/rfcs/0006-top-sites-feature.md
```diff
@@ -0,0 +1,125 @@
...
+/**
+ * Abstraction layer above the multiple storages.
+ */
+interface TopSitesStorage {
+
+  /**
+   * Return a unified list of top sites based on the given number of sites desired.
+   * If `includeFrecent` is true, fill in any missing top sites with frecent top site results.
+   */
+  fun getTopSites(totalNumberOfSites: Int, includeFrecent: Boolean): List<TopSite>
```
grigoryk, line 59: Why do you need `includeFrecent`? Where would we use this "top sites" without frecent sites? If you need that, you can just interact with the pinned sites storage directly.

Let's try to keep this as simple as possible, and only add such flags if they're clearly very necessary.

docs/rfcs/0006-top-sites-feature.md
```diff
@@ -0,0 +1,125 @@
...
+/**
+ * View-bound feature that updates the UI when changes are made.
+ */
+class TopSitesFeature(
+  val storage: TopSitesStorage,
+  val presenter: TopSitesPresenter,
+  val view: TopSitesView,
+  val defaultSites: () -> List<TopSites>
+) {
+  fun start()
```
grigoryk, line 101: What would start/stop do?

docs/rfcs/0006-top-sites-feature.md
```diff
@@ -0,0 +1,125 @@
...
+  fun stop()
+}
+```
+
+## Drawbacks
+
+* `TopSitesStorage` always pulls sites from a persistent storage (Room). This could cause an excessive amount of reads from disk - this is drawback that exists in today's implementation as well. Places has an in-memory cache that should not be affected by this.
```
grigoryk, line 108: It's fine to get this from the db, caching layers add complexity that is very nice to avoid if you don't need it.

Also, I don't think places has an in-memory cache, it reads data from the sqlite db. It's _ really quick_, since that's a `select` against a table with an index on the `frecent` score, IIRC.

docs/rfcs/0006-top-sites-feature.md
```diff
@@ -0,0 +1,125 @@
...
+  /**
+   * Update the UI.
+   */
+  fun displaySites(sites: List<TopSite>)
+
+  interface Listener {
+    /**
+     * Invoked by the UI for us to add to our storage.
+     */
+    fun onAddTopSite(topSite: TopSite)
```
grigoryk, line 41: Let's get the terminology right here. What does that mean, "addTopSite"? Do you actually mean "pin site"? I think so, since frecent sites are outside of our control, and default sites don't change.

Let's not mix-up the naming at the api level. We "get a list of top sites" (since it's a merge of three data sources), but we create "pinned sites".

docs/rfcs/0006-top-sites-feature.md
```diff
@@ -0,0 +1,125 @@
...
+   * Return a unified list of top sites based on the given number of sites desired.
+   * If `includeFrecent` is true, fill in any missing top sites with frecent top site results.
+   */
+  fun getTopSites(totalNumberOfSites: Int, includeFrecent: Boolean): List<TopSite>
+
+  interface Observer {
+    /**
+     * Invoked when changes are made to the storage.
+     */
+    fun onStorageUpdated()
```
grigoryk, line 65: Why do you need this? There's no mechanism for tying changes in the history store (e.g. new frecent sites after a sync) to this listener.

I don't think we need to live-update top sites UI as things change. Top Sites won't really change while you're looking at them (the only reason it can is if sync runs in the background and significantly changes the frecent scores).

docs/rfcs/0006-top-sites-feature.md
```diff
@@ -0,0 +1,125 @@
...
+}
+
+// Already exists in browser-storage-sync.
+interface PlacesHistoryStorage {
+  fun getTopFrecentSites(num: Int)
+}
+
+class DefaultTopSitesStorage(
+  val pinnedSitesStorage: PinnedSitesStorage,
+  val historyStorage: HistoryStorage? = null
```
grigoryk, line 76: You're the hard-coded default sites. It should also be passed in here (as a `List` or a `Set`), so that the implementation can merge them in when necessary.

grigoryk, line 76: Why this `HistoryStorage` nullable? We'll always have history storage.

docs/rfcs/0006-top-sites-feature.md
```diff
@@ -0,0 +1,125 @@
...
+) : TopSitesStorage {
+
+  /**
+   * Merge data sources here, return a single list of top sites.
+   */
+  override fun getTopSites(numFrecent): List<TopSite>
```
grigoryk, line 82: `numFrecent` should just be `num`. Application has no way to know how many frecent sites there are, and neither does it care. What it cares is how many "top sites" it wants to display, and this storage is responsible for providing that number of top sites (some pinned, some frecent, some default).

docs/rfcs/0006-top-sites-feature.md
```diff
@@ -0,0 +1,125 @@
...
+}
+
+/**
+ * Use cases can be used for adding a top from different places like a context menu.
+ */
+class TopSitesUseCases(topSitesStorage: TopSitesStorage) {
```
grigoryk, line 88: Again, `PinnedSitesUseCases`, since you're only modifying a single storage. (not frecent sites, not default sites).

docs/rfcs/0006-top-sites-feature.md
```diff
@@ -0,0 +1,125 @@
...
+  val removeTopSites: RemoveTopSiteUseCase
+}
+/**
+ * View-bound feature that updates the UI when changes are made.
+ */
+class TopSitesFeature(
+  val storage: TopSitesStorage,
+  val presenter: TopSitesPresenter,
+  val view: TopSitesView,
+  val defaultSites: () -> List<TopSites>
```
grigoryk, line 99: should be considered as part of `TopSitesStorage`.

grigoryk: Or, why do we even need `TopSitesStorage`? Can't we just merge within the feature?

docs/rfcs/0006-top-sites-feature.md
ekager, line 41: In the current (or planned) Fenix implementation of Top Sites - can a user ever add a top site from the "Top Site View"?

docs/rfcs/0006-top-sites-feature.md
```diff
@@ -0,0 +1,125 @@
...
+We can then follow the architecture of existing features by using the presenter/interactor/view model, and introduce a `TopSitesStorage` which abstract out the complexity of the multiple data sources from `PinnedSitesStorage` and `PlacesHistoryStorage`.
+
+To allow the ability for adding a top site from different parts of the app (e.g. context menus), we introduce the `TopSitesUseCases` that acts on the storage directly. If the `TopSitesFeature` is started, it will then be notified by the `Observer` to update the UI.
+
+```kotlin
+
+/**
+ * Implemented by the application for displaying onto the UI.
+ */
+interface TopSitesView {
```
ekager, line 31: Are we also going to add a default implementation of this in the feature like `ReaderViewControlsBar`/`ReaderViewControlsView` or are we planning on writing it all in Fenix?

docs/rfcs/0006-top-sites-feature.md
```diff
@@ -0,0 +1,125 @@
...
+- We can reduce the multiple implementations of the TopSitesFeature by introducing it into Android Components.
+
+## Prior art
+
+* Fenix abstraction of Top Sites: [TopSiteStorage.kt](https://github.com/mozilla-mobile/fenix/blob/3d3153039ce794df09a243968b888ae7cb856d9b/app/src/main/java/org/mozilla/fenix/components/TopSiteStorage.kt)
+* The RFC structure can be see in other components such as, [FindInPageFeature](https://github.com/mozilla-mobile/android-components/blob/master/components/feature/findinpage/src/main/java/mozilla/components/feature/findinpage/FindInPageFeature.kt)  & [SyncedTabsFeature](https://github.com/mozilla-mobile/android-components/blob/master/components/feature/syncedtabs/src/main/java/mozilla/components/feature/syncedtabs/SyncedTabsFeature.kt).
+
+## Unresolved questions
+
+* When the user removes a top site from the UI that comes from `TopSitesStorage`, we remove it from the storage. When a top site comes from frecent, what do we do? Should we be removing the frecent result?
```
ekager, line 122: Hm would it be within scope to make an exclusions list of sites the user has removed and presumably doesn't want to see as a top site? I think this would be a fairly common use case 😬

docs/rfcs/0006-top-sites-feature.md
jonalmeida: Good point! The terminology is not accurate here.

docs/rfcs/0006-top-sites-feature.md
jonalmeida: We do not know how many top sites to query for, so the UI can request the number of top sites it wants to display and the remaining can be filled by frecent.

For example, if the UI can display 10 top site tiles, we have 6 pinned sites, and with `includeFrecent` true, the remaining 4 would come from the history storage.

We took the idea from the discussion [in this PR](https://github.com/mozilla-mobile/android-components/pull/7258#issuecomment-661465105) to fill the remaining with the frecent data. If you add a new pinned site, one less frecent would show up.

docs/rfcs/0006-top-sites-feature.md
jonalmeida: TopSitesFeature will be view bound by the Android lifecycle, so those methods would nullify any binds to `TopSitesView.Listener` if it is implemented on an Android View.

docs/rfcs/0006-top-sites-feature.md
jonalmeida: Unfortunately, I believe there are already some startup perf concerns with reading databases that need to show data on the homescreen, so we may run into this sooner than later 🤷 . For example: https://github.com/mozilla-mobile/android-components/issues/7459

docs/rfcs/0006-top-sites-feature.md
jonalmeida: We kept `onAddTopSite` here since it exists on the API level and not the implementation we have (which is a pinned site). Using top sites instead of pinned sites conforms to your standard naming of the feature itself on desktop too.

Should we switch them up to have the API `onAddPinnedSite` everywhere?

docs/rfcs/0006-top-sites-feature.md
jonalmeida: TopSiteFeature will be a view bound feature is it needs to inform the UI when to display content. We have use cases currently in Fenix where you can use other UI menus to alter your top sites while you can see the top sites (see screenshot below).

For this reason we need to notify the storage observers to update the top sites list that can be displayed on the visible UI if there is one.

This is also when we added the concept of a `TopSitesUseCase` to act on the storage and not on the view-bound feature.

Screen Shot 2020-07-30 at 7 23 54 PM

docs/rfcs/0006-top-sites-feature.md
jonalmeida: This was in-error based on an idea that we may not want a history storage. What we really wanted was to have a defaultSites callback instead! 😄

docs/rfcs/0006-top-sites-feature.md
jonalmeida: Oops, typo.

docs/rfcs/0006-top-sites-feature.md
jonalmeida: We left the use cases as top sites as well based on the known problems listed in the RFC: how do we delete a frecent site?

@gabrielluong had made a good point that we don't want to alter the frecent data from history since we are just a consumer of the data and our use case (top sites) shouldn't alter the frecent data that might be valid in a different context.

docs/rfcs/0006-top-sites-feature.md
jonalmeida: Yeah, the `defaultSites` should go to the storage later.

I mentioned it above, we need the storage abstraction so that we can separate the view-bound feature from the data layer which can be used in non-view bound contexts.

docs/rfcs/0006-top-sites-feature.md
jonalmeida: I imagined there was going to be some kind of button there eventually if screen real-estate available to add a top site, we could add top sites easily when/if that UX decision comes through.

It seemed fitting to have a symmetric API, but yeah, the app doesn't have to invoke this today.

docs/rfcs/0006-top-sites-feature.md
jonalmeida: I think leaving the implementation up to the app makes sense right now. For Fenix, we would just have the existing UI implement the interface and get this feature update for free, similar to [`SyncedTabsLayout`](https://github.com/mozilla-mobile/fenix/blob/6cdf304d10826c9df521b9f59fa8415cf79d8170/app/src/main/java/org/mozilla/fenix/sync/SyncedTabsLayout.kt#L25) that implements the view.

docs/rfcs/0006-top-sites-feature.md
gabrielluong: On the Fenix side, we want to provide a user setting to disable frecent sites. So, fetching frecent sites can be optional if the users doesn't want it.

docs/rfcs/0006-top-sites-feature.md
jonalmeida: @gabrielluong mentioned this too! The implementation of a blacklist could definitely work at the `TopSitesStorage`, where does this blacklist get stored is something we should figure out.

With the frecent data coming from the history storage, adding an exclusion list would mean removing sites that show up from there. This will _add_ to our storage layer such that our exclusion list could get much larger over time as we keep asking for the next frecent site and moving it to our blacklist.

I was hoping @grigoryk had some ideas what we should could do here.

jonalmeida: I also realized we don't have an API from A-S to be able to do this today. :\

A hacky implementation would be based on our exclusion list size which assumes our exclusion list matches the frecent data:

```kotlin
// Alter request to get actual new frecent results
val requestNumOfFrecent = exclusionList.size + originalFrecentRequest size.

// Get a larger list of frecents than what we want
val frecentList = historyStorage.getTopFrecentSites(requestNumOfFrecent)

// Throw out the older frecents because we blacklisted them from previous attempts
val actualFrecentList = frecentList.take(exclusionList.size, frecentList.size) 
```

ekager: Yeah and good to keep in mind that the user will probably need to be able to edit or at least reset that list as well a la Site Permissions, Tracking Protection, Logins exceptions 😕

docs/rfcs/0006-top-sites-feature.md
grigoryk: Oh, I agree completely that remaining site slots should be filled out w/ frecents.

> we want to provide a user setting to disable frecent sites

Interesting! that makes sense then.

docs/rfcs/0006-top-sites-feature.md
grigoryk: From that discussion, it feels like the `LiveData` part was problematic, not the db access itself?

But yeah, while you'll need to get this data from somewhere on first load, subsequent loads can benefit from an in-memory cache to fill out the UI sooner. This logic can be hidden within the feature later on (or live in the app); for example, we can compute the top sites now and then as a bg job, keep our in-memory cache updated, etc... I wouldn't worry about this for now, doesn't seem like anything here will get in the way of this sort of thing.

docs/rfcs/0006-top-sites-feature.md
grigoryk: Having the API be in terms of adding a pinned site makes sense to me, since it makes it very clear what's actually happening. E.g. it's not at all clear what "addTopSite" would do for a feature which operates over three different sources of data.

You'll also be exposing pinned-sites storage to the app, right? Perhaps this isn't even necessary then? E.g. the application can query via the feature, but add via the storage. (and then re-query after adding).

docs/rfcs/0006-top-sites-feature.md
grigoryk: I see, so the feature/use-case/whatever will trigger this observer method whenever something is removed/added via the app? That seems fine.

The way I first read that, it felt like this will triggered on _any_ storage changes, which I don't think you have visibility into.

docs/rfcs/0006-top-sites-feature.md
grigoryk: Currently, "deleting frecent sites" will look like storing an exclusion list of sites that you'd filter out if they come up as results from the frecent storage. This could live in PlacesHistoryStorage right on top of a-s, or even in a-s itself, and provide a nice "ignore frecent site" api. cc @linacambridge

docs/rfcs/0006-top-sites-feature.md
grigoryk: 👍

docs/rfcs/0006-top-sites-feature.md
grigoryk: @ekager i recall that in fennec's contextual menus, "remove" was _by far_ the most popular option :)

I think what makes most sense at this level is to just assume that the frecency API will not return anything we don't want. Then you don't need to worry about counts and stuff.

Then, we'll need an exclusion API (`ignoreFrecentSite`, `resetIgnoredFrecentSites`). This could be entirely implemented within `PlacesHistoryStorage` on top of the A-S API, or within A-S itself. Or we do the A-C impl first, then move it to A-S if needed... I imagine iOS will have an exact similar flow here, and will benefit from this API.

gabrielluong: cc @linacambridge to weigh in on the exclusion API.

docs/rfcs/0006-top-sites-feature.md
gabrielluong: Yes, I think `PinnedSitesUseCases` is the correct naming here.

docs/rfcs/0006-top-sites-feature.md
gabrielluong: TODO

- [ ] Rename `onAddTopSite` to `onAddPinnedSite`

gabrielluong: There is an issue for adding a top site from home https://github.com/mozilla-mobile/fenix/issues/7543. However, we have flagged it as something that we would revisit after we have frecent top sites since it might not be necessary anymore.

docs/rfcs/0006-top-sites-feature.md
gabrielluong: I agree we need a way to better handle `defaultSites`. For a bit of context, we currently add default sites into the `PinnedSitesStorage` and set a `isDefault` boolean flag to true for these items. See Fenix implementation in https://github.com/mozilla-mobile/fenix/blob/e6a56aec36fba68ebed183c3c268126ece5135df/app/src/main/java/org/mozilla/fenix/components/TopSiteStorage.kt#L56.

One of the requirements would be to allow users to exclude/remove default sites. It is currently easily dealt with by having it inside of `PinnedSitesStorage`, but we will want to re-evaluate that solution with the consideration that default sites can change over time, and we may want to present the updated list of default sites to all users instead of only new installs. Currently, we had one instance where we had changed the default sites in Fenix, but we only cared about new installs so that meant we didn't have to worry about any database migrations and making these changes in A-C.

gabrielluong: TODO

- [ ] Change `HistoryStorage? = null` to `HistoryStorage`

docs/rfcs/0006-top-sites-feature.md
gabrielluong: TODO

- [ ] Change `numFrecent` to `limit` or `num`

gabrielluong:
I am noticing one thing that we didn't address is when the frecent sites data changes we'll need to update the UI.

grigoryk:
> I am noticing one thing that we didn't address is when the frecent sites data changes we'll need to update the UI.

I've mentioned this in one of the comments, but currently you have no way knowing when this happens. And in practice, the only change that's worth thinking through at the first pass is user "deleting" frecent sites. Other changes (like the actual set of frecent sites changing, e.g. in response to sync pulling in bunch of external browsing history) for now can be handled simply by waiting until the next time user opens up top sites UI.

docs/rfcs/0006-top-sites-feature.md
grigoryk: Having the app pass a list of default sites to the feature (or this storage wrapper) solves the migration problem. If that list changes in the app (say, after an update), top sites feature will operate over the new list. So both new installs _and_ existing installs will the new default sites. No db migrations, etc required.

docs/rfcs/0006-top-sites-feature.md
gabrielluong: Regarding the exclusion API, @ekager and Vesta also noted that Fennec and Desktop provide a "Remove from history" option in the context menu for frecent top sites.

ekager: From what I can see - Desktop has "Dismiss" and "Remove from History" on their suggestions from history on new tab page. I assume we would _want_ to at least think about these two options and come up with something flexible enough to handle them. I'm worried if we don't add something to handle this case now, it's going to be way harder to add later.

grigoryk: We should absolutely handle this case. Again, in Fennec "remove a top site" was a very, very popular action.

ekager: In Fennec - what did "Remove a top site" do though?

docs/rfcs/0006-top-sites-feature.md
linacambridge: At first, we were thinking of having a-c store the exclusion list, since it's already storing the pinned list.

But thinking more about it, I wonder if we want to push down pinned _and_ ignored sites into a-s. IOW, we end up with a new tri-state integer column in `moz_places`: -1 for "never show in top sites", 0 for "no preference, show if frecent enough", and 1 for "always show".

Then the query to fetch top sites looks something like:

```sql
-- First, select all sites that the user pinned.
SELECT url, title, pin, frecency FROM moz_places
WHERE pin = 1
UNION ALL
-- Fill the rest with top frecent sites.
SELECT url, title, pin, frecency FROM moz_places 
WHERE pin != -1 AND
      (SUBSTR(url, 1, 6) == 'https:' OR SUBSTR(url, 1, 5) == 'http:') AND
      (last_visit_date_local + last_visit_date_remote) != 0 AND
      NOT hidden
-- User-pinned sites first (1 > 0), then most frecent, up to the limit.
ORDER BY pin, frecency DESC
LIMIT :num
```

WDYT? I think if we're going to track ignored sites in a-s, we might as well go all the way and track pinned ones, too. Having an `ignoreFrecentSite` (`ignoreTopSite`?) method without a corresponding `pinTopSite` feels asymmetrical.

docs/rfcs/0006-top-sites-feature.md
linacambridge: We can add a read-through cache to the Places connection for top sites, but I think that first read on startup that populates the cache is always going to be a potential problem, no? Regardless of whether you're reading from the `Room` SQLite DB or Places SQLite DB?

docs/rfcs/0006-top-sites-feature.md
linacambridge: @gabrielluong Yep, see my comment above—I wonder if we can push both pinning and ignoring into a-s. It feels inconsistent to have Places manage the "never show this site in top sites" state, but a-c (and eventually iOS, and Desktop) manage the pinned ("always show this site in top sites") state. And then the a-s query would have enough information to do the right thing, where you wouldn't have to request `exclusionList.size + originalFrecentRequest.size` sites and then filter out the excluded ones.

docs/rfcs/0006-top-sites-feature.md
gabrielluong: My immediate thoughts were that this would be a very interest approach because we can essentially remove `PinnedSitesStorage` from A-C and just focus on the sites provided by a single storage `PlacesHistoryStorage`. 

Thinking from an outside A-C consumer perspective, I am thinking the `TopSitesFeature` would create a bit more overhead if I wanted to use my own roomDB, but would now need to tie in with the `PlacesHistoryStorage` (A-S storage - no idea if this means setting up their own A-S).

gabrielluong: We are currently working under the assumption that we have frecent top sites in Fenix by the end of August but ideally as early as mid August. I think we need to create a couple of scenarios / steps / chunks of work and see what is most feasible given the timeline, and plan any remaining follow up work as necessary. Of course, before doing that I think settling on a solution comes first.

gabrielluong: Questions I would have if we considered this approach of migrating to keep all sites in `PlacesHistoryStorage`:

- How hard would it be to migrate existing top sites from `TopSitesStorage` to `PlacesHistoryStorage`? Where do we think default top sites provided by the application should go - maybe this goes back to the defaultSites list that is provided by the application, but I think it would need its own exclusion list in the event that an user wants to exclude seeing a particular default site.
- Let's examine the current "Remove from history" context menu action in Desktop / Fennec and re-evaluate if the exclusion list is still the right solution here vs doing the same.

docs/rfcs/0006-top-sites-feature.md
jonalmeida: I'm leaning towards Grisha's thoughts - let's optimize for perf later until we know how we want this to function.

There are some [other discussions here](https://github.com/mozilla-mobile/fenix/pull/13009) about using different cache types to satisfy the visual completeness that UX is looking for. So yes, the initial read will always hurt us, the type of cache is what we can figure out later or even worse, how many layers of it (this last minute is a joke, please no :).

docs/rfcs/0006-top-sites-feature.md
jonalmeida: > I see, so the feature/use-case/whatever will trigger this observer method whenever something is removed/added via the app? That seems fine.

Correct! This is an existing pattern we have across our components that touch the UI. For example, [`FindInPageView`](https://github.com/mozilla-mobile/android-components/blob/master/components/feature/findinpage/src/main/java/mozilla/components/feature/findinpage/view/FindInPageView.kt).

docs/rfcs/0006-top-sites-feature.md
jonalmeida: > TODO
> 
>     * [ ]  Change `numFrecent` to `limit` or `num`

It should match the API from the `TopSitesStorage` since it'll be an override: 

```suggestion
  override fun getTopSites(totalNumberOfSites: Int, includeFrecent: Boolean): List<TopSite>
```

mcomella, line 82: From a performance perspective, there's a problem in the current implementation on startup. The pinned sites can only be returned from an observer. As such, on startup, UI is laid out and drawn with no pinned sites, then, when the pinned sites data is ready, it's laid out again and drawn again with the pinned sites. This 1) causes an undesirable flicker and 2) layout is expensive, particularly on low-end devices so it takes longer to reach visual completeness than if this didn't exist.

I can think of two solutions:
- have a blocking fetch API that 1) we can call early on in startup to prefetch and 2) block on for the initial UI lay out and draw on **(the line I commented on already seems to address our need** but I wanted to raise our need to you just in case it changes)**
- have an async listener API that 1) we can prefetch early on in startup and 2) will return a value even when the value is empty such that we can progressively load the homescreen: we can first inflate an empty layout without the top sites, collections, etc. and, when the data is ready, lay out and draw the complex layouts all at once

Assuming the calls are fast enough, I'm not sure which is preferably from a UI perspective. If they're slow, we may prefer option 2 to get data on the screen faster.

Ideally, all the views on the homescreen uses the same style of data loading so we only have to implement one of these solutions.

CC @csadilek @MarcLeclair

mcomella, line 82: How can we surface telemetry from this API? In particular, from a performance perspective, I'm thinking:
- how long did this query take?
- how large are the data sources we're sourcing the top sites query from? (e.g. how many pinned sites, how much history, etc.)
- how many top sites are there? (assuming it's not constant)

docs/rfcs/0006-top-sites-feature.md
```diff
@@ -0,0 +1,125 @@
...
+---
+
+* Start date: 2020-07-30
+* RFC PR: [#7931](https://github.com/mozilla-mobile/android-components/pull/7931)
+
+## Summary
+
+Create a TopSitesFeature that abstracts out the co-ordination between the multiple data storages.
+
+## Motivation
```
mcomella, line 14: About the RFC process and not this PR: I wonder if it'd be good to add a section to collect a list of "requirements". For example, I made some suggestions about performance needs and telemetry we might want. As we collect feedback and iterate, how can we make sure we're continuing to meet all of these requirements if they're not documented? I'm concerned if they become documented in prose or in the solution, we'll forget about a few and future iterations will omit them.

Also, when looking through the RFC for historical purposes, it can be helpful to understand the requirements explicitly because, if we want to make a change to the API or replace the API, we can evaluate whether the same requirements apply or not.

CC @pocmo @csadilek

docs/rfcs/0006-top-sites-feature.md
jonalmeida: Option 2 is what is in the current proposal: we notify the `TopSitesView.displaySites` which the app can implement on an Android View or daisy chain with other listeners and collect all the data before drawing it on the screen.

This line can act like a prefetch as well (we could make it a suspend as well probably) and let the app decide when is the best time to show the initial data.

jonalmeida: The implementation can emit facts for the app to collect for telemetry. [Some examples](https://github.com/mozilla-mobile/android-components/search?q=emit*Fact&unscoped_q=emit*Fact) of where we do this in other parts of the codebase.

docs/rfcs/0006-top-sites-feature.md
jonalmeida: In terms of naming, this looks a bit odd. A `TopSitesView` that has callbacks to `onAddPinnedSite`?

We can rename the view class to `PinnedSitesView`, but that as a chain reaction of needing to rename everything to "PinedSites". We will then also have a "DefaultPinedSitesStorage" that takes a `PinedSitesStorage` and `PlacesHistoryStorage`. 😃

jonalmeida: Let's punt on this renaming, because it sounds like we can get rid of the `PinedSitesStorage` when we migrate it over the history storage API and then we would have a straight-forward naming for what we call "top sites" everywhere.

jonalmeida:
Let's merge this since we got a good review and all the known issues have an action plan. Let's gooo!

jonalmeida:
bors r=csadilek,gabrielluong

PR #7942, opened 2020-07-31
Introduce concept-base component.

### LINKED ISSUE #7689: Add Profiler interface to support-base
Similar to `CrashReporting` let's create a `Profiler` interface to `support-base`. An `Engine` now exposes an optional `Profiler`... when using this from other components then we do not want to pass an `Engine` to everything, which would mean that everything depends on `concept-engine`. Instead let's create an interface in `support-base`.

Maybe it's even possible to move the `Profiler` interface from `concept-engine` ... since that is already an interface.

pocmo:
As discussed in https://mozac.org/rfc/0003-concept-base-component.

Closes #7689.


components/support/base/build.gradle
```diff
@@ -54,6 +54,7 @@ dependencies {
     implementation Dependencies.kotlin_stdlib
     implementation Dependencies.kotlin_coroutines
 
+    api project(":concept-base")
```
csadilek, line 57: Should `concept-engine` (and other modules) depend on `concept-base` directly? With this they're depending on it transitively `concept-engine` -> `support-base` -> `concept-base`, which I guess is fine too. I think I would prefer the former, to make it explicit. Not sure, wdyt?

pocmo: Yes, they should. This line here is so that consumers of `support-base` do not explicitly import the interfaces too, that seemed unnecessary.

When I wanted to make `concept-engine` depend on `concept-base` I noticed that it already depends on `support-base` (which is exactly what I wanted to avoid) :(

At least `Observable` (that could be moved) and `ObserverRegistry` (which shouldn't be moved to the concept) are causing this.

I don't see a quick fix for this situation. In the long term we may be able to untangle this once the engine can depend on the store and dispatch directly (and therefore doesn't need to be observable independently).

pocmo:
bors r=csadilek

pocmo:
bors r=csadilek

pocmo:
bors retry

Issue #7939, opened 2020-07-31
Fragment already added: SimpleDownloadDialogFragment


csadilek:
https://sentry.prod.mozaws.net/operations/fenix-fennec-nightly/issues/7281917
https://sentry.prod.mozaws.net/operations/fenix-fennec/issues/9070806/

```
java.lang.IllegalStateException: Fragment already added: SimpleDownloadDialogFragment{a94908c} (bc78f0e4-aef9-423e-9d1a-95d7a8c21279) SHOULD_DOWNLOAD_PROMPT_DIALOG}
    at androidx.fragment.app.FragmentStore.addFragment(FragmentStore.java:7)
    at androidx.fragment.app.FragmentManager.addFragment(FragmentManager.java:4)
    at androidx.fragment.app.BackStackRecord.executeOps(BackStackRecord.java:24)
    at androidx.fragment.app.FragmentManager.executeOpsTogether(FragmentManager.java:51)
    at androidx.fragment.app.FragmentManager.removeRedundantOperationsAndExecute(FragmentManager.java:10)
    at androidx.fragment.app.FragmentManager.execPendingActions(FragmentManager.java:4)
    at androidx.fragment.app.FragmentManager$4.run(FragmentManager.java:1)
    at android.os.Handler.handleCallback(Handler.java:907)
    at android.os.Handler.dispatchMessage(Handler.java:105)
    at android.os.Looper.loop(Looper.java:216)
    at android.app.ActivityThread.main(ActivityThread.java:7625)
    at java.lang.reflect.Method.invoke(Method.java)
    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:524)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:987)
```

Issue #7180, opened 2020-05-29
SimpleDownloadDialogFragment memory leak


csadilek:
I've seen this once but wasn't able to reproduce. 

<img width="402" alt="Screen Shot 2020-05-29 at 11 16 03 AM" src="https://user-images.githubusercontent.com/472523/83275904-d2651a00-a19d-11ea-8e83-a6c13f9604cb.png">

cadeyrn:
> I've seen this once but wasn't able to reproduce.

I tried it five times and was able to reproduce it five times. I reproduced it with downloading the .apk file from https://github.com/mozilla-tw/FirefoxLite/releases/tag/v2.5.1.

PR #7948, opened 2020-07-31
Closes issue #7939: Prevent SimpleDownloadDialogFragment to be added multiple times

Amejia481:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt
+++ b/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt
@@ -178,7 +178,8 @@ class DownloadsFeature(
         ).show()
     }
 
-    private fun showDialog(tab: SessionState, download: DownloadState) {
+    @VisibleForTesting
+    internal fun showDialog(tab: SessionState, download: DownloadState) {
         dialog.setDownload(download)
 
         dialog.onStartDownload = {
@@ -191,12 +192,15 @@ class DownloadsFeature(
             useCases.consumeDownload.invoke(tab.id, download.id)
         }
 
-        if (!isAlreadyADialogCreated() && fragmentManager != null) {
-            dialog.show(fragmentManager, FRAGMENT_TAG)
+        if (!dialog.isAdded && !isAlreadyADialogCreated() && fragmentManager != null) {
+            // We want to explicitly show the dialog and commit the transaction now
+            // to avoid adding the fragment multiple times.
+            dialog.showNow(fragmentManager, FRAGMENT_TAG)
         }
     }
 
-    private fun isAlreadyADialogCreated(): Boolean {
+    @VisibleForTesting
+    internal fun isAlreadyADialogCreated(): Boolean {
         return findPreviousDialogFragment() != null
     }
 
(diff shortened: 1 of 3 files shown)


components/feature/downloads/src/test/java/mozilla/components/feature/downloads/DownloadsFeatureTest.kt
```diff
@@ -411,6 +412,31 @@ class DownloadsFeatureTest {
         assertNotNull(toast)
         assertTrue(toast.contains("can’t download this file type"))
     }
+
+    @Test
+    fun `download dialog must be added once`() {
```
csadilek, line 417: nit: only once.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadsFeature.kt
```diff
@@ -191,12 +192,15 @@ class DownloadsFeature(
             useCases.consumeDownload.invoke(tab.id, download.id)
         }
 
-        if (!isAlreadyADialogCreated() && fragmentManager != null) {
-            dialog.show(fragmentManager, FRAGMENT_TAG)
+        if (!dialog.isAdded && !isAlreadyADialogCreated() && fragmentManager != null) {
```
csadilek, line 195: Fix looks OK. Were you able to reproduce the original crash at all? Do you have STRs?

csadilek: (Haven't been able to repro the crash yet and would like to verify manually as well).

Amejia481: We don't have the STR, but I think the issue was caused because we were calling [dialog.show](https://developer.android.com/reference/androidx/fragment/app/DialogFragment#show(androidx.fragment.app.FragmentManager,%20java.lang.String)) that doesn't warranty the  fragment transaction will be committed right away , allowing to dialog to re-add, for this reason we are using [dialog.showNow](https://developer.android.com/reference/androidx/fragment/app/DialogFragment#showNow(androidx.fragment.app.FragmentManager,%20java.lang.String)) that adds the fragment right away and guarding the call using `dialog.isAdded`.

Amejia481: Not sure how to replicate, but it looks like a race condition, where we added a previous dialog, but it wasn't commit when line 195 was called

csadilek: Yes, understood the patch, but was hoping we can reproduce to make sure.

csadilek: I'll look into the memory leak and see if I can "simulate" this as well somehow.

Amejia481: 👍

Amejia481: One thing that I noticed, was that most of the reported devices are low memory phones, not sure if it's connected

csadilek: OK, we verifed this by overriding show and delaying the `commit`. The solution works but we're gonna fold it into the fix for https://github.com/mozilla-mobile/android-components/issues/7180

Amejia481: 👍

csadilek:
New PR: https://github.com/mozilla-mobile/android-components/pull/7973

PR #7987, opened 2020-08-05
Bug 1645844 - Allow the webcompat-reporter to provide a product name to the reporting backend.

denschub:
This adds the ability to automatically add `browser-XXX` labels to our reports. This is important, as all android-components-based products show up as the same thing on webcompat.com - which makes our work a bit harder.

r? @pocmo 

---

### Pull Request checklist

This currently lacks test, because adding tests is a bit harder than I expected. I'm still working on this, but since correct labeling is blocking a teammate, I decided to ship this first. Feel free to shake your fist at me if you disagree. :)

### After merge

### DIFF
--- a/components/feature/webcompat-reporter/src/main/assets/extensions/webcompat-reporter/background.js
+++ b/components/feature/webcompat-reporter/src/main/assets/extensions/webcompat-reporter/background.js
@@ -9,6 +9,7 @@
 const Config = {
   newIssueEndpoint: "https://webcompat.com/issues/new",
   newIssueEndpointPref: "newIssueEndpoint",
+  productName: "",
   screenshotFormat: {
     format: "jpeg",
     quality: 75,
@@ -40,6 +41,19 @@ async function checkEndpointPref() {
     Config.newIssueEndpoint = value;
   }
 }
+
+(() => {
+  let port = browser.runtime.connectNative("mozacWebcompatReporter");
+  port.onMessage.addListener(message => {
+    if ("productName" in message) {
+      Config.productName = message.productName;
+
+      // For now, setting the productName is the only use for this port, and that's only happening
+      // once after startup, so let's disconnect the port when we're done.
+      port.disconnect();
+    }
+  });
+})();
 
 function hasFastClickPageScript() {
   const win = window.wrappedJSObject;
@@ -117,18 +131,21 @@ function getWebCompatInfoForTab(tab) {
         delete graphicsPrefs["layers.acceleration.force-enabled"];
       }
 
-      return Object.assign({}, {
-        tabId: id,
-        blockList,
-        details: Object.assign(graphicsPrefs, {
-          buildID,
-          channel,
-          frameworks,
-          GPUs,
-          hasTouchScreen,
-        }),
-        url,
-      });
+      return Object.assign(
+        {},
+        {
+          tabId: id,
+          blockList,
+          details: Object.assign(graphicsPrefs, {
+            buildID,
+            channel,
+            frameworks,
+            GPUs,
+            hasTouchScreen,
+          }),
+          url,
+        }
+      );
     }
   );
 }
@@ -143,11 +160,11 @@ async function openWebCompatTab(compatInfo) {
   const { details } = compatInfo;
   const params = {
     url: `${compatInfo.url}`,
-    utm_source: "desktop-reporter",
+    utm_source: "android-components-reporter",
     utm_campaign: "report-site-issue-button",
-    src: "desktop-reporter",
+    src: "android-components-reporter",
     details,
-    extra_labels: [],
+    extra_labels: [`browser-${Config.productName}`],
   };
 
   for (let framework of FRAMEWORK_KEYS) {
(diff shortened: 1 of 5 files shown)


components/feature/webcompat-reporter/src/main/assets/extensions/webcompat-reporter/background.js
```diff
@@ -117,18 +127,21 @@ function getWebCompatInfoForTab(tab) {
...
-        details: Object.assign(graphicsPrefs, {
-          buildID,
-          channel,
-          frameworks,
-          GPUs,
-          hasTouchScreen,
-        }),
-        url,
-      });
+      return Object.assign(
```
denschub, line 130: This was `prettier` from inside mozilla-central doing its thing. I could manually undo this, but since this will happen inside m-c anyway, it's probably a good thing to have that here as well.

denschub:
> should the extension version change as we are updating the extension code?

We don't use that information anywhere, and we don't update that extension out-of-band, so it probably doesn't matter. However, that's a good point either way, so I bumped the version by a minor step. :)

components/feature/webcompat-reporter/src/main/assets/extensions/webcompat-reporter/background.js
```diff
@@ -41,6 +42,15 @@ async function checkEndpointPref() {
   }
 }
 
+(() => {
+  let port = browser.runtime.connectNative("mozacWebcompatReporter");
+  port.onMessage.addListener(message => {
```
csadilek, line 47: It seems you only need this message a single time, right? Should we disconnect the port once the product name is known?

components/feature/webcompat-reporter/src/main/assets/extensions/webcompat-reporter/manifest.json
```diff
@@ -3,7 +3,7 @@
   "name": "WebCompat Reporter",
   "description": "Report site compatibility issues on webcompat.com",
   "author": "Thomas Wisniewski <twisniewski@mozilla.com>",
-  "version": "1.3.0",
+  "version": "1.4.0",
```
csadilek, line 6: Yes, let's always change the version when code changes. We're using GV's `ensureBuiltIn` now to install built-in extensions. It checks for version changes.

components/feature/webcompat-reporter/src/main/java/mozilla/components/feature/webcompat/reporter/WebCompatReporterFeature.kt
```diff
@@ -4,25 +4,48 @@
...
+     * @param runtime a WebExtensionRuntime.
+     * @param productName a custom product name used to automatically label reports. Defaults to
+     * "android-components".
      */
-    fun install(runtime: WebExtensionRuntime) {
+    fun install(runtime: WebExtensionRuntime, productName: String = "android-components") {
         runtime.installWebExtension(WEBCOMPAT_REPORTER_EXTENSION_ID, WEBCOMPAT_REPORTER_EXTENSION_URL,
             onSuccess = {
                 logger.debug("Installed WebCompat Reporter webextension: ${it.id}")
+                it.registerBackgroundMessageHandler(
```
csadilek, line 45: Would be nice to cover this with a unit test (see our other built-in extension features), basically to verify that we're sending down the product name. We can file a follow-up issue for the test, but ultimately we want to cover it. :)

components/feature/webcompat-reporter/src/main/assets/extensions/webcompat-reporter/background.js
denschub: Good point, we don't need that port anymore - at least for now. I updated the patch.

components/feature/webcompat-reporter/src/main/java/mozilla/components/feature/webcompat/reporter/WebCompatReporterFeature.kt
denschub: I absolutely agree. The lack of test coverage here (and, for what it's worth, for the other component, `feature-webcompat`) is worrying to me as well. The situation is a bit more complex since we try to share code between desktop and android-components (previously even Fennec), so that's a bit more annoying than I hope it to be.

I already started writing some tests both components, but I'm not yet at a point where I'd call that good, so I didn't include them here. This is still high in my priority list, and I just decided to push this PR ahead of time because this is actively blocking a coworker who is working on metric-related work. I'll absolutely continue that work, though, especially since there are more changes to this component coming soon'ish. :)

components/feature/webcompat-reporter/src/main/assets/extensions/webcompat-reporter/manifest.json
denschub: Good to know, thanks! Will make sure this keeps getting bumped in the future.

components/feature/webcompat-reporter/src/main/java/mozilla/components/feature/webcompat/reporter/WebCompatReporterFeature.kt
csadilek: OK. I filed a follow-up ticket: https://github.com/mozilla-mobile/android-components/issues/8012 so we don't forget. :)

Issue #7992, opened 2020-08-05
Disable "download link" feature, at least for html files


Mugurell:
Seen on Fenix - https://github.com/mozilla-mobile/fenix/issues/10604

Downloading htmls doesn't seem like a common scenario one would do in the mobile browser.
We should look into not showing the `Download link` option in the contextual menu for when pressing on a link in a webpage if that link if for another webpage.
Please refer to the above for more context.

Mugurell:
This was verified in Fenix. Closing the issue.

PR #8004, opened 2020-08-05
For #7992 - Don't show "Download link" for webpages links

Mugurell:
The "Download link" option in the contextual menu might not be a necessary
option if the link is for other webpages.
Speculate based on the url ending whether the link is for a webpage and prevent
showing the "Download link" in this case.

---

### Pull Request checklist

- [x] **Quality**: This PR builds and passes detekt/ktlint check.
- [x] **Tests**:  No tests. Small change.
- [x] **Changelog**: This PR does not need a changelog entry, small change.

### After merge

### DIFF
--- a/components/feature/contextmenu/src/main/java/mozilla/components/feature/contextmenu/ContextMenuCandidate.kt
+++ b/components/feature/contextmenu/src/main/java/mozilla/components/feature/contextmenu/ContextMenuCandidate.kt
@@ -290,7 +290,7 @@ data class ContextMenuCandidate(
         ) = ContextMenuCandidate(
             id = "mozac.feature.contextmenu.download_link",
             label = context.getString(R.string.mozac_feature_contextmenu_download_link),
-            showFor = { _, hitResult -> hitResult.isLink() },
+            showFor = { _, hitResult -> hitResult.isLinkForOtherThanWebpage() },
             action = { tab, hitResult ->
                 contextMenuUseCases.injectDownload(
                     tab.id,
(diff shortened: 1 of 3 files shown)


Mugurell:
Asked on riot if GV could provide more info about the URL, especially it's html type.
If that isn't feasible the best we could do is try to speculate based on the URL ending.

components/feature/contextmenu/src/main/java/mozilla/components/feature/contextmenu/ContextMenuCandidate.kt
```diff
@@ -376,6 +376,13 @@ private fun HitResult.isLink(): Boolean =
     ((this is HitResult.UNKNOWN && src.isNotEmpty()) || this is HitResult.IMAGE_SRC) &&
         getLink().startsWith("http")
 
+private fun HitResult.isLinkForOtherThanWebpage(): Boolean {
+    val link = getLink()
+    return ((this is HitResult.UNKNOWN && src.isNotEmpty()) || this is HitResult.IMAGE_SRC) &&
+        link.startsWith("http") &&
```
Amejia481, line 382: nit: I think this logic can be replace with is a `this.isLink()`

components/feature/contextmenu/src/main/java/mozilla/components/feature/contextmenu/ContextMenuCandidate.kt
```diff
@@ -376,6 +376,13 @@ private fun HitResult.isLink(): Boolean =
...
+        (link.endsWith("html") || link.endsWith("htm"))
```
Amejia481, line 383: nit: I this logic should be negated as we want to show menus on everything, except `urls` that end in `html` extensions.

Mugurell: Totally. Thanks!

Mugurell:
Thank you Arturo!

Regarding improving the detection of webpages links I don't think that is reasonably possible.
..without actually opening a connection and check the response headers like we do when we actually download from an URL.
I see Chrome also downloading htmls and got no response when asked GV on Riot if they can somehow pass the html type to us, so I'm not sure if we can push this further.

PR #8029, opened 2020-08-05
Add orientation parameter to concept-menu show

NotWoods:
Small change to concept-menu since many menus care about their orientation too. Currently the menus I'm testing with in Fenix need to use the BrowserMenuController class instead of the interface simply due to the lack of orientation.


NotWoods:
Actually going to hold off on this, we may want to remove the `width` parameter entirely.

PR #8035, opened 2020-08-06
Bugfix/#8034 Add remove and contains actions to edit toolbar

BranescuMihai:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
+++ b/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
@@ -286,6 +286,19 @@ class BrowserToolbar @JvmOverloads constructor(
     }
 
     /**
+     * Removes an action that was displayed on the right of the URL in edit mode.
+     */
+    override fun removeEditAction(action: Toolbar.Action) {
+        edit.removeEditAction(action)
+    }
+
+    /**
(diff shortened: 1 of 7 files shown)


components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
```diff
@@ -285,6 +285,19 @@ class BrowserToolbar @JvmOverloads constructor(
...
+    override fun removeEditAction(action: Toolbar.Action) {
+        edit.removeEditAction(action)
+    }
+
+    /**
+     * Checks if an action was displayed on the right of the URL in edit mode.
+     */
+    override fun containsEditAction(action: Toolbar.Action): Boolean =
+        edit.containsEditAction(action)
+
```
pocmo, line 300: I think the Fenix issue could also be modeled by using the action's `visible()` function to decide whether it should show up and use `invalidateActions()`?
Afaik the removal for page/browser actions was primarily added for addons where we permanently remove them - not switch between them.
What do you think?

BranescuMihai: @pocmo you're right, didn't check yet for any visual glitches but it feels like the `invalidateActions` is the way to go, therefore I don't think there are any AC changes needed, thanks!

PR #8047, opened 2020-08-06
Closes #8046: Remove unneeded unload handlers to make bfcache work

### LINKED ISSUE #8046: Remove unnecessary unload handlers in content scripts
We currently disconnect our native ports using unload handlers, but this is actually not necessary:

From Rob:
The extension framework already takes care of destroying ports upon unload. No need to do it manually. It's implemented using pageshow+pagehide+a bit more

Furthermore, `unload` handlers prevent the bfcache from working: https://github.com/mozilla-mobile/fenix/issues/12928#issuecomment-670102917

csadilek:
Window unload handlers block the bfcache. The extension framework already automatically disconnects ports on unload so they aren't needed in the first place.

For the reader mode content script, we are now connecting and disconnecting on pageshow/pagehide to support readerable checks when navigating (back/forward) to cached pages. We don't need to do that for reader pages, as we opted out of the cache there.

For FxA there's no difference in behaviour. I've also tested navigating back and forth during the signup flow and it still works, as expected.

For p2p, I opted not to change behaviour, but I am not familiar enough how it would be affected not that the bfcache is active.

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/accounts/src/main/assets/extensions/fxawebchannel/fxawebchannel.js
+++ b/components/feature/accounts/src/main/assets/extensions/fxawebchannel/fxawebchannel.js
@@ -16,10 +16,6 @@ port.onMessage.addListener((event) => {
   }));
 });
 
-// We should use pagehide/pageshow events instead, to better handle page navigation events.
-// See https://github.com/mozilla-mobile/android-components/issues/2984.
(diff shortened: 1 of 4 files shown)


csadilek:
bors r=jonalmeida

Issue #6099, opened 2020-02-27
Update Gradle and Android Gradle plugin


pocmo:
There are new versions available. Let's try them.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-5083)

pocmo:
Tried with:
* Gradle 6.0.1
* Android Gradle Plugin 3.6.0

```
FAILURE: Build failed with an exception.

* Where:
Settings file '/Users/sebastiankaspari/Projects/Mozilla/android-components/settings.gradle' line: 42

* What went wrong:
Could not compile settings file '/Users/sebastiankaspari/Projects/Mozilla/android-components/settings.gradle'.
> startup failed:
  settings file '/Users/sebastiankaspari/Projects/Mozilla/android-components/settings.gradle': 42: unable to resolve class Config
   @ line 42, column 37.
         gradle.rootProject.ext.config = new Config(buildconfig.componentsVersion)
```

Meh. Do not want to debug this now.

mcarare:
This was completed long ago.

PR #8049, opened 2020-08-06
For #6099 - Convert config to yaml

NotWoods:
In Gradle 6, buildSrc classes cannot be used inside settings.gradle. Inlining the `Config` class gets around this problem for us. The values inside the class have been moved to a YAML file so that it can be used by Taskcluster to check if data has changed, and used by the changelog to be a simple version overview.

We could also use a .properties file for this.


pocmo:
bors try

pocmo:
bors try

pocmo (approved):
bors r+

PR #8053, opened 2020-08-07
No issue: add command line build hints to README

interfect:
I'm not really an Android developer, so my system didn't have the Android SDK and JAVA_HOME setting required to build the project, and it wasn't quite clear how to obtain them (or that the Android SDK was actually a build dependency).

This adds material to the README that I had to work out by trial and error, and which may be helpful to the next inexperienced Android developer.

---

### Pull Request checklist

### After merge

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,6 +294,26 @@ $ cd android-components
 $ ./gradlew assemble
 ```
 
+### Configuring the Android SDK ###
+
+You will need the Android SDK installed (i.e. pointed to by the `ANDROID_SDK_ROOT` environment variable), and configured to indicate acceptance of the Android SDK licenses. This can be accomplished on Linux as follows:
+
+```
+$ wget http://dl.google.com/android/android-sdk_r24.2-linux.tgz
+$ tar -xvf android-sdk_r24.2-linux.tgz
+$ cd android-sdk-linux
+$ mkdir -p licenses
+$ echo "8933bad161af4178b1185d1a37fbf41ea5269c55" >> licenses/android-sdk-license
+$ echo "d56f5187479451eabf01fb78af6dfcb131a6481e" >> licenses/android-sdk-license
+$ echo "24333f8a63b6825ea9c5514f83c2829b004d1fee" >> licenses/android-sdk-license
+```
+
+Then run `pwd` and use the resulting value as `ANDROID_SDK_ROOT`
+
+### Configuring Java ###
+
+You will need `JAVA_HOME` set to a path to a JDK (which should not end in `/jre`).
+
 ## Android Studio ##
 
 If the environment variable `JAVA_HOME` is not defined, you will need to set it. If you would like to use the JDK installed by Android Studio, here's how to find it:


README.md
```diff
@@ -294,6 +294,26 @@ $ cd android-components
 $ ./gradlew assemble
 ```
 
+### Configuring the Android SDK ###
+
+You will need the Android SDK installed (i.e. pointed to by the `ANDROID_SDK_ROOT` environment variable), and configured to indicate acceptance of the Android SDK licenses. This can be accomplished on Linux as follows:
+
+```
```
jonalmeida, line 301: I don't think we should include command-line examples that point to a particular version numbers because it can lead to people copying the same contents are pasting it when these versions go out of date quickly. It also means that we have to maintain more documentation to keep the version numbers and hashes up-to-date.

How about we point to the official documentation on how to setup the SDK if we need to a section for it?

README.md
```diff
@@ -294,6 +294,26 @@ $ cd android-components
...
+$ echo "8933bad161af4178b1185d1a37fbf41ea5269c55" >> licenses/android-sdk-license
+$ echo "d56f5187479451eabf01fb78af6dfcb131a6481e" >> licenses/android-sdk-license
+$ echo "24333f8a63b6825ea9c5514f83c2829b004d1fee" >> licenses/android-sdk-license
+```
+
+Then run `pwd` and use the resulting value as `ANDROID_SDK_ROOT`
+
+### Configuring Java ###
+
+You will need `JAVA_HOME` set to a path to a JDK (which should not end in `/jre`).
```
jonalmeida, line 315: Android Studio also has a packaged version of `OpenJDK` so I think there is something else at play here (see review comment).

interfect: I tried to put this under the command line, non-Android-Studio heading. I was specifically trying to avoid the Android Studio 800 MB download and get the build working with as few moving parts as possible (i.e. on the system-provided JDK, entirely from the command line, with only the Android SDK components that are actually necessary, and without installing a whole IDE).

In my particular environment (on Ubuntu with Java 8 set as the default Java) I had to make sure to set JAVA_HOME to the Java 11 JDK. I'm not sure if that's because it picked up the Java 8 JDK which was incompatible somehow, or it picked up the Java 8 JRE which wasn't a JDK. (Setting JAVA_HOME to the Java 8 JDK still seemed to produce the error about not being able to find the required JDK tools.)

If it auto-magically detects the JDK for most people (maybe on other distros or when a reasonably new Java is the default) this could definitely be softened to a troubleshooting tip.

README.md
interfect: I did link a pretty old SDK version. Since I made this PR I've switched to installing the Command Line Tools and letting Gradle use them to install the SDK itself.

There isn't quite official documentation on how to install the Command Line Tools that I can find. Google mostly just wants you to use Android Studio, and all their articles on installing things already have their command line tools to install stuff with.

The command line tools download is at https://developer.android.com/studio#downloads under "Command line tools only", and the docs on how to use the sdkmanager it provides are [here](https://developer.android.com/studio/command-line/sdkmanager), but they don't seem to walk you through setting up the directory structure that makes it work. 

How about this approach: it fetches the newest Command Line Tools download URL at runtime by parsing it out of Google's site (because there's no stable URL for the most recent version).

```
mkdir android-sdk-linux
cd android-sdk-linux
mkdir -p licenses
echo "8933bad161af4178b1185d1a37fbf41ea5269c55" >> licenses/android-sdk-license
echo "d56f5187479451eabf01fb78af6dfcb131a6481e" >> licenses/android-sdk-license
echo "24333f8a63b6825ea9c5514f83c2829b004d1fee" >> licenses/android-sdk-license
mkdir cmdline-tools
cd cmdline-tools
wget "$(curl -s https://developer.android.com/studio | grep -oP "https://dl.google.com/android/repository/commandlinetools-linux-[0-9]+_latest.zip")"
unzip commandlinetools-linux-*_latest.zip
cd ..
```

jonalmeida:
In my experience, documentation like this can go stale pretty quickly in the Android world. Maybe there some additions we should make to [the general android docs for setting up the Java environment](https://github.com/mozilla-mobile/shared-docs/blob/master/android/configure_java.md) instead of the android component specific ones?

interfect:
A link to the general doc would be good. Maybe some of this can become the Linux section of that doc, which doesn't seem to be written yet.

Do I just open a PR on that repo?

jonalmeida:
Yes, that should be fine.

jonalmeida:
Based on the above discussion, I'm going to close this.

Issue #8054, opened 2020-08-07
UX changes for addons - unsupported item


Mugurell:
Based on @brampitoyo's comment from https://github.com/mozilla-mobile/fenix/issues/8578#issuecomment-652167232
and
https://app.abstract.com/share/28d0811b-ffed-4863-89c8-8aaa10acb528?mode=build&selected=3739771863-5D69D8D0-1838-4F55-B7BA-6EE363185D86&sha=e00d4f002ecef58f001903d9111f95b864daa8ae\
we need the following changes:
- use the same style as for the other favicons - [Mozac.Widgets.Favicon](https://github.com/mozilla-mobile/android-components/blob/297ed9f1fdbce91c934de824af412df2934c8074/components/ui/widgets/src/main/res/values/styles.xml#L37)
- ensure the addon icon has 24 dp width/height and uses the same glyph as the other addons icons used in menus, one that has inherent padding.
- ensure the remove icon has 20dp height.
`mozac_ic_delete` [actually has a 24dp size set](https://github.com/mozilla-mobile/android-components/blob/daa3595a9130d3b28f0fea64d5094762be2abd58/components/ui/icons/src/main/res/drawable/mozac_ic_delete.xml) but also an innate padding so it can be used as is.

brampitoyo:
@Mugurell Sorry to change the icon spec one more time. Our add-ons icon should follow the same guidelines as favicons (see https://github.com/mozilla-mobile/fenix/issues/7992).

Unfortunately, in the past, favicons look different on Top Sites, Bookmarks, History and Search Results. Sometimes they have a square background, or round background, or sometimes they have no background at all.

Now that the specs have been unified, we should follow it.

Mugurell:
To be verified in Fenix - https://github.com/mozilla-mobile/fenix/issues/8578

PR #8055, opened 2020-08-07
For #8054 - Unify addons icons design with the one for favicons

Mugurell:
This is a breaking change, specifying a different background is not possible
anymore.

Will use the same vector glyph with inherent padding as default addons icon.

---

### Pull Request checklist

- [x] **Tests**: No tests. Small UI change.
- [x] **Changelog**: This PR does include a changelog entry.
- [x] **Accessibility**: The code in this PR does not include any user facing accessibility changes.

AddonsIcons
[video](https://drive.google.com/file/d/1TLgDuCx8BwCdoof3y0ZNCP5AYTR6G8DL/view?usp=sharing)

### After merge


components/feature/addons/src/main/res/layout/mozac_feature_addons_unsupported_item.xml
```diff
@@ -12,14 +12,25 @@
...
+    <androidx.cardview.widget.CardView
+        android:id="@+id/icon_container"
+        android:layout_width="40dp"
+        android:layout_height="40dp"
         android:importantForAccessibility="no"
-        android:scaleType="fitCenter"
-        app:srcCompat="@drawable/mozac_ic_extensions"
-        app:tint="?android:attr/textColorPrimary" />
+        app:cardCornerRadius="20dp"
+        app:cardElevation="0dp">
```
pocmo, line 21: Interesting approach using `CardView` as a circular background. Is there any advantage over using a background drawable (which I assume is cheaper since it doesn't need to be part of the layouting)?

Mugurell: Thanks! Great point!

I went with the CardView just because the layout for installed addons used it already and it provided the nice round background that [Bram recommended](https://github.com/mozilla-mobile/fenix/issues/8578#issuecomment-652167232).
..background which is different from [the specs](https://app.abstract.com/share/28d0811b-ffed-4863-89c8-8aaa10acb528?mode=build&selected=3366345710-5CF8E47B-A223-4368-BE1C-4EEBF8ACDFAE&sha=e00d4f002ecef58f001903d9111f95b864daa8ae) but looks better.
Going with that light grey background would give us this 
image

Mugurell: Thinking that we could add some transparency to the grey background... just don't know how much.. I'll have to eyeball it.

Mugurell: Would need different transparency levels for dark/white themes to get the same effect.
So.. I propose sticking with the CardView.

Amejia481: We are also using a `CardView` on [the add-ons list](https://github.com/mozilla-mobile/android-components/blob/0094209ac99076787f4e2d3693fff4aa60c40472/components/feature/addons/src/main/res/layout/mozac_feature_addons_item.xml#L17), for the background we are using [the background provide by the app ](https://github.com/mozilla-mobile/android-components/blob/0094209ac99076787f4e2d3693fff4aa60c40472/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonsManagerAdapter.kt#L235), I think it would be the same in this case.

Mugurell: Based on Bram's design guidelines and after confirming with Arturo I went with removing the CardView and also the programatic background.
Thank you both!

Mugurell:
Regarding the naming of the style now used - `Mozac.Widgets.Favicon` which also uses `mozac_widget_favicon_background` and `mozac_widget_favicon_size` I though more about it and I would say it's okay to keep these as they are, mainly because:
- They are used in many other places -> hard to find a name better for all usecases
- They intent to unify the design of all favicons used all throughout the app. I'm thinking about favicons loosely here - any resource icons, like in the case of addons also.

Mugurell:
Since this is a breaking change it should be merged only after https://github.com/mozilla-mobile/fenix/pull/13523 is merged.

Amejia481:
The change looks good!
I'm doing some manual testing to verify that everything is working as expected.

brampitoyo:
@Amejia481 @Mugurell This looks amazing!

Just to confirm: are the icons inside the box (including not yet available add-ons) all sized at 24dp? It seems larger to me, but I could have assumed wrong.

Mugurell:
> Just to confirm: are the icons inside the box (including not yet available add-ons) all sized at 24dp? It seems larger to me, but I could have assumed wrong.
> 
> If all our add-ons icons are sized according to the favicon spec – which is 24dp – then this change gets an r+ from me.

It's a bit tough to get a measurement so what I've done is to use a [red border](https://pastebin.com/xKjx707z) for the addons icons in the [supported addons layout](https://pastebin.com/ZQ7Sb84F) and for the addon icons and the remove button in the [not yet supported addons layout](https://pastebin.com/Ef53CViK) with hardcoded height and width.

With the icons border having 24dp height/width and the remove icon border having 20dp height/width [**it is seen in this video**](https://drive.google.com/file/d/1vSNmMz-q-F_O-tNqIKkXXtGi3l5nD3kH/view?usp=sharing) that it matches exactly the actual dimensions of the icons.

Mugurell:
So if it seems ok with you @brampitoyo I'll merge this.

pocmo:
(unassign: arturo reviewed already, and I'll be on PTO next week)

brampitoyo:
@Mugurell I see – thanks for the clarification!

My intent is for the icon for ‘generic’ unsupported add-on, to match the size of the add-on icon used elsewhere within the app (e.g. inside the main menu, in settings, etc.)

And for the rubbish bin icon to also match the one used in settings.

Is it possible to try to match the icon size?

Otherwise, when the add-on has a logo (or a “favicon” if we put it that way), the size is perfect. And so is the size of the border surrounding it.

Thanks!

Mugurell:
Awesomely spotted @brampitoyo !
Confirmed we are using the same 24dp size for all addons icons
but we are not using the same resource as addons icon.

In [the browser menu and the parent settings screen](https://github.com/mozilla-mobile/fenix/search?q=ic_addons_extensions&unscoped_q=ic_addons_extensions) we use [ic_addons_extensions](https://github.com/mozilla-mobile/fenix/blob/88f01731d7dd6fb9820190e05c35d98c728ef4d0/app/src/main/res/drawable/ic_addons_extensions.xml).
In other places like [the addons checker notification, the unsupported addons screen](https://github.com/mozilla-mobile/android-components/search?q=mozac_ic_extensions&unscoped_q=mozac_ic_extensions) we use [mozac_ic_extensions](https://github.com/mozilla-mobile/android-components/blob/daa3595a9130d3b28f0fea64d5094762be2abd58/components/ui/icons/src/main/res/drawable/mozac_ic_extensions.xml).

The difference is that `ic_addons_extensions` has an inherent padding, as with all drawables used in menus
image

While `mozac_ic_extensions` occupies the entire canvas so it appears as bigger. The same is true for all the addons icons.
image

So I could copy the resource from Fenix into AC and use that for unsupported addons - so all the placeholder addons icons would look to be the same size,
but then the unsupported addons icons might seem smaller compared to other addons icons.

Would've been great for the images to not have any padding from the start and when needed to be added programmatically.
At this time, with different icon sizes (as a visual effect) but consistent size in their screen I'm not sure what's the best approach to unify the experience, looking forward to your input.

Mugurell:
For the trashcan we use again two separate drawable files but which are the same SVGs.
They are 24dp width/height and again have some innate padding.
Since we use them without any scaleType they'll keep the same dimensions, 24dp minus the padding, visually more like 20-21dp I'd say.

brampitoyo:
@Mugurell Aha. So the two assets are different. Thanks for pointing this out! I didn’t know that.

I recommend combining them, so `ic_addons_extensions` (the one that contains inherent padding) is the version that we use everywhere in the app. Not just in menus and Settings, but also as placeholder unsupported add-ons, in modal dialogs, etc.

Our original intention was as follows:

- If the icon contains a vector glyph – like the ones we use in menus and Settings – then it’s meant to have inherent paddings
- If the icon is a logo/branding/graphic, then it shouldn’t have any padding – it would simply occupy 100% of the alloted space

The final result is that, in some cases, the logo/branding/graphic asset would look visually larger than the vector glyph. But on the whole, the “artboard” (which includes padding) should always have one consistent size.

Mugurell:
Thank you @brampitoyo !
I've made the changes to use the same padded addons icon in all places and updated the first comment with a video showing the result.

Amejia481:
> Thank you @brampitoyo !
> I've made the changes to use the same padded addons icon in all places and updated the first comment with a video showing the result.

@Mugurell thanks all the great work. The changes look great ✅ , is the pr ready to land?

Mugurell:
> > Thank you @brampitoyo !
> > I've made the changes to use the same padded addons icon in all places and updated the first comment with a video showing the result.
> 
> @Mugurell thanks all the great work. The changes look great ✅ , is the pr ready to land?

I would say so. I was kind of waiting for the final ok from Bram.

Amejia481:
> > > Thank you @brampitoyo !
> > > I've made the changes to use the same padded addons icon in all places and updated the first comment with a video showing the result.
> > 
> > 
> > @Mugurell thanks all the great work. The changes look great ✅ , is the pr ready to land?
> 
> I would say so. I was kind of waiting for the final ok from Bram.

No problem :), feel free to land when you are ready!

brampitoyo:
@Mugurell [wrote](https://github.com/mozilla-mobile/android-components/pull/8055#issuecomment-675855099):
> …use the same padded addons icon in all places and updated the first comment with a video showing the result.

This looks perfect to me – sorry for the overdue reply!

brampitoyo (approved):

Mugurell:
bors r=Amejia481, brampitoyo

Issue #7725, opened 2020-07-15
Expose more data about why sharing a tab failed


Mugurell:
Needed for Fenix -  https://github.com/mozilla-mobile/fenix/issues/10662

I see the call to share a tab getting to [sendCommandToDeviceAsync](https://github.com/mozilla-mobile/android-components/blob/df80cf4173680a77e8e8807975a04f42400bdaa8/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FxaDeviceConstellation.kt#L105) and then to a rust call.
Unfortunately `sendCommandToDeviceAsync` returns a boolean so we'd only know if the operation succeeded or not, but in the case it failed, we would not know the reason.

Depending on the fail reason we might want to start a different flow (maybe one for authentication?), use different ux - different error messages, or log errors with better granularity.

Mugurell:
CCing @mhammond and @grigoryk since the changes needed to support this are to be done in A-S.
It would be great if instead of a boolean we would get a bigger payload with both the status and some message, details about the result of the operation.

mhammond:
I think that change would be in the android-components link above - the rust code will throw an exception if it fails, which  sendCommandToDeviceAsync converts to the bool - but IIRC, `handleFxaExceptions` should log the reason and also handle account state reasons, but @grigoryk can tell you more about that.

grigoryk:
@Mugurell take a look at https://github.com/mozilla-mobile/android-components/blob/df80cf4173680a77e8e8807975a04f42400bdaa8/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/Utils.kt#L19

This is what happens during that call. In case of auth problems, the account state machine will attempt automatic recovery, and if that fails will switch into a "auth problems" state, which is exposed to the app via AccountObserver events. Currently in Fenix that will trigger a "yellow" account state in the UI.

Most common failures you'll see here are network failures. Other than auth problems, those are the only ones actionable by the user.

We already throw in case we see an "fxa panic" exception. It may be worth it to log `FxaException.Unspecified` to Sentry, that's our only "blind spot" right now here.

Mugurell:
Thank you @grigoryk !

I see though that on iOS some specific failures are logged
- [no tabs were sent](https://github.com/mozilla-mobile/firefox-ios/blob/76faae8c6c94e22c8fbc0de72e2d06f615738e2c/Extensions/NotificationService/NotificationService.swift#L196) (Don't really know how this could happen)
[...]

PR #8082, opened 2020-08-10
For #7725 - Expose the result of sending tabs

Mugurell:
Needed for Fenix to log errors that might occur when sending tabs.
For proper logging we need to expose not just the success/failure status but
also the reason for why the operation failed, since there may be different
causes that need to be logged differently / acted upon differently.

Previously we would just fold the success/failure - true/false results of the
"share tab" operations. To be able to log exactly why sharing failed we need
the exact reasons exposed to the client.

---

### Pull Request checklist

- [x] **Accessibility**: The code in this PR follows does not include any user facing features

### After merge

### DIFF
--- a/components/concept/sync/src/main/java/mozilla/components/concept/sync/Devices.kt
+++ b/components/concept/sync/src/main/java/mozilla/components/concept/sync/Devices.kt
@@ -70,7 +70,10 @@ interface DeviceConstellation : Observable<AccountEventsObserver> {
      * @param outgoingCommand An event to send.
      * @return A [Deferred] that will be resolved with a success flag once operation is complete.
      */
-    fun sendCommandToDeviceAsync(targetDeviceId: String, outgoingCommand: DeviceCommandOutgoing): Deferred<Boolean>
+    fun sendCommandToDeviceAsync(
+        targetDeviceId: String,
+        outgoingCommand: DeviceCommandOutgoing
+    ): Deferred<FxaOperationResult>
 
     /**
      * Process a raw event, obtained via a push message or some other out-of-band mechanism.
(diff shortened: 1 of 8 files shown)


Mugurell:
@mhammond While rebasing this I saw you recently merged a nice patch to `processFxaTelemetry`.
Maybe you are a good reviewer for this?

mhammond (commented):
The code itself looks fine to me. However, I'm not a huge fan of the new return type TBH - I don't see a good reason to not just use the original exceptions - so `Deferred<Boolean>` can just become `Deferred<Unit>` and the docs changed to indicate that FxA exceptions will be thrown - IIUC, the raw exceptions already have the granularity you need here.

There's an argument that sending multiple tabs, or to multiple devices are best served using this new return type (as you return a list of them) , however, ultimately this batch sending should be handled by application-services so that the correct telemetry can be generated, which can't be the way it's set up now - see https://github.com/mozilla/application-services/issues/3402 for more.

However, you really need to take this with a grain of salt, as I'm not a reviewer of this project! You may well find that @grigoryk has an exact opposite opinion. I believe he is on PTO, so @liuche might be able to help with a formal review in the mean time.

Mugurell:
Thank you Mark!
The issues on Fenix is simple, we just need to log to Sentry why sending tabs failed but I'm not sure how to better bubble up the errors from fxa to Fenix instead of the current boolean which just signify sending failed.
We could have a wrapper over the exception or just have it unchecked and let it bubble up to clients.
Seems like a decision for maintainers of this code, will wait for Grisha.

Mugurell:
Opened a new PR based on the direction proposed by Mark - https://github.com/mozilla-mobile/android-components/pull/8344

PR #8159, opened 2020-08-19
Update Flank to v20.08.3

AaronMT:
Updating the fetch Flank command to a pinned version so it's easier to identify at a glance which version Flank is used, similar to Fenix here: https://github.com/mozilla-mobile/fenix/blob/master/taskcluster/docker/ui-tests/Dockerfile

Updating Flank to v20.08.2 https://github.com/Flank/flank/releases/tag/v20.08.2


AaronMT:
We need to touch the UI Dockerfile too. I'll make the change.

taskcluster/docker/ui-tests/Dockerfile
```diff
@@ -28,6 +28,10 @@ RUN curl https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud
     && ${TEST_TOOLS}/google-cloud-sdk/install.sh --quiet \
     && ${TEST_TOOLS}/google-cloud-sdk/bin/gcloud --quiet components update
 
+#------------------------
```
rpappalax, line 31: this is already a variable defined in the base image so we probably shouldn't also hard code the API version number here

rpappalax:
small nit, otherwise ok

taskcluster/docker/ui-tests/Dockerfile
AaronMT: Is it not a different parent though? One lives in `/automation/docker` and the other file lives in `/taskcluster/docker/` the base in `/taskcluster/docker` only deals with high level system setup. For now since we just need to touch the `/taskcluster/docker/ui-tests/Dockerfile` to update Flank, I'll just update API string to match the one above.

rpappalax: @JohanLorenzo we noticed that there is still an old zombie Dockerfile in place.  Is this still being used?  https://github.com/mozilla-mobile/android-components/blob/master/automation/docker/Dockerfile
It looks like it was replaced by /taskcluster/docker/base, etc.

JohanLorenzo: Nice finding! TL;DR: Let's remove that old file.

https://github.com/mozilla-mobile/android-components/blob/a64c68498c92b04c99b2f913ae81ebae51cc8615/automation/docker/Dockerfile was used when we didn't automatically build docker images on Taskcluster. We used to build them locally, then publish them on Dockerhub, then Taskcluster consumes them. This process was risky because we had to trust whoever/whatever machine built that image. 

In other words, we shouldn't need this Dockerfile anymore. If we still do, then we should port the bits that aren't already in https://github.com/mozilla-mobile/android-components/blob/a64c68498c92b04c99b2f913ae81ebae51cc8615/taskcluster/docker/base/Dockerfile.

AaronMT:
Pushed a new change that removes the unused Dockerfile

PR #8170, opened 2020-08-20
Update GeckoView Nightly to 81.0.20200820093209

jonalmeida:
---

Manual upgrade needed for the breaking API changes.


Amejia481:
We had some tests that were testing how we react  to `WebExtension.MetaData` be null, as now it's `NonNull` the tests needed to be updated.

Issue #7134, opened 2020-05-27
Implement Autocomplete Login Select Feature


ekager:
In #7132 I fixed the breaking changes, but I still need to actually implement the new select a login prompt feature.

PR #8178, opened 2020-08-20
For #7134 - Adds login selection to prompt feature

ekager:
---

### Pull Request checklist

### After merge


components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt
```diff
@@ -240,6 +264,20 @@ class PromptFeature private constructor(
                         prompt.dismiss()
                     }
                     activePrompt?.clear()
+                    loginPicker.dismissCurrentLoginSelect()
+                }
+        }
+
+        sessionPromptScope = store.flowScoped { flow ->
```
ekager, line 271: This was added for dismissing prompts when a new tab is selected (ie via the swipe gesture or long press to open a new tab). Is this the best thing to observe on for that?

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt
```diff
@@ -240,6 +263,20 @@ class PromptFeature private constructor(
                         prompt.dismiss()
                     }
                     activePrompt?.clear()
+                    loginPicker.dismissCurrentLoginSelect()
+                }
+        }
+
+        // Dismiss prompts when a new tab is selected.
+        sessionPromptScope = store.flowScoped { flow ->
+            flow.ifChanged { browserState -> browserState.selectedTabId }
```
jonalmeida, line 272: This would mean that we only be notified for selected sessions and not custom tabs, is that right? I think your previous version that looking for the session ID would have worked too, just the `ifChanged` needed to be simplified from: 
`ifChanged { it.engine.engineSession }` to just `ifChanged()`.

ekager:
ezgif com-video-to-gif(9)

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/login/LoginSelectBar.kt
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    private fun toggleSavedLoginsHeader(shouldExpand: Boolean) {
+        if (shouldExpand) {
+            loginsList?.isVisible = true
+            manageLoginsButton?.isVisible = true
+            loginPickerView?.hideKeyboard()
+            expandArrowHead?.rotation = ROTATE_UPSIDE_DOWN
+        } else {
+            expandArrowHead?.rotation = 0F
```
jonalmeida, line 129: ROTATE_DOWNSIDE_UP? 🙃

ekager: good thing i didnt go with FLIPPIDITY_DIPPITY

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt
ekager: Yeah I think selected sessions ignoring custom tabs should be what we want here! And this should also resolve the issue here with other prompts showing again after switching tabs https://github.com/mozilla-mobile/fenix/issues/12943#issuecomment-669069988

jonalmeida: That's fair. It would isolate our Sync users that sign-in (which uses a Custom Tab) who may have sync'd with logins, signed out, and then attempted to sign it again using the saved logins, but maybe that's too fine of an edge case.

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/login/LoginSelectBar.kt
jonalmeida: ⅄┴IԀԀIp‾⅄pIԀԀI˥Ⅎ would work but aapt might get mad at you.

Issue #8185, opened 2020-08-21
Offer a SwipeRefreshLayout trigger pull to refresh only for vertical scrolls


Mugurell:
Needed for Fenix - https://github.com/mozilla-mobile/fenix/issues/10241

Interestingly `SwipeRefreshLayout` does not check for only vertical scroll to trigger pull to refresh.

### Expected result
The pull to refresh throbber should only appear for when scrolling down and the webpage is at top.

### Actual result
- A swipe gesture more on the x axis triggers pull to refresh
- A two finger scale gesture triggers pull to refresh
- A quick scale gesture (one finger double tap and move) triggers pull to refresh
- Descended initiated nested scrolls trigger pull to refresh, for us meaning doubling the throbber with the overscroll shadow.

![SwipeRefreshIssues](https://user-images.githubusercontent.com/11428869/90861540-91dbfd80-e394-11ea-91c5-c0cb5ee96c71.gif)

[video](https://drive.google.com/file/d/1aQPidEnbxYfNxu6KQ4gikUwwoGELahlB/view?usp=sharing)

Mugurell:
Verified as fixed on Fenix - https://github.com/mozilla-mobile/fenix/issues/10241#issuecomment-690254408

PR #8186, opened 2020-08-21
For #8185 - Add a new VerticalSwipeRefreshLayout

Mugurell:
This comes to resolve many issues with SwipeRefreshLayout but based on a lot of
debugging. As a result the code is somewhat brittle.

Because we'd control responding to touches and updating the UI we need to be as
swift as possible and as such to differentiate between possible gestures I am
some simple layman's ways of doing this and GestureDetector / ScaleDetector
instances since they would add to the execution time and overall complexity,
and their callbacks would be hard to synchronize.

---

### Pull Request checklist

- [x] **Accessibility**: The code in this PR does not include any user facing accessibility features.

FixedPullToRefresh
[video](https://drive.google.com/file/d/18ix3V6G6r8qkOkK-bWM7sAZWlBfh9Uhi/view?usp=sharing)
[apks to test](https://drive.google.com/drive/folders/1cvrpb_nS1z5dbJ4IfD58esB73qd0yUMm?usp=sharing)

### After merge

### DIFF
--- a/components/ui/widgets/build.gradle
+++ b/components/ui/widgets/build.gradle
@@ -33,6 +33,7 @@ dependencies {
     implementation Dependencies.androidx_constraintlayout
     implementation Dependencies.androidx_core_ktx
     implementation Dependencies.google_material
+    implementation Dependencies.androidx_swiperefreshlayout
 
     testImplementation project(":support-test")
 
--- a/components/ui/widgets/src/main/java/mozilla/components/ui/widgets/VerticalSwipeRefreshLayout.kt
+++ b/components/ui/widgets/src/main/java/mozilla/components/ui/widgets/VerticalSwipeRefreshLayout.kt
@@ -1,0 +1,197 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.ui.widgets
+
+import android.annotation.SuppressLint
+import android.content.Context
+import android.util.AttributeSet
+import android.view.MotionEvent
+import android.view.View
+import android.view.ViewConfiguration
+import androidx.annotation.VisibleForTesting
+import androidx.swiperefreshlayout.widget.SwipeRefreshLayout
+import kotlin.math.abs
+
+/**
+ * [SwipeRefreshLayout] that filters only vertical scrolls for triggering pull to refresh.
+ *
+ * Following situations will not trigger pull to refresh:
+ * - a scroll happening more on the horizontal axis
+ * - a scale in/out gesture
+ * - a quick scale gesture
+ *
+ * To control responding to scrolls and showing the pull to refresh throbber or not
+ * use the [View.isEnabled] property.
+ */
+class VerticalSwipeRefreshLayout @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet? = null
+) : SwipeRefreshLayout(context, attrs) {
+    @VisibleForTesting
+    internal var isQuickScaleInProgress = false
+    @VisibleForTesting
+    internal var quickScaleEvents = QuickScaleEvents()
+    private var previousX = 0f
+    private var previousY = 0f
+    private val doubleTapTimeout = ViewConfiguration.getDoubleTapTimeout()
+    private val doubleTapSlop = ViewConfiguration.get(context).scaledDoubleTapSlop
+    private val doubleTapSlopSquare = doubleTapSlop * doubleTapSlop
+
+    @Suppress("ComplexMethod", "ReturnCount")
+    override fun onInterceptTouchEvent(event: MotionEvent): Boolean {
+        // Setting "isEnabled = false" is recommended for users of this ViewGroup
+        // who who are not interested in the pull to refresh functionality
+        // Setting this easily avoids executing code unneededsly before the check for "canChildScrollUp".
+        if (!isEnabled) {
+            return false
+        }
+
+        // Layman's scale gesture (with two fingers) detector.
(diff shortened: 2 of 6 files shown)


components/feature/session/src/main/java/mozilla/components/feature/session/view/VerticalSwipeRefreshLayout.kt
```diff
@@ -0,0 +1,180 @@
...
+
+    override fun onStartNestedScroll(child: View, target: View, nestedScrollAxes: Int): Boolean {
+        // Ignoring nested scrolls from descendants.
+        // Allowing descendants to trigger nested scrolls would defeat the purpose of this class
+        // and result in pull to refresh to happen for all movements on the Y axis
+        // (even as part of scale/quick scale gestures) while also doubling the throbber with the overscroll shadow.
+        return if (isEnabled) {
+            return false
+        } else {
+            callSuperOnStartNestedScroll(child, target, nestedScrollAxes)
```
NotWoods, line 99: This function is public, is the extra `call*` method needed?

components/feature/session/src/main/java/mozilla/components/feature/session/view/VerticalSwipeRefreshLayout.kt
```diff
@@ -0,0 +1,180 @@
...
+            val xDistance = abs(event.x - previousX)
+            val yDistance = abs(event.y - previousY)
+            previousX = event.x
+            previousY = event.y
+            if (xDistance > yDistance) {
+                return false
+            }
+        }
+
+        return callSuperOnInterceptTouchEvent(event)
```
NotWoods, line 88: same here

components/feature/session/src/main/java/mozilla/components/feature/session/view/VerticalSwipeRefreshLayout.kt
```diff
@@ -0,0 +1,180 @@
...
+ * use the [View.isEnabled] property.
+ */
+class VerticalSwipeRefreshLayout @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet? = null
+) : SwipeRefreshLayout(context, attrs) {
+    @VisibleForTesting
+    internal var isQuickScaleInProgress = false
+    @VisibleForTesting
+    internal var quickScaleEvents: Triple<MotionEvent?, MotionEvent?, MotionEvent?> = Triple(null, null, null)
```
NotWoods, line 34: This should be either a data class or just 3 local variables. The lack of naming makes the code hard to read. If you go the data class route, you may want to use mutable `var`s.

components/feature/session/src/main/java/mozilla/components/feature/session/view/VerticalSwipeRefreshLayout.kt
```diff
@@ -0,0 +1,180 @@
...
+    @VisibleForTesting
+    internal fun isQuickScaleInProgress(events: Triple<MotionEvent?, MotionEvent?, MotionEvent?>): Boolean {
+        return if (events.isNotNull()) {
+            isQuickScaleInProgress(events.first!!, events.second!!, events.third!!)
+        } else {
+            false
+        }
+    }
+
+    // Method closely following GestureDetectorCompat#isConsideredDoubleTap.
```
NotWoods, line 151: It's preferable to use KDoc comments here.
```suggestion
    /**
     * Method closely following GestureDetectorCompat#isConsideredDoubleTap.
     * Allows for serial inference of double taps as opposed to using callbacks.
     */
    @VisibleForTesting
```

NotWoods (changes requested):
This might be better off as a new `ui-*` component. I don't think `feature-session` is the right place for this to live in.

components/feature/session/src/main/java/mozilla/components/feature/session/view/VerticalSwipeRefreshLayout.kt
Mugurell: > This might be better off as a new `ui-*` component. I don't think `feature-session` is the right place for this to live in.

Thanks!
Confirmed with Sebastian, will move this to `ui-widgets`

Mugurell: > This function is public, is the extra `call*` method needed?

All the extra `call*` are used in tests to check if `super.*` was called.
Did not know of another way other than this proxies.

components/feature/session/src/main/java/mozilla/components/feature/session/view/VerticalSwipeRefreshLayout.kt
Mugurell: All the extra call* are used in tests to check if super.* was called.
Did not know of another way other than this proxies.

components/feature/session/src/main/java/mozilla/components/feature/session/view/VerticalSwipeRefreshLayout.kt
Mugurell: Fair. Will use a data class. Thanks!

components/feature/session/src/main/java/mozilla/components/feature/session/view/VerticalSwipeRefreshLayout.kt
Mugurell: I prefer to use kdoc/javadoc for public documentation and these type of comment for future maintainers of the code.
The above doesn't really serve as method documentation but rather as implementation specifics.

Mugurell:
Thanks you @NotWoods for helping here, sorry I hadn't responded sooner.

NotWoods (approved):
LGTM!

Mugurell:
bors r=NotWoods

Issue #7762, opened 2020-07-17
Persist downloads on disk  and restore them


Amejia481:
We want to store all the downloads preferable when they change status.

Amejia481:
Closed by https://github.com/mozilla-mobile/android-components/commit/b948ae026bede46115dcd95e2b360fb002609770

PR #8197, opened 2020-08-21
Closes issue #7762: Adds support for persisting/restoring downloads.

Amejia481:
Fenix pr https://github.com/mozilla-mobile/fenix/pull/14043
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
@@ -207,7 +207,7 @@ sealed class ContentAction : BrowserAction() {
     /**
      * Removes the [DownloadState] of the [ContentState] with the given [sessionId].
      */
-    data class ConsumeDownloadAction(val sessionId: String, val downloadId: Long) : ContentAction()
+    data class ConsumeDownloadAction(val sessionId: String, val downloadId: String) : ContentAction()
 
     /**
      * Updates the [HitResult] of the [ContentState] with the given [sessionId].
@@ -700,7 +700,7 @@ sealed class DownloadAction : BrowserAction() {
     /**
      * Updates the [BrowserState] to remove the download with the provided [downloadId].
      */
-    data class RemoveDownloadAction(val downloadId: Long) : DownloadAction()
+    data class RemoveDownloadAction(val downloadId: String) : DownloadAction()
 
     /**
      * Updates the [BrowserState] to remove all downloads.
@@ -711,6 +711,16 @@ sealed class DownloadAction : BrowserAction() {
      * Updates the provided [download] on the [BrowserState].
      */
     data class UpdateDownloadAction(val download: DownloadState) : DownloadAction()
+
+    /**
+     * Restores the [BrowserState.downloads] state from the storage.
+     */
+    object RestoreDownloadsStateAction : DownloadAction()
+
+    /**
+     * Restores the given [download] from the storage.
+     */
+    data class RestoreDownloadStateAction(val download: DownloadState) : DownloadAction()
 }
 
 /**
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/DownloadStateReducer.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/DownloadStateReducer.kt
@@ -25,6 +25,8 @@ internal object DownloadStateReducer {
             is DownloadAction.RemoveAllDownloadsAction -> {
                 state.copy(downloads = emptyMap())
             }
+            is DownloadAction.RestoreDownloadsStateAction -> state
+            is DownloadAction.RestoreDownloadStateAction -> updateDownloads(state, action.download)
         }
     }
 
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/state/content/DownloadState.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/state/content/DownloadState.kt
@@ -7,7 +7,7 @@ package mozilla.components.browser.state.state.content
 import android.os.Environment
 import android.os.Parcelable
 import kotlinx.android.parcel.Parcelize
-import kotlin.random.Random
+import java.util.UUID
 
 /**
  * Value type that represents a download request.
@@ -24,7 +24,7 @@ import kotlin.random.Random
  * @property referrerUrl The site that linked to this download.
  * @property skipConfirmation Whether or not the confirmation dialog should be shown before the download begins.
  * @property id The unique identifier of this download.
- * @property sessionId Identifier of the session that spawned the download.
+ * @property createdTime A timestamp when the download was created.
  * @
  */
 @Suppress("Deprecation")
@@ -40,8 +40,9 @@ data class DownloadState(
     val destinationDirectory: String = Environment.DIRECTORY_DOWNLOADS,
     val referrerUrl: String? = null,
     val skipConfirmation: Boolean = false,
-    val id: Long = Random.nextLong(),
-    val sessionId: String? = null
+    val id: String = UUID.randomUUID().toString(),
+    val sessionId: String? = null,
+    val createdTime: Long = System.currentTimeMillis()
 ) : Parcelable {
     val filePath: String get() =
         Environment.getExternalStoragePublicDirectory(destinationDirectory).path + "/" + fileName
@@ -49,31 +50,32 @@ data class DownloadState(
     /**
      * Status that represents every state that a download can be in.
      */
-    enum class Status {
+    @Suppress("MagicNumber")
+    enum class Status(val id: Int) {
         /**
          * Indicates that the download is in the first state after creation but not yet [DOWNLOADING].
          */
-        INITIATED,
+        INITIATED(1),
         /**
          * Indicates that an [INITIATED] download is now actively being downloaded.
          */
-        DOWNLOADING,
+        DOWNLOADING(2),
         /**
          * Indicates that the download that has been [DOWNLOADING] has been paused.
          */
-        PAUSED,
+        PAUSED(3),
         /**
          * Indicates that the download that has been [DOWNLOADING] has been cancelled.
          */
-        CANCELLED,
+        CANCELLED(4),
         /**
          * Indicates that the download that has been [DOWNLOADING] has moved to failed because
          * something unexpected has happened.
          */
-        FAILED,
+        FAILED(5),
         /**
          * Indicates that the [DOWNLOADING] download has been completed.
          */
-        COMPLETED
+        COMPLETED(6)
     }
 }
--- a/components/feature/downloads/schemas/mozilla.components.feature.downloads.db.DownloadsDatabase/1.json
+++ b/components/feature/downloads/schemas/mozilla.components.feature.downloads.db.DownloadsDatabase/1.json
@@ -1,0 +1,76 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 1,
+    "identityHash": "342d0e5d0a0fcde72b88ac4585caf842",
+    "entities": [
+      {
+        "tableName": "downloads",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`id` TEXT NOT NULL, `url` TEXT NOT NULL, `file_name` TEXT, `content_type` TEXT, `content_length` INTEGER, `status` INTEGER NOT NULL, `destination_directory` TEXT NOT NULL, `created_at` INTEGER NOT NULL, PRIMARY KEY(`id`))",
+        "fields": [
+          {
+            "fieldPath": "id",
+            "columnName": "id",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "url",
+            "columnName": "url",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "fileName",
+            "columnName": "file_name",
+            "affinity": "TEXT",
+            "notNull": false
+          },
+          {
+            "fieldPath": "contentType",
+            "columnName": "content_type",
+            "affinity": "TEXT",
+            "notNull": false
+          },
+          {
+            "fieldPath": "contentLength",
+            "columnName": "content_length",
+            "affinity": "INTEGER",
+            "notNull": false
+          },
+          {
+            "fieldPath": "status",
+            "columnName": "status",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
+          {
+            "fieldPath": "destinationDirectory",
+            "columnName": "destination_directory",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "createdAt",
+            "columnName": "created_at",
+            "affinity": "INTEGER",
+            "notNull": true
+          }
+        ],
+        "primaryKey": {
+          "columnNames": [
+            "id"
+          ],
+          "autoGenerate": false
+        },
+        "indices": [],
+        "foreignKeys": []
+      }
+    ],
+    "views": [],
+    "setupQueries": [
+      "CREATE TABLE IF NOT EXISTS room_master_table (id INTEGER PRIMARY KEY,identity_hash TEXT)",
+      "INSERT OR REPLACE INTO room_master_table (id,identity_hash) VALUES(42, '342d0e5d0a0fcde72b88ac4585caf842')"
+    ]
+  }
+}
--- a/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
+++ b/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
@@ -68,6 +68,7 @@ import mozilla.components.feature.downloads.facts.emitNotificationResumeFact
 import mozilla.components.feature.downloads.facts.emitNotificationTryAgainFact
 import mozilla.components.support.base.log.logger.Logger
 import mozilla.components.support.ktx.kotlin.sanitizeURL
+import mozilla.components.support.ktx.kotlinx.coroutines.throttleLatest
 import mozilla.components.support.utils.DownloadUtils
 import java.io.File
 import java.io.FileOutputStream
@@ -100,7 +101,7 @@ abstract class AbstractFetchDownloadService : Service() {
     internal var compatForegroundNotificationId: Int = COMPAT_DEFAULT_FOREGROUND_ID
     private val logger = Logger("AbstractFetchDownloadService")
 
-    internal var downloadJobs = mutableMapOf<Long, DownloadJobState>()
+    internal var downloadJobs = mutableMapOf<String, DownloadJobState>()
 
     // TODO Move this to browser store and make immutable:
     // https://github.com/mozilla-mobile/android-components/issues/7050
@@ -156,7 +157,7 @@ abstract class AbstractFetchDownloadService : Service() {
             @Suppress("LongMethod")
             override fun onReceive(context: Context, intent: Intent?) {
                 val downloadId =
-                    intent?.extras?.getLong(DownloadNotification.EXTRA_DOWNLOAD_ID) ?: return
+                    intent?.extras?.getString(DownloadNotification.EXTRA_DOWNLOAD_ID) ?: return
                 val currentDownloadJobState = downloadJobs[downloadId] ?: return
 
                 when (intent.action) {
@@ -246,24 +247,28 @@ abstract class AbstractFetchDownloadService : Service() {
     }
 
     override fun onStartCommand(intent: Intent?, flags: Int, startId: Int): Int {
-        val download = intent?.getLongExtra(EXTRA_DOWNLOAD_ID, -1)?.let {
+        val download = intent?.getStringExtra(EXTRA_DOWNLOAD_ID)?.let {
             store.state.downloads[it]
         } ?: return START_REDELIVER_INTENT
 
         // If the job already exists, then don't create a new ID. This can happen when calling tryAgain
         val foregroundServiceId = downloadJobs[download.id]?.foregroundServiceId ?: Random.nextInt()
 
+        val actualStatus = if (download.status == INITIATED) DOWNLOADING else download.status
+
         // Create a new job and add it, with its downloadState to the map
         val downloadJobState = DownloadJobState(
-            state = download.copy(status = DOWNLOADING),
+            state = download.copy(status = actualStatus),
             foregroundServiceId = foregroundServiceId,
-            status = DOWNLOADING
+            status = actualStatus
         )
 
         store.dispatch(DownloadAction.UpdateDownloadAction(downloadJobState.state))
 
-        downloadJobState.job = CoroutineScope(IO).launch {
-            startDownloadJob(downloadJobState)
+        if (actualStatus == DOWNLOADING) {
+            downloadJobState.job = CoroutineScope(IO).launch {
+                startDownloadJob(downloadJobState)
+            }
         }
 
         downloadJobs[download.id] = downloadJobState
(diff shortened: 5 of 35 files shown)


components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlinx/coroutines/Utils.kt
```diff
@@ -0,0 +1,37 @@
...
+import kotlinx.coroutines.launch
+
+/**
+ *
+ * Returns a function that limits the executions of the [block] function, until the [skipTime] passes,
+ * any calls before [skipTime] passes will be ignored.
+ *
+ * Credit to Terenfear https://gist.github.com/Terenfear/a84863be501d3399889455f391eeefe5
+ *
+ * @param skipTime the time to wait until the next call to [block] be processed.
```
csadilek, line 19: Let's make it clear that this is in ms, maybe even by naming the param `skipTimeInMs`.

components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlinx/coroutines/Utils.kt
```diff
@@ -0,0 +1,37 @@
...
+ * @param coroutineScope the coroutine scope where [block] will executed.
+ * @param block function to be execute.
+ */
+fun <T> throttle(
```
csadilek, line 23: Good find. I think you want to use the `throttleLatest` from the example above though? Then after skipTime you would get the latest/newest update to the store. Now you get the oldest.

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -590,6 +590,11 @@ sealed class DownloadAction : BrowserAction() {
      * Updates the provided [download] on the [BrowserState].
      */
     data class UpdateDownloadAction(val download: DownloadState) : DownloadAction()
+
+    /**
+     * Restore the [BrowserState.downloads] state from the storage.
+     */
+    object RestoreDownloadsState : DownloadAction()
```
csadilek, line 597: nit: Let's keep the `Action` postfix. Eventually, I think we can remove `Action` from all these names. I filed a ticket for that a while ago, but for now let's keep it consistent.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadMiddleware.kt
```diff
@@ -7,34 +7,100 @@ package mozilla.components.feature.downloads
...
         action: BrowserAction
     ) {
-        next(action)
         when (action) {
             is DownloadAction.AddDownloadAction -> {
-                val intent = Intent(applicationContext, downloadServiceClass)
-                intent.putExtra(DownloadManager.EXTRA_DOWNLOAD_ID, action.download.id)
-                applicationContext.startService(intent)
+                next(action)
+                scope.launch {
```
csadilek, line 52: Here we should check before we launch. No need to launch the coroutine if we don't need to add to the storage.

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -574,12 +574,12 @@ sealed class DownloadAction : BrowserAction() {
     /**
      * Updates the [BrowserState] to track the provided [download] as added.
      */
-    data class AddDownloadAction(val download: DownloadState) : DownloadAction()
+    data class AddDownloadAction(val download: DownloadState, val restored: Boolean = false) : DownloadAction()
```
csadilek, line 577: Could `restored` just be part of the `DownloadState` instead of being in the action?

components/browser/state/src/main/java/mozilla/components/browser/state/state/content/DownloadState.kt
```diff
@@ -40,40 +40,42 @@ data class DownloadState(
...
+    val createdTime: Long = System.currentTimeMillis()
 ) : Parcelable {
     val filePath: String get() =
         Environment.getExternalStoragePublicDirectory(destinationDirectory).path + "/" + fileName
 
     /**
      * Status that represents every state that a download can be in.
      */
-    enum class Status {
+    @Suppress("MagicNumber")
```
csadilek, line 53: Couldn't we just persist the String value and rely on `Enum.valueOf`. Then we don't need the `StatusConverter` etc.

components/feature/downloads/schemas/mozilla.components.feature.downloads.db.DownloadsDatabase/1.json
```diff
@@ -0,0 +1,76 @@
...
+          {
+            "fieldPath": "contentLength",
+            "columnName": "content_length",
+            "affinity": "INTEGER",
+            "notNull": false
+          },
+          {
+            "fieldPath": "status",
+            "columnName": "status",
+            "affinity": "INTEGER",
```
csadilek, line 44: See my comment / question above. :)

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadMiddleware.kt
```diff
@@ -7,34 +7,100 @@ package mozilla.components.feature.downloads
...
+                }
+            }
+
+            is DownloadAction.RestoreDownloadsState -> {
+                scope.launch {
+                    downloadStorage.getDownloads().collect { downloads ->
+                        downloads.forEach { download ->
+                            if (!store.state.downloads.containsKey(download.id)) {
+                                store.dispatch(DownloadAction.AddDownloadAction(download, true))
+                                logger.error("Download restarted from db")
```
csadilek, line 95: This shouldn't be logged as `error` and should probably be restored, not restarted?

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/db/DownloadsDatabase.kt
```diff
@@ -0,0 +1,55 @@
...
+                "mozac_downloads_database"
+            ).build().also {
+                instance = it
+            }
+        }
+    }
+}
+
+@Suppress("unused")
+internal class StatusConverter {
```
csadilek, line 43: See comment above. I think I would prefer not needing this, but I might miss why this is more practical?

components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlinx/coroutines/Utils.kt
csadilek: Another good thing to point out in the docs, is that the calls to the returned function should happen on the same thread, esp. for `throttleLatest` (see comment below). Unless we make sure we have visibility guarantees. Not a problem the way you use it now though.

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
Amejia481: Do you think that we are going to need it in the future?

components/browser/state/src/main/java/mozilla/components/browser/state/state/content/DownloadState.kt
Amejia481: It could be a bit brittle if somebody renames some `Status`, at least with `ids` is less possible, what do you think?

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadMiddleware.kt
Amejia481: Agree, It's a leftover from testing.

components/browser/state/src/main/java/mozilla/components/browser/state/state/content/DownloadState.kt
csadilek: Well, on the flip side, if someone renames / messes up the names we would get an exception/crash. If someone messes up IDs crazy things can start happening. Eitherway, we need data migration if the status fields change. Up to you :). Don't feel strongly, just thought we can simplify.

components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlinx/coroutines/Utils.kt
Amejia481: Thanks!

components/browser/state/src/main/java/mozilla/components/browser/state/state/content/DownloadState.kt
Amejia481: 👍

Amejia481:
@csadilek I addressed the comments on this commit https://github.com/mozilla-mobile/android-components/pull/8197/commits/1c374b005c647975d683faab19c32fd1fe24fdd1 :)

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
Amejia481: This is the only item left on the updated. I'm not sure if we are going to need `restored`  on `DownloadState` in the future, as now it just serves the purpose of identifying a download is coming from the storage.

components/browser/state/src/main/java/mozilla/components/browser/state/state/content/DownloadState.kt
Amejia481: For the moment, I prefer to keep it as it's :)

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadStorage.kt
```diff
@@ -0,0 +1,89 @@
...
+        downloadDao.deleteAllDownloads()
+    }
+
+    companion object {
+        /**
+         * Takes two [DownloadState] objects and the determine if they are the same, be aware this
+         * only takes into considerations fields that are being stored,
+         * not all the field on [DownloadState] are stored.
+         */
+        fun areTheSame(first: DownloadState, second: DownloadState): Boolean {
```
csadilek, line 78: I think we should move or rename this method. `DownloadStorage.areTheSame(a, b)` doesn't look right. It doesn't check if tow storages are the same, but compares `DownloadState`s. I would move to `DownloadState` or turn into an ext. function.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadStorage.kt
```diff
@@ -0,0 +1,89 @@
...
+            return first.id == second.id &&
```
csadilek, line 79: Now that we have a unique ID, comparing the ID should be good enough here. This would hide a bug if we had two downloads with the same ID but different file name.

components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlinx/coroutines/Utils.kt
```diff
@@ -0,0 +1,41 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+
+/**
+ *
+ * Returns a function that limits the executions of the [block] function, until the [skipTimeInMs]
+ * passes, then the latest value passed to [block] will be used. Any calls before [skipTimeInMs]
+ * passes will be ignored. It's spected that calls to the returned function
```
csadilek, line 16: nit: "expected"? I would write: All calls to the returned function must happen on the same thread.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadMiddleware.kt
```diff
@@ -7,34 +7,100 @@ package mozilla.components.feature.downloads
...
+    @VisibleForTesting
+    internal val downloadStorage: DownloadStorage = DownloadStorage(applicationContext)
 ) : Middleware<BrowserState, BrowserAction> {
+    private val logger = Logger("DownloadMiddleware")
 
+    private var scope = CoroutineScope(coroutineContext)
+
+    @InternalCoroutinesApi
     override fun invoke(
         store: MiddlewareStore<BrowserState, BrowserAction>,
```
csadilek, line 45: We have an API change here on master. So we need to update this to receive a `MiddlewareContext` (or simply rebase). Everything else will be fine e.g. you can still `dispatch` on it etc.

components/support/ktx/src/main/java/mozilla/components/support/ktx/kotlinx/coroutines/Utils.kt
Amejia481: 👍

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadStorage.kt
Amejia481: I didn't want to added directly to `DownloadState` as it could get confused with `equals`, because It only cares about properties that the storage take into consideration, we don't want to overwhelm the storage with every update on the `DownloadMiddleware`  if nothing that the storage cares has changed  : 
```
            is DownloadAction.UpdateDownloadAction -> {
                val updated = action.download
                store.state.downloads[updated.id]?.let { old ->
                    // To not overwhelm the storage, we only send updates that are relevant,
                    // we only care about properties, that we are stored on the storage.
                    if (!DownloadStorage.areTheSame(old, updated)) {
                        scope.launch {
                            downloadStorage.update(action.download)
                        }
                        logger.debug("Updated download ${action.download.fileName} on the storage")
                    }
                }
            }
```

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
Amejia481: 👍

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadStorage.kt
Amejia481: We are using this function to validate if something has changed, between updates. We need to verify for each field that the storage cares about, because we don't overwhelm the storage, if there is not new info, maybe `DownloadStorage` is not a good place. What do you think?


```kotlin
            is DownloadAction.UpdateDownloadAction -> {
                val updated = action.download
                store.state.downloads[updated.id]?.let { old ->
                    // To not overwhelm the storage, we only send updates that are relevant,
                    // we only care about properties, that we are stored on the storage.
                    if (!DownloadStorage.areTheSame(old, updated)) {
                        scope.launch {
                            downloadStorage.update(action.download)
                        }
                        logger.debug("Updated download ${action.download.fileName} on the storage")
                    }
                }
            }
```

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadStorage.kt
csadilek: Sure, but I don't think how it would be confused with `equals`, but you can simply rename the method too and keep it in the storage class. My comment was just about `areTheSame` not being a good name on the storage. When I read the tests I found this very confusing. `compareDownloads` etc. would have been much cleared, imo.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadStorage.kt
csadilek: OK, see comment below. I think we can just choose a clearer name and keep everything as-is otherwise.

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadMiddleware.kt
Amejia481: 👍

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadStorage.kt
Amejia481: Any suggestions for the new name?

Amejia481: What about `DownloadStorage.equals(a, b)` ?

Amejia481:
New changes are in https://github.com/mozilla-mobile/android-components/pull/8197/commits/c1a18b11dc38cde3d77e7e9f6183b58fc5604455

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
Amejia481: Added a new `RestoreDownloadAction` makes things clear :)

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/DownloadStateReducer.kt
```diff
@@ -25,6 +25,8 @@ internal object DownloadStateReducer {
             is DownloadAction.RemoveAllDownloadsAction -> {
                 state.copy(downloads = emptyMap())
             }
+            DownloadAction.RestoreDownloadsStateAction -> state
```
csadilek, line 28: nit: no `is`

We should also add a quick test case that this action doesn't affect the state :) e.g

```
    @Test
    fun `RestoreDownloadsStateAction does nothing`() {
        val store = BrowserStore(BrowserState())

        val state = store.state
        store.dispatch(DownloadAction.RestoreDownloadsStateAction).joinBlocking()
        assertSame(store.state, state)
    }
```

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadMiddleware.kt
```diff
@@ -7,35 +7,116 @@ package mozilla.components.feature.downloads
...
+            is DownloadAction.UpdateDownloadAction -> updateDownload(action.download, context.store)
+
+            is DownloadAction.RestoreDownloadsStateAction -> restoreDownloads(context.store)
+
+            is DownloadAction.RestoreDownloadStateAction -> {
+                next(action)
+                sendDownloadIntent(action.download)
+            }
+        }
+        if (action !is DownloadAction.AddDownloadAction) {
```
csadilek, line 74: That's a bug here now. `RestoreDownloadStateAction` would cause next to be called twice. I would handle this differently and have a `preProcess` function that returns whether nor not `next` should be called e.g. whether or not the original action should hit the store.

Here's an example:  https://github.com/mozilla-mobile/android-components/blob/master/components/feature/readerview/src/main/java/mozilla/components/feature/readerview/ReaderViewMiddleware.kt#L55

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/DownloadStateReducer.kt
Amejia481: 👍

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadMiddleware.kt
Amejia481: Thanks for catching this 👍

Amejia481: What do you think about 
```kotlin
        if (requireNext(context, action)) {
            next(action)
            postNextProcess(action)
        }
```

csadilek: I find that a bit confusing, to be honest :).

csadilek: So, I would check actions individually or do what we did in the `ReaderViewMiddleware` so we follow the same pattern.

Amejia481:
Pr update https://github.com/mozilla-mobile/android-components/pull/8197/commits/3b9703e1b68f82bd3357e631cf6e51ac6c73a538 :)

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt
```diff
@@ -625,7 +630,15 @@ abstract class AbstractFetchDownloadService : Service() {
     internal fun copyInChunks(downloadJobState: DownloadJobState, inStream: InputStream, outStream: OutputStream) {
         val data = ByteArray(CHUNK_SIZE)
         logger.debug("starting copyInChunks ${downloadJobState.state.url}" +
-                " currentBytesCopied ${downloadJobState.currentBytesCopied}")
+                " currentBytesCopied ${downloadJobState.state.currentBytesCopied}")
+
+        val throttleUpdatedDownload = throttleLatest<Long>(
```
csadilek, line 635: nit: `throttleUpdateDownload`

samples/browser/src/main/java/org/mozilla/samples/browser/SampleApplication.kt
```diff
@@ -58,7 +59,7 @@ class SampleApplication : Application() {
         GlobalScope.launch(Dispatchers.IO) {
             components.webAppManifestStorage.warmUpScopes(System.currentTimeMillis())
         }
-
+        components.store.dispatch(DownloadAction.RestoreDownloadsStateAction)
```
csadilek, line 62: nit: Was wondering if we should create a use case to `DownloadsUseCases` for this.

components/feature/downloads/src/test/java/mozilla/components/feature/downloads/DownloadMiddlewareTest.kt
```diff
@@ -8,34 +8,181 @@ import android.app.DownloadManager.EXTRA_DOWNLOAD_ID
...
+        val downloadInTheStore = store.state.downloads.getValue(download.id)
+
+        assertEquals(download, downloadInTheStore)
+
+        var updatedDownload = download.copy(url = "updatedURL")
+        store.dispatch(DownloadAction.UpdateDownloadAction(updatedDownload)).joinBlocking()
+
+        verify(downloadStorage).update(updatedDownload)
+
+        updatedDownload = download.copy(url = "updatedURLAgain")
```
csadilek, line 153: nit: use a test case that can happen...this seems a bit contrived?

components/feature/downloads/src/test/java/mozilla/components/feature/downloads/DownloadMiddlewareTest.kt
```diff
@@ -8,34 +8,181 @@ import android.app.DownloadManager.EXTRA_DOWNLOAD_ID
...
+        val download = DownloadState("https://mozilla.org/download", destinationDirectory = "")
+        store.dispatch(DownloadAction.AddDownloadAction(download)).joinBlocking()
+
+        store.dispatch(DownloadAction.RemoveAllDownloadsAction).joinBlocking()
+
+        verify(downloadStorage).removeAllDownloads()
+    }
+
+    @Test
+    fun `UpdateDownloadAction MUST update the storage when changes are meaningful`() = runBlockingTest {
```
csadilek, line 127: nit: when changes are needed :)

samples/browser/src/main/java/org/mozilla/samples/browser/SampleApplication.kt
Amejia481: Agree, it will hide the implementation details!

components/feature/downloads/src/test/java/mozilla/components/feature/downloads/DownloadMiddlewareTest.kt
Amejia481: 👍

components/feature/downloads/src/test/java/mozilla/components/feature/downloads/DownloadMiddlewareTest.kt
Amejia481: updated!

Amejia481:
bors r=csadilek

PR #8198, opened 2020-08-21
For #7134 - Dismiss the current select login prompt request in LoginPicker

ekager:
---

@jonalmeida i couldn't repro the crash myself but I realized I could have been dismissing the wrong request so I'm hoping this is the fix 

### Pull Request checklist

### After merge


jonalmeida:
@ekager I found a consistent set of steps to reproduce the crash I had mentioned on the other PR. I'm going to try and reproduce them with this patch as well. :)

### Steps to reproduce
  1. Go to a site.
  2. Click on the password text field.
  3. Click on a saved password to autofill (click nothing else).
  4. Open Settings from kebab menu.
  5. Go back to the browser.
  6. Crash! :boom:

ekager:
Ah I can repro. I'll update when I figure out what's happening

jonalmeida:
Hmm it still crashes for me, but I think this is something we should fix.

I noticed there is a discrepancy `consumePromptFrom` which looks for the request from the selected tab OR the custom tab. Would that might mean we have a login prompt in a custom tab that we never never dismiss since we don't support logins there?

ekager:
It looks like when you return from settings, we get two simultaneous calls to dismiss the prompt from the `dismissPromptScope` and `sessionPromptScope` which causes the crash. We could try/catch the dismiss, but it seems better to just not try to dismiss at all if we have already tried. What do you think?

ekager:
Oh and to answer your other question - this does work in custom tabs.

ekager:
Okay this is ready for review @Amejia481 . I tried to consolidate the two scopes that dismiss prompts into one and switched to using the `url` instead of the `progress`. I don't think we can use the progress because it would require us to `filter` and I was having a hard time thinking of a case with prompts that we shouldn't dismiss when the page changes. Let me know what you think (and sorry this is on a time crunch!)

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/login/LoginPicker.kt
```diff
@@ -34,21 +33,13 @@ internal class LoginPicker(
...
 
-    @VisibleForTesting
-    internal var currentRequest: PromptRequest.SelectLoginPrompt? = null
-
     override fun onLoginSelected(login: Login) {
-        (currentRequest as PromptRequest.SelectLoginPrompt).onConfirm(login)
-        currentRequest = null
+        store.consumePromptFrom(sessionId) {
+            if (it is PromptRequest.SelectLoginPrompt) it.onConfirm(login)
+        }
```
Amejia481, line 42: Should we include tests for these new behaviour?

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/login/LoginPicker.kt
```diff
@@ -58,12 +49,9 @@ internal class LoginPicker(
     }
 
     fun dismissCurrentLoginSelect() {
-        if (currentRequest != null) {
-            store.consumePromptFrom(sessionId) {
-                (it as PromptRequest.SelectLoginPrompt).onDismiss()
-            }
+        store.consumePromptFrom(sessionId) {
+            if (it is PromptRequest.SelectLoginPrompt) it.onDismiss()
```
Amejia481, line 53: Should we include tests for these new behaviour?

Amejia481:
There are some test failing on `build-feature-prompts` let's take a look :)

components/feature/prompts/src/main/java/mozilla/components/feature/prompts/login/LoginPicker.kt
```diff
@@ -57,13 +48,15 @@ internal class LoginPicker(
...
-            store.consumePromptFrom(sessionId) {
-                (it as PromptRequest.SelectLoginPrompt).onDismiss()
+    @Suppress("TooGenericExceptionCaught")
+    fun dismissCurrentLoginSelect(promptRequest: PromptRequest.SelectLoginPrompt? = null) {
+        try {
+            promptRequest?.let { it.onDismiss() } ?: store.consumePromptFrom(sessionId) {
+                if (it is PromptRequest.SelectLoginPrompt) it.onDismiss()
             }
+        } catch (e: RuntimeException) {
+            // This prompt has already been dismissed
```
Amejia481, line 58: nit: let's log the exception

ekager:
bors r=Amejia481

Issue #8200, opened 2020-08-22
"users" in add-ons manager should be renamed to "reviews"


cadeyrn:
"users" is misleading, see https://github.com/mozilla-mobile/fenix/issues/11570, already approved for Fenix (https://github.com/mozilla-mobile/fenix/issues/11570#issuecomment-665792416).

Mugurell:
To be tested in https://github.com/mozilla-mobile/fenix/issues/11570

PR #8201, opened 2020-08-22
for #8200 - renamed "users" to "reviews" in add-on manager

cadeyrn:
string change because the old string was misleading.


components/feature/addons/src/main/res/layout/mozac_feature_addons_item.xml
```diff
@@ -102,7 +102,7 @@
                 android:layout_marginStart="6dp"
                 android:textSize="12sp"
                 android:layout_gravity="center_vertical"
-                tools:text="Users: 591,642" />
+                tools:text="Reviews: 591,642" />
```
Mugurell, line 105: Nice catch!

Mugurell (approved):
LGTM, thanks!

Mugurell:
bors r=Mugurell

PR #8213, opened 2020-08-24
Avoid double layout in WidgetSiteItemView

NotWoods:
The parent class extends constraint layout, so the XML layout resource should use a `<merge>` tag.


pocmo (approved):
bors r+

PR #8226, opened 2020-08-25
Closes #8209 - Use new onTouchEventForResult GV api

### LINKED ISSUE #8209: GeckoView Nighly/Beta update failing
https://firefox-ci-tc.services.mozilla.com/tasks/GpGq8GTFS7i4fnH5iFpfIQ/runs/0/logs/https%3A%2F%2Ffirefox-ci-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FGpGq8GTFS7i4fnH5iFpfIQ%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log

There seems to be an API breakage in the latest Nightly:

```
e: /build/android-components/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/NestedGeckoView.kt: (57, 58): Return type of 'onTouchEventForResult' is not a subtype of the return type of the overridden member '@NonNull public open fun onTouchEventForResult(@NonNull p0: MotionEvent): GeckoResult<Int!> defined in org.mozilla.geckoview.GeckoView'
e: /build/android-components/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/NestedGeckoView.kt: (106, 16): Type mismatch: inferred type is GeckoResult<Int!> but Int was expected
```

Mugurell:
Changes:
- previously NestedGV#onTouchEvent event would call onTouchEventForResult where
we would make possible the integration of GV in CoordinatorLayout.
Since NestedGV#onTouchEvent does not anymore call onTouchEventForResult I've
moved all code in onTouchEvent and then call onTouchEventForResult only for
ACTION_DOWN events as snorp recommended.
- we now must wait for onTouchEventForResult#GeckoResult<Int> and only then
call startNestedScroll(..) since otherwise when executing that method in
BrowserToolbarBottomBehavior we would return false since the MotionEvent is
at that time unhandled.
- only dispatch startNestedScroll and dispatchNestedPreScroll if GV returned
INPUT_RESULT_HANDLED.

In testing I saw most of the times GeckoResult resolves in < 5ms with bigger
timeouts showing a logarithmic growth. As such I think the change should be
imperceptible to users.

[video showing the same as before behavior in Fenix](https://drive.google.com/file/d/1WiQ5bXb-dBxYDA03snYSkxApxfhr7hgQ/view?usp=sharing)

---

### After merge

### DIFF
--- a/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/NestedGeckoView.kt
+++ b/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/NestedGeckoView.kt
@@ -10,8 +10,11 @@ import androidx.annotation.VisibleForTesting
 import androidx.core.view.NestedScrollingChild
 import androidx.core.view.NestedScrollingChildHelper
 import androidx.core.view.ViewCompat
+import kotlinx.coroutines.launch
 import mozilla.components.concept.engine.EngineView
+import mozilla.components.support.ktx.android.view.toScope
 import org.mozilla.geckoview.GeckoView
+import org.mozilla.geckoview.PanZoomController.INPUT_RESULT_HANDLED
 import org.mozilla.geckoview.PanZoomController.INPUT_RESULT_UNHANDLED
 
 /**
@@ -25,7 +28,7 @@ import org.mozilla.geckoview.PanZoomController.INPUT_RESULT_UNHANDLED
  * https://github.com/takahirom/webview-in-coordinatorlayout
  */
 
-@Suppress("TooManyFunctions")
+@Suppress("TooManyFunctions", "ClickableViewAccessibility")
 open class NestedGeckoView(context: Context) : GeckoView(context), NestedScrollingChild {
 
     @VisibleForTesting
@@ -42,6 +45,8 @@ open class NestedGeckoView(context: Context) : GeckoView(context), NestedScrolli
     @VisibleForTesting
     internal var childHelper: NestedScrollingChildHelper = NestedScrollingChildHelper(this)
 
+    private val coroutineScope = toScope()
+
     /**
      * Integer indicating how user's MotionEvent was handled.
      *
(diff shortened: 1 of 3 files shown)


Mugurell:
Everything works as before for the bottom toolbar but I'm investigating some issues for when the toolbar is at top.

Mugurell:
Ready for review 😃

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/NestedGeckoView.kt
```diff
@@ -87,23 +83,46 @@ open class NestedGeckoView(context: Context) : GeckoView(context), NestedScrolli
...
-    internal fun handleEvent(event: MotionEvent): Int {
-        return super.onTouchEventForResult(event)
+    @VisibleForTesting
+    internal fun updateInputResult(event: MotionEvent) {
+        rootView.toScope().launch {
+            super.onTouchEventForResult(event).await()?.let {
+                inputResult = it
+                startNestedScroll(ViewCompat.SCROLL_AXIS_VERTICAL)
+            }
+        }
```
pocmo, line 125: It looks like this gets called from `onTouchEvent()` and therefore potentially quite often? Under the hood `toScope()` always creates a new `MainScope` and registers an `OnAttachStateChangeListener` to know when to cancel it. So we may end up creating a lot of them. Maybe we can create a scope from the View only once inside this class?

Mugurell:
Regarding the lint error - `Custom view NestedGeckoView overrides onTouchEvent but not performClick` based on https://developer.android.com/guide/topics/ui/accessibility/custom-views#custom-touch-events looks like it's safe/ok to not call performClick. GeckoView is not indended to be clicked but all of the clickable items in the webpage.
Previously we didn't override `onTouchEvent` and the `GeckoView` did not performed clicks.
Did a quick test with talkback enabled and saw the same behavior in a dev build with this patch as in the current Nightly.

Based on the above I'll look into suppressing the check.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/NestedGeckoView.kt
```diff
@@ -42,6 +45,8 @@ open class NestedGeckoView(context: Context) : GeckoView(context), NestedScrolli
     @VisibleForTesting
     internal var childHelper: NestedScrollingChildHelper = NestedScrollingChildHelper(this)
 
+    private val coroutineScope = rootView.toScope()
```
pocmo, line 48: One more question: Do we need to use `rootView` here, or could we just use `toScope()` to create a scope in this view? When *this* view gets detached then we want to cancel the scope, I assume. Since the rootView is outside of fragments (where this view here may be in), it may live much longer than anything we do here.

Mugurell: Indeed. Thanks!

pocmo (approved):
👍

PR #8221, opened 2020-08-25
Merge day!

pocmo:
This only updates the release version. Unfortunately updating beta and nightly is blocked on #8209.

PR #8239, opened 2020-08-26
Update GeckoView (Beta) to 81.0.20200825191644.

pocmo:
Follow-up from #8221. We got a new beta build and now we can update safely. :)


pocmo:
bors r=Amejia481

PR #8250, opened 2020-08-26
Close #7476: Persists lastAccess to SnapshotSerializer 

### LINKED ISSUE #7476: Add lastAccess property to SnapshotSerializer
We should try to serialize the last used property so that we don't wrongfully update the sync servers with incorrect timestamps on a cold-boot restore.

See original comment here: https://github.com/mozilla-mobile/android-components/pull/7277#issuecomment-644250888

jonalmeida:
---

Close #7476 

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/engine/middleware/LastAccessMiddleware.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/engine/middleware/LastAccessMiddleware.kt
@@ -1,0 +1,36 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.session.engine.middleware
+
+import mozilla.components.browser.state.action.BrowserAction
+import mozilla.components.browser.state.action.TabListAction
+import mozilla.components.browser.state.action.LastAccessAction
+import mozilla.components.browser.state.state.BrowserState
+import mozilla.components.browser.state.state.TabSessionState
+import mozilla.components.lib.state.Middleware
+import mozilla.components.lib.state.MiddlewareContext
+
+/**
+ * [Middleware] that handles updating the [TabSessionState.lastAccess] when a tab is selected.
+ */
+internal class LastAccessMiddleware : Middleware<BrowserState, BrowserAction> {
+    override fun invoke(
+        context: MiddlewareContext<BrowserState, BrowserAction>,
+        next: (BrowserAction) -> Unit,
+        action: BrowserAction
+    ) {
+        next(action)
+
+        if (action is TabListAction.SelectTabAction) {
+            context.dispatchUpdateActionForId(action.tabId)
+        } else if (action is TabListAction.AddTabAction && action.select) {
+            context.dispatchUpdateActionForId(action.tab.id)
+        }
+    }
+
+    private fun MiddlewareContext<BrowserState, BrowserAction>.dispatchUpdateActionForId(id: String) {
+        dispatch(LastAccessAction.UpdateLastAccessAction(id, System.currentTimeMillis()))
+    }
+}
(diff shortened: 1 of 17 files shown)


components/browser/session/src/main/java/mozilla/components/browser/session/engine/middleware/LastAccessMiddleware.kt
```diff
@@ -0,0 +1,35 @@
...
+    override fun invoke(
+        context: MiddlewareContext<BrowserState, BrowserAction>,
+        next: (BrowserAction) -> Unit,
+        action: BrowserAction
+    ) {
+        if (action is TabListAction.SelectTabAction) {
+            context.dispatchUpdateActionForId(action.tabId)
+        } else if (action is TabListAction.AddTabAction && action.select) {
+            context.dispatchUpdateActionForId(action.tab.id)
+        }
```
pocmo, line 28: nit: Just a nit... should we first select/add (invoking `next`) and *then* update the time? Especially the "add tab" case. How does this work if you update before the tab is added?

components/browser/session/src/test/java/mozilla/components/browser/session/engine/middleware/LastAccessMiddlewareTest.kt
```diff
@@ -0,0 +1,80 @@
...
+        val captor = argumentCaptor<TabSessionAction.UpdateLastAccessAction>()
+        val tab = createTab("https://mozilla.org", id = "456")
+
+        middleware.invoke(context, { nextInvoked = true }, TabListAction.AddTabAction(tab, true))
+
+        verify(store).dispatch(captor.capture())
+
+        assertEquals("456", captor.value.tabId)
+        assertTrue(captor.value.lastAccess != 0L)
+        assertTrue(nextInvoked)
```
pocmo, line 64: See question above about updating the time before the tab is added. I'd recommend testing with an actual store instance, no need to mock that in test if you want to test a middleware.

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -115,6 +115,22 @@ sealed class TabListAction : BrowserAction() {
     object RemoveAllNormalTabsAction : TabListAction()
 }
 
+/**
+ * [BrowserAction] implementations related to updating the [TabSessionState] inside [BrowserState].
+ */
+sealed class TabSessionAction : BrowserAction() {
```
pocmo, line 121: nit: TabSessionAction is very generic and may be confusing if there's nothing else in here? Others like `TrackingProtectionAction` are more specific. I fear that this action group may end up becoming a dumping ground. :)

jonalmeida: This was my fear as well. I was hesitant to put this under `TabListAction` since it wasn't acting on the list, and it didn't fit into `ContentAction` as well since it isn't content.

Maybe I should change the name to something else, like `TabSelectionAction`? 🤔

components/browser/session/src/main/java/mozilla/components/browser/session/engine/middleware/LastAccessMiddleware.kt
jonalmeida: I haven't noticed any side-effects in my testing, but I think your suggestion makes a lot more sense than the current behaviour.

components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
csadilek: It's a bit of a special case, maybe just `LastAccessAction` and `UpdateLastAccessAction` to prevent from becoming a dumping ground?

jonalmeida:
Thanks, I've addressed the comments and squash the changes into one!

jonalmeida:
I'd like to re-run this PR after https://github.com/mozilla-mobile/android-components/pull/8268 lands. I'm certain I've broken some more tests since I've changed the behaviour from reverting the changes to `SelectTabAction` where we altered the tests when that change originally landed.

jonalmeida:
The CI changes that I wanted to land first before landing this PR will [only happen on Monday](https://github.com/mozilla-mobile/android-components/pull/8268#pullrequestreview-477529803), which may be too late to get this into the next release in time.

My concern is that we may have broken tests in other components that depend on this change when [landing the original feature](https://github.com/mozilla-mobile/android-components/pull/7277) which we have not fixed correctly and CI may not run have run on them. My plan of action is to run all the component tests locally, correct them if needed, and only then merge this PR.

jonalmeida:
From local testing the `ReaderViewFeatureTest` changes that were changed in the original PR needed to be changed. I think I'm more comfortable with landing this change now.

jonalmeida:
bors r=csadilek

PR #8256, opened 2020-08-27
 Docs update (20200827-121225) [ci skip]

MickeyMoz:


pocmo (approved):
bors r+

escapewindow:
@MickeyMoz @pocmo this broke android-l10n-tooling (~700 files changed led to a too-long command line), so any strings in this PR have not been picked up.

- are there any strings that android-l10n-tooling should have picked up from this PR?
- if so, what does it mean if we don't get those addressed?
- do we plan on landing any other PRs with tons of file changes?

We're looking at moving the input of files changed into an artifact rather than the command line to address this type of thing in the future. That may take us a while to implement, however, so we should figure out the short term fallout and fixes.

MihaiTabara:
@liuche confirmed on Slack there are no `strings.xml` files within this PR so its sounds like we dodged the bullet this time. 

The last two questions remain still. What would've happened if we did have l10n strings? Can we afford dropping some on the floor in an emergency case? Also how often should we we expect PRs at this size?

PR #8268, opened 2020-08-27
Find upstream and downstream dependencies when determining which components to build (fixes #8265)

### LINKED ISSUE #8265: Task graph does not rebuild components that are dependant on other components that have changed
For this PR, I've touched `browser-session` and `browser-state`, so I would have expected all components that have those two as a dependency to be rebuilt: https://github.com/mozilla-mobile/android-components/pull/8250

Instead, I only see a handful of tasks created after the graph has been pruned: https://firefoxci.taskcluster-artifacts.net/HB9ZUbeESpGtZX1b-F61xw/0/public/target-tasks.json

According to the logs, this is the full task graph before it's reduced and there are a lot of components missing in it: https://firefoxci.taskcluster-artifacts.net/HB9ZUbeESpGtZX1b-F61xw/0/public/full-task-graph.json

For example, [`feature-findinpage`](https://github.com/mozilla-mobile/android-components/blob/master/components/feature/findinpage/build.gradle#L31) has a dependency on `browser-state` (which I modified [in my PR](https://github.com/mozilla-mobile/android-components/pull/8250/files#diff-5b5de9adf69f383f48292ad741e73676)), so I would have expected that component to be rebuilt. However, I don't see it in the `full-task-graph.json` list to begin with, so that might be why it has been ignored?

cc: @mozbhearsum or @JohanLorenzo

mozbhearsum:
I tested this locally by rerunning the decision task for #8250, which ended up with a much lengthier list of affected components:
```
2020-08-27 15:30:29,758 - INFO - Affected components: browser-icons browser-menu browser-session browser-state browser-thumbnails concept-engine feature-accounts feature-addons feature-app-links feature-awesomebar feature-containers feature-contextmenu feature-customtabs feature-downloads feature-findinpage feature-intent feature-media feature-p2p feature-privatemode feature-prompts feature-pwa feature-readerview feature-search feature-session feature-sitepermissions feature-syncedtabs feature-tab-collections feature-tabs feature-toolbar samples-browser support-ktx support-migration support-utils support-webextensions
```

The implementation is not great -- we end up running ` ./gradlew X:dependencies --configuration implementation` for every single component. On my laptop this took about 7min - it may take longer in CI. This could be improved if someone can figure out a way to more quickly dump dependencies for a single component, or even better - dump the entire dependency tree.

### DIFF
--- a/taskcluster/ac_taskgraph/loader/build_config.py
+++ b/taskcluster/ac_taskgraph/loader/build_config.py
@@ -9,6 +9,7 @@ import os
 import re
 import subprocess
 
+from collections import defaultdict
 from copy import deepcopy
 from taskgraph.loader.transform import loader as base_loader
 from taskgraph.util.memoize import memoize
@@ -45,7 +46,7 @@ cached_deps = {}
 
 # Gradle is expensive to run. Cache the results for each component.
 @memoize
-def get_deps_for_component(component):
+def get_upstream_deps_for_component(component):
     """Return the full list of local upstream dependencies of a component."""
     deps = set()
     # Parsing output like this is not ideal but bhearsum couldn't find a way
@@ -56,15 +57,15 @@ def get_deps_for_component(component):
         if line.startswith("+--- project"):
             deps.add(line.split(" ")[2])
 
+    if deps:
+        logger.info("Found upstream dependencies for '%s': %s" % (component, " ".join(sorted(deps))))
+    else:
+        logger.info("No upstream dependencies found for '%s'" % component)
     cached_deps[component] = deps
-    # So far, we only have the direct local upstream dependencies. We need to look at
-    # each of those to find _their_ upstreams as well.
-    for d in deps.copy():
-        deps.update(get_deps_for_component(d))
 
     return deps
 
-def get_affected_components(files_changed, files_affecting_components):
+def get_affected_components(files_changed, files_affecting_components, component_dependencies, reverse_component_dependencies):
     affected_components = set()
 
     # First, find the list of changed components
@@ -82,9 +83,14 @@ def get_affected_components(files_changed, files_affecting_components):
             affected_components.update(components)
 
     # Finally, go through all of the affected components and recursively
-    # find their dependencies.
+    # find their upstream and downstream dependencies.
     for c in affected_components.copy():
-        affected_components.update(get_deps_for_component(c))
+        if component_dependencies[c]:
+            logger.info("Adding upstream dependencies for '%s': %s" % (c, " ".join(sorted(component_dependencies[c]))))
+            affected_components.update(component_dependencies[c])
+        if component_dependencies[c]:
+            logger.info("Adding downstream dependencies for '%s': %s" % (c, " ".join(sorted(reverse_component_dependencies[c]))))
+            affected_components.update(reverse_component_dependencies[c])
 
     return affected_components
 
@@ -93,24 +99,32 @@ def loader(kind, path, config, params, loaded_tasks):
     # Build everything unless we have an optimization strategy (defined below).
     files_changed = []
     affected_components = ALL_COMPONENTS
+    component_dependencies = defaultdict(set)
+    reverse_component_dependencies = defaultdict(set)
+
+    for component in sorted(get_components()):
+        name = component["name"]
+        for dep_component in get_upstream_deps_for_component(name):
+            component_dependencies[name].add(dep_component)
+            reverse_component_dependencies[dep_component].add(name)
 
     if params["tasks_for"] == "github-pull-request":
         logger.info("Processing pull request %s" % params["pull_request_number"])
         files_changed = get_files_changed_pr(params["base_repository"], params["pull_request_number"])
-        affected_components = get_affected_components(files_changed, config.get("files-affecting-components"))
+        affected_components = get_affected_components(files_changed, config.get("files-affecting-components"), component_dependencies, reverse_component_dependencies)
     elif params["tasks_for"] == "github-push":
         if params["base_rev"] in _GIT_ZERO_HASHES:
             logger.warn("base_rev is a zero hash, meaning there is no previous push. Building every component...")
         else:
             logger.info("Processing push for commit range %s -> %s" % (params["base_rev"], params["head_rev"]))
             files_changed = get_files_changed_push(params["base_repository"], params["base_rev"], params["head_rev"])
-            affected_components = get_affected_components(files_changed, config.get("files-affecting-components"))
+            affected_components = get_affected_components(files_changed, config.get("files-affecting-components"), component_dependencies, reverse_component_dependencies)
 
     logger.info("Files changed: %s" % " ".join(files_changed))
     if affected_components is ALL_COMPONENTS:
         logger.info("Affected components: ALL")
     else:
-        logger.info("Affected components: %s" % " ".join(affected_components))
+        logger.info("Affected components: %s" % " ".join(sorted(affected_components)))
 
     not_for_components = config.get("not-for-components", [])
     jobs = {
--- a/taskcluster/ac_taskgraph/loader/build_config.py
+++ b/taskcluster/ac_taskgraph/loader/build_config.py
@@ -44,28 +44,40 @@ def get_components_changed(files_changed):
 
 cached_deps = {}
 
-# Gradle is expensive to run. Cache the results for each component.
-@memoize
-def get_upstream_deps_for_component(component):
+def get_upstream_deps_for_components(components):
     """Return the full list of local upstream dependencies of a component."""
     deps = set()
+    cmd = ["./gradlew"]
+    for c in sorted(components):
+        cmd.extend(["%s:dependencies" % c, "--configuration", "implementation"])
     # Parsing output like this is not ideal but bhearsum couldn't find a way
     # to get the dependencies printed in a better format. If we could convince
     # gradle to spit out JSON that would be much better.
     # This is filed as https://github.com/mozilla-mobile/android-components/issues/7814
-    for line in subprocess.check_output(['./gradlew', '%s:dependencies' % component, '--configuration', 'implementation']).splitlines():
+    current_component = None
+    for line in subprocess.check_output(cmd).splitlines():
+        # If we find the start of a new component section, update our tracking variable
+        if line.startswith("Project"):
+            if deps:
+                logger.info("Found direct upstream dependencies for component '%s': %s" % (current_component, sorted(deps)))
+            else:
+                logger.info("No direct upstream dependencies found for component '%s'" % current_component)
+            yield current_component, deps
+            deps = set()
+            current_component = line.split(":")[1]
+
+        # If we find a new local dependency, add it.
         if line.startswith("+--- project"):
             deps.add(line.split(" ")[2])
 
     if deps:
-        logger.info("Found upstream dependencies for '%s': %s" % (component, " ".join(sorted(deps))))
+        logger.info("Found direct upstream dependencies for component '%s': %s" % (current_component, sorted(deps)))
     else:
-        logger.info("No upstream dependencies found for '%s'" % component)
-    cached_deps[component] = deps
+        logger.info("No direct upstream dependencies found for component '%s'" % current_component)
+    yield current_component, deps
 
-    return deps
 
-def get_affected_components(files_changed, files_affecting_components, component_dependencies, reverse_component_dependencies):
+def get_affected_components(files_changed, files_affecting_components, upstream_component_dependencies, downstream_component_dependencies):
     affected_components = set()
 
     # First, find the list of changed components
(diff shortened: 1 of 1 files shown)


mozbhearsum:
It looks like this decision task took 10min, which should be relatively constant since we're looking at dependencies for all components (instead of just directly modified + their upstreams). As a comparison, https://github.com/mozilla-mobile/android-components/pull/8250 took ~3min in its decision task. So it's a big relative increase, but not a ridiculously long time. If there's a way to more quickly get the dependency tree, I'd like to use it, but I haven't found one.

@jonalmeida - can you sanity check the new component list for your PR in the previous comment? @JohanLorenzo - you should probably review the actual code.

jonalmeida:
@mozbhearsum thanks for the quick turnaround! I did a quick check by looking for all the component's build.gradle that had any of the two lines:
 -  `implementation project(:browser-session)`
 -  `implementation project(:browser-state)`

I de-duped that list and came up with a nearly identical list to the one you posted above. 🙂 Mine was shorter since the there are other changes in my PR that would have added more dependencies that would include more.

> It looks like this decision task took 10min, which should be relatively constant since we're looking at dependencies for all components (instead of just directly modified + their upstreams). As a comparison, #8250 took ~3min in its decision task. So it's a big relative increase, but not a ridiculously long time. If there's a way to more quickly get the dependency tree, I'd like to use it, but I haven't found one.

I wonder if it would make sense to store the full-graph.json as part of the toolchain task so that we don't have the generate it all the time? It sounds like that would be the source of the slow-down?

We could re-generate the list when we rebuild the toolchain task to keep it up-to-date.

I'm leaving the PR up to @JohanLorenzo to sign off on otherwise.

mozbhearsum:
> @mozbhearsum thanks for the quick turnaround! I did a quick check by looking for all the component's build.gradle that had any of the two lines:
> 
>     * `implementation project(:browser-session)`
> 
>     * `implementation project(:browser-state)`
> 
> 
> I de-duped that list and came up with a nearly identical list to the one you posted above. slightly_smiling_face Mine was shorter since the there are other changes in my PR that would have added more dependencies.

Hm, I ran the same thing that the decision task did -- so I would expect it would be exactly the same...what was mine missing?

> > It looks like this decision task took 10min, which should be relatively constant since we're looking at dependencies for all components (instead of just directly modified + their upstreams). As a comparison, #8250 took ~3min in its decision task. So it's a big relative increase, but not a ridiculously long time. If there's a way to more quickly get the dependency tree, I'd like to use it, but I haven't found one.
> 
> I wonder if it would make sense to store the full-graph.json as part of the toolchain task so that we don't have the generate it all the time? It sounds like that would be the source of the slow-down?
> 
> We could re-generate the list when we rebuild the toolchain task to keep it up-to-date.

Hrm. This, or something like it, might be possible. To clarify, let him double check a couple of assumptions:
* The toolchain task in question is `toolchain-linux64-android-gradle-dependencies`
* Dependencies between components don't change unless that task runs

If both of those are true, I don't see why we can't store the dependency graph as an artifact of that task. The tricky part will be finding the correct version of it, particular for PRs, which may be based on an older revision that the tip of `master`. I suspect there's a way to do this gracefully, but I'll have to consult with someone else to check on it.

In any case, I think it would be good to land this initial fix (I'm a little shocked this oversight hasn't caused bustage yet, to be honest) and come back to optimize it later (hopefully soon).

jonalmeida:
> Hm, I ran the same thing that the decision task did -- so I would expect it would be exactly the same...what was mine missing?

Sorry, I wasn't clear, it was my list that was lacking because I was lazy in my testing. 🙂  Your list looked right!

> If both of those are true, I don't see why we can't store the dependency graph as an artifact of that task. The tricky part will be finding the correct version of it, particular for PRs, which may be based on an older revision that the tip of master. I suspect there's a way to do this gracefully, but I'll have to consult with someone else to check on it.

Yes, both of those are true, but Johan will know better.

> In any case, I think it would be good to land this initial fix (I'm a little shocked this oversight hasn't caused bustage yet, to be honest) and come back to optimize it later (hopefully soon).

I agree, let's land this asap. I'm certain #8250 would have been the first change to break the build and discover this bug. I'll leave it up to strongly typed languages and good test coverage for the lack of current bustage. 😅

taskcluster/ac_taskgraph/loader/build_config.py
```diff
@@ -93,24 +99,32 @@ def loader(kind, path, config, params, loaded_tasks):
     # Build everything unless we have an optimization strategy (defined below).
     files_changed = []
     affected_components = ALL_COMPONENTS
+    component_dependencies = defaultdict(set)
+    reverse_component_dependencies = defaultdict(set)
+
+    for component in sorted(get_components()):
+        name = component["name"]
+        for dep_component in get_upstream_deps_for_component(name):
```
JohanLorenzo, line 107: I was trying to figure out what line could have increased the duration that much, and I found it 😁 

I read the previous proposal about baking this in `toolchain-linux64-android-gradle-dependencies`. I've thought about it out loud to understand the implications of such a change. While I believe the happy path easily works, supporting edge cases will be tough. I'm going to expand on this in the summary of my review. 

In the meantime, I believe there is a lower-hanging fruit. If I'm not mistaken, every time we run gradle, we spend several seconds initializing everything, even though the task we're interested in takes less than a second. My proposal is: instead of running `./gradlew X:dependencies --configuration implementation` a hundred times, we can craft a single command line that generates everything at once. I tried with this example, and it worked fine locally: `./gradlew browser-session:dependencies --configuration implementation browser-menu:dependencies --configuration implementation` (You need to repeat `--configuration implementation` for each component, otherwise you end up with an output that is too verbose).

Pros: 
 * We know we have to process all components, now. So it's easy to craft one big command.
 * We shave off the init steps that last several seconds.

Cons: 
 * The command line may be too large for gradle to comprehend. We may need to split it into 2
 * We have to make a smarter parser since the output will contain the whole tree.

I wouldn't be surprised if we decision task would take less than 7 minutes to run (instead of 10). That said, we have to test this out. I see you already have some baseline results and a testing protocol. I'm happy to defer this testing to you if you have everything set up. Otherwise, I can take care of it. What do you think @mozbhearsum ?

JohanLorenzo (commented):
Thank you all for your input, it enabled me to think about this problem more thoroughly. 

I agree on landing this quickly. Let's aim for Monday. Today looks quite risky to me because of the complexity of this patch and the induced duration of the decision task. 

I like the proposal about baking the dependency graph in `toolchain-linux64-android-gradle-dependencies` (aka `toolchain` below). I think it works in the happy path, but supporting edge cases sounds tricky. Here's what I think: 

 * Happy path: the decision task knows no `build.gradle` file was modified so the dependency tree hasn't changed. It looks up the last `toolchain` task and find the processed dependency graph. The decision task uses it. End of story, many minutes saved 👍 
 * One `build.gradle` file has changed. The decision task doesn't reuse the output of the last `toolchain` task, because a new one will be spun up. The decision task will have to process the dependency tree (just like it does now) to create the right graph. Then the toolchain task will process the dependency tree once more to output the artifact. That sounds suboptimal.
 * No `build.gradle` file has changed, but the previous push did change it and the `toolchain` isn't completed yet. I'm unsure how we can simply fix this problem. I worked on spec'ing a similar behavoir in https://github.com/mozilla-releng/releng-rfcs/pull/32. In my opinion, we should fix this in mozilla-central first and then let android-components get the changes. That said, I'm unsure whether this project is a priority because of the last reorg.

So, in my opinion, let's generate the dependency tree in the decision task and optimize it. I think I have a low hanging fruit. I described it below. Please let me know what you all think 🙂

taskcluster/ac_taskgraph/loader/build_config.py
```diff
@@ -82,9 +83,14 @@ def get_affected_components(files_changed, files_affecting_components):
...
 
     # Finally, go through all of the affected components and recursively
-    # find their dependencies.
+    # find their upstream and downstream dependencies.
     for c in affected_components.copy():
-        affected_components.update(get_deps_for_component(c))
+        if component_dependencies[c]:
+            logger.info("Adding upstream dependencies for '%s': %s" % (c, " ".join(sorted(component_dependencies[c]))))
+            affected_components.update(component_dependencies[c])
+        if component_dependencies[c]:
```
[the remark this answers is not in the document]
mozbhearsum: Whoops, good catch!

taskcluster/ac_taskgraph/loader/build_config.py
```diff
@@ -82,9 +83,14 @@ def get_affected_components(files_changed, files_affecting_components):
             affected_components.update(components)
 
     # Finally, go through all of the affected components and recursively
-    # find their dependencies.
+    # find their upstream and downstream dependencies.
     for c in affected_components.copy():
-        affected_components.update(get_deps_for_component(c))
+        if component_dependencies[c]:
+            logger.info("Adding upstream dependencies for '%s': %s" % (c, " ".join(sorted(component_dependencies[c]))))
```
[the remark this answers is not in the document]
mozbhearsum: This is a good idea, let's do it.

taskcluster/ac_taskgraph/loader/build_config.py
```diff
@@ -56,15 +57,15 @@ def get_deps_for_component(component):
         if line.startswith("+--- project"):
             deps.add(line.split(" ")[2])
 
+    if deps:
+        logger.info("Found upstream dependencies for '%s': %s" % (component, " ".join(sorted(deps))))
```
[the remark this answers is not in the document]
mozbhearsum: Yeah, that's probably a good idea.

taskcluster/ac_taskgraph/loader/build_config.py
mozbhearsum: To be clear: you're suggesting we run one gradle command for all (or as many as we can) components at once, instead of one at a time? I think that makes sense, and might greatly reduce the pressure for any additional optimization. I'll dig into it more.

mozbhearsum: I tried this, and it basically reduces gradle time to nothing on my laptop -- like, <5s to get the entire tree.

mozbhearsum:
For the record - the decision task for the push I just did took less than 3min - so I don't think we'll need any further optimization here, unless there's an issue with the current patch that I missed.

taskcluster/ac_taskgraph/loader/build_config.py
```diff
@@ -9,6 +9,7 @@
 import re
 import subprocess
 
+from collections import defaultdict
 from copy import deepcopy
 from taskgraph.loader.transform import loader as base_loader
 from taskgraph.util.memoize import memoize
```
JohanLorenzo, line 15: Nit: `memoize is not used anymore, I think.

taskcluster/ac_taskgraph/loader/build_config.py
```diff
@@ -43,28 +44,40 @@ def get_components_changed(files_changed):
...
-    cached_deps[component] = deps
-    # So far, we only have the direct local upstream dependencies. We need to look at
-    # each of those to find _their_ upstreams as well.
-    for d in deps.copy():
-        deps.update(get_deps_for_component(d))
+    if deps:
+        logger.info("Found direct upstream dependencies for component '%s': %s" % (current_component, sorted(deps)))
+    else:
+        logger.info("No direct upstream dependencies found for component '%s'" % current_component)
+    yield current_component, deps
```
JohanLorenzo, line 77: Good idea about yielding results here.

taskcluster/ac_taskgraph/loader/build_config.py
```diff
@@ -43,28 +44,40 @@ def get_components_changed(files_changed):
...
+        cmd.extend(["%s:dependencies" % c, "--configuration", "implementation"])
     # Parsing output like this is not ideal but bhearsum couldn't find a way
     # to get the dependencies printed in a better format. If we could convince
     # gradle to spit out JSON that would be much better.
     # This is filed as https://github.com/mozilla-mobile/android-components/issues/7814
-    for line in subprocess.check_output(['./gradlew', '%s:dependencies' % component, '--configuration', 'implementation']).splitlines():
+    current_component = None
+    for line in subprocess.check_output(cmd).splitlines():
+        # If we find the start of a new component section, update our tracking variable
+        if line.startswith("Project"):
```
JohanLorenzo, line 60: The parser you wrote is much simpler than the one I had in mind. Excellent! 👍

JohanLorenzo (approved):
Wonderful! This looks great to me!

taskcluster/ac_taskgraph/loader/build_config.py
mozbhearsum: I'm just glad gradle uses some basic ascii formatting to help denote sections :laughing:

mozbhearsum:
bors r=JohanLorenzo

Issue #8009, opened 2020-08-05
Catch exceptions in Context.hasCamera


csadilek:
We fixed this for the QR scanner in https://github.com/mozilla-mobile/android-components/pull/7912, but we also have a generic `Context.hasCamera` check that can crash on some devices:
https://github.com/mozilla-mobile/android-components/pull/7912

This is used by the `SearchFragment` in Fenix. Let's just fall back and return `false` in case the check crashes.

PR #8284, opened 2020-08-28
Issue #8009: Handle all throwables in hasCamera check

csadilek:
We're seeing some devices that throw `AssertionErrors` here! 💯 🤦 

Let's not crash in those cases either and return false:
https://github.com/mozilla-mobile/fenix/issues/13956


csadilek:
bors r=Amejia481

Issue #8271, opened 2020-08-27
Test failure RestoreDownloadsState MUST populate the store with items in the storage


Archaeopteryx:
From https://github.com/mozilla-mobile/android-components/pull/8197

CC @Amejia481 @csadilek 

Log: https://treeherder.mozilla.org/logviewer.html#?job_id=314222448&repo=android-components

[task 2020-08-27T22:17:08.550Z]   TEST: RestoreDownloadsState MUST populate the store with items in the storage
[task 2020-08-27T22:17:08.550Z]     [Robolectric] mozilla.components.feature.downloads.DownloadMiddlewareTest.RestoreDownloadsState MUST populate the store with items in the storage: sdk=28; resources=binary
[task 2020-08-27T22:17:08.647Z]     I/MonitoringInstr: Instrumentation started!
[task 2020-08-27T22:17:08.647Z]     I/MonitoringInstr: Setting context classloader to 'org.robolectric.internal.bytecode.SandboxClassLoader@52d6d490', Original: 'org.robolectric.internal.bytecode.SandboxClassLoader@52d6d490'
[task 2020-08-27T22:17:08.647Z]     D [DownloadMiddleware] Download restarted from the storage null
[task 2020-08-27T22:17:08.647Z]     D [DownloadMiddleware] Sending download intent null
[task 2020-08-27T22:17:08.647Z]   FAILURE
[task 2020-08-27T22:17:08.647Z] 
[task 2020-08-27T22:17:08.648Z] java.util.NoSuchElementException: List is empty.
[task 2020-08-27T22:17:08.648Z] 	at kotlin.collections.CollectionsKt___CollectionsKt.first(_Collections.kt:206)
[task 2020-08-27T22:17:08.648Z] 	at kotlin.collections.CollectionsKt___CollectionsKt.first(_Collections.kt:190)
[task 2020-08-27T22:17:08.648Z] 	at mozilla.components.feature.downloads.DownloadMiddlewareTest$RestoreDownloadsState MUST populate the store with items in the storage$1.invokeSuspend(DownloadMiddlewareTest.kt:190)
[task 2020-08-27T22:17:08.648Z] 	at mozilla.components.feature.downloads.DownloadMiddlewareTest$RestoreDownloadsState MUST populate the store with items in the storage$1.invoke(DownloadMiddlewareTest.kt)
[...]

PR #8293, opened 2020-08-31
Closes #8292: DownloadMiddleware uses context.dispatch on wrong thread

### LINKED ISSUE #8292: DownloadMiddleware dispatches on wrong thread via context.dispatch
Cause of: https://github.com/mozilla-mobile/fenix/issues/14486

`restoreDownloads` switches to an IO thread and can't use `context.dispatch` which is synchronous on the store thread.  Regression of: https://github.com/mozilla-mobile/android-components/issues/8271

csadilek:
@Amejia481 when reviewing https://github.com/mozilla-mobile/android-components/issues/8271 I missed that we're switching to an IO thread in `restoreDownloads`. My mistake.

We're not allowing a synchronous dispatch on the store from another thread. Undoing that and adding a fix for the intermittent test failure.

### DIFF
--- a/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadMiddleware.kt
+++ b/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadMiddleware.kt
@@ -101,7 +101,7 @@ class DownloadMiddleware(
         downloadStorage.getDownloads().collect { downloads ->
             downloads.forEach { download ->
                 if (!context.state.downloads.containsKey(download.id)) {
-                    context.dispatch(DownloadAction.RestoreDownloadStateAction(download))
+                    context.store.dispatch(DownloadAction.RestoreDownloadStateAction(download))
                     logger.debug("Download restarted from the storage ${download.fileName}")
                 }
             }
--- a/components/feature/downloads/src/test/java/mozilla/components/feature/downloads/DownloadMiddlewareTest.kt
+++ b/components/feature/downloads/src/test/java/mozilla/components/feature/downloads/DownloadMiddlewareTest.kt
@@ -8,8 +8,14 @@ import android.app.DownloadManager.EXTRA_DOWNLOAD_ID
 import android.content.Context
 import android.content.Intent
 import androidx.test.ext.junit.runners.AndroidJUnit4
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.cancel
 import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.test.TestCoroutineDispatcher
+import kotlinx.coroutines.test.resetMain
 import kotlinx.coroutines.test.runBlockingTest
+import kotlinx.coroutines.test.setMain
 import mozilla.components.browser.state.action.DownloadAction
 import mozilla.components.browser.state.state.BrowserState
 import mozilla.components.browser.state.state.content.DownloadState
@@ -21,8 +27,10 @@ import mozilla.components.support.test.argumentCaptor
 import mozilla.components.support.test.mock
 import mozilla.components.support.test.whenever
 import mozilla.components.support.test.ext.joinBlocking
+import org.junit.After
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertTrue
+import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.mockito.Mockito.verify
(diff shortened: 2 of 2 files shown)


Amejia481:
I've already put a pr :) https://github.com/mozilla-mobile/android-components/pull/8291/files

Amejia481:
bors retry

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadMiddleware.kt
```diff
@@ -101,7 +101,7 @@ class DownloadMiddleware(
         downloadStorage.getDownloads().collect { downloads ->
             downloads.forEach { download ->
                 if (!context.state.downloads.containsKey(download.id)) {
-                    context.dispatch(DownloadAction.RestoreDownloadStateAction(download))
+                    context.store.dispatch(DownloadAction.RestoreDownloadStateAction(download))
```
pocmo, line 104: This was the quick fix. But I'd recommend not passing the `MiddlewareContext` to this Coroutine at all since accessing it is not correct in any case.

Amejia481: Thanks for the feedback, could you further explain to understand better the issue?, what approach do you recommend in this case?

pocmo: pass the result of `context.store`, a `BrowserStore` around to coroutines like here and not the context object. Just to avoid using the wrong methods completely. :)

Amejia481: Something like this?
```kotlin
    private fun restoreDownloads(context: MiddlewareContext<BrowserState, BrowserAction>) {
        val store = context.store
        scope.launch {
            downloadStorage.getDownloads().collect { downloads ->
                downloads.forEach { download ->
                    if (!store.state.downloads.containsKey(download.id)) {
                        store.dispatch(DownloadAction.RestoreDownloadStateAction(download))
                        logger.debug("Download restarted from the storage ${download.fileName}")
                    }
                }
            }
        }
    }
```

components/feature/downloads/src/test/java/mozilla/components/feature/downloads/DownloadMiddlewareTest.kt
```diff
@@ -186,6 +214,7 @@ class DownloadMiddlewareTest {
         assertTrue(store.state.downloads.isEmpty())
 
         store.dispatch(DownloadAction.RestoreDownloadsStateAction).joinBlocking()
+        dispatcher.advanceUntilIdle()
```
csadilek, line 218: @Amejia481 I've just noticed we need both a `dispatcher.advanceUntilIdle` and also a `store.waitUnitldle` here to fix the intermittent test failure. The first one to make sure we've launched the coroutine in `restoreDownload` that dispatches the action. The second one to make sure the action has been processed by the store. We should have landed both PRs :).

Right now if we simulate a slow processing of the action the test would still fail.

Amejia481: Thanks, I will put a pr to fix!

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadMiddleware.kt
pocmo: I'd move this earlier to the call `restoreDownloads(context.store)`, then you have no way of accessing `context` inside the `launch` block :)

components/feature/downloads/src/test/java/mozilla/components/feature/downloads/DownloadMiddlewareTest.kt
pocmo: P
T
O

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadMiddleware.kt
Amejia481: That's even nicer!!!

csadilek: @pocmo @Amejia481 but with this pattern we have the inverse problem. If we pass the store around in the middleware then devs tend to call `store.dispatch` when they really want the synchronous `context.dispatch` in the middleware. (That happened last week too)

Maybe we should rename the latter to `syncDispatch` to make this more obvious.

pocmo: Oh yeah, I only meant in this specific case where everything the function does is inside a coroutine.

```Kotlin
private fun restoreDownloads(...) = scope.launch {
  ...
}
```

PR #8296, opened 2020-08-31
Improve qr support

Mugurell:
Our scanCompleteListener was being leaked by an AsyncTask.
Make sure we won't get called after onStop anymore.

Refactor the heavy AsyncTask to a cancellable coroutine and allow for an easy
resume of decoding the images.

### Pull Request checklist

- [x] **Changelog**: This PR does not need one

### After merge

### DIFF
--- a/components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt
+++ b/components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt
@@ -30,7 +30,6 @@ import android.hardware.camera2.CameraManager
 import android.hardware.camera2.CaptureRequest
 import android.media.Image
 import android.media.ImageReader
-import android.os.AsyncTask
 import android.os.Bundle
 import android.os.Handler
 import android.os.HandlerThread
@@ -43,6 +42,7 @@ import android.view.View
 import android.view.ViewGroup
 import android.widget.TextView
 import androidx.annotation.StringRes
+import androidx.annotation.VisibleForTesting
 import androidx.core.content.ContextCompat.getColor
 import androidx.fragment.app.Fragment
 import com.google.zxing.BinaryBitmap
@@ -50,6 +50,9 @@ import com.google.zxing.MultiFormatReader
 import com.google.zxing.NotFoundException
 import com.google.zxing.PlanarYUVLuminanceSource
 import com.google.zxing.common.HybridBinarizer
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
 import mozilla.components.feature.qr.views.AutoFitTextureView
 import mozilla.components.feature.qr.views.CustomViewFinder
 import mozilla.components.support.base.log.logger.Logger
@@ -75,7 +78,9 @@ import kotlin.math.min
 @Suppress("TooManyFunctions", "LargeClass")
 class QrFragment : Fragment() {
     private val logger = Logger("mozac-qr")
-
+    @VisibleForTesting
+    internal var multiFormatReader = MultiFormatReader()
+    private val coroutineScope = CoroutineScope(Dispatchers.Default)
     /**
      * [TextureView.SurfaceTextureListener] handles several lifecycle events on a [TextureView].
      */
@@ -192,12 +197,15 @@ class QrFragment : Fragment() {
             try {
                 image = reader.acquireNextImage()
                 val availableImage = image
-                if (availableImage != null) {
+                if (availableImage != null && scanCompleteListener != null) {
                     val source = readImageSource(availableImage)
                     val bitmap = BinaryBitmap(HybridBinarizer(source))
                     if (qrState == STATE_FIND_QRCODE) {
                         qrState = STATE_DECODE_PROGRESS
-                        AsyncScanningTask(scanCompleteListener).execute(bitmap)
+
+                        coroutineScope.launch {
+                            tryScanningBitmap(bitmap)
+                        }
                     }
                 }
             } finally {
@@ -239,6 +247,13 @@ class QrFragment : Fragment() {
         closeCamera()
         stopBackgroundThread()
         super.onPause()
+    }
+
+    override fun onStop() {
+        // Ensure we'll continue tracking qr codes when the user returns to the application
+        qrState = STATE_FIND_QRCODE
+
+        super.onStop()
     }
 
     internal fun startBackgroundThread() {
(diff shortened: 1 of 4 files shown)


components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt
```diff
@@ -593,33 +616,39 @@ class QrFragment : Fragment() {
...
-        private val scanCompleteListener: OnScanCompleteListener?,
-        private val multiFormatReader: MultiFormatReader = MultiFormatReader()
-    ) : AsyncTask<BinaryBitmap, Void, Void>() {
-
-        override fun doInBackground(vararg bitmaps: BinaryBitmap): Void? {
-            return processImage(bitmaps[0])
+    @VisibleForTesting
+    internal suspend fun tryScanningBitmap(bitmap: BinaryBitmap, scanCompleteListener: OnScanCompleteListener?) {
+        scanCompleteListener?.let {
+            withContext(Dispatchers.Default) {
```
csadilek, line 622: We're switching context here and then again in `decodeBitmap`. Looks like we can skip it here?

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt
```diff
@@ -211,6 +224,7 @@ class QrFragment : Fragment() {
     }
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        coroutineScope = view.toScope()
```
csadilek, line 227: We used to decode bitmaps on a background thread and now this is happening on the main thread. In case decoding is slow, we may introduce ANRs. It's fast on my devices here, but wondering if should switch to a different scope? Wdyt?

components/feature/qr/src/test/java/mozilla/components/feature/qr/QrFragmentTest.kt
```diff
@@ -82,6 +83,18 @@ class QrFragmentTest {
         verify(qrFragment).tryOpenCamera(anyInt(), anyInt(), anyBoolean())
     }
 
+    @Test
+    fun `onStop cancels current decode operations and reset state`() = runBlocking {
```
csadilek, line 87: nit: and resets state

components/feature/qr/src/test/java/mozilla/components/feature/qr/QrFragmentTest.kt
```diff
@@ -99,62 +112,82 @@ class QrFragmentTest {
     }
 
     @Test
-    fun `async scanning task invokes listener on successful qr scan`() {
+    fun `listener invocation on successful qr scan`() = runBlocking {
```
csadilek, line 115: nit: maybe: listener is invoked on...

Mugurell: Sure, thanks!

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt
Mugurell: I thought of decoupling them, make sure that anything `tryScanningBitmap` will call will run on a background thread.
But indeed, probably yagni. Thanks!

csadilek: > I thought of decoupling them, make sure that anything tryScanningBitmap will call will run on a background thread.

OK, got it. Did you see my comment below re: "background thread". It seems right now we run this on the main thread?

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt
Mugurell: `onImageAvailable` still comes on a background thread.
I used the `coroutineScope` scoped to Fragment's View's scope just to have the benefit of structured concurrency.
Though if we cancel the coroutine either way in onStop we'll not really need it...
Will create a new Dispatchers.Default scope and use that just in `launch`. Should help on simplifying the code and tests, thanks!

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt
Mugurell: Previously we would just call `tryScanningBitmap` from Main but it would then run on a background thread because of `withContext(Dispatchers.Default)`.
To simplify things now I call `tryScanningBitmap` straight from `Dispatchers.Default` and continue execution in the background until `onScanComplete` posts to Main in case of a successfully identified QR code.

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt
```diff
@@ -192,12 +200,15 @@ class QrFragment : Fragment() {
...
                 val availableImage = image
-                if (availableImage != null) {
+                if (availableImage != null && scanCompleteListener != null) {
                     val source = readImageSource(availableImage)
                     val bitmap = BinaryBitmap(HybridBinarizer(source))
                     if (qrState == STATE_FIND_QRCODE) {
                         qrState = STATE_DECODE_PROGRESS
-                        AsyncScanningTask(scanCompleteListener).execute(bitmap)
+
+                        currentDecodeOperation = coroutineScope.launch {
```
csadilek, line 209: @Mugurell Since the scope is using `Dispatchers.Default`, it can have more than one active coroutine. If `tryScanningBitmap` is still running when the next `onImageAvailable` call comes in we'd be launching a second / third one, but we only keep track of the last one we launched (`currentDecodeOperation`). You can simulate this by adding a `delay` after `tryScanningBitmap`.

In `onStop` we would then only cancel the last one we launched but not the other ones that could still be running.

This doesn't really have any consequences as you set the `scanComplete` listener to null and other results that still come in would be ignored. But still this doesn't quite behave as intended?

I see the following options:
- Don't stop the jobs, let it finish and ignore results if the listener is null.
- Collect all jobs and cancel them all on stop.

**And these I wouldn't implement:**
- Limit to one coroutine/thread -- Wouldn't do that because that will make scanning slower unnecessarily
- Cancel previous coroutine -- Wouldn't do that either because we could lose a good result as the camera is moving

csadilek (commented):
Thanks for the updates, @Mugurell. I found one more thing below worth discussing before landing this!

csadilek (approved):
There's one comment inline, but I think it's not critical to fix. ✅ so you can land when ready.

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt
Mugurell: Thank you for helping with this and sorry I hadn't responded earlier.
Postponed this a bit to think about the best way but forgot to update.

Based on you feedback I went with not trying to cancel coroutines but rather rely on null checks for `scanCompleteListener` for whether to continue execution or not.
Did not know of a small, clean way of collecting the coroutines and allow for them to automatically be disgarded when completed and if not for a small change this would've brought a bit too high of a complexity.
Also removed the `scanCompleteListener` as a parameter from `tryScanningBitmap` [since that would've leaked it](https://github.com/mozilla-mobile/android-components/blob/2a060eb2b9f80216690db1684cf0cfcaf4a83b5a/components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt#L617).

Would appreciate just a small pass over this changes, thanks again!

components/feature/qr/src/test/java/mozilla/components/feature/qr/QrFragmentTest.kt
```diff
@@ -82,6 +82,16 @@ class QrFragmentTest {
         verify(qrFragment).tryOpenCamera(anyInt(), anyInt(), anyBoolean())
     }
 
+    @Test
+    fun `onStop cancels current decode operations and resets state`() {
```
csadilek, line 86: nit: just `onStop resets state`, as it no longer cancels the decode operation.

components/feature/qr/src/main/java/mozilla/components/feature/qr/QrFragment.kt
csadilek: @Mugurell OK, thanks, looks good! nit: we can remove the `currentDecodeOperation` field now, right, and also update the test name?

One more thing we could do defensively is set the `scanCompleteListener` to null once we have a successful scan. Since we're potentially doing that on multiple threads we could end up with multiple calls to `onScanComplete`. We could do in the `QrFeature` where we remove the fragment (similar to `onStop`). Wdyt?

Mugurell: Indeed, I think there might be an edgecase here in which more Runnables about `onScanComplete` would be posted on Main.
.. Based on which we might again do inside `QrFeature#onScanComplete()`:
>            removeQrFragment()
>            onScanResult(result)

Went with nulling the `scanCompleteListener` in `QrFeature#onScanComplete()`.
Wanted to also add a test for this but.. seems like the invocation inside scanCompleteListener are not recorded.

csadilek: 👍

csadilek (approved):

Mugurell:
bors r=csadilek

Issue #8308, opened 2020-09-01
Let AutoSave respond to state changes


pocmo:
Issue #8255 will change how we deal with `EngineSessionState`. Instead of requesting it from an `EngineSession` via `saveState()`, we will get it via `Observer.onState()` whenever there is a new state.

This means that in `AutoSave` we can react to this change and can save the state to disk when it changed. Some other `AutoSave` options like `periodicallyInForeground` may no longer be needed since there's no need to save an unchanged state again.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-7196)

csadilek:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1793188

Change performed by the [Move to Bugzilla add-on](https://addons.mozilla.org/en-US/firefox/addon/move-to-bugzilla/).

PR #8310, opened 2020-09-01
Lazily restore engine sessions after content process kill or crash.

pocmo:
* Instead of keeping the `EngineSessionState` inside `EngineSession`, we now always attach it to `EngineState` and also do not clear it anymore.
* If the content process gets killed we now just suspend affected `EngineSession` instances. They will automatically and lazily get restored from the last `EngineSessionState` once needed.
* On a content process crash we now mark the `EngineState` as crashed and suspend the `EngineSession`. We will not restore the `EngineSession` until explicitly restored by the application (Fenix shows UI to the user for that).

Once this lands we can also change the `AutoSave` behavior since we now get a new state pushed and know that we can save it. I filed #8308 for that.

### DIFF
--- a/components/feature/session/src/main/java/mozilla/components/feature/session/engine/EngineViewPresenter.kt
+++ b/components/feature/session/src/main/java/mozilla/components/feature/session/engine/EngineViewPresenter.kt
@@ -4,6 +4,7 @@
 
 package mozilla.components.feature.session.engine
 
+import android.view.View
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.flow.collect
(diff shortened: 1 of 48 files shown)


pocmo:
The code is ready and works nicely in sample browser. But I am running into a white screen of doom issue in Fenix. I filed the following GeckoView bug for it: https://bugzilla.mozilla.org/show_bug.cgi?id=1662400

I also need to update some more tests.

So "draft" for now.

pocmo:
Depends on #8320

components/feature/session/src/main/java/mozilla/components/feature/session/engine/EngineViewPresenter.kt
```diff
@@ -56,6 +64,20 @@ internal class EngineViewPresenter(
...
+
+        if (tab.engineState.crashed) {
+            engineView.release()
+            actualView.visibility = View.GONE
+            return
+        } else {
+            actualView.visibility = View.VISIBLE
+        }
+
+        if (tab.content.firstContentfulPaint) {
```
csadilek, line 77: This seems unnecessary as we already set View.VISIBLE when we get here?

pocmo:
bors r=csadilek

PR #8344, opened 2020-09-07
For #7725 - Log FxaExceptions to our crash reporters if sending to device fails

Mugurell:
This brings parity with iOS which already is logging such exceptions (see https://github.com/mozilla-mobile/fenix/issues/10662)

---

### Pull Request checklist

### After merge


components/concept/sync/src/main/java/mozilla/components/concept/sync/Devices.kt
```diff
@@ -68,9 +68,10 @@ interface DeviceConstellation : Observable<AccountEventsObserver> {
      * Send a command to a specified device.
      * @param targetDeviceId A device ID of the recipient.
      * @param outgoingCommand An event to send.
-     * @return A [Deferred] that will be resolved with a success flag once operation is complete.
+     * @return A [Deferred] that will be resolved when the operation is complete.
+     * @throws Exception if the operation could not be completed.
```
mhammond, line 72: It might be worth clarifying that it will (should?) only throw `FxaException` exceptions (here and below)

mhammond (approved):
This looks good to me, thanks! But there will need to be a matching PR for Fenix before it lands, otherwise I fear these now unhandled exceptions will cause Fenix to crash. I don't know enough about Kotlin to know whether that Fenix caller will be able to examine the exceptions correctly (ie, whether it will know the exact types so it can differentiate them correctly for Sentry) so it's probably important we know for sure this approach will work OK. So I'll approve this, but please don't land it before someone (eg, @liuche, or even me in a pinch ;) has approved the matching Fenix PR.

Mugurell:
Thanks @mhammond 
The Fenix ticket already has a PR waiting - https://github.com/mozilla-mobile/fenix/pull/14810

components/concept/sync/src/main/java/mozilla/components/concept/sync/Devices.kt
Mugurell: Thanks!
Changed `Exception` occurrences to `FxaException`.

mhammond:
I think the assignee should be @Mugurell ?

Mugurell:
@mhammond @grigoryk I rebased the patch after the recent "Account manager state machine refactoring".
Would be great if you could take another look to confirm everything seems ok.
I tested and on the Fenix side everything is good.

components/feature/accounts-push/src/main/java/mozilla/components/feature/accounts/push/SendTabUseCases.kt
```diff
@@ -125,22 +126,17 @@ class SendTabUseCases(
 
         private suspend inline fun sendToAll(
             block: (Collection<Device>) -> List<Pair<Device, TabData>>
-        ): Boolean {
+        ) {
             // Filter devices to send tab capable ones.
             filterSendTabDevices(accountManager) { constellation, devices ->
-                // Get a list of device-tab combinations that we want to send.
-                return block(devices).map { (device, tab) ->
+                return block(devices).forEach { (device, tab) ->
```
mhammond, line 132: is this 'return' redundant?

mhammond (commented):
This looks right to me, but please get Grisha's review!

components/feature/accounts-push/src/main/java/mozilla/components/feature/accounts/push/SendTabUseCases.kt
Mugurell: It is, thanks!

liuche:
@st3fan could you get this pushed through? It's been open for a while.

st3fan:
I would like to ask @grigoryk for an additional review of this patch

st3fan:
@Mugurell can you bring this patch up to date?

liuche:
Hey @grigoryk @jonalmeida for the send tab issues that y'all are running into, can we get this PR reviewed and landed? It might have helpful logs as well.

The only conflicts I see are in the changelog so it's ready for review.

grigoryk:
@liuche taking a look at both PRs now. I don't think they'll help with debugging our current problems though (since this is dealing with the Send part, which works well atm).

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/Utils.kt
```diff
@@ -73,6 +73,17 @@ suspend fun handleFxaExceptions(logger: Logger, operation: String, block: () ->
     })
 }
 
+/**
+ * Helper method that intercepts [FxaException]s thrown when executing [block].
+ * - Adds own logic for handling such exceptions
+ * - Retrows the unaltered exceptions to callers of this method
+ */
+suspend fun executeAndGetFxaExceptions(logger: Logger, operation: String, block: () -> Unit) {
```
grigoryk, line 81: If your plan is to just re-throw whatever it is `block` throws, you don't need this wrapper at all. Just execute `block` directly!

grigoryk (changes requested):
While this approach works, I don't think it's the one we should take. Specifically, forcing consumers to deal with FxaExceptions is a potential foot gun (they must handle auth errors correctly, re-throw panics, swallow network errors, etc), leaks a pretty low-level detail all the way to the top of our stack, and ultimately complicates our implementation of these APIs.

The problem you're trying to solve is that a `boolean` result for `sendCommandToDevice` doesn't provide enough information about what actually happened, making it difficult to log an exact failure to Sentry. You can solve that by introducing a richer result, e.g.:

```
enum class SendCommandResult {
  Success,
  NetworkFailure,
  AuthFailure,
  OtherFailure
}
```

Or consider a `sealed class` if you'd like to associate more data with these exceptions, e.g.:
```
sealed class SendCommandResult {
   object Success : SendCommandResult()
   sealed class Failure : SendCommandResult() {
      data class Network(val code: Int) : Failure()
      object AuthFailure : Failure()
      data class OtherFailure(val details: String?) : Failure()
   }
}
```

You will then convert an FxaException into one of these, _while also_ letting `handleFxaException` do its thing - e.g. inform the state machine of auth failures, re-throw panics (these _will be_ reported to Sentry as uncaught crashes).

Optionally within the `SendToAllUseCase` class, if you choose, you may aggregate these results into something like the following:

```
enum class SendCommandToAllResult {
  Success,
  PartialFailure,
  Failure
}
```

But to be honest, I think that's unnecessary. We can infer health of this feature by looking at individual send results.

In order to report these to Sentry, you should create a new Exception class wrapping your `Failure`:

`class SendCommandException(val failure: SendCommandResult.Failure) : Exception(failure.toString())`

and then submit that to Sentry.

All of these changes can be contained entirely within `FxaDeviceConstellation` - it already has the `crashReporter` instance.

This way you won't need any changes in Fenix, and you don't even have to change any public API - e.g. you can just keep the `Boolean` interface externally, and not expose the internal rich `SendTabResult` at all, just use that for detailed error reporting.

... or you can expose it to consumers, with a hope that they can then build a nicer UI (e.g. Fenix will be able to call out a network failure in a Snackbar notification! that seems quite nice). But, consider doing that in a follow-up, e.g. this isn't central to your goal of better error reporting via Sentry.

Mugurell:
Thank you @grigoryk !

I now have `handleFxaExceptions` returning a `FxaException?` which when in `sendCommandToDevice` I just wrap in a `SendCommandException` and submit it to the crash reporter.

The `FxaException?` coming from `handleFxaExceptions` already has the appropriate subtypes for better identifying the exception cause so I didn't create another `SendCommandResult`. I can also wrap these FxaExceptions in new `SendCommandResult` or just map their stacktrace to new appropriate subtypes if you think this is needed.

Mugurell:
> @liuche taking a look at both PRs now. I don't think they'll help with debugging our current problems though (since this is dealing with the Send part, which works well atm).

The intent for this, based on https://github.com/mozilla-mobile/fenix/issues/10662 was just to have parity with iOS which would log such exceptions, if they were to happen.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/Utils.kt
```diff
@@ -64,12 +64,20 @@ suspend fun <T> handleFxaExceptions(
...
+ * @param operation short description of the operation to be completed. Used in logging.
+ * @param block operation to be completed.
+ *
+ * @return [FxaException] with more details about why the operation failed or
+ *
+ * [null] if the operation completed successfully.
  */
-suspend fun handleFxaExceptions(logger: Logger, operation: String, block: () -> Unit): Boolean {
-    return handleFxaExceptions(logger, operation, { false }, {
+suspend fun handleFxaExceptions(logger: Logger, operation: String, block: () -> Unit): FxaException? {
```
grigoryk, line 77: You don't need to change this method. Instead, in `sendCommandToDevice` change how `handleFxaExceptions` is used:

```
        val result = handleFxaExceptions(logger, "sending device command", { error -> error }) {
            when (outgoingCommand) {
                is DeviceCommandOutgoing.SendTab -> {
                    account.sendSingleTab(targetDeviceId, outgoingCommand.title, outgoingCommand.url)
                    SyncTelemetry.processFxaTelemetry(account.gatherTelemetry(), crashReporter)
                }
                else -> logger.debug("Skipped sending unsupported command type: $outgoingCommand")
            }
            null
        }
```

And that should be pretty much it!

grigoryk (commented):
Nice! You can simplify this even further, see my comment.

components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/Utils.kt
Mugurell: Thanks!

grigoryk (approved):
Looks good, thanks!

PR #8358, opened 2020-09-09
SessionStorage: Write state as stream to disk instead of building model in memory first.

pocmo:
With this change we are using `JsonWriter` to write `BrowserState` directly to disk using a stream. Previously we first build a `JSONObject` tree in memory. This essentially duplicates the state and that could become big if a user had a lot of tabs open and not much RAM.

### DIFF
--- a/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSessionState.kt
+++ b/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSessionState.kt
@@ -4,6 +4,7 @@
 
 package mozilla.components.concept.engine
 
+import android.util.JsonWriter
 import org.json.JSONObject
 
 /**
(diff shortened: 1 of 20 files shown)


components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSessionState.kt
```diff
@@ -18,5 +19,14 @@ interface EngineSessionState {
      * When reading JSON from disk [Engine.createSessionState] can be used to turn it back into an [EngineSessionState]
      * instance.
      */
+    @Deprecated("Use writeTo() instead.")
```
Amejia481, line 22: nit: Should we also include when this will be deleted?

Amejia481 (approved):
🚢

components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSessionState.kt
pocmo: I didn't want to have a fixed date for removing this since we first need time to migrate some of our own components code over. My primary goal was to avoid any new code using this. :)
But I filed a new issue for eventually migrating everything and will also link this in a comment: https://github.com/mozilla-mobile/android-components/issues/8370

pocmo:
bors r=Amejia481

pocmo:
bors r=Amejia481

pocmo:
bors retry

pocmo:
bors retry

Issue #8319, opened 2020-09-02
Update to Kotlin 1.4


pocmo:
✨

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-7205)

Amejia481:
I think we can close this :)

PR #8359, opened 2020-09-09
Issue #8319: Update to Kotlin 1.4 and Coroutines 1.3.9.

pocmo:
Let's see what taskcluster says. I am able to build sample browser locally with those changes.


pocmo:
```
[task 2020-09-09T13:30:12.377Z] Exception: task post-signing/post-signing-nightly has too many dependencies (100 > 99)
```

RelEng is currently trying to fix this.

components/browser/engine-gecko/src/test/java/mozilla/components/browser/engine/gecko/webnotifications/GeckoWebNotificationDelegateTest.kt
```diff
@@ -1,47 +0,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
```
jonalmeida, line 1: We're removing these tests for the Kotlin update?

pocmo: At first I wanted to fix them since using a null value for a non-null variable in Kotlin now throws `NullPointerException` instead of `IllegalStateException` as before, ... then I noticed that they do not actually test anything:

What happens is that `GeckoViewWebNotification` is mocked (so everything returns null). Once we pass it into `onShowNotification()` the code will call `toWebNotification()` on the mock. This will throw since `tag` is not nullable, but that's what the mock returns for `tag`. Then back in the test we catch that exception and do an assert on the message. So we do not really test our code at all. What we are testing is that the Kotlin compiler adds a null check with a specific message, a good test, but now for us? :)

jonalmeida: That sounds quite reasonable, thanks!

components/feature/push/src/main/java/mozilla/components/feature/push/Connection.kt
```diff
@@ -235,6 +234,8 @@ internal class RustPushConnection(
             )
 
             return DecryptedMessage(scope, data)
+        } else {
+            return null
```
jonalmeida, line 238: Interesting that the previous compiler didn't catch this! I wonder if it did some implicit return null.

jonalmeida (approved):

pocmo:
Not ready yet. :)

jonalmeida:
Oops!

pocmo:
Looks like the last things left are some issues with Glean. The team currently has a work week. End of the week we want to look at getting the Gradle plugin working with Glean, maybe that will also fix issues here: https://github.com/mozilla-mobile/android-components/pull/8360.

pocmo:
Added that change to #8360.

Issue #8365, opened 2020-09-09
Decision task is broken: post-signing tasks have more than 100 deps


JohanLorenzo:
Error: https://firefox-ci-tc.services.mozilla.com/tasks/YaVmsW-8QAmad7G1Gzb3nw/runs/0/logs/https%3A%2F%2Ffirefox-ci-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FYaVmsW-8QAmad7G1Gzb3nw%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log#L3099

```
[task 2020-09-09T13:08:06.876Z] 2020-09-09 13:08:06,875 - ERROR - Error loading tasks for kind post-signing:
[task 2020-09-09T13:08:06.877Z] Traceback (most recent call last):
[task 2020-09-09T13:08:06.878Z]   File "/builds/worker/.local/lib/python2.7/site-packages/taskgraph/generator.py", line 272, in _run
[task 2020-09-09T13:08:06.879Z]     parameters, list(all_tasks.values()), self._write_artifacts,
[task 2020-09-09T13:08:06.880Z]   File "/builds/worker/.local/lib/python2.7/site-packages/taskgraph/generator.py", line 76, in load_tasks
[task 2020-09-09T13:08:06.881Z]     for task_dict in transforms(trans_config, inputs)]
[task 2020-09-09T13:08:06.882Z]   File "/builds/worker/.local/lib/python2.7/site-packages/taskgraph/transforms/task.py", line 1131, in check_run_task_caches
[task 2020-09-09T13:08:06.883Z]     for task in tasks:
[task 2020-09-09T13:08:06.885Z]   File "/builds/worker/.local/lib/python2.7/site-packages/taskgraph/transforms/task.py", line 1079, in check_task_dependencies
[task 2020-09-09T13:08:06.886Z]     MAX_DEPENDENCIES))
[task 2020-09-09T13:08:06.886Z] Exception: task post-signing/post-signing-nightly has too many dependencies (100 > 99)
[task 2020-09-09T13:08:06.887Z] Traceback (most recent call last):
[task 2020-09-09T13:08:06.888Z]   File "/builds/worker/.local/lib/python2.7/site-packages/taskgraph/main.py", line 442, in main
[task 2020-09-09T13:08:06.889Z]     args.command(vars(args))
[...]

PR #8374, opened 2020-09-10
Chunk post-signing tasks 

JohanLorenzo:
---

Fixes https://github.com/mozilla-mobile/android-components/issues/8365

I'm not super proud of this patch, but it does the job and produces the intended behavior. Let me explain. Unlike what we have in mozilla-central with the `dummy` tasks, reference-browser has a custom implementation. The `complete` and now the `post-signing` kinds use this implementation. Just like in mozilla-central: it chunks tasks once one reached the max number of dependencies. However, it also creates an "umbrella" task that depends on all chunks. This umbrella task is of the same kind. This specificity made `multi_dep` unhappy because it enforces the beetmover tasks to have one dependency per kind:

https://github.com/mozilla-mobile/android-components/blob/7f8475537a42ec77b933157b5e2b4f5ade07398a/taskcluster/ac_taskgraph/loader/multi_dep.py#L50-L52

So what I ended doing: I added an attribute to the chunk tasks in order to know which one is the final one (aka the "umbrella" one) so that it becomes the only dependency of that kind. I don't know if this is the most elegant way of solving this issue, but taskgraph-diff tells me it works as expected: Beetmover keep waiting on the `post-signing-nightly` task for instance, and that one waits on `post-signing-nightly-1` and `post-signing-nightly-2` which depend on the signing tasks. I didn't see any side effect. 

### Pull Request checklist

### After merge

### DIFF
--- a/taskcluster/ac_taskgraph/transforms/chunk.py
+++ b/taskcluster/ac_taskgraph/transforms/chunk.py
@@ -24,6 +24,7 @@ def build_task_definition(orig_task, deps, count):
         task['treeherder']['symbol'] = add_suffix(
             task['treeherder']['symbol'], "-{}".format(count))
 
+    task["attributes"]["is_final_chunked_task"] = False
     return task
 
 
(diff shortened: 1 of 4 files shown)


taskcluster/ac_taskgraph/transforms/chunk.py
```diff
@@ -39,7 +40,9 @@ def add_dependencies(config, tasks):
         chunked_labels = {}
 
         # sort for deterministic chunking
-        dep_labels = sorted(task.pop('soft-dependencies', []))
+        dep_labels = task.pop('soft-dependencies', [])
+        dep_labels.extend(task.get('dependencies', {}).keys())
```
JohanLorenzo, line 44: post-signing doesn't define `soft-dependencies` that's why I added this line.

MihaiTabara (approved):

JohanLorenzo:
bors r=MihaiTabara

Issue #7558, opened 2020-06-29
Expose getBookmarkUrlForKeyword A-S api


grigoryk:
https://github.com/tinoism/application-services/commit/faaa8aba4dc95de9b7e5e8fd0e0ab1f89acec4a7 added this API, but we never exposed it in A-C. Let's do that, so we can add keywords support to the fenix awesomebar.

Mugurell:
Based on https://github.com/mozilla-mobile/android-components/pull/8375#issuecomment-704144847 this won't be needed. Closing the ticket.

PR #8375, opened 2020-09-10
For #7558 - Expose getBookmarkUrlForKeyword(String)

Mugurell:
This new method allows to query AS for if a String (containing one or more
words) is set as a keyword for an existing bookmark.

AS does not yet exposes the keyword in the "Bookmark"s objects so users will
not be able to see/add/modify keywords but with this new method they can be
used if they were set on another platform and then the bookmarks were synced
on an Android device.

The searchTerms for which we query bookmarks can have one or more words, with
the first followed by a space character.
If more words are used we'll try to:
- match the first word against a bookmark keyword
- use the rest of the String as a query parameter. Only if the bookmark url
contains "%s" as a placeholder that we can replace.
Otherwise "%s" (case insensitive) will be removed from the bookmark url.

---

### Pull Request checklist

- [x] **Quality**: This PR builds and passes detekt/ktlint checks
- [x] **Tests**: This PR includes thorough tests
- [x] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/android-components/blob/master/docs/changelog.md)
- [x] **Accessibility**: The code in this PR or does not include any user facing features

BookmarkKeywordsInFenix
[video demoing bookmark keywords in Fenix](https://drive.google.com/file/d/14Z9KEp8IgtkhQbs-dNjEgN_OKZNDmr83/view?usp=sharing)

### After merge

### DIFF
--- a/components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesBookmarksStorage.kt
+++ b/components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesBookmarksStorage.kt
@@ -16,6 +16,7 @@ import mozilla.components.concept.sync.SyncAuthInfo
 import mozilla.components.concept.sync.SyncStatus
 import mozilla.components.concept.sync.SyncableStore
 import mozilla.components.support.base.log.logger.Logger
+import mozilla.components.support.utils.URLStringUtils
 import org.json.JSONObject
 
 /**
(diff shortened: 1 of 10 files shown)


Amejia481:
@Mugurell  is this already compatible with https://bugzilla.mozilla.org/show_bug.cgi?id=1653261?

Amejia481:
Or do we can land this without anything for it?

Mugurell:
> @Mugurell is this already compatible with https://bugzilla.mozilla.org/show_bug.cgi?id=1653261?

Thanks!

I though of landing this first since it seems like there is a lot of work still needed for search engine aliases
and it would be a great addition to Fenix's functionality although we don't have full support (from AS) even for bookmark keywords (we still need https://github.com/mozilla-mobile/fenix/issues/12100)

I understand there are some concerns about different approaches - mobile / desktop, so I'd propose to wait a bit to make sure we have a consistent approach.

Amejia481:
👍

components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesBookmarksStorage.kt
```diff
@@ -63,6 +64,50 @@ open class PlacesBookmarksStorage(context: Context) : PlacesStorage(context), Bo
...
+    @Suppress("ComplexMethod")
+    override suspend fun getBookmarkUrlForKeyword(searchTerms: String): String? {
+        return withContext(readScope.coroutineContext) {
+            if (searchTerms.isBlank()) {
+                return@withContext null
+            }
+
+            if (URLStringUtils.isURLLike(searchTerms)) {
+                return@withContext null
+            }
```
Amejia481, line 86: nit: We could combine this too in one condition :)

```kotlin
 if (searchTerms.isBlank() || URLStringUtils.isURLLike(searchTerms)) {
                return@withContext null
}
```

components/feature/search/src/main/java/mozilla/components/feature/search/SearchUseCases.kt
```diff
@@ -20,15 +24,19 @@ import mozilla.components.browser.state.store.BrowserStore
  * this (optional) lambda will be invoked to create a [Session]. The default implementation creates a [Session] and adds
  * it to the [SessionManager].
  */
+@Suppress("LongParameterList")
 class SearchUseCases(
     context: Context,
     store: BrowserStore,
     searchEngineManager: SearchEngineManager,
     sessionManager: SessionManager,
+    private val bookmarksStorage: BookmarksStorage,
```
Amejia481, line 33: I'm not sure if this should be mandatory as not all browsers will have `Bookmarks`, but all will do search for example [Reference browser](https://github.com/mozilla-mobile/reference-browser) doesn't have bookmarks but it need search

Mugurell:
Thank you @Amejia481 for helping with this!
I understand though that we are transitioning from bookmark keywords to custom search engines so this can be dropped.

PR #8380, opened 2020-09-11
No issue: add script to get a valid mc hash from current ac checkout

mcomella:
---

Purpose: when building across repositories, e.g. fenix with a local a-c,
sometimes the build will fail with errors unrelated to your changes because of
a version mismatch: there were breaking changes in the dependency repository
that are not represented in the dependent repository. This script aims to
address that.

### Pull Request checklist

### After merge

### DIFF
--- a/tools/get_cross_repo_hashes.py
+++ b/tools/get_cross_repo_hashes.py
@@ -47,6 +47,9 @@ to get valid hashes between the major repositories:
 1) fenix checkout -> a-c hash
     build & install Fenix, hit 3-dot -> Settings -> About Firefox Nightly
     and read hash at the end of the "AC:" line
+    OR
+    after building a release build, visit `mozilla.components.Build` from your
+    Fenix repo in Android Studio and go-to-definition on `gitHash`
 
 2) a-c checkout (current) -> GV Nightly version
     AUTOMATED: run this script
--- a/tools/get_cross_repo_hashes.py
+++ b/tools/get_cross_repo_hashes.py
@@ -6,7 +6,6 @@
 """See USAGE for details.
 
 Potential script improvements:
-- AUTOMATE: a-c check out -> GV versions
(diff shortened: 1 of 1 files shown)


tools/get_cross_repo_hashes.py
```diff
@@ -0,0 +1,183 @@
...
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+"""See USAGE for details.
+
+Potential script improvements:
+- AUTOMATE: fenix check out -> a-c hash
+- (later) Support more than nightly
+- (later) Would this be more usable as a website?
+  Would require automating extraction of a-c hash
+  from Fenix apk or other mechanism
```
pocmo, line 13: That's interesting. I was also thinking about a website some time ago. I was wondering whether we could use the treeherder API to find releases and maybe have our build attach some more information about the builds as an artifact to the builds tasks.

pocmo (approved):
Nice 👍 

bors r+

tools/get_cross_repo_hashes.py
mcomella: For the purposes of finding a valid set of revisions for fenix -> ac -> mc, I was thinking it would be a website that takes the latest Nightly build and creates the mapping, e.g.
```
2020.09.08 -> <fenix-hash> <ac-hash> <mc-hash>
2020.09.07 -> <fenix-hash> <ac-hash> <mc-hash>
```

> maybe have our build attach some more information about the builds as an artifact to the builds tasks.

@pocmo Do you mean we'd attach something like the to the Nightly artifact Taskcluster page? e.g. https://firefox-ci-tc.services.mozilla.com/tasks/index/mobile.v2.fenix.nightly.2020.09.11.latest/arm64-v8a ? I think that would address the problem and avoid making a separate, hard-to-remember website. 😅

PR #8384, opened 2020-09-11
No issue: publish git hash in published pom XML.

mcomella:
This is inspired by how GV publishes commit information in their build
POMs.
    
This commit will allow devs to map an ac version to the ac revision that
built it. This will also enable the tools/get_cross_repo_hashes.py script
to map fenix -> ac -> mc.

---

### Pull Request checklist

### After merge

### DIFF
--- a/publish.gradle
+++ b/publish.gradle
@@ -14,6 +14,13 @@ static def getLocalPublicationTimestamp() {
     return date.format('yyyyMMddHHmmss')
 }
 
+static def getGitHash() {
+    String[] cmd = ['git', 'rev-parse', '--verify', 'HEAD'] // via https://stackoverflow.com/a/949391/2219998
+    def process = Runtime.getRuntime().exec(cmd)
+    process.waitFor()
+    return process.inputStream.getText()
+}
+
 ext.configurePublish = { groupIdArg, artifactIdArg, descriptionArg ->
     apply plugin: 'digital.wup.android-maven-publish'
 
@@ -53,6 +60,7 @@ ext.configurePublish = { groupIdArg, artifactIdArg, descriptionArg ->
                         connection = libVcsUrl
                         developerConnection = libVcsUrl
                         url = libUrl
+                        tag = getGitHash()
                     }
                 }
             }
--- a/publish.gradle
+++ b/publish.gradle
@@ -14,6 +14,7 @@ static def getLocalPublicationTimestamp() {
     return date.format('yyyyMMddHHmmss')
 }
 
+// This gitHash functionality is duplicated in :support-base build.gradle.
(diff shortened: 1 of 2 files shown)


publish.gradle
```diff
@@ -53,6 +60,7 @@ ext.configurePublish = { groupIdArg, artifactIdArg, descriptionArg ->
                         connection = libVcsUrl
                         developerConnection = libVcsUrl
                         url = libUrl
+                        tag = getGitHash()
```
pocmo, line 63: Oh, wow, that was much easier than I thought. :)

publish.gradle
```diff
@@ -14,6 +14,13 @@ static def getLocalPublicationTimestamp() {
     return date.format('yyyyMMddHHmmss')
 }
 
+static def getGitHash() {
+    String[] cmd = ['git', 'rev-parse', '--verify', 'HEAD'] // via https://stackoverflow.com/a/949391/2219998
+    def process = Runtime.getRuntime().exec(cmd)
+    process.waitFor()
+    return process.inputStream.getText()
+}
+
```
pocmo, line 23: We also have code like this in build.gradle of `support-base`, maybe both could call the same?
https://github.com/mozilla-mobile/android-components/blob/9d8bcc23c391e69876761262736a56c77e995a20/components/support/base/build.gradle#L11-L18

But not a blocker. :)

pocmo (approved):

publish.gradle
mcomella: I thought the same thing, lol

publish.gradle
mcomella: I don't know enough about groovy to make this easy, especially because one uses the short arg and the other does not, and I'm already way outside the thing I'm supposed to be working on 😅 so I'll settle for comments that indicate that this is duplicated.

mcomella: I realized that trimming might be a good idea – I didn't notice the POM file looks like this currently:
```
    <tag>52ac53ba150cd7753b1dd0080689741eb82e4dd9
</tag>
```

So I'll add one more commit that does the trim.

mcomella:
Maybe wait for Monday to land this? I'm concerned a bug could break publication and then we'll all be out b/c of the weekend. That being said, I'd love for it to land today because I could make more concrete progress on the script sooner. :)

pocmo (approved):
Alright, I'll land it and will trigger Nightlies. :)

bors r+

pocmo:
Can see it in Nightly here :)
https://nightly.maven.mozilla.org/maven2/org/mozilla/components/browser-awesomebar/59.0.20200914093656/browser-awesomebar-59.0.20200914093656.pom

PR #8388, opened 2020-09-11
Closes #8323: Add-on permission dialog does not prompt  for host permissions

### LINKED ISSUE #8323: Add-on permission dialog does not prompt for host permissions
In https://github.com/mozilla-mobile/android-components/issues/4871 we created a mapping from permissions to their translations.

One thing we missed is host permissions. We just added two new add-ons to Nightly, which only have host permissions. This currently causes us to display an empty "It requires your permission to:".

For example Video Background Play Fix requires the following permissions:
```
"*://*.youtube.com/*"
"*://*.youtube-nocookie.com/*"
"*://*.vimeo.com/*"
```

Since we don't have a mapping we don't display any of them. We should fix this before configuring these new add-ons in beta and release.

Desktop displays the following:

image

Amejia481:
Adapted algorithm from the [desktop implementation](https://searchfox.org/mozilla-central/source/toolkit/components/extensions/Extension.jsm#1488-1568) docs related to host permissions [[1]](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/permissions#Host_permissions) [[2]](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Match_patterns)

Fenix pr https://github.com/mozilla-mobile/fenix/pull/15002

---

### Pull Request checklist

### After merge

### DIFF
--- a/components/feature/addons/src/main/java/mozilla/components/feature/addons/Addon.kt
+++ b/components/feature/addons/src/main/java/mozilla/components/feature/addons/Addon.kt
@@ -4,7 +4,9 @@
 
 package mozilla.components.feature.addons
 
+import android.content.Context
 import android.os.Parcelable
+import androidx.core.net.toUri
 import kotlinx.android.parcel.Parcelize
 
 /**
@@ -107,12 +109,12 @@ data class Addon(
     ) : Parcelable
 
     /**
-     * Returns a list of id resources per each item on the [permissions] list.
+     * Returns a list of localized Strings per each item on the [permissions] list.
+     * @param context A context reference.
      */
-    fun translatePermissions(): List<Int> {
-        return localizePermissions(permissions)
+    fun translatePermissions(context: Context): List<String> {
+        return localizePermissions(permissions, context)
     }
-
     /**
      * Returns whether or not this [Addon] is currently installed.
      */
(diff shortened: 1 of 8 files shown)


components/feature/addons/src/main/res/values/strings.xml
```diff
@@ -7,6 +7,18 @@
...
+    <!-- Description for accessing multiple sites permission. %1$d will be replaced by an integer indicating the number of additional hosts for which this webextension is requesting permission. -->
+    <string name="mozac_feature_addons_permissions_to_many_sites_description_plural" tools:ignore="PluralsCandidate">Access your data on %1$d other sites</string>
+    <!-- Description for accessing multiple domains permission. %1$s will be replaced by an integer indicating the number of additional hosts for which this webextension is requesting permission. -->
+    <string name="mozac_feature_addons_permissions_to_many_domains_description" tools:ignore="PluralsCandidate">Access your data on %1$d other domain</string>
+    <!-- Description for accessing multiple domains permission. %1$s will be replaced by an integer indicating the number of additional hosts for which this webextension is requesting permission. -->
+    <string name="mozac_feature_addons_permissions_to_many_domains_description_plural" tools:ignore="PluralsCandidate">Access your data on %1$d other domains</string>
```
Amejia481, line 21: @Pike would you mind taking a look at these strings before landing? :)

components/feature/addons/src/main/java/mozilla/components/feature/addons/Addon.kt
```diff
@@ -193,8 +194,118 @@ data class Addon(
...
+            var siteCount = 0
+
+            loop@ for ((permission, stringId) in permissionsToTranslations) {
+                var host = permission.toUri().host ?: ""
+                when {
+                    stringId.isDomainAccessPermission() -> {
+                        ++domainCount
+                        host = host.removePrefix("*.")
+
+                        if (domainCount > 4) continue@loop
```
rocketsroger, line 251: Maybe we should assign 4 to a private const val? so we can update quickly if we decided to switch to a different number.

components/feature/addons/src/main/res/values/strings.xml
```diff
@@ -7,6 +7,18 @@
...
     <!-- Description for all_urls add-on permission. -->
     <string name="mozac_feature_addons_permissions_all_urls_description">Access your data for all websites</string>
+    <!-- Description for one site access add-on permission. %1$s  will be replaced by the DNS host name for which a web extension is requesting access (e.g., www.mozilla.org). -->
+    <string name="mozac_feature_addons_permissions_one_site_description">Access your data for %1$s</string>
+    <!-- Description for accessing multiple sites in one domain add-on permission. %1$s  will be replaced by the DNS domain for which a web extension is requesting access (e.g., mozilla.org). -->
+    <string name="mozac_feature_addons_permissions_sites_in_domain_description">Access your data for sites in the %1$s domain</string>
+    <!-- Description for accessing multiple sites permission. %1$d will be replaced by an integer indicating the number of additional hosts for which this webextension is requesting permission. -->
+    <string name="mozac_feature_addons_permissions_to_many_sites_description" tools:ignore="PluralsCandidate">Access your data on %1$d other site</string>
```
rocketsroger, line 15: NIT: Maybe add _single so it's easier for translation? Same with others below

rocketsroger (approved):
Looks good.  Some minor comments.  Thanks!

components/feature/addons/src/main/java/mozilla/components/feature/addons/Addon.kt
Amejia481: 👍

components/feature/addons/src/main/java/mozilla/components/feature/addons/Addon.kt
```diff
@@ -193,8 +194,118 @@ data class Addon(
          * @param permissions The list of permissions to be localized. Valid permissions can be found in
          * https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/permissions#API_permissions
          */
```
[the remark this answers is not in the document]
Amejia481: 👍

components/feature/addons/src/main/res/values/strings.xml
Amejia481: 👍

components/feature/addons/src/main/res/values/strings.xml
jonalmeida: @Delphine might be able to help here instead.

Amejia481: Thanks :)

Delphine: So I see a few things here:
* line 15-21: all comments are the same, so it's hard to understand how to localize these, and which string is supposed to be singular or plural
* Seems like they are assuming that all languages handle plurals like English?
* We unfortunately don't currently handle plural rules for Android
We should just do like we have usually done for these kinds of strings in the past, even though it's less than ideal (see the suggestions given here: https://github.com/mozilla-l10n/android-l10n/pull/11#discussion_r271636736)
thanks!

Amejia481: > line 15-21: all comments are the same, so it's hard to understand how to localize these, and which string is supposed to be singular or plural

Sorry my bad, I assumed that add it `_plural` to the end of the entry will be sufficient, I will fix comments :)

> Seems like they are assuming that all languages handle plurals like English?

Yes, we have a entry for singular and another entry for plural. Should I be doing something extra?

> We unfortunately don't currently handle plural rules for Android
> We should just do like we have usually done for these kinds of strings in the past, even though it's less than ideal (see the suggestions given here: mozilla-l10n/android-l10n#11 (comment))
> thanks!

Thanks, the approach that we were using it is adding an entry for plural and singular, similar to what we are doing. I will update the commends to make it more noticeable.

components/feature/addons/src/main/res/values/strings.xml
```diff
@@ -7,6 +7,24 @@
...
+      then we will show another collapsed entry saying "Access your data on 1 other domain". This entry it's for the singular case, when the add-on is only accessing one extra domain. -->
+    <string name="mozac_feature_addons_permissions_one_extra_domain_description" tools:ignore="PluralsCandidate">Access your data on 1 other domain</string>
+    <!-- When an add-on requires access to more than 4 domains, for example the add-on requires access for 6 domains. We will show the first 4 domains in individual entries, as in mozac_feature_addons_permissions_sites_in_domain_description,
+     then we will show another collapsed entry saying "Access your data on 2 other domains". This entry it's for the plural case, when the add-on is accessing more than one extra domain.
+     %1$s will be replaced by an integer indicating the number of additional domains for which this web extension is requesting permission. -->
+    <string name="mozac_feature_addons_permissions_extra_domains_description_plural" tools:ignore="PluralsCandidate">Access your data on %1$d other domains</string>
```
Amejia481, line 27: @Delphine I updated the pr, adding more descriptive comments. What do you think?

Delphine: Thanks for the update.
* line 19: comment doesn't match string
Seems like other than that, it should be good to go

Amejia481: That's correct, I will fix it thanks!

Amejia481:
bors r=rocketsroger,delphine

PR #8407, opened 2020-09-15
Closes #8406: Create separate usecases for removing all normal/private tabs.

### LINKED ISSUE #8406: Create separate usecases for removing all normal/private tabs.
Needed to integrate https://github.com/mozilla-mobile/android-components/issues/3265 into Fenix.

Right now we only have `RemoveAllTabsOfTypeUseCase`. Internally this maps to `filter` & `remove`. For one that is not very efficient. But more importantly this makes things like "undo" & "recently closed tabs" more complicated because there's not a single action to process.

Unfortunately this doesn't exist on `SessionManager` yet. And I do not want to introduce new events in `SessionManager.Observer`. So I think a good workaround could be to still do the `filter` & `remove` inside `SessionManager` but dispatch a single action to `BrowserStore`. Our migration tests should make sure that both are still in sync.

pocmo:
Undo (https://github.com/mozilla-mobile/android-components/issues/8406) will be much easier to implement if there's a single action for removing all normal/private tabs instead of mapping this to multiple `remove` calls.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -14,6 +14,11 @@ permalink: /changelog/
 
 * **browser-engine-gecko**, **browser-engine-gecko-beta**, **browser-engine-gecko-nightly**
   * 🚒 Bug fixed [issue #8431](https://github.com/mozilla-mobile/android-components/issues/8431) update `Session.trackerBlockingEnabled` and `SessionState#trackingProtection#enabled` with the initial tracking protection state.
+* **feature-tabs**
+  * Added `TabsUseCases.removeNormalTabs()` and `TabsUseCases.removePrivateTabs()`.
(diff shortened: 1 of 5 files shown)


rocketsroger (approved):

docs/changelog.md
```diff
@@ -14,7 +14,9 @@ permalink: /changelog/
 
 * **feature-downloads**
   * 🚒 Bug fixed [issue #8354](https://github.com/mozilla-mobile/android-components/issues/8354) Do not restart FAILED downloads.
-
+* **feature-tabs**
```
rocketsroger, line 17: Please move this to the latest version

pocmo:
bors r=rocketsroger

Issue #8417, opened 2020-09-16
SessionStorage: BrowserState with selected tab id for tab that doesn't exist


pocmo:
We switched from saving a `SessionManager.Snapshot` to a `BrowserState` in #7867 / 9e60a62119e4d0cb352642940faaec9da00fe1e1.

We now see the "selected tab" check fail. As a workaround I'll write a patch that clears the selected tab id in the state we are writing to disk, in that situation.

Nevertheless we should investigate how we end up in such a situation. Our reducer definitely doesn't do any checks when setting the selected tab id. But maybe this is intentional for "out of order" actions? Let's find out.

https://sentry.prod.mozaws.net/operations/firefox-beta/issues/9527493

```Kotlin
        if (state.normalTabs.isEmpty()) {
            clear()
            return true
        }

        if (state.selectedTabId != null) {
            requireNotNull(state.selectedTab) {
                "Selected session must exist"
            }
        }
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FNXV2-7291)

pocmo:
Related: https://github.com/mozilla-mobile/fenix/issues/15008

csadilek:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1793186

Change performed by the [Move to Bugzilla add-on](https://addons.mozilla.org/en-US/firefox/addon/move-to-bugzilla/).

PR #8419, opened 2020-09-16
Issue #8417: SessionStorage: Remove invalid selected tab id before saving state to disk.

pocmo:


Amejia481 (approved):
Looks good!
Should we include this change in the changelog?

pocmo:
I think I'd keep this as an implementation detail. Depending on what comes out of #8419 we may re-introduce the previous check.

bors r=Amejia481

PR #8434, opened 2020-09-17
Closes #8207: Add Delegate and Browser Session State for MediaSession API

### LINKED ISSUE #8207: Add Delegate and Browser Session State for MediaSession API
Gecko recently added support for platform level MediaControl and MediaSession API.  Integrate this support in Android Components.

rocketsroger:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/browser/session/src/main/java/mozilla/components/browser/session/engine/EngineObserver.kt
+++ b/components/browser/session/src/main/java/mozilla/components/browser/session/engine/EngineObserver.kt
@@ -19,6 +19,7 @@ import mozilla.components.browser.state.action.ContentAction
 import mozilla.components.browser.state.action.CrashAction
 import mozilla.components.browser.state.action.EngineAction
 import mozilla.components.browser.state.action.MediaAction
+import mozilla.components.browser.state.action.MediaSessionAction
 import mozilla.components.browser.state.action.TrackingProtectionAction
 import mozilla.components.browser.state.state.BrowserState
 import mozilla.components.browser.state.state.content.DownloadState
@@ -32,6 +33,7 @@ import mozilla.components.concept.engine.history.HistoryItem
 import mozilla.components.concept.engine.manifest.WebAppManifest
 import mozilla.components.concept.engine.media.Media
 import mozilla.components.concept.engine.media.RecordingDevice
+import mozilla.components.concept.engine.mediasession.MediaSession
 import mozilla.components.concept.engine.permission.PermissionRequest
 import mozilla.components.concept.engine.prompt.PromptRequest
 import mozilla.components.concept.engine.window.WindowRequest
@@ -306,6 +308,60 @@ internal class EngineObserver(
         }
 
         mediaMap.remove(media)
+    }
+
+    override fun onMediaActivated(mediaSessionController: MediaSession.Controller) {
+        store?.dispatch(MediaSessionAction.ActivatedMediaSessionAction(
+            session.id,
+            mediaSessionController
+        ))
+    }
+
+    override fun onMediaDeactivated() {
+        store?.dispatch(MediaSessionAction.DeactivatedMediaSessionAction(session.id))
+    }
+
+    override fun onMediaMetadataChanged(metadata: MediaSession.Metadata) {
+        store?.dispatch(MediaSessionAction.UpdateMediaMetadataAction(session.id, metadata))
+    }
+
+    override fun onMediaPlaybackStateChanged(playbackState: MediaSession.PlaybackState) {
+        store?.dispatch(MediaSessionAction.UpdateMediaPlaybackStateAction(
+            session.id,
+            playbackState
+        ))
+    }
+
+    override fun onMediaFeatureChanged(features: MediaSession.Feature) {
+        store?.dispatch(MediaSessionAction.UpdateMediaFeatureAction(
+            session.id,
+            features
+        ))
+    }
+
+    override fun onMediaPositionStateChanged(positionState: MediaSession.PositionState) {
+        store?.dispatch(MediaSessionAction.UpdateMediaPositionStateAction(
+            session.id,
+            positionState
+        ))
+    }
+
+    override fun onMediaMuteChanged(muted: Boolean) {
+        store?.dispatch(MediaSessionAction.UpdateMediaMutedAction(
+            session.id,
+            muted
+        ))
+    }
+
+    override fun onMediaFullscreenChanged(
+        fullscreen: Boolean,
+        elementMetadata: MediaSession.ElementMetadata?
+    ) {
+        store?.dispatch(MediaSessionAction.UpdateMediaFullscreenAction(
+            session.id,
+            fullscreen,
+            elementMetadata
+        ))
     }
 
     override fun onWebAppManifestLoaded(manifest: WebAppManifest) {
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/MediaSessionReducer.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/MediaSessionReducer.kt
@@ -1,0 +1,128 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.state.reducer
+
+import mozilla.components.browser.state.action.MediaSessionAction
+import mozilla.components.browser.state.state.BrowserState
+import mozilla.components.browser.state.state.SessionState
+import mozilla.components.browser.state.state.MediaSessionState
+import mozilla.components.concept.engine.mediasession.MediaSession
+
+internal object MediaSessionReducer {
+    /**
+     * [MediaSessionAction] Reducer function for modifying the [MediaSessionState] of a [SessionState].
+     */
+    fun reduce(state: BrowserState, action: MediaSessionAction): BrowserState {
+        return when (action) {
+            is MediaSessionAction.ActivatedMediaSessionAction ->
+                state.addMediaSession(action.tabId, action.mediaSessionController)
+            is MediaSessionAction.DeactivatedMediaSessionAction ->
+                state.removeMediaSession(action.tabId)
+            is MediaSessionAction.UpdateMediaMetadataAction ->
+                state.updateMediaMetadata(action.tabId, action.metadata)
+            is MediaSessionAction.UpdateMediaPlaybackStateAction ->
+                state.updatePlaybackState(action.tabId, action.playbackState)
+            is MediaSessionAction.UpdateMediaFeatureAction ->
+                state.updateMediaFeature(action.tabId, action.features)
+            is MediaSessionAction.UpdateMediaPositionStateAction ->
+                state.updatePositionState(action.tabId, action.positionState)
+            is MediaSessionAction.UpdateMediaMutedAction ->
+                state.updateMuted(action.tabId, action.muted)
+            is MediaSessionAction.UpdateMediaFullscreenAction ->
+                state.updateFullscreen(
+                    action.tabId,
+                    action.fullScreen,
+                    action.elementMetadata
+                )
+        }
+    }
+}
+
+private fun BrowserState.addMediaSession(
+    tabId: String,
+    mediaSessionController: MediaSession.Controller
+): BrowserState {
+    return updateTabState(tabId) { current ->
+        current.createCopy(mediaSessionState = MediaSessionState(
+            controller = mediaSessionController
+        ))
+    }
+}
+
+private fun BrowserState.removeMediaSession(
+    tabId: String
+): BrowserState {
+    return updateTabState(tabId) { current ->
+        current.createCopy(mediaSessionState = null)
+    }
+}
+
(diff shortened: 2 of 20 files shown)


rocketsroger:
Adding tests.

components/browser/session/src/main/java/mozilla/components/browser/session/engine/EngineObserver.kt
```diff
@@ -303,6 +305,73 @@ internal class EngineObserver(
         mediaMap.remove(media)
     }
 
+    override fun onMediaActivated(mediaSession: MediaSession) {
+        val store = store ?: return
+        store.dispatch(MediaSessionAction.ActivatedMediaSessionAction(session.id, mediaSession))
```
pocmo, line 310: nit: those can be written as `store?.dispatch(..)` - with that you do not need the check above.

components/browser/state/src/main/java/mozilla/components/browser/state/state/BrowserState.kt
```diff
@@ -31,6 +31,7 @@ data class BrowserState(
     val containers: Map<String, ContainerState> = emptyMap(),
     val extensions: Map<String, WebExtensionState> = emptyMap(),
     val media: MediaState = MediaState(),
+    val mediaSession: MediaSessionState = MediaSessionState(),
```
pocmo, line 34: My biggest question here is probably whether we should move this to `SessionState` and give every `SessionState` an optional `MediaSessionState?` instead of having one global `MediaSessionState`. Some advantage I see:
* We are independent from event ordering issues. If GeckoView will activate/deactivate sessions in an unexpected order then we may end up keeping the deactive state here and override the activate state.
* It's closer to the state in GV and we can support multiple sessions in the future.
* Getting the global state is not hard: We can get it from the list of tabs. And if that is too expensive then we can have a middleware that will aggregate a state for us (like we do for media right now, just a bit simpler).
* I feel like this could also make the `EngineSession.Observer` / `EngineObserver` / `BrowserStore` flow simpler since you do not need to pipe all the things through it every time (like everyone of them gets the whole session object and you need to write multiple values from it to make sure you got everything) and instead can only pass the changed data along (e.g. only the metadata).

What do you think?

rocketsroger: Understood.  I can see the advantages to move this into the `SessionState`.  I'll update this PR.  Thanks!

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/MediaSessionReducer.kt
```diff
@@ -0,0 +1,142 @@
...
+    }
+}
+
+private fun BrowserState.updateMediaSession(
+    tabId: String,
+    mediaSession: MediaSession
+): BrowserState {
+    return updateTabState(tabId) { current ->
+        current.createCopy(mediaSessionState = current.mediaSessionState.copy(
+            sessionId = tabId,
```
pocmo, line 45: It seems unnecessary to have the sessionId in the state object since it is already a child of the tab.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/mediasession/GeckoMediaSession.kt
```diff
@@ -0,0 +1,28 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.engine.gecko.mediasession
+
+import mozilla.components.concept.engine.mediasession.MediaSession
+import org.mozilla.geckoview.MediaSession as GeckoViewMediaSession
+
+internal class GeckoMediaSession(
+    private val geckoViewMediaSession: GeckoViewMediaSession
+) : MediaSession() {
```
pocmo, line 12: I wonder if we even need this `MediaSession` interface and the `GeckoMediaSession` implementations: In the reducer we seem to only take things like the `controller` or the `isActive` flag, so maybe we could just pass them through the observer to the store and keep all the media session details in GV?

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/MediaSessionReducer.kt
```diff
@@ -0,0 +1,142 @@
...
+internal object MediaSessionReducer {
+    /**
+     * [MediaSessionAction] Reducer function for modifying the [MediaSessionState] of a [SessionState].
+     */
+    fun reduce(state: BrowserState, action: MediaSessionAction): BrowserState {
+        return when (action) {
+            is MediaSessionAction.ActivatedMediaSessionAction ->
+                state.updateMediaSession(action.tabId, action.mediaSession)
+            is MediaSessionAction.DeactivatedMediaSessionAction ->
+                state.updateMediaSession(action.tabId, action.mediaSession)
```
pocmo, line 22: Instead of marking the media session as active `true/false`. Could the MediaState be nullable and we set it or remove it? I guess there's no need to keep an `MediaState` around if there's no media session.

components/browser/state/src/main/java/mozilla/components/browser/state/state/MediaSessionState.kt
```diff
@@ -0,0 +1,37 @@
...
+ * @property metadata The [MediaSession.Metadata] for this media session.
+ * @property elementMetadata The [MediaSession.ElementMetadata] for this media session.
+ * @property playbackState The current simplified [MediaSession.PlaybackState] of this media session.
+ * @property features The [MediaSession.Feature] for this media session.
+ * @property positionState The current simplified [MediaSession.PositionState] of this media session.
+ * @property muted True if media session is muted.
+ * @property fullscreen True if media session is fullscreen.
+ */
+data class MediaSessionState(
+    val id: String = UUID.randomUUID().toString(),
```
pocmo, line 26: Does this state need an ID?

components/browser/state/src/main/java/mozilla/components/browser/state/state/MediaSessionState.kt
```diff
@@ -0,0 +1,37 @@
...
+    val sessionId: String? = null,
```
pocmo, line 27: As mentioned above, I think this is now unnecessary since we attach the state to a session.

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/MediaSessionReducer.kt
rocketsroger: That's true, this is redundant.  I'll remove.

components/browser/state/src/main/java/mozilla/components/browser/state/state/MediaSessionState.kt
rocketsroger: Agreed, I'll remove

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/MediaSessionReducer.kt
rocketsroger: Sure, I'll make `MediaSessionState` nullable.  When activated, I'll create a new `MediaSessionState`.  When deactivated, I'll set it to null.

components/browser/state/src/main/java/mozilla/components/browser/state/state/MediaSessionState.kt
rocketsroger: I'm not sure, probably not.  I'll remove it.  If we need it in the future I can add it back in.

components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/mediasession/GeckoMediaSession.kt
rocketsroger: Updated to remove `GeckoMediaSession`.  Kept `MediaSession` to keep controller interface and data classes in one place.  What do you think?

components/browser/session/src/main/java/mozilla/components/browser/session/engine/EngineObserver.kt
```diff
@@ -305,6 +307,85 @@ internal class EngineObserver(
...
+    }
+
+    override fun onMediaPlaybackStateChanged(
+        mediaSessionController: MediaSession.Controller,
+        playbackState: MediaSession.PlaybackState
+    ) {
+        store?.dispatch(MediaSessionAction.UpdateMediaPlaybackStateAction(
+            session.id,
+            mediaSessionController,
+            playbackState
```
pocmo, line 339: Do we need to pass the `mediaSessionController` along every time? Does it change after the activate call?

config/detekt-baseline.xml
```diff
@@ -313,6 +313,14 @@
...
+    <ID>UndocumentedPublicFunction:EngineSession.kt$EngineSession.Observer$fun onMediaMetadataChanged( mediaSessionController: MediaSession.Controller, metadata: MediaSession.Metadata )</ID>
+    <ID>UndocumentedPublicFunction:EngineSession.kt$EngineSession.Observer$fun onMediaFeatureChanged( mediaSessionController: MediaSession.Controller, features: MediaSession.Feature )</ID>
+    <ID>UndocumentedPublicFunction:EngineSession.kt$EngineSession.Observer$fun onMediaPlaybackStateChanged( mediaSessionController: MediaSession.Controller, playbackState: MediaSession.PlaybackState )</ID>
+    <ID>UndocumentedPublicFunction:EngineSession.kt$EngineSession.Observer$fun onMediaPositionStateChanged( mediaSessionController: MediaSession.Controller, positionState: MediaSession.PositionState )</ID>
+    <ID>UndocumentedPublicFunction:EngineSession.kt$EngineSession.Observer$fun onMediaMuteChanged( mediaSessionController: MediaSession.Controller, muted: Boolean )</ID>
+    <ID>UndocumentedPublicFunction:EngineSession.kt$EngineSession.Observer$fun onMediaFullscreenChanged( mediaSessionController: MediaSession.Controller, fullscreen: Boolean, elementMetadata: MediaSession.ElementMetadata? )</ID>
```
pocmo, line 323: Could we just document them instead of adding more to the baseline file? That file was only needed to get us started. :)

pocmo (commented):
I think that looks pretty great already. Mostly only wondering about the controller being passed along for every callback. Is that needed?

config/detekt-baseline.xml
rocketsroger: Sure, I"ll document them :)

components/browser/session/src/main/java/mozilla/components/browser/session/engine/EngineObserver.kt
rocketsroger: I'm assuming the controller does not change after activate.  I'll double check.

rocketsroger: Confirmed, no change is expected.  I'll remove passing in `mediaSessionController` after activate.

components/browser/state/src/main/java/mozilla/components/browser/state/state/CustomTabSessionState.kt
```diff
@@ -25,6 +26,7 @@ data class CustomTabSessionState(
     val config: CustomTabConfig,
     override val engineState: EngineState = EngineState(),
     override val extensionState: Map<String, WebExtensionState> = emptyMap(),
+    override val mediaSessionState: MediaSessionState? = MediaSessionState(),
```
pocmo, line 29: Shouldn't this be `null` by default?

components/browser/state/src/main/java/mozilla/components/browser/state/state/MediaSessionState.kt
```diff
@@ -0,0 +1,30 @@
...
+ * @property metadata The [MediaSession.Metadata] for this media session.
+ * @property elementMetadata The [MediaSession.ElementMetadata] for this media session.
+ * @property playbackState The current simplified [MediaSession.PlaybackState] of this media session.
+ * @property features The [MediaSession.Feature] for this media session.
+ * @property positionState The current simplified [MediaSession.PositionState] of this media session.
+ * @property muted True if media session is muted.
+ * @property fullscreen True if media session is fullscreen.
+ */
+data class MediaSessionState(
+    val controller: MediaSession.Controller? = null,
```
pocmo, line 22: It looks like we will get the controller with the activate callback, so this will probably be never null?

pocmo (commented):
Looks pretty good. Only two small things.

components/browser/state/src/main/java/mozilla/components/browser/state/state/CustomTabSessionState.kt
rocketsroger: Yes, I agree.  I'll update

components/browser/state/src/main/java/mozilla/components/browser/state/state/MediaSessionState.kt
rocketsroger: Yes, I've confirmed that there's no usecase that will create a `MediaSessionState` without controller.  I'll fix.  Thanks,

pocmo (approved):

Issue #7865, opened 2020-07-27
Define how default search engines should get loaded and what the fallback mechanism should be


pocmo:
RFC: #7655

The basic mechanism to load and determine the default search engines should be:
* Query MLS for region (or read from cache)
* Use region to load search engines using `list.json`.

However we may need a fallback mechanism to use something else (e.g. locale, see also #1720) in case we can't query MLS or it just takes too long.

**Defining this mechanism may require an RFC of its own**. Let's also look at desktop and iOS and as needed involve the stakeholders to define the best approach.

PR #8437, opened 2020-09-18
RFC for determining the region of the user for picking default search engines.

pocmo:
Issue #7865.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -41,6 +41,7 @@ Before contributing, please review our [Community Participation Guidelines](http
 * [0003 - Adding a `concept-base` component]({{ site.baseurl }}/rfc/0003-concept-base-component)
 * [0004 - Introducing a Top Sites Feature]({{ site.baseurl }}/rfc/0004-top-sites-feature)
 * [0005 - Migrate feature-sitepermissions to be compatible with the new GeckoView permission API]({{ site.baseurl }}/rfc/0005-migrate-sitepermission-ac-store-geckoview-store)
+* [0006 - Determining the set of default search engines]({{ site.baseurl }}/rfc/0006-search-defaults)
 
 ### Presentations
 
--- a/docs/rfcs/0006-search-defaults.md
+++ b/docs/rfcs/0006-search-defaults.md
@@ -1,0 +1,101 @@
+---
+layout: page
+title: Determining the set of default search engines
+permalink: /rfc/0000-search-defaults
+---
+
+* Start date: 2020-09-D8
+* RFC PR: [#8437](https://github.com/mozilla-mobile/android-components/pull/8437)
+
+## Summary
+
+Determining the set of default search engines based on the user's home region.
+
+## Motivation
+
+Our configuration of default search engines ([list.json](https://github.com/mozilla-mobile/android-components/blob/master/components/browser/search/src/main/assets/search/list.json)) is based on the region and language or locale of the user.
+
+To provide our users with the best choice of search engines and to satisfy the needs of our partnerships, we want to determine (and recognize changes of) the user's region as accurately as possible.
+
+The scope of this document is limited to the resolving and caching of the user's region and using it to determine the default search engines.
+
+### Problems and constraints
+
+* **Fast**: A search engine needs to be available on application start. The search engine icon is used in the application UI and once the user starts typing we need to be able to provide search suggestions and ultimately perform the search.
+
+* **Accurate**: While the locale can be queried from the system, the actual region can be obscured or just harder to determine accurately.
+
+* **Change**: So far, on mobile, we have been doing only a single region lookup. In reality the region of a user can change over time. Especially when the initial installation happened outside of the "home region" of the user then this user remained in the "wrong" region until the app was reinstalled.
+
+* **Privacy**: Not all sources can be accessed at any time. While GPS may be the most accurate source for the user's region, it also requires a runtime permission, that we wouldn't want to ask for on the first app launch.
+
+## Guide-level explanation
+
+### Determining the region
+
+While there are many available sources for determining the region of the user, we will focus on using the Mozilla Location Service (MLS) as the primary source for now. MLS can determine the region of the user by doing a GeoIP lookup. Additionally it supports determining the location based on Bluetooth, cell tower and WiFI access point data. However so far on mobile we have only been using the GeoIP lookup and have not sent any data to MLS.
+
+### Fallback
+
+Other than previously in Fenix we are **not** using the locale as a fallback provider for the region. Instead we will fallback to use the default from the configuration (e.g. `default` in `list.json`).
+
+### Validation
+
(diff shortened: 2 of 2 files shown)


docs/rfcs/0000-search-defaults.md
```diff
@@ -0,0 +1,246 @@
...
+
+The first _validator_ that we want to ship uses the timezone of the user. Similar to the implementation in the desktop version of Firefox ([[1]](https://searchfox.org/mozilla-central/source/toolkit/modules/Region.jsm#189-194), [[2]](https://searchfox.org/mozilla-central/source/toolkit/modules/Region.jsm#736-752)) users with a `US` region and not a `US` timezone will not get `US` assigned as _current_ or _home_ region. In our implementation this behavior can be mapped to a _validator_ that returns `0` for `US` if the user is not in a `US` timezone.
+
+### Timeouts
+
+On the very first app launch we will have no _home_ region set yet. In order to get a quick response, and to not block the search functionality, we will have to apply a stricter timeout to the region resolution. This primarily affects sources that require a network request, like MLS. Fennec used a [10 second timeout](https://searchfox.org/mozilla-esr68/rev/4b79eedfe95ef427490a16e7c96f3a0930ab0c32/mobile/android/base/java/org/mozilla/gecko/search/SearchEngineManager.java#399-400), which we will use as starting point.
+
```
daleharvey, line 101: When you say timeout, are you planning to block the Search engines from being shown until this timeout is hit?

If so thats one thing I would consider rethinking, particularly when the default fallback is going to be a network request to MLS on a mobile device, on desktop we have switched that path to use the default region on startup and always treat region updates as background / async updates. Whether you have realtime updates may affect this on first run but even so I would expect on mobile we dont see the month long sessions without a restart that we can get on desktop right?

daleharvey:
Yeh this looks good, I would mostly echo the other comments made by mconnor + standard8 on Slack and added one about the startup flow, at least on desktop found it useful to get rid of any blocking but still let the region update, we dont actually redraw anything if its already shown to the user and so havent seen any problems with the realtime updates, but also a "hasBeenShownToUser" flag could also work

docs/rfcs/0000-search-defaults.md
pocmo: I sketched some code for this today and came to the same conclusion: There's no need for a timeout. Like you said, we can either use the home region we already have, or the default. And then always do the lookup request  + react to region updates, which are rare.
I haven't thought about very long running sessions, but also expect them to be unlikely on mobile. Does desktop handle this? I briefly looked at region.jsm, but couldn't see anything except the failure retry.

daleharvey: The patch for that is @ https://phabricator.services.mozilla.com/D91214

As was proposed doing here, we passively listen for geolocation updates and use that to update the region, but in case the user never uses any geolocation we schedule an updateTimer for once a week

mikedeboer:
Comment I made about the previous commit: I think it may be worth thinking a bit more about why you’re adding the scoring system here; using it for external systems that we really know nothing about is a little bit like adding a bit of leeway for it to be slightly less wrong in case our assumptions about the external service are wrong.
Of course we can tweak these multipliers later on, but experience tells me we seldom do.
What has worked in practice so far, in my (humble) experience, has been to ask each service to fetch the location and pick the one that’s most accurate as they are returned. So we have a ranking, where GNSS leads and MLS is second. and whichever wins the race gets to set the default. If GNSS is the last to return with a result, it’ll re-set the region.
Additionally, as on desktop, I think that we should be able to change the region during the app’s runtime. That might be super hard, I just don’t know.

docs/rfcs/0000-search-defaults.md
```diff
@@ -0,0 +1,101 @@
...
+* Current [SearchEngineManager in Android Components](https://github.com/mozilla-mobile/android-components/blob/08880314f56d73691b3cd909d5dee199bba4ed0b/components/browser/search/src/main/java/mozilla/components/browser/search/SearchEngineManager.kt#L28)
+* [FenixSearchEngineProvider](https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/components/searchengine/FenixSearchEngineProvider.kt) in Fenix
+* [SearchService](https://searchfox.org/mozilla-central/source/toolkit/components/search/SearchService.jsm) in Firefox (desktop)
+* [SearchEngines.swift](https://github.com/mozilla-mobile/firefox-ios/blob/main/Client/Frontend/Browser/SearchEngines.swift) in Firefox for iOS
+
+## Unresolved questions
+
+* What kind of telemetry could or should we collect? What data would help us tweak this? How can we confirm accuracy?
+
+* Is the 2 weeks period before switching the _home_ region, that we borrowed from desktop, acceptable for mobile too?
```
Amejia481, line 99: Is this the same time frame that iOS is using?

docs/rfcs/0000-search-defaults.md
```diff
@@ -0,0 +1,101 @@
...
+
+* Would it help to expose the _current_ and _home_ region in Fenix in a debug screen or in the "about" screen?
```
Amejia481, line 101: It could be helpful for debugging future situations

Amejia481 (approved):

docs/rfcs/0000-search-defaults.md
pocmo: From what I have seen iOS doesn't query MLS at all so far.

docs/rfcs/0000-search-defaults.md
```diff
@@ -0,0 +1,101 @@
...
+
+The new `SearchMiddleware` (see [RFC 2 about the new search architecture](https://mozac.org/rfc/0002-search-state-in-browser-store)) will listen to the `RegionState` update, load the matching default search engines using the search storage, and dispatch an action updating the `SearchState`.
+
+## Rationale and alternatives
+
+* An earlier draft of this RFC described using multiple region sources (GNSS, Network and SIM country, Google Play Services) and weighting them. This requires collecting more real world data about the accuracy of those sources and how they drift from what MLS determines, before shipping an implementation and therefore this idea was discarded here. It may be the topic for a separate, future RFC.
+
+## Prior art
+
+* [Timezone check in Firefox desktop](https://searchfox.org/mozilla-central/source/toolkit/modules/Region.jsm#180)
```
Standard8, line 85: Just a nit: you might want to change these to perma links available on the searchfox UI, e.g. that line is already out of date. Should probably be:

https://searchfox.org/mozilla-central/rev/f82d5c549f046cb64ce5602bfd894b7ae807c8f8/toolkit/modules/Region.jsm#204

docs/rfcs/0000-search-defaults.md
Standard8: Desktop has exposed the home region at least in about:support and it has been useful there when we've been looking at issues with people having the wrong/unexpected search engines.

docs/rfcs/0000-search-defaults.md
Amejia481: 👍

PR #8447, opened 2020-09-21
Closes #7719: Remove all default headers from GeckoViewFetchClient.

### LINKED ISSUE #7719: Are defult header needed GeckoViewFetchClient?
After chatting with @snorp he suggested that we shouldn't be passing [default headers ](https://github.com/mozilla-mobile/android-components/blob/589e186f8779594fb0641f6bc8bc00efa2da11f7/components/browser/engine-gecko-nightly/src/main/java/mozilla/components/browser/engine/gecko/fetch/GeckoViewFetchClient.kt#L49) to the `GeckoWebExecutor` as `GeckoView` is already providing any default headers and passing them could cause issue for some sites. Do we really need them?

pocmo:
Looking at all the linked issues I decided we should get rid of all default headers for the GeckoView client implementation.


pocmo:
bors try

Amejia481 (approved):
👍

Amejia481:
bors try

pocmo:
bors r=Amejia481

PR #8448, opened 2020-09-21
Feature-Parity with Desktop for the WebCompat Reporter, other minor things.

denschub:
This PR brings the android-components "Report Site Issue" feature to the same level of features as the one shipped in desktop. It also adds some tests and did a bit of cleanup.

---

I strongly suggest reviewing this by commits instead of the full diff view, that'll make your life a lot easier. :) Short explanation for each commit, in the order they appear on GitHub:

1. This is just a small adjustment to the name of the WebExtensions in our component. I've noticed that all other WebExtensions are prefixed with "Mozilla Android Components - ", but ours were not. There was no good reason for this, so I adjusted our build tooling to make sure we follow that.
2. This enables automated screenshots. We're shipping the same to desktop, but are using `tabs.captureTab()` instead of `tabs.captureVisibleTab()` there. The rest of the sources match the one shipped in mozilla-central.
3. This adds the `tabExtras` API that we're also shipping on desktop. This allows us to gather a bit more useful information to see why a broken site is broken at report time, for example checking if content blocking blocked things etc. This code also ships on desktop.
4. Is adding some basic tests, covering the Kotlin side only. Motivated by the recent changes in testing policy to m-c, we're currently re-evaluating how and where we run our tests, so there might be some follow-ups here, but this does, at least, cover the native part of the components.
5. Same as 4, but for the interventions part.
6. Is syncing down `v15.1.0` of the interventions addon that has been shipped to central as a last-second follow-up to 15.0.0 last week. The PiP changes do not affect AC/Fenix, but as you know, we share the same codebase between all platforms.

@csadilek, do you mind having a look at the tests (or the entire PR)? I can't say I ever wrote Kotlin-tests before, so this might be bad. :)

Also, I'm very sorry for batching so many changes together. We had some interesting dependency-issues, so this was the easiest way I could get these things together. If you want, I'm happy to split this PR into three parts: Changes to the reporter, changes to the Interventions addon, syncing v15.1.0. Just let me know!

### Pull Request checklist

### After merge


components/feature/webcompat-reporter/src/main/java/mozilla/components/feature/webcompat/reporter/WebCompatReporterFeature.kt
```diff
@@ -21,6 +23,14 @@ object WebCompatReporterFeature {
     internal const val WEBCOMPAT_REPORTER_EXTENSION_URL = "resource://android/assets/extensions/webcompat-reporter/"
     internal const val WEBCOMPAT_REPORTER_MESSAGING_ID = "mozacWebcompatReporter"
 
+    @VisibleForTesting
+    // This is an internal var to make it mutable for unit testing purposes only
+    internal var extensionController = WebExtensionController(
```
Amejia481, line 28: The annotation above `@VisibleForTesting` make sure if someone tries to use this property outside of a testing context will get a compilation error :)

components/feature/webcompat-reporter/src/test/java/mozilla/components/feature/webcompat/reporter/WebCompatReporterFeatureTest.kt
```diff
@@ -0,0 +1,115 @@
...
+@RunWith(AndroidJUnit4::class)
+class WebCompatReporterFeatureTest {
+
+    @Before
+    fun setup() {
+        WebExtensionController.installedExtensions.clear()
+    }
+
+    @Test
+    fun `install installs the webextension`() {
```
Amejia481, line 35: nit: `install installs`

Amejia481 (approved):
It looks good to me, I left some comments for some nits  :)
Christian will be back tomorrow from PTO :)

components/feature/webcompat-reporter/src/test/java/mozilla/components/feature/webcompat/reporter/WebCompatReporterFeatureTest.kt
denschub: that actually was in purpose, as the method is called `install` and it `installs the webextension`. But yeah, that's weird, I'll make that `installs the webextension`.

components/feature/webcompat-reporter/src/main/java/mozilla/components/feature/webcompat/reporter/WebCompatReporterFeature.kt
denschub: I don't think I understand your comment, sorry. It was my intent that nobody except for us in tests can modify this, and I shamelessly copy-pasted the annotation, including the comment, [over from `feature-readerview`](https://github.com/mozilla-mobile/android-components/blob/cb73e4f9564dd1e888241a621843c9383fd7cd6e/components/feature/readerview/src/main/java/mozilla/components/feature/readerview/ReaderViewFeature.kt#L60-L61). Do you think that's a bad idea?

Amejia481: Do not worry, the comment is OK :). My intention was just to let you know that when somebody sees the annotation already knows by annotation is self, that the property shouldn't be use outside of this context, the annotation itself is documenting the intention and enforcing it by a compilation check :)

components/feature/webcompat-reporter/src/test/java/mozilla/components/feature/webcompat/reporter/WebCompatReporterFeatureTest.kt
Amejia481: Ok, now I got it :)

components/feature/webcompat-reporter/src/main/java/mozilla/components/feature/webcompat/reporter/WebCompatReporterFeature.kt
denschub: Ahhh, alright. Thanks for that clarification! :)

components/feature/webcompat-reporter/src/test/java/mozilla/components/feature/webcompat/reporter/WebCompatReporterFeatureTest.kt
denschub: I changed it either way, that'll confuse a lot of folks otherwise. Thanks for the feedback!

components/feature/webcompat-reporter/src/main/java/mozilla/components/feature/webcompat/reporter/WebCompatReporterFeature.kt
csadilek: My comment here is meant in addition to `VisibleForTesting`. It's not just visible (internal), it's also a var (not a val) for testing purposes. Maybe I should have written "This is an internal **var**` :)

components/feature/webcompat-reporter/src/main/java/mozilla/components/feature/webcompat/reporter/WebCompatReporterFeature.kt
```diff
@@ -39,16 +49,15 @@ object WebCompatReporterFeature {
      * "android-components".
      */
     fun install(runtime: WebExtensionRuntime, productName: String = "android-components") {
-        runtime.installWebExtension(WEBCOMPAT_REPORTER_EXTENSION_ID, WEBCOMPAT_REPORTER_EXTENSION_URL,
+        extensionController.install(runtime,
             onSuccess = {
                 logger.debug("Installed WebCompat Reporter webextension: ${it.id}")
-                it.registerBackgroundMessageHandler(
-                    WEBCOMPAT_REPORTER_MESSAGING_ID,
+                extensionController.registerBackgroundMessageHandler(
```
csadilek, line 55: I think it's better to use `it.registerBackgroundMessageHandler` here, where `it` is the installed extension. Otherwise you're referencing  `this` (via `this.extensionController`) in the lambda. In other places this is problematic because we'd be leaking the feature instance. In your case this doesn't matter as your feature is a `object` (singleton), but I would still change it.

components/feature/webcompat-reporter/src/test/java/mozilla/components/feature/webcompat/reporter/WebCompatReporterFeatureTest.kt
```diff
@@ -0,0 +1,115 @@
...
+
+        val port: Port = mock()
+        val message = argumentCaptor<JSONObject>()
+        messageHandler.value.onPortConnected(port)
+        verify(port).postMessage(message.capture())
+
+        val productNameMessage = JSONObject().put("productName", "android-components")
+        verify(port, times(1)).postMessage(message.capture())
+
+        Assert.assertEquals(productNameMessage.toString(), message.value.toString())
```
csadilek, line 67: nit:  We usually import `import org.junit.Assert.assertEquals` directly so you don't have to write `Assert.assertEquals`, but just `assertEquals`.

csadilek (changes requested):
Thanks @denschub for adding test coverage. I found one small thing I'd recommend changing, but then we can 🚢 .

components/feature/webcompat-reporter/src/main/java/mozilla/components/feature/webcompat/reporter/WebCompatReporterFeature.kt
csadilek: So, it would just be `it.registerBackgroundMessageHandler(WEBCOMPAT_REPORTER_MESSAGING_ID, WebcompatReporterBackgroundMessageHandler(productName))`.

csadilek: Another alternative is to call `extensionsController.registerbackgroundMessageHandler()` before `extensionController.install` which will also work, as the the controller supports that, it will register once installed.

components/feature/webcompat-reporter/src/test/java/mozilla/components/feature/webcompat/reporter/WebCompatReporterFeatureTest.kt
```diff
@@ -0,0 +1,116 @@
...
+
+        val port: Port = mock()
+        val message = argumentCaptor<JSONObject>()
+        messageHandler.value.onPortConnected(port)
+        verify(port).postMessage(message.capture())
+
+        val productNameMessage = JSONObject().put("productName", "test")
+        verify(port, times(1)).postMessage(message.capture())
+
+        Assert.assertEquals(productNameMessage.toString(), message.value.toString())
```
csadilek, line 88: nit: leftover `Assert.` :)

csadilek (approved):
Looks good to me now! :)

components/feature/webcompat-reporter/src/test/java/mozilla/components/feature/webcompat/reporter/WebCompatReporterFeatureTest.kt
denschub: whoops! That's gone now.

csadilek:
@denschub looks like there's a lint error now:

`WebCompatReporterFeatureTest.kt:18:1: Unused import`

denschub:
Already force-pushed a fix (and ran ktlint and detekt locally to verify). That one time I didn't run it because I just removed one forgotten leftover... classic. :)

Issue #8456, opened 2020-09-21
Don't process add request for downloads already added


Amejia481:
Related Fenix issue https://github.com/mozilla-mobile/fenix/issues/15012

Amejia481:
Closed by https://github.com/mozilla-mobile/android-components/commit/df2b072299c895b863620c96d9b650e30421e5d3

PR #8457, opened 2020-09-21
Closes issue #8456: Don't process add request for downloads already added

Amejia481:
The only explanation that could find of why we are trying to add a duplicated id in the db is that it could be the case that we are getting multiple `AddDownloadAction`s on the `DownloadMiddleware`. For this reason, I'm ignoring any `AddDownloadAction` from an id that is already on `store.state.downloads`.

---

### Pull Request checklist

### After merge


components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadMiddleware.kt
```diff
@@ -55,21 +55,21 @@ class DownloadMiddleware(
...
-                    downloadStorage.add(action.download)
-                    logger.debug("Added download ${action.download.fileName} to the storage")
-                }
-                sendDownloadIntent(action.download)
-            }
-            is DownloadAction.RestoreDownloadStateAction -> {
-                sendDownloadIntent(action.download)
-            }
+            is DownloadAction.AddDownloadAction -> sendDownloadIntent(action.download)
+            is DownloadAction.RestoreDownloadStateAction -> sendDownloadIntent(action.download)
```
csadilek, line 72: nit: we could combine the two branches now since they're the same `AddDownloadAction, RestoreDownloadStateAction ->`?

csadilek (approved):
Not sure how this would happen but this is a good check to have. The only way I can see is if our `onStartDownload` / `onAppSelected` lambdas get invoked twice because they both start the download and dispatch the event.

In all other cases dispatching a new event is done with a new `DownloadState` object with a new ID. So, it has to be that somehow?

components/feature/downloads/src/main/java/mozilla/components/feature/downloads/DownloadMiddleware.kt
Amejia481: The problem is that we get a compilation error, as both actions are has `action.download` but when we combine them they don't have a hierarchy relationship parent that have the `download` property.

PR #8465, opened 2020-09-22
Mergify: Use smart strict merge.

pocmo:
With just "true" mergify updates all pull requests that are ready to merge. But after merging one,
it needs  to update all remaining ones again, and so on. Using "smart" mergify will only update
one pull request, merge it and then move to the next one.


jonalmeida (approved):
Nice!

pocmo:
Was wondering why this never got merged ... but the check says:
> This pull request must be merged manually because it modifies Mergify configuration

I guess this makes sense.

PR #8480, opened 2020-09-23
AddonCollectionProvider: Allow configuring custom collection owner / clean old cache files

csadilek:
This allows customizing the add-on collection owner by specifying either user ID or name, see: 
https://addons-server.readthedocs.io/en/latest/topics/api/collections.html#collection-add-ons-list (user parameter) (To test in Sample Browser, my user "16314372" and collection "test" can be used)

This is needed for https://github.com/mozilla-mobile/fenix/issues/14034

The second commit here makes sure we delete unused cache files. Currently when switching collections we cause left-over cache files that will never get deleted.

I've also refactored the tests a bit and added more coverage.

### DIFF
--- a/components/feature/addons/src/main/java/mozilla/components/feature/addons/amo/AddonCollectionProvider.kt
+++ b/components/feature/addons/src/main/java/mozilla/components/feature/addons/amo/AddonCollectionProvider.kt
@@ -17,6 +17,7 @@ import mozilla.components.concept.fetch.isSuccess
 import mozilla.components.feature.addons.Addon
 import mozilla.components.feature.addons.AddonsProvider
 import mozilla.components.support.base.log.logger.Logger
+import mozilla.components.support.ktx.kotlin.sanitizeFileName
 import mozilla.components.support.ktx.kotlin.sanitizeURL
 import mozilla.components.support.ktx.util.readAndDeserialize
 import mozilla.components.support.ktx.util.writeString
@@ -32,6 +33,7 @@ import java.util.Locale
 
 internal const val API_VERSION = "api/v4"
 internal const val DEFAULT_SERVER_URL = "https://addons.mozilla.org"
+internal const val DEFAULT_COLLECTION_USER = "mozilla"
 internal const val DEFAULT_COLLECTION_NAME = "7e8d6dc651b54ab385fb8791bf9dac"
 internal const val COLLECTION_FILE_NAME = "mozilla_components_addon_collection_%s.json"
 internal const val MINUTE_IN_MS = 60 * 1000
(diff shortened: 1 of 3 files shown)


components/feature/addons/src/main/java/mozilla/components/feature/addons/amo/AddonCollectionProvider.kt
```diff
@@ -176,6 +185,22 @@ class AddonCollectionProvider(
         }
     }
 
+    /**
+     * Deletes cache files from previous (now unused) collections.
+     */
+    @VisibleForTesting
+    internal fun deleteUnusedCacheFiles() {
+        val currentCacheFileName = getCacheFileName()
```
csadilek, line 193: We don't need to use our lock here as we're not touching files that are currently in use.

docs/changelog.md
```diff
@@ -17,7 +17,20 @@ permalink: /changelog/
 * **feature-tabs**
   * Added `TabsUseCases.UndoTabRemovalUseCase` for undoing the removal of tabs.
 * **feature-webcompat-reporter**
-  * Added the ability to automatically add a screenshot as well as more technical details when submitting a WebCompat report.  
+  * Added the ability to automatically add a screenshot as well as more technical details when submitting a WebCompat report.
+* **feature-addons**
+  * `AddonCollectionProvider` now supports configuring a custom collection owner (via user ID or name). ⚠️ This is a breaking change for call sites that don't rely on named arguments.
```
Amejia481, line 22: nit: Should this be int two separated lines? One for the new functionality and the other for the breaking change api.

components/feature/addons/src/test/java/AddonCollectionProviderTest.kt
```diff
@@ -219,6 +219,17 @@ class AddonCollectionProviderTest {
         verify(cachingProvider).writeToDiskCache(jsonResponse)
     }
 
+    @Test
```
Amejia481, line 222: Should we also add an extra test to verify that we are only deleting add-on collection files?

Amejia481 (commented):
So far it looks good, just one question and a nit. I will do a manual test to verify before ✅

Amejia481 (approved):
Working like a cat 🐱 

image

docs/changelog.md
csadilek: Hm but it's the same change and we usually keep those together. I agree this look weird though, so I've formatted it differently now in line with what we usually do.

components/feature/addons/src/test/java/AddonCollectionProviderTest.kt
csadilek: Yeah, good idea. I've added a test for this now: https://github.com/mozilla-mobile/android-components/pull/8480/files#diff-c1cd858b6129959a62115bf8d2f4229cR235

csadilek:
@Amejia481 updated and ready.

docs/changelog.md

components/feature/addons/src/test/java/AddonCollectionProviderTest.kt
Amejia481: Looks great!

Amejia481 (approved):
🚢

PR #8545, opened 2020-09-29
Closes #8488: Treat temporarily installed add-ons as supported

### LINKED ISSUE #8488: Display temporary addons as "supported"
Right now addons installed with `web-ext` don't show up in the supported list, this prevents addons authors from opening the options page and it's a confusing experience.

We recently landed `WebExtension.MetaData.temporary` that AC can use to detect when an extension is installed temporarely.

csadilek:
This is to make sure we show add-on temporarily installed via web-ext as "supported."


Amejia481 (approved):
🚢

Issue #8511, opened 2020-09-25
Fix CrashHandlerService#IntentService deprecation in Android 11 


Amejia481:
See the [docs for how to replace it](https://developer.android.com/reference/android/app/IntentService)

pocmo:
@rocketsroger converted the crash handler service into a job service in https://github.com/mozilla-mobile/android-components/issues/6058, but we had to back it out.

rocketsroger:
> @rocketsroger converted the crash handler service into a job service in #6058, but we had to back it out.

Yes it was causing the ANR which it shouldn't be I backed it until I can look into it more
If you want I can take a look at this next week.

PR #8547, opened 2020-09-29
For #8511: Fix CrashHandlerService IntentService deprecation in Android 11

rocketsroger:
---

### Pull Request checklist

### After merge

### DIFF
--- a/components/lib/crash/src/main/java/mozilla/components/lib/crash/handler/CrashHandlerService.kt
+++ b/components/lib/crash/src/main/java/mozilla/components/lib/crash/handler/CrashHandlerService.kt
@@ -4,42 +4,69 @@
 
 package mozilla.components.lib.crash.handler
 
-import android.app.IntentService
+import android.app.Service
 import android.content.Intent
-import mozilla.components.lib.crash.CrashReporter
+import android.os.Build
+import android.os.IBinder
+import androidx.annotation.VisibleForTesting
+import androidx.annotation.VisibleForTesting.PRIVATE
+import androidx.core.app.NotificationCompat
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
 import mozilla.components.lib.crash.Crash
+import mozilla.components.lib.crash.CrashReporter
+import mozilla.components.lib.crash.R
+import mozilla.components.lib.crash.notification.CrashNotification
+import mozilla.components.support.base.ids.SharedIdsHelper
 
-private const val WORKER_THREAD_NAME = "CrashHandlerService"
+private const val NOTIFICATION_TAG = "mozac.lib.crash.handlecrash"
 
 /**
  * Service receiving native code crashes (from GeckoView).
  */
-class CrashHandlerService : IntentService(WORKER_THREAD_NAME) {
-    private val logger by lazy { CrashReporter
-        .requireInstance
-        .logger
+class CrashHandlerService : Service() {
+    private val crashReporter: CrashReporter by lazy { CrashReporter.requireInstance }
+
+    override fun onStartCommand(intent: Intent, flags: Int, startId: Int): Int {
+        crashReporter.logger.error("CrashHandlerService received native code crash")
+        handleCrashIntent(intent)
+
+        return START_NOT_STICKY
     }
 
-    public override fun onHandleIntent(intent: Intent?) {
-        if (intent == null) {
-            return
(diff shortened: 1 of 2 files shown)


rocketsroger:
Tried to use JobIntentService but enqueue work from background service causes ANR.  Foreground service is the only way to go.

components/lib/crash/src/main/java/mozilla/components/lib/crash/handler/CrashHandlerService.kt
```diff
@@ -4,42 +4,57 @@
...
+            startForeground(notificationId, notification)
         }
 
-        logger.error("CrashHandlerService received native code crash")
-
+        NotificationManagerCompat.from(this).cancel(NOTIFICATION_TAG, NOTIFICATION_ID)
         intent.extras?.let { extras ->
             val crash = Crash.NativeCodeCrash.fromBundle(extras)
+            CrashReporter.requireInstance.onCrash(this, crash)
+        } ?: crashReporter.logger.error("Received intent with null extras")
```
pocmo, line 50: `IntentService` did run this on a background thread and now we are running this on the main thread. Could this be a problem? Are we potentially blocking the main thread here?

rocketsroger: We are probably blocking the main thread here.  I didn't think about this case since by the time we're here there's no UI thread.  I'll look into this and see if we should not run this in the main thread.

rocketsroger: Agreed, I'll move this into the IO thread.

components/lib/crash/src/test/java/mozilla/components/lib/crash/handler/CrashHandlerServiceTest.kt
```diff
@@ -48,52 +54,54 @@ class CrashHandlerServiceTest {
...
-        verify(reporter)?.onCrash(any(), any())
-        verify(reporter)?.sendCrashReport(any(), any())
-        verify(reporter, never())?.sendNonFatalCrashIntent(any(), any())
+        CoroutineScope(IO).launch {
+            intent!!.putExtra("fatal", true)
+            service!!.onStartCommand(intent!!, 0, 0)
+            verify(reporter)!!.onCrash(any(), any())
+            verify(reporter)!!.sendCrashReport(any(), any())
+            verify(reporter, never())!!.sendNonFatalCrashIntent(any(), any())
+        }
```
pocmo, line 92: Should this be `runBlocking` in the test? Otherwise this will likely run after the test method has already completed and never actually get verified. And even if it fails then it would crash in the thread of the coroutine and not here.

components/lib/crash/src/main/java/mozilla/components/lib/crash/handler/CrashHandlerService.kt
```diff
@@ -4,42 +4,60 @@
...
+    private val crashReporter: CrashReporter by lazy { CrashReporter.requireInstance }
+
+    override fun onStartCommand(intent: Intent, flags: Int, startId: Int): Int {
+        crashReporter.logger.error("CrashHandlerService received native code crash")
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            val channel = CrashNotification.ensureChannelExists(this)
+            val notification = NotificationCompat.Builder(this, channel)
+                .setContentTitle(getString(R.string.mozac_lib_crash_dialog_title,
+                    crashReporter.promptConfiguration.organizationName))
```
pocmo, line 36: I using the "crash dialog title" here the right choice?

components/lib/crash/src/test/java/mozilla/components/lib/crash/handler/CrashHandlerServiceTest.kt
rocketsroger: My mistake.  I'll update so tests can inject their own scope.

components/lib/crash/src/main/java/mozilla/components/lib/crash/handler/CrashHandlerService.kt
rocketsroger: I think so.  The name of the string is misleading.  This is actually the string we use for the crash notification.  And since this foreground service notification is not visible to the user, (The foreground service ends quick enough that no notification is shown) I chose this to match the crash notification that we put up for the user.

pocmo (approved):

Issue #8356, opened 2020-09-08
[BrowserGestureDetector] Crash in [@ java.lang.IllegalArgumentException: at kotlin.jvm.internal.ArrayIteratorKt.throwParameterIsNullException(ArrayIterator.kt:5)]


kbrosnan:
Crash report: https://crash-stats.mozilla.org/report/index/3e85c676-c77c-4ca8-ab56-1504d0200908

Java stack trace:
```
java.lang.IllegalArgumentException
	at kotlin.jvm.internal.ArrayIteratorKt.throwParameterIsNullException(ArrayIterator.kt:5)
	at kotlin.jvm.internal.ArrayIteratorKt.checkParameterIsNotNull(ArrayIterator.kt:1)
	at mozilla.components.browser.toolbar.behavior.BrowserGestureDetector$CustomScrollDetectorListener.onScroll(Unknown Source:2)
	at android.view.GestureDetector.onTouchEvent(GestureDetector.java:625)
	at mozilla.components.browser.toolbar.behavior.BrowserGestureDetector.handleTouchEvent$browser_toolbar_release(BrowserGestureDetector.kt:7)
	at mozilla.components.browser.toolbar.behavior.BrowserToolbarBottomBehavior.onInterceptTouchEvent(BrowserToolbarBottomBehavior.kt:2)
	at mozilla.components.browser.toolbar.behavior.BrowserToolbarBottomBehavior.onInterceptTouchEvent(BrowserToolbarBottomBehavior.kt:1)
	at androidx.coordinatorlayout.widget.CoordinatorLayout.performIntercept(CoordinatorLayout.java:13)
	at androidx.coordinatorlayout.widget.CoordinatorLayout.onInterceptTouchEvent(CoordinatorLayout.java:3)
	at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2573)
	at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3030)
	at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2719)
	at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3030)
	at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2719)
	at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3030)
	at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2719)
[...]

PR #8553, opened 2020-09-30
For #8356 - Null check down event in BrowserGestureDetector

Mugurell:
Last ditch effort to prevent against crashes caused by ACTION_DOWN event being
null in GestureDetector.SimpleOnGestureListener#onScroll.

We can use the fact that the parameters are platform types and so can consider
the first parameter (that should be the ACTION_DOWN MotionEvent) being null so
that we won't automatically crash in such situations.
This will result in the toolbar not being animated as a result of a scroll
gesture by the user.

More investigations about why the ACTION_DOWN event is null are to be done in
android-components/issues/8552.

---

### Pull Request checklist

- [x] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended).
- [x] **Tests**: No tests. Simple null check added.
- [x] **Changelog**: This PR does not need a changelog entry.

### After merge


Amejia481 (approved):

components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/behavior/BrowserGestureDetector.kt
```diff
@@ -29,14 +29,19 @@ internal class BrowserGestureDetector(
...
                 listener.onScroll?.invoke(distanceX, distanceY)
 
-                if (abs(currentEvent.y - previousEvent.y) >= abs(currentEvent.x - previousEvent.x)) {
-                    listener.onVerticalScroll?.invoke(distanceY)
-                } else {
-                    listener.onHorizontalScroll?.invoke(distanceX)
+                // We got many crashes because of the initial event - ACTION_DOWN being null.
+                // Investigations to be continued in android-components/issues/8552.
+                // In the meantime we'll protect against this with a simple null check.
+                if (previousEvent != null) {
```
Amejia481, line 39: Should we add a test as part of the change?

Mugurell: Added one that would fail with the same stacktrace before the patch.

Mugurell:
bors r=Amejia481

Issue #8554, opened 2020-09-30
Migrate sitePermissions from Observer to Flow


codrut-topliceanu:
Will fix https://github.com/mozilla-mobile/fenix/issues/14637
At the moment SitePermissionsFeature uses an Observer to check if a permissionRequest for the current session has been called. This does currently does not handle the case where multiple requests are made from embedded elements. 
In trying to fix this I suggest moving away from observing Session to using Kotlin Flow with ContentState.

PR #8555, opened 2020-09-30
For #8554 - Migrate permissionFeature to KotlinFlow

codrut-topliceanu:
---

For https://github.com/mozilla-mobile/android-components/issues/8554
to fix https://github.com/mozilla-mobile/fenix/issues/14637
### Pull Request checklist

### After merge


Amejia481 (commented):
So far so good 👍 . I did a first pass, and left some comments. I will continue tomorrow :)

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
```diff
@@ -116,27 +180,31 @@ class SitePermissionsFeature(
     @Suppress("MaxLineLength")
     override fun onPermissionsResult(permissions: Array<String>, grantResults: IntArray) {
         sessionManager.runWithSessionIdOrSelected(sessionId) { session ->
```
Amejia481, line 182: Now that we are use `BrowserStore`, I think we are not going to need `sessionManager`

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
```diff
@@ -116,27 +180,31 @@ class SitePermissionsFeature(
...
-            session.appPermissionRequest.consume { permissionsRequest ->
 
+            val appPermissionRequest =
+                store.state.findTabOrCustomTabOrSelectedTab(customTabId)?.content?.appPermissionRequestsList?.find {
```
Amejia481, line 185: nit: As this `store.state.findTabOrCustomTabOrSelectedTab(customTabId)?.content?` repeats in the code maybe we can have a helper function. What do you think?

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsFeature.kt
```diff
@@ -150,61 +218,71 @@ class SitePermissionsFeature(
...
+        consumePermissionRequest(permissionRequest)
     }
 
-    internal fun onPositiveButtonPress(sessionId: String, shouldStore: Boolean) {
+    internal fun onPositiveButtonPress(
+        permissionRequest: PermissionRequest,
+        sessionId: String,
+        shouldStore: Boolean
+    ) {
         sessionManager.runWithSession(sessionId) { session ->
```
Amejia481, line 236: Same as https://github.com/mozilla-mobile/android-components/pull/8555/files#r498277307 :)

Amejia481 (commented):
Looks good keep up the good work 💪 !

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/ContentStateReducer.kt
```diff
@@ -128,6 +129,62 @@ internal object ContentStateReducer {
...
+            is ContentAction.UpdatePermissionsRequest -> updateContentState(
+                state,
+                action.sessionId
+            ) {
+                if (!it.permissionRequestsList.doesContain(action.permissionRequest)) {
+                    val currentList = it.permissionRequestsList.toMutableList()
+                    currentList.add(action.permissionRequest)
+                    it.copy(
+                        permissionRequestsList = currentList
+                    )
```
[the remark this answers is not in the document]
codrut-topliceanu: I used `list + element` shorthand (as per the previous comment).

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsDialogFragment.kt
```diff
@@ -217,5 +221,6 @@ internal class SitePermissionsDialogFragment : AppCompatDialogFragment() {
             fragment.arguments = arguments
             return fragment
         }
+        lateinit var permissionRequest: PermissionRequest
```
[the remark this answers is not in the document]
codrut-topliceanu: Won't this dialog last for a few seconds, until the user dismisses it or presses allow/don't allow?

components/browser/state/src/main/java/mozilla/components/browser/state/ext/PermissionRequest.kt
```diff
@@ -0,0 +1,16 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+package mozilla.components.browser.state.ext
+import mozilla.components.concept.engine.permission.PermissionRequest
+
+/**
+ * An improved version of [List.contains] built to be used in [ContentStateReducer] to make sure
+ * no duplicate requests are added to the state.
+ * */
+fun List<PermissionRequest>.doesContain(permissionRequest: PermissionRequest): Boolean {
```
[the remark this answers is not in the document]
codrut-topliceanu: I tried this initially but Android Studio tells me that "extension is shadowed by a member", so using my extension `.contains` will actually default to the standard `contains()`.

codrut-topliceanu:
Found some issues, moving it back to draft

components/browser/state/src/main/java/mozilla/components/browser/state/ext/PermissionRequest.kt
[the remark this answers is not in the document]
Amejia481: I think we have to use the override operator keyword in front of the extension, maybe this example could [help](https://github.com/mozilla-mobile/android-components/blob/5cd371eebc9e2dfef8cc45472dd52a72b5a36558/components/concept/fetch/src/main/java/mozilla/components/concept/fetch/Headers.kt#L39)

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsDialogFragment.kt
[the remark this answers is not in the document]
Amejia481: Yes, but if we hold the reference and Android kills the app and tries to restored it, this value will no be present and the app will crash when it tries to used it. To see this behaviour go developer options and activate "Don't keep activities", then trigger the dialog, with the dialog open and switch to another and come back to this app.

components/browser/state/src/main/java/mozilla/components/browser/state/ext/PermissionRequest.kt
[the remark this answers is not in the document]
codrut-topliceanu: `operator fun List<PermissionRequest>.contains(permissionRequest: PermissionRequest): Boolean{
    return this.any {
        it.uri == permissionRequest.uri &&
                it.permissions == permissionRequest.permissions
    }
}`
This is what I've tried in my extension file, but I get the warning/error I mentioned. The example you've given is from a Iterable class, PermissionRequest isn't that type of class.

I can add the following snippet to the PermissionRequest class, 
`operator fun contains(list:List<PermissionRequest>):Boolean{
        return list.any {
            it.uri == this.uri &&
                    it.permissions == this.permissions
        }
    }`

...but I don't really like the syntax because it would look like `permissionRequest contains List<PermissionRequest>` -> `list in request`

To make it work and look properly I believe I'd have to create a custom list class for List<PermissionRequest>.... but that defeats the purpose of simplicity. 
Please let me know if I'm missing something as I don't have much experience with operator overloading.

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsDialogFragment.kt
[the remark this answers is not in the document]
codrut-topliceanu: While working on implementing your suggestion I decided to test how it works as is. Turns out with "Don't keep activities" enabled, the dialog fragment just gets destroyed and triggers the `onDismiss` thus consuming the permissionRequest which, I think, is a satisfactory behaviour. Did it work differently when you tested it?

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/ContentStateReducer.kt
```diff
@@ -128,6 +129,66 @@ internal object ContentStateReducer {
             is ContentAction.UpdateHistoryStateAction -> updateContentState(state, action.sessionId) {
                 it.copy(history = HistoryState(action.historyList, action.currentIndex))
             }
+            is ContentAction.UpdatePermissionsRequest -> updateContentState(
+                state,
+                action.sessionId
+            ) {
+                if (!it.permissionRequestsList.doesContain(action.permissionRequest)) {
+                    it.copy(
+                        permissionRequestsList = it.permissionRequestsList + action.permissionRequest
```
Amejia481, line 138: nit: could be do `permissionRequestsList +=action.permissionRequest`? :)

Amejia481: The same could apply to other variants :)

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsDialogFragment.kt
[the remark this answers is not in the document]
Amejia481: This behaviour is happening because `permissionRequest` is static it belongs to the companion object, holding static values is not a good idea as it could cause memory leaks and other issues. 
 
image

components/browser/state/src/main/java/mozilla/components/browser/state/ext/PermissionRequest.kt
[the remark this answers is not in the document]
Amejia481: Got it, the issue because we have an extension function :( , what do you think about `containsPermission` or `hasPermission` ?

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/ContentStateReducer.kt
codrut-topliceanu: `permissionRequestList` is actually a named argument there.
 
image

components/feature/sitepermissions/src/main/java/mozilla/components/feature/sitepermissions/SitePermissionsDialogFragment.kt
```diff
@@ -70,6 +71,7 @@ internal class SitePermissionsDialogFragment : AppCompatDialogFragment() {
 
     internal var feature: SitePermissionsFeature? = null
     internal var userSelectionCheckBox: Boolean = false
+    internal var permissionRequest: PermissionRequest? = null
```
Amejia481, line 74: I think it will be better if we don't keep this reference as mentioned on https://github.com/mozilla-mobile/android-components/pull/8555#discussion_r497764502 this value will be not available in situations where Android recreates the fragment, using a null able type we are getting away from the null pointer exception, but we are introducing a new bug:

1. With don't keep activities active, trigger the prompt.
2. Leave the app and go to another app.
3. Return to the app, you will see the prompt being recreated.
4. Tap either "Don't allow" or "Allow" buttons.

None of these actions are taking place, this creates a false expectancy for users to believe the permission was `allowed` or `denied` and later the dialog is going to show again.  

 The issue is that `permissionRequest` is `null`, because the fragment was recreated and this value wasn't re assigned to avoid that I believe we could use an alternative as suggested on https://github.com/mozilla-mobile/android-components/pull/8555#discussion_r497764502 of storing an identifier of the permission, in the fragment arguments, and later passing it to the `feature`, similar to what we are doing with `sessionId`,  as an example you could take a look at all the state of this fragment like `sessionId`, `icon`, `dialogGravity` and others, all of them are following the same pattern to avoid similar issues, you could find more info about this pattern [here](https://stackoverflow.com/a/9245510/5533820) and for a more expanded version [here](https://developer.android.com/reference/android/app/Fragment.html#lifecycle).

If you would like to schedule a meeting to chat about it just let me know happy to help :)

Amejia481:
Apart from the comments above the pr looks pretty, I think we can start writing tests for it :)

components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/permission/GeckoPermissionRequest.kt
```diff
@@ -31,7 +32,8 @@ import org.mozilla.geckoview.GeckoSession.PermissionDelegate.PERMISSION_PERSISTE
  */
```
Amejia481, line 32: nit: Let's add the docs for the new property :)

Amejia481: Same for other variants

components/concept/engine/src/main/java/mozilla/components/concept/engine/permission/PermissionRequest.kt
```diff
@@ -14,6 +14,8 @@ interface PermissionRequest {
      */
     val uri: String?
 
+    val id: String
```
Amejia481, line 17: nit: Let's add docs for this property :)

Amejia481 (approved):
LGTM 👍🏽 

Let's fix the failing task and squash the commits.

```
[task 2020-10-26T14:00:08.677Z] > Task :concept-engine:compileReleaseUnitTestKotlin FAILED
[task 2020-10-26T14:00:08.677Z] e: /builds/worker/checkouts/src/components/concept/engine/src/test/java/mozilla/components/concept/engine/permission/PermissionRequestTest.kt: (73, 13): Class 'MockPermissionRequest' is not abstract and does not implement abstract member public abstract val id: String defined in mozilla.components.concept.engine.permission.PermissionRequest
[task 2020-10-26T14:00:08.677Z] 
[task 2020-10-26T14:00:08.677Z] > Task :concept-engine:dexDebug
[task 2020-10-26T14:00:08.677Z] > Task :ui-icons:bundleLibRuntimeToJarRelease
[task 2020-10-26T14:00:08.677Z] > Task :support-utils:bundleLibResRelease
[task 2020-10-26T14:00:08.777Z] 
[task 2020-10-26T14:00:08.777Z] FAILURE: Build failed with an exception.
[task 2020-10-26T14:00:08.777Z] 
[task 2020-10-26T14:00:08.777Z] * What went wrong:
[task 2020-10-26T14:00:08.777Z] Execution failed for task ':concept-engine:compileReleaseUnitTestKotlin'.
[task 2020-10-26T14:00:08.777Z] > Compilation error. See log for more details
```

PR #8561, opened 2020-09-30
Closes #8529: Dispatch a generic InitAction from BrowserStore

### LINKED ISSUE #8529: BrowserStore: Dispatch an init action that a Middleware can react to
Some `Middleware` needs to do some initialization. For example loading data from disk and then dispatching an action to add the data to the state. The `Middleware` cannot do that on its own because it first needs a reference to the store. We solved this by introducing init actions.. but then every `Middleware` needs its own. Instead the store could dispatch a generic init action that every interested `Middleware` can listen to.

csadilek:
This introduces a new `InitAction` which is dispatched when the Store is ready before other actions are being processed. In middlewares we can now "observe" this action and trigger initialization logic.

### DIFF
--- a/components/browser/state/src/main/java/mozilla/components/browser/state/state/CustomTabSessionState.kt
+++ b/components/browser/state/src/main/java/mozilla/components/browser/state/state/CustomTabSessionState.kt
@@ -4,6 +4,7 @@
 
 package mozilla.components.browser.state.state
 
+import mozilla.components.concept.engine.EngineSession
 import java.util.UUID
 
 /**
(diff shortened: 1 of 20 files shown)


components/browser/state/src/main/java/mozilla/components/browser/state/state/CustomTabSessionState.kt
```diff
@@ -52,12 +52,14 @@ fun createCustomTab(
...
+    contextId: String? = null,
+    engineState: EngineState = EngineState()
 ): CustomTabSessionState {
     return CustomTabSessionState(
         id = id,
         content = ContentState(url),
         config = config,
-        contextId = contextId
+        contextId = contextId,
+        engineState = engineState
```
pocmo, line 63: nit: The nice thing about `create[Custom]Tab()` is that you can use them as a "builder" without knowing about the internal state structure. Maybe we could add parameters for `EngineSession` (and if needed the `EngineSessionState`) and then assemble this internally (like we do for `ContentState`. Then a caller can assemble a tab without having to care about the `EngineState` wrapper. But it's also only a nit and we can do that later. :)

pocmo (approved):
✨

components/browser/state/src/main/java/mozilla/components/browser/state/state/CustomTabSessionState.kt
csadilek: Ah yeah, good point. That's inconsistent now. I will try and fix.

csadilek: OK, updated!

Issue #8535, opened 2020-09-29
Optimize multi dispatch, specifically SessionManager.restore()


pocmo:
We see `SessionManager.restore()` take quite some time. This is primarily caused by us doing a lot of sync dispatch calls. There are multiple options how we could fix that. The obvious choice is to let restore() use batch actions. Another option may be to have a generic `batch { }` that will group actions together and perform them in order - assuming that most of the overhead comes from the context switches.

mcomella:
iiuc, this work is for addressing the slowdown that occurs for having many tabs open on cold startup https://github.com/mozilla-mobile/android-components/issues/7304

Issue #7304, opened 2020-06-09
FXP-1281 ⁃ Having a larger number of tabs slows down cold startup


agi90:
While doing an unrelated startup performance investigation, I noticed that having a large number of tabs significantly slows down cold startup (tested on latest Fenix nightly).

This is the test I was running:

```
PACKAGE=org.mozilla.fenix.nightly
PAGE=https://www.google.com

echo "Warming up..."
adb shell pm clear $PACKAGE
adb shell am start -a android.intent.action.VIEW -d $PAGE
sleep 120

adb shell am force-stop $PACKAGE
adb shell am kill-all
adb logcat -c

echo "testing..."
while true; do
  adb shell am start -a android.intent.action.VIEW -d $PAGE
  sleep 15
  adb shell "input keyevent HOME"
  sleep 1
  adb shell am force-stop $PACKAGE
done
```

An empirical result is below, x axis is number of tabs, y axis is startup time to fragment rendered in milliseconds, purple points are raw measurements and green points are average [0, x] with error bars.

![test](https://user-images.githubusercontent.com/4297388/84110313-78ecbe80-a9d9-11ea-9a33-ce4b30e0d20f.png)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXP-1281)

agi90:
After looking into it more, this seems to be (partially) caused by `registerWebExtension`. I modified the test to not open a new tab every time (pressing `BACK` instead of `HOME` in the above test). And the increase in startup time is still noticeable on nightly, but not on the branch that uses `installBuiltIn`. See below graph (green = nightly, purple = `installBuiltIn`)

![what](https://user-images.githubusercontent.com/4297388/84116430-b1de6080-a9e4-11ea-9dcc-e6922bb47017.png)

agi90:
Possibly relevant for the above https://bugzilla.mozilla.org/show_bug.cgi?id=1639978

agi90:
More graphs! I ran the `installBuiltIn` test overnight to make sure the non-increase wasn't an accident and it doesn't seem to be, after 1000 restarts the startup time is still consistent.
![not-an-accident](https://user-images.githubusercontent.com/4297388/84158951-554d6680-aa21-11ea-9075-6a4be8cc93b8.png)

agi90:
After further testing the `registerWebExtension` vs `installBuiltIn` problem seems to be a red herring (had a bad local build). So the only difference here is the number of tabs.

[...]

PR #8589, opened 2020-10-01
Issue #8535, #7304: Prevent dispatching individual actions to restore tabs

csadilek:
First draft to validate performance impact. Instead of dispatching individual actions to update the state of the restored tab we make sure that the initial state is correct, based on the Snapshot.

`RestoreAction` is already a "bulk action", but tabs didn't include the full initial state and subsequent actions were required, which is fixed here.


jonalmeida (approved):
I took a few samples with this patch with 150 open tabs where `SessionManager.restore` took ~300ms.

We're now seeing much lower numbers (in milliseconds): 37.9, 30.4, 45.7, 32.8

This averages to 36.7ms.

pocmo (approved):

csadilek:
@pocmo @jonalmeida I added a commit to here to introduce a state flag to indicate that restore is complete. Not sure if the name is ideal? We only need to set this to true once and for now applications should dispatch the action. Later, we can introduce a `RestoreUseCase`.

pocmo (approved):
That'll do. 👍

Issue #8540, opened 2020-09-29
Build tree only takes implementation dependencies into account


pocmo:
See the following PR:
https://github.com/mozilla-mobile/android-components/pull/8509

It modifies the `browser-state` component. And this is the build group that was generated:
https://firefox-ci-tc.services.mozilla.com/tasks/groups/T4jWfVTTRKW7YrcCKCPHXA

I noticed that _at least_ `feature-media` is missing, which has an indirect dependency on `browser-state`:
- depends on browser-session: https://github.com/mozilla-mobile/android-components/blob/master/components/feature/media/build.gradle#L37
- which depends on browser-state: https://github.com/mozilla-mobile/android-components/blob/master/components/browser/session/build.gradle#L31

pocmo:
CC @JohanLorenzo @mozbhearsum

pocmo:
I think we should declare a direct dependency, because we use it directly in `feature-media`. Nevertheless I think it should have built? 🤔

mozbhearsum:
Quick thoughts (I'll dig more later): I _think_ this is because we use `implementation` dependencies find finding them, and `browser-session` has an `api` dependency on `browser-state`.

Two possible ideas to solve:
1) Change the dependency type (or add it to `implementation` too).
2) Look at `implementation` and `api` dependencies when calculating them in taskgraph. This would probably mean running two commands in https://github.com/mozilla-mobile/android-components/blob/master/taskcluster/ac_taskgraph/loader/build_config.py#L46 instead of one, unless we can make gradle do both in one invocation.

mozbhearsum:
I had a look around the tree and I found a few other cases where we may have non-implementation dependencies that matter here:
https://github.com/mozilla-mobile/android-components/blob/2bc8627e160196c544bb8584cf926765fce4889c/samples/browser/build.gradle#L131
https://github.com/mozilla-mobile/android-components/blob/master/components/browser/engine-gecko-beta/build.gradle#L74
https://github.com/mozilla-mobile/android-components/blob/master/components/browser/engine-gecko/build.gradle#L73
https://github.com/mozilla-mobile/android-components/blob/master/components/browser/engine-gecko-nightly/build.gradle#L74
https://github.com/mozilla-mobile/android-components/blob/2bc8627e160196c544bb8584cf926765fce4889c/components/lib/crash/build.gradle#L76
[...]

PR #8594, opened 2020-10-02
Add a few more configurations to check for dependencies

mozbhearsum:
Additional follow-up for #8540.

### DIFF
--- a/taskcluster/ac_taskgraph/loader/build_config.py
+++ b/taskcluster/ac_taskgraph/loader/build_config.py
@@ -27,7 +27,16 @@ _GIT_ZERO_HASHES = (
     "0000000000000000000000000000000000000000000000000000000000000000", # sha256
 )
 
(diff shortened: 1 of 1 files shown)


mozbhearsum:
I'm not sure why the `gecko*Implementation` configurations don't work. It's not worth debugging right now.

taskcluster/ac_taskgraph/loader/build_config.py
```diff
@@ -27,7 +27,16 @@
     "0000000000000000000000000000000000000000000000000000000000000000", # sha256
 )
 
-CONFIGURATIONS_WITH_DEPENDENCIES = ("implementation", "api")
+CONFIGURATIONS_WITH_DEPENDENCIES = (
+    "api",
+    "compileOnly",
```
JohanLorenzo, line 32: This makes sense to me after reading https://github.com/mozilla-mobile/android-components/issues/8540#issuecomment-700942877.

JohanLorenzo (approved):
:shipit:

PR #8628, opened 2020-10-06
Closes #7976: Use different port for article and reader pages

### LINKED ISSUE #7976: Use different port name for content and reader pages
Sometimes we see the disconnect (article page port) and the connect (reader page port) events arrive out of order. When the disconnect of the content page happens after the reader page connected, we remove the port and can no longer communicate with the reader page. The only consequence is that the appearance controls don't work for the reader page. This is the cause of: https://github.com/mozilla-mobile/fenix/issues/13213

Let's fix this by either using different names for the arguably different ports or include the port ID as well.

csadilek:
This refactors reader view to use two different ports/messaging channels. One for communicating with regular pages and one for active reader pages. This was a leftover from the very first version where we didn't have extension page support and had to transform the existing DOM. Using the same port name causes the problem described in https://github.com/mozilla-mobile/android-components/issues/7976#issue-673131926. There's no guarantee that the port to the original content page disconnects before the new one (with the same name) to the reader page connects.

In addition to separating the messaging, this PR includes some cleanup:
- `WebExtensionController.install` now invokes the success callback if the extension is already installed. This is in line with how the underlying API works
- The workaround to refresh the reader page opened on startup is no longer needed as was removed


jonalmeida (approved):
📖 📑

Issue #8367, opened 2020-09-09
Expose Shimmed category on tracking protection categories 


Amejia481:
As we are now using the [shims library](https://groups.google.com/g/mozilla.dev.platform/c/28eWGe4s5a80) from the [web compact extension](https://github.com/mozilla-mobile/android-components/commit/5e9e03a5f9ece1c7b7150a625234e6784f5e4385), we need to exposed this new category for apps to be aware that some resources can be shimmed. Related GV bug https://bugzilla.mozilla.org/show_bug.cgi?id=1663756

Amejia481:
Closed by https://github.com/mozilla-mobile/android-components/commit/3e68528074251168276c3b6d0deb16a18ec961c3

PR #8630, opened 2020-10-06
Closes issue #8367: Add SHIMMED TrackingCategory

Amejia481:
As we are now using the [shims library](https://groups.google.com/g/mozilla.dev.platform/c/28eWGe4s5a80) from the [web compact extension](https://github.com/mozilla-mobile/android-components/commit/5e9e03a5f9ece1c7b7150a625234e6784f5e4385), we need to exposed this new category for apps to be aware that some resources can be replaced with shimmed files instead of been blocked. You can find more info in the Fenix related issue https://github.com/mozilla-mobile/fenix/issues/14071

### Pull Request checklist

### After merge


components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
```diff
@@ -298,6 +298,11 @@ abstract class EngineSession(
              */
             SCRIPTS_AND_SUB_RESOURCES(1 shl 31),
 
+            /**
+             * Indicates that content that would have been blocked has instead been replaced with a shim.
+             */
+            SHIMMED(0),
```
csadilek, line 365: I think we should update the comment here to indicate that `SHIMMED` is only used to read and there's no need to actively configure it via `select`, or otherwise.

Since we use the same value as NONE, configuring it with select wouldn't make sense, but it's still possible using our current API i.e. you could write `TrackingProtectionPolicy.select(arrayOf(TrackingCategory.SHIMMED)` which we wouldn't want to do?

Amejia481:
> This pull request has conflicts when rebasing. Could you fix it @Amejia481? 🙏

Ready :)

csadilek (approved):
Just one nit to clarify API. 👍

components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
Amejia481: That's right, we only want to use it for categorization propose. I will add the comment thanks :)

components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
```diff
@@ -359,6 +359,13 @@ abstract class EngineSession(
              */
             SCRIPTS_AND_SUB_RESOURCES(1 shl 31),
 
+            /**
+             * Indicates that content that would have been blocked has instead been replaced with a shim.
+             * Be aware, this category is only used for categorization purpose, any uses
+             * outside of its primary motivation will not have any effect.
+             */
+            SHIMMED(NONE.id),
```
Amejia481, line 367: @csadilek the suggestion is for the updated version? :)

components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
Amejia481: Updated :)

csadilek (approved):

PR #8647, opened 2020-10-07
Closes #8548: Display icons of temporary extensions

### LINKED ISSUE #8548: Display icon of temporary extensions
Follow-up to https://github.com/mozilla-mobile/android-components/issues/8488.

We can get to the icon of a temporary extension via `WebExtension.MetaData.icon.getBitmap()`.

csadilek:
Displays icons of temporary installed extensions (installed via web-ext). Our Addons class is immutable and only carrying data so I didn't want to add a lambda to it (similar to our implementation in `Action`), especially as it has to remain Parcelable. So, I implemented this as part of `AddonManager` instead, plus a bunch of tests :).


Amejia481 (approved):
🚢 !

Issue #8527, opened 2020-09-28
AuthException: Missing key info / ISE: missing OAuthScopedKey


grigoryk:
See https://github.com/mozilla-mobile/fenix/issues/15501

https://sentry.prod.mozaws.net/operations/firefox-nightly/issues/6336621/?referrer=github_plugin

```
AuthException: Missing key info
    at mozilla.components.service.fxa.manager.FxaAccountManager.maybeUpdateSyncAuthInfoCache(FxaAccountManager.kt:23)
    at mozilla.components.service.fxa.manager.FxaAccountManager$maybeUpdateSyncAuthInfoCache$1.invokeSuspend
    at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:3)
    at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:21)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
...
(2 additional frame(s) were not displayed)

Missing key info
```

Access token we get via getAccessToken doesn't have a sync key...

grigoryk:
https://github.com/mozilla-mobile/android-components/pull/7856/files#diff-b9f70f704cf34d237544a7470a408904L15 removed this exception class entirely, and replaced this particular error with `throw IllegalStateException("missing OAuthScopedKey")`

Volume of the AuthExceptions started to disappear as that PR rolled out, however the ^^ IllegalStateException appeared: https://sentry.prod.mozaws.net/operations/firefox-nightly/issues/9659200/?query=is%3Aunresolved%20missing%20OAuthScopedKey

grigoryk:
key for the access scope [is obtained](https://github.com/mozilla/application-services/blob/aa493ae84e637a83ea36a16250517b534376d5db/components/fxa-client/src/oauth.rs#L78) via the account state: `state.scoped_keys.get(scope)`

Seems like in this case, the `scoped_keys` map doesn't have an entry for `scope`.

grigoryk:
`scoped_keys` is populated as we're finalizing authentication during `completeOAuthFlow`, in [`handle_oauth_response`](https://github.com/mozilla/application-services/blob/aa493ae84e637a83ea36a16250517b534376d5db/components/fxa-client/src/oauth.rs#L332-L343)

[...]

PR #8650, opened 2020-10-08
For #8527, Closes #8195 - Make 'account' immutable in the FxA manager

### LINKED ISSUE #8195: Consider not re-initilizing account after disconnecting
When logging-out, or cleaning up after any auth flow failures we first `disconnect` the account, and then we create an entirely new instance of it. Reading what `disconnect` does, it shouldn't be necessary to entirely start from scratch. I don't think there's any harm in doing this, so consider this purely as a cleanup.

grigoryk:
This should make it impossible for what we consider a theoretical possibility in #8527 to happen - account changing underneath us, for some reason. It's now immutable.

Code responsible for initializing the account is factored out into a separate class, which makes it easy to have the account init'd at the class level when necessary without mutation.

### Pull Request checklist

### After merge


csadilek (approved):
🔎 🕵️ 

Let's see if we spot something, but I like the refactoring of `FxaAccountManager` either way!

grigoryk:
Tested this out locally using samples-sync app, everything functions as it should. Caught one problem (account was `closed` on disconnect), which I've fixed.

PR #8655, opened 2020-10-08
Search: Add storage for bundled search engines and search middleware

pocmo:
The next pieces.

* `BundledSearchEnginesStorage` is the new replacement for `AssetsSearchEngineProvider` - but still with very similar code.
* `SearchMiddleware` reacts to the region update and loads the bundled search engines, as well as handles CRUD for custom search engines.

I still need to write some more tests, but wanted to put this up already for starting the review process.


components/feature/search/src/main/java/mozilla/components/feature/search/middleware/SearchMiddleware.kt
```diff
@@ -0,0 +1,168 @@
...
+        loadMetadata(store)
+        loadCustomSearchEngines(store)
+    }
+
+    private fun loadMetadata(
+        store: Store<BrowserState, BrowserAction>
+    ) = GlobalScope.launch(ioDispatcher) {
+        val id = metadataStorage.getDefaultSearchEngineId()
+        if (id != null) {
+            store.dispatch(SearchAction.SetDefaultSearchEngineAction(id))
```
csadilek, line 64: I have question here to make sure I understood the flow. When `SearchMiddleware` receives the `InitAction` it will read the default search engine ID here and dispatch it to the store via `SetDefaultSearchEngineAction`.

When `RegionMiddleware` receives the `InitAction` it starts determining the region and dispatches to the store via `SetRegionAction`, which is "intercepted" above here in `SearchMiddleware` to trigger loading search engines from the bundle storage.

Is that correct? It might be worth documenting this init flow in the KDocs of this Middleware.

csadilek (approved):
🔥 Looks great! Made a first pass. Some nits, and a question below.

I can go over `BundleSearchEngineStore` again tomorrow and compare with our current `AssetsSearchEngineProvider` but I didn't find anything so far!

components/browser/state/src/test/java/mozilla/components/browser/state/action/SearchActionTest.kt
```diff
@@ -19,93 +20,168 @@ import org.junit.Test
...
-        assertNotNull(searchEnginesMap[searchEngineList[1].id])
+        val searchEngines = store.state.search.customSearchEngines
+        assertFalse(searchEngines.isEmpty())
+        assertEquals(2, searchEngines.size)
+        assertEquals(engine1, searchEngines[0])
+        assertEquals(engine2, searchEngines[1])
     }
 
     @Test
     fun `SetCustomSearchEngineAction sets a new custom search engine`() {
```
csadilek, line 79: nit: This action is now `UpdateCustomSearchEngineAction`?

components/browser/state/src/main/java/mozilla/components/browser/state/reducer/SearchReducer.kt
```diff
@@ -16,61 +14,77 @@ internal object SearchReducer {
...
-        searchEngines = search.searchEngines + (searchEngine.id to searchEngine)
+        region = action.regionState
+    ))
+}
+
+private fun BrowserState.updateCustomSearchEngine(
+    action: SearchAction.UpdateCustomSearchEngineAction
+): BrowserState {
+    var updated = false
+    val searchEngines = search.customSearchEngines.map {
```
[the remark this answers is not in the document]
pocmo: Ah, yes, I like this more.

components/feature/search/src/main/java/mozilla/components/feature/search/middleware/SearchMiddleware.kt
pocmo: Yeah, to summarize:

**InitAction**
- _RegionMiddleware_ - Determines region of user and dispatches `SetRegionAction`
- _SearchMiddleware_ - Loads custom search engines and dispatches `SetCustomSearchEngines`. And loads default search engine ID and dispatches `SetDefaultSearchEngineAction`. (Both only happens if the user has picked a default, or has added custom search engines).

**SetRegionAction**
- _SearchMiddleware_ - Loads the bundled search engines for that region and dispatches `SetRegionSearchEngines`.


We could have done all search engine loading when observing `SetRegionAction` in `SearchMiddleware`, but it felt unnecessary to wait since the other values are independent from the region. Something to consider though. It may make it easier to, for example, force a load of the default search engine of the user if it isn't found in either search engine list (e.g. the user may have a new _home_ region where this search engine is not in the configuration).

components/feature/search/src/test/java/mozilla/components/feature/search/middleware/SearchMiddlewareTest.kt
```diff
@@ -0,0 +1,62 @@
...
+
+        store.dispatch(SearchAction.SetRegionAction(
+            RegionState("US", "US")
+        )).joinBlocking()
+
+        dispatcher.advanceUntilIdle()
+        // Why?
+        Thread.sleep(1000)
+
+        store.waitUntilIdle()
```
pocmo, line 58: @csadilek I need a second pair of eyes here. I do not understand why this is not working without the sleep: I first dispatch the action blocking, then I wait for the dispatcher to be idle (which executes a task that loads the engines and dispatches a new action), then wait for the store to be idle (to wait for the search engines to be added). We have a bunch of other tests doing this. But here it fails without the sleep. With some logging I see the loading of the search engines not being completed yet and `advanceUntilIdle()` apparently not waiting for it.  It feels like I am missing something obvious?!

csadilek: 👀

csadilek: @pocmo OK, I found the problem now. There's another nested dispatcher which is independent of the tests in `loadSearchEnginesFromList` (it used and independent IO dispatcher which we're not blocking on in the test). One min pushing up a fix.

csadilek: @pocmo fix to use the same dispatcher: https://github.com/csadilek/android-components/commit/4d41373082137765f8e191136f0453636e44d1f2

pocmo: Thank you! I still don't fully understand what's going on, I think. :)

I cherry-picked your patch and did fix the tests. The other tests that I wrote in the meantime still fail. Maybe there's a similar problem with the custom engine storage too. 🤔

components/feature/search/src/main/java/mozilla/components/feature/search/middleware/SearchMiddleware.kt
csadilek: 👍

csadilek (approved):

PR #8663, opened 2020-10-09
Introduce log statements in pageStart/pageStop for FNPRMS

csadilek:
In https://github.com/mozilla-mobile/fenix/pull/15747 we're making sure to strip debug logs from release builds. FNPRMS is currently depending on debug logs from GeckoView which it would then no longer see: https://github.com/mozilla-mobile/FNPRMS/blob/master/times.py#L149

This is a temporary workaround as we're in the process of replacing FNPRMS.

@mcomella this should be good enough with a quick update to times.py, but let me know if you want different logs.


mcomella:
Thanks for driving this, csadilek – it's really easy to work off of your WIP here.

I confirmed these log statements occur at the same time as the ones in GeckoView:
> 2020-10-09 15:59:49.471 1966-1966/? D/GeckoSession: handleMessage GeckoView:PageStart uri=about:blank
2020-10-09 15:59:49.471 1966-1966/? I/GeckoEngineSession: GeckoView:PageStart
2020-10-09 15:59:49.732 1966-1966/? D/GeckoSession: handleMessage GeckoView:PageStart uri=https://example.com/
2020-10-09 15:59:49.732 1966-1966/? I/GeckoEngineSession: GeckoView:PageStart
>
> 2020-10-09 15:59:49.482 1966-1966/? D/GeckoSession: handleMessage GeckoView:PageStop uri=null
2020-10-09 15:59:49.482 1966-1966/? I/GeckoEngineSession: GeckoView:PageStop
2020-10-09 15:59:49.929 1966-1966/? D/GeckoSession: handleMessage GeckoView:PageStop uri=null
2020-10-09 15:59:49.929 1966-1966/? I/GeckoEngineSession: GeckoView:PageStop

However, it'd be ideal if we can make these log statements identical to GeckoView's so that they're 1) backwards compatible (i.e. we don't need to remember to run an older version of FNPRMS on builds before date X or bake that into the code) and 2) so we don't need to make any changes to FNPRMS as making changes introduces risk and coordination.

I'm going to open a new PR that keeps the log messages identical and test it on FNPRMS. If this is undesirable, I guess let me know after the long weekend. 😅

csadilek:
@mcomella Thanks for verifying. We can't make the log statements identical because we don't want to include the URI! So, I figured since we have to change FNPRMS anyway, it wouldn't matter?

csadilek:
Updated so FNPRMS can remain backward compatible and won't need changes. Ticket to remove the logs is already filed and referenced in the comments.

mcomella (approved):
I confirmed that this works with FNPRMS without making changes to it. Caveat: the PR to remove the logs must land at the same time because FNPRMS breaks when it has the duplicate log messages.

Thanks for driving this forward!

